PR #3181, opened 2021-03-19
Adding test materialization, implement for data tests

### LINKED ISSUE #3154: Move data test "rendering" out of python compilation step
## Description
We currently wrap data tests with a CTE during the node compilation step
https://github.com/fishtown-analytics/dbt/blob/24e4b75c354e0f6bc5af7b51c4bb220428c84d5b/core/dbt/compilation.py#L362-L372
This ties our hands for desired changes to tests. We should be able to align data tests with how models are compiled using a test materialization. The goal of this issue is to maintain current implementation and reporting of data tests, but compile them similarly to models.

This will affect execution as well, since we execute the sql directly and simply return the amount of rows in the result.
https://github.com/fishtown-analytics/dbt/blob/24e4b75c354e0f6bc5af7b51c4bb220428c84d5b/core/dbt/task/test.py#L82-L83
We can use prior art of model execution to help here as well.

kwigley:
resolves #3154 

### Description

This PR pulls data test logic out of a CTE injected during compilation and puts it in a new `test` materialization. This mimics existing data test logic (simple checking of `count(*)`), subsequent changes will add more functionality.

### Checklist

### DIFF
--- a/core/dbt/task/test.py
+++ b/core/dbt/task/test.py
@@ -90,7 +90,13 @@ class TestRunner(CompileRunner):
                 .format(context)
             )
 
-        MacroGenerator(materialization_macro, context)()
+        # generate materialization macro
+        # simple `select(*)` of the compiled test node
+        macro_func = MacroGenerator(materialization_macro, context)
+        # execute materialization macro
+        macro_func()
+        # load results from context
+        # could eventually be returned directly by materialization
         result = context['load_result']('main')
         table = result['table']
         num_rows = len(table.rows)
--- a/core/dbt/task/test.py
+++ b/core/dbt/task/test.py
@@ -16,7 +16,13 @@ from dbt.contracts.graph.parsed import (
     ParsedSchemaTestNode,
 )
 from dbt.contracts.results import RunResult, TestStatus
-from dbt.exceptions import raise_compiler_error, InternalException
+from dbt.context.providers import generate_runtime_model
+from dbt.clients.jinja import MacroGenerator
+from dbt.exceptions import (
+    raise_compiler_error,
+    InternalException,
+    missing_materialization
+)
 from dbt.graph import (
     ResourceTypeSelector,
     SelectionSpec,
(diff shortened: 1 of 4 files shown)


core/dbt/include/global_project/macros/materializations/test/data.sql
```diff
@@ -0,0 +1,10 @@
+{%- materialization test, default -%}
```
jtcohen6, line 1: I notice that the file name is `materializations/test/data.sql`, and I think this is rightly just `materializations/test.sql`. We'll want to use one & the same materialization for _both_ ~data~ one-off and ~schema~ generic/reusable tests.

jtcohen6 (commented):
I know you're still finalizing the last few pieces, but I couldn't help myself, so I took this for a spin. Everything worked the way I'd hoped! I was able to override the `test` materialization locally, with both default and adapter-specific implementations, and dbt preferred my own each time.

core/dbt/include/global_project/macros/materializations/test/data.sql
kwigley: No, please help yourself! Thanks for taking it for a spin. I'll update the name accordingly :)

core/dbt/task/test.py
```diff
@@ -78,9 +67,47 @@ def execute_schema_test(self, test: CompiledSchemaTestNode):
...
+        if num_rows != 1:
+            num_cols = len(table.columns)
+            # since we just wrapped our query in `select count(*)`, we are in
+            # big trouble!
+            raise InternalException(
+                f"dbt internally failed to execute {test.unique_id}: "
+                f"Returned {num_rows} rows and {num_cols} cols, but expected "
+                f"1 row and 1 column"
+            )
+        return int(table[0][0])
```
jtcohen6, line 106: Rather than just returning a single number to the `execute` method, I imagine the materialization-calling method should construct and return the full `RunResult`, similar to [`_build_run_model_result`](https://github.com/fishtown-analytics/dbt/blob/440a5e49e274b7c98a726c1acdd202fbf1063df1/core/dbt/task/run.py#L192). It seems like that will enable us to manage the dynamic `message` interpolation, as well as recording the `adapter_response`.

Of course, those can and should be new issues! It's cool to see the path taking shape.

jtcohen6 (approved):
All good in my book!

nathaniel-may (approved):
LGTM

gshank (approved):

PR #3204, opened 2021-03-25
dev env clean up and improvements

### LINKED ISSUE #3194: tox updates

kwigley:
resolves #3194 

### Description

- update development dependencies (default to using the latest and greatest)
- update code and installed a new pytest plugin named `pytest-dotenv` to load `test.env` secrets in testing environment in order to avoid running integration tests in a docker container locally.
- remove `scripts/dtr.py` because it is an abstraction we shouldn't need anymore, default to using pytest to run individual tests, tox to run full test suites, and Makefile for commonly used commands
- update `tox.ini` with better defaults and test envs, we are really close to only needing python and tox installed on your machine to be able to do everything related to development with dbt :) 
- update `Makefile` to run target on local machine by default, not in a docker container. Still provided option to run in a docker container using `USE_DOCKER=true` in case folks don't have a certain python version installed.
- update `CONTRIBUTING.md` to reflect changes
- CI runs a lot fast

### Future ideas
- use `tox` to create development virtualenvs with everything necessary to develop (example `tox --devenv env -e py38` is the same thing as `python3.8 -m venv env && source env/bin/activate && pip install -r dev-requirements.txt -r editable-requirements.txt`)
- create a VSCode dev container for folks who use VSCode
- remove initial step to setup postgres database for integration tests
  - explore using something like https://github.com/avast/pytest-docker

### Checklist


test/rpc/conftest.py
```diff
@@ -57,8 +58,13 @@ def profiles_root(tmpdir):
 def project_root(tmpdir):
     return tmpdir.mkdir('project')
 
-
 def postgres_profile_data(unique_schema):
+    database = 'database'
+    if sys.platform == 'win32':
+        datbase = 'localhost'
```
gshank, line 64: I think this is supposed to be 'database'

tox.ini
```diff
@@ -1,388 +1,58 @@
...
-
-[testenv:integration-redshift-py38]
+[testenv:flake8]
 basepython = python3.8
-passenv = *
-setenv =
-    HOME=/home/tox
-commands = /bin/bash -c '{envpython} -m pytest --durations 0 -v -m profile_redshift {posargs} -n4 test/integration/*'
+skip_install = true
+commands = flake8 --select=E,W,F --ignore=W504,E741 \
```
gshank, line 15: Can we change flake8 to allow longer lines here?

CONTRIBUTING.md
```diff
@@ -128,100 +127,103 @@ Docker and docker-compose are both used in testing. Specific instructions for yo
...
-dbt's unit tests and Python linter can be run with:
+There are a few methods for running tests locally.
 
-```
-make test-unit
-```
+#### Makefile
 
-To run the Postgres + Python 3.6 integration tests, you'll have to do one extra step of setting up the test database:
+There mulitple targets in the Makefile to run common test suites and code
```
gshank, line 188: "There are multiple"

gshank (changes requested):
This is great! A lot of good cleanup. Found a couple of typos and was wondering if we could take this opportunity to change line length checks?

test/rpc/conftest.py
kwigley: ah ya, but who develops on windows anyway 🥴

gshank: True :-)

tox.ini
kwigley: ya, just set it to 99 !

nathaniel-may (commented):
I'm actually unfamiliar with the vast majority of changes you're proposing here so I'm not hitting approve, but I looked through it all and didn't see any red flags.

> we are really close to only needing python and tox installed on your machine to be able to do everything related to development with dbt
I am so excited for this future. 

Very excited about these changes. Thanks for all the hard work!

core/setup.py
```diff
@@ -64,7 +64,7 @@ def read(fname):
         'sqlparse>=0.2.3,<0.4',
         'networkx>=2.3,<3',
         'minimal-snowplow-tracker==0.0.2',
-        'colorama>=0.3.9,<0.4.4',
+        'colorama>=0.3.9,<0.4.5',
```
kwigley, line 67: We had to pin this before because the package didn't provide the source distribution on PyPI, but they have since added it! 👍 https://pypi.org/project/colorama/0.4.4/#files

test/integration/015_cli_invocation_tests/test_cli_invocation.py
```diff
@@ -149,15 +149,15 @@ def test_postgres_dbt_commands_with_cwd_as_project_dir(self):
 
     @use_profile('postgres')
     def test_postgres_dbt_commands_with_randomdir_as_project_dir(self):
-        workdir = os.getcwd()
+        workdir = self.test_root_dir
```
kwigley, line 152: Without this change, integration tests failed on MacOS because of how the os handles temp dir/file. Functionally, `os.getcwd()` and `self.test_root_dir` are the same value. `os.getcwd()` returns a fully resolved filepath while `self.test_root_dir` is a string value pointing to a symlinked path that is equivalent to `os.getcwd()`.

gshank (approved):

jtcohen6 (approved):
This is amazing! 🎤  &rarr; ⬇️ 

I was able to run locally with `pytest`, and it was a breeze.

CONTRIBUTING.md
```diff
@@ -128,100 +127,105 @@ Docker and docker-compose are both used in testing. Specific instructions for yo
 
 For testing, and later in the examples in this document, you may want to have `psql` available so you can poke around in the database and see what happened. We recommend that you use [homebrew](https://brew.sh/) for that on macOS, and your package manager on Linux. You can install any version of the postgres client that you'd like. On macOS, with homebrew setup, you can run:
```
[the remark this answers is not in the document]
kwigley: Docker is still required to create and run a testing postgres database, the scripts to setup the database for testing (`test/setup_db.sh`) is specific to the database created by docker-compose. So it it still a requirement to setup a postgres database, but no longer needed to invoke tests locally. I'll try to make this more clear!

Issue #3217, opened 2021-03-31
Move the save point for partial parsing


gshank:
Ticket #3163 switched from using ParseResults to save the first stage of parsing artifacts to using the Manifest. The pickle file that saves the state of parsing for reloading with the partial parsing flag is now a Manifest object.

This ticket will take that beginning work and move the save point to the end of parsing, after the "patching" and "process_manifest" steps. This will result in much faster execution with the partial parsing flag.

PR #3219, opened 2021-04-01
Use Manifest instead of ParseResult [#3163]

### LINKED ISSUE #3163: Refactor partial parsing to cover more startup costs
The current implementation of partial parsing saves a pickle file during parsing, before the 'create_manifest' step in which the following occurs 1) patch_sources (where source tests are processed), 2) patch_nodes, 3) patch macros, 4) process_sources, 5) process_refs, and 6) process_docs. If we refactor parsing so that the parsing state is saved after these things happen, then startup costs when using partial parsing will be substantially reduced. In our large parsing test project, those steps take 70 seconds out of a total of 193 in a sample run. We wouldn't be able to eliminate all of that time, but eliminating most of it would result in almost a 1/3 reduction in time.

Because the current implementation does not have to handle re-processing sources, refs, and docs, we would need to create new code to handle doing those tasks for only the changed files.

At some point we may also look at whether a pickle file is the best alternative for saving and restoring state, but that is not currently part of this project.

gshank:
resolves #3163

### Description

As part of the performance initiative we are working on saving more of the parsing state to be reused when partial parsing is turned on. This ticket switches to use a Manifest object instead of a ParseResult object to store parsing state. The eventual goal is to be able to reload a previously fully parsed Manifest.

### Checklist


core/dbt/parser/manifest.py
```diff
@@ -367,82 +318,132 @@ def read_parse_results(self) -> Optional[ParseResult]:
...
-        process_refs(manifest, project_name)
-        process_docs(manifest, self.root_project)
-
-    def create_manifest(self) -> Manifest:
-        # before we do anything else, patch the sources. This mutates
-        # results.disabled, so it needs to come before the final 'disabled'
-        # list is created
+        process_sources(self.manifest, project_name)
+        process_refs(self.manifest, project_name)
+        process_docs(self.manifest, self.root_project)
```
jtcohen6, line 341: If I understand it right, our next goal (in #3217) is to move these methods to _before_ the partial-parsing save point. Today, dbt fully re-processes every single source, ref, and description, every single time. Ideally, partial-parsing should only be re-processing nodes that have changed—whether the change occurred in a `.sql` file, `.yml` file, or `.md` file associated with that node.

Do I have that right?

jtcohen6 (commented):
So cool! I left a few initial notes as I was trying to read through + understand the changes. Thank you for all the inline comments along the way :)

core/dbt/parser/manifest.py
gshank: Yes, that's correct.

core/dbt/parser/manifest.py
```diff
@@ -367,82 +318,132 @@ def read_parse_results(self) -> Optional[ParseResult]:
...
             time.perf_counter() - start_process
         )
 
-        return manifest
+        return self.manifest
+
+    # TODO: this should be calculated per-file based on the vars() calls made in
+    # parsing, so changing one var doesn't invalidate everything. also there should
+    # be something like that for env_var - currently changing env_vars in way that
+    # impact graph selection or configs will result in weird test failures.
```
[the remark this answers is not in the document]
gshank: The comment was already, I just moved it :) But it's a good comment.

core/dbt/contracts/graph/manifest.py
```diff
@@ -908,6 +957,212 @@ def merge_from_artifact(
...
+            )
+
+    # This is called by ManifestLoader._get_cached/parse_with_cache,
+    # which handles updating the ManifestLoader results with information
+    # from the "old_manifest", i.e. the pickle file if the checksums are
+    # the same.
+    def sanitized_update(
+        self,
+        source_file: SourceFile,
+        old_manifest: Any,
```
kwigley, line 1099: Just a note, you can do something like
```suggestion
        old_manifest: Optional["Manifest"],
```
here to refer to itself as a type.

kwigley (approved):
Looks good!

kwigley:
I'm not 100% on what is causing the Windows tests to fail, can we re-run it to see if it is just a transient error?

jtcohen6:
> I'm not 100% on what is causing the Windows tests to fail, can we re-run it to see if it is just a transient error?
> 
>> can we re-run it
> 
> I'm actually not 100% how to do this haha

I've re-run it twice, and it's failing reliably (though inscrutably) on [this line](https://dev.azure.com/fishtown-analytics/dbt/_build/results?buildId=3155&view=logs&j=237693ef-88fc-5a71-c0ba-028c2d175baa&t=2e54d75c-0a78-5d9c-6a48-5b20a2da8e82&l=145):
```
test/unit/test_config.py::TestProject::test_cycle Windows fatal exception: stack overflow
```

core/dbt/contracts/graph/manifest.py
gshank: Thanks! I've seen code using the string class name but didn't think of it for here.

core/dbt/parser/manifest.py
```diff
@@ -367,82 +318,132 @@ def read_parse_results(self) -> Optional[ParseResult]:
...
+    # finally, we should hash the actual profile used, not just root project +
+    # profiles.yml + relevant args. While sufficient, it is definitely overkill.
```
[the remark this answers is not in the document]
jtcohen6: This is a case where we're being overzealous, and fully re-parsing the project when we could make partial parsing smarter. Therefore, this is in the "slow + correct" bucket, rather than the "fast + incorrect" bucket. We shouldn't lose track of this, but for now, we should be prioritizing edge cases in the latter category.

jtcohen6 (approved):
All good in local testing!

PR #3221, opened 2021-04-01
Add Redshift parameter to create tables with backup option specified

### LINKED ISSUE #2367: Redshift create table with backup option
### Describe the feature
The `CREATE TABLE` statement in Redshift has a configuration option that controls whether or not the table should be included in automated and manual cluster snapshots.  The default setting for this option is `YES`.  At present this is not configurable in DBT and the default is used for all table and incremental materialisations.

By allowing configuration of this parameter in DBT models we can save processing time when creating snapshots and restoring from snapshots and to reduce storage space on Amazon Simple Storage Service.

There is no way to get around this except to turn off automated snapshots at a cluster level.

> To disable automated snapshots, set the retention period to zero. If you disable automated snapshots, Amazon Redshift stops taking snapshots and deletes any existing automated snapshots for the cluster.

### Additional context
This feature is Redshift specific.  The feature is similar but not exactly the same as Snowflakes `TRANSIENT` table.

### Who will this benefit?
This will benefit all DBT users who use Redshift and who want to control storage costs and to speed up snapshot creation and restores.

P.S. If you think this is worthwhile and in keeping with DBT's principles, then I would love to work on this issue with a PR.

dlb8685:
resolves #2367 

### Description
This PR adds a `backup` parameter to `RedshiftConfig` defaulting to true, which controls whether tables are included in any snapshots. It uses the value of this parameter in the Redshift table adapter to insert a "BACKUP NO" clause into the table creation statement of tables or models where `backup: false` is specified.

The new tests confirm that that DDL will only contain a "BACKUP NO" clause if this parameter is explicitly set to false. It also confirms that views are not affected by the change.

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
 - Fix for EOL SQL comments prevent entire line execution ([#2731](https://github.com/fishtown-analytics/dbt/issues/2731), [#2974](https://github.com/fishtown-analytics/dbt/pull/2974))
 - Add optional `merge_update_columns` config to specify columns to update for `merge` statements in BigQuery and Snowflake ([#1862](https://github.com/fishtown-analytics/dbt/issues/1862), [#3100](https://github.com/fishtown-analytics/dbt/pull/3100))
 - Use query comment JSON as job labels for BigQuery adapter when `query-comment.job-label` is set to `true` ([#2483](https://github.com/fishtown-analytics/dbt/issues/2483)), ([#3145](https://github.com/fishtown-analytics/dbt/pull/3145))
+- Add optional Redshift parameter to create tables with BACKUP NO set, to exclude them from snapshots. ([#2367](https://github.com/fishtown-analytics/dbt/issues/2367))
 
 ### Under the hood
 - Add dependabot configuration for alerting maintainers about keeping dependencies up to date and secure. ([#3061](https://github.com/fishtown-analytics/dbt/issues/3061), [#3062](https://github.com/fishtown-analytics/dbt/pull/3062))
@@ -33,6 +34,7 @@ Contributors:
 - [@prratek](https://github.com/prratek) ([#3100](https://github.com/fishtown-analytics/dbt/pull/3100))
 - [@techytushar](https://github.com/techytushar) ([#3158](https://github.com/fishtown-analytics/dbt/pull/3158))
 - [@cgopalan](https://github.com/cgopalan) ([#3165](https://github.com/fishtown-analytics/dbt/pull/3165))
+- [@dlb8685](https://github.com/dlb8685/)
 
 ## dbt 0.19.1 (Release TBD)
 
--- a/test/integration/034_redshift_test/test_backup_table_option.py
+++ b/test/integration/034_redshift_test/test_backup_table_option.py
@@ -35,15 +35,16 @@ class TestBackupTableOption(DBTIntegrationTest):
             ddl_statement = ddl_file.readlines()
             self.assertEqual('backup no' not in ' '.join(ddl_statement).lower(), backup_is_expected)
         # Use information schema to confirm backup is set correctly on new table (0 if BACKUP NO is present)
-        check = "select distinct backup from stv_tbl_perm where name = '{}'".format(test_table_name)
-        info_schema_result = self.run_sql(check, fetch='all')
-        self.assertEqual(info_schema_result[0][0], int(backup_is_expected))
-        self.assertEqual(len(info_schema_result), 1)
+        if 'view' not in test_table_name:
+            check = "select distinct backup from stv_tbl_perm where name = '{}'".format(test_table_name)
+            info_schema_result = self.run_sql(check, fetch='all')
+            self.assertEqual(info_schema_result[0][0], int(backup_is_expected))
+            self.assertEqual(len(info_schema_result), 1)
 
     @use_profile('redshift')
     def test__redshift_backup_table_option(self):
         self.assertEqual(len(self.run_dbt(["seed"])), 1)
-        self.assertEqual(len(self.run_dbt()), 3)
+        self.assertEqual(len(self.run_dbt()), 4)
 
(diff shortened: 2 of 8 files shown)


dlb8685:
@jtcohen6 Thanks for the guidance in the [issue](https://github.com/fishtown-analytics/dbt/issues/2367) discussion. I figured I was close enough to start a PR and let you (or the most appropriate person) comment here. 100% let me know if there's more you have in mind on the integration testing.

Everything seemed straightforward to me besides "and once via `dbt_project.yml`". Do you mean to update that within the test and re-run dbt? Is there a good example of doing that somewhere else that I could look at?

Lastly, there is a system table `stv_tbl_perm` that can be queried to see how `backup` is set. It looks like whatever user is running the CircleCI tests doesn't have access to that table, and that is why the Redshift check failed. Let me know if someone can update the permissions accordingly, or if you want to have that part removed.

test/integration/034_redshift_test/test_backup_table_option.py
```diff
@@ -0,0 +1,59 @@
...
+                'quote_columns': False,
+            }
+        }
+
+    def project_config_append(self, new_name, new_element):
+        self._project_config[new_name] = new_element
+
+    def check_backup_param_template(self, test_table_name, backup_is_expected):
+        # Use raw DDL statement to confirm backup is set correctly on new table
+        with open('target/run/test/models_backup/{}.sql'.format(test_table_name), 'r') as ddl_file:
```
jtcohen6, line 34: I know this is a pattern present in other integration tests (I see it in 020, 028, 029). I still hesitated about this, because the run file path (`target/run/.../*.sql`) is not _exactly_ contracted. After thinking about it more, this would still be a highly visible user-facing change, and so if we were to ever change it, this test failure would be a nice way to call that out. Just leaving this comment for our future reference.

test/integration/034_redshift_test/test_backup_table_option.py
```diff
@@ -0,0 +1,59 @@
...
+            ddl_statement = ddl_file.readlines()
+            self.assertEqual('backup no' not in ' '.join(ddl_statement).lower(), backup_is_expected)
+        # Use information schema to confirm backup is set correctly on new table (0 if BACKUP NO is present)
+        if 'view' not in test_table_name:
+            check = "select distinct backup from stv_tbl_perm where name = '{}'".format(test_table_name)
```
jtcohen6, line 39: It looks like [only superusers](https://docs.aws.amazon.com/redshift/latest/dg/r_STV_TBL_PERM.html) can read from `stv_tbl_perm`, which is too bad. I'd rather not give the CircleCI/Azure user access, so let's remove this piece of the integration test (even though it offers compelling confirmation).

test/integration/034_redshift_test/test_backup_table_option.py
```diff
@@ -0,0 +1,59 @@
...
+
+    @property
+    def project_config(self):
+        return {
+            'config-version': 2,
+            'data-paths': [self.dir('seed')],
+            'seeds': {
+                'quote_columns': False,
+            }
+        }
```
jtcohen6, line 27: > Everything seemed straightforward to me besides "and once via `dbt_project.yml`". Do you mean to update that within the test and re-run dbt? Is there a good example of doing that somewhere else that I could look at?

Yes, all I meant here was creating another test case, with only one difference:
```python
    @property
    def project_config(self):
        return {
            'config-version': 2,
            'data-paths': [self.dir('seed')],
            'seeds': {
                'quote_columns': False,
            },
            'models': {
                'backup': False
            }
        }
```
Since this should affect `model_backup_undefined` by causing it to include `BACKUP NO`. I see you've defined the `project_config_append` method, was that to accomplish this?

CHANGELOG.md
```diff
@@ -15,6 +15,7 @@
 - Fix for EOL SQL comments prevent entire line execution ([#2731](https://github.com/fishtown-analytics/dbt/issues/2731), [#2974](https://github.com/fishtown-analytics/dbt/pull/2974))
 - Add optional `merge_update_columns` config to specify columns to update for `merge` statements in BigQuery and Snowflake ([#1862](https://github.com/fishtown-analytics/dbt/issues/1862), [#3100](https://github.com/fishtown-analytics/dbt/pull/3100))
 - Use query comment JSON as job labels for BigQuery adapter when `query-comment.job-label` is set to `true` ([#2483](https://github.com/fishtown-analytics/dbt/issues/2483)), ([#3145](https://github.com/fishtown-analytics/dbt/pull/3145))
+- Add optional Redshift parameter to create tables with BACKUP NO set, to exclude them from snapshots. ([#2367](https://github.com/fishtown-analytics/dbt/issues/2367))
```
jtcohen6, line 18: Nit: we like to include both the issue # and PR # in the changelog entry
```suggestion
- Add optional Redshift parameter to create tables with BACKUP NO set, to exclude them from snapshots. ([#2367](https://github.com/fishtown-analytics/dbt/issues/2367), [#3221](https://github.com/fishtown-analytics/dbt/pull/3221))
```

jtcohen6 (commented):
@dlb8685 This is looking really good! So sorry for the delay on my end. Just a few small comments, then I think this will be good to ship.

jtcohen6:
@dlb8685 Is this still something you'd be interested in contributing? I'd still welcome the addition to `dbt-redshift`. If not, I'm going to close the PR for the time being

dlb8685:
@jtcohen6 Thank you for following up. To be honest this slipped my mind. I do want to bring this home though. It will probably be the middle of next week before I have time to look into your feedback though -- I'm taking a long weekend off for Labor Day. I will keep you posted.

jtcohen6:
@dlb8685 That's fine by me!

jtcohen6:
@dlb8685 You got it :)

https://github.com/dbt-labs/dbt-redshift/issues/18#issuecomment-941178025

closing in favor of a PR over there!

PR #3235, opened 2021-04-07
Be less greedy in test selection expansion

### LINKED ISSUE #2891: Run tests IFF all first-order parents are selected
### Describe the feature
See prior discussion: #1827, #2132, misc `state:modified` conversations

Run a test only if _all_ first-order parents are included in the selection criteria. This would be most immediately relevant for:
- `relationships` tests
- data tests with multiple parents

This would also handle cases where a schema test has a property (`tag:`, `state:`) that its one parent model does not have. Since its single parent would not be included in the selection criteria, given the proposed condition, the test should not run.

- How useful vs. confusing is this idea?
- Do we change default behavior? Or add a flag (`--all-parents`) to enable this?

### Who will this benefit?
Users who want finer-grained control over which tests they're running, especially:
- using complex selection syntax
- running in CI environments

jtcohen6:
resolves #2891

### Problem

Let's say I have two models (`model_a` + `model_b`) and a source table (`my_src.my_tbl`). There's a unique test defined on a column in `model_a`, plus two relationships tests—one to `model_b`, one to `my_src.my_tbl`—for a total of three tests.

<details><summary>YAML</summary>

```yml
version: 2

sources:
  - name: my_src
    tables:
      - name: my_tbl

models:
  - name: model_a
    columns:
      - name: id
        tests:
          - unique
          - relationships:
              to: ref('model_b')
              field: id
          - relationships:
              to: source('my_src', 'my_tbl')
              field: id
```

</details>

Today, the following command will select _all three_ tests to run:
```bash
$ dbt test -m model_a
```
That's because dbt test selection is always **greedy** today, and it performs what I call a "magic jump" (the actual function is named `expand_selection`). Put simply, dbt says: I see that you've selected model_a, I see that the task is `test`, so really I'm going to select all tests that are first-order children of model_a. (Doesn't that sound more like `--select model_a+1 --resource-type test`? Too bad!)

The upshot is, that's really nice syntax if what you're after is `dbt run -m model_a && dbt test -m model_a` (or someday soon: `dbt build -s model_a`.) But in the scenario outlined above, since the relationships tests are first-order children of model_a, they get run too—even if model_b is missing! And this puts people in all kinds of awkward situations.

In v0.19.0, we offered a resolution to this conundrum in the CI context with `dbt test --defer`: If model_b is missing in my sandbox schema, but present in the comparison/production schema, use that one instead! But we could still do a better job here of making test selection intuitive, and making it work outside of CI/deferral contexts, too.

### Proposal

Test selection should be less greedy when it comes to multi-parent tests. In particular:
1. Include a multi-parent test if and only if **all** its parents are selected (not current behavior)
2. Exclude a multi-parent test if **any** parent is excluded (current behavior)

To accomplish this, the PR adds a more formal notion of "greedy" expanded selection:
- A `greedy` property to the `SelectionCriteria` class, which is set to `False` for inclusion and `True` for exclusion. That way, `TestSelector.expand_selection` can _always_ be greedy for exclusion ("any parent"), but inclusion can be more cautious ("all parents").
- A test that would be included greedily/indirectly (1+ parent is selected) is instead kept at arm's length, either until all its parents make an appearance (included), or until all selection criteria have been sorted through and a parent is still missing (not included).

In order to make selection expansion more consistent, this PR also moves expansion out of the test task and into the default selector logic. This way, `dbt ls -s [criteria] --resource-type test` should always return the tests that would be executed by `dbt test -m [criteria]`.

Per the changes in this PR, here's what happens in various scenarios:
- `dbt test`: Run all tests.
- `dbt test -m model_a`: Run _just_ the unique test. Don't be greedy.
- `dbt test -m model_a model_b`: Run the unique test + the relationships test between model_a + model_b. We "expand" to the relationships test because both parents are selected.
- `dbt test -m model_a source:*`: Run the unique test + the relationships test between model_a + my_src.my_tbl
- `dbt test --exclude model_b`: Skips any tests that are first-order children of `model_b`. Exclusion _should_ be greedy!
- `dbt test -m model_a+`: Run all tests. All three tests are selected directly, as children of model_a. There's no magic jump / expansion needed. (In a CI context, it's common to `dbt test -m state:modified+`, so deferral-for-tests still has a role to play.)

Brain teasers [updated!]:

- `dbt test -m model_a,test_name:unique`: Just the unique test on model_a
- `dbt test -m tag:special,test_name:relationships`:
    - If I've tagged the relationships test, or the column it's defined on, then this will run. In either case, the test itself is tagged `'special'` (we can check this in the manifest).
    - If model_a + model_b are _both_ tagged `'special'`, their relationships test will run.
    
</details>

### Questions

- Does any of the scenarios above feel weird, unexpected, or unintuitive?
- Should expansion treat **source** parents differently from model/seed/snapshot parents? Generally speaking, dbt assumes that sources are always present in the database, so we could argue that it's fair to greedily expand selection to tests with one selected parent and another unselected source parent. Concretely: In the scenario above, should `dbt test -m model_a` also run the relationships test between model_a + my_src.my_tbl?

### Checklist


gshank (approved):
Looks good.

kwigley (approved):
🎉

kwigley:
Do you mind rebasing before merging? I'm hoping that will make all checks green :)

jtcohen6:
Thanks @gshank @kwigley!

Tagging @drewbanin because I know he has a few thoughts :) and I want to give you a chance to air 'em if you'd like!

In effect, with this PR, test selection boils down to:

(A test node has been selected OR all of a test node's direct parents have been selected) AND NOT (a test node has been excluded OR any of a test node's direct parents has been excluded)

core/dbt/graph/selector.py
```diff
@@ -162,12 +187,53 @@ def filter_selection(self, selected: Set[UniqueId]) -> Set[UniqueId]:
...
+        #    for later and see if its other parents show up.
+        # We use this for INCLUSION.
+
+        direct_nodes = selected
+        indirect_nodes = set()
+
+        for unique_id in self.graph.select_successors(selected):
+            if unique_id in self.manifest.nodes:
+                node = self.manifest.nodes[unique_id]
+                if node.resource_type == NodeType.Test:
```
drewbanin, line 211: if there's a single thing that i don't feel great about in this PR, it's this line! I think that the majority of node selection is resource type-agnostic, and the incidence of Test in here is setting off all of my "weird abstraction" alarm bells. I don't know if the thing to do is:
1. Make tests less weird / not a node type, such that this logic wouldn't need to exist (but what?)
2. Make selection logic resource type-specific, such that selectors do fundamentally different things for different node types (this feels bad)

So, not claiming this as a blocker or challenge to overcome - I think this change is behavior is a real net-positive, but it would be good for us to come up with an abstraction over / reimagination of tests that doesn't _require_ this kind of logic to exist!

core/dbt/graph/selector.py
```diff
@@ -162,12 +187,53 @@ def filter_selection(self, selected: Set[UniqueId]) -> Set[UniqueId]:
...
+        #
+        # GREEDY mode: If ANY parent is selected, select the test. We use this for EXCLUSION.
+        #
+        # NOT GREEDY mode:
+        #  - If ALL parents are selected, select the test.
+        #  - If ANY parent is missing, return it separately. We'll keep it around
+        #    for later and see if its other parents show up.
+        # We use this for INCLUSION.
+
+        direct_nodes = selected
```
drewbanin, line 205: I think that using `selected` directly here might not be a _great_ idea because `direct_nodes` is just a pointer to the `selected` function argument. On line 214 below, we're calling the `direct_nodes.add` function which is going to mutate the `selected` variable and could possibly cause problems in any function calls up the call stack.

This would _not_ cause any problems today since the caller of this method looks like:
```
        direct_nodes, indirect_nodes = self.expand_selection(
            selected=(collected | neighbors),
            greedy=spec.greedy
        )
```

so, the `selected` argument is an expression and is not assigned to a variable. Imagine if this logic changed in the future to look more like:

```
        to_select = (collected | neighbors)
        direct_nodes, indirect_nodes = self.expand_selection(
            selected=to_select,
            greedy=spec.greedy
        )
        some_function(to_select)
```

If this happened, then the value of `to_select` would be changed after the call to `expand_selection`! That kind of behavior would probably not be expected, and it would probably be a tough bug to track down!

Good news is that it's a quick change, I think you can make a copy of the set by changing this line to:
```
direct_nodes = set(selected)
```

That way, this method won't have any "side effects"

core/dbt/graph/selector.py
```diff
@@ -162,12 +187,53 @@ def filter_selection(self, selected: Set[UniqueId]) -> Set[UniqueId]:
...
+
+        return direct_nodes, indirect_nodes
+
+    def incorporate_indirect_nodes(
+        self, selected: Set[UniqueId], greedy_nodes: Set[UniqueId] = set()
+    ) -> Set[UniqueId]:
+        # Check tests previously selected indirectly to see if ALL their
+        # parents are now present.
+
+        for unique_id in greedy_nodes:
```
drewbanin, line 227: similar feedback to the side-effect comment above - can we make a copy of `selected` before mutating it?

drewbanin: Also, there's some complexity here! We are both adding elements to the `selected` set while also checking if `node.depends_on.nodes` is a subset of `selected`. Is that going to have any funny/unexpected impacts?

I need to dig in here to understand if this would be an issue, but I'm just imagining:

```
selected: {A}

Node A:
 - Depends on: []

Node B:
 - Depends on: [A]

Node C:
 - Depends on [A, B]
```

If `greedy_nodes == [B, C]`, then:
 - loop iter 0: B gets added to `selected`
 - loop iter 1: C gets added to `selected`
 - Return value is `{A, B, C}`

If the order is reversed and `greedy_nodes == [C, B]`, then:
 - loop iter 0: C does not get added to `selected`
 - loop iter 1: B does not get added to `selected`
 - Return value is `{A}`

I actually... don't know if that's an issue based on how we use this method, would need to dig in a lot further to understand, but wanted to mention it in case it sparks any thoughts!

drewbanin (commented):
I dropped a couple of comments in here, but nothing major. These tests look really thorough - nice work! Let me know if you have any thoughts/questions or if I am way off-base on anything

core/dbt/graph/selector.py
jtcohen6: Yeah, this is a really fair point. The goal of this PR was to make selection _more_ consistent overall, while at the same time acknowledging that (a) there exists this special thing, which I'm calling "indirect selection," and (b) only test nodes qualify for it.

I formalized that by adding a super-basic function to the `selector` module, `can_select_indirectly`, which returns `True` for tests and `False` otherwise. For now, it's just tests, but if we ever added another resource type that's sort-of-a-node, sort-of-an-attribute-of-another-node, we could include it in this function. And if we decided to introduce some kind of runtime variable / flag that turned on indirect selection for everything, that'd be the place to do it.

core/dbt/graph/selector.py
jtcohen6: You're 100% right! Including `direct_nodes = selected` was my attempt at not mutating the original argument value, but that's obviously not what it does. Wrapping with `set()` makes a ton of sense.

core/dbt/graph/selector.py
jtcohen6: Definitely not my intention here for order to matter here. I rewrote so that it instantiates and mutates a different variable from the value passed in as the argument. A test cannot be parent to another test, so for the specific case tackled in this PR, I think we're in the clear—but this was a great thing to call out and change, in the spirit of making indirect selection as much of its own thing as possible, not uniquely suited to tests.

ags2121:
do we know what DBT version this commit will be included in?

jtcohen6:
@ags2121 This will be in v0.21. You can try it out with v0.21.0-rc1, which we released yesterday :)

ags2121:
thanks! I'm seeing this PR tagged in the 0.20.0 release here under the test section:
https://github.com/dbt-labs/dbt/releases/tag/v0.20.0

was this PR not included in 0.20.0 then?

jtcohen6:
@ags2121 So sorry! I was getting this PR confused with #3833, which is in v0.21.0-rc1. This PR indeed shipped in v0.20 :)

ags2121:
ah great! thanks 👍

Issue #1269, opened 2019-01-30
Add namespacing for dbt resources


drewbanin:
Placeholder issue for thoughts related to namespacing. When we're able to prioritize work on namespacing, let's convert this to a more actionable issue:

Questions:
 - how can we namespace models to avoid ambiguity?
 - what do package namespaces look like for models?
 - is there a way to "import" resources from a package, rather than injecting everything into the toplevel context?

Other similar questions / use cases welcomed :)

danimcgoo:
Hey @drewbanin, tagging looks like it'll do the trick. thanks fo the tip!

heisencoder:
Hi @drewbanin 

Is this "Help Wanted" still current?  I'm interested in supporting this feature in our internal projects because we have 100s of sql files that have a hierarchy of subdirectories under the "models" directory to help keep things semantically organized.  Also, we have several SQL files with the same name but that exist in different directories (as illustrated above with the sales.sql example).

Offhand, I think that this feature needs to be enabled with a flag somewhere to avoid getting an error during the compile phase when two SQL files have the same name (i.e., the "dbt found two resources with the name" CompilationException.  One place to add this flag would be in dbt_project.yml at the top-level.  Maybe like this:

dbt_project.yml:
```yaml
  use_reference_namespace = true
```

The next trick is to thread this parameter through to the places where the unique identifiers are generated.  As one example, in parser/base_sql.py `BaseSqlParser.parse_sql_node`, a `unique_id` is generated by joining the resource_type, package_name, and node name (i.e., filename base) using '.'.  This `unique_id` is later split apart in utils.py `id_matches()`.  As such, I don't think that we can use the '.' character as the separate the namespace.  Instead, maybe it could be a '/', just like the directory separator:

```python
select * from {{ ref('shopify_uk/sales') }}
union all
select * from {{ ref('shopify_fr/sales') }}
```

So, I think one solution would approximately be to update the `get_path` methods (in parser/schemas.py and parser/base.py) to take a node instead of name as the last parameter, and then conditionally include the directory prefix in the name based on the `use_reference_namespace` flag.  All these directories would be relative to the directory specified in dbt_project.yml, and could include subdirectories of subdirectories (e.,g, `ref('shopify_uk/sub_dir/sales')`).

Thoughts?  Thanks!

heisencoder:
[...]

PR #3247, opened 2021-04-12
Support dots in model names: Make FQN model selectors work with them

### LINKED ISSUE #3246: Support dots in model names
### Describe the bug
In our dbt project we are managing multiple schemas. We are namespacing the models with the standard SQL syntax `<schema>.<relation>.sql` and following [this approach](https://discourse.getdbt.com/t/extracting-schema-and-model-names-from-the-filename/575). It works quite well for the most part, however when specifying model selectors by FQN (`dbt run --model myschema.mytable`) the matcher is unable to locate the desired models

### Steps To Reproduce
Using a model or seed with dots in their name and trying to run it via FQN selector (`dbt run --model model.with.dots`) will trigger the issue.

### Expected behavior
`model.with.dots` runs successfully

Log output is:
```
The selector 'model.with.dots' does not match any nodes and will be ignored
```

### System information
- [ ] postgres
- [ ] redshift
- [ ] bigquery
- [x] snowflake

```
installed version: 0.19.1
   latest version: 0.19.1

Up to date!

Plugins:
  - bigquery: 0.19.1
  - postgres: 0.19.1
  - redshift: 0.19.1
  - snowflake: 0.19.1

```

```
OS Name: Microsoft Windows 10 Enterprise
OS Version: 10.0.18363 N/A Build 18363
```
`Python 3.6.13 :: Anaconda, Inc.`

### Additional context
Related to #1269

monti-python:
resolves #3246

### Description

Support dots in model names as namespace separators. In our dbt projects we are managing multiple schemas. We are namespacing the models with the standard SQL syntax `<schema>.<relation>.sql` and following [this approach](https://discourse.getdbt.com/t/extracting-schema-and-model-names-from-the-filename/575). It works quite well for the most part, however when specifying model selectors by FQN (dbt run --model myschema.mytable) the matcher is unable to locate the desired models

Somehow also related to #1269 

### Checklist


jtcohen6 (commented):
@monti-python This is neat! Thanks for the contribution :)

I know your primary ask here is around models, and that makes sense. I tested this with a few different resource types:
1. Works for: `model.with.dots`, `seed.with.dots`, `data.test.with.dots`. Nice! These are most of the resources that define their name by their file path, and where this functionality is most useful (avoiding collisions across namespaces).
2. Does not work for: `analysis.with.dots`, `source:src.with.dots`, `exposure:exp.with.dots`. These actually _do_ work in every other meaningful way—just not for `fqn:`/`source:`/`exposure:`-based selection.
3. Impossible to define: `snapshot.with.dots`, since the name comes from the Jinja tag (same problem with, e.g., `macro.with.dots`). We can reasonably consider this a hard limit on which resource names can contain dots, at least for the foreseeable future.

For consistency, I'd obviously prefer to extend this support to as many resource types as possible—but if you're at the limits of your interest here, I'd be equally happy with opening those as a new issue and documenting the limitation.

Separately: I love the unit tests that you've added. In the better-safe-than-sorry spirit, could I convince you to add an integration test with dot-named models, seeds, and tests? I just want to be sure that we don't, in a future version of dbt, accidentally regress behind what dbt _already_ supports today.

monti-python:
Thanks a lot for reviewing @jtcohen6 !

Regarding your remarks, I think this is indeed mostly interesting for models to avoid collisions, as other objects such as sources are inherently anchored to their relevant _namespaces_ . As you mentioned, I think it'd be more convenient to open new separate issues for other resources.

I've simplified the FQN matching logic and extended the following integration tests:
  * `024_custom_schema_test`: I've included a replica of our current setup. Schema and relation are parsed from the model name using custom macros, and all models are created accordingly. The test also exports the seed files to their corresponding schemas.
  * `007_graph_selection_tests` and `005_simple_seed_test`: Included support for dots in model names without schema/relation parsing. Only relevant for postgres because snowflake and bigquery will complain if there are dots in the relation names

Personally I think this feature is useful only when combined with custom parsing of `db`, `schema` and `alias`, therefore the latter integration tests might not be of relevance. Let me know what you think

jtcohen6 (approved):
@monti-python Really nice work on those tests! Thanks so much for this contribution, and for going above & beyond to carry this functionality into the future

PR #3257, opened 2021-04-13
Feature: test config parity

### LINKED ISSUE #3252: Support `enabled` configuration for generic tests
Today, it is possible to disable a one-off test by including `{{ config(enabled = false) }}` in its `.sql` file.

1. Make it possible to disable specific instances of generic tests via an `enabled` configuration/argument:
```yml
models:
  - name: model_a
    columns:
      - name: id
        tests:
          - unique:
              enabled: false
```

2. Make it possible to disable all instances of a generic test:
```sql
{% test not_null(model, column_name), adapter = 'default' %}

    {{ config(enabled = false) }}

    select * from {{ model }}

{% endtest %}
```

To be clear: Achieving item 2 is less important than item 1. The idea here is _not_ that the `not_null` "test macro" is itself disabled and missing from the project, but rather that all specific instances of the `not_null` test inherit the default configuration/argument of `enabled = false`. It's a matter of convenience, more than anything.

_Note: The ability to set "default" behavior for custom schema tests, by including `config()` in the test definition, was an undocumented (though reasonable) feature. There is a regression for it in v0.19.1. See #3240._

### LINKED ISSUE #3253: Tests are configurable from dbt_project.yml
### Describe the feature

This issue brings tests up to parity with other resource types in terms of being able to configure them from `dbt_project.yml`. In particular, this is essential for being able to control the behavior of tests installed from packages.

**dbt_project.yml**
```yml
# top-level: both bespoke + generic tests
tests:
  +enabled: true
  +severity: error
  +tags: ["tag_one", "tag_two"]
  
# for specific instances of generic tests defined on models
models:
  my_project:
    my_subfolder:
      +tests:
        enabled: true
        severity: warn

# for specific instances of generic tests defined on seeds
seeds:
  installed_package:
    +tests:
      enabled: false

# for specific instances of generic tests defined on snapshots
snapshots:
  +tests:
    tags: ["tag_three"]

# for specific instances of generic tests defined on sources
sources:
  +tests:
    severity: warn
```

### What does it mean to configure...
1. A bespoke test (e.g. `test/my_test.sql`)? Same it does today when I put `{{ config(severity = 'warn') }}` at the top of `test/my_test.sql`.
2. A generic test (e.g. `unique`)? I should be able to add _either_ `{{ config(severity = 'warn') }}` to my test block definition, or `severity=warn` as a default argument in the test signature. All specific instances of the `unique` test will now be warn-severity by default, unless the specific test overrides as below. (Note: There was a regression for this in v0.19.1, see #3240).
3. A specific instance of a generic test (e.g. `unique_model_a_id`)? Same as it does today when I add an "argument" to the test definition:
```yml
        tests:
          - unique:
               severity: warn
```

In this way, for generic tests, configs are like a special type of test block (macro) argument. You could think about this issue as being a way to "pass" arguments to a bunch of instantiated generic tests all at once, right from `dbt_project.yml`. That said, there are a few differences between configs and arguments for generic tests:
- Configs apply to **all** tests, bespoke and generic. Arguments are specific to one generic test block (e.g. the `values` for `accepted_values`)
- Only config values can be passed from `dbt_project.yml`
- How the `unique_id` is constructed for a specific test node (see #3254)

### Hierarchy

1. Most specific always wins: The `config()` block in a data test, or the value passed to a specific instance of a generic test where it's defined in (e.g.) a `models/*.yml` file
2. The default values set within the generic test block (either by `config()` block or by default arguments in the signature)
3. Config in `dbt_project.yml`, applying to _all_ tests
4. The configs of installed packages (1-3)

Edited: I previously had 2 last, but this is going to be difficult given the way this is implemented today. I don't mind the idea that the generic test's default values are still more specific than `dbt_project.yml` config. I'm also not sure how to think about #3256 relative to the ordering above.

### Additional context

Today, the only supported configs for tests are `enabled`, `severity`, and `tags` ([docs](https://docs.getdbt.com/reference/data-test-configs)). Generic tests support `severity` + `tags`, and they should `enabled` (https://github.com/fishtown-analytics/dbt/issues/3252). We hope to add net-new configs very soon as well (e.g. `where`, `limit`).

jtcohen6:
resolves #3252
resolves #3253

This PR makes it possible to:
- Disable schema tests where they're defined
- Configure tests from `dbt_project.yml`

**models/*.yml**:
```yml
      tests:
        - not_null:
            enabled: true
        - accepted_values:
            values: [1,2,3]
            enabled: false
```
**dbt_project.yml**:
```yml
tests:
  my_project:
    +severity: warn
  installed_package:
    +enabled: false
```

### Differences from proposal in #3253

- Rather than configuring schema tests underneath each of models/sources/etc, all tests are configured under `tests:`. I think that's okay.
- (After we solve v0.19.1 regression): Generic tests' default configs/arguments will override configurations set in `dbt_project.yml`. That's not the order I had originally envisioned, but again I think that's okay; it's up for debate which one is actually more generic / specific.

### Next issues

- Net-new configs (https://github.com/fishtown-analytics/dbt/issues/3258)
- Test FQNs could be a lot better, to make configuration easier (https://github.com/fishtown-analytics/dbt/issues/3259)

### Checklist

### DIFF
--- a/core/dbt/parser/schema_test_builders.py
+++ b/core/dbt/parser/schema_test_builders.py
@@ -184,8 +184,8 @@ class TestBuilder(Generic[Testable]):
         r'((?P<test_namespace>([a-zA-Z_][0-9a-zA-Z_]*))\.)?'
         r'(?P<test_name>([a-zA-Z_][0-9a-zA-Z_]*))'
     )
-    # map magic keys to default values
-    MODIFIER_ARGS = {'severity': 'ERROR', 'tags': []}
+    # kwargs representing test configs
+    MODIFIER_ARGS = ('severity', 'tags', 'enabled')
 
     def __init__(
         self,
@@ -217,11 +217,12 @@ class TestBuilder(Generic[Testable]):
         self.name: str = groups['test_name']
         self.namespace: str = groups['test_namespace']
         self.modifiers: Dict[str, Any] = {}
-        for key, default in self.MODIFIER_ARGS.items():
-            value = self.args.pop(key, default)
+        for key in self.MODIFIER_ARGS:
+            value = self.args.pop(key, None)
             if isinstance(value, str):
                 value = get_rendered(value, render_ctx)
-            self.modifiers[key] = value
+            if value is not None:
+                self.modifiers[key] = value
 
         if self.namespace is not None:
             self.package_name = self.namespace
@@ -267,8 +268,15 @@ class TestBuilder(Generic[Testable]):
             test_args['column_name'] = name
         return test_name, test_args
 
-    def severity(self) -> str:
-        return self.modifiers.get('severity', 'ERROR').upper()
+    def enabled(self) -> Optional[bool]:
+        return self.modifiers.get('enabled')
+
+    def severity(self) -> Optional[str]:
+        sev = self.modifiers.get('severity')
+        if sev:
+            return sev.upper()
+        else:
+            return None
 
     def tags(self) -> List[str]:
         tags = self.modifiers.get('tags', [])
@@ -303,15 +311,21 @@ class TestBuilder(Generic[Testable]):
             name = '{}_{}'.format(self.namespace, name)
         return get_nice_schema_test_name(name, self.target.name, self.args)
 
+    def construct_config(self) -> str:
+        configs = ",".join([f"{key}={value}" for key, value in self.modifiers.items()])
+        if configs:
+            return f"{{{{ config({configs}) }}}}"
+        else:
+            return ""
+
     # this is the 'raw_sql' that's used in 'render_update' and execution
     # of the test macro
     def build_raw_sql(self) -> str:
         return (
-            "{{{{ config(severity='{severity}') }}}}"
-            "{{{{ {macro}(**{kwargs_name}) }}}}"
+            "{config}{{{{ {macro}(**{kwargs_name}) }}}}"
         ).format(
             macro=self.macro_name(),
-            severity=self.severity(),
+            config=self.construct_config(),
             kwargs_name=SCHEMA_TEST_KWARGS_NAME,
         )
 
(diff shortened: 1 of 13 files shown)


core/dbt/parser/schema_test_builders.py
```diff
@@ -217,11 +217,12 @@ def __init__(
...
         self.namespace: str = groups['test_namespace']
         self.modifiers: Dict[str, Any] = {}
-        for key, default in self.MODIFIER_ARGS.items():
-            value = self.args.pop(key, default)
+        for key in self.MODIFIER_ARGS:
+            value = self.args.pop(key, None)
             if isinstance(value, str):
-                value = get_rendered(value, render_ctx)
-            self.modifiers[key] = value
+                value = get_rendered(value, render_ctx, native=True)
```
kwigley, line 223: ah!

gshank:
I'm wondering about the use of ResourceType.Seed.... It applies in some cases to everything in a schema file, not just tests. I introduced a ParseFileType in the load_all_files branch. Perhaps we should straighten out the ambiguous use of 'tests' for all schema files things versus just tests.

jtcohen6:
@gshank Could you say more about that? I just did a quick search for all the places `NodeType.Test` crops up, and all of them seem to be specific to data tests + schema tests. Granted, there are some wonky things we do to construct schema tests while reading schema (non-project `.yml`) files...

kwigley (approved):

core/dbt/parser/schema_test_builders.py
```diff
@@ -303,15 +311,25 @@ def get_test_name(self) -> Tuple[str, str]:
             name = '{}_{}'.format(self.namespace, name)
         return get_nice_schema_test_name(name, self.target.name, self.args)
 
+    def construct_config(self) -> str:
+        configs = ",".join([
+            f"{key}=" + (f"'{value}'" if isinstance(value, str) else str(value))
```
jtcohen6, line 316: @kwigley This is actually the line that fixed the `Undefined` Jinja error, because I was accidentally passing the config down on line 329 as:
```
{{ config(severity = warn) }}
```
Instead of `severity = 'warn'`. At the same time, we can't quote values that need to be non-string types (namely, `enabled` is boolean). This is a pretty janky way of writing that, I'm sure there's a more pythonic / f-string-smart way to do it.

Issue #1173, opened 2018-12-05
Schema test blocks


drewbanin:
### Describe the feature
dbt should support blocks that represent schema test definitions. A rough sketch of what these could look like:

```
{% test schema_test_name(....arguments) %}

{{ config(...test_configs) }}

select ...

{% endtest %}
```

### Implementation ideas:
 - The `name` of the test block should dictate how the calling schema.yml file calls this test
 - The `config` block should help provide a rendered name and description for the resulting test node
 - The test should return a set of failing rows for the test, rather than a count(*) directly

Important considerations:
 - databases like BigQuery charge based on data scanned, so a naive translation of existing `count(*)` style tests to a `select *` style test could be really expensive. Let's keep that in mind as we build out this functionality

### Configuration ideas:
To support dynamic test node names and descriptions, the `config` block should accept two configs:
 - name
 - description

A quick and imperfect example:
```
{% test unique(model, column_name) %}

{{ config(
  name="test_unique_{{ model.name }}.{{ column_name }}",
  description="Assert that {{ column_name }} is unique in {{ model }}"
) }}
....
```

I know, I know, curlies inside of curlies. Let's figure out a sensible way to make these names and descriptions easy to write.

When this feature is live, we should deprecate the old-style schema test macros and mark them for removal in a future release.

### Additional context
- The compiled SQL would be more inspectable in a query runner
  - We should think hard about how to 1) generate a test query and 2) generate a diagnosis/debugging query from the original select statement
- We could build-in support for things like a maxerror count, severity, or enable/disable flags at a `config` level
- We could add support for recording failing rows in a table for inspection (#903)

jtcohen6:
I wanted to share some of our latest thinking on schema test blocks, which we're hoping to include in v0.20.0. Necessary caveats: We're still working through the details and various implications; this is just the tip of the iceberg; anything below is still subject to change.

### Semantics

First off, I'd like to ditch the names "schema" and "data" tests, and many of the current distinctions between them. The two should provide the dbt developer two angles into the same functionality:
1. It's easy to define a one-off test as a SQL file. Write a query that returns the rows you don't want from the specific model(s). It's executed by `dbt test` every time.
2. If you want to reuse that test query for multiple models/columns, rather than copying and pasting, you can refactor the SQL to accept arguments. This is the reusable/generic/special-sort-of-macro `test` block proposed by this issue, and it would take the place of the `test__` macro currently used.

This issue is about the latter.

[...]

PR #3261, opened 2021-04-13
Add `test` Jinja tag

kwigley:
addresses part of #1173

### Description

This PR adds a `test` Jinja tag. This gives us the ability to treat tests differently than very specifically defined macros. There is no change in behavior with this PR.

### Checklist

### DIFF
--- a/core/dbt/compilation.py
+++ b/core/dbt/compilation.py
@@ -250,9 +250,6 @@ class Compiler:
 
         return str(parsed)
 
-    def _get_dbt_test_name(self) -> str:
-        return 'dbt__cte__internal_test'
-
     # This method is called by the 'compile_node' method. Starting
     # from the node that it is passed in, it will recursively call
     # itself using the 'extra_ctes'.  The 'ephemeral' models do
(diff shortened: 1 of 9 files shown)


core/dbt/compilation.py
```diff
@@ -250,9 +250,6 @@ def _inject_ctes_into_sql(self, sql: str, ctes: List[InjectedCTE]) -> str:
 
         return str(parsed)
 
-    def _get_dbt_test_name(self) -> str:
```
kwigley, line 253: Wee, since we don't prepend tests with `count(*)` when "compiling" this can go away!

iknox-fa (approved):

core/dbt/utils.py
```diff
@@ -101,6 +101,15 @@ def get_docs_macro_name(docs_name, with_prefix=True):
         return docs_name
 
 
+def get_test_macro_name(test_name, with_prefix=True):
+    name = f'test_{test_name}'
+
+    if with_prefix:
```
iknox-fa, line 107: Nit:  Use ternary here?

iknox-fa (approved):

jtcohen6 (approved):
This is backwards compatible in a very clever way, and I had to spend a few minutes thinking about it to really get it. Nice work!

A few things worth calling out, either because I think we should eventually document them, or because it reorients my own thinking:
- Test blocks are still technically macro nodes, just like materializations. As long as that's true, some of our other work around "configuring" generic tests (e.g. from `dbt_project.yml`) will have to work a bit differently. I actually think that's ok!
- A test node named `my_test` cannot coexist with a macro named `test_my_test`. I suppose the same is true of a materialization named `my_mtzn` and a macro named `materialization_my_mtzn`.
- If users are commonly calling other macros from within a test block, it's more likely that they will encounter the v0.19.1 regression (`config()` block or `ref` not properly accounted for), which @gshank is fixing by sticking the top-level test macro into the capture context. Once more, I think that's ok: We can tell users they need to hoist ref/source/config within the main test block to achieve the functionality they're after.

gshank (approved):

nathaniel-may (approved):
This is very slick! thanks for adding all the comments 🥇

PR #3117, opened 2021-02-19
git package: warn if specified revision master, main

jtcohen6:
Follow up from #3057, #3104

Previously, if a git package was unpinned, dbt tried to install it from `master` (hard coded). In #3104, this was fixed to used `HEAD` (default branch, regardless of name) and continue to warn if unpinned.

We'd like to warn as well if the user specifies the package `revision` as `'master'` or `'main'`, since either almost certainly represents the default branch. Users can disable the warning with `warn-unpinned: false`.

### Example

```yml
packages:
  - git: https://github.com/fishtown-analytics/dbt-codegen 
    revision: master
  - git: https://github.com/tailsdotcom/dbt_artifacts   
    revision: main
  - git: https://gitlab.com/gitlab-data/snowflake_spend
  - package: fishtown-analytics/audit_helper
    version: 0.3.0
```

<details>
<summary> <code>$ dbt deps</code> </summary>

```
Running with dbt=0.19.0
WARNING: The git package "https://github.com/fishtown-analytics/dbt-codegen"
	is pinned to the "master" branch.
	This can introduce breaking changes into your project without warning!

See https://docs.getdbt.com/docs/package-management#section-specifying-package-versions
WARNING: The git package "https://github.com/tailsdotcom/dbt_artifacts"
	is pinned to the "main" branch.
	This can introduce breaking changes into your project without warning!

See https://docs.getdbt.com/docs/package-management#section-specifying-package-versions
WARNING: The git package "https://gitlab.com/gitlab-data/snowflake_spend"
	is not pinned, using HEAD (default branch).
	This can introduce breaking changes into your project without warning!

See https://docs.getdbt.com/docs/package-management#section-specifying-package-versions
WARNING: The git package "https://github.com/fishtown-analytics/dbt-codegen"
	is pinned to the "master" branch.
	This can introduce breaking changes into your project without warning!

See https://docs.getdbt.com/docs/package-management#section-specifying-package-versions
WARNING: The git package "https://github.com/tailsdotcom/dbt_artifacts"
	is pinned to the "main" branch.
	This can introduce breaking changes into your project without warning!

See https://docs.getdbt.com/docs/package-management#section-specifying-package-versions
WARNING: The git package "https://gitlab.com/gitlab-data/snowflake_spend"
	is not pinned, using HEAD (default branch).
	This can introduce breaking changes into your project without warning!

See https://docs.getdbt.com/docs/package-management#section-specifying-package-versions
[...]

PR #3275, opened 2021-04-18
Feature/3117 dbt deps timeout

TeddyCr:
resolves #

### Description
resolves #3117 ([dbt deps hanging for hours](https://github.com/fishtown-analytics/dbt/issues/3177))

In this PR we added a `timeout` parameter to the `requests.get()` method to prevent requests from hanging for hours.

### Checklist


core/dbt/clients/registry.py
```diff
@@ -28,11 +28,14 @@ def wrapper(*args, **kwargs):
             attempt += 1
             try:
                 return fn(*args, **kwargs)
+            except requests.exceptions.Timeout as exc_timeout:
+                raise requests.exceptions.Timeout(
+                    'Request timeout, server has not issued a response for 60 seconds'
+                ) from exc_timeout
```
jtcohen6, line 34: Should this raise its own `Timeout`-specific error message after hanging for 60 seconds, or simply kick back to retry in the line below?

```python
except (requests.exceptions.ConnectionError, requests.exceptions.Timeout) as exc:
```
I think either could be a good answer!

test/unit/test_registry_get_request_timeout.py
```diff
@@ -0,0 +1,9 @@
+import unittest
+from requests.exceptions import Timeout
+
+from dbt.clients.registry import _get
+
+class testRegistryGetRequestTimeout(unittest.TestCase):
+    def test_registry_request_timeout(self):
+        # using non routable IP to test timeout logic in the _get function
+        self.assertRaises(Timeout, _get, '', 'http://10.255.255.1')
```
jtcohen6, line 9: Related to above, it looks like [the failing Windows test](https://dev.azure.com/fishtown-analytics/dbt/_build/results?buildId=3252&view=logs&j=1c778e8e-596b-5a0a-22c2-460364725d7c&t=1607d490-6a74-5e3f-97a9-72eb94e82d47) is skipping the `Timeout` error, retrying 5 times, and finally raising `dbt.exceptions.RegistryException: Unable to connect to registry hub`.

Let's do one of:
- Figure out why Windows is skipping the `Timeout` exception and jumping into retries
- Amend the code above to raise a consistent retry-based exception
- Edit this test to check for either `Timeout` or `RegistryException`

jtcohen6 (commented):
@TeddyCr This looks solid! Thanks so much for contributing, and sorry for the delay on my end. I left a few comments prompted by the failing Windows CI test.

core/dbt/clients/registry.py
TeddyCr: My initial thought was to have 2 distinct error handling, but since, it appears that, the azure pipeline does not raise a `Timeout`error, it might be safer to handle both in a generic error handling block as you wrote above.

test/unit/test_registry_get_request_timeout.py
TeddyCr: Based on the above update to the timeout error handling I went ahead and made a generic error catching test.

jtcohen6 (approved):
Thanks @TeddyCr!

Issue #2593, opened 2020-06-25
Storing test results in the database


jtcohen6:
### Describe the feature

dbt testing is powerful; being notified when a test fails can be life-saving; debugging those failed tests could be easier!

There are two long-lived issues related to this subject ([#517](https://github.com/fishtown-analytics/dbt/issues/517) and [#903](https://github.com/fishtown-analytics/dbt/issues/903)), and dozens of valuable comments on both. After discussing with several colleagues and a few community members, we've arrived at an approach I feel good about.

The goal here isn't to build a dimensional model of historical data surrounding dbt invocations. I think we can and should build out capacity for those longitudinal analyses, but it should be powered by dbt artifacts (manifest and run results) rather than adding records to database tables (a la [logging](https://github.com/fishtown-analytics/dbt-event-logging) package).

Rather, the goal is to add tooling that can be of immediate benefit to dbt developers, for whom debugging tests is a slog that's harder than it needs to be. We want to provide the post-test investigator with a high-level overview of `dbt test` statuses, and the errant records from any failing tests, within a query or two.

A lot of this work is going to build on top of v2 schema tests ([#1173](https://github.com/fishtown-analytics/dbt/issues/1173)) and data tests ([#2274](https://github.com/fishtown-analytics/dbt/issues/2274)), both of which are on the [1.0.0 docket](https://github.com/fishtown-analytics/dbt/labels/1.0.0). So while this feature is not coming in the next month or two, it is firmly on the roadmap.

## Spec: v2 schema tests

To get where we want to go, we need to talk a little about our planned changes to schema tests.

Let's imagine a modified version of the current `not_null` schema test. Several small things are different, the most significant of which is that the test returns a set of records instead of just `count(*)`, similar to how data tests work today:

```sql
{% test not_null(model, column_name) %}

    {% set name = 'not_null__' ~ model.name ~ '__' ~ column_name %}

    {% set description %} Assert that {{ column_name }} is never null in {{ model }} {% endset %}

    {% set fail_msg %} Found {{ result }} null values of {{ model }}.{{ column_name }} {% endset %}
        
    {{ config(name=name, description=description, fail_msg=fail_msg) }}
    
    select * from {{ model }}
    where {{ column_name }} is null

{% endmacro %}
```

Let's say we have two `not_null` tests defined in our project, in a file called `resources.yml`:

```yaml
version: 2

models:
  - name: customers
    colums:
      - name: id
        tests:
          - not_null

  - name: orders
[...]

PR #3286, opened 2021-04-22
use test materialization for schema/generic tests

### LINKED ISSUE #3192: Use test materialization when executing ✨generic✨ tests
### Current Behavior
generic (aka schema) tests are currently executed using the python api directly.
https://github.com/fishtown-analytics/dbt/blob/77c10713a325d2bee91d1822951ce5d91ccc3278/core/dbt/task/test.py#L84-L85

### Desired Behavior
Schema tests should be executed and handled by the same test materialization used for data tests. The goal here is to maintain existing behavior handling results (simply interpreting number of rows returned by `count(*)` for the test)

### Things to consider
Returning results for most schema tests is straight forward, we can simply remove `count(*)` and return the results of the query fulfilling the test, the materialization will take care of the rest. In some cases, this is not straight forward for tests that are simple pass/fail and do not necessarily return results.

Example:
https://github.com/fishtown-analytics/dbt/blob/749f87397ec1e0a270b2e09bd8dbeb71862fdb81/test/integration/005_simple_seed_test/macros/schema_test.sql

We will have to create a way to interpret different types of results as part of this.

kwigley:
resolves #3192 

### Description

- convert schema/generic tests to use `test` jijna tag
- update schema/generic tests to return rows that include failures instead of a count of failures (this still needs some love, but will help support #2593)

### Checklist


test/integration/022_bigquery_test/macros/test_creation.sql
```diff
@@ -15,4 +15,4 @@
         select 1 as error
     {% endif %}
```
jtcohen6, line 16: Does this just need the same treatment as above?
```sql
select '{{ table.type }} does not match expected value {{ type }}'
where '{{ table.type }}' != '{{ type }}'
```

test/integration/056_column_type_tests/macros/test_is_type.sql
```diff
@@ -65,5 +64,5 @@
         {% endif %}
     {% endfor %}
     {% do log('bad columns: ' ~ bad_columns, info=True) %}
-    select {{ bad_columns | length }} as pass_fail
-{% endmacro %}
+    select * from unnest(string_to_array('{{ bad_columns|join(',') }}', ','))
```
jtcohen6, line 67: fancy! :)

This works well on Postgres. If we needed to run it on any database, I think we could do:
```sql
{% for bad_column in bad_columns %}
  select '{{ bad_column }}' as bad_column
  {{ 'union all' if not loop.last }}
{% endfor %}
```

test/integration/056_column_type_tests/test_column_types.py
```diff
@@ -20,7 +21,7 @@ def models(self):
     def test_postgres_column_types(self):
         self.run_and_test()
 
-
+@pytest.mark.skip("Temporarily skipping while implementing the handling of test results.")
```
jtcohen6, line 24: 👍

jtcohen6 (commented):
This is looking really good!

One thing we can think about before / as part of #2593: _What_ data should each generic test query return? For the `not_null` test, the entire row (`*`) makes sense, but I could imagine the `unique` and `accepted_values` test including the count of each duplicated/unaccepted value, too. That does not need to block merging this PR, just something on my mind.

test/integration/022_bigquery_test/macros/partition_metadata.sql
```diff
@@ -1,22 +1,18 @@
...
-        select 0 as success
-    {% else %}
-        -- actual: {{ actual }}
-        -- expected: {{ expected }}
-        select 1 as error
-    {% endif %}
+    {% set success = 1 if model and actual == expected else 0 %}
+
+    select 'Expected {{ expected }}, but got {{ actual }}' as validation_error
+    where {{ success }} = 0
```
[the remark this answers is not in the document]
kwigley: The `where` hack actually doesn't work on bigquery :( 

`Query without FROM clause cannot have a WHERE clause at `[..]`

test/integration/056_column_type_tests/macros/test_is_type.sql
kwigley: I totally agree! The piece here that I'm struggling to handle correctly is if there are no bad columns (`bad_columns == []`), in this case, no sql will be rendered.

test/integration/022_bigquery_test/macros/partition_metadata.sql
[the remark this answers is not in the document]
jtcohen6: Heh, ok, try this:
```sql
select 'Expected {{ expected }}, but got {{ actual }}' as validation_error
from (select true)
where {{ success }} = 0
```

The `where` test config should work a little differently—that would be filtering/rewriting the actual `{{ model }}` reference (as a subquery). There would definitely be a `from` clause.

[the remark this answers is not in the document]
kwigley: `from`!!! ya ya ya

test/integration/056_column_type_tests/macros/test_is_type.sql
jtcohen6: Ohh right! So this is perhaps a bit jankier, but perhaps something like:
```
{% for bad_column in bad_columns %}
  select '{{ bad_column }}' as bad_column union all
{% endfor %}
  select * from (select 1 limit 0)
```

If we only run this test on postgres, I like your answer better :)

iknox-fa (approved):
LGTM

gshank (approved):

jtcohen6 (approved):
```
+78 −422
```
LGr8TM!

Issue #1132, opened 2018-11-15
Incremental models should handle column additions/deletions


drewbanin:
### Feature description
dbt should add or remove columns in the destination table for incremental models to make the target schema match the schema produced by the model SQL. dbt already has some existing logic that's responsible for "expanding" column types to make model SQL and the target table share a compatible schema. dbt should use a similar mechanism to add new columns and delete extraneous columns as required to make the target schema and the model SQL identical.

Ideas:
 - can we show a warning when this happens?
 - dbt should obviously add columns, but should it also delete them? Should that be configurable?

Caveats:
 - adding and removing columns on BQ is difficult
 - old records will be backfilled with `null

### Who will this benefit?
users of incremental models

elexisvenator:
Well.. this will open a can of worms.

Personally I am against any implicit smartness in this area by dbt. I feel this will only lead to confusion about why data has unexpected values.

Some other considerations:

* How do older columns get backfilled if they are meant to have data in them? If whatever logic in the model is able to detect those rows, then you are looking at what is essentially a slower version of a table refresh
* Columns that should default to something other than null? eg Booleans or non-nullable foreign keys. Perhaps something in the `schema.yml` to specify a default value?
* What is the value of `{{ this }}`? Given almost every incremental model uses that macro to detect whether a row needs updating, that is kindof important.

bobzoller:
dbt handling schema changes automatically is akin to how fivetran/stitchdata ingestors behave.  the upside is they don't stop ingesting, the downside is you inevitably end up chasing down issues and retroactively.

I think I'd prefer being forced to deal with the discrepancy when it happens (eg dbt should throw an error).  My only issue with the current behavior is there seemed to be some cases where it would fail silently and others where it would throw an error.

Perhaps there's still value so long as you have to opt-in to this behavior?

clrcrl:
I think it's undesirable that there is currently no warning whatsoever, however I tend to agree with @elexisvenator that implicit smartness may not be the way to do.

A happy medium might be to have dbt warn you, and provide migration statements that need to be run. Something along the lines of:
```
The incremental model "analytics"."events" contains column(s) that do not exist in the target table. These columns WERE NOT appended to the target table.
To append these columns:
* If backfilling this data is important, run a full-refresh on this table, or,
* If backfilling this data is not important, and it is desirable to preserve the existing target table, manually update the target table to include the new column(s), by running the following SQL:
alter table "analytics"."events"
[...]

PR #3288, opened 2021-04-22
Feature/incremental schema changes

matt-winkler:
addresses #1132 

### Description

Goal is to identify whether there are different schemas in the source and target relations during an incremental run. The allowed behaviors are `['full_refresh', 'ignore', 'fail', 'append', 'sync']` specified in a new config option called `on_schema_change` on the incremental.sql materialization. 

New macros added to helpers.sql:

- incremental_validate_on_schema_change: defaults to 'ignore' if None or invalid values are specified
- diff_columns: detects differences available columns based on name only (i.e. doesn't check for data type differences)
- sync_columns: runs ALTER TABLE statements depending on the value of `on_schema_change` and whether the source / target column sets differ.

### Checklist


jtcohen6:
@cla-bot check

core/dbt/include/global_project/macros/materializations/helpers.sql
```diff
@@ -72,3 +72,77 @@
...
+     {{ return(on_schema_change) }}
+   
+   {% endif %}
+
+{% endmacro %}
+
+{% macro diff_columns(array_one, array_two) %}
+  {% set result = [] %}
+   {%- for col in array_one -%} 
+     {%- if col not in array_two -%}
```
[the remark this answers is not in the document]
matt-winkler: Thank you for pointing this out! I was wondering if this might be an issue since we want to detect name changes only at this point.

core/dbt/include/global_project/macros/materializations/helpers.sql
```diff
@@ -72,3 +72,77 @@
...
+{% macro sync_columns(source_relation, target_relation, on_schema_change='append') %}
+  
+  {%- set source_columns = adapter.get_columns_in_relation(source_relation) -%}
+  {%- set target_columns = adapter.get_columns_in_relation(target_relation) -%}
+  {%- set add_to_target_arr = diff_columns(source_columns, target_columns) -%}
+  {%- set remove_from_target_arr = diff_columns(target_columns, source_columns) -%}
+
+  {%- if on_schema_change == 'append' -%}
+    {%- for col in add_to_target_arr -%}
+       {%- set build_sql = 'ALTER TABLE ' + target_relation.schema+'.'+target_relation.name + ' ADD COLUMN ' + col.name + ' ' + col.dtype -%}
```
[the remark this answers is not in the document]
matt-winkler: Great catch!

CHANGELOG.md
```diff
@@ -22,6 +22,7 @@
 - Support disabling schema tests, and configuring tests from `dbt_project.yml` ([#3252](https://github.com/fishtown-analytics/dbt/issues/3252),
 [#3253](https://github.com/fishtown-analytics/dbt/issues/3253), [#3257](https://github.com/fishtown-analytics/dbt/pull/3257))
 - Add Jinja tag for tests ([#1173](https://github.com/fishtown-analytics/dbt/issues/1173), [#3261](https://github.com/fishtown-analytics/dbt/pull/3261))
+- Support detecting schema changes on incremental models per [#1132](https://github.com/fishtown-analytics/dbt/issues/1132)
```
jtcohen6, line 25: Nit: we like to include both the issue # and the PR # in the changelog entry
```suggestion
- Support detecting schema changes on incremental models ([#1132](https://github.com/fishtown-analytics/dbt/issues/1132), [#3288](https://github.com/fishtown-analytics/dbt/issues/3288))
```
Please also add yourself to the list of contributors below! It's well deserved :)

core/dbt/include/global_project/macros/materializations/helpers.sql
```diff
@@ -72,3 +72,96 @@
...
+  {% set result = [] %}
+  
+  {% for col in columns %}
+    {{ result.append(col.column) }}
+  {% endfor %}
+  
+  {{ return(result) }}
+{% endmacro %}
+
+{% macro diff_columns(source_columns, target_columns) %}
```
jtcohen6, line 99: I just searched for "Jinja set difference" and, of course, came across https://github.com/fishtown-analytics/dbt/issues/1894 :)

I do wonder if something like is better implemented as a python method. Let's think about the right balance between clarity and confusion. Python obfuscates logic away from user space, but it can express certain logical operations much more cleanly.

core/dbt/include/global_project/macros/materializations/incremental/incremental.sql
```diff
@@ -25,12 +27,34 @@
...
+          }}
+      
+        {# unless we ignore, run the sync operation per the config #}
+        {% elif on_schema_change != 'ignore' %}
+        
+          {% set schema_changes = sync_columns(tmp_relation, target_relation, on_schema_change) %}
+
+        {% endif %}
+
+      {% endif %}
```
jtcohen6, line 50: Could we wrap this in a macro? We'll need to add this logic to every `incremental` materialization (including Snowflake and BigQuery), and I think it will be DRYer to call a macro rather than copy-paste.

I do think it makes a ton of sense for each incremental materialization to be responsible for including logic like:
```sql
{% set on_schema_change = incremental_validate_on_schema_change(config.get('on_schema_change')) %}
{% do process_schema_changes(on_schema_change, tmp_relation, target_relation) %}
```
Where `process_schema_changes` is a macro that wraps logic for the `on_schema_change`-specific behavior, whether raising the compilation error or running `sync_columns`.

core/dbt/include/global_project/macros/materializations/helpers.sql
```diff
@@ -72,3 +72,96 @@
...
+  {% set result = [] %}
+  {% set source_names = get_column_names(source_columns) %}
+  {% set target_names = get_column_names(target_columns) %}
+   
+   {# check whether the name attribute exists in the target, but dont worry about data type differences #}
```
jtcohen6, line 104: What happens if there _is_ a data type difference? It's possible that the model will fail its incremental build, though it's also possible that the database will neatly handle by coercing the new data into the old data type.

My immediate inclination is to be strict here: If the data types are incompatible, we should consider that a schema change. Depending on the user-supplied value of `on_schema_change`, that could mean raising a compilation error, or attempting to handle within `sync_columns` (i.e. via [`adapter.alter_column_type`](https://github.com/fishtown-analytics/dbt/blob/8c55e744b8e86b282dcea75e5efcf7c15a7142b9/core/dbt/include/global_project/macros/adapters/common.sql#L123)).

(We also have some logic built into the column object that can help with ensuring that numeric / varchar sizes don't get detected as changes. The incremental materialization [already handles](https://github.com/fishtown-analytics/dbt/blob/8c55e744b8e86b282dcea75e5efcf7c15a7142b9/core/dbt/include/global_project/macros/materializations/incremental/incremental.sql#L30-L32) the case where a string column is wider in the new data than in the existing table.)

core/dbt/include/global_project/macros/materializations/helpers.sql
```diff
@@ -72,3 +72,96 @@
...
+  {% elif on_schema_change == 'sync' %}
+    {%- for col in add_to_target_arr -%}
+       {%- set build_sql = 'ALTER TABLE ' + target_relation.database + '.' + target_relation.schema+'.'+target_relation.name + ' ADD COLUMN ' + col.name + ' ' + col.dtype -%}
+       {%- do run_query(build_sql) -%}
+    {%- endfor -%}
+
+    {%- for col in remove_from_target_arr -%}
+      {%- set build_sql = 'ALTER TABLE ' + target_relation.database + '.' + target_relation.schema+'.'+target_relation.name + ' DROP COLUMN ' + col.name -%}
+      {%- do run_query(build_sql) -%}
+    {%- endfor -%}
```
jtcohen6, line 154: 1. These need to be adapter-dispatched macros, since the specific DML syntax is going to look different on different adapters/databases.
2. On many adapters, it's possible to add/remove multiple columns in a single `ALTER TABLE` statement. I think that's preferable, for concision in general, and for adapters that don't use transactions in particular—if any of the column additions/deletions fail, the whole operation will roll back. The only tricky thing is getting the commas right. E.g. in PostgreSQL:
```sql
alter table dbt_jcohen.incremental
    add column fun int,
    add column funny varchar(10),
    drop column yelly,
    drop column hilarious;
```

So I think it would make sense to add a new macro to [`macros/adapters/common`](https://github.com/fishtown-analytics/dbt/blob/develop/core/dbt/include/global_project/macros/adapters/common.sql):
```sql
{% macro alter_relation_add_remove_columns(relation, add_columns = none, remove_columns = none) -%}
  {{ return(adapter.dispatch('alter_relation_add_column')(relation, add_columns, remove_columns)) }}
{% endmacro %}

{% macro default__alter_relation_add_remove_columns(relation, add_columns = none, remove_columns = none) -%}
  {% set sql -%}
      alter {{ relation.type }} {{ relation }}
          {% for column in add_columns %}
            add column {{ column.name }} {{ column.data_type }} {{ ',' if not loop.last }}
          {% endfor %}
          {{ ',' if remove_columns %}
          {% for column in remove_columns %}
            drop column {{ column.name }}
          {% endfor %}
  {%- endset %}
  {{ return(run_query(sql)) }}
{% endmacro %}
```
Then the logic here can just look like:
```sql
{%- if on_schema_change == 'append' -%}
   {%- do alter_relation_add_remove_columns(target_relation, add_to_target_arr) -%}
{% elif on_schema_change == 'sync' %}
   {%- do alter_relation_add_remove_columns(target_relation, add_to_target_arr, remove_from_target_arr) -%}
{% endif %}
```

core/dbt/include/global_project/macros/materializations/helpers.sql
```diff
@@ -72,3 +72,96 @@
   {% endif %}
   {% do return(config_full_refresh) %}
 {% endmacro %}
+
+
+{% macro incremental_validate_on_schema_change(on_schema_change, default_value='ignore') %}
+   
+   {% if on_schema_change not in ['sync', 'append', 'fail', 'ignore'] %}
```
jtcohen6, line 79: - Whoa, `sync` is very cool! Could we make this even more explicit: `sync_all_columns`?
- I hesitate with the name `append`, since that (to my mind) always means adding _rows_, rather than adding _columns_. I get what you mean—add columns without removing any—but I wonder if we could find a different word here (or be more verbose: `add_new_columns`).
- The original proposal in #1132 also had a `refresh` option, motivated by the cases where users would want to fully backfill new columns (rather than null values for old rows). What do you think? Better to `'fail'` and give the user an explicit error telling them to manually refresh? I could go either way here.

core/dbt/include/global_project/macros/materializations/helpers.sql
```diff
@@ -72,3 +72,96 @@
...
+  {% endif %}
+
+  {{ 
+      return(
+             {
+              'columns_added': add_to_target_arr,
+              'columns_removed': remove_from_target_arr
+             }
+          )
+  }}
```
jtcohen6, line 165: I like the idea of logging some of this information (to debug, not to info/stdout). Something roughly like:
```
Schema change detected. dbt will {{ on_schema_change }} by adding {{ columns_added }} and removing {{ columns_removed }}.
```

jtcohen6 (commented):
@matt-winkler This is incredible! Way to take a 2.5-year-old dream and turn it into reality :)

This works wonderfully for me locally on Postgres/Redshift. I left a handful of comments below, largely around extending this pattern to other adapters. We'll definitely want to add some integration tests, too. I'd be thrilled to help however I can.

core/dbt/include/global_project/macros/materializations/helpers.sql
matt-winkler: @jtcohen6 On board with those naming changes (`sync` --> `sync_all_columns` and `append` --> `add_new_columns`). In general I like a bit of verbosity up to a point :). 

On the third point about refresh behavior, I think that should be handled in a separate unit of work, and that this current approach should update the columns and notify the user of what happened. I have a start on some additional code that would backfill columns according to data type (e.g. an INT column gets -1 for everything), with those default values specified in a mapping input to the materialization config.. Are you thinking we'd (optionally) backfill with actual data as well?

core/dbt/include/global_project/macros/materializations/helpers.sql
matt-winkler: Right on - will make these updates - I was going to ask how to correct this for each database provider, so thanks for commenting on that proactively!

core/dbt/include/global_project/macros/materializations/incremental/incremental.sql
matt-winkler: @jtcohen6 up good call to abstract this.

core/dbt/include/global_project/macros/materializations/helpers.sql
jtcohen6: For simplicity's sake, I think the default behavior for syncing/adding columns should be the database's default behavior, i.e. backfilling with null values. (I'm hesitant about having users supply a full mapping of every column and its default value in the case of schema evolution, though we could consider adding `default` as a property in the model's `columns` YAML dict.)

I do think, if a user wants the _actual_ values, that should require a full refresh. The question is, should full refreshing the table, in the case that a schema change is detected, be an optional strategy of `on_schema_change`? Or must a user manually pass the `--full-refresh` flag / set the `should_full_refresh` config?

core/dbt/include/global_project/macros/materializations/helpers.sql
matt-winkler: @jtcohen6 Yeah haha this was the piece I was most on-the-fence about in the PR. The scenario you described where the data type changes in the source in such a way that it's incompatible with the target makes sense and is a vulnerability of the current approach, so I agree we should handle that explicitly in the code for `sync_all_columns`. At the very least this should trigger a message to the user. It feels like dropping and recreating the target column with the new type from the source may be a bridge too far, because I *don't* think we should encourage users to mindlessly change data types and expect the system to "handle it". Agree?

core/dbt/include/global_project/macros/materializations/helpers.sql
matt-winkler: Yeah I was a bit annoyed to have to write this hehe. This approach seems relatively low risk for pushing the enhancement to incremental logic forward, and I'd be happy to dive into adding a python method to enhance the concision in the coming weeks.

CHANGELOG.md
matt-winkler: 10-4, thanks for the commentary.

core/dbt/include/global_project/macros/materializations/helpers.sql
matt-winkler: Good call.

core/dbt/include/global_project/macros/materializations/helpers.sql
matt-winkler: I could see use cases where making that automated full refresh configurable would be valuable. That would actually be relatively straightforward to program with a config like you described.

matt-winkler: So, NULLs when new columns are added is the new default behavior, and optionally users can trigger a full refresh when either `append_new_columns` or `sync_all_columns` is specified.

core/dbt/include/global_project/macros/materializations/helpers.sql
jtcohen6: Depending on the database, it may support altering column types in place (without needing to drop the column + recreate it). Of course, Redshift only supports that for varchar resizing, but other databases are a little cleverer (e.g. [Postgres](https://www.postgresql.org/docs/9.3/ddl-alter.html#AEN2819), [Snowflake](https://docs.snowflake.com/en/sql-reference/sql/alter-table-column.html)) and, in the event of an unsupported type switch, they would throw a more specific error message.

The choice here feels like, Should we try to implicit alter the type (which may or may not return an error), or just raise an error? I'd be happy if the v1 of this cannot account for type changes and always raises an explicit exception.

core/dbt/include/global_project/macros/materializations/helpers.sql
jtcohen6: Hm, I think what I'm sayig is, the `full_refresh` option is orthogonal to sync/add columns, and it also lets us keep the sync/add options simpler.

By making `full_refresh` another potential value of `on_schema_change`, we're letting the user pick any of:
- Just add new columns (backfilled with nulls)
- Sync new + existing columns (backfilled with nulls)
- Rebuild the entire table from scratch (real backfill, but also $$$)
- Ignore (current behavior)
- Aggressively raise an error

The status quo is, the user needs to manually determine if there's a schema change, and if so, pass `--full-refresh`. This would be an improvement insofar as the incremental model would only full refresh if dbt determines it needs to.

matt-winkler: Ooooh! So we'd have the possible values of `on_schema_change` be `['fail', 'ignore', 'sync_all_columns', 'append_new_columns', 'full_refresh']` then?

jtcohen6: That's what I'm thinking... what say you?

matt-winkler: Makes sense! In my brain the "full refresh in the case of schema change" functionality was a *separate* config set as a True/False, so comes to the same thing behavior wise, but this is a cleaner way to expose it to the user.

matt-winkler:
@jtcohen6 It would be great to get your further commentary on updates and where updates to each database adapter need to go. I'm guessing those changes need to go into `/plugins`? Also, I'm not aware of any ability to access Redshift for testing - let me know if you don't think that's necessary or it setting up a Redshift connection seems worthwhile.

core/dbt/include/global_project/macros/materializations/incremental/incremental.sql
```diff
@@ -7,15 +7,31 @@
...
 
+  {% set on_schema_change = incremental_validate_on_schema_change(config.get('on_schema_change')) %}
+
   {{ run_hooks(pre_hooks, inside_transaction=False) }}
 
   -- `BEGIN` happens here:
   {{ run_hooks(pre_hooks, inside_transaction=True) }}
 
+  {% do run_query(create_table_as(True, tmp_relation, sql)) %}
+  {% set schema_changed = check_for_schema_changes(tmp_relation, target_relation) %}
```
jtcohen6, line 18: Hm hm hm. On second thought, I don't think we want to create the temp table _before_ checking to see if we should full refresh for `should_full_refresh()` or `existing_relation.is_view` reasons. And if we do need to full-refresh for schema-change reasons, we'll want to swap the existing table with a new _permanent_ table, rather than a new _temp_ table.

Let's think more about the right way to structure this logical flow. I bet we can come up with something that makes sense and reads clearly.

jtcohen6 (commented):
I'm really excited about how far along this is already. I definitely want to take another deep dive into the code here next week, and start thinking about how want to test this.

> It would be great to get your further commentary on updates and where updates to each database adapter need to go. I'm guessing those changes need to go into `/plugins`?

Totally, I think it looks like:
- Including the same `trigger_full_refresh` + `process_schema_changes` logic in the incremental materializations for [Snowflake](https://github.com/fishtown-analytics/dbt/blob/develop/plugins/snowflake/dbt/include/snowflake/macros/materializations/incremental.sql) + [BigQuery](https://github.com/fishtown-analytics/dbt/blob/develop/plugins/bigquery/dbt/include/bigquery/macros/materializations/incremental.sql), just as you've included it in the default implementation
- If/as needed, reimplementing `alter_relation_add_remove_columns` for Snowflake and/or BigQuery
- Add `on_schema_change: str = 'ignore'` to the [`NodeConfig` class](https://github.com/fishtown-analytics/dbt/blob/5956a64b0109f7b33681d37fa49364e900b730ac/core/dbt/contracts/graph/model_config.py#L349). This makes it possible for (a) users to set it as a config in `dbt_project.yml`, (b) it to show up in the `configs` dict in `manifest.json` (hooray for metadata!)

> Also, I'm not aware of any ability to access Redshift for testing - let me know if you don't think that's necessary or it setting up a Redshift connection seems worthwhile.

We have a sandboxed Redshift cluster that we use for dbt integration testing (including the CI jobs that run on every commit you push here), I can get you access.

matt-winkler:
Hey @jtcohen6 I made some additional changes here, including to the main incremental materialization for Snowflake. There are still several if/else conditions nested in there that we could potentially clean up, but I didn't see obvious ways to simplify that logic further. 

I am thinking that once we feel good about the logic here, we can layer additional detection of changes based on names OR data types and trigger the same behavior as here. 

Testing is the other area where I could use your brain - I am seeing some unit and integration tests fail with this, but they are (seemingly) unrelated to the changes I'm making. But, that could easily be lack of awareness on my part.

core/dbt/include/global_project/macros/materializations/incremental/incremental.sql
```diff
@@ -28,13 +32,23 @@
...
       {% do to_drop.append(backup_relation) %}
+  
   {% else %}
-      {% set tmp_relation = make_temp_relation(target_relation) %}
       {% do run_query(create_table_as(True, tmp_relation, sql)) %}
+      {% set schema_changed = check_for_schema_changes(tmp_relation, target_relation) %}
+  
+      {% if schema_changed and on_schema_change == 'full_refresh' %}
+        {# placeholder for full refresh from permanent table #}
+      {% endif %}
```
jtcohen6, line 42: Okay, I've thought about it some more, and I think we have two options in the case of `on_schema_change == 'full_refresh'`:
1. Create the temp table first, compare the schemas. If the schemas mismatch, proceed down a copy of the full-refresh conditional path: abandon the temp table, create the _full_ table, proceed with swap/atomic replacement as appropriate. (Downside: we'll need to reimplement the logic.) If the schemas match, proceed down incremental conditional branch.
2. Compare the schema _before_ creating the temp or full table. How? I think we could use [`get_columns_in_query`](https://github.com/fishtown-analytics/dbt/blob/85e2c897945e6372fe3657075b2c5ea4b068d3e0/core/dbt/include/global_project/macros/adapters/common.sql#L1-L15), to return just the list of column names. Then we can compare against the current table; if they don't match, proceed down usual full-refresh path; if they do match, proceed down usual incremental path. No need to reimplement the logic.
```sql
{% set schema_change_full_refresh = False %}
{% if on_schema_change == 'full_refresh' %}
  {# could wrap the logic below in a macro / adapt check_for_schema_changes to account for it #}
  {% set source_columns = get_columns_in_query(sql) %}
  {% set target_columns = adapter.get_columns_in_relation(target_relation) %}
  {%- set source_not_in_target = diff_columns(source_columns, target_columns) -%}
  {%- set target_not_in_source = diff_columns(target_columns, source_columns) -%}
  {%- set schema_change_full_refresh = (source_not_in_target or target_not_in_source) %}

{% if existing_relation is none %}
      {% set build_sql = create_table_as(False, target_relation, sql) %}

{% elif should_full_refresh() or existing_relation.is_view or schema_change_full_refresh %}
-- normal full-refresh 

{% else %}
-- normal incremental
```

I'm inclined toward option #2, since it minimizes the duplication of logic, _but_ it would be an acknowledgment that `on_schema_change: full_refresh` will not be able to detect data type mismatches.

core/dbt/include/global_project/macros/materializations/incremental/incremental.sql
```diff
@@ -28,13 +32,23 @@
       {% set build_sql = create_table_as(False, intermediate_relation, sql) %}
       {% set need_swap = true %}
       {% do to_drop.append(backup_relation) %}
+  
   {% else %}
-      {% set tmp_relation = make_temp_relation(target_relation) %}
       {% do run_query(create_table_as(True, tmp_relation, sql)) %}
+      {% set schema_changed = check_for_schema_changes(tmp_relation, target_relation) %}
```
jtcohen6, line 38: This macro needs to run introspective queries, so it shouldn't execute if `on_schema_change = 'ignore'`—or if, per comment below, `on_schema_change = 'full_refresh'` and the schema has already been checked to trigger standard full-refresh logical pathway.

You could either or embed that check within these macros, or add it as an explicit check here in the materialization. I think I prefer the explicitness of having it here:
```sql
{% do run_query(create_table_as(True, tmp_relation, sql)) %}
{% if on_schema_change not in ('ignore', 'full_refresh') %}
  {% set schema_changed = check_for_schema_changes(tmp_relation, target_relation) %}
  {% do process_schema_changes(schema_changed, on_schema_change, tmp_relation, target_relation) %}
{% endif %}

File: plugins/snowflake/dbt/include/snowflake/macros/materializations/helpers.sql [hunk 13]
```diff
@@ -0,0 +1,121 @@
+{% macro incremental_validate_on_schema_change(on_schema_change, default_value='ignore') %}
```
jtcohen6, line 1: Are any of the macros in this file different from the ones defined in `global_project`? If not, you don't need to define them again here, Snowflake will use the default versions. (I see `snowflake_diff_columns` below, but I don't actually think it's different from `diff_columns`.)

core/dbt/include/global_project/macros/materializations/helpers.sql
jtcohen6: I'd be happy with punting on data type differences for the v1 of this, documenting it as a known limitation, opening a ticket to come back to it

core/dbt/include/global_project/macros/materializations/helpers.sql
```diff
@@ -72,3 +72,141 @@
   {% endif %}
   {% do return(config_full_refresh) %}
 {% endmacro %}
+
+
+{% macro incremental_validate_on_schema_change(on_schema_change, default_value='ignore') %}
```
jtcohen6, line 77: There's enough here to warrant a new file. Let's move all of these macros (from here down) into `global_project/macros/materializations/incremental/on_schema_change.sql`

jtcohen6 (commented):
@matt-winkler Excited about progress here. Let me know how I can help get this over the finish line.

1. Let's get the logic right for the default materialization, and then make it as consistent as possible (few if any macro reimplementations) across the Snowflake + BigQuery materializations as well. Namely that looks like a consistent process for handling `full_refresh` and `ignore` (do nothing).
2. I'd be happy punting on data type discrepancies, and prioritize a clean first cut of this with column names only.

As far as testing goes: The test failures are resulting from the addition of `on_schema_change` to the `NodeConfig` with a default value. There's several places in our unit tests where dbt compares actual node configs against dictionaries of expected values; we'll just need to add it to those dictionaries.

There's another, easier out here: While I'd like to see us be strict and consistent about including all official, documented configs in the `NodeConfig`, it's still technically possible to set configs within a model's `config()` and `dbt_project.yml` even if it's _not_ an official node config. That's how, e.g., `incremental_strategy` and `unique_key` work. Given how closely related the `on_schema_change` config is to those, I'd be fine with rolling back [this change](https://github.com/fishtown-analytics/dbt/commit/1549af6b3d9e7db4cb73cd1727a1d19252382548#diff-007a86f768b23e1516e6d253d71656b96d46672fc7432dbc81d1d9445c98ca05R397) accordingly, and it should magically fix the unit tests.

In any case, we'll want to add integration tests of our own, to verify all the potential values and behaviors of `on_schema_change`.

plugins/snowflake/dbt/include/snowflake/macros/materializations/incremental.sql
```diff
@@ -37,6 +37,7 @@
 
   {#-- Validate early so we don't run SQL if the strategy is invalid --#}
   {% set strategy = dbt_snowflake_validate_get_incremental_strategy(config) -%}
+  {% set on_schema_change = incremental_validate_on_schema_change(config.get('on_schema_change')) %}
```
jtcohen6, line 40: If we choose to roll back the addition of `on_schema_change` as a NodeConfig, we'll need to set the default here instead:
```suggestion
  {% set on_schema_change = incremental_validate_on_schema_change(config.get('on_schema_change'), default = 'ignore') %}
```

core/dbt/include/global_project/macros/materializations/helpers.sql
matt-winkler: Ok good deal, moving that code for postgres. I also got some good news during testing (on Snowflake, at least) that we essentially get the check on data type differences for free, because the `in` returns `False` when a checked column doesn't match the data type of what's in the reference list.

plugins/snowflake/dbt/include/snowflake/macros/materializations/helpers.sql
matt-winkler: Got it - moving the logic from snowflake to `global_project`.

jtcohen6:
Closing in favor of #3387, which is the narrower version we're planning to ship in v0.21

PR #3296, opened 2021-04-26
Include parent adapters in dispatch

### LINKED ISSUE #2923: Use all adapter names for adapter.dispatch
### Describe the feature

The `dbt-redshift` adapter plugin inherits from the `dbt-postgres` adapter plugin. This reflects the fact that they share a lot of code: connection mechanism (`psycopg2`), basic syntax, types, materializations, etc. (This can be obfuscated by the fact that their shared implementations of macros are also often the `default__` dbt implementations.)

Today, when dbt dispatches `package_name.macro_name` macro on Redshift, it looks for:

1. `my_project.redshift__macro_name`
2. `my_project.redshift__macro_name`
3. `package_name.postgres__concat`
4. `package_name.postgres__concat`

I think it should instead look for:

1. `my_project.redshift__macro_name`
2. `my_project.postgres__macro_name`
3. `my_project.default__macro_name`
4. `package_name.redshift__macro_name`
5. `package_name.postgres__macro_name`
6. `package_name.default__macro_name`

This is a possibility we weighed in #2679, to the point that we make it possible as a one-line change:
https://github.com/fishtown-analytics/dbt/blob/5ba5271da99bc1ef7fbad2f7d0b45634087300fa/core/dbt/context/providers.py#L104-L109

### What's the risk?

There are cases today where Redshift uses the `default__` implementation and Postgres uses a custom `postgres__` implementation, e.g. [`dbt_utils.dateadd`](https://github.com/fishtown-analytics/dbt-utils/blob/master/macros/cross_db_utils/dateadd.sql).

I don't think this is a breaking change per se, because Redshift largely supports PostgreSQL syntax, but we may want to add implementations that explicitly preserve existing behavior:

```sql
{% macro redshift__dateadd(datepart, interval, from_date_or_timestamp) %}
    {% do return(default__dateadd(datepart, interval, from_date_or_timestamp)) %}
{% endmacro %}
```

### Who will this benefit?
This will enable maintainers of similar-yet-slightly-different plugins, e.g. `dbt-sqlserver` and `dbt-synapse`, who can then "share" macro implementations between parent and child adapters. See https://github.com/dbt-msft/dbt-synapse/issues/18.

It's possible to imagine many more adapter inheritances in the future:
- a dedicated `dbt-databricks` or `dbt-delta` that inherits from `dbt-spark`
- a `dbt-hive` that inherits from `dbt-spark` or `dbt-presto`
- a `dbt-materialize` that inherits from `dbt-postgres`

jtcohen6:
resolves #2923

### Description

Include parent adapters in dispatch search order, i.e. `dbt-redshift` will look for dispatched implementations in the following order:
1. `redshift__`
2. `postgres__`
3. `default__`

This should make it easier for maintainers of plugins that inherit from parent adapters and wish to reuse their macros, namely:
- `dbt-synapse` (inherits from `dbt-sqlserver`)
- `dbt-materialize` (inherits from `dbt-postgres`)
- future?????: `dbt-databricks` (`dbt-spark`), `dbt-trino` / `dbt-starburst` (`dbt-presto`)

Note that this will require change in dbt-utils (and any other package) wherever `dbt-redshift` expects to use the `default__` version instead of the `postgres__` version. The require change would simply be wrapping / passing from `redshift__` to `default__`.

### TODO

~I really don't understand the MagicMock unit test that's failing, and would definitely need help sorting this out :)~

I resolved this by switching `get_adapter_type_names(self._config.credentials.type)`  to `get_adapter_type_names(self._adapter.type())`. In the failing unit test (`test/unit/test_context.py::test_resolve_default`), the former was `'redshift'` while the latter was `'postgres'`. The distinction here is very subtle—I see that the credentials `type` is used elsewhere in `providers.py`—and it's not clear to me whether we should expect them to diverge in any real-world (non-mock) scenarios.

### Checklist


gshank:
I think in practice self._config.credentials.type and self._adapter.type probably amount to the same thing. There's probably some value in being consistent in the way we call it though.

I think the problem is probably that the test is wrong. It certainly looks like it intends to test the postgres_adapter, but the 'type' that's being returned for self._config.credentials.type is 'redshift'. I'll take a stab at trying to fix the test.

gshank (approved):

jtcohen6:
@gshank Thanks for adding the test fix! I'm going to merge this in the hopes of sneaking it into the v0.20.0 beta going out next week—that way, we can start preparing changes in `dbt-utils` and other dispatch-heavy packages accordingly.

PR #3134, opened 2021-03-01
(explicitly) find the executable before running run_cmd

majidaldo:
resolves #3035

### Checklist
 - [x] I have signed the [CLA](https://docs.getdbt.com/docs/contributor-license-agreements)
 - [x] I have run this code in development and it appears to resolve the stated issue
 - [x] This PR includes tests, or tests are not required/relevant for this PR
   ran test_system_client.py
 - [x] I have updated the `CHANGELOG.md` and added information about my change to the "dbt next" section.

PR #3299, opened 2021-04-27
Use shutil.which to find executable path

### LINKED ISSUE #3035: git.bat not picked up on dbt deps
ref. https://github.com/fishtown-analytics/dbt/pull/1110

```
> dbt deps
Encountered an error:
Could not find command, ensure it is in the user's PATH: "git"
```

I have reasons to wrap my executables. On Windows I wrap my executables in `.bat` scripts (to get a similar effect to `.sh`). But then[ `Popen( ['git'] )`](https://github.com/fishtown-analytics/dbt/blob/42a85ac39f34b058678fd0c03ff8e8d2835d2808/core/dbt/clients/system.py#L419) doesn't work. `Popen( ['git.bat'])` works. 

So the platform-independent way to get `git` to work is to use  `shutil.which` to get the full path to the exe. `Popen([ which('git') ], )` works.

jtcohen6:
resolves #3035
closes #3134

### Description

To accommodate Windows users who wrap executables, use `shutil.which()` to find + append the `cmd` executable path.

### Checklist


jtcohen6:
@majidaldo I rearranged the logic so that dbt can still raise more helpful OS-specific errors. Could you now help me by:
- Testing that this change still resolves the issue for your local setup
- Advising on if there's a simple way to mock a wrapped executable, for the purposes of unit testing this functionality (I think [here](https://github.com/fishtown-analytics/dbt/blob/a260d4e25b9b1e27449b9b6a97ca121eca573176/test/unit/test_system_client.py#L52)). I want to make sure that this fix works for you in the long run, and the only way to do that is by adding a test case!

jtcohen6:
I received confirmation over in #3134 that this fix resolves the initially stated issue

kwigley (approved):
👍!

PR #3034, opened 2021-01-26
Parsing performance

gshank:
Resolves #2862 (partially -- there is still more performance work to come)

This work implements a number of changes to improve performance, including supporting libyaml, caching mapped_fields in the classes for 'from_dict', removing deepcopy on fqn_search, moving validation out of the 'from_dict' methods, amd special handling for dbt internal not_null and unique tests.  Uses a new TestMacroNamespace instead of original in order to limit the number of macros in the context.

In addition it replaces the Hologram 'from_dict' and 'to_dict ' calls with Mashumaro calls, which compiles those methods into each subclass. 

### Checklist
 - [x] I have signed the [CLA](https://docs.getdbt.com/docs/contributor-license-agreements)
 - [x] I have run this code in development and it appears to resolve the stated issue
 - [x] This PR includes tests, or tests are not required/relevant for this PR
 - [x] I have updated the `CHANGELOG.md` and added information about my change to the "dbt next" section.

PR #3302, opened 2021-04-27
Support descriptions on tests

### LINKED ISSUE #3249: Configurable description in yaml config for generic tests
### Describe the feature
Add the ability to define descriptions in yaml files. Support the ability to use the rendered values of the test args (`{{ model }}` and `{{ column_name }}` by default)

Consider the following example:
```
# macros/reusable_tests.sql
{% test not_null(model, column_name), adapter = 'default' %}

    select * from {{ model }}

{% endtest %}

# macros/schema.yml
tests:
 - name: not_null
   adapter: default
   description: "Assert that {{ column_name }} is not null in {{ model }}"
   args:
     - name: model
       description: "The model to test"
     - name: column_name
       description: "The column or expression to test"
```

Additional context in #1173.

kwigley:
resolves #3249 

### Description

This PR uses the same patching mechanism to apply updates to test nodes after parsing test properties. This currently supports descriptions for generic tests by applying the description to all nodes representing an instance of the test. When rendering the description for the test node, we expose the column name and model the test is running on in the context (`column_name` and `model`). This also supports doc blocks for both generic and bespoke test descriptions. yaml files can be defined in the test directory to achieve this.

### Checklist

### DIFF
--- a/core/dbt/contracts/graph/manifest.py
+++ b/core/dbt/contracts/graph/manifest.py
@@ -417,7 +417,7 @@ T = TypeVar('T', bound=GraphMemberNode)
 
 def _update_into(dest: MutableMapping[str, T], new_item: T):
     """Update dest to overwrite whatever is at dest[new_item.unique_id] with
-    new_itme. There must be an existing value to overwrite, and they two nodes
+    new_item. There must be an existing value to overwrite, and they two nodes
     must have the same original file path.
     """
     unique_id = new_item.unique_id
@@ -730,7 +730,7 @@ class Manifest(MacroMethods):
         # Q: could we save patches by node unique_ids instead, or convert
         # between names and node ids?
         for node in self.nodes.values():
-            patch = self.patches.pop(node.name, None)
+            patch = self.patches.pop(node.patch_lookup_key, None)
             if not patch:
                 continue
 
@@ -745,7 +745,6 @@ class Manifest(MacroMethods):
                     raise_invalid_patch(
                         node, patch.yaml_key, patch.original_file_path
                     )
-
             node.patch(patch)
 
         # If anything is left in self.patches, it means that the node for
--- a/core/dbt/contracts/graph/manifest.py
+++ b/core/dbt/contracts/graph/manifest.py
@@ -729,8 +729,12 @@ class Manifest(MacroMethods):
         # were ok with doing an O(n*m) search (one nodes scan per patch)
         # Q: could we save patches by node unique_ids instead, or convert
         # between names and node ids?
+        patch_keys = set(self.patches.keys())
+        used_patch_keys = set()
+
         for node in self.nodes.values():
-            patch = self.patches.pop(node.patch_lookup_key, None)
+            patch_lookup_key = node.patch_lookup_key
+            patch = self.patches.get(patch_lookup_key, None)
             if not patch:
                 continue
 
@@ -746,11 +750,15 @@ class Manifest(MacroMethods):
                         node, patch.yaml_key, patch.original_file_path
                     )
             node.patch(patch)
+            used_patch_keys.add(patch_lookup_key)
 
-        # If anything is left in self.patches, it means that the node for
+        unused_patch_keys = patch_keys - used_patch_keys
+
+        # if there are any unused patch names, it means that the node for
         # that patch wasn't found.
-        if self.patches:
-            for patch in self.patches.values():
+        if unused_patch_keys:
+            for patch_key in unused_patch_keys:
+                patch = self.patches[patch_key]
                 # since patches aren't nodes, we can't use the existing
                 # target_not_found warning
                 logger.debug((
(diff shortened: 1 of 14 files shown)


kwigley:
@cla-bot check

core/dbt/parser/manifest.py
```diff
@@ -749,12 +743,42 @@ def _process_docs_for_exposure(
...
+                target_model_name = model[0]
+            elif len(model) == 2:
+                _, target_model_name = model
+            else:
+                raise dbt.exceptions.InternalException(
+                    f'Refs and sources should always be 1 or 2 arguments - got {len(model)}'
+                )
+
+            ctx['column_name'] = node.column_name
+            ctx['model'] = target_model_name
```
kwigley, line 774: blah hardcoded values set directly on the context, I tried to avoid this without success

core/dbt/contracts/graph/manifest.py
```diff
@@ -729,8 +729,12 @@ def patch_nodes(self) -> None:
...
         # between names and node ids?
+        patch_keys = set(self.patches.keys())
+        used_patch_keys = set()
+
         for node in self.nodes.values():
-            patch = self.patches.pop(node.name, None)
+            patch_lookup_key = node.patch_lookup_key
+            patch = self.patches.get(patch_lookup_key, None)
             if not patch:
                 continue
```
kwigley, line 739: This is new/diverging behavior! Previously, a patch can only be applied once to a single node. The new behavior allows patches to be applied to multiple nodes to support applying the same patch to multiple instances of the same generic test.

core/dbt/contracts/graph/parsed.py
```diff
@@ -445,7 +449,7 @@ class ParsedMacro(UnparsedBaseNode, HasUniqueID):
     def local_vars(self):
         return {}
 
-    def patch(self, patch: ParsedMacroPatch):
+    def patch(self, patch: Union[ParsedMacroPatch]):
```
kwigley, line 452:
```suggestion
    def patch(self, patch: ParsedMacroPatch):
```

core/dbt/parser/manifest.py
jtcohen6: Ohh I see your comment above. Okay, that makes sense!

core/dbt/node_types.py
```diff
@@ -46,7 +46,8 @@ def documentable(cls) -> List['NodeType']:
             cls.Source,
             cls.Macro,
             cls.Analysis,
-            cls.Exposure
+            cls.Exposure,
+            cls.Test
```
jtcohen6, line 50: Confirmed that both types of tests (generic + bespoke) can use `doc` blocks in their descriptions, nice!

jtcohen6 (approved):
This is wonderful!

Some follow-on issues that should come from here:
- Per the [comment here](https://github.com/fishtown-analytics/dbt/issues/3249#issuecomment-832696074), we should including the test `description` (if available) in the info/stdout logging when a test fails.
- Let's update the builtin generic tests to have nice descriptions! I'd be happy to take a swing at this.

All in all, the work here LGTM.

gshank:
Something else to think about, if we only allow 'column_name' and 'model' in those descriptions, is not doing jinja rendering but simple a python string replace, i.e. replace {{ model}} with the model name (and the same for the column_name). It's not consistent with the way we handle other attributes, but it would be a lot faster.

jtcohen6:
@gshank I really appreciate you raising the performance concern here! It's tremendously justified. Let's figure out a way to achieve the functionality we want _without_ bloating `process_docs`.

I think we can take similar "shortcuts" for builtin `unique` and `not_null` test descriptions, similar to what we did in #3034: The descriptions will be set _within_ dbt, and they are (as you say) basically just python f-string operations. Conceivably, in the same way we "just set" configs for those tests, we could "just write" the test description.

All told, I don't see it as a common occurrence for users to edit generic test properties in development: most tests come from within dbt or within packages; they may define a handful of custom generic tests, but they're unlikely to have a footprint on the same scale as `unique` or `not_null`.

core/dbt/parser/manifest.py
```diff
@@ -749,12 +743,42 @@ def _process_docs_for_exposure(
...
-        ctx = generate_runtime_docs(
-            config,
-            node,
-            manifest,
-            config.project_name,
-        )
+        if isinstance(node, ParsedSchemaTestNode):
+            # Idealing we should be able to put the test kwargs in the context,
+            # but kwargs have already been processed at this point
+            # ex: 'model' --> "{{ ref('model') }}"
```
jtcohen6, line 749: Just to clarify my understanding: The `test_metadata` that lives on the test node, and contains `kwargs` as well, is _not_ Jinja-rendered. E.g:
```json
"test_metadata": {
    "name": "accepted_values",
    "kwargs": {
        "values": [
            "usd"
        ],
        "column_name": "currency_code",
        "model": "{{ ref('stg_ticket_tailor__orders') }}"
    },
    "namespace": null
},
```
In order to include `kwargs` in the test description, we'd need to upgrade the `description` renderer to use the _full_ Jinja rendering context (refs, macros, all that). Is that right?

Issue #3265, opened 2021-04-15
Show unexpected values when accepted_values test fails


brylie:
When an `accepted_values` test fails, print the _unexpected_ values.

I am currently writing a test against a column with many possible values, where we want to know if any new values appear in the column that might affect our analytics.

When the `accepted_values` test fails, it shows a message like:

> Got N results, expected 0.

However, it doesn't display information about the unexpected results

Showing the unexpected results would make it easier to investigate and modify the test.

I can help contribute to this feature.

jtcohen6:
@brylie What a cool idea! I think that the requisite pieces may just be falling into place, so that we could make this happen.

Back in #2593, I made a detailed proposal for how tests could store their (failing) results in database tables. To date, each schema test query has returned a single numeric value—`count(*)` or whatever else—so it would have been impossible to save and present an actual table of failing values. In order to make that happen, we've been working on creating a standard test "materialization" (#3181), and implementing it for ~schema~ generic tests as well (#3192). Those generic tests will now return a set of rows—whether that's simply rows from the table where a column is null, or rows from an aggregate query that rolls up counts by unacceptable values. 

All of which is to say, if we're doing all that and can save those results in a database table, in addition to grabbing the total failure count/value, we could also print a few of those results to stdout using good old [`results.print_table()`](https://docs.getdbt.com/reference/dbt-jinja-functions/run_query):

```bash
Failure in test accepted_values__orders__order_status (models/resources.yml)
  Found 7 unacceptable values of orders.order_status, erroring because >0.
  
    | VALUE_FIELD |      COUNT |
    | ----------- | ---------- |
    | blue        | 17,452,443 |
    | red         |  6,105,277 |
    | green       |  1,111,547 |
  
    For full results:
    -------------------------------------------------------------------------
    select * from dev_jcohen__dbt_test_results.accepted_values__orders__order_status
    -------------------------------------------------------------------------

  compiled SQL at target/compiled/models/resources.yml/schema_test/not_null__orders__order_status.sql
```

I think we'd put in limitations like:
- Print top 5 rows, and up to 5 columns
- An optional property of generic test blocks that defines `order_by` (in the case above, `order by 2 desc`)

Let me know what you think about the above, and let's see if we can sneak in this (or something like it) alongside the other test functionality coming in v0.20.0 :)

brylie:
Yeah, that seems like a good idea.

[...]

PR #3316, opened 2021-05-04
dbt test --store-failures

### LINKED ISSUE #517: compiled dbt test output could be more helpful
Consider compiling a simple `select` statement (w/ a `limit`) that can be directly run to find test failures. Right now, you need to change a `select count(*)` to `select *` and sometimes remove groups, add limits, etc.

### LINKED ISSUE #903: Record failing rows for tests into an auditable table
### Feature description
dbt should record rows that fail schema or data tests into tables, to be used for auditing. There are two general ways this could work:

1. insert into an `errors` table with a schema like:
 - run started at
 - invocation id
 - test name
 - failing row primary key / json

2. Create a table for each test failure in a scratch schema, matching the schema of the test query that failed

The baked-in schema tests will need to be rewritten to provide debuggable information in these tables.

This should probably be an opt-in feature for tests. It could be configured either in `schema.yml`, though it would also be good to configure whole swaths of tests at once.

### Who will this benefit?
dbt users could use this table to quickly determine why their tests failed. At present, it's surprisingly difficult to debug a failed test.

### LINKED ISSUE #2593: Storing test results in the database

jtcohen6:
Closes #517
Closes #903
Closes #2593

I also played around with #3265 (print the top 5 failures to stdout, and include in `run_results.json`). This work here is definitely the prerequisite for that, but we should tackle it separately if we decide it's something we want to do.

### Biggest choices
- This is all turned on via a flag, `--store-results`
- Hard-coded default schema for tests: `dbt_test__audit`. Normally, this means the schema is `my_target_schema_dbt_test__audit`; if someone is using `generate_schema_name_for_env`, it will be just `dbt_test__audit`.
    - Note: Because of the work in #3257, users can override this value by setting a `tests: +schema` config in `dbt_project.yml`. So I feel less bad about hard coding it.
- The original proposal in #2593 (and my first cut of this) would have dbt totally blow away the test-audit schema each time. As I said here previously: "This statelessness is actually _desirable_, but we should be immensely careful whenever dbt auto-drops anything." Well, that auto-dropping scared me a bit; now, dbt will drop-if-exists and recreate each test audit _table_, but not the whole schema. It's not an atomic operation, since that's not the point; it's not quite as dramatic as clobbering the whole schema, but it accomplishes our goal.

### Other comments
- I introduced a new node property, `relational()`, which is the same as `refable()` but includes tests IFF `flags.STORE_RESULTS` is turned on. dbt now uses `relational()` to decide which nodes inform the schemas it will create and when it will write the `relation_name` property to the manifest.
- I rewrote the queries defining a few of dbt's builtin generic tests, so that the failing records are more helpful/interesting.
- A table is not dropped if its test passes, i.e. if it contains zero rows. We could add that pretty easily, if it's something we want. Worth considering:
    - What if a test has a pass/fail condition _other_ than `count(*) = 0`? It's possible that a passing test may still have rows in the database that a user would want to inspect
    - Potential bloat to database metadata or block storage. (This isn't additive across invocations, since the test table is dropped and recreated on a subsequent store-failure test, but old/deleted tests would still hang around.)
- There's no centralized "results" table, as proposed in #2593, that contains one row for every test, its status info, and the location of its results table. The question is, do we need this? These days, all of that information is available (and more reliably accessed) from the `run_results.json` artifact.

```
Running with dbt=0.20.0-b1
Found 1 model, 3 tests, 0 snapshots, 0 analyses, 162 macros, 0 operations, 0 seed files, 1 source, 0 exposures

22:23:09 | Concurrency: 8 threads (target='bigquery')
22:23:09 |
22:23:09 | 1 of 3 START test my_custom_bespoke.................................. [RUN]
22:23:09 | 2 of 3 START test not_null_my_model_fun.............................. [RUN]
22:23:09 | 3 of 3 START test unique_my_model_fun................................ [RUN]
22:23:14 | 1 of 3 WARN 1 my_custom_bespoke...................................... [WARN 1 in 4.69s]
22:23:14 | 3 of 3 FAIL 1 unique_my_model_fun.................................... [FAIL 1 in 4.91s]
22:23:14 | 2 of 3 PASS not_null_my_model_fun.................................... [PASS in 4.98s]
22:23:14 |
22:23:14 | Finished running 3 tests in 10.98s.

Completed with 1 error and 1 warning:

Failure in test unique_my_model_fun (models/schema.yml)
  Got 1 result, expected 0.

  compiled SQL at target/run/my_project/models/schema.yml/schema_test/unique_my_model_fun.sql

  See test failures:
  --------------------------------------------------------------------------------
  select * from `dbt-test-env`.`dbt_jcohen___dbt_test_audit`.`unique_my_model_fun`
  --------------------------------------------------------------------------------

Warning in test my_custom_bespoke (tests/my_custom_bespoke.sql)
  Got 1 result, expected 0.

  compiled SQL at target/run/my_project/tests/my_custom_bespoke.sql

  See test failures:
  ------------------------------------------------------------------------------
  select * from `dbt-test-env`.`dbt_jcohen___dbt_test_audit`.`my_custom_bespoke`
  ------------------------------------------------------------------------------
```

### Checklist


core/dbt/include/global_project/macros/materializations/test.sql
```diff
@@ -1,10 +1,46 @@
...
+    {% set old_relation = adapter.get_relation(database=database, schema=schema, identifier=identifier) %}
+    {% set target_relation = api.Relation.create(
+        identifier=identifier, schema=schema, database=database, type='table') -%} %}
+    
+    {% if old_relation %}
+        {% do adapter.drop_relation(old_relation) %}
+    {% endif %}
+    
+    {% call statement(auto_begin=True) %}
+        {{ create_table_as(False, target_relation, sql) }}
```
drewbanin, line 17: this is 🔥

core/dbt/node_types.py
```diff
@@ -49,6 +50,10 @@ def documentable(cls) -> List['NodeType']:
             cls.Exposure
         ]
 
+    @classmethod
+    def relational(cls) -> List['NodeType']:
+        return cls.refable() + ([cls.Test] if flags.STORE_FAILURES else [])
```
drewbanin, line 55: this is.... not ideal... i think. Is there a way that we can push this logic around `flags.STORE_FAILURES` into the caller? Something about changing the behavior of this class method based on a global flag feels pretty fraught

core/dbt/parser/schema_test_builders.py
```diff
@@ -41,14 +41,25 @@ def get_nice_schema_test_name(
...
         label = unique
     else:
         label = hashlib.md5(unique.encode('utf-8')).hexdigest()
 
-    filename = '{}_{}_{}'.format(test_type, test_name, label)
-    name = '{}_{}_{}'.format(test_type, test_name, unique)
+    cutoff = 30
+    test_full_name = '{}_{}'.format(test_type, test_name)
+    if len(test_full_name) <= cutoff:
+        test_prenom = test_full_name
```
drewbanin, line 57: let's uh... use english.... for variable names :)

core/dbt/parser/schema_test_builders.py
```diff
@@ -231,6 +242,10 @@ def __init__(
         self.compiled_name: str = compiled_name
         self.fqn_name: str = fqn_name
 
+        # use hashed name as alias if too long
+        if compiled_name != fqn_name:
+            self.modifiers['alias'] = compiled_name
```
drewbanin, line 247: really clever & really cooll!

core/dbt/task/runnable.py
```diff
@@ -524,8 +524,6 @@ def create_schema(relation: BaseRelation) -> None:
 
                 db_schema = (db_lower, schema.lower())
                 if db_schema not in existing_schemas_lowered:
-                    existing_schemas_lowered.add(db_schema)
```
drewbanin, line 527: is there a reason why this disappeared?

drewbanin (commented):
Some comments dropped inline! Nice work @jtcohen6!

core/dbt/task/runnable.py
jtcohen6: nope :) thanks for catching it!

core/dbt/parser/schema_test_builders.py
jtcohen6: heh. I decided to reimplement this logic anyway, since it was getting a little confused

core/dbt/flags.py
```diff
@@ -64,11 +65,12 @@ def reset():
     PARTIAL_PARSE = False
     MP_CONTEXT = _get_context()
     USE_COLORS = True
+    STORE_FAILURES = False
```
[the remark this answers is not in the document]
jtcohen6: definitely!

core/dbt/node_types.py
jtcohen6: Good call. I removed `relational` as a class method and instead implemented this logic as a [`should_store_failures` boolean property of parsed nodes](https://github.com/fishtown-analytics/dbt/pull/3316/files#diff-2865acf9f8c91b43775b17533e2f5ac56f9f6a7bf2fff497e0bc84edaf4af8daR118-R123).

Unfortunately, it's duplicative with logic in the [`should_store_failures()` Jinja macro](https://github.com/fishtown-analytics/dbt/pull/3316/files#diff-4bef0f1be924c18da1624194a38efd20176db97cd0d0bfa6d68f460c7f9e7848R77-R83) used in the `test` materialization, which takes after the `should_full_refresh()` macro. I don't know a good way around this: We need this property accessible in python, for deciding which schemas need to be created as part of the task. I also see merit in making the logic really explicit to users, and allowing them to override it if they'd like. And the places in the codebase where we call Jinja macros from python are... not ideal...

jtcohen6:
Updated:
- Tests accept a `store_failures` config which can be set in all the usual places. This works like `full_refresh`: If set, the config takes precedence over the `--store-failures` flag, so a test either _always_ or _never_ stores its failures. If the config is not set, then the flag takes effect.
- Cleaned up the test aliasing logic.

Known limitations of the current approach:
- Duplicated logic between `should_full_refresh` node property and `should_full_refresh()` Jinja macro (see [comment above](https://github.com/fishtown-analytics/dbt/pull/3316#discussion_r633113784))
- Generic test nodes with too-long names (64+ characters) don't use custom `alias` names or respect the `generate_alias_name()` macro. Instead, they use an alias guaranteed to be both short enough (<64 char) and unique (contains hash). Users _can_ specify `alias`/`schema`/`database` configs for those tests from `dbt_project.yml`, but not in the `.yml` files that specify them. (Should we accept these as `MODIFIER_ARGS`? I'm willing to punt on this for now)

kwigley (approved):
❤️ this is awesome! looks good to me

sphinks:
Great feature! Any idea when it will be released?

jtcohen6:
@sphinks This will be in v0.20.0, for which we'll have a first release candidate very shortly

jranks123:
Love this, thanks everyone <3

jranks123:
@jtcohen6 This seems to imply it only stores failures - are there any plans to also store non-failures, i.e passes?

jtcohen6:
@jranks123 The test table will store whatever the test query returns. In general, test queries pass when they return zero rows—so the table will still be created, it will just have nothing in it.

In v0.20.0, tests can also have a few new configs—`fail_calc` and `error_if`—such that it's possible to return rows from a test query and still pass, so long as the `fail_calc` does not meet the `error_if` criterion. In that case, `--store-failures` would still cause the test to store the results of its query in a table, even though the test does not ultimately raise an error.

PR #3327, opened 2021-05-08
Separate compiled_path in manifest + printer

### LINKED ISSUE #1985: Database Error messages include `run` sql path instead of `compiled` SQL path
### Describe the bug
Database Error messages include `run` sql path instead of `compiled` SQL path.

### Steps to reproduce
* Run a model with bad SQL

### Expected results
* Database Error returned with a path to the `compiled` file (this was the case for v0.14.4)
```
Database Error in model test_date_trunc (models/cross_db_utils/test_date_trunc.sql)
  000904 (42000): 01908e37-0358-2460-0000-0e690422b63a: SQL compilation error: error line 11 at position 22
  invalid identifier 'UPDATED_AT'
  compiled SQL at target/compiled/dbt_utils_integration_tests/cross_db_utils/test_date_trunc.sql
```
(note: `target/compiled`)

### Actual results
* Database Error returned with a path to the `run` fil
```
Database Error in model test_date_trunc (models/cross_db_utils/test_date_trunc.sql)
  000904 (42000): SQL compilation error: error line 11 at position 22
  invalid identifier 'UPDATED_AT'
  compiled SQL at target/run/dbt_utils_integration_tests/cross_db_utils/test_date_trunc.sql
```
(note: `target/run`)
If applicable, add screenshots or log output to help explain your problem.

v0.15.0

### Additional context
Reason I don’t love it: I frequently copy the filepath of the failing SQL, type atom [cmd + v] to open the file, copy it into a query runner and try to run it. I in fact _don’t_ want the DDL when I copy it.

jtcohen6:
resolves #1985

### Description

Adds a new node property, `compiled_path`. Previously, dbt would only use one property, `build_path`, which would represent the `target/compiled` path for compile-only tasks (`compile`, `docs generate`) and the `target/run` path for execute tasks (`run`, `test`, etc). Given that these paths represent truly different things, I think it makes sense to have them be separate!

The immediate motivation is fixing a regression from many versions ago that caused dbt to print the "Run SQL" file path, rather than the "Compiled SQL" file path, as context for model database errors and test failures.

### Checklist

### DIFF
--- a/core/dbt/compilation.py
+++ b/core/dbt/compilation.py
@@ -442,7 +442,7 @@ class Compiler:
         logger.debug(f'Writing injected SQL for node "{node.unique_id}"')
 
         if node.compiled_sql:
-            node.build_path = node.write_node(
+            node.compiled_path = node.write_node(
                 self.config.target_path,
                 'compiled',
                 node.compiled_sql
--- a/core/dbt/contracts/graph/parsed.py
+++ b/core/dbt/contracts/graph/parsed.py
@@ -179,6 +179,7 @@ class ParsedNodeDefaults(ParsedNodeMandatory):
     meta: Dict[str, Any] = field(default_factory=dict)
     docs: Docs = field(default_factory=Docs)
     patch_path: Optional[str] = None
+    compiled_path: Optional[str] = None
     build_path: Optional[str] = None
     deferred: bool = False
     unrendered_config: Dict[str, Any] = field(default_factory=dict)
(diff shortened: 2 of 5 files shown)


gshank:
As far as the manifest schema, I've also added (for partial parsing) a 'created_at' field to nodes. I've been meaning to raise the issue of the schema...

kwigley:
🤔 I could be convinced either way whether to bump the schema version for adding new fields, in these case we are sort of changing what a field means and also adding a new field. I'm leaning towards keeping same version and publishing the new schema with the new field and same version 🤷

kwigley (approved):
We can follow up with schema changes as part of releasing 0.20.0!

jtcohen6:
Silly me! I just merged this in without a changelog entry. I'll follow up with that right now.

PR #3335, opened 2021-05-10
Feature/schema tests are more unique

### LINKED ISSUE #3254: Better unique_id construction for specific instances of generic tests
### Describe the feature

(Yes, I know I need to come up with a better phrase than "specific instances of generic tests")

Users will know that they cannot:
- Have two bespoke tests named the same
- Have two generic tests named the same
- Have a bespoke test and a generic test named the same (!)

But they should _not_ be responsible for worrying about namespace clashes between specific instances of generic tests!

Over in #2308 (which I'm closing in favor of this new issue), we had a great example of this. A `not_null` test on `base.extension_id` and `base_extension.id` both get the same:
- unique_id: `test.my_package.not_null_base_extension_id`
- fqn: `test.schema_test.not_null_base_extension_id`

Only one of these tests can be selected, executed, seen in `manifest.json`, etc.

### Q & A

_Do they need to be globally unique?_ No, I think we'd rather the `unique_id` is consistent from run to run. It should be created from the components of the test definition / instantiation.
_Do they need to be human-readable?_ I don't think this is *so* important. We already hash the file names of schema tests with too many arguments. In a world where generic tests produce really good **descriptions** (#3249) and **failure messages**, the node name / unique_id feels less important: But we better do a good job of printing those descriptions out to stdout / artifacts!

### Potential solutions

Components:
- test name (e.g. `not_null`, `unique`)
- `model`
- `column_name` (if available)
- Other non-config arguments., e.g. `values` for `accepted_values`, `to` + `field` for `relationships`
    - Why not configs? I don't think changing the severity of a test should change the `unique_id` of that test node. That has the effect of making it seem, to dbt, that you've removed and added a genuinely different test. (I do think that's the effect of changing the `values` in `accepted_values`, however.) If this distinction gives us too much trouble, however, we should throw it away and figure out something simpler.

Options:

- Concatenate using a character uncommon in SQL databases (_not_ `_`), such as `.` or `-`. Unfortunately, those characters are also unwieldy in python. How wild would it be to use URL encoding for this?
- Hash together the concatenation of the test components
- Some combination: `not_null__base__extension_id__0a9c45d3f7560da45492b85afa1f41b5`, i.e. just enough to be readable plus `md5(concat('not_null','-','model=model.my_package.base','-','column_name=extension_id'))`

iknox-fa:
resolves #3254

### Description
Adds a 10 character hash to the unique_id of all schema (generic) tests.  This ensures that if a test differs only in it's hash components, it will still be run instead of clobbered by the last test defined.

~There's a few bits I'd like feedback about from the team.  I tagged them with TODOs, but I'll put a comment in the PR as well.~ <- complete.

### Checklist


core/dbt/exceptions.py
```diff
@@ -772,6 +772,10 @@ def raise_duplicate_resource_name(node_1, node_2):
     elif node_1.resource_type == NodeType.Documentation:
         get_func = 'doc("{}")'.format(duped_name)
     elif node_1.resource_type == NodeType.Test and 'schema' in node_1.tags:
+        # TODO: This causes duplicate schema test errors to be ignored!!
+        # This PR should ensure this never gets called when the tests are actually different,
+        # but removing this breaks 50+ tests that currently have the exact same schema
+        # test defined more than once. :/
```
iknox-fa, line 778: Anyone else know why this is the way it is?  Git blame isn't helping much.

kwigley: I don't have anything to add to _why_ this exists, but wanted to shine light on something I noticed.

The weird part is that when you remove that short circuiting logic, **source** tests defined once raise a duplicate resource error. At first glance, it looks like we add source tests twice:

https://github.com/fishtown-analytics/dbt/blob/8d39ef16b67fcc3c30e8a18751b4e64831ec9aaf/core/dbt/parser/sources.py#L128-L132

https://github.com/fishtown-analytics/dbt/blob/8d39ef16b67fcc3c30e8a18751b4e64831ec9aaf/core/dbt/parser/schemas.py#L530-L533

I think our tests are doing their job in this case! We should reconcile the duplicate logic adding schema tests on sources, I think this is residual code from swapping parse results with directly using the manifest (#3219)

gshank: That's pretty strange and not something I've looked at before. As far as I know, it's not related to using the manifest directly, because the sources and nodes were just copied into the manifest and then things proceeded and the copy step just vanished. But I could be wrong.

iknox-fa: > Do you remember which ones in particular were breaking?
@jtcohen6 I do not, but it's easy enough to generate one.  I remember thinking at the time I discovered it that, whatever is going on there was too large to include as part of this ticket.

@kwigley Ohh interesting about the source tests!  I'll do some digging and see what's happening there.  It's possible those are the test fails I was seeing.

core/dbt/parser/schemas.py
```diff
@@ -351,6 +352,30 @@ def create_test_node(
...
     ) -> ParsedSchemaTestNode:
 
+        # TODO: (PR input needed) Should this be applied to all tests?
+        # In theory we already check non-schema tests for "unique" name uniqueness in
+        # core/dbt/contracts/graph/manifest.py @ _check_duplicates()
+
+        HASH_LENGTH = 10
+
+        if 'schema' in tags:
+            # TODO: (PR input needed) is this a complete set of uniquie-ifying data?
```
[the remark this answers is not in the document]
iknox-fa: > What do you think about just hashing together everything in test_metadata.kwargs? 
@jtcohen6 Hashing the entire test_metadata is a little tricky because there's no guarantee of order.  We can enforce an order by doing a recursive sorting/frozendict/whatever, but it _might_ be a bit of a hit performance wise.  It might not.. I have no idea how many times we'd be likely to call this in any given run, nor am I aware of the full extent of the size that `test_metadata` can reach.  I'm happy to try it out though.

> Those are both valid tests, but the only way they look different to this `unique_id` construction is in their name
From an operational point of view, I think that's fine.  as long as `unique_id` is _actually_ unique. Then everything should "just work", although I agree it's not a very helpful / human-friendly.

core/dbt/exceptions.py
kwigley: > > Do you remember which ones in particular were breaking?
> > @jtcohen6 I do not, but it's easy enough to generate one.  I remember thinking at the time I discovered it that, whatever is going on there was too large to include as part of this ticket.
> 
> @kwigley Ohh interesting about the source tests! I'll do some digging and see what's happening there. It's possible those are the test fails I was seeing.

Awesome! I'm going to mark the review as "Request Changes" just so I know when to take another pass if you come back with anything new after digging in 👍

iknox-fa: Digging in it appears that @kwigley nailed it.  All the errors were due to source tests being added to the manifest x2.
It seems out of scope here, so I made a [bug ticket](https://github.com/fishtown-analytics/dbt/issues/3347) for it.

core/dbt/parser/schemas.py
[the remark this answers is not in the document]
iknox-fa: I went ahead and implemented a quick semantic hashing algo (AKA sort-it-then-hash-it) so we can include the entire test_metadata dict.

core/dbt/parser/schemas.py
```diff
@@ -351,6 +352,30 @@ def create_test_node(
         column_name: Optional[str],
     ) -> ParsedSchemaTestNode:
 
+        # TODO: (PR input needed) Should this be applied to all tests?
+        # In theory we already check non-schema tests for "unique" name uniqueness in
+        # core/dbt/contracts/graph/manifest.py @ _check_duplicates()
+
+        HASH_LENGTH = 10
+
+        if 'schema' in tags:
```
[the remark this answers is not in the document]
iknox-fa: Updated

core/dbt/parser/schemas.py
```diff
@@ -351,6 +352,23 @@ def create_test_node(
         column_name: Optional[str],
     ) -> ParsedSchemaTestNode:
 
+        HASH_LENGTH = 10
+
+        def get_hashable_md(
+            data: Union[str, int, float, List, Dict]
+        ) -> Union[str, List, Dict]:
+            if type(data) == dict:
+                return {k: get_hashable_md(data[k]) for k in sorted(data.keys())}  # type: ignore
```
nathaniel-may, line 361: What's the logic behind sorting here? We're returning a dictionary which will be unordered.

core/dbt/parser/schemas.py
```diff
@@ -351,6 +352,23 @@ def create_test_node(
         column_name: Optional[str],
     ) -> ParsedSchemaTestNode:
 
+        HASH_LENGTH = 10
+
+        def get_hashable_md(
```
nathaniel-may, line 357: I was hoping there would be a way to do this hash in a type-agnostic way instead, but I couldn't figure one out. Womp womp.

core/dbt/parser/schemas.py
```diff
@@ -351,6 +352,23 @@ def create_test_node(
...
+        ) -> Union[str, List, Dict]:
+            if type(data) == dict:
+                return {k: get_hashable_md(data[k]) for k in sorted(data.keys())}  # type: ignore
+            elif type(data) == list:
+                return [get_hashable_md(val) for val in data]  # type: ignore
+            else:
+                return str(data)
+
+        hashable_metadata = repr(get_hashable_md(test_metadata))
+        hash_list = filter(None, [name, hashable_metadata])
```
nathaniel-may, line 368: I don't think I understand this line. Using None for our filter function means it's using the identity function, so either these values are booleans, or no filtering is occurring here.

core/dbt/parser/schemas.py
nathaniel-may: On another note it would be nice to have a comment on this method.

core/dbt/parser/schemas.py
```diff
@@ -351,6 +352,23 @@ def create_test_node(
...
+        hash_string = ''.join(hash_list).encode('utf-8')
+        test_hash = md5(hash_string).hexdigest()[-HASH_LENGTH:]
```
nathaniel-may, line 370: It looks like we could use `hexdigest(length)` and have the library choose the value instead of pulling from the end manually.

test/integration/029_docs_generate_tests/test_docs_generate.py
```diff
@@ -1333,7 +1333,7 @@ def expected_seeded_manifest(self, model_database=None, quote_model=False):
                     'checksum': self._checksum_file(seed_path),
                     'unrendered_config': unrendered_seed_config,
                 },
-                'test.test.not_null_model_id': {
+                'test.test.not_null_model_id.8ab00aacfe': {
```
nathaniel-may, line 1336: Are we going to have to manually write down these hashes for all these tests in the future? Not sure how this test works.

nathaniel-may (changes requested):
Some relatively minor points I'd like to discuss before merging. Otherwise it looks pretty good.

core/dbt/parser/schemas.py
iknox-fa: Ah, good catch.  It's a hold-over from a previous iteration where we picked specific items from test_metadata to hash instead of hashing the whole thing.  It was basically my way of DRYing up all the "dummy_value"s in a list of this:
```
hash_list.append(test_metadata.get("thing_1", "dummy_value"))
hash_list.append(test_metadata.get("thing_2", "dummy_value"))
hash_list.append(test_metadata.get("thing_3", "dummy_value"))
hash_list.append(test_metadata.get("thing_4", "dummy_value"))
```

I don't _think_ `name` or `hashable_metadata` could ever be `None` (test running right now to confirm)

core/dbt/parser/schemas.py
iknox-fa: Yeah, there's not a lot of great options that are performant.  I'll add a comment to make clear what's happening.

core/dbt/parser/schemas.py
iknox-fa: That's only for hashers that support variable length hashes like BLAKE2. 
```
>>> hashlib.md5(b"foobar").hexdigest(5)
  File "<stdin>", line 1, in <module>
TypeError: HASH.hexdigest() takes no arguments (1 given)
```

core/dbt/parser/schemas.py
iknox-fa: Dicts in python are [**insertion ordered** since 3.6](https://docs.python.org/3.6/whatsnew/3.6.html#new-dict-implementation) (that's why we're ok with returning a dict as a final output of the function).  That said, I didn't have much confidence that the order would be consistent given it's generated... somewhere else.  It would also be an easy thing to break because of that so I added an extra sort.

test/integration/029_docs_generate_tests/test_docs_generate.py
iknox-fa: Probably for the time being we will.  IMO this add-a-hash solution needs to be replaced with either:
- something human readable
- something better instrumented so we don't have to manage these in tests/etc.

Either way, I think it'll be a larger project to re-work the way node-naming works so this feels stop-gap-ish.

core/dbt/parser/schemas.py
iknox-fa: This might make it clearer:
```
> /Users/iknox/.pyenv/versions/dbt-dev/lib/python3.8/site-packages/dbt/parser/schemas.py(381)create_test_node()
(Pdb) dct_1 = {'c':'C', 'a':'A', 'b':'B'}
(Pdb) dct_2 = {'b':'B', 'a':'A', 'c':'C'}
(Pdb) (get_hashable_md(dct_1), get_hashable_md(dct_2))
({'a': 'A', 'b': 'B', 'c': 'C'}, {'a': 'A', 'b': 'B', 'c': 'C'})
(Pdb) (get_hashable_md_no_keysort(dct_1), get_hashable_md_no_keysort(dct_2))
({'c': 'C', 'a': 'A', 'b': 'B'}, {'b': 'B', 'a': 'A', 'c': 'C'})
```

core/dbt/parser/schemas.py
iknox-fa: Confirmed and updated.

core/dbt/parser/schemas.py
iknox-fa: Added and updated.

core/dbt/parser/schemas.py
nathaniel-may: I didn't know that! Good fact to file away for later. And because we're using it as a string later, the order matters here. It does mean that if they swap dict implementations in a future version of python this might break on a version upgrade. But I'm not actually that worried about this.

test/integration/029_docs_generate_tests/test_docs_generate.py
nathaniel-may: Maybe we should file a tech debt ticket and reference this PR so we don't forget.

nathaniel-may (approved):

core/dbt/parser/schemas.py
iknox-fa: Yeah, it's been this way since 3.6 and documented as a language feature since 3.7-- I don't see it disappearing any time soon.

test/integration/029_docs_generate_tests/test_docs_generate.py
iknox-fa: [Doneski](https://github.com/fishtown-analytics/dbt/issues/3348)

gshank (approved):
Lots of good comments already! Looks good.

kwigley (approved):
lgtm 👍

jtcohen6 (approved):

Issue #3321, opened 2021-05-05
User-supplied fail_calc for tests


jtcohen6:
### Describe the feature

Today, the `'test'` materialization hard-codes `count(*)` as the way to calculate failures from a test:

https://github.com/fishtown-analytics/dbt/blob/26fb58bd1b08218781b182918f5ed4dec3f735d9/core/dbt/include/global_project/macros/materializations/test.sql#L4-L7

I think that makes sense 90% of the time in the general case, but we want to give users the ability to customize the "failure calculation" if they'd like it to be something other than `count(*)`. This is important for backward compatibility, since schema tests could previously calculate and return whatever numeric value they wanted. In the wild, this could be as simple as `sum(column)` instead of `count(*)`, or it could be as complex as the [`dbt_utils.equality`](https://github.com/fishtown-analytics/dbt-utils/blob/489a44bb7ebad1a6d9ba762fdcd7dc6a9fea7e0d/macros/schema_tests/equality.sql#L68-L73) test:

```sql
select count(*) from unioned) +
        (select abs(
            (select count(*) from a_minus_b) -
            (select count(*) from b_minus_a)
            )
```

I'm hopeful this is quite straightforward to implement—it's just a matter of pulling in the `fail_calc` and templating it into the materialization.

### Questions

- Should `fail_calc` be a test _config_ or a test _property_? I lean toward _property_, since I think this is an essential component of the test definition and less like something that wants to be set for many different types of tests at once, e.g. from `dbt_project.yml`. (In a post-#2401 world, this is hopefully a less meaningful distinction!)
- Could this have potentially strange interactions with `%` values of `warn_if` / `error_if` (#3258)? Yes! I don't think we need to solve for every edge case there now.

jtcohen6:
Update: We're going to make this a _test config_, following the pattern sketched out in #3258.

I don't think it makes a _ton_ of sense for users to override the default configs set by the generic test definition, but hey, they'll be able to if they want to. I could even see this being compelling—let's say you want the `unique` test to calculate failure as _the number of original rows containing a duplicate_, rather than _the length of the set of duplicate values_.

jtcohen6:
Resolved by #3392

PR #3336, opened 2021-05-11
New test configs: where, limit, warn_if, error_if, fail_calc

### LINKED ISSUE #3258: Net-new test configs
### Describe the feature

Let's add configs for:
- `where`
- `limit`
- `warn_if`, `error_if` (picking up from https://github.com/fishtown-analytics/dbt/issues/2219 + https://github.com/fishtown-analytics/dbt/issues/2982)

Following the work in #3257, these will be configurable from `config()` + `dbt_project.yml` + `.yml` instantiation (for generic tests).

### Mechanism

1. Add to TestConfig:
https://github.com/fishtown-analytics/dbt/blob/7203123d56e42007d569a03c76bea0f351fb1611/core/dbt/contracts/graph/model_config.py#L437-L440

2. Add to schema test builder (along with methods lower down):
https://github.com/fishtown-analytics/dbt/blob/7203123d56e42007d569a03c76bea0f351fb1611/core/dbt/parser/schema_test_builders.py#L187-L188

3. Add to shortcut method for "fast" schema tests:
https://github.com/fishtown-analytics/dbt/blob/7203123d56e42007d569a03c76bea0f351fb1611/core/dbt/parser/schemas.py#L490-L491

jtcohen6:
resolves #3258, #3321

### Description

```yml
tests:
  +warn_if: ">10"
  +error_if: ">100"
  +where: "date_col = current_date"
  +limit: 10
  +fail_calc: "count(*)"
```

Adds ~four~ five new test configs:
- `limit`: Simple enough, templated out in the materialization. This is mostly a complement for #3316
- `where`: The way I did this was _unbelievably_ hacky—see the test cases for yourself—but _it works_, and in a way that should be backwards compatible with all existing schema/generic tests, without them needing to change any part of their SQL definition. (This config doesn't make sense for one-off tests.)
- `warn_if`, `error_if`: The user supplies a python-evaluable string (e.g. `>=3`, `<5`, `==0`, `!=0`) and dbt will compare the fail count/calc against it.
    - By default, set to `>0`. (I realize now this should probably be `!=0`)
    - Interaction with `severity`: A little tricky, but I actually think they work reasonably together. By default, `severity: error`, and dbt checks the `error_if` condition first; if not error, then check the `warn_if` condition; if the result meets neither, it passes the test. If the user sets `severity: warn`, dbt will skip over the `error_if` condition entirely and jump straight to `warn_if`.
- `fail_calc`: https://github.com/fishtown-analytics/dbt/issues/3321

### TODO
- Add integration tests
- Cleanup: Could we wrap the `where` logic in a special internal macro, so long as it's included in the schema test parsing context?
- Cleanup: Any way to DRY up the manual-set configs within `unique` + `not_null` shortcut?

### For the future

I got stuck trying to implement relative/percentage `warn_if` and `error_if`. That's ok! That piece, while compelling, can very easily come in a later issue.
- If the user supplies `error_if: >5%`, it's easy enough to `error_if.replace('%', '/100')` and compare against `>5/100` instead.
- The issue was, I couldn't figure out how to grab the denominator, i.e. the count of the "full model" (ideally with the `where` condition included). At materialization time, we can access `test_metadata.kwargs.model`, but this is an _unrendered_ Jinja string, and even with the `as_native` filter, I couldn't figure out how to "extra-render" it.

### Checklist


core/dbt/parser/schema_test_builders.py
```diff
@@ -278,6 +280,18 @@ def severity(self) -> Optional[str]:
         else:
             return None
 
+    def where(self) -> Optional[str]:
+        return self.modifiers.get('where')
+
+    def limit(self) -> Optional[str]:
```
jtcohen6, line 286:
```suggestion
    def limit(self) -> Optional[int]:
```

jtcohen6:
Also need to update:

https://github.com/fishtown-analytics/dbt/blob/85e2c897945e6372fe3657075b2c5ea4b068d3e0/core/dbt/contracts/graph/compiled.py#L122-L126

core/dbt/parser/schemas.py
```diff
@@ -488,11 +488,28 @@ def render_test_update(self, node, config, builder):
         if (macro_unique_id in
                 ['macro.dbt.test_not_null', 'macro.dbt.test_unique']):
             self.update_parsed_node(node, config)
+            # manually set configs
             if builder.severity() is not None:
                 node.unrendered_config['severity'] = builder.severity()
                 node.config['severity'] = builder.severity()
             if builder.enabled() is not None:
+                node.unrendered_config['severity'] = builder.severity()
```
jtcohen6, line 496:
```suggestion
                node.unrendered_config['enabled'] = builder.enabled()
```

core/dbt/parser/schema_test_builders.py
```diff
@@ -185,7 +185,9 @@ class TestBuilder(Generic[Testable]):
         r'(?P<test_name>([a-zA-Z_][0-9a-zA-Z_]*))'
     )
     # kwargs representing test configs
-    MODIFIER_ARGS = ('severity', 'tags', 'enabled')
+    MODIFIER_ARGS = (
+        'severity', 'tags', 'enabled', 'where', 'limit', 'warn_if', 'error_if'
```
jtcohen6, line 189:
```suggestion
        'severity', 'tags', 'enabled', 'where', 'limit', 'warn_if', 'error_if', 'fail_calc'
```

core/dbt/task/test.py
```diff
@@ -79,37 +92,52 @@ def execute_test(
...
             )
-        return int(table[0][0])
+
+        test_result_data: PrimitiveDict = dict(zip(table.column_names, table.rows[0]))
+
+        return TestResult.from_dict(test_result_data)
 
     def execute(self, test: CompiledTestNode, manifest: Manifest):
-        failed_rows = self.execute_test(test, manifest)
+        result = self.execute_test(test, manifest)
```
jtcohen6, line 113: Do we need to pass through the `error_if` and `warn_if` strings as columns in the result? Or could we access them below as properties of the test node? e.g.
```
message = f'Got {num_errors}, configured to fail if {test.config.error_if}'
message = f'Got {num_errors}, configured to warn if {test.config.warn_if}'
```

core/dbt/contracts/graph/compiled.py
```diff
@@ -122,7 +122,17 @@ class CompiledSchemaTestNode(CompiledNode, HasTestMetadata):
...
+            self.unrendered_config.get('where') ==
+            other.unrendered_config.get('where') and
+            self.unrendered_config.get('limit') ==
+            other.unrendered_config.get('limit') and
+            self.unrendered_config.get('fail_calc') ==
+            other.unrendered_config.get('fail_calc') and
+            self.unrendered_config.get('error_if') ==
+            other.unrendered_config.get('error_if') and
+            self.unrendered_config.get('warn_if') ==
+            other.unrendered_config.get('warn_if')
```
jtcohen6, line 135: Here's what I was doing over in #3316:
```python
def same_config(self, other) -> bool:
        comparisons = [
            self.unrendered_config.get(modifier) == other.unrendered_config.get(modifier) or (
                self.unrendered_config.get(modifier) is None and
                other.unrendered_config.get(modifier) is None
            )
            for modifier in ('severity', 'store_failures', ...)
        ]
        return all(comparisons)
```
I'm not sure if that's better or worse. Might not work at all. Checking for `None` does seem important though

test/integration/008_schema_tests_test/test_schema_v2_tests.py
```diff
@@ -414,8 +412,8 @@ def test_postgres_collision_test_names_get_hash(self):
 
         # both tests have the same unique id except for the hash
         expected_unique_ids = [
-            'test.test.not_null_base_extension_id.2dbb9627b6',
-            'test.test.not_null_base_extension_id.d70fc39f40'
-            ]
+            'test.test.not_null_base_extension_id.4a9d96018d',
+            'test.test.not_null_base_extension_id.60bbea9027'
```
jtcohen6, line 416: Do you know why the `unique_id`s have changed? We shouldn't be including any configs in the hash, right?

core/dbt/task/test.py
```diff
@@ -79,37 +92,52 @@ def execute_test(
...
+        num_cols = len(table.columns)
+        if num_cols != 5:
+            raise InternalException(
+                f"dbt internally failed to execute {test.unique_id}: "
+                f"Returned {num_cols} columns, but expected "
+                f"5 columns"
             )
-        return int(table[0][0])
+
+        test_result_data: PrimitiveDict = dict(zip(table.column_names, table.rows[0]))
```
jtcohen6, line 108: Seeing this in the Snowflake test:
```
Unhandled error while executing test.test.unique_view_model_id.d4e7c60bfe
Field "validation_errors" of type builtins.int is missing in dbt.task.test.TestResult instance
```
I'm guessing this is because Snowflake returns all uppercase identifiers, so the key is `VALIDATION_ERRORS`. I'm sure there's a way to handle that

core/dbt/task/test.py
```diff
@@ -79,37 +92,52 @@ def execute_test(
...
 
         return RunResult(
             node=test,
             status=status,
             timing=[],
             thread_id=thread_id,
             execution_time=0,
-            message=int(failed_rows),
-            adapter_response={}
+            message=message,
```
jtcohen6, line 139: I am _all for_ making the message an actual message. There are some places in our test suite that expect `message` to be an integer. Could we add a new property, `failures`, that can be that integer value? This would also represent a breaking artifact change (`run_results.json`) that we should communicate

It _could be_ the same value as in `adapter_response.rows_affected`, but I'd rather have this be a first-class test property, and a thing that's consistent across adapters.

jtcohen6 (commented):
This is looking really good! I haven't had the chance for an exhaustive review, so I just dropped some quick comments where things are standing out to me.

core/dbt/contracts/graph/compiled.py
kwigley: ya! was planning on going back to do this, love it.

test/integration/008_schema_tests_test/test_schema_v2_tests.py
kwigley: The model kwarg changed with the `where` config. The hash includes kwargs.

core/dbt/task/printer.py
```diff
@@ -128,11 +128,11 @@ def print_test_result_line(
         color = ui.green
         logger_fn = logger.info
     elif result.status == TestStatus.Warn:
-        info = 'WARN {}'.format(result.message)
+        info = 'WARN'
```
jtcohen6, line 131: I think parity with current would be something like `'WARN {}'.format(result.failures)`. Do you have a reason to prefer a clean `WARN` / `FAIL` info line? If so, I'm open to it!

jtcohen6 (commented):
Looking good! And looks like a lot of failing tests that remain are checking `result.message` against an integer, so just need to switch over to `result.failures`

core/dbt/task/printer.py
kwigley: Adding a new field to run results for one type of task is going to turn into an entire refactor of how we handle task results. I'm shelving this for this issue.

jtcohen6: That's a good enough reason for me

core/dbt/parser/schema_test_builders.py
```diff
@@ -334,10 +351,13 @@ def build_raw_sql(self) -> str:
         )
```
jtcohen6, line 351: We need to reorder `"{config}{{{{ {macro}(**{kwargs_name}) }}}}"` to be `{{{{ {macro}(**{kwargs_name}) }}}}{config}`, so that specific configs (supplied by the user in the modifiers) override the generic configs set in the test macro.
```python
  # this is the 'raw_sql' that's used in 'render_update' and execution
  # of the test macro
  # config needs to be rendered last to take precedence over default configs
  # set in the macro (test definition)
  def build_raw_sql(self) -> str:
      return (
          "{{{{ {macro}(**{kwargs_name}) }}}}{config}"
      ).format(
          macro=self.macro_name(),
          config=self.construct_config(),
          kwargs_name=SCHEMA_TEST_KWARGS_NAME,
      )
```

core/dbt/contracts/graph/compiled.py
```diff
@@ -119,12 +119,7 @@ class CompiledSchemaTestNode(CompiledNode, HasTestMetadata):
     column_name: Optional[str] = None
     config: TestConfig = field(default_factory=TestConfig)
 
-    def same_config(self, other) -> bool:
-        return (
-            self.unrendered_config.get('severity') ==
-            other.unrendered_config.get('severity')
-        )
-
+    # TODO: this is unused
```
jtcohen6, line 122: Yes... indeed it is. I confirmed that `dbt test -m state:modified` works as expected, leveraging the `unrendered_config`.

core/dbt/contracts/graph/model_config.py
```diff
@@ -438,6 +438,34 @@ class SeedConfig(NodeConfig):
...
     materialized: str = 'test'
     severity: Severity = Severity('ERROR')
+    where: Optional[str] = None
+    limit: Optional[int] = None
+    fail_calc: str = "count(*)"
+    warn_if: str = "!= 0"
+    error_if: str = "!= 0"
+
+    @classmethod
+    def same_contents(
```
jtcohen6, line 448: Ah! I see now that this is the proper place for it. Nice work, this logic is much clearer.

jtcohen6 (commented):
Took this for a ride, it's looking really good!
- I found one thing we need to switch for config inheritance, which I didn't realize back in #3257.
- I was able to reimplement the tests in `dbt_utils` through a combination of `where` + `fail_calc`.
- I realize I originally opened this PR... so I'm going to mark it as review-ready, hah. Feel free to close and reopen in a new PR if you'd like.

PR #3343, opened 2021-05-11
Add a better error for undefined macros

### LINKED ISSUE #2999: Raise a warning for required, but uninstalled packages
### Describe the feature
The error message for an uninstalled package comes straight out of the Jinja library:
```
Compilation Error in analysis scratch (analysis/scratch.sql)
    'dbt_utils' is undefined
```

This error message _is_ strictly true, but I think it might be worth taking the time to write a more explicit warning message for uninstalled packages

```
Warning: The following packages are listed as required in `packages.yml` but have not been installed
- dbt_utils

Run `dbt deps` to install these packages
```

Since we have the expected packages in `packages.yml` and the installed packages in `dbt_modules`, this should be possible, right? And not that hard?

Note: I've only got it as a warning so that the places where the package is referenced still causes an error, but I'm open to having this as an error instead

Keep as is

### Who will this benefit?
Confused users

No ty :)

jaypeedevlin:
resolves #2999 

### Description

Adds a new `UndefinedMacroException` and raises this when an undefined macro is caught.

### Checklist


jtcohen6:
@jaypeedevlin I just left a comment over in #2999! I think we can do both :)

core/dbt/config/runtime.py
```diff
@@ -328,10 +328,19 @@ def load_dependencies(self) -> Mapping[str, 'RuntimeConfig']:
...
             all_projects = {self.project_name: self}
             internal_packages = get_include_paths(self.credentials.type)
-            project_paths = itertools.chain(
-                internal_packages,
-                self._get_project_directories()
-            )
+            # raise exception if fewer installed packages than in packages.yml
+            package_dirs = self._get_project_directories()
+            count_packages_specified = len(self.packages.get('packages', 0))
+            count_packages_installed = len(tuple(package_dirs))
```
jtcohen6, line 334: Ah, I think (at least some of) the failing tests are because `_get_project_directories()` returns a generator object, so we can only access it once. No sense setting it to a var...
```suggestion
            count_packages_specified = len(self.packages.packages)
            count_packages_installed = len(tuple(self._get_project_directories()))
```

core/dbt/config/runtime.py
```diff
@@ -328,10 +328,19 @@ def load_dependencies(self) -> Mapping[str, 'RuntimeConfig']:
...
+            if count_packages_specified > count_packages_installed:
+                raise_compiler_error(
+                    f'dbt found {count_packages_specified} package(s) '
+                    f'specified in packages.yml, but only '
+                    f'{count_packages_installed} package(s) installed '
+                    f'in {self.modules_path}. Run "dbt deps" to '
+                    f'install package dependencies.'
+                )
+            project_paths = itertools.chain(internal_packages, package_dirs)
```
jtcohen6, line 343: ...if we just need to call it again down here
```suggestion
            project_paths = itertools.chain(
                internal_packages,
                self._get_project_directories()
            )
```

jtcohen6 (commented):
Okay, I think I figured out why the tests are failing! There's one integration test in particular, however, that _expects_ to fail when its package is missing, and raise a helpful error message, before it actually runs `deps`:

https://github.com/fishtown-analytics/dbt/blob/bbab6c2361139073db5259d98f12addae7302b4c/test/integration/059_source_overrides_test/test_source_overrides.py#L78-L84

I think we could either:
- Split apart the test into a Failing and Passing version. In the former, the `packages_config` should be empty (thereby passing the check we're adding in this PR), and we can still confirm that source `overrides` raise a helpful error message if their associated package is missing. Then, the Passing version has a reasonable `packages_config` and runs `deps` right away.
- Remove that piece of the test. With this amazing new error message, no one should ever have this problem ever again!

jaypeedevlin:
Thanks for the pointers, we're looking in _much_ better shape now.  Here's what I have as outstanding:

- mypy is failing over the `self.packages.packages` call in `runtime.py`.  I tried replacing `.packages` with `.get('packages', [])` but it seemed like sometimes that object _isn't_ a dict so I'm not sure how to resolve this.
- There's a pesky error coming out of the 109 set of tests that I'm not sure how to handle.
- There are a series of tests around documentation in the 029 set — initially these were failing because they were missing calls to `dbt deps`, but now those calls are being made there are extra models included that are causing test failures.  I see two potential paths forward, and I'm interested in your thoughts:
  - Removing the packages config altogether — I've tested and this does solve the problem, but I'm not sure if it means the tests aren't doing all we want them to do.
  - Changing the default expected models, and then adding extra data to the expected manifests (but I'm not sure how to get this extra data if I went that route).

jtcohen6 (commented):
Nice, we're really close!

I left comments for `mypy` (let's cheat) and `029_docs_generate` (seems like we were never using the package at all).

As far as the `100_rpc_test`, here is the problematic bit:

https://github.com/fishtown-analytics/dbt/blob/8ac5cdd2e1b5de14b6e341188a2d3498af305687/test/integration/100_rpc_test/test_rpc.py#L1143-L1151

Namely, line 1146 returns:
```
E   AssertionError: 'error' != 'ready'
E   - error
E   + ready
```

Why? Now that we raise a compilation error as soon as we detect a mismatch between `packages.yml` and `dbt_modules`, the RPC server will _really aggressively_ raise that error on startup/sighup. Previously, we tested starting up an RPC server and checking that it's "ready" (even though it's missing a dependency), now it returns "error" status:
```
{'result': {'logs': [], 'state': 'error', 'error': {'message': 'Compilation Error\n  dbt found 1 package(s) specified in packages.yml, but only 0 package(s) installed in dbt_modules. Run "dbt deps" to install package dependencies.'}, 'timestamp': '2021-05-17T22:00:29.817312Z', 'pid': 62478}, 'id': 1, 'jsonrpc': '2.0'}
```
This is a positive change IMO! So we just need to rework the test accordingly:
```python
    def _check_start_predeps(self):
        self.assertFalse(os.path.exists('./dbt_modules'))
        status = self.assertIsResult(self.query('status').json())
        # will return an error because defined dependency is missing
        self.assertEqual(status['state'], 'error')
        return status
```

jtcohen6:
Ok, nothing makes you feel like you're on the right track quite like some mysterious RPC test errors! I figured out what's going on here for `test/rpcs/test_deps.py`, which makes sense. I'm not sure about the others, but they're passing when I run locally on your branch, so I'm hopeful it's just guilt by association.

This is, I realize, an opinionated change to how the RPC server works. Today, if there are packages defined and uninstalled, `dbt rpc` is still happy to spin up a server and wait for requests. Unless, of course, the uninstalled package is cause for an implicit Jinja parse error—the very thing we seek to ameliorate. So, I see this PR taking a principled stand that it's better to raise a very early, very explicit error as soon as a user interacts with the RPC server, and allow no further activity until they run `deps`.

Here's what that looks like in the test case:

https://github.com/fishtown-analytics/dbt/blob/30fed8d4218ce79d324ec41741fd41b79f678e9d/test/rpc/test_deps.py#L16-L29

We need:
- To pass `criteria='error'` as an additional keyword argument to `get_querier()`. That tells the RPC test utility, which is responsible for spinning up the RPC server, that we're not expecting it to return a `'ready'` status right from the get go.
- Contrary to the comment, we should _not_ be able to run queries at startup if there are packages missing, given the error message that this PR seeks to raise. So we should change `async_wait_for_result()` to `is_error()` in line 26—that is, it can still try to run that query, but it should expect an error in response.
- The status query should still work fine, just the comment is wrong: the status should be something _negative_.

jaypeedevlin:
Interesting — I still had a single failing test locally.  I'm going to fix the merge conflicts and see if it passes on a second go!

jtcohen6 (approved):
@jaypeedevlin This sure looks good to me! Thanks so much for the contribution :)

PR #3358, opened 2021-05-15
removing get_missing_var from ParseVar class

jpmmcneill:
resolves https://github.com/fishtown-analytics/dbt/issues/3320

### Description

Gets rid of a function that only returned `None` that appears to be causing the compiler having issues raising descriptive exceptions when loops and vars are involved.

Getting rid of this _seems_ to get everything working for me (ie. get more verbose errors showing up - I have tested this in jaffle_shop). I have no context as to why this odd return None function was added here so there might be an some conditional logic to make sure that the original case isn't being blown out of the water here.

### Checklist

### DIFF
--- a/core/dbt/context/providers.py
+++ b/core/dbt/context/providers.py
@@ -570,9 +570,7 @@ class ModelConfiguredVar(Var):
 
 
 class ParseVar(ModelConfiguredVar):
-    def get_missing_var(self, var_name):
-        # in the parser, just always return None.
-        return None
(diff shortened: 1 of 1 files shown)


jpmmcneill:
The initial naive change has lead to unit tests failing. I'll look into the reasons for failure and update below if I make progress

jpmmcneill:
Closing this issue for the moment, I may come back to it in future.

PR #3363, opened 2021-05-17
Rework dispatch, with backwards compatibility for current use

### LINKED ISSUE #3362: Test parsing context: package + macro deps + dispatch, oh my
Follow-on from https://github.com/fishtown-analytics/dbt/pull/3345 (I know, I hoped we were done with this one too!)

For better and for worse (definitely better!), `dbt-utils` does a lot of things with macro nesting and dispatching within custom schema test definitions. So, it's pretty good surface area for testing this functionality. This is on me for not testing #3345 against `dbt-utils` before approving #3345.

The problem seems to be when a package macro's dispatched implementation calls, in turn, another dispatched package macro. The second macro isn't included in the test parsing context. Currently, we're seeing this issue with `dbt_utils.test_recency`, the _default implementation of which_ in turn [calls the `dbt_utils.current_timestamp` macro](https://github.com/fishtown-analytics/dbt-utils/blob/489a44bb7ebad1a6d9ba762fdcd7dc6a9fea7e0d/macros/schema_tests/recency.sql#L15):
```
Encountered an error:
Compilation Error in test dbt_utils_recency_test_recency_day__today__1 (models/schema_tests/schema.yml)
  'dict object' has no attribute 'current_timestamp'
  > in macro test_recency (macros/schema_tests/recency.sql)
  > called by macro default__test_recency (macros/schema_tests/recency.sql)
  > called by test dbt_utils_recency_test_recency_day__today__1 (models/schema_tests/schema.yml)
```

I've opened a draft PR (#3363) that adds a reproduction case to `008_schema_tests_test`. I'm hopeful that this is as simple as altering the logic we've already put in place to recurse through `depends_on.macros` and add to the test parsing context.

Once we have a fix for that, we'll also need to cherry-pick it (and the fix from #3345) onto `0.19.latest` for inclusion in v0.19.2-rc2.

### LINKED ISSUE #3383: Rework adapter.dispatch(), remove packages arg
We've run into some issues (#3362) with statically analyzing macros, which is required to resolve v0.19.1 bugs, continue our performance work, and enable compelling features in the future.

The big blocker is that `adapter.dispatch()` is not-quite statically analyzable today, since the `packages` argument can accept a Jinja macro. We're working on a fix that will be backwards compatible for how we know this is used in the wild today, by packages such as [dbt-utils](https://github.com/fishtown-analytics/dbt-utils/blob/master/macros/cross_db_utils/_get_utils_namespaces.sql), [fivetran-utils](https://github.com/fivetran/dbt_fivetran_utils/blob/master/macros/_get_utils_namespaces.sql), and [dbt-expectations](https://github.com/calogica/dbt-expectations/blob/main/macros/utils/_get_namespaces.sql).

Namely, so long as the user sets a variable named `<package_name>_dispatch_list`, dbt will set `packages` to the equivalent of `var('<package_name>_dispatch_list', []) + ['<package_name']`. All open source packages I know of that lean on dispatch have followed this convention for the `_get_namespaces()` macro that they currently pass to the `packages` argument of dispatched macros.

In the meantime, we also want to change `dispatch()` to:
- Safeguard how users are actually getting value from it
- Ensure that all macro dependencies in a dbt project are statically analyzable

### Proposal

**The Jinja macro `adapter.dispatch()` should only take one argument, `macro_name` (a string literal).** Package maintainers shouldn't need to think about what kind of vars/macros to pass into the `packages` arg at all.

For users, instead of setting a magic variable name (e.g. `dbt_utils_dispatch_list`), we should support a new config specification in `dbt_project.yml`:
```yml
dispatch:
  - macro_namespace: dbt_utils
    search_order: ['spark_utils', 'dbt_utils']
  - macro_namespace: fivetran_utils
    search_order: ['my_project', 'dbt_utils', 'fivetran_utils']
```
When dbt goes to dispatch a macro, it uses the `search_order` corresponding to the `macro_namespace` (package) of the dispatching macro. This accomplishes the same functionality as the current `packages` + vars setup, but with much less confusion, and it centralizes all macro "routing" logic in the project file. (If a dispatching macro does not have a "route" set in the project `dispatch` config, it just searches within its own project—the same as current behavior when `packages` is `None`.)

This should also make the syntax for `dispatch` a little less onerous:
```sql
{% macro my_macro(field) %}
  {{ return(adapter.dispatch('my_macro')(field)) }}
{% endmacro %}
```

Whether the python `dispatch` argument still takes two arguments, with `search_order` passed in as `packages`, is an implementation detail. But **in v0.20.0, we should start raising a deprecation warning if the _user_ passes a value to the `packages` argument of the `adapter.dispatch()` Jinja macro**.

And in a future version of dbt, once everyone has had a chance to upgrade to this new way of dispatching, we should remove the Jinja AST-parsing logic that #3363 is adding for backwards compatibility.

jtcohen6:
resolves #3362
resolves #3383 

This just adds a reproduction case for now. I've confirmed that, when I run pytest locally, it raises the same error we're seeing in the wild:

```
E       dbt.exceptions.CompilationException: Compilation Error in test pkg_and_dispatch_model_c_ (test-context-models/schema.yml)
E         'dict object' has no attribute 'current_timestamp'
E
E         > in macro test_pkg_and_dispatch (macros/custom_test.sql)
E         > called by macro default__test_pkg_and_dispatch (macros/custom_test.sql)
E         > called by test pkg_and_dispatch_model_c_ (test-context-models/schema.yml)
```

### Checklist


jtcohen6 (commented):
I confirmed that this offers backwards compatibility for the `dbt-utils` package. Hooray! I moved a little of the logic around, to account for when `packages` is a positional arg and also a `Call`, and I touched up the error message. (Looks like I messed up the flake8 linting, sorry about that.)

I'm so happy to be adding this logic and ensure projects have a smooth upgrade path to v0.19.2 + v0.20.0; I also want to get "the right way" to do this in place, and immediately recommend it to everyone who's upgrading. I laid out my sketch of a longer-term sustainable solution over in https://github.com/fishtown-analytics/dbt/issues/3383.

core/dbt/clients/jinja_static.py
```diff
@@ -0,0 +1,204 @@
...
+#          dyn_args=None, dyn_kwargs=None)
+#     )
+#   ],
+#   dyn_args=None,
+#   dyn_kwargs=None
+# )
+def statically_parse_adapter_dispatch(func_call, ctx, db_wrapper):
+    possible_macro_calls = []
+    # This captures an adapter.dispatch('<macro_name>') call.
+
```
jtcohen6, line 90: Can we leave some comments below around:
- Which logic offers backwards compatibility in 0.19.2 and should be removed in a future version
- Which pieces need to remain for good, to enable dispatching during static analysis

core/dbt/clients/jinja_static.py
```diff
@@ -0,0 +1,204 @@
...
+                    possible_macro_calls.append(func_name)
+                else:
+                    raise_compiler_error(f"The macro_name parameter ({kwarg.value.value}) "
+                                         "to adapter.dispatch was not a string")
+            elif kwarg.key == 'macro_namespace':
+                if type(kwarg.value).__name__ == 'Const':
+                    macro_namespace = kwarg.value.value
+                else:
+                    raise_compiler_error(f"The macro_namespace parameter ({kwarg.value.value}) "
+                                         "to adapter.dispatch was not a string")
```
jtcohen6, line 128: If `kwarg.value` is not a Const, then `kwarg.value.value` is not always defined. E.g. if `macro_namespace` is set to a list, then `kwarg.value` is `List(items=[Const(value='dbt_utils')])`. So we get a bad error here:
```
'List' object has no attribute 'value'
```
I'm not sure if there's some way to reliably unpack the value. In any case, we could adapt the error message to be more like:
 ```python
 kwarg_type = type(kwarg.value).__name__
 The macro_namespace parameter to adapter.dispatch is a {kwarg_type}, not a string
 ```

jtcohen6 (commented):
These changes are looking really good! I confirmed they worked with `dbt-utils`, both with:
- The setup as it exists today
- The changes we'll want to make to support the new `dispatch` project config

I left a few comments, nothing major.

Next steps from here:
- Merge this PR into `develop` (or a new one, it doesn't have to be the one I stubbed out last week! haha)
- Cherry-pick this PR onto `0.19.latest`
- Create a new PR, for inclusion in `develop` (v0.20) _only_, that raises a deprecation warning if the user either provides the `packages` kwarg or passes a list to the second positional arg

gshank (approved):

PR #3374, opened 2021-05-19
Add experimental parser behind flag

### LINKED ISSUE #3307: Package dbt-jinja so it can be distributed with dbt
Package dbt-jinja so it can be imported by dbt. Initially we are thinking about packaging it separately so that pip can find the correct wheel at dbt install time, and so that users don't need any additional system dependencies such as a C compiler.

nathaniel-may:
resolves #3307

### Description

Adds a flag `--use-experimental-parser` which defaults to `False`. When used, dbt will run the static extractor powered by tree-sitter on all model files and skip jinja rendering when possible.

In a separate PR I plan on writing integration tests so that the behavior when this flag is turned on is verified in CI. Right now CI only verifies that dbt installs with the new dependencies, and that the flag's default value of `False` doesn't break anything.

### Checklist


core/dbt/parser/models.py
```diff
@@ -17,3 +20,48 @@ def resource_type(self) -> NodeType:
     @classmethod
     def get_compiled_path(cls, block: FileBlock):
         return block.path.relative_path
+
+    def render_update(
+        self, node: IntermediateNode, config: ContextConfig
+    ) -> None:
+
+        # normal dbt run
+        if not flags.USE_EXPERIMENTAL_PARSER:
```
jtcohen6, line 29: Something about how we're setting the flag value doesn't seem quite right. `flags.USE_EXPERIMENTAL_PARSER` has been returning `False` here, even when I run (in our internal analytics project):
```
$ dbt --use-experimental-parser parse --no-version-check
```

core/dbt/main.py
```diff
@@ -998,12 +998,23 @@ def parse_args(args, cls=DBTArgumentParser):
     # if set, extract all models and blocks with the jinja block extractor, and
     # verify that we don't fail anywhere the actual jinja parser passes. The
     # reverse (passing files that ends up failing jinja) is fine.
+    # TODO remove?
```
jtcohen6, line 1001: 👍

nathaniel-may: I might wait to remove it till the integration test PR since it's used in a way that I should also remove some plumbing in our tests while I'm in there.

core/dbt/parser/models.py
nathaniel-may: Good catch! This was a "find and replace" error 🙃

nathaniel-may:
reminder to reviewers: `test_tree_sitter_jinja.py` and `extractor.py` which make up 90% of this PR have already been code reviewed in our live session and have had almost no changes since then. Any additional feedback is still welcome there of course.

core/setup.py
```diff
@@ -73,6 +73,8 @@ def read(fname):
         'networkx>=2.3,<3',
         'packaging~=20.9',
         'sqlparse>=0.2.3,<0.4',
+        'tree-sitter==0.19.0',
+        'tree-sitter-jinja2==0.1.0a1',
```
kwigley, line 77: I know I mentioned before that this should probably be an "opt-in" dependency, but I think I'm leaning towards not doing that. I'm pretty happy with how we packaged this up, so I'm down to keep this in the list of required dependencies! 👍

core/dbt/tree_sitter_jinja/__init__.py
```diff
@@ -0,0 +1 @@
+from . import extractor  # noqa
```
kwigley, line 1: (optional) I want to say you can remove this line, I believe it is redundant if I understand how python importing work.

core/dbt/tree_sitter_jinja/README.md
```diff
@@ -0,0 +1,38 @@
...
+
+The extractor uses the typed abstract syntax tree to easily identify all the refs, sources, and configs present and extract them to a dictionary.
+
+## Tests
+
+- Tests are in `test/unit/test_tree_sitter_jinja.py` and run with dbt unit tests
+
+## Future
+
+- This module will eventually be rewritten in Rust for the added type safety
```
kwigley, line 38: 🙏

core/dbt/tree_sitter_jinja/extractor.py
```diff
@@ -0,0 +1,292 @@
...
+    return list(filter(lambda x: x.is_named, node.children))
+
+
+def text_from_node(source_bytes, node):
+    return source_bytes[node.start_byte:node.end_byte].decode('utf8')
+
+
+def strip_quotes(text):
+    if text:
+        return text[1:-1]
```
kwigley, line 33: (optional) might be able to use python's built-in [strip](https://docs.python.org/3/library/stdtypes.html#str.strip) here!

kwigley (approved):
Looks good!

core/setup.py
jtcohen6: That's good by me! If we feel good about this installing on all supported systems/architectures/python versions, let's include it. We can call it out in the RC and see if we hear back about any installation issues.

gshank (approved):
LGTM!

core/setup.py
nathaniel-may: I want to avoid making this a special case if we can. Plus if people have install issues with the RC (which I don't expect) we can fix them before the general release.

core/dbt/tree_sitter_jinja/__init__.py
nathaniel-may: You're right! I love the chance to remove more code.

core/dbt/tree_sitter_jinja/extractor.py
nathaniel-may: Just checked this out and I think it's better to use the explicit indexing because of the following edge case which removes too many characters:
```
>>> x = '"nested "quotes""'
>>> x.strip('"')
'nested "quotes'
```

PR #3376, opened 2021-05-20
Allow to use a costum field as check-cols updated_at

### LINKED ISSUE #1844: Snapshots with check option - allow custom date field to be used when setting dbt_date_from and dbt_date_to fields.
### Overview of Change
Essentially we would like to be able to supply the following config so that a snapshot with check strategy honours the updated_at field that we supply when setting the dbt_date_from/to fields.
```
 {{
        config(
          target_database='analytics',
          target_schema='snapshots',
          unique_key='id',
          strategy='check',
          updated_at='updated_at',
        )
    }}
```

We would like to use the snapshot feature to manage type 2 changes on our dimensions (products, customers etc).  

We need to use check mode (not timestamp) so we can look for record attributes that change (as well as new/deleted records).

In our staging layer we have a dw_load_datetime column on every staged dimension that represents when the data was processed.  This is the timestamp from Airflow and importantly it allows us to replay historic runs (e.g. reload data for the past 3 days).

Question - can we tell dbt to use this dw_load_datetime in the snapshot when setting the dbt_date_from and dbt_date_to fields?

Currently dbt uses the current time to do this. We can see you support this use case when in timestamp mode - but we weren’t sure why this couldn’t also apply in check mode?

PJGaetan:
resolves #1844

### Description

Allows the use of `updated_at` field while using `check` strategy in a snapshot.
The timestamp used to update `dbt_valid_from` will come from the costum field instead of the default current timestamp.

I have added a test to ensure that rows are invalidated with the right timestamp. Not sure it is in the right place to add the test but I didn't know how to pick up `relation_b` otherwise.

### Checklist


CHANGELOG.md
```diff
@@ -41,6 +41,7 @@ Contributors:
 - Add native support for Postgres index creation ([#804](https://github.com/fishtown-analytics/dbt/issues/804), [3106](https://github.com/fishtown-analytics/dbt/pull/3106))
 - Less greedy test selection: expand to select unselected tests if and only if all parents are selected ([#2891](https://github.com/fishtown-analytics/dbt/issues/2891), [#3235](https://github.com/fishtown-analytics/dbt/pull/3235))
 - Prevent locks in Redshift during full refresh in incremental materialization. ([#2426](https://github.com/fishtown-analytics/dbt/issues/2426), [#2998](https://github.com/fishtown-analytics/dbt/pull/2998))
+Add optional `updated_at` config parameter to invalidate rows with the inputed column timestamp in snapshot using `check` strategy. ([#1844](https://github.com/fishtown-analytics/dbt/issues/1844), [#3376](https://github.com/fishtown-analytics/dbt/pull/3376))
```
jtcohen6, line 44: Just one small suggestion on the change wording, to make it clear what the current/default behavior is:

```suggestion
- Support optional `updated_at` config parameter with `check` strategy snapshots. If not supplied, will use current timestamp (default). ([#1844](https://github.com/fishtown-analytics/dbt/issues/1844), [#3376](https://github.com/fishtown-analytics/dbt/pull/3376))
```

jtcohen6 (commented):
@PJGaetan This looks great to me!

I left a small suggestion on the changelog. Could you also:
- Pull in `develop`
- Move your changelog note up to `v0.20.0` (rather than `v0.20.0b1`)
- Add yourself to the list of Contributors

CHANGELOG.md
PJGaetan: Definitely fells better this way

jtcohen6 (approved):
Thanks for the contribution @PJGaetan!

Issue #3356, opened 2021-05-14
Add keys to list output, parametrize


jtcohen6:
Add `unique_id` + `original_file_path` to the set of ALLOWED_KEYS returned by `dbt ls --output json` / RPC `list` method:

https://github.com/fishtown-analytics/dbt/blob/9693170eb9d8bba68abe6006571f1a80ed184d47/core/dbt/task/list.py#L32-L41

Edited: updated `path` above to be `original_file_path`

jtcohen6:
Resolved by #3384

Issue #3210, opened 2021-03-30
Remove the distinction between --select and --models flags when working with node types


joellabes:
### Describe the feature
It would be less confusing if all nodes were selectable using a single flag, instead of having to switch back and forth. 

### Describe alternatives you've considered
Getting over it

### Additional context
I imagine it would make more sense to settle on `--select and -s` because they're more broadly applicable. The migration might be uncomfortable, but could be a slow-boiled frog in the same vein as `config-version`: 
- A minor version of dbt where `--select` is the default version in the docs etc for new users and `-m` users keep going without any issues
- Some minor versions of dbt where `-m` users get a WARN during compilation 
- A future version of dbt (maybe 1.0?) where `-m` is deprecated and no longer respected. 

Or, just add `-s` as an alias and leave the old ones around forever; they're not really doing any harm! 

### Who will this benefit?
New users in particular (this confused me [~9 months ago](https://getdbt.slack.com/archives/C2JRRQDTL/p1595802905302700)), but it also just bit me again 😑.

### Are you interested in contributing this feature?
Sure! I'm sure I could work it out by looking at how the `-model` alias is currently handled. Most of the work feels like it would be in the deprecation warnings

jtcohen6:
@joellabes I agree! In all likelihood, I think we should:
- Switch all commands / tasks to use `-s, --select`
- Keep supporting `-m, --models`, forever, as an (undocumented) alias for `-s, --select` 

There are two things we'll need to sort out before we do this:

### Trivial: list

The `list` task / `dbt ls` command is the only one that supports both `-m, --models` and `-s, --select`, and they do slightly different things. Instead of `-m` and `-s` being direct aliases,  `--models xyz` is functionally equivalent to `--resource-type model --select xyz`. I think we could deprecate `--models` and still have it perform that same functionality. We already do a good job today of raising an exception if `--models` is passed alongside either `--select` or `--resource-type`, since they're mutually exclusive.

### Trickier: test selection

`dbt test --models` has a "magic jump" baked in, such that `dbt test -m model_a` _really_ means, "Select all the tests that directly depend on model_a`, i.e. `dbt ls -s model_a+1 --resource-type test`. It's a bit unintuitive, but it also enables really compelling syntax for "run and test model_a." But I think it _only_ makes more conceptual sense for `test -m`, i.e. execute tests that directly apply to a model, even if the tests themselves are not directly selected—whereas `test -s`, to me, sounds like it should execute only the tests that are directly selected.

[...]

PR #3384, opened 2021-05-21
Add `ls` to RPC server

### LINKED ISSUE #3311: Add `list` RPC method
_picks up from #3269_

Add a `list` method to the RPC server.

~Can we run it synchronously / single-threaded, to avoid delays associated with polling?~ Running this sync won't be in scope for the initial effort. The most important thing is wiring it up and making sure the inputs + outputs match the `dbt ls` CLI command.

iknox-fa:
resolves #3311 and #3356

### Description
Adds support for the `list` method in the RPC server
Adds new keys for `unique_id` and `original_file_path` to list task when invoked with json output

### Checklist


core/dbt/contracts/rpc.py
```diff
@@ -63,6 +63,16 @@ class RPCCompileParameters(RPCParameters):
     state: Optional[str] = None
 
 
+@dataclass
+class RPCListParameters(RPCParameters):
+    resource_types: Optional[List[str]] = None
+    models: Union[None, str, List[str]] = None
+    exclude: Union[None, str, List[str]] = None
+    select: Optional[List[str]] = None
```
jtcohen6, line 71: I believe the type of `select` should be `Union[None, str, List[str]]` as well. It should work just the same as `models` and `exclude`.

jtcohen6:
This is working well for me in local testing!

A few comments:
- As I noted above, could we switch the type of `select` to accept a string as well as a list? It should work exactly the same as `models` and `exclude`
- For parity with the existing `list` task, I expected to see [an error](https://github.com/fishtown-analytics/dbt/blob/4d5d0e2150ff2776f5171eb229c367af839b9a35/core/dbt/task/list.py#L46-L55) if `models` is set alongside `select` or `resource_types`, since these are mutually exclusive arguments. This shouldn't block merging the PR if it would be tricky to implement.
- The p0 resolution to #3356 is all we need for now. Could we open a new issue to capture the p1 requirement,  parametrizing the set of returned keys?

core/dbt/contracts/rpc.py
iknox-fa: Fixed!

iknox-fa:
> This is working well for me in local testing!
> 
> A few comments:
> 
> * As I noted above, could we switch the type of `select` to accept a string as well as a list? It should work exactly the same as `models` and `exclude`
Done!
> * For parity with the existing `list` task, I expected to see [an error](https://github.com/fishtown-analytics/dbt/blob/4d5d0e2150ff2776f5171eb229c367af839b9a35/core/dbt/task/list.py#L46-L55) if `models` is set alongside `select` or `resource_types`, since these are mutually exclusive arguments. This shouldn't block merging the PR if it would be tricky to implement.
Done!
> * The p0 resolution to #3356 is all we need for now. Could we open a new issue to capture the p1 requirement,  parametrizing the set of returned keys?
Done!

test/integration/047_dbt_ls_test/test_ls.py
```diff
@@ -94,6 +94,8 @@ def expect_snapshot_output(self):
                     'alias': None,
                     'check_cols': None,
                 },
+                'unique_id': 'snapshot.test.my_snapshot',
+                'original_file_path': 'snapshots/snapshot.sql',
```
jtcohen6, line 98: Looks like this is raising an error on Windows because paths are different. I think we have [`normalize()`](https://github.com/fishtown-analytics/dbt/blob/0f018ea5dd678d37668c819552328b3254c2f220/test/integration/base.py#L32-L42) + [`Normalized()`](https://github.com/fishtown-analytics/dbt/blob/0f018ea5dd678d37668c819552328b3254c2f220/test/integration/base.py#L45-L56) for this purpose.

jtcohen6 (commented):
- There's one last failing test on Windows, I think because of file paths
- I confirmed that passing both `models` and `select` raises an error (good!), passing both `resource_types` params still doesn't raise the expected error (`"models" and "resource_type" are mutually exclusive arguments`). This isn't the most important thing: we're thinking of retiring the `--models` flag/param soon, anyway (#3210).

Once the failing test is sorted, I'm quite happy with where this is at. Nice work!

test/integration/047_dbt_ls_test/test_ls.py
iknox-fa: Done!

iknox-fa:
> * There's one last failing test on Windows, I think because of file paths
Fixed!
> * I confirmed that passing both `models` and `select` raises an error (good!), passing both `resource_types` params still doesn't raise the expected error (`"models" and "resource_type" are mutually exclusive arguments`). This isn't the most important thing: we're thinking of retiring the `--models` flag/param soon, anyway (#3210).
Fixed!
> 
> Once the failing test is sorted, I'm quite happy with where this is at. Nice work!

kwigley (approved):
👍

jtcohen6 (approved):
lgtm!

PR #2928, opened 2020-12-03
Support require_partition_filter and partition_expiration_days in BQ

yu-iskw:
resolves #1843

### Description
We will support `require_partition_filter` and `partition_expiration_days` in the BigQuery adapter.

### Checklist
 - [x] I have signed the [CLA](https://docs.getdbt.com/docs/contributor-license-agreements)
 - [x] I have run this code in development and it appears to resolve the stated issue
 - [x] This PR includes tests, or tests are not required/relevant for this PR
 - [x] I have updated the `CHANGELOG.md` and added information about my change to the "dbt next" section.

PR #3386, opened 2021-05-23
BQ partition configs support

### LINKED ISSUE #3016: [BQ] Support new partition configs within incremental strategies
### Describe the feature
Picking up from #2928, ensure that `require_partition_filter` works with all the permutations of incremental models on BigQuery. Anyone is welcome to pick up this contribution for `v0.20.0`!

### The `merge` strategy

We need the merge condition to be
```sql
on 
            DBT_INTERNAL_SOURCE.[unique_key] = DBT_INTERNAL_DEST.[unique_key]
            and DBT_INTERNAL_DEST.[partition_col] is not null
```

This could be accomplished by passing an additional `predicate` to `get_merge_sql` [here](https://github.com/fishtown-analytics/dbt/blob/1f927a374c8bd52a12a20d892fed9d59cffd04f4/plugins/bigquery/dbt/include/bigquery/macros/materializations/incremental.sql#L147), something like:
```sql
{% is_partition_filter_required = config.get('require_partition_filter', false) %}
{% set predicates = [] %}
{% if is_partition_filter_required %}
  {% set partition_filter %} ({{ partition_by.field }} is not null or {{ partition_by.field }} is null) {% endset %}
  {% do predicates.append(partition_filter) %}
{% endif %}

{% set build_sql = get_merge_sql(target_relation, source_sql, unique_key, dest_columns, predicates) %}
```

This is a bit of a hack—filtering only in [this sense](https://www.reddit.com/r/funny/comments/1ornz8/these_two_books_contain_the_sum_total_of_all/)—but to be honest there isn't any straightforward way dbt can know in advance the specific partitions it's merging into. For that, you should use...

### The `insert_overwrite` strategy

The current error comes in step 2:

https://github.com/fishtown-analytics/dbt/blob/1f927a374c8bd52a12a20d892fed9d59cffd04f4/plugins/bigquery/dbt/include/bigquery/macros/materializations/incremental.sql#L63-L68

We need to either:
- Change this query to include a `where {{ partition_by.field }} is not null`, thereby satisfying the filter requirement
- Change the temp table to _not_ set `require_partition_by`. Should this be a more general rule, that if `temporary = True`, the `create table` statement shouldn't set `require_partition_filter = True`? I think it would make good sense.

- Not supporting `require_partition_filter` for incremental models. I think we definitely should!

### Related unsolved questions
- Default schema tests will not work on tables with `require_partition_filter` set
- Related addition in #2928: How should we advise on whether to set `partition_expiration_days` for incremental models? Personally, I'm not so sure

### Who will this benefit?
- BigQuery users with large-volume datasets

prratek:
resolves #3016 

### Description

#2928 added support for `require_partition_filter` and `partition_expiration_days` in the BigQuery adapter. As outlined in #3016, this PR ensure that the new config options work with all incremental strategies available for BigQuery today.

### Checklist

### DIFF
--- a/plugins/bigquery/dbt/include/bigquery/macros/materializations/incremental.sql
+++ b/plugins/bigquery/dbt/include/bigquery/macros/materializations/incremental.sql
@@ -122,26 +122,26 @@
                 {% do exceptions.raise_compiler_error(missing_partition_msg) %}
             {% endif %}
 
-        {% set build_sql = bq_insert_overwrite(
-            tmp_relation,
-            target_relation,
-            sql,
-            unique_key,
-            partition_by,
-            partitions,
-            dest_columns) %}
+            {% set build_sql = bq_insert_overwrite(
+                tmp_relation,
+                target_relation,
+                sql,
+                unique_key,
+                partition_by,
+                partitions,
+                dest_columns) %}
 
-    {% else %}
-        {#-- wrap sql in parens to make it a subquery --#}
-        {%- set source_sql -%}
-            (
-                {{sql}}
-            )
-        {%- endset -%}
+        {% else %}
+            {#-- wrap sql in parens to make it a subquery --#}
+            {%- set source_sql -%}
+                (
+                    {{sql}}
+                )
+            {%- endset -%}
 
-    {% set build_sql = get_merge_sql(target_relation, source_sql, unique_key, dest_columns) %}
+            {% set build_sql = get_merge_sql(target_relation, source_sql, unique_key, dest_columns) %}
 
-    {% endif %}
+        {% endif %}
 
     {% endif %}
 
--- a/plugins/bigquery/dbt/include/bigquery/macros/materializations/incremental.sql
+++ b/plugins/bigquery/dbt/include/bigquery/macros/materializations/incremental.sql
@@ -92,6 +92,7 @@
     {%- set raw_partition_by = config.get('partition_by', none) -%}
     {%- set partition_by = adapter.parse_partition_by(raw_partition_by) -%}
     {%- set partitions = config.get('partitions', none) -%}
+    {%- set is_partition_filter_required = config.get('require_partition_filter', false) -%}
     {%- set cluster_by = config.get('cluster_by', none) -%}
 
     {{ run_hooks(pre_hooks) }}
@@ -135,11 +136,19 @@
             {#-- wrap sql in parens to make it a subquery --#}
             {%- set source_sql -%}
                 (
-                    {{sql}}
+                    {{ sql }}
                 )
             {%- endset -%}
 
-            {% set build_sql = get_merge_sql(target_relation, source_sql, unique_key, dest_columns) %}
+            {% set predicates = [] %}
+            {% if is_partition_filter_required %}
+                {%- set partition_filter -%}
+                    (DBT_INTERNAL_DEST.{{ partition_by.field }} is not null or DBT_INTERNAL_DEST.{{ partition_by.field }} is null)
+                {%- endset -%}
+                {% do predicates.append(partition_filter) %}
+            {% endif %}
+
+            {% set build_sql = get_merge_sql(target_relation, source_sql, unique_key, dest_columns, predicates) %}
 
         {% endif %}
 
--- a/plugins/bigquery/dbt/include/bigquery/macros/materializations/incremental.sql
+++ b/plugins/bigquery/dbt/include/bigquery/macros/materializations/incremental.sql
@@ -1,126 +1,126 @@
 
 {% macro dbt_bigquery_validate_get_incremental_strategy(config) %}
-  {#-- Find and validate the incremental strategy #}
-  {%- set strategy = config.get("incremental_strategy", default="merge") -%}
+    {#-- Find and validate the incremental strategy #}
+    {%- set strategy = config.get("incremental_strategy", default="merge") -%}
 
-  {% set invalid_strategy_msg -%}
-    Invalid incremental strategy provided: {{ strategy }}
-    Expected one of: 'merge', 'insert_overwrite'
-  {%- endset %}
-  {% if strategy not in ['merge', 'insert_overwrite'] %}
-    {% do exceptions.raise_compiler_error(invalid_strategy_msg) %}
-  {% endif %}
+    {% set invalid_strategy_msg -%}
+        Invalid incremental strategy provided: {{ strategy }}
+        Expected one of: 'merge', 'insert_overwrite'
+    {%- endset %}
+    {% if strategy not in ['merge', 'insert_overwrite'] %}
+        {% do exceptions.raise_compiler_error(invalid_strategy_msg) %}
+    {% endif %}
 
-  {% do return(strategy) %}
+    {% do return(strategy) %}
 {% endmacro %}
 
 
 {% macro bq_insert_overwrite(tmp_relation, target_relation, sql, unique_key, partition_by, partitions, dest_columns) %}
 
-  {% if partitions is not none and partitions != [] %} {# static #}
+    {% if partitions is not none and partitions != [] %} {# static #}
 
-      {% set predicate -%}
-          {{ partition_by.render(alias='DBT_INTERNAL_DEST') }} in (
-              {{ partitions | join (', ') }}
-          )
-      {%- endset %}
+        {% set predicate -%}
+            {{ partition_by.render(alias='DBT_INTERNAL_DEST') }} in (
+                {{ partitions | join (', ') }}
+            )
+        {%- endset %}
 
-      {%- set source_sql -%}
-        (
-          {{sql}}
-        )
-      {%- endset -%}
+        {%- set source_sql -%}
+            (
+                {{sql}}
+            )
+        {%- endset -%}
 
-      {{ get_insert_overwrite_merge_sql(target_relation, source_sql, dest_columns, [predicate], include_sql_header=true) }}
+        {{ get_insert_overwrite_merge_sql(target_relation, source_sql, dest_columns, [predicate], include_sql_header=true) }}
 
-  {% else %} {# dynamic #}
+    {% else %} {# dynamic #}
 
-      {% set predicate -%}
-          {{ partition_by.render(alias='DBT_INTERNAL_DEST') }} in unnest(dbt_partitions_for_replacement)
-      {%- endset %}
+        {% set predicate -%}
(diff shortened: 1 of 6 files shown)


plugins/bigquery/dbt/include/bigquery/macros/materializations/incremental.sql
```diff
@@ -1,160 +1,169 @@
...
+            {%- set source_sql -%}
+                (
+                    {{ sql }}
+                )
+            {%- endset -%}
+
+            {% set predicates = [] %}
+            {% if is_partition_filter_required %}
+                {%- set partition_filter -%}
+                    (DBT_INTERNAL_DEST.{{ partition_by.field }} is not null or DBT_INTERNAL_DEST.{{ partition_by.field }} is null)
```
prratek, line 146: I wasn't sure if there was a better way to get the `DBT_INTERNAL_DEST` alias in there than just typing it out. I tried `{{ partition_by.render(alias='DBT_INTERNAL_DEST') }}` but that renders to `timestamp_trunc(my_partition_col, date)` when partitioning by date on a timestamp col, and BigQuery doesn't like that as a partition filter.

prratek:
Okay a few notes:
1. Sorry for the large diff! The inconsistent indentation was driving me nuts 
2. I'd love some guidance on writing a test case to make sure this works for the `merge` strategy. I tested this locally by putting together a minimal dbt project with an incremental model that uses these configs and I've worked with the integration test suite once before but still don't feel too comfortable navigating everything under `test/integration`

plugins/bigquery/dbt/include/bigquery/macros/materializations/incremental.sql
jtcohen6: > and BigQuery doesn't like that as a partition filter

Really? It works as a `partition by` expression, but not as a partition-pruning filter? That's... too bad.

The approach you've taken here seems fine by me

jtcohen6:
@prratek Thanks so much for taking this on!

As far as testing this functionality, the right place might just be an extension of [`022_bigquery_test/test_incremental_strategies.py`](https://github.com/fishtown-analytics/dbt/blob/develop/test/integration/022_bigquery_test/test_incremental_strategies.py). There are a bunch of models [here](https://github.com/fishtown-analytics/dbt/tree/develop/test/integration/022_bigquery_test/incremental-strategy-models) that mix and match incremental strategies and partition configurations. I think you could extend the [`TestBigQueryScripting` case](https://github.com/fishtown-analytics/dbt/blob/4d5d0e2150ff2776f5171eb229c367af839b9a35/test/integration/022_bigquery_test/test_incremental_strategies.py#L3), with one crucial adjustment: turn on `require_partition_filter` for all the models.

```python
@property
def project_config(self):
    return {
        'config-version': 2,
        'seeds': {
            '+quote_columns': False,
        },
        'models': {
            '+require_partition_filter': True
        },
    }
```

If every model / strategy / partition combo runs with that config turned on, we'll know that this change has been successful. (Conversely, without the change in this PR, some of those models should fail.)

prratek:
That makes sense! Couple of thoughts:
1. I saw your comment in #3016 about having temp tables not require a partition filter more generally and that seems reasonable. Could that be accomplished by just changing [this](https://github.com/fishtown-analytics/dbt/issues/3016) `if` statement? Something like:
```python
if config.get('require_partition_filter') and not temporary:
```
2. [This](https://github.com/fishtown-analytics/dbt/blob/develop/core/dbt/include/global_project/macros/adapters/common.sql#L46) colon looks suspicious - is that valid jinja?

jtcohen6:
I can't see what line you linked to in item 1; I think we're on the same page, but just to be safe, I'll say what I'm thinking.

The operative logic is going to be in the BigQuery plugin, rather than the default, since BigQuery implements its own [`bigquery__create_table_as`](https://github.com/fishtown-analytics/dbt/blob/d89e1d7f850ba5a48e66cfabf0dcfe14afc32ccc/plugins/bigquery/dbt/include/bigquery/macros/adapters.sql#L42). That macro calls [`bigquery_table_options`](https://github.com/fishtown-analytics/dbt/blob/d89e1d7f850ba5a48e66cfabf0dcfe14afc32ccc/plugins/bigquery/dbt/include/bigquery/macros/adapters.sql#L31), which in turn shells out to a python adapter method, [`get_table_options`](https://github.com/fishtown-analytics/dbt/blob/d89e1d7f850ba5a48e66cfabf0dcfe14afc32ccc/plugins/bigquery/dbt/adapters/bigquery/impl.py#L772). In particular, we'd want to change this line in the way you recommend:
https://github.com/fishtown-analytics/dbt/blob/d89e1d7f850ba5a48e66cfabf0dcfe14afc32ccc/plugins/bigquery/dbt/adapters/bigquery/impl.py#L797-L799
```python
 if config.get('require_partition_filter') and not temporary:
       opts['require_partition_filter'] = config.get(
                 'require_partition_filter') 
```

There's something funny about the way we've implemented temp tables on BigQuery, because (a) our implementation predated "true" temp tables in BQ, and (b) "true" temp tables are only supported in scripting-style queries, which the `snapshot` materialization can't quite be today. So instead of "true" temp tables, dbt creates _real_ tables with 12-hour expiration windows. That doesn't change the substance of the change you need to make here, I just wanted to make sure you weren't staring at the code in confusion.

For the integration test, rather than turning on `require_partition_filter` for the existing test case, could you create a new test case that inherits from the existing one, with `require_partition_filter` turned on? I want to make sure this works both with and without the config for all incremental strategies.

Last but not least, as far as that colon in Jinja: I hear you! It works either way, though.
```jinja
{% if True: %} select 1 as fun {% endif %}
{% if True %} select 1 as fun {% endif %}
```
I figure some folks prefer the colon because it looks more like a python conditional:
```python
if True:
  return "select 1 as fun"
```

prratek:
Sorry it took me a while to get back to this, but I think I'm almost there:
1. For the integration tests, I had to navigate the fact that BigQuery can't use the result of a subquery to prune partitions and that the query used to get the most recent row from the table must itself also include a partition filter. The test as it stands feels a little hacky but it works. Open to any suggestions!
2. @jtcohen6 do you know what `_dbt_max_partition` is [here](https://github.com/fishtown-analytics/dbt/blob/develop/test/integration/022_bigquery_test/incremental-strategy-models/incremental_overwrite_date.sql#L40)? A couple of the models in that directory use it and it doesn't prune partitions but I'd want to understand what it's doing before changing anything.
3. Was there anything else you think I should write tests for? You mentioned a possible test case to ensure the static `insert_overwrite` strategy works and I think [this model](https://github.com/fishtown-analytics/dbt/blob/develop/test/integration/022_bigquery_test/incremental-strategy-models/incremental_overwrite_partitions.sql) should cover that.

jtcohen6:
> 1. For the integration tests, I had to navigate the fact that BigQuery can't use the result of a subquery to prune partitions and that the query used to get the most recent row from the table must itself also include a partition filter. The test as it stands feels a little hacky but it works. Open to any suggestions!

Nice work on this! Looks similar to what we've done on other projects/packages for BigQuery.

> 2. @jtcohen6 do you know what `_dbt_max_partition` is [here](https://github.com/fishtown-analytics/dbt/blob/develop/test/integration/022_bigquery_test/incremental-strategy-models/incremental_overwrite_date.sql#L40)? A couple of the models in that directory use it and it doesn't prune partitions but I'd want to understand what it's doing before changing anything.

In the "dynamic" version of `insert_overwrite`, `_dbt_max_partition` is [included in the script](https://github.com/fishtown-analytics/dbt/blob/14507a283e6acd710363aab9fc8fbcac1658002d/plugins/bigquery/dbt/include/bigquery/macros/materializations/incremental.sql#L50-L53) to calculate the max partition value. So its query may not be pruning partitions, it's just `select max(partition_col)`, but I'd think the resulting value _should_ effectively prune partitions. If it doesn't, well... we already know that this approach to `_dbt_max_partition` is not ideal: https://github.com/fishtown-analytics/dbt/issues/2278

> 3. Was there anything else you think I should write tests for? You mentioned a possible test case to ensure the static `insert_overwrite` strategy works and I think [this model](https://github.com/fishtown-analytics/dbt/blob/develop/test/integration/022_bigquery_test/incremental-strategy-models/incremental_overwrite_partitions.sql) should cover that.

I think you've got it covered!

jtcohen6:
@prratek I looked a bit more into this, and I've found something pretty wacky: wrapping a datetime-type partition column in `date(...)` and `timestamp_trunc(..., day)` both work just fine for partition filtering and elimination, but when I wrap it in `datetime_trunc(...)`, it doesn't work!

```sql
-- models/my_model.sql
{{
    config(
        materialized="incremental",
        incremental_strategy="insert_overwrite",
        partition_by={
            "field": "date_time",
            "data_type": "datetime"
        },
        require_partition_filter = True
    )
}}

select 1 as id, cast('2020-01-01' as datetime) as date_time
```

```sql
-- excerpted from insert_overwrite script, which I copy-pasted into BigQuery console to confirm

merge into `my_model` as DBT_INTERNAL_DEST
        using ( ... ) as DBT_INTERNAL_SOURCE
        on FALSE

    when not matched by source
         and datetime_trunc(DBT_INTERNAL_DEST.date_time, day) in ('2020-01-01'')
        then delete

    when not matched then insert
        (`id`, `date_time`)
    values
        (`id`, `date_time`)
```
```
  Query error: Cannot query over table 'my_model' without a filter over column(s) 'date_time' that can be used for partition elimination at [42:5]
  compiled SQL at target/run/my_project/models/my_model.sql
```

This sure looks like a BigQuery bug, doesn't it?

In the meantime, we could work around it by adding another filter:
```
    when not matched by source
         and datetime_trunc(DBT_INTERNAL_DEST.date_time, day) in ('2020-01-01', '2020-01-02')
         and DBT_INTERNAL_DEST.date_time is not null
        then delete
```
This enables the query to succeed, but it will still process more bytes than it should, since `datetime_trunc` doesn't work for partition elimination. To repeat, I'm _only_ seeing this issue with `datetime_trunc`; both `date` and `timestamp` column types/functions seem to work just fine with partition filtering.

jtcohen6:
@prratek Is this something you'd still be interested in contributing? I'd be willing to find a way to work around the failing test re: filtering with `datetime_trunc`, since that really does feel like an undocumented limitation (or bug) with BigQuery.

jtcohen6:
Closing in favor of https://github.com/dbt-labs/dbt-bigquery/pull/65

PR #3387, opened 2021-05-24
Feature/incremental non full refresh

matt-winkler:
addresses #1132 

### Description
Goal is to identify whether there are different schemas in the source and target relations during an incremental run. The allowed behaviors are ['ignore', 'fail', 'append', 'sync'] specified in a new config option called on_schema_change on the incremental.sql materialization.

### New macros added to /core/dbt/include/global_project/macros/materializations/incremental/on_schema_change.sql:

`incremental_validate_on_schema_change`: defaults to 'ignore' if None or invalid values are specified
`sync_schemas`: runs ALTER TABLE statements depending on the value of on_schema_change and whether the source / target column sets differ.
`process_schema_changes`: orchestrator macro that determines which others in this module should be called based on configuration.
`get_column_names`: Extracts col.column values from a list of columns
`diff_arrays`: detects the difference between two arrays
`diff_columns`: detects the difference between two arrays of `Columns`, and calls `diff_arrays`
`check_for_schema_changes`: Determines if a source and target schema are different due to the presence / absence of columns and / or data type differences
 
### Checklist


matt-winkler:
@jtcohen6 I've run through this for Postgres / Snowflake / Bigquery, exluding the `full_refresh` behavior per our discussion offline.

plugins/bigquery/dbt/include/bigquery/macros/adapters.sql
```diff
@@ -128,6 +128,37 @@
   {% do adapter.rename_relation(from_relation, to_relation) %}
 {% endmacro %}
 
+{% macro alter_relation_add_remove_columns(relation, add_columns = none, remove_columns = none) -%}
+  {{ return(adapter.dispatch('alter_relation_add_remove_columns')(relation, add_columns, remove_columns)) }}
+{% endmacro %}
```
jtcohen6, line 133: You can remove this dispatch call and just keep the `bigquery__` implementation below

core/dbt/include/global_project/macros/materializations/incremental/incremental.sql
```diff
@@ -28,13 +38,32 @@
...
+    
+    {% if on_schema_change != 'ignore' %}
+      {% set schema_changed = check_for_schema_changes(tmp_relation, target_relation) %}
+      {% do log('schema changed: %s' % schema_changed, info=true) %}
+      {% if schema_changed %}
+        {% do process_schema_changes(on_schema_change, tmp_relation, existing_relation) %}
+        {% set build_sql = incremental_upsert(tmp_relation, target_relation, unique_key=unique_key) %}
+      
+      {% else %}
+        {% do adapter.expand_target_column_types(
```
jtcohen6, line 54: Let's always `expand_target_column_types` before `check_for_schema_changes`, even if it means running a few extra metadata queries, in order to avoid over-aggressively dropping and recreating columns that just need to be varchar/numeric-type expanded

core/dbt/include/global_project/macros/materializations/incremental/on_schema_change.sql
```diff
@@ -0,0 +1,138 @@
...
+  {%- set target_columns = adapter.get_columns_in_relation(target_relation) -%}
+  {%- set source_not_in_target = diff_columns(source_columns, target_columns) -%}
+  {%- set target_not_in_source = diff_columns(target_columns, source_columns) -%}
+
+  {% if source_not_in_target != [] %}
+    {% set schema_changed = True %}
+  {% elif target_not_in_source != [] %}
+    {% set schema_changed = True %}
+  {% endif %}
+
```
jtcohen6, line 76: Could we centralize all schema-change-related logging in this macro, to avoid repeating it in the materialization code?

plugins/bigquery/dbt/include/bigquery/macros/materializations/incremental.sql
```diff
@@ -94,54 +137,69 @@
...
-     {% else %}
-       {#-- wrap sql in parens to make it a subquery --#}
-       {%- set source_sql -%}
-         (
-           {{sql}}
-         )
-       {%- endset -%}
-
-       {% set build_sql = get_merge_sql(target_relation, source_sql, unique_key, dest_columns) %}
+     {% set tmp_relation = make_temp_relation(target_relation) %}
```
jtcohen6, line 165: **Stylistically,** I think we can reorder the logic below to avoid repetition:
```sql
{% if on_schema_change != 'ignore' %}
  {# only create the temp table if it's needed. can we make this a free query? #}
  {% do run_query(create_table_as(True, tmp_relation, sql)) %}
  {% set schema_changed = check_for_schema_changes(tmp_relation, target_relation) %}
  {% if schema_changed %}
    {% do process_schema_changes(on_schema_change, tmp_relation, existing_relation) %}
  {% endif %}
{% endif %}
{% set dest_columns = adapter.get_columns_in_relation(existing_relation) %}
{% set build_sql = dbt_bq_get_incremental_sql(strategy, tmp_relation, target_relation, sql, unique_key, partition_by, partitions, dest_columns) %}
```

**Substantively,** there is something subtle going on here: We're creating a temp table first thing, in order to power schema comparison. I think it's fair to document that the "price" of dynamic schema evolution is the need to produce that temp table, but today, dbt does _not_ produce a temp table for the `merge` strategy, and produces that temp table as part of a script in the `insert_overwrite` strategy.

I'm wondering if we could adjust the temp table creation to be a "free" ($0) query, i.e. by adding a `where false limit 0` condition. So like:
```sql
{% set tmp_sql %}
  select * from ({{ sql }}) where false limit 0
{% endset %}
{% do run_query(create_table_as(True, tmp_relation, tmp_sql)) %}
```

plugins/snowflake/dbt/include/snowflake/macros/materializations/incremental.sql
```diff
@@ -46,20 +47,44 @@
...
     {% do adapter.drop_relation(existing_relation) %}
     {% set build_sql = create_table_as(False, target_relation, sql) %}
+  
   {% elif full_refresh_mode %}
     {% set build_sql = create_table_as(False, target_relation, sql) %}
+  
   {% else %}
     {% do run_query(create_table_as(True, tmp_relation, sql)) %}
-    {% do adapter.expand_target_column_types(
+
```
jtcohen6, line 62: As with the default materialization, let's consolidate this logic to:
1. Always expand target column types first thing
2. Check for schema changes if and only if `on_schema_change != 'ignore'`. If there are schema changes, process them, endif.
3. Grab `dest_columns` and set `build_sql`

jtcohen6 (commented):
Coming along nicely! Thanks for the work shoring up the progress so far, outside the trickier implications of `on_schema_change: full_refresh`.

I left some comments around consolidating logic. I'd like to keep the "footprint" on materializations as light as possible, since that's where we have the least modularity, greatest complexity, and most duplication across adapters.

Let me know how I can help as you get into unit + integration tests!

matt-winkler:
@jtcohen6 updated the unit tests here that aren't impacted by the NodeConfig change on incrementals A few still failing on snapshots but they appear unrelated to this code.

core/dbt/include/global_project/macros/materializations/incremental/on_schema_change.sql
```diff
@@ -0,0 +1,138 @@
...
+  {%- set remove_from_target_arr = diff_columns(target_columns, source_columns) -%}
+  
+  -- Validates on_schema_change config vs. whether there are column differences
+  {% if on_schema_change=='append_new_columns' and add_to_target_arr == [] %}
+    
+    {{ 
+        exceptions.raise_compiler_error('append_new_columns was set, but no new columns to append. 
+              This can occur when columns are removed from the source dataset unintentionally.
+              Review the schemas in the source and target relations, and consider re-running with the --full-refresh option.') 
+    }}
```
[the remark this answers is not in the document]
matt-winkler: This occurs specifically when 1) `append_new_columns` is set AND 2) There are columns removed from the source (meaning we trip `schema_changed = True`). My thought was that adding an explicit error here would be helpful to give context into why that approach fails. An alternative may be uploading only the columns in the source that do exist in the destination.

core/dbt/include/global_project/macros/materializations/incremental/on_schema_change.sql
```diff
@@ -0,0 +1,138 @@
...
+
+  {% set result = [] %}
+  
+  {% for col in columns %}
+    {{ result.append(col.column) }}
+  {% endfor %}
+  
+  {{ return(result) }}
+
+{% endmacro %}
```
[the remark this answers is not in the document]
matt-winkler: Boom, `map` for the win.

core/dbt/include/global_project/macros/materializations/incremental/incremental.sql
matt-winkler: Great call - I tested this on Snowflake and works beautifully.

core/dbt/include/global_project/macros/materializations/incremental/on_schema_change.sql
matt-winkler: done

core/dbt/include/global_project/macros/adapters/common.sql
```diff
@@ -311,3 +311,35 @@
...
+    
+      alter {{ relation.type }} {{ relation }}
+          {% if add_columns %}
+             add 
+          {% endif %} 
+          {% for column in add_columns %}
+            column {{ column.name }} {{ column.data_type }}{{ ',' if not loop.last }}
+          {% endfor %}
+          
+          {{ ', ' if add_columns and remove_columns }}
```
entechlog, line 330: I did a test run with a column add and column drop, default__alter_relation_add_remove_columns got dispatched to take care of the alters and snowflake did not like the comma separated alters.

Here is the config and sql in database side 

```sql
alter table xyz
            add column TEST_COL2 character varying(250), 
            drop column TEST_COL1
```

```
{{ 
    config(
	  alias = 'xyz',
      materialized='incremental',
      incremental_strategy='merge',
      unique_key='record_id',
	  transient=false,
      full_refresh=false,
      on_schema_change='sync_all_columns'
  )
}}

```

core/dbt/include/global_project/macros/materializations/incremental/on_schema_change.sql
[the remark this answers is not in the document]
matt-winkler: On further testing, I found that and additional `| list` was needed

plugins/bigquery/dbt/include/bigquery/macros/materializations/incremental.sql
matt-winkler: Implemented this change on snowflake and rolling to other providers.

plugins/snowflake/dbt/include/snowflake/macros/materializations/incremental.sql
matt-winkler: this is done for snowflake and porting to the other adapters

matt-winkler:
@jtcohen6 Feeling about near the finish line here. I've added integration tests for postgres/snowflake/bigquery. The questions I have at this point: 

- Do you think an explicit integration test for the `fail` config option is needed? If so, are you aware of any existing examples of something like this, where there is no target model populated because of the failure?
- I added a configuration option (boolean) called `alter_column_types`, which can be applied when `sync_all_columns` is specified. My thought was to make this explicit for Bigquery users who are syncing schemas, because changing a data type is an expensive operation, then make the config options consistent across databases.

jtcohen6 (commented):
@matt-winkler Thanks for all the work cleaning this up and adding integration tests. This feels really, really close.

I took another look today, started by leaving some comments, ended up iterating locally for a bit and teasing apart a few pieces in order to really understand what each one is doing. I'm going to open a PR against your branch with some proposed changes, which we can take/leave/discuss.

Let's get this in for v0.21.0-b1.

jtcohen6:
@matt-winkler After you resolve conflicts, I think this is good to get into `develop` + v0.21.0-b1

core/dbt/include/global_project/macros/materializations/incremental/incremental.sql
```diff
@@ -28,22 +32,37 @@
...
+=======
   {% elif existing_relation.is_view or should_full_refresh() %}
+>>>>>>> develop
       {% set intermediate_relation = existing_relation.incorporate(path={"identifier": tmp_identifier}) %}
       {% set backup_relation = existing_relation.incorporate(path={"identifier": backup_identifier}) %}
 
       {% set build_sql = create_table_as(False, intermediate_relation, sql) %}
       {% set need_swap = true %}
       {% do to_drop.append(backup_relation) %}
+
```
jtcohen6, line 57: Looks like this got caught in the merge, and is causing the failure in `035_changing_relation_type_test` for Postgres + Redshift
```suggestion
{% elif trigger_full_refresh %}
      {#-- Make sure the backup doesn't exist so we don't encounter issues with the rename below #}
      {% set tmp_identifier = model['name'] + '__dbt_tmp' %}
      {% set backup_identifier = model['name'] + '__dbt_backup' %}
      {% set intermediate_relation = existing_relation.incorporate(path={"identifier": tmp_identifier}) %}
      {% set backup_relation = existing_relation.incorporate(path={"identifier": backup_identifier}) %}

      {% set build_sql = create_table_as(False, intermediate_relation, sql) %}
      {% set need_swap = true %}
      {% do to_drop.append(backup_relation) %}
```

jtcohen6 (approved):
Let's ship it :)

core/dbt/include/global_project/macros/materializations/incremental/helpers.sql
```diff
@@ -1,5 +1,6 @@
 
 {% macro incremental_upsert(tmp_relation, target_relation, unique_key=none, statement_name="main") %}
+
```
b-luu, line 3: @matt-winkler @jtcohen6 do you have context on why the existing "target" columns are used as `dest_columns` below ?
Why not use the "tmp" increment's own columns ?

Sorry, I stumbled on this PR because 1. I was trying to implement schema evolution on incremental runs as well and 2. I had issues with some of my increments not necessarily having all the columns of the existing "target".

So I've used tmp_relation instead in the line below in a custom materialization on my side.

If the reason is unclear, I'll open an issue to have a broader discussion on this topic.

core/dbt/include/global_project/macros/materializations/incremental/on_schema_change.sql
```diff
@@ -0,0 +1,164 @@
...
+     {% endif %}
+  
+  {% endif %}
+
+  {% set schema_change_message %}
+    In {{ target_relation }}:
+        Schema change approach: {{ on_schema_change }}
+        Columns added: {{ add_to_target_arr }}
+        Columns removed: {{ remove_from_target_arr }}
+        Data types changed: {{ new_target_types }}
```
b-luu, line 125: Warning: `remove_from_target_arr` and `new_target_types` are only set `if on_schema_change == 'sync_all_columns'`.

This should raise an exception `if on_schema_change == 'append_new_columns'`

b-luu (commented):
This is brillant !!

Thanks a lot for your work here @matt-winkler & @jtcohen6 !

(I needed this !)

I've mainly added a note in `sync_column_schemas` as I think the current implementation would break if configured in `append` mode.

Also a question (for context) in `incremental_upsert`, but I can open a new issue for that if preferred ?

core/dbt/include/global_project/macros/materializations/incremental/helpers.sql
matt-winkler: Hi @b-luu here are my thoughts on use of target columns for each scenario:

- If we `ignore` or `fail` in the case of a schema change (or leave off the `on_schema_change` entirely), we assume that the UPSERT is targeted towards the existing columns in the target table. Basically the thinking here is that the target schema *should not* be changing under these configuration options.
- If we `append_new_columns` or `sync_all_columns`, the column addition / syncing actions have already been performed by the time we reach this point in the code.
- I'd be interested to see more examples of the issue you describe @b-luu to see if we're missing something here; let us know!

@jtcohen6 Please comment if the above seems off

core/dbt/include/global_project/macros/materializations/incremental/on_schema_change.sql
matt-winkler: @b-luu The assumption here is that when someone only wants to append new columns, we don't need to be concerned with existing column data type changes, or columns that have been removed from the upstream model vs. the target incremental model. Those scenarios are handled with the `sync_all_columns` configuration option.

@jtcohen6 Please comment if the above seems off

PR #3396, opened 2021-05-26
Creating test workflow in Actions

leahwicz:
### Description

Starting to move our tests into Actions. I moved all the tests that were in Azure Pipelines over (all the windows tests) and then added Mac tests to everything except Postgres (some complexities there- I left a TODO). 

My plan is to let this run as a not required check for a week or 2 and then make this required and remove the Azure Pipelines tests

### Checklist

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -38,8 +38,42 @@ jobs:
     - name: 'Run unit tests'
       run: python -m tox -e py -- -v
       
-  PostgresIntegrationTest:
+  PostgresIntegrationTest-Windows:
     runs-on: 'windows-latest'
+    needs: UnitTest
+    steps:
+    - uses: actions/checkout@v2
+    - name: 'Install postgresql and set up database'
+      shell: pwsh
+      run: |
+          $serviceName = Get-Service -Name postgresql*
+          Set-Service -InputObject $serviceName -StartupType Automatic
+          Start-Service -InputObject $serviceName
+          & $env:PGBIN\createdb.exe -U postgres dbt
+          & $env:PGBIN\psql.exe  -U postgres -c "CREATE ROLE root WITH PASSWORD 'password';"
+          & $env:PGBIN\psql.exe  -U postgres -c "ALTER ROLE root WITH LOGIN;"
+          & $env:PGBIN\psql.exe  -U postgres -c "GRANT CREATE, CONNECT ON DATABASE dbt TO root WITH GRANT OPTION;"
+          & $env:PGBIN\psql.exe  -U postgres -c "CREATE ROLE noaccess WITH PASSWORD 'password' NOSUPERUSER;"
+          & $env:PGBIN\psql.exe  -U postgres -c "ALTER ROLE noaccess WITH LOGIN;"
+          & $env:PGBIN\psql.exe  -U postgres -c "GRANT CONNECT ON DATABASE dbt TO noaccess;"
+      
+    - name: Setup Python
+      uses: actions/setup-python@v2.2.2
+      with:
+        python-version: '3.7'
+        architecture: 'x64'
+        
+    - name: 'Install dependencies'
+      run: python -m pip install --upgrade pip && pip install tox
+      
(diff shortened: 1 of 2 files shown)


.github/workflows/tests.yml
```diff
@@ -0,0 +1,160 @@
+# This is a workflow to run our unit and integration tests for windows and mac
+
+name: Integration Tests
```
iknox-fa, line 3: NBD, but maybe change this to reflect we're also running unit tests here?

.github/workflows/tests.yml
```diff
@@ -0,0 +1,160 @@
...
+    steps:
+    - uses: actions/checkout@v2
+    - name: 'Install postgresql and set up database'
+      shell: pwsh
+      run: |
+          $serviceName = Get-Service -Name postgresql*
+          Set-Service -InputObject $serviceName -StartupType Automatic
+          Start-Service -InputObject $serviceName
+          & $env:PGBIN\createdb.exe -U postgres dbt
+          & $env:PGBIN\psql.exe  -U postgres -c "CREATE ROLE root WITH PASSWORD 'password';"
```
iknox-fa, line 53: I'm sure it's appropriately sandboxed, but I always cringe when I see a plain-text password.  Would it be possible to add this to secrets?

iknox-fa:
This looks great-- I've very excited to move towards a unified build system!  One thought (and I'm not sure if this is in scope here or not) but it might make sense to add the linter to all this.

.github/workflows/tests.yml
```diff
@@ -0,0 +1,158 @@
...
+    strategy:
+      matrix:
+        os: [windows-latest, ubuntu-latest, macos-latest]
+    runs-on: ${{ matrix.os }}
+    steps:
+    - uses: actions/checkout@v2
+    - name: Setup Python
+      uses: actions/setup-python@v2.2.2
+      with:
+        python-version: '3.7'
```
nathaniel-may: We probably want to set the python version in the matrix for 3.6, 3.7, and 3.8

.github/workflows/tests.yml
```diff
@@ -0,0 +1,158 @@
...
+        architecture: 'x64'
+        
+    - name: 'Install dependencies'
+      run: python -m pip install --upgrade pip && pip install tox
+      
+    - name: 'Run integration tests'
+      run: python -m tox -e py-postgres -- -v -n4
+      
+  # These three are all similar except secure environment variables, which MUST be passed along to their tasks,
+  # but there's probably a better way to do this!
```
nathaniel-may, line 90: probably but this is good enough for me for now.

nathaniel-may (changes requested):

.github/workflows/tests.yml
jtcohen6: +3.9, which is supported by `dbt-core==0.19.0`, and by all plugins in this repo as of v0.20

leahwicz:
> This looks great-- I've very excited to move towards a unified build system! One thought (and I'm not sure if this is in scope here or not) but it might make sense to add the linter to all this.

Did you just mean adding a linting step? That makes sense and I can do that- just copied over the step CircleCI was doing

.github/workflows/tests.yml
leahwicz: I think I'll save this for the next iteration on this file since I had 3.9 and it was failing so I need to dig into that. To keep this simple for the moment, I'll do just version 3.7 and then build upon it before actually depending on this

iknox-fa:
> > This looks great-- I've very excited to move towards a unified build system! One thought (and I'm not sure if this is in scope here or not) but it might make sense to add the linter to all this.
> 
> Did you just mean adding a linting step? That makes sense and I can do that- just copied over the step CircleCI was doing
Yep, that's what I meant!

iknox-fa (approved):
LGTM!

jtcohen6 (approved):
This looks awesome! Excited to be moving the center of CI gravity into a shared workflow

nathaniel-may (approved):
LGTM

PR #3417, opened 2021-06-02
Updating triggers to not have duplicate runs

leahwicz:
### Description

Using `pull_request_target` and `pull_request` is causing duplicate triggers. Removing `pull_request_target` since that is less secure of an option. The trade off is that for PR's from a fork we will have to manually approve they run the tests. If this starts to hinder contributors, we can go back and reconsider the trigger type.

I also added CI for a few more important branches for `push` 

### Checklist

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -4,17 +4,18 @@ name: dbt Tests
 
 # Triggers
 on:
-  # Triggers the workflow on push or pull request events but only for the develop branch
+  # Triggers the workflow on push or pull request events and also adds a manual trigger
   push:
-    branches: [ develop ]
+    branches: 
+      - 'develop' 
+      - '*.latest' 
+      - 'releases/*'
   pull_request:
     branches: 
       - 'develop' 
       - '*.latest' 
       - 'pr/*'
       - 'releases/*'
-  pull_request_target:  
-    branches: [ develop ]
   # Allows you to run this workflow manually from the Actions tab
   workflow_dispatch:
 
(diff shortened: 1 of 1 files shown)


jtcohen6:
> for PR's from a fork we will have to manually approve they run the tests

What does the manual approval process look like? Definitely ideal version is, the tests just run. If one of us has to press a button, that's still better than checking out a commit locally and pushing to a branch named `pr/1234`.

leahwicz:
> > for PR's from a fork we will have to manually approve they run the tests
> 
> What does the manual approval process look like? Definitely ideal version is, the tests just run. If one of us has to press a UI button, that's still better than checking out a commit locally and pushing to a branch named `pr/1234`.

@jtcohen6 It would be a button in the UI. We can make it automatic if we want to but we expose our Snowflake/Redshift/Bigquery test creds which I guess we have been doing anyway (looks like they are just for integration tests). It's a safer practice to do this but if we aren't too concerned with those test creds we can just make it automatic

nathaniel-may:
I'm ok with it being more permissive for now and we can reign it in when it becomes a problem. Mostly because we'll have to do more steps than this to actually protect ourselves so we might as well enjoy the convenience for now too.

jtcohen6:
> if we want to but we expose our Snowflake/Redshift/Bigquery test creds which I guess we have been doing anyway (looks like they are just for integration tests)

Right on—these are totally sandboxed instances, with no sensitive data, and aggressive resource limits to prevent any bad behavior

leahwicz:
@nathaniel-may @jtcohen6 ok I changed it to trigger for forks too so we won't need any manual intervention

jtcohen6 (approved):

PR #3432, opened 2021-06-04
Save doc file node references and use in partial parsing

### LINKED ISSUE #3425: Save information about doc references for partial parsing
If a document file changes when the partial parse flag is on, we can't identify the nodes that use that document. Save the information about referenced docs in the node, so that when the document changes we can schedule those nodes for re-parsing.

gshank:
resolves #3425

### Description

When doc references are processed in parsing, save the unique_id of the node with the reference in the 'nodes' list in the doc source file. When the doc is changed and processed during partial parse, it will use the list of nodes to re-parse and update the docs.

### Checklist

### DIFF
--- a/core/dbt/parser/manifest.py
+++ b/core/dbt/parser/manifest.py
@@ -255,8 +255,8 @@ class ManifestLoader:
             # load up the Lookup objects to resolve them by name, so the SourceFiles store
             # the unique_id instead of the name. Sources are loaded from yaml files, so
             # aren't in place yet
-            self.manifest.ref_lookup
-            self.manifest.doc_lookup
+            self.manifest.rebuild_ref_lookup()
+            self.manifest.rebuild_doc_lookup()
 
             # Load yaml files
             parser_types = [SchemaParser]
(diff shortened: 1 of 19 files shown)


core/dbt/parser/manifest.py
```diff
@@ -255,8 +255,8 @@ def load(self):
             # load up the Lookup objects to resolve them by name, so the SourceFiles store
             # the unique_id instead of the name. Sources are loaded from yaml files, so
             # aren't in place yet
-            self.manifest.ref_lookup
-            self.manifest.doc_lookup
+            self.manifest.rebuild_ref_lookup()
+            self.manifest.rebuild_doc_lookup
```
kwigley, line 259:
```suggestion
            self.manifest.rebuild_doc_lookup()
```

kwigley (approved):
I didn't test this, but code changes look good to me.

jtcohen6 (approved):
This works for me in local testing. Nice work!

Intuitively, I was expecting an approach whereby docs blocks are included in the `depends_on` of each node. I don't have a strong reason to prefer it, though, and I'm very happy with this approach as well.

core/dbt/parser/manifest.py
gshank: Good catch.

gshank:
I initially started using a docs list in the depends_on, but that involved a lot of changes (to tests particularly) and would be a schema change. Also we'd have to loop though ALL of the nodes, sources, docs, and macros looking for the depends_on.docs. Storing them in the docs file means that we can skip that.

jtcohen6:
> Storing them in the docs file means that we can skip that.

Makes sense to me!

PR #3460, opened 2021-06-15
Add minimal validation of schema file yaml prior to partial parsing

### LINKED ISSUE #3246: Support dots in model names

gshank:
resolves #3246

### Description

Partial parsing expects a few things from the dictionary contents of a schema yaml file, but validation hasn't been performed yet, so certain types of errors in the yaml would result in unhelpful errors, such as "key error". This adds validation that the value of the expected yaml keys is a list, that the list contains dictionaries, and that the dictionaries contain a 'name' key.

### Checklist

### DIFF
--- a/core/dbt/parser/read_files.py
+++ b/core/dbt/parser/read_files.py
@@ -3,7 +3,8 @@ from dbt.contracts.files import (
     FilePath, ParseFileType, SourceFile, FileHash, AnySourceFile, SchemaSourceFile
 )
 
-from dbt.parser.schemas import yaml_from_file
+from dbt.parser.schemas import yaml_from_file, schema_file_keys, check_format_version
+from dbt.exceptions import CompilationException
 from dbt.parser.search import FilesystemSearcher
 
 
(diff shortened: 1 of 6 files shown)


core/dbt/parser/read_files.py
```diff
@@ -3,7 +3,8 @@
     FilePath, ParseFileType, SourceFile, FileHash, AnySourceFile, SchemaSourceFile
 )
 
-from dbt.parser.schemas import yaml_from_file
+from dbt.parser.schemas import yaml_from_file, schema_file_keys, check_format_version
+from dbt.exceptions import CompilationException
```
jtcohen6, line 7: For later, not now: We may want a cleaner distinction between compile-time exceptions (Jinja rendering, SQL templating) and validation/parse-time exceptions. This would be one of the latter.

core/dbt/parser/read_files.py
```diff
@@ -17,11 +18,36 @@ def load_source_file(
...
-    if parse_file_type == ParseFileType.Schema:
-        source_file.dfy = yaml_from_file(source_file)
+    if parse_file_type == ParseFileType.Schema and source_file.contents:
+        dfy = yaml_from_file(source_file)
+        validate_yaml(source_file.path.original_file_path, dfy)
+        source_file.dfy = dfy
     return source_file
 
 
+# Do some minimal validation of the yaml in a schema file.
```
jtcohen6, line 28: Are we doing this validation on _all_ yaml files, or just the ones we've identified to be new/changed during partial parsing?

jtcohen6 (commented):
Before:
```
Encountered an error:
string indices must be integers
```
After:
```
Encountered an error:
Compilation Error
  The schema file at models/schema.yml is invalid because the value of 'models' is not a list
```
Definitely an improvement!

kwigley (approved):

core/dbt/parser/read_files.py
gshank: I thought about this for a while, and in theory we could get a way with just doing it on changed and new files, but then the validation would have to be done in the partial parsing code, which felt wrong, and also yaml files with these errors wouldn't get any benefit from the better errors. Also we'd have different error messages depending on whether or not somebody had changed partial parsing.

PR #3461, opened 2021-06-15
dispatch logic of new test materialization

swanderz:
fixes: #3465
allow custom adapters to override the core test select statement functionality

 @jtcohen6, https://github.com/dbt-msft/dbt-sqlserver/pull/141 is telling us that we'll need to dispatch some of the logic in the new test materialization. T-SQL doesn't like direct boolean comparisons as columns, you have to wrap them in a `CASE WHEN`

### Checklist

### DIFF
--- a/core/dbt/include/global_project/macros/materializations/test.sql
+++ b/core/dbt/include/global_project/macros/materializations/test.sql
@@ -1,3 +1,19 @@
+{% macro test_body(main_sql, fail_calc, warn_if, error_if, limit) -%}
+  {{ adapter.dispatch('test_body')(main_sql, fail_calc, warn_if, error_if, limit) }}
+{%- endmacro %}
+
+
+{% macro default__test_body(main_sql, fail_calc, warn_if, error_if, limit) -%}
+    select
+      {{ fail_calc }} as failures,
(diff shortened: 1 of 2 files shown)


core/dbt/include/global_project/macros/materializations/test.sql
```diff
@@ -1,3 +1,19 @@
+{% macro test_body(main_sql, fail_calc, warn_if, error_if, limit) -%}
```
jtcohen6, line 1: Let's name this macro something that doesn't start with prefix `test_`, since that's the prior indicator of a test-definition macro. How about: `get_test_sql`?

jtcohen6 (commented):
Thanks for testing the prerelease @swanderz!

As it is, you can override the entire `'test'` materialization on dbt-sqlserver / dbt-synapse. I agree, though, that if the only blocker is a teensy bit of syntax, we should wrap it in a macro and dispatch.

core/dbt/include/global_project/macros/materializations/test.sql
swanderz: done!

jtcohen6 (approved):
Thanks for being on top of it @swanderz! I'll cherry-pick onto `0.20.latest` after merging.

PR #3468, opened 2021-06-16
Feature/faster materializations with fewer transactions

### LINKED ISSUE #3112: dbt runs slowly on Redshift RA3 nodes due to excessive commits
### Describe the bug
This is similar to #2748, but is Redshift-specific. However, I think users of other databases would benefit from slightly faster runs.

The next-gen Redshift RA3 nodes seem to have very slow COMMITs, due to the fact that:
- Redshift only processes a single transaction/COMMIT at a time, and
- RA3 nodes write their commits to S3; this makes each COMMIT take about a second (🤮)

[Many dbt users](https://getdbt.slack.com/archives/CJARVS0RY/p1609958080328900) have reported excessively slow runs creating views on RA3 nodes. In my experience, creating views while running dbt in multiple threads can take upwards of 20 seconds of clock time (see logs below).

There are at least 4 transactions as part of the [view materialization](https://github.com/fishtown-analytics/dbt/blob/develop/core/dbt/include/global_project/macros/materializations/view/view.sql):
1. Drop the intermediate relation
2. Drop the temp relation
3. Create new view; swap names with old view
4. Drop old view

Best case, on RA3, this sequence will take 4 seconds per view; in multiple threads, as these sequences stack up in the transaction queue, it will take much longer (in clock time).

### Steps To Reproduce
On a Redshift cluster with RA3 nodes (I'm on RA3.xplus x4), execute `dbt run` in multiple threads in a project with many views.

### Expected behavior
Views should ideally take at most a second or two to be materialized. 

Above, the four transactions to create a view could be compressed into a single transaction. I don't think this would have other side-effects, as any one of these operations failing would cause the run step to fail anyway.

To achieve this, I think we would have to modify the materialization macro(s) so dropping the other relations happens inside the same commit. It may be sufficient to add an explicit BEGIN in the materialization.

### Potential workarounds
On dbt v0.18+, we can avoid rebuilding views using the new selection syntax. However, this introduces some complexity and the potential for deployment errors when views change.

As an alternative, we can use more ephemeral models and fewer views, if the views do not need to be exposed in other contexts; this can make debugging more difficult.

Here's a recent run in dbt cloud:
image

Investigating the logs, it's clear the bottleneck is committing the transaction. This is each statement to execute the `START view model analytics.logistics_ops_sectors` line above:
image

Source for logs is `svl_statementtext`:
```sql
with 
    statements as (

        select 
            *,
            (datediff(milliseconds, starttime, endtime)::float / 1000)::numeric(10, 3) as dur_s,
            (datediff(milliseconds, lag(endtime, 1) over (order by starttime), starttime)::float /
            1000)::numeric(10, 3) as wait_s

    from svl_statementtext
    where 
        -- matches specific prod run above
        starttime >= '2021-02-17 08:04:17.124262'
        and starttime <= '2021-02-17 08:04:44.216802'
        and pid = 17738
        and sequence = 0

    order by starttime asc

),
    totals as (

        select
            null::int, 
            null::int, 
            null::int, 
            null::varchar, 
            getdate()::timestamp, 
            null::timestamp, 
            null::int, 
            null::varchar,
            ' --------- TOTAL ---------'::varchar,
            sum(dur_s), 
            sum(wait_s)

        from statements

 ),
    unioned as (

         select * from statements
         union all
         select * from totals

)
select * from unioned order by starttime
```
### System information
- [ ] postgres
- [x] redshift
- [ ] bigquery
- [ ] snowflake

```
0.17.1
```

Windows, but reproduces on dbt cloud
3.7.9

tconbeer:
resolves #3112 

### Description

This PR updates the view and table materializations in the `core/.../global_project` to check the relation cache for `__dbt_tmp` and `__dbt_backup` relations instead of blindly trying to drop those relations. The additional transactions required to drop those temp relations slows down dbt runs, especially on Redshift/RA3, where commits are slow.

I've also removed unused code from those same files (`set exists_as_table = ...` etc).

Unit and pg integration tests pass on my machine; I've added to an existing integration test for this specific edge case of preexisting temp relations.

This PR should not affect Snowflake or BQ, since those adapters have their own materializations. 

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,9 @@ Contributors:
 
 ## dbt 0.20.0 (Release TBD)
 
+### Under the hood
+- Improve view and table materialization performance by checking relational cache before attempting to drop temp relations ([#3112](https://github.com/fishtown-analytics/dbt/issues/3112), [#3468](https://github.com/fishtown-analytics/dbt/pull/3468))
+
 ## dbt 0.20.0rc1 (June 04, 2021)
 
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,18 +7,17 @@
 
 ### Under the hood
 
+- Improve default view and table materialization performance by checking relational cache before attempting to drop temp relations ([#3112](https://github.com/fishtown-analytics/dbt/issues/3112), [#3468](https://github.com/fishtown-analytics/dbt/pull/3468))
+
 Contributors:
 - [@kostek-pl](https://github.com/kostek-pl) ([#3236](https://github.com/fishtown-analytics/dbt/pull/3408))
+- [@tconbeer](https://github.com/tconbeer) [#3468](https://github.com/fishtown-analytics/dbt/pull/3468))
 
 ## dbt 0.20.0 (Release TBD)
 
 ### Fixes
 
 - Handle quoted values within test configs, such as `where` ([#3458](https://github.com/fishtown-analytics/dbt/issues/3458), [#3459](https://github.com/fishtown-analytics/dbt/pull/3459))
-
-### Under the hood
-
-- Improve default view and table materialization performance by checking relational cache before attempting to drop temp relations ([#3112](https://github.com/fishtown-analytics/dbt/issues/3112), [#3468](https://github.com/fishtown-analytics/dbt/pull/3468))
 
 ## dbt 0.20.0rc1 (June 04, 2021)
 
--- a/core/dbt/include/global_project/macros/materializations/incremental/incremental.sql
+++ b/core/dbt/include/global_project/macros/materializations/incremental/incremental.sql
@@ -6,6 +6,22 @@
   {% set target_relation = this.incorporate(type='table') %}
   {% set existing_relation = load_relation(this) %}
 
+  {% set tmp_identifier = model['name'] + '__dbt_tmp' %}
+  {% set backup_identifier = model['name'] + "__dbt_backup" %}
+
+  -- the intermediate_ and backup_ relations should not already exist in the database; get_relation
+  -- will return None in that case. Otherwise, we get a relation that we can drop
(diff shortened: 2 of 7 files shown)


CHANGELOG.md
```diff
@@ -16,6 +16,10 @@ Contributors:
 
 - Handle quoted values within test configs, such as `where` ([#3458](https://github.com/fishtown-analytics/dbt/issues/3458), [#3459](https://github.com/fishtown-analytics/dbt/pull/3459))
 
+### Under the hood
+
+- Improve default view and table materialization performance by checking relational cache before attempting to drop temp relations ([#3112](https://github.com/fishtown-analytics/dbt/issues/3112), [#3468](https://github.com/fishtown-analytics/dbt/pull/3468))
```
jtcohen6, line 21: Could you move this up to `v0.21`, and add yourself to the list of Contributors there?

This is a net-new change, so I'd prefer to have it in the next version, rather than trying to sneak it in between v0.20.0rc1 and final. We're going to be disciplined about keeping the v0.21 scope narrow and putting out prereleases early & often.

jtcohen6 (commented):
This is looking great, @tconbeer. Thanks for taking it on, and for adding the test case.

For consistency's sake, let's also update the code in the default `incremental` materialization:
https://github.com/fishtown-analytics/dbt/blob/9faa019a0780132c0cd908c1b6b875edc7f041f5/core/dbt/include/global_project/macros/materializations/incremental/incremental.sql#L25-L26
It's less pressing there than in `table` and `view`—in practice, it should only be relevant for full-refresh runs—but I'd like to maintain as much coordination as possible among code that the three materializations have in common.

CHANGELOG.md
tconbeer: totally. Didn't know where to put this, 0.21 makes more sense.

jtcohen6 (approved):
LGTM @tconbeer! Thank you so much for the contribution :)

Issue #3005, opened 2021-01-11
Move starter project for `init` task into dbt repo


jtcohen6:
Rather than storing the starter project cloned by `dbt init` in a separate repository ([`dbt-starter-project`](https://github.com/fishtown-analytics/dbt-starter-project)), we should store that code directly in the main `dbt` repo.

Originally, we stored the starter project separately so that we could make changes to it more frequently than we could cut new releases of dbt. Instead, over the past four years, we've cut many more releases to dbt than we have merged updates to the starter project :)

This helps us avoid annoying behavior when a specific dbt version requires a specific breaking change to the starter project (e.g. https://github.com/fishtown-analytics/dbt-starter-project/pull/8, https://github.com/fishtown-analytics/dbt-starter-project/issues/10).

We could take one of two implementation approaches:
1. Store starter project files in the https://github.com/fishtown-analytics/dbt repository, `dbt init` works more or less the same by cloning from the repo @ the tag of the current dbt version
2. Ship starter project files _within_ each version of dbt, rewrite `dbt init` to just copy files (no cloning from remote)

### Questions
- Should we allow adapter plugins to override / reimplement the starter project? We already enable them to provide a sample profile for first-time users, which enables `dbt init --adapter [adapter]` (mostly: #2814).

### Describe alternatives you've considered
- Continue maintaining separate repo with starter project. Develop more rigorous plan for versions + tags.

leahwicz:
* IDE does not use the `init` command
   * Check what their `init` command does vs what we do 
   * Why are they doing that?

ssassi:
Have you considered using https://github.com/cookiecutter/cookiecutter?

jtcohen6:
@ssassi Yes, we were just talking about it over in #3474 (which was just merged, and really closed this issue). I think there's a lot `cookiecutter` could get us. It's one step beyond what `dbt init` needs to do today, but if/as we add more complexity, we may decide it's worth the switch.

Issue #3361, opened 2021-05-16
Split up adapters to encourage adapter-specific installation


jtcohen6:
Today, `pip install dbt` + `brew install dbt` actually install four python packages and their dependencies, including `dbt-core`:
- `dbt-postgres`
- `dbt-redshift`
- `dbt-snowflake`
- `dbt-bigquery`

In [our docs](https://docs.getdbt.com/docs/available-adapters), we now encourage users to install the specific adapter plugin they plan to use whenever possible, to avoid possible installation issues caused by another adapter plugin that they're not even using, but which is included among the four above.

That recommendation is backed up by each of those individual plugins being available on PyPi today. To further ease the path, we should consider creating and maintaining:
- Separate Homebrew installations for specific plugins
- Separate Docker images for specific plugins

At the same time, we should reinvest in a "just install dbt" experience, especially for the benefit of newer users who want to take it for a spin before knowing the exact adapter they plan to use. In my mind, the ideal would be an interactive experience that asks users which adapters they want to install (perhaps recommending Postgres by default), and which offers detailed tips for getting started, similar to current `dbt init` command. We would then repurpose `pip install dbt`/`brew install dbt` to offer that experience, in place of the bundled installation that exists today.

jtcohen6:
Related question: Should we also move the four plugins included this repository into four separate repositories, ahead of v1.0?

gshank:
I think it makes sense to spin off redshift, snowflake, and bigquery into their own repositories. I'm less certain about Postgres. The postgres plugin has a lot of the default behavior, so it feels like we'd end up having to prereq it for the other plugins. Also the dbt repository by itself wouldn't be functional, which feels kind of wrong.

Separate Docker images feels reasonable. 

I'm not very familiar with how customizable the pip install and brew install processes are, but more seamless installation process is certainly a worthwhile goal.

leahwicz:
Work we would need to do:
* Create 3-4 new repos and pull the code into each
* Split out the integration tests for those
* Create docker images for each
* Split out the Homebrew install

* Hardest part- spitting up the integration tests
* Not too hard - different Docker images

Open questions:
* What about postgres?
* Do we rename `dbt` to `dbt core`? 
* How would install work?

leahwicz:
Maybe we can split this up into smaller issues/goals

jtcohen6:
We're going to give this a go in two weeks. Latest thinking:
- **Codebases:**
[...]

PR #3474, opened 2021-06-18
Move starter project into `dbt` repo

leahwicz:
Addresses issue #3005 

### Description
I moved the starter_project from its own repo into this repo. I copied the starter project from the `dbt-yml-config-version-2` tag.

This will stop cloning the project from the other repo and instead will package the starter project within the dbt release and then copy it to the given directory. 

### Checklist


leahwicz:
@jtcohen6 - do we need a changelog for changes that aren't customer facing?

@kwigley - can we test packaging this up for distribution to see if it works? I just really want to check that the files are included.

core/setup.py
```diff
@@ -38,17 +38,7 @@ def read(fname):
...
-            'include/index.html',
-            'include/global_project/dbt_project.yml',
-            'include/global_project/docs/*.md',
-            'include/global_project/macros/*.sql',
-            'include/global_project/macros/**/*.sql',
-            'include/global_project/macros/**/**/*.sql',
-            'py.typed',
-        ]
-    },
+    include_package_data = True,
```
leahwicz, line 41: A couple articles I read said to use `include_package_data` instead of `package_data` with the MANIFEST.in file so that's why I made this change. You don't have to list the files then.

Example:
https://newbedev.com/how-include-static-files-to-setuptools-python-package

gshank (commented):
LGTM!

core/dbt/include/starter_project/dbt_project.yml
```diff
@@ -0,0 +1,38 @@
...
+analysis-paths: ["analysis"]
+test-paths: ["tests"]
+data-paths: ["data"]
+macro-paths: ["macros"]
+snapshot-paths: ["snapshots"]
+
+target-path: "target"  # directory which will store compiled SQL files
+clean-targets:         # directories to be removed by `dbt clean`
+    - "target"
+    - "dbt_modules"
```
jtcohen6, line 25: Nit: can we make these 2-space tabs? That's the [style guide](https://github.com/fishtown-analytics/corp/blob/master/dbt_style_guide.md#yaml-style-guide)

core/dbt/include/starter_project/dbt_project.yml
```diff
@@ -0,0 +1,38 @@
...
+# Full documentation: https://docs.getdbt.com/docs/configuring-models
+
+# In this example config, we tell dbt to build all models in the example/ directory
+# as tables. These settings can be overridden in the individual model files
+# using the `{{ config(...) }}` macro.
+models:
+  my_new_project:
+      # Applies to all files under models/example/
+      example:
+          materialized: view
```
jtcohen6, line 38: As above with 2-space tabs, and let's also add the (optional) [+ prefix](https://docs.getdbt.com/reference/resource-configs/plus-prefix), which serves as a decorator for configs. (Not strictly necessary for `materialized`, which is only a string, but would be required if it were a dictionary.)
```suggestion
    # Config indicated by + and applies to all files under models/example/
    example:
      +materialized: view
```

core/dbt/include/starter_project/models/example/schema.yml
```diff
@@ -0,0 +1,21 @@
+
```
jtcohen6, line 1: 2-spaced tabs here as well :)

core/dbt/include/starter_project/dbt_project.yml
```diff
@@ -0,0 +1,38 @@
+
+# Name your project! Project names should contain only lowercase characters
+# and underscores. A good package name should reflect your organization's
+# name or the intended use of these models
+name: 'my_new_project'
+version: '1.0.0'
+config-version: 2
+
+# This setting configures which "profile" dbt uses for this project.
+profile: 'default'
```
jtcohen6, line 10: I would _love_ if we could figure out a way to automatically populate `name` and `profile` with the `project_name` argument supplied to `dbt init`. Right now, that `project_name` arg is just used to name the file directory, but not the actual project name. Pretty confusing! Also, we tend to discourage using a profile named `default`, and yet here we are...

I know all we're really doing here is `shutil.copytree`. Is there any sane way to try editing the files after copying?

For reference, I was just triaging a related issue, so I mentioned this over there as well: https://github.com/fishtown-analytics/dbt/issues/3462#issuecomment-865314756

core/setup.py
jtcohen6: I really like it! We did the same in dbt-spark a few months ago: https://github.com/fishtown-analytics/dbt-spark/pull/151. To quote a wise person, "This is much easier."

jtcohen6 (commented):
Nice work!! I just took this for a spin, the new approach seems great.

- Did we lose the model files (`models/example/*.sql`) themselves?
- Have you checked in with our Experience colleagues about whether this approach is something they could leverage in the future (rather than reimplementing)?
- I left a comment around some related-but-separate improvements to `init`. It shouldn't be considered blocking to this PR, but if we could find a neat way to do it, it will have been a long time coming.
- While taking this for a spin, I also played around with a fix to https://github.com/fishtown-analytics/dbt/issues/2814. I'll open a separate PR for that based off this branch.

core/dbt/include/starter_project/README.md
```diff
@@ -0,0 +1,15 @@
...
+
+Try running the following commands:
+- dbt run
+- dbt test
+
+
+### Resources:
+- Learn more about dbt [in the docs](https://docs.getdbt.com/docs/introduction)
+- Check out [Discourse](https://discourse.getdbt.com/) for commonly asked questions and answers
+- Join the [chat](http://slack.getdbt.com/) on Slack for live discussions and support
```
jtcohen6, line 13:
```suggestion
- Join the [chat](https://community.getdbt.com/) on Slack for live discussions and support
```

core/dbt/include/starter_project/dbt_project.yml
gshank: We have a lot of expertise with jinja templates :). Perhaps we could make a dbt_project jinja template and substitute the project name. Or just do a python substitution if it's simpler. I guess jinja would only make sense if there are other ways we could leverage it to customize the project file.

jtcohen6: Heh, you're not wrong. Back in the day, we did a lot of customization for client projects over here: https://github.com/fishtown-analytics/dbt-init

The purpose there was cross-applying some pretty opinionated configs for a given adapter. I think, for our purposes, we shouldn't go much further beyond `name` and `profile`.

leahwicz:
>     * Did we lose the model files (`models/example/*.sql`) themselves?

🤦‍♀️ I had my global .gitignore file set to filter out sql and I missed this so good catch

core/dbt/include/starter_project/dbt_project.yml
leahwicz: I'll take a crack at doing this in another PR to follow this one up. So we want `project_name` to be the `name` and `profile` field in the dbt_project.yml file (just to verify)?

core/setup.py
kwigley: 🙏

core/dbt/include/starter_project/dbt_project.yml
jtcohen6: Yes, that's right!

kwigley (approved):
👍 looks good!

I confirmed that the source and wheel distributions include the starter project files. In the `dbt/core` directory, I ran `python setup.py sdist bdist_wheel` which creates a source (`*.tar.gz`) and a wheel (`*.whl`) distribution. I installed these in separate temporary virtualenvs and confirmed the project existed and  `init` command worked for both.

I think if we want a comprehensive project initialization, we could consider using a scaffolding tool like [cookiecutter](https://cookiecutter.readthedocs.io/en/1.7.2/cookiecutter.html#module-cookiecutter.main). This might be overkill for the current starter project thought.

jtcohen6:
> I think if we want a comprehensive project initialization, we could consider using a scaffolding tool like [cookiecutter](cookiecutter). This might be overkill for the current starter project thought.

We've had this idea before: https://github.com/fishtown-analytics/dbt-init/issues/31. I'm open to it! Agree that it's probably not necessary for now, but might be a good idea if we want `init` to do something more complex than just copy files (https://github.com/fishtown-analytics/dbt/issues/3462). I also like how interactive `cookiecutter` is with setup Q&As, it inspired some of my thinking about an interactive install experience over in #3361.

PR #3488, opened 2021-06-22
Feature/dbt build

### LINKED ISSUE #2743: Generalized `dbt build` command
See also: #1054, #1227, #2234, [this comment](https://github.com/fishtown-analytics/dbt/issues/2203#issuecomment-611736503)

### Describe the feature
Each dbt node-resource type has a task-command associated with it:
* models = `dbt run`
* tests = `dbt test`
* seeds = `dbt seed`
* snapshots = `dbt snapshot`
* sources = `dbt source snapshot-freshness`

Additionally, there could be a generalized command `dbt build`<sup>1</sup> that would step through a DAG of multiple resource types and "build" them accordingly.

What would this look like? I imagine an argument syntax similar to [`dbt ls`](https://docs.getdbt.com/reference/commands/list/), i.e.
```
dbt build --select ... --exclude ... --resource-type ...
```

<sup>1</sup> _name subject to change, though for the ultimate command of the data **build** tool, it'd be hard to think of one more apropos..._

### Example

Let's imagine we had `model_a` that depends on a source (`my_source.table`) and a seed (`my_seed`), a snapshot (`my_snapshot`) of `model_a`, and then `model_b` which selected from `my_snapshot`. Of course, we also have tests on many of them. Roughly:

```
my_source.table --> my_seed --> model_a --> my_snapshot --> model_b
```

Within a single invocation, `dbt build` would go through motions analogous to running the following dbt commands. It would only proceed to the next numerical steps if all upstream steps succeed:

1a. `dbt seed my_seed`
1b. `dbt source snapshot-freshness --select my_source.table`
2a. `dbt test --models my_seed`
2b. `dbt test --models source:my_source.table`
3. `dbt run --models model_a`
4. `dbt test --models model_a`
5. `dbt snapshot --select my_snapshot`
6. `dbt test --models my_snapshot`
7. `dbt run --models model_b`
8. `dbt test --models model_b`

### Complexities

* Some of these tasks are already DAG aware (`run`, `test`, `snapshot`), some are not (`seed`, `snapshot-freshness`)
* Commands support several different flags
    * How to expose when a flag is being used, and when it isn't?
    * What about same-named flags that do subtly different things across commands? e.g. `dbt run --full-refresh` vs. `dbt seed --full-refresh`
* Node types are _just about_ 1:1 with task types, though `dbt test` almost feels like an exception. Technically, `dbt test` operations on test nodes, but other node types can be passed into its selection syntax, with selector expansion as the last step, so it "feels" like you're testing a model or a snapshot. (Edit: this behavior may someday change.)
* This risks a lot of our existing intuitions that come from having resource types nicely delineated. Put differently: what if it all just falls apart?
    * What if it works _so_ well that 90% of dbt deployments are just `dbt build`? Should we be weary of creating one command to rule them all?

* Doing a more particularized version of this, e.g. `dbt run+test` (as outlined in linked issues)
* Not doing this at all, and leaving the federation of one resource type = one command/invocation. Is this a _good_ abstraction that we should fight to keep?

### Who will this benefit?
* Bigger, more complex projects who want to run subsets of different resource types. Today, that can only be accomplished through complex selection syntax leveraging tags. YAML selectors improves this somewhat, but they're not the answer.
* Projects with snapshots that participate in the middle of the DAG
* Deployments that want to test upstream models before running downstream models, so as to alert earlier and save compute time/$$ in the event of failure

iknox-fa:
resolves #2743

### Description
Adds a `dbt build` command that does "all the things" for a given project.  More specifically:

- models invoke logic from `dbt run`
- tests invoke logic from `dbt test`
- seeds invoke logic from `dbt seed`
- snapshots invoke logic from `dbt snapshot`

This is all handled in dependency order

### Checklist
 - [] I have updated the `CHANGELOG.md` and added information about my change to the "dbt next" section.


gshank:
Are there tests?

iknox-fa:
> Are there tests?

Yep.  Trying to get them merged into this branch now but git is being a pain.  I'll ping you when they get in there.

Issue #3452, opened 2021-06-10
[Tracking] `dbt build` command


leahwicz:
Tracking issue for the generalized `dbt build` command outlined in issue https://github.com/fishtown-analytics/dbt/issues/2743

- [x] [RAT](https://github.com/fishtown-analytics/dbt/issues/3415)
- [x] Demo
- [x] Ensure produced artifacts (`manifest.json`, `run_results.json`, `sources.json`) make sense
- [ ] #3597
- [x] Exclude `source snapshot-freshness` for now
- [ ] #3596
- [ ] #3595
- [x] Additional resource types (~1 left)
- [x] #3598

~~ - [ ] Refactor cmd line library ~~ --> still a smart thing to do, but not needed for this effort

jtcohen6:
> Parity for flags

Ideal is that `build` supports the superset of all flags. If a flag is supported by any one of the sub-tasks, it should be supported by `build`, too. Tasks that can use the flag, observe it; tasks that can't use it, ignore it.

> Ensure produced artifacts (run_results.json + sources.json) make sense

Should `dbt build` also include `dbt docs generate` (and produce `catalog.json` as well)? This one is funny, I'd want to think

jtcohen6:
I'm going to close this tracking issue, in favor of keeping #3595, #3596, #3597 open and tagged for the next cycle epic + release milestone. In the future, let's open all tracking issues as epics, which enables us to nest an epic for a specific work-area (e.g. `dbt build`) inside larger release + cycle epics.

PR #3490, opened 2021-06-22
dbt `build`

### LINKED ISSUE #2743: Generalized `dbt build` command

iknox-fa:
resolves #2743

### Description
Adds a `dbt build` command that does "all the things" for a given project.  More specifically:

- models invoke logic from `dbt run`
- tests invoke logic from `dbt test`
- seeds invoke logic from `dbt seed`
- snapshots invoke logic from `dbt snapshot`

This is all handled in dependency order.  ~~Test failures block further downstream processing.~~
Test failure blocking was de-scoped due to testing issues.

### Checklist


core/dbt/graph/queue.py
```diff
@@ -37,7 +37,8 @@ def __init__(
         # this lock controls most things
         self.lock = threading.Lock()
         # store the 'score' of each node as a number. Lower is higher priority.
-        self._scores = self._calculate_scores()
+        # TODO: incorporate _include_in_cost (or remove dead code, still needed?)
+        self._scores = {y: x for x, y in enumerate(nx.topological_sort(self.graph))}
```
jtcohen6, line 41: This is code we haven't touched for some time, perhaps since https://github.com/fishtown-analytics/dbt/pull/1157.

https://blog.getdbt.com/how-we-made-dbt-runs-30--faster/

> This is a key insight, and it’s the most meaningful performance optimization in dbt today. Unfortunately, the logic required to intelligently parallelize the execution of the DAG is complicated, and we can’t just throw topological sort at this problem. To understand why, consider the output of the topological sort algorithm. Suppose we had two threads ready to start executing models. The first thread can definitely get started on model A, but what should the second thread do? It can’t process model K until model A has completed, but that information isn’t present in the topological ordering of the DAG. Topological sort just isn’t applicable in a multi-threaded use case.

Does this still hold true?

core/dbt/main.py
```diff
@@ -377,6 +378,30 @@ def _build_init_subparser(subparsers, base_subparser):
     return sub
 
 
+def _build_build_subparser(subparsers, base_subparser):
+    sub = subparsers.add_parser(
+        'build',
+        parents=[base_subparser],
+        help='''
+        Runs the whole shebang!  @TODO: better description
```
jtcohen6, line 386: :)

jtcohen6 (commented):
Nice work so far!

> Test failures block further downstream processing.

Does this PR include the logic for skipping a downstream model, if a test on an upstream model fails? I didn't see that happen when I took for a spin locally.

core/dbt/task/build.py
```diff
@@ -0,0 +1,45 @@
...
+    """
+    The Build task processes all assets of a given process and attempts to 'build'
+    them in an opinionated fashion.  Every resource type outlined in RUNNER_MAP
+    will be processed by the mapped runner class.
+
+    I.E. a resource of type Model is handled by the ModelRunner which is imported
+    as run_model_runner.
+    """
+
+    # TODO: is this list complete?
```
nathaniel-may, line 23: Is this something we can answer now?

nathaniel-may (changes requested):
Looks good! I like that it ended up being a nice and concise change. What I can't quite tell from the code is where the ordering happens, and what happens when a test fails. Maybe we can add some comments to that effect?

nathaniel-may:
Do we have an integration test that hits all the types with the build command?

core/dbt/task/build.py
iknox-fa: I think so!  @jerco does this represent all of the node types that we would like to process when we run the build command?

core/dbt/graph/queue.py
iknox-fa: Updated!

iknox-fa:
> Do we have an integration test that hits all the types with the build command?

We do now :)

test/unit/test_registry_get_request_exception.py
```diff
@@ -6,4 +6,4 @@
 class testRegistryGetRequestException(unittest.TestCase):
     def test_registry_request_error_catching(self):
         # using non routable IP to test connection error logic in the _get function
-        self.assertRaises(RegistryException, _get, '', 'http://10.255.255.1')
+        self.assertRaises(RegistryException, _get, '', 'http://0.0.0.0')
```
iknox-fa, line 9: A little bonus update-- this should make this test a lot faster (0.0.0.0 doesn't even get as far as the DNS level before it fails routing)

core/dbt/task/build.py
jtcohen6: Yes! This is everything

core/dbt/contracts/graph/parsed.py
```diff
@@ -592,7 +592,8 @@ class ParsedSourceDefinition(
     UnparsedBaseNode,
     HasUniqueID,
     HasRelationMetadata,
-    HasFqn
+    HasFqn,
```
leahwicz, line 595: I'm not a Python person, but this looks weird to me to add the comma and the newline. Is that a formatting thing?

core/dbt/graph/queue.py
```diff
@@ -56,30 +53,59 @@ def _include_in_cost(self, node_id: UniqueId) -> bool:
...
 
-    def _calculate_scores(self) -> Dict[UniqueId, int]:
-        """Calculate the 'value' of each node in the graph based on how many
-        blocking descendants it has. We use this score for the internal
-        priority queue's ordering, so the quality of this metric is important.
+    @staticmethod
+    def _grouped_topological_sort(
+        graph: nx.DiGraph,
+    ) -> Generator[List[str], None, None]:
+        """Topological sort of given graph that groups ties.
```
leahwicz, line 60: Nice comment block 💯

leahwicz (approved):

core/dbt/contracts/graph/parsed.py
iknox-fa: Yeah-- I'm starting to use an auto-formatter called Black which is opinionated.  Not sure how much I like it though.
:)

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 ## dbt 0.21.0
 
 ### Features
+- Add `dbt build` command to run Models, Tests, Seeds, and snapshots in DAG order. ([#2743] (https://github.com/dbt-labs/dbt/issues/2743))
```
jtcohen6, line 4: Nit, sorry:
```suggestion
- Add `dbt build` command to run models, tests, seeds, and snapshots in DAG order. ([#2743] (https://github.com/dbt-labs/dbt/issues/2743), [#3490] (https://github.com/dbt-labs/dbt/issues/3490))

File: core/dbt/task/build.py [hunk 8]
```diff
@@ -0,0 +1,43 @@
...
+from .snapshot import SnapshotRunner as snapshot_model_runner
+from .seed import SeedRunner as seed_runner
+from .test import TestRunner as test_runner
+
+from dbt.graph import ResourceTypeSelector
+from dbt.exceptions import InternalException
+from dbt.node_types import NodeType
+
+
+class BuildTask(CompileTask):
```
jtcohen6, line 13: Love how simple & straightforward this code ended up being!

jtcohen6 (approved):
### substance
- Artifact output looks great! (I spent a few minutes visually inspecting `run_results.json`.) Given that tests can now be skipped (if their upstream model fails), for completeness' sake, we should probably adjust the [`TestStatus`](https://github.com/dbt-labs/dbt/blob/f460d275baa4df8cf8713f27e198a3577d6ea71c/core/dbt/contracts/results.py#L76) contract accordingly.
- Ultimately, I'd like `dbt build` to use `--select` rather than `--models`. I'd actually like to make that switch across the board v0.21 (#3210). I don't want to block this PR any more than necessary, so as long as that feels like a simple thing to come later, I'm happy moving forward with it as is.

### process
- Do we want new issues for test-fail-block-downstream and RPC/CLI flag parity, or to keep using #3452 for this?
- Could we get a WIP draft PR for the test-blocking feature (since we got so close, and may just need to debug a few integration tests)?

jtcohen6 (changes requested):
Sorry, I should've caught this above. I was just testing this a bit more, and I ran into an unexpected behavior. If I cause one model to fail (by introducing a syntax error), rather than continuing on and skipping downstream nodes, the error results in the end of the invocation—no more nodes run. This is the behavior I'd expect from `--fail-fast`, but not from a standard `build` invocation without supplying that flag.

Here's the log output of `dbt build` (without `--fail-fast`):
```
00:45:53 | 1 of 12 ERROR creating view model dbt_jcohen.stg_trello_boards....... [ERROR in 2.43s]
Encountered an error:
00:45:55 | 5 of 12 OK created view model dbt_jcohen.stg_trello_members.......... [SUCCESS 1 in 4.54s]
Runtime Error
  Database Error in model stg_trello_boards (models/staging/trello/stg_trello_boards.sql)
    001003 (42000): SQL compilation error:
    syntax error line 22 at position 4 unexpected 'a'.
    parse error line 27 at position 4 near '<EOF>'.
    syntax error line 26 at position 0 unexpected 'select'.
    compiled SQL at target/run/fishtown_internal_analytics/models/staging/trello/stg_trello_boards.sql
00:45:55 | 4 of 12 OK created view model dbt_jcohen.stg_trello_cards............ [SUCCESS 1 in 4.54s]
00:45:55 | 3 of 12 ERROR creating view model dbt_jcohen.stg_trello_card_members. [ERROR in 4.55s]
00:45:55 | 2 of 12 ERROR creating view model dbt_jcohen.stg_trello_card_labels.. [ERROR in 4.55s]
```
Here's `dbt build --fail-fast`:
```
00:46:13 | 1 of 12 ERROR creating view model dbt_jcohen.stg_trello_boards....... [ERROR in 2.49s]
00:46:15 | CANCEL query model.fishtown_internal_analytics.stg_trello_boards..... [CANCEL]
00:46:15 | CANCEL query model.fishtown_internal_analytics.stg_trello_card_labels [CANCEL]
00:46:15 | CANCEL query model.fishtown_internal_analytics.stg_trello_card_members [CANCEL]
00:46:15 | CANCEL query model.fishtown_internal_analytics.stg_trello_cards...... [CANCEL]
00:46:15 | CANCEL query model.fishtown_internal_analytics.stg_trello_members.... [CANCEL]
00:46:15 | 2 of 12 OK created view model dbt_jcohen.stg_trello_card_labels...... [SUCCESS 1 in 4.73s]
00:46:15 | 3 of 12 OK created view model dbt_jcohen.stg_trello_card_members..... [SUCCESS 1 in 4.73s]
00:46:16 | 5 of 12 ERROR creating view model dbt_jcohen.stg_trello_members...... [ERROR in 4.92s]
00:46:16 | 4 of 12 ERROR creating view model dbt_jcohen.stg_trello_cards........ [ERROR in 4.94s]

Database Error in model stg_trello_boards (models/staging/trello/stg_trello_boards.sql)
  001003 (42000): SQL compilation error:
```

CHANGELOG.md
iknox-fa: Fixed!

kwigley (approved):
looks good!

nathaniel-may (approved):
🐑

jtcohen6 (approved):
We had a chance to catch up on the behavior above and discuss next steps. For now, I'm conditionally approving this PR, with the understanding that we'll be opening new issues to track the pieces that are missing:
- [critical] On node failure: correctly write `run_results.json` (missing today); parity with `dbt run` for logging (showing `skipped` + run summary)
- [highly desirable] Tests block other node children, test failures skip those children
- [completeness] CLI flag parity with existing commands
- [completeness] RPC `build` method

@iknox-fa Could you please link to those issues here after merging?

Excited to get the first rough cut of `dbt build` in for v0.21.0b1!

PR #3495, opened 2021-06-24
Update project load tracking to include experimental parser info

### LINKED ISSUE #3438: Update project load tracking for v0.20 parsing changes
### Describe the feature

Let's add fields to `track_project_load` so we can account for / disambiguate between:
- Whether partial parsing is enabled (we use `is_partial_parse_enabled` for this today)
- Total number of files in the project (we use `path_count` for this today)
- Number of files actually re-parsed by partial parsing (this is what `path_count` represents in v0.20rc1, we should keep these separate)
- Whether the experimental parser is enabled
- Number of files eligible to be parsed by experimental parser (just models today, but could be more in the future)
- Number of files actually parsed by experimental parser

https://github.com/fishtown-analytics/dbt/blob/abe8e839458d0288f76cb6950e79c77e3b1627cc/core/dbt/parser/manifest.py#L599-L617

kwigley:
resolves #3438 

### Description

I reset this branch several times while trying to add this data, this is what I landed on that doesn't involve refactoring pieces of partial parsing. Will pull logic in to `0.20.latest` branch after approval.

- `parsed_path_count` -> path_count can represent the size of the project and amount of files eligible to be parsed while parsed_path_count will be the actual amount of files parsed
- `is_static_analysis_enabled` -> set if flag for experimental parser is turned on
- `static_analysis_path_count` -> number of files eligible to be parsed by experimental parser
- `static_analysis_parsed_path_count` -> number of files actually parsed by experimental parser

### Checklist

### DIFF
--- a/core/dbt/parser/manifest.py
+++ b/core/dbt/parser/manifest.py
@@ -31,7 +31,7 @@ from dbt.parser.read_files import read_files, load_source_file
 from dbt.parser.partial import PartialParsing
 from dbt.contracts.graph.compiled import ManifestNode
 from dbt.contracts.graph.manifest import (
-    Manifest, Disabled, MacroManifest, ManifestStateCheck
+    Manifest, Disabled, MacroManifest, ManifestStateCheck, ParsingInfo
 )
 from dbt.contracts.graph.parsed import (
     ParsedSourceDefinition, ParsedNode, ParsedMacro, ColumnInfo, ParsedExposure
@@ -210,12 +210,15 @@ class ManifestLoader:
                 # files are different, we need to create a new set of
                 # project_parser_files.
                 project_parser_files = partial_parsing.get_parsing_files()
-                self.manifest = self.saved_manifest
                 self.partially_parsing = True
+
+            self.manifest = self.saved_manifest
+
+        if self.manifest._parsing_info is None:
+            self.manifest._parsing_info = ParsingInfo()
 
         if skip_parsing:
             logger.info("Partial parsing enabled, no changes found, skipping parsing")
-            self.manifest = self.saved_manifest
         else:
             # Load Macros
             # We need to parse the macros first, so they're resolvable when
@@ -304,10 +307,11 @@ class ManifestLoader:
             self.process_sources(self.root_project.project_name)
             self.process_refs(self.root_project.project_name)
             self.process_docs(self.root_project)
+
+            # update tracking data
             self._perf_info.process_manifest_elapsed = (
                 time.perf_counter() - start_process
             )
-            # update tracking data with static parsing info
             self._perf_info.static_analysis_parsed_path_count = (
                 self.manifest._parsing_info.static_analysis_parsed_path_count
             )
--- a/core/dbt/parser/manifest.py
+++ b/core/dbt/parser/manifest.py
@@ -71,7 +71,7 @@ DEFAULT_PARTIAL_PARSE = False
 class ParserInfo(dbtClassMixin):
     parser: str
     elapsed: float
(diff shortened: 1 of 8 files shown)


core/dbt/parser/manifest.py
```diff
@@ -614,6 +629,9 @@ def track_project_load(self):
             "is_partial_parse_enabled": (
                 self._perf_info.is_partial_parse_enabled
             ),
+            "is_static_analysis_enabled": flags.USE_EXPERIMENTAL_PARSER,
```
jtcohen6, line 632: Could we add `is_static_analysis_enabled` to the output of `dbt parse` (`target/perf_info.json`) as well?

core/dbt/parser/models.py
```diff
@@ -22,8 +22,9 @@ def get_compiled_path(cls, block: FileBlock):
         return block.path.relative_path
 
     def render_update(
-        self, node: IntermediateNode, config: ContextConfig
+        self, node: ParsedModelNode, config: ContextConfig
     ) -> None:
+        node.meta["is_statically_extractable"] = True
```
jtcohen6, line 27: I don't think we can use `meta` for this. That's a property that users can set in yaml files ([docs](https://docs.getdbt.com/reference/resource-properties/meta)). If the user sets any yaml properties for the node, even if they don't set any `meta` properties, the node drops `{'is_statically_extractable': True}` during yaml parsing [here](https://github.com/fishtown-analytics/dbt/pull/3495/files#diff-a25a77362db78f4ddc0675bcbd4775e158631076c9918ce5b64cfcd9f6db6bceR268). The upshot is that we significantly undercount the number of statically extractable + extracted paths.

Is there a different preexisting, internal-use node property we could set here? Or another way to tag the node (without adding having to adjust the manifest schema)?

I'd also be open to adjusting the `v2` manifest schema so that, instead of just adding a standalone `created_at` property, we added a `parsing` dictionary (which could include `created_at`, `statically_extractable`, `statically_extracted`, total parse time, etc).

jtcohen6 (changes requested):
- Event looks good, I confirmed the data flows all the way through as expected
- The updates to `path_count` + `parsed_path_count` are right on
- We're undercounting for `static_analysis_path_count` + `static_analysis_parsed_path_count` because of how we're using the `node.meta` property right now. Let's find another way to do this

core/dbt/parser/models.py
gshank: One option might be to store the information in an attribute/dictionary that is filtered out in serialization so that we don't need to update the manifest schema. I couldn't do that with 'created_at' because it had to be in the partial_parse.msgpack file, but I think the information we're discussion here doesn't need to be stored.

I suspect that we're going to keep running into this "new attribute vs updating schema" issue, so we might want to think about something like the 'parsing' dictionary Jeremy proposes, maybe 'internal' or something, where we can put data that probably is not relevant except internally.

gshank: Oh, I see Kyle already is filtering out that info. Nevermind :)

gshank:
Did we decide not to track the path counts on a project/parser basis? I see the total path_count is saved from the number of files.

Also we had discussed storing whether partial parsing is actually running ('partially_parsing'). Is that something we want to save?

jtcohen6:
> Also we had discussed storing whether partial parsing is actually running ('partially_parsing'). Is that something we want to save?

@gshank I think we can infer this from whether `path_count != parsed_path_count`

jtcohen6 (approved):
Nice work! Numbers are looking right on.

gshank (approved):

PR #3498, opened 2021-06-26
Feature/support sql dot jinja

### LINKED ISSUE #3484: [Feature] Support `.jinja`, `.jinja2`, `.j2` file extensions for `.sql` files
### Describe the feature
Provide the option to allow `*.sql.jinja` files as alternative to `*.sql` in all places where DBT uses jinja in SQL (I believe that is all places excluding the target folder).

This allows users to easily configure their IDE to handle both plain `sql` files as well as dbt `sql(.jinja)` files as intended. Currently (at least with pycharm) I have to choose between having all `sql` files interpreted as jinja or as plain `sql`.

I've looked into options to only apply jinja interpretation to the `dbt/` part of my project via my IDE but that is not possible. Not only that I'd prefer a configuration that would have the IDE interpret files properly by default, or at least irregardless of project-specific configuration. This one simple change would ensure that.

### Who will this benefit?
Everybody 😉

Sure

Luttik:
resolves #3484

### Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -85,7 +85,6 @@ target/
 
 # pycharm
 .idea/
-venv/
 
 # AWS credentials
 .aws/
@@ -94,3 +93,7 @@ venv/
 
 # vscode
 .vscode/
+
+# Virtual environments
+venv/
+.venv/
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## dbt 0.21.0 (Release TBD)
 
 ### Features
+- Now supports the `.sql.jinja` file extension in addition to the current `.sql` file extension. With the same behaviour. ([#3484](https://github.com/fishtown-analytics/dbt/issues/3484), [#3498](https://github.com/fishtown-analytics/dbt/pull/3498))
 
 ### Fixes
 - Fix docs generation for cross-db sources in REDSHIFT RA3 node ([#3236](https://github.com/fishtown-analytics/dbt/issues/3236), [#3408](https://github.com/fishtown-analytics/dbt/pull/3408))
--- a/core/dbt/clients/system.py
+++ b/core/dbt/clients/system.py
@@ -1,5 +1,4 @@
 import errno
-import fnmatch
 import json
 import os
 import os.path
@@ -29,14 +28,14 @@ else:
 def find_matching(
     root_path: str,
     relative_paths_to_search: List[str],
-    file_pattern: str,
+    file_regex: str,
 ) -> List[Dict[str, str]]:
     """
     Given an absolute `root_path`, a list of relative paths to that
-    absolute root path (`relative_paths_to_search`), and a `file_pattern`
-    like '*.sql', returns information about the files. For example:
+    absolute root path (`relative_paths_to_search`), and a `file_regex`
+    like '.*\\.sql', returns information about the files. For example:
 
-    > find_matching('/root/path', ['models'], '*.sql')
+    > find_matching('/root/path', ['models'], '.*\\.sql')
 
       [ { 'absolute_path': '/root/path/models/model_one.sql',
           'relative_path': 'model_one.sql',
@@ -47,8 +46,7 @@ def find_matching(
     """
     matching = []
     root_path = os.path.normpath(root_path)
-    regex = fnmatch.translate(file_pattern)
-    reobj = re.compile(regex, re.IGNORECASE)
+    reobj = re.compile(file_regex, re.IGNORECASE)
 
     for relative_path_to_search in relative_paths_to_search:
         absolute_path_to_search = os.path.join(
@@ -61,7 +59,7 @@ def find_matching(
                 relative_path = os.path.relpath(
                     absolute_path, absolute_path_to_search
                 )
-                if reobj.match(local_file):
+                if reobj.fullmatch(local_file):
                     matching.append({
                         'searched_path': relative_path_to_search,
                         'absolute_path': absolute_path,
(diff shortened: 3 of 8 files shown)


Luttik:
@jtcohen6 Can you help me with some questions during this process? First one: I couldn't find `dbt next` in the changelog. So I prepended the file with this section is that the correct behavior or should I move this?

Luttik:
@jtcohen6 Yet another question. I see that a unittest is failing (seems like the same unittest in 5 different scenarios) but I fail to see how this is affected by my changes. Could you point me in the right direction to get me to use your test flow with breakpoints? Simply running pytest will not work in the debt scenario :/

jtcohen6:
@Luttik Thanks for the contribution!

As far as debugging test failures: You can insert `breakpoint()` / `idpb.set_trace()` and then run with the `-s` flag (shortcut for `--capture=no`, though I remember it as short for "show").

In this case, it's a pretty thorny unit test that failed, since it "magically mocks" a python dictionary into a set of model `.sql` files. In so doing, it takes a couple clever shortcuts, with the upshot that dbt (given the changes in this PR) was reading those model files _twice_. Hence, the duplication error: `dbt found two resources with the name "model_one"`.

I think I found a straightforward fix, by changing this line:
https://github.com/fishtown-analytics/dbt/blob/4d246567b937f9707ed748992912972c4698e068/test/unit/test_graph.py#L78
To be more prescriptive:
```python
if iter_self.extension != '.sql':
```
Voila:
```
$ python3 -m pytest test/unit/test_graph.py
======================================================= test session starts =======================================================
platform darwin -- Python 3.9.5, pytest-6.2.4, py-1.10.0, pluggy-0.13.1
rootdir: /Users/jerco/dev/product/dbt, configfile: tox.ini
plugins: logbook-1.2.0, flaky-3.7.0, xdist-2.3.0, dotenv-0.5.2, forked-1.3.0
collected 6 items

test/unit/test_graph.py ......                                                                                              [100%]

======================================================== 6 passed in 2.46s ========================================================
```

Luttik:
No clue how I closed this :/

Luttik:
This isn't working yet. When running this on a project I get the error 
`Model 'model1.sql' (model_name1.sql.jinja) depends on a node named 'model2' which was not found`  if I switch the file extension. The issue weirdly seems to be solved by renaming `model1.sql.jinja` to `model1.sql` instead of doing to same to `model2`... I think that the model should be named `model1` rather than `model1.sql` (is this the case). However, I cannot find where the model-names are generated from the file_path...

It seems like the `read_files` function (which I changed) is only used in `ManifestLoader.load`  (which feels wrong to me...) but this is an almost 140 lines long function with a lot of connected classes and functions. I don't think I have the time to dive deep enough into this code to get to the bottom of this issue :/

Luttik:
@jtcohen6 I'm just tagging you such that somebody on your theme is up to date on my comment above 😉

jtcohen6 (commented):
Hey @Luttik, I took another look and figured out that the new regexes don't interact well with resources that have dots (`.`) in their names. In particular:
- dbt isn't picking up [`seed.with.dots.csv`](https://github.com/dbt-labs/dbt/blob/develop/test/integration/005_simple_seed_test/data-dottedseed/seed.with.dots.csv), causing the failure in `005_simple_seed_test`
- dbt isn't picking up [`alternative.users.sql`](https://github.com/dbt-labs/dbt/blob/develop/test/integration/007_graph_selection_tests/models/alternative.users.sql), causing the failures in `007_graph_selection_tests`

Naming resources with dots is not a very common pattern, but it's not explicitly discouraged and relied on by a few, including the community member who recently added integration tests for it (https://github.com/dbt-labs/dbt/pull/3247).

Could we rework the regexes so that dotted names are still be supported, while also supporting the `.sql.jinja` file extension?

Luttik:
Sure, I could have sworn that the pattern already disallowed dots. But maybe that was a conclusion that I drew before figuring out that it first converts to linux path before converting to regex.

Luttik:
Ok so the tests succeed. The remaining steps are:

- Manual testing 
- Adding new tests.

I am doing this as a hobby but I expect to be able to work on this in the coming weekend.

Luttik:
So the good thing is that the test suite now reflects the issues that I was experiencing in an actual project. The bad aspect is that I need to unravel the monster method that is `dbt.parser.manifest.ManifestLoader.load`. Which is the issue that I mentioned above... @jtcohen6 You've been really helpful so far. Any pointers?

jtcohen6:
@Luttik Sorry for the delay on my end here. The failing tests you're running into are pretty thorny. I can try to take a look when I have a chance to put on my serious debugging hat.

Luttik:
@jtcohen6 Yeah, I thought the same. It looked to me that it might even be beneficial to start with refactoring the `ManifestLoader` (especially the load function) a bit. I think that it could pay huge dividends in terms of overall code quality and interpretability and would make debugging issues like these much easier.

Would you guys be open to refactors like that?

jtcohen6:
There's definitely a lot of code that's changed since this PR was opened. I'd also say that, insofar as this one got bogged down in tricky unit tests, which mock manifests and the like — we're actively looking into refactoring our approaches to unit/integration testing, to prioritize usability.

@dbt-labs/core-language I leave this one up to you — some of the changes that @Luttik is proposing may yield dividends, by introducing the kinds of abstractions / separations (i.e. between file loading vs. manifest construction) that we've been discussing over the past few weeks.

Luttik:
@jtcohen6 Thanks!

gshank:
Do you still want help with this?  The code has changed so you'd probably need to freshen this up.

Luttik:
Hé @gshank, 

I think that this would be a very relevant change. You'd expect something like this to be a fairly small change too... I can definitely freshen this up, but that only seems relevant if https://github.com/orgs/dbt-labs/teams/core-language has some resources to spend on this too as @jtcohen6 indicated.

Issue #3413, opened 2021-06-02
Remove agate dependency?


jtcohen6:
### Describe the feature

As time goes on, our reliance on the `agate` package for light-weight data frames may be providing us more grief than gain.
- Type inference by `agate` causes headaches in the RPC server (#2666, #2984), and it necessitates one of my least favorite configs, [seed `column_types` overrides](https://docs.getdbt.com/reference/resource-configs/column_types)
- We've pinned `agate` to an older version to avoid a second-order dependency that bricks a lot of installation (#3160, #3161)

That said, there's still a lot `agate` does get us:
- Type inference for seeds that's better than nothing
- A reasonable programming interface for tables of info dynamically returned from the database. Many dbt internals, packages, and community members have tapped directly into agate methods, e.g. `print_table()` ([docs](https://docs.getdbt.com/reference/dbt-jinja-functions/run_query))
- It's not pandas

Any changes we make here have a _lot_ of potential visibility and surface area.

### Describe alternatives you've considered
I'm not sure... are there other packages that could occupy the same niche for us? Is this something we should look to hand-roll? I'm not convinced that the problems we're encountering, especially around data type inference, have "easy" solutions, if only we could be writing the code for them ourselves.

### Additional context
This affects all databases. While `agate` functionality is obscured from many users, I imagine that a nontrivial percentage of the overall user base depends on `agate`-specific methods today, whether via their own Jinja code or by leveraging macros from a package.

I'm adding this to the v1.0 list, not because we should definitely do it, but because we should consider whether we can comfortably include the current `agate` dependency in our stable major-version release.

### Who will this benefit?
- RPC users &rarr; dbt Cloud IDE users
- Maintainers of dbt-core
- The workings of `run_query` are wrapped in more mystique today than they ultimately should be. While `agate` provides an okay syntax for working with query results, it's not ideal. Along with work around `execute`, we could improve the experience of templating model SQL based on the results from introspective queries.

jtcohen6:
We can't estimate this right now. Instead, let's spike:
- all the places `agate` is used
- what we need out of a type inferencer
- backwards compatibility options for documented methods on `agate.Table` objects

Some options:
- re-roll the functional bits we need from `agate`
- fork or vendor `agate` to fix dependency/upgrade issues, and give ourselves points of intervention in the future

jtcohen6:
[...]

Issue #2666, opened 2020-07-29
rpc returns empty string values as null values


DylanBaker:
### Describe the bug
When using the rpc (and dbt Cloud IDE where this was discovered), values that are empty strings are being returned and displayed as null values.

### Steps To Reproduce
Using the IDE or rpc, run `select '' and fun`. The following object is returned by the rpc:
```...
        "results": [
            {
                "logs": [],
                "raw_sql": "select '' as fun\n",
                "compiled_sql": "select '' as fun\n",
...
                "table": {
                    "column_names": [
                        "FUN"
                    ],
                    "rows": [
                        [
                            null
                        ]
                    ]
                }
...
```

### Expected behavior
I would expect a response that looks like:
```
                    "rows": [
                        [
                            ""
                        ]
                    ]
```

### Screenshots and log output
If applicable, add screenshots or log output to help explain your problem.

### System information
**Which database are you using dbt with?**
- [ ] postgres
- [ ] redshift
- [ ] bigquery
- [x] snowflake
- [ ] other (specify: ____________)

**The output of `dbt --version`:**
0.16.1

**The operating system you're using:**
Mac

davidharting:
@jtcohen6 I have spent some trying to locate this bug, and was wondering if you could gut check if you think I may be on the right path.

(I am testing this against a local Postgres database, but I believe the issue is database agnostic based on what I believe to be the cause.)

I have been using `select '' as blank`, as suggested in the issue as a simplified reproduction of the bug. 

I have honed in on `dbt.core.clients.AgateHelper.table_from_rows` as what I believe to be the location of problem. 
https://github.com/fishtown-analytics/dbt/blob/ebd761e3dce2eb33bb0ccf807742c1c5f931d583/core/dbt/clients/agate_helper.py#L61

As an input, the `row` parameter contains the correct, empty string representation of the result. 

However, when we convert this into an `agate.Table` at the end of that method, that result contains a `null` instead of an empty string, and the `column_type` is `Number`.

I initially thought that our [default type tester](https://github.com/fishtown-analytics/dbt/blob/ebd761e3dce2eb33bb0ccf807742c1c5f931d583/core/dbt/clients/agate_helper.py#L38). 

```python
def build_type_tester(text_columns: Iterable[str]) -> agate.TypeTester:
    types = [
        agate.data_types.Number(null_values=('null', '')),
        agate.data_types.Date(null_values=('null', ''),
                              date_format='%Y-%m-%d'),
[...]

PR #3499, opened 2021-06-26
Prevent Agate from coercing values in query result sets

### LINKED ISSUE #2984: RPC server sometimes returns the wrong type
### Describe the bug

While using the RPC server, some values are returned as the wrong type. It looks like strings are being cast to floats somehow.

### Steps To Reproduce

1. Start up the rpc server (`dbt rpc`)
2. Make a POST request with the following body:
```json
{
    "jsonrpc": "2.0",
    "method": "run_sql",
    "id": "test3",
    "params": {
        "timeout": 60,
        "name": "test query",
        "sql": "d2l0aCBiYXNlIGFzICgKCiAgc2VsZWN0CgogICAgJ3Rlc3QnOjp2YXJjaGFyIGFzIHRlc3RfdmFyY2hhciwKCiAgICAyNTo6dmFyY2hhciBhcyB0ZXN0X2ludF92YXJjaGFyLAoKICAgICctMjUnOjp2YXJjaGFyIGFzIHRoaXNfd29ya3MKCikKCnNlbGVjdCAKCiAgKiwKCiAgTFBBRCh0ZXN0X3ZhcmNoYXIsIDEwLCAnMCcpIGFzIHBhZF92YXJjaGFyLAoKICBMUEFEKHRlc3RfaW50X3ZhcmNoYXIsIDEwLCAnMCcpIGFzIHBhZF9pbnRfdmFyY2hhciwKCiAgTFBBRCh0aGlzX3dvcmtzLCAxMCwgJzAnKSBhcyBwYWRfdGhpc193b3JrcwoKZnJvbSBiYXNlCg=="
    }
}
```

The SQL here is the base64 encoded version of:
```sql
with base as (

  select

    'test'::varchar as test_varchar,

    25::varchar as test_int_varchar,

    '-25'::varchar as this_works

)

select 

  *,

  LPAD(test_varchar, 10, '0') as pad_varchar,

  LPAD(test_int_varchar, 10, '0') as pad_int_varchar,

  LPAD(this_works, 10, '0') as pad_this_works

from base
```

3. Copy the `request_token` from the response
4. Make another request with the following body:
```json
{
    "jsonrpc": "2.0",
    "method": "poll",
    "id": "test4",
    "params": {
        "request_token": <the-copied-request-token>,
        "logs": true,
        "logs_start": 0
    }
}
```

5. Inspect the rows that are returned. You should see something like this:
```
"rows": [
    [
        "test",
        25.0,
        -25.0,
        "000000test",
        25.0,
        "0000000-25"
    ]
]
```

6. If you run the same query in psql, you will get the following output:
```
 test_varchar | test_int_varchar | this_works | pad_varchar | pad_int_varchar | pad_this_works
--------------+------------------+------------+-------------+-----------------+----------------
 test         | 25               | -25        | 000000test  | 0000000025      | 0000000-25
```

7. Notice how `pad_int_varchar` is padded since it is represented as a varchar, and not an integer

### Expected behavior

I would expect that all of the rows returned would be strings (since they were casted), and not floats. As seen in step 5, the second to last row is not padded since it is being represented as a float.

I don't have any other relevant logs or output, but I'm happy to add more of the RPC response to this issue if that is helpful!

### System information
- [x] postgres
- [ ] redshift
- [ ] bigquery
- [ ] snowflake

```
installed version: 0.18.1
   latest version: 0.18.1

Up to date!

Plugins:
  - postgres: 0.18.1
  - spark: 0.18.1.1
```

macOS Catalina
Version 10.15.7

Python 3.7.7

### Additional context

N/A

drewbanin:
resolves #2984, #2666
related to https://github.com/fishtown-analytics/dbt/issues/3413

### Description

dbt uses Agate to represent internal dataframes (eg. seed file tabular representations, or in-memory query result sets). Agate's type coercions are interesting and helpful for identifying column types in csv files, but they can lead to incorrect/confusing results when applied to query result sets. This PR prevents type coercion from happening when marshaling a query resultset into an Agate table.

### Example

 I used the following macro to debug / reproduce some type coercion issues locally (using Snowflake):
```
{% macro debug() %}

    {% set query %}

        select '0005' as stringnum, 'false' as stringbool, 10 as intfield, false as boolfield, '01T00000aabbccdd' as dtstring
        union all
        select '0006' as stringnum, 'true' as stringbool, 20 as intfield, true as boolfield, '01T00000aabbccde' as dtstring

    {% endset %}

    {% set res = run_query(query) %}
    {% do res.print_table() %}

{% endmacro %}
```

**Before this change:**

| STRINGNUM (bad)  | STRINGBOOL (bad) | INTFIELD | BOOLFIELD |            DTSTRING (would be funny if it wasn't so bad) |
| --------- | ---------- | -------- | --------- | ------------------- |
|         5 |      False |       10 |     False | 0101-01-01 00:00:00 |
|         6 |       True |       20 |      True | 0101-01-01 00:00:00 |

**After this change:**

| STRINGNUM | STRINGBOOL | INTFIELD | BOOLFIELD | DTSTRING         |
| --------- | ---------- | -------- | --------- | ---------------- |
| 0005      | false      |       10 |     False | 01T00000aabbccdd |
| 0006      | true       |       20 |      True | 01T00000aabbccde |

### About the implementation
This PR assumes that we _never_ want to coerce types in tables returned by the database, which I _think_ is correct, but would welcome any holes that can be poked in that assumption. I was originally going to try to use the [cursor.description](https://www.python.org/dev/peps/pep-0249/#description) provided by dbapi implementation to map columns in a query resultset to Python types and then use those Python types to "force" types for all of the columns in the Agate table. In practice, this would require us to add some logic to every adapter plugin which felt both onerous and unnecessary. Instead, we can just peek at the actual query results and "force" string columns in the resultset to be retained as strings in the Agate table that dbt produces. This works because:
 - If the database is returning a string value in a field, we can feel 100% confident that the right type for the column in Agate is a string
 - Columns in a database result cannot be of mixed types, so if we find one string value in a column, we know the whole column should be treated as a string
 - Agate won't coerce a not-string into an unexpected type (eg. an `int` returned by the database can't become a bool in Agate)

### What else
I tried to make a surgical change here. By putting this logic in the agate_helper client, all adapter plugins should be able to take advantage of the change. Longer term, I think we should consider making a less-surgical change that removes Agate entirely. Let's pick up that discussion in #3413 though :)

### Checklist


core/dbt/clients/agate_helper.py
```diff
@@ -84,21 +93,36 @@ def table_from_data(data, column_names: Iterable[str]) -> agate.Table:
...
+    coerced to '5'). Additionally, this method does not coerce values to
+    None (eg. '' or 'null' will retain their string literal representations).
+    """
 
     rows = []
+    text_only_columns = set()
     for _row in data:
         row = []
-        for value in list(_row.values()):
+        for col_name in column_names:
```
drewbanin, line 113: I think our existing implementation assumed that the order of keys in a dictionary was guaranteed... but IIRC that's not true for all versions of Python. My Python is rusty, so maybe this change is not necessary. I can revert if need be :)

core/dbt/clients/agate_helper.py
```diff
@@ -84,21 +93,36 @@ def table_from_data(data, column_names: Iterable[str]) -> agate.Table:
...
+            value = _row[col_name]
             if isinstance(value, (dict, list, tuple)):
-                row.append(json.dumps(value, cls=dbt.utils.JSONEncoder))
-            else:
-                row.append(value)
+                # Represent container types as json strings
+                value = json.dumps(value, cls=dbt.utils.JSONEncoder)
+                text_only_columns.add(col_name)
```
drewbanin, line 118: I was originally just going to peek at the first row in the result set, but it's possible that the value in a string column returned by the database could be `NULL`. This implementation looks at all of the values in the result set, which we're actually already doing anyway to json-encode dict/list/tuple types.

CHANGELOG.md
```diff
@@ -21,6 +21,7 @@ Contributors:
 ### Fixes
 
 - Handle quoted values within test configs, such as `where` ([#3458](https://github.com/fishtown-analytics/dbt/issues/3458), [#3459](https://github.com/fishtown-analytics/dbt/pull/3459))
+- Fix type coercion issues when fetching query result sets ([#2984](https://github.com/fishtown-analytics/dbt/issues/2984), [#3499](https://github.com/fishtown-analytics/dbt/pull/3499))
```
drewbanin, line 24: Is 0.20.0 the right home for this? Or should I slot it into 0.21.0?

jtcohen6: I'd prefer 0.21, rather than sneaking it in during RC time. Even though it is technically a bug fix, I imagine it _might_ be breaking behavior for folks who do a lot of custom query-results munging with Jinja + agate.

core/dbt/clients/agate_helper.py
```diff
@@ -66,7 +70,13 @@ def table_from_rows(
     if text_only_columns is None:
         column_types = DEFAULT_TYPE_TESTER
     else:
-        column_types = build_type_tester(text_only_columns)
+        # If text_only_columns are present, prevent coercing empty string or
+        # literal 'null' strings to a None representation.
```
jtcohen6, line 74: Looks good!
```sql
select '' as empty, null as verynull, 'null' as stringnull
```
Before:
```json
"table": {
    "column_names": [
        "empty",
        "verynull",
        "stringnull"
    ],
    "rows": [
        [
            null,
            null,
            null
        ]
    ]
},
```
After:
```json
"table": {
    "column_names": [
        "empty",
        "verynull",
        "stringnull"
    ],
    "rows": [
        [
            "",
            null,
            "null"
        ]
    ]
},
```

core/dbt/clients/agate_helper.py
jtcohen6: Not my area, but I do remember hearing that this changed in recent versions of python. Indeed [this blog post](https://softwaremaniacs.org/blog/2020/02/05/dicts-ordered/) suggests that dictionaries are ordered by insertion as of python 3.6, and that this is officially guaranteed as of python 3.7. dbt currently requires py36 or higher.

jtcohen6 (approved):
Thanks for pinch-hitting :)

This LGr8TM. I'm going to tag someone else on the Core team for code review. I do think this should go into v0.21, for which we're planning to cut a first beta prerelease in two weeks.

gshank (approved):
Looks good to me!

drewbanin:
Right on, thanks. I updated the changelog to include this fix in the 0.21 section. What's the right base branch for this PR? Is it `develop` or something else?

jtcohen6:
`develop` is right!

barberscott:
Would https://github.com/dbt-labs/dbt/issues/2897 also be in scope for this one?

PR #3505, opened 2021-06-28
Expand partial parsing tests; fix macro partial parsing [#3449]

### LINKED ISSUE #3449: Add test cases for partial parsing
Write test cases for edge cases and coverage of partial parsing.

- [ ] Test macros
- [ ] Test data tests
- [ ] Test analyses
- [ ] Test snapshots
- [ ] Test seeds

gshank:
resolves #3449

### Description

Add some tests for partial parsing. New code to handle partially parsing nested macros, including storing the test unique ids in the SchemaSourceFile 'tests' dictionary in a a format that saves the connection between the test and the particular schema file block that caused the test to be created. 

### Checklist

### DIFF
--- a/core/dbt/contracts/files.py
+++ b/core/dbt/contracts/files.py
@@ -214,7 +214,7 @@ class SourceFile(BaseSourceFile):
 class SchemaSourceFile(BaseSourceFile):
     dfy: Dict[str, Any] = field(default_factory=dict)
     # these are in the manifest.nodes dictionary
-    tests: List[str] = field(default_factory=list)
+    tests: Dict[str, Any] = field(default_factory=dict)
     sources: List[str] = field(default_factory=list)
     exposures: List[str] = field(default_factory=list)
     # node patches contain models, seeds, snapshots, analyses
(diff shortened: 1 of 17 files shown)


jtcohen6 (commented):
Looking good as far as:
- Tested nested macro dependencies with [this gist](https://gist.github.com/jtcohen6/0cf551655dd2d3e415da8252ce82c386), and it works!
- Also: edited a custom generic test definition by adding a `config()`, confirmed that the change propagated.

Issue:
- Removing tests from a `.yml` file does not actually delete the test nodes. If I add a `unique` test on a column, then replace it with a `not_null` test, a subsequent `dbt test` will run both. I remember discussing this previously, but I tested with 0.20.0rc1 and didn't find the same issue, so I think this may be a regression.

gshank:
I created a test case which replaces a model unique test with a model not_null test and it worked. Could you share the files that you were using so I can recreate?

jtcohen6:
@gshank Hm ok! Here's a [gist](https://gist.github.com/jtcohen6/83467c1771d9bce5d893bcf976ae723d) with reproduction case. I just tested on this branch again to confirm.

core/dbt/contracts/files.py
```diff
@@ -255,5 +255,44 @@ def __post_serialize__(self, dct):
...
+    def get_tests(self, yaml_key, name):
+        if yaml_key in self.tests:
+            if name in self.tests[yaml_key]:
+                return self.tests[yaml_key][name]
+        return []
+
+    def get_key_and_name_for_test(self, test_unique_id):
+        yaml_key = None
+        block_name = None
+        for key in self.tests.keys():
```
leahwicz, line 281: Do we think perf is going to take a hit on cycling through these loops?

jtcohen6 (approved):
> Here's a gist with reproduction case. I just tested on this branch again to confirm.

Tested again with the latest code (from fd0d951), confirmed that this is fixed.

LGTM

core/dbt/contracts/files.py
gshank: I don't think so. The dictionaries are pretty small and Python is pretty efficient at doing those things. I believe it's actually less than was happening before these changes.

core/dbt/contracts/graph/manifest.py
```diff
@@ -1031,10 +1051,11 @@ def add_node_nofile(self, node: ManifestNodes):
         _check_duplicates(node, self.nodes)
         self.nodes[node.unique_id] = node
 
-    def add_node(self, source_file: AnySourceFile, node: ManifestNodes):
+    def add_node(self, source_file: AnySourceFile, node: ManifestNodes, test_from=None):
         self.add_node_nofile(node)
         if isinstance(source_file, SchemaSourceFile):
-            source_file.tests.append(node.unique_id)
+            assert test_from
```
kwigley, line 1057: Do we want to keep this assertion in? What is a failure mode that would trip this assertion?

kwigley (approved):
Looks good, thanks for the additional test cases. I left one question for you.

core/dbt/contracts/graph/manifest.py
gshank: It's to keep mypy happy. Calling add_test without a test_from would fail. It shouldn't happen in practice, but mypy doesn't know that.

PR #3509, opened 2021-06-29
Update create_adapter_plugins.py

### LINKED ISSUE #3002: Update create_adapter_plugins.py
Improvements to [`create_adapter_plugins.py`](https://github.com/fishtown-analytics/dbt/blob/dev/kiyoshi-kuromiya/core/scripts/create_adapter_plugins.py). Some of these are things we should do now, some are things we should seek to automate for future releases.

### Versioning
- [ ] Bump [default core version](https://github.com/fishtown-analytics/dbt/blob/30f473a2b1e261e3f4b712564df854f8a95134f3/core/scripts/create_adapter_plugins.py#L235) to `0.19.0-rc1`, add to `.bumpversion.cfg`
- [ ] Create file `dbt/adapters/[adapter]/__version__.py`, to enable reporting from `dbt --version`
- [ ] Codify our recommended practice: plugin versions should match dbt versions, with additional suffixes as needed. (E.g. [here's how dbt-spark does it](https://github.com/fishtown-analytics/dbt-spark/blob/2f7c2dcb419b66d463bb83212287bd29280d6964/setup.py#L15-L38).) Change default `package-version` from `0.0.1` to match the core version, and add setup functions to effectuate.

### Testing
- [ ] Testing: Add scaffolding for `test/[adapter].dbtspec`, modeled off the [postgres spec](https://github.com/fishtown-analytics/dbt-adapter-tests/blob/master/specs/postgres.dbtspec)
- [ ] Add scaffolding for `dbt/include/[adapter]/sample_profiles.yml`, which dbt will use to populate `~/.dbt/profiles.yml` if a user installs the plugin and runs `dbt init --adapter [adapter]` for the first time. (We should make this work if they just run `dbt init`, too: https://github.com/fishtown-analytics/dbt/issues/2814)
- [ ] Should the script accept/expect a `connection_keys` argument? If we did, we could use that to populate both of the above. Currently, [we expect the user to add in these fields](https://docs.getdbt.com/docs/contributing/building-a-new-adapter/#the-credentials-class) after running the scaffolding script.

### Future: dbt Cloud

Eventually, we may want community plugin maintainers of community plugins to include:
- "blessed" dockerfiles of runtime environments
- JSONSchema of connection profile spec (same info as above)

This will enable us to support the plugin in dbt Cloud. We're still a ways away from that, but it's worth thinking about now in terms of the scaffolding we provide.

jtcohen6:
resolves #3002

- Bump default dbt version, add to `.bumpversion.cfg`
- Add `__version__.py`, `sample_profiles.yml`, `<myadapter>.dbtspec`
- Use `find_namespace_packages` and `MANIFEST.in`
- This feels like a good candidate for `cookiecutter`! There's _so_ many arguments to this script, and I could imagine getting fancier and asking for credential members as well (to populate in `connections.py`, `sample_profiles.yml`, and `<myadapter>.dbtspec`). I may take a crack at that, but I don't want to impede progress in the meantime.

This PR does _not_ include all the opinionated (and fairly tricky) logic that we have in `dbt-spark` + `dbt-presto` around:
- Dynamically grabbing the package version from `__version__.py` (as long as both are being bumped...)
- Guaranteeing that the package version matches the dbt-core version

### Checklist

### DIFF
--- a/core/scripts/create_adapter_plugins.py
+++ b/core/scripts/create_adapter_plugins.py
@@ -284,12 +284,12 @@ def parse_args(argv=None):
     parser.add_argument('adapter')
     parser.add_argument('--title-case', '-t', default=None)
     parser.add_argument('--dependency', action='append')
-    parser.add_argument('--dbt-core-version', default='0.20.0rc1')
+    parser.add_argument('--dbt-core-version', default='0.20.0rc2')
     parser.add_argument('--email')
     parser.add_argument('--author')
     parser.add_argument('--url')
     parser.add_argument('--sql', action='store_true')
-    parser.add_argument('--package-version', default='0.20.0rc1')
+    parser.add_argument('--package-version', default='0.20.0rc2')
     parser.add_argument('--project-version', default='1.0')
     parser.add_argument(
         '--no-dependency', action='store_false', dest='set_dependency'
--- a/core/scripts/create_adapter_plugins.py
+++ b/core/scripts/create_adapter_plugins.py
@@ -14,12 +14,12 @@ macro-paths: ["macros"]
 
 SETUP_PY_TEMPLATE = '''
 #!/usr/bin/env python
-from setuptools import find_packages
-from setuptools import setup
+from setuptools import find_namespace_packages, setup
 
 package_name = "dbt-{adapter}"
+# make sure this always matches dbt/adapters/{adapter}/__version__.py
 package_version = "{version}"
-description = """The {adapter} adapter plugin for dbt (data build tool)"""
+description = """The {adapter} adapter plugin for dbt"""
 
 setup(
     name=package_name,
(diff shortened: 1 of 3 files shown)


gshank (approved):
This is not an area that I'm familiar with, but it looks pretty straightforward.

core/scripts/create_adapter_plugins.py
```diff
@@ -232,12 +284,12 @@ def parse_args(argv=None):
     parser.add_argument('adapter')
     parser.add_argument('--title-case', '-t', default=None)
     parser.add_argument('--dependency', action='append')
-    parser.add_argument('--dbt-core-version', default='0.16.1rc1')
+    parser.add_argument('--dbt-core-version', default='0.20.0rc1')
```
kwigley, line 287: 😮

kwigley (approved):

core/scripts/create_adapter_plugins.py
jtcohen6: I know!

jtcohen6: going to update this to 0.20.0 actually

PR #3512, opened 2021-06-29
Experimental Parser: Change Python for Rust

nathaniel-may:
Merges [PR#3497](https://github.com/fishtown-analytics/dbt/pull/3497)

### Checklist


core/setup.py
```diff
@@ -73,10 +73,9 @@ def read(fname):
         'networkx>=2.3,<3',
         'packaging~=20.9',
         'sqlparse>=0.2.3,<0.4',
-        'tree-sitter==0.19.0',
-        'tree-sitter-jinja2==0.1.0a1',
-        'typing-extensions>=3.7.4,<3.8',
-        'werkzeug>=0.15,<2.0',
+        'dbt-extractor==0.2.0',
+        'typing-extensions>=3.7.4,<3.11',
+        'werkzeug>=0.15,<3.0',
```
kwigley, line 78: hmm, I'm not opposed, _but_ should we widen this dependency range in this PR? I think we did on `develop`, but not in `0.20.latest`

kwigley (approved):

core/setup.py
nathaniel-may: It looks like it matches develop. Am I missing something?

nathaniel-may: Oh you mean that we should consider putting it back to the previous narrower range here.

kwigley: ya ya, `'werkzeug>=0.15,<2.0'`

nathaniel-may: I'm not familiar with our usage of that package. What are the pros and cons in your mind?

kwigley: It is a dependency of the rpc server.. I'm not sure we've thoroughly tested any versions of `werkzeug`> `2.0.0` though

nathaniel-may: If we don't need it I'm happy to reduce it. Since I merged this, do you want to open a new PR with that?

PR #3526, opened 2021-07-01
'+' config prefix handling whitespace

### LINKED ISSUE #3494: '+' config prefix should handle whitespace
### Describe the bug
Macros are not working in post-hooks in `dbt_project.yml` but working as a model config

### Steps To Reproduce
1. Create an empty dbt project with `dbt init`
2. Create a macro in `macros/grants_test.sql`. I'm using this for testing:
```
{% macro grants_test(rel) %}
ALTER TABLE {{ rel }} SET OPTIONS (description="testing...")
{% endmacro %}
```
3. Modify the `dbt_project.yml` to add a post-hook. Based on the sample project `dbt init` creates I changed it to this:
```
models:
  my_new_project:
      # Applies to all files under models/example/
      example:
          + post-hook: [
            '{{ grants_test(this) }}'
          ]
          materialized: view
```

### Expected behavior
I would expect the post-hook to use the `grants_test` macro correctly. It actually works if instead we add it as a model config in for example: `models/example/my_first_dbt_model.sql`

```
Running with dbt=0.19.0
Encountered an error:
Compilation Error
  Could not render {{ grants_test(this) }}: 'grants_test' is undefined
```

### System information
- [ ] postgres
- [ ] redshift
- [x] bigquery
- [ ] snowflake

```
installed version: 0.19.0
   latest version: 0.19.1

Plugins:
  - redshift: 0.19.0
  - bigquery: 0.19.0
  - snowflake: 0.19.0
  - postgres: 0.19.0
```

Ubuntu 20.04

Python 3.8.5

jmriego:
resolves #3494

### Description

There is not really supported to add a whitespace after the plus sign in the configuration yml but the error is not intuitive. There was a fix for the eager rendering of macros and this extends this fix for when there's a space between the plus sign and the key (eg. `+ post-hook`)

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Fixes
 - Fix docs generation for cross-db sources in REDSHIFT RA3 node ([#3236](https://github.com/fishtown-analytics/dbt/issues/3236), [#3408](https://github.com/fishtown-analytics/dbt/pull/3408))
 - Fix type coercion issues when fetching query result sets ([#2984](https://github.com/fishtown-analytics/dbt/issues/2984), [#3499](https://github.com/fishtown-analytics/dbt/pull/3499))
+- Handle whitespace after a plus sign on the project config ([#3526](https://github.com/dbt-labs/dbt/pull/3526))
 
 ### Under the hood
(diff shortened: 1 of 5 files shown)


jtcohen6:
@jmriego I think you're close!

https://github.com/dbt-labs/dbt/blob/83b98c8ebf6620fd033af54cdede3d3578d0f5de/core/dbt/context/context_config.py#L99-L100

```python
 if key.startswith('+'): 
     result[key[1:].strip()] = deepcopy(value) 
```

jtcohen6 (approved):
Thanks for this one @jmriego!

PR #3528, opened 2021-07-01
add correctness tracking for experimental parser

### LINKED ISSUE #3503: Add experimental parser correctness checks to tracking
### Describe the feature
`1 / n` parses with the experimental parser should run both python jinja parsing and the experimental parser, compare results and send back whether they match or not. Frequency should be determined with an rng.

Not having this information

### Additional context
We want to make sure the experimental parser is 100% in line with the existing python jinja rendering before we consider turning it on by default. 

### Who will this benefit?
dbt maintainers

### Relevant Tickets
#3438

nathaniel-may:
resolves #3503

### Description
Adds tracking for experimental parser correctness that includes what kind of mismatch occurred.

### Checklist

### DIFF
--- a/core/dbt/parser/models.py
+++ b/core/dbt/parser/models.py
@@ -7,7 +7,7 @@ from dbt.parser.search import FileBlock
 from dbt_extractor import ExtractionError, py_extract_from_source  # type: ignore
 import itertools
 import random
-from typing import Any, Dict, List, Tuple
+from typing import Any, Dict, List, Set, Tuple
 
 
 class ModelParser(SimpleSQLParser[ParsedModelNode]):
@@ -35,7 +35,7 @@ class ModelParser(SimpleSQLParser[ParsedModelNode]):
         # run the experimental parser if the flag is on or if we're sampling
         if flags.USE_EXPERIMENTAL_PARSER or sample:
             try:
-                experimentally_parsed = py_extract_from_source(node.raw_sql)
+                experimentally_parsed: Dict[str, List[Any]] = py_extract_from_source(node.raw_sql)
 
                 # second config format
                 config_calls: List[Dict[str, str]] = []
@@ -43,7 +43,7 @@ class ModelParser(SimpleSQLParser[ParsedModelNode]):
                     config_calls.append({c[0]: c[1]})
 
                 # format sources TODO change extractor to match this type
-                source_calls = []
+                source_calls: List[List[str]] = []
                 for s in experimentally_parsed['sources']:
                     source_calls.append([s[0], s[1]])
                 experimentally_parsed['sources'] = source_calls
@@ -57,7 +57,7 @@ class ModelParser(SimpleSQLParser[ParsedModelNode]):
             super().render_update(node, config)
             # if we're sampling, compare for correctness
             if sample:
-                result = set()
+                result: Set[str] = set()
                 # experimental parser couldn't parse
                 if isinstance(experimentally_parsed, Exception):
                     result.add("01_experimental_parser_cannot_parse")
--- a/core/dbt/parser/models.py
+++ b/core/dbt/parser/models.py
@@ -5,6 +5,7 @@ from dbt.node_types import NodeType
 from dbt.parser.base import SimpleSQLParser
 from dbt.parser.search import FileBlock
 from dbt_extractor import ExtractionError, py_extract_from_source  # type: ignore
+import random
 
 
 class ModelParser(SimpleSQLParser[ParsedModelNode]):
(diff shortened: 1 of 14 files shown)


core/dbt/parser/models.py
```diff
@@ -26,46 +27,109 @@ def render_update(
...
-                # when we support hooks
-                self.update_parsed_node(node, config)
+                    # dedup values
+                    value = list(set(value))
 
-                # udpate the unrendered config with values from the file
-                # values from yaml files are in there already
-                node.unrendered_config.update(dict(res['configs']))
+                    # set sample results
+                    # TODO set this somewhere so it can be sent
```
jtcohen6, line 106: Right on, @gshank or @kwigley could help with setting up snowplow tracking for this.

We could swing this with just a standard structured event (no schema addition necessary), but I think we'll want a new validation schema that checks for the possible outcome values you've defined above.

nathaniel-may: Yeah I have no idea what you might expect to see in the database, so these are just wet noodle values I flung at the wall. What values make sense to you?

nathaniel-may:
Full suite passed with the flag on by default. Reverted flag setting.

core/dbt/parser/models.py
```diff
@@ -26,46 +27,106 @@ def render_update(
...
+        # normal dbt run
+        if not flags.USE_EXPERIMENTAL_PARSER:
+            # normal rendering
+            super().render_update(node, config)
+            # if we're sampling, compare for correctness
+            if sample:
+                value = []
+                # experimental parser couldn't parse
+                if isinstance(experimentally_parsed, Exception):
+                    value += ["01_experimental_parser_couldn't_parse"]
```
jtcohen6, line 61: Nit: that apostrophe is going to be annoying
```suggestion
                    value += ["01_experimental_parser_cannot_parse"]
```

core/dbt/parser/models.py
```diff
@@ -26,46 +27,106 @@ def render_update(
...
+                        if c not in node.sources:
+                            value += ["04_false_positive_source_value"]
+                            break
 
-                # udpate the unrendered config with values from the file
-                # values from yaml files are in there already
-                node.unrendered_config.update(dict(res['configs']))
+                    # look for missed sources
+                    for c in node.sources:
+                        if c not in experimentally_parsed['configs']:
```
jtcohen6, line 83:
```suggestion
                        if c not in experimentally_parsed['sources']:
```

core/dbt/parser/models.py
```diff
@@ -26,46 +27,106 @@ def render_update(
...
-                for c in res['configs']:
+                for c in experimentally_parsed['configs']:
                     config_calls.append({c[0]: c[1]})
 
-                config._config_calls = config_calls
+                # format sources TODO change extractor to match this type
+                source_calls = []
+                for s in experimentally_parsed['sources']:
+                    source_calls.append([s[0], s[1]])
+                experimentally_parsed['sources'] = source_calls
```
jtcohen6, line 47: Need this for `experimentally_parsed['configs'] = config_calls` above as well

core/dbt/parser/models.py
```diff
@@ -26,46 +27,106 @@ def render_update(
...
-                res = py_extract_from_source(node.raw_sql)
+                experimentally_parsed = py_extract_from_source(node.raw_sql)
 
-                # since it doesn't need python jinja, fit the refs, sources, and configs
-                # into the node. Down the line the rest of the node will be updated with
-                # this information. (e.g. depends_on etc.)
+                # second config format
                 config_calls = []
-                for c in res['configs']:
+                for c in experimentally_parsed['configs']:
```
jtcohen6, line 40: Hmm I'm seeing some misalignments in the sense that the experimental parser returns
```
[{'unique_key': 'event_id'}, {'materialized': 'incremental'}]
```
Whereas traditional parsing returns
```
[{'unique_key': 'event_id', 'materialized': 'incremental'}]
```
With the resulting value of: `['02_false_positive_config_value', '03_missed_config_value']`

core/dbt/parser/models.py
```diff
@@ -26,46 +27,106 @@ def render_update(
...
+                            value += ["07_missed_ref_value"]
+                            break
 
-            # exception was thrown by dbt jinja extractor meaning it can't
-            # handle this source. fall back to python jinja rendering.
-            except ExtractionError:
-                super().render_update(node, config)
+                    # dedup values
+                    value = list(set(value))
+
```
jtcohen6, line 101: If `value` is empty here at the end, what do you think of sticking in some kind of `"00_success"` or `"00_exact_match"`?

core/dbt/parser/models.py
jtcohen6: The ones you picked make a lot of sense to me! Just a few tweaks

jtcohen6 (commented):
Neat-o! I took this for a spin with our internal analytics project, and dug into some of the discrepancies.

Before merging this in, we need to:
- Instrument a new tracking event
- Run this through all the models in our internal project, and confirm everything comes up exactly matching

core/dbt/parser/models.py
nathaniel-may: good catch!

core/dbt/parser/models.py
nathaniel-may: very good point

core/dbt/parser/models.py
nathaniel-may: configs are weird and need to be in two forms. one for each of the places where we need to set them when the flag is active: `config._config_calls` and `node.unrendered_config.update`. So I purposefully don't overwrite the original format here.

_However_ this correctly points out that my comparison is not using the correct one of the two options so I fixed that.

core/dbt/parser/models.py
nathaniel-may: yup. I think the above comment should address this now. Thanks for the detailed example!

core/dbt/parser/models.py
nathaniel-may: yes we should totally do that. added.

core/dbt/parser/models.py
nathaniel-may: I was wrong, this needed to be handled completely separately. Thanks for pointing it out!

nathaniel-may:
Tested locally with our internal analytics project and it correctly compares for all parseable model files.

jtcohen6 (commented):
Thanks for confirming that this matches expectations for our internal project.

Let's wire up a schema + tracking event!

core/dbt/parser/models.py
```diff
@@ -26,46 +27,111 @@ def render_update(
...
+                    # if there are no errors, return a success value
+                    if not result:
+                        result = set(["00_exact_match"])
+
+                    # set sample results. results are cumulative across the dbt run,
+                    # so we will only know there exists a false positive config value
+                    # somewhere in the project, not how many files have false positve
+                    # configs. this means a perfectly correct sample will return a
+                    # result of {'00_exact_match'}.
+                    self.manifest._parsing_info.static_analysis_result.update(result)
```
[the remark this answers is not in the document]
nathaniel-may: just changed it up to fire one event per sampled file.

nathaniel-may:
These CI tests were expected to fail. They will hopefully pass when I uncomment the iglu address once the new schema has been added.

nathaniel-may:
rebased on branch `test-tweaks` to apply changes here before `test-tweaks` gets merged into develop. Only reason is to continue working while Circle is having issues.

core/dbt/parser/models.py
```diff
@@ -26,46 +31,122 @@ def render_update(
...
+                    # fire a tracking event. this fires one event for every sample
+                    # so that we have data on a per file basis. Not only can we expect
+                    # no false positives or misses, we can expect the number model
+                    # files parseable by the experimental parser to match our internal
+                    # testing.
+                    tracking.track_experimental_parser_sample({
+                        "project_id": self.root_project.hashed_name(),
+                        "file_id": utils.get_hash(node),
+                        "status": result
+                    })
```
jtcohen6, line 123: Let's also track if the result is `01_experimental_parser_cannot_parse`!

nathaniel-may:
closing in favor of [#3553](https://github.com/dbt-labs/dbt/pull/3553)

PR #3536, opened 2021-07-06
dbt init: add gitignore, ignore pycache

jtcohen6:
Follow-up to #3474, #3483

### Description

- Exclude `__pycache__` (if present, as it is for me) from `dbt init`
- Include `.gitignore` for starter dbt projects

### Checklist


leahwicz:
I think `.gitignore` needs to be added to the [Manifest.in](https://github.com/dbt-labs/dbt/blob/develop/core/MANIFEST.in) file as well for the cases where we don't have the source code and this is packaged up. It needs to know to include those files

leahwicz (approved):

PR #3537, opened 2021-07-06
table and view materializations should rename from old_relation

### LINKED ISSUE #2161: Use old_relation for renaming in default materializations
### Describe the bug
In the default [view](https://github.com/fishtown-analytics/dbt/blob/dev/barbara-gittings/core/dbt/include/global_project/macros/materializations/view/view.sql#L50) and [table](https://github.com/fishtown-analytics/dbt/blob/dev/barbara-gittings/core/dbt/include/global_project/macros/materializations/table/table.sql#L45) materializations, dbt should rename `old_relation` to `backup_relation`, instead of renaming `target_relation` to `backup_relation`. The issue here is that, if `old_relation.type != target_relation.type`, dbt will run the wrong `alter ... rename` statement.

I think this is as simple as switching these to
```sql
  {% if old_relation is not none %}
    {{ adapter.rename_relation(old_relation, backup_relation) }}
  {% endif %}
```

### Steps To Reproduce
- Create + run a view model
- Change config to table + rerun

This does not break on any of the core adapters:
- Redshift and Postgres use `alter ... rename`, but they support `alter table` for both tables + views
- Snowflake and BigQuery don't use `alter ... rename`, they instead use `create or replace`

It does break on Presto, which distinguishes between `alter view` and `alter table`. dbt will attempt to run `alter table` (because `target_relation.type` is `table`) against an existing view and return an error.

### Expected behavior
dbt should use the type of the old relation when it attempts to rename it.

```
dbt --version
installed version: 0.15.2
   latest version: 0.15.2

Up to date!
```

danielefrigo:
in this way the materialization should manage changes from view to table and reverse

resolves #2161

### Checklist


jtcohen6:
@danielefrigo Looks great, thanks for this! Could you:
- Change the base branch to be `develop`
- Add a Changelog entry under `dbt 0.21.0`
- Add yourself to the list of contributors

PR #3549, opened 2021-07-08
Tweak function to return a value instead of asserting

nathaniel-may:
### Description

This looks like a very silly change. Why would we want this?

Well, when the helper function asserts, our tests are locked into the interface of the function itself. Our test _must fail_ if this equality check fails. However, with the addition of the new [experimental parser tracking event](https://github.com/dbt-labs/dbt/pull/3528), our tracking events are non-deterministic. This is a quick change for us to be able to assert that the set of fired events must be `A` or `B` (with or without the experimental parser sampling).

Although I do generally believe in pushing assert statements to the top-level test, this very simple change isn't a long term brilliant solution to efficiently testing non-deterministic tracking.

### Checklist

### DIFF
--- a/test/integration/033_event_tracking_test/test_events.py
+++ b/test/integration/033_event_tracking_test/test_events.py
@@ -83,10 +83,7 @@ class TestEventTracking(DBTIntegrationTest):
             else:
                 populated_contexts.append(context)
 
-        self.assertEqual(
-            ordered_contexts,
-            populated_contexts
-        )
+        return ordered_contexts == populated_contexts
 
     def load_context(self):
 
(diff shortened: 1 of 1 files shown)


leahwicz (approved):

kwigley (commented):
code changes look good, do you know if CI is being flaky or if this is a real failure?

nathaniel-may:
@kwigley I think it's being flaky? It's almost always failing on partial parsing, but not the same test every time.

nathaniel-may:
that is to say, there are exactly zero tests that always failed with this change through all the CI reruns I triggered last night.

PR #3554, opened 2021-07-09
Source freshness task node selection and cli command parity

### LINKED ISSUE #2987: Parity for source freshness task/command
### Describe the feature

`dbt source snapshot-freshness` is quite limited and a bit verbose relative to other dbt task-commands.

1. Source freshness should support full node selection syntax, including things like:
    - `--select tag:something` (#2369)
    - `--exclude` and `--selector` (#2729)
    - `--select +downstream_resource`, i.e. check freshness of all sources upstream of `downstream_resource`. I think this would be especially compelling for exposures!

2. Should this be renamed to avoid confusion with `dbt snapshot`? I think yes, and it could be as simple as calling it `dbt source freshness` instead—although that's not a verb! Open to suggestions.

We'd obviously do the renaming in a way that's backwards-compatible. Ideally, we'd rename and add functionality in the same release.

Leaving `dbt source snapshot-freshness` as is: clunky to type, wonky to use

### Additional context
Is this feature database-specific? Which database(s) is/are relevant? Please include any other relevant context here.

### Who will this benefit?
All current and potential users of source freshness, which is a relatively underused + under-appreciated feature! Maybe in part because it feels a bit clunky :)

kwigley:
resolves #2987 

### Description
- adds support for node selection (`--select`, `--exclude`, `--selector`)
- aligns selection syntax with other commands: `dbt source freshness --select raw_payments` -> `dbt source freshness --select source:raw_payments`
- rename cli and rpc command to `source freshness`
- continue to support `source snapshot-freshness` via cli and rpc

Note: There is no deprecation warning for using the old command name, it wasn't as simple as I initially thought (#3565)

### Checklist

### DIFF
--- a/core/dbt/task/freshness.py
+++ b/core/dbt/task/freshness.py
@@ -137,9 +137,15 @@ class FreshnessTask(GraphRunnableTask):
         return False
 
     def get_selection_spec(self) -> SelectionSpec:
+        """Generates a selection spec from task arguments to use when processing
+        graph. A SelectionSpec describes what nodes to select when creating 
+        queue from graph of nodes.
+        """
         if self.args.selector_name:
+            # use pre-defined selector (--selector) to create selection spec
             spec = self.config.get_selector(self.args.selector_name)
         else:
+            # use --select and --exclude args to create selection spec
             spec = parse_difference(self.args.select, self.args.exclude)
         return spec
 
--- a/core/dbt/task/freshness.py
+++ b/core/dbt/task/freshness.py
@@ -137,9 +137,9 @@ class FreshnessTask(GraphRunnableTask):
         return False
 
     def get_selection_spec(self) -> SelectionSpec:
-        """Generates a selection spec from task arguments to use when processing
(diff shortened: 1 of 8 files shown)


core/dbt/task/freshness.py
```diff
@@ -137,11 +137,10 @@ def raise_on_first_error(self):
         return False
 
     def get_selection_spec(self) -> SelectionSpec:
```
nathaniel-may, line 139: This is kind of non-obvious what's happening here. A comment could be useful.

nathaniel-may (approved):
Looks good! I like the added test cases. I think it could use more comments specifically around describing how each function fits in to the bigger picture and exactly what the goal of each test case is. Otherwise lgtm.

leahwicz (approved):

jtcohen6 (approved):
This LGTM! Nice work here.

Worth calling out—probably as a Breaking Change in the changelog—that the sui generis selection syntax previously supported by the freshness task is now going to work like all other selection syntax. So `-s src_name` is now `-s source:src_name`.

PR #3558, opened 2021-07-11
Suggestions for #3387

jtcohen6:
### Biggest changes
- Remove as much (repeated) logic as possible from the materialization
- Change info-level logging to be debug-level
- Refactor tests to be reused across adapters
- Add tests and macros for Redshift
- ❗ Remove `alter_column_types` boolean config, in favor of the opinion that `on_schema_change: sync_all_columns` should always alter column types. This is potentially controversial, so I included it in a totally separate commit that we can easily revert :)

### For reference
After these changes, here is the function of each macro:
- `process_schema_changes`: This is the main point of entry, and it handles the entire schema change check-and-handle process, from beginning to end. That includes ignoring if `ignore`, erroring if `fail`, and calling other macros as needed. Each incremental materialization calls this macro alone. (Exception: BigQuery needs a little custom logic, since it's not a given that the temp relation of new records will exist, and there are cost implications to creating it when it's not needed.)
- `check_for_schema_changes`: Accepts `source_relation` + `target_relation`, returns `changes_dict`, prints a debug log message.
- `alter_relation_add_remove_columns`: Accepts `relation` and lists of `add_columns` + `remove_columns`, templates and runs the appropriate DML. This is the bit of logic that varies across adapters, so it's the only dispatched macro of the bunch (so far). I recombined `add column` + `drop column` into a single macro here because both Postgres + BigQuery can run them within the same `alter table` statement, so I think we should take that as the high-water-mark implementation. It's easy enough to reimplement for Redshift + Snowflake, which need to run separate statements.
- `sync_column_schemas`: "Do the thing." Accepts `on_schema_change` + `target_relation` + `schema_changes_dict`, calls `alter_relation_add_remove_columns` and `alter_column_type` as appropriate for the `on_schema_change` strategy, prints a debug log message of changes made.

### Housekeeping
- Move Changelog entry up to v0.21 Features
- Delete `067` test, redundant with `069_test_incremental_schema_tests`. I'd love to actually rename this to `069_incremental_schema_change_tests`, but I don't want to mess up git diffs for the review process.

### Next steps
- Do you buy it? Anything stand out as a totally unreasonable way of doing this?
- Merge/rebase the base branch against `develop`
- Get this in for v0.21.0b1
- Write some prerelease docs :)

### Checklist


matt-winkler:
@jtcohen6 Thanks for putting this together - I'm good with this strategy. Some notes on additional changes and a question: 

**Additional changes**
- Updated the logging in `on_schema_change` for the fail case to tell users they have the option to use `append_new_columns` OR `sync_all_columns` depending on their use case.  
- Updated `test_incremental_schema.py` to compare the source and target models in `self.assertTablesEqual()`
- Updated the integration tests to explicitly specify the data types, as this resolves issues with Snowflake/Bigquery/Redshift related to the column types that are mapped by default.   

**Question**
- With the proposal to ADD/DROP on Bigquery in order to accomplish a type change, I'm wondering if we will create a situation that destroys historical data in the target incremental table. If that's the case we should probably stick with the current behavior even if it's a more expensive query to perform a type change on BQ.   
- What do we need to do to document this? Update dbt docs?

jtcohen6:
> With the proposal to ADD/DROP on Bigquery in order to accomplish a type change, I'm wondering if we will create a situation that destroys historical data in the target incremental table. If that's the case we should probably stick with the current behavior even if it's a more expensive query to perform a type change on BQ.

We discussed this a bit offline. Although BigQuery now supports `alter table add | drop column`, it doesn't _really_ support this in a way that saves $$. That's ok—`sync_all_columns` will require a full re-scan of the existing table, that should still be cheaper than a full re-scan of source data, which is what `--full-refresh` requires today. At such time as BigQuery adds real support for `alter table alter column type`, we'll really be in business.

> What do we need to do to document this? Update dbt docs?

Yes, I think we'll want a new section here: https://docs.getdbt.com/docs/building-a-dbt-project/building-models/configuring-incremental-models

PR #3559, opened 2021-07-12
Feature: state:modified.macros

### LINKED ISSUE #2704: Subselectors for state:modified
This is a follow-on from the initial proposals in #2465, #2641. Most of the required work is around exposing + sugaring the foundational work in #2695.

The way state comparison works is by identifying discrepancies between two manifests. When comparing between a past prod manifest and the current development manifest, discrepancies can be the result of two things:

1. Changes made to the project in development
2. Env-aware logic that causes different behavior based on the `target`, env vars, etc.

We're going to do our best to capture *only* changes that are the result of development. If someone's project has tons of intricate env-aware logic, they'll run more models than they want (i.e. more false positives). So we're giving them the option to turn off some knobs, in the form of more-specific subselectors.

## Subselectors

There is potential for overlap: a single change can trigger multiple modification categories.

**`state:modified.contents`**:

- Models: raw file contents changed
- Snapshots: raw file contents changed
- Data tests: raw file contents changed
- Analyses: raw file contents changed
- Seeds: raw file contents changed
    - However, if the file is >1 MB, we cannot compare raw file contents, so we raise a warning and just compare based on file path instead.

This alone would get a lot of people what they want! It's basically "just hash the files," excluding YAML configuration.

**`state:modified.configs`**:

- Models: changes to materialized, quoting, bind, transient, sort/dist, partition_by, incremental_strategy...
    - This category captures changes in `dbt_project.yml` or `{{config()}}` blocks. If the changes are made in a `{{config()}}` block, they will also be picked up as **content** changes.
    - If someone has env-aware logic for materialized, where a model is a view in dev and a table in prod, they will not want to include this.
- Snapshots: unique_key, strategy, ...
- Seeds: quoting, column_types
- Schema tests: severity changes

**`state:modified.descriptions`**:

- If `persist_docs` is turned on for a node, description changes count as modifications. (If just columns, just column descriptions; just relations, top-level descriptions; if both, then both.)

**`state:modified.database_representations`**:

- Models: changes to the *configured* `database`, `schema`, `identifier`. This value represents the manual input only, and it's *different* from the *resolved* database representation, which depends on the `target` and `generate_x_name` macros.
    - If someone manually sets `schema = target.schema`, or `schema = target.schema ~ '_suffix'` instead of using the `generate_schema_name` macro, that _will_ register as a change between environments and they'll want to turn this off.
    - Depending on the `generate_x_name` logic and the current environment, a chance to the *configured* value may not actually change the database representation. We'll still register it as a modification.
- Seeds: treated the same as models.
- Sources: `database`, `schema`, or `identifier` has changed. If someone has env-aware definitions, they'll want to turn this off.
- Snapshots: treated the same as sources.

## Default behavior

I think `state:modified` should include _all_ changes from _all_ the categories above. The question mark is whether `database_representations` should be included in the default, since this is the area where people do the most custom things, and it's the knob that will likely be switched off most often. For the sake of clarity, I think it's best to have the `state:modified` selector be a superset of all `modified` subselectors.

## Future art

**`state:modified.macros`**:

- A macro's raw contents have changed
- By extension, `state:modified.macros+` would include all downstream models, tests, etc. that call (directly or indirectly) a macro that has changed
- This also includes implicit macro dependencies such as `generate_schema_name`

**`state:modified.vars`**:

- A var value has changed
- By extension, `state:modified.vars+` would include all downstream models, tests, etc. that call (directly or indirectly) a var that has changed

We will update `state:modified` to include both of these as well.

### LINKED ISSUE #3278: Use depends_on.macros as input to state:modified
During state comparison, dbt knows which macros have been modified, but all it does today is log a warning if it finds one:

https://github.com/fishtown-analytics/dbt/blob/f73202734c08b85da739818af2aff74dcbc8638a/core/dbt/graph/selector_methods.py#L435-L445

Why just raise a warning? Well, even though we've long had `depends_on.macros` for _models_, we didn't have it for _macros_—so we didn't have a reliable way to know if a model depends on a macro that depends on a (changed) macro.

For an unrelated purpose, #3272 added a full `depends_on.macros` for macros, too. We could use this (now reliable) `depends_on.macros` as an input to `state:modified`, to select models that call (directly or indirectly) a modified macro.

I think this would look like:
- Find the set of modified macros (we have this today)
- Recursively find the "extended" set of macros that depend on a modified macro
- Find the set of all nodes that depend on any macro in the extended set

I think this could be a simpler way to get at some of the functionality that #2714 described, though the proposal there is much more complex.

jtcohen6:
resolves #2704
resolves #3278
resolves [this tweet](https://twitter.com/tjwaterman99/status/1420809652548296704)

### Description

If a macro's SQL is modified, mark any node that depends on it (or depends on a macro that depends on it) as modified.

While I was here, it was trivial to make sub-selectors (finally) happen:
- `state:modified.body`
- `state:modified.configs`
- `state:modified.persisted_descriptions`
- ~`state:modified.database_representations`~ `state:modified.relation`
- `state:modified.macros`

### For review
- I don't really know how to write recursive functions :)
- Let's revisit some of the nomenclature here (vs. the original proposal in #2704). Even though it's a mouthful, I prefer `persisted_descriptions` vs. just `descriptions`, to be clear it's the subset with `persist_docs` enabled. On the other hand, I'm not sure about `modified.database_representations` vs. something shorter & sweeter like `modified.relations`.
- There are a few more `same_x` checks (`same_fqn`, `same_quoting`, `same_external`) that are resource-specific and wouldn't have dedicated subselectors in this first round. I think that's ok—these aren't ones we've heard as much demand for, and they're still included in the `state:modified` superset—just worth calling out.

### Checklist

### DIFF
--- a/core/dbt/graph/selector_methods.py
+++ b/core/dbt/graph/selector_methods.py
@@ -381,7 +381,7 @@ class TestTypeSelectorMethod(SelectorMethod):
 class StateSelectorMethod(SelectorMethod):
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
-        self.macros_were_modified: Optional[List[str]] = None
+        self.modified_macros: Optional[List[str]] = None
 
     def _macros_modified(self) -> List[str]:
         # we checked in the caller!
@@ -394,38 +394,42 @@ class StateSelectorMethod(SelectorMethod):
 
         modified = []
         for uid, macro in new_macros.items():
-            name = f'{macro.package_name}.{macro.name}'
             if uid in old_macros:
                 old_macro = old_macros[uid]
                 if macro.macro_sql != old_macro.macro_sql:
-                    modified.append(f'{name} changed')
+                    modified.append(uid)
             else:
-                modified.append(f'{name} added')
+                modified.append(uid)
 
         for uid, macro in old_macros.items():
             if uid not in new_macros:
-                modified.append(f'{macro.package_name}.{macro.name} removed')
+                modified.append(uid)
 
-        return modified[:3]
+        return modified
+
+    def recursively_check_macros_modified(self, node):
+        for macro_uid in node.depends_on.macros:
+            if macro_uid in self.modified_macros:
+                return True
+            macro = self.manifest.macros[macro_uid]
+            if len(macro.depends_on.macros) > 0:
+                return self.recursively_check_macros_modified(macro)
+            else:
+                return False
+        return False
 
     def check_modified(
         self,
         old: Optional[SelectorTarget],
         new: SelectorTarget,
     ) -> bool:
-        # check if there are any changes in macros, if so, log a warning the
-        # first time
-        if self.macros_were_modified is None:
-            self.macros_were_modified = self._macros_modified()
-            if self.macros_were_modified:
-                log_str = ', '.join(self.macros_were_modified)
-                logger.warning(warning_tag(
-                    f'During a state comparison, dbt detected a change in '
-                    f'macros. This will not be marked as a modification. Some '
-                    f'macros: {log_str}'
-                ))
+        # check if there are any changes in macros, if so, log them the first time
+        if self.modified_macros is None:
+            self.modified_macros = self._macros_modified()
 
-        return not new.same_contents(old)  # type: ignore
+        different_contents = not new.same_contents(old)
+        upstream_macro_change = self.recursively_check_macros_modified(new)
+        return different_contents or upstream_macro_change  # type: ignore
 
     def check_new(
         self,
--- a/core/dbt/graph/selector_methods.py
+++ b/core/dbt/graph/selector_methods.py
@@ -408,6 +408,10 @@ class StateSelectorMethod(SelectorMethod):
         return modified
 
     def recursively_check_macros_modified(self, node):
+        # check if there are any changes in macros the first time
+        if self.modified_macros is None:
+            self.modified_macros = self._macros_modified()
+        
         for macro_uid in node.depends_on.macros:
             if macro_uid in self.modified_macros:
                 return True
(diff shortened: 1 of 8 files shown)


core/dbt/graph/selector_methods.py
```diff
@@ -394,44 +392,71 @@ def _macros_modified(self) -> List[str]:
...
+        if hasattr(new, "same_database_representation"):
+            return not new.same_database_representation(old)  # type: ignore
+        else:
+            return False
+
+    def check_modified_macros(self, old: Optional[SelectorTarget], new: SelectorTarget) -> bool:
+        return self.recursively_check_macros_modified(new)
+
+    def check_new(self, old: Optional[SelectorTarget], new: SelectorTarget) -> bool:
         return old is None
```
nathaniel-may, line 460: I know this was here already, but it seems pretty unnecessary. Maybe change `L471` to the function body of this method.

core/dbt/graph/selector_methods.py
```diff
@@ -443,8 +468,13 @@ def search(
             )
 
         state_checks = {
-            'modified': self.check_modified,
             'new': self.check_new,
```
nathaniel-may, line 471:
```suggestion
            # maybe with some sort of comment here.
            'new': old is None,
```

test/integration/062_defer_state_test/changed_models/table_model.sql
```diff
@@ -1,2 +1,3 @@
 {{ config(materialized='table') }}
 select * from {{ ref('ephemeral_model') }}
+-- {{ my_macro() }}
```
nathaniel-may, line 3: Just want to double check that these should be sql commented but not jinja commented.

nathaniel-may (changes requested):

core/dbt/graph/selector_methods.py
```diff
@@ -394,44 +392,71 @@ def _macros_modified(self) -> List[str]:
...
+    def check_modified_database_representations(
+        self, old: Optional[SelectorTarget], new: SelectorTarget
     ) -> bool:
+        if hasattr(new, "same_database_representation"):
+            return not new.same_database_representation(old)  # type: ignore
+        else:
+            return False
+
+    def check_modified_macros(self, old: Optional[SelectorTarget], new: SelectorTarget) -> bool:
+        return self.recursively_check_macros_modified(new)
```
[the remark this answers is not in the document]
kwigley: Unfortunately, this function is called with 2 args by default
https://github.com/dbt-labs/dbt/blob/0efaaf7dafccb3329ade084a14218e73e8e83b08/core/dbt/graph/selector_methods.py#L468-L469
^^ `checker` here is assigned a function by looking up in this map
https://github.com/dbt-labs/dbt/blob/4fe16a67e89ee0d5ba1602358d97fa60e9634b87/core/dbt/graph/selector_methods.py#L470-L478
Dunno the best way to handle this, but a pythonic convention for args or variables that won't be used is `_` 🤷 

```suggestion
    def check_modified_macros(self, _: Optional[SelectorTarget], new: SelectorTarget) -> bool:
        return self.recursively_check_macros_modified(new)
```

core/dbt/graph/selector_methods.py
kwigley: I think this needs to be a function/lambda still. maybe?
```suggestion
            'new': lambda old, _: old is None,
```

kwigley (approved):
nice! looks good

core/dbt/graph/selector_methods.py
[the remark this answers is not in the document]
nathaniel-may: yeah I'll take the underscore given the situation. But also, oof.

core/dbt/graph/selector_methods.py
nathaniel-may: ah yeah that might be right.

core/dbt/graph/selector_methods.py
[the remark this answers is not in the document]
kwigley: oof is right

test/integration/062_defer_state_test/changed_models/table_model.sql
jtcohen6: Yup! I still want the macro to be picked up as a dependency, just without actually doing anything. I can make this clearer with a comment

core/dbt/graph/selector_methods.py
jtcohen6: lambda's a neat idea. I'll leave a comment, too

core/dbt/graph/selector_methods.py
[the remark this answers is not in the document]
jtcohen6: pythonic it is!

jtcohen6:
Thanks for the reviews @nathaniel-may @kwigley! I added some comments where the code wasn't clear.

I ended up renaming `state:modified.database_representations` to `state:modified.relation`. I think I like it better that way. Relation is a wonkier word, but it's [one that dbt users should be familiar with](https://docs.getdbt.com/reference/dbt-classes#relation) as indicating `database.schema.table`.

I spent slightly too long thinking about which of these sub-selector names should be singular and which should be plural. I decided to use "per node" as the frame, such that:
- one `body` per node
- multiple `configs` per node
- multiple `persisted_descriptions` per node (potentially relation + columns)
- one `relation` per node
- multiple `macros` that a node may depend on

nathaniel-may (approved):

PR #3566, opened 2021-07-13
Adding missing v0.20.0 files back to `develop`

leahwicz:
### Description

When releasing v0.20.0RC2 and v0.20.0, we ran CI where it committed straight to the release branch and not to `develop`. This PR ports over those changes so we don't miss them as we move forward.

❗ As an action item, I'm going to make sure CI always commits to `develop` and then cherry-picks the changes to the release branch so we don't have this issue going forward. @kwigley if you have thoughts on this , let me know.

### Checklist
 ~- [ ] I have run this code in development and it appears to resolve the stated issue~
 ~- [ ] This PR includes tests, or tests are not required/relevant for this PR~


.bumpversion.cfg
```diff
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 0.20.0rc1
+current_version = 0.20.0
```
jtcohen6, line 2: This will be bumped to `0.21.0b1` when we cut v0.20.0-b1 next week. What do you think about bumping this to `0.20.1a1` for the time being, on `develop` only? Silly and unnecessary?

jtcohen6 (approved):
One question that's more of a comment. Overall this LGTM.

Perhaps it makes sense to stage the release into two commits:
1. add docker requirements, update changelog
2. bumpversion

Then the `dbt-release` process can cherry-pick the first of those commits onto `develop`, too, as one of its final steps (to avoid annoying rollbacks if the release process fails partway through).

.bumpversion.cfg
leahwicz: The only concern that I would have is that it might confuse people that there is an actual `0.21.0a1` release but if we never do alpha releases then I guess that would be less confusing

leahwicz:
> Perhaps it makes sense to stage the release into two commits:

I like the idea of making the commits separate. We can do that. I think what I envision the release process looking like is making all these commits on `develop` and then cherry-picking them to the release branch instead. We should easily be able to flip that direction today.

I want to add logic in the workflow to check for if steps have been run already so it make it easy to rerun if a failure happens. For this it should be easy to check if we already did a commit and cherry-pick and skip ahead if that is the case

.bumpversion.cfg
kwigley: not silly! I've had the same thought, we can determine if folks are installing directly from source this way as well.

kwigley (approved):

.bumpversion.cfg
leahwicz: Alright! Let's do it!

.bumpversion.cfg
```diff
@@ -1,23 +1,27 @@
 [bumpversion]
-current_version = 0.20.0rc1
+current_version = 0.21.0a1
 parse = (?P<major>\d+)
 	\.(?P<minor>\d+)
 	\.(?P<patch>\d+)
-	((?P<prerelease>[a-z]+)(?P<num>\d+))?
```
leahwicz, line 6: @kwigley that worked! Thanks!

PR #3583, opened 2021-07-15
Refactor relationships test to support where config

### LINKED ISSUE #3579: `where` argument for relationship test produces invalid sql
### Describe the bug
While testing out the new `where` config with a relationships test, I get a SQL complication error.  The issue seems to be that one of the subqueries is aliased twice, see below.

### Steps To Reproduce
- Find an existing relationships test
- Add `where: 1 = 1` as a config to the test
- Run the test.

### Expected behavior
The test to run without error.

For the following YAML:

```yaml
version: 2

models:
  - name: stg_jaffle_shop__orders
    description: "this is the orders table"
    columns:
      - name: customer_id
        tests:
          - relationships:
              to: stg_jaffle_shop__customers
              field: customer_id
              where: 1 = 1
```

The following is run against the DB

```sql
select
      count(*) as failures,
      count(*) != 0 as should_warn,
      count(*) != 0 as should_error
    from (
      
    
    

select
    child.customer_id

from (select * from JD_DBT.DEV_JOSH.stg_jaffle_shop__orders where 1 = 1) stg_jaffle_shop__orders as child

left join stg_jaffle_shop__customers as parent
    on child.customer_id = parent.customer_id

where child.customer_id is not null
  and parent.customer_id is null



      
    ) dbt_internal_test
```

The issue specifically seems to be the `stg_jaffle_shop__orders as child` where the table name is used as an alias and then it's attempt to be aliased again with `as child`.  The error returned is

```
  001003 (42000): SQL compilation error:
  syntax error line 13 at position 97 unexpected 'as'.
  syntax error line 24 at position 4 unexpected ')'.
  ```

### System information
- [ ] postgres
- [ ] redshift
- [ ] bigquery
- [X] snowflake

```
installed version: 0.20.0
   latest version: 0.20.0

Up to date!

Plugins:
  - bigquery: 0.20.0
  - snowflake: 0.20.0
  - redshift: 0.20.0
  - postgres: 0.20.0
```

**The operating system you're using:** Mac

**The output of `python --version`:** Python 3.8.10

jtcohen6:
resolves #3579

@jaypeedevlin Does this make sense to you?

Any tests that alias `{{ model }}` themselves will run into this issue with the `where` config. A better way of doing this might be to give the option of providing a custom alias when templating `{{ model }}`, though I'm not sure how exactly we'd go about pulling that off.

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,11 @@ Contributors:
 - [@JLDLaughlin](https://github.com/JLDLaughlin) ([#3473](https://github.com/fishtown-analytics/dbt/pull/3473))
 - [@jmriego](https://github.com/jmriego) ([#3526](https://github.com/dbt-labs/dbt/pull/3526))
 
+## dbt 0.20.1 (Release TBD)
+
+### Fixes
+- Fix `where` config with `relationships` test by refactoring test SQL ([#3579](https://github.com/fishtown-analytics/dbt/issues/3579), [#3579](https://github.com/fishtown-analytics/dbt/pull/3579))
+
 ## dbt 0.20.0 (July 12, 2021)
 
 ### Fixes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,7 +23,7 @@ Contributors:
 ## dbt 0.20.1 (Release TBD)
 
 ### Fixes
-- Fix `where` config with `relationships` test by refactoring test SQL ([#3579](https://github.com/fishtown-analytics/dbt/issues/3579), [#3579](https://github.com/fishtown-analytics/dbt/pull/3579))
+- Fix `where` config with `relationships` test by refactoring test SQL ([#3579](https://github.com/fishtown-analytics/dbt/issues/3579), [#3583](https://github.com/fishtown-analytics/dbt/pull/3583))
 
 ## dbt 0.20.0 (July 12, 2021)
 
--- a/core/dbt/include/global_project/macros/schema_tests/relationships.sql
+++ b/core/dbt/include/global_project/macros/schema_tests/relationships.sql
@@ -1,16 +1,23 @@
 
 {% macro default__test_relationships(model, column_name, to, field) %}
 
+with child as (
+    select * from {{ model }}
+    where {{ column_name }} is not null
+),
+
+parent as (
+    select * from {{ to }}
+)
+
 select
     child.{{ column_name }}
 
-from {{ model }} as child
-
-left join {{ to }} as parent
+from child
+left join parent
     on child.{{ column_name }} = parent.{{ field }}
 
-where child.{{ column_name }} is not null
-  and parent.{{ field }} is null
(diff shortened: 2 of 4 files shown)


jaypeedevlin:
@jtcohen6 this is a neat way of getting this fixed, nice work!

jtcohen6:
@jaypeedevlin Thanks for taking a look!

I know that CTEs nested within a subquery aren't supported by all adapters. It's possible that this will require some adapter maintainers to (re)implement `<adapter>__test_relationships` for 0.20.1, even if they didn't have to for 0.20.0. I think that's ok—we haven't made adapter-level guarantees around breaking changes—I'll just make sure to call it out in the changelog.

kwigley (approved):
nice, looks good!

Not 100% sure what is up with the test suite, will poke around a little.

gshank:
The event tests have a bug and also seem to be randomly missing an active user. I've been seeing that in my branch too.

jtcohen6:
> Not 100% sure what is up with the test suite, will poke around a little.

I believe both are non-deterministic failures:
- `Rate limits exceeded` on BigQuery (should we try to add this to the [set of transient errors](https://github.com/dbt-labs/dbt/blob/1d37c4e55504a9021507c013f6d1232b30ba409a/plugins/bigquery/dbt/adapters/bigquery/connections.py#L44-L49)?
- The `experimental_parser` event isn't jiving with some unit tests + `test/integration/033_event_tracking_test` (https://github.com/dbt-labs/dbt/issues/3561)

gshank (approved):

leahwicz:
>     * `Rate limits exceeded` on BigQuery (should we try to add this to the [set of transient errors](https://github.com/dbt-labs/dbt/blob/1d37c4e55504a9021507c013f6d1232b30ba409a/plugins/bigquery/dbt/adapters/bigquery/connections.py#L44-L49)?

The list you linked to is called `RETRYABLE_ERRORS` and if we truly do retry on those errors, we are most likely going to hit the same error again unless we back-off and try again later

jtcohen6:
> we are most likely going to hit the same error again unless we back-off and try again later

Good point—I was grasping at a hopeful link there, probably not the right one.

I just quickly looked at docs for this error: https://cloud.google.com/bigquery/docs/troubleshoot-quotas#concurrent_queries_quota_errors

> To prevent this error, switch the query to batch mode. The query will be queued and will execute when resources are available. Batch queries don't count towards your concurrent rate limit, which can make it easier to start many queries at once.

We already have a credential option for `priority: batch` ([here](https://github.com/dbt-labs/dbt/blob/1d37c4e55504a9021507c013f6d1232b30ba409a/plugins/bigquery/dbt/adapters/bigquery/connections.py#L323-L325)). Adding this to our BigQuery credentials for integration tests ([here](https://github.com/dbt-labs/dbt/blob/1d37c4e55504a9021507c013f6d1232b30ba409a/test/integration/base.py#L245-L267)) could be a quick win.

Issue #3502, opened 2021-06-28
Turn on performance improvements by default


jtcohen6:
Namely:
- partial parsing
- experimental parser

This will require accounting for all known "fast-and-incorrect" edge cases and either (or both) of:
- fixing them
- documenting them, and what to do when they happen

This does not require accounting for all known "slow-but-correct" edge cases. While strictly better, those can come later, and they won't yield incorrect results in the meantime.

leahwicz:
* Can we expand the experimental parsing scope?
* Depends on the results we get back from metrics

leahwicz:
Where is the flag for this? In the profile?

leahwicz:
Let's split this into 2 different tickets:
* Experimental parsing -> would be a 1 
* Partial parsing -> would be a 4-8 with test updates and some additional code changes

jtcohen6:
> Where is the flag for this? In the profile?

Yes! I think CLI flag + profile + env var (+ other configuration mechanism?). I'm looking at our discussion in #2990 re: coordinating configs.

> Can we expand the experimental parsing scope?

One thing I had discussed with @nathaniel-may last week: rather than simply setting `use_experimental_parser: True` for everyone, I think we'll actually want to:
- Move proven parts of the experimental parser into the default parser (and delete old default code) — it's no longer experimental, everyone gets to have it
- Keep around the `use_experimental_parser` flag/config/var, still `False` by default, as a way for users to always opt into the next phase of not-yet-proven experimentation

Noting this here because it's different from the way we'd turn on partial parsing by default—that we'd more straightforwardly accomplish by setting the default value of `partial_parse` to `True`.

leahwicz:
The experimental parser remaining work is in this ticket: https://github.com/dbt-labs/dbt/issues/3377

jtcohen6:
Remaining **must-haves** before we turn these on by default. We can fix/handle (ideal), or document limitations with path to turn off (less desirable).

- **Partial parsing** "fast-and-incorrect" edge cases:
    - `generate_x_name` macro changes (https://github.com/dbt-labs/dbt/issues/3810)
    - env vars needed at parse time (ref/source/config or yaml property) have changed
    - volatile variables (`run_started_at`, `invocation_id`, etc) needed at parse time (https://github.com/dbt-labs/dbt/issues/2330)

- **Experimental parser:**
    - Detect if user has overridden ref/source/config macros; if so, disable experimental parser
    
As far as actually turning on:
[...]

PR #3602, opened 2021-07-20
Add Performance Regression Testing [Rust]

### LINKED ISSUE #3054: Add Performance Regression Tests
### Describe the feature
Add testing to detect whether key performance metrics are within an acceptable threshold.

### Additional context
While aiming to improve the startup time of dbt with large projects, adding regression testing will prevent us from making future changes that slow down startup times below our acceptable threshold. Testing the parse time of large projects is a good place to start with the additions of the [dbt-timing-project](https://github.com/fishtown-analytics/dbt-timing-project) and the `dbt parse` command.

### Who will this benefit?
This will benefit all users because it protects them from performance regressions in future releases. This benefits dbt contributors because it will catch performance regressions before merging and releasing.

nathaniel-may:
resolves #3054

### Description

Adds the first iteration of performance regression testing. The test suite consists of the following components:

1. Target dbt projects that demonstrate performance strain
2. Rust runner with target dbt commands, and regression detection logic
3. Github Actions workflow for orchestration across branches

These tests will run nightly via github actions, and can be triggered through the UI. They will run all of these commands on all of the dbt projects with dbt installed from develop and with dbt installed from our baseline branch which is currently set to `0.20.latest`. It will trigger a failure if the dev branch is slower than the baseline version by a set threshold. Failures could be reported via email, slack and whatever else we think of.

This PR is the first to include Rust in this repo. However, you can rest assured that contributors do not need to install the rust tool chain to continue contributing to dbt. this code relies on dbt, not the other way around.

This PR Is to be merged with the dummy projects. The regressions it is catching now are likely due to the fact the dummy projects do not induce any stress on dbt and thus have pretty high variance in their run times. Immediately after, I will put up a PR with at least one real project that may have thousands of files, but no other changes. From there the exact threshold numbers may still need to be tuned, but the variance in run times should be far more reasonable.

### Suggested Path to Review this PR:
1. Look at `.github/workflows/performance.yml`. This is the highest level orchestration of the tests and outlines the workflow of how we are conducting these regression tests
2. Look at `performance/runner/src/main.rs`. This is the entry point to the rust runner that we saw in the github action. It has two sub commands `measure` and `calculate`. First look at the exit criteria here before moving on to the implementation of these sub commands.
3. Look at `performance/runner/src/measure.rs::measure` since this is the entry point for the sub-command that is called first in the workflow. 
4. Look at `performance/runner/src/calculate.rs::regressions` since this is the entry point for the other sub-command. 

### Requested Feedback
- At the highest level, is this approach the way we want to do this? If there's an alternative way you think we should try, now is the time to say something.
- Are there any improvements I can make to my github action workflow? I'm not a power user yet and I would love to see if there are any improvements we can make to this one.
- Look at the output of the action. Are there any changes to the artifacts or stdout that you would like to see?
- Where should I add more comments? I especially want to add comments to places if you don't know _why_ something is happening.
- Is any code particularly ugly? I love a good refactor.
- Are there any places you think might be particularly terrible to debug? I tried my best to make our future debugging efforts as painless as possible.
- Anywhere that my Rust could improve. I'm not fluent in all the corners of Rust yet so if you see something I could be doing better, let me know.

### TODO: 
- [x] define precise failure characteristics
- [x] test actions workflow
- [x] write result comparison logic
- [x] use a real arg library
- [ ] remove PR workflow trigger (intended for development only here)
- [x] add error on warning to the test runner job
- [ ] [Optional] pull commands from json config file instead of hard coding in source

### Checklist


performance/runner/src/main.rs
```diff
@@ -0,0 +1,70 @@
...
+
+#[derive(Debug, Clone)]
+struct Metric<'a> {
+    name: &'a str,
+    prepare: &'a str,
+    cmd: &'a str,
+}
+
+impl Metric<'_> {
+    fn outfile(&self, project: &str) -> String {
```
leahwicz, line 13: Rust newbie but does this just return the name of the results file for the specific metric?

performance/runner/src/main.rs
```diff
@@ -0,0 +1,70 @@
+use std::{env, fs, io};
+use std::process::{Command, ExitStatus, exit};
+
+
+#[derive(Debug, Clone)]
+struct Metric<'a> {
+    name: &'a str,
+    prepare: &'a str,
```
leahwicz, line 8: Is `prepare` just like a pre-step that is run before the test starts? It looks like that is the arg name below

performance/runner/src/main.rs
nathaniel-may: that's exactly what this function does 😃

performance/runner/src/main.rs
nathaniel-may: yeah it's part of hyperfine. there's also cleanup which I can add if it turns out we need that too.

.github/workflows/performance.yml
```diff
@@ -0,0 +1,111 @@
...
+    name: Measure develop Branch
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/download-artifact@v2
+        with:
+          name: runner
+      - name: checkout develop
+        uses: actions/checkout@v2
+        with:
+          ref: 'develop'
```
leahwicz, line 67: Maybe we don't specify `develop` here and instead let schedule runs use `develop` and then also allow us to pick the branch we want in manual runs. We can have the scheduled run default to `develop`. With manual runs you can pick a branch and then it will default checkout that

nathaniel-may: that sounds great! .....how do you write that down in a workflow?

.github/workflows/performance.yml
```diff
@@ -0,0 +1,120 @@
...
+      - 'performance-regression-testing'
+  schedule:
+    # runs every day at 10:05pm
+    - cron:  '5 22 * * *'
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
+
+jobs:
+
+  test-rust:
```
leahwicz, line 17: What does the `test-rust` and `build-rust` steps do for us? I don't fully understand why they are needed for the perf testing. I know nothing about Cargo so it might be worth adding comments so it's more apparent

nathaniel-may: added comments 👍

.github/workflows/performance.yml
```diff
@@ -0,0 +1,158 @@
...
+      - name: run
+        run: ./runner measure -b dev -p ${{ github.workspace }}/performance/projects/
+      - uses: actions/upload-artifact@v2
+        with:
+          name: dev-results
+          path: performance/results/
+
+  # run the performance measurements on the latest release branch
+  # this part takes by far the longest, so we do everything we can first
+  # so the job fails fast.
```
leahwicz, line 91: It will be good to explain here why you are checking out `dbt` twice: 1) the runner is from HEAD 2) the `*.latest` branch is to test on

.github/workflows/performance.yml
```diff
@@ -0,0 +1,158 @@
...
+          path: performance/results/
+
+  # detect regressions on the output generated from measuring
+  # the two branches. Exits with non-zero code if a regression is detected.
+  calculate-regressions:
+    needs: [measure-dev, measure-baseline]
+    name: Compare Results
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
```
leahwicz, line 137: Do you need to checkout the repo in this step?

performance/projects/01_mini_project_dont_merge/dbt_project.yml
```diff
@@ -0,0 +1,38 @@
+
```
leahwicz, line 1: The name of this file makes me thing you don't want this in the repo...is that correct?

leahwicz:
Overall this LGTM as an approach. Same with the GitHub Action. I would be good with getting this in as a V1 and seeing it run over a couple weeks and watching to see what we think.

I noticed in the runs you have done, there is a regression caught. Is that actually true?

performance/projects/01_mini_project_dont_merge/dbt_project.yml
nathaniel-may: That's correct. I'll replace it with one with like 2000 files which will take an eternity to run. But not while I'm developing 🙃

nathaniel-may:
@leahwicz I think so? at least by the median runtime threshold. I think the one I set for `stddev` metric is too small if we even want to track that at all. I'll have to dig in a little more.

.github/workflows/performance.yml
nathaniel-may: it looks like I don't! good catch.

.github/workflows/performance.yml
nathaniel-may: added comments here to that effect.

nathaniel-may:
@leahwicz I actually think because the dummy projects are so small, the times aren't _actually_ measuring parse time since that probably takes up a small percentage of the end-to-end dbt process here. Once we merge a bigger project we'll have a better idea on regressions. I updated the description to reflect this.

performance/README.md
```diff
@@ -0,0 +1,16 @@
+# Performance Regression Testing
+This directory includes dbt project setups to test on and a test runner written in Rust which runs specific dbt commands on each of the projects. Orchestration is done via the GitHub Action workflow in `/.github/workflows/performance.yml`. The workflow is scheduled to run every night, but it can also be triggered manually.
+
+The github workflow hardcodes our baseline branch for performance metrics as `0.20.latest`. As future versions become faster, this branch will be updated to hold us to those new standards.
+
+## Adding a new dbt project
+Just make a new directory under `performance/projects/`. It will automatically be picked up by the tests.
```
gshank, line 7: I'm wondering if we really want to check in the thousands of files associated with the performance tests? Could we put the projects we want to test into separate repos?

gshank (commented):
In general the code looks very clean and I find the commenting good. I'm not really up to critiquing the Rust though :)

performance/README.md
nathaniel-may: yes. we absolutely could do that. The advantage would be that we don't have to download all those files when we checkout dbt, but it's also another repository to manage. I just measured that a dbt project with 2k files is roughly 6.3k to download which I don't consider a deal breaker. But I'm honestly open to both approaches.

performance/runner/src/measure.rs
```diff
@@ -0,0 +1,89 @@
...
+        2. Pair `n` projects with `m` metrics for a total of n*m pairs
+        3. Run hyperfine on each project-metric pair
+    */
+
+    // To add a new metric to the test suite, simply define it in this list:
+    // TODO: This could be read from a config file in a future version.
+    let metrics: Vec<Metric> = vec![Metric {
+        name: "parse",
+        prepare: "rm -rf target/",
+        cmd: "dbt parse --no-version-check",
```
jtcohen6, line 39: If there's a way to compare command-specific output, I'd be excited to add:
```
dbt --partial-parse parse
dbt --use-experimental-parser parse
```
For the moment, I don't think it's worth a lot of additional effort to add that in, since we plan on turning on both of these by default pretty soon (#3502).

performance/README.md
jtcohen6: I like that this offers general support for performance-testing any project. I'm thinking about specific projects that we've known to parse slowly, and how great it would be to stress-test a putative improvement in a more-rigorous manner.

As far as running this nightly: If we're going to use generated projects, I'd be open to including the [project-generation code](https://github.com/dbt-labs/dbt-timing-project) in this repo, and running the project generation as a step in the `performance` workflow.

performance/README.md
```diff
@@ -0,0 +1,16 @@
+# Performance Regression Testing
+This directory includes dbt project setups to test on and a test runner written in Rust which runs specific dbt commands on each of the projects. Orchestration is done via the GitHub Action workflow in `/.github/workflows/performance.yml`. The workflow is scheduled to run every night, but it can also be triggered manually.
+
+The github workflow hardcodes our baseline branch for performance metrics as `0.20.latest`. As future versions become faster, this branch will be updated to hold us to those new standards.
```
jtcohen6, line 4: We'll need some way to keep this up to date. We already have `0.21.latest`, though there hasn't yet been a final release from that branch.

I'm wondering if it would be preferable to install the latest stable release of dbt, rather than checking out a specific branch?

jtcohen6 (approved):
Very cool! I left a few questions for my own edification, none of which should block merging. I'm excited to have this in the codebase + testing suite.

performance/README.md
nathaniel-may: @jtcohen6 that is a good next step, however most places that do this kind of performance work (and publish it) usually check in projects to prove the metrics aren't changing because the project is changing. We could go against the grain on that one, but I'm inclined to check it in if it doesn't cause any tangible issues.

performance/runner/src/measure.rs
nathaniel-may: totally. Hyperfine can already run two things and compare them just like that. So to build that, I would have a third sub command that follows a different flow, but reuses pretty much all the same types so it wouldn't be hard to build at all if we wanted to track the performance differences. It wouldn't really be regression testing, but I think it's still valuable.

performance/README.md
nathaniel-may: yeah we could do that instead. Definitely less manual than this approach.

performance/README.md
jtcohen6: Ok cool, I buy that (and remember you having told me this). I'm all for doing this in a way that follows existing practice.

performance/README.md
nathaniel-may: Maybe we should switch to this approach after a version or two so we know we haven't introduced any slippage in the metrics we're measuring. 🤔

PR #3604, opened 2021-07-21
Fix flaky test related to tracking events

### LINKED ISSUE #3561: Fix flaky tracking integration tests
Postgres integration test suite 033 matches expected and actual tracking events. However, the experimental parser fires its tracking event non-deterministically. Commits 7a32129, and c43c5fa were supposed to fix this, yet [circleci](https://app.circleci.com/pipelines/github/dbt-labs/dbt/3023/workflows/cee21f06-bbc4-4b47-a12b-339259f430b9/jobs/44132) still sees inconsistent failures. 

This set of tests may not actually be all that useful as is and we may want to rework how we test events going forward anyway.

kwigley:
resolves #3561

### Description
- ~skips~ comment out tests related to event tracking, these are testing non-deterministic behavior
- allow null users when calling tracking event code, skip sending events downstream if this is the case

### Checklist

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -10,7 +10,7 @@ on:
       - 'develop' 
       - '*.latest' 
       - 'releases/*'
-  pull_request_target:
+  pull_request:
     branches: 
       - 'develop' 
       - '*.latest' 
--- a/test/integration/033_event_tracking_test/test_events.py
+++ b/test/integration/033_event_tracking_test/test_events.py
@@ -1,987 +1,986 @@
-from test.integration.base import DBTIntegrationTest, use_profile
-import pytest
-import hashlib
-import os
+# NOTE: turning off event tracking tests!
+# from test.integration.base import DBTIntegrationTest, use_profile
+# import hashlib
+# import os
 
-from unittest.mock import call, ANY, patch
+# from unittest.mock import call, ANY, patch
 
-import dbt.exceptions
-import dbt.version
-import dbt.tracking
-import dbt.utils
-
-pytestmark = pytest.mark.skip("Experimental parser generates events non-deterministically, skipping.")
-
-# immutably creates a new array with the value inserted at the index
-def inserted(value, index, arr):
-    x = []
-    for i in range(0, len(arr)):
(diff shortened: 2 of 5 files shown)


test/integration/033_event_tracking_test/test_events.py
```diff
@@ -9,6 +10,7 @@
 import dbt.tracking
 import dbt.utils
 
+pytestmark = pytest.mark.skip("Experimental parser generates events non-deterministically, skipping.")
```
leahwicz, line 13: Does this skip all the tests in this file? I don't see `pytestmark` being used anywhere so I'm not sure how that works

kwigley: That is correct, it marks all tests in the module as skipped!

https://docs.pytest.org/en/stable/reference.html#globalvar-pytestmark

leahwicz (approved):

test/integration/033_event_tracking_test/test_events.py
kwigley: Well, I thought it did. https://github.com/dbt-labs/dbt/pull/3604/checks?check_run_id=3124252846

nathaniel-may (approved):

leahwicz (approved):

nathaniel-may (approved):
The fmt changes seem unintentional, but it doesn't actually matter. lgtm.

.github/workflows/tests.yml
```diff
@@ -6,28 +6,29 @@ name: dbt Tests
...
-    branches: 
-      - 'develop' 
-      - '*.latest' 
-      - 'pr/*'
-      - 'releases/*'
+    branches:
+      - "develop"
+      - "*.latest"
+      - "releases/*"
+  pull_request:
```
kwigley, line 13: @leahwicz wanted to run this change by you. I was confused why tests weren't picking up my change, I realized the behavior for the `pull_request_target` event checks out the target branch for the checkout step. I think this is what we want, I'd like to confirm though!

kwigley:
Created #3631 to address root cause.

PR #3607, opened 2021-07-21
Skateboard for local filesystem adapter [WIP]!

iknox-fa:
resolves #

### Checklist

### DIFF
--- a/core/dbt/config/profile.py
+++ b/core/dbt/config/profile.py
@@ -4,7 +4,7 @@ import os
 
 from dbt.dataclass_schema import ValidationError
 
-from dbt.clients.system import load_file_contents
+from dbt.clients.storage_adaptor import StorageAdapter as SA
 from dbt.clients.yaml_helper import load_yaml_text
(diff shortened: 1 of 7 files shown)


core/dbt/config/profile.py
```diff
@@ -52,7 +52,7 @@ def read_profile(profiles_dir: str) -> Dict[str, Any]:
     contents = None
     if os.path.isfile(path):
         try:
-            contents = load_file_contents(path, strip=False)
+            contents = SA.adapter.load_file_contents(path, strip=False)  # type: ignore
```
cmcarthur, line 55: I think that profile reading should not go through the SA as strange as that sounds, or at least not go through the same load_file_contents method.

In a thin client world, the thin client will sync project file changes only -- the profile will need to come from somewhere else, and won't be stored alongside project files.

iknox-fa:
CLosing PR (re-orging the git work for this effort to make it easier to PR)

PR #3608, opened 2021-07-21
Fix error handling in `dbt build`

### LINKED ISSUE #3598: dbt build: Correct non-fast-failing behavior to match other tasks such as run
`dbt build` is exiting earlier than it should when faced with a broken project.  This leads to an inconsistent user experience, and an incomplete process artifacts (`run_results.json`).  Work done to correct this seemed to lead into larger discussions about the expected behavior of the `fail-fast` flag. 

References:
- https://github.com/dbt-labs/dbt/pull/3490#pullrequestreview-705849186
- https://github.com/dbt-labs/dbt/pull/3490#pullrequestreview-709735773

jtcohen6:
resolves #3598

### Description

I saw that the `BuildTask` was inheriting from `CompileTask`, which sets:
https://github.com/dbt-labs/dbt/blob/9f716b31b3e0ca57a3722809f8c3405a7b5752f8/core/dbt/task/compile.py#L36-L38

That's the cause of the weird behavior I was seeing when reviewing #3490. We could just override that method for the BuildTask, but what I think we really want is for `dbt build` to operate at parity with `dbt run`, and take all of its cues—including how it handles hooks, deferral, etc, etc. I confirmed that `TestTask`, `SeedTask`, `SnapshotTask` also inherit from `RunTask`, so it feels appropriate that `BuildTask` would as well. Let me know if I'm totally off base here.

Tiny thing: I confirmed that a failed model will be handled and downstream resources will be skipped. Now that tests can be skipped, in addition to models/snapshots, I figured we should tweak the `SKIP` log message to be more inclusive of tests.

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 ## dbt 0.21.0 (Release TBD)
 
 ### Features
-- Add `dbt build` command to run models, tests, seeds, and snapshots in DAG order. ([#2743] (https://github.com/dbt-labs/dbt/issues/2743), [#3490] (https://github.com/dbt-labs/dbt/issues/3490))
+- Add `dbt build` command to run models, tests, seeds, and snapshots in DAG order. ([#2743] (https://github.com/dbt-labs/dbt/issues/2743), [#3490] (https://github.com/dbt-labs/dbt/issues/3490), [#3608](https://github.com/dbt-labs/dbt/issues/3608))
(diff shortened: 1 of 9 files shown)


iknox-fa:
Good catch! I think I chose to inherit from `CompileTask` ealy on when I thought I needed the selection spec stuff found there.

iknox-fa (approved):

jtcohen6:
Some failing tests, but they all look related to the flakey issues we're currently tackling elsewhere, and unrelated to the changes in this PR.

Issue #3564, opened 2021-07-13
Variables not picked up by tests


Limess:
### Describe the bug
Compilation fails when using variables defined in dbt_project.yaml in tests.

This was working prior to the upgrade to 0.20.0

### Steps To Reproduce
Variable in dbt_project.toml

```yaml
vars:
  signal_dbt:
    # the number of days we are legally allowed to retain data
    tracking_event_retention_period_days: 1460
```

Test:

```yaml
        tests:
          - not_null:
              where: 'day >= CURRENT_DATE - interval ''{{ var("tracking_event_retention_period_days") }} days'''
```

### Expected behavior
The variable to be used from dbt_project.toml

### Screenshots and log output

```
Running with dbt=0.20.0
Encountered an error:
Compilation Error
  Invalid test config given in models/marts/core/core.yml:
  	Required var 'tracking_event_retention_period_days' not found in config:
  Vars supplied to <Configuration> = {}
  	@: UnparsedNodeUpdate(original_file_path='model...ne)
```

### System information
**Which database are you using dbt with?**
- [ ] postgres
- [x] redshift
- [ ] bigquery
- [ ] snowflake
- [ ] other (specify: ____________)

**The output of `dbt --version`:**
```
installed version: 0.20.0
   latest version: 0.20.0

Up to date!

Plugins:
  - redshift: 0.20.0
  - postgres: 0.20.0

```

**The operating system you're using:**

Mac OS big Sur 11.2.3

**The output of `python --version`:**

3.8.5

jtcohen6:
Thanks for the quick report @Limess! I was able to reproduce this locally.

Given that this works if I pass the `var` from the CLI:
```
dbt test --vars 'tracking_event_retention_period_days: 1460'
```

This appears to be a bug with the rendering context used for test arguments here:
https://github.com/dbt-labs/dbt/blob/f460d275baa4df8cf8713f27e198a3577d6ea71c/core/dbt/parser/schema_test_builders.py#L231

I'm hopeful that this is a quick fix, and something we could easily include in a patch release.

codigo-ergo-sum:
We are experiencing the same issue when testing an upgrade to dbt .20 as it's very common behavior for us to use vars defined in dbt_project.yml in our tests.

An example of a test that is failing for us:

`
      - dbt_utils.expression_is_true:
          severity: warn
          tags:
            - known_fail_or_warn
          expression: "parent_classification IN ('Revenue', 'Expense')"
[...]

Issue #3557, opened 2021-07-11
Reorganize nodes, node.config in manifest


jtcohen6:
### Describe the feature

Our work on #2401 promises to rationalize the ways that node configs interact with node-level properties. In the general case, a config should be something set by the user within a file's Jinja `config()` block or in a yaml file (including `dbt_project.yml`). Some of those configs sit inside the `node.config` object, waiting to be accessed within the Jinja context via `config.get()`; other configs will be copied over to node-level properties (`tags` today, `description` and `meta` coming soon).

This also gives us the opportunity to rationalize the ways that nodes and their configs are represented today in `manifest.json`:

- There should be a separate `ModelConfig` that inherits from base [`NodeConfig`](https://github.com/dbt-labs/dbt/blob/f460d275baa4df8cf8713f27e198a3577d6ea71c/core/dbt/contracts/graph/model_config.py#L324).
- Configs should only be set on the resource types for which they are relevant. E.g. `full_refresh` on models + seeds only.
- If a config is empty/null, it _should not_ be included in `node.config`. (This could trim down `manifest.json`.)
- Adapter-specific configs with not-empty default values _should_ be included in `node.config`. For simplicity, adapter configs apply to all nodes. (This would add some bulk to `manifest.json`.)
- All official, documented configs (e.g. `incremental_strategy`, `unique_key` for models) should be specified in a python dataclass, and their default values stored in `manifest.json` accordingly. (Today, `incremental_strategy` default values are set [in Jinja macros](https://github.com/dbt-labs/dbt/blob/f460d275baa4df8cf8713f27e198a3577d6ea71c/plugins/snowflake/dbt/include/snowflake/macros/materializations/incremental.sql#L4). It would be amazing, though perhaps more tricky than it's worth, to include the default `incremental_strategy` config on models that are `materialized: incremental`, and only those models.)
- Resources that aren't persisted in the database (i.e. `analyses`) shouldn't have node configs like `materialized`/`database`/`schema`. (As of v0.20, tests now use the `'test'` materialization, and `--store-failures` means they can have an associated database representation, so those configs can be relevant. Again, it would be amazing to set those configs if and only if `store_failures` is enabled.)
- **Big question:** Right now, the top-level keys in the manifest for project resources are `nodes`, `sources`, `macros`, `docs`, `exposures` (honorable mentions: `selectors`, `disabled`). Should `nodes` be split up into `models`, `tests`, `seeds`, `snapshots`, `analyses`?

[...]

PR #3616, opened 2021-07-23
Configs in schema files

### LINKED ISSUE #2401: Set configs in schema.yml files
### Describe the feature
Let's bring the `schema.yml` spec and node configs together! Right now, some configs can only be set for a node in the node's `.sql` file with `config()` or in the `dbt_project.yml` file. These configs include:
 - materialized
 - enabled
 - tags
 - database-specific configs (bind, sort, dist, etc)
 - etc

These configs _can not_ be configured in the schema.yml file. At the same time, there are some configs that can _only_ be set in a `schema.yml` file:
 - description
 - columns (tags, descriptions, tests, etc)
 - tags (for sources)

Instead, it would be ideal if these configs could be supplied in either place, with a sensible and well-defined precedence rules.

### Example usage:

```
-- models/my_model.sql
/*
  Configure the description and columns for a model inline in a config.
  Note: I'd probably advise against doing this, but it should be possible
*/

{{ config(
  description="This is my model description",
  columns=[
    [
      {"name": "id", "description": "The primary key"}
    ]
) }}

select ...
```

```
# models/schema.yml
version: 2
models:
 - name: my_model
   config:
      tags: ["nightly"]
      materialized: view
```

Some configs, like `tags` or `freshness` in a source are set at the root level of the source. Today, this looks like:
```
version: 2
sources:
 - name: my_source
   tags: [my_tag]
```

We should continue to support this syntax, but it would also be good to allow a `config` level key which scopes these configurations, eg:

```
version: 2
sources:
 - name: my_source
   config:
      tags: [my_tag]
```

We should _not_ allow these configs to be configured at both the root-level and the `config` level for parsing simplicity.

### Additional context
I fear that this might be a really big departure from how parsing works today. Some fields, like `description` are patched into model objects. Would this feature require us to drastically change how we parse nodes in dbt?

gshank:
resolves #2401

### Description

This adds the ability to set configs in schema files for models, seeds, snapshots, and analyses. It does not address configs in macros, sources, or tests, which will be addressed separately.

### Checklist


gshank:
Notes on implementation:
I moved the add_patch, add_macro_patch, and add_source_patch code into the schema parser file because they were more appropriate there.
I changed the 'render_test_update' code for unique and not_null to make better use of the 'update_parsed_node_config' method.
I added a 'meta' attribute to NodeConfig because there were not very many code implications, but did not change any of the existing 'meta' attributes because the behavior was inconsistent.

jtcohen6 (commented):
This is great work! I had a lot of fun taking this for a spin. We'll want to significantly revise our documentation on nodes vs. properties, and that's a good thing :)

I'm going to spend some more time playing around with this. For the moment, my biggest comment is that **we should merge `node.config.meta` into `node.meta`**. It doesn't make sense to leave these as separate, unrelated values. This might have some confusing behavior, so let's find ways to do reasonable things. I'm thinking we should treat the `meta` key in the `models` dict as being at the "same level as" the `config.meta` key, and raise an error if both are set. For example:
```yml
models:
  - name: my_model
    meta:
      something: blabla
    config:
      meta:
        something: lala
```
We should raise an error here, since we don't know whether to prefer `blabla` or `lala`. We should do it even if those `meta` dictionaries contain different keys. Any in-file `config(meta = ...)` should take precedence.

I'm also going to open some follow-on issues that are outside the scope of this specific PR, related to the larger reconciliation effort proposed in #2401 and kicked off here.
- Support `config:` on tests. That might get us "for free" the ability to set the `meta` property on tests (https://github.com/dbt-labs/dbt/issues/3629)
- Support `config:` on sources, and then turn existing node-level properties into configs that can be set in `dbt_project.yml` + `config` blocks. When I say this, I'm _especially_ thinking about source properties like `database`, `loaded_at_field`, etc. (I think it would make sense to reserve some properties—`description`, `columns`, `tests`—rather than try to make those work as configs... though there are some valid reasons to want those, too!)
- Support the same `config:` block-style syntax in `dbt_project.yml`, as a substitute for the `+` prefix that everyone finds confusing :)

core/dbt/parser/schemas.py
```diff
@@ -360,37 +362,10 @@ def render_test_update(self, node, config, builder):
...
-            if builder.error_if is not None:
-                node.unrendered_config['error_if'] = builder.error_if
-                node.config['error_if'] = builder.error_if
-            if builder.fail_calc is not None:
-                node.unrendered_config['fail_calc'] = builder.fail_calc
-                node.config['fail_calc'] = builder.fail_calc
-            if builder.store_failures is not None:
-                node.unrendered_config['store_failures'] = builder.store_failures
-                node.config['store_failures'] = builder.store_failures
+            config_call_dict = builder.get_static_config()
```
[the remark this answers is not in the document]
gshank: As far as I can tell from reading the code, the unrendered_config only has actual unrendered_config from dbt_project. The config from SQL files is rendered by jinja in the rendering process. So we probably do need to put config in schema files through a get_rendered step, which this pr is not doing yet. We *could* save the schema file config dictionary in rendered and unrendered form, but I think we'd have to statically parse config in SQL files to get unrendered config from there.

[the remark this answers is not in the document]
gshank: And we could save the unrendered config from tests too, but it would require refactoring that update_parsed_node(_config) codepath, because it doesn't have a source of unrendered_config for nodes.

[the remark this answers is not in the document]
gshank: Note: I think the experimental parser doesn't render configs either.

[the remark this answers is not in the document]
jtcohen6: > So we probably do need to put config in schema files through a get_rendered step, which this pr is not doing yet.

Ah. I think we'll want them to have the same rendering context as other preexisting `schema.yml` properties (e.g. `meta`, `tags`).

> We could save the schema file config dictionary in rendered and unrendered form

Yes, this is what we do for config from `dbt_project.yml`. It's not a must-have right now, just a way to keep making `state:modified` comparisons better and better.

> I think we'd have to statically parse config in SQL files to get unrendered config from there... I think the experimental parser doesn't render configs either.

Yup, I think this will come from making our tree-sitter grammar able to parse unrendered configs, detect Jinja expressions, and then pass those expressions in for rendering. This is what I'd love to work toward, ultimately: parse/extract _all_ unrendered configs, then render all configs in one big Jinja-rendering step.

jtcohen6:
Update on my comment above, with links to the promised issues:
- The same, but for sources: https://github.com/dbt-labs/dbt/issues/3662
- Symmetry with project file: https://github.com/dbt-labs/dbt/issues/3663
- Support `config` for tests, including `meta`: looks like we might tackle that as part of this PR!

core/dbt/parser/schemas.py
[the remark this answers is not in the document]
jtcohen6: Noted that long con here: https://github.com/dbt-labs/dbt/issues/3680

jtcohen6:
As discussed, there are three outstanding pieces for this PR:

- Make sure configs are being rendered (using standard `.yml` file context). Ideally, we'd capture the un-rendered versions in `unrendered_config`, as we do for config set in `dbt_project.yml`.
- Merge `node.config.meta` --> `node.meta`
- Support a `config:` dict nested under tests. Support grabbing test configs from `config:` dict or top-level keys. Raise an exception if the same test config is set in both the `config:` dict and as a top-level key.

Those are all things we want for feature completeness in v0.21. They don't have to be blockers to merging this specific PR; I'd be happy to open follow-on issue(s), if they'd require nontrivial work and we'd rather get this PR merged now.

gshank:
For the node.meta, do we want to have two copies of the meta, one in node.meta and one in mode.config.meta or just one?

For test config, pretty much everything goes in TestConfig now. So I assume it will stay there, but we should do some validation/munging of things set under the 'config' key and things not set under it?

Also TestConfig inherits from NodeConfig now, but the only things that it looks like are actually supported from it are 'tags' and 'enabled'. This feels like a good time to change TestConfig to *not* inherit from NodeConfig. The attributes that are in NodeConfig and TestConfig but don't look supported in TestConfig: persist_docs, post_hook, pre_hook, quoting, column_types, alias, schema, database, full_refresh, on schema_change. Should they all be removed from TestConfig?

gshank:
Oh, the rendering was actually already being done by the render_data codepath in the schema parser. The test configs were an explicit exception for some reason.

gshank:
I suspect that config for schema test nodes and data test nodes is (or ought to be) rather different. Can those configs that aren't settable for schema tests be set for data tests?

jtcohen6:
> For the node.meta, do we want to have two copies of the meta, one in node.meta and one in node.config.meta or just one?

For backwards compatibility, I'm thinking we should have it in both places, with identical contents. (Folks are used to accessing `node.meta` from the [`graph`](https://docs.getdbt.com/reference/dbt-jinja-functions/graph) Jinja context variable, and from `manifest.json`.)

And raise an exception if the same meta property is set like so:
```yml
models:
  - name: model_a
    meta:
      lala: haha
    config:
      meta:
        lala: haha
```

> For test config, pretty much everything goes in TestConfig now. So I assume it will stay there, but we should do some validation/munging of things set under the 'config' key and things not set under it?

That sounds right. Just to triple-check, this is what I have in mind:
```yml
tests:
  - unique:
      config:
        severity: warn
```
I think we should do just enough validation to raise an error if the same config is set under `config:` (new) and as a top-level key (which we need to support for backwards compatibility):
```yml
tests:
  - unique:
      severity: warn
      config:
        severity: warn
```

> Also TestConfig inherits from NodeConfig now, but the only things that it looks like are actually supported from it are 'tags' and 'enabled'. This feels like a good time to change TestConfig to not inherit from NodeConfig. The attributes that are in NodeConfig and TestConfig but don't look supported in TestConfig: persist_docs, post_hook, pre_hook, quoting, column_types, alias, schema, database, full_refresh, on schema_change. Should they all be removed from TestConfig?

I'd be fine with making `TestConfig` no longer inherit from `NodeConfig`, though what I'd _really_ like to do is make a separate `ModelConfig` that inherits from `NodeConfig` and includes model-only configs (e.g. `on_schema_change`) there, so that they aren't inherited by other resource types where they don't make sense (#3557).

Tests do use the `alias`, `schema`, `database` configs, now that it's possible to store test failures as tables in the database.

> Oh, the rendering was actually already being done by the render_data codepath in the schema parser. The test configs were an explicit exception for some reason.

Ah, ok. I think we fixed that for test arguments in #3564, right? My sense is that test modifiers have been using a subtly different rendering context for a long time...

> I suspect that config for schema test nodes and data test nodes is (or ought to be) rather different. Can those configs that aren't settable for schema tests be set for data tests?

They should be the same! We're trying to unify tests as much as possible, and a big part of that is having both test types support all the same configs.

core/dbt/parser/schema_test_builders.py
```diff
@@ -224,13 +224,20 @@ def __init__(
         groups = match.groupdict()
         self.name: str = groups['test_name']
         self.namespace: str = groups['test_namespace']
-        self.modifiers: Dict[str, Any] = {}
-        for key in self.MODIFIER_ARGS:
+        self.config: Dict[str, Any] = {}
```
jtcohen6, line 227: Need to add support for `config:` dict for `unique` + `not_null` as well. This currently raises an error:
```yml
- unique:
    config:
      severity: warn
```
```
macro 'dbt_macro__test_unique' takes no keyword argument 'config'
```

core/dbt/parser/schema_test_builders.py
```diff
@@ -241,7 +248,7 @@ def __init__(
 
         # use hashed name as alias if too long
         if compiled_name != fqn_name:
-            self.modifiers['alias'] = compiled_name
+            self.config['alias'] = compiled_name
```
jtcohen6, line 251: This is an inspired choice from #3316. Let's check to see if there's a user-supplied value of `config['alias']` before overriding it here.

core/dbt/parser/schema_test_builders.py
```diff
@@ -224,13 +224,20 @@ def __init__(
...
+        for key in self.CONFIG_ARGS:
             value = self.args.pop(key, None)
+            # 'modifier' config could be either top level arg or in config
+            if not value and 'config' in self.args:
+                value = self.args['config'].pop(key, None)
             if isinstance(value, str):
                 value = get_rendered(value, render_ctx, native=True)
             if value is not None:
-                self.modifiers[key] = value
+                self.config[key] = value
```
jtcohen6, line 237: Could we raise a validation error if the same key exists in `config:` and in top-level arguments?
```yml
- accepted_values:
    values: [1,2]
    where: "2=2"
    config:
      where: "1=1"
```

jtcohen6 (commented):
This is really really close! Thanks for rolling with the latest round of feedback, it's just some minor cleanup from here.

core/dbt/parser/schema_test_builders.py
gshank: I can't recreate this. You're sure you were using the branch?

jtcohen6: I must've been doing something funny! Just tried this again with the latest version of this branch, and it works just fine

core/dbt/contracts/graph/compiled.py
```diff
@@ -109,15 +109,19 @@ class CompiledSnapshotNode(CompiledNode):
 @dataclass
 class CompiledDataTestNode(CompiledNode):
     resource_type: NodeType = field(metadata={'restrict': [NodeType.Test]})
-    config: TestConfig = field(default_factory=TestConfig)
+    # Was not able to make mypy happy and keep the code working. We need to
+    # refactor the various configs.
+    config: TestConfig = field(default_factory=TestConfig)  # type:ignore
```
nathaniel-may, line 114: It looks like your suggestion to remove TestConfig as a subclass of NodeConfig would fix this, but it looks like we'd have to stop pulling that information from `field` and use a function that mypy can follow the types of.

Can we update the comment to reference the problem / possible fix here just so the information doesn't get lost?

core/dbt/contracts/graph/model_config.py
```diff
@@ -355,10 +356,38 @@ class SourceConfig(BaseConfig):
 
 
 @dataclass
-class NodeConfig(BaseConfig):
+class NodeAndTestConfig(BaseConfig):
```
nathaniel-may, line 359: This is a superclass of `NodeConfig` and `TestConfig`. I don't see it used explicitly beyond the inheritance, so is this just for defining common fields in one place? If so, we might want to make this an `abstract base class` so it's clear there will never be any instances of this type floating around.

core/dbt/contracts/graph/parsed.py
```diff
@@ -229,6 +228,11 @@ class ParsedNode(ParsedNodeDefaults, ParsedNodeMixins, SerializableType):
     def _serialize(self):
         return self.to_dict()
 
+    def __post_serialize__(self, dct):
+        if 'config_call_dict' in dct:
+            del dct['config_call_dict']
+        return dct
```
nathaniel-may, line 234: I don't really understand what's happening here.

core/dbt/contracts/graph/parsed.py
```diff
@@ -258,10 +262,16 @@ def _deserialize(cls, dct: Dict[str, int]):
             return cls.from_dict(dct)
 
     def _persist_column_docs(self) -> bool:
-        return bool(self.config.persist_docs.get('columns'))
+        if hasattr(self.config, 'persist_docs'):
+            assert isinstance(self.config, NodeConfig)
```
nathaniel-may, line 266: I get that this is a mypy trick, but if this asserts false, it'll raise to our users right? Might want to put a user-friendly exception before the assert.

core/dbt/contracts/graph/parsed.py
```diff
@@ -258,10 +262,16 @@ def _deserialize(cls, dct: Dict[str, int]):
...
+            return bool(self.config.persist_docs.get('columns'))
+        return False
 
     def _persist_relation_docs(self) -> bool:
-        return bool(self.config.persist_docs.get('relation'))
+        if hasattr(self.config, 'persist_docs'):
+            assert isinstance(self.config, NodeConfig)
```
nathaniel-may, line 272: same with this assert

core/dbt/contracts/graph/parsed.py
```diff
@@ -411,15 +421,19 @@ class HasTestMetadata(dbtClassMixin):
 @dataclass
 class ParsedDataTestNode(ParsedNode):
     resource_type: NodeType = field(metadata={'restrict': [NodeType.Test]})
-    config: TestConfig = field(default_factory=TestConfig)
+    # Was not able to make mypy happy and keep the code working. We need to
+    # refactor the various configs.
+    config: TestConfig = field(default_factory=TestConfig)  # type: ignore
```
nathaniel-may, line 426: Updating comment here too.

core/dbt/parser/base.py
```diff
@@ -294,38 +293,59 @@ def update_parsed_node_name(
...
+        # config calls from SQL files
+        config_dict = config.build_config_dict(patch_config_dict=patch_config_dict)
+
+        # Set tags on node provided in config blocks. Tags are additive, so even if
+        # config has been built before, we don't have to reset tags in the parsed_node.
         model_tags = config_dict.get('tags', [])
-        parsed_node.tags.extend(model_tags)
+        for tag in model_tags:
+            if tag not in parsed_node.tags:
+                parsed_node.tags.append(tag)
```
nathaniel-may, line 315: I think we should be referencing `MergeBehavior` here and using a centralized function to do this merging. I used `ContextConfig::_add_config_call` in my experimental parser changes.

core/dbt/parser/schema_test_builders.py
```diff
@@ -282,46 +293,88 @@ def extract_test_args(test, name=None) -> Tuple[str, Dict[str, Any]]:
...
+
+    @property
+    def database(self) -> Optional[str]:
+        return self.config.get('database')
+
+    @property
+    def schema(self) -> Optional[str]:
+        return self.config.get('schema')
+
+    def get_static_config(self):
```
nathaniel-may, line 346: The way this method body is defined makes me think there's a better way to do this. My guess is that should entail a bigger refactor though. What are your thoughts on this, @gshank?

core/dbt/parser/schemas.py
```diff
@@ -314,7 +316,7 @@ def _parse_generic_test(
         # is not necessarily this package's name
         fqn = self.get_fqn(fqn_path, builder.fqn_name)
 
-        # this is the config that is used in render_update
+        # this is the ContextConfig that is used in render_update
```
nathaniel-may, line 319: Is it possible to use mypy to make this assertion here?

core/dbt/parser/schemas.py
```diff
@@ -786,6 +798,9 @@ class NodePatchParser(
     def parse_patch(
         self, block: TargetBlock[NodeTarget], refs: ParserRef
     ) -> None:
+        # We're not passing the ParsedNodePatch around anymore, so we
+        # could possibly skip creating one. Leaving here for now for
+        # code consistency.
```
nathaniel-may, line 803: imo, less code is better in this case.

nathaniel-may (commented):
I have a little trouble following this area of the code base so I think it would be helpful to have a "walkthrough" or "suggested reviewer path" in the PR description to help follow the flow of big multi-file changes like this.

I don't think it's a problem for this PR but in the future I think removals of deprecated items should have their own PR and commit so they're easier to track and document.

Leaving my review as `Comment` because I don't fully understand the change enough to approve, but also didn't see anything concerning enough to block on.

gshank:
Notes for reviewers:

The changes are all inserted into a pre-existing flow, they don't call each other. There are a number of chunks. All of these "nodes" (using the term loosely) already had Config objects of some kind.

Prior to this ticket "Config" attributes (except for tests)  were set in the dbt_project.yml file and in SQL files which ended up in the various node Config objects when 'update_parsed_node_(config)' was called by render_update.

The schema file stuff is mostly set in the "patch" methods, because those "nodes" already exist and schema "patches" just update existing nodes (except for sources, which we're not touching here). So the newest piece is where we call 'update_parsed_node_config' when doing the node patching

Originally a piece of patching (add_patch, I think) was done in Manifest methods, but they now require schema parsing methods, so I had to move that code into the schema parser (which is where patching is initiated).

When the patching code is called, if there's 'config', it calls 'update_parsed_node_config'. Which re-does what happened the first time (when the node was created) except it also passes in the config dictionary from the patch

So we have to run the update_parsed_node_config twice when we have config in schema files, but there really wasn't any easy way around that.

There's also some noise in there, because I got rid of the deprecated 'vars' attribute. And also added 'meta' to all the config classes.

core/dbt/contracts/graph/model_config.py
gshank: I'm pretty sure that it can't be an abstract base class because it's an intermediate class in the hierarchy, not an actual base class.

core/dbt/contracts/graph/parsed.py
gshank: We don't want the 'config_call_dict' in the serialized output. We needed to save it for when we reprocess the config dictionaries, but it shouldn't be saved in artifacts or the partial parse manifest.

core/dbt/contracts/graph/parsed.py
gshank: The 'assert' is for situations where we will never encounter an actual error, it is only to make mypy happy. So there is no need for a user friendly exception.

core/dbt/contracts/graph/parsed.py
gshank: See above.

core/dbt/parser/base.py
gshank: This is only because tags are copied to top level node. It's a special exception and does not need to use the MergeBehavior.

core/dbt/parser/schema_test_builders.py
gshank: I agree that this is not a good way to do this and it should be refactored. We will probably be doing more config work on tests in the future and that would be a good time to do it.

core/dbt/parser/schemas.py
gshank: Sure. Have changed the following line to say config: ContextConfig

core/dbt/parser/schemas.py
```diff
@@ -674,7 +649,14 @@ def parse(self) -> List[TestBlock]:
             if is_override:
                 data['path'] = self.yaml.path.original_file_path
                 patch = self._target_from_dict(SourcePatch, data)
-                self.manifest.add_source_patch(self.yaml.file, patch)
+                assert isinstance(self.yaml.file, SchemaSourceFile)
```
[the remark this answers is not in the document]
gshank: These asserts are only for mypy. They are only used in cases where there is no chance of them occurring, so user friendly error messages are not necessary.

core/dbt/parser/schemas.py
gshank: I always like removing code :) But it's more changes than I want to do at this point. We can fix next time this is touched.

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@
 ### Features
 - Add `dbt build` command to run models, tests, seeds, and snapshots in DAG order. ([#2743](https://github.com/dbt-labs/dbt/issues/2743), [#3490](https://github.com/dbt-labs/dbt/issues/3490), [#3608](https://github.com/dbt-labs/dbt/issues/3608))
 - Introduce `on_schema_change` config to detect and handle schema changes on incremental models ([#1132](https://github.com/fishtown-analytics/dbt/issues/1132), [#3387](https://github.com/fishtown-analytics/dbt/issues/3387))
+- Enable setting configs in schema files for models, seeds, snapshots, analyses, tests ([#2401](https://github.com/dbt-labs/dbt/issues/2401), [#3616](https://github.com/dbt-labs/dbt/pull/3616))
```
jtcohen6, line 13: Could you move this up to be under `dbt 0.21.0 (Release TBD)`?

jtcohen6 (approved):
Giving this a thumbs-up from an end-user functional perspective. There are likely to be a handful of small things that we want to touch up as people start using this. I'm excited to get it into beta.

🚀 🚀

CHANGELOG.md
gshank: Yes.

core/dbt/contracts/graph/compiled.py
nathaniel-may: Ahh gotcha. I played around with it myself a bit and got it wrong. Definitely fine to push off.

core/dbt/contracts/graph/model_config.py
nathaniel-may: ugh that's good point. nevermind then.

core/dbt/contracts/graph/parsed.py
nathaniel-may: ahh that makes sense. Thanks for the explanation. I think a comment to that effect would help.

core/dbt/contracts/graph/parsed.py
nathaniel-may: I just don't see any mechanical proof that we could never reach this state. I might be being overly cautious, but code that throws makes me nervous.

core/dbt/parser/base.py
nathaniel-may: 🤯 ok that's a little unexpected. So by the end of this, the top-level node will have all the tags from all the sub nodes in the tree?

core/dbt/parser/schemas.py
nathaniel-may: I'm also in favor of a teeny tiny PR to make this change 🙂

nathaniel-may (approved):

kwigley (approved):

avaitla:
Does this mean you could add a unique test for a column in the models .sql file rather than having to put it into the .yaml file? Do you have a quick example / doc to show how that might look?

brad-meru:
These aren't being captured in dbt Docs. The only way that seems to work is to config in the model.

It would also be great if column tags made it into Docs.

PR #3617, opened 2021-07-23
Secret scrubbing for env variables

### LINKED ISSUE #3440: Scrub environment variables in target + logs
### Describe the feature

Could we scrub environment variables (values templated by `env_var`) from `target/` and `logs/`?

We could consider doing this for:
- All environment variables
- Environment variables prefixed with `SECRET_`

I don't have a great sense of what would be involved here. The best approach would tie into each actual `env_var` call. A less-good approach would grep strings for env vars before writing to files.

The primary motivation for this request is dbt Cloud. dbt Core feels like the best place to handle it, however, and many Core users have asked for it ([e.g.](https://github.com/fishtown-analytics/dbt-spark/issues/147#issuecomment-843627607)).

### Additional context

- Not specific to any database
- Scrubbing environment variables from JSON artifacts also scrubs them in dbt-docs

### Who will this benefit?

- More complex, security-conscious deployments

leahwicz:
resolves #3440 

### Description

This change allows users to prefix their environment variables with `DBT_ENV_SECRET_` which we will then treat as secrets and scrub them from the logs and console output.

### Checklist


jtcohen6 (approved):
lgtm! nice work adding the test.

anyone have a reason to prefer something other than `*****`?

leahwicz:
> anyone have a reason to prefer something other than `*****`?

I know this is industry standard for CI/CD systems so it would be easy to keep aligned with that

PR #3623, opened 2021-07-24
fewer adapters will need to re-implemnt basic_load_csv_rows

### LINKED ISSUE #3622: Allow custom adapters to replace default param marker in seed's basic_load_csv_rows
### Describe the feature
Rather than overriding the entire `basic_load_csv_rows macro`, allow dbt adapters to override just the SQL parameter marker.

- `:p` (see [dbt-oracle's `seed.sql`](https://github.com/techindicium/dbt-oracle/blob/fa9718809840ee73e6072f483233f5150cc9986c/dbt/include/oracle/macros/materializations/seed/seed.sql#L20-22))
- `?` (see [dbt-sqlserver's `seed.sql`](https://github.com/dbt-msft/dbt-sqlserver/blob/75ff846ae0fbb93c41e7f93c087a24a1e19bb7e0/dbt/include/sqlserver/macros/materializations/seed/seed.sql#L34), [dbt-ibmdb2's `seed.sql`](https://github.com/kaysef/dbt-ibm-db2/blob/b01548c2057d6df76cb285d0f4168a6c8fdef064/dbt/include/ibm_dbt/macros/materializations/seed/seed.sql#L18) and other as of yet unreleased database adapters...)

overriding the whole macro is what's currently done

### Additional context

relevant for the [pyodbc](https://github.com/mkleehammer/pyodbc), [cx_Oracle](https://oracle.github.io/python-cx_Oracle/),  and [JayDeBeApi](https://github.com/baztian/jaydebeapi) drivers 

### Who will this benefit?

This benefits adapter maintainers not having to keep the `basic_load_csv_rows` macro up to date and keep their respective adapter repo code footprint smaller, thus more easily maintained and compatible with future changes

PR coming right up!

swanderz:
resolves #3622

### Description

Implements a `get_binding_char` macro, which allows for easy substitution. This will allow dbt-sqlserver, dbt-oracle, and other adapters to only implement `get_binding_char` and not overload the entire `basic_load_csv_rows`

### Questions

1. are we at the point yet of obviating `basic_load_csv_rows` in favor of `default__load_csv_rows` where `get_batch_size()` is another dispatched macro?
2. related to above are dispatched macros the best way to store the param string and `default_batch_size`? perhaps it's cleaner to add this as a `SQLAdapterClass` property?
3. I thought it a bit strange to that 3 different macros are defines, followed by their `default__` implementations. I followed this pattern, but it feels weird to have `get_binding_char` so far from `default__get_binding_char`. any strong opinions? 

### Checklist


setup.py
```diff
@@ -24,7 +24,7 @@
 
 
 package_name = "dbt"
-package_version = "0.21.0a1"
+package_version = "0.20.0"
```
jtcohen6, line 27: Could you switch this back? I'm guessing you changed to test with current plugin versions

CHANGELOG.md
```diff
@@ -36,6 +36,12 @@ Contributors:
 - Fix `store_failures` config when defined as a modifier for `unique` and `not_null` tests ([#3575](https://github.com/fishtown-analytics/dbt/issues/3575), [#3577](https://github.com/fishtown-analytics/dbt/pull/3577))
 - Fix `where` config with `relationships` test by refactoring test SQL. Note: The default `relationships` test now includes CTEs, and may need reimplementing on adapters that don't support CTEs nested inside subqueries. ([#3579](https://github.com/fishtown-analytics/dbt/issues/3579), [#3583](https://github.com/fishtown-analytics/dbt/pull/3583))
 
+### Under the hood
+- Allow the default seed macro's SQL parameter, `%s`, to be replaced by dispatching a new macro, `get_binding_char()`. This enables adapters with parameter marker characters such as `?` to not have to override `basic_load_csv_rows`.  ([#3622](https://github.com/fishtown-analytics/dbt/issues/3622), [#3623](https://github.com/fishtown-analytics/dbt/pull/3623))
+
+
+Contributors:
+- [@swanderz](https://github.com/swanderz) [#3623](https://github.com/fishtown-analytics/dbt/pull/3623)
```
jtcohen6, line 44: Could you move this up to v0.21?

jtcohen6 (commented):
I'm all about it—let's make these macros more modular, and make the lives of adapter plugin maintainers easier.

Answers to your questions:
1. Agreed, it's a bit silly that the only difference between `default__load_csv_rows` and `basic_load_csv_rows` is setting the batch size to 10000. I'd be open to new dispatched macros for `get_csv_batch_size()` and `get_csv_insert_into_sql()`.
2. I gave this a bit of thought. Our general rule is: All explicit SQL should be wrapped in Jinja macros. All implicit or non-SQL behavior of databases should be set in python. There are adapter plugins that leverage multiple connection options—e.g. `dbt-spark` uses `pyodbc` for some, not others—which is an argument in favor of adding a `SQLConnectionManager` property/method. This one is tricky, but I think a SQL param _is_ SQL, sort of, so I'm happy setting it in a Jinja macro.
3. No strong preference here. In other places, I think we tend to define a dispatched macro, immediately followed by its default implementation, and I agree that's a bit easier to read.

swanderz:
took another stab at refactoring. The diff is ugly (the pedant in my is toying with two separate PRs) so make this better.

I decided to keep the params for `load_csv_rows()` at two, so adapter maintainers don't have breaking changes. However, I'm not in love with the hacky workaround of introducing`get_batch_size()` with a `var()` call. On the same note, I'm not in love with the `batch_size` parameter name, as it isn't immediately clear from the name as to whether it means the maximum number of rows to include in each query or the max number of SQL params to include in each query. The truth is the former, so recently we merged https://github.com/dbt-msft/dbt-sqlserver/pull/151 which will auto-calculate batch size.

swanderz:
@jtcohen6, I responded to your feedback, but I'm not sure why some of these tests are failing

core/dbt/include/global_project/macros/materializations/seed/seed.sql
```diff
@@ -47,6 +42,21 @@
...
+  {{ return('%s') }}
+{% endmacro %}
+
+{% macro get_batch_size() -%}
+  {{ adapter.dispatch('get_batch_size')() }}
+{%- endmacro %}
+
+{% macro default__get_batch_size() %}
+  {{ return(var("batch_size", 10000)) }}
+{% endmacro %}
```
jtcohen6, line 59: Definitely agree with turning this from an embedded hard-coded value into a standalone macro!

I don't think I see the need for the `var` here. As it is, users can override/reimplement `get_batch_size()` in their root projects, and dbt will prefer their version. So I'd suggestion just:
```suggestion
{% macro default__get_batch_size() %}
  {{ return(10000) }}
{% endmacro %}
```

jtcohen6: I'm trying to decide if this needs to be a _dispatched_ macro. How common would it be that an adapter needs to change the default from 10k because of some implementation detail of their database / python client / pyodbc / etc? (I _think_ I follow the reasoning in https://github.com/dbt-msft/dbt-sqlserver/pull/151, which would seem to be a case in point.)

There's very little downside to dispatching, only more optionality at the cost of (very minor) complexity, so I don't really have any opposition.

jtcohen6 (commented):
@swanderz Looking good! So sorry for the delay getting back to you on this one.

When you get a chance, could you pull the latest changes from `develop`? Both in terms of the changelog placement, and also because we just overhauled our CI testing suite. I'd be curious to see if the same tests are failing as the ones you saw a few weeks ago.

jtcohen6:
Going to close and reopen just to trigger additional adapter tests

jtcohen6 (approved):
Thanks for this @swanderz!

PR #3625, opened 2021-07-25
Initial

### LINKED ISSUE #3462: Add command to auto-populate `profiles.yml`
### Describe the feature
This may have been suggested already but I couldn't find a duplicate issue from a quick search.

Configuring the `profiles.yml` is non-beginner friendly, and creating instructions on how to do so is complicated especially due to the differences between Mac and Windows environments. Introducing a dbt command which would add a profile to `profiles.yml` (and create the file if needed) using the name from the `dbt_project.yml` would be extremely useful. In additional, it is typical that only the schema, username and password are unique between individual `profiles.yml`s, so a default set of values specified in e.g. `dbt_project.yml` could reduce the number of options required to be entered.

Creating my own script to do this.

### Who will this benefit?
Anyone who introduces dbt to new users frequently and struggles to write a simple and comprehensive set of instructions for configuring a `profiles.yml`.

With some guidance, sure.

NiallRees:
Resolves #3462

### Description

When `dbt init` is run in a pre-existing project directory, use an interactive CLI to configure a connection profile.

### Checklist

### DIFF
--- a/core/dbt/main.py
+++ b/core/dbt/main.py
@@ -256,7 +256,6 @@ def run_from_args(parsed):
 
     with track_run(task):
         results = task.run()
-
     return task, results
 
 
@@ -361,7 +360,7 @@ def _build_init_subparser(subparsers, base_subparser):
         '''
     )
     sub.add_argument(
-        'project_name',
+        '--project_name',
         type=str,
         help='''
         Name of the new project
--- a/core/dbt/task/init.py
+++ b/core/dbt/task/init.py
@@ -1,5 +1,9 @@
+import copy
 import os
 import shutil
+import yaml
+
+import click
 
 import dbt.config
 import dbt.clients.system
@@ -10,7 +14,7 @@ from dbt.logger import GLOBAL_LOGGER as logger
 
 from dbt.include.starter_project import PACKAGE_PATH as starter_project_directory
 
-from dbt.task.base import BaseTask
+from dbt.task.base import BaseTask, move_to_nearest_project_dir
 
 DOCS_URL = 'https://docs.getdbt.com/docs/configure-your-profile'
 SLACK_URL = 'https://community.getdbt.com/'
@@ -54,19 +58,18 @@ class InitTask(BaseTask):
             return True
         return False
 
-    def create_profiles_file(self, profiles_file, sample_adapter):
+    def create_sample_profiles_file(self, profiles_file, adapter):
         # Line below raises an exception if the specified adapter is not found
-        load_plugin(sample_adapter)
-        adapter_path = get_include_paths(sample_adapter)[0]
+        load_plugin(adapter)
+        adapter_path = get_include_paths(adapter)[0]
         sample_profiles_path = adapter_path / 'sample_profiles.yml'
 
         if not sample_profiles_path.exists():
-            logger.debug(f"No sample profile found for {sample_adapter}, skipping")
+            logger.debug(f"No sample profile found for {adapter}, skipping")
             return False
 
         if not os.path.exists(profiles_file):
-            msg = "With sample profiles.yml for {}"
-            logger.info(msg.format(sample_adapter))
+            logger.info(f"With sample profiles.yml for {adapter}")
             shutil.copyfile(sample_profiles_path, profiles_file)
             return True
 
@@ -83,29 +86,117 @@ class InitTask(BaseTask):
             slack_url=SLACK_URL
         )
 
+    def generate_target_from_input(self, target_options, target={}):
+        target_options_local = copy.deepcopy(target_options)
+        # value = click.prompt('Please enter a valid integer', type=int)
+        for key, value in target_options_local.items():
+            if not key.startswith("_"):
+                if isinstance(value, str) and (value[0] + value[-1] == "[]"):
+                    hide_input = key == "password"
+                    target[key] = click.prompt(
+                        f"{key} ({value[1:-1]})", hide_input=hide_input
+                    )
+                else:
+                    target[key] = target_options_local[key]
+            if key.startswith("_choose"):
+                choice_type = key[8:]
+                option_list = list(value.keys())
+                options_msg = "\n".join([
+                    f"[{n+1}] {v}" for n, v in enumerate(option_list)
+                ])
+                click.echo(options_msg)
+                numeric_choice = click.prompt(
+                    f"desired {choice_type} option (enter a number)", type=int
+                )
+                choice = option_list[numeric_choice - 1]
+                target = self.generate_target_from_input(
+                    target_options_local[key][choice], target
+                )
+        return target
+
+    def get_profile_name_from_current_project(self):
+        with open("dbt_project.yml") as f:
+            dbt_project = yaml.load(f)
+        return dbt_project["profile"]
+
+    def write_profile(self, profiles_file, profile, profile_name=None):
+        if not profile_name:
+            profile_name = self.get_profile_name_from_current_project()
+        if os.path.exists(profiles_file):
+            with open(profiles_file, "r+") as f:
+                profiles = yaml.load(f) or {}
+                profiles[profile_name] = profile
+                f.seek(0)
+                yaml.dump(profiles, f)
+        else:
+            profiles = {profile_name: profile}
+            with open(profiles_file, "w") as f:
+                yaml.dump(profiles, f)
+
+    def configure_profile_from_scratch(self, selected_adapter):
+        # Line below raises an exception if the specified adapter is not found
+        load_plugin(selected_adapter)
+        adapter_path = get_include_paths(selected_adapter)[0]
+        target_options_path = adapter_path / 'target_options.yml'
+        profiles_file = os.path.join(dbt.config.PROFILES_DIR, 'profiles.yml')
+
+        if not target_options_path.exists():
+            logger.info(f"No options found for {selected_adapter}, using " +
+                        "sample profiles instead. Make sure to update it at" +
+                        "{profiles_file}.")
+            self.create_sample_profiles_file(profiles_file, selected_adapter)
+        else:
+            logger.info(f"Using {selected_adapter} profile options.")
+            with open(target_options_path) as f:
+                target_options = yaml.load(f)
+            target = self.generate_target_from_input(target_options)
+            profile = {
+                "outputs": {
+                    "dev": target
+                },
+                "target": "dev"
+            }
+            self.write_profile(profiles_file, profile)
+
+    def configure_profile_using_defaults(self, selected_adapter):
+        raise(NotImplementedError())
+
     def run(self):
-        project_dir = self.args.project_name
-        sample_adapter = self.args.adapter
-        if not sample_adapter:
-            try:
-                # pick first one available, often postgres
-                sample_adapter = next(_get_adapter_plugin_names())
-            except StopIteration:
-                logger.debug("No adapters installed, skipping")
-
+        selected_adapter = self.args.adapter
         profiles_dir = dbt.config.PROFILES_DIR
-        profiles_file = os.path.join(profiles_dir, 'profiles.yml')
-
         self.create_profiles_dir(profiles_dir)
-        if sample_adapter:
-            self.create_profiles_file(profiles_file, sample_adapter)
 
-        if os.path.exists(project_dir):
-            raise RuntimeError("directory {} already exists!".format(
-                project_dir
-            ))
+        # Determine whether we're initializing a new project or configuring a
+        # profile for an existing one
+        if self.args.project_name:
+            project_dir = self.args.project_name
+            if os.path.exists(project_dir):
+                raise RuntimeError("directory {} already exists!".format(
+                    project_dir
+                ))
 
-        self.copy_starter_repo(project_dir)
+            self.copy_starter_repo(project_dir)
 
-        addendum = self.get_addendum(project_dir, profiles_dir)
-        logger.info(addendum)
+            addendum = self.get_addendum(project_dir, profiles_dir)
+            logger.info(addendum)
+            if not selected_adapter:
+                try:
+                    # pick first one available, often postgres
+                    selected_adapter = next(_get_adapter_plugin_names())
+                except StopIteration:
+                    logger.debug("No adapters installed, skipping")
+            self.configure_profile_from_scratch(
+                selected_adapter
+            )
+        else:
+            logger.info("Setting up your profile.")
+            move_to_nearest_project_dir(self.args)
+            if os.path.exists("target_defaults.yml"):
+                self.configure_profile_using_defaults()
+            else:
+                if not selected_adapter:
+                    raise RuntimeError("No adapter specified.")
+                logger.info("Configuring from scratch.")
+                self.configure_profile_from_scratch(
+                    selected_adapter
+                )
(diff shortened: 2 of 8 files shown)


NiallRees:
This is very much draft @jtcohen6 but would nevertheless appreciate your review on A. the behaviour and B. the implementation approach. I'm struggling to work out how to easily run the `debug` task from within the `init` task, if you have any ideas on that too! The task interfaces seem to be designed only to be run from the `main.py` context.

core/dbt/main.py
```diff
@@ -361,7 +360,7 @@ def _build_init_subparser(subparsers, base_subparser):
         '''
     )
     sub.add_argument(
-        'project_name',
+        '--project_name',
```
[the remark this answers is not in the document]
jtcohen6: Better yet, could we do away with the `--project-name` and `--adapter` arguments entirely, and simply `click.prompt()` for these things too??

plugins/snowflake/dbt/include/snowflake/target_options.yml
```diff
@@ -0,0 +1,15 @@
+type: snowflake
+account: '[account id + region (if applicable)]'
+user: '[username]'
+_choose_authentication_type:
+  password:
+    password: '[password]'
+  keypair:
+    private_key_path: '[path/to/private.key]'
+    private_key_passphrase: '[passphrase for the private key, if key is encrypted]'
```
jtcohen6, line 9: Really cool!

jtcohen6 (commented):
@NiallRees This is slick!!

I left some comments below where I was working through each step in my mind. Then I took a step back, thought through the ideal flow, and threw together a quick flowchart ([whimsical link](https://whimsical.com/RtVTJWrkXUyYWjxdtumFZ4@2Ux7TurymN2M5kLojspD)):

Screen Shot 2021-08-10 at 10 31 49 AM

What do you think? In particular, I'd love to:
- Prefer `click.prompt()` over CLI flags/args wherever possible
- Include backwards-compatible behavior for the old way (`sample_profiles.yml`), while adapter plugin maintainers upgrade to the new way (`target_options.yml`)
- Ensure that `dbt init` _never_ fails or raises an error. It's fine if it needs to skip steps due to missing information

NiallRees:
> @NiallRees This is slick!!
> 
> I left some comments below where I was working through each step in my mind. Then I took a step back, thought through the ideal flow, and threw together a quick flowchart ([whimsical link](https://whimsical.com/RtVTJWrkXUyYWjxdtumFZ4@2Ux7TurymN2M5kLojspD)):
> 
> Screen Shot 2021-08-10 at 10 31 49 AM
> 
> What do you think? In particular, I'd love to:
> 
> * Prefer `click.prompt()` over CLI flags/args wherever possible
> * Include backwards-compatible behavior for the old way (`sample_profiles.yml`), while adapter plugin maintainers upgrade to the new way (`target_options.yml`)
> * Ensure that `dbt init` _never_ fails or raises an error. It's fine if it needs to skip steps due to missing information

I think yes to all

I believe this addresses all three of your points. `dbt init` now takes no CLI arguments/flags and prompts as needed. If a `target_options.yml` is unavailable, `sample_profiles.yml` will be used. One key change on the latter is that it is no longer copied wholesale, but can be added to an existing `profiles.yml`. Curious for your thoughts on that.

core/dbt/task/init.py
```diff
@@ -83,29 +86,117 @@ def get_addendum(self, project_name, profiles_path):
...
+            profile = {
+                "outputs": {
+                    "dev": target
+                },
+                "target": "dev"
+            }
+            self.write_profile(profiles_file, profile)
+
+    def configure_profile_using_defaults(self, selected_adapter):
+        raise(NotImplementedError())
```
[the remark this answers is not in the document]
NiallRees: I actually switched over to the idea of using a `profile_template.yml` file which provides greater flexibility to orgs wishing to use several targets. An example `profile_template.yml`:

```
vars:
  - snowflake_username
  - snowflake_password
profile:
  my_profile:
    outputs:
      dev:
        type: snowflake
        account: asdf
        user: "{{ vars.snowflake_username }}"
        password: "{{ vars.snowflake_password }}"
        role: asdf
        database: asdf
        warehouse: asdf
        schema: asdf
        threads: 4
        client_session_keep_alive: false
    target: dev
```

What do you think?

core/dbt/clients/yaml_helper.py
```diff
@@ -1,17 +1,16 @@
 import dbt.exceptions
 from typing import Any, Dict, Optional
-import yaml
-import yaml.scanner
+import oyaml as yaml
```
NiallRees, line 3: I replaced yaml with oyaml in order to retain ordering when prompting for user input. Rather than keep both I just replaced every reference with oyaml. It may well be preferred that we leave all other imports as-is, let me know.

NiallRees:
Ready for your 👀 again @jtcohen6. Once I've got a thumbs up on what this is doing I'll get some integration tests written up. Still curious for your thoughts on running `dbt debug` automatically following the profile setup, and if your ideas on the best way to do that.

core/dbt/task/init.py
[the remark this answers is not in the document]
jtcohen6: Very cool! I like it. Maybe just call these `prompts` instead of `vars`? The latter feels like an overloaded term in dbt

jtcohen6:
Great work here @NiallRees! It's so cool to see this coming together, and in a way that's honestly _fun_ to use.

Some misc feedback:

- Let's overwrite the `name` and `profile` in `dbt_project.yml` with the input to `What is the desired project name?` Otherwise, the profile created for new projects is always called `default`. (It will also resolve https://github.com/dbt-labs/dbt/issues/3677 by the by.) Simplest would be to move `dbt_project.yml` out from [the starter project file](https://github.com/dbt-labs/dbt/blob/develop/core/dbt/include/starter_project/dbt_project.yml) and into a templated f-string; I'm sure there's a better way, that still manages to preserve the comments/hints.
- Numeric profile fields (`threads`, `port`) are wrapped in single quotes, so these targets fail on validation. Is there a way to avoid this? Check the type of user inputs?
- I'm unsure about appending the contents of `sample_profiles.yml` to `profiles.yml`, for adapters that do not yet support `target_options.yml`. It messes up the YAML formatting right now (I'm sure that's fixable), but also, many sample profiles make use of comments (e.g. [dbt-presto's](https://github.com/dbt-labs/dbt-presto/blob/master/dbt/include/presto/sample_profiles.yml), [spark's](https://github.com/dbt-labs/dbt-spark/blob/master/dbt/include/spark/sample_profiles.yml)), and those would not be copied, leaving entries like `method: none` with no explanation, or both entries from an either/or choice.
- The desired behavior around `init+debug` is basically just `dbt init && dbt debug`, right? I'm sure you could lightly refactor the debug task's methods, so that the init task could import and call `test_connection()`... but what do you think about an even simpler solution? In our log messages to say `Profile ... written to ...`, we could also include a note like, `Run "dbt debug" to validate your new connection profile`.

@leahwicz I'd be curious to get your take on this, since you and I did some work on `init` recently. Namely:
- We'll need to make an engineering judgment on whether the `yaml` &rarr; `oyaml` replacement across the board is something we'd be comfortable with; `oyaml` [describes itself](https://github.com/wimglenn/oyaml) as a drop-in replacement for PyYAML that preserves dict ordering.
- Let's think about the "programmatic" entry-points to `init`, e.g. for the dbt Cloud IDE (which doesn't use the real `init` task today, but for consistency's sake, I'd feel better if it could). Do we need to preserve a "streamlined," CLI flag-based version of `init` that _just_ scaffolds the basic file structure, using a provided project name, and doesn't worry about `profiles.yml` at all?

core/dbt/task/init.py
```diff
@@ -83,29 +88,196 @@ def get_addendum(self, project_name, profiles_path):
...
+
+    def ask_for_adapter_choice(self) -> str:
+        """Ask the user which adapter (database) they'd like to use."""
+        click.echo("Which database would you like to use?")
+        available_adapters = list(_get_adapter_plugin_names())
+        click.echo("\n".join([
+            f"[{n+1}] {v}" for n, v in enumerate(available_adapters)
+        ]))
+        numeric_choice = click.prompt("Enter a number", type=click.INT)
+        return available_adapters[numeric_choice - 1]
```
jtcohen6, line 244: This could be a neat opportunity for a hint like:
```
Which database adapter would you like to use?

[1] presto
[2] postgres
[3] snowflake
[4] redshift
[5] bigquery

(Don't see the one you want? https://docs.getdbt.com/docs/available-
adapters)

Enter a number:
```

core/dbt/clients/yaml_helper.py
leahwicz: @kwigley do you have any thoughts here?

NiallRees:
> * Let's think about the "programmatic" entry-points to `init`, e.g. for the dbt Cloud IDE (which doesn't use the real `init` task today, but for consistency's sake, I'd feel better if it could). Do we need to preserve a "streamlined," CLI flag-based version of `init` that _just_ scaffolds the basic file structure, using a provided project name, and doesn't worry about `profiles.yml` at all?

@jtcohen6 @leahwicz Just so I know where to head - should this PR need to concern itself with this, given that you've said dbt Cloud doesn't use the init task currently?

test/integration/040_init_test/test_init.py
```diff
@@ -26,18 +29,32 @@ def models(self):
...
+            "localhost",
+            5432,
+            "test_user",
+            "test_password",
+            "test_db",
+            "test_schema",
+        ]
+        self.run_dbt(['init', '--profiles-dir', 'dbt-profile'], profiles_dir=False)
+        manager.assert_has_calls([
+            call.confirm('The profile test already exists in /Users/niall.woodward/.dbt/profiles.yml. Continue and overwrite it?'),
```
NiallRees, line 51: I'm struggling to work out how to get the profile directory to be one that is consistent between testing environments. I think it relates to https://github.com/dbt-labs/dbt/pull/1527 - but would definitely appreciate it if one of the core maintainers could take a look as the test running code looks to be quite in-depth.

core/dbt/task/init.py
```diff
@@ -20,11 +28,7 @@
 IGNORE_FILES = ["__init__.py", "__pycache__"]
 
 ON_COMPLETE_MESSAGE = """
-Your new dbt project "{project_name}" was created! If this is your first time
-using dbt, you'll need to set up your profiles.yml file -- this file will tell dbt how
-to connect to your database. You can find this file by running:
-
-  {open_cmd} {profiles_path}
+Your new dbt project "{project_name}" was created!
```
jtcohen6, line 31: (low priority)

This message feels like _lot_ of text after the interactive bite-sized chunks. I wonder if we can do something cool with `click`, to either:
- [clear](https://click.palletsprojects.com/en/8.0.x/api/#click.clear) the terminal, leaving only the welcome message
- prompting with each item one by one, so that the user "acknowledges" each: project created!, here's a link to the docs, need help?, happy modeling!

core/dbt/clients/yaml_helper.py
jtcohen6: It looks like the switch from `yaml` to `oyaml` has broken one highly specific test, which checks the sorting behavior of the `toyaml` Jinja context method. The `sort_keys` argument is not being respected by `oyaml.safe_dump`. Glad we have a test for it!

https://github.com/dbt-labs/dbt/blob/3789acc5a7b3f71b4e333ac6e235c62ee0c957f5/test/integration/013_context_var_tests/tests/to_yaml.sql#L5

https://github.com/dbt-labs/dbt/blob/3789acc5a7b3f71b4e333ac6e235c62ee0c957f5/core/dbt/context/base.py#L416

I think my preference would probably be to avoid switching from `yaml` to `oyaml` wherever possible. I'm also wondering if there's another way we can preserve prompt order for `target_options.yml`, even if it requires an extra attribute

core/dbt/task/init.py
```diff
@@ -40,40 +44,68 @@
...
-        if not os.path.exists(profiles_file):
-            msg = "With sample profiles.yml for {}"
-            logger.info(msg.format(sample_adapter))
-            shutil.copyfile(sample_profiles_path, profiles_file)
-            return True
-
-        return False
+            logger.debug(f"No sample profile found for {adapter}.")
+        else:
+            with open(sample_profiles_path, "r") as f:
```
jtcohen6, line 85: Here and elsewhere, we probably want to lead on the `dbt.clients.system` module for cross-OS support / error handling. I'm thinking about `load_file_contents` and `write_file` in particular... though we'll need to adjust that method to support the `r+` and `a` modes you're using here.

cc @leahwicz: I'm pretty fuzzy on this stuff, so you should correct me if I'm wrong here. My sense is, the more we can lean on `clients.system` methods for all file operations, the better-served we'll be in a world with storage adapters etc. Alternatively, we could simply say that since `init` is a CLI-only task, we don't care about needing to support it outside of the local file system.

core/dbt/task/init.py
```diff
@@ -84,29 +116,209 @@ def get_addendum(self, project_name, profiles_path):
...
+            self.copy_starter_repo(project_name)
+            os.chdir(project_name)
+            with open("dbt_project.yml", "r+") as f:
+                content = f"{f.read()}".format(
+                    project_name=project_name,
+                    profile_name=project_name
+                )
+                f.seek(0)
+                f.write(content)
+                f.truncate()
```
jtcohen6, line 316: >> Let's think about the "programmatic" entry-points to `init`, e.g. for the dbt Cloud IDE (which doesn't use the real `init` task today, but for consistency's sake, I'd feel better if it could). Do we need to preserve a "streamlined," CLI flag-based version of `init` that _just_ scaffolds the basic file structure, using a provided project name, and doesn't worry about `profiles.yml` at all?

> @jtcohen6 @leahwicz Just so I know where to head - should this PR need to concern itself with this, given that you've said dbt Cloud doesn't use the init task currently?

I think you've done it! This snippet of code is exactly what we'd want in the "programmatic" version. I could imagine wrapping this into a `create_project_files` method, re-adding a flag to init like `--starter-project-only`, and then supporting this workflow as:
```python
    def run(self):
        """Entry point for the init task."""
        if self.args.starter_project_only:
            project_name = self.args.starter_project_only
            create_project_files(project_name)
            return
            
        ... otherwise proceed interactively ...
```
```
dbt init --starter-project-only my_cool_project_name
```
Or, the equivalent interface into the dbt-core API ;)

@leahwicz What do you think? I don't think we need to make that change in this PR, but I'm satisfied knowing it would be a small lift.

jtcohen6 (commented):
@NiallRees Amazing work :)

I only managed to find one bug, which is (I think) a quick fix in the `create_profile_using_profile_template` method.

@leahwicz I think this is ready for review from an engineer on the Core team! The big question for me is around leveraging `dbt.clients.system` methods in lieu of of direct file access.

core/dbt/task/init.py
```diff
@@ -84,29 +116,209 @@ def get_addendum(self, project_name, profiles_path):
...
+        if profile_name in profiles.keys():
+            response = click.confirm(
+                f"The profile {profile_name} already exists in "
+                f"{profiles_file}. Continue and overwrite it?"
+            )
+            return response
+        else:
+            return True
+
+    def create_profile_using_profile_template(self):
```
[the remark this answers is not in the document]
NiallRees: 1. It retrieves the `profile_name` from `profile_template.yml`, I'm not sure it needs to be passed from the run entry point - there is a possibility that `profile_template.yml`'s profile name differs from that of the project.
2. Good idea

PR #3630, opened 2021-07-26
fix awkward exception being raised by a yml file with all comments

### LINKED ISSUE #3568: Full-commented-out yaml file contents return ugly error
### Describe the bug

A user reports seeing this error when trying to parse their project after upgrading to v0.20.0:
```
Encountered an error:
argument of type 'NoneType' is not iterable
```

This is the stacktrace:
```
File "/usr/local/Cellar/dbt/0.20.0_1/libexec/lib/python3.8/site-packages/dbt/parser/schemas.py", line 560, in check_format_version
    if 'version' not in yaml_dct:
TypeError: argument of type 'NoneType' is not iterable
```

That's very surprising! The error cropping up in this line:
https://github.com/dbt-labs/dbt/blob/f460d275baa4df8cf8713f27e198a3577d6ea71c/core/dbt/parser/schemas.py#L560

Means that `yaml_dct` must be set to `None`, somehow. That method is called here:
https://github.com/dbt-labs/dbt/blob/f460d275baa4df8cf8713f27e198a3577d6ea71c/core/dbt/parser/read_files.py#L21-L23

It's pretty hard to imagine [`yaml_from_file`](https://github.com/dbt-labs/dbt/blob/f460d275baa4df8cf8713f27e198a3577d6ea71c/core/dbt/parser/schemas.py#L99) returning `None`. I've tried a couple different things to trip this error (an empty `.yml` file, malformed contents, simply the word `'None'`), but none (hah) of my attemps has worked. The only way I've managed to reproduce it is by stepping directly into the `validate_yaml` or `yaml_from_file` method calls and setting the yaml dictionary to `None`.

One important hint: The user reports that this project runs fine on earlier versions of dbt. It's only with the upgrade to v0.20.0 that they started seeing this error.

### Steps To Reproduce
???

### Expected behavior
Definitely not this error!

```
/usr/local/Cellar/dbt/0.20.0_1/
```
(i.e. dbt v0.20.0 installed via Homebrew)

nathaniel-may:
resolves #3568

### Description

Adds handling for `None` response from yaml loader. Triggered in 0.20.0 by files with all comments.

### Requested Feedback

Are there more places I should be putting yml files with all comments in them? It looks like all the different file types go through the same pipeline, but if there's an exception to that we should test it too.

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 - Add full node selection to source freshness command and align selection syntax with other tasks (`dbt source freshness --select source_name` --> `dbt source freshness --select source:souce_name`) and rename `dbt source snapshot-freshness` -> `dbt source freshness`. ([#2987](https://github.com/dbt-labs/dbt/issues/2987), [#3554](https://github.com/dbt-labs/dbt/pull/3554))
 
 ### Fixes
+- Fix exception on yml files with all comments [3568](https://github.com/dbt-labs/dbt/issues/3568)
 - Fix docs generation for cross-db sources in REDSHIFT RA3 node ([#3236](https://github.com/fishtown-analytics/dbt/issues/3236), [#3408](https://github.com/fishtown-analytics/dbt/pull/3408))
 - Fix type coercion issues when fetching query result sets ([#2984](https://github.com/fishtown-analytics/dbt/issues/2984), [#3499](https://github.com/fishtown-analytics/dbt/pull/3499))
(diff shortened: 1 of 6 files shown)


iknox-fa (approved):
LGTM!

kwigley (approved):

gshank (approved):
Looks good to me.

nathaniel-may:
rebased onto develop for that sweet sweet green.

PR #3646, opened 2021-07-29
Use SchemaParser render context to render test configs

kwigley:
resolves #3564 

### Description

Instead of creating a new context for the test builder to jinja render test configs, we can just use the context from the schema parser that has the right context properties to support `var()` with project configured vars in addition to cli vars.

### Checklist
 - [x] I have signed the [CLA](https://docs.getdbt.com/docs/contributor-license-agreements)
 - [x] I have run this code in development and it appears to resolve the stated issue
 - [x] This PR includes tests, or tests are not required/relevant for this PR
 - [ ] I have updated the `CHANGELOG.md` and added information about my change to the "dbt next" section.

PR #3650, opened 2021-07-29
[backport] Use SchemaParser render context to render test configs (#3646)

kwigley:
* use available context when rendering test configs

* add test

* update changelog


leahwicz (approved):

kwigley:
merging as admin because failing test will be addressed by #3651

PR #3669, opened 2021-08-02
Moving CI to GitHub actions

### LINKED ISSUE #3688: Move CI to GitHub Actions
see title 😄 

- helpful for all of CI to be located in one place
- logs and artifacts are easily accessible for the same UI the codebase lives in
- noticeable increase in speed
- ability to control concurrency if we notice parallel test runs overloading sandbox warehouses
- a safer way to run CI for PRs from forked repos
- support for windows, macos, and ubuntu runners

kwigley:
### Description

resolves #3688

This PR moves our CI to GitHub actions. I created the following 2 workflows.

### main.yml
This workflow is the "main" CI workflow. This will run on all PRs, pushes to mainline branches ('develop', '*.latest`, etc.), and when manually dispatched in the GitHub UI. Today this workflow run static code checks (mypy, flake8) and unit tests. This could also include other lint checks, formatting checks, etc. This also builds, tests, and stores python distributions (wheel and source) as well, this can be used in other workflows (build docker image, publishing, etc.). This workflow runs for PRs from forked repositories and GitHub Actions does not pass secrets to forks, so this workflow would fail if it tried to access them.

### integration.yml
**This workflow can be triggers by untrusted actors.**

Be default, only integration test affected by the code changes in the PRs will be run. The mapping of file changes and tests run can be found here: https://github.com/dbt-labs/dbt/blob/0d556de6994891561b408fe53ba8c521bc165730/.github/workflows/integration-pr.yml#L43-L55
If someone would like to run all tests or tests for a specific adapter, you can add a label to the pr to do so.
`test all` -> run all integration tests
`test ${adapter}` -> run tests for that specific ${adapter}

It also requires an additional check when the PR is from a forked repository. Right now, that additional check is a PR label `ok to test`. This extra check is needed to ensure that outside contributors do not exploit any secrets or compute. Since this action runs on the `pull_request_target` event, it is important to note that the action runs in the context of the "target" of a PR. So, for a PR pointing at the develop branch, the action will run on the develop branch, but also include all secrets and even a github token with write permissions. We explicitly checkout the head of the branch that the PR is coming from in order to run checks against the code to be merged. Since only folks with write access to the repo can add labels to PRs, we can ensure that the PR is coming from a trusted actor. Please see https://securitylab.github.com/research/github-actions-preventing-pwn-requests/ for more info why we should do this.

### Additional notes
While testing these new workflows, I discovered an issue with python3.9 unit tests. I have turned them off for the time being and created #3689 to follow up with a fix. https://github.com/dbt-labs/dbt/blob/0d556de6994891561b408fe53ba8c521bc165730/.github/workflows/main.yml#L34 

I think there is still work to be done to increase the reliability and trust in our test suite, but I believe that this PR will contribute towards that.

### Checklist


kwigley:
Still need to update repo secrets with the relevant warehouse credentials in order to integration tests to work.

.github/workflows/integration-pr.yml
```diff
@@ -0,0 +1,221 @@
...
+            snowflake:
+              - 'core/**'
+              - 'plugins/snowflake/**'
+            bigquery:
+              - 'core/**'
+              - 'plugins/bigquery/**'
+            redshift:
+              - 'core/**'
+              - 'plugins/redshift/**'
+              - 'plugins/postgres/**'
```
kwigley, line 55: This is included because redshift code inherits a lot of things from the postgres implementation.

.github/workflows/main.yml
```diff
@@ -0,0 +1,71 @@
...
+jobs:
+  test:
+    name: ${{ matrix.tox-env }} ${{ matrix.python-version }} / ${{ matrix.os }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest]
+        tox-env: [unit, mypy]
+        python-version: [3.6, 3.7, 3.8] # TODO: add support unit testing python 3.9
```
kwigley, line 34:
```suggestion
        python-version: [3.6, 3.7, 3.8] # TODO: support unit testing for python 3.9
```

.github/workflows/build.yml
```diff
@@ -0,0 +1,98 @@
...
+  build:
+    name: Build python
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out the repository
+        uses: actions/checkout@v2
+        with:
+          persist-credentials: false
+
+      - name: Set up Python ${{ matrix.python-version }}
```
leahwicz, line 35: There is no matrix for python version in this job and below you hard code 3.8 as the version. Are you only building with 3.8 and if so then this can be removed

.github/workflows/integration-pr.yml
```diff
@@ -0,0 +1,198 @@
...
+
+  test:
+    name: ${{ matrix.adapter }} ${{ matrix.python-version }} / ${{ matrix.os }}
+
+    if: >-
+      github.event.pull_request.head.repo.full_name == github.repository ||
```
leahwicz, line 61: Is this checking if it's from a fork?

.github/workflows/integration-pr.yml
```diff
@@ -0,0 +1,198 @@
...
+
+concurrency:
+  group: integration-pr-${{ github.event_name }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  pre-checks:
+    runs-on: ubuntu-latest
+    outputs:
+      has-postgres-changes: ${{ steps.check-changes.outputs.postgres == 'true' }}
```
leahwicz, line 28: Does the `path-filter` action update these output variables?

.github/workflows/integration-pr.yml
```diff
@@ -0,0 +1,198 @@
...
+  post-checks:
+    runs-on: ubuntu-latest
+    needs: test
+    permissions:
+      pull-requests: write
+    steps:
+      - name: Needs permission PR comment
+        if: >-
+          needs.test.result == 'skipped' &&
+          github.event_name == 'pull_request_target' &&
```
leahwicz, line 188: This seems redundant since the only event that can trigger this is a 'pull_request_target' event

.github/workflows/main.yml
```diff
@@ -0,0 +1,78 @@
...
+            python-version: 3.6
+          - os: ubuntu-latest
+            tox-env: mypy
+            python-version: 3.7
+          - os: ubuntu-latest
+            tox-env: mypy
+            python-version: 3.8
+          - os: ubuntu-latest
+            tox-env: mypy
+            python-version: 3.9
```
leahwicz, line 50: You don't have `3.9` in the above list so does this work?

docker-compose.yml
```diff
@@ -19,7 +19,7 @@ services:
         GROUP_ID: ${GROUP_ID:-}
     command: "/root/.virtualenvs/dbt/bin/pytest"
     environment:
-      DOCKER_TEST_DATABASE_HOST: "database"
+      POSTRES_TEST_HOST: "database"
```
leahwicz, line 22: Spelling mistake: POSTGRES_TEST_HOST

.github/workflows/integration-pr.yml
```diff
@@ -0,0 +1,198 @@
+# **This is not a trusted workflow.**
+# **what?** - This workflow runs integration tests for supported OS,
+# python versions and core adapters. It will only run tests for
+# related file changes. Use the `test all` and `test ${adapter}`
+# label to manually run all or additional tests. Use `ok to test`
+# label to mark PRs from forked repositories that are safe to run integration
+# tests for. Requires secrets to run against different warehouses.
+# **why?** - This checks the functionality of dbt from a user's perspective,
+# provides code coverage, and attempts to catch functional regressions.
```
leahwicz, line 9: For the code coverage aspect, does it just print the results to the console? Can we have it output as an artifact?

.github/workflows/build.yml
kwigley: ah ya good call!

.github/workflows/integration-pr.yml
kwigley: This is confirming that the PR is _not_ coming from a fork (repo of the pull request == repo where the action is running), and if it _is_ coming from a fork, we confirm that we have a label 'ok to test' present.

kwigley: So no need for label for contributors with write access! We can explore automating the 'ok to test' label as well.

.github/workflows/integration-pr.yml
kwigley: ya, that's correct. I added a comment in the code pointing to its docs

https://github.com/marketplace/actions/paths-changes-filter

.github/workflows/integration-pr.yml
kwigley: Good call, I went through a few iterations and never cleaned this up. rm'ing 👍

.github/workflows/main.yml
kwigley: Yes, it does. The items in the `include` list are added after the matrix if evaluated and flattened I believe. I went back and forth this specific matrix. I think it is fine to just run flake8 and mypy once, with a single python version. I'll update it so that is the case.

docker-compose.yml
kwigley: 🙏

.github/workflows/integration-pr.yml
kwigley: Oh, I thought I'd be able to set up coverage reporting, but it is trickier than I thought! I'll remove that statement for now.

Here is the coverage tool that integrates with pytest that I was trying to use: https://pytest-cov.readthedocs.io/en/latest/reporting.html

I actually like how codecov works, you can send many coverage reports for the same PR run and it will combine them on server side and we are able to set up checks that _new_ code is has test coverage. This is awesome because it will work great with unit tests and integration tests in different workflows and it is possible only a subset of integration tests could run for a PR. I like this approach vs looking at project wide code coverage! (https://github.com/codecov/codecov-action)

This is also a pretty cool idea for reporting tests failures in the codebase, though, it could get noisy: https://github.com/utgwkk/pytest-github-actions-annotate-failures

.github/workflows/integration-pr.yml
```diff
@@ -0,0 +1,196 @@
+# **This is not a trusted workflow.**
+# **what?** - This workflow runs integration tests for supported OS,
+# python versions and core adapters. It will only run tests for
+# related file changes. Use the `test all` and `test ${adapter}`
```
gshank, line 4: Does "related file changes" here just mean adapters that have changes? Because in previous job we did have functionality that would attempt to only run the tests that were necessary to cover code changes.

gshank (commented):
This looks good to me, but I know very little about this area.

leahwicz (approved):

.github/workflows/integration-pr.yml
kwigley: Yes, this is just for adapters. So it is not super helpful for folks working in the core codebase, since it could affect all adapters. That would be pretty neat if we could only run a subset of tests related to actual code changes though.

.github/actions/setup-postgres-macos/action.yml
```diff
@@ -0,0 +1,24 @@
...
+        while [ $i -gt 0 ]; do
+            echo "Check PostgreSQL service status"
+            eval $COMMAND && break
+            ((i--))
+            if [ $i == 0 ]; then
+                echo "PostgreSQL service not ready, all attempts exhausted"
+                exit 1
+            fi
+            echo "PostgreSQL service not ready, wait 10 more sec, attempts left: $i"
+            sleep 10
```
nathaniel-may, line 21: fairly minor, but I find this ordering easier to read.
```suggestion
        while [ $i -gt -1 ]; do
            if [ $i == 0 ]; then
                echo "PostgreSQL service not ready, all attempts exhausted"
                exit 1
            fi
            echo "Check PostgreSQL service status"
            eval $COMMAND && break
            echo "PostgreSQL service not ready, wait 10 more sec, attempts left: $i"
            sleep 10
            ((i--))
```

.github/actions/setup-postgres-windows/setup_db.sh
```diff
@@ -0,0 +1 @@
+../../../test/setup_db.sh
```
nathaniel-may, line 1: Since this is the same in a few places should we just make it one action under `setup-postgres-all`? I'm not opposed to this either though if it feels better organized to you.

Makefile
```diff
@@ -42,31 +42,31 @@ integration-postgres: .env ## Runs postgres integration tests with py38.
 
 .PHONY: integration-postgres-fail-fast
 integration-postgres-fail-fast: .env ## Runs postgres integration tests with py38 in "fail fast" mode.
-	$(DOCKER_CMD) tox -e py38-postgres -- -x -nauto
+	$(DOCKER_CMD) tox -e py38-postgres -- -x
```
nathaniel-may, line 45: what was -nauto?

core/setup.py
```diff
@@ -2,6 +2,7 @@
 import os
 import sys
 
+
```
nathaniel-may, line 5: let's remove this fmt change from this PR.

test/integration/base.py
```diff
@@ -243,7 +243,7 @@ def snowflake_profile(self):
         }
 
     def bigquery_profile(self):
-        credentials_json_str = os.getenv('BIGQUERY_SERVICE_ACCOUNT_JSON').replace("'", '')
+        credentials_json_str = os.getenv('BIGQUERY_TEST_SERVICE_ACCOUNT_JSON').replace("'", '')
```
nathaniel-may, line 246: I know this isn't code you added, but why is this replace necessary? might be worth a comment if we know.

test/unit/test_parser.py
```diff
@@ -69,6 +70,9 @@ def _generate_macros(self):
     def setUp(self):
         dbt.flags.STRICT_MODE = True
         dbt.flags.WARN_ERROR = True
+        # HACK: this is needed since tracking events can 
+        # be sent when using the model parser
+        tracking.do_not_track()
```
nathaniel-may, line 75: We should probably put a ticket number for a relevant tech debt ticket in here.

nathaniel-may (changes requested):
This is FANTASTIC. I am so looking forward to having shiny new CI. I do have some security questions around the untrusted workflows that I left below as well as an onslaught of random questions. I would also like to request that we take out the fmt changes to all the md files to reduce bloat and have a cleaner history.

.github/workflows/integration-pr.yml
nathaniel-may: It'll still help for when we make changes to top level md files, or the soon to be merged performance directory.

.github/actions/setup-postgres-windows/setup_db.sh
kwigley: I was a little limited in what I could do here. I wasn't able to figure out how to run conditional steps in the same actions. I attempted to make a single action, and only run the step for the particular OS it was targeting. For some reason, I couldn't use something like: `if: os == 'windows'`. I landed on splitting up the actions per OS and moving the conditional logic in the workflow https://github.com/dbt-labs/dbt/blob/bdc42e3d040e68122ab5a8b643f87597f459d69a/.github/workflows/integration.yml#L50-L66

I don't think I had to symlink this file though! I think I could just de `bash test/setup_db.sh` in the action, but I wasn't 100%. I'll try this out, but I won't consider it a blocker for merging.

nathaniel-may: Fair enough. It's not actually that awkward.

Makefile
kwigley: It is an argument for a pytest plugin that is used to split up tests and run in separate processes ([docs](https://github.com/pytest-dev/pytest-xdist#speed-up-test-runs-by-sending-tests-to-multiple-cpus)). It does not play nicely with the the `-x` arg (exit instantly on first error or failed test).

kwigley: I take that back, this is fine! It doesn't work with `-s` (turns of stdout capturing) I'll clean up these changes, they aren't related anyway.

.github/workflows/integration-pr.yml
```diff
@@ -0,0 +1,196 @@
...
+# label to manually run all or additional tests. Use `ok to test`
+# label to mark PRs from forked repositories that are safe to run integration
+# tests for. Requires secrets to run against different warehouses.
```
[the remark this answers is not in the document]
iknox-fa: I also have this question (and in general with the use of labels since they can be added at any time, not just at PR creation).

scripts/build-dist.sh
```diff
@@ -1,11 +1,12 @@
-#!/bin/bash -eo pipefail
+#!/bin/bash 
+
+set -eo pipefail
 
 DBT_PATH="$( cd "$(dirname "$0")/.." ; pwd -P )"
 
 PYTHON_BIN=${PYTHON_BIN:-python}
 
-echo $SCRIPTPATH
```
iknox-fa, line 7: Why was this removed?  I don't see where $SCRIPTPATH is defined, but I assumed it used to output something...

iknox-fa (approved):
A few comments, but nothing to stop a merge.
👍

scripts/build-dist.sh
kwigley: I confirmed that this was just echoing a newline in CI currently, it isn't set as env var anywhere that I can tell in the current CI.

https://circleci.com/api/v1.1/project/github/dbt-labs/dbt/46195/output/102/0?file=true&allocation-id=6115a0dab070f52e9c5bdfa7-0-build%2F5A11EAF0

It is the line right before `/home/tox/build_venv//bin/python` aka `$PYTHON_BIN`
```console
  WARNING: The script flake8 is installed in '/home/tox/build_venv/bin' which is not on PATH.
  Consider adding this directory to PATH or, if you prefer to suppress this warning, use --no-warn-script-location.
Successfully installed backcall-0.2.0 backports.entry-points-selectable-1.1.0 bleach-4.0.0 bump2version-1.0.1 bumpversion-0.6.0 click-8.0.1 decorator-5.0.9 distlib-0.3.2 docutils-0.17.1 execnet-1.9.0 filelock-3.0.12 flake8-3.9.2 flaky-3.7.0 freezegun-0.3.12 iniconfig-1.1.1 ipdb-0.13.9 ipython-7.26.0 ipython-genutils-0.2.0 jedi-0.18.0 matplotlib-inline-0.1.2 mccabe-0.6.1 mypy-0.782 mypy-extensions-0.4.3 parso-0.8.2 pep517-0.11.0 pexpect-4.8.0 pickleshare-0.7.5 pip-tools-6.2.0 pkginfo-1.7.1 platformdirs-2.2.0 pluggy-0.13.1 prompt-toolkit-3.0.19 ptyprocess-0.7.0 py-1.10.0 pycodestyle-2.7.0 pyflakes-2.3.1 pygments-2.9.0 pytest-6.2.4 pytest-dotenv-0.5.2 pytest-forked-1.3.0 pytest-logbook-1.2.0 pytest-xdist-2.3.0 python-dotenv-0.19.0 readme-renderer-29.0 requests-toolbelt-0.9.1 rfc3986-1.5.0 toml-0.10.2 tomli-1.2.1 tox-3.24.1 tqdm-4.62.0 traitlets-5.0.5 twine-3.4.2 typed-ast-1.4.3 virtualenv-20.7.2 wcwidth-0.2.5 webencodings-0.5.1 wheel-0.37.0

/home/tox/build_venv//bin/python
+ rm -rf /home/dbt_test_user/project/dist
+ rm -rf /home/dbt_test_user/project/build
+ mkdir -p /home/dbt_test_user/project/dist
+ for SUBPATH in core plugins/postgres plugins/redshift plugins/bigquery plugins/snowflake
+ rm -rf /home/dbt_test_user/project/core/dist
+ rm -rf /home/dbt_test_user/project/core/build
+ cd /home/dbt_test_user/project/core
+ /home/tox/build_venv//bin/python setup.py sdist bdist_wheel
running sdist
running egg_info
```

.github/workflows/integration.yml
```diff
@@ -0,0 +1,114 @@
...
+        if: |
+          matrix.adapter == 'postgres' &&
+          runner.os == 'macOS'
+        uses: ./.github/actions/setup-postgres-macos
+
+      - name: Set up postgres (windows)
+        if: |
+          matrix.adapter == 'postgres' &&
+          runner.os == 'Windows'
+        uses: ./.github/actions/setup-postgres-windows
```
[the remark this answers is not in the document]
kwigley: confirmed that this path works fine!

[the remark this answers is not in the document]
kwigley: I think because we are using a bash shell?

PR #3691, opened 2021-08-04
#3682 Changed how tables and views are generated to be able to use differen…

### LINKED ISSUE #3682: BigQuery KMS should not be used on VIEWS
### Describe the bug
If I add one kms key for the entire project that should be used for all models, like below. It will blow up, because you can't add KMS encryption to views. ( Unless I'm missing something here )

```yml
models:
  test:
    +kms_key_name: 'projects/<project>/locations/<location>/keyRings/key-ring/cryptoKeys/key'
```

### Steps To Reproduce
Add kms_key_name to the entire project or specific folder that contains views.

### Expected behavior
It should skip views, and only add the encryption key to the tables.

### System information
- [ ] postgres
- [ ] redshift
- [x] bigquery
- [ ] snowflake

```
dbt --version
installed version: 0.20.0
   latest version: 0.20.0

Up to date!

Plugins:
  - bigquery: 0.20.0
  - snowflake: 0.20.0
  - redshift: 0.20.0
  - postgres: 0.20.0
```

MAC

### Additional context
Should probably be in its own issue, but I would like to be able to add default encryption key to the entire dataset that DBT creates. Maybe this is possible, but I didn't find anything about it :)

Cheers!

AndreasTA-AW:
…t options

resolves #3682

### Description
I added a new function called common where common options can be put, otherwise views and tables are separated from each other to be able to add either for both or a single one of them.

Not sure if it's good enough or what you say. I ran the tests but didn't touch them.

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@
   - to simplify config (default usage of `copy_materialization='table'` if is is not found in global or local config)
   - to let copy several source tables into single target table at a time. ([Google doc reference](https://cloud.google.com/bigquery/docs/managing-tables#copying_multiple_source_tables))
 
+### Fixes
+
+- Separated table vs view configuration for BigQuery since some configuration is not possible to set for tables vs views. ([#3691](https://github.com/dbt-labs/dbt/issues/3682))
+
 ### Under the hood
 
 - Use GitHub Actions for CI ([#3688](https://github.com/dbt-labs/dbt/issues/3688), [#3669](https://github.com/dbt-labs/dbt/pull/3669))
@@ -15,6 +19,11 @@ Contributors:
 - [@xemuliam](https://github.com/xemuliam) ([#3606](https://github.com/dbt-labs/dbt/pull/3606))
 
 ## dbt 0.21.0b2 (August 19, 2021)
+
+### Fixes
+
+
+## dbt 0.21.0b1 (August 03, 2021)
 
 ### Features
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,11 +20,6 @@ Contributors:
 
 ## dbt 0.21.0b2 (August 19, 2021)
 
-### Fixes
-
-
-## dbt 0.21.0b1 (August 03, 2021)
-
 ### Features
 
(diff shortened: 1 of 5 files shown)


AndreasTA-AW:
I'll ping here. Would be nice if someone looks at this. I will sign the agreement when we think this will eventually be merged :)

jtcohen6 (commented):
Thanks for this @AndreasTA-AW! The code looks good to me. Appreciate the ping, and sorry for the delay getting back to you.

As far as testing this change, I think a unit test in the style of [these ones](https://github.com/dbt-labs/dbt/blob/9e796671dd55d4781284d36c035d1db19641cd80/test/unit/test_bigquery_adapter.py#L843-L869) should be sufficient. How about something like:
```python
    def test_table_kms_key_name(self):
        adapter = self.get_adapter('oauth')
        mock_config = create_autospec(
            RuntimeConfigObject)
        config={'kms_key_name': 'some_key'}
        mock_config.get.side_effect = lambda name: config.get(name)

        expected = {
            'kms_key_name': "'some_key'"
        }
        actual = adapter.get_table_options(mock_config, node={}, temporary=False)
        self.assertEqual(expected, actual)

        
    def test_view_kms_key_name(self):
        adapter = self.get_adapter('oauth')
        mock_config = create_autospec(
            RuntimeConfigObject)
        config={'kms_key_name': 'some_key'}
        mock_config.get.side_effect = lambda name: config.get(name)

        expected = {}
        actual = adapter.get_view_options(mock_config, node={})
        self.assertEqual(expected, actual)
```
After adding that test, could you:
- Pull changes from `develop`
- Add a changelog entry below `## dbt 0.21.0 (Release TBD)`
- Sign the CLA

AndreasTA-AW:
The tests looks great to me, I'll add those and make sure they pass. I'll then finish the rest of the tasks you added.

Thanks for looking in to it!

jtcohen6:
@AndreasTA-AW Will you be able to sign the CLA? The PR looks in good shape, so that's the only blocker to merging this in.

AndreasTA-AW:
Hi!
I'm actually waiting to get a GO on that, hopefully it will be solved anytime.

AndreasTA-AW:
@jtcohen6 Sorry for the delay, I got an OK to sign it now, so it should all be fixed.

jtcohen6 (approved):
Thanks for the contribution @AndreasTA-AW!

PR #3693, opened 2021-08-04
Basic local FS functionality

iknox-fa:
PR for first segment of local FS storage adapter.
Contains:
- adapter logic
- storage client (loads adapter)
- typing work for hinting and mypy
- integration into codebase in all places where the old system client was used.

This breaks a few unit tests and 1 integration test.  The unit test failures are utilizing MagicMocks which don't seem to play nice as file paths for some reason.  The integration test failure is in partial parsing testing, needs further investigation


core/dbt/adapters/internal_storage/local_filesystem.py
```diff
@@ -0,0 +1,218 @@
...
+
+    return file_content
+
+
+def write(
+    path: str,
+    content: Union[str, bytes, None],
+    overwrite: bool = False,
+) -> bool:
+    """Writes the given content out to a resource on the filesystem.
```
leahwicz, line 69: This comment block is great 💯 !

core/dbt/adapters/internal_storage/local_filesystem.py
```diff
@@ -0,0 +1,218 @@
...
+        `True` for success, `False` otherwise.
+
+    """
+    # TODO: double check I hit all possible permutations here! (IK)
+
+    # create a concrete path object
+    path: Path = Path(path)
+
+    # handle overwrite errors for files and directories
+    if path.exists() and (overwrite is False or path.is_dir() is True):
```
leahwicz, line 99: If directory already exists, that essentially is a no-op. Would you consider that not successful? I honestly don't know how that effects things in the code, but it just seemed odd to me

core/dbt/adapters/internal_storage/local_filesystem.py
```diff
@@ -0,0 +1,218 @@
...
+
+    Args:
+        path: Full path of resource to be deleted
+
+    """
+    # create concrete path object
+    path: Path = Path(path)
+
+    # ensure resource to be deleted exists
+    if not path.exists():
```
leahwicz, line 160: Same question here, if it doesn't already exist, would you call it not successful?

core/dbt/adapters/internal_storage/local_filesystem.py
```diff
@@ -0,0 +1,218 @@
...
+        return False
+
+    # remove files
+    if path.is_file():
+        path.unlink()
+
+    # remove directories recuersivly, surprisingly obnoxious to do in a cross-platform safe manner
```
leahwicz, line 167:
```suggestion
    # remove directories recursively, surprisingly obnoxious to do in a cross-platform safe manner
```

core/dbt/clients/storage.py
```diff
@@ -0,0 +1,27 @@
+from importlib import import_module
+from os import getenv
+from dbt.logger import GLOBAL_LOGGER as logger
+from .storage_adapter_type import StorageAdapterType
+from typing import cast
+
+# TODO:
+# ensure configured adapter has the correct module signature
+# provide better not-ready error messagin
```
leahwicz, line 9:
```suggestion
# provide better not-ready error messaging
```

leahwicz (approved):
Other than those few comments, LGTM!

core/dbt/adapters/TODO.md
```diff
@@ -0,0 +1 @@
+Move the warehouse adapters to a subfolder of adapters (adapters/warehouse).
```
jtcohen6, line 1: These are in `plugins/` right now (and three of them are soon to leave this repo). Let's think about whether `local_filesystem` makes more sense under `plugins/` or as part of `core/`.

Also... we'll need to come up with a new word for this, since none of warehouse, database, query engine, etc is quite perfect. To date, I've just been using "adapter," but obviously we've got adapters for more things now :)

core/dbt/adapters/internal_storage/local_filesystem.py
```diff
@@ -1,557 +1,218 @@
...
+    # conditionally strip whitespace
+    file_content = file_content.strip() if strip else file_content
+
+    return file_content
+
+
+def write(
+    path: str,
+    content: Union[str, bytes, None],
+    overwrite: bool = False,
```
gshank, line 67: I'm not sure that "overwrite" should be off by default. Generally when we write out a file we really do want to replace it if it already exists. I think this ended up here because of combining two formerly different methods. It's the reason that the partial parsing tests are failing, because the partial parse msgpack file isn't being written out. And I think that would be the case for a lot of other cases, but that tends not to be the sort of thing we cover in our tests.

iknox-fa: Ahh.. Good to know.  Yeah we can switch that up.

core/dbt/adapters/TODO.md
iknox-fa: @jerco That makes sense.  I'll do some folder juggling and see where we land.  As far as naming goes.. I've been leaning towards "warehouse adapters" and "storage adapters", but I don't hold that opinion too strongly.

core/dbt/adapters/internal_storage/local_filesystem.py
iknox-fa: Yeah it's a kind of funny interaction to model out.  I do consider it a "fail" (hence it returns false), but one that may be expected under certain circumstances.  For example, you can do things like to populate a directory with files (say our snowplow event cookie)

```
if SA.write('path/that/might/be/here', None):
    SA.write('path/that/might/be/here/'important_cookie_file.json, TEMPLATE_STRING)
```

core/dbt/adapters/TODO.md
kwigley: lmk if you want to chat plugins! I think I have a good sense how that could look

kwigley (approved):
looks good! just left some questions for you

core/dbt/adapters/internal_storage/local_filesystem.py
iknox-fa: Thanks for correcting my spelling errors :). I should have read more carefully.

core/dbt/adapters/internal_storage/local_filesystem.py
```diff
@@ -0,0 +1,218 @@
...
+
+    Overwrites are only supported for files and are toggled by the overwrite parameter.
+
+    All logical cases outside those outlined above will result in failure
+
+    Args:
+        path: Full path of resource to be written.
+        content: Data to be written.
+        overwrite: Wether or not to overwrite if a file already exists at this path.
+        parser: A parser to apply to file data.
```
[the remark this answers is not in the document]
iknox-fa: I was going to build common parsers (json, msgpack, etc) into the adapter but that ended up being more of a hassle than it was worth for now .. I'll fix that mention.

core/dbt/clients/storage_adapter_type.py
```diff
@@ -0,0 +1,37 @@
...
+    ) -> bool:
+        pass
+
+    @staticmethod
+    def delete(path: str) -> bool:
+        pass
+
+    @staticmethod
+    def find() -> None:
+        pass
```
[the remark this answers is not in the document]
iknox-fa: No, it's a reminder to me that I need to incorporate the `find_matching` method from the `dbt.client.system`

nathaniel-may:
pulling myself off because it looks like you have plenty of reviews. Let me know if you need me for any part of this though.

PR #3707, opened 2021-08-08
add support for execution project for BigQuery adapter

Kayrnt:
resolves #

### Description

add support for execution project for BigQuery adapter inside the profile adapter configuration

### Checklist


jtcohen6 (commented):
@Kayrnt The change here looks good! Sorry for the delayed response.

I was able to test this locally, and confirm that dbt queries use the `execution_project` for billing/execution, even as they create objects in the default project.

**Naming:** Just curious, what was your rationale for picking `execution_project` over some of the other names I've seen proposed (e.g. `billing_project`, `jobuser_project`)? I think I like `execution_project`, just want to make sure we're making a decision we won't regret later on :)

**Testing:** We have access to an alternate BigQuery project within our integration testing environment. So we could create [an additional BigQuery profile target](https://github.com/dbt-labs/dbt/blob/51a76d0d63efd8a393ace482c6c93afcf75f446c/test/integration/base.py#L247) that sets the `execution_project` to [`alternative_database()`](https://github.com/dbt-labs/dbt/blob/51a76d0d63efd8a393ace482c6c93afcf75f446c/test/integration/base.py#L318), run some models/tests/etc, and then query [`INFORMATION_SCHEMA.JOBS_BY_PROJECT`](https://cloud.google.com/bigquery/docs/information-schema-jobs) filtering on the alternative `project_id`. That arbitrary query could happen with the [`run_sql_bigquery`](https://github.com/dbt-labs/dbt/blob/51a76d0d63efd8a393ace482c6c93afcf75f446c/test/integration/base.py#L646) method.

Does writing that test feel like a feasible next step for you, or would you want some additional help/pointers?

Kayrnt:
Hello @jtcohen6, great to read to that it works as intended so far.

*Naming*
To be honest, I didn't really thought that much through as that's the naming we use and refer to internally at Teads since we  started to need to decorrelate both kind of projects on our queries. Reading the documentation while writing issue description, I didn't feel like alternatives were very explicit:

- `billing_project`: it feels like related to billing while slots are bought as commitments which are global to the organisation and then are related to projects through reservations & assignments. `assignment_project` (or similar) doesn't feel right as it requires a clear understanding of the resource hierarchy in my opinion.

- `jobuser_project`: When I read it implies I link "jobuser" to the query execution because it's what the BigQuery role allows to. Again I'm not sure it's clear for everyone without documentation.

- `execution_project`: it reads as it means it's the project used for (query) execution. Of course it might bring some confusion to people as the other config parameter is "project" whereas we use on our end `datasets_project` (vs `execution_project`) when it comes to the project used for storing the datasets & tables.

I didn't settle for a specific naming yet I couldn't find a more appropriate name.

*Testing*
Thanks for all the details, I think I should be able to write it.
I'll likely be able to find some spare time by the end of the week so unless there is a time constraint for you, I can take it.

Kayrnt:
I tried to implement the test but I fall short at least on a part.
Here's was my plan so far:
- Add a `bigquery_alternate` profile to base
- Create a new test in the suite that would run a simple model (a simple `select 1`)
- Create a test that would read the `INFORMATION_SCHEMA` filtering on the job_id from the model and check that job_project is the same as the one from the alternate database

However I'm not sure how to retrieve the job_id on a DBT run inside the test. Is it doable?

test/integration/022_bigquery_test/test_bigquery_execution_project.py
```diff
@@ -0,0 +1,21 @@
...
+        return "execution-project-models"
+
+    @use_profile('bigquery_alternate')
+    def test__bigquery_execute_project(self):
+        results = self.run_dbt(['run', '--models', 'model'])
+        self.assertEqual(len(results), 1)
+        execution_project = os.environ['BIGQUERY_TEST_ALT_DATABASE']
+        self.run_dbt(['test', '--vars',
+                      '{ project_id: %s, job_id: %s }'%(execution_project, "")],
+                     expect_pass=True)
```
jtcohen6, line 21: It isn't possible to access the `job_id`. If anonymous usage tracking were turned on, you could get the `invocation_id` (which is persisted as a job label), but we have it turned off for integration tests.

So I think your best bet is to pass in `self.unique_schema()` and check for something like:
```sql
select project_id from `region-us`.INFORMATION_SCHEMA.JOBS_BY_PROJECT
where date(creation_time) = current_date
  and destination_table.dataset_id = 'test16308377135774383339_bigquery_test_022'
```
And make sure that project_id matches `self.alternative_database`. Whether you'd rather perform that comparison in a custom generic (dbt) test, or use the arbitrary SQL runner (`run_sql_bigquery`) and perform an `assert` check in the python code here, is up to you.

test/integration/base.py
```diff
@@ -266,6 +266,12 @@ def bigquery_profile(self):
             }
         }
 
+    def bigquery_alternate_profile(self):
+        bigquery_profile = self.bigquery_profile()
+        execution_project = self.alternative_database
+        bigquery_profile['test']['outputs']['default2']['execution_project'] = execution_project
+        return bigquery_profile
```
jtcohen6, line 273: Rather than defining an additional profile, I think you can just add a new target above that also defines `execution_project`. Then you can just pass `--target alternate` as an additional argument in the test.
```python
        return {
            'config': {
                'send_anonymous_usage_stats': False
            },
            'test': {
                'outputs': {
                    'default2': {
                        'type': 'bigquery',
                        'method': 'service-account-json',
                        'threads': 1,
                        'project': project_id,
                        'keyfile_json': credentials,
                        'schema': self.unique_schema(),
                    },
                    'alternate': {
                        'type': 'bigquery',
                        'method': 'service-account-json',
                        'threads': 1,
                        'project': project_id,
                        'keyfile_json': credentials,
                        'schema': self.unique_schema(),
                        'execution_project': self.alternative_database,
                    },
                },
                'target': 'default2'
            }
        }
```

test/integration/022_bigquery_test/execution-project-models/model.sql
```diff
@@ -0,0 +1,5 @@
+{{ config(materialized='incremental') }}
+select 1 as id
+{% if is_incremental() %}
+this is a syntax error!
+{% endif %}
```
jtcohen6, line 5: I imagine you copy-pasted this from a different test case. I think the query can just be `select 1 as id`, no need for the extra complexity.

jtcohen6 (commented):
Nice work, I think the test case is really close!

test/integration/022_bigquery_test/test_bigquery_execution_project.py
Kayrnt: OK I tried to use `self.unique_schema()`, I'm not confident it's meant to be used, is it what you had in mind?
I wented for a custom generic DBT test as I felt it make it "reusable" (though I hardly think at this point we would have to with the same macro).

test/integration/base.py
Kayrnt: Ok I updated that part 👍

test/integration/022_bigquery_test/execution-project-models/model.sql
Kayrnt: Indeed it's copy pasted I was wondering if the incremental check could be useful as I saw it elsewhere

Kayrnt:
@jtcohen6 it looks like tests are now passing, is it meeting your expectations?

jtcohen6 (commented):
@Kayrnt Looks great! Thanks for making those updates, the test case is really clear & straightforward.

Could you add an entry to the [Changelog](https://github.com/dbt-labs/dbt/blob/develop/CHANGELOG.md#dbt-0210-release-tbd) (under v0.21.0 - Features), and add yourself to the list of contributors? :)

Kayrnt:
Sure, I updated the changelog file

jtcohen6 (approved):
Thanks for the contribution, @Kayrnt!

Kayrnt:
Great! I'm looking forward to use it 👍

PR #3716, opened 2021-08-10
Use is_relational check for schema caching

### LINKED ISSUE #3715: Avoid caching schemas for tests when store_failures is not enabled
In dbt v0.20.0, tests are now configured by default to use a schema named `dbt_test__audit`, which is useful for the "store failures" feature. On some databases, this is causing dbt to run an extra metadata query when `--store-failures` is _not_ enabled.

On Snowflake, because of how we list objects to populate the relational cache, such a metadata query often appears in the query history as a failed statement (`show terse objects in database.dbt_test__audit`), leading to user confusion.

https://github.com/dbt-labs/dbt/blob/43642956a21571e4589d3f7b9a7ad69439c6d4cc/plugins/snowflake/dbt/adapters/snowflake/impl.py#L135-L141

I view this as a low-priority cosmetic issue, with a simple fix: We could change the logic in `_get_cache_schemas` to check for _relational_ nodes (ones that map to database objects) instead of _executable_ nodes (ones dbt knows how to compile/execute):

https://github.com/dbt-labs/dbt/blob/43642956a21571e4589d3f7b9a7ad69439c6d4cc/core/dbt/adapters/base/impl.py#L304-L316

```python
            for node in manifest.nodes.values()
            if (
                node.is_relational and not node.is_ephemeral_model
            )
```

The `is_relational` method already includes logic around whether to consider tests relational:
https://github.com/dbt-labs/dbt/blob/43642956a21571e4589d3f7b9a7ad69439c6d4cc/core/dbt/contracts/graph/parsed.py#L127-L133

And this is the same method we now use inside runnable tasks, to determine which schemas need creating:
https://github.com/dbt-labs/dbt/blob/43642956a21571e4589d3f7b9a7ad69439c6d4cc/core/dbt/task/runnable.py#L460-L462

This would remove `Analysis`, `Operation`, `Documentation`, and `RPCCall` from the schema-caching list, but frankly it doesn't make much sense that they were there in the first place!

jtcohen6:
resolves #3715 (check out that issue for a full description)

### Checklist
 - [x] This PR includes tests, or tests are not required/relevant for this PR — good enough that existing tests pass? ✅ 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@
 ### Features
 - Capture changes to macros in `state:modified`. Introduce new `state:` sub-selectors: `modified.body`, `modified.configs`, `modified.persisted_descriptions`, `modified.relation`, `modified.macros` ([#2704](https://github.com/dbt-labs/dbt/issues/2704), [#3278](https://github.com/dbt-labs/dbt/issues/3278), [#3559](https://github.com/dbt-labs/dbt/issues/3559))
 
+### Fixes
+- Fix for RPC requests that raise a RecursionError when serializing Undefined values as JSON ([#3464](https://github.com/dbt-labs/dbt/issues/3464), [#3687](https://github.com/dbt-labs/dbt/pull/3687))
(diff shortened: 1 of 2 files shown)


kwigley (approved):

gshank (approved):
Looks good.

PR #3724, opened 2021-08-11
Combine integration test workflows

kwigley:
- Combines the integration test workflows from #3669
- splits up unit tests and code linting so it is clear what each workflow is responsible for

This PR builds on top of #3669. In #3669 there are two workflows for integration tests, one for PRs and one for pushes to release branches (so after merging PRs). I don't think there is any reason to split these up, the thing to keep in mind is that when the workflow runs for PRs using the `pull_request_target` event, the workflow will run in the context of the "target" branch of the PR. It will also run for PRs from forked repositories and allow access to secrets configured for the repository. To ensure that PRs are from trusted contributors, the integration test workflow will check for a label on the PR that is applied by someone with write access to the repository. I combined the necessary checks into one workflow for simplicity and to combat code drift.

The other change in this PR is splitting up unit and linting into separate workflow. No functional changes, purely cosmetic. I like this because you can name exactly what the workflow's goal is and the workflow "matrix" is less complex.


.github/workflows/build.yml
```diff
@@ -40,16 +56,24 @@ jobs:
...
         run: ls -lh dist/
 
+      - name: Check distribution descriptions
+        run: |
+          twine check dist/*
+
+      - name: Check wheel contents
+        run: |
+          check-wheel-contents dist/*.whl --ignore W007,W008
+
```
kwigley, line 76: moved this up, since this only needs to be checked once, not for different OSs

kwigley: also, if these checks fail then it makes sense to fail early and not try to verify installation at all

.github/workflows/integration.yml
```diff
@@ -15,33 +27,96 @@ on:
...
+      contains(github.event.pull_request.labels.*.name, 'ok to test')
     runs-on: ${{ matrix.os }}
+    needs: pre-checks
     strategy:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
         adapter: [postgres, snowflake, redshift, bigquery]
-        python-version: [3.6, 3.7, 3.8, 3.9]
+        python-version: [3.6, 3.7, 3.8, 3.9] # TODO: only run python 3.8 on PRs
```
kwigley, line 100: this TODO is a nice-to-have IMO

.github/workflows/integration.yml
```diff
@@ -15,33 +27,96 @@ on:
...
 
     env:
       TOXENV: integration-${{ matrix.adapter }}
       PYTEST_ADDOPTS: "-v --color=yes -nauto"
 
     steps:
       - name: Check out the repository
```
kwigley, line 107:
```suggestion
      - name: Check out the repository (non-PR)
```
??

nathaniel-may: maybe a name that describes what it is rather than what it isn't?

.github/workflows/integration.yml
```diff
@@ -15,33 +27,96 @@ on:
...
+              - 'core/**'
+              - 'plugins/redshift/**'
+              - 'plugins/postgres/**'
+
   test:
     name: ${{ matrix.adapter }} ${{ matrix.python-version }} / ${{ matrix.os }}
+    # run if not a pull request or has a label to mark as safe to test
+    if: >-
+      github.event_name != 'pull_request_target' ||
+      contains(github.event.pull_request.labels.*.name, 'ok to test')
```
nathaniel-may, line 92: do we want this check to be higher up in the workflow?

.github/workflows/integration.yml
```diff
@@ -76,11 +151,29 @@ jobs:
...
+        # run if postgres integration tests
+        # and is not a PR or has relevant changes/specified label
+        if: >-
+          matrix.adapter == 'postgres' &&
+          (
+            github.event_name != 'pull_request_target' ||
+            needs.pre-checks.outputs.has-postgres-changes ||
+            contains(github.event.pull_request.labels.*.name, 'test all') ||
+            contains(github.event.pull_request.labels.*.name, 'test postgres')
+          )
```
nathaniel-may, line 163: My instinct is to pull this out into something reusable but I don't think there's a good option for that is there?

nathaniel-may (approved):
These changes seem like a solid improvement from your previous iteration. Nice work.

The way your description is worded it looks like this could have been two smaller PRs into the feature which might have been easier to go through. Just something to keep in mind for next time.

.github/workflows/unit.yml
```diff
@@ -17,32 +22,30 @@ on:
       - "*.latest"
       - "releases/*"
   pull_request:
+    branches-ignore:
```
leahwicz, line 25: Do we have unit tests running in another workflow for PRs to `develop`?

kwigley: oh wow, I misunderstood what branches-ignore what doing. This is the only workflow running unit tests! removing this branches-ignore code from all workflows

.github/workflows/integration.yml
kwigley: I think this is fine, this will skip the job not the workflow which is intentional. If this job is skipped, I'd still like the last job to run and post a comment on the PR that it needs someone from Core to mark as safe to test.

.github/workflows/integration.yml
kwigley: I think we are limited by what yaml provides here :( I can't think of a way to reuse this.

PR #3729, opened 2021-08-11
Retry GitHub download failures

leahwicz:
resolves https://github.com/dbt-labs/dbt/issues/3546

### Description

We are seeing `dbt deps` fail intermittently when trying to reach out to GitHub. We do retry logic for when we reach out to the registry hub so we want to try the same retry logic for GitHub to see if it improves the success rate of `dbt deps`.

### Checklist

### DIFF
--- a/core/dbt/clients/registry.py
+++ b/core/dbt/clients/registry.py
@@ -1,5 +1,5 @@
 import requests
-from dbt.utils import memoized, _exception_retry as exception_retry
+from dbt.utils import memoized, _connection_exception_retry as connection_exception_retry
 from dbt.logger import GLOBAL_LOGGER as logger
 import os
 
@@ -16,7 +16,10 @@ def _get_url(url, registry_base_url=None):
     return '{}{}'.format(registry_base_url, url)
 
 
-@exception_retry
+def _get_with_retries(path, registry_base_url=None):
+    return connection_exception_retry(lambda: _get(path, registry_base_url), 5)
+
+
 def _get(path, registry_base_url=None):
     url = _get_url(path, registry_base_url)
     logger.debug('Making package registry request: GET {}'.format(url))
@@ -28,22 +31,22 @@ def _get(path, registry_base_url=None):
 
 
 def index(registry_base_url=None):
-    return _get('api/v1/index.json', registry_base_url)
+    return _get_with_retries('api/v1/index.json', registry_base_url)
 
 
 index_cached = memoized(index)
 
 
 def packages(registry_base_url=None):
-    return _get('api/v1/packages.json', registry_base_url)
+    return _get_with_retries('api/v1/packages.json', registry_base_url)
 
 
 def package(name, registry_base_url=None):
-    return _get('api/v1/{}.json'.format(name), registry_base_url)
+    return _get_with_retries('api/v1/{}.json'.format(name), registry_base_url)
 
 
 def package_version(name, version, registry_base_url=None):
-    return _get('api/v1/{}/{}.json'.format(name, version), registry_base_url)
+    return _get_with_retries('api/v1/{}/{}.json'.format(name, version), registry_base_url)
 
 
 def get_available_versions(name):
--- a/core/dbt/clients/registry.py
+++ b/core/dbt/clients/registry.py
@@ -1,10 +1,7 @@
-from functools import wraps
 import requests
-from dbt.exceptions import RegistryException
-from dbt.utils import memoized
+from dbt.utils import memoized, _exception_retry as exception_retry
 from dbt.logger import GLOBAL_LOGGER as logger
 import os
-import time
 
 if os.getenv('DBT_PACKAGE_HUB_URL'):
     DEFAULT_REGISTRY_BASE_URL = os.getenv('DBT_PACKAGE_HUB_URL')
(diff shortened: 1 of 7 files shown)


core/dbt/clients/registry.py
```diff
@@ -19,26 +16,7 @@ def _get_url(url, registry_base_url=None):
...
-        while True:
-            attempt += 1
-            try:
-                return fn(*args, **kwargs)
-            except (requests.exceptions.ConnectionError, requests.exceptions.Timeout) as exc:
-                if attempt < max_attempts:
-                    time.sleep(1)
-                    continue
-                raise RegistryException(
-                    'Unable to connect to registry hub'
```
leahwicz, line 36: TBD: Change this message if this is the right implementation overall

jtcohen6 (commented):
Hooray!! Glad we're reusing + extending the existing decorator. The code looks pretty good to me, but I should not be the arbiter :)

Also interested to hear folks' thoughts about how best to test - perhaps mocking a package with a nonexistent tarball URL? I know [`test_registry_get_request_exception`](https://github.com/dbt-labs/dbt/blob/develop/test/unit/test_registry_get_request_exception.py) added some unneeded baggage to our unit testing.

core/dbt/utils.py
```diff
@@ -599,3 +603,22 @@ def __getitem__(self, name: str) -> Any:
 
     def __contains__(self, name) -> bool:
         return any((name in entry for entry in self._itersource()))
+
+
+def _exception_retry(fn):
+    @functools.wraps(fn)
+    def wrapper(*args, **kwargs):
+        max_attempts = 5
```
nathaniel-may, line 611: I'd rather have this as a function parameter than a hard coded number.

nathaniel-may (commented):
The python checks out. However, on a personal note, I find this wrapping pattern to be particularly opaque.

Here is an example of how I might suggest refactoring this:
```
# no juggling of indexes which is extremely error prone
def _exception_retry(lambda_fn, retries):
    if retries <= 0:
        raise RegistryException('Unable to connect to registry hub')
    else: 
        try:
            return lambda_fn()
        except (requests.exceptions.ConnectionError, requests.exceptions.Timeout):
            time.sleep(1)
            _exception_retry(lambda_fn, retries - 1)

# example call:
# accepting a lambda function as a parameter allows the outer 
# function to run it regardless of what parameters it takes
_exception_retry(lambda: 1+2, 5)
```
(the original exception is lost here, but that can be put back. I didn't write that part because it obfuscated the point I'm trying to make in the review)

In both ways of writing this, I would suggest the following tests:
1. pass a function that always throws these exceptions and check that you get the right exception back
2. pass a function that throws twice then returns correctly and check that you get the correct response

core/dbt/utils.py
```diff
@@ -599,3 +603,22 @@ def __getitem__(self, name: str) -> Any:
...
+        attempt = 0
+        while True:
+            attempt += 1
+            try:
+                return fn(*args, **kwargs)
+            except (requests.exceptions.ConnectionError, requests.exceptions.Timeout) as exc:
+                if attempt < max_attempts:
+                    time.sleep(1)
```
gshank, line 619: Do you think we might want to do some logging? Something like: logger.debug(f"Attempt {attempt} to connect.").

test/unit/test_core_dbt_utils.py
```diff
@@ -0,0 +1,41 @@
...
+
+class testCoreDbtUtils(unittest.TestCase):
+    def test_connection_exception_retry_none(self):
+        Counter._reset()
+        connection_exception_retry(lambda: Counter._add(), 5)
+        self.assertEqual(1, counter)
+
+    def test_connection_exception_retry_max(self):
+        Counter._reset()
+        self.assertRaises(ConnectionException, connection_exception_retry(lambda: Counter._add_with_exception(), 5))
```
leahwicz, line 16: This `assertRaises` doesn't work for me where it doesn't catch the exception (its the and the `ConnectionException` that is listed) and the test fails because an exception is hit. Can anyone see why that would be the case instead of it just succeeding that it was raised?

test/unit/test_core_dbt_utils.py
```diff
@@ -0,0 +1,41 @@
+import requests
+import unittest
+
+from dbt.exceptions import ConnectionException
+from dbt.utils import _connection_exception_retry as connection_exception_retry
+
+
+class testCoreDbtUtils(unittest.TestCase):
```
kwigley, line 8: nit 😄 
```suggestion
class TestCoreDbtUtils(unittest.TestCase):
```

core/dbt/clients/registry.py
```diff
@@ -19,26 +16,10 @@ def _get_url(url, registry_base_url=None):
...
-                ) from exc
-    return wrapper
-
-
-@_wrap_exceptions
+def _get_with_retries(path, registry_base_url=None):
+    return connection_exception_retry(lambda: _get(path, registry_base_url), 5)
```
kwigley, line 20: Not a required change, but I _think_ this might be a good place to use `functools.partial`!(https://docs.python.org/3/library/functools.html#functools.partial)

```suggestion
    get_fn = partial(_get, path, registry_base_url)
    return connection_exception_retry(get_fn, 5)
```

core/dbt/clients/system.py
```diff
@@ -441,6 +440,12 @@ def run_cmd(
     return out, err
 
 
+def download_with_retries(
+    url: str, path: str, timeout: Optional[Union[float, tuple]] = None
+) -> None:
+    connection_exception_retry(lambda: download(url, path, timeout), 5)
```
kwigley, line 446: same comment about `partial` here as well

kwigley (changes requested):

kwigley (approved):

PR #3732, opened 2021-08-11
Swap dummy perf testing projects for a real one

nathaniel-may:
### Description

To test [PR#6302](https://github.com/dbt-labs/dbt/pull/3602) I used dummy projects which are currently merged. This PR ONLY swaps out those target projects. There are no code changes here. There is only one target project now and it has 2000 model files.

As a best practice, performance testing targets are generally checked in instead of generated. Even though this example is from a generated project, to show without a doubt that the generation does not change between runs, we simply check in the generated files. In the future we could consider separating out our performance work from the main repository, however as of this change, these files only add an additional ~6k to our codebase which will have a negligible effect on our contributors.

### Checklist


leahwicz:
This failed integration test is the exact thing we should look at when we go to fix the tests this cycle 😢

leahwicz (approved):

PR #3749, opened 2021-08-13
Basic local FS functionality, pt 2

iknox-fa:
PR for second segment of local FS storage adapter.
Contains:

- Cleaned up `dbt.clients.system` module
- More/better integration with the rest of the codebase-- 100% of `system` module's file system functions have been swapped with the SA ones (excepting `find_matching`)
- Re-worked `write` method params to match previous system
- Removed many `os.path` refs replacing with `Pathlib` 

This breaks a few tests (unit and integration). The unit test failures are utilizing MagicMocks which don't seem to play nice as file paths for some reason. The integration test failures are mostly from the cli test suite and need further investigation.


iknox-fa:
re-opening

PR #3758, opened 2021-08-16
[#3757] Produce better information about partial parsing version mismatches

### LINKED ISSUE #3757: Unexpected occurrences of partial parsing version mismatch
From dbt community slack:

JC, I'm now using the 0.20.1 actual release and I'm still seeing occasions of Unable to do partial parsing because of a dbt version mismatch on runs.  It's not happening on every run, but it's definitely happening with regularity.
I'm having trouble generating a reproducible case (which is why I haven't opened an issue), but it's happening (with this project anyway) at least 1 in 3 runs.  I can't see anything that seems relevant in the logs, but I wanted to call it out in case (a) you hear of this elsewhere and (b) there's anything I can do to give you more context to understand why this might be happening.

Two consecutive commands were shared, both reporting the same version, with the second one claiming a version mismatch.

gshank:
resolves #3757

### Description

Gathers more information in order to debug spurious version mismatches which prevent partial parsing.

### Checklist

### DIFF
--- a/core/dbt/parser/manifest.py
+++ b/core/dbt/parser/manifest.py
@@ -478,6 +478,12 @@ class ManifestLoader:
         path = os.path.join(self.root_project.target_path,
                             PARTIAL_PARSE_FILE_NAME)
         try:
+            # This shouldn't be necessary, but we have gotten bug reports (#3757) of the
+            # saved manifest not matching the code version.
+            if self.manifest.metadata.dbt_version != __version__:
+                logger.debug("Manifest metadata did not contain correct version. "
+                             f"Contained '{self.manifest.metadata.dbt_version}' instead.")
+                self.manifest.metadata.dbt_version = __version__
             manifest_msgpack = self.manifest.to_msgpack()
             make_directory(os.path.dirname(path))
             with open(path, 'wb') as fp:
@@ -493,7 +499,10 @@ class ManifestLoader:
         reparse_reason = None
 
         if manifest.metadata.dbt_version != __version__:
-            logger.info("Unable to do partial parsing because of a dbt version mismatch")
+            # #3757 log both versions because of reports of invalid cases of mismatch.
+            logger.info("Unable to do partial parsing because of a dbt version mismatch. "
+                        f"Saved manifest version: {manifest.metadata.dbt_version}. "
+                        f"Current version: {__version__}.")
             # If the version is wrong, the other checks might not work
             return False, ReparseReason.version_mismatch
         if self.manifest.state_check.vars_hash != manifest.state_check.vars_hash:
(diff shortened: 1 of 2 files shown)


core/dbt/parser/manifest.py
```diff
@@ -493,7 +499,10 @@ def is_partial_parsable(self, manifest: Manifest) -> Tuple[bool, Optional[str]]:
         reparse_reason = None
 
         if manifest.metadata.dbt_version != __version__:
-            logger.info("Unable to do partial parsing because of a dbt version mismatch")
+            # #3757 log both versions because of reports of invalid cases of mismatch.
+            logger.info("Unable to do partial parsing because of a dbt version mismatch. "
+                        "Saved manifest version: {manifest.metadata.dbt_version}. "
+                        "Current version: {__version__}.")
```
leahwicz, line 505: Does this work in Python without the `f` for inserting variables? I honestly don't know so asking for myself

leahwicz (approved):

core/dbt/parser/manifest.py
```diff
@@ -478,6 +478,12 @@ def write_manifest_for_partial_parse(self):
         path = os.path.join(self.root_project.target_path,
                             PARTIAL_PARSE_FILE_NAME)
         try:
+            # This shouldn't be necessary, but we have gotten bug reports (#3757) of the
+            # saved manifest not matching the code version.
+            if self.manifest.metadata.dbt_version != __version__:
+                logger.debug("Manifest metadata did not contain correct version. "
+                             "Contained '{self.manifest.metadata.dbt_version}' instead.")
```
kwigley, line 485:
```suggestion
                             f"Contained '{self.manifest.metadata.dbt_version}' instead.")
```

core/dbt/parser/manifest.py
kwigley: Ya, good call!

core/dbt/parser/manifest.py
```diff
@@ -478,6 +478,12 @@ def write_manifest_for_partial_parse(self):
...
+                self.manifest.dbt_version = __version__
```
kwigley, line 486: Should this be
```suggestion
                self.manifest.metadata.dbt_version = __version__
```

kwigley (dismissed):

nathaniel-may (approved):

core/dbt/parser/manifest.py
gshank: Thanks!

core/dbt/parser/manifest.py
gshank: Arggh... Hard to test, but should have tried harder :)

gshank:
Tested with 2 different version. Message: Unable to do partial parsing because of a dbt version mismatch. Saved manifest version: 0.20.1. Current version: 0.21.0b1.

PR #3760, opened 2021-08-16
Basic local FS functionality, pt 2

iknox-fa:
PR for second segment of local FS storage adapter.
Contains:

- Cleaned up `dbt.clients.system` module
- More/better integration with the rest of the codebase-- 100% of `system` module's file system functions have been swapped with the SA ones (excepting `find_matching`)
- Re-worked `write` method params to match previous system
- Removed many `os.path` refs replacing with `Pathlib` 

This breaks a few tests (unit and integration). The unit test failures are utilizing MagicMocks which don't seem to play nice as file paths for some reason. The integration test failures are mostly from the cli test suite and need further investigation.

### DIFF
--- a/core/dbt/adapters/internal_storage/local_filesystem.py
+++ b/core/dbt/adapters/internal_storage/local_filesystem.py
@@ -2,7 +2,7 @@ from pathlib import Path
 from shutil import rmtree
 from stat import S_IRUSR, S_IWUSR
 from sys import platform
-from typing import Any, Union
+from typing import Union
 from typing_extensions import Literal
 
 from dbt.logger import GLOBAL_LOGGER as logger
--- a/core/dbt/adapters/internal_storage/local_filesystem.py
+++ b/core/dbt/adapters/internal_storage/local_filesystem.py
@@ -64,7 +64,7 @@ def read(
 def write(
     path: str,
     content: Union[str, bytes, None],
-    overwrite: bool = False,
+    overwrite: bool = True,
 ) -> bool:
     """Writes the given content out to a resource on the filesystem.
 
@@ -90,8 +90,6 @@ def write(
         `True` for success, `False` otherwise.
 
     """
-    # TODO: double check I hit all possible permutations here! (IK)
-
     # create a concrete path object
     path: Path = Path(path)
 
@@ -152,6 +150,9 @@ def delete(path: str) -> bool:
     Args:
         path: Full path of resource to be deleted
 
+    Returns:
+        `True` for success, `False` otherwise.
+
     """
     # create concrete path object
     path: Path = Path(path)
@@ -177,10 +178,6 @@ def delete(path: str) -> bool:
             rmtree(path)
 
     return True
-
-
-def find():
-    pass
 
 
 def info(path: str) -> Union[dict, Literal[False]]:
--- a/core/dbt/adapters/internal_storage/local_filesystem.py
+++ b/core/dbt/adapters/internal_storage/local_filesystem.py
@@ -42,7 +42,7 @@ def read(
     """Reads the content of a file on the filesystem.
 
     Args:
-        path: Full path of file to be read.
+        path: Path of resource to be read (full path or relative to cwd).
         strip: Wether or not to strip whitespace.
 
     Returns:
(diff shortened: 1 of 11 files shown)


core/dbt/adapters/internal_storage/local_filesystem.py
```diff
@@ -64,7 +64,7 @@ def read(
 def write(
     path: str,
     content: Union[str, bytes, None],
-    overwrite: bool = False,
+    overwrite: bool = True,
```
leahwicz, line 67: Maybe add a comment as to why we chose to default to true for the future

core/dbt/logger.py
```diff
@@ -360,8 +361,8 @@ def process(self, record):
 
 
 def make_log_dir_if_missing(log_dir):
-    import dbt.clients.system
-    dbt.clients.system.make_directory(log_dir)
+    # N.B:  Storage adapters can't be used in the logger (circular imports)
```
leahwicz, line 364: Oh this is an interesting point. I guess we can always write logs locally and then if they need to be stored somewhere else for cloud, we can move them

nathaniel-may: might want to add "syncing logs" as something to think about when we work on structured logging.

nathaniel-may: Alternatively, we should consider the storage adapter taking over logging. It might be a bit involved, but changing all the log calls to be SA log calls would really only involve finagling import statements. I don't particularly like logs going somewhere else outside the configured SA.

nathaniel-may (commented):
This looks pretty good! I like all the code deletion and simplification of file system interactions.

One thing I would consider is having a custom exception hierarchy for the storage adapter and passing or raising structured exceptions so they can more easily be handled, and displayed to users in a meaningful way. If we're going to separate system IO into its own module, we should guard the boundary from leaking implementation details via exceptions too.

core/dbt/adapters/internal_storage/local_filesystem.py
gshank: Do we actually have any places where we don't want to overwrite the existing file?

core/dbt/logger.py
gshank: I agree that storage adapters need to be used for logging (somehow) for the long term. Probably not necessary for this stage though.  Another issue is that I'm pretty sure that we will need to put different files in different places eventually. For example the logs, json artifacts, partial parsing file, compiled SQL, all have different use cases that will probably require different storage destinations.

core/dbt/deps/local.py
```diff
@@ -45,25 +45,12 @@ def _fetch_metadata(self, project, renderer):
...
-            system.make_symlink(src_path, dest_path)
-
-        else:
-            logger.debug('  Symlinks are not available on this '
-                         'OS, copying dependency.')
-            shutil.copytree(src_path, dest_path)
+        local_SA.delete(dest_path)
+        src_path.rename(dest_path)
+        # TODO: is it ok to remove symlinking?
+        # Symlinks aren't really a thing outside of filesystems and will be hard to model in SAs
```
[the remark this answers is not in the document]
gshank: Nate: good point. I'm pretty sure that we will want to have symlinks for local execution for performance reasons if nothing else. For server side... Hmm. I'm not sure we would be copying external packages anywhere. Load from git repos, possibly?

core/dbt/logger.py
nathaniel-may: Having that all centralized will help manage all the different destinations. If we don't merge logging into SAs we should definitely make a ticket to track it for when we're ready to.

iknox-fa: +1 to all of this.. right now we have a chicken/egg problem in that we use the logger in the SA to log requests so we can't import the logger which is importing the SA, which is importing the logger... etc...

We can certainly re-work how that goes, but not for this effort.

core/dbt/adapters/internal_storage/local_filesystem.py
iknox-fa: @leahwicz TBH, I didn't choose the default-- this function reflects the way the code was written before SAs.  The change here just reflects that I didn't get it right the first time (which @gerda pointed out)

iknox-fa:
> One thing I would consider is having a custom exception hierarchy for the storage adapter and passing or raising structured exceptions so they can more easily be handled, and displayed to users in a meaningful way. If we're going to separate system IO into its own module, we should guard the boundary from leaking implementation details via exceptions too.

I totally agree.  I think it's best to tackle after we finish whatever updates we're going to end up making vis-a-vis structured logging first, just so we don't have to re-write a bunch of it next cycle.  I'll make sure this ends up in a follow-on ticket or otherwise documented as a thing to try,

PR #3775, opened 2021-08-19
Rebase from `develop`

leahwicz:
Pulling in all changes from `develop` for the next beta release


nathaniel-may (approved):
Took a look at all the commit messages. I don't see anything that shouldn't be in this beta.

PR #3780, opened 2021-08-20
Slim down integration test matrix

kwigley:
#3688 continued

### Description

Last PR I promise. This PR adds the ability to dynamically generate a job matrix based on how the workflow was triggered (PR, push, schedule). See `.github/scripts/integration-test-matrix.js` for how I generated the matrix. This cut down the number of test runs from 48 to 24, and even fewer for PRs depending on what labels are applied and what files are changed ⚡

Additionally:
- stores logs generated from tests
- ensures `PYTEST_ADDOPTS` is passed as an env var from the gha environment -> tox -> pytest, this allows us to control num of processed pytest runs with to parallelize tests and other flags useful for CI.
- organization tweaks, I took some common practices from other python projects as well as other repos in the dbt-labs org

### Checklist


kwigley:
I'll probably end up slimming down all other workflow triggers, not just the pull trigger. I think testing all python versions on linux and all other operating systems on one version of python is a good enough test coverage. I don't anticipate issues coming up in a specific version of python and a specific operating system that isn't covered by this matrix.

kwigley:
Example for now python code changes: https://github.com/dbt-labs/dbt/runs/3385389694
Example for only postgres changes: https://github.com/dbt-labs/dbt/runs/3385398187

.github/scripts/integration-test-matrix.js
```diff
@@ -0,0 +1,90 @@
...
+
+    for (const adapter of supportedAdapters) {
+      if (
+        changes.includes(adapter) ||
+        testAllLabel ||
+        labels.includes(`test ${adapter}`)
+      ) {
+        for (const pythonVersion of supportedPythonVersions) {
+          if (
+            pythonVersion === defaultPythonVersion ||
```
kwigley, line 23: This will only run integration tests for python3.8, unless you apply labels to PR to test other python versions (`test python3.6`). I'm not sure if this is a good default, so I'm happy to change this!

.github/scripts/integration-test-matrix.js
```diff
@@ -0,0 +1,90 @@
...
+  const defaultPythonVersion = "3.8";
+  const supportedPythonVersions = ["3.6", "3.7", "3.8", "3.9"];
+  const supportedAdapters = ["snowflake", "postgres", "bigquery", "redshift"];
+
+  if (context.eventName.includes("pull_request")) {
+    // if PR, generate matrix based on files changed and PR labels
+    const changes = JSON.parse(process.env.CHANGES);
+    const labels = context.payload.pull_request.labels.map(({ name }) => name);
+    console.log("labels", labels);
+    console.log("changes", labels);
```
leahwicz, line 11: logging the wrong thing (changes)

.github/scripts/integration-test-matrix.js
```diff
@@ -0,0 +1,90 @@
...
+    const testAllLabel = labels.includes("test all");
+    const include = [];
+
+    for (const adapter of supportedAdapters) {
+      if (
+        changes.includes(adapter) ||
```
leahwicz, line 17: What format is `changes` in? Is this looking for the adapter name in a list of files, commit messages, etc.?

.github/scripts/integration-test-matrix.js
leahwicz: I'm good with defaulting to just 3.8 for PRs and have all the version run for a schedule run or a commit

.github/scripts/integration-test-matrix.js
```diff
@@ -0,0 +1,90 @@
...
+      }
+    }
+
+    console.log("matrix", { include });
+
+    return {
+      include,
+    };
+  }
+
```
leahwicz, line 60: So this is for non-PRs below here? I would comment on that just so its obvious

.github/workflows/main.yml
```diff
@@ -0,0 +1,223 @@
...
+      fail-fast: false
+      matrix:
+        toxenv: [flake8, mypy]
+
+    env:
+      TOXENV: ${{ matrix.toxenv }}
+      PYTEST_ADDOPTS: "-v --color=yes"
+
+    steps:
+      - name: Check out the repository
```
leahwicz, line 54: Why do you checkout twice?

.github/workflows/main.yml
```diff
@@ -0,0 +1,223 @@
...
+      fail-fast: false
+      matrix:
+        python-version: [3.6, 3.7, 3.8] # TODO: support unit testing for python 3.9 (https://github.com/dbt-labs/dbt/issues/3689)
+
+    env:
+      TOXENV: "unit"
+      PYTEST_ADDOPTS: "-v --color=yes"
+
+    steps:
+      - name: Check out the repository
```
leahwicz, line 98: Same here

.github/scripts/integration-test-matrix.js
kwigley: it is a list of warehouse adapter names, I added a comment above to clarify! I also added some comments in the workflow as well https://github.com/dbt-labs/dbt/blob/ff9c8c8d854ff59be17414b68f472df7d1ed58b2/.github/workflows/integration.yml#L79-L103

.github/scripts/integration-test-matrix.js
kwigley: added a comment! good call

.github/workflows/main.yml
kwigley: rm'd it! that was left in by accident

leahwicz (approved):

PR #3791, opened 2021-08-23
[#3210] Make --models and --select synonyms, except for 'ls'

### LINKED ISSUE #3210: Remove the distinction between --select and --models flags when working with node types

gshank:
resolves #3210

### Description

Some dbt commands used '--select' and some used '--models', but they worked the same under the covers. This makes the commands synonyms, although '--select' is preferred.  The 'ls' command has slightly different behavior for the two commands which is untouched.

### Checklist

### DIFF
--- a/core/dbt/main.py
+++ b/core/dbt/main.py
@@ -615,17 +615,10 @@ def _add_common_selector_arguments(sub):
     )
 
 
-def _add_selection_arguments(*subparsers, **kwargs):
-    models_name = kwargs.get('models_name', 'models')
+def _add_selection_arguments(*subparsers):
     for sub in subparsers:
-        if models_name == 'models':
-            _add_models_argument(sub)
-        elif models_name == 'select':
-            # these still get stored in 'models', so they present the same
-            # interface to the task
-            _add_select_argument(sub)
-        else:
-            raise InternalException(f'Unknown models style {models_name}')
+        _add_models_argument(sub)
+        _add_select_argument(sub)
         _add_common_selector_arguments(sub)
 
 
@@ -1075,8 +1068,8 @@ def parse_args(args, cls=DBTArgumentParser):
                           rpc_sub, seed_sub, parse_sub, build_sub)
     # --models, --exclude
     # list_sub sets up its own arguments.
-    _add_selection_arguments(build_sub, run_sub, compile_sub, generate_sub, test_sub)
-    _add_selection_arguments(snapshot_sub, seed_sub, models_name='select')
+    _add_selection_arguments(
+        build_sub, run_sub, compile_sub, generate_sub, test_sub, snapshot_sub, seed_sub)
     # --defer
     _add_defer_argument(run_sub, test_sub, build_sub)
     # --full-refresh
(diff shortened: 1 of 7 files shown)


leahwicz (approved):
Is the `ls` command just handled as its own separate thing?

core/dbt/main.py
```diff
@@ -615,17 +615,10 @@ def _add_common_selector_arguments(sub):
...
 
-def _add_selection_arguments(*subparsers, **kwargs):
-    models_name = kwargs.get('models_name', 'models')
+def _add_selection_arguments(*subparsers):
     for sub in subparsers:
-        if models_name == 'models':
-            _add_models_argument(sub)
-        elif models_name == 'select':
-            # these still get stored in 'models', so they present the same
-            # interface to the task
```
jtcohen6, line 625: Could we change this, from both storing in `args.models` to both storing in `args.select`? To my mind, this is most relevant for:
- task code (e.g. [compile](https://github.com/dbt-labs/dbt/blob/7fa14b6948d48bfaf521a2a24587e2ec79e1e89a/core/dbt/task/compile.py#L44))
- the `args` dict that gets stored in `run_results.json`, which right now is always called `models` (for both `--models` and `--select`)

Since we're going to start documenting `select` across the board, and only keep `--models` support for backwards compatibility, I'd prefer that the other interfaces reflect the `select` nomenclature. But I'd want to know how big a lift that would be.

jtcohen6 (commented):
Looks good! One comment about switching out `models` for `select` in other parts of the codebase. If it's a bigger lift than I think it is, we can pass.

core/dbt/main.py
gshank: I'll take a look at what changes would be involved.

leahwicz: @gshank @jtcohen6 if this PR is good, lets merge this in and then tackle the possible additional work in a new PR

gshank: Oops. Sorry. Didn't see Leah's comment and the extra work is done.

jtcohen6 (approved):

gshank:
Leah: The 'ls' command uses both '--select' and '--models' with somewhat different behavior, so we're leaving the behavior alone.

PR #3792, opened 2021-08-23
Handle failing tests for `build` command

### LINKED ISSUE #3597: Change the way we handle tests (if parent test fails then don't run non-test children)

kwigley:
resolves #3597 

### Description

This PR add additional edges to the DAG representation of node dependencies for a compiled project. It "just works" with how node selection is designed. The major takeaway here is we want a given node to depend on tests belonging to any upstream nodes. The caveat is that the test can only depend on nodes that are upstream of the given node in order to create this dependency, else this could open up the possibility of circular dependencies (ew, we love DAGs). This ensures that when running tasks that execute multiple different types of resources (`dbt build`), models will depend on all upstream dependencies having passing tests, else it will be skipped during execution.

### Checklist

### DIFF
--- a/core/dbt/compilation.py
+++ b/core/dbt/compilation.py
@@ -395,15 +395,20 @@ class Compiler:
                 if node.resource_type != NodeType.Test:
                     test_dependencies = manifest.get_tests_for_node(dependency)
                     for test_dependency in test_dependencies:
-                        linker.dependency(
-                            node.unique_id,
-                            test_dependency
-                        )
+                        if (
+                           test_dependency not in node.depends_on_nodes and
+                           node.unique_id not in manifest.nodes[test_dependency].depends_on_nodes
+                        ):
+                            linker.dependency(
+                                node.unique_id,
+                                test_dependency
+                            )
 
                 linker.dependency(
                     node.unique_id,
                     (manifest.nodes[dependency].unique_id)
                 )
+
             elif dependency in manifest.sources:
                 linker.dependency(
                     node.unique_id,
@@ -421,6 +426,7 @@ class Compiler:
             self.link_node(linker, exposure, manifest)
 
         print(linker.graph.edges)
+
         cycle = linker.find_cycles()
 
         if cycle:
--- a/core/dbt/compilation.py
+++ b/core/dbt/compilation.py
@@ -390,9 +390,16 @@ class Compiler:
         self, linker: Linker, node: GraphMemberNode, manifest: Manifest
     ):
         linker.add_node(node.unique_id)
-
         for dependency in node.depends_on_nodes:
             if dependency in manifest.nodes:
+                if node.resource_type != NodeType.Test:
+                    test_dependencies = manifest.get_tests_for_node(dependency)
+                    for test_dependency in test_dependencies:
+                        linker.dependency(
+                            node.unique_id,
+                            test_dependency
+                        )
+
                 linker.dependency(
                     node.unique_id,
                     (manifest.nodes[dependency].unique_id)
@@ -412,8 +419,8 @@ class Compiler:
             self.link_node(linker, node, manifest)
         for exposure in manifest.exposures.values():
             self.link_node(linker, exposure, manifest)
-            # linker.add_node(exposure.unique_id)
 
+        print(linker.graph.edges)
         cycle = linker.find_cycles()
 
         if cycle:
--- a/core/dbt/compilation.py
+++ b/core/dbt/compilation.py
@@ -158,13 +158,23 @@ class Linker:
                 resolved_graph.add_edge(predecessor, node)
                 predecessor_tests = manifest.get_tests_for_node(predecessor)
                 for test in predecessor_tests:
-                    test_predecessors = set([
-                        n for n in nx.traversal.bfs_tree(self.graph, test, reverse=True)
-                        if n != test
-                    ])
+                    # test_predecessors = set([
+                    #     n for n in nx.traversal.bfs_tree(self.graph, test, reverse=True)
+                    #     if n != test
+                    # ])
+                    test_depends_on = set(manifest.nodes[test].depends_on_nodes)
+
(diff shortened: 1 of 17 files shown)


kwigley:
I've put this through a lot of testing locally, but would love some help making sure this covers all use cases. Please call out any specific use cases we should add some automated testing for as well!

core/dbt/compilation.py
```diff
@@ -142,11 +154,77 @@ def write_graph(self, outfile: str, manifest: Manifest):
...
+                    upstream_tests += _get_tests_for_node(
+                        manifest,
+                        upstream_node
+                    )
+
+                for upstream_test in upstream_tests:
+                    # Get the set of all nodes that the test depends on
+                    # including the upstream_node itself. This is necessary
+                    # because tests can depend on multiple nodes (ex:
+                    # relationship tests). Test nodes do not distiquish
```
jtcohen6, line 210:
```suggestion
                    # relationship tests). Test nodes do not distinguish
```

jtcohen6 (commented):
This is awesome! Totally working locally. I confirmed it finds a cycle when it should, too.

I'd definitely like to see a more complex test cases that includes tests with multiple, non-consecutive parents (`relationships` or custom ~data~ ~bespoke~ singular tests), i.e. the edge case that you were most recently working to handle over the last few weeks.

gshank (approved):
This looks really good! And excellent comments too.

I do wonder if we could store information in the nodes so that we don't have to loop through the entire list of manifest nodes, but that's something we can look at in the future.

core/dbt/compilation.py
```diff
@@ -142,11 +154,77 @@ def write_graph(self, outfile: str, manifest: Manifest):
...
+        #   |        \/
+        #  \/      test 2
+        # test1
+        #
+        # Produce the following graph:
+        # model1 --> model2 --> model3
+        #   |         |         /\ /\
+        #   |        \/         |  |
+        #  \/      test2 -------   |
+        # test1 -------------------
```
nathaniel-may, line 180: Love this comment.

test/integration/069_build_test/tests-failing/test.yml
```diff
@@ -0,0 +1,18 @@
+version: 2
+
+models:
+  - name: model_0
+    columns:
+      - name: iso3
+        tests:
+          - unique
+          - not_null
```
nathaniel-may, line 9: do these create the tests with the graph? 

Maybe we should add a readme to this test to describe what's going on.

nathaniel-may (approved):
Very clean. I agree with the previous reviewers that a larger more rigorous test project would be useful here, but otherwise it LGTM.

iknox-fa:
This all looks great @kwigley!  My only concern with this approach (and the reason I didn't use something similar) is that doing a lot of graph re-creation may chew through more CPU than we'd like (especially given networkx doesn't utilize multiple cores IIRC).  Have you A/B tested this on some large  project a-la `internal-analytics`?  Might be worth checking so we know what to expect.

core/dbt/compilation.py
```diff
@@ -142,11 +154,77 @@ def write_graph(self, outfile: str, manifest: Manifest):
         include all nodes in their corresponding graph entries.
         """
         out_graph = self.graph.copy()
-        for node_id in self.graph.nodes():
+        for node_id in self.graph:
             data = manifest.expect(node_id).to_dict(omit_none=True)
             out_graph.add_node(node_id, **data)
         nx.write_gpickle(out_graph, outfile)
 
+    def resolve_graph(self, manifest: Manifest):
```
iknox-fa, line 162: We haven't been doing this consistently (I include myself in this), but the idiomatic way to do this type hint should probably include `-> None`.  Super minor, but mypy will carp about it when (if) we ever crank up the strictness.

kwigley:
> This looks really good! And excellent comments too.
> 
> I do wonder if we could store information in the nodes so that we don't have to loop through the entire list of manifest nodes, but that's something we can look at in the future.

@gshank I agree, there is room for refactor. This leads to my only concern that @iknox-fa noted, perf. I'll get a sense of timing and report any further concerns.

core/dbt/compilation.py
```diff
@@ -412,13 +424,80 @@ def link_graph(self, linker: Linker, manifest: Manifest):
...
+        self.resolve_graph(linker, manifest)
+
+    def resolve_graph(self, linker: Linker, manifest: Manifest) -> None:
+        """ This method adds additional edges to the DAG. For a given non-test
+        executable node, add an edge from an upstream test to the given node if
+        the set of nodes the test depends on is a proper/strict subset of the
+        upstream nodes for the given node. """
+
+        # Given a graph:
+        # model1 --> model2 --> model3
```
leahwicz, line 442: Great visualization!

core/dbt/compilation.py
```diff
@@ -107,6 +107,18 @@ def _extend_prepended_ctes(prepended_ctes, new_prepended_ctes):
...
 
 
+def _get_tests_for_node(manifest: Manifest, unique_id: UniqueID) -> List[UniqueID]:
+    """ Get a list of tests that depend on the node with the
+    provided unique id """
+
+    return [
+        node.unique_id
+        for _, node in manifest.nodes.items()
+        if node.resource_type == NodeType.Test and
```
leahwicz, line 117: I don't understand Python syntax but does the `if` not need to be indented under the `for` loop?

kwigley: Nope, not here 😎 This is just one long line using [list comprehension](https://docs.python.org/3/tutorial/datastructures.html#list-comprehensions) that I broke up into multiple lines! It is almost equivalent to:
```python
result_list = []
for _, node in manifest.nodes.items():
    if node.resource_type == NodeType.Test and unique_id in node.depends_on_nodes:
        result_list.append(node.unique_id)
return result_list
```

PR #3794, opened 2021-08-24
Add test results publishing as artifact

leahwicz:
@kwigley I was tinkering around with adding test results and publishing them as an artifact. Not sure what you think, but it's a first step to keeping track of failures. My thought is to put them in Snowflake and then modeling some dashboards for us. We can chat about this if you don't think adding this is worth it. I'm sure there is a 3rd party tool for this but I want to work `dbt` into our workflow in places where it makes sense.


kwigley (approved):
👍

PR #3825, opened 2021-08-26
Feature: Add support for Package name changes on the Hub

annafil:
Partially resolves https://github.com/dbt-labs/hub.getdbt.com/issues/810

### Description

This PR adds a warning message in the user logs when running dbt deps when the process encounters "redirectnamespace" or "redirectname" in the package index.json. 

#### Implementation details
- There is a bunch of nested if logic that is needed because those variables can exist in a variety of combinations: "redirectnamespace" can point to a new GitHub org name, "redirectname" can point to a new repo name, or both at the same time. This code should also handle the condition where these variables are present, but null. 
- Formatting and more Pythonic suggestions very welcome :) 

#### Current undesirable behavior

- the part of the codebase that has access to the package index.json variables seems to get executed twice, resulting in the following in a user's info log: 
Screen Shot 2021-08-26 at 10 30 31 AM

Anyone know why this is happening? 🙏 😬 I spent the bulk of my time on this PR following breadcrumbs to figure this out, and traced it to this line: https://github.com/dbt-labs/dbt/blob/09ea989d81e5272d9fe9e5b5b30b1a441e3da349/core/dbt/task/deps.py#L53

#### What this doesn't do (yet)

Notably, this PR doesn't actually implement any redirects on behalf of the user, it merely tells folks to change their package.yml configuration. A few things that still need to be done, but are out of scope for this PR: 
- throw a more useful error when someone encounters a mixture of dbt-labs/dbt_utils and fishtown-analytics/dbt_utils in their package.yml (e.g. they have moved over to the dbt-labs version of our packages, but another package maintainer who depends on dbt_utils didn't). Right now we just tell folks dbt_utils has a conflict with a dependency, and no other useful information. 
- (assuming we want this behavior) let Core automatically resolve any redirects in a way that is seamless for the user, knowing that it's unlikely we will ever get everyone's projects and packages moved over due to a name change. 

### Checklist

### DIFF
--- a/core/dbt/deprecations.py
+++ b/core/dbt/deprecations.py
@@ -131,6 +131,14 @@ class AdapterMacroDeprecation(DBTDeprecation):
     '''
 
 
+class PackageRedirectDeprecation(DBTDeprecation):
+    _name = 'package-redirect'
+    _description = '''\
+    The `{old_name}` package is deprecatd in favor of `{new_name}`. Please update
+    your `packages.yml` configuration to use `{new_name}` instead.
+    '''
+
+
 _adapter_renamed_description = """\
 The adapter function `adapter.{old_name}` is deprecated and will be removed in
 a future release of dbt. Please use `adapter.{new_name}` instead.
@@ -176,6 +184,7 @@ deprecations_list: List[DBTDeprecation] = [
     ModelsKeyNonModelDeprecation(),
     ExecuteMacrosReleaseDeprecation(),
     AdapterMacroDeprecation(),
+    PackageRedirectDeprecation()
 ]
 
 deprecations: Dict[str, DBTDeprecation] = {
--- a/core/dbt/deprecations.py
+++ b/core/dbt/deprecations.py
@@ -134,7 +134,7 @@ class AdapterMacroDeprecation(DBTDeprecation):
 class PackageRedirectDeprecation(DBTDeprecation):
     _name = 'package-redirect'
     _description = '''\
-    The `{old_name}` package is deprecatd in favor of `{new_name}`. Please update
+    The `{old_name}` package is deprecated in favor of `{new_name}`. Please update
     your `packages.yml` configuration to use `{new_name}` instead.
     '''
 
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -40,7 +40,7 @@ permissions: read-all
 
 # will cancel previous workflows triggered by the same event and for the same ref
 concurrency:
-  group: ${{ github.workflow }}-${{ github.event_name }}-${{ github.ref }}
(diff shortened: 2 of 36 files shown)


core/dbt/deprecations.py
```diff
@@ -176,6 +184,7 @@ def warn(name, *args, **kwargs):
     ModelsKeyNonModelDeprecation(),
     ExecuteMacrosReleaseDeprecation(),
     AdapterMacroDeprecation(),
+    PackageRedirectDeprecation()
```
jtcohen6, line 187: As far as testing this, we can take a cue from the existing [deprecation tests](https://github.com/dbt-labs/dbt/blob/develop/test/integration/012_deprecation_tests/test_deprecations.py):
- set up the conditions for the deprecation to raise
- assert that it raises as a warning
- assert that the warning becomes an error in [`--warn-error`](https://docs.getdbt.com/reference/global-cli-flags#warnings-as-errors) mode (called `strict` here for historical reasons)
- that the warning/exception message says what you think it should say

So, you can add this or something like it to the bottom of `test_deprecations.py`. The methods prefixed with `test_postgres_` will be picked up during Postgres integration tests.
```python
class TestPackageRedirectDeprecation(BaseTestDeprecations):
    @property
    def models(self):
        return self.dir('where-were-going-we-dont-need-models')

    @property
    def packages_config(self):
        return {
            "packages": [
                {
                    'package': 'fishtown-analytics/dbt_utils',
                    'version': '0.7.0'
                }
            ]
        }
    
    @use_profile('postgres')
    def test_postgres_package_redirect(self):
        self.assertEqual(deprecations.active_deprecations, set())
        self.run_dbt(['deps'], strict=False)
        expected = {'package-redirect'}
        self.assertEqual(expected, deprecations.active_deprecations)

    @use_profile('postgres')
    def test_postgres_package_redirect_fail(self):
        self.assertEqual(deprecations.active_deprecations, set())
        with self.assertRaises(dbt.exceptions.CompilationException) as exc:
            self.run_dbt(['deps'], strict=True)
        exc_str = ' '.join(str(exc.exception).split())  # flatten all whitespace
        expected = "The `fishtown-analytics/dbt_utils` package is deprecated in favor of `dbt-labs/dbt_utils`"
        assert expected in exc_str
```

jtcohen6 (commented):
Looking really good! One typo, and one recommended addition to the deprecations testing suite.

annafil:
OK thanks to @jtcohen6 this now has tests, checks are passing and I merged in the latest state of develop with @sungchun12's changes ✨ @leahwicz Think this is ready for review?

jtcohen6 (approved):

leahwicz (approved):

annafil:
Mergy mergy? 🙏

kwigley (approved):
nice! ❤️ UX improvement

Issue #3735, opened 2021-08-12
dbt debug avoid dependency checking 


leoch20:
### Describe the feature
We deployed dbt in AWS managed Airflow MWAA. Currently we perform some system checks to ensure the deployment is successful. Among them is `dbt debug` . The issue is that **git** is not install in MWAA and there is no reason to install it (please correct me if there is a need to have it). Because git is not install, every time we run the system checks `dbt debug` fails. 

It would be great if we could manually bypass checking the dependencies so that if debug fails we know it is not because of missing git.

jtcohen6:
@leoch20 Thanks for opening! Honestly, I'm of two minds about this one.

On the one hand, you're completely right that it's possible to use dbt _without_ having git installed, assuming you don't need to install any `git`-based packages.

On the other hand, `dbt debug` is really intended as a helper check for CLI users who have installed dbt locally, to help them make sure they're ready to start up with dbt development—not a step in automated orchestration processes. I can appreciate that, if something is going to fail in Airflow (e.g. a faulty database connection), it's nicer to fail with the clear error early on, rather than a slightly more-buried error during the early steps of a run. But it's going to fail either way.

(I could imagine a local user who doesn't want to have `git` installed, but that seems much less likely.)

Having said all that, the codepath here is quite straightforward:
https://github.com/dbt-labs/dbt/blob/d2dbe6afe4ecff25e4fa8ba471716bc8bd0ef760/core/dbt/task/debug.py#L124-L126

I don't think it would be a big lift to add [debug task flags](https://github.com/dbt-labs/dbt/blob/d2dbe6afe4ecff25e4fa8ba471716bc8bd0ef760/core/dbt/task/debug.py#L124-L126) like:
```
dbt debug --skip-configuration
dbt debug --skip-dependencies
dbt debug --skip-connection
```
And check before running each check:
```python
    if not self.args.skip_configuration:
        self.test_configuration()
    if not self.args.skip_dependencies:
        self.test_dependencies()
    if not self.args.skip_connection:
        self.test_connection()
```
So, if this is something you'd like to open a PR for, I'd be happy to review it.

PR #3832, opened 2021-08-30
Add colourful count of pass/fail tests in dbt debug

### LINKED ISSUE #3831: Show a pass/fail count for all steps tested during dbt debug, to avoid issues being pushed below the fold
### Describe the feature
When running `dbt debug` in a short terminal window (e.g. in VS Code on a laptop), issues can be pushed out of the viewport. In this example, `dbt_project.yml` is missing:
image

In fairness, there is an error message there but at a glance it just looks like the cwd prompt. The green `OK connection ok` message above absolutely smokes it in a competition for attention. 

```
$dbt debug
Running with dbt=0.21.0-b1
...
Using dbt_project.yml file at C:\WF\LP\data-insights\dbt-core\dbt_project.yml

Configuration:
  dbt_project.yml file [ERROR not found]

Required dependencies:
 - git [OK found]

Connection:
...
  Connection test: OK connection ok

3/4 checks passed #in red 
Could not load dbt_project.yml
```

Reading the error message that's already on screen. 

### Additional context
This bit me hard the first time I ran debug. It just caught me again but I was wiser this time. 

I'm not sure if it's 4 checks, or whether it's also checking things like Python version and path. Let me know if it's more complex than just the 4 that get an OK/ERROR outcome.

### Who will this benefit?
Most of all, new users on their first run through. 

Yep! I've already had a look around the region and it looks pretty straightforward.

joellabes:
resolves #3831 

### Description
With a failing check:
image

With no failing checks:
image

### Style comment

That red is sorta hard to see. I would have preferred to use yellow, but that would be inconsistent. 
### Checklist

### DIFF
--- a/core/dbt/task/debug.py
+++ b/core/dbt/task/debug.py
@@ -125,6 +125,10 @@ class DebugTask(BaseTask):
         self.test_dependencies()
         self.test_connection()
 
+        if self.any_failure:
+            print(red('{}/4 checks failed'.format(len(self.messages))))
+        else:
+            print(green('All checks passed!'))
         for message in self.messages:
             print(message)
             print('')
@@ -368,7 +372,7 @@ class DebugTask(BaseTask):
         print('Connection:')
         for k, v in self.profile.credentials.connection_info():
             print('  {}: {}'.format(k, v))
-        print('  Connection test: {}'.format(self._connection_result()))
+        print('  Connection test: [{}]'.format(self._connection_result()))
         print('')
 
     @classmethod
--- a/core/dbt/task/debug.py
+++ b/core/dbt/task/debug.py
@@ -130,7 +130,7 @@ class DebugTask(BaseTask):
             print(red(f"{(pluralize(len(self.messages), 'check'))} failed:"))
         else:
             print(green('All checks passed!'))
-        
+
         for message in self.messages:
             print(message)
             print('')
@@ -328,8 +328,8 @@ class DebugTask(BaseTask):
         if self.project_fail_details == FILE_NOT_FOUND:
             return
         msg = (
-            f'Project loading failed for the following reason: ' \
-            f'\n{self.project_fail_details} ' \
+            f'Project loading failed for the following reason:'
+            f'\n{self.project_fail_details}'
             f'\n'
         )
         self.messages.append(msg)
@@ -342,8 +342,8 @@ class DebugTask(BaseTask):
         if self.profile_fail_details == FILE_NOT_FOUND:
             return
         msg = (
-            f'Profile loading failed for the following reason: ' \
-            f'\n{self.profile_fail_details} ' \
+            f'Profile loading failed for the following reason:'
+            f'\n{self.profile_fail_details}'
             f'\n'
         )
         self.messages.append(msg)
--- a/core/dbt/task/debug.py
+++ b/core/dbt/task/debug.py
@@ -16,6 +16,7 @@ from dbt.context.target import generate_target_context
 from dbt.clients.yaml_helper import load_yaml_text
 from dbt.links import ProfileConfigDocs
 from dbt.ui import green, red
+from dbt.utils import pluralize
 from dbt.version import get_installed_version
 
 from dbt.task.base import BaseTask, get_nearest_project_dir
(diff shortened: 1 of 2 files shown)


core/dbt/task/debug.py
```diff
@@ -125,6 +125,10 @@ def run(self):
         self.test_dependencies()
         self.test_connection()
 
+        if self.any_failure:
+            print(red('{}/4 checks failed'.format(len(self.messages))))
```
jtcohen6, line 129: Clever approach, minimally invasive, using the `dbt.ui` color functions—I like it!

One easily fixable problem: `_log_project_fail()` and `_log_profile_fail()` do their own `print`ing directly, rather than appending to the `self.messages` queue. They should append to `messages` so that we can count failures appropriately.

One additional thought: Rather than hard-coding `4` here as the total number of checks, we could keep a running tally that increments with each check we perform. I'm thinking about this in two ways:
- If the specified `profile` is invalid, we won't run a connection test, so really the user will only see 3 checks, 1 of which failed
- The proposal in #3735 would enable a more "targeted" debug that runs fewer checks

jtcohen6: Following both recommendations above, I've pasted below a quick diff to make 'em happen. You don't have to use—or, feel free to use and rename any/all variables—I just wanted to make sure that I was asking for easily doable things.
```
$ git diff core/dbt/task/debug.py
diff --git a/core/dbt/task/debug.py b/core/dbt/task/debug.py
index 0cb1e8960..0e53518ac 100644
--- a/core/dbt/task/debug.py
+++ b/core/dbt/task/debug.py
@@ -91,6 +91,7 @@ class DebugTask(BaseTask):
         self.project_fail_details = ''
         self.any_failure = False
         self.messages: List[str] = []
+        self.check_counter: int = 0

     @property
     def project_profile(self):
@@ -125,13 +126,12 @@ class DebugTask(BaseTask):
         self.test_dependencies()
         self.test_connection()

+        for message in self.messages:
+            print(message)
         if self.any_failure:
-            print(red('{}/4 checks failed'.format(len(self.messages))))
+            print(red('{}/{} checks failed'.format(len(self.messages), self.check_counter)))
         else:
             print(green('All checks passed!'))
-        for message in self.messages:
-            print(message)
-            print('')

         return not self.any_failure

@@ -289,6 +289,7 @@ class DebugTask(BaseTask):
         return green('OK found and valid')

     def test_git(self):
+        self.check_counter += 1
         try:
             dbt.clients.system.run_cmd(os.getcwd(), ['git', '--help'])
         except dbt.exceptions.ExecutableError as exc:
@@ -319,26 +320,34 @@ class DebugTask(BaseTask):
         self._log_profile_fail()

     def _log_project_fail(self):
+        self.check_counter += 1
         if not self.project_fail_details:
             return

         self.any_failure = True
         if self.project_fail_details == FILE_NOT_FOUND:
             return
-        print('Project loading failed for the following reason:')
-        print(self.project_fail_details)
-        print('')
+        msg = (
+            f'Project loading failed for the following reason: ' \
+            f'\n {self.project_fail_details} ' \
+            f'\n'
+        )
+        self.messages.append(msg)

     def _log_profile_fail(self):
+        self.check_counter += 1
         if not self.profile_fail_details:
             return

         self.any_failure = True
         if self.profile_fail_details == FILE_NOT_FOUND:
             return
-        print('Profile loading failed for the following reason:')
-        print(self.profile_fail_details)
-        print('')
+        msg = (
+            f'Profile loading failed for the following reason: ' \
+            f'\n {self.profile_fail_details} ' \
+            f'\n'
+        )
+        self.messages.append(msg)

     @staticmethod
     def attempt_connection(profile):
@@ -359,6 +368,7 @@ class DebugTask(BaseTask):
         return None

     def _connection_result(self):
+        self.check_counter += 1
         result = self.attempt_connection(self.profile)
         if result is not None:
             self.messages.append(result)
```

jtcohen6 (commented):
Nice work! Definitely feels like a step in the right UX direction. I left a few comments + suggestions.

Thanks, as always, for picking up what you're putting down :)

core/dbt/task/debug.py
joellabes: Instead of keeping a running tally, would it be fair to just say `1 check failed` and trade the accurate count problem for a pluralisation problem?

jtcohen6: Yes... that's just simpler overall, and generally better, isn't it?

jtcohen6:
Going to close + re-open just to trigger adapter tests, now that I've applied the `ok to test` label

jtcohen6 (approved):
lgtm! thanks for the contribution :)

Issue #3596, opened 2021-07-20
dbt build: Parity for flags


iknox-fa:
Allow any flags set to be passed in to any of the tasks run in an invocation of `dbt build`.  In the case that a flag might apply to multiple types of tasks it applies to all tasks (it is an opinionated build command, after all)

jtcohen6:
- Any flag that is supported by `run`, `test`, `seed`, `snapshot`
- `--resource-type` (like `list`)
- Ensure that the RPC method has parameters to support those flags as well

PR #3833, opened 2021-08-30
Feature/3723 greedy flag test selection

### LINKED ISSUE #3723: Support a --greedy flag that turns on all indirect selection. Otherwise, log nodes that weren't selected indirectly
### Describe the feature
When I use `dbt test --models some_model` or `dbt test --models +exposure:some_exposure_name` it would be nice if tests like `relationships:` that refer to other models were explicitly skipped, rather than just silently refusing to go.

I now know that in the case of specific models I can just use graph operators to make sure all of the tests do run, so instead of the above I would use `dbt test --models some_model+1`.  Interestingly, however, I have not been able to make this work with exposures in the same way. (I think this is because exposures are already to the right of the DAG, so the + behaves differently or is simply redundant in that case.) 

### Additional context
The danger here is simply in not knowing. It's fine if certain tests don't run for certain reasons. It's less fine that there's zero warning when it happens. To my mind this would be a great implementation of the `Skipped` statistic when testing.

I was caught unaware by the behavior shift in 0.20.0, so I didn't know that my core assumption around testing, namely: _"When I test a model, all tests that I've put against that model will be run"_, no longer held true. I now understand how and (broadly) why the behavior shifted, but I still think there's a valid point to be made here. 

I thought my tests were all passing, because none were failing and none were being skipped. All I saw was green, so I was pressing on with the project. Once I figured out what was happening and ran all of the tests, I found out the data I had been working with was flawed in ways I thought I was actively protecting against. 

### Who will this benefit?
Going forward, it will be beneficial for myself and others to be told that certain tests are being skipped when testing specific models without graph operators, or testing through exposures.

An analyst or engineer who sees that 100% of the tests they're running against a model are passing might think that is brilliant news; everything works. They might later (hopefully not too much later) discover that yes 100% of the running tests passed, but some that were not being run at all would have failed spectacularly.

An analyst or engineer who sees that 100% of the tests they're running against a model are either passing or being skipped will immediately look into why the tests are being skipped, and will not face the same kind of danger.

I don't think I'm the best person to write code around this, but I am happy to discuss it further if there would be value in that.

joellabes:
resolves #3723 

### Description
This restores the old behaviour for magic test expansion on an opt-in basis via a new `dbt test --greedy` flag. It also makes the new behaviour of only testing nodes where all parents are in scope more visible, by logging the tests that were excluded. 

### Conversation topics
- I've taken huge guesses at how this should be implemented, based on `--store-failures`. I don't _really_ understand what I did in the flags zone, so let me know if I'm off base. 
- Should this also apply to `dbt ls`? I think yes - in my head, `ls` is an interactive debugger before applying the selection logic and they should be equivalent. If so, should it only be allowed in conjunction with `--resource-type test`?
- Should this apply to `--selector`? I think not, you'd pass `--greedy` in separately?
- Should there be a way to suppress the warnings if you know what you're doing? I'm not wed to this, but I can imagine them getting tedious, especially if it warns about a big list of nodes. I don't know where you'd put it though - passing a `--i-know-about-greedy-leave-me-alone` flag every time is equally tedious. Maybe an environment variable a la `DBT_PROFILES_DIR`? 
- There aren't any tests yet - please point me in the right direction for that

### Checklist

- [ ] I have updated the documentation site

### DIFF
--- a/core/dbt/graph/cli.py
+++ b/core/dbt/graph/cli.py
@@ -1,4 +1,5 @@
 # special support for CLI argument parsing.
+from dbt import flags
 import itertools
 from dbt.clients.yaml_helper import yaml, Loader, Dumper  # noqa: F401
 
@@ -66,7 +67,7 @@ def parse_union_from_default(
 def parse_difference(
     include: Optional[List[str]], exclude: Optional[List[str]]
 ) -> SelectionDifference:
-    included = parse_union_from_default(include, DEFAULT_INCLUDES)
+    included = parse_union_from_default(include, DEFAULT_INCLUDES, greedy=bool(flags.GREEDY))
     excluded = parse_union_from_default(exclude, DEFAULT_EXCLUDES, greedy=True)
     return SelectionDifference(components=[included, excluded])
 
@@ -180,7 +181,7 @@ def parse_union_definition(definition: Dict[str, Any]) -> SelectionSpec:
     union_def_parts = _get_list_dicts(definition, 'union')
     include, exclude = _parse_include_exclude_subdefs(union_def_parts)
 
-    union = SelectionUnion(components=include)
+    union = SelectionUnion(components=include, greedy_warning=False)
 
     if exclude is None:
         union.raw = definition
@@ -188,7 +189,8 @@ def parse_union_definition(definition: Dict[str, Any]) -> SelectionSpec:
     else:
         return SelectionDifference(
             components=[union, exclude],
-            raw=definition
+            raw=definition,
+            greedy_warning=False
         )
 
 
@@ -197,7 +199,7 @@ def parse_intersection_definition(
 ) -> SelectionSpec:
     intersection_def_parts = _get_list_dicts(definition, 'intersection')
     include, exclude = _parse_include_exclude_subdefs(intersection_def_parts)
-    intersection = SelectionIntersection(components=include)
+    intersection = SelectionIntersection(components=include, greedy_warning=False)
 
     if exclude is None:
         intersection.raw = definition
@@ -205,7 +207,8 @@ def parse_intersection_definition(
     else:
         return SelectionDifference(
             components=[intersection, exclude],
-            raw=definition
+            raw=definition,
+            greedy_warning=False
         )
 
 
@@ -239,7 +242,7 @@ def parse_dict_definition(definition: Dict[str, Any]) -> SelectionSpec:
     if diff_arg is None:
         return base
     else:
-        return SelectionDifference(components=[base, diff_arg])
+        return SelectionDifference(components=[base, diff_arg], greedy_warning=False)
 
 
 def parse_from_definition(
--- a/core/dbt/graph/selector.py
+++ b/core/dbt/graph/selector.py
@@ -1,4 +1,3 @@
-
 from typing import Set, List, Optional, Tuple
 
 from .graph import Graph, UniqueId
@@ -28,6 +27,24 @@ def alert_non_existence(raw_spec, nodes):
             f"The selection criterion '{str(raw_spec)}' does not match"
             f" any nodes"
         )
+
+
+def alert_unused_nodes(raw_spec, node_names):
+    summary_nodes_str = ("\n  - ").join(node_names[:3])
+    debug_nodes_str = ("\n  - ").join(node_names)
+    and_more_str = f"\n  - and {len(node_names) - 3} more" if len(node_names) > 4 else ""
+    summary_msg = (
+        f"\nSome tests were excluded because at least one parent is not selected. "
+        f"Use the --greedy flag to include them."
+        f"\n  - {summary_nodes_str}{and_more_str}"
+    )
+    logger.info(summary_msg)
+    if len(node_names) > 4:
+        debug_msg = (
+            f"Full list of tests that were excluded:"
+            f"\n  - {debug_nodes_str}"
+        )
+        logger.debug(debug_msg)
 
 
 def can_select_indirectly(node):
@@ -151,16 +168,16 @@ class NodeSelector(MethodManager):
 
         return direct_nodes, indirect_nodes
 
-    def select_nodes(self, spec: SelectionSpec) -> Set[UniqueId]:
+    def select_nodes(self, spec: SelectionSpec) -> Tuple[Set[UniqueId], Set[UniqueId]]:
         """Select the nodes in the graph according to the spec.
 
         This is the main point of entry for turning a spec into a set of nodes:
         - Recurse through spec, select by criteria, combine by set operation
         - Return final (unfiltered) selection set
         """
-
         direct_nodes, indirect_nodes = self.select_nodes_recursively(spec)
-        return direct_nodes
+        indirect_only = indirect_nodes.difference(direct_nodes)
+        return direct_nodes, indirect_only
 
     def _is_graph_member(self, unique_id: UniqueId) -> bool:
         if unique_id in self.manifest.sources:
@@ -213,6 +230,8 @@ class NodeSelector(MethodManager):
         #  - If ANY parent is missing, return it separately. We'll keep it around
         #    for later and see if its other parents show up.
         # We use this for INCLUSION.
+        # Users can also opt in to inclusive GREEDY mode by passing --greedy flag,
+        # or by specifying `greedy: true` in a yaml selector
 
         direct_nodes = set(selected)
         indirect_nodes = set()
@@ -251,15 +270,24 @@ class NodeSelector(MethodManager):
 
             - node selection. Based on the include/exclude sets, the set
                 of matched unique IDs is returned
-                - expand the graph at each leaf node, before combination
-                    - selectors might override this. for example, this is where
-                        tests are added
+                - includes direct + indirect selection (for tests)
             - filtering:
                 - selectors can filter the nodes after all of them have been
                   selected
         """
-        selected_nodes = self.select_nodes(spec)
+        selected_nodes, indirect_only = self.select_nodes(spec)
         filtered_nodes = self.filter_selection(selected_nodes)
+
+        if indirect_only:
+            filtered_unused_nodes = self.filter_selection(indirect_only)
+            if filtered_unused_nodes and spec.greedy_warning:
+                # log anything that didn't make the cut
+                unused_node_names = []
+                for unique_id in filtered_unused_nodes:
+                    name = self.manifest.nodes[unique_id].name
+                    unused_node_names.append(name)
+                alert_unused_nodes(spec, unused_node_names)
+
         return filtered_nodes
 
     def get_graph_queue(self, spec: SelectionSpec) -> GraphQueue:
--- a/core/dbt/main.py
+++ b/core/dbt/main.py
@@ -406,6 +406,14 @@ def _build_build_subparser(subparsers, base_subparser):
         Store test results (failing rows) in the database
         '''
     )
+    sub.add_argument(
+        '--greedy',
+        action='store_true',
+        help='''
+        Select all tests that touch the selected resources,
+        even if they also depend on unselected resources
+        '''
+    )
     resource_values: List[str] = [
         str(s) for s in build_task.BuildTask.ALL_RESOURCE_VALUES
     ] + ['all']
@@ -637,7 +645,7 @@ def _add_table_mutability_arguments(*subparsers):
             '--full-refresh',
             action='store_true',
             help='''
-            If specified, DBT will drop incremental models and
+            If specified, dbt will drop incremental models and
             fully-recalculate the incremental table from the model definition.
             '''
         )
@@ -751,6 +759,14 @@ def _build_test_subparser(subparsers, base_subparser):
         action='store_true',
         help='''
         Store test results (failing rows) in the database
+        '''
+    )
+    sub.add_argument(
+        '--greedy',
+        action='store_true',
+        help='''
+        Select all tests that touch the selected resources,
+        even if they also depend on unselected resources
         '''
     )
 
@@ -877,6 +893,14 @@ def _build_list_subparser(subparsers, base_subparser):
         ''',
         metavar='SELECTOR',
         required=False,
+    )
+    sub.add_argument(
+        '--greedy',
+        action='store_true',
+        help='''
+        Select all tests that touch the selected resources,
+        even if they also depend on unselected resources
+        '''
     )
     _add_common_selector_arguments(sub)
 
--- a/test/integration/066_test_selection_tests/test_selection_expansion.py
+++ b/test/integration/066_test_selection_tests/test_selection_expansion.py
@@ -1,4 +1,5 @@
 from test.integration.base import DBTIntegrationTest, FakeArgs, use_profile
+import yaml
 
 from dbt.task.test import TestTask
 from dbt.task.list import ListTask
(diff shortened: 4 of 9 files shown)


core/dbt/graph/selector.py
```diff
@@ -258,8 +258,21 @@ def get_selected(self, spec: SelectionSpec) -> Set[UniqueId]:
...
+        if indirect_only:
+            filtered_unused_nodes = self.filter_selection(indirect_only)
+            # log anything that didn't make the cut
+            if filtered_unused_nodes:
+                unused_node_names = []
+                for unique_id in filtered_unused_nodes:
+                    name = self.manifest.nodes[unique_id].name
+                    unused_node_names.append(name)
+                unused_nodes_str = ("\n  - ").join(unused_node_names)
+                msg = f"\nSome resources were excluded because at least one parent is missing:\n  - {unused_nodes_str}\nUse the --greedy flag to include them"
```
joellabes, line 273: This is the opposite to the phrasing you had in the sample code you wrote. I switched it because it was confusing to say "these were selected but they're not going to run", but I'm not super happy with how it reads now either.

core/dbt/main.py
```diff
@@ -727,6 +727,13 @@ def _build_test_subparser(subparsers, base_subparser):
         Store test results (failing rows) in the database
         '''
     )
+    sub.add_argument(
+        '--greedy',
+        action='store_true',
+        help='''
+        Run all tests that touch the selected models, even if they also depend on unselected models
```
joellabes, line 734: This is probably not strictly true either - if it is the foreign table in a relationships test, that test won't be run even though this model touches it.

core/dbt/graph/cli.py
```diff
@@ -66,7 +67,7 @@ def parse_union_from_default(
 def parse_difference(
     include: Optional[List[str]], exclude: Optional[List[str]]
 ) -> SelectionDifference:
-    included = parse_union_from_default(include, DEFAULT_INCLUDES)
+    included = parse_union_from_default(include, DEFAULT_INCLUDES, greedy=flags.GREEDY)
```
jtcohen6, line 70: This captures all standard selection (`--select`, `--exclude`) perfectly.

To make the `--greedy` flag work with a yaml `--selector` too, we'd need to add a similar bit of logic that passes `flags.GREEDY` into each method called by [`parse_from_definition`](https://github.com/dbt-labs/dbt/blob/51a76d0d63efd8a393ace482c6c93afcf75f446c/core/dbt/graph/cli.py#L245). That's fairly straightforward for two of the conditions—`SelectionCriteria.from_single_spec(definition, greedy=flags.GREEDY)` and `SelectionCriteria.selection_criteria_from_dict(definition, dct, greedy=flags.GREEDY)` (called by `parse_dict_definition`) already accept `greedy` as an argument—but we'd need to add `greedy` arg support for `parse_union_definition` and `parse_intersection_definition` as well.

So: I'm thinking more about yaml selectors, and how they intentionally _don't_ (or _shouldn't_) intersect with additional CLI-provided selection criteria, and about the fact that they can define `greedy` themselves (!). I'll say more in my overall review comment.

jtcohen6 (commented):
### quick comments

- Really nice work on this one! You've managed to do this quite surgically.
- I agree that `dbt ls` should also support the `--greedy` flag (as should `dbt build`, #3596). However, `dbt ls` does something kinda funny & kinda good to swallow info-level logging, so that it _only_ outputs resource types. The `Some resources were excluded` message will still appear, it will just be written to `logs/dbt.log` instead of stdout.
- As far as testing this: I added a whole new suite, `test/integration/066_test_selection_tests`, just for the change in #3235. So I think you can add a new test case there, passing the greedy flag and saying "jk we want them all"!

### longer-form thoughts

It might get somewhat tiresome to keep seeing the info message. The canonical way to do this would be by supporting both `--greedy` and `--no-greedy` flags. If the user passes `--no-greedy`, that reinforces the default behavior and turns off the info logging. Ultimately, a bit of extra logging doesn't feel like a huge deal, and it's much better to show the message than show _nothing_ and confuse lots of people (i.e. what we're doing today). Unless this speaks to you in a big way, I'd be happy to call that out of scope for the current PR, and welcome a new `good first issue` + contribution when another community member feels strongly :)

Another thought I had: More than passing an on/off flag, it feels like the most-intentional way to "lock in" any type of resource selection should be a **yaml selector.** Believe it or not, based on the way we've implemented this, _it's already possible to set greediness within selectors_, and at a finer level of granularity:
```yml
selectors:
  - name: all_for_some_not_for_all
    definition:
      union:
        - method: fqn
          value: model_a
          greedy: false
        - method: fqn
          value: model_b
          greedy: true
```

There's one bit of extra logic we'd need, inside [`selection_criteria_from_dict`](https://github.com/dbt-labs/dbt/blob/51a76d0d63efd8a393ace482c6c93afcf75f446c/core/dbt/graph/selector_spec.py#L106):
```python
    @classmethod
    def selection_criteria_from_dict(
        cls, raw: Any, dct: Dict[str, Any], greedy: bool = False
    ) -> 'SelectionCriteria':
        if 'value' not in dct:
            raise RuntimeException(
                f'Invalid node spec "{raw}" - no search value!'
            )
        method_name, method_arguments = cls.parse_method(dct)
        
        # is 'greedy' defined in the yaml selector definition?
        # if so, use it
        if 'greedy' in dct:
            is_greedy = dct['greedy']
        
        # the 'greedy' arg to this method is False by default
        # we could edit cli.py to pass flags.GREEDY into that arg, if we want
        # the --greedy flag to interact with yaml selectors, either by
        # taking precedence or changing the default/fallback value
        else:
            is_greedy = greedy

        parents_depth = _match_to_int(dct, 'parents_depth')
        children_depth = _match_to_int(dct, 'children_depth')
        return cls(
            raw=raw,
            method=method_name,
            method_arguments=method_arguments,
            value=dct['value'],
            childrens_parents=bool(dct.get('childrens_parents')),
            parents=bool(dct.get('parents')),
            parents_depth=parents_depth,
            children=bool(dct.get('children')),
            children_depth=children_depth,
            greedy=is_greedy
        )
```
Questions:
- Should yaml selectors interact with the `--greedy` flag at all, or just always fall back to the default of `False`?
- If they do respect the `greedy` flag, this logic would still prioritize any `greedy` values set _inside_ the yaml selector, in version control, over the value set by the `--greedy` flag. Does that seem like the right precedence order? I'm thinking about this like `--full-refresh` + `--store-failures`, which change the global default but should be preceded by any resource-specific settings (`full_refresh` + `store_failures` configs).
- If they don't respect the `greedy` flag at all, and/or if they define greediness for all criteria, could we find a way to turn off the info logging about greediness if the user has specified `--selector`? By using a yaml selector, the user is in in "full control" mode. This might require a bit of tricky plumbing, for an effect that's ultimately more cosmetic than functional, but it feels promising as an abstraction...

joellabes:
OK I think I have `ls` going. I've added a couple of tests as starters and I know they're failing, but am struggling to interpret the [output](https://github.com/dbt-labs/dbt/pull/3833/checks?check_run_id=3494193736) so if you can give me some pointers that'd be great 🙏

core/dbt/graph/selector.py
```diff
@@ -258,8 +258,25 @@ def get_selected(self, spec: SelectionSpec) -> Set[UniqueId]:
...
+            if filtered_unused_nodes:
+                unused_node_names = []
+                for unique_id in filtered_unused_nodes:
+                    name = self.manifest.nodes[unique_id].name
+                    unused_node_names.append(name)
+                unused_nodes_str = ("\n  - ").join(unused_node_names)
+                msg = (
+                    f"\nSome resources were excluded because at least one parent is missing:"
+                    f"\n  - {unused_nodes_str}"
+                    f"\nUse the --greedy flag to include them"
```
jtcohen6, line 276: I'm on the fence between:
- it's worthwhile surfacing enough info to make clear what's happening
- taking up prime real estate with a list of potentially dozens/hundreds of not-greedily-included tests

What do you think about splitting this into two log messages:
- [info level] "Some resources were excluded because at least one parent is ~missing~ not selected. Use the --greedy flag to include them."
- [debug level] List every specific resource that was excluded because it had at least one parent unseelcted

Also - as I typed this, I realized "missing" may imply "missing from the database," when really we mean "not currently selected"

test/integration/066_test_selection_tests/test_selection_expansion.py
```diff
@@ -228,3 +232,32 @@ def test__postgres__exclude_data_test_tag(self):
...
+        greedy = True
+        expected = [
+            'cf_a_b', 'cf_a_src', 'just_a',
+            'relationships_model_a_fun__fun__ref_model_b_',
+            'relationships_model_a_fun__fun__source_my_src_my_tbl_',
+            'unique_model_a_fun'
+        ]
+            
+        self.list_tests_and_assert(select, exclude, expected, greedy)
+        self.run_tests_and_assert(select, exclude, expected, greedy)
```
jtcohen6, line 249: The `run_tests_and_assert` method takes other arguments (`schema` and `data`), so we need to specify the `greedy` arg by name, rather than just ordinal position:
```suggestion
        self.run_tests_and_assert(select, exclude, expected, greedy=greedy)
```

test/integration/066_test_selection_tests/test_selection_expansion.py
```diff
@@ -228,3 +232,32 @@ def test__postgres__exclude_data_test_tag(self):
...
+        exclude = 'test_name:unique'
+        greedy = True
+        expected = [
+            'cf_a_b', 'cf_a_src', 'just_a',
+            'relationships_model_a_fun__fun__ref_model_b_',
+            'relationships_model_a_fun__fun__source_my_src_my_tbl_',
+        ]
+            
+        self.list_tests_and_assert(select, exclude, expected, greedy)
+        self.run_tests_and_assert(select, exclude, expected, greedy)
```
jtcohen6, line 263:
```suggestion
        self.run_tests_and_assert(select, exclude, expected, greedy=greedy)
```

core/dbt/graph/selector.py
```diff
@@ -258,8 +258,25 @@ def get_selected(self, spec: SelectionSpec) -> Set[UniqueId]:
                 - selectors can filter the nodes after all of them have been
                   selected
         """
-        selected_nodes = self.select_nodes(spec)
+        selected_nodes, indirect_only = self.select_nodes(spec)
         filtered_nodes = self.filter_selection(selected_nodes)
+
+        if indirect_only:
```
jtcohen6, line 264: It could be a neat move to wrap this logging logic in another method, defined higher up, and then call it down here, similar to `alert_non_existence`.

I also bring that up because I think the `expect_exists` property (attached to [`BaseSelectionGroup`](https://github.com/dbt-labs/dbt/blob/9af071bfe481eb41cde4e68cf9ff35b812aa629d/core/dbt/graph/selector_spec.py#L164)) could be a neat example for what we want to do around a `raise_greedy_warning`, which would raise a warning for CLI-provided selection criteria, but not for yaml selector-provided selection criteria that define `greedy: true|false` explicitly. The `alert_non_existence` method is called `if spec.expect_exists` [here](https://github.com/dbt-labs/dbt/blob/9af071bfe481eb41cde4e68cf9ff35b812aa629d/core/dbt/graph/selector.py#L149-L150).

jtcohen6 (commented):
Quick fixes for the tests, and a few more comments :)

core/dbt/graph/selector.py
joellabes: Good shout on _missing_ vs _not selected_.

How about listing the first, say, 3 at info level and the whole lot at debug:
```
Some resources were excluded because at least one parent is not selected:
- excluded_test_1
- excluded_test_2
- excluded_test_3
- and 4 more
Use the --greedy flag to include them
```

My main concern is balancing the needs of someone who's brand new and doesn't know where to go for debug-level messages (esp in the IDE), while not annoying long-time power users.

test/integration/066_test_selection_tests/test_selection_expansion.py
joellabes: doh

core/dbt/graph/selector.py
joellabes: On the note of naming, will the "resources" always be tests? I'm assuming so since the `--greedy` flag is only available in the `dbt test` context. In which case could we give it a more concrete name?

jtcohen6: I'm happy with the balanced approach: listing the first three "and X more." What do you think about making the message inline rather than a list?

Today, the only resources that can be selected indirectly are tests; that's stated [here](https://github.com/dbt-labs/dbt/blob/df7ec3fb37396e67b7986cea99a370bba7c54ca0/core/dbt/graph/selector.py#L33-L42). The `--greedy` flag should be available to both `dbt test` and `dbt build`, but in both cases, the only things that show up in this list will be tests.

We did implement this in such a way as to minimize hard-coding that this is for tests only; while I can't think of one right now, we might eventually extend this to other resource types, though. But I'm happy to hard-code "tests" in the log message, to avoid user confusion, and update it if/when we widen the net of indirect selection.

joellabes: >What do you think about making the message inline rather than a list?

I just gave it a go, but found it very hard to visually parse when they all ran together instead of being newline delimited. Unless you feel really strongly that it should be all together, I'd rather not. 

Have hardcoded tests into the message for now 👍

joellabes:
@jtcohen6 I've just had a stab at collating all of the discussion topics here, to make sure I've understood everything

Still to do: 
- [ ] Flesh out some more integration tests
- [ ] Support greediness in selectors
   - [ ] using the [one bit of extra logic we'd need, inside `selection_criteria_from_dict`](https://github.com/dbt-labs/dbt/pull/3833#pullrequestreview-742570069)
- [ ] Suppress unused node warnings when a YAML selector is in use:
   - [x] [Move logging logic to own method](https://github.com/dbt-labs/dbt/pull/3833/files#diff-b4756f21e0fa65b77bb36828994d3a758360b5acd05bbb9526b5bfe399194617R31-R53)
   - [ ] Add a `raise_greedy_warning` property to `BaseSelectionGroup` (?) 
   - [ ] set `raise_greedy_warning` to true by default, override the spec generated by [`get_selection_spec`](https://github.com/dbt-labs/dbt/blob/9af071bfe481eb41cde4e68cf9ff35b812aa629d/core/dbt/task/compile.py#L42) to false because it means a YAML selector was provided (?)
- [ ] Add `--greedy` support to `--build` (but depending on when this gets wrapped up, it's more likely to be something I punt on and leave for whichever brave soul takes on #3596 at large)

Not to do: 
- [Make greedy flag available in conjunction with `--selector`](https://github.com/dbt-labs/dbt/pull/3833/files/833b2dc96bbb667198d62a16ea46614757c057dc#r699235206)
- Add a `--no-greedy` flag to suppress the (now slimmer) warnings

Does that sound right to you?

jtcohen6:
@joellabes Nice! That looks right to me. Thanks for keeping track of it all.

At this point, I think the current PR is just about merge-able as is. If you can get all the code checks + unit tests passing, once you're happy with the integration test coverage for the `--greedy` behavior, I'd love to get this in (in its current form) for v0.21.0-rc1.

Then, we can open and scope a separate issue to support greediness + yaml selectors. That will require some plumbing, but I do think the needed changes are fairly specific + self-contained. Also, I'd be okay with that functionality coming in a patch release or future version (v1.0).

Does that sound ok to you?

joellabes:
@jtcohen6 that split sounds great. Just to give you an update, I've been out of action this last week because it was my last few days at EP. Now I'm about to head off on holiday for a few days with no computer. 

I'm totally happy to close off these loose ends when I come back, but if you want to get the current work into .21 then I also won't be offended if you jump in.

jtcohen6:
@joellabes I think I resolved the outstanding TODOs, including support for the `greedy` property in yaml selectors. I did _not_ add a `--no-greedy` flag. If someone comes along who feels strongly, we can open a good first issue for it.

Congrats on wrapping up @ EP, and hope you enjoy your holiday! The CLA bot is having a tough time with the fact that you changed your email (I think?). I've manually checked to confirm that, yes indeed, you've signed our CLA :)

I think this PR is just about good to go. I hope to get it in for v0.21.0-rc1.

jtcohen6 (approved):

jtcohen6:
Aiii, looks like I need the CLA to pass in order to merge.

@joellabes Apologies in advance: I'm going to rebase this PR, and scrub your now-defunct email from the commits. In no way do I mean to take credit for the hard work you've put in here.

joellabes:
Ahhhh sorry! That final commit was actually some week-old unstaged work I found while cleaning out my laptop, but I'd already disassociated my work email from the account 😬 I'll fix that up when I'm back. Thanks for getting this over the line! (Especially since it was all the cleanup of my messes)

PR #3850, opened 2021-09-02
deepcopy args when passed down to rpc pask

### LINKED ISSUE #3846: `dbt ls` causes `dbt deps` to fail
### Describe the bug
Once you run `dbt ls` the following `dbt deps` commands will fail

### Steps To Reproduce
```# start the server
dbt rpc

# send a list request
curl -X POST http://localhost:8580/jsonrpc -d '{"jsonrpc": "2.0", "method": "list", "id": "1", "params": {"select": "debug"}}' -H 'Content-Type: application/json'

# send a deps request
curl -X POST http://localhost:8580/jsonrpc -d '{"jsonrpc": "2.0", "method": "deps", "id": "1"}' -H 'Content-Type: application/json'
```

### Expected behavior
A successful `dbt deps` run

Currently `dbt deps` returns
```
{"error": {"code": -32000, "message": "Server error", "data": {"type": "ValueError", "args": ["signal only works in main thread of the main interpreter"], "message": "signal only works in main thread of the main interpreter"}}, "id": "1", "jsonrpc": "2.0"}
```

Stack trace:
```
  File "/Users/kwigley/workspace/dbt/env/lib/python3.8/site-packages/jsonrpc/manager.py", line 113, in _get_responses
    result = method(*request.args, **request.kwargs)
  File "/Users/kwigley/workspace/dbt/core/dbt/rpc/task_handler.py", line 511, in __call__
    return self.handle(kwargs)
  File "/Users/kwigley/workspace/dbt/core/dbt/rpc/task_handler.py", line 503, in handle
    return self.handle_singlethreaded(kwargs, flags)
  File "/Users/kwigley/workspace/dbt/core/dbt/rpc/task_handler.py", line 427, in handle_singlethreaded
    self.process.task_exec()
  File "/Users/kwigley/workspace/dbt/core/dbt/rpc/task_handler.py", line 84, in task_exec
    signal.signal(signal.SIGTERM, sigterm_handler)
  File "/Users/kwigley/.asdf/installs/python/3.8.11/lib/python3.8/signal.py", line 47, in signal
    handler = _signal.signal(_enum_to_int(signalnum), _enum_to_int(handler))
ValueError: signal only works in main thread
```

### System information
- [ ] postgres
- [ ] redshift
- [ ] bigquery
- [ ] snowflake

```
<output goes here>
```

kwigley:
also resolves #3846

### Description

This makes sure that RPC tasks are initialized with a copy of `self.args`. _IF_ they are mutated further down the call stack, it won't affect the high level RPC process management task and be passed to subsequent RPC tasks.

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@
 
 - Support BigQuery-specific aliases `target_dataset` and `target_project` in snapshot configs ([#3694](https://github.com/dbt-labs/dbt/issues/3694), [#3834](https://github.com/dbt-labs/dbt/pull/3834))
 - `dbt debug` shows a summary of whether all checks passed or not ([#3831](https://github.com/dbt-labs/dbt/issues/3831), [#3832](https://github.com/dbt-labs/dbt/issues/3831))
-- Fix issue when running the `deps` task after the `list` task in the RPC server ([#3846](https://github.com/dbt-labs/dbt/issues/3846), [#3848](https://github.com/dbt-labs/dbt/pull/3848))
+- Fix issue when running the `deps` task after the `list` task in the RPC server ([#3846](https://github.com/dbt-labs/dbt/issues/3846), [#3848](https://github.com/dbt-labs/dbt/pull/3848), [#3850](https://github.com/dbt-labs/dbt/pull/3850))
 - Fix bug with initializing a dataclass that inherits from `typing.Protocol`, specifically for `dbt.config.profile.Profile` ([#3843](https://github.com/dbt-labs/dbt/issues/3843), [#3855](https://github.com/dbt-labs/dbt/pull/3855))
 
 ### Under the hood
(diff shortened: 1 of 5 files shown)


gshank:
Do you think we should remove the deepcopy from the RemoteMethod init method in core/dbt/rpc/method.py?  I guess it doesn't hurt, but it is an extra copy now.

kwigley:
> Do you think we should remove the deepcopy from the RemoteMethod init method in core/dbt/rpc/method.py? I guess it doesn't hurt, but it is an extra copy now.

@gshank good point, I'll remove it. It can be pretty expensive operation. I also like knowing that the object owns the args passed in, instead of creating a copy in the constructor.

nathaniel-may (approved):
LGTM

gshank (approved):
Looks good!

Issue #3830, opened 2021-08-30
dbt extensions / dbt jinja extensions


jpeak-intellify:
### Describe the feature

As a dbt user I should be able to install extensions through `pip` such that I can:
 - `pip install dbt-ext-myextension` OR
 - `pip install dbt-jinja-ext-myextension`

And the `dbt` cli autoloads these extensions [similar to pytest](https://docs.pytest.org/en/6.2.x/writing_plugins.html#making-your-plugin-installable-by-others) allowing control of the event hooks and the Jinja environment for custom behaviour.

### Describe alternatives you've considered

I am trying to automate overriding the `generate_database_name` and `generate_schema_name` macros but even if I install a dbt-package with my custom macro logic I still need to manually shim these macros in each repo. 

### Additional context

N/A

### Who will this benefit?

Power users that have an architecture or naming conventions dictated by business users that has outrun what dbt Out-Of-Box (OOB) can achieve.

Ideally allowing the Jinja environment to be extended would be an end goal here. It looks like the `do` command is already being used from `jinja2.ext.do`. https://github.com/dbt-labs/dbt/blob/develop/core/dbt/clients/jinja.py#L489

My ideal UX/DX would be:

- to develop a plugin with naming convention `dbt-jinja-ext-myextension`.
- `dbt` then uses the [`import_metalib.distributions()` functionality like `pluggy` does](https://github.com/pytest-dev/pluggy/blob/main/src/pluggy/_manager.py#L277) for autodiscovery.
- From that list of discovered plugins we have the array of strings to pass to the Jinja Environment.
- From there it is standard Jinja Extensions

This leverages prior art for plugin design. 
It is minimally invasive (leaving plugins for dbt event hooks as a separate scope of work).
It leverages the Jinja Extension community as the scope of what can be done.
Authoring a python extension better leverages python tooling to maintain that code than Jinja macros in a dbt package.

### Are you interested in contributing this feature?

Happy to help contribute this feature. Wanted to discuss scope of work before starting work though to fit in the roadmap.

For reference I was a [maintainer of `vcrpy`](https://github.com/kevin1024/vcrpy/commits?author=neozenith) API testing library for 6 months in 2019.

jpeak-intellify:
Thanks @jtcohen6 for the comprehensive context catchup.

# Big Ideas 

[...]

PR #3851, opened 2021-09-02
Specify macro_namespace of global_project dispatch macros

### LINKED ISSUE #3456: Allow dispatch & overrides of global macros via packages
### Describe the feature
As a dbt package developer, I'd like to be able to override a built-in (global) macro like `create_table_as` or `is_incremental` in a package and have projects that import the package transparently use the built-in macro without any changes to their code. 
One way to accomplish this would be to allow dbt to apply its package dispatch functionality to global macros. 
@jtcohen6 has suggested the following syntax in this [Slack thread](https://getdbt.slack.com/archives/C2JRRQDTL/p1623365118031500).

```yaml
dispatch:
  - macro_namespace: dbt
    search_order: ['my_package', 'dbt']
```
This would effectively inject `my_package` in the dispatch `search_path` for global macros in the `dbt` project.
It would also maintain the current behavior where dbt looks for overrides of global macros in the current root project and should follow the precedence described for materializations [here](https://docs.getdbt.com/docs/guides/creating-new-materializations#materialization-precedence).

So, e.g. if someone
- overwrote `create_table_as` locally in  the root project only, dbt would use the local version
- imported a package that overwrote `create_table_as` only, dbt would use the package version (over the global version)
- overwrote `create_table_as` locally in  the root project *and* imported a package that overwrote `create_table_as`, dbt would use the local version in the root project, over the package version and over the global version.

Another use case, in addition to overrides of the `default__` version of macros, would be for a package to supply alternate versions of global macros for specific platforms. Ideally such platform specific versions of global macros should belong in the respective plugin's global project, but there may be use cases for company-specific overrides to particular platforms that don't make sense in the plugin project. 

An alternative would be to wrap the macro exposed by the package locally. This would require project owners to be aware of the override and maintain a local wrapper macro for each override.

### Who will this benefit?
This would benefit package maintainers, particular those maintaining internal, company specific packages that override default behavior with customized implementations.

jtcohen6:
resolves #3456

Fuller description in the issue. Big idea is, you can reimplement a built-in global macro (including your adapter's built-in version) with a version that's defined in a package.

```yml
# dbt_project.yml
dispatch:
  - macro_namespace: dbt
    search_order: ['my_root_project', 'my_cool_package', 'dbt']
```

I think this _just works_ for existing functionality. I'll create some tests for the net-new functionality described above.

Following the conversation in #3830, I also added dispatching for `generate_schema_name` and `generate_alias_name`. Funny enough, `generate_database_name` was already dispatched.

This wasn't on our v1.0 list, but it feels quite 1.0-y, so I'd be excited to get it in before then.

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@
 - Allow the default seed macro's SQL parameter, `%s`, to be replaced by dispatching a new macro, `get_binding_char()`. This enables adapters with parameter marker characters such as `?` to not have to override `basic_load_csv_rows`. ([#3622](https://github.com/fishtown-analytics/dbt/issues/3622), [#3623](https://github.com/fishtown-analytics/dbt/pull/3623))
 - Alert users on package rename ([hub.getdbt.com#180](https://github.com/dbt-labs/hub.getdbt.com/issues/810), [#3825](https://github.com/dbt-labs/dbt/pull/3825))
 - Add `adapter_unique_id` to invocation context in anonymous usage tracking, to better understand dbt adoption ([#3713](https://github.com/dbt-labs/dbt/issues/3713), [#3796](https://github.com/dbt-labs/dbt/issues/3796))
+- Specify `macro_namespace = 'dbt'` for all dispatched macros in the global project, making it possible to dispatch to macro implementations defined in packages. Dispatch `generate_schema_name` and `generate_alias_name` ([#3456](https://github.com/dbt-labs/dbt/issues/3456), [#3851](https://github.com/dbt-labs/dbt/issues/3851))
 
 Contributors:
 
--- a/core/dbt/include/global_project/macros/adapters/common.sql
+++ b/core/dbt/include/global_project/macros/adapters/common.sql
@@ -1,5 +1,5 @@
 {% macro get_columns_in_query(select_sql) -%}
-  {{ return(adapter.dispatch('get_columns_in_query')(select_sql)) }}
+  {{ return(adapter.dispatch('get_columns_in_query', 'dbt')(select_sql)) }}
 {% endmacro %}
 
 {% macro default__get_columns_in_query(select_sql) %}
@@ -15,7 +15,7 @@
 {% endmacro %}
 
 {% macro create_schema(relation) -%}
-  {{ adapter.dispatch('create_schema')(relation) }}
+  {{ adapter.dispatch('create_schema', 'dbt')(relation) }}
 {% endmacro %}
 
 {% macro default__create_schema(relation) -%}
@@ -25,7 +25,7 @@
 {% endmacro %}
 
 {% macro drop_schema(relation) -%}
-  {{ adapter.dispatch('drop_schema')(relation) }}
+  {{ adapter.dispatch('drop_schema', 'dbt')(relation) }}
 {% endmacro %}
 
 {% macro default__drop_schema(relation) -%}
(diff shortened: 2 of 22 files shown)


gshank (approved):
This feels like it ought to have always worked this way. LGTM!

jpeak-intellify:
Really struggled to `pip install` just this branch to evaluate it which I have used these strategies before:
https://stackoverflow.com/questions/20101834/pip-install-from-git-repo-branch

Turns out I had to git clone the repo locally, checkout the branch then:
```
pip install --upgrade pip
pip install --upgrade setuptools wheel twine check-wheel-contents
cd scripts
. ./build-dist.sh
```

Then in my target toy dbt project:
```
python3 -m venv .venv
. ./.venv/bin/activate
pip install -U pip
cd path/to/dbt/dist
pip install *.whl
cd -
```

So I finally got our `dbt_bigcorp_util` installed and it dispatched but that package has sub macros that could not resolve and it's not obvious as a (novice) package developer how to handle this.

Error output:
```
λ dbt compile --no-version-check                             
Running with dbt=0.21.0-b1
* Deprecation Warning: The "packages" argument of adapter.dispatch() has been
deprecated. Use the "macro_namespace" argument instead.
Raised during dispatch for: except
For more information, see:
https://docs.getdbt.com/reference/dbt-jinja-functions/dispatch
Encountered an error:
Compilation Error in macro default__generate_database_name (macros/administration/generate_database_name.sql)
  'resolve_database_name' is undefined
  
  > in macro generate_database_name (macros/etc/get_custom_database.sql)
  > called by macro default__generate_database_name (macros/administration/generate_database_name.sql)
```

Does that make sense?

jpeak-intellify:
Following up, I got it working by name spacing calls to the internal macros with the name of the package.

Eg
`default__generate_database_name` now internally makes a call to `dbt_bigcorp_utils.resolve_database_name` instead of the bare `resolve_database_name`.

I thought it was something simple on my end.

This looks good to use immediately for our use case. 👍

jpeak-intellify:
Just following up here to save time explaining the context of the feature/bug I am experiencing before opening an issue where I have to repeat the context.

`dbt_project.yml`
```yaml
dispatch:
  - macro_namespace: dbt
    search_order: ['dbt_sample', 'my_macro_package', 'dbt']
```

But because I refactored **all** of my macros into `my_macro_package`, my `macros/` folder in my root `dbt_sample` project is empty.

So the above config gives the following error message:
```
Running with dbt=0.21.0
Encountered an error:
Compilation Error
  In dispatch: Could not find package 'dbt_sample'
```

The only workaround I can see is to change it to the following if the root project has no macros:
`dbt_project.yml`
```yaml
dispatch:
  - macro_namespace: dbt
    search_order: ['my_macro_package', 'dbt']
```

I'm not a huge fan of having to change preference config to match the existence of filesystem content. Feels like an opportunity for automation instead of manually twizzling config to suit.

If it is already reported and being addressed, a nudge to the existing issue would be appreciated so I can follow.
Otherwise if it is non-trivial, I'll create the full issue ticket to track the flow of work.

jtcohen6:
@jpeak-intellify Appreciate you putting the comment here! I _think_ we made a semi-related change that will incidentally fix this (https://github.com/dbt-labs/dbt-core/pull/4114), [by catching the mid-dispatch `Compilation Error` instead of raising it [here](https://github.com/dbt-labs/dbt-core/pull/4114/files#diff-d14b7d00e52da5da96bfb3bdef414e9289d923efe0a9cdadd716d792c7199535R166-R168).

We've backported that fix for v0.21.1, which is currently available as a release candidate. Could you take that for a spin, and see if it resolves the issue?
```
pip install dbt-<adapter>==0.21.1rc1
```
(where `<adapter>` is whichever one of postgres/redshift/snowflake/bigquery/etc that you're using)

jpeak-intellify:
Love your work. Yeah that nailed the issue. 👏

PR #3852, opened 2021-09-02
fix prerelease imports with loose version semantics

sungchun12:
resolves issue in comment link below

### Description

https://github.com/dbt-labs/dbt/pull/3759#issuecomment-911785073

### Checklist

- [x] I have signed the [CLA](https://docs.getdbt.com/docs/contributor-license-agreements)
- [x] I have run this code in development and it appears to resolve the stated issue
- [x] This PR includes tests, or tests are not required/relevant for this PR
- [x] I have updated the `CHANGELOG.md` and added information about my change to the "dbt next" section.

PR #3759, opened 2021-08-16
Better dbt packages version logging to aid in upgrading outdated packages

sungchun12:
resolves #3560 #3763

### Description

- For publically available dbt packages through the [package registry](https://hub.getdbt.com/), provide clear dbt package versions being installed vs. the latest version where applicable.
- Guides the developer to update older dbt packages
```yaml
packages:
  - package: dbt-labs/dbt_utils
    version: 0.7.0
  - package: dbt-labs/codegen
    version: 0.4.0
  - package: calogica/dbt_expectations
    version: 0.4.0
```
```bash
# display which specific packages are available for upgrade
# display everything that is up to date
$ dbt deps
Running with dbt=0.21.0-b1
Installing dbt-labs/dbt_utils@0.7.0
  Installed from version 0.7.0
  Latest hub registry version 0.7.1
Installing dbt-labs/codegen@0.4.0
  Installed from version 0.4.0
  Up to date!
Installing calogica/dbt_expectations@0.4.0
  Installed from version 0.4.0
  Latest hub registry version 0.4.1
Installing calogica/dbt_date@0.4.0
  Installed from version 0.4.0
  Up to date!

Upgrades available for: ['dbt-labs/dbt_utils', 'calogica/dbt_expectations']
```
image

```yaml
packages:
  - package: dbt-labs/dbt_utils
    version: 0.7.1
  - package: dbt-labs/codegen
    version: 0.4.0
  - package: calogica/dbt_expectations
    version: 0.4.1
```
```bash
$ dbt deps
Running with dbt=0.21.0-b1
Installing dbt-labs/dbt_utils@0.7.1
  Installed from version 0.7.1
  Up to date!
Installing dbt-labs/codegen@0.4.0
  Installed from version 0.4.0
  Up to date!
Installing calogica/dbt_expectations@0.4.1
  Installed from version 0.4.1
  Up to date!
Installing calogica/dbt_date@0.4.0
  Installed from version 0.4.0
  Up to date!
```
image

**Out of scope**: git-installed and local dbt packages

**Notes**:
- Tested `get_version_latest()` method call in` test/unit/test_deps.py`. This is only relevant for hub registry packages

**Open Questions**
- Test logging outputs in `test/rpc/test_deps.py`? I looked across several test files relevant to `deps` and didn't see any related to validating logging outputs. Can you provide guidance on where to best insert logging test cases/is it needed here given we aren't testing that today?

### Checklist
[...]

PR #3856, opened 2021-09-02
Fix dbt deps sorting behavior for non-standard version semantics

sungchun12:
resolves bugs from #3852 #3759 

### Description
When prereleases are included in the package versions list, it doesn't sort properly. The merged pull requests above had an issue with either being too strict or too loose when sorting version lists. It’s difficult to ensure this can capture all the unique ways people version their packages. [`StrictVersion`](https://www.python.org/dev/peps/pep-0386/#distutils) is too strict for packages and does not tolerate any deviation from the standard(think: `-` in the prerelease tag). [`LooseVersion`](https://www.python.org/dev/peps/pep-0386/#distutils)is too loose and not properly sorting prerelease versions BEFORE the latest stable version. 

This can apply to examples like this that exist in the [wild](https://hub.getdbt.com/mjirv/dbt_datamocktool/0.1.1-beta-fishtown/): `1.0.4-pre, 1.0.4-beta-fishtown`

This does NOT effect packages that aren't utilizing prereleases. 

With this solution, it iterates on the original `filter_installable` approach: https://github.com/dbt-labs/dbt/pull/3759#discussion_r697274347

However, this leverages a dictionary that sorts the formatted version strings first and then brings in the ORIGINAL version strings. If the ORIGINAL version string is not used, it can not properly download the package. Example: `0.5.0a1` will be interpreted as `0.5.0-a1` from dbt Hub and it will fail. (edit: example references how transforming a version number with [`from_version_string`](https://github.com/dbt-labs/dbt/blob/develop/core/dbt/semver.py#L94) using [`to_version_string`](https://github.com/dbt-labs/dbt/blob/develop/core/dbt/semver.py#L72) formats the version string in a way that isn't 1:1 with what exists in dbt Hub)
image

This solution also interprets the version number tag as a PRERELEASE and is lesser than a stable version.

Packages with known issues exemplifying current prerelease sorting behavior below. You'll notice the majority of these issues are with 3rd party packages. 
- [re-data/re_data](https://hub.getdbt.com/re-data/re_data/0.1.1-alpha/)
- [tailsdotcom/dbt_artifacts](https://hub.getdbt.com/tailsdotcom/dbt_artifacts/0.4.5a2/)
- [mjirv/dbt_datamocktool](https://hub.getdbt.com/mjirv/dbt_datamocktool/0.1.1-beta-fishtown/)
- [dbt-labs/adwords](https://hub.getdbt.com/dbt-labs/adwords/0.2.2a/)
- [Datavault-UK/dbtvault](https://hub.getdbt.com/datavault-uk/dbtvault/0.7.6-b1/)

**Example packages config**
image
**Current Behavior**
image

**Expected Behavior**

image

### Checklist

### DIFF
--- a/core/dbt/semver.py
+++ b/core/dbt/semver.py
@@ -3,7 +3,6 @@ import re
 from typing import List
 
 from packaging import version as packaging_version
-from distutils.version import LooseVersion
 
 from dbt.exceptions import VersionsNotCompatibleException
 import dbt.utils
@@ -439,9 +438,14 @@ def filter_installable(
         install_prerelease: bool
 ) -> List[str]:
     installable = []
+    installable_dict = {}
     for version_string in versions:
         version = VersionSpecifier.from_version_string(version_string)
         if install_prerelease or not version.prerelease:
-            installable.append(version_string)
-    installable.sort(key=LooseVersion)
-    return installable
+            installable.append(version)
+            installable_dict[str(version)] = version_string
+    sorted_installable = sorted(installable)
+    sorted_installable_original_version_string = [
+        installable_dict.get(str(version)) for version in sorted_installable
+    ]
+    return sorted_installable_original_version_string
--- a/core/dbt/semver.py
+++ b/core/dbt/semver.py
@@ -445,7 +445,7 @@ def filter_installable(
             installable.append(version)
             installable_dict[str(version)] = version_string
     sorted_installable = sorted(installable)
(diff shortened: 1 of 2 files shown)


jtcohen6 (approved):
@sungchun12 This looks great to me!

Thanks for the thorough appraisal of each permutation. You've managed to go deeper on this code than anyone has in a long time.

PR #3859, opened 2021-09-03
Use a macro for where subquery in tests

### LINKED ISSUE #3857: 0.20.1 tests `where` config does not support models with periods in the name
### Describe the bug

I'm trying to execute in BigQuery a `not_null` test on a model that includes a period in its name. Here is the relevant part of the `schema.yml`:
```
models:
  - name: staging.my_table_name
    columns:
      - name: reset_length
        tests:
          - not_null:
              where: "reset_unit != 'LIFETIME'"
```
The generated SQL for the test uses the model name as the alias, unquoted, which is not valid BigQuery SQL:
```
select
      count(*) as failures,
      count(*) != 0 as should_warn,
      count(*) != 0 as should_error
    from (
select *
from (select * from `project-name`.`dataset_name`.`staging__my_table_name` where reset_unit != 'LIFETIME') staging.my_table_name
where reset_length is null
    ) dbt_internal_test
```

### Steps To Reproduce
Create a model with a period in its name, for example `staging.my_table.sql` and include a `not_null` test with a `where` condition in the schema.

### Expected behavior
Expected behavior is valid SQL that test not-NULL values of columns under the given WHERE condition.

If applicable, add screenshots or log output to help explain your problem.

### System information
- [ ] postgres
- [ ] redshift
- [x] bigquery
- [ ] snowflake

```
installed version: 0.20.1
   latest version: 0.20.1

Up to date!

Plugins:
  - bigquery: 0.20.1
  - snowflake: 0.20.1
  - redshift: 0.20.1
  - postgres: 0.20.1
```

mac OS X 11.4 big sur

`Python 3.8.5`

jtcohen6:
resolves #3857

The previous way was good enough to get started. Now, let's define and use a macro instead to:
- make `test_metadata.model` in `manifest.json` much more readable
- enable adapter plugins + users to reimplement the where subquery logic if they desire

Also, let's just name the subquery something generic (such as `dbt_subquery`), rather than trying to alias the subquery with the model name. This wasn't really necessary (I overcorrected with regard to backwards compatibility), and it caused the issue in #3857 if the model's non-aliased name has special characters. The lowest-effort solution would have been to use `config.get('alias')` instead of the `identifier`, but I figured, while we're here, let's do a better thing.

TODO:
- [x] Fix all existing tests (pretty annoying with the `unique_id` hash suffix)
- [x] Add test case for specific bug in #3857

### Checklist


kwigley (approved):
nice, this definitely reads a lot better than what was there before!

PR #3865, opened 2021-09-09
Feature/3448 user defined default selection criteria

### LINKED ISSUE #3448: User-defined default selection criteria
_Credit where due: This was @aescay's idea many months ago!_

### Describe the feature

We've increasingly heard a desire from community members to limit `dbt run` in development to only a subset of their project, or to always exclude a set of infrequently run models. Trying to solve for this with `enabled` gets tricky quickly, because disabled resources cannot participate in the DAG and will raise dependency errors accordingly.

Instead, what if there were a way to redefine the default selection criteria? I think the right construct for this is a [yaml selector](https://docs.getdbt.com/reference/node-selection/yaml-selectors) (as in `selectors.yml`). We could support a new selector property, `default: true|false`, or check for a selector named `default`. The former is almost certaintly better, we'd just need to raise an error if multiple selectors have `default: true`.

If specified, that selector should override the default includes + excludes defined here:

https://github.com/fishtown-analytics/dbt/blob/abe8e839458d0288f76cb6950e79c77e3b1627cc/core/dbt/graph/cli.py#L22-L23

Then, `dbt run` would have the same effect as `dbt run --selector my_default_selector`.

### Questions

1. How to combine the user-defined default with additional selection criteria passed via CLI syntax? Should the two be combined, or should the CLI criteria entirely override the default? I'm leaning toward total override for two reasons:
    - It's what happens today: `dbt ls` really means ` dbt ls --select fqn:* source:* exposure:*`, as soon as the user says `-s something_else`, the command becomes `dbt ls -s something_else`
    - Override feels like the only way to make sense of `dbt ls --selector not_the_default`
2. Can we support different default selectors in different environments? I think it makes a lot of sense to perma-exclude certain models in development, but still select them in production. As it turns out, `selectors.yml` already supports Jinja today, so this could be as simple as:
```yml
selectors:
  - name: prod
    description: Select everything in prod
    definition: 'fqn:* source:* exposure:*'
  - name: dev
    description: Avoid unpleasant surprises in dev
    definition:
      union:
        - 'fqn:* source:* exposure:*'
        - exclude:
            - unpleasant
            - surprises
```

TeddyCr:
resolves #3448 

### Description
This PR introduces a `default` field in the `selectors.yml` allowing users to define 1 default selector that will execute the definition on each `dbt run`, `dbt compile`, `dbt list`, or `dbt source freshness`.

### Checklist


TeddyCr:
> Thank you for your pull request and welcome to our community. We could not parse the GitHub identity of the following contributors: **Teddy Crepineau**.
> This is most likely caused by a git client misconfiguration; please make sure to:
> 
> 1. check if your git client is configured with an email to sign commits `git config --list | grep email`
> 2. If not, set it up using `git config --global user.email email@example.com`
> 3. Make sure that the git commit email is configured in your GitHub account settings, see https://github.com/settings/emails

@jtcohen6 do I need to re-write the author info on all the commits (i.e. the last two commits happened after I ran the commands in the cla-bot message)?

jtcohen6:
@TeddyCr Nice work!! So excited to see this PR :)

Yes, I think you may need to re-write those commits so that the associated git identity (email) matches the one in your GitHub profile (which has signed the CLA)

core/dbt/config/project.py
```diff
@@ -645,13 +645,25 @@ def from_project_root(
...
+    def get_selector(self, name: str) -> Union[SelectionSpec, bool]:
         if name not in self.selectors:
             raise RuntimeException(
                 f'Could not find selector named {name}, expected one of '
                 f'{list(self.selectors)}'
             )
-        return self.selectors[name]
+        return self.selectors[name]["definition"]
+
+    def get_default_selector(self) -> Union[SelectionSpec, bool, None]:
```
jtcohen6, line 656: Makes sense that `get_default_selector` returns either `SelectionSpec` or `None`. Is there ever a case where `get_selector` or `get_default_selector` returns `bool`?

core/dbt/config/selectors.py
```diff
@@ -118,6 +119,25 @@ def selector_config_from_data(
...
+        if selector.default is True and default_set is False:
+            default_set = True
+            default_selector_name = selector.name
+            continue
+        if selector.default is True and default_set is True:
+            raise DbtSelectorsError(
+                "Error when parsing the selector file. "
+                "Default key is set to True multiple times.\n"
+                "Default with value set to True found in "
+                f"{default_selector_name} and {selector.name}"
```
jtcohen6, line 137: Nice! Let's make it clear that `default: true` is found in multiple different selectors. Perhaps something like:
```python
                f"Found multiple selectors with `default: true`:"
                f"{default_selector_name} and {selector.name}"
```

jtcohen6 (commented):
@TeddyCr Agh, sorry you're in a tricky git state! I think either (a) rebasing or (b) opening a new PR from a new branch is the way to go. I'm happy to help with the rebase/reopen if you'd like.

All in all, this PR looks really solid, and I'm thrilled with the change and test coverage you've put together. I left a few small comments.

Small thing: Could you also renumber the new integration test from `101_` to `073_`? (`100_rpc_test` is a bit of a special case.)

core/dbt/config/project.py
TeddyCr: @jtcohen6 that is correct, though, mypy complained the expected return type for these two should include bool.

TeddyCr:
> @TeddyCr Agh, sorry you're in a tricky git state! I think either (a) rebasing or (b) opening a new PR from a new branch is the way to go. I'm happy to help with the rebase/reopen if you'd like.
> 
> All in all, this PR looks really solid, and I'm thrilled with the change and test coverage you've put together. I left a few small comments.
> 
> Small thing: Could you also renumber the new integration test from `101_` to `073_`? (`100_rpc_test` is a bit of a special case.)

Perfect, let me know once you are done with the review, this way I'll make a new PR from a new branch with the updates -- if that works with you.  🙂

core/dbt/config/project.py
jtcohen6: Ah, classic mypy. Ok!

jtcohen6:
@TeddyCr This works brilliantly in local testing. I think you're good to close, rebase, and reopen :)

My only thoughts are around logging:
- I left one comment about rewording the error message if multiple selectors with `default: true` are found
- I'm also wondering if we should include a log message to avoid confusing users who type unqualified `dbt run`, and who might be surprised to see dbt only run a few models:
```
$ dbt run
Running with dbt=0.21.0-b1
Found 123 models, 456 tests, 3 snapshots, 3 analyses, 789 macros, 1 operation, 10 seed files, 200 sources, 20 exposures
Using default selector 'my_default_selector'

09:26:51 | Concurrency: 8 threads (target='dev')
09:26:51 |
09:26:51 | 1 of 6 START view model dbt_jcohen.default_selected_model................. [RUN]
```
What do you think? I'm always hesitant to add stdout clutter, but if it could prevent end-user confusion, it feels worth doing.

jtcohen6:
@TeddyCr This works brilliantly in local testing. I think you're good to close, rebase, and reopen :)

My only thoughts are around logging:
- I left one comment about rewording the error message if multiple selectors with `default: true` are found
- I'm also wondering if we should include a log message to avoid confusing users who type unqualified `dbt run`, and who might be surprised to see dbt only run a few models:
```
$ dbt run
Running with dbt=0.21.0-b1
Found 123 models, 456 tests, 3 snapshots, 3 analyses, 789 macros, 1 operation, 10 seed files, 200 sources, 20 exposures
Using default selector 'my_default_selector'

09:26:51 | Concurrency: 8 threads (target='dev')
09:26:51 |
09:26:51 | 1 of 6 START view model dbt_jcohen.default_selected_model................. [RUN]
```
What do you think? I'm always hesitant to add stdout clutter, but if it could prevent end-user confusion, it feels worth doing.

TeddyCr:
> @TeddyCr This works brilliantly in local testing. I think you're good to close, rebase, and reopen :)
> 
> My only thoughts are around logging:
> 
> * I left one comment about rewording the error message if multiple selectors with `default: true` are found
> * I'm also wondering if we should include a log message to avoid confusing users who type unqualified `dbt run`, and who might be surprised to see dbt only run a few models:
> 
> ```
> $ dbt run
> Running with dbt=0.21.0-b1
> Found 123 models, 456 tests, 3 snapshots, 3 analyses, 789 macros, 1 operation, 10 seed files, 200 sources, 20 exposures
> Using default selector 'my_default_selector'
> 
> 09:26:51 | Concurrency: 8 threads (target='dev')
> 09:26:51 |
> 09:26:51 | 1 of 6 START view model dbt_jcohen.default_selected_model................. [RUN]
> ```
> 
> What do you think? I'm always hesitant to add stdout clutter, but if it could prevent end-user confusion, it feels worth doing.

PR moved [here](https://github.com/dbt-labs/dbt/pull/3875) with fixes

PR #3870, opened 2021-09-10
Schema check testing

leahwicz:
Adding an Action that will compare our current official schema in the `schemas.getdbt.com` repo against the generated schema from the ref that the Action runs on. This will show us if there are changes in the artifact schemas that we need to alert Metadata to. 

If a mismatch is found, the diff is uploaded as an artifact.

Future work: it would be nice to have this either create a ticket or ping on Slack or something so that it is more obvious but this is the first step to getting there.

FYI: it looks like we do have schema changes right now compared to what's in the repo so it's successfully detecting it

### DIFF
--- a/.github/workflows/schema-check.yml
+++ b/.github/workflows/schema-check.yml
@@ -6,7 +6,8 @@ on:
   pull_request:
 
 env:
-  LATEST_SCHEMA_PATH: ${{github.workspace}}/new_schemas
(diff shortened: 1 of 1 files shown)


leahwicz:
@kwigley I'm curious what you think about this approach. I wanted to do the check in this repo and not in the release repo so I took your existing python code and morphed into into Actions scripts

jtcohen6 (approved):
lgtm!

PR #3875, opened 2021-09-13
3448 user defined default selection criteria

### LINKED ISSUE #3448: User-defined default selection criteria

TeddyCr:
resolves #3448 

### Description
This PR introduces a `default` field in the `selectors.yml` allowing users to define 1 default selector that will execute the definition on each `dbt run`, `dbt compile`, `dbt list`, or `dbt source freshness`.

This is a clean PR of #3865 fixing git status after rebased perform to fix CLA error.

### Checklist


jtcohen6:
Closing and reopening to trigger adapter integration tests

core/dbt/config/project.py
```diff
@@ -645,13 +646,26 @@ def from_project_root(
...
+
+    def get_default_selector(self) -> Union[SelectionSpec, bool, None]:
+        """This function fetch the default selector to use on `dbt run` (if any)
+        :return: either a selector if default is set or None
+        :rtype: Union[SelectionSpec, None]
+        """
+        for selector_name, selector in self.selectors.items():
+            if selector["default"] is True:
+                name = selector_name
+                print_timestamped_line(f'Using default selector {name}')
```
jtcohen6, line 665: Really happy to have logging that will help avoid user confusion. Two quick thoughts:
- [cosmetic] I think this should just be `logger.info` rather than `print_timestamped_line`. It feels more like the `Found 5 models, ...` line (which doesn't include a timestamp), versus the actual execution steps that print with timestamps below it.
- [functional] We don't want to log this line every time the user has a default selector _defined_, only if it's being _used_ (i.e. no CLI `--select` or `--exclude` args). So I think this logging needs to go in the `get_selection_spec` task method instead, in the second conditional branch.

Since that's one more line of code to add in three places, I'm thinking: All of these tasks now use the same exact definition of `get_selection_spec` (thanks to #3554). What do you think of just updating the `GraphRunnableTask` method, so that instead of implementing as an `abstractmethod` it actually defines the logic that `compile`, `list`, and `freshness` will all use? Then we can delete the duplicated code from `compile`, `list`, and `freshness`, since they'll just inherit from `GraphRunnableTask` (just as `run`, `test`, etc inherit from `compile`).

https://github.com/dbt-labs/dbt/blob/3effade26603a38fa71b10990a82e3318faad7b8/core/dbt/task/runnable.py#L106-L110

jtcohen6 (commented):
@TeddyCr This looks really really good! One last comment re: logging, and (if you feel up to it) DRYing up our task method code a bit.

Thanks for reopening the new PR and passing the CLA check, sorry again for the bother.

core/dbt/config/project.py
TeddyCr: Do you know why list.py had `spec = parse_difference(self.selector, ...` in `def get_selection_spec` while freshness and compile have `spec = parse_difference(self.args.select` ?

jtcohen6: @TeddyCr Ah, it's just because `list` accepts both `models` and `select`. So options are:
- reimplement `get_selection_spec` for `list` alone
- implement `selector` as a property on `GraphRunnableTask`, and override just for `ListTask`

I'm leaning toward the latter

jtcohen6 (approved):
@TeddyCr Thanks for all the amazing work here! I'd really like to get this change in ahead of v0.21.0-rc1, so I'm going to merge this PR as soon as the tests pass. Then, I'll open a follow-on PR just to clean up the last bit of logging and duplicative task logic.

PR #3877, opened 2021-09-13
Make ModelParser detect macro overrides for ref, source, and config.

### LINKED ISSUE #3866: Turn on experimental parser by default
Pre-reqs:
- [ ] Macros that override `ref` and `source`
- [ ] Adding second flag to put this logic behind instead

nathaniel-may:
resolves #3866

### Description

The static parser (formerly known as the experimental parser) relies on a static semantic meaning of `ref`, `source`, and `config` calls. If this semantic meaning is changed by the user by defining their own macro by the same name, the experimental parser would by default read the model file with incorrect assumptions. So instead, the model parser must fall back to jinja rendering in order for the integration of the static parser to work for all projects.

The way this is implemented, it checks for all macros in scope for this particular file (i.e. - from the root project scope and from the current project scope) and falls back if any of the banned macros are defined. Because of this implementation, if any dependencies in the project have these banned macro definitions, we can still use the experimental parser for model files within the user's project itself if they didn't also create their own macro overrides for these calls.

I found this change particularly difficult to test cleanly, but it turns out we have preexisting art around using debug log entries to detect internal state so that's the route I went when it came to testing this functionality. There's also a small unit test that doesn't test much, but overall I'm happy with the coverage this change has.

### Reviewers

The meat of the change is contained to two files:
- `core/dbt/parser/models.py`
- `test/integration/072_experimental_parser_tests/test_all_experimental_parser.py`

Most of the other files are integration test setup and are nearly trivial to review.

### TODO
- [x] Integration test that detects ref source and config overrides

### Checklist


jtcohen6:
Excited to see the quick progress here! 🚀 

Is our current plan to include this in v0.21, and then use the updated sampling from v0.21 usage to verify whether we can turn on the experimental parser by default in v1.0?

nathaniel-may:
@jtcohen6 I believe so! @leahwicz can confirm though.

Turns out this might be tricky to test, so that's going to be the one snag I have to get past.

leahwicz:
@jtcohen6 @nathaniel-may I would like to get this into the RC but if it's not ready by EOW, I don't want the RC to be held up on this. Let's play it by ear and see where we are by Thursday and make the call

nathaniel-may:
The simplest way to test this is in the unit test I wrote here: test/unit/test_parser.py::ExperimentalModelParserTest. It's kind of bare bones compared to the coverage I would like to have, but I really can't find a sane way to get an integration test up and running for macro detection. If anyone has ideas, I'm all ears.

nathaniel-may:
To be clear, this strategy for macro detection relies on the fact that the manifest has the names of the macros populated before the model parser is executed. This is very unlikely to change, but if it did, we would get all greens even with this unit test and the experimental parser would actually be broken.

nathaniel-may:
Just to wrap up the comment thread here, the current integration tests cover the whole stack from project files -> model parser run. Any change outside the model parser that would prevent us from detecting macros at parse time would be caught by these tests.

core/dbt/parser/models.py
```diff
@@ -141,7 +124,102 @@ def render_update(
...
+
+        return reduce(
+            lambda z, key: z or (key in self.manifest.macros),
+            all_banned_macro_keys,
+            False
+        )
+
+
+# returns a list of string codes to be sent as a tracking event
+def _get_sample_result(
```
gshank, line 167: Moving this code out into a separate function was a good idea.

gshank (approved):
Looks good! Glad you were able to test it.

leahwicz (approved):

Issue #3259, opened 2021-04-13
Improve test FQNs


jtcohen6:
### Describe the feature

The FQNs for schema + data tests today are sort of silly, especially schema tests.
```
$ dbt ls --resource-type test
my_other_project.data_test.my_test
my_project.schema_test.accepted_values_model_a_id__1__2__3
my_project.subfolder.data_test.my_test
my_project.schema_test.my_test_model_a_
```
This means that, from the project file, these tests have to be configured sort of awkardly:
```yml
tests:
  my_project:
    schema_test:
      my_test_model_a_:
        +enabled: false
  my_other_project:
    data_test:
      +severity: warn
```

Ideally, we'd remove the data test / schema test nomenclature (#3234) and rationalize the FQNs.

- ~Bespoke~ Singular test FQNs are straightforward: just project + file path
- Generic test blocks: just project + name? Or fuller file-path FQNs?
- Specific instantiations of generic tests: ideally, this might include target FQNs (e.g. `my_project.models.subfolder.model_a.id.unique`), to enable things like:
```yml
tests:
  models:
    subfolder:
      model_a:
        +enabled: false
        id:
          +enabled: true
```
But that may be overkill for the first cut of this! The main use case for configuring tests from `dbt_project.yml` is disabling/controlling behavior from a package. Maybe neater package-level FQNs is all we need.

In any case, we could start making the adjustments here:
https://github.com/fishtown-analytics/dbt/blob/7ec5c122e17ab7e1a5dd1e1cab36527159564a0f/core/dbt/parser/schemas.py#L439-L444

jtcohen6:
resolved by #3880

Issue #3347, opened 2021-05-12
Schema tests for sources are added to the manifest twice.


iknox-fa:
### Describe the bug
Tests that are applied to sources are added to the manifest twice.  Once in the [source patcher](https://github.com/fishtown-analytics/dbt/blob/8d39ef16b67fcc3c30e8a18751b4e64831ec9aaf/core/dbt/parser/sources.py#L128-L132) and once in the  [schema parser](https://github.com/fishtown-analytics/dbt/blob/8d39ef16b67fcc3c30e8a18751b4e64831ec9aaf/core/dbt/parser/schemas.py#L530-L533).

This wasn't caught before due to the early return in the [exception method that should have been thrown when this occurs](https://github.com/fishtown-analytics/dbt/blob/8d39ef16b67fcc3c30e8a18751b4e64831ec9aaf/core/dbt/exceptions.py#L774).

### Steps To Reproduce
This behavior doesn't break functionality, however it's a performance hit and reduces our ability to throw helpful errors when duplicated tests are found.

To see what's actually happening start by commenting out lines noted in core/dbt/exceptions.py and run a dbt project that has tests applied to a source.

### Expected behavior
dbt  adds source tests to the manifest only once.

### System information
**Which database are you using dbt with?**
- [x] postgres
- [ ] redshift
- [ ] bigquery
- [ ] snowflake
- [ ] other (specify: ____________)

**The output of `dbt --version`:**
```
 0.20.0-b1
```

**The operating system you're using:**
macOS

**The output of `python --version`:**
Python 3.8.8

PR #3880, opened 2021-09-14
Rename tests: singular + generic

### LINKED ISSUE #3234: Rename "schema" + "data" tests
For v0.20.0, we want to make tests more powerful and (crucially) much more consistent. Today, there are real functional differences between "schema" tests and "data" tests. Instead, tests should just be **tests**. They'll still have two points of entry, but this should reflect only the dbt developer's choices to trade off clarity and reusability.

As such, we're thinking about renaming:
- schema test &rarr; **generic test**
    - alternatives: "macro test," "reusable test"
- data test &rarr; ~**bespoke test**~ **singular test**
    - alternatives: "file test," "one-off test"

This renaming has real implications for:
- Internal objects (e.g. `ParsedDataTestNode`, `ParsedSchemTestNode`). This shouldn't have any implications for end users.
- File names, e.g. builtin generic tests are defined in [`global_project/macros/schema_tests/`](https://github.com/fishtown-analytics/dbt/tree/develop/core/dbt/include/global_project/macros/schema_tests). This is cosmetic only.
- Resource FQNs + output of `dbt ls`. (Today, the `unique_id` is `test.my_project.unique_model_a_fun`, but the fqn is `my_project.schema_test.unique_model_a_fun`.)
- Selection criteria: this represents a functional change, so I'll say more below

### Selection criteria

- `dbt test --data`, `dbt test -m test_type:data`
- `dbt test --schema`, `dbt test -m test_type:schema`

Should instead become:

- `dbt test -m test_type:singular`
- `dbt test -m test_type:generic`

Notes:
- We should make this backward compatible by continuing to supporting `schema` + `data` as aliases for the names we ultimately decide
- We should nix the CLI flag form entirely, in favor of the `test_type` method. I'm open to whether we should continue supporting the `--schema` and `--data` flags for backwards compatibility.

- Leaving the old names in place

jtcohen6:
resolves #3234

### Basics

- Rename `data` &rarr; `singular`, `schema` &rarr; `generic`
- Retain backwards compatibility for `test_type:schema` + `test_type:data`

### Opinionated choices
- ❗ Use only the word `test` in FQN construction, rather than `schema_test` or `data_test`. It's still possible to tell the difference in `manifest.json` via `test_metadata`, but overall I want the two to feel more consistent!
- ❗ ❗  Remove support for `--data` and `--schema` flags to `dbt test` entirely. Remove `data` + `schema` tags from tests.

### TODOs
- I have at least one failing test having to do with a duplicated `unique_id`. It might be https://github.com/dbt-labs/dbt/issues/3347, I'm not yet sure.
- Since I'm already touching test FQNs, why not make them even better? (#3259) I may decide that's out of scope here.
- I renamed file names containing `schema_test` or `data_test` when they are dbt-core modules. I'd prefer to not rename integration tests for now, to avoid big git diffs.

### Checklist


core/dbt/parser/schemas.py
```diff
@@ -325,8 +325,6 @@ def _parse_generic_test(
             'kwargs': builder.args,
         }
         tags = sorted(set(itertools.chain(tags, builder.tags())))
-        if 'schema' not in tags:
```
leahwicz, line 328: This seems randomly removed but maybe it isn't?

core/dbt/parser/sources.py
```diff
@@ -269,11 +269,6 @@ def parse_source_test(
             tags=tags,
             column_name=column_name
         )
-        # we can't go through result.add_node - no file... instead!
-        if node.config.enabled:
```
leahwicz, line 273: Same here. What does this have to do with the rename? Honestly asking to learn

jtcohen6: This is the fix for #3347, in which we noticed a while ago that source tests are added to the manifest _twice_.

Things I don't know:
- why that doesn't cause a functional error currently
- why that _started_ causing a functional error when I made the other changes in this PR, i.e. renamed things

Despite not knowing those things, I feel reasonably good about making this fix. I can move the commit (0adec32) into a separate PR if that would make things clearer.

core/dbt/parser/schemas.py
jtcohen6: Super fair question!

In v0.18, we introduced a selection method, `test_type:`, that allowed users to select just schema tests (`dbt test --models test_type:schema`) or data tests (`dbt test --models test_type:data`).

Prior to that, we supported the same functionality, but in a pretty silly way. dbt would secretly add tags—`'schema'` to schema tests, `'data'` to data tests—and then users could pass a flag `dbt test --schema` to select all tests tagged "schema," or `dbt test --data` to select all tests tagged "data." Funny enough, that would include schema tests manually tagged "data," too, or data tests that selected from models tagged "schema." Not a good system.

This PR takes the opinionated view that we should do away with that legacy approach to the functionality, since we've had a better approach (`test_type` selector) for several versions now. To that end, there's some backwards compatibility for the newer approach, and total deprecation fro the older approach:
- Stop secretly adding `'schema'` and `'data'` tags to tests
- Deprecate `--schema` and `--data` flags entirely
- Continue supporting `test_type:schema` + `test_type:data` as aliases for `test_type:generic` and `test_type:singular`, respectively

gshank (approved):
Looks good!

PR #3882, opened 2021-09-14
Collab PR for macro detection integration tests

nathaniel-may:
Run this test with the following command:
```
python3 -m pytest -m profile_postgres test/integration/072_experimental_parser_tests/test_all_experimental_parser.py::TestRefOverrideExperimentalParser
```

This is the print debug lines I see:
```
RENDER_UPDATE CALLED
FLAG ON
macros with ref in name:
['macro.dbt.should_full_refresh']
```

I expect to see `macro.test.ref` in the list of macros with ref in the name since it's defined in `macros/ref.sql` but I don't see it.

I verified that the macro detection does indeed work by making a dummy project locally and just running `dbt parse` on it, so I'm not sure why it's not working in this integration test setting.

I attempted to specify the macro path in the project config since that didn't seem to be set by default, but that didn't work either.

Let me know if you have any thoughts as to why this isn't detecting the macro!

### DIFF
--- a/core/dbt/parser/models.py
+++ b/core/dbt/parser/models.py
@@ -1,7 +1,7 @@
 from dbt.context.context_config import ContextConfig
 from dbt.contracts.graph.parsed import ParsedModelNode
 import dbt.flags as flags
-import dbt.tracking
+from dbt.logger import GLOBAL_LOGGER as logger
 from dbt.node_types import NodeType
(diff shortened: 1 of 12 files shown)


nathaniel-may:
@kwigley figured it out! macro-paths is relative to the test definition.

nathaniel-may:
merging into original feature branch.

PR #3884, opened 2021-09-15
Build task arg parity

### LINKED ISSUE #3596: dbt build: Parity for flags

kwigley:
resolves #3596 

### Description

Adds:
- `--resource-type` param to cli and rpc task
- ~ensures that `-models` and `--select` are mutually exclusive (aligns with `list` task behavior)~
	- ~additionally, `--models` will only select nodes of type `model` (again, to align with `list` task behavior)~
- adds `--store-failures` param to cli task

### Checklist

### DIFF
--- a/core/dbt/task/build.py
+++ b/core/dbt/task/build.py
@@ -4,12 +4,8 @@ from .seed import SeedRunner as seed_runner
 from .test import TestRunner as test_runner
 
 from dbt.contracts.results import NodeStatus
-from dbt.exceptions import RuntimeException, InternalException
-from dbt.graph import (
-    parse_difference,
-    ResourceTypeSelector,
-    SelectionSpec,
-)
+from dbt.exceptions import InternalException
+from dbt.graph import ResourceTypeSelector
 from dbt.node_types import NodeType
 from dbt.task.test import TestSelector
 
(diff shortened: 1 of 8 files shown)


core/dbt/task/build.py
```diff
@@ -4,40 +4,96 @@
...
+
+    @property
+    def resource_types(self):
+        if self.args.models:
+            return [NodeType.Model]
+
+        if not self.args.resource_types:
+            return list(self.ALL_RESOURCE_VALUES)
+
+        values = set(self.args.resource_types)
```
jtcohen6, line 56: Agree that parity with the `list` task is a good guideline to follow. I have half a mind to just entirely remove support for the `--models` flag in `dbt build`, since it's a totally new command in v0.21 and we shouldn't have to worry about backwards compatibility. I think either is a fair way to go, and I hesitate to make you rewrite a bunch of code given that I think you've done a sensible thing.

FWIW my plan is to just about entirely remove `--models` from our documentation (https://github.com/dbt-labs/docs.getdbt.com/pull/802), except for changelog notes and a note about backwards compatibility in `list`.

kwigley: let's think about our future selves and not support it with `build` from the start. not a huge lift at all!

gshank (approved):
Looks good to me.

PR #3888, opened 2021-09-15
[#3563] Don't reload and validate schema files if they haven't changed

### LINKED ISSUE #3563: Skip yaml validation for files that are completely unchanged
Thanks to https://github.com/dbt-labs/dbt/pull/3460, when dbt reads yaml files, it also performs minimal validation and stores their intermediate representation in dictionary form. This enables us to:
- Return good error messages early & often
- Better separate file loading from manifest construction
- In partial parsing: only re-parse nodes affected by specific lines changed in yaml files, rather than re-parsing all resources implicated in the entire yaml file

At the same time, partial parsing runs need to read and validate every yaml file in the project, every time. This should still be _really_ fast, even for big projects, so long as the user has `libyaml` installed (C bindings for `PyYAML`). I'm going to make sure we document this.

Would it be possible to have the best of both worlds, by storing a hash of the yaml file, and skip re-reading / validating it entirely if it's completely unchanged?

gshank:
resolves #3563

### Description

Store the modification time for files and for schema files compare the times and skip reloading if the files haven't changed. This only addresses schema files because we are already storing the dictionary representation in the saved files. If we wanted to do the same thing for SQL files we would have to store the contents.

### Checklist

### DIFF
--- a/core/dbt/clients/system.py
+++ b/core/dbt/clients/system.py
@@ -30,7 +30,7 @@ def find_matching(
     root_path: str,
     relative_paths_to_search: List[str],
     file_pattern: str,
-) -> List[Dict[str, str]]:
+) -> List[Dict[str, Any]]:
     """
     Given an absolute `root_path`, a list of relative paths to that
     absolute root path (`relative_paths_to_search`), and a `file_pattern`
@@ -61,11 +61,19 @@ def find_matching(
                 relative_path = os.path.relpath(
                     absolute_path, absolute_path_to_search
                 )
+                modification_time = 0.0
+                try:
+                    modification_time = os.path.getmtime(absolute_path)
+                except OSError:
+                    logger.exception(
+                        f"Error retrieving modification time for file {absolute_path}"
+                    )
                 if reobj.match(local_file):
                     matching.append({
                         'searched_path': relative_path_to_search,
                         'absolute_path': absolute_path,
                         'relative_path': relative_path,
+                        'modification_time': modification_time,
                     })
 
     return matching
--- a/core/dbt/parser/hooks.py
+++ b/core/dbt/parser/hooks.py
@@ -72,10 +72,13 @@ class HookParser(SimpleParser[HookBlock, ParsedHookNode]):
 
     # Hooks are only in the dbt_project.yml file for the project
     def get_path(self) -> FilePath:
+        # There ought to be an existing file object for this, but
+        # until that is implemented use a dummy modification time
         path = FilePath(
             project_root=self.project.project_root,
             searched_path='.',
             relative_path='dbt_project.yml',
+            modification_time=0.0,
         )
         return path
 
--- a/core/dbt/parser/read_files.py
+++ b/core/dbt/parser/read_files.py
@@ -12,13 +12,27 @@ from typing import Optional
 # This loads the files contents and creates the SourceFile object
 def load_source_file(
         path: FilePath, parse_file_type: ParseFileType,
-        project_name: str) -> Optional[AnySourceFile]:
-    file_contents = load_file_contents(path.absolute_path, strip=False)
-    checksum = FileHash.from_contents(file_contents)
+        project_name: str, saved_files,) -> Optional[AnySourceFile]:
+
     sf_cls = SchemaSourceFile if parse_file_type == ParseFileType.Schema else SourceFile
-    source_file = sf_cls(path=path, checksum=checksum,
+    source_file = sf_cls(path=path, checksum=FileHash.empty(),
                          parse_file_type=parse_file_type, project_name=project_name)
-    source_file.contents = file_contents.strip()
+
+    skip_loading_schema_file = False
+    if (parse_file_type == ParseFileType.Schema and
+            saved_files and source_file.file_id in saved_files):
+        old_source_file = saved_files[source_file.file_id]
+        if (source_file.path.modification_time != 0.0 and
+                old_source_file.path.modification_time == source_file.path.modification_time):
+            source_file.checksum = old_source_file.checksum
+            source_file.dfy = old_source_file.dfy
+            skip_loading_schema_file = True
+
+    if not skip_loading_schema_file:
+        file_contents = load_file_contents(path.absolute_path, strip=False)
+        source_file.checksum = FileHash.from_contents(file_contents)
+        source_file.contents = file_contents.strip()
+
     if parse_file_type == ParseFileType.Schema and source_file.contents:
         dfy = yaml_from_file(source_file)
         if dfy:
@@ -69,7 +83,7 @@ def load_seed_source_file(match: FilePath, project_name) -> SourceFile:
 
 # Use the FilesystemSearcher to get a bunch of FilePaths, then turn
 # them into a bunch of FileSource objects
-def get_source_files(project, paths, extension, parse_file_type):
+def get_source_files(project, paths, extension, parse_file_type, saved_files):
     # file path list
     fp_list = list(FilesystemSearcher(
         project, paths, extension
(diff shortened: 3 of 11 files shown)


core/dbt/clients/system.py
```diff
@@ -61,11 +61,13 @@ def find_matching(
                 relative_path = os.path.relpath(
                     absolute_path, absolute_path_to_search
                 )
+                modification_time = os.path.getmtime(absolute_path)
```
nathaniel-may, line 64: It looks like `getmtime` can raise an `OSError`. We should probably explicitly handle that case.

core/dbt/parser/hooks.py
```diff
@@ -76,6 +77,7 @@ def get_path(self) -> FilePath:
             project_root=self.project.project_root,
             searched_path='.',
             relative_path='dbt_project.yml',
+            modification_time=time.time(),  # this is wrong, but temporary
```
nathaniel-may, line 80: Can we update the comment to be a little more descriptive? Why do we need this now, and under what conditions will it change?

nathaniel-may (changes requested):
Looks like a nice and simple change. I just want us to handle the edge case where the accessing the last modified time throws.

core/dbt/parser/hooks.py
gshank: I've updated the comment.

core/dbt/clients/system.py
gshank: It might be a good idea, but this code is all being reimplemented in the filesystem adapter, so this is going to be temporary. Also we just got the list of files from os.path, so there's not much of a chance that the file isn't there.

nathaniel-may: I agree that it's unlikely to come up, but I think it also throws if the metadata doesn't exist, or is inaccessible independent of the file existing. Given that dbt runs on a ton of machines, I'd like to handle all our edge cases if possible especially since this is an exception we can eat. If the last updated timestamp isn't there, just reload it like normal.

core/dbt/parser/hooks.py
```diff
@@ -72,10 +73,13 @@ def transform(self, node):
 
     # Hooks are only in the dbt_project.yml file for the project
     def get_path(self) -> FilePath:
+        # There ought to be an existing file object for this, but
+        # until that is implemented use a dummy modification time
```
nathaniel-may, line 77: If we're using dummy times and the times might not always be accessible, it seems like a good use case for an `Optional` value.

gshank: This is only for this one use case for creating a temporary FilePath (when we should really have some kind of a file object). Other FilePath objects should always have a modification time, so this should not be optional.  We could read the modification time for this file but that would be a waste since the FilePath is not persistent and is just discarded.

nathaniel-may: To me it just doesn't look like the last updated time is actually guaranteed since the function we use to collect it throws. If it were something we could always have on every os, it wouldn't have a reason to throw.

core/dbt/clients/system.py
gshank: There won't be any thrown exceptions. I checked that getmtime works correctly on Windows. And it certainly works on Mac and Linux. We don't support any other platforms. The test suite has been run on all of our integration environments and it works correctly.

That file just came from os.walk. So if it doesn't exist it could only be from a bug in os.walk, which is unlikely and would have all kinds of other implications. The code to get those files has been there since the beginning and if there was a possibility of the file not being there we would have seen errors before.

I agree that exceptions should be caught when they're possible. This is not a place where they're possible. Adding an unnecessary exception complicates the code and is overly defensive.

kwigley: I land right in the middle, it shouldn't happen but if it does, let's raise something a _little_ more helpful maybe?
```suggestion
				try:
                	modification_time = os.path.getmtime(absolute_path)
                except OSError as exc:
                	# this shouldn't be reachable, raise something a little more helpful
                	raise dbt.exceptions.InternalException(
        				f'Unhandled exception getting modified time for file "{absolute_path}": {exc}'
    				) from exc

```

core/dbt/parser/read_files.py
```diff
@@ -12,13 +12,26 @@
...
+
     sf_cls = SchemaSourceFile if parse_file_type == ParseFileType.Schema else SourceFile
-    source_file = sf_cls(path=path, checksum=checksum,
+    source_file = sf_cls(path=path, checksum=FileHash.empty(),
                          parse_file_type=parse_file_type, project_name=project_name)
-    source_file.contents = file_contents.strip()
+
+    skip_loading_schema_file = False
+    if (saved_files and source_file.file_id in saved_files and
+            parse_file_type == ParseFileType.Schema):
```
kwigley, line 23: nit: suggesting a better short circuit
```suggestion
    if (parse_file_type == ParseFileType.Schema and
    		saved_files and source_file.file_id in saved_files):
```

test/unit/test_graph.py
```diff
@@ -137,6 +138,7 @@ def _mock_hook_path(self):
                 searched_path='.',
                 project_root=os.path.normcase(os.getcwd()),
                 relative_path='dbt_project.yml',
+                modification_time=time.time(),
```
kwigley, line 141: nit: I'd try to not use `time.time()` during testing, this value will change every time you run tests. This won't cause any issues today, but its something I don't want to be copied for other tests. Can we use a dummy float value here?

kwigley: This comment goes for the other places `time.time()` is used in tests as well

kwigley (dismissed):

core/dbt/clients/system.py
nathaniel-may: `getmtime` is dependent on the filesystem not the OS and we definitely do not test across all the filesystems our users are on. Hence, my strong preference to code defensively here.

@kwigley if we're going to catch it, I feel like we should just eat the exception by passing None. If the modification time doesn't exist, then dbt would fallback to the original behavior of reloading. But I will agree, throwing your suggested exception at least lets us debug it if it happens, but it would still happen.

kwigley: Ah, ok. That's a distinction I didn't grok in my first pass, I thought that this would only raise if something went really bad or some race condition where we read a file and the file disappeared when trying to get the modified time. I don't have a sense of how often this could raise... but catching, logging and moving along seems appropriate. Though, I'm sure something will throw elsewhere :)
```suggestion
                modification_time = time.time() # <- I don't think this can be None?
                try:
                    modification_time = os.path.getmtime(absolute_path)
                except OSError as exc:
                    logger.exception('uh oh')
```

kwigley: I think `None` makes more sense though

core/dbt/parser/read_files.py
```diff
@@ -12,13 +12,26 @@
...
-    source_file = sf_cls(path=path, checksum=checksum,
+    source_file = sf_cls(path=path, checksum=FileHash.empty(),
                          parse_file_type=parse_file_type, project_name=project_name)
-    source_file.contents = file_contents.strip()
+
+    skip_loading_schema_file = False
+    if (parse_file_type == ParseFileType.Schema and
+            saved_files and source_file.file_id in saved_files):
+        old_source_file = saved_files[source_file.file_id]
+        if old_source_file.path.modification_time == source_file.path.modification_time:
```
nathaniel-may, line 25:
```suggestion
        if source_file.path.modification_time != 0.0 and old_source_file.path.modification_time == source_file.path.modification_time:
```

nathaniel-may: if we ever detect the default value, we want to reload.

kwigley (commented):
looks good! deferring approval to @nathaniel-may

nathaniel-may (approved):
This looks good to me! Thanks for the updates.

PR #3892, opened 2021-09-16
Fix logging for default selector

jtcohen6:
Follow-on to #3875, the actual functionality of which I was eager to get into v0.21 :)

In particular, this fixes https://github.com/dbt-labs/dbt/pull/3875#discussion_r708094168. Only log "Using default selector" when the default selector is in fact being used, not every time one has been defined.

While we're here: Acknowledging that the `get_selection_spec` method is now 99% the same across the three tasks that implement it, move the method definition up to `GraphRunnableTask` (replacing the abstract method), and remove the duplicated code from the child tasks.


kwigley (approved):
nice! love deleting code

PR #3893, opened 2021-09-16
removed deprication for materialization-return and replaced it with a…

### LINKED ISSUE #2798: Enact deprecations
### Describe the feature

- As of dbt v0.18.0, we collect anonymous usage stats for deprecation warnings.
- We were already planning to end support for `config-version: 1` of `dbt_project.yml` in dbt v0.19.0 (per implementation of #2735)
- We should officially deprecate:
    - `dbt-project-yaml-1`: this is far & away the most common deprecation warning, and we still see it popping up in ~40% of projects running v0.18.0. Users may have upgraded their own projects but still need to upgrade installed packages. In any case, this is definitely the most controversial of our proposed deprecations.
    - `column-quoting-unset`: 11% of projects
    - `models-key-mismatch`: 2% of projects
    - `materialization-return`: <1% of projects
    - `not-a-dictionary`: 0% of projects
- We should _not_ deprecate:
    - `adapter-macro`: this is the second-most-common deprecation warning, and we only introduced it in v0.18 with the switch to `adapter.dispatch`. We should give users at least one more minor version to upgrade.

- Leaving these as warnings until v1.0.0, or forever

### Who will this benefit?
- Us as dbt maintainers in the short term, as we can reduce the surface area of code that patched older syntaxes
- Users, in the long run, as we continue to crystallize the UX of dbt project development

emmyoop:
resolves #2798

### Description

Enact deprecation for `materialization-return`.  Removed the deprecation warning and replaced it with an exception. 

Looking for feedback to see if there's anything I'm overlooking (other than updating the CHANGELOG.md).  I tested locally and ran all tests.

### Checklist


gshank (approved):
Looks good to me. I can't think of anything you're missing.

test/integration/053_custom_materialization/test_custom_materialization.py
```diff
@@ -98,3 +98,19 @@ def test_postgres_default_dependency(self):
...
+        return {
+            'config-version': 2,
+            'macro-paths': ['override-view-return-no-relation']
+        }
+
+    @use_profile('postgres')
+    def test_postgres_default_dependency(self):
+        self.run_dbt(['deps'])
+        self.run_dbt(['run'], expect_pass=False)
+
```
jtcohen6, line 115: If you want, you can check the results for the explicit error message you're expecting:
```python
        results = self.run_dbt(['run'], expect_pass=False)
        assert 'did not explicitly return a list of relations' in results[0].message
```
I wouldn't say that's required here, though it could be a nice way to ensure we're providing users with a helpful error message, and not just failing silently.

jtcohen6 (commented):
Looks great! Left one small comment that you can choose to take or leave.

Thanks also for splitting up the various deprecations into sub-tasks. Feeling much more organized :)

Let's hold off on merging this into `develop` until we cut v0.21.0-rc1.

CHANGELOG.md
```diff
@@ -1,3 +1,13 @@
+## dbt 1.0.0 (Release TBD)
+
+### Features
+
+### Fixes
+
+### Under the hood
+
+- Enact deprecation for `materialization-return` and replace deprecation warning with an exception. ([#3896](https://github.com/dbt-labs/dbt/issues/3896))
+
```
jtcohen6, line 10: 🎉

CHANGELOG.md
```diff
@@ -1,3 +1,13 @@
+## dbt 1.0.0 (Release TBD)
```
jtcohen6, line 1: 🎉

PR #3894, opened 2021-09-16
Add timing and thread info to sources.json artifact

### LINKED ISSUE #3804: Results in sources.json missing contextual info present in run_results.json
Establish parity between `sources.json` and `run_results.json`, especially for per-result information:
- `thread_id`
- `execution_time`
- `timing`

The relevant classes here:
https://github.com/dbt-labs/dbt/blob/ab06149c81d36889cdf9dff5b5903ae0d7be92e2/core/dbt/contracts/results.py#L91-L108

https://github.com/dbt-labs/dbt/blob/ab06149c81d36889cdf9dff5b5903ae0d7be92e2/core/dbt/contracts/results.py#L255-L301

It _looks_ like the source result classes inherit from generic node results, and then the output classes do not. So it's possible that the results themselves already include these properties, and it's just a matter of adding them into the output.

drewbanin:
resolves #3804

### Description

This PR adds the following fields to the `sources.json` artifact:
 - timing (same format as run_results.json)
 - execution_time
 - thread_id

Big question here is: is this the right change to make, or should we pursue moving this information into the `run_results.json` file instead? I think that adding it to `sources.json` is a net-positive, it isn't a breaking change, and it doesn't preclude us from migrating this metadata into the `run_results.json` file in the future.... so why not? :)

Small question: if we did want to merge this, where should it live in the changelog?

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 - Add support for execution project on BigQuery through profile configuration ([#3707](https://github.com/dbt-labs/dbt/issues/3707), [#3708](https://github.com/dbt-labs/dbt/issues/3708))
 - Skip downstream nodes during the `build` task when a test fails. ([#3597](https://github.com/dbt-labs/dbt/issues/3597), [#3792](https://github.com/dbt-labs/dbt/pull/3792))
 - Added default field in the `selectors.yml` to allow user to define default selector ([#3448](https://github.com/dbt-labs/dbt/issues/3448#issuecomment-907611470))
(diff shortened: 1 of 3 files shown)


jtcohen6 (approved):
I can respond to the substantive point over in https://github.com/dbt-labs/dbt/issues/3804#issuecomment-921086339. In the meantime, this feels like an uncontroversial improvement. Happy to get it in & under the wire for v0.21.0! We'll just want to bump the schema version for `sources.json` in v0.21.

Logistically speaking, you can just merge to `develop` once the tests pass. We're planning to merge `develop` into `0.21.latest` ahead of cutting v0.21.0-rc1.

drewbanin:
Right on! I just updated the changelog and put this in for 0.21.0 and will merge when all is green :)

PR #3912, opened 2021-09-17
add log_var macro

### LINKED ISSUE #3911: new jinja function: log var
### Describe the feature

As I mention in https://github.com/bastienboutonnet/vscode-dbt/pull/14

> in Python 3.8+ I'm now very dependent on the f-string and walrus-operator related move:
> ```
> > foo = 'my string'
> > print(f'{foo=}')
> foo='my string'
> ```
> 
> I find myself doing be below in Jinja all the time and missing my Pythonic way of doing it. A snippet would is the lowest-hanging fruit, although I'd love a `log_var()` macro in dbt, someday!
> ```
> {{ log('drop_script: ' ~ drop_script, info=True) }}
> ```

Making it available as a snippet for VSCode, Atom and/or dbtCloud?

### Who will this benefit?

anyone who spends a lot of time writing dbt-jinja.

Yes! PR imminent...

swanderz:
resolves #3911

### Description

Add a new `log_var` dbt-jinja macro. My implementation only works for Python 3.8 and above. getting a variable's name in python<=3.7 is way trickier. my initial idea for the PR would be jinja instead of python like `msg = 'drop_script: ' ~ drop_script`, but not sure how I'd do that.

### Checklist

### DIFF
--- a/core/dbt/context/base.py
+++ b/core/dbt/context/base.py
@@ -447,6 +447,31 @@ class BaseContext(metaclass=ContextMeta):
         else:
             logger.debug(msg)
         return ''
+    
+    @contextmember
+    @staticmethod
+    def log_var(var: str, sep = ': ', info: bool = False) -> str:
+        """Given a variable,
+        logs the variable name and value separated by sep
+
+        :param var: The variable to log
+        :param sep: how name and value are concatenated
+        :param info: If `False`, write to the log file. If `True`, write to
+            both the log file and stdout.
+
+        > macros/my_log_var_macro.sql
+
+            {% macro some_macro(arg1) %}
+              {{ log(arg1) }}
+            {% endmacro %}"
+        """
+        msg = f'{var=}' # py3.8+ only
+
+        if info:
+            logger.info(msg)
+        else:
+            logger.debug(msg)
+        return ''
 
     @contextproperty
     def run_started_at(self) -> Optional[datetime.datetime]:


core/dbt/context/base.py
```diff
@@ -447,6 +447,31 @@ def log(msg: str, info: bool = False) -> str:
...
+
+        :param var: The variable to log
+        :param sep: how name and value are concatenated
+        :param info: If `False`, write to the log file. If `True`, write to
+            both the log file and stdout.
+
+        > macros/my_log_var_macro.sql
+
+            {% macro some_macro(arg1) %}
+              {{ log(arg1) }}
```
jtcohen6, line 465:
```suggestion
              {{ log_var(arg1) }}
```

core/dbt/context/base.py
```diff
@@ -447,6 +447,31 @@ def log(msg: str, info: bool = False) -> str:
...
+            {% endmacro %}"
+        """
+        msg = f'{var=}' # py3.8+ only
```
jtcohen6, line 468: We'll obviously want to clearly document that this is only supported by python 3.8 and above. I think we may need to go a step further, though, to prevent this from bricking dbt on older python versions.

As a case in point, unit tests are running just fine for me locally (3.9) and [in the py38 GitHub Action](https://github.com/dbt-labs/dbt/pull/3912/checks?check_run_id=3634994100), but they're running into big issues in [3.6](https://github.com/dbt-labs/dbt/pull/3912/checks?check_run_id=3634994028) and [3.7](https://github.com/dbt-labs/dbt/pull/3912/checks?check_run_id=3634994062).

I wonder if something like this can do the trick?
```python
        if sys.version_info[0] == 3 and sys.verson_info[1] < 8:
            msg = f'{var}' # better than nothing? or just log "do not use this!"
        else:
            msg = f'{var=}' # py3.8+ only
```

jtcohen6 (commented):
A very handy shortcut!

A big first step will be un-bricking py36 + py37. Once that's sorted:
- You'll need to add `log_var` to the [set of required context keys](https://github.com/dbt-labs/dbt/blob/6563d09ba7da855c2ed05498d7909bb793f9c4a1/test/unit/test_context.py#L174-L190) in `test/unit/test_context.py`
- For integration testing, [`013_context_var_tests`](https://github.com/dbt-labs/dbt/tree/develop/test/integration/013_context_var_tests) could be a good home for this. I think [`run_dbt_and_capture`](https://github.com/dbt-labs/dbt/blob/6563d09ba7da855c2ed05498d7909bb793f9c4a1/test/integration/base.py#L591) will come in handy, since it lets you capture and examine stdout.

swanderz:
can probably be rolled in #4184

PR #3922, opened 2021-09-20
Feature/catalog relational objects

### LINKED ISSUE #3920: Generate catalog for relational, non-ephemeral nodes only
Similar to what #3716 did for schema caching, we should do for cataloging, too. If we expect a node to exist as an object in the database, add its database-schema to the set of namespaces to catalog; if it's ephemeral or otherwise non-relational, ignore it.

I think this would be as simple as updating the logic in `_get_catalog_schemas` to check for `node.is_relational and not node.is_ephemeral_model`, as we do in [`_get_cache_schemas`](https://github.com/dbt-labs/dbt/blob/05dea18b625253625c135cbf37311d0206a8cc79/core/dbt/adapters/base/impl.py#L303-L314):

https://github.com/dbt-labs/dbt/blob/05dea18b625253625c135cbf37311d0206a8cc79/core/dbt/adapters/base/impl.py#L316-L337

dave-connors-3:
resolves #3920 

### Description

This PR changes the base behavior for generating the map of schemas used in the `dbt docs generate` step to only those that contain relational, non-ephemeral nodes. Current behavior checks all nodes, which includes tests, and can result in errors when used in conjunction with custom `generate_database_name` behavior in a project. 

Note: when `store_failures` is set to true, this logic will interpret tests as relational nodes, and therefore can trigger a similar error depending on the default `generate_database_name` behavior in the project.

### Checklist


jtcohen6 (approved):
lgtm!

Could you pull `develop`, and add a note to the changelog under `1.0.0 - Under the hood`? As well as adding a `Contributors` section, and adding yourself to it? :)

jtcohen6 (approved):
thanks for this one, @dave-connors-3 :)

PR #3939, opened 2021-09-22
bump sampling back to 100

### LINKED ISSUE #3377: [Tracking] Experimental Parser (Tree Sitter)
## Proof of Concept (Completed 04/20/21)

- [x]  [Write tree-sitter grammar for subset of jinja and unit test.](https://github.com/fishtown-analytics/dbt/issues/3215)
- [x]  [Extract refs sources and configs from model file.](https://github.com/fishtown-analytics/dbt/issues/3215)
- [x]  Detect false positives and misses with sample projects- both must be zero.
- [x]  Determine if the percent of model files affected is worth it to include in the main product. (result = 60%)

## 0.20.0-RC1  (Completed 05/20/21)

- [x]  [dbt with the experimental parser passes unit tests](https://github.com/fishtown-analytics/dbt/issues/3306)
- [x]  [Run internal analytics project with and without the experimental parser. Manifest files must compare as exactly the same.](https://github.com/fishtown-analytics/dbt/issues/3306)
- [x] [dbt with the experimental parser passes integration tests](https://github.com/fishtown-analytics/dbt/issues/3306)
- [x]  [Make a packaging strategy for the new parser + dbt.](https://github.com/fishtown-analytics/dbt/issues/3307)
- [x]  Split tree-sitter repository from extractor
- [x]  [Create separate CI pipeline for tree-sitter repository to create wheels and upload to pypi ](https://github.com/fishtown-analytics/dbt/issues/3307)(Thank you, Kyle!)
- [x]  Rerun integration tests once the tree-sitter dependency is installable from pypi.
- [x]  [Add `--use-experimental-parser` flag to core, defaulting to false.](https://github.com/fishtown-analytics/dbt/issues/3307)

## 0.20.0-RC2 (Completed 06/29/21)
- [x] rewrite extractor and tests in rust. ([PR](https://github.com/fishtown-analytics/dbt-parser-generator/pull/12))
- [x] write PyO3 wrapper to expose to python ABI (https://github.com/fishtown-analytics/dbt-parser-generator/issues/19)
- [x] depend on tree-sitter-jinja2 not via private repo branch (publish on crates.io, or public repo (easiest)
- [x] add github actions to package extractor (https://github.com/fishtown-analytics/dbt-parser-generator/issues/18)
- [x] add step to publish to pypi to github actions (https://github.com/fishtown-analytics/dbt-parser-generator/issues/18)
- [x] rerun project test runner with rust extractor (https://github.com/fishtown-analytics/dbt-parser-generator/issues/20)
- [x] in dbt remove extractor.py, both tree-sitter dependencies, and instead depend on the rust extractor package from pypi [PR](https://github.com/fishtown-analytics/dbt/pull/3497)

## 0.20.0

- [x]  Add snowplow reporting to sometimes run both and send back if they match or not. (https://github.com/fishtown-analytics/dbt/issues/3503)
- [x]  Add counter for how many model files were "experimentally parsed" vs "jinja rendered" to the perf file. (https://github.com/fishtown-analytics/dbt/issues/3438)
- [x]  Add snowplow reporting to track how many models can be parsed without jinja rendering. (https://github.com/fishtown-analytics/dbt/issues/3438)

## Before flag is on by default

- [x]  Rewrite extractor in Rust for type safety and confidence in making future changes. Will need to decide to keep Rust code in the dbt repository, or depend on the extractor as an outside library. (Went with outside library)
- [x]  Make sure the experimental parser can detect when ref source or config has been overridden with a macro and can skip everything instead of running. New test for this necessary. Hard prereq for turning on by default.
- [ ] Apply feedback from beta testers.
- N/A  ~Work with Experience and Platform teams to create and execute a beta-testing plan with the `--use-experimental-parser` flag.~ This is no longer necessary with the current structure of the flag and tracking.

nathaniel-may:
resolves #3377

### Description

- adds new flag `--no-static-parse` to the cli, env var, and profile config
- turns on static parsing by default
- disables sampling, but leaves the logic intact for when we expand the static parser to include new experimental features that we need to sample for correctness

### Checklist


nathaniel-may:
I have to sign off, but the only thing I can't get to pass yet is the testing the `DBT_NO_STATIC_PARSER` environment flag in `test_postgres_env_no_static_parser`.

Opening the PR for reviews though so I can address any concerns when I return from PTO.

core/dbt/main.py
```diff
@@ -1066,14 +1066,24 @@ def parse_args(args, cls=DBTArgumentParser):
         help=argparse.SUPPRESS,
     )
 
-    # if set, will use the tree-sitter-jinja2 parser and extractor instead of
-    # jinja rendering when possible.
+    # if set, will use the latest features from the static parser instead of
+    # the stable static parser..
```
jtcohen6, line 1070: 👍

core/dbt/parser/models.py
```diff
@@ -28,80 +28,39 @@ def resource_type(self) -> NodeType:
...
-                        "status": result
-                    })
-
-        # if the --use-experimental-parser flag was set, and the experimental parser succeeded
-        elif isinstance(experimentally_parsed, Dict):
+        statically_parsed: Optional[Union[str, Dict[str, List[Any]]]] = None
+        experimental_sample: Optional[Union[str, Dict[str, List[Any]]]] = None
+
+        # sample the experimental parser during a normal run that hasn't explicitly
+        # turned of static parsing.
```
jtcohen6, line 49:
```suggestion
        # turned off static parsing.
```

core/dbt/parser/models.py
```diff
@@ -114,29 +73,104 @@ def render_update(
...
-            super().render_update(node, config)
-        # the experimental parser ran on this model and failed.
-        # fall back to python jinja rendering.
+        # if the static parser didn't succeed, fall back to jinja
         else:
+            # jinja rendering
+            super().render_update(node, config)
             logger.debug(
-                f"1602: parser fallback to jinja because of extractor failure for {node.path}"
+                f"1602: parser fallback to jinja rendering on {node.path}"
```
jtcohen6, line 91: Now that we're turning on the experimental parser by default, worth calling out: we see a _lot_ of this debug message in full re-parses. You figure ~40% of models, in a project with 1k models in it, means ~400 log lines like this. That's not necessarily a bad thing, and it's definitely preferable to the other (very misleading) log line that we currently print once per node (https://github.com/dbt-labs/dbt/issues/3137)

core/dbt/parser/models.py
```diff
@@ -114,29 +73,104 @@ def render_update(
...
+        # if we want information on what features are barring the static
+        # parser from reading model files, this is where we would add that
+        # since that information is stored in the `ExtractionError`.
+        except ExtractionError:
+            logger.debug(
+                f"1603: static parser failed on {node.path}"
+            )
+            return "cannot_parse"
+
+    def run_experimental_parser(
```
jtcohen6, line 143: It looks like this is an exact dupe of the logic in `run_static_parser`, except subbing out "static" for "experimental." I get that, at a future point where we have more experimental features to sample/enable, we'll want to do other things here. Is this how we want the code to be in the meantime? Honest q, I'm happy with either answer

jtcohen6 (commented):
Broadly, looking really good! No blockers from me, just a couple of questions

core/dbt/parser/models.py
nathaniel-may: If the team wanted this to logic to be deduplicated, I can totally get behind that. However, the rationale for this decision was to leave the scaffolding for what this feature actually needs to maximize maintainability. It's not just for the rest of the team, but also for future me when I forget what the plan was when we go to change this a few months down the line 😅

nathaniel-may:
To resolve my previous comment, @gshank helped me figure out how to handle the environment variable issue I was having. Everything should work exactly as intended now.

core/dbt/parser/models.py
jtcohen6: Ah, I see how it's needed for tests. Ok, it's decidedly debug-level logging, I don't feel a need to be too precious about taking up space. We can reserve the right to swing back later and cut it.

jtcohen6 (approved):

gshank (approved):
This looks good. I do agree that all of those debug lines in the logs could make for a lot of noise. Could you open a testing ticket to add a 'test' log output? Just so we don't forget about it.

nathaniel-may:
#3977

PR #3955, opened 2021-09-27
Feature: nullable error_after in source

### LINKED ISSUE #3874: For source freshness, allow null error_after with non-null warning_after
### Describe the feature
I have a few sources where it would be great to have:
- non-null `warning_after`
- null `error_after`

Today, if you leave `error_after` empty while specifying a `warning_after` value or if you set `error_after: null`, the `error_after` value defaults to the parent value.

Not the same, but a hacky workaround is to set a high `error_after` value that should never hit.

### Who will this benefit?
I haven't seen too many people ask about this, but I'd imagine there must be many that would use this! Especially if highlighted in the docs as a use case.

No. But if I'm the only one asking for it, I'll reconsider lol.

kadero:
resolves #3874 

### Description
#TODO

### Checklist

### DIFF
--- a/core/dbt/parser/sources.py
+++ b/core/dbt/parser/sources.py
@@ -348,6 +348,12 @@ def merge_freshness(
     base: Optional[FreshnessThreshold], update: Optional[FreshnessThreshold]
 ) -> Optional[FreshnessThreshold]:
     if base is not None and update is not None:
+        if base.error_after is None and update.error_after is None:
+            return base.merged(update)
+        if base.error_after is None and update.error_after is not None:
+            return update
+        if update is None:
+            return None
         return base.merged(update)
     elif base is None and update is not None:
         return update
--- a/core/dbt/contracts/graph/unparsed.py
+++ b/core/dbt/contracts/graph/unparsed.py
@@ -162,20 +162,22 @@ class TimePeriod(StrEnum):
 
 
 @dataclass
-class Time(dbtClassMixin, Replaceable):
-    count: int
-    period: TimePeriod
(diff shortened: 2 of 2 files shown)


kadero:
Hello @jtcohen6 @lchoward :wave:,

Here is a first proposition based on what we agreed on [the initial issue](https://github.com/dbt-labs/dbt/issues/3874).

For now, the changes reflect what we agreed [here](https://github.com/dbt-labs/dbt/issues/3874#issuecomment-922546181), and the tests were fixed naively.

Before going further and adding new tests (units and integrations), I would like your thought on the current implementation

Many thanks :pray:

leahwicz:
@gshank @nathaniel-may this falls into your domain so would you be able to take a look at this please and see what you think?

core/dbt/parser/sources.py
```diff
@@ -344,11 +345,32 @@ def get_unused_msg(
...
+) -> Optional[Time]:
+    if base is not None and update is not None:
+        return base.merged(update)
+    elif base is None and update is not None:
+        return update
+    elif base is not None and update is None:
+        return base
+    else:
+        # base and update are none
+        return None
```
nathaniel-may, line 359: Here's a more terse way to express this function that leans more into EAFP:

```suggestion
    try:
        return base.merged(update)
    except AttributeError:
        return update or base
```

nathaniel-may (commented):
The Python looks good to me and I only had one minor suggestion. I'm not deeply familiar with this corner of the code base yet so I'm going to withhold an explicit approval for now. Happy to do a more time consuming deep dive if we need to though.

PR #3959, opened 2021-09-27
Er/3899 models key mismatch

emmyoop:
resolves # 3899

### Description

Give a warning when a node is specified under the wrong type in `schema.yml` and don't continue processing.

Does this still need to be handled - I couldn't produce the behavior: Throw an exception for any model/seed/snapshot that has a `unique_id` of `None` and is not disabled.

As part of this change `manifest._disabled` was also removed in favor of the existing `manifest.disabled`.

### Checklist

### DIFF
--- a/core/dbt/contracts/graph/manifest.py
+++ b/core/dbt/contracts/graph/manifest.py
@@ -1008,7 +1008,6 @@ class Manifest(MacroMethods, DataClassMessagePackMixin, dbtClassMixin):
     def add_disabled_nofile(self, node: CompileResultNode):
         self.disabled.append(node)
 
-
     def add_disabled(self, source_file: AnySourceFile, node: CompileResultNode, test_from=None):
         self.add_disabled_nofile(node)
         if isinstance(source_file, SchemaSourceFile):
--- a/core/dbt/contracts/graph/manifest.py
+++ b/core/dbt/contracts/graph/manifest.py
@@ -567,7 +567,6 @@ class Manifest(MacroMethods, DataClassMessagePackMixin, dbtClassMixin):
     # Moved from the ParseResult object
     source_patches: MutableMapping[SourceKey, SourcePatch] = field(default_factory=dict)
     # following is from ParseResult
-    _disabled: MutableMapping[str, List[CompileResultNode]] = field(default_factory=dict)
     _doc_lookup: Optional[DocLookup] = field(
         default=None, metadata={'serialize': lambda x: None, 'deserialize': lambda x: None}
     )
@@ -1007,10 +1006,8 @@ class Manifest(MacroMethods, DataClassMessagePackMixin, dbtClassMixin):
         source_file.exposures.append(exposure.unique_id)
 
     def add_disabled_nofile(self, node: CompileResultNode):
-        if node.unique_id in self._disabled:
-            self._disabled[node.unique_id].append(node)
-        else:
-            self._disabled[node.unique_id] = [node]
+        self.disabled.append(node)
+
 
     def add_disabled(self, source_file: AnySourceFile, node: CompileResultNode, test_from=None):
         self.add_disabled_nofile(node)
@@ -1048,7 +1045,6 @@ class Manifest(MacroMethods, DataClassMessagePackMixin, dbtClassMixin):
             self.flat_graph,
             self.state_check,
             self.source_patches,
-            self._disabled,
             self._doc_lookup,
             self._source_lookup,
             self._ref_lookup,
--- a/core/dbt/contracts/graph/manifest.py
+++ b/core/dbt/contracts/graph/manifest.py
@@ -404,7 +404,6 @@ class NameSearcher(Generic[N]):
 
     def search(self, haystack) -> Optional[N]:
         """Find an entry in the given iterable by name."""
-        # breakpoint()
         for model in haystack.values():
             if self._matches(model):
                 return model
@@ -738,7 +737,7 @@ class Manifest(MacroMethods, DataClassMessagePackMixin, dbtClassMixin):
             exposures={k: _deepcopy(v) for k, v in self.exposures.items()},
             selectors={k: _deepcopy(v) for k, v in self.selectors.items()},
             metadata=self.metadata,
-            disabled=[_deepcopy(n) for n in self.disabled],
+            disabled={k: _deepcopy(v) for k, v in self.disabled.items()},
             files={k: _deepcopy(v) for k, v in self.files.items()},
             state_check=_deepcopy(self.state_check),
         )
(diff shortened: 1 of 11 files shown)


core/dbt/parser/schemas.py
```diff
@@ -816,6 +816,24 @@ def parse_patch(
         source_file: SchemaSourceFile = self.yaml.file
         if patch.yaml_key in ['models', 'seeds', 'snapshots']:
             unique_id = self.manifest.ref_lookup.get_unique_id(patch.name, None)
+            if unique_id:
+                resource_type = unique_id.split('.')[0]
+                if (unique_id.split('.')[0] + 's') != patch.yaml_key:
```
jtcohen6, line 821: We need to use [`NodeType.pluralize()`](https://github.com/dbt-labs/dbt/blob/6854e6746406d87e3e800d7e535c85aac5bf6d2e/core/dbt/node_types.py#L52) here, primarily because the plural of `analysis` is `analyses`.

Could we do something like:
```suggestion
                resource_type = NodeType(unique_id.split('.')[0])
                if resource_type.pluralize() != patch.yaml_key:
```

As a guide, [here's](https://github.com/dbt-labs/dbt/blob/745ae3e64ac03fedb19228b31252fea05eb0ed37/core/dbt/contracts/graph/manifest.py#L685-L695) the code in `0.19.latest`... In fact, it looks like we already have an exception for this: [`raise_invalid_patch`](https://github.com/dbt-labs/dbt/blob/6854e6746406d87e3e800d7e535c85aac5bf6d2e/core/dbt/exceptions.py#L927-L942). Perhaps we can turn that into a `warn_or_error`?

core/dbt/parser/schemas.py
```diff
@@ -816,6 +816,24 @@ def parse_patch(
...
+                        f"under the {resource_type} key instead.")
+
+                    return
+
+            else:
+                if not self.manifest.find_disabled_by_name(patch.name, None):
+                    # need to raise an exception when a refable node does not
+                    # have a unique id and is not disabled
+                    # todo: do we still want to do this?
+                    raise raise_no_unique_id(patch)
```
jtcohen6, line 836: I don't believe we raise an exception for this today, so I don't think we should start. Unfortunately, this is likely to break a lot of projects, and prevent them from upgrading to v1.0 :(

jtcohen6 (commented):
Nice work pulling on the thread for this one. I left a few small comments

core/dbt/exceptions.py
```diff
@@ -849,6 +849,15 @@ def raise_patch_targets_not_found(patches):
     )
 
 
+# todo: fix error message to be useful & add test
+def raise_no_unique_id(patch):
```
jtcohen6, line 853: If we're not calling this anywhere, I think we can avoid adding it / causing future confusion

core/dbt/exceptions.py
```diff
@@ -924,22 +933,20 @@ def raise_unrecognized_credentials_type(typename, supported_types):
     )
 
 
-def raise_invalid_patch(
-    node, patch_section: str, patch_path: str,
-) -> NoReturn:
+def warn_invalid_patch(patch, resource_type):
```
jtcohen6, line 936: Nice! I'm just going to leave a few cosmetic suggestions below

core/dbt/exceptions.py
```diff
@@ -924,22 +933,20 @@ def raise_unrecognized_credentials_type(typename, supported_types):
...
-        {patch_path}.
+        '{patch.name}' is a {resource_type} node, but it is
+        specified in the {patch.yaml_key} section of
+        {patch.original_file_path}.
 
 
 
-        To fix this error, place the `{node.name}`
-        specification under the {node.resource_type.pluralize()} key instead.
+        To fix this error, place the `{patch.name}`
```
jtcohen6, line 945:
```suggestion
        {patch.original_file_path}.
        To fix this error, place the `{patch.name}`

Comment (jtcohen6) [inline core/dbt/exceptions.py:945]:
```python
        {patch.original_file_path}. To fix this error, place the `{patch.name}
```
(no line breaks)

core/dbt/exceptions.py
```diff
@@ -924,22 +933,20 @@ def raise_unrecognized_credentials_type(typename, supported_types):
...
+        specification under the {resource_type.pluralize()} key instead.
         '''
     )
-    raise_compiler_error(msg, node)
+    warn_or_error(msg)
```
jtcohen6, line 949:
```suggestion

    warn_or_error(msg, log_fmt=warning_tag('{}'))

File: core/dbt/contracts/graph/manifest.py [hunk 7]
```diff
@@ -1007,10 +1006,7 @@ def add_exposure(self, source_file: SchemaSourceFile, exposure: ParsedExposure):
         source_file.exposures.append(exposure.unique_id)
 
     def add_disabled_nofile(self, node: CompileResultNode):
-        if node.unique_id in self._disabled:
-            self._disabled[node.unique_id].append(node)
-        else:
-            self._disabled[node.unique_id] = [node]
+        self.disabled.append(node)
```
jtcohen6, line 1009: What's changed here? Are we no longer patching properties for disabled nodes?

jtcohen6 (commented):
Nice work! Feels really close

core/dbt/contracts/graph/manifest.py
gshank: I don't think we've ever patched disabled nodes. That's the way that it worked when I first saw it anyway.

gshank (changes requested):

core/dbt/contracts/graph/manifest.py
gshank: I'm not sure that feels right though. I woudn't have no objection to patching disabled nodes. They do come out in the manifest.json.

core/dbt/contracts/graph/manifest.py
```diff
@@ -567,7 +567,6 @@ class Manifest(MacroMethods, DataClassMessagePackMixin, dbtClassMixin):
     # Moved from the ParseResult object
     source_patches: MutableMapping[SourceKey, SourcePatch] = field(default_factory=dict)
     # following is from ParseResult
-    _disabled: MutableMapping[str, List[CompileResultNode]] = field(default_factory=dict)
```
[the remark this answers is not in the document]
gshank: @jtcohen6  Disabled does show up in manifest.json, so this would be a manifest schema change. Do you agree with what I propose here? If we're going to do it, for 1.0 is the right time.

core/dbt/contracts/graph/manifest.py
gshank: But a lookup won't find them (since they're not included there), so we'd have to separately look them up in disabled.

core/dbt/contracts/graph/manifest.py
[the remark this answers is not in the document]
jtcohen6: Your proposal makes sense to me! We should feel well within our rights to make these kind of schema changes ahead of 1.0

gshank (approved):
Looks good!

PR #3970, opened 2021-09-29
Add Jinja Sampling to Stable Static Parser

nathaniel-may:
[DRAFT]
- [ ] Finish code
- [ ] Point at develop after [PR#3939](https://github.com/dbt-labs/dbt/pull/3939) is merged

### Description

Infrequently sample jinja parsing to compare against the new static parser. Without this, we'll have no data to compare the parity of jinja rendering and the stable static parser for the 1.0 version. This sampling is separate from the sampling pathway for the experimental parser, and both can be done at the same time. Since we expect a huge number of people to upgrade to 1.0, even though we have done our due diligence and expect the static parser to match parity with jinja rendering, this tracking information will allow us to validate our prior work while the feature is turned on by default for users.

The trade off here is that adding a second sampling path adds another bit of complexity to the `render_update` method and jinja rendering is often two orders of magnitude greater than the static parser making sampling extremely expensive. Right now the plan is to sample 1/5000 model files instead of 1/100 but this sample rate is flexible. A lower rate gives us less feedback and impacts users less, and a higher rate gives us more feedback sooner at a performance cost to users.

### Reviewers

In order to jinja render separately, I deep copy both the `node` and `config` objects and pass them to the super class's `render_update` method. What do you think about this approach? I'm trying to avoid mutating the objects I'm working on currently for correctness, but I'm also not using the jinja rendering for the run when I have it so that subsequent user runs of dbt are deterministic and not dependent on the sampling rng.

There's no great way to test the tracking logic, so please take a good look at `_get_stable_sample_result`. 

### Checklist


core/dbt/parser/models.py
```diff
@@ -288,3 +322,54 @@ def _get_sample_result(
...
+    sample_node: ParsedModelNode,
+    sample_config: ContextConfig,
+    node: ParsedModelNode,
+    config: ContextConfig
+) -> List[str]:
+    result: List[str] = []
+    # look for false positive configs
+    for k in config._config_call_dict:
+        if k not in config._config_call_dict:
+            result += ["82_stable_false_positive_config_value"]
```
jtcohen6, line 338: ah, a different error code! really good. I was way overcomplicating this in my head

nathaniel-may: I try to make things easy for you when I can 🙂

jtcohen6:
Small thing, while we're hanging out in the model parser's `render_update`: I just realized that this line got lost in the shuffle (or did we remove it intentionally?):

https://github.com/dbt-labs/dbt/blob/733f35f5e133327e8e54477e94166ec76eec926e/core/dbt/parser/models.py#L32

We [include that value](https://github.com/dbt-labs/dbt/blob/38eb46dfc32525c5834d7b092a64456eaccb18df/core/dbt/parser/manifest.py#L102) when we fire off the total # of files "eligible" for static parsing, as the denominator when calculating the efficacy of our current grammar for one project.

nathaniel-may:
@jtcohen6, It's definitely not the plan to remove that line. It looks like it's still there on [develop](https://github.com/dbt-labs/dbt/blob/c25b7a1143a2285ba82c18ffef7dec5c0f8c480b/core/dbt/parser/models.py#L90) and in [this branch](https://github.com/dbt-labs/dbt/blob/ca69e842579a60d54c9f42d7c9fe7ce779daa48d/core/dbt/parser/models.py#L104). So I suspect it's just the PR view that makes it look like it disappeared.

jtcohen6:
@nathaniel-may That's `static_analysis_parsed_path_count` (numerator), which is distinct from `static_analysis_path_count` (denominator). I think `static_analysis_path_count` was incidentally removed in https://github.com/dbt-labs/dbt/commit/39447055d31dc05b754cf97f2ff445de28dff4dc#diff-dc357e8548e098dd9f293a19d206d90baf4796d1f2e05e237342334fd6657356L32.

I observed this because I parsed a project with `develop`, inspected `target/perf_info.json`, and noticed that `static_analysis_path_count` is 0, at the same time that `static_analysis_parsed_path_count` is >0 and `is_static_analysis_enabled` is True.

nathaniel-may:
My mistake. Good catch though! I opened [PR#3981](https://github.com/dbt-labs/dbt/pull/3981).

nathaniel-may:
rebased so PR#3981 is reflected here now.

gshank (approved):

PR #3972, opened 2021-09-29
Scrub strings, not AttributeErrors

### LINKED ISSUE #3971: Nasty error with log scrubbing in v0.21.0rc2
In the dbt Cloud IDE, prefixed env var log scrubbing is failing with:
```
'AttributeError' object has no attribute 'replace'
```

The processor code is:
https://github.com/dbt-labs/dbt/blob/571beb13d91d9e0c065082f6a6d9f0293666b782/core/dbt/logger.py#L345-L348

I have _not_ been able to reproduce this locally using the RPC server. There are half a dozen or so places in dbt-core where we raise an `AttributeError`, and half a dozen more where we try to handle one. I'm not sure if it's any of those.

That said, I think we might just be able to add handling for this, to avoid tripping this particular error, by wrapping the message in `str()` first. That's exactly what we do in the other log processor where we call `.replace()`:

https://github.com/dbt-labs/dbt/blob/571beb13d91d9e0c065082f6a6d9f0293666b782/core/dbt/logger.py#L459-L464

jtcohen6:
resolves #3971

As soon as this is merged into `develop` and dbt Cloud is redeployed, we can test this there using the `HEAD` version of dbt. If it works, I'd feel comfortable cherry-picking onto `0.21.latest` and including in v0.21.0 (final), without another RC.

### Checklist

### DIFF
--- a/core/dbt/logger.py
+++ b/core/dbt/logger.py
@@ -344,9 +344,8 @@ class TimestampNamed(logbook.Processor):
 
 class ScrubSecrets(logbook.Processor):
     def process(self, record):
-        subbed = str(record.message)
         for secret in get_secret_env():
-            record.message = subbed.replace(secret, "*****")
+            record.message = str(record.message).replace(secret, "*****")
 
 
 logger = logbook.Logger('dbt')
--- a/core/dbt/logger.py
+++ b/core/dbt/logger.py
@@ -344,8 +344,9 @@ class TimestampNamed(logbook.Processor):
 
 class ScrubSecrets(logbook.Processor):
     def process(self, record):
+        subbed = str(record.message)
         for secret in get_secret_env():
-            record.message = record.message.replace(secret, "*****")
+            record.message = subbed.replace(secret, "*****")
 
 
 logger = logbook.Logger('dbt')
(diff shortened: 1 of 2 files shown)


core/dbt/logger.py
```diff
@@ -344,8 +344,9 @@ def process(self, record):
 
 class ScrubSecrets(logbook.Processor):
     def process(self, record):
+        subbed = str(record.message)
         for secret in get_secret_env():
-            record.message = record.message.replace(secret, "*****")
+            record.message = subbed.replace(secret, "*****")
```
jtcohen6, line 349: definitely open to a smarter / more pythonic way to do this!

gshank: Looks fine to me, but I'm no expert and phyonic-ness.

gshank (approved):

core/dbt/logger.py
ryancharris: FWIW the DBT_ENV_SECRET vars are prepended with `********`

iknox-fa: It's perfectly reasonable IMO.. If I was being terribly pedantic I might suggest skipping assigning the stringified value to a var like so:

```
for secret in get_secret_env():
            record.message = str(record.message).replace(secret, "*****")
            
```

nathaniel-may: I agree with @iknox-fa and would like to skip the assignment. But also won't block merging on that.

nathaniel-may (approved):
LGTM

nathaniel-may:
I'm guessing we're not running mypy somewhere important because it looks like it should be asserted here: https://github.com/dbt-labs/dbt/blob/df30bf1135dc159acb87e9e92fc11105079a030d/core/dbt/logger.py#L62

jtcohen6:
> I'm guessing we're not running mypy somewhere important

yeah... https://github.com/dbt-labs/dbt/issues/3203

PR #3982, opened 2021-09-30
[#3810] Skip partial parsing if certain macros have changed

### LINKED ISSUE #3810: Trigger full reparse if changes to widely used built-in macros
There is a special category of macro that:
- dbt uses at parse time, across almost all resources in the project
- is not explicitly called within project code, such that its dependency is not grabbed during static analysis; or is excluded from `depends_on.macros` because of being so widely used

Those macros are:
- `ref`
- `source`
- `config`
- `execute` (?)
- `generate_schema_name`
- `generate_database_name`
- `generate_alias_name`

The [builtins](https://docs.getdbt.com/reference/dbt-jinja-functions/builtins) are fairly self-explanatory (though `execute` is a weird inclusion in those docs).

As for `generate_x_name` macros: The code that uses these macros is _incredibly_ thorny, pretty old, and I think it has to do with a method named `generate_generate_component_name_macro`. I don't mind hard-coding that these seven need special treatment, and that any detected changes to them should be grounds for triggering a full re-parse (with logging + tracking to show for it).

Opened as a bug because this currently yields incorrect results :)

gshank:
resolves #3810

### Description

Checks for changed or new macros for ref, config, source, and the generate_ macros, and switches to a full re-parse.

### Checklist

### DIFF
--- a/core/dbt/parser/manifest.py
+++ b/core/dbt/parser/manifest.py
@@ -29,7 +29,7 @@ from dbt.context.configured import generate_macro_context
 from dbt.context.providers import ParseProvider
 from dbt.contracts.files import FileHash, ParseFileType, SchemaSourceFile
 from dbt.parser.read_files import read_files, load_source_file
-from dbt.parser.partial import PartialParsing
+from dbt.parser.partial import PartialParsing, special_override_macros
 from dbt.contracts.graph.compiled import ManifestNode
 from dbt.contracts.graph.manifest import (
     Manifest, Disabled, MacroManifest, ManifestStateCheck, ParsingInfo
@@ -130,7 +130,9 @@ class ManifestLoader:
         self.root_project: RuntimeConfig = root_project
         self.all_projects: Mapping[str, Project] = all_projects
         self.manifest: Manifest = Manifest()
+        self.new_manifest = self.manifest
         self.manifest.metadata = root_project.get_metadata()
+        self.macro_resolver = None  # built after macros are loaded
         self.started_at = int(time.time())
         # This is a MacroQueryStringSetter callable, which is called
         # later after we set the MacroManifest in the adapter. It sets
@@ -149,6 +151,7 @@ class ManifestLoader:
         # We need to know if we're actually partially parsing. It could
         # have been enabled, but not happening because of some issue.
         self.partially_parsing = False
+        self.partial_parser = None
 
         # This is a saved manifest from a previous run that's used for partial parsing
         self.saved_manifest: Optional[Manifest] = self.read_manifest_for_partial_parse()
@@ -207,13 +210,14 @@ class ManifestLoader:
             saved_files = self.saved_manifest.files
         for project in self.all_projects.values():
             read_files(project, self.manifest.files, project_parser_files, saved_files)
+        orig_project_parser_files = project_parser_files
         self._perf_info.path_count = len(self.manifest.files)
         self._perf_info.read_files_elapsed = (time.perf_counter() - start_read_files)
 
         skip_parsing = False
         if self.saved_manifest is not None:
-            partial_parsing = PartialParsing(self.saved_manifest, self.manifest.files)
-            skip_parsing = partial_parsing.skip_parsing()
+            self.partial_parser = PartialParsing(self.saved_manifest, self.manifest.files)
+            skip_parsing = self.partial_parser.skip_parsing()
             if skip_parsing:
                 # nothing changed, so we don't need to generate project_parser_files
                 self.manifest = self.saved_manifest
@@ -223,7 +227,7 @@ class ManifestLoader:
                 # files are different, we need to create a new set of
                 # project_parser_files.
                 try:
-                    project_parser_files = partial_parsing.get_parsing_files()
+                    project_parser_files = self.partial_parser.get_parsing_files()
                     self.partially_parsing = True
                     self.manifest = self.saved_manifest
                 except Exception:
@@ -245,7 +249,7 @@ class ManifestLoader:
 
                     # get file info for local logs
                     parse_file_type = None
-                    file_id = partial_parsing.processing_file
+                    file_id = self.partial_parser.processing_file
                     if file_id and file_id in self.manifest.files:
                         old_file = self.manifest.files[file_id]
                         parse_file_type = old_file.parse_file_type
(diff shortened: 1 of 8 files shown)


core/dbt/parser/manifest.py
```diff
@@ -270,20 +274,17 @@ def load(self):
...
-                    parser.parse_file(block)
-                    # increment parsed path count for performance tracking
-                    self._perf_info.parsed_path_count = self._perf_info.parsed_path_count + 1
-            # Look at changed macros and update the macro.depends_on.macros
-            self.macro_depends_on()
+            self.load_and_parse_macros(project_parser_files)
+
+            # If we're partially parsing check that certain macros have not been changed
+            if self.partially_parsing and self.skip_partial_parsing_because_of_macros():
+                logger.info("Partial parsing enabled, but skipping due to macro changes.")
```
jtcohen6, line 281: Would it be possible to include the name of the changed/added/deleted macro in this log message? I'm thinking something like:
```
Change detected to macro generate_schema_name. Starting full parse.
```
Or if we cannot tell:
```
Change detected to override macro used during parsing. Starting full parse.
```

jtcohen6: I think this is also the right spot to fire off a `track_partial_parser` event with a new `ReparseReason` (something like `special_override_macro`)

jtcohen6 (commented):
Looking good! Just a comment about the log message and tracking

core/dbt/parser/manifest.py
```diff
@@ -440,20 +459,23 @@ def parse_project(
...
-    # Loop through macros in the manifest and statically parse
-    # the 'macro_sql' to find depends_on.macros
-    def macro_depends_on(self):
+    # This should only be called after the macros have been loaded
+    def build_macro_resolver(self):
         internal_package_names = get_adapter_package_names(
             self.root_project.credentials.type
         )
-        macro_resolver = MacroResolver(
+        self.macro_resolver = MacroResolver(
```
nathaniel-may, line 467: I don't think I completely understand this change. Why didn't we set `self.macro_resolver` before, and why are we doing it now?

core/dbt/parser/manifest.py
```diff
@@ -270,20 +274,17 @@ def load(self):
...
-                    block = FileBlock(self.manifest.files[file_id])
-                    parser.parse_file(block)
-                    # increment parsed path count for performance tracking
-                    self._perf_info.parsed_path_count = self._perf_info.parsed_path_count + 1
-            # Look at changed macros and update the macro.depends_on.macros
-            self.macro_depends_on()
+            self.load_and_parse_macros(project_parser_files)
+
+            # If we're partially parsing check that certain macros have not been changed
+            if self.partially_parsing and self.skip_partial_parsing_because_of_macros():
```
nathaniel-may, line 280: We're checking if we should skip partial parsing here, but we also check above. Can we merge them into one check? I like the interface above where we just have one clear check.

core/dbt/parser/manifest.py
```diff
@@ -207,13 +210,14 @@ def load(self):
...
         self._perf_info.path_count = len(self.manifest.files)
         self._perf_info.read_files_elapsed = (time.perf_counter() - start_read_files)
 
         skip_parsing = False
         if self.saved_manifest is not None:
-            partial_parsing = PartialParsing(self.saved_manifest, self.manifest.files)
-            skip_parsing = partial_parsing.skip_parsing()
+            self.partial_parser = PartialParsing(self.saved_manifest, self.manifest.files)
+            skip_parsing = self.partial_parser.skip_parsing()
             if skip_parsing:
```
nathaniel-may, line 221: I really like the way this reads. There's another skip check down below for these new macro overrides. Can we fold that check inside `skip_parsing()`? Reading here, I feel like if `skip_parsing` is `False` we know it is safe to partial parse.

nathaniel-may (changes requested):
For the most part this looks pretty good! I requested some additional test coverage, and asked some questions about if we can make some parts a bit more streamlined.

core/dbt/parser/manifest.py
gshank: Technically we don't actually know that a particular macro has changed, we know that the file that contains that macro has changed. I'm a little afraid that if we specify the macro by name we'd get people complaining that they hadn't changed that macro... Should I put the macro file name in the message instead?

test/integration/068_partial_parsing_tests/test_partial_parsing.py
```diff
@@ -468,3 +472,33 @@ def test_postgres_nested_macros(self):
...
+    @use_profile('postgres')
+    def test_postgres_skip_macros(self):
+
+        # initial run so we have a msgpack file
+        results = self.run_dbt()
+
+        # add a new ref override macro
+        shutil.copyfile('extra-files/ref_override.sql', 'macros-macros/ref_override.sql')
+        results, log_output = self.run_dbt_and_capture(['--partial-parse', 'run'])
+        self.assertTrue('skipping due to macro changes' in log_output)
```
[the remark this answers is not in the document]
gshank: We don't have numerical codes for any of those error messages at this point. I don't think it makes sense to do just one of them, and this is not the appropriate time to add codes to all of them.

core/dbt/parser/partial.py
```diff
@@ -32,10 +33,21 @@
...
+]
+
+
 # Partial parsing. Create a diff of files from saved manifest and current
 # files and produce a project_parser_file dictionary to drive parsing of
 # only the necessary changes.
 # Will produce a 'skip_parsing' method, and a project_parser_file dictionary
+# All file objects from the new manifest are deepcopied, because we need
+# to preserve an unchanged file object in case we need to drop back to a
+# a full parse (such as for certain macro changes)
```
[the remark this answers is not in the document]
gshank: We're copying the sourcefile. We need to leave all of the information in the file objects unchanged because they need to be unchanged if we need to restore the the originally read files because of aborting the partial parse. The other alternative would be to use to_dict() and restore from a dictionary, but to_dict() on files doesn't have the contents.

core/dbt/parser/manifest.py
gshank: Because before this we only used the macro_resolver in one place in the ManifestLoader and now we're using it in two places.

core/dbt/parser/manifest.py
gshank: No, we can't merge them into one place. The new place to skip requires that the macros be loaded, which they aren't in the first place. Yes, this makes the code more complicated but there's no way around it if we want to do this.

test/integration/068_partial_parsing_tests/test_partial_parsing.py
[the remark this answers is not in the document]
nathaniel-may: alright. we'll just change it later.

core/dbt/parser/partial.py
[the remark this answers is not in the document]
nathaniel-may: What's the worst case scenario for this then? Do we end up deep copying all source files in a project?

core/dbt/parser/manifest.py
nathaniel-may: Okay figured I'd check.

test/integration/068_partial_parsing_tests/extra-files/ref_override2.sql
```diff
@@ -0,0 +1,4 @@
+- Macro to override ref xxxx
+{% macro ref(modelname) %}
```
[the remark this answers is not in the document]
gshank: I've added a test to check that the special_override_macros are correct. We don't need to test each name individually because for the purposes of this code there are only two variations and both are tested.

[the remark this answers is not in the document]
nathaniel-may: Ok checking that `special_override_macros` is what we expect is good enough for me.

core/dbt/parser/manifest.py
jtcohen6: Got it. I like the idea of including the file name. This sounds a bit clunky, but it feels closest to the meaning we want to get across:
```
Change detected in {file_name}, which defines a macro widely used in parsing. Starting full parse.
```

core/dbt/parser/partial.py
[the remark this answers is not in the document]
nathaniel-may: Talked with @gshank and it looks like these calls to `deepcopy` only happen on the changed files so there won't be very many calls to them. I'd like to be able to bypass these copies and trigger a reparse directly after parsing macros, but there's enough mutation to the nodes to schedule parsing that we can't easily shortcut at that stage.

To me, it looks like the performance impact here is acceptable since it is proportional to the number of files changed which should be expected for users.

nathaniel-may (approved):

PR #4015, opened 2021-10-06
Creating version bump action

leahwicz:
### Description

This creates an Action that bumps the version and Docker requirement file dependencies 

### Checklist

### DIFF
--- a/.github/workflows/version-bump.yml
+++ b/.github/workflows/version-bump.yml
@@ -1,0 +1,106 @@
+# **what?**
+# This workflow will take a version number and a dry run flag. With that
+# it will run versionbump to update the version number everywhere in the 
+# code base and then generate an update Docker requirements file. If this
+# is a dry run, a draft PR will open with the changes. If this isn't a dry
+# run, the changes will be committed to the branch this is run on.
+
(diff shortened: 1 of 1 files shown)


.github/workflows/version-bump.yml
```diff
@@ -0,0 +1,106 @@
...
+  workflow_dispatch:
+  repository_dispatch:
+    types: [version-bump]
+  
+# Default version number and dry run setting in case we have to run this
+# Action manually and not from a repository_dispatch event.
+# These should be updated if that is the case.
+env:
+  version_number_default: '0.22.0'
+  is_dry_run_default: true
```
kwigley, line 29: Instead of codifying this, can we just use inputs for the manual workflow dispatch trigger?
https://docs.github.com/en/actions/learn-github-actions/events-that-trigger-workflows#example-workflow-configuration

the trigger above would look something like:
```
on:
  workflow_dispatch:
    inputs:
      version_number:
        required: true
      is_dry_run:
        required: false
        default: 'true'
```

kwigley (approved):
Looks good! Left a suggestion, but approved otherwise 👍

PR #4048, opened 2021-10-12
Fix tracking bug for jinja sampling

### LINKED ISSUE #4038: [Static parser] hook ref missing from sample during comparison
When a model includes a novel `ref()` inside a project-configured hook, the static parser works correctly, but the comparison for sampling/tracking does not.

### Reproduction case
```sql
-- models/model_a.sql
select 1 as id
```
```sql
-- models/subdir/model_b.sql
```
```sql
-- models/model_c.sql
select 1 as id
```
```sql
-- macros/ref_my_other_model.sql
{% macro ref_my_other_model() %}
  select * from {{ ref('model_c') }}
{% endmacro %}
```
```yml
# dbt_project.yml
models:
  <project_name>:
    subdir:
      +post-hook: "{{ ref_my_other_model() }}"
```

We get the correct results in `node.refs`, both with and without the experimental parser enabled:

```bash
$ dbt --no-partial-parse ls -s model_b --output json --output-keys refs
{"refs": [["model_a"], ["model_c"]]}
```
```bash
$ dbt --use-experimental-parser  --no-partial-parse ls -s model_b --output json --output-keys refs
{"refs": [["model_c"], ["model_a"]]}
```

### Diagnostic

The issue is when the tracking takes place. In order to have an apples-to-apples comparison between the static parser output and the Jinja-rendered output, we'd need some way to do this step _before_ the sample comparison:

https://github.com/dbt-labs/dbt/blob/c009485de223cb86eab2a1db201c02cfea0ceaa2/core/dbt/parser/models.py#L110-L113

Whereas, currently, we perform the comparison using _only_ the output of static extraction:

https://github.com/dbt-labs/dbt/blob/c009485de223cb86eab2a1db201c02cfea0ceaa2/core/dbt/parser/models.py#L216-L220

nathaniel-may:
resolves #4038

### Description
Comparison for tracking events should be done before we apply any changes. See ticket description for details.

### Reviewers
I just reordered existing code so that the sample comparison is the first thing that's done in the function instead of the last.

### Checklist

### DIFF
--- a/core/dbt/parser/models.py
+++ b/core/dbt/parser/models.py
@@ -106,7 +106,7 @@ class ModelParser(SimpleSQLParser[ParsedModelNode]):
             # This means we skip sampling for 40% of the 1/5000 samples. We could run the
             # sampling rng here, but the effect would be the same since we would only roll
             # it 40% of the time. So I've opted to keep all the rng code colocated above.
-            if stable_sample:
+            if stable_sample and not flags.USE_EXPERIMENTAL_PARSER:
                 logger.debug(f"1611: conducting full jinja rendering sample on {node.path}")
                 # if this will _never_ mutate anything `self` we could avoid these deep copies,
                 # but we can't really guarantee that going forward.


nathaniel-may:
Chatted with @jtcohen6 and we figured out that we actually want to be doing this comparison on fully formed nodes. The experimental parser will likely be attempting to expand the capacity of the existing static parser so we cannot compare the two together naively. Instead will have to populate a complete copy of the node with everything after experimentally parsing and check if they match. Thankfully, this is a much sounder way to approach this comparison anyway.

PR #4070, opened 2021-10-15
fixed error_id typo and added associated tests

### LINKED ISSUE #4051: [Bug] test not_null: error_if not working

### Current Behavior

I have a test within a model that has the following syntax:
- name: column name
   description: description
   tests:
     - not_null:
           - error_if: '>1'
           
This test was working okay in the 0.20 version but when I upgraded to the 0.21 version when I run `dbt test` it fails with the following error:
`Got 1 result, configured to fail if  != 0`

### Expected Behavior

The expected behaviour is that the test only fails when there is more than one null value in the model, that is to say, that the test allows only one null value.

### Steps To Reproduce

1. Have a model with a source that has a filed with one null value
2. Have a test for `not_null`
3. Add the `error_if: '>1'`  to that test
4. run `dbt test`
5. If you have none or 1 null value in that column the test should pass, else if you have more than 1 null, the test should throw an error.

### Relevant log output

```shell
Got 1 result, configured to fail if  != 0
```

### Environment

```markdown
- OS:macOS Big Sur
- Python: 3.8.8
- dbt: 0.21.0
```

bigquery

### Additional Context

_No response_

emmyoop:
resolves #4051

### Description

- fixed bug where `error_id` existed instead of `error_if`
- added tests around `error_if`
- removed extra `alias` check

### Checklist


emmyoop:
Will need to be backported to `0.21.latest`

PR #4114, opened 2021-10-22
Add project name to default search packages

### LINKED ISSUE #4106: [FEATURE] Set all macro namespace by default to `['<name_of_your_root_project>', '<macro_namespace>']`
As suggested by @jtcohen6 in [comment](https://github.com/dbt-labs/dbt-core/issues/4098#issuecomment-947678409), change this https://github.com/dbt-labs/dbt-core/blob/280e9ad9c9bf914b5bbd1aa14629a6708c921546/core/dbt/context/providers.py#L146-L147

to:

``` python
            if not search_packages and namespace in self._adapter.config.dependencies:
                search_packages = [self.config.project_name, namespace]
```

With this change, by default we prefer macro overwrites for packages (like `spark__create_table_as`) defined in the root project over the ones defined in the package. The user can change the preference behavior by defining the search order explicitly:

``` yml
# dbt_project.yml
dispatch:
  - macro_namespace: dbt
    search_order: ['<package name>, '<root project name>']
```

In the old behavior the macros for packages in the root folder are not preferred over the ones in the packages, needing the user to explicitly set the search order to:

``` yml
# dbt_project.yml
dispatch:
  - macro_namespace: dbt
    search_order: ['<root project name', '<package name>]
```

We think that the proposed change is a more sensible default.

JCZuurmond:
We prefer macros in the project over the ones in the namespace (package)

resolves #4106 

### Description

This changes will choose macros in the root project over the ones in the namespace.

### Checklist

### DIFF
--- a/test/integration/016_macro_tests/test_macros.py
+++ b/test/integration/016_macro_tests/test_macros.py
@@ -119,6 +119,11 @@ class TestMacroOverrideBuiltin(DBTIntegrationTest):
 
 
 class TestMacroOverridePackage(DBTIntegrationTest):
+    """
+    The macro in `override-postgres-get-columns-macros` should override the
+    `get_columns_in_relation` macro by default.
+    """
+
     @property
     def schema(self):
         return "test_macros_016"
@@ -142,6 +147,12 @@ class TestMacroOverridePackage(DBTIntegrationTest):
 
 
 class TestMacroNotOverridePackage(DBTIntegrationTest):
+    """
+    The macro in `override-postgres-get-columns-macros` does NOT override the
+    `get_columns_in_relation` macro because we tell dispatch to not look at the
+    postgres macros.
+    """
+
     @property
     def schema(self):
         return "test_macros_016"


test/integration/016_macro_tests/test_macros.py
```diff
@@ -118,6 +118,64 @@ def test_postgres_overrides(self):
...
+            'config-version': 2,
+            'macro-paths': ['override-postgres-get-columns-macros'],
+            'dispatch': [{'macro_namespace': 'dbt', 'search_order': ['dbt']}],
+        }
+
+    @use_profile('postgres')
+    def test_postgres_overrides(self):
+        # the first time, the model doesn't exist
+        self.run_dbt(expect_pass=False)
+        self.run_dbt(expect_pass=False)
```
chapmanjacobd, line 176: should the second time be expect pass True ?

jtcohen6: There's a check in the model that raises an explicit error if the macro is overridden, so I believe it fails both times:
https://github.com/dbt-labs/dbt-core/blob/f20e83a32b99540f15d1b8ab051a781f03243d2d/test/integration/016_macro_tests/override-get-columns-models/model.sql#L1-L5

JCZuurmond: As explained in the docstring class, the run in this test should fail because the override does **not** happen. We need the override to overrule the raise compile error that @jtcohen6  showed.

```
    The macro in `override-postgres-get-columns-macros` does NOT override the
    `get_columns_in_relation` macro because we tell dispatch to not look at the
    postgres macros.
```

PR #4133, opened 2021-10-26
Raise error on `pip install dbt`

### LINKED ISSUE #4100: What should `pip install dbt` do after v1?
Split off from https://github.com/dbt-labs/dbt-core/issues/3968

Everyone will be able and encouraged to do the following after v1:
- `pip install dbt-core`
- `pip install dbt-<adapter>`

The big question is, what to do with the PyPi package named `dbt`, defined in this repo as a thing _separate_ from `dbt-core`? Note that this is just a cosmetic consideration—it shouldn't change anything about how we're packaging/distributing the underlying components—we just know that a lot of people are still using `pip install dbt`. Starting in v1, this could:

1. **Full backwards compatibility, with warning:** Raise a deprecation warning, then install `dbt-core`, `dbt-postgres`, `dbt-redshift`, `dbt-bigquery`. This might be tricky if we've released a newer patch/prerelease of dbt-core than the others, but I think our compatibility operator (`~=`) could work here
2. **Very limited backwards compatibility, with warning**: Raise a deprecation warning, then install only the code contained in this repository (`dbt-core` + `dbt-postgres`, or `dbt-core` only), knowing that most people will encounter errors and have to switch)
3. **Raise an explicit error:** If all of that is too tricky to figure out, we should keep it simple and just raise a good old fashioned error message when someone tries `pip install dbt` (unqualified) or `pip install dbt==1.0.0`: "Going forward, you must install dbt-<adapter`"

I'm leaning toward the first option right now. I'm very open to other opinions.

jtcohen6:
resolves #4100

- Raise an explicit error on `pip install dbt`
- Include richer metadata for `dbt-core` (in `core/setup.py`)

```
As of v1.0.0, `pip install dbt` is no longer supported.
  Instead, please use either:
    - `pip install dbt-core`, for core functionality
    - `pip install dbt-<adapter>`, to use dbt with your database, platform, or query engine
  See full list: https://docs.getdbt.com/docs/available-adapters
```

### What's good

- Simple, clear, to the point
- No need for circular-ish dependencies (`dbt-core` &rarr; `dbt-snowflake` &rarr; `dbt`)
- Rips off the band-aid

### What's bad

- This will break production pipelines that depend on `pip install dbt`
- This will break any python programs that have unbounded dependencies on `dbt` (rather than `dbt-core`)
- Cloning this repo + `pip install .` will not work, it needs to be cloning this repo + `pip install -r requirements.txt` (what dbt Cloud `HEAD` does) or `pip install -r editable-requirements.txt` (what we tell contributors to do)
- `pip install git+https://github.com/dbt-labs/dbt-core` will not work, it needs to be `git+https://github.com/dbt-labs/dbt.git#egg=dbt-core&subdirectory=core` (we can document this)

### Uncertainties

- Will PyPi metadata still work? Will we be able to push and recognize this package as `dbt`?
- I wasn't able to get clear deprecation warnings working, and I wasn't able to find very clear resources online, but that's not to say that it isn't possible

### Checklist


leahwicz (approved):

jtcohen6:
I'm _pretty_ sure this change will break PyPi metadata, so we might need to hand-roll that instead. Maybe we could test by pushing a `dbt==1.00b1` prerelease after merging?

leahwicz:
@jtcohen6 we have the test PyPi account we can use to try this out

jtcohen6:
@leahwicz ah really good thought!! I forgot all about that

kwigley:
@jtcohen6 tested locally and ended up pushing a small change to `setup.py` (https://github.com/dbt-labs/dbt-core/pull/4133/commits/cf3089df943f5d5f9a95d58eba55cdfb6f0c85b8). 

In order for this to work as expected, we should only build a source distribution when publishing to PyPI (`python setup.py sdist`). Distributing a wheel file won't run the code to warn the user.

kwigley (approved):

jtcohen6:
@kwigley Thanks for the magic one-liner there. I did some reorg enabled by that, including some edits to `build-dist`. Let me know if I'm being too clever/fancy there.

Let's try building `dbt=1.0.0-b1` and pushing to Test PyPi, before we go live-live with this

jtcohen6:
Now that I've reenabled the source-only distribution of `dbt`, it looks like I need to update some more pieces of the `main.py` CI test to avoid trying to install it, since it will raise an error :)

joellabes:
I'm not convinced that 
>    - `pip install dbt-core`, for core functionality

fulfils the "clear" component of 
>Simple, clear, to the point.

Why? Because I don't understand it 😅 . I'm reading it as "to access the critical functionality of dbt", but I think you want me to read it as "access the functionality of the dbt core project". Is that right?

Who is the target market for dbt-core on its own? I assume it's people building things that integrate with the dbt ecosystem, as opposed to analytics engineers building a dbt project? As a generic end user, "connect to a database" is definitely "~core~ critical functionality" so I wouldn't necessarily be warned off of trying to install it. 

Noodling on some alternatives:
- `pip install dbt-core`, for core functionality without connecting to a database
- `pip install dbt-core`, for developers of dbt integrations

jtcohen6:
> `pip install dbt-core`, for developers of dbt integrations

I like this! I'll also move it so that it's after `pip install dbt-<adapter>`, which we expect to be relevant in 95%+ of cases.

Issue #4135, opened 2021-10-26
[Bug] Huge slowdown of dbt initialization for projects with many tests since v0.21


dvalchev:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

This is a continuation of the closed #4012. 
Since the upgrade from v0.20 to v0.21, when the project has a lot of schema tests, dbt startup time is severely increased. In one of the projects (343 models, 3882 tests) the startup time increased from 7 seconds to over 3 minutes! Using the timing recording, I managed to track down the root cause to #4022 (`compilation.py:114(<listcomp>)`). In addition, it seems that the `digraph.py:638(add_edges_from)` also performs significantly worse as well. 
With #4012, the first issue seems to be resolved and the startup gets faster but is still impacted by the `digraph.py:638(add_edges_from)`. 
Disabling all tests makes the performance on par with 0.20.

Here are the details from each version tests. I used pre-parsed simple `dbt compile` to showcase the issue:
### 0.20.2

* **dbt.log**
```
2021-10-26 13:09:56.378448 (MainThread): Running with dbt=0.20.2
2021-10-26 13:09:57.385906 (MainThread): running dbt with arguments Namespace(cls=<class 'dbt.task.compile.CompileTask'>, debug=False, defer=None, exclude=None, full_refresh=False, log_cache_events=False, log_format='default', models=None, parse_only=False, partial_parse=None, profile=None, profiles_dir='C:\\Users\\dvalchev\\.dbt', project_dir=None, record_timing_info='..\\timimng-compile-0.20.txt', rpc_method='compile', selector_name=None, single_threaded=False, state=None, strict=False, target=None, test_new_parser=False, threads=15, use_cache=True, use_colors=None, use_experimental_parser=False, vars='{}', version_check=True, warn_error=False, which='compile', write_json=True)
2021-10-26 13:09:57.387901 (MainThread): Tracking: do not track
2021-10-26 13:09:59.117617 (MainThread): Partial parsing enabled: 0 files deleted, 0 files added, 0 files changed.
2021-10-26 13:09:59.117617 (MainThread): Partial parsing enabled, no changes found, skipping parsing
2021-10-26 13:10:01.176648 (MainThread): write_gpickle is deprecated and will be removed in 3.0.Use ``pickle.dump(G, path, protocol)``
[...]

PR #4143, opened 2021-10-27
Add extra graph edges for `build` only

jtcohen6:
Short-term resolution for #4135. My hope is that, with a change this simple, we can backport it to `0.21.latest` for last-minute inclusion in v0.21.1

In order to enable desirable test-blocking behavior for `dbt build`, we add graph edges between upstream tests and downstream models. In very large projects, this has performance implications, since we're passing a much larger graph into `networkx` methods. Let's only add those extra edges for `build`, where they really get us something, and restore previous performance for other tasks, which don't use them.

The long-term resolution to this issue could look like:
- Reusing previously compiled graph (as in partial parsing)
- Cleverly constructing a partial graph, based on ancestors and descendents of actually selected nodes, to significantly speed up `dbt build -s one_model`
- Reworking our execution queue to obviate the need for transitive closure

I realized, counterintuitively, that in many cases, preserving transitive edges doesn't really change the result of a dbt operation. That said, the rigor of dbt's DAG is one of its most important conceptual features, so I don't think we can remove this right now.

In the process of using snakeviz, I saw a lot of opportunity for performance work beyond our tight focus this year (project parsing / manifest construction). There are methods in graph compilation and node selection that take multiple seconds in large projects, and (no surprise) I added at least one of them in #3235. Let's swing back for more after v1.

### Checklist


leahwicz (approved):

iknox-fa (approved):
LGTM

PR #4052, opened 2021-10-13
Er/3250 generic tests

emmyoop:
resolves #3250

### Description

Adding ability to have generic tests in the `/tests` directory.

- added new ParseFileType `GenericTest` and renamed `Test` to be `SingularTest`
- added new GenericTestParser
- parse generic tests in `/tests` directly after macros and before everything else.

### Checklist

- [x] I have signed the [CLA](https://docs.getdbt.com/docs/contributor-license-agreements)
- [x] I have run this code in development and it appears to resolve the stated issue
- [x] This PR includes tests, or tests are not required/relevant for this PR
- [x] I have updated the `CHANGELOG.md` and added information about my change

PR #3858, opened 2021-09-03
Reorganize global macros

jtcohen6:
Follow-on to #3851. I got carried away and started:
- reorganizing `global_project` macros into smaller, more clearly named files
- dispatching more macros, especially snapshot materialization helpers, to enable more modular plugin/package overrides

There are zero functional changes intended in this PR. Is this even a good idea? Is the improved organization worth the loss in `git blame` history? I'm open to feedback.

### Checklist

- [ ] I have signed the [CLA](https://docs.getdbt.com/docs/contributor-license-agreements)
- [ ] I have run this code in development and it appears to resolve the stated issue
- [ ] This PR includes tests, or tests are not required/relevant for this PR
- [ ] I have updated the `CHANGELOG.md` and added information about my change to the "dbt next" section.

PR #4154, opened 2021-10-28
Reorganize global project (macros)

jtcohen6:
Picking up from #3858. I got the itch for some autumnal cleaning. It's just too hard today to find the macros I'm looking for, or to even know where they might be. I'd really like this to be better for v1.

Dream scenario: We also document (`description` + `args`) the most important global macros, especially ones that users are likely to override/reimplement. That can come after rc1, and it can be a team effort :)

Trade-off:
- Much easier to find these macros in the future
- Much harder to use `git blame` to understand when + why they've changed in the past (but they don't change so so often, and usually not one critical line at a time, save for pesky syntax errors)

Also:
- dispatch more macros
- move generic tests (just the `test` part) into `tests/generic`, following #4052, with a comment explaining why most of the actual SQL lives separately — **do we like this?**

No intention for functional change in this PR. Very open to feedback!

### Checklist

- ~This PR includes tests, or tests are not required/relevant for this PR~ — existing tests all passing (🤞)


core/dbt/include/global_project/macros/adapters/relation.sql
```diff
@@ -0,0 +1,83 @@
...
+{% endmacro %}
+
+{# here for backwards compatibility #}
+{% macro load_relation(relation) %}
+  {% do return(adapter.get_relation(
+    database=relation.database,
+    schema=relation.schema,
+    identifier=relation.identifier
+  )) -%}
+{% endmacro %}
```
swanderz, line 83: in a post 1.0 world, do we need to keep these?

core/dbt/include/global_project/macros/etc/column_helpers.sql
```diff
@@ -0,0 +1,59 @@
+{% macro column_list(columns) %}
+  {%- for col in columns %}
+    {{ col.name }} {% if not loop.last %},{% endif %}
+  {% endfor -%}
+{% endmacro %}
```
swanderz, line 5: perhaps justifying this PR further, but this is the first time I've seen this macro before... ah! `/global_project/macros/materializations/helpers.sql`! guess i never looked there before... now I'm racking my brain for instances of this macro not being used where it should be.

core/dbt/include/global_project/macros/etc/configs.sql
```diff
@@ -0,0 +1,21 @@
...
+{% endmacro %}
+
+
+{% macro should_store_failures() %}
+  {% set config_store_failures = config.get('store_failures') %}
+  {% if config_store_failures is none %}
+    {% set config_store_failures = flags.STORE_FAILURES %}
+  {% endif %}
+  {% do return(config_store_failures) %}
+{% endmacro %}
```
swanderz, line 21: is this used outside of the test materialization? for me this `configs.sql` is the once file i feel iffy about existing, but I see where you're coming from.

core/dbt/include/global_project/macros/etc/column_helpers.sql
```diff
@@ -0,0 +1,59 @@
...
+   {# --check whether the name attribute exists in the target - this does not perform a data type check #}
+   {% for sc in source_columns %}
+     {% if sc.name not in target_names %}
+        {{ result.append(sc) }}
+     {% endif %}
+   {% endfor %}
+  
+  {{ return(result) }}
+
+{% endmacro %}
```
swanderz, line 42: curious (but no strong opinion), on the home for helper macros that are only currently used for a specific materialization... I like the categorization but maybe unnecessary complexity to go to another file when no other macros are using it?

core/dbt/include/global_project/macros/materializations/models/table/create_table.sql
```diff
@@ -0,0 +1,16 @@
...
+
+  {{ sql_header if sql_header is not none }}
+
+  create {% if temporary: -%}temporary{%- endif %} table
+    {{ relation.include(database=(not temporary), schema=(not temporary)) }}
+  as (
+    {{ sql }}
+  );
+
+{% endmacro %}
```
swanderz, line 16: love that these macros be broken into their own files! nit: can we make it `create_table_as.sql`? and do the same thing for `create_view_as` macro (`create_view_as.sql`)?

test/integration/029_docs_generate_tests/test_docs_generate.py
```diff
@@ -574,7 +574,7 @@ def _verify_generic_macro_structure(self, manifest):
             if k not in {'macro_sql'}
         }
         # Windows means we can't hard-code these.
-        helpers_path = Normalized('macros/materializations/helpers.sql')
+        helpers_path = Normalized('macros/etc/column_helpers.sql')
```
swanderz, line 577: this scares me. should I be scared?

swanderz (changes requested):
thanks @jtcohen6! I can tell you put a lot of thought into this.

I definitely support this change and that it ships as part of 1.0. I'd like to believe these macros will be more stable in 1.0 so the git history is not as important.

Random thought, but with this new macro taxonomy, I'm imagining just looking at an adapters `include/` file tree, would give you the ways that an adapter differs from the core/postgres implementation.

### questions
- would now (or at least pre-1.0) be a good time to rename adapter macros based on whether they return sql strings or actually execute SQL (`get_x_sql` vs `alter_x_do_y`) ([previous discussion in dbt Slack](https://getdbt.slack.com/archives/C50NEBJGG/p1627961286009800))
- what's the ideal ordering of macros?
    - alphabetically?
    - most used to least used?
    - those without `default__` versions first?
- related to above, anyway we can sandbox the macros that are required to be implemented for a custom adapter? or maybe we update the scaffolding to give the filetree and files for macros they have to implement? I guess I'd be interested to see what an adapter's filetree would look like in this new model...

core/dbt/include/global_project/docs/overview.md
```diff
@@ -38,6 +38,6 @@ Note that you can also right-click on models to interactively filter and explore
 - [What is dbt](https://docs.getdbt.com/docs/overview)?
 - Read the [dbt viewpoint](https://docs.getdbt.com/docs/viewpoint)
 - [Installation](https://docs.getdbt.com/docs/installation)
-- Join the [chat](https://community.getdbt.com/) on Slack for live questions and support.
+- Join the [dbt Community](https://www.getdbt.com/community/) for questions and discussion
```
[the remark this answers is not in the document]
jtcohen6: This is what's included by default in the dbt-docs site `Overview`, until/unless users provide a [custom overview](https://docs.getdbt.com/docs/building-a-dbt-project/documentation#setting-a-custom-overview). It could be simpler / have less info in it — is that what you're thinking?

core/dbt/include/global_project/macros/adapters/columns.sql
```diff
@@ -0,0 +1,88 @@
...
+  {{ return(adapter.dispatch('get_columns_in_relation', 'dbt')(relation)) }}
+{% endmacro %}
+
+{% macro sql_convert_columns_in_relation(table) -%}
+  {% set columns = [] %}
+  {% for row in table %}
+    {% do columns.append(api.Column(*row)) %}
+  {% endfor %}
+  {{ return(columns) }}
+{% endmacro %}
```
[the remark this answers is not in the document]
jtcohen6: agree!

core/dbt/include/global_project/macros/adapters/relation.sql
jtcohen6: My preference would be to:
- call out that they're just pass-throughs for other macros
- avoid calling them in any global materializations/macros
- not to remove them entirely — I don't see a ton of harm in keeping them around, and it's one more bump in the road for plugins/projects upgrading

core/dbt/include/global_project/macros/etc/column_helpers.sql
jtcohen6: Honestly, we've probably reinvented this wheel a dozen+ times. I didn't realize it existed until I did the work for this PR either. We have some utilities in core that we've used for various materializations, but some of them aren't even being used anymore. At this point, we could delete them entirely, or to keep them around forever—either feels like a valid approach.

core/dbt/include/global_project/macros/etc/configs.sql
jtcohen6: To me, `should_store_failures()` feels very closely related to `should_full_refresh()`. They have almost identical logic. The `should_full_refresh()` macro is used by both the incremental model materialization + seeds. Do you think it would make more sense to move `configs.sql` to `macros/materializations/` instead of `macros/etc`?

core/dbt/include/global_project/macros/etc/column_helpers.sql
jtcohen6: Yeah, good thought. As above, the question for me is:
1. Should we consider these "private" methods that exist for internal (global materialization) use only, not to be documented, and used at users' own risk
2. Should dbt-core's global project become more of a global repository of handy, nice-to-have helper macros like this one
3. Somewhere in between: these are mostly internal methods, they're useful for adapter plugin maintainers when extending `dbt-core` to new databases, but they shouldn't be called / relied on any old project

When I put it like that, I actually find myself leaning toward 2 or 3, rather than 1. Which is to say, they should probably live in a file like `materializations/internal_helpers`

core/dbt/include/global_project/macros/etc/column_helpers.sql
jtcohen6: I don't see `column_list` or `column_list_for_create_table` being used in any of our global or plugin macros/materializations, so I think I'm just going to snip these out

core/dbt/include/global_project/macros/etc/configs.sql
jtcohen6: I've decided, yes, it would make more sense to have these in `macros/materializations/configs.sql`

core/dbt/include/global_project/macros/materializations/models/table/create_table.sql
jtcohen6: yes, absolutely!

test/integration/029_docs_generate_tests/test_docs_generate.py
jtcohen6: No cause for fear. The comment a few lines up:
```
        # just test a known global macro to avoid having to update this every
        # time they change.
```
We want to verify that global/internal macros are included in `manifest.json`, and to do that we just test that one of them exists, with all of its properties, including its path.

jtcohen6:
@swanderz Thank you for the detailed + thoughtful feedback!!

Questions:

> would now (or at least pre-1.0) be a good time to rename adapter macros based on whether they return sql strings or actually execute SQL (`get_x_sql` vs `alter_x_do_y`)?

Yes. I think `create_table_as` and `create_view_as` are the biggest apostates from this convention. I just updated them to be `get_create_table_as_sql` and `get_create_view_as_sql` by default, with the old macros serving as passthroughs for backwards compatibility.

> what's the ideal ordering of macros? alphabetically? most used to least used? those without `default__` versions first?

Good question! My perspective has been:
- We should split up macros into enough files that they are grouped more meaningfully, and so that you don't need to scroll very much or at all (<100 lines)
- Materializations should always be alone in their own files
- Within each (much smaller) file, macros should proceed in as "logical" of an order as we can muster. The dispatched macro should precede the `default__` implementation. If `macro_a` is called by `macro_b`, `macro_a` should precede `macro_b`.
- Beyond that, I don't have super-strong feelings. I think this reorganization is a _very_ partial step on the way toward real macro discovery/discoverability, and we'll need other tools to help.

> related to above, anyway we can sandbox the macros that are required to be implemented for a custom adapter? or maybe we update the scaffolding to give the filetree and files for macros they have to implement? I guess I'd be interested to see what an adapter's filetree would look like in this new model...

Yes! All of these macros are sandboxed in `global_project/macros/adapters`. These are the macros we expect an adapter plugin will need or want to override, since it's behavior that differs across databases.

Right now, I have the macros grouped by function (`columns` vs. `freshness` vs. `persist_docs`), but we could instead group them by which ones have default implementations vs. which _must_ be implemented for a new custom adapter. I think my preference is, as you say, to update the adapter scaffolding to use the same functional file tree when they override/reimplement, rather than one huge file named `adapters.sql`.

Of course, some adapters will need to reimplement the core materializations in addition to the macros in the `adapters/` subdirectory, but we should aspire to evade that necessity as often as we can.

To your earlier comment, all of these tweaks together will enable someone to quickly glance at an adapter's `include/<adapter>/macros` directory to see how much it differs from core/postgres/default.

core/dbt/include/global_project/macros/adapters/relation.sql
jtcohen6: In retrospect, `load_relation` actually offers a real convenience. I'm going to add it back

core/dbt/include/global_project/docs/overview.md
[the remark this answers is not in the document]
swanderz: sure. but does the `global_project` need it? `global_project` doesn't show up in anyones dbt docs, nor are any of the macros documented? I guess this file is just present in all scaffolded dbt projects?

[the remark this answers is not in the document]
jtcohen6: That's right - this is how it gets into users' project, by default.

It's an opinionated choice to exclude internal/global macros from the dbt-docs site; they're included in `manifest.json`, just like every other macro. I'd like to document global macros (perhaps a community-sourced effort??), which we can use to power other cool metadata-y things down the line

leahwicz (approved):

PR #4160, opened 2021-10-28
Add test edges for improper subsets, too

### LINKED ISSUE #4158: In `build`, tests should always block downstream, even when parent sets are equal
Last one, I promise. I should've caught this a while ago, but better late than never.
```sql
-- models/model_a.sql
select null as id
```
```sql
-- models/model_b.sql
select * from {{ ref('model_b') }}
```
```yml
-- models/schema.yml
version: 2
models:
  - name: model_a
    columns:
      - name: id
        tests:
          - not_null
```

I'd expect:
```
21:32:06 | 1 of 3 START view model dbt_jcohen.model_a........................... [RUN]
21:32:06 | 1 of 3 OK created view model dbt_jcohen.model_a...................... [CREATE VIEW in 0.06s]
21:32:06 | 2 of 3 START test not_null_model_a_id................................ [RUN]
21:32:06 | 2 of 3 FAIL 1 not_null_model_a_id.................................... [FAIL 1 in 0.03s]
21:32:06 | 3 of 3 SKIP relation dbt_jcohen.model_b.............................. [SKIP]
```

But in v0.21.0, it's:
```
21:31:38 | 1 of 3 START view model dbt_jcohen.model_a........................... [RUN]
21:31:38 | 1 of 3 OK created view model dbt_jcohen.model_a...................... [CREATE VIEW in 0.06s]
21:31:38 | 2 of 3 START view model dbt_jcohen.model_b........................... [RUN]
21:31:38 | 3 of 3 START test not_null_model_a_id................................ [RUN]
21:31:38 | 3 of 3 FAIL 1 not_null_model_a_id.................................... [FAIL 1 in 0.04s]
21:31:38 | 2 of 3 OK created view model dbt_jcohen.model_b...................... [CREATE VIEW in 0.05s]
```

The issue comes up because we require a strict/proper subset when adding dependency links. Because `model_b` and the test `not_null_model_a_id` have the same set of parents (`model_a`), the test's dependencies aren't a strict subset of the model's dependencies, and we don't add a link between `not_null_model_a_id --> model_b`.

There's a reason we opted for strict/proper subset, but I'm not exactly sure what it is, and I've managed to get all our tests passing when I change the logic to use a non-strict/improper subset.

If I'm on the right track (never guaranteed!), I'd love to fix this for v0.21. It really threw me for a loop earlier today, and it goes against the documented/expected behavior of `build`.

jtcohen6:
resolves #4158

- Don't require a strict/proper subset when adding edges between tests and downstream models
- Add simple integration test that fails with current behavior, passes with change

There's a reason we opted for strict/proper subset, because there was _some_ case in which it caused a circular dependency. I'm trying to figure out what that case was.

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -96,7 +96,9 @@ Contributors:
 - Added support for tests on databases that lack real boolean types. ([#4084](https://github.com/dbt-labs/dbt-core/issues/4084))
 - Scrub secrets coming from `CommandError`s so they don't get exposed in logs. ([#4138](https://github.com/dbt-labs/dbt-core/pull/4138))
 - Syntax fix in `alter_relation_add_remove_columns` if only removing columns in `on_schema_change: sync_all_columns` ([#4147](https://github.com/dbt-labs/dbt-core/issues/4147))
+- Increase performance of graph subset selection ([#4135](https://github.com/dbt-labs/dbt-core/issues/4135),[#4155](https://github.com/dbt-labs/dbt-core/pull/4155))
 - Add downstream test edges for `build` task _only_. Restore previous graph construction, compilation performance, and node selection behavior (`test+`) for all other tasks ([#4135](https://github.com/dbt-labs/dbt-core/issues/4135), [#4143](https://github.com/dbt-labs/dbt-core/pull/4143))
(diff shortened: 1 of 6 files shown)


iknox-fa (approved):
LGTM

PR #4161, opened 2021-10-28
Adapter logging interface

nathaniel-may:
waiting for #4137 to merge first

### Description

Adapter maintainers currently use our old logging system. This exposes the new eventing system to them so that they have to make minimal changes to stay up-to-date for v1. Adapter maintainers will not have to touch every log call site to use the new system.

### Checklist

### DIFF
--- a/core/dbt/events/README.md
+++ b/core/dbt/events/README.md
@@ -7,3 +7,16 @@ The event module provides types that represent what is happening in dbt in `even
 
 # Adding a New Event
 In `events.types` add a new class that represents the new event. This may be a simple class with no values, or it may be a dataclass with some values to construct downstream messaging. Only include the data necessary to construct this message within this class. You must extend all destinations (e.g. - if your log message belongs on the cli, extend `CliEventABC`) as well as the loglevel this event belongs to.
+
+# Adapter Maintainers
+To integrate existing log messages from adapters, you likely have a line of code like this in your adapter already:
+```python
+from dbt.logger import GLOBAL_LOGGER as logger
+```
+
+Simply change it to these two lines with your adapter's database name, and all your existing call sites will now use the new system for v1.0:
+```python
+from dbt.events import AdapterLogger
+logger = AdapterLogger("<database name>")
+# e.g. AdapterLogger("Snowflake")
+```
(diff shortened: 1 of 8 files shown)


nathaniel-may:
I should probably add the `exception` log level here too.

nathaniel-may:
re-requesting prior approvals because I made a significant change since then.

core/dbt/events/adapter_endpoint.py
```diff
@@ -0,0 +1,86 @@
...
+        event.exc_info = exc_info
+        event.stack_info = stack_info
+        event.extra = extra
+
+        fire_event(event)
+
+    def exception(
+        self,
+        msg: str,
+        exc_info: Any = True,  # this default is what makes this method different
```
emmyoop, line 76: I like this call out.

emmyoop (approved):

nathaniel-may:
Should this object have some kind of doc string? What should it say?

leahwicz (approved):

core/dbt/events/README.md
```diff
@@ -7,3 +7,16 @@ The event module provides types that represent what is happening in dbt in `even
...
+To integrate existing log messages from adapters, you likely have a line of code like this in your adapter already:
+```python
+from dbt.logger import GLOBAL_LOGGER as logger
+```
+
+Simply change it to these two lines with your adapter's database name, and all your existing call sites will now use the new system for v1.0:
+```python
+from dbt.events import AdapterLogger
+logger = AdapterLogger("<database name>")
+# e.g. AdapterLogger("Snowflake")
```
jtcohen6, line 21: 👍

jtcohen6 (approved):
The interface and explainer look great to me.

Could this PR also do the two-line work of switching over the Postgres adapter? It looks like there's one spot where it imports `dbt.logger` and should instead use `AdapterLogger("Postgres")`? That could be good as far as having a concrete example to point to when cutting over more adapters.

If you'd rather than come in a subsequent PR, that's also fine by me.

nathaniel-may:
Logbook is causing issues when I moved the Postgres adapter over to the new interface. Instead of spinning my wheels investigating, I migrated all these changes into #4176 because Logbook isn't a problem over there.

nathaniel-may:
Tests passed over there so I'm verifying that it works over here for record.

PR #4162, opened 2021-10-29
[#3885] Partially parse when environment variables in schema files change

### LINKED ISSUE #3885: Env vars + partial parsing

gshank:
resolves #3885

### Description

A previous pull request handled env_vars in SQL files. This pull request handles env_vars in schema files. Prior to this yaml dictionaries were rendered by deep_map all at once. In order to flag individual entries in a schema file for partial parsing rendering, the deep_map rendering has been split so that each individual entry is rendered (except for tests and descriptions, which are rendered later). In addition env_vars are trapped by test rendering in several places. This does not address the possibility of env_vars in descriptions.

### Checklist


core/dbt/parser/schemas.py
```diff
@@ -326,19 +341,35 @@ def _parse_generic_test(
...
+        self.manifest.env_vars.update(env_vars)
+        if schema_file_id in self.manifest.files:
+            schema_file = self.manifest.files[schema_file_id]
+            if isinstance(target, UnpatchedSourceDefinition):
+                search_name = target.source.name
+                yaml_key = target.source.yaml_key
+                if '.' in search_name:  # source file definitions
+                    (search_name, _) = search_name.split('.')
+            else:
+                search_name = target.search_name
```
jtcohen6, line 360: I was trying something pretty wacky, and managed to break this:
```
version: 2
models:
  - name: model_a
    columns:
      - name: id
        tests:
          - unique:
              enabled: "{{ env_var('ENABLED', True) }}"
```
```
  File "/Users/jerco/dev/product/dbt/core/dbt/parser/manifest.py", line 439, in parse_project
    parser.parse_file(block, dct=dct)
  File "/Users/jerco/dev/product/dbt/core/dbt/parser/schemas.py", line 507, in parse_file
    self.parse_tests(test_block)
  File "/Users/jerco/dev/product/dbt/core/dbt/parser/schemas.py", line 483, in parse_tests
    self.parse_column_tests(block, column)
  File "/Users/jerco/dev/product/dbt/core/dbt/parser/schemas.py", line 207, in parse_column_tests
    self.parse_test(block, test, column)
  File "/Users/jerco/dev/product/dbt/core/dbt/parser/schemas.py", line 479, in parse_test
    self.parse_node(block)
  File "/Users/jerco/dev/product/dbt/core/dbt/parser/schemas.py", line 417, in parse_node
    node = self._parse_generic_test(
  File "/Users/jerco/dev/product/dbt/core/dbt/parser/schemas.py", line 295, in _parse_generic_test
    self.store_env_vars(target, schema_file_id, self.schema_yaml_vars.env_vars)
  File "/Users/jerco/dev/product/dbt/core/dbt/parser/schemas.py", line 360, in store_env_vars
    search_name = target.search_name
```

core/dbt/context/base.py
```diff
@@ -21,6 +21,38 @@
 import datetime
 import re
 
+# Contexts in dbt Core
```
jtcohen6, line 24: This comment is simply outstandingly helpful. Thank you x1000

core/dbt/parser/schemas.py
gshank: Will make a test case for it.

gshank: This should be fixed. I also added throwing an exception for when there is no node to match a patch, because part of the test case was not doing what it was supposed to.

core/dbt/parser/generic_test_builders.py
```diff
@@ -226,6 +226,12 @@ def __init__(
         self.namespace: str = groups['test_namespace']
         self.config: Dict[str, Any] = {}
 
+        # This code removes keys identified as config args from the test entry
+        # dictionary. The keys remaining in the 'args' dictionary will be
+        # "kwargs", or keyword args that are passed to the test macro.
+        # The "kwargs" are not rendered into strings until compilation time.
+        # The "configs" are rendered here (since they were not rendered back
+        # in the 'get_key_dicts' methods in the schema parsers).
```
jtcohen6, line 234: Thanks for the comment here. This is quite confusing behavior, and we should aim to eventually refactor it. I just realized one more way it can confuse: test kwargs _are_ included when creating the test node's name and `unique_id`, which are created at parse time, even though they're not _really_ rendered until execution time.

There's some conceptual merit to this distinction. A test with different kwargs is a _different test_ (e.g. `accepted_values` checking for `1,2,3` instead of `a,b,c`), whereas a test with different configs is just the same test being configured differently. But this is a fuzzy boundary at best, with some odd consequences. I think we should spend more time rethinking it after v1.

jtcohen6 (commented):
From my end-to-end functional testing, this is swell. It's amazing how many places we let users put env vars; this PR manages to nab just about all of them. I'll happily document the limitation about `description` fields—we can swing back around for those if/when there's need for it.

I'm also deeply grateful for all the comments and annotations explaining the intricacies of our many contexts and spots for rendering. Compared to where we were in January, I feel like we've finally manage to excavate a massive paleontological creature. The project for next year will be finding ways to put the pieces together in a way that's much more sensible for us, and consistent for users.

Another member of the Core eng team should give this code a look before we merge, given the number of things it touches. From my perspective, this gets the job done.

jtcohen6 (approved):
We weren't able to give this a line-by-line review. At this point, it's more important that we have it in for the release candidate, where it can be thoroughly tested over the next 3-4 weeks.

Nice work on this one! I know it involved touching a lot of codepaths. Serious props for leaving them better than you found them.

core/dbt/parser/schemas.py
```diff
@@ -827,8 +886,11 @@ def parse_patch(
                     # re-application of the patch in partial parsing.
                     node.patch_path = source_file.file_id
             else:
-                # Should we issue a warning message here?
-                return
+                raise ParsingException(
+                    f"Did not find matching node for patch with name '{patch.name}' "
+                    f"in the '{patch.yaml_key}' section of "
+                    f"file '{source_file.path.original_file_path}'"
+                )
```
jtcohen6, line 893: @gshank This is on me for not catching before we merged. This is going to be a breaking change for many users. Historically, we've supported the ability to define patches for nodes that don't exist; they simply aren't applied.

Any chance we could switch this from an exception to a debug-level warning message, and keep going?

gshank: Yes, that would be fine. We still create the tests for non-existent nodes, which seems kind of wrong, but it won't affect the env_var parsing.

jtcohen6: Agree, that does seem sort of wrong. Those tests will exist in the manifest, but I don't think they'll actually be executed. For later :shrug:

PR #4169, opened 2021-10-29
config call sites

emmyoop:
### Description

Modifying call sites for structured logging in core/dbt/config

### Checklist

### DIFF
--- a/core/dbt/events/types.py
+++ b/core/dbt/events/types.py
@@ -313,10 +313,9 @@ class MissingProfileTarget(InfoLevel, CliEventABC):
         return f"target not specified in profile '{self.profile_name}', using '{self.target_name}'"
 
 
-#  TODO: this should have exc_info=True
 @dataclass
-class ProfileLoadError(DebugLevel, CliEventABC):
-    exc: Exception
+class ProfileLoadError(ShowException, DebugLevel, CliEventABC):
+    exc: Exception = Exception('')
 
     def cli_msg(self) -> str:
         return f"Profile not loaded due to error: {self.exc}"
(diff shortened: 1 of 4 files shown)


core/dbt/events/types.py
```diff
@@ -299,6 +299,36 @@ def cli_msg(self) -> str:
...
+    def cli_msg(self) -> str:
+        return f"target not specified in profile '{self.profile_name}', using '{self.target_name}'"
+
+
+@dataclass
+class ProfileLoadError(ShowException, DebugLevel, CliEventABC):
+    exc: Exception = Exception('')
+
+    def cli_msg(self) -> str:
+        return f"Profile not loaded due to error: {self.exc}"
```
emmyoop, line 316: @nathaniel-may I'm not sure this was the intention or is I'm missing something.  Because `ShowException` has defaults it throws a `Fields without default values cannot appear after fields with default values` error unless a default is provided for any fields in the type.  Is this how we want to handle it?  Or is there something I'm missing?

nathaniel-may: I tried to recreate locally, and the code you have here passes mypy checks and runs fine for me when I create an instance of this class. How did you get that exception to throw?

nathaniel-may: @iknox-fa figured it out. The lib we use for 3.6 compat with dataclasses is the issue. We probably will have to set all the values to their defaults again in the subclasses until we drop support for 3.6. Let's TODO setting these defaults to remove them after dropping 3.6.

nathaniel-may (approved):

nathaniel-may:
@iknox-fa is going to fix our `ShowException` issue on `feature/structured-logging` so let's wait to merge till we can remove the default and use the new `ShowException` class.

iknox-fa:
> @iknox-fa is going to fix our `ShowException` issue on `feature/structured-logging` so let's wait to merge till we can remove the default and use the new `ShowException` class.

Fixed!

PR #4176, opened 2021-11-01
Structlog

nathaniel-may:
### Description

Adds structlog to the event module as the new logging library. The existing logger will be removed once all the call sites have been modified to use the event module.

Before:
log-before

After:
log-after

Notable differences:
- Time is in UTC
- Log levels are always displayed
- Output has color

This PR also includes the new logging interface for adapters from #4161. The tests had an issue with Logbook, so I moved the changes here where there is no Logbook to save time.

### Todo:

- [ ] visually inspect on a windows machine with and without colorama
- [ ] figure out how to use local time if we want it to match the old time stamps

### Checklist


core/dbt/events/functions.py
```diff
@@ -1,117 +1,128 @@
...
-log = structlog.get_logger()
+    LOG = structlog.get_logger()
 
 
 # top-level method for accessing the new eventing system
 # this is where all the side effects happen branched by event type
 # (i.e. - mutating the event history, printing to stdout, logging
 # to files, etc.)
 def fire_event(e: Event) -> None:
+    if LOG is None:
```
nathaniel-may, line 85: This is evaluates to True

core/dbt/main.py
```diff
@@ -240,6 +241,9 @@ def run_from_args(parsed):
     if task.config is not None:
         log_path = getattr(task.config, 'log_path', None)
     # we can finally set the file logger up
+    # TODO move as a part of #4179
+    breakpoint()
+    setup_event_logger(log_path or 'logs')
```
nathaniel-may, line 246: even though this is called first

core/dbt/events/functions.py
```diff
@@ -1,117 +1,128 @@
...
+            structlog.processors.format_exc_info,
+            structlog.stdlib.ProcessorFormatter.wrap_for_formatter,
+        ],
+        logger_factory=structlog.stdlib.LoggerFactory(),
+        wrapper_class=structlog.stdlib.BoundLogger,
+        cache_logger_on_first_use=True,
+    )
 
-log = structlog.get_logger()
+    LOG = structlog.get_logger()
```
nathaniel-may, line 77: Why doesn't this get set?

emmyoop: This creates a new local instance of LOG and doesn't modify the global instance.

kwigley:
```suggestion
    global LOG
    LOG = structlog.get_logger()
```
I think this should do what you expect

nathaniel-may: That did it! Thanks!

Now I am going to read on how Python manages scope because this is not how this works in a lot of other languages.

kwigley: Sounds about right 🐍

nathaniel-may:
closing in favor of #4231

PR #4180, opened 2021-11-01
Dataclass compatibility

iknox-fa:
The `dataclass` library we're using for 3.6 compatibility causes mypy/pylance errors when a subclass doesn't have default values  for a member var, but the parent class does.

Switched to using `__post_init__()` instead.

### DIFF
--- a/core/dbt/events/types.py
+++ b/core/dbt/events/types.py
@@ -33,9 +33,10 @@ class ErrorLevel():
 
 @dataclass
 class ShowException():
-    exc_info: Any = None
-    stack_info: Any = None
-    extra: Any = None
+    def __post_init__(self):
+        self.exc_info: Any = None
+        self.stack_info: Any = None
+        self.extra: Any = None
 
 
 # The following classes represent the data necessary to describe a
--- a/core/dbt/events/types.py
+++ b/core/dbt/events/types.py
@@ -33,6 +33,10 @@ class ErrorLevel():
 
 @dataclass
 class ShowException():
+    # N.B.:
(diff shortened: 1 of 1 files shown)


core/dbt/events/types.py
```diff
@@ -33,9 +33,10 @@ def level_tag(self) -> str:
 
 @dataclass
 class ShowException():
-    exc_info: Any = None
-    stack_info: Any = None
-    extra: Any = None
+    def __post_init__(self):
+        self.exc_info: Any = None
+        self.stack_info: Any = None
+        self.extra: Any = None
```
nathaniel-may, line 39: Since this runs after the init method, do we have to check that the values haven't been overridden before setting them back to None?

nathaniel-may (changes requested):

core/dbt/events/types.py
iknox-fa: We could I suppose, seems a little un-necessary given we always set the values in `cli_msg()` which , by virtue of being a method on the event class will always happen _after_ `__post_init__()`.

Let me know if you still would like that changed, I don't really care about a little extra guard-railing here.

nathaniel-may: oh that makes sense. let's keep it simple then. Maybe add a comment to this effect, but this should be good.

nathaniel-may (approved):

core/dbt/events/types.py
iknox-fa: 👍

iknox-fa: I added a comment block so it's clear what's happening there.

PR #4183, opened 2021-11-02
task call sites part 1

emmyoop:
### Description

Modifying call sites for structured logging in core/dbt/task

This directory has a lot of logging so splitting it up for the purposes of smaller reviews.

### Checklist

- [ ] I have signed the [CLA](https://docs.getdbt.com/docs/contributor-license-agreements)
- [ ] I have run this code in development and it appears to resolve the stated issue
- [ ] This PR includes tests, or tests are not required/relevant for this PR
- [ ] I have updated the `CHANGELOG.md` and added information about my change

PR #4188, opened 2021-11-02
task callsites: part 2

emmyoop:
### Description

Modifying call sites for structured logging in core/dbt/task

#4183 is Part 1 of the same dir

### Checklist


core/dbt/events/types.py
```diff
@@ -636,15 +638,210 @@ class InvalidDisabledSourceInTestNode(WarnLevel, CliEventABC):
...
+    if execution_time is None:
+        status_time = ""
+    else:
+        status_time = " in {execution_time:0.2f}s".format(
+            execution_time=execution_time)
+
+    output = "{justified} [{status}{status_time}]".format(
+        justified=justified, status=status, status_time=status_time)
+
+    return output
```
emmyoop, line 816: Thoughts on this living here like this?

core/dbt/events/types.py
```diff
@@ -636,15 +638,210 @@ class InvalidDisabledSourceInTestNode(WarnLevel, CliEventABC):
...
+    total: int
+    truncate: bool
+
+    def cli_msg(self) -> str:
+        msg = "START hook: {self.statement}"
+        return format_fancy_output_line(msg=msg,
+                                        status='RUN',
+                                        index=self.index,
+                                        total=self.total,
+                                        truncate=self.truncate)
```
emmyoop, line 843: An alternative to this approach of one type per message is a general log statement for each level: Error, Warn and Info that is then shared across messages in `core/dbt/task/printer.py`.  That goes against what we've been doing elsewhere though.

core/dbt/task/printer.py
```diff
@@ -67,14 +71,11 @@ def get_counts(flat_nodes) -> str:
 
 
 def print_start_line(description: str, index: int, total: int) -> None:
-    msg = "START {}".format(description)
-    print_fancy_output_line(msg, 'RUN', logger.info, index, total)
```
emmyoop, line 71: `print_fancy_output_line` is used in multiple places and needs to be replaced.

core/dbt/events/types.py
nathaniel-may: Yeah that's an alternative I considered. It doesn't really solve the problem of having development time notice when our public interface changes. Leveraging mypy via these types gives us a little of that even though it might look a bit messy at times.

core/dbt/events/types.py
nathaniel-may: I'm fine with this. I would also be fine with it living in another module and we just import it for now till we can migrate it in elegantly. I actually much prefer when a lot of this logic is in functions and colocated with the messages themselves.

core/dbt/task/printer.py
nathaniel-may: if it's used all over the place, leaving it where it is and just importing it might be the way to go for now. I'd just put a TODO on the import statement.

emmyoop: I need to tweak it slightly as it uses the `logger_fn` parameter which is dynamically determining log levels.  I moved it since it needs to be modified anyways.

nathaniel-may: oohhhhhh good point. I didn't catch that.

core/dbt/task/base.py
```diff
@@ -392,14 +391,11 @@ def on_skip(self):
...
-                    node_name,
-                    self.node_index,
-                    self.num_nodes,
-                    self.skip_cause
-                )
+                fire_event(PrintSkipBecauseError(schema=schema_name,
+                                                 relation=node_name,
+                                                 index=self.node_index,
+                                                 total=self.num_nodes))
+                print_run_result_error(result=self.skip_cause, newline=False)
```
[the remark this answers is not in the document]
nathaniel-may: (there's a few places like this in this PR)

core/dbt/task/freshness.py
```diff
@@ -35,10 +36,41 @@ def on_skip(self):
...
+                                            table_name=table_name,
+                                            index=self.node_index,
+                                            total=self.num_nodes,
+                                            execution_time=result.execution_time))
+        else:
+            fire_event(PrintHookEndPassLine(source_name=source_name,
+                                            table_name=table_name,
+                                            index=self.node_index,
+                                            total=self.num_nodes,
+                                            execution_time=result.execution_time))
```
nathaniel-may, line 73: this looks like it was pretty non-trivial to change. I really like what you did with it here: It brings the important stuff to the top, and buries the stuff we don't care about as much. Nice job.

core/dbt/events/types.py
```diff
@@ -994,6 +997,649 @@ def cli_msg(self) -> str:
         return "Done."
 
 
+@dataclass
+class ServingDocsPort(InfoLevel, CliEventABC):
+    port: str
+
+    def cli_msg(self) -> str:
+        return f"Serving docs at 0.0.0.0:{self.port}"
```
nathaniel-may, line 1005: is there a place this address is used in the code? can we pass it in instead of having to maintain that the message is in sync?

core/dbt/events/types.py
```diff
@@ -994,6 +997,649 @@ def cli_msg(self) -> str:
...
+    if execution_time is None:
+        status_time = ""
+    else:
+        status_time = " in {execution_time:0.2f}s".format(
+            execution_time=execution_time)
+
+    output = "{justified} [{status}{status_time}]".format(
+        justified=justified, status=status, status_time=status_time)
+
+    return output
```
nathaniel-may, line 1168: I'm thinking we should move this into `functions.py` so that everything here is a type definition. Does that sound reasonable to you?

nathaniel-may (changes requested):
Looks good! Only a few minor things.

core/dbt/events/types.py
emmyoop: That creates a circular dependancy.  You're right that it doesn't belong in types though.  Maybe a new module in /events?  Then all the formatting (coloring etc) can end up there eventually too.  `format.py`?

core/dbt/events/types.py
emmyoop: Not in the form of a variable but I can use one and sync it up.

core/dbt/events/types.py
nathaniel-may: sure! I dig it. Let's make it happen in this PR just because we can.

core/dbt/events/types.py
nathaniel-may: I think that would be a great improvement that's worth doing now since it doesn't change the message output.

nathaniel-may (approved):
Perfect. Thanks!

PR #4189, opened 2021-11-02
Struct log for adapter call sites

iknox-fa:
Updated logging call sites for `core/dbt/adapters`


core/dbt/adapters/cache.py
```diff
@@ -330,14 +334,15 @@ def add(self, relation):
         :param BaseRelation relation: The underlying relation.
         """
         cached = _CachedRelation(relation)
-        logger.debug('Adding relation: {!s}'.format(cached))
-
-        lazy_log('before adding: {!s}', self.dump_graph)
+        fire_event(AddRelation(relation=cached))
+        # TODO: conditionally fire this event (logger.disabled, if it was std python logger)
+        fire_event(DumpBeforeAddGraph(graph=self.dump_graph()))
```
nathaniel-may, line 339: OH I see what `lazy_log` is doing here. I think it's guarding against _log levels_. If we're on info level, but this is debug level, because of strict evaluation, the long running `dump graph` will get called even though the message will never be displayed. The way the event module works right now, we force the message early so lazy logging won't help. I think this might be worth solving inside `fire_event` so we as developers don't have to decide whether the construction of a message is long running or not, it only constructs each message as they are needed.

core/dbt/events/stubs.py
```diff
@@ -0,0 +1,44 @@
+from typing import (
```
nathaniel-may, line 1: is the plan for this file to grow a lot for now and then we can delete it when we turn on mypy everywhere?

core/dbt/events/types.py
```diff
@@ -284,21 +286,303 @@ def cli_msg(self) -> str:
...
+    def cli_msg(self) -> str:
+        return f"Renaming relation {self.old_key} to {self.new_key}"
+
+
+@dataclass
+class DumpBeforeAddGraph(DebugLevel, CliEventABC):
+    graph: Dict[str, List[str]]
+
+    def cli_msg(self) -> str:
+        return f"before adding: {str(self.graph)}"
```
nathaniel-may, line 545: Here's the kind of thing I think we should be doing:

```python
from typing import Callable, Dict, List


def dump_graph() -> Dict[str, List[str]]:
    raise Exception('stub')

@dataclass
class Foo():
    graph_thunk:  Callable[[], Dict[str, List[str]]]

    def bar(self) -> str:
        # this runs into a mypy bug: https://github.com/python/mypy/issues/6910
        # I'll leave it up to you how you want to get around it.
        return f"some msg with the graph:\n{self.graph_thunk()}"

Foo(graph_thunk=dump_graph)
```

test/integration/051_query_comments_test/test_query_comments.py
```diff
@@ -70,10 +71,12 @@ def run_get_json(self, expect_pass=True):
...
     def query_comment(self, model_name, log):
+        # N.B: a temporary string replacement regex to strip the HH:MM:SS from the log line if present.
+        # TODO: make this go away when structured logging is stable 
+        log_msg = re.sub("(?:[01]\d|2[0123]):(?:[012345]\d):(?:[012345]\d \| )", "", log['message'])
         prefix = 'On {}: '.format(model_name)
-
-        if log['message'].startswith(prefix):
-            msg = log['message'][len(prefix):]
+        if log_msg.startswith(prefix):
+            msg = log_msg[len(prefix):]
```
nathaniel-may, line 79: What's this addition about?

nathaniel-may (changes requested):
We definitely need to address the lazy log situation which I completely didn't understand until now. Of course the obvious solution runs us into a mypy bug so this will be fun to handle. I'll put up a PR to make message evaluation lazy in fire_event, so our solution here can assume that's already in.

nathaniel-may:
Please see  #4210 for the preserved laziness in `fire_event`.

core/dbt/adapters/cache.py
```diff
@@ -330,14 +334,12 @@ def add(self, relation):
...
-
-        lazy_log('before adding: {!s}', self.dump_graph)
+        fire_event(AddRelation(relation=cached))
+        fire_event(DumpBeforeAddGraph(graph_func=self.dump_graph))
 
         with self.lock:
             self._setdefault(cached)
-
-        lazy_log('after adding: {!s}', self.dump_graph)
+        fire_event(DumpAfterAddGraph(graph_func=self.dump_graph()))
```
nathaniel-may, line 342:
```suggestion
        fire_event(DumpAfterAddGraph(graph_func=self.dump_graph))
```

nathaniel-may (changes requested):
Other than this straggler I think we're good. Mypy should have caught it though...

nathaniel-may (approved):

PR #4190, opened 2021-11-02
Show Exception should trigger a stack trace

nathaniel-may:
### Description

When we see a logline with just `exc_info=True` we're just Extending `ShowException` which I think is good because it's the most common form. Without this change we really should be setting `exc_info=True` in the concrete types which this change makes unnecessary.

### Checklist


emmyoop (approved):
Makes sense.

leahwicz (approved):

iknox-fa (approved):
LGTM

PR #4197, opened 2021-11-03
Add get_where_subquery to test namespace

### LINKED ISSUE #4195: [v0.21] Add get_where_subquery to generic test rendering context
This is a bug resulting from the switch in https://github.com/dbt-labs/dbt-core/pull/3859.

Create any generic test containing code like:

```sql
{%- set relation = api.Relation.create(schema=model.schema, identifier=model.table) -%}
{%- set columns = adapter.get_columns_in_relation(relation) -%}
```
Idea being, we want to get the columns present in the model being tested. This worked in previous versions of dbt-core, but starting in v0.21, it leads to an error like:
```
Field "path" of type dbt.contracts.relation.Path in dbt.adapters.postgres.relation.PostgresRelation has invalid value {'database': Undefined, 'schema': Undefined, 'identifier': Undefined}
```

The simplest fix here, from the user's perspective, is to switch the logic above to:
```sql
{%- set relation = api.Relation.create(schema=model.schema, identifier=model.table) if execute -%}
{%- set columns = adapter.get_columns_in_relation(relation) -%}
```

That has no functional effect, since `get_columns_in_relation` doesn't run at parse time anyway.

From our perspective, I _think_ the move is to add `get_where_subquery` to the rendering context for generic tests. Right now, the macro is (I believe) missing from the rendering context, so it's returning `None` instead of a relation, thus leading to the validation error above when we try to create a relation from the properties of `model`. If the macro is in the rendering context, it should return the actual result of `get_where_subquery` instead.

If this is a simple fix, we should try to get it in for v0.21.1 (which has a first release candidate going out today).

Note that this type of generic test will _not_ work if the `where` config is defined, as `get_where_subquery(ref(...))` will return a string (`select * from ref(...) where ...`), not a relation.

jtcohen6:
resolves #4195, naively

I'm definitely open to better ideas for how to go about this!

- https://github.com/dbt-labs/dbt-core/pull/3859 added a macro, `get_where_subquery`, to use when constructing generic tests. Now, the `model` object in tests is set to the output of `get_where_subquery(ref(...))` or `get_where_subquery(source(..., ...))`, instead of a very ugly Jinja string.
- Therefore, `get_where_subquery` needs to be present in the test context / macro namespace in order for particular generic tests to work, if they introspect `model` / expect it to be a relation. I added a test case that demonstrates this behavior.

In a sense, all generic tests _do_ depend on `get_where_subquery`, in the similar way to how they depend on `ref()` or `source()`. Should we _actually_ add `get_where_subquery` to the `depends_on.macros` for these generic test macros? If a user provided a custom (override) version of `get_where_subquery`, with partial parsing enabled, would we need to re-parse all those tests? 99% of the time, no; with generic tests like the ones that raise this issue, _technically yes_. It's a quite narrow edge case.

### Checklist

### DIFF
--- a/core/dbt/include/global_project/macros/etc/where_subquery.sql
+++ b/core/dbt/include/global_project/macros/etc/where_subquery.sql
@@ -1,5 +1,5 @@
 {% macro get_where_subquery(relation) -%}
-    {% do return(adapter.dispatch('get_where_subquery')(relation)) %}
+    {% do return(adapter.dispatch('get_where_subquery', 'dbt')(relation)) %}
 {%- endmacro %}
 
 {% macro default__get_where_subquery(relation) -%}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -90,6 +90,11 @@ Contributors:
 ## dbt 0.21.1 (Release TBD)
 
 ### Fixes
+- Add `get_where_subquery` to test macro namespace, fixing tests  
+
+## dbt 0.21.1rc1 (November 03, 2021)
+
+### Fixes
 - Performance: Use child_map to find tests for nodes in resolve_graph ([#4012](https://github.com/dbt-labs/dbt/issues/4012), [#4022](https://github.com/dbt-labs/dbt/pull/4022))
 - Switch `unique_field` from abstractproperty to optional property. Add docstring ([#4025](https://github.com/dbt-labs/dbt/issues/4025), [#4028](https://github.com/dbt-labs/dbt/pull/4028))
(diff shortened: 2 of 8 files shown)


gshank (approved):
Looks good! Seems like a pretty simple change, and is definitely better than some other kludgey approaches we've seen. I can envision providing some additional test macros in the future, but if that ever happens we'll worry about it then

core/dbt/include/global_project/macros/etc/where_subquery.sql
```diff
@@ -1,5 +1,5 @@
 {% macro get_where_subquery(relation) -%}
-    {% do return(adapter.dispatch('get_where_subquery')(relation)) %}
+    {% do return(adapter.dispatch('get_where_subquery', 'dbt')(relation)) %}
```
jtcohen6, line 2: I figured, while we're here, might as well

PR #4203, opened 2021-11-03
scrub the secrets

nathaniel-may:
### Description

Uses the same strategy that we use today via LogBook just via a simple function call. I decided not to use the built-in scrubbing from structlog because it's made to scrub named fields, which we are not doing here. We only have the values.

### Checklist

### DIFF
--- a/core/dbt/events/functions.py
+++ b/core/dbt/events/functions.py
@@ -2,6 +2,24 @@
 import dbt.logger as logger  # type: ignore # TODO eventually remove dependency on this logger
 from dbt.events.history import EVENT_HISTORY
 from dbt.events.types import CliEventABC, Event, ShowException
+import os
+from typing import List
+
+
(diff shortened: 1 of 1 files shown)


emmyoop (commented):
Looks good - FYI there's a commit in the main branch ahead of when this branched off that uses scrubbing in `exceptions.py`

leahwicz (approved):

jtcohen6 (approved):

emmyoop:
Tests also import `SECRET_ENV_PREFIX` to build secrets so it should probably stay a var.

nathaniel-may:
Thanks for the heads up @emmyoop. It looks like the commit in `exceptions.py` is going to work just fine with the new system. I also agree that since the var is used in tests I should be referencing it as well. We can move it out of the old logger module later.

PR #4209, opened 2021-11-04
test/integration call sites

emmyoop:
### Description

Modifying call sites for structured logging in test/integration

Only `base.py` actually produces logs and those are all this PR touches.  log_manager still appears in multiple places.

### Checklist

### DIFF
--- a/core/dbt/events/types.py
+++ b/core/dbt/events/types.py
@@ -1722,36 +1722,11 @@ class GeneralWarningException(WarnLevel, CliEventABC):
 
 
 @dataclass
-class IntegrationTestSetupFailure(ShowException, ErrorLevel, CliEventABC):
+class IntegrationTestMessage(ShowException, ErrorLevel, CliEventABC):
     msg: str
 
     def cli_msg(self) -> str:
         return self.msg
-
-
-@dataclass
-class IntegrationTestTearDownFailure(ShowException, ErrorLevel, CliEventABC):
-    dir: str
-
-    def cli_msg(self) -> str:
-        return f"Could not clean up after test - {self.dir} not removable"
-
-
-@dataclass
-class IntegrationTestInvokingWithArgs(InfoLevel, CliEventABC):
-    final_args: List[str]
-
-    def cli_msg(self) -> str:
-        return f"Invoking dbt with {self.final_args}"
-
-
-@dataclass
-class IntegrationTestTestConnection(DebugLevel, CliEventABC):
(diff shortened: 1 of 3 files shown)


test/integration/base.py
```diff
@@ -322,7 +327,7 @@ def setUp(self):
                 'test_original_source_path={0.test_original_source_path}',
                 'test_root_dir={0.test_root_dir}'
             )).format(self)
-            logger.exception(msg)
+            fire_event(IntegrationTestSetupFailure(msg=msg))
```
nathaniel-may, line 330: Ok so I'm actually tempted to make an escape hatch event for tests. Like one class we just pass strings to since they will only ever be used for debugging tests.

If we want test messages to be typed like this, I think I want them to be in a separate types module within tests...

What do you think?

nathaniel-may (commented):
I want to discuss our strategy for addressing test logs since I'm not convinced it should be the same here.

test/integration/base.py
nathaniel-may: @iknox-fa I'd like your thoughts too.

emmyoop: I lean towards the escape hatch class.  Right now these are the only 4 logs generated in any tests.  The escape hatch makes it easy to add more logs when testing if needed without changing code elsewhere to add new types, which i like.

iknox-fa: Yeah, that makes sense to me, especially putting it in a separate module for separation of concerns.

iknox-fa (approved):
Looks good to me!

nathaniel-may (approved):

core/dbt/events/types.py
```diff
@@ -1721,6 +1721,14 @@ def cli_msg(self) -> str:
         return str(self.exc)
 
 
+@dataclass
+class IntegrationTestMessage(ShowException, ErrorLevel, CliEventABC):
+    msg: str
+
+    def cli_msg(self) -> str:
+        return self.msg
```
nathaniel-may, line 1729: On second thought, what do you think about doing something like this?
```suggestion
        return f"Integration Test: self.msg"
```

nathaniel-may (changes requested):

test/integration/base.py
emmyoop: With current logging using different levels I ended up creating a new module and standard messages to go with it.

nathaniel-may (approved):
Beautiful

PR #4210, opened 2021-11-04
lazy logging in event module

nathaniel-may:
### Description

#4189 ran into a lazy logging function we previously didn't understand. Now we understand it has a pretty significant purpose, and we need to plumb laziness through the event module to prevent rendering huge messages that will never be written to a user.

### Checklist

### DIFF
--- a/core/dbt/events/functions.py
+++ b/core/dbt/events/functions.py
@@ -41,7 +41,6 @@ def gen_msg(e: CliEventABC) -> Generator[str, None, None]:
 # to files, etc.)
 def fire_event(e: Event) -> None:
     EVENT_HISTORY.append(e)
-    debug = flags.DEBUG
     if isinstance(e, CliEventABC):
         msg = gen_msg(e)
         if e.level_tag() == 'test' and not isinstance(e, ShowException):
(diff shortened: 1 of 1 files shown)


core/dbt/events/functions.py
```diff
@@ -22,57 +22,65 @@ def scrub_secrets(msg: str, secrets: List[str]) -> str:
     return scrubbed
 
 
+# this exists because some log messages are actually expensive to build.
+# for example, many debug messages call `dump_graph()` and we don't want to
+# do that in the event that those messages are never going to be sent to
+# the user because we are only logging info-level events.
+def lazy_log_msg(e: CliEventABC) -> Callable[[], str]:
+    return lambda: scrub_secrets(e.cli_msg(), env_secrets())
+
```
iknox-fa, line 31: This looks great @nathaniel-may!  One thought, could we move `logger.timestamped_line` into this?  it would dry up the big route-all-the-things block below.

iknox-fa (approved):
Looks good to me!

core/dbt/events/functions.py
nathaniel-may: oh good call. `timestamped_line` actually shouldn't be called at all because structlog is adding those now. I'll drop that in #4176.

emmyoop (approved):

nathaniel-may:
Well I thought about it this morning, and this definitely didn't solve the problem. Because we're forcing the message construction at the call sites regardless of debug level, we kicked the can down the road and still have the issue. The bit that I just added that really does work, is the `if ... and debug and ...` parts of `fire_event`.

Because I was testing generators anyway, I figured I'd throw memoization to message construction in `gen_msg`, but that's just a bonus not part of the solution here.

core/dbt/events/functions.py
nathaniel-may: jk I threw it in here.

emmyoop (approved):
looks good.

PR #4055, opened 2021-10-13
Add Structured Logging

nathaniel-may:
resolves #3359

### Description

The `structured-logging` branch is a feature branch and small incremental changes should be merged via PR into this branch. Tests need to pass for each merge into this feature branch. There are many integration tests that rely on certain wording of log messages, and it is important that these tests remain intact by converting them to use codes incrementally.

#### Phase 1:

- [x] merge initial framework design
    - #4137
- [x] add adapter interface
    - ~#4161~ #4176
- [x] change every single callsite to the existing logger in the app and tests
    - #4164
    - #4163
    - #4165
    - #4166
    - #4169
    - #4183
    - #4199
    - #4219
- [x] plumb the cli color flag so structlog respects it
    - #4176
- [x] secret scrubbing
    - #4203
- [x] send all logs to stdout and dbt.log through some logger
    - ~#4176~ #4231
- [x] hook in the option for json logs
    - ~#4176~ #4231
- [x] update tests
    - #4257
- [ ] update adapter repos: (requires this PR to be merged first)
    - [redshift](https://github.com/dbt-labs/dbt-redshift/pull/34)
    - [snowflake](https://github.com/dbt-labs/dbt-snowflake/pull/42)
    - [bigquery](https://github.com/dbt-labs/dbt-bigquery/pull/55)
    - [spark](https://github.com/dbt-labs/dbt-spark/pull/251)

#### Phase 2:

- [ ] add additional log lines for node status reporting
- [ ] create scalable in-memory event history
- [ ] improve message wording
- [ ] fine tune `structlog` configs including look and feel
- [ ] add codes to each event
- [ ] add thread id to each event
- [ ] make sure we can distinguish logs from different runs (preserve process id like it is today, or allow some sort of user markers)
- [ ] upgrade rpc backdoor away from env vars ala #4055

### Checklist

- [x] I have signed the [CLA](https://docs.getdbt.com/docs/contributor-license-agreements)
- [ ] I have run this code in development and it appears to resolve the stated issue
- [ ] This PR includes tests, or tests are not required/relevant for this PR
- [ ] I have updated the `CHANGELOG.md` and added information about my change

PR #4212, opened 2021-11-05
[#3885] Skip partial parsing if project env vars change

### LINKED ISSUE #3885: Env vars + partial parsing

gshank:
resolves #3885

### Description

This pull request captures the env vars when the dbt_project.yml file is rendered and saves a checksum of them in the manifest.state_check. When partial parsing is attempted it will compare the checksums and skip partial parsing if they have changed.

Profile env vars have also been handled. Note: only the specified profile is actually rendered.

### Checklist

### DIFF
--- a/core/dbt/config/project.py
+++ b/core/dbt/config/project.py
@@ -284,6 +284,7 @@ class PartialProject(RenderComponents):
             selectors_dict=rendered_selectors,
         )
 
+    # Called by 'collect_parts' in RuntimeConfig
     def render(self, renderer: DbtProjectYamlRenderer) -> 'Project':
         try:
             rendered = self.get_rendered(renderer)
@@ -397,6 +398,8 @@ class PartialProject(RenderComponents):
             vars_dict = cfg.vars
 
         vars_value = VarProvider(vars_dict)
+        # There will never be any project_env_vars when it's first created
+        project_env_vars: Dict[str, Any] = {}
         on_run_start: List[str] = value_or(cfg.on_run_start, [])
         on_run_end: List[str] = value_or(cfg.on_run_end, [])
 
@@ -444,6 +447,7 @@ class PartialProject(RenderComponents):
             vars=vars_value,
             config_version=cfg.config_version,
             unrendered=unrendered,
+            project_env_vars=project_env_vars,
         )
         # sanity check - this means an internal issue
         project.validate()
@@ -556,6 +560,7 @@ class Project:
     query_comment: QueryComment
     config_version: int
     unrendered: RenderComponents
+    project_env_vars: Dict[str, Any]
 
     @property
     def all_source_paths(self) -> List[str]:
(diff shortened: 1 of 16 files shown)


jtcohen6 (commented):
Just spent some time playing around with these changes. Looking good so far!

core/dbt/parser/manifest.py
```diff
@@ -78,6 +78,8 @@ class ReparseReason(StrEnum):
     project_config_changed = '06_project_config_changed'
     load_file_failure = '07_load_file_failure'
     exception = '08_exception'
+    proj_env_vars_changed = '09_project_env_vars_changed'
+    prof_env_vars_changed = '10_profile_env_vars_changed'
```
jtcohen6, line 82: 👍

jtcohen6 (commented):
Working great in local testing. Good to go from my perspective.

Let's make sure someone is able to give this PR + #4162 a quick code review on Monday morning, so that we can have both in `rc1`.

core/dbt/config/project.py
```diff
@@ -397,6 +398,8 @@ def create_project(self, rendered: RenderComponents) -> 'Project':
             vars_dict = cfg.vars
 
         vars_value = VarProvider(vars_dict)
+        # There will never be any project_env_vars when it's first created
+        project_env_vars: Dict[str, Any] = {}
```
emmyoop, line 402: Is this true?  With the new interactive `init` you can technically put in env_vars.

core/dbt/parser/manifest.py
```diff
@@ -553,6 +555,18 @@ def is_partial_parsable(self, manifest: Manifest) -> Tuple[bool, Optional[str]]:
...
+        if self.manifest.state_check.project_env_vars_hash != \
+                manifest.state_check.project_env_vars_hash:
+            logger.info("Unable to do partial parsing because env vars "
+                        "used in dbt_project.yml have changed")
+            valid = False
+            reparse_reason = ReparseReason.proj_env_vars_changed
+        if self.manifest.state_check.profile_env_vars_hash != \
+                manifest.state_check.profile_env_vars_hash:
+            logger.info("Unable to do partial parsing because env vars "
+                        "used in profiles.yml have changed")
```
emmyoop, line 567: Depending on what goes in first (this or #4055) , this will need to be updated to the new struct logging as well as anywhere else logging is happening.  Nothing to do now - more of a reminder for later.

emmyoop (approved):
LGTM but I'd be happier is someone else took a look too.

core/dbt/config/project.py
gshank: Yes. The project_env_vars only come from rendering, which happens after the object is created. So whether or not env vars exist before the dbt code starts up doesn't matter.

core/dbt/parser/manifest.py
gshank: I think we should have a walkthrough next week since everything is all crunchy right now.

Issue #4179, opened 2021-11-01
Loosely Couple CLI


nathaniel-may:
### Is there an existing feature request for this?

- [X] I have searched the existing issues

### Describe the Feature

I would like to see the dbt cli loosely coupled from the rest of the core code base. This way the cli could be parsed into an object, and the rest of the code can access that object's fields. Today we have the cli parsing arguments and mutating global state without saving the original values, which makes it difficult to respect the inputs users have given us later in the flow.

Additionally, users have noticed how slow the CLI is to start up (e.g. - #4627), and separating out the CLI would allow us to more easily return quickly without loading unnecessary resources when we invoke dbt with invalid arguments or just ask for the version.

### Describe alternatives you've considered

We could just keep adding global mutations of values ad-hoc every time we need them. It's just messy and keeps the cli unnecessary tightly coupled.

### Who will this benefit?

dbt core devs and community contributors.

### Are you interested in contributing this feature?

Yes

nathaniel-may:
Realistically, this same approach should be applied to the cli, all environment variables, and reading files, probably in that order.

jtcohen6:
I know we've discussed this previously: As part of splitting out / rewriting the CLI, we should also consider alternatives to `argparse`. It can do some surprising things today:
- "Skip" arguments in highly confusing ways (https://github.com/dbt-labs/dbt-core/issues/4645)
- "Complete" abbreviated flag names (has caused confusion between similarly-named-yet-different flags, e.g. `--select` + `--selector`)

PR #4219, opened 2021-11-05
Deal with callsites in Main

nathaniel-may:
### Description

There were a few options to deal with the fact that many of these call sites might not have a logger set up, but I decided to go with the least interesting one: do it the same way as everywhere else. Structlog will be created using the defaults, then overriden once we have the information from the command that the user entered. If it has not been configured yet it will use the unconfigured one. The advantage here is that we will not have to change these log call sites once we refactor the cli in #4179. There is an edge case where a keyboard interrupt or an exception could occur before the default initialization of Structlog which happens on inport of the module, but I think it's acceptable to be a little splody when the whole python app hasn't even loaded yet.

### Checklist

### DIFF
--- a/core/dbt/events/types.py
+++ b/core/dbt/events/types.py
@@ -1,4 +1,5 @@
 from abc import ABCMeta, abstractmethod
+import argparse
 from dataclasses import dataclass
 from typing import Any, List, Optional, Dict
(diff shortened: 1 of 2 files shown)


emmyoop (approved):
looks good.

PR #4236, opened 2021-11-08
Issue message instead of exception when patch does not have a matching node

### LINKED ISSUE #3885: Env vars + partial parsing

gshank:
resolves #3885

### Description

Implementation of the env var partial parsing code included a change to throw an exception when there is no matching node for a patch. This reverts that and switches to issuing a message.

### Checklist

### DIFF
--- a/core/dbt/parser/schemas.py
+++ b/core/dbt/parser/schemas.py
@@ -57,6 +57,7 @@ from dbt.parser.generic_test_builders import (
     TestBuilder, GenericTestBlock, TargetBlock, YamlBlock,
     TestBlock, Testable
 )
+from dbt.ui import warning_tag
 from dbt.utils import (
     get_pseudo_test_path, coerce_dict_str
 )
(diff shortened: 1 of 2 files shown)


core/dbt/parser/schemas.py
```diff
@@ -886,11 +887,13 @@ def parse_patch(
                     # re-application of the patch in partial parsing.
                     node.patch_path = source_file.file_id
             else:
-                raise ParsingException(
+                msg = (
                     f"Did not find matching node for patch with name '{patch.name}' "
                     f"in the '{patch.yaml_key}' section of "
                     f"file '{source_file.path.original_file_path}'"
                 )
+                logger.info(msg)
```
jtcohen6, line 895: As a compromise, what do you think about making this:
```python
warn_or_error(msg, log_fmt=warning_tag('{}'))
```

That's what we do in other places when people have unused resources, or tests that depend on a disabled node

jtcohen6 (commented):
Thanks for the quick work! One small comment, otherwise good to go

core/dbt/parser/schemas.py
gshank: Ok

jtcohen6:
I'm just running the `dbt-adapter-tests` suite with `dbt-spark` to be extra sure that this change will fix integration tests over there. I hoped to do it in CircleCI, but they're having a major outage right now, so I'm doing it locally (and slowly) instead. As soon as that happens, I'll give final approval and we can merge.

jtcohen6 (approved):
good to go!

Issue #4245, opened 2021-11-09
[Bug] Table-level freshness should inherit from schema-level freshness when schema-level is null and table-level is missing


jtalmi:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Given the set of four sources seen below:

```
version: 2

sources:
  - name: my_src
    tables:
      - name: src_freshness_default
      - name: src_freshness_disabled
        freshness: null

  - name: my_src2
    freshness: null
    tables:
      - name: src_freshness_default
 
```

The resulting freshness will be:

**my_src.src_freshness_default**: `{'warn_after': None, 'error_after': None, 'filter': None}`
**my_src.src_freshness_disabled**: `None`
**my_src2.src_freshness_default**: `{'warn_after': None, 'error_after': None, 'filter': None}`
**my_src2.src_freshness_disabled**: `None`

my_src2.src_freshness_default does not inherit the `freshness: null`  property from the schema-level when freshness is missing.

### Expected Behavior

When table-level freshness is missing and schema-level freshness is set to `null`, table-level freshness should inherit from schema-level and be set to None instead of the default missing value of `{'warn_after': None, 'error_after': None, 'filter': None}`

### Steps To Reproduce

Using dbt == 0.21, create a .yml file using the above yml and run this as a run operation or analysis:

```
{% for source in graph.sources.values() %}

  schema: {{ source.schema }}
  name: {{ source.name }}
  freshness: {{ source.freshness }}

{% endfor %}
```

### Relevant log output

```shell
schema: my_src
  name: src_freshness_default
  freshness: {'warn_after': None, 'error_after': None, 'filter': None}

  schema: my_src
  name: src_freshness_disabled
  freshness: None

  schema: my_src2
  name: src_freshness_default
  freshness: {'warn_after': None, 'error_after': None, 'filter': None}

  schema: my_src2
  name: src_freshness_default
  freshness: None
```

### Environment

```markdown
- OS: Mac OS Big Sur
- Python: 3.8
- dbt: 0.21
```

### What database are you using dbt with?

snowflake

### Additional Context

https://getdbt.slack.com/archives/C50NEBJGG/p1634829239001900

jtcohen6:
@jtalmi Thanks for opening the issue, and the PR!

PR #4246, opened 2021-11-09
Table-level freshness should inherit from schema-level freshness when table-level is missing and schema-level is null

jtalmi:
resolves https://github.com/dbt-labs/dbt-core/issues/4245

### Description

Currently, when schema-level (base) freshness is set to `null` and table-level freshness is missing (i.e. set to default), merge_freshness returns `{'warn_after': None, 'error_after': None, 'filter': None}`, which is the default value when freshness is missing.

In this case, the source's freshness should equal `None`, as it should inherit from the schema-level freshness, which is set to `null`

### Checklist

### DIFF
--- a/core/dbt/parser/sources.py
+++ b/core/dbt/parser/sources.py
@@ -364,9 +364,9 @@ def merge_freshness(
         merged_freshness.error_after = merged_error_after
         merged_freshness.warn_after = merged_warn_after
         return merged_freshness
-    elif base is None and update is not None:
-        return update
     elif base is None and not update:
         return None
+    elif base is None and update is not None:
+        return update
     else:
         return None
--- a/core/dbt/parser/sources.py
+++ b/core/dbt/parser/sources.py
@@ -364,7 +364,9 @@ def merge_freshness(
         merged_freshness.error_after = merged_error_after
         merged_freshness.warn_after = merged_warn_after
         return merged_freshness
-    elif base is None and not update:
+    elif base is None and update is not None:
         return update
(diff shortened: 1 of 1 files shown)


jtcohen6:
@jtalmi It looks like the CLA bot is taking issue with the GitHub author on these commits. Any chance you can rebase with `git commit --amend --author="Author Name <email@address.com>"`, so that the email matches the email with which you signed the CLA?

jtalmi:
done, let me take a look at adding a test tonight.

jtcohen6:
@jtalmi Could you try squashing/rebasing your previous commits (the ones with `Jonathan Talmi` as the author), into the commits with `jtalmi` as the author, and force pushing? That will hopefully make the CLA bot happy.

Here's an idea for a really basic test. It requires importing a `dbt.parser.sources` method into `test_contracts_graph_unparsed.py`, but it's the cleanest cut I could manage. In [`TestFreshnessThreshold`](https://github.com/dbt-labs/dbt-core/blob/434f3d26789d4bed3f0057c7c7d898e9dbf6687f/test/unit/test_contracts_graph_unparsed.py#L173):
```python
from dbt.parser.sources import merge_freshness

...
class TestFreshnessThreshold(ContractTestCase):
    ...
    def test_merged(self):
    ...
    def test_merged_null_parent(self):
        # parent - null
        t1 = None
        # child - default
        t2 = self.ContractType()
        expected = None
        actual = merge_freshness(t1, t2)
        self.assertEqual(expected, actual)

PR #4249, opened 2021-11-09
Dbt init with provided project name

### LINKED ISSUE #4156: Re-add support for `dbt init <project_name>`
Folks are used to being able to pass project name as a first positional argument to the `dbt init` subcommand. The new approach (new for v1!) is much more fun + interactive, but for muscle memory's sake, let's see if we can't find a way to keep supporting `dbt init my_new_project`, too.

If passed, I think this argument should be used as the implicit answer for "What is the desired project name?" We should set `project_name` accordingly [here](https://github.com/dbt-labs/dbt-core/blob/178f74b753aab0c40d03799b3287a13f35740248/core/dbt/task/init.py#L301), and skip the prompt.

Related ask that we should foresee:
- Ability to skip profile setup entirely, and just create scaffolding for a new project (`--skip-profile-setup`?). This could be used by tools which manage connection setup themselves, such as dbt Cloud
- Ability to create project scaffolding in the current working directory, rather than always creating a new subfolder (#4141)

kadero:
resolves #4156

### Description

Aim of this PR is to support `dbt init <project_name>` and to be able to skip the adapter configuration via `dbt init -s`. Of course, we can also combine the arguments `<project_name>` and `-s`.

Examples:
`dbt init jaffle_shop`
image

`dbt init -s`
image

`dbt init jaffle_shop -s`
image

### Checklist


core/dbt/task/init.py
```diff
@@ -302,26 +302,33 @@ def run(self):
...
+        self.copy_starter_repo(project_name)
+        os.chdir(project_name)
+        with open("dbt_project.yml", "r+") as f:
+            content = f"{f.read()}".format(
+                project_name=project_name,
+                profile_name=project_name
+            )
+            f.seek(0)
+            f.write(content)
+            f.truncate()
```
jtcohen6, line 328: I think it would be great, in the future, to add another log/event here to explain what's happening. That way:
```
$ dbt init my_new_project -s
11:54:50 | [ info  ] | Running with dbt=1.0.0-b2
11:54:51 | [ info  ] | Created starter project named 'my_new_project'
```
That can come in another future pass

jtcohen6 (approved):
Looks great to me, @kadero! I'm going to sneak this in for rc1. Thanks so much for another solid contribution :)

PR #4255, opened 2021-11-09
Env var shim to enable legacy logger

jtcohen6:
resolves https://github.com/dbt-labs/dbt-rpc/issues/60

### Description

We need an escape hatch to use the legacy logger instead, for plugins with complex integrations that aren't worth cutting over (namely, `dbt-rpc`).

I coded this up using an env var, but I think the best version probably looks like an on/off switch (function) in the `logger` module that can be flipped by packages importing `dbt-core` (namely, `dbt-rpc`).

### Checklist

- ~I have updated the `CHANGELOG.md` and added information about my change~

### DIFF
--- a/core/dbt/events/functions.py
+++ b/core/dbt/events/functions.py
@@ -4,7 +4,7 @@ import dbt.events.functions as this  # don't worry I hate it too.
 from dbt.events.types import Cli, Event, File, ShowException
 import dbt.flags as flags
 # TODO this will need to move eventually
-from dbt.logger import SECRET_ENV_PREFIX, make_log_dir_if_missing
+from dbt.logger import SECRET_ENV_PREFIX, make_log_dir_if_missing, GLOBAL_LOGGER
 import io
 from io import StringIO
 import json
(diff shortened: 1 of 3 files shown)


nathaniel-may (commented):
So I'm not totally clear on the approach here. The thing you're turning on is the redirect from the std lib logger. Does the rpc server only make stdlib calls instead of going through our global instance of logbook? If so, this makes sense. If not, I think we should find a different solution.

When this is turned on, it looks like the stdout log format starts using the default std lib logger formatter like so:
`2021-11-09 16:06:45.724968 (MainThread): 11:06:43 | [ info  ] | Done.`

Not a problem, it's just going to look weird.

I also couldn't get an example of this to work locally, but I suspect that we could see some duplicate logs like we did last night when testing the feature branch before we turned off the redirect.

jtcohen6:
@nathaniel-may I tested locally, and I'm pretty sure we just need the conditional check on `logger.disable()`. I'll run integration tests from `dbt-rpc` off this branch again to be sure.

jtcohen6:
Postgres integration tests are failing, in the same way they currently are for the base branch

nathaniel-may:
being a little pedantic, but would you consider `ENABLE_LEGACY_LOGGER` instead of `USE_LEGACY_LOGGER` only because the latter implies to me that we won't be using the new logger at all, which is not true.

jtcohen6:
Just rebased against base branch. Just to be safe, I'll also kick off another run in https://github.com/dbt-labs/dbt-rpc/pull/61

nathaniel-may (approved):

jtcohen6:
All tests passing, including [corresponding dbt-rpc tests](https://github.com/dbt-labs/dbt-rpc/runs/4158219241?check_suite_focus=true)

PR #4256, opened 2021-11-09
Smart Source Freshness Runs

### LINKED ISSUE #4050: Selection method based on source freshness: new `max_loaded_at`, new data
See also: https://github.com/dbt-labs/dbt/issues/2465#issuecomment-763776415,  https://github.com/dbt-labs/dbt/issues/2743#issuecomment-841230958, https://github.com/dbt-labs/dbt/issues/3804#issuecomment-921115079, https://github.com/dbt-labs/dbt/issues/3862#issuecomment-916828824, [Benn's tweet](https://twitter.com/bennstancil/status/1433959272891785216)

Use the `sources.json` artifact to determine which sources have new data since the last dbt build.

Imagine:
```
$ dbt source freshness
$ dbt build -s sources:fresh+ --state my/previous/run/artifacts
```

This would require handling _multiple_ `sources.json`:
- the version from the last run in the `--state` directory
- the one from just now in the `target/` directory

Then, dbt would compare the high watermarks recorded in each, and select all source tables with a greater/changed `max_loaded_at`.

### Questions

- Should we reconsider the inclusion of the `freshness` task within `dbt build`? I think it would be _tricky_ to check source freshness _and_ perform dynamic selection _and_ build those now-selected resources, all in one invocation. It would require blurring lines that are pretty well delineated today (selection &rarr; execution &rarr; artifacts).
- Should we reconsider the consolidation of `sources.json` into `run_results.json`? It seems like the separation would actually be a plus, for the sake of this feature, but I'm sure we could manage either way
- Would it be better to perform this comparison _outside_ dbt-core (e.g. dbt Cloud metadata API), and then simply pass the set of sources that have new data? If we know that `source:stripe` and `source:salesforce.account` have new data, it's easy enough to then define a job with `dbt build -s source:stripe+ source:salesforce.account+`.

### Big idea

Coupled with `state:modified` and `result:<status>` (https://github.com/dbt-labs/dbt/pull/4017), you could dynamically select the part of your DAG that's relevant by comparing the current state of your sources/project and the artifacts from your last run.

```yml
selectors:
  - name: one_selector_to_rule_them_all
    default: true
    description: Select sub-DAG based on new source data, changed logic, &/or failure in last run
    definition:
      union:
      
        # check freshness + test all sources
        - 'source:*'
      
        # new source data + table/incremental
        - intersection:
            - union:
                - 'config.materialized:table'
                - 'config.materialized:incremental'
            - method: source
              value: fresh  # imagine
              children: true
      
        # changed logic (need this for views + MVs too)
        - method: state
          value: modified
          children: true

        # failed last time
        - union:
            - 'result:error'
            - 'result:skipped'
```
```
$ export DBT_ARTIFACT_STATE_PATH=my/previous/run/artifacts
$ dbt source freshness
$ dbt build
```

Schedule that to run every 5 minutes, set up solid slack notifications on failures, and put on a good album.

While this is very cool, it does take its cue from the source data loaders—whenever they've ingested new rows, we're going to build downstream tables. This is the inverse of defining freshness / latency expectations for specific "final" models, and using that as the primary input to deployment/scheduling. That still seems like a cool and important thing to do... perhaps via `exposures`??

sungchun12:
resolves #4050

### Description

When a user wants to run dbt models based on freshest data possible, they can do so in an elegant way: `dbt build --select source_status:fresher+ --defer --state <previous state path>`

The ultimate goal is to save precious time and money spent reprocessing stale data. Just because the extract and load process is finished doesn't mean it provided fresher data.

## Example Use Case
**dbt Cloud Job A:** runs source freshness every 30 minutes. It will serve as the previous state `sources.json`.
Command Steps:
- `dbt source freshness`
- `dbt build`

**dbt Cloud Job B:** runs every 35 minutes and defers to Job A above. If the `max_loaded_at` date for the current state source freshness check is greater than the `max_loaded_at` for the previous state, this command will run the downstream models in scope. The job should only have to look within the `./target` folder of Job A. Using the `build` command tests the sources for quality, so if those checks fail, downstream nodes will not run if you use [fail fast](https://docs.getdbt.com/reference/global-configs#failing-fast).
Command Steps:
- `dbt source freshness`
- `dbt build --select source_status:fresher+ --defer --state <previous state path>`

image

**Considerations:**
- Will dbt Cloud have to add new functionality to store previous state given that dbt command steps are stateless by default? It shouldn't. Since [4721](https://github.com/dbt-labs/dbt-cloud/pull/4721), dbt Cloud will include sources.json (if available) in the artifacts it copies during deferral. We'll avoid the issue we experienced with the result:<status> selector + run_results.json, because the relevant artifact in this case is sources.json — which doesn't get overridden by subsequent job steps, namely docs generate

### Checklist


jtcohen6:
@sungchun12 @anaisvaillant Very cool, very quick work!

I took a quick look, and I noticed a difference between the implementation here and the idea described in #4050, though please tell me if I'm missing something.

The proposal in #4050 is to compare max_loaded_at between `<state_path>/sources.json` (previous run) and `<target_path>/sources.json`, to figure out if the "bookmark" for that source has moved. Two distinctions:
- We need to read in two different `sources.json` files. I think that _might_ look like registering _both_ `PreviousState.sources` (using `PreviousState.path`), and also something like `CurrentState.sources` (using `target_path`). The latter isn't something that exists just yet; we might need to think a bit about where it ought to be defined and pull in.
- I think we want the source's `max_loaded_at`, rather than its `status`. It could be fresh one or both times, without having moved the bookmark.

Once we have a way to register both, and perform the comparison between their `max_loaded_at`, I envision something like (pseudo-code):
```python
class SourceBookmarkSelectorMethod(SelectorMethod):
    def search(
        self, included_nodes: Set[UniqueId], selector: str
    ) -> Iterator[UniqueId]:
        if self.previous_state is None or self.previous_state.sources is None
        or self.current_state is None or self.current_state.sources is None:
            raise InternalException(
                'Comparison sources missing in {self.previous_state.path} or {self.current_state.path}
            )
        # select any "current" source (just freshness-checked) that has a greater max_loaded_at
        # than the same source (previously freshness-checked), or if the source did not previously exist
        previous_sources = self.previous_state.sources
        current_sources = self.current_state.sources
        matches = set(
            current_source.unique_id for current_source in current_sources
            if (
                current_source.unique_id not in previous_sources or
                current_source.max_loaded_at > previous_sources[current_source.unique_id].max_loaded_at
            )
        )
        for node, real_node in self.all_nodes(included_nodes):
            if node in matches:
                yield node
```

sungchun12:
@jtcohen6 You raise a solid point that a `source_status:pass+` without a `max_loaded_at` bookmark quickly becomes a wasteful run of children nodes if it's run every 5 minutes. I'll figure out where to inject `CurrentState.sources`. 

**Options on selector syntax and mechanics:** @jtcohen6, **Do you agree with my recommendation?** 
- `dbt run --select source_status_max_loaded_at:pass+` This provides explicit intuition for the developer upfront that source status matters, and it'll only run children nodes if the max loaded at date has a date greater than the previous source freshness date. I still see use cases where `warn` and `error` statuses are used to run children nodes rather than `max_loaded_at` alone.  However, it's a long name for a selector flag. **Sung's Recommendation**

- `dbt run --select source_status:pass+`. This implies BOTH the source status AND the max loaded at date are explicit considerations before children nodes are run. Still the same flag. However, a developer will have to look at documentation to understand that both mechanisms are at play. 

- Two separate selectors that include `source_status:pass+` AND `source_status_max_loaded_at: pass+`. This way, the developer can choose their own mechanism to run children nodes on status criteria OR status and max loaded at criteria. 

- I want to avoid using `sources:fresh+` because it requires the developer to research on their own what mechanisms are used to run children nodes.

jtcohen6:
@sungchun12 I think I agree with your first recommendation. I want to hear a bit more about:

> This provides explicit intuition for the developer upfront that source status matters... I still see use cases where `warn` and `error` statuses are used to run children nodes rather than `max_loaded_at` alone.

While source freshness statuses (`pass`, `warn`, `error`) matter tremendously for team SLAs and downstream metadata (e.g. exposure status reporting), I don't see how I'd want to use them to inform selection criteria. From my point of view, I just want the comparison between old + new `max_loaded_at` values. Does source X have new data since the last time I ran my models? If so, I want to select models downstream of source X to re-run; if not, I don't.

I feel that holds true regardless of whether the old `max_loaded_at` was past the `warn` threshold, the new `max_loaded_at` is past the `error` threshold, or both are safely within my freshness expectations and `pass`. In all cases, I just want to know whether the bookmark has moved.

sungchun12:
@jtcohen6 Thanks for the thorough reply! I also updated the selector flag to be shorter and coherent `source_refresh:pass+`. I'll provide examples below of how this feature will be used in the wild in response to your point on `max_loaded_at` criteria. 

I agree that previous state `pass, warn, error` thresholds are irrelevant for source refresh jobs. I still recommend `pass, warn, error` thresholds are relevant for **current state**.

**Open Questions**
- Do you agree with this recommendation?
- When does this feature need to be merged to make it to the first release of `v1.0.0`?

**Previous State dbt job**
```
$ dbt source freshness # this job only checks data freshness
```

**Current State dbt job**
```
$ dbt source freshness # run this again to compare `max_loaded_at` to the previous state
$ dbt build -s source_refresh:pass+ source_refresh:warn+ --state my/previous/run/artifacts # refresh children nodes of sources that have `pass` or `warn` AND have fresher `max_loaded_at` dates ONLY in the current state dbt job
```

Today, a dbt developer has an intuition that children nodes of sources should run based on `pass, warn, error`. For the majority, `pass, warn` are the driving criteria to run children nodes. If this feature for source refreshes removes that criteria and only considers `max_loaded_at`, it takes away that driving criteria. Including the `pass, warn` criteria as concurrent selectors gives the developer more choices to mix and match use cases. 

I don't see a compelling use case where a source refresh run should do the below, assuming all sources have `error`.

**Previous State dbt job**
```
$ dbt source freshness # this job only checks data freshness
```

**Current State dbt job**
```
$ dbt source freshness # run this again to compare `max_loaded_at` to the previous state. All sources have fresher `max_loaded_at` dates, but all are `error`
$ dbt build -s sources:fresh+ --state my/previous/run/artifacts # refresh children nodes of sources that have fresher `max_loaded_at` dates even if they are all `error`
```

Developers are expecting source refresh runs to be **fresh AND meeting thresholds(ex: `pass, warn`)** not only fresh. 

I anticipate developers getting confused on why children nodes are being run if all sources have freshness errors. I also imagine lots of downstream tests failing because data isn't meeting the source `pass` criteria. However, if you have something like the below, this gives clarity to the logs and troubleshooting process. 
```
$ dbt source freshness # run this again to compare `max_loaded_at` to the previous state. All sources have fresher `max_loaded_at` dates, but all are `error`
$ dbt build -s source_refresh:error+ --state my/previous/run/artifacts # refresh children nodes of sources that have fresher `max_loaded_at` dates explicitly knowing sources are not passing freshness
```

sungchun12:
@jtcohen6, can you take a look at our implementation because it's changing the behavior of how `dbt source freshness` is now a stateful command across 2 invocations?

Also, I added a `path_utils.py` to more easily import a global path for the `archive_sources` path for previous invocations of `dbt source freshness`.

Before we proceed with testing, @matt-winkler, @anaisvaillant, and I want to make sure the above mechanics won't mess with the promise of how `v1` should "look and feel".

sungchun12:
@jtcohen6 Any feedback on our approach before we start testing?

jtcohen6:
@sungchun12 Thanks for all your work on this! I know you're waiting on me for the go-ahead. I'd like to carve out some time to dive into what's at play in the latest implementation. I'd also like to bring one of the foks on the Execution sub-team to ride along with me. In the meantime, I really appreciate your patience, and all the thought you've been putting into this

matt-winkler:
@jtcohen6 anyone on the Execution team been assigned to this to date? IMO this is a large step towards solving for coordination with data loaders vs. using external orchestration tools, so am pretty excited to see it take further shape!

core/dbt/task/freshness.py
```diff
@@ -190,6 +190,15 @@ def result_path(self):
             return os.path.realpath(self.args.output)
         else:
             return os.path.join(self.config.target_path, RESULT_FILE_NAME)
+    
+    def archive_path(self):
```
jtcohen6, line 194: Not a fan of this behavior, and it won't be needed if we use `target_path` + `state_path` (instead of `state_path` + `state_path/archive_sources`)

core/dbt/graph/selector_methods.py
```diff
@@ -577,6 +578,39 @@ def search(
...
+    def search(
+        self, included_nodes: Set[UniqueId], selector: str
+    ) -> Iterator[UniqueId]:
+
+        if self.previous_state is None or \
+           self.previous_state.current_sources is None or \
+           self.previous_state.archive_sources is None:
+            raise InternalException(
+                'No previous state comparison freshness results in sources.json'
+            )
```
jtcohen6, line 591: If we're missing either source result (previous or current), is it better to raise this explicit error, or to return all sources as selected?

This question could apply equally to `state:` and `result:` selection; we might end up with different answers for each, though I'd obviously prefer consistency.

jtcohen6 (commented):
Spent a few hours playing around with this — it's very neat! — if a bit janky to set up, and to debug when things go wrong. It emanates "big if true" energy.

Let's keep talking about this one, especially as it overlaps with some other conversations we're having right now around partial DAG execution (cc @schottj @barryaron).

> Will dbt Cloud have to add new functionality to store previous state given that dbt command steps are stateless by default?

I don't think so, thanks to the recent changes you all advocated for, to power `result`-based selection. This one might work even more cleanly, since we don't need to worry about `sources.json` being overwritten by subsequent/unrelated run steps. Just check the "Run Source Freshness" box, write a yaml selector, `dbt build`, and away we go...

cc @ChenyuLInx: I think you'll find this one interesting, given your recent work on the `state:` selection method. I'd love to grab some time and get you up to speed.

core/dbt/graph/selector_methods.py
```diff
@@ -577,6 +578,39 @@ def search(
...
+        }
+
+        matches = set()
+        for unique_id in current_state_sources:
+            if unique_id not in archive_state_sources:
+                matches.add(unique_id)
+            elif selector == 'yes' and current_state_sources.get(unique_id) > archive_state_sources.get(unique_id):
+                matches.add(unique_id)
+            elif selector == 'no' and current_state_sources.get(unique_id) <= archive_state_sources.get(unique_id):
+                matches.add(unique_id)
```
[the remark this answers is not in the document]
jtcohen6: Ah - I now remember discussing the case where someone does _not_ want to build models downstream of a source, unless all tables in that source are fresh, so as to avoid uncoordinated updates.

Then, the inclusion of `source_fresh:no` exists to support:
```
dbt build -s source_fresh:yes --exclude source_fresh:no
```

Even then, I think this is likely to break down. Suppose we have three runs (A, B, C), each of which checks source freshness and defers to the previous run. If half of the sources update between A &rarr; B, and half of the sources update between B &rarr; C — we don't end up updating any downstream models, because in neither comparison (A &rarr; B, B &rarr; C) do _all_ source tables appear "fresher" than they were before.

To actually solve for that aim, I think we'd need to coordinate source tables with previous model build times. ("Have all source tables upstream of model X been updated since model X last run?") That information would be available from the previous `run_results.json` — but now we're talking about even more moving pieces, plus factoring in graph traversal (?).

At this level of strictness, someone would be better off triggering `dbt build source:specific_src+` when (and only when) they've confirmed that `specific_src` has finished updating all tables.

core/dbt/contracts/state.py
```diff
@@ -1,15 +1,21 @@
 from pathlib import Path
 from .graph.manifest import WritableManifest
+
 from .results import RunResultsArtifact
+from .results import FreshnessExecutionResultArtifact
 from typing import Optional
+
 from dbt.exceptions import IncompatibleSchemaException
 
+from dbt.path_utils import FRESHNESS_RESULT_ARCHIVE_DIR_NAME
```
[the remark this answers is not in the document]
matt-winkler: @jtcohen6 Ah interesting, we'd considered something like this option before the latest iteration, though that was also hard coded. 

A few additional points to consider.

- How will the user manage previous and current state paths? Wouldn't using `target` for both of those cause dbt to overwrite the state it needs to make the freshness determination per the intended use here?
- What does the availability of these artifacts look like in a dbt cloud deployment job? Do we (theoretically) have access to both the previous and current state, and what are the configurations in which that may not be true?

core/dbt/task/freshness.py
matt-winkler: Love it - `archive_sources` didn't feel great :). Noting a question from above about how a user should configure `target_path` + `state_path` to accomplish this.

core/dbt/graph/selector_methods.py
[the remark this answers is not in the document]
sungchun12: Agreed on removing the complex orchestration example from scope.

We will change the flag to something like: `source_fresher`

Example: `dbt build --select source_fresher`

core/dbt/contracts/state.py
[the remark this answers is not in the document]
jtcohen6: Good questions!

> How will the user manage previous and current state paths? Wouldn't using target for both of those cause dbt to overwrite the state it needs to make the freshness determination per the intended use here?

I believe the user should manage _previous_ state via the same "state" configuration (`--state`/`DBT_ARTIFACT_STATE`) as they do today. I'm advocating for using `target/` as the default path for _current_ state _only_ — i.e. source freshness results, based on the `source freshness` step that just ran. Because `sources.json` is a standalone artifact (unlike `run_results.json`) that is only written by the `source freshness` task, dbt will only overwrite the "current state" (`target/sources.json`) if you run another `dbt source freshness` task. That actually seems right to me! You now have even-more-current freshness results.

> What does the availability of these artifacts look like in a dbt cloud deployment job? Do we (theoretically) have access to both the previous and current state, and what are the configurations in which that may not be true?

Yes! If job deferral is set up, dbt Cloud makes "previous" state accessible by copying artifacts from the deferred-to job and setting `DBT_ARTIFACT_STATE`. Then, "current" state should always be accessible, insofar as the `target/` directory (from previous run steps) is accessible.

Worth confirming / testing to be sure. When this PR is merge-ready, we could test (after merging into `main`) by using the `development (HEAD)` version of dbt Core in dbt Cloud.

sungchun12:
@matt-winkler @anaisvaillant I updated this pull request by: removing the archive path mechanic, introducing a CurrentState class to default to the `target/sources.json` path for current state, and verifying basic functionality works. 

Got the basic commands to work below. We'll have to think through `source_fresher:` and `source_fresher:+` vs. `source_fresher` and `source_fresher+` implementations.

We'll also have think through how to enable users or not to change the current state source freshness path for this `source_fresher` run(think: `custom-path-that's-not-target/sources.json`)

We can talk details this week!

image
image

core/dbt/contracts/state.py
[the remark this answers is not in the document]
sungchun12: @matt-winkler I implemented the current state target path `sources.json` logic and verified that custom target paths work!

You can branch off this branch, and we can merge your branch into this branch when your changes are complete.

image
image
image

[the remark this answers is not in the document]
sungchun12: FYI: I removed the `CurrentState` class implementation as it would force a refactoring of many other method/class arguments to include it. That's more scope that doesn't fit with this work.

matt-winkler:
Hey @sungchun12 After working through this, I'm no longer sure `--select source_fresher+` is the right selection syntax. That looks like a `model` node selection to me, and I don't know if we want to disturb the default behavior there. It would be weird for someone to have a model in their project called `source_fresher`, but still :). 

Instead, what makes sense to me is a hybrid of the approaches we've taken so far: `--select source_status:fresher+`. In my brain, this tells the user intuitively that they're in the mode of selection based on status of their sources, and those sources must be fresher than the last `dbt source freshness` run in order to be selected.
 
Interested in your thoughts here.

sungchun12:
@matt-winkler that is such a great call out. That totally looks like model selection syntax and would confuse a developer several times over. I like your suggestion better!

I tested it out locally and the look and feel is great!
image

I'll go ahead and put together a detailed testing plan and let you and @anaisvaillant know more in notion.

@jtcohen6 is it possible to get this part of [v1.1](https://github.com/dbt-labs/dbt-core/milestone/70)?

I expect us this get this merged in by 3/18/2022.

Any other testing framework changes that we should incorporate? I noticed we're moving everything to a brand new framework.

core/dbt/contracts/state.py
```diff
@@ -1,30 +1,48 @@
...
+        self.sources: Optional[FreshnessExecutionResultArtifact] = None
+        self.sources_current: Optional[FreshnessExecutionResultArtifact] = None
 
-        manifest_path = self.path / "manifest.json"
+        manifest_path = self.path / 'manifest.json'
         if manifest_path.exists() and manifest_path.is_file():
             try:
-                # we want to bail with an error if schema versions don't match
-                self.manifest = WritableManifest.read_and_check_versions(str(manifest_path))
+                self.manifest = WritableManifest.read(str(manifest_path))
```
jtcohen6, line 21: We want this to use `read_and_check_versions` (instead of `read`), to raise a better exception when the version mismatch is preempted by a contract mismatch. See https://github.com/dbt-labs/dbt-core/pull/4497.

If you swap out all instances of `read` for `read_and_check_versions`, that will get `test_postgres_previous_version_manifest` passing again.

core/dbt/contracts/state.py
```diff
@@ -1,30 +1,48 @@
...
 from .graph.manifest import WritableManifest
 from .results import RunResultsArtifact
+from .results import FreshnessExecutionResultArtifact
 from typing import Optional
 from dbt.exceptions import IncompatibleSchemaException
 
 
 class PreviousState:
-    def __init__(self, path: Path):
+    def __init__(self, path: Path, current_path: Path):
```
jtcohen6, line 10: Two choices to get the unit tests passing:
1. Don't require `current_path` as an argument to initialize `PreviousState`, by setting the default arg value here to `None` and by adding an `if self.sources_current` conditional check below (line 34). This is the approach I prefer, as it's much simpler.
2. Every time we initialize `PreviousState` (including [`test/unit/test_graph_selector_methods.py:790`](https://github.com/dbt-labs/dbt-core/blob/5071b00baac3ff2a9bcce565d09661605de52654/test/unit/test_graph_selector_methods.py#L790)), pass in a placeholder value for `current_path`

test/integration/042_sources_tests/test_source_fresher_state.py
```diff
@@ -0,0 +1,685 @@
...
+import yaml
+
+from dbt.exceptions import ParsingException
+import dbt.tracking
+import dbt.version
+from test.integration.base import DBTIntegrationTest, use_profile, AnyFloat, \
+    AnyStringWith
+
+
+class BaseSourcesTest(DBTIntegrationTest):
```
jtcohen6, line 14: We've _just_ converted `042_sources_tests` to use the new `pytest`-based framework (https://github.com/dbt-labs/dbt-core/pull/4935).

I'd say you have two valid options:
- Try rewriting these tests to use the new framework — I think it's a lot nicer! perhaps not all that different, for what you're trying to do here
- Pop this test over to `062_defer_state_tests`, which hasn't yet been converted (and is arguably closer kin for this functionality, anyway)

jtcohen6 (commented):
Ok @sungchun12 @matt-winkler @anaisvaillant: I'm down to give this a go, as "experimental" functionality, and call it out in the docs as such. It's going to have rough edges, and it's no substitute for longer-term thinking about how we can strategically run sub-DAGs of larger projects — but I believe it meets an immediate need, and it will build appetite among users for even more creative, bigger-picture thinking on this subject.

Would you be able to:
- Get the tests passing (comments)
- Add a changelog entry using `changie` (explained [here](https://github.com/dbt-labs/dbt-core/blob/main/CONTRIBUTING.md#adding-changelog-entry))

core/dbt/contracts/state.py
sungchun12: I went with choice 2 as `current_path` is best set within `runnable.py`. If not, we'll have to do lot of overhead implementations to avoid circular logic.

core/dbt/contracts/state.py
sungchun12: Done!

test/integration/042_sources_tests/test_source_fresher_state.py
sungchun12: We'll use the new pytest framework. It's worth learning upfront for many more PRs to come 😉

sungchun12:
Hey @matt-winkler @anaisvaillant,

I added delegated test cases as comments. Since this is a new pytest framework, I won't pre-optimize by creating fixture handlers and fancy setup utilities upfront. We can see where we can build reusable components as we develop our own test code. 

All specific plans and notes are contained within our internal notion including due dates: [here](https://www.notion.so/dbtlabs/Testing-plans-03b1c65f9f5b476088c02bd4b6048fcc)

tests/functional/sources/test_source_fresher_state.py
```diff
@@ -0,0 +1,727 @@
...
+    error_models__schema_yml,
+    models__newly_added_model_sql,
+    models__newly_added_error_model_sql,
+)
+
+
+# TODO: We may create utility classes to handle reusable fixtures.
+def copy_to_previous_state():
+    shutil.copyfile("target/manifest.json", "previous_state/manifest.json")
+    shutil.copyfile("target/run_results.json", "previous_state/run_results.json")
```
jtcohen6, line 23: Not sure if Windows will be happy about this. We do have a `copy_file` utility [here](https://github.com/dbt-labs/dbt-core/blob/83612a98b733a02b78c55648123456650ee43a04/core/dbt/tests/util.py#L97), but it looks like it's doing pretty much the same thing.

tests/functional/sources/test_source_fresher_state.py
```diff
@@ -0,0 +1,727 @@
...
+
+    def __repr__(self):
+        return "AnyStringWith<{!r}>".format(self.contains)
+
+
+class AnyFloat:
+    """Any float. Use this in assertEqual() calls to assert that it is a float."""
+
+    def __eq__(self, other):
+        return isinstance(other, float)
```
jtcohen6, line 48: These were copied from https://github.com/dbt-labs/dbt-core/pull/4935. Maybe they eventually want to move into `dbt/core/tests/util.py`? (cc @gshank)

tests/functional/sources/test_source_fresher_state.py
```diff
@@ -0,0 +1,727 @@
...
+            for c in ("favorite_color", "id", "first_name", "email", "ip_address", "updated_at")
+        )
+        kwargs = {
+            "schema": project.test_schema,
+            "time": timestr,
+            "id": insert_id,
+            "source": project.adapter.quote("source"),
+            "quoted_columns": quoted_columns,
+        }
+        raw_sql = """INSERT INTO {schema}.{source}
```
jtcohen6, line 83: Note for future reader: This probably isn't going to fly on other databases, but that's okay, the functionality being tested here / added in this PR is node selection, not adapter-specific

jtcohen6 (approved):
Thanks for all the hard work here @sungchun12 @matt-winkler @anaisvaillant!

The test coverage looks solid—thanks for all the extra work to adapt to the latest framework. We may want some more annotation and light refactoring, but I'm appreciative of the work that you all have already put into this, with the desire to ship this first experimental version in v1.1.

Let's ship it :)

PR #4266, opened 2021-11-10
Minor Cleanup of Structured Logging Module

nathaniel-may:
### Description

1 - Cleaned up some of the code.

2- Fixed the adapter interface in here since post-cleanup it wasn't type checking. In the new system, the following logger call would have thrown a runtime exception before this change but is almost certainly present in the wild since it was found in our adapter tests:
`logger.info("ran this sql: {}", sql)`

~Rather than re-implement every way a python logger could be called, I just make the std lib logger construct the message component of the log line and hand it back with a stringio buffer.~ <- this strategy did not work because the std lib logger does not use new-style string formatting, only old-style.

I dug into some logger libraries and found that this string formatting is really the only thing that `*args` is doing. Python prefers it this way so it can be done lazily which we are already doing via a function. I apply these arguments with new-style formatting via `str::format` to maintain compatibility with the former logbook logger.

### Checklist

### DIFF
--- a/core/dbt/events/adapter_endpoint.py
+++ b/core/dbt/events/adapter_endpoint.py
@@ -17,7 +17,7 @@ class AdapterLogger():
         stack_info: Any = None,
         extra: Any = None
     ):
-        event = AdapterEventDebug(name=self.name, raw_msg=msg)
+        event = AdapterEventDebug(self.name, msg)
 
         event.exc_info = exc_info
         event.stack_info = stack_info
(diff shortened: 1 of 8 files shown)


leahwicz (approved):

emmyoop (approved):
Love how much cleaner it is.

emmyoop (approved):
The adapter fixes you implemented made me realize we should have generic adapter types, similar to what's in `test_types.py` for new adapters(or anyone inclined to make more updates than the few lines).  Seems off to have new adapters use the old logging format.

iknox-fa (changes requested):
Just swap to using the built-in get method on those dicts and 👍

nathaniel-may:
@emmyoop if I'm understanding you correctly, I think that's exactly what we have already with the subclasses of `types.py::AdapterEventBase`. Do you mean something different than that?

iknox-fa (commented):
LGTM!

emmyoop (approved):

iknox-fa (approved):

PR #4268, opened 2021-11-10
add codes to log events

nathaniel-may:
wait to merge into the feature branch till #4266 is in.

### Description

Uses mypy to enforce codes that uniquely identify each log event. This allows consumers of structured logs to rely on these codes while giving developers the freedom to change class names.

The proposed format for codes is one capital letter followed by 3 digits:
    aka `^[A-Z][0-9]{3}` E.g. - `A123`, `E003` etc.

The letter is determined by where the log line was fired. e.g. - parsing, node status, sql, adapter, deps etc.

### TODO

- [x] get format approved by team
- [x] add concrete codes to each class
- [ ] change base branch of PR back to feature/logging-phase2

### Checklist

### DIFF
--- a/core/dbt/events/base_types.py
+++ b/core/dbt/events/base_types.py
@@ -1,6 +1,6 @@
 from abc import ABCMeta, abstractmethod
-from datetime import datetime
 from dataclasses import dataclass
+from datetime import datetime
 import os
 from typing import Any
 
--- a/core/dbt/events/base_types.py
+++ b/core/dbt/events/base_types.py
@@ -5,7 +5,11 @@ import os
 from typing import Any
 
 
-# types to represent log levels
+# # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # #
+# These base types define the _required structure_ for the concrete event #
+# types defined in types.py                                               #
+# # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # #
+
 
 # in preparation for #3977
 class TestLevel():
--- a/core/dbt/events/base_types.py
+++ b/core/dbt/events/base_types.py
@@ -61,18 +61,18 @@ class Event(metaclass=ABCMeta):
     @classmethod
     @abstractmethod
     def code(cls) -> str:
-        raise Exception("msg not implemented for event")
+        raise Exception("code() not implemented for event")
 
     # do not define this yourself. inherit it from one of the above level types.
     @abstractmethod
     def level_tag(self) -> str:
-        raise Exception("level_tag not implemented for event")
+        raise Exception("level_tag() not implemented for event")
 
     # Solely the human readable message. Timestamps and formatting will be added by the logger.
     # Must override yourself
     @abstractmethod
     def message(self) -> str:
-        raise Exception("msg not implemented for event")
+        raise Exception("message() not implemented for event")
 
     # returns a dictionary representation of the event fields. You must specify which of the
     # available messages you would like to use (i.e. - e.message, e.cli_msg(), e.file_msg())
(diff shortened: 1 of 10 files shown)


leahwicz (approved):

core/dbt/events/types.py
```diff
@@ -266,7 +266,7 @@ def message(self) -> str:
 
 @dataclass
 class SystemReportReturnCode(DebugLevel, Cli, File):
-    code: int
+    returncode: int
 
     def message(self) -> str:
         return f"command return code={self.code}"
```
emmyoop, line 272: Should use the renamed version - `self.returncode`

core/dbt/events/types.py
```diff
@@ -266,7 +266,7 @@ def message(self) -> str:
 
 @dataclass
 class SystemReportReturnCode(DebugLevel, Cli, File):
-    code: int
+    returncode: int
```
emmyoop, line 269: Why is this an int instead of a string?

emmyoop (changes requested):
Is there a difference between `code` and `returncode`?

core/dbt/events/types.py
nathaniel-may: yes! thank you.

core/dbt/events/types.py
nathaniel-may: I think I just guessed. I'll change it to a string. Mypy should tell us once we put all the concrete codes in place and satisfy those abstract method errors.

nathaniel-may:
@emmyoop `returncode` is specific to the log event `SystemReportReturnCode`. `code` on the otherhand is a new field that mypy will enforce is on every log event to uniquely identify each event.

iknox-fa:
A proposal for event code prefixes:
| Code | Description          |
|------|----------------------|
| A    | Pre-project loading  |
| B    | DB adapter setup     |
| C    | Project parsing      |
| D    | Deps generation      |
| E    | Node testing         |
| F    | Node processing      |
| G    | Misc                 |
|      |                      |

core/dbt/events/base_types.py
```diff
@@ -0,0 +1,102 @@
...
+    # fields that should be on all events with their default implementations
+    ts: datetime = datetime.now()
+    pid: int = os.getpid()
+
+    # four digit string code that uniquely identifies this type of event
+    # uniqueness and valid characters are enforced by tests
+    @classmethod
+    @abstractmethod
+    def code(cls) -> str:
+        raise Exception("code() not implemented for event")
```
nathaniel-may, line 64: Here's the abstract method that's doing the enforcement via mypy. I couldn't figure out how to make mypy check that subclasses have defined an abstract property even though the internet seems to think you can do that. You'll need to do one of two things to get this to work:
1. figure out how to make an abstract property such that if you delete `code` from one of the concrete classes, mypy fails.
2. change all the codes from fields to functions that override this one in each of the concrete classes.

core/dbt/events/types.py
```diff
@@ -30,11 +30,12 @@
 # | W    | Node testing        |
 # | Y    | Post processing     |
 # | Z    | Misc                |
+# | T    | Test only           |
 #
 # The basic idea is that event codes roughly translate to the natural order of running a dbt task
 
 # can't use ABCs with @dataclass because of https://github.com/python/mypy/issues/5374
-@dataclass
+@dataclass  # type: ignore
```
nathaniel-may, line 38: oh wow I didn't even think of doing this. we can get rid of the comment that says we can't do this then.

core/dbt/events/base_types.py
nathaniel-may: I just tried this by commenting out the code on one of the concrete tests and running mypy, and it worked. I'm not sure why though. @iknox-fa do you know why a @dataclass field named code in the concrete classes counts as overriding this abstract method??

iknox-fa: Because it's now an abstract property.... and data fields are just automagic properties generated by an automagic `__init__` method.

nathaniel-may: too implicit for my liking, but I'm glad it works.

PR #4270, opened 2021-11-11
Feature/3997 profiles test selection flag

### LINKED ISSUE #3997: Expand --greedy test selection to have a flag in profile.yml file
### Describe the feature
Starting in 0.20.0 test selection was changed to be less greedy meaning that things like relationship tests which depend on more than one model are not selected by default on single model test runs. @joellabes has a great [PR](https://github.com/dbt-labs/dbt/pull/3833) to enable a `--greedy` flag in 0.21.0. It would be a great if you could enable this feature the same way you could enable partial parse or other configuration options and make it the default. 

The alternative, which is the current system, is to just have `--greedy ` on each test run. There's nothing explicitly wrong with this but it is tedious.

### Additional context
Is this feature database-specific? Which database(s) is/are relevant? Please include any other relevant context here.

### Who will this benefit?
To my mind this will benefit a lot of dbt users; its always better imho to over test rather than undertest and since we have many tests that rely on multiple models, we prefer greedy test selection to less greedy selection. Of course this is a matter of preference though I know from seeing other people asking on slack that others feel similarly.

I've never contributed to a public repo actually but I love using dbt so if I could get some help getting started I'd be super excited to contribute!

VersusFacit:
resolves #3997 

### Description

This is an "extra-credit" roundup of Jerco's comments in PR #4104 . Namely, this unifies the indirect_selection and eagerly_expand predicates. It also uses an enum. The other defining change is the movement towards a flag that can be a DBT_ prefixed env var. In deferring the boolean-ification of this value, it makes the logic quite a bit easier to follow: you're tracing an Enum through the whole thing rather than demanding a sort of pseudo-boolean just for sake of conformity.  

All unit and integration tests relevant to this flag are passing locally; includes newly added unit tests and already-present integration tests 066_* from the last PR.

Open concerns:
- What would be the next release candidate for the changelog? 

### Checklist

### DIFF
--- a/core/dbt/graph/selector_spec.py
+++ b/core/dbt/graph/selector_spec.py
@@ -1,6 +1,5 @@
 import os
 import re
-import enum
 from abc import ABCMeta, abstractmethod
 from dataclasses import dataclass
 from dbt.dataclass_schema import StrEnum
@@ -23,9 +22,11 @@ RAW_SELECTOR_PATTERN = re.compile(
 )
 SELECTOR_METHOD_SEPARATOR = '.'
 
+
 class IndirectSelection(StrEnum):
     Eager = 'eager'
     Cautious = 'cautious'
+
 
 def _probably_path(value: str):
     """Decide if value is probably a path. Windows has two path separators, so
@@ -109,7 +110,8 @@ class SelectionCriteria:
 
     @classmethod
     def selection_criteria_from_dict(
-        cls, raw: Any, dct: Dict[str, Any], indirect_selection: IndirectSelection = IndirectSelection.Eager
+        cls, raw: Any, dct: Dict[str, Any],
+        indirect_selection: IndirectSelection = IndirectSelection.Eager
     ) -> 'SelectionCriteria':
         if 'value' not in dct:
             raise RuntimeException(
@@ -159,7 +161,10 @@ class SelectionCriteria:
         return dct
 
     @classmethod
-    def from_single_spec(cls, raw: str, indirect_selection: IndirectSelection = IndirectSelection.Eager) -> 'SelectionCriteria':
+    def from_single_spec(
+        cls, raw: str,
+        indirect_selection: IndirectSelection = IndirectSelection.Eager
+    ) -> 'SelectionCriteria':
         result = RAW_SELECTOR_PATTERN.match(raw)
         if result is None:
             # bad spec!
--- a/core/dbt/graph/selector_spec.py
+++ b/core/dbt/graph/selector_spec.py
@@ -1,7 +1,9 @@
 import os
 import re
+import enum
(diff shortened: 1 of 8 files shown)


core/dbt/graph/selector_spec.py
```diff
@@ -116,14 +123,9 @@ def selection_criteria_from_dict(
...
-        if indirect_selection:
-            if indirect_selection in ['eager', 'cautious']:
-                eagerly_expand = indirect_selection != 'cautious'
-            else:
-                raise RuntimeException(
-                    f'indirect_selection value "{indirect_selection}" is not valid!'
-                )
+        indirect_selection = dct.get('indirect_selection', None) or indirect_selection
+        if indirect_selection and indirect_selection not in ['eager', 'cautious']:
+            raise RuntimeException(f'indirect_selection value "{indirect_selection}" is invalid!')
```
jtcohen6, line 128: Rather than redefining the acceptable values here, we can use the same enum class for validation:
```python
        indirect_selection = IndirectSelection(
            dct.get('indirect_selection', None) or indirect_selection
        )
```
Then, the error raised is:
```
17:17:49 | [ error ] | Encountered an error:
'bad_value' is not a valid IndirectSelection
```
Instead of:
```
17:14:56 | [ error ] | Encountered an error:
Runtime Error
  Could not read selector file data: Runtime Error
    indirect_selection value "bad_value" is invalid!
```
That feels clear enough to me. If you wanted to keep the extra bit of context, we could try catching + wrapping the exception.

jtcohen6 (commented):
@VersusFacit Thanks for the extra work here! I agree, I think the late bool-ification makes the codepaths much easier to follow, and will make it that much easier when someone comes along to add a third indirect selection mode.

I left one comment. As far as the changelog, feel free to add an `rc2` section, and stick this `Under the hood`, [like so](https://github.com/dbt-labs/dbt-core/pull/4271/files#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4edR1-R4). (Assuming that PR gets merged first, you can just pull from `main`.)

core/dbt/graph/selector_spec.py
VersusFacit: Oh wow. Gonna be honest, I just didn't know the semantics around enums with confidence. Did some local testing and I find it a mental stretch that a StrEnum constructor _can take_ a StrEnum parameter. Not what I was expecting. Suffice to say, I'm onboard with this.

VersusFacit:
Addressed your comment and added a CHANGELOG line (also, just the heck of it, cherry-picked in Gerda's hunk in hopes of avoiding merge conflicts 🤞 )

jtcohen6 (approved):
Brilliant!

PR #4283, opened 2021-11-16
playing with agate column types

### LINKED ISSUE #4282: override seed column types when inferred to be of type `agate_helper.ISODateTime`

### Current Behavior

this issue came up with dbt-firebolt because the JDBC driver doesn't like non-string bindings for prepared statements. My goal is to cast them all to strings. 

the issue came up when using [dbt-adapter-tests's base.csv](https://github.com/dbt-labs/dbt-adapter-tests/blob/29356d9a07529e1a835ffdd422d94ad44a005b6f/pytest_dbt_adapter/builtin.py#L17-L29). IIRC, this issue did not happen in dbt-core 0.20.0

[line 49](https://github.com/dbt-labs/dbt-core/blob/d833c8e45a5486dcd07f3417926a88c6edb6fd43/core/dbt/clients/agate_helper.py#L49) of `agate_helper.build_type_tester()` (snippet below) results in an `agate_table` where `some_date` to be of type : `ISODateTime`. commenting out line 49 fixes this problem, as it allows the values in agate table to be overridden.

https://github.com/dbt-labs/dbt-core/blob/d833c8e45a5486dcd07f3417926a88c6edb6fd43/core/dbt/clients/agate_helper.py#L43-L54

it is insufficient to  modify `adapter.convert_datetime_type` because that will only change the column's type, not the actual values. though, in my head, I thought that this would work.

```python
@classmethod
def convert_datetime_type(cls, agate_table: agate.Table, col_idx: int) -> str:
    # there's an issue with timestamp currently
    return 'TEXT'
```

#### logging set up
here's the log statements I added to `seed.sql`

```sql
{% macro firebolt__reset_csv_table(model,
                                   full_refresh,
                                   old_relation,
                                   agate_table) %}
    {{ log('agate_table col types: ' ~ agate_table.column_types, info=True) }}
    {{ log('agate_table schema: ' ~ agate_table.print_structure(), info=True) }}
    {{ log('agate_table vals: ' ~ agate_table.print_table(), info=True) }}
....
{% macro firebolt__load_csv_rows(model, agate_table) %}

  {% set batch_size = get_batch_size() %}

  {% set cols_sql = get_seed_column_quoted_csv(model, agate_table.column_names) %}
  {% set bindings = [] %}

  {% set statements = [] %}

  {% for chunk in agate_table.rows | batch(batch_size) %}
      {% set bindings = [] %}

      {% for row in chunk %}
            {{ log('row: ' ~ row, info=True) }}
          {% do bindings.extend(row) %}
      {% endfor %}
```

#### logging output

```
col_name: some_date
inferred_type: TEXT
type: TEXT
row: <agate.Row: (Decimal('1'), 'Easton', datetime.datetime(1981, 5, 20, 6, 46, 51))>
```

however, if I comment out line 49 from above, then not only is the column type changed to be of type `TEXT` but also the parameters in the `INSERT INTO ... VALUES` statement also become strings.

### Expected Behavior

I expect that `adapter.convert_datetime_type()` would allow any seed column that's datetime-like to be overrided to the returned type. However, this does not work if the column type is inferred to be `agate_helper.ISODateTime`.

### Environment

```markdown
- OS: macOS `11.6.1`
- Python: `3.9.5`
- dbt: `0.21.1rc1`
```

`firebolt==0.20.2`

dataders:
resolves #4282

### Description

trying to get things working

### Checklist

### DIFF
--- a/core/dbt/clients/agate_helper.py
+++ b/core/dbt/clients/agate_helper.py
@@ -46,7 +46,6 @@ def build_type_tester(
                               date_format='%Y-%m-%d'),
         agate.data_types.DateTime(null_values=('null', ''),
                                   datetime_format='%Y-%m-%d %H:%M:%S'),
-        ISODateTime(null_values=('null', '')),
         agate.data_types.Boolean(true_values=('true',),
                                  false_values=('false',),
                                  null_values=('null', '')),
--- a/core/dbt/adapters/base/impl.py
+++ b/core/dbt/adapters/base/impl.py
@@ -23,7 +23,9 @@ from dbt.adapters.protocol import (
     AdapterConfig,
     ConnectionManagerProtocol,
 )
-from dbt.clients.agate_helper import empty_table, merge_tables, table_from_rows
+from dbt.clients.agate_helper import (
+    empty_table, merge_tables, table_from_rows, ISODateTime
+)
 from dbt.clients.jinja import MacroGenerator
 from dbt.contracts.graph.compiled import (
     CompileResultNode, CompiledSeedNode
@@ -925,6 +927,7 @@ class BaseAdapter(metaclass=AdapterMeta):
             (agate.Number, cls.convert_number_type),
             (agate.Boolean, cls.convert_boolean_type),
             (agate.DateTime, cls.convert_datetime_type),
+            (ISODateTime, cls.convert_datetime_type),
             (agate.Date, cls.convert_date_type),
             (agate.TimeDelta, cls.convert_time_type),
         ]


core/dbt/clients/agate_helper.py
```diff
@@ -46,7 +46,6 @@ def build_type_tester(
                               date_format='%Y-%m-%d'),
         agate.data_types.DateTime(null_values=('null', ''),
                                   datetime_format='%Y-%m-%d %H:%M:%S'),
-        ISODateTime(null_values=('null', '')),
```
jtcohen6, line 49: This scares me—feels likely to be a subtle breaking change

jtcohen6 (commented):
Dug into this code as I sought to understand the related issue, ended up leaving a more detailed comment over there: https://github.com/dbt-labs/dbt-core/issues/4282

PR #4285, opened 2021-11-16
Avoid errors when missing column in yaml doc

### LINKED ISSUE #4151: [Bug] Need more helpful error message when persist_docs encounters invalid column names

### Current Behavior

For a given model, if the YML documentation contains a column name not found in the model and `persist_docs` is enabled, the build fails with the following error:

```
Database Error in model my_model (path/to/my_model.sql)
  column "<col_name>" of relation "my_model" does not exist
  compiled SQL at target/run/path/to/my_model.sql
```

This message misleads the user to look in the SQL file to resolve the issue, even though that's not where the error lies. (Assuming we take the _model_ to be canonical and the _documentation_ to be derivative: presumably, most people wait to document until after building, although, like me, they might \*cough\* fail to remember to update their docs after later changing the model.)

### Expected Behavior

It would be more helpful to direct the user to the YML file, where the problem actually exists, and to explicitly say that the error occurred during the persist docs step. Along the lines of

```
Error persisting docs for model my_model
  column "<col_name>" defined in path/to/my_model_docs.yml 
  does not exist in path/to/my_model.sql
```

### Steps To Reproduce

Put the following model and its docs file in a project, and run the model.
```sql
{{ config(
    materialized = 'table',
    persist_docs={"relation": true, "columns": true}
) -}}

with sample_data as (

{% for item in ['row1', 'row2', 'row3'] %}
    {% if not loop.first %}union all {% endif %}select
        '{{ item }}' as field1,
        '{{ item }}' as field2
{% endfor %}
)

select
    field1 as alias1,
    field2 as alias2
from sample_data
```
```yml
version: 2

models:
  - name: bad_file
    columns:
      - name: alias1
        description: This field will persist
      - name: field2
        description: This field will fail
```

You'll get
```
Completed with 1 error and 0 warnings:

Database Error in model bad_file (models/marts/core/bad_file.sql)
  column "field2" of relation "bad_file" does not exist
  compiled SQL at target/run/ae_dw/models/marts/core/bad_file.sql
```

### Relevant log output

_No response_

### Environment

```markdown
- OS: Mac 11.6
- Python: 3.9.2
- dbt: 0.21.0
```

redshift

### Additional Context

See [this Stackoverflow question](https://stackoverflow.com/questions/69710700/why-am-i-getting-a-column-x-of-relation-y-does-not-exist-error-when-runnin), which led to a ticket to dbt support (18116), in which @jeremyyeo helped me resolve the issue.

kadero:
resolves #4151 

### Description
As described in the issue, we just need to update the postgres/redshift adapter.

I reproduced the issue described in the issue:
image

With the fix, the `dbt run` pass:
image

And the wrong fiied is not persisted:

image

### Checklist

### DIFF
--- a/test/integration/060_persist_docs_tests/test_persist_docs.py
+++ b/test/integration/060_persist_docs_tests/test_persist_docs.py
@@ -95,3 +95,32 @@ class TestPersistDocs(BasePersistDocsTest):
     def test_postgres_comments(self):
         self.run_has_comments_pglike()
 
+class TestPersistDocsColumnMissing(BasePersistDocsTest):
+    @property
+    def project_config(self):
+        return {
+            'config-version': 2,
+            'models': {
+                'test': {
+                    '+persist_docs': {
+                        "columns": True,
+                    },
+                }
+            }
+        }
+
+    @property
+    def models(self):
+        return 'models-column-missing'
+
+    @use_profile('postgres')
+    def test_postgres_missing_column(self):
+        self.run_dbt()
+        self.run_dbt(['docs', 'generate'])
+        with open('target/catalog.json') as fp:
+            catalog_data = json.load(fp)
+        assert 'nodes' in catalog_data
+
+        table_node = catalog_data['nodes']['model.test.missing_column']
+        table_id_comment = table_node['columns']['id']['comment']
+        assert table_id_comment.startswith('test id column description')
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,14 @@
 ## dbt-core 1.0.0rc2 (TBD)
 
+### Features
+- Avoid error when missing column in YAML description ([#4151](https://github.com/dbt-labs/dbt-core/issues/4151), [#4285](https://github.com/dbt-labs/dbt-core/pull/4285))
+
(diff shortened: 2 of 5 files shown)


jtcohen6 (commented):
@kadero Looks great! Any chance you could add an integration test case for this behavior? It could look exactly like the one in the Snowflake adapter [here](https://github.com/dbt-labs/dbt-snowflake/blob/e3769b36551a5040a74b80994277e1d8925c3848/tests/integration/persist_docs_tests/test_persist_docs.py#L142-L170)

test/integration/060_persist_docs_tests/test_persist_docs.py
```diff
@@ -95,3 +95,32 @@ def run_has_comments_pglike(self):
     def test_postgres_comments(self):
         self.run_has_comments_pglike()
 
+class TestPersistDocsColumnMissing(BasePersistDocsTest):
```
jtcohen6, line 98: 👍 Confirmed this test fails without the fix, and passes with it

jtcohen6 (approved):
@kadero Thank you for the contribution, as always!

PR #4288, opened 2021-11-16
Adding release workflow

leahwicz:
### Description

This is a GitHub Action that tests, builds, and packages `dbt-core` and `dbt-postgres` and then releases them to GitHub and PyPi.

Example workflow: https://github.com/leahwicz/dbt/actions/runs/1468971423
Example release: https://github.com/leahwicz/dbt/releases

I would like to try this out for RC2 🎉 

### Checklist

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,198 @@
+# **what?**
+# Take the given commit, run unit tests specifically on that sha, build and 
+# package it, and then release to GitHub and PyPi with that specific build
+
+# **why?**
+# Ensure an automated and tested release process
+
+# **when?**
+# This will only run manually with a given sha and version
+
+name: Release to GitHub and PyPi
+
+on:
+  workflow_dispatch:
+    inputs:
+      sha:
+       description: 'The last commit sha in the release'
+       required: true
+      version_number:
+       description: 'The release version number (i.e. 1.0.0b1)'
+       required: true
+
+defaults:
+  run:
+    shell: bash
+
+jobs:
+  unit:
+    name: Unit test
+
+    runs-on: ubuntu-latest
+
+    env:
+      TOXENV: "unit"
+
+    steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,198 @@
...
+    shell: bash
+
+jobs:
+  unit:
+    name: Unit test
+
+    runs-on: ubuntu-latest
+
+    env:
+      TOXENV: "unit"
```
leahwicz, line 34: Is unit testing enough verification for moving forward with the release or should we run some integration test? We should only be releasing when the last commit is all 🟢 in Actions but that relies on us doing our due diligence and checking

.github/workflows/release.yml
```diff
@@ -0,0 +1,198 @@
...
+          echo ::set-output name=isPrerelease::$IS_PRERELEASE
+
+      - name: Creating GitHub Release
+        uses: softprops/action-gh-release@v1
+        with:
+          name: dbt-core v${{github.event.inputs.version_number}}
+          tag_name: v${{github.event.inputs.version_number}}
+          prerelease: ${{ steps.release_type.outputs.isPrerelease }}
+          target_commitish: ${{github.event.inputs.sha}}
+          body: |
```
leahwicz, line 173: We will still have to paste in the release notes we want manually but for now that seemed like an ok start. We have been talking about changing how we do the Changelog anyway so maybe once we have a plan there we can figure how to automatically pull them in

.github/workflows/release.yml
jtcohen6: Unit tests should catch the very obvious issues. Of course, if there are failing integration tests on the `main`/release branch, we should know about them

.github/workflows/release.yml
jtcohen6: That's fine. We can also edit the GitHub release, and update release notes after the fact

PR #4291, opened 2021-11-17
Clean up project config rename deprecation warnings

### LINKED ISSUE #4276: Refactor logging for `PackageRedirectDeprecation`
I want to see two _different_ log messages: one each for `source-paths` &rarr; `model-paths` and `data-paths` &rarr; `seed-paths`. To achieve this, they need to use different deprecation names, since `active_deprecations` prevents the same deprecation from being shown more than once.

At least, it should! After our structured logging changes, I'm now seeing the same message twice:

```
$ dbt <command>
17:26:58 | [ info  ] | Running with dbt=1.0.0-rc1
17:26:58 | [ warn  ] | * Deprecation Warning: The `source-paths` config has been deprecated in favor of
`model-paths`. Please update your `dbt_project.yml` configuration to reflect
this change.
17:26:58 | [ warn  ] | * Deprecation Warning: The `source-paths` config has been deprecated in favor of
`model-paths`. Please update your `dbt_project.yml` configuration to reflect
this change.
```

So this issue is:
- [ ] Split up the `PackageRedirectDeprecation` into two separate deprecations
- [ ] Figure out why the same deprecation warning is being shown twice

This is pretty important, insofar as _everyone_ is going to see these messages as soon as they upgrade to v1, and we want the upgrade experience to feel as smooth as possible.

jtcohen6:
resolves #4276

- Create different deprecation warnings for each of `source-paths` and `data-paths` being renamed, so that both will display for users newly upgraded to v1
- Add a null handler to the global `FILE_LOG` when it's first instantiated, to prevent the un-configured `FILE_LOG` and `STDOUT_LOG` from _both_ writing to stdout. This shouldn't have any functional impact, since the file logger isn't meant to be doing anything until it's configured, anyway

Also:
- Add `\n\n` to deprecation warning. `ui.line_wrap_message()` doesn't work perfectly with the new logging format just yet (it doesn't account for space taken up by timestamp + log level), so I'd prefer a cleaner break.
- In `ui.line_wrap_message()`, prevent `textwrap.fill()` from breaking on hyphens. This is awkward for `PackageRedirectDeprecation`, which includes hyphen-bearing package names.

Before:
```
$ dbt test
07:16:14 | [ info  ] | Running with dbt=1.0.0-rc1
07:16:14 | [ warn  ] | * Deprecation Warning: The `source-paths` config has been deprecated in favor of
`model-paths`. Please update your `dbt_project.yml` configuration to reflect
this change.
07:16:14 | [ warn  ] | * Deprecation Warning: The `source-paths` config has been deprecated in favor of
`model-paths`. Please update your `dbt_project.yml` configuration to reflect
this change.
```

After:
```
$ dbt test
06:50:56 | [ info  ] | Running with dbt=1.0.0-rc1
06:50:56 | [ warn  ] | * Deprecation Warning:
The `source-paths` config has been deprecated in favor of `model-paths`. Please
update your `dbt_project.yml` configuration to reflect this change.
06:50:56 | [ warn  ] | * Deprecation Warning:
The `data-paths` config has been deprecated in favor of `seed-paths`. Please
update your `dbt_project.yml` configuration to reflect this change.
```

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,8 +4,9 @@
 - Avoid error when missing column in YAML description ([#4151](https://github.com/dbt-labs/dbt-core/issues/4151), [#4285](https://github.com/dbt-labs/dbt-core/pull/4285))
 
 ### Under the hood
-Add --indirect-selection parameter to profiles.yml and builtin DBT_ env vars; stringified parameter to enable multi-modal use ([#3997](https://github.com/dbt-labs/dbt-core/issues/3997), [PR #4270](https://github.com/dbt-labs/dbt-core/pull/4270))
+- Add --indirect-selection parameter to profiles.yml and builtin DBT_ env vars; stringified parameter to enable multi-modal use ([#3997](https://github.com/dbt-labs/dbt-core/issues/3997), [#4270](https://github.com/dbt-labs/dbt-core/pull/4270))
 - Fix filesystem searcher test failure on Python 3.9 ([#3689](https://github.com/dbt-labs/dbt-core/issues/3689), [#4271](https://github.com/dbt-labs/dbt-core/pull/4271))
+- Clean up deprecation warnings shown for `dbt_project.yml` config renames ([#4276](https://github.com/dbt-labs/dbt-core/issues/4276), [#4291](https://github.com/dbt-labs/dbt-core/pull/4291))
 
 Contributors:
 - [@kadero](https://github.com/kadero) ([#4285](https://github.com/dbt-labs/dbt-core/pull/4285))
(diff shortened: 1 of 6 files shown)


nathaniel-may (approved):
I noticed this when testing on one of my old projects and I found it annoying. Nice fix!

I do think a lot of this logic should actually live in the new `events` module, but I'm ok not doing that refactor right now. The bottom of the deprecation warnings looks like it drops out into events [here](https://github.com/dbt-labs/dbt-core/blob/9d73304c1a39a4c4a4caca8b44645169b4ea9c4e/core/dbt/exceptions.py#L996), but I think most of the rest of this logic could live in there too. I've added it to the stretch goals list for phase 2. We can make a ticket for it if we don't get to it.

jtcohen6:
Agreed - we treat deprecations _sorta_ like exceptions, and we should eventually move all the deprecation / exception / `warn_error` logic into `events`. That can be TK

Issue #4260, opened 2021-11-09
Structured Logging Phase 2


nathaniel-may:
### Describe the Feature

Individual tasks are ordered by expected release with high priority unfinished items at the top of the list.

### 1.0.0-RC2 (released 11/22)
- [x] (size: M) (@nathaniel-may) cleanup code #4266
- [x] (size: XL) (@iknox-fa) add codes to each event ~~#4268~~ #4319
- [x] (size: S) (@nathaniel-may) add versioning ~#4272~ #4289 
- [x] (size: M) (@nathaniel-may) fix adapter interface #4266
- [x] (size: S) (@emmyoop) include event dataclass attributes in json-formatted logs `{code: ..., msg: ...  data:{<goes here>}}` #4301  
- [x] (size: S) (@jtcohen6) fix log duplication in subsequent tests #4275 

### 1.0.0-RC3 (released 11/30)
- [x] (size: XXL) (@emmyoop) add additional log lines for node status reporting ([details](https://github.com/dbt-labs/dbt-core/issues/3359#issuecomment-934428619))  #4336
- [x] (size: M) (@iknox-fa) Does something graceful when it runs out of memory (small machines shouldn't explode because of logging), is queryable by a potential future server-wrapper #4358
- [x] (size: S) (@nathaniel-may) add thread name to each event (cheap way to distinguish dbt runs from each other, tracks dbt threads when we go multithreaded) (make sure windows is handled) #4353
- [x] (size: XL) (@nathaniel-may) stabilize JSON serialization.
- [x] (size: M) (@gshank) make sure we can distinguish logs from different runs Impl - Existing InvocationID. If tracking is off, it isn't created today so we would need to create it. (to add it to logging, add it as a [class property on the superclass](https://github.com/dbt-labs/dbt-core/blob/e1a2e8d9f542b15d283b3085ee867e882d85f5b7/core/dbt/events/base_types.py#L55-L59)) (previously considered: preserve process id like it is today, global run_id reported at start, or allow some sort of user markers)  #4340

### 1.0.0 (goal: merge by Fri 12/03)
**Cleanup:**
- [x] (size: S) (@jtcohen6) Find a way to configure logger _before_ firing any events, e.g. avoid text logs preceding JSON logs https://github.com/dbt-labs/dbt-core/pull/4341
- [x] (size: S) (@emmyoop) turn on verbose cache logging (every event fired by `cache.py`) if and only if the `--log-cache-events` flag is turned on #4369
[...]

PR #4293, opened 2021-11-17
Misc logging cleanup + commentary

jtcohen6:
I played around with the logging module this morning from an end-user perspective. It was a lot of fun! There are some fixes we should definitely merge; some questions are open for discussion. I tried to keep the commits separate + clean, so we could split up and cherry-pick accordingly. This is all in the interest of making dbt look + feel "like dbt" by v1 final.

### Critical fixes
- #4266 turned on JSON-formatted logging everywhere all the time, _except_ for the legacy logger. Re-enable the choice between `create_json_log_line` and `create_text_log_line` based on `this.format_json`

### Touch-ups + cosmetic fixes
- For events whose messages were prefixed with `\n`, instead precede them with `fire_event(EmptyLine())`. This is much cleaner to look at, since all (non-wrapped) log lines get the same frontmatter (timestamp + log level)
- I happened to notice that `ConcurrencyLine` could be more structured (unless there's a good reason why it wasn't)
- If we're going to include log levels on the CLI (more on this below), disable duplicative `WARNING` prefixes in messages / formatted via `warning_tag`

### Commentary
- **Log levels:** These are amazing to have in the File. I find they add a bit too much noise on the CLI. I'd be happy with either (a) turning them off for CLI logging, or (b) colorizing `warn`/yellow + `error`/red to raise the signal from the noise. (Also, I believe #4266 removed width-fixing for log levels, so I added that back in.)
- **Colorization:** Ideally, we'd have a way to turn this off for both File + JSON-formatted logging. This would require a little bit of plumbing: `dbt.ui.color` would need information from the `events` module, beyond just `flags.USE_COLORS`. Since color methods are mostly called in messages defined on event classes, I wonder if this could work by adding a `use_color(self) -> bool` method to the `File` + `Cli` base types, and reimplementing the color methods to check `self.use_color` (where self = event class).
- **EmptyLine:** Just like colorization, this is great for CLI + text formatting, but we'd ideally have a way to turn it off for File or JSON logging. (This is what `logbook` accomplished via the `TextOnly()` wrapper; I realize I was a bit inconsistent in whether I used this.)

Broadly, I find that I want:
- color, concision, clean spacing on the CLI
- colorless + compact verbosity in the File + JSON-formatted

### Checklist

### DIFF
--- a/core/dbt/events/functions.py
+++ b/core/dbt/events/functions.py
@@ -3,6 +3,7 @@ from colorama import Style
 import dbt.events.functions as this  # don't worry I hate it too.
 from dbt.events.base_types import Cli, Event, File, ShowException
 import dbt.flags as flags
+import dbt.ui as ui
 # TODO this will need to move eventually
 from dbt.logger import SECRET_ENV_PREFIX, make_log_dir_if_missing, GLOBAL_LOGGER
 import io
@@ -110,6 +111,19 @@ def scrub_secrets(msg: str, secrets: List[str]) -> str:
 T_Event = TypeVar('T_Event', bound=Event)
 
 
+def format_level(level):
+    fixed_width = level.ljust(5)
+    # TODO turn off all color everwhere for file + JSON
+    if not this.format_color:
+        return fixed_width
+    elif level == 'warn':
+        return ui.yellow(fixed_width)
+    elif level == 'error':
+        return ui.red(fixed_width)
+    else:
+        return fixed_width
+
+
 # translates an Event to a completely formatted text-based log line
 # you have to specify which message you want. (i.e. - e.message, e.cli_msg(), e.file_msg())
 # type hinting everything as strings so we don't get any unintentional string conversions via str()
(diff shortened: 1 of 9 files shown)


jtcohen6:
@nathaniel-may @emmyoop @iknox-fa Tagging you here for visibility. The critical fix noted above is the only thing we _must_ have for RC2. I know the rest of this is lower priority than the other work we've itemized in #4260.

nathaniel-may:
> If we're going to include log levels on the CLI (more on this below), disable duplicative WARNING prefixes in messages / formatted via warning_tag

This is a good reason why I think the warning stuff should live in the events mod eventually.

emmyoop (approved):
LGTM

core/dbt/events/functions.py
```diff
@@ -223,7 +237,11 @@ def fire_event(e: Event) -> None:
 
     # always logs debug level regardless of user input
     if isinstance(e, File):
-        log_line = create_json_log_line(e, msg_fn=lambda x: x.file_msg())
+        log_line = (
+            create_json_log_line(e, msg_fn=lambda x: x.file_msg())
+            if this.format_json else
+            create_text_log_line(e, msg_fn=lambda x: x.file_msg())
+        )
```
nathaniel-may, line 244: let's encapsulate this into a function since it's used 3 times.
```suggestion
        log_line = create_log_line(json=this.format_json, e=e, msg_fn=lambda x: x.file_msg())
```
where
```
def create_log_line(json: bool, e: T_Event, msg_fn: Callable[[T_Event], str]) -> str:
    return (
        create_json_log_line(e, msg_fn)
        if this.format_json else
        create_text_log_line(e, msg_fn)
    )
```

jtcohen6:
Closing in favor of https://github.com/dbt-labs/dbt-core/pull/4388

PR #4301, opened 2021-11-17
structured logging: add data attributes to json log output

emmyoop:
### Description

Add the dataclass attributes to a data key in the json log output.

### Checklist

### DIFF
--- a/core/dbt/events/functions.py
+++ b/core/dbt/events/functions.py
@@ -14,7 +14,7 @@ import logging
 from logging import Logger
 from logging.handlers import RotatingFileHandler
 import os
-from typing import Callable, List, Union
+from typing import Any, Callable, Dict, List, Optional, Union
 from dataclasses import _FIELD_BASE  # type: ignore[attr-defined]
 
 
@@ -121,7 +121,9 @@ def event_to_dict(e: T_Event, msg_fn: Callable[[T_Event], str]) -> dict:
         'ts': e.get_ts(),
         'pid': e.get_pid(),
         'msg': msg_fn(e),
-        'level': level
+        'level': level,
+        'data': Optional[Dict[str, Any]],
+        'event_data_serialized': True
     }
 
 
@@ -161,9 +163,10 @@ def create_json_log_line(e: T_Event, msg_fn: Callable[[T_Event], str]) -> str:
     except TypeError:
         # the only key currently throwing errors is 'data'.  Expand this list
         # as needed if new issues pop up
-        safe_values = {k: v for (k, v) in values.items() if k not in ('data')}
+        values['data'] = None
+        values['event_data_serialized'] = False
         log_line = json.dumps(
-            {k: scrub_secrets(v, env_secrets()) for (k, v) in safe_values.items()},
+            {k: scrub_secrets(v, env_secrets()) for (k, v) in values.items()},
             sort_keys=True
         )
     return log_line
--- a/core/dbt/events/functions.py
+++ b/core/dbt/events/functions.py
@@ -141,6 +141,7 @@ def create_text_log_line(e: T_Event, msg_fn: Callable[[T_Event], str]) -> str:
 def create_json_log_line(e: T_Event, msg_fn: Callable[[T_Event], str]) -> str:
     values = event_to_dict(e, lambda x: scrub_secrets(msg_fn(x), env_secrets()))
     values['ts'] = e.get_ts().isoformat()
+    values['data'] = {k: scrub_secrets(str(v), env_secrets()) for (k, v) in e.__dict__.items()}
     log_line = json.dumps(values, sort_keys=True)
     return log_line
 
--- a/core/dbt/events/functions.py
+++ b/core/dbt/events/functions.py
@@ -137,15 +137,6 @@ def create_text_log_line(e: T_Event, msg_fn: Callable[[T_Event], str]) -> str:
     return log_line
 
 
-# classes and sets are returned that cannot be natively encoded to json
-def set_default(obj):
-    if hasattr(obj, '__dict__'):
-        return obj.__dict__
-    if isinstance(obj, set):
-        return list(obj)
-    return str(obj)
-
-
 # translates an Event to a completely formatted json log line
 # you have to specify which message you want. (i.e. - e.message, e.cli_msg(), e.file_msg())
 def create_json_log_line(e: T_Event, msg_fn: Callable[[T_Event], str]) -> str:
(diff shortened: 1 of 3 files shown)


emmyoop:
Example output:
```
{"data": {"v": "=1.0.0-rc1"}, "level": "info ", "msg": "Running with dbt=1.0.0-rc1", "pid": 23222, "ts": "2021-11-17T11:35:28.328580"}
{"data": {"stat_line": "5 models, 4 tests, 0 snapshots, 0 analyses, 166 macros, 0 operations, 1 seed file, 0 sources, 0 exposures, 0 metrics"}, "level": "info ", "msg": "Found 5 models, 4 tests, 0 snapshots, 0 analyses, 166 macros, 0 operations, 1 seed file, 0 sources, 0 exposures, 0 metrics", "pid": 23222, "ts": "2021-11-17T11:35:28.328580"}
{"data": {}, "level": "info ", "msg": "", "pid": 23222, "ts": "2021-11-17T11:35:28.328580"}
```

core/dbt/events/functions.py
```diff
@@ -129,6 +129,7 @@ def create_text_log_line(e: T_Event, msg_fn: Callable[[T_Event], str]) -> str:
 def create_json_log_line(e: T_Event, msg_fn: Callable[[T_Event], str]) -> str:
     values = e.to_dict(scrub_secrets(msg_fn(e), env_secrets()))
     values['ts'] = e.ts.isoformat()
+    values['data'] = {k: scrub_secrets(str(v), env_secrets()) for (k, v) in e.__dict__.items()}
```
nathaniel-may, line 132: This is where my python knowledge is a little weak. I know there are lots of ways to access class attributes and I'm not sure what the tradeoffs of `e.__dict__.items()` is compared to the others. Do you have thoughts on this choice?

emmyoop: I should use `vars()` here instead of `__dict__`.  My understanding is there's isn't much of a real difference between the two in terms of what they're doing.  Were you thinking of another way to solve this?

nathaniel-may: not in particular no. I am curious what @iknox-fa thinks though since he might know more of the alternatives.

iknox-fa: Oh there be dragons there, maybe.  

`vars()` and `_dict_()` both do the same thing, but I'm not sure it's what you want there.  They contain all writeable attributes of an object which means it will be scrubbing some stuff you might not want to scrub (log levels, tags, anything else we hang on an event, etc).

If the scrub method handles that gracefully there's no problems though.

emmyoop: `__dict__` only returns the writeable attributes on the subclass.  `inspect.getmembers(e)` shows `__dict__` is just  `{'conn_type': 'postgres', 'conn_name': 'master'}` for the `NewConnection` event.

```@dataclass
class NewConnection(DebugLevel, Cli, File):
    conn_type: str
    conn_name: str

    def message(self) -> str:
        return f'Acquiring new {self.conn_type} connection "{self.conn_name}"'
```

Am I right that since this is just an instance of `NewConnection` that I'm executing on we haven't attached the other writable attributes yet so we're only scrubbing what we care about?

iknox-fa: > __dict__ only returns the writeable attributes on the subclass

So just to be clear, this isn't generally true, however is here because of the rest of the function as @nathaniel-may points out.

> Am I right that since this is just an instance of NewConnection that I'm executing on we haven't attached the other writable attributes yet so we're only scrubbing what we care about?

Yes, but only by the virtue that we only added fields and all fields are ones we wish to scrub.  Here's an example where it could break stuff.

```
from dataclasses import dataclass, _FIELD_BASE
from pprint import pprint as pp

@dataclass
class NewConnection():
    conn_type: str
    conn_name: str

    def something_unrelated(self):
        self.thing_we_use_somewhere_else = 173

    def message(self) -> str:
        return f'Acquiring new {self.conn_type} connection "{self.conn_name}"'

conn = NewConnection(conn_type="postgres", conn_name="localhost")
conn.something_unrelated()


things_to_scrub = conn.__dict__
pp(things_to_scrub)
# > {'conn_name': 'localhost', 'conn_type': 'postgres','thing_we_use_somewhere_else': 173}
```

So some dev added something to the class and didn't know it was going to be scrubbed and now we try to scrub an int.  No bueno.
Luckily, python gives us some more dunder methods specific to dataclasses we can leverage like so:
```
# continued from above
safer_things_to_scrub = {x:getattr(conn, x) for x,y in conn.__dataclass_fields__.items() if type(y._field_type) == _FIELD_BASE}
pp(safer_things_to_scrub)
# > {'conn_name': 'localhost', 'conn_type': 'postgres'}```

Comment (nathaniel-may) [reply core/dbt/events/functions.py]: So right now we're only scrubbing the message field, not any of the other fields which is going to be a problem once we add these fields. I think we should move scrubbing to the final stringified log line instead. Do you think you could add that to this PR, @emmyoop?

Comment (nathaniel-may) [reply core/dbt/events/functions.py]: @iknox-fa that looks like exactly what we want. Although right now not every class is a dataclass so we might get an `AttributeError` on `x.__dataclass_fields__`. I'm fine making everything a dataclass that might not be currently, as long as it's not abstract because of that mypy bug.

Comment (iknox-fa) [reply core/dbt/events/functions.py]: @nathaniel-may Oh good point.  You can't use dataclass features without... dataclasses.  This may also run afoul of our 3.6 dataclass library.  I'm def getting into the nitty-gritty of how DCs work and they may not have implemented the full 3.7+ api on those.

Comment (emmyoop) [reply core/dbt/events/functions.py]: What about just a simple 
```
if hasattr(e, '__dataclass_fields'):
        values['data2'] = {x:str(getattr(e, x)) for x,y in e.__dataclass_fields__.items() if type(y._field_type) == _FIELD_BASE}
```
We officially deprecated 3.6 (right?) for 1.0.0 so that should be fine.

Comment (nathaniel-may) [reply core/dbt/events/functions.py]: yeah that pattern looks totally reasonable.

File: core/dbt/events/functions.py [hunk 2]
```diff
@@ -124,12 +125,33 @@ def create_text_log_line(e: T_Event, msg_fn: Callable[[T_Event], str]) -> str:
...
+        values['data'] = {
+            x: getattr(e, x) for x, y
+            in e.__dataclass_fields__.items()  # type: ignore[attr-defined]
+            if type(y._field_type) == _FIELD_BASE
+        }
+    else:
+        values['data'] = None
+    log_line = json.dumps(
+        {k: scrub_secrets(v, env_secrets()) for (k, v) in values.items()},
+        default=lambda x: set_default(x),
```
emmyoop, line 152: @iknox-fa thoughts on this?  Since our attributes are a bit all over the place in terms of type, many are not serializable.  I don't like what I did but I'm not sure what the better alternative is.  Any advice would be appreciated!

core/dbt/events/functions.py
iknox-fa: Yep that works too!

core/dbt/events/functions.py
iknox-fa: Good question!  I would probably go for a try/except on `json.JSONDecodeError`.  You could also catch `TypeError` exceptions if we thought someone would use non-json types in the keys of the dict (this seems _highly_ unlikely)

core/dbt/events/functions.py
```diff
@@ -127,9 +128,32 @@ def create_text_log_line(e: T_Event, msg_fn: Callable[[T_Event], str]) -> str:
...
+    # need to catch if any data is not serializable but still make sure as much of
+    # the logs go out as possible
+    try:
+        log_line = json.dumps(
+            {k: scrub_secrets(v, env_secrets()) for (k, v) in values.items()},
+            sort_keys=True
+        )
+    except TypeError:
+        # the only key currently throwing errors is 'data'.  Expand this list
+        # as needed if new issues pop up
```
iknox-fa, line 151: We should probably have some indicator that part of the event record is missing from the json log.  Can we add a 
`values["event_data_failed_serailize"] = True` or something similar?

core/dbt/events/functions.py
```diff
@@ -127,9 +128,32 @@ def create_text_log_line(e: T_Event, msg_fn: Callable[[T_Event], str]) -> str:
...
+        safe_values = {k: v for (k, v) in values.items() if k not in ('data')}
```
iknox-fa, line 152: Since you're not using `values` after this point you can just pop the `data` key.  
```
except...
    values.pop("data")
    log_line = blahblah(in values.items())
 ```

iknox-fa (approved):
LGTM

PR #4305, opened 2021-11-18
Fix adapter failures due to string formatting issues

nathaniel-may:
### Description

This is an alternative solution to dbt-redshift#43. Prior to this change, log messages passed to the AdapterLogger that had brackets in them would trigger a KeyError when we tried to format with an empty set of arguments. This solution is preferred to dbt-redshift#43 because it fixes an existing interface problem.

### Checklist

### DIFF
--- a/test/unit/test_events.py
+++ b/test/unit/test_events.py
@@ -1,4 +1,5 @@
 from dbt.events import AdapterLogger
+from dbt.events.types import AdapterEventDebug
 from unittest import TestCase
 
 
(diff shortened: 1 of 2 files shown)


jtcohen6 (approved):
Nice catch!

Confirmed this fixed `test_redshift_column_quotes` (one of the nightly failures) when I ran it locally with this branch

test/unit/test_events.py
```diff
@@ -9,11 +10,38 @@ def setUp(self):
...
+        self.assertTrue("1 2 3" in event.message())
+
+        # tests that it doesn't throw
+        logger.debug("boop{x}boop")
+
+        # enters lower in the call stack to test that it formats correctly
+        # in this case it's that we didn't attempt to replace anything since there
+        # were no args passed after the initial message
+        event = AdapterEventDebug(name="dbt_tests", base_msg="boop{x}boop", args=())
+        self.assertTrue("boop{x}boop" in event.message())
```
emmyoop, line 47: 👍

emmyoop (approved):

PR #4311, opened 2021-11-19
SecretContext for secret env vars, profiles + packages only

### LINKED ISSUE #4310: Stricter secret env vars: allow in `profiles.yml` + `packages.yml`, disallow elsewhere
In v0.21, we implemented "secret" env vars (prefixed with `DBT_ENV_SECRET_`) that enabled users to ensure that their configured credentials or access tokens would be scrubbed from logs.

While this is a convenience for many users, it's still possible to use those env vars in all the ways + places that `{{ env_var() }}` can be used in dbt projects—in arbitrary macros, query results, models that write to the database, etc.

After further discussion and deliberation, we've identified two tightly constrained use cases for "secret" env vars:
- environment-based configuration of connection profiles (in `profiles.yml`)
- git access tokens for private packages (in `packages.yml`)

As such, we should restrict use of secret-prefixed env vars, enabling them for use in those two files _only_, and raising an explicit error if they're used elsewhere. This enables project administrators to unblock their colleagues from connecting to the database and accessing dbt packages, without exposing those secrets to misuse.

Questions we're considering now, and will likely answer in the future:
- Should this tight constraint be the default behavior? Should it be possible to re-enable more "permissive" use?
- Are there other use cases we aren't capturing above? (As soon as secrets are allowed in `dbt_project.yml` or macros, they're effectively allowed everywhere)

jtcohen6:
resolves #4310

### Description
- Adds `SecretContext` for use by the `ProfileRenderer` (`profiles.yml`) + `PackageRenderer` (`packages.yml`)
- In all other rendering contexts, the `env_var` context member will raise an exception (`disallow_secret_env_var`) if it encounters a var prefixed with `DBT_ENV_SECRET_`

Adjust integration tests accordingly:
- check for `ParsingException` if secret env var is used in a model
- ensure that secret env vars _can_ be used in `profiles.yml` + `packages.yml`
- ensure that the value of a secret env var is scrubbed from the logs (even if defined as a string literal, not via `env_var()`)

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,12 +1,13 @@
 ## dbt-core 1.0.0rc2 (TBD)
 
+### Breaking changes
+- Restrict secret env vars (prefixed `DBT_ENV_SECRET_`) to `profiles.yml` + `packages.yml` _only_. Raise an exception if a secret env var is used elsewhere ([#4310](https://github.com/dbt-labs/dbt-core/issues/4310), [#4311](https://github.com/dbt-labs/dbt-core/pull/4311))
+- Reorder arguments to `config.get()` so that `default` is second ([#4273](https://github.com/dbt-labs/dbt-core/issues/4273), [#4297](https://github.com/dbt-labs/dbt-core/pull/4297))
+
 ### Features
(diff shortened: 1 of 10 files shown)


gshank (approved):
Looks good. Nice work :)

jtcohen6:
@gshank Couldn't have done it without you. The code reorg + annotations re: contexts are what made this possible

PR #4322, opened 2021-11-22
[#4253] Support partial parsing of env_vars in metrics definitions

### LINKED ISSUE #4253: Support partial parsing with env_vars in metrics definition
Because the metrics feature and the support for partially parsing yaml entries with changed env vars happened in parallel, there is no code in core/dbt/parser/partial.py to re-parse on env var changes for metrics entries.

gshank:
resolves #4253

### Description

Because of parallel development of the metrics feature and partial parsing with environment variables, some code was missing to support re-parsing metrics when env_vars change.

### Checklist

### DIFF
--- a/test/integration/068_partial_parsing_tests/test-files/env_var_metrics.yml
+++ b/test/integration/068_partial_parsing_tests/test-files/env_var_metrics.yml
@@ -1,0 +1,30 @@
+version: 2
+
+metrics:
+
+  - model: "ref('people')"
+    name: number_of_people
+    description: Total count of people
+    label: "Number of people"
+    type: count
+    sql: "*"
+    timestamp: created_at
+    time_grains: [day, week, month]
+    dimensions:
+      - favorite_color
+      - loves_dbt
+    meta:
+        my_meta: '{{ env_var("ENV_VAR_METRICS") }}'
+
+  - model: "ref('people')"
+    name: collective_tenure
+    description: Total number of years of team experience
+    label: "Collective tenure"
+    type: sum
+    sql: tenure
+    timestamp: created_at
+    time_grains: [day]
+    filters:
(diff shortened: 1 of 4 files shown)


test/integration/068_partial_parsing_tests/test-files/env_var_metrics.yml
```diff
@@ -0,0 +1,30 @@
+version: 2
+
+metrics:
+
+  - model: "ref('people')"
+    name: number_of_people
+    description: Total count of people
+    label: "Number of people"
+    type: count
+    sql: "*"
```
jtcohen6, line 10: Should I anticipate this working for non-`meta` fields as well?

I set an env var for `sql` here like:
```yml
    sql: '{{ env_var("ENV_VAR_SQL", "*") }}'
```
Then:
```
$ dbt ls --resource-type metric --output json --output-keys name,sql
{"name": "number_of_people", "sql": "*"}
```
Looks good. But it doesn't seem to update when the env var changes:
```
$ ENV_VAR_SQL=person_id dbt ls --resource-type metric --
{"name": "number_of_people", "sql": "*"}
```

jtcohen6 (commented):
Code change makes sense - one question for rendering of other top-level metric properties (non-`meta` fields)

test/integration/068_partial_parsing_tests/test-files/env_var_metrics.yml
gshank: Strange. The metrics nodes are only rendered in one place, so I assumed just checking one would be enough. I'll check this out.

gshank: I think this should be working now.

PR #4326, opened 2021-11-23
make json serialization overridable for events

nathaniel-may:
### Description

This PR does three things:

1. Instead of inspecting the Python dictionary values for scrubability before converting to a string, convert the whole log line to a json string and scrub the string.

which requires

2. More graceful handling of events with non json serializable field types, by making the serialization function overridable.

which implies the need for

3. A test that attempts to serialize one instance of each concrete event type into JSON. This is not a perfect test becuase some of these events are annotated with the `Any` type so mypy nor this test will catch if some instances of that class that we make somewhere in the code base are not serializable. However, in that case instead of exploding or hanging, users will see something like `{ ... "some_key": "JSON_SERIALIZE_FAILED: MyClass", ... }`.

### Checklist

### DIFF
--- a/core/dbt/events/functions.py
+++ b/core/dbt/events/functions.py
@@ -13,7 +13,6 @@ import logbook
 import logging
 from logging import Logger
 from logging.handlers import RotatingFileHandler
-import numbers
 import os
 import uuid
 from typing import Any, Callable, Dict, List, Optional, Set, Tuple, Union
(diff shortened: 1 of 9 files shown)


emmyoop:
Can you add the `test all` label?

nathaniel-may:
> `TypeError: Object of type VersionSpecifier is not JSON serializable`
ahh, I see why you caught `TypeError`. I'm going to try and fix this with a different approach and see what you think.

core/dbt/events/functions.py
```diff
@@ -181,17 +153,8 @@ def create_json_log_line(e: T_Event, msg_fn: Callable[[T_Event], str]) -> str:
...
-        log_line = json.dumps(scrub_dict_secrets(values), sort_keys=True)
-    except TypeError:
-        # the only key currently throwing errors is 'data'.  Expand this list
-        # as needed if new issues pop up
-        values['data'] = None
-        values['event_data_serialized'] = False
-        log_line = json.dumps(scrub_dict_secrets(values), sort_keys=True)
-    return log_line
+    raw_log_line = json.dumps(values, sort_keys=True)
+    return scrub_secrets(raw_log_line, env_secrets())
```
emmyoop, line 157: I think you may still need the try statement to build the raw log line.  Then scrub on the return.  The `data` field has unpredictable contents.

nathaniel-may: The data field has very predictable contents! We wrote them all down in `types.py`. I think we just have to handle them better. I'll propose something here and see what you think.

nathaniel-may:
having trouble debugging the test failures here. I'll try again in the morning.

nathaniel-may:
@emmyoop helped me figure out that the hanging tests actually error out on a local `dbt run` which was helpful. Now that I've fixed the root causes of those exceptions, both the tests and `dbt run` hang. I'll have to come back to this later.

nathaniel-may:
hooray! Just the windows tests fail. Adding a test for serialization now.

leahwicz:
The changes LGTM but the Adapter tests seem to be failing so I'm concerned about that

gshank:
In the rest of our code we use Mashumaro for serialization ('to_dict' is usually mashumaro). It's too late to use it for this now, but we might want to look at the advantages/disadvantages for that later.

nathaniel-may:
@leahwicz the adapter tests are only the windows ones which is a known issue.

@gshank good call we should definitely do that.

emmyoop (approved):

leahwicz (approved):

PR #441, opened 2017-05-22
colorized dbt output

drewbanin:
TODO:
- [x] Test on windows
- [x] Add option to turn this off?

PR #4327, opened 2021-11-23
add new interop tests for black-box json log schema testing

nathaniel-may:
### Description

This PR is a proposal for a new type of test: interop tests. These are different than integration and unit tests because we have to write new CI to run them. Namely, for this first one, to compile Rust as a down stream consumer of our log files.

I chose to create the black-box logging test in Rust because Rust has support for statically derived json parsers. Although these derived parsers can parse all valid json values, some schemas are just ugly and I want our initial proposed design to fit into a clean, well-designed schema to make it as easy as possible for all downstream consumers to read our logs regardless of their language of choice.

ugly example:
```
{ "a": 1 }
{ "b": 1 }
{ "c": "string" }
```
Rust type modeling:
```
struct LogLine {
    a: Option<f64>
    b: Option<f64>
    c: Option<String>
}
```
This type is particularly ugly because it allows for completely empty log lines which we will never actually output. There's nothing technically wrong with this, but it doesn't exactly fill me with confidence either. This test forces us to confront our schema in a centralized way outside of the Python implementation the way many of our consumers will.

### Todo

- [x] Use `serde_json` to check serialization loop
- [x] Add new Action which makes a dbt run that outputs json files and passes the location to the rust tests
- [ ] add a runnable dbt project to get a better sample of logs
- [ ] model all possible sub fields as structs
- [ ] add tests for as many expected values as possible

### Checklist


emmyoop (approved):
Love this idea.

nathaniel-may:
Just leaving a note for myself that the latest commits require this to be rebased onto #4381 and #4384 once they get merged into main.

test/integration/030_statement_test/test_statements.py
```diff
@@ -34,15 +34,3 @@ def test_postgres_statements(self):
         self.assertEqual(len(results), 1)
 
         self.assertTablesEqual("statement_actual", "statement_expected")
-
-    @use_profile("presto")
```
jtcohen6, line 38: 👍

.github/workflows/structured-logging-schema-check.yml
```diff
@@ -0,0 +1,69 @@
...
...dependency_test test/integration/007_graph_selection_tests test/integration/008_schema_tests_test test/integration/009_data_tests_test test/integration/010_permission_tests test/integration/011_invalid_model_tests test/integration/012_deprecation_tests test/integration/013_context_var_tests test/integration/014_hook_tests test/integration/015_cli_invocation_tests test/integration/016_macro_tests test/integration/017_runtime_materialization_tests test/integration/018_adapter_ddl_tests test/integration/019_analysis_tests test/integration/020_ephemeral_test test/integration/021_concurrency_test test/integration/023_exit_codes_test test/integration/024_custom_schema_test test/integration/025_duplicate_model_test test/integration/025_timezones_test test/integration/026_aliases_test test/integration/027_cycle_test test/integration/028_cli_vars test/integration/029_docs_generate_tests test/integration/030_statement_test test/integration/031_thread_count_test test/integration/033_event_tracking_test test/integration/035_changing_relation_type_test test/integration/035_docs_blocks test/integration/037_external_reference_test test/integration/038_caching_test test/integration/039_config_test
```
jtcohen6, line 62: FWIW I managed to successfully run most of the excluded tests locally:
- I ran into hanging in `042_sources_test`, which I fixed with the change in https://github.com/dbt-labs/dbt-core/pull/441
- `test/integration/040_init_test/test_init.py` fails - I think it's fair to say `init` will work weirdly with JSON-formatted logs

I'd be interested in attempting to rerun with that fix (and excluding `040_init_test`)

jtcohen6 (commented):
I think highest priority is fixing the issues uncovered by this testing before the final release. Then, I'd love to get this merged in

.github/workflows/structured-logging-schema-check.yml
jtcohen6: There's one more hanging test, `test/integration/062_defer_state_test/test_run_results_state.py`, that I'm looking into right now

test/interop/log_parsing/src/main.rs
```diff
@@ -0,0 +1,209 @@
...
+mod tests {
+    use chrono::{DateTime, Utc};
+    use serde::{Deserialize, Serialize};
+    use std::env;
+    use std::fs::File;
+    use std::io::{self, BufRead};
+    use std::path::Path;
+    use walkdir::WalkDir;
+
+    //
```
leahwicz, line 15: Nit: assuming there is supposed to be a comment or something here

.github/workflows/structured-logging-schema-check.yml
```diff
@@ -0,0 +1,69 @@
...
+name: Structured Logging Schema Check
+on:
+  push:
+    branches:
+      - "main"
+      - "*.latest"
+      - "releases/*"
+  pull_request:
+  workflow_dispatch:
+
```
leahwicz, line 19: I would add `permissions: read-all` like [here](https://github.com/dbt-labs/dbt-core/blob/main/.github/workflows/main.yml#L27). This scope the GitHub token down so you can't write back to the repo

.github/workflows/structured-logging-schema-check.yml
```diff
@@ -0,0 +1,69 @@
...
+      # turns warnings into errors
+      RUSTFLAGS: "-D warnings"
+      # points tests to the log file
+      LOG_DIR: "/home/runner/work/dbt-core/dbt-core/logs"
+      # tells integration tests to output into json format
+      DBT_LOG_FORMAT: 'json'
+    steps:
+
+      - name: checkout dev
+        uses: actions/checkout@v2
```
leahwicz, line 35:
```suggestion
        uses: actions/checkout@v2
        with:
          persist-credentials: false
```

This might be overkill but then gets rid of the token which I don't think is needed anymore

leahwicz (approved):
LGTM but I would be interested in walking through the Rust with you as a learning exercise since not all of it makes sense to me

nathaniel-may:
rebased on the main, running all integration tests with json log format now.

PR #4332, opened 2021-11-24
Fix static parser tracking logic

jtcohen6:
### Background

I noticed a few weird things this morning:
- We have _way_ too many `cannot_parse` results in anonymous usage tracking for v1 — many more than I'd expect if we're only sampling one out of every 50k models
- Using latest v1, `dbt --no-partial-parse parse` ("cold parse") in our internal analytics project is taking 15-16 seconds, instead of 7-8 seconds. Indeed, that extra time is spent in _tracking_:

Screenshot 2021-11-24 at 12 23 28

### Fix

We're firing a tracking event _every single time_ the stable parser `cannot_parse` (or `has_banned_macro`). Instead, we should only populate the tracking result if we're sampling. When I make this change, a "cold parse" of our internal analytics project takes 7-8 seconds again. Hooray!

While here:
- Consolidated the logic, so this check + result population happens as part of the "fall back to jinja" conditional branch
- We should be checking `experimental_sample`, too, and firing a differently coded event depending (`0x`vs. `8x`)

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 
 ### Under the hood
 - Change some CompilationExceptions to ParsingExceptions ([#4254](http://github.com/dbt-labs/dbt-core/issues/4254), [#4328](https://github.com/dbt-core/pull/4328))
+- Reorder logic for static parser sampling to speed up model parsing ([#4332](https://github.com/dbt-labs/dbt-core/pull/4332))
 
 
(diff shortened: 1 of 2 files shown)


gshank (approved):
Looks good to me.

nathaniel-may (approved):
Oof. good catch! thanks for fixing this!

PR #4336, opened 2021-11-24
add node_info to relevant logs

### LINKED ISSUE #4324: Add node_status reporting to logging
### Big Picture
Provide a more complete picture of the node status in a point in time through logs.  Create a dict `node_info` that will contain the status at the time the log fired.

### Existing Events
- add a new dict with node_info
- ensure that all node-related events include unique_id.  Acts as both a unique identifier and a way to look up any node information in manifest.json
- Provide more structure by adding `node_info` to existing messages where we fire and event when starting/finishing node events.

### New Events
- add a few new granular events to fill in the gaps between Start + Finish — though we're currently limited in the level of granularity, since we can't (yet) fire custom event types within the materialization

### dict structure
```
  "node_info": 
        {"node_path": "models/slow.sql", 
        "node_name": "slow", 
        "resource_type": "model", 
        "node_materialized": "table", 
        "node_started_at": "2021-10-05T13:17:28.018613", 
        "unique_id": "model.my_new_project.slow", 
        "node_finished_at": "2021-10-05T13:17:29.134126", 
        "node_status": "passed", 
        "run_state": "running"}
```

[Original details](https://github.com/dbt-labs/dbt-core/issues/3359#issuecomment-934428619)

related to #4260 

Things left to do:

- [x] node_status - create new contract for "running" nodes and Union it with the result contract
    - "running" contract could be started (very first event for a node), compiling [here](https://github.com/dbt-labs/dbt-core/blob/4eb17b57fbcacb2c63a252b412a64e1c2e944adc/core/dbt/task/base.py#L282-L286), executing [here](https://github.com/dbt-labs/dbt-core/blob/4eb17b57fbcacb2c63a252b412a64e1c2e944adc/core/dbt/task/base.py#L291-L293) - and fire new events when the node reaches those points
- [x] start/end time need to be populated.  Needs to persist across events for a unique id.
- [ ] modify return of node_info after #4326 is merged in
- [ ] create Union of types for `report-node_data` instead of using `Any` - currently involves some circular imports
- [x] fire new events for node status changes
- [x] modify current events to include `node_info` as needed

emmyoop:
resolves #4324 

### Description

First pass - any feedback on approach welcome!

### Checklist


core/dbt/events/base_types.py
```diff
@@ -57,6 +56,9 @@ class Event(metaclass=ABCMeta):
     log_version: int = 1
     ts: Optional[datetime] = None  # use getter for non-optional
     pid: Optional[int] = None  # use getter for non-optional
+    report_node_data: Optional[Any] = None
+    node_status: Optional[str] = None
+    node_state: Optional[str] = None
```
nathaniel-may, line 61: could this be modeled with a single class? If it's all or nothing, splitting out the optionals like this is a looser type fit. Something like this:
```python
node_info: Optional[NodeInfo]
```
where `NodeInfo` has report_node_data, node_status, and node_state.

core/dbt/events/base_types.py
```diff
@@ -88,6 +90,21 @@ def get_pid(self) -> int:
...
+                "node_path": self.report_node_data.path,
+                "node_name": self.report_node_data.name,
+                "resource_type": self.report_node_data.resource_type,
+                "node_materialized": self.report_node_data.config.materialized,
+                "node_started_at": "TODO",
+                "unique_id": self.report_node_data.unique_id,
+                "node_finished_at": "TODO",
+                "node_status": self.node_status,
+                "run_state": self.node_state
+            }
```
nathaniel-may, line 106: I'd rather see this as a class definition than a dictionary so we can dump the `Any` type up top.

core/dbt/events/format.py
```diff
@@ -3,6 +3,22 @@
...
+    "success": "success",
+    "failed": "failed"
+}
+
+node_statuses = {
+    "running": "running",
+    "pass": "pass",
+    "error": "error",
+    "fail": "fail"
+}
```
nathaniel-may, line 19: It might feel a little boiler platey, but I think a better way to model this would be to have an ABC `NodeState` with subclasses for each of these. Those subclasses can override `__str__()` to get the string via something like `str(Queued())`. This way we can pass around `Queued()` and have mypy check it everywhere until the last minute where we need to turn it into a string.

core/dbt/events/functions.py
```diff
@@ -144,16 +144,20 @@ def scrub_dict_secrets(values: Dict) -> Dict:
...
+    event_dict = {
+        'type': 'log_line',
         'log_version': e.log_version,
         'ts': e.get_ts(),
         'pid': e.get_pid(),
         'msg': msg_fn(e),
         'level': level,
         'data': Optional[Dict[str, Any]],
-        'event_data_serialized': True
+        'event_data_serialized': True,
```
nathaniel-may, line 155: what is `event_data_serialized`? I saw it in a different PR and removed it. Maybe I shouldn't have.

core/dbt/events/functions.py
```diff
@@ -178,6 +182,7 @@ def create_json_log_line(e: T_Event, msg_fn: Callable[[T_Event], str]) -> str:
             in e.__dataclass_fields__.items()  # type: ignore[attr-defined]
             if type(y._field_type) == _FIELD_BASE
         }
+        values['data']['type'] = 'dataclass_attr'
```
nathaniel-may, line 185: hmm I don't think this is quite what we're looking for here. `type` should communicate a schema i.e. what fields and types of values to expect, however this seems to communicate the lack of a schema.

core/dbt/events/types.py
```diff
@@ -2439,6 +2462,18 @@ def message(self) -> str:
         return str(self.exc)
 
 
+# TODO: new event
```
jtcohen6, line 2465: This event seems duplicative with `NodeStart` + `PrintStartLine`. Is this new event needed, if we can support the same node details on those?

core/dbt/events/types.py
```diff
@@ -2161,6 +2178,9 @@ def message(self) -> str:
 @dataclass
 class NodeStart(DebugLevel, Cli, File):
     unique_id: str
+    report_node_data: ParsedModelNode
+    node_status: str = node_statuses['running']
+    node_state: str = node_states['started']
```
nathaniel-may, line 2183: How would you feel about making a new ABC superclass that's below `Event` that communicates that this, and other, concrete events must have these three fields? That way we can have mypy enforce these fields on some but not others which lets us get rid of them as Optional on `Event`.

core/dbt/events/format.py
jtcohen6, line 19: - I'm thinking this should be pulled off the [result contracts](https://github.com/dbt-labs/dbt-core/blob/7f2d3cd24f8715d3e36873a1f0049d8a764b2df5/core/dbt/contracts/results.py#L61-L89), rather than re-defined here
- Can we consolidate these into one `status`? I'm thinking we don't need separate `state` + `status`, since every node's progression should look like:
    - `started`
    - `compiling` (if applicable)
    - `executing` / `running` (neither word is perfect, both work)
    - final result status (one of the contracted options)

nathaniel-may (commented):
I left a bunch of comments regarding some alternatives for the core type modeling of this. I think if we get that right, the rest will come easy.

core/dbt/events/base_types.py
```diff
@@ -88,6 +90,21 @@ def get_pid(self) -> int:
             self.pid = os.getpid()
         return self.pid
 
+    def get_node_info(self) -> Any:
+        if self.report_node_data:
+            return {
+                "type": 'node_status',
```
jtcohen6, line 96: Not convinced we need this — I can't think of any events right now that should have node info, and aren't `'node_status'` in one way or another

core/dbt/events/types.py
```diff
@@ -1785,28 +1792,38 @@ def message(self) -> str:
...
     total: int
     truncate: bool
+    status: str = 'RUN'
     code: str = "Z032"
 
     def message(self) -> str:
         msg = f"START hook: {self.statement}"
         return format_fancy_output_line(msg=msg,
-                                        status='RUN',
+                                        status=self.status,
```
jtcohen6, line 1826: Definitely an improvement to structure it like this. It is a bit odd that we call this `RUN` instead of something like `START`, but for this specific event type, it's really just a question of CLI cosmetics

core/dbt/events/functions.py
emmyoop: #4326 removes the need for it.

core/dbt/events/types.py
emmyoop: Makes sense.

core/dbt/events/base_types.py
jtcohen6: Realized this is here for functional/plumbing reasons, not user-facing context

core/dbt/events/base_types.py
```diff
@@ -88,6 +90,21 @@ def get_pid(self) -> int:
...
+                "node_path": self.report_node_data.path,
+                "node_name": self.report_node_data.name,
+                "resource_type": self.report_node_data.resource_type,
+                "node_materialized": self.report_node_data.config.materialized,
+                "node_started_at": "TODO",
```
jtcohen6, line 101: Maybe tricky - we'll need to pull `node_started_at` off the [context manager](https://github.com/dbt-labs/dbt-core/blob/7f2d3cd24f8715d3e36873a1f0049d8a764b2df5/core/dbt/task/runnable.py#L207-L215) (?) for events fired while the node is running

core/dbt/events/types.py
```diff
@@ -2170,6 +2190,9 @@ def message(self) -> str:
 @dataclass
 class NodeFinished(DebugLevel, Cli, File):
     unique_id: str
+    report_node_data: ParsedModelNode
+    node_status: str = node_statuses['pass']
+    node_state: str = node_states['success']
```
jtcohen6, line 2195: I wonder if it makes sense to include the full [contracted `RunResult`](https://github.com/dbt-labs/dbt-core/blob/7f2d3cd24f8715d3e36873a1f0049d8a764b2df5/core/dbt/contracts/results.py#L117-L127) object in `NodeFinished` as well — even if we don't use it to template out the CLI message or specific `node_info` — that way, users can access its serialized representation from `data` within structured logs

jtcohen6 (commented):
Nice work pulling this together! I think we're heading in the right overall direction. I left some scattered thoughts, would be happy to discuss in more detail when I'm back online a bit later

core/dbt/events/types.py
emmyoop: In all honesty it was just a new event I added to start the work.  Then found `NodeStart`.  Have not taken the time to go back and see if it was needed - it's probably not.

core/dbt/events/format.py
nathaniel-may: This sounds good to me.

emmyoop: I agree as I wrote this.  However, I think i named these poorly.  I want to make sure I'm not missing the original intention.   A `node_state` and a `run_status` were what was originally laid out as needed.  `node_state` being state of the specific node processing (`started`, `compiling`, `executing`, `skipped`, `failed`, `success`) and a separate `run-status` (`running`, `pass`, `error`, `fail`)

core/dbt/events/functions.py
nathaniel-may: that's the PR I removed it in so that's good lol.

core/dbt/events/format.py
nathaniel-may: JK I think Jeremy had a better idea since it sounds like these are already defined somewhere else in core. Reusing those definitions, but keeping the string translations in here is probably a better way to go.

nathaniel-may: I'm not sure here. When Jeremy comes back online I think he'll know more.

jtcohen6: I think the middle two (`compiling` + `executing`) imply the existence of new events that we'll need to add [here](https://github.com/dbt-labs/dbt-core/blob/7f2d3cd24f8715d3e36873a1f0049d8a764b2df5/core/dbt/task/base.py#L279-L297)

jtcohen6: Ah, sorry, I missed the question above earlier on. My read of `run_state`:
- This is the name of the field in older structured logging. It's only ever `running` or `internal`, neither of which is very helpful.
- I don't much like the name. If we _were_ to have a field named `run_state`, that should mean the state of the overall `run`: task level, all nodes, not specific to this particular node. This isn't a very interesting field — I think it's fair to assume that, if we're firing an event that reports on the status of a specific node, the overall run state has got to be `running`. Of course, we should + do fire other events when the run concludes, including its final status.
- At first, it feels like we need different fields for (a) the state of the currently running node, and (b) the final status of the node once it finishes running. Given that these are effectively non-overlapping categories, I think we can easily combine them into a single `node_status`. An alternative approach would be to have a `current_node_state` that reports `finished` once it has finished and the `node_status` is available. I'm inclined to prefer the sparser option, and I don't think we're overloading its meaning by having it represent the status of the node both while running and once finished.

@drewbanin feel free to weigh in if you disagree with any of the above!

nathaniel-may: oh right I ran into `running` vs `internal` in our tests so we talked about this a bit. I like the sound of merging these into a single field.

core/dbt/events/base_types.py
```diff
@@ -37,6 +36,20 @@ def level_tag(self) -> str:
         return "error"
 
 
+class Node():
```
nathaniel-may, line 39: using `@dataclass` would be a bit cleaner here than using an `__init__()` method.

core/dbt/events/base_types.py
```diff
@@ -88,6 +102,18 @@ def get_pid(self) -> int:
             self.pid = os.getpid()
         return self.pid
 
+    def get_node_info(self):
+        return None
+
+
+class NodeInfo(Event, metaclass=ABCMeta):
+    report_node_data: Any
```
nathaniel-may, line 110: hmm lets see if we can make this type annotation tighter than `Any`. Looking at the call sites below, I think they're all subclasses of some parent Node type?

core/dbt/events/base_types.py
```diff
@@ -88,6 +102,18 @@ def get_pid(self) -> int:
             self.pid = os.getpid()
         return self.pid
 
+    def get_node_info(self):
+        return None
+
+
+class NodeInfo(Event, metaclass=ABCMeta):
```
nathaniel-may, line 109: it also looks like you're using `@dataclass` notation, so you should add the decorator. You'll need `# type: ignore` because of a mypy bug on using that decorator on ABCs.

core/dbt/events/base_types.py
```diff
@@ -88,6 +102,18 @@ def get_pid(self) -> int:
...
+    report_node_data: Any
+    node_status: str
+    run_state: str
+
+    def get_node_info(self):
+        return vars(Node(self.report_node_data))
```
nathaniel-may, line 115: So this is straying from the nominal typing we're using in most of the event module to use structural typing instead. If we don't have a way to explicitly describe the fields we care about returning to users, this is ok for now, but it means that if we make a change on the node types, it might blow our logging interface and break downstream consumers without us really knowing.

core/dbt/events/types.py
```diff
@@ -1785,22 +1791,32 @@ def message(self) -> str:
 
 @dataclass
 class PrintStartLine(InfoLevel, Cli, File):
+    # TODO: add node info
+    # report_node_data: ParsedModelNode
     description: str
     index: int
     total: int
+    status: str = 'RUN'
```
nathaniel-may, line 1799: this doesn't inherit from `NodeInfo`. What's this status for?

nathaniel-may (changes requested):
We're getting there! definitely some great improvements keeping these changes on track to use nominal typing. Let me know what you think of these suggestions and we'll keep tweaking from there.

core/dbt/events/types.py
emmyoop: This is literally for the output in `format_fancy_output_line`.  I pulled it out originally to share values but shouldn't as it will change the text.  Changing messages text is beyond the scope of this specific task so I'll change it back to how it was to reduce confusion.

core/dbt/events/base_types.py
emmyoop: I'm not following.  What would you expect here?

core/dbt/events/types.py
nathaniel-may: ahh. makes sense. might be worth it to drop a TODO to that effect.

core/dbt/events/base_types.py
```diff
@@ -88,6 +101,29 @@ def get_pid(self) -> int:
             self.pid = os.getpid()
         return self.pid
 
+    def get_node_info(self):
+        return None
```
nathaniel-may, line 105: this doesn't look right.

core/dbt/events/base_types.py
```diff
@@ -88,6 +101,29 @@ def get_pid(self) -> int:
...
+            node_path=self.report_node_data.path,
+            node_name=self.report_node_data.name,
+            resource_type=self.report_node_data.resource_type,
+            materialized=self.report_node_data.config.materialized,
+            node_started_at="TODO",
+            unique_id=self.report_node_data.unique_id,
+            node_finished_at="TODO",
+            node_status="TODO"  # self.node_status  # pull from contract
+        )
+        return vars(node_info)  #TODO: can just return node_info after #4326 is merged in
```
nathaniel-may, line 125: yes totally! I'll let you know when this PR is in.

core/dbt/events/base_types.py
```diff
@@ -88,6 +101,29 @@ def get_pid(self) -> int:
...
+        node_info = Node(
+            node_path=self.report_node_data.path,
+            node_name=self.report_node_data.name,
+            resource_type=self.report_node_data.resource_type,
+            materialized=self.report_node_data.config.materialized,
+            node_started_at="TODO",
+            unique_id=self.report_node_data.unique_id,
+            node_finished_at="TODO",
+            node_status="TODO"  # self.node_status  # pull from contract
+        )
```
nathaniel-may, line 124: the fields you're accessing on `report_node_data` like `path`, `name`, `resource_type`, `config.materialized` etc. ARE the type that we've defined with the type hint `Any` above. I suspect we can represent this with a union of node types instead.

nathaniel-may (commented):
Thanks for chatting over zoom! here's some comments.

core/dbt/task/runnable.py
```diff
@@ -206,17 +206,34 @@ def call_runner(self, runner):
         with RUNNING_STATE, uid_context:
             startctx = TimestampNamed('node_started_at')
             index = self.index_offset(runner.node_index)
+            runner.node.config['started_at'] = datetime.utcnow().isoformat()
```
jtcohen6, line 209: Is there any possibility that these could interfere with user-provided configs for `started_at` + `finished_at`? If so, could we name these in such a way that collisions are extremely unlikely (e.g. `dbt_internal__started_at`)?

core/dbt/events/functions.py
emmyoop: I'm not really seeing what the schema here is since it is just any attributes that were passed in.  I know we've defined them in the event classes but each one is different.

core/dbt/task/runnable.py
```diff
@@ -206,17 +206,36 @@ def call_runner(self, runner):
...
+                    fire_event(
+                        NodeFinished(
+                            report_node_data=runner.node,
+                            unique_id=runner.node.unique_id,
+                            run_result=result
+                        )
+                    )
+            del runner.node._event_status["dbt_internal__started_at"]
+            del runner.node._event_status["dbt_internal__finished_at"]
+            del runner.node._event_status["node_status"]
```
nathaniel-may, line 239: this is kind of awkward but I wouldn't be surprised if it's necessary. I think it's worth commenting though since it's odd.

core/dbt/task/base.py
```diff
@@ -286,6 +287,13 @@ def skip_result(self, node, message):
     def compile_and_execute(self, manifest, ctx):
         result = None
         with self.adapter.connection_for(self.node):
+            ctx.node._event_status['node_status'] = str(RunningStatus.Compiling)
```
nathaniel-may, line 290: is this string representation of this class name user visible? the conversion to user-visible strings should live in the events module if possible.

core/dbt/events/functions.py
nathaniel-may: my guess is that "any attributes that were passed in" is a distributed type instead of a centralized type. The schema is written down, it's just written across the whole codebase not in one place. It exists even if it's annoying to pin down. I can help pin this down in the future. we may have to release with this kind of unstructured for now though.

core/dbt/events/base_types.py
nathaniel-may: what is this needed for?

core/dbt/events/functions.py
```diff
@@ -155,9 +156,14 @@ def event_to_dict(e: T_Event, msg_fn: Callable[[T_Event], str]) -> dict:
         'data': Optional[Dict[str, Any]],
         'event_data_serialized': True,
         'invocation_id': e.get_invocation_id(),
-        'thread_name': e.get_thread_name()
+        'thread_name': e.get_thread_name(),
+        'node_info': e.get_node_info()  # TODO: update to just `node_info` and more calling this to
+                                        # `event_to_serializable_dict` after #4326 gets merged in.
+                                        # also remove refrence to node_info in Event class
```
nathaniel-may, line 162: highighting this TODO since that PR is soon to be merged in.

core/dbt/contracts/graph/parsed.py
```diff
@@ -626,6 +629,12 @@ class ParsedSourceDefinition(
     unrendered_config: Dict[str, Any] = field(default_factory=dict)
     relation_name: Optional[str] = None
     created_at: float = field(default_factory=lambda: time.time())
+    _event_status: Dict[str, Any] = field(default_factory=dict)
+
+    def __post_serialize__(self, dct):
+        if '_event_status' in dct:
+            del dct['_event_status']
+        return dct
```
nathaniel-may, line 637: what is this and when is it called?

nathaniel-may (changes requested):
I'm guessing we can delete the file `trash.py`. Added some questions.

core/dbt/events/base_types.py
emmyoop: @nathaniel-may right now in the `event_dict` I have `'node_info': e.get_node_info() ` so without this you trigger an Attribute Error.  There's a comment in `functions.py` that this will go away after I can leverage `event_to_serializable_dict` in #4326.

core/dbt/contracts/graph/parsed.py
emmyoop: @gshank can you speak to this?

core/dbt/events/base_types.py
nathaniel-may: ok perfect

core/dbt/task/base.py
gshank: The problem is that the node_status can contain either RunningStatus or NodeStatus, so it's just more work to get that straight.

core/dbt/task/runnable.py
gshank: Jeremy asked for this when we were putting these keys in the config, but we don't need it now that I changed it to put the keys in an event_status dict instead. So they names could be changed back.

core/dbt/contracts/graph/parsed.py
gshank: __post_serialize is a method that's called by the mashumaro serialization code. This removes the '_event_status' dictionary from the serialized form of the class (like manifest.json) because we should only need while running.

core/dbt/task/runnable.py
nathaniel-may: it's the deleting of all the keys that seems odd to me.

core/dbt/task/base.py
nathaniel-may: it is more work, but I don't want us to change a classname and break an app parsing our logs.

core/dbt/task/runnable.py
gshank: I guess I was concerned that there might be situations where the nodes are reused and the status information is leftover from a previous run. I'm not sure if that can happen or not...  Once we're really solid on what information we want to store there we could change it to a class and just reset it.

core/dbt/task/base.py
gshank: I'm not sure what you're asking about here. When enums like NodeStatus are serialized, I think they're always strings.

nathaniel-may (approved):
Approving with the intent that we'll do more cleanup after we cut the RC.

PR #4340, opened 2021-11-24
[#4337] Always create an invocation_id, even when not tracking

### LINKED ISSUE #4337: Ensure that all dbt invocations have an invocation_id
This is a sub-issue of #4260. 

In order to make sure we can distinguish logs from different runs we should always have an invocation_id. If tracking is off, it isn't created today so we would need to create it. (to add it to logging, add it as a class property on the superclass) (previously considered: preserve process id like it is today, global run_id reported at start, or allow some sort of user markers).

gshank:
resolves #4337

### Description

For logging purposes we always want an invocation_id, even when we're not tracking. This pull request decouples the invocation_id from dbt.tracking.active_user.

I've added a 'get_invocation_id' class method to the base Event class, but I put the generic 'get_invocation_id' and the global variable invocation_id in dbt.events.functions, because that felt more appropriate and I don't think we want to import the base_types file everywhere that we're setting the invocation_id. That meant that I had to do the import of get_invocation_id in the Event.get_invocation_id function to avoid circular imports. Let me know if you see a better way of doing that.

### Checklist

### DIFF
--- a/core/dbt/context/base.py
+++ b/core/dbt/context/base.py
@@ -15,7 +15,7 @@ from dbt.exceptions import (
     raise_compiler_error, MacroReturn, raise_parsing_error, disallow_secret_env_var
 )
 from dbt.logger import SECRET_ENV_PREFIX
-from dbt.events.functions import fire_event
+from dbt.events.functions import fire_event, get_invocation_id
 from dbt.events.types import MacroEventInfo, MacroEventDebug
 from dbt.version import __version__ as dbt_version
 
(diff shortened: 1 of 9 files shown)


nathaniel-may (changes requested):
This looks great! The last step to get it into the log messages would be to add `invocation_id: e.get_invocation_id()` to [this dict](https://github.com/dbt-labs/dbt-core/blob/6a3ddfb5e22b4edae6cf325718e09978e55a5584/core/dbt/events/functions.py#L149-L156)

nathaniel-may (approved):

PR #4341, opened 2021-11-25
Logging: restore previous (small) behaviors

jtcohen6:
From #4260:

> - [ ] (size: S) Find a way to configure logger _before_ firing any events, e.g. avoid text logs preceding JSON logs

```
$ dbt --log-format json run
{"data": null, "event_data_serialized": false, "invocation_id": "455dcb72-fcb5-41dd-bc64-4e0279a45a7c", "level": "info", "log_version": 1, "msg": "Running with dbt=1.0.0-rc2", "pid": 81044, "ts": "2021-11-25T14:56:50.296716"}
{"data": {"code": "Z046", "log_fmt": null, "msg": "* Deprecation Warning:\nThe `source-paths` config has been deprecated in favor of `model-paths`. Please\nupdate your `dbt_project.yml` configuration to reflect this change."}, "event_data_serialized": true, "invocation_id": "455dcb72-fcb5-41dd-bc64-4e0279a45a7c", "level": "warn", "log_version": 1, "msg": "* Deprecation Warning:\nThe `source-paths` config has been deprecated in favor of `model-paths`. Please\nupdate your `dbt_project.yml` configuration to reflect this change.", "pid": 81044, "ts": "2021-11-25T14:56:50.358948"}
{"data": {"code": "Z046", "log_fmt": null, "msg": "* Deprecation Warning:\nThe `data-paths` config has been deprecated in favor of `seed-paths`. Please\nupdate your `dbt_project.yml` configuration to reflect this change."}, "event_data_serialized": true, "invocation_id": "455dcb72-fcb5-41dd-bc64-4e0279a45a7c", "level": "warn", "log_version": 1, "msg": "* Deprecation Warning:\nThe `data-paths` config has been deprecated in favor of `seed-paths`. Please\nupdate your `dbt_project.yml` configuration to reflect this change.", "pid": 81044, "ts": "2021-11-25T14:56:50.360507"}
{"data": {"code": "W006", "stat_line": "2 models, 1 test, 0 snapshots, 0 analyses, 355 macros, 0 operations, 0 seed files, 0 sources, 0 exposures, 1 metric"}, "event_data_serialized": true, "invocation_id": "455dcb72-fcb5-41dd-bc64-4e0279a45a7c", "level": "info", "log_version": 1, "msg": "Found 2 models, 1 test, 0 snapshots, 0 analyses, 355 macros, 0 operations, 0 seed files, 0 sources, 0 exposures, 1 metric", "pid": 81044, "ts": "2021-11-25T14:56:50.485785"}
```

> - [ ] (size: S) feature parity for `dbt ls` + workflow automation: set stdout logger to WARN level instead of INFO (see [here](https://github.com/dbt-labs/dbt-core/blob/22416980d19888d7bbbcfcda639c16d37a80b948/core/dbt/logger.py#L517-L521) + [here](https://github.com/dbt-labs/dbt-core/blob/22416980d19888d7bbbcfcda639c16d37a80b948/core/dbt/task/list.py#L55-L59)), or better yet replace [`print()` 
here](https://github.com/dbt-labs/dbt-core/blob/22416980d19888d7bbbcfcda639c16d37a80b948/core/dbt/task/list.py#L151) with an actual fired event that can be filtered on + piped out (??)

```
$ dbt ls
14:57:26 | [ warn  ] | * Deprecation Warning:
The `source-paths` config has been deprecated in favor of `model-paths`. Please
update your `dbt_project.yml` configuration to reflect this change.
14:57:26 | [ warn  ] | * Deprecation Warning:
The `data-paths` config has been deprecated in favor of `seed-paths`. Please
update your `dbt_project.yml` configuration to reflect this change.
testy.ephemeral_model
testy.my_model
```

For both of these, I took the goal as providing exact parity with legacy logging behavior, with as small a lift as possible. That means, if we hook into legacy logger methods from `main.py` by way of task methods in a few subtle ways (`pre_init_hook`, `set_log_format`), we should set the same configs for the new event logger. If we hate this and want to keep config centralized, we can take up a more serious refactor.

> - [ ] (size: S) include snowplow tracking events in file logging (not cli) when available (i.e. without breaking `dbt clean`)

This was just a matter of adding `File` back to tracking events. I'm able to `dbt clean`, with `logs/` as a clean target, without dbt throwing up an error that it can't find the file it expects to write to. If I had to guess, the switch from `WatchedFileHandler` to `RotatingFileHandler` may have fixed this?

### Checklist

### DIFF
--- a/core/dbt/events/functions.py
+++ b/core/dbt/events/functions.py
@@ -48,7 +48,7 @@ def setup_event_logger(log_path):
     this.format_color = True if flags.USE_COLORS else False
     # TODO this default should live somewhere better
     log_dest = os.path.join(log_path, 'dbt.log')
-    level = logging.DEBUG if flags.DEBUG else logging.INFO
+    level = logging.DEBUG if flags.DEBUG else this.STDOUT_LOG.level
 
     # overwrite the STDOUT_LOG logger with the configured one
     this.STDOUT_LOG = logging.getLogger('configured_std_out')
--- a/core/dbt/events/functions.py
+++ b/core/dbt/events/functions.py
@@ -40,7 +40,7 @@ format_json = False
 invocation_id: Optional[str] = None
 
 
-def setup_event_logger(log_path):
+def setup_event_logger(log_path, level_override=None):
     make_log_dir_if_missing(log_path)
     this.format_json = flags.LOG_FORMAT == 'json'
     # USE_COLORS can be None if the app just started and the cli flags
@@ -48,7 +48,7 @@ def setup_event_logger(log_path):
     this.format_color = True if flags.USE_COLORS else False
     # TODO this default should live somewhere better
     log_dest = os.path.join(log_path, 'dbt.log')
-    level = logging.DEBUG if flags.DEBUG else this.STDOUT_LOG.level
+    level = level_override or (logging.DEBUG if flags.DEBUG else logging.INFO)
 
     # overwrite the STDOUT_LOG logger with the configured one
     this.STDOUT_LOG = logging.getLogger('configured_std_out')
--- a/core/dbt/events/types.py
+++ b/core/dbt/events/types.py
@@ -137,7 +137,7 @@ class MainReportArgs(DebugLevel, Cli, File):
 
 
 @dataclass
-class MainTrackingUserState(DebugLevel, Cli):
+class MainTrackingUserState(DebugLevel, Cli, File):
     user_state: str
     code: str = "A003"
 
@@ -2365,7 +2365,7 @@ class DisableTracking(WarnLevel, Cli, File):
 
 
 @dataclass
-class SendingEvent(DebugLevel, Cli):
+class SendingEvent(DebugLevel, Cli, File):
     kwargs: str
     code: str = "Z040"
 
--- a/core/dbt/task/base.py
+++ b/core/dbt/task/base.py
@@ -15,6 +15,7 @@ from dbt.exceptions import (
     InternalException
 )
 from dbt.logger import log_manager
+import dbt.events.functions as event_logger
 from dbt.events.functions import fire_event
 from dbt.events.types import (
     DbtProjectError, DbtProjectErrorException, DbtProfileError, DbtProfileErrorException,
@@ -64,6 +65,7 @@ class BaseTask(metaclass=ABCMeta):
         """A hook called before the task is initialized."""
         if args.log_format == 'json':
             log_manager.format_json()
+            event_logger.format_json = True
         else:
             log_manager.format_text()
 
@@ -71,6 +73,7 @@ class BaseTask(metaclass=ABCMeta):
     def set_log_format(cls):
         if flags.LOG_FORMAT == 'json':
             log_manager.format_json()
+            event_logger.format_json = True
         else:
             log_manager.format_text()
 
--- a/core/dbt/task/base.py
+++ b/core/dbt/task/base.py
@@ -15,6 +15,7 @@ from dbt.exceptions import (
     InternalException
 )
 from dbt.logger import log_manager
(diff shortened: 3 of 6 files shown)


core/dbt/task/list.py
```diff
@@ -56,6 +58,7 @@ def __init__(self, args, config):
     def pre_init_hook(cls, args):
         """A hook called before the task is initialized."""
         log_manager.stderr_console()
+        event_logger.STDOUT_LOG.level = logging.WARN
```
nathaniel-may, line 61: I'm definitely not a fan of reaching in and modifying `STDOUT_LOG` like this, but I'm fine with it as a quick fix. Can we just leave a comment here stating that this is a hack and describe what the right way is?

core/dbt/task/base.py
```diff
@@ -64,13 +65,15 @@ def pre_init_hook(cls, args):
         """A hook called before the task is initialized."""
         if args.log_format == 'json':
             log_manager.format_json()
+            event_logger.format_json = True
         else:
             log_manager.format_text()
```
nathaniel-may, line 70: it seems like these blocks won't need to exists once we refactor the early logs right? can we comment to that effect so the refactor goes smoother?

core/dbt/events/types.py
```diff
@@ -2365,7 +2365,7 @@ def message(self) -> str:
 
 
 @dataclass
-class SendingEvent(DebugLevel, Cli):
+class SendingEvent(DebugLevel, Cli, File):
```
nathaniel-may, line 2368: awesome! glad this works now 💪

core/dbt/events/functions.py
```diff
@@ -48,7 +48,7 @@ def setup_event_logger(log_path):
     this.format_color = True if flags.USE_COLORS else False
     # TODO this default should live somewhere better
     log_dest = os.path.join(log_path, 'dbt.log')
-    level = logging.DEBUG if flags.DEBUG else logging.INFO
+    level = logging.DEBUG if flags.DEBUG else this.STDOUT_LOG.level
```
nathaniel-may, line 51: If we set it to `WARN` below what's the path for setting it back to `INFO` for runs without the `--debug` flag? I don't want those runs to be stuck on warn instead of info.

nathaniel-may (changes requested):
requesting changes just so I can have a question answered and some comments. Otherwise this seems like an ok fix to me.

core/dbt/task/list.py
nathaniel-may: actually, are there info-level logs we're trying to filter out? I'm not sure why we're setting it to warn here.

core/dbt/events/functions.py
jtcohen6: Yeah, totally fair question. We could set it to `INFO` within the base (non-`list`) task `pre_init_hook`? Perhaps even include the `debug` logic there too?

Ultimately, we'll want to refactor this so that the logger is being instantiated/configured as soon as flags are available, and before all of this other stuff is happening

nathaniel-may: ok so now that I think about it, wouldn't it get reset to info when we call `setup_event_logger` if we DON'T change this line? do we need it to stay warn longer than that?

If so maybe we could change the signature of this function:
```
def setup_event_logger(log_path, level_override=None):
    ...
        level = level_override or logging.DEBUG if flags.DEBUG else logging.INFO
    ...
```

jtcohen6: @nathaniel-may Just gave this approach a try, I do think it's slightly better

nathaniel-may (approved):
Thanks for all the comments! I think this level_override looks pretty good. Happy for it to stick around for commands like list if we need it.

PR #4348, opened 2021-11-27
Make `read_files` behave functional and let `filesystem_search` accept multiple extensions

### LINKED ISSUE #4343: [Refactor] Make `read_files` stateless/functional

### Describe the ~Feature~ refactor

The [`read_files`](https://github.com/dbt-labs/dbt-core/blob/e8c267275e306d40fab3e2e661ed2b648bac29db/core/dbt/parser/read_files.py#L126) function changes the [state of the manifest](https://github.com/dbt-labs/dbt-core/blob/a9b4316346e3c25b82fc59fa1f090e14b57a65db/core/dbt/parser/manifest.py#L228) by [adding files to the `self.manifest.files`](https://github.com/dbt-labs/dbt-core/blob/e8c267275e306d40fab3e2e661ed2b648bac29db/core/dbt/parser/read_files.py#L117) attribute. Also, it changes the `project_parser_files` variable (local to the `manifest.load` method), which is created just before the call to `read_files`. 

This makes the code hard to read and have (unexpected) side effects. My suggestion would be to let `read_files` return a **new** dict that returns the `project_parser_files`. Then, when it is called inside the manifest, the manifest can add the found files to the `self.manifest.files` attribute.

NA, it's a refactor

### Who will this benefit?

Those who read dbt's source code and use dbt programmatically.

Yes

### Anything else?

It's not too important, this is part of my process of understanding dbt's internals. In this way I hope to contribute back to the dbt project.

JCZuurmond:
resolves #4343

### Description

Make `read_files` behave functional and let `filesystem_search` accept multiple extensions.

### Why am I working on this?

I would like to build a tool to test dbt macros. The tool should use `dbt-core` to retrieve macros, I think, it should use the `Manifest` class for this. Therefore, I started what the `ManifestBuilder` does, and was puzzled about what state  - like which files are read when - is added where to the `Manifest`. This rewrite pushes some of the building logic back into the `ManifestBuilder` class so that it is explicit where state is added to the manifest, instead of implicit in a function external to the `ManifestBuilder` class.

### Checklist

- [x] ~I have run this code in development and it appears to resolve the stated issue~ NA, it's a refactor
- [x] ~This PR includes tests, or tests are not required/relevant for this PR~ NA, it's a refactor

### DIFF
--- a/core/dbt/parser/manifest.py
+++ b/core/dbt/parser/manifest.py
@@ -220,12 +220,23 @@ class ManifestLoader:
         # of parsers to lists of file strings. The file strings are
         # used to get the SourceFiles from the manifest files.
         start_read_files = time.perf_counter()
-        project_parser_files = {}
         saved_files = {}
         if self.saved_manifest:
             saved_files = self.saved_manifest.files
-        for project in self.all_projects.values():
-            read_files(project, self.manifest.files, project_parser_files, saved_files)
+
+        project_parser_files = {
+            project.project_name: read_files(project, saved_files)
+            for project in self.all_projects.values()
+        }
+
+        for files in project_parser_files.values():
+            for file in files:
+                saved_files[file.file_id] = file
+        project_parser_files = {
+            project_name: list(map(lambda file: file.file_id, files))
+            for project_name, files in project_parser_files.items()
+        }
+
         orig_project_parser_files = project_parser_files
         self._perf_info.path_count = len(self.manifest.files)
         self._perf_info.read_files_elapsed = (time.perf_counter() - start_read_files)
--- a/core/dbt/parser/manifest.py
+++ b/core/dbt/parser/manifest.py
@@ -229,10 +229,13 @@ class ManifestLoader:
             for project in self.all_projects.values()
         }
 
-        for project_files in project_parser_files.values():
-            for files in project_files.values():
-                for file in files:
-                    self.manifest.files[file.file_id] = file
+        all_files = {
+            file.file_id: file
+            for project_files in project_parser_files.values()
+            for files in project_files.values()
+            for file in files
+        }
+        self.manifest.files.update(all_files)
 
         project_parser_files = {
             project_name: {
--- a/core/dbt/parser/manifest.py
+++ b/core/dbt/parser/manifest.py
@@ -229,12 +229,17 @@ class ManifestLoader:
             for project in self.all_projects.values()
         }
 
-        for files in project_parser_files.values():
-            for file in files:
-                saved_files[file.file_id] = file
+        for project_files in project_parser_files.values():
+            for files in project_files.values():
+                for file in files:
+                    self.manifest.files[file.file_id] = file
+
         project_parser_files = {
-            project_name: list(map(lambda file: file.file_id, files))
-            for project_name, files in project_parser_files.items()
+            project_name: {
+                parser_name: list(map(lambda file: file.file_id, files))
+                for parser_name, files in project_files.items()
+            }
+            for project_name, project_files in project_parser_files.items()
         }
 
         orig_project_parser_files = project_parser_files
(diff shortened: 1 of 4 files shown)


core/dbt/parser/read_files.py
```diff
@@ -108,68 +117,44 @@ def get_source_files(project, paths, extension, parse_file_type, saved_files):
...
-    source_files = get_source_files(
-        project, dirs, extension, parse_ft, saved_files
-    )
-    for sf in source_files:
-        files[sf.file_id] = sf
-        parser_files.append(sf.file_id)
-    return parser_files
-
-
 # This needs to read files for multiple projects, so the 'files'
```
JCZuurmond, line 120: I think this command can be removed. They are valid questions, but it does not help the user to understand the code better. If this is something of concern, we should create an issue for it.

core/dbt/parser/manifest.py
```diff
@@ -220,12 +220,23 @@ def load(self):
...
+
+        project_parser_files = {
+            project.project_name: read_files(project, saved_files)
+            for project in self.all_projects.values()
+        }
+
+        for files in project_parser_files.values():
+            for file in files:
+                saved_files[file.file_id] = file
+        project_parser_files = {
```
JCZuurmond, line 235: I am not _really_ happy with this dict comprehension. 

My recommendation would be to refactor `project_parser_files` to have `list[SourceFile]` as values instead of `list[str]` (which are the file ids). It would make the code below simpler as that is that is fetching the `SourceFile` from the `saved_files` all the time through the `file_id`. Memory wise it should not matter _too much_, since the `saved_files` and `project_parser_files` will have _pointers_ to the same `SourceFile`.

JCZuurmond: while we are at it: a refactor to use `Project` instead of `str` (`Project.project_name`) as keys might also make sense

core/dbt/parser/read_files.py
```diff
@@ -108,68 +117,44 @@ def get_source_files(project, paths, extension, parse_file_type, saved_files):
...
-    # Also read .yaml files for schema files. Might be better to change
-    # 'read_files_for_parser' accept an array in the future.
-    yaml_files = read_files_for_parser(
-        project, files, project.all_source_paths, '.yaml', ParseFileType.Schema, saved_files
-    )
-    project_files['SchemaParser'].extend(yaml_files)
-
-    # Store the parser files for this particular project
-    parser_files[project.project_name] = project_files
+def read_files(
```
JCZuurmond, line 124: Not that this is a **breaking** API change of this function. What is dbt's policy regarding such functions, as the public API of dbt is the CLI, but this function might still be used by some users that call `dbt-core` programatically?

JCZuurmond (commented):
Tell me what you think! Besides feedback on the PR, I would also like to hear if you would recommend using the `Manifest` and `ManifestBuilder` to fetch the macros from dbt.

JCZuurmond:
I'll fix the unit tests, the essence of the PR is there though

ismailsimsek:
+1 for the changes, im also looking into a simpler way to test dbt macros. looking at current unit and integration tests somehow its too complex to get testing setup

my test plan is like: 
1-load macros 
2-compile macro
3-validate compiled sql

gshank:
I completely understand why you want to do this; it makes a lot of sense for your use case. Unfortunately we are planning a number of changes to this area of the code in the fairly near future, and your changes do not play that well with those plans. I think it would be difficult to merge these changes and maintain the interface that you're requesting. In addition we are looking at establishing more formal interfaces when using dbt as a library; right now we don't officially support calling dbt as a library, which these changes require.

As far as testing macros, that's also something that we're actively working on right now, so it would be better to wait until we've done that work before establishing a supported interface in this area.

Thanks for all the work you've put in around this, and keep your eyes open for future updates. Hopefully we can provide you with the functionality that you need.

JCZuurmond:
@gshank : Sure, you know what is best for the whole code base.

One remark: you could [cherry pick](https://git-scm.com/docs/git-cherry-pick) the first four commits. It implements what was already suggested in the code:

>         Also read .yaml files for schema files. Might be better to change 'read_files_for_parser' accept an array in the future.

and it reduces code repetition

PR #4353, opened 2021-11-29
add thread_name to structured logging

nathaniel-may:
### Description

Adds the following field that was present in pre-1.0 json logging.

```
{ ... "thread_name": "MainThread", ... }
```

This uses the python-native thread name instead of the kernel assigned `thread_id` or `ident` for two reasons: 1 - it's what dbt used before, and 2 - kernel ids are OS specific (there are some cases where fetching these from linux doesn't get the right long value) and they can be reused by the kernel within the same running dbt process.

### Checklist

### DIFF
--- a/core/dbt/events/functions.py
+++ b/core/dbt/events/functions.py
@@ -154,7 +154,8 @@ def event_to_dict(e: T_Event, msg_fn: Callable[[T_Event], str]) -> dict:
         'level': level,
         'data': Optional[Dict[str, Any]],
         'event_data_serialized': True,
-        'invocation_id': e.get_invocation_id()
+        'invocation_id': e.get_invocation_id(),
+        'thread_name': e.get_thread_name()
     }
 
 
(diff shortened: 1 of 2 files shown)


emmyoop (approved):
Nice and simple.

core/dbt/events/functions.py
```diff
@@ -154,7 +154,8 @@ def event_to_dict(e: T_Event, msg_fn: Callable[[T_Event], str]) -> dict:
         'level': level,
         'data': Optional[Dict[str, Any]],
         'event_data_serialized': True,
-        'invocation_id': e.get_invocation_id()
+        'invocation_id': e.get_invocation_id(),
+        'thread_name': e.get_thread_name()
```
jtcohen6, line 158: (for later) Let's think about splitting `create_text_log_line` into `create_text_stdout_log_line` and `create_text_file_log_line`. I like how the thread name has been historically included in `logs/dbt.log`, but it feels like a lot of extra clutter on the CLI.

jtcohen6 (approved):

core/dbt/events/functions.py
nathaniel-may: #4354

PR #4358, opened 2021-11-29
Event buffer for structlog

### LINKED ISSUE #4320: Scalable in-memory event history
Does something graceful when it runs out of memory (small machines shouldn't explode because of logging), is queryable by a potential future server-wrapper.

iknox-fa:
resolves #4320

### Description
Adds a global event history buffer using a FIFO strategy to keep memory usage reasonable.  Can be used when wrapping dbt or in general to access past events.

### Checklist

- [] I have updated the `CHANGELOG.md` and added information about my change

### DIFF
--- a/core/dbt/events/functions.py
+++ b/core/dbt/events/functions.py
@@ -2,7 +2,7 @@
 from colorama import Style
 import dbt.events.functions as this  # don't worry I hate it too.
 from dbt.events.base_types import Cli, Event, File, ShowException
-from dbt.events.types import T_Event
+from dbt.events.types import EventBufferFull, T_Event
 import dbt.flags as flags
 # TODO this will need to move eventually
 from dbt.logger import SECRET_ENV_PREFIX, make_log_dir_if_missing, GLOBAL_LOGGER
@@ -16,9 +16,13 @@ from logging.handlers import RotatingFileHandler
 import numbers
 import os
 import uuid
-from typing import Any, Callable, Dict, List, Optional, Set, Tuple, Union
+from typing import Any, Callable, Dict, List, Optional, Set, Tuple, Union, Deque
 from dataclasses import _FIELD_BASE  # type: ignore[attr-defined]
+from collections import deque
 
+# create the global event history buffer with a max size of 1M records
+global EVENT_HISTORY
+EVENT_HISTORY: Deque = deque(maxlen=1000000)
 
 # create the global file logger with no configuration
 global FILE_LOG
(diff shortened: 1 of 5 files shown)


core/dbt/events/functions.py
```diff
@@ -16,9 +16,14 @@
...
 import uuid
-from typing import Any, Callable, Dict, List, Optional, Set, Tuple, Union
+from typing import Any, Callable, Dict, List, Optional, Set, Tuple, Union, Deque
 from dataclasses import _FIELD_BASE  # type: ignore[attr-defined]
+from collections import deque
 
+# create the global event history buffer with a max size of 100k records
+# TODO: make the maxlen something configurable from the command line via args(?)
+global EVENT_HISTORY
+EVENT_HISTORY: Deque = deque(maxlen=100000)
```
iknox-fa, line 26: @kwigley is this going to fit the bill for what's needed in Client/Server?  It should make a globally available deque of events so you can do whatever's needed.

core/dbt/events/functions.py
```diff
@@ -281,7 +286,15 @@ def send_exc_to_logger(
...
 # to files, etc.)
 def fire_event(e: Event) -> None:
-    # TODO manage history in phase 2:  EVENT_HISTORY.append(e)
+
+    # if and only if the event history deque will be completely filled by this event
+    # fire warning that old events are now being dropped
+    global EVENT_HISTORY
+    if len(EVENT_HISTORY) == (100000 - 1):
+        EVENT_HISTORY.append(e)
+        fire_event(EventBufferFull())
```
nathaniel-may, line 295: what if someone is running on a machine so small that memory runs out before this hard-coded limit? would we never log it?

core/dbt/events/functions.py
```diff
@@ -281,7 +286,15 @@ def send_exc_to_logger(
...
+    else:
```
nathaniel-may, line 296: I'd prefer this if/else to be in the form:
```python
if we_need_to_log:
    log()

append()
```

test/unit/test_events.py
```diff
@@ -70,4 +74,24 @@ def test_event_codes(self):
...
+    def test_buffer_populates(self):
+        fire_event(UnitTestInfo(msg="Test Event 1"))
+        fire_event(UnitTestInfo(msg="Test Event 2"))
+        self.assertTrue(
+            EVENT_HISTORY.count(UnitTestInfo(msg='Test Event 1', code='T006')) == 1
+        )
+
+    # ensure events drop from the front of the buffer when buffer maxsize is reached
+    def test_buffer_FIFOs(self):
+        for n in range(0,100001):
```
iknox-fa, line 90: FYI -- This is adding ~10 seconds to the unit test run time... can be fixed by correcting TODO on `core/dbt/events/functions.py` line 24.

nathaniel-may (changes requested):
one question and one minor change

test/unit/test_events.py
nathaniel-may: ahh. making it configurable so we can test with a smaller queue.

core/dbt/events/functions.py
kwigley: awesome! Ya, I believe so.

leahwicz (approved):

core/dbt/events/functions.py
leahwicz: I'm ok looking into this after the RC3 since I feel like it will be a rare case

kwigley (approved):

leahwicz:
@nathaniel-may we can merge w/o the 3.9 Mac test succeeding. It looks like a problem with Actions. The Macs are known to have issues...

nathaniel-may (approved):

PR #4361, opened 2021-11-30
better dataclass field handling

nathaniel-may:
### Description

when I changed around the way the data field was populated it never actually populated it. this fixes that.

### Checklist

### DIFF
--- a/core/dbt/events/functions.py
+++ b/core/dbt/events/functions.py
@@ -133,9 +133,9 @@ def event_to_serializable_dict(
     if hasattr(e, '__dataclass_fields__'):
         for field, value in dataclasses.asdict(e).items():  # type: ignore[attr-defined]
             _json_value = e.fields_to_json(value)
-      
+
             if isinstance(e, NodeInfo):
-                node_info = asdict(e.get_node_info())
+                node_info = dataclasses.asdict(e.get_node_info())
 
             if not isinstance(_json_value, Exception):
                 data[field] = _json_value
--- a/core/dbt/events/functions.py
+++ b/core/dbt/events/functions.py
@@ -17,7 +17,7 @@ from logging.handlers import RotatingFileHandler
 import os
 import uuid
 from typing import Any, Callable, Dict, List, Optional, Union
-from dataclasses import _FIELD_BASE  # type: ignore[attr-defined]
+import dataclasses
 
 
 # create the global file logger with no configuration
@@ -123,14 +123,13 @@ def event_to_serializable_dict(
 ) -> Dict[str, Any]:
     data = dict()
     if hasattr(e, '__dataclass_fields__'):
-        for field, value in e.__dataclass_fields__.items():  # type: ignore[attr-defined]
-            if type(value._field_type) != _FIELD_BASE:
-                _json_value = e.fields_to_json(value)
+        for field, value in dataclasses.asdict(e).items():  # type: ignore[attr-defined]
+            _json_value = e.fields_to_json(value)
 
-                if not isinstance(_json_value, Exception):
-                    data[field] = _json_value
-                else:
-                    data[field] = f"JSON_SERIALIZE_FAILED: {type(value).__name__, 'NA'}"
+            if not isinstance(_json_value, Exception):
+                data[field] = _json_value
+            else:
+                data[field] = f"JSON_SERIALIZE_FAILED: {type(value).__name__, 'NA'}"
 
     return {
         'log_version': e.log_version,
--- a/core/dbt/adapters/cache.py
+++ b/core/dbt/adapters/cache.py
@@ -335,11 +335,11 @@ class RelationsCache:
         """
         cached = _CachedRelation(relation)
         fire_event(AddRelation(relation=cached))
-        fire_event(DumpBeforeAddGraph(graph_func=self.dump_graph))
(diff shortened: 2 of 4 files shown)


leahwicz (approved):

emmyoop (approved):

core/dbt/events/functions.py
```diff
@@ -129,14 +129,13 @@ def event_to_serializable_dict(
...
     if hasattr(e, '__dataclass_fields__'):
-        for field, value in e.__dataclass_fields__.items():  # type: ignore[attr-defined]
-            if type(value._field_type) != _FIELD_BASE:
-                _json_value = e.fields_to_json(value)
-
-                if not isinstance(_json_value, Exception):
-                    data[field] = _json_value
-                else:
-                    data[field] = f"JSON_SERIALIZE_FAILED: {type(value).__name__, 'NA'}"
+        for field, value in dataclasses.asdict(e).items():  # type: ignore[attr-defined]
```
jtcohen6, line 132: This is failing in integration tests where we use `run_dbt_and_capture` with debug-level + JSON-formatted logs.

In particular, it's raising this error:
```
*** TypeError: cannot pickle '_thread.RLock' object
```
I wrapped this in a `try`/`except TypeError`, and identified the offending event as `DumpBeforeAddGraph` (`E031`). The logic that was previously here worked, the new logic raises the type error.
```python
ipdb> e
DumpBeforeAddGraph(graph_func=<bound method RelationsCache.dump_graph of <dbt.adapters.cache.RelationsCache object at 0x1104d56a0>>, code='E031')
ipdb> hasattr(e, '__dataclass_fields__')
True
ipdb> dataclasses.asdict(e)
*** TypeError: cannot pickle '_thread.RLock' object
ipdb> dataclasses.asdict(e).items()
*** TypeError: cannot pickle '_thread.RLock' object
ipdb> e.__dataclass_fields__
{'graph_func': Field(name='graph_func',type=typing.Callable[[], typing.Dict[str, typing.List[str]]],default=<dataclasses._MISSING_TYPE object at 0x10c9bd4c0>,default_factory=<dataclasses._MISSING_TYPE object at 0x10c9bd4c0>,init=True,repr=True,hash=None,compare=True,metadata=mappingproxy({}),_field_type=_FIELD), 'code': Field(name='code',type=<class 'str'>,default='E031',default_factory=<dataclasses._MISSING_TYPE object at 0x10c9bd4c0>,init=True,repr=True,hash=None,compare=True,metadata=mappingproxy({}),_field_type=_FIELD)}
ipdb> e.__dataclass_fields__.items()
dict_items([('graph_func', Field(name='graph_func',type=typing.Callable[[], typing.Dict[str, typing.List[str]]],default=<dataclasses._MISSING_TYPE object at 0x10c9bd4c0>,default_factory=<dataclasses._MISSING_TYPE object at 0x10c9bd4c0>,init=True,repr=True,hash=None,compare=True,metadata=mappingproxy({}),_field_type=_FIELD)), ('code', Field(name='code',type=<class 'str'>,default='E031',default_factory=<dataclasses._MISSING_TYPE object at 0x10c9bd4c0>,init=True,repr=True,hash=None,compare=True,metadata=mappingproxy({}),_field_type=_FIELD))])
```

jtcohen6: I see we have logic in `DumpBeforeAddGraph` to override JSON serialization for `graph_func` — but it looks like we need that logic to apply one step earlier, in `dataclasses.asdict(e)`, right before we call `fields_to_json`

jtcohen6: To @gshank's point in https://github.com/dbt-labs/dbt-core/pull/4326#issuecomment-982097428, I wonder if we'd be able to get this by using `to_dict()` (mashumaro) instead of native `dataclasses.asdict()`. Would that enable us to define a custom `__pre_deserialize__` method that converts function calls to strings?

It's probably too late for that refactor now — though given how we've centralized event functionality, it _could_ be a lighter lift than we think. I guess I'd just say, there's some hard-won wisdom on this team about python dataclasses, validation, and (de)serialization :)

gshank: The reason that they were passing along the function reference instead of the data was that they assumed putting the actual data into the log message would  be a performance hit even when they weren't logging. Is that actually true though? And it seems like with the current framework it would be called anyway. Maybe we can skip that complication and just put the dump_graph call into the fire_event.

gshank: With regards to mashumaro serialization, in the rest of the code we include it with 'dbtClassMixin', but that includes some things we wouldn't need for this use case. It's possible we could just include the mashumaro 'DataClassJsonMixin' and get the 'to_json' method for free.

nathaniel-may: Turns out function and method values are totally fine to pass to `dataclasses.asdict()`. What's not fine is to try to operate on _shared values_ between threads which is what these references to the functions are. This is why we're getting `_thread.RLock` and not something like `<bound method A.foo of <class '__main__.A'>>` I suspect that Mashumaro would not help us in this particular case.

@gshank is right that we don't need this optimization today since I believe every debug log message is logged to the file regardless of the `--debug` flag. I'll try moving that around since the produced value shouldn't be shared by the threads but we'll see.

PR #4379, opened 2021-12-01
[#4354] Different output for console and file logs

### LINKED ISSUE #4354: Text vs JSON and File vs CLI logs

### Describe the Feature

re [this comment](https://github.com/dbt-labs/dbt-core/pull/4353#discussion_r758603782), text log lines might want to be different per destination. We may want text log lines to filter from json log lines, but I'm not sure about that. Just something to consider.

_No response_

### Who will this benefit?

_No response_

_No response_

### Anything else?

_No response_

gshank:
resolves #4354

### Description

Create a different format for file logs than for console logs. 

### Checklist

### DIFF
--- a/core/dbt/events/functions.py
+++ b/core/dbt/events/functions.py
@@ -183,7 +183,10 @@ def create_file_text_log_line(e: T_Event, msg_fn: Callable[[T_Event], str]) -> s
     level: str = e.level_tag() if len(e.level_tag()) == 5 else f"{e.level_tag()} "
     thread = ''
     if threading.current_thread().getName():
-        thread = f' ({threading.current_thread().getName()}):'
+        thread_name = threading.current_thread().getName()
+        thread_name = thread_name[:10]
+        thread_name = thread_name.ljust(10, ' ')
+        thread = f' [{thread_name}]:'
     log_line = log_line + f"{color_tag}{ts} [{level}]{thread} {scrubbed_msg}"
     return log_line
 
--- a/core/dbt/events/functions.py
+++ b/core/dbt/events/functions.py
@@ -3,7 +3,7 @@ from colorama import Style
 from datetime import datetime
 import dbt.events.functions as this  # don't worry I hate it too.
 from dbt.events.base_types import Cli, Event, File, ShowException, NodeInfo, Cache
-from dbt.events.types import EventBufferFull, T_Event
+from dbt.events.types import EventBufferFull, T_Event, MainReportVersion
 import dbt.flags as flags
 # TODO this will need to move eventually
 from dbt.logger import SECRET_ENV_PREFIX, make_log_dir_if_missing, GLOBAL_LOGGER
@@ -17,6 +17,7 @@ import sys
 from logging.handlers import RotatingFileHandler
 import os
 import uuid
+import threading
 from typing import Any, Callable, Dict, List, Optional, Union
 import dataclasses
 from collections import deque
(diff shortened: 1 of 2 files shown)


core/dbt/events/functions.py
```diff
@@ -171,6 +172,22 @@ def create_text_log_line(e: T_Event, msg_fn: Callable[[T_Event], str]) -> str:
     return log_line
 
 
+def create_file_text_log_line(e: T_Event, msg_fn: Callable[[T_Event], str]) -> str:
```
jtcohen6, line 175: What do you think about:
- padding all thread names to be equal width (up to 10 characters, the length of `MainThread`)
- including `level` in the file logger — I imagine it might be helpful to scan/search for `error` + `warn` messages

core/dbt/events/functions.py
```diff
@@ -171,6 +172,22 @@ def create_text_log_line(e: T_Event, msg_fn: Callable[[T_Event], str]) -> str:
...
+    log_line: str = ''
+    # Create a separator if this is the beginning of an invocation
+    if type(e) == MainReportVersion:
+        separator = 30 * '='
+        log_line = f'\n\n{separator} {e.get_ts()} | {get_invocation_id()} {separator}\n'
```
jtcohen6, line 180: yes!! with the `invocation_id` too!! I really like this

@nathaniel-may may have thoughts about implementation. Given that `MainReportVersion` really truly is a one-of-a-kind event, and ought to be the very first one fired, this approach makes some sense to me

jtcohen6 (commented):
took for a quick spin, I like the way it's looking!

core/dbt/events/functions.py
nathaniel-may: yeah if it's truly one of a kind and only happens in this text-file combo, this approach looks like the right way to do it.

core/dbt/main.py
```diff
@@ -221,24 +221,22 @@ def track_run(task):
...
     if task.config is not None:
         log_path = getattr(task.config, 'log_path', None)
-    # we can finally set the file logger up
     log_manager.set_path(log_path)
+    # if 'list' task: set stdout to WARN instead of INFO
+    level_override = parsed.cls.pre_init_hook(parsed)
     setup_event_logger(log_path or 'logs', level_override)
+
+    fire_event(MainReportVersion(v=str(dbt.version.installed)))
+    fire_event(MainReportArgs(args=parsed))
```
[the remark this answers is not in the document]
gshank: Yeah. I personally kind of prefer to have the 'Running with' message right above 'Found...' message, but it's certainly a matter of opinion.

core/dbt/events/functions.py
```diff
@@ -162,7 +163,7 @@ def event_to_serializable_dict(
 # translates an Event to a completely formatted text-based log line
 # you have to specify which message you want. (i.e. - e.message, e.cli_msg(), e.file_msg())
 # type hinting everything as strings so we don't get any unintentional string conversions via str()
-def create_text_log_line(e: T_Event, msg_fn: Callable[[T_Event], str]) -> str:
+def create_stdout_text_log_line(e: T_Event, msg_fn: Callable[[T_Event], str]) -> str:
```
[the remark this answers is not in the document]
gshank: Ok. When I did that the pipe between timestamp and message didn't seem necessary, so I made it two spaces.

core/dbt/events/functions.py
gshank: I empathize with the desire to make them even. It would be slightly better to read. But there is also a thread name ThreadPoolExecutor-0_0.  Should I just truncate that one?

gshank: I truncated ThreadPoolExecutor-0_0 to ThreadPool, and changed to the square brackets. Making the prefix consistent does help with the reading. Still not entirely happy with the padded spaces in the thread name. We could translate MainThread to 'Thread-0' and ThreadPool to 'Thread-P'. Is it possible for the thread number to go higher than 9?

Anyway, take a look and see what you think

jtcohen6: I definitely like this more than the current version!

Switch to square brackets is good. Spacing is a teensy bit awkward, but I do think it makes it easier to read.

> Is it possible for the thread number to go higher than 9?

In theory, users can set it as high as they want, but I don't think we should worry about it being 3+ digits, let alone >4

core/dbt/main.py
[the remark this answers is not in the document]
jtcohen6: That's a fair point!

core/dbt/events/functions.py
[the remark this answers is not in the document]
jtcohen6: Two spaces looks clean! I like it:
```
$ dbt run
12:00:00  Running with dbt=1.0.0-rc3
12:00:00  Found 2 models, 0 tests, 0 snapshots, 0 analyses, 168 macros, 0 operations, 0 seed files, 0 sources, 0 exposures, 0 metrics
12:00:00
12:00:00  Concurrency: 5 threads (target='dev')
12:00:00
12:00:00  1 of 2 START table model dbt_jcohen.my_table.................................... [RUN]
12:00:00  1 of 2 OK created table model dbt_jcohen.my_table............................... [SELECT 1 in 0.06s]
12:00:00  2 of 2 START view model dbt_jcohen.my_view...................................... [RUN]
12:00:00  2 of 2 OK created view model dbt_jcohen.my_view................................. [CREATE VIEW in 0.03s]
12:00:00
12:00:00  Finished running 1 table model, 1 view model in 0.26s.
12:00:00
12:00:00  Completed successfully
12:00:00
Done. PASS=2 WARN=0 ERROR=0 SKIP=0 TOTAL=2
```
Previous versions of dbt used a pipe for messages that had timestamps. I think this is a good change.

jtcohen6 (approved):
I'm a fan of these changes

PR #4380, opened 2021-12-01
Add windows OS error suppressing for temp dir cleanups

iknox-fa:
resolves # 4261

### Description
Our new logging keeps files open in such a way that causes a problem for a test when executed in a windows-based env.  It's apparently "a thing" involving the builtin `tempfile::TemporaryDirectory` and has a work-around in python 3.10.  This PR replicates that behavior for all python versions we support.

### Checklist

### DIFF
--- a/test/integration/015_cli_invocation_tests/test_cli_invocation.py
+++ b/test/integration/015_cli_invocation_tests/test_cli_invocation.py
@@ -41,9 +41,17 @@ def temporary_working_directory() -> str:
     out : str
         The temporary working directory.
     """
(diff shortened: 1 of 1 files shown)


nathaniel-may (approved):
Ok cool so it looks like windows users won't have a problem when they install it, it's just a problem with the way we run our tests.

leahwicz (approved):

leahwicz:
Good catch!

iknox-fa:
> Ok cool so it looks like windows users won't have a problem when they install it, it's just a problem with the way we run our tests.

yeppers

PR #4381, opened 2021-12-01
move event code up a level

nathaniel-may:
### Description

code was previously under data. but I think it makes more sense at the top level.

### Checklist

### DIFF
--- a/core/dbt/events/functions.py
+++ b/core/dbt/events/functions.py
@@ -132,12 +132,12 @@ def event_to_serializable_dict(
     data = dict()
     node_info = dict()
     if hasattr(e, '__dataclass_fields__'):
+        if isinstance(e, NodeInfo):
+            node_info = dataclasses.asdict(e.get_node_info())
+
         for field, value in dataclasses.asdict(e).items():  # type: ignore[attr-defined]
             if field not in ["code", "report_node_data"]:
                 _json_value = e.fields_to_json(value)
-
-                if isinstance(e, NodeInfo):
-                    node_info = dataclasses.asdict(e.get_node_info())
 
                 if not isinstance(_json_value, Exception):
                     data[field] = _json_value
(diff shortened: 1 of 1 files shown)


emmyoop:
This makes a lot of sense but do we want to also pull out the `NodeInfo` from `data` as well?  Not sure there's value in duplicating it.

nathaniel-may:
@emmyoop I'm only seeing node info once at the top level already. Where are you seeing it duplicated?

example from a run I just did:
```
{
  "code": "Z042",
  "data": {},
  "invocation_id": "05a5ac50-eb4d-44ea-9928-7c2a67396788",
  "level": "debug",
  "log_version": 1,
  "msg": "Flushing usage events",
  "node_info": {},
  "pid": 48311,
  "thread_name": "MainThread",
  "ts": "2021-12-01T16:11:30.643316",
  "type": "log_line"
}
```

emmyoop:
`report_node_data` ends up in `data` which is what we use to build `node_info`.  You'll only see it for events with non empty `node_info`.

nathaniel-may:
Ahh I see. I just added the ignore for it.

core/dbt/events/functions.py
```diff
@@ -133,15 +133,16 @@ def event_to_serializable_dict(
...
     if hasattr(e, '__dataclass_fields__'):
         for field, value in dataclasses.asdict(e).items():  # type: ignore[attr-defined]
-            _json_value = e.fields_to_json(value)
+            if field not in ["code", "report_node_data"]:
+                _json_value = e.fields_to_json(value)
 
-            if isinstance(e, NodeInfo):
-                node_info = dataclasses.asdict(e.get_node_info())
+                if isinstance(e, NodeInfo):
+                    node_info = dataclasses.asdict(e.get_node_info())
```
emmyoop, line 140: I know I wrote this - but is there a reason it can't be above the `for` on 135 so we don't hit it possibly multiple times?

nathaniel-may: haha yeah let's fix that here.

emmyoop (approved):
Look good!

PR #4385, opened 2021-12-02
Fix release process

leahwicz:
### Description

Excluding the `dbt-<version>` artifacts from being uploaded and then tested later. We only want to test and release `dbt-core` and `dbt-postgres` going forward. Testing `dbt` just results in an error that it's deprecated.  

### Checklist

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -95,7 +95,9 @@ jobs:
       - uses: actions/upload-artifact@v2
         with:
           name: dist
-          path: dist/
+          path: |
+            dist/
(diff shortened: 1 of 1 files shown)


jtcohen6 (approved):
We will want to release `dbt==1.0.0` exactly once, and then never again. I think it's fine to do that manually

leahwicz:
> We may also want to consider waiting to do that until Monday

Wait to release `dbt-1.0.0` since it will just say deprecated?

jtcohen6:
> Wait to release dbt-1.0.0 since it will just say deprecated?

`pip install dbt` will actually raise an error

PR #4388, opened 2021-12-02
A few final logging touch-ups

jtcohen6:
Pick up from things I noticed in #4293. These are tiny non-functional / cosmetic changes only:

- Remove two unused events (they were removed in #4104)
- More structured `ConcurrencyLine`
- Instead of starting or ending messages with `\n`, precede/follow them with `EmptyLine()` events. **Note:** The legacy logger used to wrap these in `with TextOnly()`, in order to avoid writing them to JSON-formatted logs. I don't care too much about the legacy logger, but I bet we could do something very simple to achieve the same behavior with the new event logger, that is, avoid firing `EmptyLine()` events when JSON formatting is enabled.

### Checklist

- ~I have updated the `CHANGELOG.md` and added information about my change~ — centralizing in #4389


core/dbt/task/runnable.py
```diff
@@ -461,7 +459,9 @@ def run(self):
             )
 
         if len(self._flattened_nodes) == 0:
-            warn_or_error("\nWARNING: Nothing to do. Try checking your model "
+            with TextOnly():
+                fire_event(EmptyLine())
+            warn_or_error("WARNING: Nothing to do. Try checking your model "
```
emmyoop, line 464: Take out the "WARNING" text too since we want to remove that duplication.

core/dbt/events/types.py
```diff
@@ -863,7 +829,7 @@ class InvalidVarsYAML(ErrorLevel, Cli, File):
     code: str = "A008"
 
     def message(self) -> str:
-        return "The YAML provided in the --vars argument is not valid.\n"
+        return "The YAML provided in the --vars argument is not valid."
 
 
 # TODO: Remove? (appears to be uncalled)
```
emmyoop, line 835: Can we remove these too?

emmyoop (changes requested):
Just a couple more cleanup spots.

core/dbt/events/types.py
emmyoop: This highlighted wrong.  I'm referring to `HandleInternalException`, `CatchRunException`, `DetailsHandleGenericException`

nathaniel-may:
> ... avoid firing EmptyLine() events when JSON formatting is enabled.
yup. we really should do this.

core/dbt/task/runnable.py
jtcohen6: We're no longer including the log level in the CLI output (since https://github.com/dbt-labs/dbt-core/pull/4379), so it's not as duplicative as it used to be. But I can sub this out to use [`ui.warning_tag`](https://github.com/dbt-labs/dbt-core/blob/29f504e201ba512c30f59de648a00c54ac24863f/core/dbt/ui.py#L68)

core/dbt/events/types.py
jtcohen6: yep can do!

core/dbt/task/runnable.py
jtcohen6: I realize I changed `warning_tag` to do nothing! Let me change it to do something again

jtcohen6: I'm thinking we could use these for deprecation warnings too, just to give them a little colorful pop like the others:
Screenshot 2021-12-02 at 18 20 05

jtcohen6: @emmyoop let me know if you hate it!

core/dbt/events/types.py
jtcohen6: also removing `MessageHandleGenericException`, since it doesn't have a code and appears to be uncalled

core/dbt/deprecations.py
```diff
@@ -62,7 +62,7 @@ class PackageInstallPathDeprecation(DBTDeprecation):
 
 class ConfigPathDeprecation(DBTDeprecation):
     _description = '''\
-    The `{deprecated_path}` config has been deprecated in favor of `{exp_path}`.
+    The `{deprecated_path}` config has been renamed to `{exp_path}`.
```
emmyoop, line 65: Love the wording updates.

emmyoop (approved):

PR #4405, opened 2021-12-02
Rollover + backup for `dbt.log`

### LINKED ISSUE #4404: `dbt.log` should rollover at 10 mb, as before
https://docs.python.org/3/library/logging.handlers.html#rotatingfilehandler

https://github.com/dbt-labs/dbt-core/blob/74fbaa18cd7522161b8e0db97393c73d869a0498/core/dbt/logger.py#L386-L387

jtcohen6:
resolves #4404

- Configs available from the good ol' built-in `RotatingFileHandler`. Match [the old values](https://github.com/dbt-labs/dbt-core/blob/74fbaa18cd7522161b8e0db97393c73d869a0498/core/dbt/logger.py#L386-L387) exactly
- We also used to mess with `delay` + `emit`, to buffer events before the log file location was known/ready. It all makes sense! The way we're doing it now forces us to be much more intentional about when configuration happens, which is a net positive
- Confirmed this worked locally by setting `maxSize=1000` + `dbt run` :)

### DIFF
--- a/core/dbt/events/functions.py
+++ b/core/dbt/events/functions.py
@@ -82,7 +82,12 @@ def setup_event_logger(log_path, level_override=None):
 
     file_passthrough_formatter = logging.Formatter(fmt=FORMAT)
 
-    file_handler = RotatingFileHandler(filename=log_dest, encoding='utf8')
+    file_handler = RotatingFileHandler(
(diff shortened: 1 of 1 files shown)


nathaniel-may (approved):
Amazing

PR #4410, opened 2021-12-02
use reference keys instead of relations

nathaniel-may:
### Description

our Relation types override the behavior of deepcopy which means they cannot go through the json serialization pipeline which leverages asdict. This is because the dict_factory that asdict takes is not applied till the end of processing when deepcopy is internally used. The way we've overridden the behavior raises an exception when we try to serialize these values as part of a dataclass. Using reference keys instead is easier to serialize since it's just a named tuple.

### Checklist

### DIFF
--- a/core/dbt/events/types.py
+++ b/core/dbt/events/types.py
@@ -1,11 +1,11 @@
 import argparse
 from dataclasses import dataclass
+from dbt.adapters.reference_keys import _make_key, _ReferenceKey
 from dbt.events.stubs import (
     _CachedRelation,
     BaseRelation,
     ParsedModelNode,
     ParsedHookNode,
-    _ReferenceKey,
     RunResult
 )
 from dbt import ui
@@ -506,7 +506,7 @@ class CacheMiss(DebugLevel, Cli, File):
 class ListRelations(DebugLevel, Cli, File):
     database: Optional[str]
     schema: str
-    relations: List[BaseRelation]
+    relations: List[_ReferenceKey]
     code: str = "E014"
 
     def message(self) -> str:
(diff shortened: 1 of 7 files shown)


nathaniel-may:
integration tests won't really hit all the changes here. I'm going to make sure they pass locally with `DBT_LOG_FORMAT=json make integration` before merging.

leahwicz (approved):
LGTM

gshank:
It looks like the ListRelations event takes a list of relations, so we need to run the make_key thing on each element of the list

core/dbt/events/types.py
```diff
@@ -2499,8 +2491,8 @@ def message(self) -> str:
     SQLQueryStatus(status="", elapsed=0.1)
     SQLCommit(conn_name="")
     ColTypeChange(orig_type="", new_type="", table="")
-    SchemaCreation(relation=BaseRelation())
-    SchemaDrop(relation=BaseRelation())
+    SchemaCreation(relation=_make_key(BaseRelation()))
```
gshank, line 2494: I would have expected these lines to be: SchemaCreating(relation=_ReferenceKey). Is there some reason to call the _make_key function here?

nathaniel-may: it's just shorter. this code is never run, just evaluated by mypy. happy to change it though if you think it's better.

gshank: It's fine for now :)

gshank (approved):

nathaniel-may:
it's failing integration tests with json formatted logs locally so I'll be pushing a few more changes to fix those issues.

nathaniel-may:
now it hangs [here](https://github.com/dbt-labs/dbt-core/blob/b4beacdb92e01f963110457788fd11a2d914d680/core/dbt/task/runnable.py#L388) when `DBT_LOG_FORMAT=json`

nathaniel-may:
I'm going to merge anyway to get these fixes in. I can open a new pr to fix this hang once I can nail it down.

PR #4434, opened 2021-12-03
point latest version check to dbt-core package

nathaniel-may:
### Description

Because we renamed our package from `dbt` to `dbt-core` when we split out the adapters, our latest version check is wrong. Users will see this in the following message for v1.0.0:

```
% dbt --version
installed version: 1.0.0
   latest version: 0.21.1

Your version of dbt is ahead of the latest release!

Plugins:
  - snowflake: 1.0.0
```

You can manually check that this change works with the following python shell commands:
```
>>> import requests
>>> requests.get("https://pypi.org/pypi/dbt/json").json()['info']['version']
'0.21.1'
>>> requests.get("https://pypi.org/pypi/dbt-core/json").json()['info']['version']
'1.0.0'
```

### Checklist

### DIFF
--- a/core/dbt/version.py
+++ b/core/dbt/version.py
@@ -11,7 +11,7 @@ import dbt.exceptions
 import dbt.semver
 
 
-PYPI_VERSION_URL = 'https://pypi.org/pypi/dbt/json'
+PYPI_VERSION_URL = 'https://pypi.org/pypi/dbt-core/json'
 
 
 def get_latest_version():


leahwicz:
CC: @jtcohen6 we probably want to patch this right away

leahwicz (approved):

jtcohen6:
@leahwicz @nathaniel-may As soon as we push `dbt==1.0.0` to PyPi — the one that will start raising a descriptive error on `pip install dbt` — then `requests.get("https://pypi.org/pypi/dbt/json").json()['info']['version']` will start returning v1.0.0, right? And the resulting message (with current logic, before this change) will say `Up to date!`?

So:
- I'm glad we've already fixed this for all future versions
- We should actually be in ok shape for v1.0.0 _specifically_ (we're not planning to update the package named `dbt` beyond that)
- Now that plugin version numbers can differ slightly from dbt-core version numbers, this feels like a good opportunity to also add a compatibility check to `dbt --version`. Imagine something like:

```
% dbt --version
installed version: 1.0.1
   latest version: 1.0.1

Up to date!

Plugins:
  - snowflake: 1.0.0 — compatible
```
vs.
```
% dbt --version
installed version: 1.0.1
   latest version: 1.0.1

Up to date!

Plugins:
  - snowflake: 0.21.1 — not compatible!
```

nathaniel-may:
@jtcohen6 
> then requests.get("https://pypi.org/pypi/dbt/json").json()['info']['version'] will start returning v1.0.0, right?
No. We pushed dbt-core@1.0.0 not dbt@1.0.0 so that line will return '0.21.1'. The python code in the description was run after we pushed to pypi. The reason we fixed it so fast is that when people download 1.0.0 today (try it!) it will say that their version is ahead of the latest. In order for it to say it's up-to-date we have to push 1.0.1 with this tiny change.

That being said, I definitely want to add the compatibility messaging.

jtcohen6:
@nathaniel-may Let's check in again after we push `dbt==1.0.0` to PyPi (happening today). I'm thinking that, when we do, this (un-patched) code will start returning `1.0.0`:
https://github.com/dbt-labs/dbt-core/blob/9ed9936c84ef5a2706ad15f9757fb83a9c14139b/core/dbt/version.py#L14-L25

nathaniel-may:
@jtcohen6 [we did it Friday](https://pypi.org/project/dbt-core/). Is there something we should have done to handle unbounded `pip install dbt` that we should instead do today?

jtcohen6:
@nathaniel-may I'm talking about https://pypi.org/project/dbt/, not https://pypi.org/project/dbt-core/

nathaniel-may:
OH. You mean we plan on pushing 1.0.0 code to BOTH `dbt` and `dbt-core`. Unusual choice, but yes that will work. Sorry for the confusion.

jtcohen6:
Yeah! There's backstory for this in #4100

PR #4463, opened 2021-12-09
made change to test of str

### LINKED ISSUE #4433: [CORE-15] `SQLQueryStatus` event expects `get_response` to return `AdapterResponse` object only
See https://github.com/dbt-labs/dbt-spark/pull/265 for context

My proposal would be:

- **v1.0.1:** Check the type of the value returned by `get_response`, then either use its `_message` attribute (if `AdapterResponse`) or itself (if `str`)
- **v1.1.0:** Enforce that the `get_response` abstract method must return an `AdapterResponse` — not a bad change, and it's _effectively_ true, but we should wait for a minor version release, with clearer communication ahead of time

McKnight-42:
resolves #4433 

### Description
redefined status param of SQLQueryStatus to typecheck. string which passes on ._message value of AdapterResponse or the str value sent by adapter plugin.

Worked with Nate going through process of dealing with issues as they come in with minimal context, talked through several potential ways to solve the issue and even found some that would cause a bigger change than currently needed so had to take into account what might be worth time investment of task. 

-- Co Author @nathaniel-may 

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ### Fixes
 - Fix wrong url in the dbt docs overview homepage ([#4442](https://github.com/dbt-labs/dbt-core/pull/4442))
+- Fix redefined status param of SQLQueryStatus to typecheck the string which passes on `._message` value of `AdapterResponse` or the `str` value sent by adapter plugin.  ([#4463](https://github.com/dbt-labs/dbt-core/pull/4463#issuecomment-990174166)) 
 
 Contributors:
 - [remoyson](https://github.com/remoyson) ([#4442](https://github.com/dbt-labs/dbt-core/pull/4442))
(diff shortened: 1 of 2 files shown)


nathaniel-may:
The other alternative we considered was to have the status field of `SQLQueryStatus` take a value of type `Union[AdapterResponse, str]` and do the to string logic inside the event module. However, we would have had to move the definition of the `AdapterResponse` class to avoid circular imports, and override the `asdict` method to make json serialization work for `AdapterResponse` values. Since the plan is to eventually enforce the `get_response` method to return an `AdapterResponse`, such that we could rely on calling `._message`, making the more complex change didn't seem worth it for the short term.

`__str__` is appropriately defined for `AdapterResponse` so calling `str` at the call site is both simple and a graceful way to handle this for a patch in v1.0.1.

nathaniel-may (approved):
Let's just add a changelog entry before we merge.

leahwicz:
@McKnight-42 this change will need to be backported to the `1.0.latest` branch. If you need help with that, let me know

McKnight-42:
@leahwicz would love to set up time to walk through that process

PR #4465, opened 2021-12-09
[#4464] Check specifically for generic node type for some partial parsing actions

### LINKED ISSUE #4464: [CORE-19] Partial parsing: check for generic or singular test node
Error encountered from logged events:

AttributeError: 'SourceFile' object has no attribute 'get_key_and_name_for_test'

On line:
    (key, name) = schema_file.get_key_and_name_for_test(node.unique_id)

gshank:
resolves #4464

### Description

There are two types of nodes with a resource_type of 'test': generic test nodes and singular test nodes. The generic tests comes from schema files and the singular tests from SQL files in the 'tests' directory. This fixes the detection of generic test nodes when partially parsing.

### Checklist

### DIFF
--- a/core/dbt/parser/manifest.py
+++ b/core/dbt/parser/manifest.py
@@ -246,7 +246,7 @@ class ManifestLoader:
                     project_parser_files = self.partial_parser.get_parsing_files()
                     self.partially_parsing = True
                     self.manifest = self.saved_manifest
-                except Exception:
+                except Exception as exc:
                     # pp_files should still be the full set and manifest is new manifest,
                     # since get_parsing_files failed
                     fire_event(PartialParsingFullReparseBecauseOfError())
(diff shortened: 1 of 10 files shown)


core/dbt/parser/manifest.py
```diff
@@ -284,6 +284,9 @@ def load(self):
                         exc_info['full_reparse_reason'] = ReparseReason.exception
                         dbt.tracking.track_partial_parser(exc_info)
 
+                    if os.environ.get('DBT_PP_TEST'):
```
leahwicz, line 287: Is this for testing purposes?

gshank: Yes. There have been a couple of partial parsing errors that I didn't catch even though there were tests, because of the code that traps the exceptions and does a full parse. So I wanted to make sure we see the exceptions in the test cases.

nathaniel-may: This OS-level trap door makes me think we should split some of this up into better functions or methods. But the tactic definitely works.

nathaniel-may (approved):
I like all the test coverage.

PR #4466, opened 2021-12-09
feat: allow nested YAML in profile templates

### LINKED ISSUE #4462: [Feature] Allow dbt init to use profile templates with nested configurations

### Describe the Feature

Some adapters may have complex profiles with nested objects and arrays, and currently it's not possible to capture user input of such values with `dbt init` and [`profile_template.yml`](https://docs.getdbt.com/docs/contributing/building-a-new-adapter#profile_templateyml)

An example of such a profile is what [dbt-sqlite](https://github.com/codeforkjeff/dbt-sqlite) requires:

```yaml
default:
  outputs:

    dev:
      type: sqlite
      threads: 1
      database: <database name>
      schema: 'main'
      schemas_and_paths:
        main: '/my_project/data/etl.db'
      schema_directory: '/my_project/data'
      extensions:
        - '/path/to/sqlite-digest/digest.so'

    prod:
      type: sqlite
      threads: 1
      database: <database name>
      schema: 'main'
      schemas_and_paths:
        main: '/my_project/data/etl.db'
      schema_directory: '/my_project/data'
      extensions:
        - '/path/to/sqlite-digest/digest.so'

  target: dev
```

[Source](https://github.com/codeforkjeff/dbt-sqlite/blob/f4f0f04047fcfc4af96fd12f1fbc76e0eb4b6a89/dbt/include/sqlite/sample_profiles.yml)

- The `extensions` array will have as many entries as SQLite extensions the user needs.
- The `schemas_and_paths` object can have any number of key-value pairs.

So, my proposal is to add a new `yaml` type option to `profile_template.yml` prompts and use that to parse input as YAML that is  written as such to the profiles file.

1. Make arrays and objects `fixed` in the `profile_template.yml` file. This makes the user edit the `~/.dbt/profiles.yml` instead of relying solely on input.

2. A different (and harder to implement) spec where object and array prompts are walked recursively based on something like a JSON schema for the profile.

### Who will this benefit?

Users of `dbt-sqlite` (or any adapter with such configuration requirements) initializing new dbt projects.

Yes!

### Anything else?

If you're interested, I can start a PR soon 😄

edgarrmondragon:
resolves #4462

### Description

Allow YAML objects and arrays to be parsed from user input or passed directly from default values in order to fill the target profile.

### Checklist

### DIFF
--- a/core/dbt/task/init.py
+++ b/core/dbt/task/init.py
@@ -3,7 +3,7 @@ import os
 from pathlib import Path
 import re
 import shutil
-from typing import Optional
+from typing import Any, Mapping, Optional, Sequence
 
 import yaml
 import click
@@ -48,13 +48,30 @@ Need help? Don't hesitate to reach out to us via GitHub issues or on Slack:
 Happy modeling!
 """
 
+class YAMLType(click.ParamType):
+    """Custom YAML type for click parameters and prompts."""
+
+    name = "yaml"
+
+    def convert(
+        self,
+        value: Any,
+        param: Optional[click.Parameter],
+        ctx: Optional[click.Context],
+    ) -> Any:
+        if isinstance(value, str):
+            return yaml.safe_load(value)
+        return value
+
+
 # https://click.palletsprojects.com/en/8.0.x/api/?highlight=float#types
 click_type_mapping = {
     "string": click.STRING,
     "int": click.INT,
     "float": click.FLOAT,
     "bool": click.BOOL,
-    None: None
+    "yaml": YAMLType(),
+    None: None,
 }
 
 


NiallRees:
I do like the flexibility that this would offer and it definitely resolves the issue you raised! Will leave a comment around making this more interactive on the issue.

core/dbt/task/init.py
```diff
@@ -48,13 +48,30 @@
...
+    name = "yaml"
+
+    def convert(
+        self,
+        value: Any,
+        param: Optional[click.Parameter],
+        ctx: Optional[click.Context],
+    ) -> Any:
+        if isinstance(value, str):
+            return yaml.safe_load(value)
```
nathaniel-may, line 63: If `value` is an empty string, `yaml.safe_load` will return `None`. I'm pointing it out because we had a previous production bug because of this. Not sure how the `click` library handles `None` values.

nathaniel-may (commented):
I realize this is still in draft, but I'm leaving a small note here while you work on it.

PR #4474, opened 2021-12-13
copy over windows compat logic for colored log output

### LINKED ISSUE #4443: [CORE-16] [Bug] Uncolored stdout when using the `--use-colors` flag

### Current Behavior

When running models via the CLI with the `--use-colors` flag, the output is still uncolored.
`dbt --use-colors run --select <models>`

### Expected Behavior

I would expect the output to be colored.

### Steps To Reproduce

Running dbt via the CLI, execute: `dbt --use-colors run --select <models>`

### Relevant log output

_No response_

### Environment

```markdown
- OS: macOS BigSur 11.6.1
- Python: 3.9
- dbt: 1.0.0 (installed via dbt-bigquery)
```

bigquery

### Additional Context

I am running this in a Docker container via a shell script and the full command, including Docker commands, is:
```
docker run --rm \
  -e DBT_DEV_PROJECT=$DBT_DEV_PROJECT \
  -p 8080:8080 \
  -v $HOME/.config/:/root/.config \
  -v ${DBT_HOME}/dbt:/dbt \
  $target_image \
  "$@"
```
where the values passed ("$@") are `dbt --use-colors run --select <models>`

If I rebuild my image using dbt v0.21.0 then the same command results in colored output.

nathaniel-may:
### Description

I originally added this PR to try and solve compatibility issues for shells not displaying color in #4443 , but it turns out those are at the shell level (not system) which is inaccessible to Colorama. While exploring that, it looks like we didn't copy over the existing Colorama windows compatibility logic from the old logging module, so this PR puts that back in place.

### Checklist


nathaniel-may:
Tested this change as is, and it definitely doesn't work.

nathaniel-may:
this didn't work either, but we should probably still use the same colorama setup.

nathaniel-may:
To address the previous comments, those commits were attempts to solve #4443 that failed. However, this PR is still open because I think it's valuable to include the windows color setup that was included in the legacy logger.

emmyoop (approved):
LGTM

PR #4482, opened 2021-12-14
updated DepsStartPackageInstall event to use package name

### LINKED ISSUE #4461: [CORE-17] [Bug] dbt deps doesn't show which packages are installing in the logs

### Current Behavior

When I run `dbt deps` using version 1.0.0, I don't see which packages are installing corresponding to their version.

image

### Expected Behavior

I should see something like the below.

image

### Steps To Reproduce

1. Setup a dbt project in cloud or local pc using `v1.0.0`
2. Run `dbt deps`
3. View output without packages corresponding to versions

### Relevant log output

```shell
See screenshots above
```

### Environment

```markdown
- OS: ???
- Python: ???
- dbt: 1.0.0(latest)

dbt Cloud in use
```

snowflake

### Additional Context

_No response_

emmyoop:
resolves #4461 

### Description

Updated `DepsStartPackageInstall` to use package name instead of version number.  Log output is now

```
(env) emily@Emily-Rockman jaffle-shop % dbt deps
12:53:38  Running with dbt=1.0.0
12:53:39  Installing dbt-labs/dbt_utils
12:53:39    Installed from version 0.8.0
12:53:39    Up to date!
```

### Checklist

### DIFF
--- a/core/dbt/events/types.py
+++ b/core/dbt/events/types.py
@@ -1414,11 +1414,11 @@ class DepsNoPackagesFound(InfoLevel, Cli, File):
 
 @dataclass
 class DepsStartPackageInstall(InfoLevel, Cli, File):
-    package: str
+    package_name: str
     code: str = "M014"
 
     def message(self) -> str:
-        return f"Installing {self.package}"
+        return f"Installing {self.package_name}"
 
 
 @dataclass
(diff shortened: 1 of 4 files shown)


core/dbt/events/types.py
```diff
@@ -1414,11 +1414,11 @@ def message(self) -> str:
 
 @dataclass
 class DepsStartPackageInstall(InfoLevel, Cli, File):
-    package: str
+    package_name: str
     code: str = "M014"
 
     def message(self) -> str:
-        return f"Installing {self.package}"
+        return f"Installing {self.package_name}"
```
nathaniel-may, line 1421: An alternative is to take the package value and in the message method, call `{self.package.name}`. It's a little more work but its an improvement from passing strings which I think we should avoid when possible.

nathaniel-may (approved):
Looks good! I left an alternative I think would be slightly better. If it's not worth the work, this is definitely good enough to merge in.

PR #4483, opened 2021-12-14
Simplify Log Destinations

nathaniel-may:
### Description

- Simplifies destination logic from opt in to opt out. Only 9 events opt out of file logging.
- Removes the destination-specific human readable message distinction since it was never used
- Previously, since every event had to have a destination the inherited class `Event` was on top of the destination types. Now since that's no longer true, I put `Event` on top of the log levels since every event inherits from a log level. If I didn't do this we'd have to add `Event` to every concrete class which would be less than ideal.

### Reviewers
There are a lot of lines that just remove `, Cli, File` from concrete classes. To see the more important code, consider looking at each commit since I sectioned these changes off into one commit.

### Checklist

### DIFF
--- a/core/dbt/events/base_types.py
+++ b/core/dbt/events/base_types.py
@@ -11,32 +11,6 @@ from typing import Any, Optional
 # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # #
 
 
-# in preparation for #3977
-class TestLevel():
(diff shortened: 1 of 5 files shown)


gshank (approved):
Looks like some good cleanup.

emmyoop (approved):
Love the reduced repetition!

PR #4486, opened 2021-12-14
upgrade mypy to latest

nathaniel-may:
### Description

From our version to this version we gain:
- M1 wheels for developers
- `TypeGuard` from 3.10 (and typing extensions) which can be used to replace some uses of `isinstance` for type narrowing
- type narrowing with `type()` not just `isinstance`
- significant performance improvements
- adds the `--exclude` flag for specifying modules
- official 3.9 support
- many bug fixes

The new version typechecks our code base just fine. I suspect it will also typecheck reasonable code that it previously rejected making our lives a tad easier when we work with mypy.

### Checklist

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -3,7 +3,7 @@ flake8
 flaky
 freezegun==0.3.12
 ipdb
-mypy==0.782
+mypy==0.910
 pip-tools
 pytest
 pytest-dotenv


gshank:
I got the fair locally too. You might need to remove the .tox mypy directory, run 'make dev' and then run 'make mypy'.

nathaniel-may:
Since this PR mypy has released v0.930. We should use that instead when we come back to this PR.

gshank:
We use "namespace packages" for our adapter directories (no __init__.py files), so we need to use the --namespace-packages flag. And possibly specify directories, I'm a little fuzzy on that.  The __init__.py files are no longer required as of Python 3.3, but mypy doesn't support them all that well. More information: https://github.com/python/mypy/issues/1645, https://github.com/python/mypy/issues/5759.

The adapters will not work with an __init__.py file in the dbt/adapters directory, because that makes dbt.adapters a package and prevents the adapters from being installed in dbt.adapters.postgres, etc.

leahwicz:
@nathaniel-may can we close this PR out since we merged Ian's?

nathaniel-may:
closing in favor of #5171

Issue #3968, opened 2021-09-29
Adapter-specific installations


jtcohen6:
- Stop updating PyPi package `dbt` (i.e. `pip install dbt`)
    - Remove `setup.py` for package `dbt`
    - Ideally, start raising some kind of deprecation warning?
- Remove Homebrew formula + Docker image for `dbt`
- Add Homebrew formulae and Docker images for:
    - `dbt-core`
    - `dbt-postgres`
    - `dbt-redshift`
    - `dbt-snowflake`
    - `dbt-bigquery`
    - `dbt-spark`

jtcohen6:
Update: `pip install dbt` _shouldn't_ just stay stuck on v0.21.0 forever. We need to decide what this should do starting with v1.0.0. I think we should raise some kind of error/warning encouraging folks to switch to `pip install dbt-<adapter>`. In addition to that warning/error, should we:
- Install all the same packages as before
- Install just `dbt-core` (likely to raise an `Adapter <name> not found!` error)
- Install nothing (an error in the truest sense)

iknox-fa:
@jtcohen6 Whatever we do with `pip install dbt` should probably be reflected in `brew install dbt` as well yes?

jtcohen6:
@iknox-fa Thanks for those links!

### pip

After spending some more time with this, I'm leaning toward _deprecating_ rather than _disabling_. I don't think we gain a lot by bricking production pipelines that depend on (unpinned) `pip install dbt`, on the day that we release v1.0. However, we should include a warning message (during install) to the effect of:
```
"pip install dbt" is deprecated, and will stop supporting new versions on [date].
Please install your specific database adapter as `pip install dbt-<adapter> instead.
Link to docs: ...
```

Then, a few months/versions later, we should feel comfortable _disabling_ the `dbt` package by raising an error. What do you think?

### homebrew

Homebrew feels like a different beast. I don't _think_ anyone's production pipelines should depend on `brew install dbt` (weird if they did!), but certainly many users have already tapped `dbt-labs/dbt` and installed the `dbt` formula. I don't have an amazing understanding of the way Homebrew works, but I'm imagining existing users could do something like:
```bash
brew unlink dbt
brew link dbt-bigquery
```
And new users could do:
```bash
brew tap dbt-labs/dbt # rename to dbt-core?
brew install dbt-bigquery
# win: no need to install snowflake dependencies
```

[...]

PR #4487, opened 2021-12-14
New Dockerfile

### LINKED ISSUE #4495: [Feature] Update Dockerfile and docker image release process

### Describe the Feature

Our Dockerfile/image release process has been at least somewhat broken since we divided the codebase into `dbt-core` + `<dbt-adaptername>`.  This ticket encompasses the work required to create a new Dockerfile and implement a new image building process coupled to Github Actions.

N/A

### Who will this benefit?

Anyone who uses Docker

_No response_

### Anything else?

_No response_

iknox-fa:
resolves - Partially resolves #4495

### Description
Updated dockerfile.  Images created should retain the previous dockerfile's functionality without requiring all the scripting and env var juggling.

### Checklist


iknox-fa:
@leahwicz do we have an issue for this?  My github-issue-searching-fu is failing me if we do.

docker/Dockerfile
```diff
@@ -1,34 +1,105 @@
...
+RUN python -m pip install "git+https://github.com/dbt-labs/${dbt_snowflake_ref}#egg=dbt-snowflake" --no-cache-dir
+RUN python -m pip install "git+https://github.com/dbt-labs/${dbt_core_ref}#egg=dbt-core&subdirectory=core" --no-cache-dir
+
+
+##
+# dbt-spark
+##
+FROM base as dbt-spark
+RUN python -m pip install "git+https://github.com/dbt-labs/${dbt_spark_ref}#egg=dbt-spark" --no-cache-dir
+RUN python -m pip install "git+https://github.com/dbt-labs/${dbt_core_ref}#egg=dbt-core&subdirectory=core" --no-cache-dir
```
jtcohen6, line 92: Will this handle the fact that `dbt-spark` has ["extra" dependency requirements](https://github.com/dbt-labs/dbt-spark/blob/04d1cd4b4cfad68581a03a954942b26e9fec71b3/setup.py#L89-L93)? It's only usable if installed as `dbt-spark[PyHive]`, `dbt-spark[ODBC]`, or `dbt-spark[all]` (both)

docker/Dockerfile
```diff
@@ -1,34 +1,105 @@
...
-ARG WHEEL_REQUIREMENTS_SRC_PATH
-ARG DIST_PATH
+##
+# base image (abstract)
+##
+FROM python:3.9-slim-bullseye as base
 
+# Build args
+# TODO: Can we automate this so it doesn't need to be updated with each major rev?
+# N.B. dbt-postgres is currently found in the core codebase so a value of dbt-core@1.0.latest is correct
```
jtcohen6, line 13: When we bump the ~major~ minor version from `1.0` to `1.1`, we'll need to update `1.0.latest` to `1.1.latest`.

Is there any way we can automate this, via [bumpversion config](https://github.com/dbt-labs/dbt-core/blob/main/.bumpversion.cfg) or [`version-bump` GitHub Action](https://github.com/dbt-labs/dbt-core/blob/main/.github/workflows/version-bump.yml)?

docker/README.md
```diff
@@ -0,0 +1,57 @@
...
+To build an image named "my-other-dbt" that supports bigquery using core version 0.21.latest and the bigquery adapter version 1.0.0b1:
+```
+cd dbt-core/docker
+docker build \
+  --tag my-other-dbt  \
+  --target dbt-bigquery \
+  --build-arg dbt_bigquery_ref=dbt-bigquery@v1.0.0b1 \
+  --build-arg dbt_core_ref=dbt-core@0.21.latest  \
+.
+```
```
jtcohen6, line 49: Could we also include an example that builds a custom Dockerfile / image for a different database adapter? Say `dbt-myadapter`. From what we said the other day, I imagine it could look something like:
```
##
# dbt-snowflake
##
FROM dbt-core as dbt-myadapter
RUN python -m pip install dbt-myadapter # for latest stable
# or git repo for latest changes
```

jtcohen6 (commented):
Nice work! I left a few comments/questions.

I don't think we ever opened a new ticket for this, after closing https://github.com/dbt-labs/dbt-core/issues/3968#issuecomment-962633412. Let's definitely do that. The ticket should also include, incorporating image builds into the release process (GitHub Action) for dbt-core and each adapter plugin.

docker/Dockerfile
iknox-fa: Oh good catch.  It won't in it's current state but I can incorporate that.  When we build images to push to the hub which of those three spark options do we want to include?

docker/Dockerfile
iknox-fa: I haven't found an easy way to do that yet (hence the todo).  But I can put some more time into it today.

docker/README.md
jtcohen6: Love that idea

docker/Dockerfile
jtcohen6: Let's do `[all]`, if possible (though `ODBC` won't be all that useful without an ODBC driver installed). The reason these dependencies are `extras` is that they can be kinda tricky to install, but if we can have both, there's no reason not to

iknox-fa: Updated-- the default is currently set to `all`, but if that's not the best choice let me know, it's easy to change.

docker/README.md
iknox-fa: It's implemented now and tested with a few third party adapters.

docker/Dockerfile
iknox-fa: I played around with a few different options and just adding the `Dockerfile` to `.bumpversion.cfg` seems to be the best solution that isn't terribly janky.  I think this is fine, although it does mean that the definition of what's considered "up to date" will only be as accurate as the last time you sourced the `Dockerfile`. 

I'd really prefer that it dynamically set that at build time, but Docker doesn't really give you much in the way of tools for that sort of thing outside of external scripting, which is also something I'd rather avoid.

Does this work for you @jtcohen6 ?

jtcohen6: Including Dockerfile in bumpversion makes a lot of sense to me. Were you able to confirm that the bumpversion step captured the branch upgrade above? Or would we need to switch this from a branch specification to a specific version tag?

docker/README.md
```diff
@@ -0,0 +1,83 @@
...
+
+## Special cases
+There are a few special cases worth noting:
+* The `dbt-spark` database adapter comes in three different versions named `PyHive`, `ODBC`, and the default `all`.  If you wish to overide this you can use the `--build-arg` flag with the value of `dbt_spark_version=<version_name>`.  See the [docs](https://docs.getdbt.com/reference/warehouse-profiles/spark-profile) for more information.
+
+## Running an image in a container:
+The `ENTRYPOINT` for this Dockerfile is the command `dbt` so you can bind-mount your project to `/usr/app` and use dbt as normal:
+```
+docker run --mount type=bind,source=path/to/project,target=/usr/app my-dbt ls
```
kwigley, line 81: You'll likely need to mount the dbt profile and allow the container to use the host's network as well. I've used this in practice.

```bash
# cd to dbt project
# docker run -it -v path/to/project:/usr/app -v ~/.dbt:/root/.dbt my-dbt run
docker run -it -v $(pwd):/usr/app -v ~/.dbt:/root/.dbt my-dbt run 
```

docker/Dockerfile
```diff
@@ -1,34 +1,127 @@
...
+##
 
-FROM $BASE_IMAGE
-ARG BASE_REQUIREMENTS_SRC_PATH
-ARG WHEEL_REQUIREMENTS_SRC_PATH
-ARG DIST_PATH
+##
+# base image (abstract)
+##
+FROM python:3.9-slim-bullseye as base
```
kwigley, line 9: not a requirement since we never used it for our purposes, but is there a reason we removed the ability to override the base image with a build arg?

docker/Dockerfile
```diff
@@ -1,34 +1,127 @@
...
+# Set docker basics
+WORKDIR /usr/app/dbt/
 VOLUME /usr/app
 ENTRYPOINT ["dbt"]
+
+##
+# dbt-core
+##
+FROM base as dbt-core
+RUN python -m pip install "git+https://github.com/dbt-labs/${dbt_core_ref}#egg=dbt-core&subdirectory=core" --no-cache-dir
```
kwigley, line 55: I can't think of a reason why we would need an image with just dbt-core installed? Can't hurt to have, but I was wondering if you had a use case in mind.

docker/Dockerfile
```diff
@@ -1,34 +1,127 @@
...
 
+# Build args
+# N.B. The refs updated automagically every release via bumpversion
+ARG dbt_core_ref=dbt-core@v1.0.0
+ARG dbt_postgres_ref=dbt-core@v1.0.0
+ARG dbt_redshift_ref=dbt-redshift@v1.0.0
+ARG dbt_bigquery_ref=dbt-bigquery@v1.0.0
+ARG dbt_snowflake_ref=dbt-snowflake@v1.0.0
+ARG dbt_spark_ref=dbt-spark@v1.0.0
+ARG dbt_spark_version=all
```
kwigley, line 19: 👍 nice, "all" is the right default here

kwigley (approved):
This dockerfile looks good and works as I expected. I left some questions for you, but it works at the end of the day. We lost the ability for reproducible builds by using pip and letting it resolve dependencies, but I actually think that is fine!

When it comes to automating the build process with GHA, I think it will be a lot easier if each database adapter repository had an individual Dockerfile responsible. I think dbt-core should be database agnostic as possible and that includes build processes. Automated process for database specific python packages lives in the repository that holds that database adapter, I think we could relatively easily hook into that build process for building docker images.

That said, we can build images and publish them for 1.0 today with this merged manually and I don't think ^^ should block that!

docker/README.md
kwigley: or bind mounts like you mentioned! I think they are the newer better way of doing this.

docker/Dockerfile
iknox-fa: I actually had to switch to a tag based selection process as I couldn't get bumpversion to recognize `latest` as a branch.  AFAICT this should work equally well and if someone wants to manually specify the ref they can still use anything supported by pip.

docker/README.md
iknox-fa: > You'll likely need to mount the dbt profile and allow the container to use the host's network as well. I've used this in practice.
Good point-- I'll update the docs to reflect that
> or bind mounts like you mentioned! I think they are the newer better way of doing this.
Fun fact:  `-v` and `--mount` are functionally identical per [the docs](https://docs.docker.com/storage/bind-mounts/#start-a-container-with-a-bind-mount)

docker/Dockerfile
iknox-fa: Just because I didn't see a use case where we needed it.  :)
I'm happy to put it back if you have feels about it.

kwigley: all good!

docker/Dockerfile
iknox-fa: I was imagining that people might want to inherit a clean up-to-date copy of `dbt-core` as part of their own images.  Easily accomplished with `FROM dbtlabs:dbt-core as dbt-core`

docker/Dockerfile
```diff
@@ -1,34 +1,127 @@
...
+##
+FROM base as dbt-core
+RUN python -m pip install "git+https://github.com/dbt-labs/${dbt_core_ref}#egg=dbt-core&subdirectory=core" --no-cache-dir
+
+##
+# dbt-postgres
+##
+FROM base as dbt-postgres
+RUN python -m pip install "git+https://github.com/dbt-labs/${dbt_postgres_ref}#egg=dbt-postgres&subdirectory=plugins/postgres" --no-cache-dir
+RUN python -m pip install "git+https://github.com/dbt-labs/${dbt_core_ref}#egg=dbt-core&subdirectory=core" --no-cache-dir
```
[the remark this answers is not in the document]
iknox-fa: > 🌶️ take, I don't think we need to worry about the specifying the dbt-core version for each database adapter specific installation

I completely agree.  That functionality was added because I can see it being useful as a debugging tool, we don't have to build images for every combination (but we could).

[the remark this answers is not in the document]
iknox-fa: > I'm unsure how the order of the pip install lines with the --no-cache-dir work in practice as well. The command to install dbt-postgres will use pip to resolve which compatible version of dbt-core to install and all of its dependencies. The command to install dbt-core will install dbt-core with the version using the ref specified and all of its dependencies again. I think combining this on one line will allow pip to install everything and resolve dependency versions once.

So there's sort of two things to unpack here:

1. Do we want to support mix-n-matching db adapter + core versions
  My take:  Yes, if it's possible to do without causing problems for builds that don't mix versions, although I'm happy to re-asses that.
2: Multiple run commands require multiple dependency resolutions, which is inefficient.
  My take:  This is a good point-- I should be able to combine them into a single RUN.

iknox-fa:
> This dockerfile looks good and works as I expected. I left some questions for you, but it works at the end of the day. We lost the ability for reproducible builds by using pip and letting it resolve dependencies, but I actually think that is fine!

Given that part of what we get from reproducible builds is an added level of personal safety* I'd like to unpack this a little further with the team at some point

\* a fact I never considered until I did some reading on repro builds this morning.

> When it comes to automating the build process with GHA, I think it will be a lot easier if each database adapter repository had an individual Dockerfile responsible for its own docker image. I think dbt-core should be database agnostic as possible and that includes build processes. Automated processes for database specific python packages live in the repository that houses that database adapter, I think we could relatively easily hook into that build process for building docker images at release time.

I was imagining the opposite paradigm where the dockerfile might move out of all the current repos and lives on it's own in a separate repo and supports all the docker images we need to build.  This kind of falls into best practices of separation of concerns and DRY IMO.  Can you say a bit more about why multiple dockers is easier?

docker/Dockerfile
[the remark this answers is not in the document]
kwigley: 1. sounds good to me!
2. I think I more meant that I'm not sure if `pip install dbt-core dbt-postgres` and `pip install dbt-core && dbt-postgres` will result in the same thing. Maybe it is! I haven't dug into pip in some time.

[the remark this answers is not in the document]
iknox-fa: Oh I see what you're saying... I'll do some playing around and see if that has the same result.  I still like the idea of putting them in the same RUN layer though.

PR #4501, opened 2021-12-16
update log message to use adapter name

### LINKED ISSUE #4480: Cancelling Runs Using CTRL-C In DBT 1.0.0 Triggers Verbose Error Message

### Current Behavior

When I start running a valid model using my valid _dbt_ configuration and then cancel it using Ctrl-C in bash, the error message that appears is:

```The <bound method BaseAdapter.type of <class 'dbt.adapters.bigquery.impl.BigQueryAdapter'>> adapter does not support query cancellation. Some queries may still be running!```

### Expected Behavior

This seems like a regression over the earlier error message in dbt 0.19.1:
```The big query adapter does not support query cancellation. Some queries may still be running!```

### Steps To Reproduce

1.  From a valid BigQuery configuration profile and project with at least one model, enter _dbt run_ in bash.
2. Before the command finishes running, use Ctrl-C to interrupt execution.

### Relevant log output

_No response_

### Environment

```markdown
- OS: Mac OS X 11.4 (20F71)
- Python: Python 3.8.10
- dbt: 1.0.0 (along with dbt-bigquery 1.0.0)
```

bigquery

### Additional Context

I believe that [this line](https://github.com/dbt-labs/dbt-core/blob/b3039fdc7629b5027096e92ed9b7612419cf3fea/core/dbt/task/runnable.py#L362) is at the root of the issue. 

If you change the code from
 ```fire_event(QueryCancelationUnsupported(type=adapter.type))``` 
to 
```fire_event(QueryCancelationUnsupported(type=adapter.type()))```

 the error message returns to the 0.19.1 version.  It seems like the existing code assigns the actual method to QueryCancelationUnsupported.type, which is then passed into the error message, while adding parentheses is required to actually call the function and fetch the string name of the adapter so it can be passed to the error message.

emmyoop:
resolves #4480 

### Description

Reimplement log message to use adapter name instead of the object method.

This should really pass the Adapter object and then get the type() in the logging event but it's creating circular imports once again.

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 - Fix wrong url in the dbt docs overview homepage ([#4442](https://github.com/dbt-labs/dbt-core/pull/4442))
 - Fix redefined status param of SQLQueryStatus to typecheck the string which passes on `._message` value of `AdapterResponse` or the `str` value sent by adapter plugin.  ([#4463](https://github.com/dbt-labs/dbt-core/pull/4463#issuecomment-990174166))
 - Fix `DepsStartPackageInstall` event to use package name instead of version number. ([#4482](https://github.com/dbt-labs/dbt-core/pull/4482))
+- Reimplement log message to use adapter name instead of the object method. ([#4501](https://github.com/dbt-labs/dbt-core/pull/4501))
 
 Contributors:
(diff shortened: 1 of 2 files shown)


nathaniel-may (approved):
This is the kind of error mypy should pick up in the future once we tune the way it runs for our project.

PR #4505, opened 2021-12-16
Use mashumaro for serializing logging events

### LINKED ISSUE #4504: [CT-19] Logging serialization: use mashumaro for consistency
The recent logging refactor used  thePython dataclasses asdict method for serialization. For consistency with the rest of the codebase and for performance, switch to using mashumaro.

gshank:
resolves #4504

### Description

Use mashumaro to serialize json logging.

### Checklist


gshank:
Mashumaro required that the Event class be a dataclass.  Changing the Event class to a dataclass with attributes with defaults means that the attributes of subclasses all have to have defaults too (because non-default attributes can only come before attributes with defaults). In some cases Event subclasses were being constructed without keyword names, which caused the passed in value to be used for the first attribute (i.e. log_version) which didn't work, so all constructors needed to use keyword arguments.

Mashumaro also expects to serialize everything in the case, so it didn't work to have various node objects in the event classes. I switched to putting the node_info construction code in the nodes themselves and passing the node_info the even classes instead of passing in the various node objects. I would have liked to make _event_status and node_info into a separate class, but mypy really didn't like that. The stubs.py file was removed.

I created a separate core/dbt/events/serialization.py file for handling the mashumaro serialization, since these classes didn't need some of the complications in core/dbt/dataclass_schema.py. It's also a good place to put other special serialization code if we need it.

nathaniel-may:
Integration tests with json serialization on are hanging in the interop tests. I suspect there's an exception being raised in threads again.

nathaniel-may:
I get a few failures too when I run `DBT_LOG_FORMAT=json make integration` locally. But it's definitely hanging sometime after the 73% mark.

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 - Reimplement log message to use adapter name instead of the object method. ([#4501](https://github.com/dbt-labs/dbt-core/pull/4501))
 - Issue better error message for incompatible schemas ([#4470](https://github.com/dbt-labs/dbt-core/pull/4442), [#4497](https://github.com/dbt-labs/dbt-core/pull/4497))
 - Remove secrets from error related to packages. ([#4507](https://github.com/dbt-labs/dbt-core/pull/4507))
+- Some structured logging serialization cleanup ([#4504](https://github.com/dbt-labs/dbt-core/issues/4504), [#4505](https://github.com/dbt-labs/dbt-core/pull/4505))
```
nathaniel-may, line 10: Maybe mention mashumaro in the message?

core/dbt/contracts/graph/parsed.py
```diff
@@ -211,6 +211,20 @@ def write_node(self, target_path: str, subdirectory: str, payload: str):
...
+            "node_path": self.path,
+            "node_name": self.name,
+            "unique_id": self.unique_id,
+            "resource_type": self.resource_type.value,
+            "materialized": self.config.get('materialized'),
+            "node_status": str(self._event_status.get('node_status')),
+            "node_started_at": self._event_status.get("started_at"),
+            "node_finished_at": self._event_status.get("finished_at")
+        }
+        return node_info
```
nathaniel-may, line 226: Does this mean we have multiple definitions for the node info structure? why do we need this one?

core/dbt/contracts/graph/parsed.py
```diff
@@ -738,6 +752,19 @@ def has_freshness(self):
...
+        node_info = {
+            "node_path": self.path,
+            "node_name": self.name,
+            "unique_id": self.unique_id,
+            "resource_type": self.resource_type.value,
+            "node_status": str(self._event_status.get('node_status')),
+            "node_started_at": self._event_status.get("started_at"),
+            "node_finished_at": self._event_status.get("finished_at")
+        }
+        return node_info
```
nathaniel-may, line 766: Same comment around multiple node_info definitions

core/dbt/events/README.md
```diff
@@ -29,28 +29,10 @@ class PartialParsingDeletedExposure(DebugLevel, Cli, File):
...
-        return f"{self.node_name} had overly verbose result of {run_result}"
-
-    @classmethod
-    def asdict(cls, data: list) -> dict:
-        return dict((k, str(v)) for k, v in data)
-
-```
 
-All values other than `code` and `report_node_data` will be included in the `data` node of the json log output.
+All values other than `code` and `node_info` will be included in the `data` node of the json log output.
```
nathaniel-may, line 35: Thanks for getting the README too!

core/dbt/events/base_types.py
```diff
@@ -42,31 +30,28 @@ def __post_init__(self):
...
     log_version: int = 1
     ts: Optional[datetime] = None  # use getter for non-optional
     ts_rfc3339: Optional[str] = None  # use getter for non-optional
     pid: Optional[int] = None  # use getter for non-optional
-    node_info: Optional[Node]
 
     # four digit string code that uniquely identifies this type of event
     # uniqueness and valid characters are enforced by tests
-    @abstractproperty
+    @property
```
nathaniel-may, line 44: This is an issue. Removing it as abstract means mypy can no longer detect that events are missing codes. You can see this by commenting out a code field in types.py and running mypy.

core/dbt/events/base_types.py
```diff
@@ -167,3 +119,8 @@ class NoFile():
 # prevents an event from going to stdout
 class NoStdOut():
     pass
+
+
+@dataclass
+class NodeInfo():
+    node_info: Dict[str, Any] = field(default_factory=dict)
```
nathaniel-may, line 126: Lets add a comment to this class to describe what it represents

core/dbt/events/functions.py
```diff
@@ -138,25 +137,21 @@ def event_to_serializable_dict(
...
+        log_line = e.to_dict()
+    except AttributeError as exc:
         event_type = type(e).__name__
         raise Exception(  # TODO this may hang async threads
-            f"type {event_type} is not serializable to json."
-            f" First make sure that the call sites for {event_type} match the type hints"
-            f" and if they do, you can override the dataclass method `asdict` in {event_type} in"
-            " types.py to define your own serialization function to a dictionary of valid json"
-            " types"
+            f"type {event_type} is not serializable. {str(exc)}"
```
nathaniel-may, line 147: When do we expect this exception to be hit with the new implementation? What do we do to convert something that's not serializable to something that is inside the event module?

core/dbt/events/functions.py
```diff
@@ -138,25 +137,21 @@ def event_to_serializable_dict(
...
         )
 
     if isinstance(e, NodeInfo):
-        node_info = dataclasses.asdict(e.get_node_info())
+        node_info = e.node_info
 
     for field, value in log_line.items():  # type: ignore[attr-defined]
-        if field not in ["code", "report_node_data"]:
+        if field not in ["code", "node_info", "log_version", "ts", "ts_rfc3339", "pid"]:
```
nathaniel-may, line 154: should we generate this list at runtime instead? If we add a new field on `Event` we'd have to remember to put it here too.

core/dbt/events/functions.py
```diff
@@ -315,7 +310,7 @@ def fire_event(e: Event) -> None:
     global EVENT_HISTORY
     if len(EVENT_HISTORY) == (flags.EVENT_BUFFER_SIZE - 1):
         EVENT_HISTORY.append(e)
-        fire_event(EventBufferFull())
+        EVENT_HISTORY.append(EventBufferFull())
```
nathaniel-may, line 313: I don't think this is right. This won't log that the event history is full, it will just append it to the history.

core/dbt/events/serialization.py
```diff
@@ -0,0 +1,40 @@
...
+            out += "Z"
+        return out
+
+    def deserialize(self, value):
+        return (
+            value if isinstance(value, datetime) else parse(cast(str, value))
+        )
+
+
+class dbtClassEventMixin(DataClassDictMixin):
```
nathaniel-may, line 34: can we change the name of this class? I know there's precedence for it, but I'd rather it be called `Serializable` or something more descriptive to why it's there and what it adds

core/dbt/events/test_types.py
```diff
@@ -14,7 +14,7 @@
 
 @dataclass
 class IntegrationTestInfo(InfoLevel, NoFile):
-    msg: str
+    msg: str = ''
```
nathaniel-may, line 17: I don't like having defaults we hope are never going to get used. But since it's just test events I can live with this if we really need it.

core/dbt/events/types.py
```diff
@@ -109,16 +101,16 @@ def message(self) -> str:
 
 @dataclass
 class MainEncounteredError(ErrorLevel, NoFile):
-    e: BaseException
+    e: str = ''
```
nathaniel-may, line 104: This undermines one of the major points of the event module. mypy should be able to enforce that we are providing real values for each of the expected fields. Yes mypy knows that we have a value for each of the fields we care about, but we'll never know if this default is used. And if this default is used something is very wrong.

core/dbt/utils.py
```diff
@@ -613,3 +615,34 @@ def _connection_exception_retry(fn, max_attempts: int, attempt: int = 0):
...
+    var_args.update(flag_dict)
+    dict_args = {}
+    # remove args keys that clutter up the dictionary
+    for key in var_args:
+        if key == 'cls':
+            continue
+        if var_args[key] is None:
+            continue
+        # TODO: add more default_false_keys
+        default_false_keys = (
```
nathaniel-may, line 634: don't we have a definition of this in the flags module already?

core/dbt/utils.py
```diff
@@ -613,3 +615,34 @@ def _connection_exception_retry(fn, max_attempts: int, attempt: int = 0):
             _connection_exception_retry(fn, max_attempts, attempt + 1)
         else:
             raise ConnectionException('External connection exception occurred: ' + str(exc))
+
+
+def args_to_dict(args):
```
nathaniel-may, line 620: can we leave a comment why this function is required? I'm not entirely sure what it's for.

test/unit/test_events.py
```diff
@@ -22,6 +21,8 @@
 def get_all_subclasses(cls):
     all_subclasses = []
     for subclass in cls.__subclasses__():
+        if subclass in [NodeInfo, AdapterEventBase, TestLevel, DebugLevel, WarnLevel, InfoLevel, ErrorLevel]:
```
nathaniel-may, line 24: The reason we didn't need this before is because we filtered out ABCs for checks. I'd prefer to keep that approach since it means we won't have to remember to update this list. We can do that filtering here since I don't think we'll ever need the ABCs from this function.

nathaniel-may (changes requested):
I have a couple of concerns with this approach:
- mypy can no longer catch malformed events such as events that have no `code` field.
- mypy can no longer catch when we fail to provide event attributes even when mypy runs everywhere
- it looks like when an event takes a type that isn't serializable we are forced to convert it to a serializable type _outside_ the event module which leaks a core abstraction of the event module back out to the rest of the code base

core/dbt/events/base_types.py
gshank: But if we leave it as an abstractproperty then we can't make it a dataclass, which mashumaro requires.

gshank: Checking for a code could be tested instead of relying on mypy.

gshank:
"it looks like when an event takes a type that isn't serializable we are forced to convert it to a serializable type outside the event module which leaks a core abstraction of the event module back out to the rest of the code base"

I'm not sure that I agree that it is the event module's job to enable serialization for outside classes. In any case, serializing classes that are not needed for the event message feels unnecessary, since it gets thrown away anyway. I think that we're better off providing only the information that is needed for the event/message.

gshank:
"mypy can no longer catch when we fail to provide event attributes even when mypy runs everywhere"

Providing defaults for attributes of subclasses is something that the Python dataclasses module forces you to do.  I think it's probably more important to consistently make the classes dataclasses than make a couple of methods abstractmethods, when executing those methods without an override will throw an error just like the abstract base class.

nathaniel-may:
The way I see it, the event module is tasked with translating dbt's internal representations into a structured output that is versioned as an API so that downstream applications can rely on it. Serialization and schema conformity are both major parts of creating this API to an enterprise quality. I want this code to be written in such a way that if you are not modifying the event module you do not need to know anything about how it was built or what it needs to continue to conform to the logging API version.

If using dataclasses forces us to weaken our development-time guarantees for this API, we should switch tactics and do something else.

core/dbt/events/types.py
gshank: The problem is that we don't know what exception class is being passed in, so I thought it was safer in that case to stringify the exceptions. Which is all that we end up doing with it anyway. We can serialize a BaseException instead, but it amounts to the same thing.

core/dbt/utils.py
gshank: A definition of what? This is just code that was moved to utils from task/runnable.py

test/unit/test_events.py
gshank: ABC filtering wasn't working with the Event class as a dataclass.

nathaniel-may: Ahh right.

core/dbt/utils.py
gshank: It's there because we get errors with some of the fields when dict-ifying, like PosixPath and the 'cls' key. This code was used to create the 'args' part of the run_results, and this just reuses that code.

core/dbt/utils.py
nathaniel-may: gotcha. I was thinking of a different dictionary

core/dbt/events/types.py
nathaniel-may: yeah this one is a special case because it really could be any subclass of `BaseException`. I still like taking a `BaseException` here because it reduces the surface area of valid `MainEncounteredError` values which makes mypy more helpful. The runtime effect is only the same if we are sure we are always stringifying an exception at every call site which mypy can enforce when structured this way.

core/dbt/events/base_types.py
nathaniel-may: If this is the case, I don't think Mashumaro is a good fit for serialization in this module.

nathaniel-may: Is there a way we could make something lower on the hierarchy inherit from the serialization class? like the loglevel classes maybe?

CHANGELOG.md
gshank: Done

core/dbt/contracts/graph/parsed.py
gshank: Yeah. I made an attempt to put it in a separate class to be included but dataclasses makes the order of default and non-default attributes irritating.  It's certainly better to only have one though, so I tweaked it and made it work.

core/dbt/contracts/graph/parsed.py
gshank: See above.

core/dbt/events/types.py
gshank: I changed this to be a BseException again. I'm not convinced it's an inmprovement...

core/dbt/utils.py
gshank: I've added a comment to the function.

core/dbt/events/types.py
```diff
@@ -1775,7 +1727,7 @@ def message(self) -> str:
                                         index=self.index,
                                         total=self.total,
                                         execution_time=self.execution_time,
-                                        truncate=self.truncate)
+                                        truncate=True)
```
nathaniel-may, line 1730: nice catch. no need to thread this value through the log structure.

core/dbt/events/types.py
```diff
@@ -1843,8 +1792,7 @@ class PrintWarnTestResult(WarnLevel, NodeInfo):
     index: int
     num_models: int
     execution_time: int
-    failures: List[str]
-    report_node_data: ParsedModelNode
```
nathaniel-may, line 1847: how does `report_node_data` work now?

core/dbt/events/types.py
```diff
@@ -2132,17 +2068,12 @@ def message(self) -> str:
 @dataclass
 class NodeFinished(DebugLevel, NodeInfo):
     unique_id: str
-    report_node_data: ParsedModelNode
-    run_result: RunResult
+    run_result: Dict[str, Any]
```
nathaniel-may, line 2071: what's wrong with using `RunResult` instead of a dict

core/dbt/utils.py
```diff
@@ -613,3 +615,40 @@ def _connection_exception_retry(fn, max_attempts: int, attempt: int = 0):
...
+    var_args.update(flag_dict)
+    dict_args = {}
+    # remove args keys that clutter up the dictionary
+    for key in var_args:
+        if key == 'cls':
+            continue
+        if var_args[key] is None:
+            continue
+        # TODO: add more default_false_keys
+        default_false_keys = (
```
nathaniel-may, line 640: what happens if we forget to update this?

test/unit/test_events.py
nathaniel-may, line 24: lets add a comment here that if the test breaks because of abcs that this list might have to be updated

test/unit/test_events.py
```diff
@@ -424,7 +432,7 @@ def test_all_serializable(self):
 
         # if we have everything we need to test, try to serialize everything
         for event in sample_values:
-            d = event_to_serializable_dict(event, lambda _: event.get_ts_rfc3339())
+            d = event_to_serializable_dict(event)
```
nathaniel-may, line 435: just noting here for posterity that this reduction in expressiveness of `event_to_serializable_dict` is fine because it's only called by json log lines which all use the same timestamp formatting. It was originally intended to be called by log messages going to stdout too but that's unnecessary flexibility now.

core/dbt/events/serialization.py
```diff
@@ -0,0 +1,49 @@
...
+    def deserialize(self, value):
+        return (Exception(value))
+
+
+class DateTimeSerialization(SerializationStrategy):
+    def serialize(self, value):
+        out = value.isoformat()
+        # Assume UTC if timezone is missing
+        if value.tzinfo is None:
+            out += "Z"
```
nathaniel-may, line 25: let's centralize this logic re: my comment on the `Event` base class

nathaniel-may (changes requested):
These recent commits address all 3 of my major concerns from my last review. Mypy is still able to be helpful in the ways it was planned to and there's a way to override serialization in `serialization.py`. The following remarks are much smaller issues likely with much simpler solutions. Thanks for all the hard work here!

core/dbt/adapters/factory.py
```diff
@@ -66,7 +66,7 @@ def load_plugin(self, name: str) -> Type[Credentials]:
             # if we failed to import the target module in particular, inform
             # the user about it via a runtime error
             if exc.name == 'dbt.adapters.' + name:
-                fire_event(AdapterImportError(exc=exc))
+                fire_event(AdapterImportError(exc=str(exc)))
```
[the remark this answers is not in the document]
gshank: Good catch. Fixed.

core/dbt/events/base_types.py
```diff
@@ -59,6 +41,10 @@ class Event(metaclass=ABCMeta):
     def code() -> str:
         raise Exception("code() not implemented for event")
 
+    @abstractmethod
+    def to_dict(self):
```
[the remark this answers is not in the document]
gshank: Added.

core/dbt/events/base_types.py
```diff
@@ -167,3 +121,8 @@ class NoFile():
 # prevents an event from going to stdout
 class NoStdOut():
     pass
+
+
+@dataclass
+class NodeInfo():
+    node_info: Dict[str, Any]
```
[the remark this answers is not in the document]
gshank: I don't think it's worthwhile to do that at this time. The dictionary is only created in one place by one method.

core/dbt/events/functions.py
```diff
@@ -18,10 +17,11 @@
 import os
 import uuid
 import threading
-from typing import Any, Callable, Dict, List, Optional, Union
-import dataclasses
+from typing import Any, Dict, List, Optional, Union
 from collections import deque
 
+global LOG_VERSION
+LOG_VERSION = 2
```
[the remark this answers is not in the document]
gshank: This was a typo. Changing back to 1.

core/dbt/events/types.py
gshank: There are some complications in serializing RunResult which don't seem appropriate to pull into the logging code.

core/dbt/utils.py
gshank: Nothing. It's just to limit the dictionary to the interesting bits, and is totally unrelated to this pull request.

test/unit/test_events.py
gshank: ok. Done.

core/dbt/events/serialization.py
gshank: This is not really related to the ts calls. I'm not sure what you mean by this comment.

nathaniel-may: Right now this PR has two places where we serialize datetimes, and they are not being serialized the same way. One here, and one in `functions.py::get_ts_rfc3339`. What I think should happen is to replace the body of this function with the contents of  `get_ts_rfc3339`, remove `get_ts_rfc3339`, and replace existing calls to it with calls to this serialization method.

core/dbt/events/base_types.py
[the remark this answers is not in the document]
nathaniel-may: I disagree that it's not worth it. I'd rather rely on mypy now than wait till the integrity of this dictionary breaks something to fix it.

nathaniel-may (changes requested):
Added some additional comments. There are still 5 unaddressed comments from my previous review.

core/dbt/events/serialization.py
gshank: The serialization of datetimes in the events matches the serialization of datetimes that we do in the artifacts. The serialization that's done in the functions.py code is for visual presentation in the logs. These are different purposes, in my opinion.

core/dbt/events/functions.py
gshank: Have changed this.

core/dbt/events/serialization.py
gshank: This name is in line with the other dbt mashumaro serialization class, which is dbtClassMixin.

core/dbt/events/test_types.py
gshank: I forgot to update the test_types file. Have done this now.

core/dbt/events/base_types.py
gshank: Done.

core/dbt/events/base_types.py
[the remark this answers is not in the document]
gshank: I don't see any situation in which the "integrity of the dictionary" will break something. The dictionary is only created in one place in the NodeInfoMixin class. My imagination fails at imagining a situation in which mypy would save us from anything here.

core/dbt/events/serialization.py
gshank: On second thought, I think we could use some standardization in serializing the datetimes, but I'd like to do that as a separate ticket because it requires moving around some things that are not related to this pull request. I've opened #4608 for that.

core/dbt/events/functions.py
gshank: We don't actually have any fields on Event now. I've changed this to delete 'code' and 'node_info' from the dictionary, but I'm wondering if it wouldn't be better to just leave node_info in the data and not move it up.

core/dbt/events/base_types.py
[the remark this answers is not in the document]
gshank: You could open a ticket to change this if you think it's worth it, but since I don't I'm not going to do it as part of this pull request.

nathaniel-may:
Just finished a quick sync with @gshank to go over the remaining comments. Here are the outcomes from that conversation:

- The `node_info` field is going to move underneath `data`. This will require an API version increase.
- The serialization strategy for datetime objects in serialization.py seems to be currently unused so it's going to be removed with a comment for now to avoid confusion between other datetime serializations. Cleaning up these across dbt-core is recorded in #4608
- The "node_info" value is actually not being constructed here, it's just being passed through as a dictionary, so let's skip turning it into a nominal type here for now.
- `RunResult` is a little tougher to serialize with mashumaro. Let's comment what makes that difficult and leave it as a dictionary value.

core/dbt/events/types.py
gshank: I misremembered the issue here. It was circular imports. You had added a stubs file to address that, but mashumaro doesn't work with the fake classes.

nathaniel-may (approved):
There's just that one value test for the log version that still needs to be updated, but this looks good to me!

PR #4507, opened 2021-12-16
scrub message of secrets

### LINKED ISSUE #4508: Scrub secret env vars from targeted exceptions
In v1:
- secret env vars (prefixed `DBT_ENV_SECRET_`) are for use exclusively in `packages.yml` + `profiles.yml`
- We scrub those secret values from log messages produced by all events

However, exceptions still do not pass through the event logger, and so they may still raise/print unscrubbed information. This came up in a test case with `bad_package_spec()`.

- Temporary fix (#4507): Let's scrub all exceptions that are raised by `deps` for `git`-based packages, since this is the use case we know about
- Long-term fix: Exception messages should pass through the same centralized event logger

emmyoop:
resolves #4508

### Description

Scrub `bad_package_spec()` of secrets it may contain in output.

### Checklist

### DIFF
--- a/core/dbt/exceptions.py
+++ b/core/dbt/exceptions.py
@@ -399,7 +399,7 @@ class CommandError(RuntimeException):
     def __init__(self, cwd, cmd, message='Error running command'):
         super().__init__(message)
         self.cwd = cwd
-        cmd = scrub_secrets(cmd, env_secrets())
+        cmd = scrub_secrets(str(cmd), env_secrets())
         self.cmd = cmd
         self.args = (cwd, cmd, message)
 
@@ -467,8 +467,7 @@ def raise_dependency_error(msg) -> NoReturn:
     raise DependencyException(scrub_secrets(msg, env_secrets()))
 
 
-def raise_cloning_problem(repo) -> NoReturn:
-    repo = scrub_secrets(repo, env_secrets())
+def raise_git_cloning_problem(repo) -> NoReturn:
     msg = '''\
     Something went wrong while cloning {}
     Check the debug logs for more information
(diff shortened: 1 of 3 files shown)


emmyoop:
I have not checked that this works locally yet but wanted to get up the PR.

core/dbt/exceptions.py
```diff
@@ -692,9 +692,11 @@ def missing_materialization(model, adapter_type):
 
 
 def bad_package_spec(repo, spec, error_message):
-    raise InternalException(
-        "Error checking out spec='{}' for repo {}\n{}".format(
-            spec, repo, error_message))
+    msg = "Error checking out spec='{}' for repo {}\n{}".format(spec, repo, error_message)
+    for secret in get_secret_env():
+        msg = str(msg).replace(secret, "*****")
```
nathaniel-may, line 697: There's a function for this we can reuse: `functions.py::scrub_secrets` if you call it like this:
```
scrub_secrets(msg, env_secrets())
```

nathaniel-may: In any case, events like this shouldn't be leaking secrets anyway because we are already running this secret scrub across the entire JSON string. And if they're not using json logs, ad-hoc fields like this aren't put into the message.

nathaniel-may: Is there a ticket associated with this leaked secret we're trying to catch?

nathaniel-may: oops my bad. Just catching that this is an exception not an event. this should totally work as a stop gap.

jtcohen6:
To be safe, let's also add scrubbing logic to all potential exceptions raised in the `dbt.deps.git` and `dbt.clients.git` modules:
- `warn_or_error`
- `raise_dependency_error`
- `RuntimeException`

There's two other worrying spots where we `raise` directly:
- [here](https://github.com/dbt-labs/dbt-core/blob/067b861d30bb03618ea163b811ab007127bfe2af/core/dbt/clients/git.py#L138-L141)
- [here](https://github.com/dbt-labs/dbt-core/blob/067b861d30bb03618ea163b811ab007127bfe2af/core/dbt/clients/git.py#L24-L25)

Best move might be wrapping / converting those into real `RuntimeException`.

leahwicz (approved):

emmyoop:
Converting [this one](https://github.com/dbt-labs/dbt-core/blob/067b861d30bb03618ea163b811ab007127bfe2af/core/dbt/clients/git.py#L24-L25) into a `RuntimeException` caused a lot of test failures that I was having trouble tracking down.  My fix just still scrubs the output but is a `CommandResultError`.

PR #4548, opened 2022-01-04
Mirroring issues to Jira

leahwicz:
### Description

These are initial GitHub Actions that will create and close issues in Jira when they are creates/closed in GitHub. 

Future PR: mirror labels and assignees 

### Checklist

### DIFF
--- a/.github/workflows/jira-closed.yml
+++ b/.github/workflows/jira-closed.yml
@@ -1,0 +1,40 @@
+# **what?**
+# Closes mirrored Jira issue
+
+# **why?**
+# Jira needs to be kept up-to-date
+
+# **when?**
+# On issue closing, deletion
+
+name: Jira Issue Closed
+
+on:
+  issues:
+    types: [closed, deleted]
+
+jobs:
+  close-issue:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Jira login
+        uses: atlassian/gajira-login@master
+        env:
+          JIRA_API_TOKEN: ${{ secrets.JIRA_API_TOKEN }}
+          JIRA_BASE_URL: ${{ secrets.JIRA_BASE_URL }}
+          JIRA_USER_EMAIL: ${{ secrets.JIRA_USER_EMAIL }}
+
+      - name: Extract issue from title
+        id: extract
+        env:
+          TITLE: "${{ github.event.issue.title }}"
+        run: |
+          jira=$(echo -n $TITLE | awk '{print $1}' | awk -F'[][]' '{print $2}')
+          echo ::set-output name=jira::$jira
+
+      - name: Transition issue in Jira
+        if: startsWith(steps.extract.outputs.jira, 'CORE-')
+        uses: atlassian/gajira-transition@master
+        with:
+          issue: "${{ steps.extract.outputs.jira }}"
+          transition: "Done"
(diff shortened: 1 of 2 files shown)


.github/workflows/jira-closed.yml
```diff
@@ -0,0 +1,40 @@
...
+        run: |
+          jira=$(echo -n $TITLE | awk '{print $1}' | awk -F'[][]' '{print $2}')
+          echo ::set-output name=jira::$jira
+
+      - name: Transition issue in Jira
+        if: startsWith(steps.extract.outputs.jira, 'CORE-')
+        uses: atlassian/gajira-transition@master
+        with:
+          issue: "${{ steps.extract.outputs.jira }}"
+          transition: "Done"
```
nathaniel-may, line 40: Is this the only time we're going to want to use this logic? should this be an action that can be reused?

.github/workflows/jira-creation.yml
```diff
@@ -0,0 +1,51 @@
...
+          fields: '{ "labels": ["core"], "customfield_10045": "${{ github.event.issue.html_url }}", "customfield_10046": ${{ github.event.issue.number }} }'
+ 
+      - name: Update Issue Title
+        run: |
+          curl --request PATCH \
+          --url https://api.github.com/repos/${{ github.repository }}/issues/${{ github.event.issue.number }} \
+          --header 'authorization: Bearer ${{ secrets.GITHUB_TOKEN }}' \
+          --header 'content-type: application/json' \
+          --header 'Accept: application/vnd.github.v3+json' \
+          --data '{ "title": "[${{ steps.create.outputs.issue }}] ${{ github.event.issue.title }}" }'
```
nathaniel-may, line 51: if we split out the above title parsing logic into its own action it would be nice for the creation logic to live along side it so that we don't have to hunt down all the usages of serialization and deserialization if we ever want to change something.

jtcohen6 (approved):
Thanks for setting these up! Let's give 'em a whirl

nathaniel-may (commented):
I'm guessing that we're not mirroring jira issues into github issues which might form an infinite loop with jira-creation.yml right?

.github/workflows/jira-closed.yml
leahwicz: Do you mean the transition logic? Or do you mean the extract and transition logic together?

.github/workflows/jira-creation.yml
leahwicz: My current thought is to work on the other workflows (i.e. label mirroring, assignee mirroring) next and then identify what I'm copying and pasting and then look at the refactor out into their own actions then. It makes sense to do that but it isn't the easiest thing to create these customer actions so I want to make sure we really do need them before I start factoring them out

gshank (approved):
Looks good!

.github/workflows/jira-closed.yml
nathaniel-may: This code looks like it's doing string parsing work and I imagine it won't be the only place we'll need it.
```
jira=$(echo -n $TITLE | awk '{print $1}' | awk -F'[][]' '{print $2}')
          echo ::set-output name=jira::$jira
```

.github/workflows/jira-creation.yml
nathaniel-may: Sounds reasonable to me.

.github/workflows/jira-closed.yml
leahwicz: I see what you mean. I'm going to save this for the next iteration as well just to see if this works well and also if we do reuse this

Issue #4438, opened 2021-12-06
Add adapter compatibility messaging to dbt --version


nathaniel-may:
### Is there an existing feature request for this?

- [X] I have searched the existing issues

### Describe the Feature

Now that plugins are not shipped alongside dbt-core, users can have incompatible adapters installed alongside dbt-core. We should tell them this when they run `dbt --version` and use green / red color to make it clear. Original proposal by @jtcohen6 can be found [here](https://github.com/dbt-labs/dbt-core/pull/4434#issuecomment-986298586) in #4434 .

```
% dbt --version
installed version: 1.0.1
   latest version: 1.0.1

Up to date!

Plugins:
  - snowflake: 1.0.0 — compatible
  ```
  
  ```
% dbt --version
installed version: 1.0.1
   latest version: 1.0.1

Up to date!

Plugins:
  - snowflake: 0.21.1 — requires update!
  
To update dbt-snowflake please run:
    pip install dbt-snowflake --upgrade

  ```

jtcohen6:
yes!! you beat me to it :)

nkyuray:
Hi, 

I want to work on this, but this will be the first for me. If you give some insights on this issue and help, would be happy to work on this.

leahwicz:
@nkyuray sorry we missed your message! Were you still thinking of working on this?

nkyuray:
Hi @leahwicz, yes absolutely:D

joellabes:
:portalparrot: from #4582, re: 
> get[ting] plugin-specific "up to date" / "out of date" checks

>We could also look into adding checks for each installed plugin, to see if they have a newer compatible version available—at the teensy cost of a bit more semver logic, and a few additional requests to the PyPi API.

I think that this change would also resolve a nitpick I have with the current phrasing - "compatible" and "out of date" are not necessarily the two states I'd describe, because "Compatible" feels like weirdly non-specific word to reassure me. 

If I'm on dbt-core 1.0.1, here's what I think would be useful to see for different variants of dbt-snowflake:
- 0.21.1: [red] Out of date! (and the existing guidance on how to upgrade)
- 1.0.0: [yellow] Update available! (and the existing guidance on how to upgrade)
- 1.0.1: [green] Up to date!

With a PR already in flight, I don't necessarily want to move your cheese @nkyuray - let me know what you think. 

[...]

PR #4536, opened 2021-12-28
Validate project names in interactive dbt init

amirkdv:
resolves #4490

### Description

This PR adds project name validation to `dbt init`.

- workflow: ask the user to provide a valid project name until they do.
- new integration tests
- supported scenarios:
  - `dbt init`
  - `dbt init -s`
  - `dbt init [name]`
  - `dbt init [name] -s`

### Checklist

- [x] I have signed the [CLA](https://docs.getdbt.com/docs/contributor-license-agreements)
- [x] I have run this code in development and it appears to resolve the stated issue
- [x] This PR includes tests, or tests are not required/relevant for this PR
- [x] I have updated the `CHANGELOG.md` and added information about my change

PR #4565, opened 2022-01-11
adapter compability messaging added.

nkyuray:
### Description

adapter compatibility messaging added. 
https://github.com/dbt-labs/dbt-core/issues/4438#issue-1072291308

### Checklist

### DIFF
--- a/core/dbt/version.py
+++ b/core/dbt/version.py
@@ -10,6 +10,8 @@ import requests
 import dbt.exceptions
 import dbt.semver
 
+from dbt.ui import green, red
+from dbt import flags
 
 PYPI_VERSION_URL = 'https://pypi.org/pypi/dbt-core/json'
 
@@ -30,6 +32,8 @@ def get_installed_version():
 
 
 def get_version_information():
+    flags.USE_COLORS = True if not flags.USE_COLORS else None
+
     installed = get_installed_version()
     latest = get_latest_version()
 
@@ -44,16 +48,26 @@ def get_version_information():
 
     plugin_version_msg = "Plugins:\n"
     for plugin_name, version in _get_dbt_plugins_info():
-        plugin_version_msg += '  - {plugin_name}: {version}\n'.format(
-            plugin_name=plugin_name, version=version
-        )
+        version_major = dbt.semver.VersionSpecifier.from_version_string(version).major
+        plugin_update_msg = ''
+        if installed.major == version_major:
+            compatibility_msg = green('compatible')
+        else:
+            compatibility_msg = red('requires update!')
+            plugin_update_msg = ("  To update dbt-{} please run:\n"
+                                 "    pip install dbt-{} --upgrade\n\n".format(plugin_name, plugin_name))
+
+        plugin_version_msg += ("  - {}: {} - {}\n"
+                               "{}").format(plugin_name, version, compatibility_msg, plugin_update_msg)
+
     if latest is None:
         return ("{}The latest version of dbt could not be determined!\n"
                 "Make sure that the following URL is accessible:\n{}\n\n{}"
-                .format(version_msg, PYPI_VERSION_URL, plugin_version_msg))
+                .format(version_msg, PYPI_VERSION_URL, plugin_version_msg)
+                )
 
     if installed == latest:
-        return "{}Up to date!\n\n{}".format(version_msg, plugin_version_msg)
+        return f"{version_msg}{green('Up to date!')}\n\n{plugin_version_msg}"
 
     elif installed > latest:
         return ("{}Your version of dbt is ahead of the latest "
@@ -91,7 +105,7 @@ def _get_dbt_plugins_info():
                 f'dbt.adapters.{plugin_name}.__version__'
             )
         except ImportError:
-            # not an adpater
+            # not an adapter
             continue
         yield plugin_name, mod.version
 
--- a/core/dbt/version.py
+++ b/core/dbt/version.py
@@ -16,9 +16,9 @@ from dbt import flags
 PYPI_VERSION_URL = 'https://pypi.org/pypi/dbt-core/json'
 
 
-def get_latest_version():
+def get_latest_version(version_url: str = PYPI_VERSION_URL):
     try:
-        resp = requests.get(PYPI_VERSION_URL)
+        resp = requests.get(version_url)
         data = resp.json()
         version_string = data['info']['version']
     except (json.JSONDecodeError, KeyError, requests.RequestException):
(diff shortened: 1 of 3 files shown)


ChenyuLInx:
Hey @nkyuray, thanks for contributing this! I will have more bandwidth to test this next Tuesday

core/dbt/version.py
```diff
@@ -44,16 +48,26 @@ def get_version_information():
...
-            plugin_name=plugin_name, version=version
-        )
+        version_major = dbt.semver.VersionSpecifier.from_version_string(version).major
+        plugin_update_msg = ''
+        if installed.major == version_major:
+            compatibility_msg = green('compatible')
+        else:
+            compatibility_msg = red('requires update!')
+            plugin_update_msg = ("  To update dbt-{} please run:\n"
+                                 "    pip install dbt-{} --upgrade\n\n".format(plugin_name, plugin_name))
```
joellabes, line 58: Love the really actionable feedback here! My only comment is that not all users install dbt via pip - I really like the idea of making the instructions easy, but it'd be confusing to give pip instructions to someone who installed dbt via Homebrew etc.

Not sure how to square that circle; perhaps we have to fall back to linking them to the [docs](https://docs.getdbt.com/dbt-cli/install/overview)?

ChenyuLInx: Looks like the pip message is provided as an example in the original feature request. But I agree with @joellabes 's fallback to docs idea here. It would make later updates easier also.

nkyuray: Hi @ChenyuLInx @joellabes 

Ok, then simply we need to put a link of related docs page, right?
Besides the yellow one will be added also.

Thanks.

ChenyuLInx: That would be perfect! Thanks @nkyuray

ChenyuLInx:
@nkyuray One more thing, don't forget to add yourself to the changelog. So we can give you credit on the release note(we are looking for ways to automate this, but right now you have to do it manually). Here's a [example](https://github.com/dbt-labs/dbt-core/pull/4536/files).

ChenyuLInx:
@nkyuray, we are thinking of getting this into the 1.0.2 release that aims to have an rc version later next week. We think more folks would appreciate and benefit from what you are adding here! Do you think you get time to work on it before that? If not we can either merge this in and leave the rest as an improvement or I am happy to update it a bit if you are okay with it!

nkyuray:
Hi @ChenyuLInx,

Sorry I couldn't take time to work on this, but I will probably finalize it on Friday.
Thanks.

ChenyuLInx:
> Hi @ChenyuLInx,
> 
> Sorry I couldn't take time to work on this, but I will probably finalize it on Friday. Thanks.

Hi @nkyuray, please don't say sorry. Thank you so much for contributing this!!

core/dbt/version.py
```diff
@@ -44,16 +48,30 @@ def get_version_information():
...
-        plugin_version_msg += '  - {plugin_name}: {version}\n'.format(
-            plugin_name=plugin_name, version=version
-        )
+        plugin_version = dbt.semver.VersionSpecifier.from_version_string(version)
+        plugin_update_msg = ''
+        if installed == plugin_version:
+            compatibility_msg = green('Up to date!')
+        else:
+            if installed.major == plugin_version.major:
+                compatibility_msg = yellow('Update available!')
```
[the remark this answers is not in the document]
jtcohen6: Agreed! I think we could take one of two approaches here:
- Compare `{major}.{minor}` only, to ensure that the plugin version is compatible with the `dbt-core` version
- Run PyPi requests for each installed plugin, to see what its latest available version is

In the future we'd want a combination of both, right? Goal being to say whether installed plugins are `Latest compatible` with the installed `dbt-core` version

jtcohen6:
> I just run the tests and noticed a few minor flake8 checks failed. Currently, you have to run it by yourself. We are going to have some tutorial on getting this into your pre-commit hooks soon.

Per CI run, it looks like the flake8 failures are just:
```
core/dbt/version.py:62:100: E501 line too long (110 > 99 characters)
core/dbt/version.py:65:100: E501 line too long (103 > 99 characters)
```

ChenyuLInx (approved):
updated compare and added pulling from PYPI. @nkyuray  Thank you so much for doing this!!! Sorry we really want this change in 1.0.2 so I just changed it a bit myself.

leahwicz (approved):

core/dbt/version.py
```diff
@@ -44,16 +52,40 @@ def get_version_information():
...
     for plugin_name, version in _get_dbt_plugins_info():
-        plugin_version_msg += '  - {plugin_name}: {version}\n'.format(
-            plugin_name=plugin_name, version=version
-        )
+        plugin_version = dbt.semver.VersionSpecifier.from_version_string(version)
+        latest_plugin_version = get_latest_version(version_url=get_package_pypi_url(plugin_name))
+        plugin_update_msg = ''
+        if installed == plugin_version or (
+            latest_plugin_version and plugin_version == latest_plugin_version
+        ):
```
jtcohen6, line 60: It's possible that plugins will put out patch releases _before_ `dbt-core` puts out a patch release. Imagine:
- `dbt-core==1.0.1` and `dbt-synapse==1.0.2` are both available
- I have `dbt-synapse==1.0.1` installed locally
- `installed == plugin_version` will return `True`, at the same time, there's a compatible plugin update available

core/dbt/version.py
```diff
@@ -44,16 +52,40 @@ def get_version_information():
...
+            compatibility_msg = green('Up to date!')
+        else:
+            if latest_plugin_version:
+                if installed.major == plugin_version.major:
```
jtcohen6, line 64: `dbt-core` + plugins are compatible at the minor-version level. So I think this might want to be:
```suggestion
                if installed.minor == plugin_version.minor:
```
If `installed.minor != plugin_version.minor`, we'd want to raise `red('Not compatible!')`

jtcohen6 (commented):
This is a great change! Thank you @nkyuray for the contribution, and @ChenyuLInx for seeing this through.

I think there are a few small intricacies that may require us to revisit this logic in the future. I left two comments with edge cases that come to mind. If those seem like quick fixes, all the better. In either case, I'm happy to move forward with the current logic for v1.0.2rc1.

core/dbt/version.py
ChenyuLInx: I didn't realize this case, and the rule in the next comment, I will open a new issue and get these changes in

PR #4585, opened 2022-01-18
drop support for Python 3.7.0 and 3.7.1

### LINKED ISSUE #4584: [CT-56] Remove support for Python 3.7.0, 3.7.1
As raised in #4564, running dbt v1.0 with Python 3.7.0 or 3.7.1 results in a Python exception being thrown. We should disallow installing dbt with these versions of Python. This issue is resolved with Python 3.7.2.

nathaniel-may:
resolves #4584

Running dbt v1.0.0 and v0.21 (and possibly other dbt versions) with these Python versions causes a Python exception to be thrown. This disallows installing dbt with when running pip with these versions.

### Checklist

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -80,5 +80,5 @@ setup(
         'Programming Language :: Python :: 3.8',
         'Programming Language :: Python :: 3.9',
     ],
-    python_requires=">=3.7",
+    python_requires=">=3.7.2",
 )


leahwicz (approved):

jtcohen6:
Should this go in for `1.0.2`? Generally, we shouldn't raise dependency lower bounds in patch releases, though it would be better to raise a more helpful error than the one currently encountered. My sense, though, is that bumping `python_requires` will only _prevent_ installation if users are running 3.7.0/3.7.1; it won't actually raise a more helpful error at runtime if already installed.

nathaniel-may:
@jtcohen6 that's correct. I can see good reasons for including in either version:

- 1.0.2 - for everyone updating from a version prior to 1.0 to 1.0.latest, this will be a useful check for them.
- 1.1.0 - for everyone currently on 1.0 updating to the latest patch release this would be a smoother transition.

Which version would you like this to go in?

jtcohen6:
I buy it!

nathaniel-may:
@leahwicz we could also backport this to 0.21 since we know that version is affected as well, but I don't think patches were in the plan for versions prior to 1.0.

leahwicz:
@nathaniel-may we only want to patch 0.21 for security bugs at this point so let's not backport this to that release.

For `1.0.2` I would be hesitant to backport it there as well and just wait for `1.1.0`. For patches, we really should only be doing bug fixes and I don't consider this a bug fix. If you feel differently, lets chat and we can figure out a plan ahead

nathaniel-may:
That's fair. Normally I'd argue that this fixes a bug because it would prevent users from issuing `dbt run` and seeing an exception. _However_, because we're not backporting it to 0.21 it will just install 0.21 instead and that will blow up in the same way anyway.

Happy to wait till `1.1.0` for this one since the only affected users are people using Python v3.7.0 and v3.7.1 which should be fairly rare at this point.

jtcohen6:
@nathaniel-may I realized we actually want this change to `core/setup.py`, not `setup.py` (which is for the now-deprecated `dbt` package)!

I'm fine with the reasoning above, so I switched #4584 to the `1.1.0` milestone.

Could we make sure this gets a changelog entry for `1.1.0`? We shouldn't expect this to affect many people, but any time we up a lower bound (especially Python version), we should record it IMO

PR #4587, opened 2022-01-19
Clarify "incompatible package version" error msg

### LINKED ISSUE #4586: [CT-57] [Feature] Clarify what to do when getting a version incompatibility error

### Describe the Feature

When trying to run dbt with a package that's incompatible due to require-dbt-version, the error message says "Check the requirements for the 'dbt_utils' package, or run dbt again with--no-version-check". 

I think "Check the requirements" is too ambiguous - we could direct them to check for an updated version of the package. 

Extra for experts _could_ be to go and check the hub automatically (if it's a hosted package), but that would slow down the response due to the network overhead and potentially couple some things too close together 🤔 

### Who will this benefit?

People who are confused when upgrading dbt versions. Post 1.0 this should be less of an issue, but there's a lot of users who are still to update to 1.0 for the first time so there's still time for this to be relevant. 

Definitely if it's just a phrasing change; maybe if we start futzing around with the hub. 

### Anything else?

_No response_

joellabes:
resolves #4586 

### Description
Gives an actionable error message when a package isn’t compatible with the current version of dbt core.

### Checklist

### DIFF
--- a/core/dbt/config/project.py
+++ b/core/dbt/config/project.py
@@ -45,7 +45,7 @@ INVALID_VERSION_ERROR = """\
 This version of dbt is not supported with the '{package}' package.
   Installed version of dbt: {installed}
   Required version of dbt for '{package}': {version_spec}
-Check the requirements for the '{package}' package, or run dbt again with \
+Check for an updated version of the '{package}' package, or run dbt again with \
 --no-version-check
 """
 
@@ -54,7 +54,7 @@ IMPOSSIBLE_VERSION_ERROR = """\
 The package version requirement can never be satisfied for the '{package}
 package.
   Required versions of dbt for '{package}': {version_spec}
-Check the requirements for the '{package}' package, or run dbt again with \
+Check for an updated version of the '{package}' package, or run dbt again with \
 --no-version-check
 """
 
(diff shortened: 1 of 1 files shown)


core/dbt/config/project.py
```diff
@@ -45,7 +45,7 @@
 This version of dbt is not supported with the '{package}' package.
   Installed version of dbt: {installed}
   Required version of dbt for '{package}': {version_spec}
-Check the requirements for the '{package}' package, or run dbt again with \
+Check for an updated version of the '{package}' package, or run dbt again with \
```
leahwicz, line 48: Would there be any scenario where we would need an older version of a package and not the newest? Honestly not sure so just asking

leahwicz (approved):

core/dbt/config/project.py
joellabes: Interesting! I guess if you started the 1.0 upgrade process, changed your dbt-utils to 0.8.0, then abandoned ship and went back to 0.21, then you'd run into the same error message and should technically go to an _older_ version...

This assumes that you merged your changes into main before you were confident in the entire migration which sounds ill-advised, but is possible.

joellabes: @leahwicz i about it more, I'd also be on board with the phrasing "a different version" if we really wanted to protect against that edge case

leahwicz: Thanks @joellabes! I wasn't sure how likely it was to happen. I just was worried about giving too specific of instructions and have them be incorrect for an edge case

leahwicz:
Thanks @joellabes! If you're good with this then I'll merge it in

joellabes:
@leahwicz yeah let's do it!

PR #4602, opened 2022-01-20
Performance Regression Testing Revamp Stage 1

### LINKED ISSUE #4021: [CT-14] Enable Regression Testing Notifications
### Description
In #3054 we introduced the basic functionality of regression testing. This ticket aims to track the next stage of development with the following goals:

- [ ] developers are notified internally via slack when a nightly run surpasses some threshold.
- [ ] performance measurements are directly comparable to product goals
- [ ] performance measurements for released versions are locked in at release time
- [ ] measure performance metrics for compilation as well as parsing

### Additional context
#4012 is a perfect example of a performance regression we would like to catch in development, but would require adding compilation to our measured metrics. 

### Implementation
There are two stages of implementation. Data collection should be done first so that we have more data to work with in the notification stage. Adding data collection for compilation performance requires adding a database to the tests which adds a bit of necessary complexity.

1. Data collection
- [ ] Make sure the generated artifacts contain all the fields we would like to track
    - [x] timestamp
    - [x] total parse time
    - [x] baseline & dev stddev parse value
    - [x] baseline & dev median parse value
    - [ ] baseline & dev stddev compile value
    - [ ] baseline & dev median compile value
    - [ ] ms / file or number of files in project
    - [ ] branch names
    - [ ] raw runs (in case we want to do our own stats later in dbt)
- [ ] Ship this data to snowflake

2. Notification
- [ ] Remove regression detection from rust runner (this logic will live in dbt)
- [ ] Decide thresholds for notifications
- [ ] Implement an actual notification via failing dbt test or looker
- [ ] Add performance regression watermarking to the release process

nathaniel-may:
resolves #4021 

### Description

This is the first part of a multi-PR feature. Merging this will have no effect on the code base, but should be reviewed in full because this code will be activated by the actions that are under development in #4872. The goal of the overarching feature is to run performance regression tests on PRs with minimal failures due to noise. This work shifts the entire strategy for detecting performance regressions away from fully modeling performance characteristics on each test, to sampling and comparing to a baseline model. The strategy and the math to support it is outlined in detail in the README.

To start, this PR only includes a single test project with 2000 model files for the performance of `dbt parse`, but once the actions are merged it will be able to run efficiently on each PR. The next step is to add more commands and projects. This stage is outlined in #4768.

### Reviewers
Please read the README in detail. It is intended to provide all the information necessary to develop this without me. If there's something that's unclear, please bring it up.

To navigate the runner, start in `main::run_app` which handles all the IO, then hit the two top-level functions: `fs::model` and `calculate::regressions`. The split between fs and calculate modules isn't perfect and I could probably have some better naming of these functions. Suggestions are welcome.

The runner is not perfect, and I have left TODOs for all the places I think we should improve. It will be worked on in subsequent stages of this feature so there will be time to fix more of these. If anything looks below the code quality we want to merge, I am happy to make those changes in this PR.

~Please also review the 1 change PR #4875 if it is not already merged. It updates the test project by regenerating it with a ref. I split it out because it touches 2000 files.~

### Tasks
- [x] regenerate test project to be less wrong (https://github.com/dbt-labs/dbt-timing-project/pull/14)
- [x] manually model 1.0.latest and (#4764) commit baseline json file to main (this branch)
- [x] wrap up runner sampling code path
- [x] merge baseline file to 1.0.latest (#4847)
- [x] modify model action to accurately obtain version information from both manual triggers and release triggers. (tested in a personal sandbox repo [here](https://github.com/nathaniel-may/gha-sandbox/runs/5498358423?check_suite_focus=true)) 
- [x] polish README (remove adding projects and commands and replace with phase 2 ticket link)
- [ ] backport new README to 1.0.latest branch as well so it has the stale info warning. (to be done after pr is approved to include any readme changes)

### Checklist


.github/workflows/performance.yml
```diff
@@ -1,175 +1,151 @@
...
-      - uses: actions/download-artifact@v2
-        with:
-          name: runner
-      - name: change permissions
-        run: chmod +x ./runner
-      - name: run runner
-        run: ./runner measure -b baseline -p ${{ github.workspace }}/performance/projects/
-      - uses: actions/upload-artifact@v2
+      - name: Upload Results
+        uses: actions/upload-artifact@v2
```
jtcohen6, line 107: Sweet! Confirming that we'll have this data saved to / available from S3?

performance/README.md
```diff
@@ -1,18 +1,69 @@
...
+| 2 σ | 1 in 44        |                         |
+| 3 σ | 1 in 741       |  evidence               |
+| 4 σ | 1 in 31,574    |                         |
+| 5 σ | 1 in 3,486,914 |  discovery              |
+
+When detecting performance regressions that trigger alerts, block PRs, or delay releases we want to be conservative enough that detections are infrequently triggered by noise, but not so conservative as to miss most actual regressions. This test suite uses a 3 sigma standard so that only about 1 in every 700 runs is expected to fail the performance regression test suite due to expected variance in our measurements.
+
+### Concrete Example
+
+The following example data was collected locally on a macbook pro using the same tools included in this repository.
```
jtcohen6, line 37: Wouldn't we want to collect the sample mean + stddev on the "same machine" as what will be running in CI? I know that's technically impossible, since GHA is a cloud service using VMs — but it's still possible to match the same basic architecture and memory characteristics, right? Versus a macbook pro running locally

jtcohen6 (commented):
This is really cool! The stats make sense to me, though still might be worth a sense-check with some of our better-informed colleagues. I appreciate how much more efficient this approach will be, enabling us to avoid re-taking the baseline measurement each time.

One question: Are we still only checking "full" parse (`dbt parse`)? How big of a lift would it be to add more? I think it's critically important that we start checking for regressions in a basket of representative commands. That's motivated by the performance regressions that have cropped up over the past few months, which eluded our narrow focus on project parsing _only_.

Proposed commands:
1. `dbt --no-partial-parse parse`: "full" parse
2. `dbt --partial-parse parse`: partial parse with no file changes, requires preserving `partial_parse.msgpack` from step 1
3. `dbt compile --exclude fqn:*`: includes graph compilation
4. `dbt build --exclude fqn:*`: includes graph compilation with extra test edges
5. `dbt compile --select node_0`: includes graph compilation + adapter cache construction

.github/workflows/performance.yml
nathaniel-may: not at first. the first version of this test suite is going to add a commit to the repository with a json file for the full modeling of each release. The samples are printed in github actions but ultimately not saved.

If we want any of the above to be available in s3 we can track that work with a separate ticket.

performance/README.md
nathaniel-may: you're totally correct. The macbook numbers are all I have right now so I used those to give concrete numbers to the abstract concepts. Your comment here tells me that I should replace these with github action numbers once I have them though.

nathaniel-may:
@jtcohen6 sorry for the delay!

Yes you're absolutely right that we _need_ more metrics for this to be useful. The reason I haven't added much else yet, is that `dbt parse` doesn't need a connected warehouse so it was the easiest to get off the ground with.

That being said, the next step is to fill out the tests with the commands you suggested. I've logged #4768 with all the next steps required to get that off the ground.

.github/workflows/model_performance.yml
```diff
@@ -0,0 +1,99 @@
+name: Model Performance Characteristics
+# Schedule triggers
+on:
+  # TODO this should hit after each release.
+  # this is just to make the first few baselines manually
+  pull_request:
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
+
```
leahwicz, line 9: We should scope down the permissions for the token here
https://github.blog/changelog/2021-04-20-github-actions-control-permissions-for-github_token/

.github/workflows/model_performance.yml
```diff
@@ -0,0 +1,99 @@
+name: Model Performance Characteristics
```
leahwicz, line 1: Nit: for the Actions, we have been doing the "what", "why", "when" comment sections so lets stick to that format

nathaniel-may: pushed first draft of these comments.

.github/workflows/model_performance.yml
nathaniel-may: this will eventually commit to the repository. Does that mean we'll need `action: write` and that's it?

.github/workflows/model_performance.yml
```diff
@@ -0,0 +1,168 @@
...
+      commit_results:
+        description: commit results directly to the above branch when finished? (true/false)
+        default: false
+
+env:
+  RUNNER_CACHE_PATH: performance/runner/target/release/runner
+
+# both jobs need to write
+permissions:
+  content: write
```
leahwicz, line 38:
```suggestion
  contents: write
```

.github/workflows/model_performance.yml
```diff
@@ -0,0 +1,168 @@
...
+  # run the performance measurements on the current or default branch
+  # TODO get the version from pypi instead
+  model:
+    # depends on `latest-runner` as a separate job so that failures in this job do not prevent
+    # a successfully tested and built binary from being cached.
+    needs: [latest-runner]
+    name: Model a release
+    runs-on: ubuntu-latest
+    steps:
+      steps:
```
leahwicz, line 112: Remove extra `steps`

performance/README.md
```diff
@@ -1,18 +1,101 @@
...
+- A GitHub action for modeling the performance distribution for a new release: `/.github/workflows/model_performance.yml`.
+- A GitHub action for sampling performance of dbt at your commit and comparing it against a previous release: `/.github/workflows/sample_performance.yml`.
+
+## Adding Test Scenarios
+
+A clear process for maintainers and community members to add new performance testing targets will exist after the next stage of the test suite is complete. For details, see #4768.
+
+## Investigating Regressions
+
+If your commit has failed one of the performance regression tests, it does not necessarily mean your commit has a performance regression. However, the observed runtime value was so much slower than the expected value that it was unlikely to be random noise. This means that any commit after the release it is being compared against through this failing commit might contain the cause. Start by investigating the failing commit and working your way backwards.
```
leahwicz, line 27: > This means that any commit after the release it is being compared against through this failing commit might contain the cause. 

This is a confusing statement to me. I know what you mean b/c we have talked but can you clarify here?

performance/runner/src/calculate.rs
```diff
@@ -1,278 +1,169 @@
...
-                    }
-                })
-        })
-        .collect::<Result<Vec<MeasurementGroup>, IOError>>()
-        .or_else(|e| Err(CalculateError::CalculateIOError(e)))?;
+        .map(|(_, x)| x)
+        .collect();
+
+    // check that we have at least one baseline
+    if baselines.is_empty() {
```
leahwicz, line 48: If this is a new minor version that we just cut a new release branch for, what will happen in this instance?

leahwicz: And by what will happen, I mean will we be able to find a baseline?

performance/runner/src/calculate.rs
```diff
@@ -1,278 +1,169 @@
...
-            system: 1.00,
-            min: 1.00,
-            max: 1.00,
-            times: vec![],
+        let baseline = Baseline {
+            version: Version::new(9, 9, 9),
+            metric: metric.clone(),
+            ts: Utc::now(),
+            measurement: Measurement {
+                command: "some command".to_owned(),
```
leahwicz, line 91: `to_owned` - what does this do? This is more of a curiosity question lol

performance/runner/src/types.rs
```diff
@@ -0,0 +1,218 @@
...
+#[derive(
+    Debug, Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Hash, DeserializeFromStr, SerializeDisplay,
+)]
+pub struct Version {
+    pub major: i32,
+    pub minor: i32,
+    pub patch: i32,
+}
+
+impl FromStr for Version {
```
leahwicz, line 85: As a validation check, do you get in the string `1.1.0` and not `v1.1.0` since I know that `v` is appended in some places like the tag

performance/runner/src/types.rs
```diff
@@ -0,0 +1,218 @@
...
+        let filestem = path.file_stem().map_or_else(
+            || Err(IOError::BadFilestemError(path.clone())),
+            |stem| Ok(stem.to_string_lossy().to_string()),
+        )?;
+
+        let metric = Metric::from_str(&filestem)?;
+
+        // TODO use result values not panics
+        match &measurement.times[..] {
+            [] => panic!("found a sample with no measurement"),
```
leahwicz, line 158: `panic!` - haha I like this as a coding concept

leahwicz: Not sure how it differs from throwing an exception

leahwicz:
Overall this LGTM! 

Two thing that I would feel like would be good to add to the README is: 
* What happens/example of when a release is completed and now we want to set a baseline for that release- what runs, on what branch, what gets committed, etc.
* What happens when we test a commit- you mention this in the Concrete Example but it doesn't mention scenario details such as: I commit to `main` vs I'm backporting to `1.0.latest`, if I'm on `1.0.latest` then I'm being compared to the latest final release or RC? Basically the decision tree of what is the baseline that I'm getting compared to

leahwicz (approved):

performance/runner/src/calculate.rs
nathaniel-may: I'll add a comment to this effect, but if we branch off of main, there will be some old baselines in main. This check was mostly useful for initial development so I'll add that it can be modified once it's all merged and running.

performance/runner/src/calculate.rs
nathaniel-may: Literal strings in your Rust source code are references that live for the lifetime of your program (they can always be fetched from the binary). Because of all the lovely guarantees the rust typechecker makes for you around references, sometimes its necessary or just cleaner to own the string instead of borrowing it. `to_owned()` allocates the same sequence of utf-8 bytes on the heap behind a different pointer and lets you pass it around independently.

In this case when we make a value of type `Measurement` if I didn't use `to_owned` I would have to explicitly handle reference lifetimes in the code which is unnecessary complication for what we're doing here. Instead I just made a copy of the string and owned it to avoid that.

nathaniel-may: This is pretty much the same thing every other language does, but you don't have the control to choose to use a pointer instead of a copy. In this case it probably looks like like unnecessary complication, but when you're doing anything more complicated than passing a string value it's a real life saver.

performance/runner/src/fs.rs
```diff
@@ -0,0 +1,349 @@
...
+use chrono::prelude::*;
+use serde::de::DeserializeOwned;
+use std::fs::DirEntry;
+use std::io;
+use std::path::{Path, PathBuf};
+use std::process::{Command, ExitStatus};
+use std::str::FromStr;
+use std::{cmp, fs};
+
+// TODO these should not be defined here anymore. they need to be split at the github action level.
```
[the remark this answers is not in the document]
nathaniel-may: It's a bullet in #4768

performance/runner/src/types.rs
nathaniel-may: yeah I handle the v in the action right now. But I could move that logic in here instead to handle either. I get the non-v parsing for free from my type definition but if we include both the v and no v variants I would have to write it myself which isn't very hard. Might be better to just seamlessly handle both.

kwigley (approved):

performance/runner/src/fs.rs
```diff
@@ -0,0 +1,349 @@
...
+        .arg("--show-output")
+        .status() // use spawn() here instead for more information
+        .or_else(|e| Err(IOError::CommandErr(Some(e))))
+}
+
+// Attempt to delete the directory and its contents. If it doesn't exist we'll just recreate it anyway.
+fn clear_dir(dir: &dyn AsRef<Path>) -> Result<(), io::Error> {
+    match fs::remove_dir_all(dir) {
+        // whether it existed or not, create the directory.
+        _ => fs::create_dir(dir),
```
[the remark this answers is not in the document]
nathaniel-may: what do you mean by parent directory? this should just be deleting a directory and all of its contents (child directories included) then recreating it empty.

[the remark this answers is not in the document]
kwigley: ah ya, poor wording on my part

`create_dir("parent/child")` -> `mkdir parent/child` will fail if parent dir doesn't exist
`create_dir_all("parent/child")` -> `mkdir -p parent/child`

not an issue here if `dir` here is just 1 level ever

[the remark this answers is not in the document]
kwigley: no preference on match vs combinator

performance/runner/src/calculate.rs
leahwicz: Oh interesting. I see. `to_owned` is kinda a funny name but that making the copy instead of passing the same pointer around here makes sense

performance/runner/src/types.rs
leahwicz: Either way works just as long as we are explicit about what we are expecting

performance/runner/src/fs.rs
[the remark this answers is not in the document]
nathaniel-may: right. if users give me a dud path my only option is to give a clear error. which I think is what you're pointing out here: a bad path doesn't have a great error story here. I'll improve that.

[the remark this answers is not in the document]
nathaniel-may: 1- the `and_then` syntax isn't semantically equivalent. oops.
2- I changed it to create any missing parents as well instead of an error. equivalent of `mkdir -p`

performance/runner/src/types.rs
nathaniel-may: added a TODO noting this.

performance/README.md
nathaniel-may: revised.

PR #4607, opened 2022-01-21
update cache action

kwigley:
update performance regression action to using cached runner

### DIFF
--- a/.github/workflows/performance.yml
+++ b/.github/workflows/performance.yml
@@ -132,25 +132,23 @@ jobs:
     name: Measure Baseline Branch
     runs-on: ubuntu-latest
     steps:
-      - name: checkout latest
-        uses: actions/checkout@v2
-        with:
-          ref: "0.20.latest"
       - name: Setup Python
         uses: actions/setup-python@v2.2.2
         with:
           python-version: "3.8"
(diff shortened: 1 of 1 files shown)


.github/workflows/performance.yml
```diff
@@ -30,8 +32,8 @@ jobs:
       - uses: actions/cache@v2
         id: cache
         with:
-          path: performance/runner/target/release/runner
-          key: ${{ runner.os }}-${{ hashFiles('performance/runner/Cargo.toml') }}-${{ hashFiles('performance/runner/src/*') }}
+          path: ${{ env.RUNNER_CACHE_PATH }}
```
nathaniel-may, line 35: I would have NEVER thought of this.

.github/workflows/performance.yml
```diff
@@ -30,8 +32,8 @@ jobs:
...
+          key: ${{ runner.os }}-${{ hashFiles('performance/runner/Cargo.toml', 'performance/runner/src/*') }}
```
nathaniel-may, line 36: and this is me just being a dumb dumb. Thanks 😄

nathaniel-may (approved):
Thanks!!

Issue #4577, opened 2022-01-17
[CT-49] [Bug] dbt deps does not handle/retry on a malformed response from hub


barberscott:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

If in the process of running `dbt deps` dbt receives a malformed response from `hub.getdbt.com` deps will fail:

```
Traceback (most recent call last):
  File "/usr/src/app/sinter/clients/dbt.py", line 1200, in call
    dbt_main.handle(command + extra_args)
  File "/usr/local/lib/python3.8/dist-packages/dbt/main.py", line 152, in handle
    res, success = handle_and_check(args)
  File "/usr/local/lib/python3.8/dist-packages/dbt/main.py", line 192, in handle_and_check
    task, res = run_from_args(parsed)
  File "/usr/local/lib/python3.8/dist-packages/dbt/main.py", line 246, in run_from_args
    results = task.run()
  File "/usr/local/lib/python3.8/dist-packages/dbt/task/deps.py", line 56, in run
    final_deps = resolve_packages(packages, self.config)
  File "/usr/local/lib/python3.8/dist-packages/dbt/deps/resolver.py", line 135, in resolve_packages
    target = final[package].resolved().fetch_metadata(config, renderer)
  File "/usr/local/lib/python3.8/dist-packages/dbt/deps/registry.py", line 121, in resolved
    self._check_in_index()
  File "/usr/local/lib/python3.8/dist-packages/dbt/deps/registry.py", line 92, in _check_in_index
    if self.package not in index:
TypeError: argument of type 'NoneType' is not iterable
```

while attempting to parse the JSON response which, presumably, is bad.

### Expected Behavior

- Log the bad JSON so that we know what hub sent
- Retry similar to how retry happens for failures to connect and for empty responses

emmyoop:
@barberscott thanks again for the detailed issue!  This is a result of a big of a bug associated with our retry logic.

We should be checking here if `resp.json()` is `None` not just the `resp` object.

https://github.com/dbt-labs/dbt-core/blob/7798f932a09a12f1993ea9cb781a20707383df96/core/dbt/clients/registry.py#L36-L39

If it is `None` we should retry.

This issue is part of our next release (1.0.2) and I'm actively working on it!  Similar to #4579 I'm going to need to think a bit more on what, if anything, should be logged in this case.  Possibly the entire response?

leahwicz:
Backport here: #4649

Issue #4601, opened 2022-01-20
[CT-77] [Bug] dbt deps should keep retrying if registry response is malformed (bug regression)


BrJan:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Running dbt deps in a nightly job failed after retrying downloading the packages. The Max retries is listed as 5 and was currently at 0.

### Expected Behavior

The download is retried until the max retries is reached or the package is downloaded successfully.

### Steps To Reproduce

This cannot be reproduced on demand. It is an intermittent failure case for dbt deps

### Relevant log output

```shell
2022-01-20 10:07:06.950387 (MainThread): 10:07:06  Making package registry request: GET https://hub.getdbt.com/api/v1/fivetran/fivetran_utils.json
2022-01-20 10:07:07.058992 (MainThread): 10:07:07  Response from registry: GET https://hub.getdbt.com/api/v1/fivetran/fivetran_utils.json 200
2022-01-20 10:07:07.059729 (MainThread): 10:07:07  Sending event: {'category': 'dbt', 'action': 'invocation', 'label': 'end', 'context': [<snowplow_tracker.self_describing_json.SelfDescribingJson object at 0x7efcdd10f580>, <snowplow_tracker.self_describing_json.SelfDescribingJson object at 0x7efcdd10f8b0>, <snowplow_tracker.self_describing_json.SelfDescribingJson object at 0x7efcdd10f4f0>]}
2022-01-20 10:07:07.060251 (MainThread): 10:07:07  Flushing usage events
2022-01-20 10:07:07.083491 (MainThread): argument of type 'NoneType' is not iterable

Datadog:
Traceback (most recent call last):
  File "/usr/src/app/sinter/clients/dbt.py", line 1200, in call
    dbt_main.handle(command + extra_args)
  File "/usr/local/lib/python3.8/dist-packages/dbt/main.py", line 159, in handle
    res, success = handle_and_check(args)
  File "/usr/local/lib/python3.8/dist-packages/dbt/main.py", line 205, in handle_and_check
    task, res = run_from_args(parsed)
  File "/usr/local/lib/python3.8/dist-packages/dbt/main.py", line 258, in run_from_args
    results = task.run()
  File "/usr/local/lib/python3.8/dist-packages/dbt/task/deps.py", line 53, in run
    final_deps = resolve_packages(packages, self.config)
[...]

PR #4609, opened 2022-01-21
Retry after failure to download or failure to open files

### LINKED ISSUE #4579: [CT-51] [Bug] dbt deps does not handle bad git tarball download

### Current Behavior

When downloading the tar file from github, it is possible for the tar file to be malformed or incomplete (the likely most common case is that the tar file in the local filesystem actually contains an error message instead of it being a real tar file) and when this happens deps will throw the following error:

```
  File "/usr/src/app/sinter/clients/dbt.py", line 1200, in call
    dbt_main.handle(command + extra_args)
  File "/usr/local/lib/python3.8/dist-packages/dbt/main.py", line 159, in handle
    res, success = handle_and_check(args)
  File "/usr/local/lib/python3.8/dist-packages/dbt/main.py", line 205, in handle_and_check
    task, res = run_from_args(parsed)
  File "/usr/local/lib/python3.8/dist-packages/dbt/main.py", line 258, in run_from_args
    results = task.run()
  File "/usr/local/lib/python3.8/dist-packages/dbt/task/deps.py", line 66, in run
    package.install(self.config, renderer)
  File "/usr/local/lib/python3.8/dist-packages/dbt/deps/registry.py", line 74, in install
    system.untar_package(tar_path, deps_path, package_name)
  File "/usr/local/lib/python3.8/dist-packages/dbt/clients/system.py", line 489, in untar_package
    with tarfile.open(tar_path, 'r') as tarball:
  File "/usr/lib/python3.8/tarfile.py", line 1608, in open
    raise ReadError("file could not be opened successfully")
tarfile.ReadError: file could not be opened successfully
```

This typically is indicative of a transient problem with github itself but in the context of dbt Cloud will cause a run to fail when a retry could possibly succeed.  

### Expected Behavior

- Identify the bad tar file and retry the download, perhaps by checking for a successful list (tar -tzf foo.tar.gz >/dev/null or something similar) or, rather than spend 2x the effort (check, then untar) simply handle the failure to untar.

- Optionally log a certain number of trailing bytes/characters in the case where the tar file is bad so we can understand, when this happens, whether it's just a truncated tar (download is being cut off) or something like an error message in a json response being pumped into the tar file since, I think, in the download iterating over what github is giving us and dropping it into a .tar.gz file.

### Steps To Reproduce

_No response_

### Relevant log output

_No response_

### Environment

_No response_

_No response_

### Additional Context

_No response_

emmyoop:
resolves #4579, #4577

### Description

- Add retry logic when there's a problem extracting the tarball.
- Fix retry logic when a response is `NoneType`.
- Add relevant unit tests.

### Checklist

### DIFF
--- a/core/dbt/clients/registry.py
+++ b/core/dbt/clients/registry.py
@@ -33,6 +33,11 @@ def _get(path, registry_base_url=None):
     resp = requests.get(url, timeout=30)
     fire_event(RegistryProgressGETResponse(url=url, resp_code=resp.status_code))
     resp.raise_for_status()
+
+    # It is unexpected for the content of the response to be None so if it is, raising this error
+    # will cause this function to retry (if called within _get_with_retries) and hopefully get
+    # a response.  This seems to happen when there's an issue with the Hub.
+    # See https://github.com/dbt-labs/dbt-core/issues/4577
     if resp.json() is None:
         raise requests.exceptions.ContentDecodingError(
             'Request error: The response is None', response=resp
--- a/core/dbt/clients/registry.py
+++ b/core/dbt/clients/registry.py
@@ -33,7 +33,7 @@ def _get(path, registry_base_url=None):
     resp = requests.get(url, timeout=30)
     fire_event(RegistryProgressGETResponse(url=url, resp_code=resp.status_code))
     resp.raise_for_status()
-    if resp is None:
+    if resp.json() is None:
         raise requests.exceptions.ContentDecodingError(
             'Request error: The response is None', response=resp
         )
--- a/core/dbt/clients/system.py
+++ b/core/dbt/clients/system.py
@@ -485,7 +485,7 @@ def untar_package(
 ) -> None:
     tar_path = convert_path(tar_path)
     tar_dir_name = None
-    with tarfile.open(tar_path, 'r') as tarball:
+    with tarfile.open(tar_path, 'r:gz') as tarball:
         tarball.extractall(dest_dir)
         tar_dir_name = os.path.commonprefix(tarball.getnames())
     if rename_to:
--- a/core/dbt/deps/registry.py
+++ b/core/dbt/deps/registry.py
@@ -1,4 +1,5 @@
 import os
+import functools
 from typing import List
 
 from dbt import semver
@@ -14,6 +15,7 @@ from dbt.exceptions import (
     DependencyException,
     package_not_found,
 )
+from dbt.utils import _connection_exception_retry as connection_exception_retry
 
 
 class RegistryPackageMixin:
@@ -68,9 +70,28 @@ class RegistryPinnedPackage(RegistryPackageMixin, PinnedPackage):
         system.make_directory(os.path.dirname(tar_path))
 
         download_url = metadata.downloads.tarball
-        system.download_with_retries(download_url, tar_path)
         deps_path = project.packages_install_path
         package_name = self.get_project_name(project, renderer)
+
+        download_untar_fn = functools.partial(
+            self.download_and_untar,
+            download_url,
+            tar_path,
+            deps_path,
+            package_name
+        )
+        connection_exception_retry(download_untar_fn, 5)
+
+    def download_and_untar(self, download_url, tar_path, deps_path, package_name):
+        """
+        Sometimes the download of the files fails and we want to retry.  Sometimes the
+        download appears successful but the file did not make it through as expected
+        (generally due to a github incident).  Either way we want to retry downloading
+        and untarring to see if we can get a success.  Call this within
+        `_connection_exception_retry`
+        """
+
+        system.download(download_url, tar_path)
         system.untar_package(tar_path, deps_path, package_name)
 
 
--- a/core/dbt/exceptions.py
+++ b/core/dbt/exceptions.py
@@ -774,6 +774,10 @@ def system_error(operation_name):
 
 
 class ConnectionException(Exception):
+    """
+    There was a problem with the connection that returned a bad response,
+    timed out, or resulted in a file that is corrupt.
+    """
     pass
 
 
--- a/core/dbt/utils.py
+++ b/core/dbt/utils.py
@@ -601,7 +601,9 @@ class MultiDict(Mapping[str, Any]):
 
 def _connection_exception_retry(fn, max_attempts: int, attempt: int = 0):
     """Attempts to run a function that makes an external call, if the call fails
-    on a connection error, timeout or other exception, it will be tried up to 5 more times.
+    on a connection error, timeout or decompression issue, it will be tried up to 5 more times.
+    See https://github.com/dbt-labs/dbt-core/issues/4579 for context on this decompression issues
+    specifically.
     """
     try:
         return fn()
--- a/core/dbt/utils.py
+++ b/core/dbt/utils.py
@@ -10,6 +10,7 @@ import jinja2
 import json
 import os
 import requests
+from tarfile import ReadError
 import time
 from pathlib import PosixPath, WindowsPath
 
@@ -600,7 +601,7 @@ class MultiDict(Mapping[str, Any]):
 
 def _connection_exception_retry(fn, max_attempts: int, attempt: int = 0):
     """Attempts to run a function that makes an external call, if the call fails
-    on a connection error or timeout, it will be tried up to 5 more times.
+    on a connection error, timeout or other exception, it will be tried up to 5 more times.
     """
     try:
         return fn()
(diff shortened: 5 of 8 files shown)


core/dbt/deps/registry.py
```diff
@@ -68,9 +70,28 @@ def install(self, project, renderer):
...
         package_name = self.get_project_name(project, renderer)
+
+        download_untar_fn = functools.partial(
+            self.download_and_untar,
+            download_url,
+            tar_path,
+            deps_path,
+            package_name
+        )
+        connection_exception_retry(download_untar_fn, 5)
```
nathaniel-may, line 83: It might help with testing and minimizing retry logic to separate downloading and unpacking the tarball into two distinct steps. If the error is with unpacking, redoing the download would be unnecessary.

nathaniel-may (commented):
I know this is still in draft but I wanted to address your questions:

The part that's hard to test is the download. I'd separate that out so that you can test unpacking the tarball with an integration test that commits a tarball as part of the test. You can do one well-formed tarball and one mal-formed one to test both sides.

As for exception messaging, I might consider making this a longer one. Listing that this might be an issue with github and to check their status page, if it's a new release to consider logging a ticket with the package maintainers, and maybe even include steps or a link to steps that show how to manually install the dependency and to log a bug with the core team if a manual install worked.

core/dbt/deps/registry.py
emmyoop: @nathaniel-may we do possibly need to retry the download when unpacking the tarball fails.  We have data from the last github incident where only about half of the runs failed during the incident implying that some jobs were able to run `dbt deps` successfully.  It's impossible to know why but seems worth it to retry downloading the file in case something went wrong there.

That said, pulling them apart will make it easier to test so I'll incorporate both for the purpose of testing.

core/dbt/clients/system.py
```diff
@@ -485,7 +485,7 @@ def untar_package(
 ) -> None:
     tar_path = convert_path(tar_path)
     tar_dir_name = None
-    with tarfile.open(tar_path, 'r') as tarball:
+    with tarfile.open(tar_path, 'r:gz') as tarball:
```
emmyoop, line 488: The exception we were raising was in a section where the code was figuring out the compression type.  Since we know it's `gz`, set it and hopefully bubble up a better exception.

test/unit/test_system_client.py
```diff
@@ -185,3 +188,55 @@ def tearDown(self):
...
+            with NamedTemporaryFile(
+                prefix='a', suffix='.txt', dir=self.tempdir
+            ) as file_a:
+                file_a.write(b'some text in the text file')
+                relative_file_a = os.path.basename(file_a.name)
+                with tarfile.open(fileobj=named_tar_file, mode='w:gz') as tar:
+                    tar.addfile(tarfile.TarInfo(relative_file_a), open(file_a.name))
+
+        assert tarfile.is_tarfile(tar.name)
+        dbt.clients.system.untar_package(tar_file_full_path, self.tempdest)
```
nathaniel-may, line 220: I would at least comment here that THIS is the thing we're testing so that future failures are easier to debug.

test/unit/test_system_client.py
```diff
@@ -185,3 +188,55 @@ def tearDown(self):
...
+            ) as file_a:
+                file_a.write(b'some text in the text file')
+                txt_file_name = file_a.name
+                file_path= os.path.dirname(txt_file_name)
+                tar_file_path = os.path.join(file_path, 'mypackage.2.tar.gz')
+        os.rename(txt_file_name, tar_file_path)
+        with self.assertRaises(tarfile.ReadError) as exc:
+            dbt.clients.system.untar_package(tar_file_path, self.tempdest)
+
+    def test_untar_package_empty(self):
```
nathaniel-may, line 236: Let's comment all the tests with the situation they're testing, and the expected results that we're asserting.

test/unit/test_system_client.py
```diff
@@ -185,3 +188,55 @@ def tearDown(self):
             shutil.rmtree(self.base_dir)
         except:
             pass
+
+
+class TestUntarPackage(unittest.TestCase):
```
nathaniel-may, line 193: Love the strategies you've employed here. Making the tarballs in the tests is much cleaner than my previous suggestion of committing tarball files.

test/unit/test_core_dbt_utils.py
```diff
@@ -22,11 +23,21 @@ def test_connection_exception_retry_success(self):
...
+        Counter._reset()
+        connection_exception_retry(lambda: Counter._add_with_timeout(), 5)
+        self.assertEqual(2, counter) # 2 = original attempt plus 1 retry
+
     def test_connection_exception_retry_success_none_response(self):
         Counter._reset()
         connection_exception_retry(lambda: Counter._add_with_none_exception(), 5)
         self.assertEqual(2, counter) # 2 = original attempt returned None, plus 1 retry
 
+    def test_connection_exception_retry_success_failed_untar(self):
```
nathaniel-may, line 36: Love a good unit test.

core/dbt/utils.py
```diff
@@ -600,14 +601,15 @@ def __contains__(self, name) -> bool:
 
 def _connection_exception_retry(fn, max_attempts: int, attempt: int = 0):
     """Attempts to run a function that makes an external call, if the call fails
-    on a connection error or timeout, it will be tried up to 5 more times.
+    on a connection error, timeout or other exception, it will be tried up to 5 more times.
```
nathaniel-may, line 604: are "other exceptions" all related to decoding/reading? should this comment be more specific? It might be able to provide historical context by including a ticket number.

core/dbt/exceptions.py
```diff
@@ -774,6 +774,10 @@ def system_error(operation_name):
 
 
 class ConnectionException(Exception):
+    """
```
nathaniel-may, line 777: 🎉

core/dbt/deps/registry.py
```diff
@@ -68,9 +70,28 @@ def install(self, project, renderer):
...
+
+    def download_and_untar(self, download_url, tar_path, deps_path, package_name):
+        """
+        Sometimes the download of the files fails and we want to retry.  Sometimes the
+        download appears successful but the file did not make it through as expected
+        (generally due to a github incident).  Either way we want to retry downloading
+        and untarring to see if we can get a success.  Call this within
+        `_connection_exception_retry`
+        """
```
nathaniel-may, line 92: 🎉

core/dbt/clients/registry.py
```diff
@@ -33,7 +33,7 @@ def _get(path, registry_base_url=None):
     resp = requests.get(url, timeout=30)
     fire_event(RegistryProgressGETResponse(url=url, resp_code=resp.status_code))
     resp.raise_for_status()
-    if resp is None:
+    if resp.json() is None:
```
nathaniel-may, line 36: I think this is the only line I don't totally understand. Can you describe what's going on here?

emmyoop: You can get a response that is not `None` but `response.json()` is `None`.  This was originally reported as needed to be retried (why raising `ContentDecodingError` was added) by #4601.  #4577 was reported because the problem was never actually resolved because we were never falling into this `if` statement.  This change will force the retires as expected.

test/unit/test_system_client.py
```diff
@@ -185,3 +188,55 @@ def tearDown(self):
...
+        self.tempdest = mkdtemp(dir=self.base_dir)
+    
+    def tearDown(self):
+        try:
+            shutil.rmtree(self.base_dir)
+        except:
+            pass
+
+    def test_untar_package_success(self):
+        with NamedTemporaryFile(
```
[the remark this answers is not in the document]
emmyoop: They're all actually different so I'll leave them in the test but comment what's happening for clarity.

core/dbt/clients/registry.py
nathaniel-may: gotcha. So what does this `json()` method do?  what does it signify when it returns `None`

emmyoop: `resp` is a response object (attributes of content, headers, status_code, etc) and the `json()` method returns the json-encoded  _content_ of the response (or raises an error).  

In some cases it is a completely legitimate response.  You made a valid request but there's no data for what you requested.  We do not expect and empty response.  The content of the response should either not be None or we should get an error (and catch said error with `resp.raise_for_status()` on line 35). From what's been gathered it looks like this is happening when something is wrong with the hub.  It's honestly difficult to replicate.  Catching it and logging the full response will hopefully be helpful too.

nathaniel-may: ahh this makes sense. I might add a comment here about having no json encoded data and that being unexpected.

emmyoop: Good point.  Added.

gshank (approved):
Looks very clean, and nicely commented.

nathaniel-may (approved):
Great work!

PR #4610, opened 2022-01-22
[#4554] Don't require a profile for dbt deps and clean commands

### LINKED ISSUE #4554: [Bug] dbt deps

### Current Behavior

We have defined a profile like that :
```
    prod:
      type: redshift
      threads: "{{ env_var('DBT_THREADS', 1)| int }}"
      host: "{{ env_var('REDSHIFT_HOST') }}"
      port: 5439
      user: "{{ env_var('REDSHIFT_USER') }}"
      pass: "{{ env_var('REDSHIFT_PASSWORD') }}"
      dbname: "{{ env_var('REDSHIFT_DATABASE') }}"
      schema: "bi"
```
If one of the env is not defined  ( ex : REDSHIFT_DATABASE) we now have the following error message while running `dbt deps`
```
13:26:05  Encountered an error:
Parsing Error
  Env var required but not provided: 'REDSHIFT_DATABASE'
```
If it is defined with a dummy data, it works properly. ( So it does not check connection)

### Expected Behavior

`dbt deps`should not return any issue when a env variable is not defined.

### Steps To Reproduce

1. Define a profile with env variable 
2. Do not define the given env variable 
3. run `dbt deps`

### Relevant log output

_No response_

### Environment

```markdown
- OS:container python:3.9.6-slim-buster
- Python: 3.9.6
- dbt:1.0.0
```

redshift

### Additional Context

_No response_

gshank:
resolves #4554

### Description

Changing a profile parsing exception from a CompilationException to a ParsingException caused a failure in the code that intended to trap the error.

Switch to not loading the profile when we don't need it, and remove code that incorrectly reported the nature of the error.

### Checklist

### DIFF
--- a/test/integration/006_simple_dependency_tests/test_simple_dependency.py
+++ b/test/integration/006_simple_dependency_tests/test_simple_dependency.py
@@ -246,3 +246,48 @@ class TestSimpleDependencyNoProfile(TestSimpleDependency):
         with tempfile.TemporaryDirectory() as tmpdir:
             result = self.run_dbt(["clean", "--profiles-dir", tmpdir])
         return result
+
+class TestSimpleDependencyBadProfile(DBTIntegrationTest):
+
+    @property
+    def schema(self):
+        return "simple_dependency_006"
+
+    @property
+    def models(self):
+        return "models"
+
+    def postgres_profile(self):
+        # Need to set the environment variable here initially because
+        # the unittest setup does a load_config.
+        os.environ['PROFILE_TEST_HOST'] = self.database_host
+        return {
+            'config': {
+                'send_anonymous_usage_stats': False
+            },
+            'test': {
+                'outputs': {
+                    'default2': {
+                        'type': 'postgres',
+                        'threads': 4,
+                        'host': "{{ env_var('PROFILE_TEST_HOST') }}",
+                        'port': 5432,
+                        'user': 'root',
+                        'pass': 'password',
+                        'dbname': 'dbt',
+                        'schema': self.unique_schema()
+                    },
+                },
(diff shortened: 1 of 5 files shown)


gshank:
Prior to this pull request for the dbt deps and clean commands we would attempt to load the profile and if it failed would mis-report an error and then proceed with an UnsetProfile. The code would also set 'send_anonymous_usage_stats' to False for cases where the profile failed to load, which meant that for this edge case we had a different default for that attribute. Since we refactored the UserConfig/flags code, having an 'unset' user config no longer makes sense. It also doesn't make a lot of sense to go to extra work to validate profiles (and issue incorrect error messages) when the profile isn't required for the command.

This change simplifies the '_get_rendered_profile' code to use the UnsetProfile directly, and set the user_config separately.

test/integration/006_simple_dependency_tests/test_simple_dependency.py
```diff
@@ -246,3 +246,48 @@ def run_clean(self):
...
+
+    @use_profile('postgres')
+    def test_postgres_deps_bad_profile(self):
+        del os.environ['PROFILE_TEST_HOST']
+        self.run_dbt(["deps"])
+
+    @use_profile('postgres')
+    def test_postgres_clean_bad_profile(self):
+        del os.environ['PROFILE_TEST_HOST']
+        self.run_dbt(["clean"])
```
emmyoop, line 293: This test doesn't really make sense to have here since it's for `dbt clean` but since we're going to rearrange them all soon I'm not sure how much it matters.

emmyoop (approved):
One comment about a test location but otherwise looks good.

PR #4615, opened 2022-01-24
Changing Jira mirroring workflows to point to shared Actions

leahwicz:
### Description

I moved all Jira mirroring Actions into a repo that can be referenced by all our repos and reused so we don't need to worry about copying and pasting the same logic everywhere. We can just point each repo to look at this https://github.com/dbt-labs/jira-actions/tree/main/.github/workflows

### Checklist


nathaniel-may (approved):
Will there be any potential gaps we should manually check to make sure we didn't miss anything when we switch over to the new system?

ChenyuLInx (approved):

leahwicz:
> Will there be any potential gaps we should manually check to make sure we didn't miss anything when we switch over to the new system?

@nathaniel-may by gaps do you mean like a loss in functionality or new regression introduced?

nathaniel-may:
> do you mean like a loss in functionality or new regression introduced?

@leahwicz I mean the actions missed a ticket because it was submitted while the switch was happening.

leahwicz:
Oh gotcha @nathaniel-may! The creation and labeling will not miss a beat since the Action itself is only being edited so it will still trigger either with the old or new logic which does the same thing. The close workflow is currently broken in the repo so renaming it won't matter since it wasn't working to begin with 😞.

leahwicz:
@cla-bot check

PR #4616, opened 2022-01-24
Docker release CT-3

iknox-fa:
resolves #CT-3

### Description
Github actions to release a matrix of docker images to github packages and handle tagging/untagging of latest/ minor.latest versions.

### Checklist


nathaniel-may:
It's going to take a lot of additional context gathering for me to give a useful review on this PR. In the interest of time, I'm going to pull myself off as a reviewer but if you're in a pinch later you can add me back and I'll see what I can do.

.github/actions/latest-wrangler/README.md
```diff
@@ -0,0 +1,51 @@
...
+
+
+## Outputs
+| Output | Description |
+| - | - |
+| `latest` | Wether or not the new container should be tagged 'latest'|
+| `minor_latest` | Wether or not the new container should be tagged major.minor.latest |
+
+## Example workflow
+TODO
```
leahwicz, line 22: Not sure if this `TODO` is supposed to be here

.github/workflows/release_docker.yml
```diff
@@ -0,0 +1,114 @@
+# **what?**
+# This workflow will generate a series of docker images for dbt and push them to the github container registry
+
+# **why?**
+# Docker images for dbt are used in a number of important places throughout the dbt ecosystem.  This is how we keep those images up-to-date.
+
+# **when?**
+# This is triggered manually 
+
+# **TODO**
```
leahwicz, line 10: I'm not trying to be a pest pointing out the TODO and if they are for a follow up PR just close all these out lol

.github/workflows/release_docker.yml
```diff
@@ -0,0 +1,114 @@
...
+          echo "::set-output name=build_arg_value::"$(echo ${{ github.event.inputs.package }} | sed 's/postgres/core/g')
+
+      - name: Log in to the GHCR
+        uses: docker/login-action@v1
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Build and push PATCH
```
leahwicz, line 81: The name of this step is slightly confusing to me with the word "Patch". This is building and pushing the image with just with the given version number as the tag, correct? Basically what we would do if latest wasn't involved?

.github/workflows/release_docker.yml
```diff
@@ -0,0 +1,114 @@
...
+        uses: docker/build-push-action@v2
+        if: ${{ needs.get_version_meta.outputs.latest == 'True' }}
+        with:
+          file: docker/Dockerfile
+          push: True
+          target: ${{ github.event.inputs.package }}
+          build-args: |
+            ${{ steps.build_arg.outputs.build_arg_name }}_ref=${{ steps.build_arg.outputs.build_arg_value }}@v${{ github.event.inputs.version_number }}
+          tags: |
+            ghcr.io/dbt-labs/${{ github.event.inputs.package }}:latest
```
leahwicz, line 114: These last 3 steps all build the same image but then push it with a different tag, correct? Does the `tags` field take multiple tags? Can we build and push once and figure out the tags to use in a previous step?

.github/actions/latest-wrangler/README.md
```diff
@@ -0,0 +1,51 @@
+# Github package 'latest' tag wrangler for containers
+## Usage
+
+Plug in the necessary inputs to determine if the container being built should be tagged 'latest; at the package level, for example `dbt-redshift:latest`.
```
leahwicz, line 4: Could you add some more about `latest` vs `minor_latest`? I want to make sure we capture why we need both tags and not just 1 (I could see this coming up as a question from a new hire or somebody outside of the team)

.github/actions/latest-wrangler/README.md
iknox-fa: Whoops.  Good catch

.github/actions/latest-wrangler/main.py
```diff
@@ -0,0 +1,97 @@
...
+    if package_request.status_code == 404:
+        if halt_on_missing:
+            sys.exit(1)
+        else:
+            # everything is the latest if the package doesn't exist
+            print(f"::set-output name=latest::{True}")
+            print(f"::set-output name=minor_latest::{True}")
+            sys.exit(0)
+
+    # TODO: verify package meta is "correct"
```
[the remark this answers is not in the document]
iknox-fa: This one is not a leftover.  I think it would be wise to do a more formal validation of the state of the packages repo given we have certain expectations and if they aren't met it will break.  That said, I forgot to take out the todo and write a ticket as a follow up.  [Done now](https://github.com/dbt-labs/dbt-core/issues/4640) and removed from the code.

.github/workflows/release_docker.yml
iknox-fa: They are all for a follow up PR and should be considered nice-to-haves.  I've updated the wordage so that's a bit more clear.

.github/workflows/release_docker.yml
iknox-fa: Yeah.  I'll make this more clear.  I refer to it as pushing the PATCH image because it's tag will be the full semver all the way down to the PATCH level.
`MAJOR.MINOR.PATCH`

.github/workflows/release_docker.yml
iknox-fa: It does and we could, but docker is smart enough not to push or build anything that it already has at the given location so the extraneous build/pushes are essentially no-ops as long as they are contained in the same job.

.github/actions/latest-wrangler/README.md
iknox-fa: TBH, I'm not 100% sure why we _need_ the `minor_latest` beyond the idea that we might have breaking changes between a `1.0.latest` and a `1.1.latest`?  @jerco would be a better person to answer that.

.github/workflows/release_docker.yml
leahwicz: Ohhhh ok good to know. I thought we were building again every time and that is what bothered me so 👍 with keeping it this way

leahwicz (approved):
This all LGTM so thanks for incorporating the feedback. I'm going to approve but let's make sure @jtcohen6 addresses the last open comment. You can merge and always open another PR to clean-up but I just don't want to lose that discussion if we do

.github/actions/latest-wrangler/README.md
jtcohen6: Sorry I missed this comment a few days ago! Rationale is to provide consistency with how we're thinking about patch + minor upgrades elsewhere (e.g. in dbt Cloud).

There are two premises after v1.0:
- Use patches as soon as they're available
- Minor releases won't have breaking changes (we've committed to this!), but they will touch significantly more code than patch releases

As such, we should let users decide when they want to perform the minor version upgrade. Each minor version will be officially supported for 12 months after its initial release.

So I wouldn't be surprised if some users want to stick with `1.0.latest`, for a few weeks or a month, even after the `1.1.0` release. The inclusion of a `1.0.latest` tag will give them the opportunity to coordinate upgrades at a time of their choosing. During that period, if we have to cut a "security" patch for v1.0 (security bug, dependency/installation issue), they'll still get it automatically, as they should.

.github/actions/latest-wrangler/action.yml
```diff
@@ -0,0 +1,20 @@
...
+    description: "Semver of the container being built (I.E. 1.0.4)"
+    required: true
+  gh_token: 
+    description: "Auth token for github (must have view packages scope)"
+    required: true
+outputs:
+  latest:
+    description: "Wether or not built container should be tagged latest (bool)"
+  minor_latest:
+    description: "Wether or not built container should be tagged minor.latest (bool)"
```
kwigley, line 17: General question.. I've never known what to use as return types for GHA & scripting things. This returns `True` | `False`, but does it make sense to return 0 | 1 or other bool representations?

kwigley (approved):
Just general question, looks good! I know you're brewing on release thoughts, looking forward to hearing what you're thinking about

.github/actions/latest-wrangler/action.yml
```diff
@@ -0,0 +1,20 @@
+name: "Github package 'latest' tag wrangler for containers"
```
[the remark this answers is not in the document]
iknox-fa: Ohhh.. good call.

.github/actions/latest-wrangler/main.py
```diff
@@ -0,0 +1,98 @@
...
+    package = os.environ["INPUT_PACKAGE"]
+    new_version = parse(os.environ["INPUT_NEW_VERSION"])
+    gh_token = os.environ["INPUT_GH_TOKEN"]
+    halt_on_missing = strtobool(os.environ.get("INPUT_HALT_ON_MISSING", "False"))
+
+    # get package metadata from github
+    package_request = requests.get(
+        f"https://api.github.com/orgs/dbt-labs/packages/container/{package}/versions",
+        auth=("", gh_token),
+    )
```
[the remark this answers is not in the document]
iknox-fa: I'd rather leave docker dependent only on what docker images already exist, but I hold that opinion very loosely.  In the long term I'd actually like to use GH releases as the source of truth for this, but it won't make sense to do that until we have a more complete release system (as opposed to this work which was meant to only address how Docker gets released)

.github/actions/latest-wrangler/action.yml
iknox-fa: Ok, so this is something that took a while to sort out, but the TL;DR is that it doesn't matter.  GHA is all written in javascript, which types _everything_ dynamically so any variable you set  as an output operates in the same way.  

I've defaulted to using string representations by default because all values [coming from bash are untyped](https://tldp.org/LDP/abs/html/untyped.html) and treated as strings.  It's one less thing to have to remember. :)

PR #4618, opened 2022-01-24
[#2479] Allow unique_id to take a list

### LINKED ISSUE #2479: [CT-96] Allow unique_key for incremental materializations to take a list
### Describe the feature
Right now when creating a model in dbt with materialization set to `incremental` you can pass in a single column to `unique key` that will act as the `key` for merging.

In an ideal world you would be able to pass in multiple columns as there are many cases where a table has more than one column that defines it's primary key.

The simplest solution would be to change to `unique_key` to take in a list (in addition to a string for backwards compatibility) and create the `predicates` for the merge based on the list vs. just the single column.

This might not be ideal as the param name `unique_key` implies a single key. Alternatives would be adding a new optional parameter `unique_key_list` or `unique_keys` that always take a list and eventually deprecate the `unique_key` parameter.

Not necessarily other alternatives but another thing to consider is the use of `unique_key` throughout the dbt project. It would stand to reason that whatever change is made here would apply to all other usages of `unique_key`. This can be done in one large roll-out or in stages such as with merges first, then upserts, then snapshots, etc. 

### Additional context
This feature should work across all databases.

### Who will this benefit?
Hopefully most dbt users. Currently the only workaround for this is using dbt_utils.surrogate_key which a) doesn't work for BigQuery and b) should ideally be an out of the box dbt feature.

gshank:
resolves #2479

### Description

This derives from pull request #4159.

### Checklist


jtcohen6 (commented):
This works with both Snowflake + BigQuery when I run locally. As a follow-up to merging this PR, we should look to add a simple integration tests to both of those plugins:
```sql
{{ config(
  materialized = 'incremental',
  unique_key = ['id1', 'id2']
)}}

select 1 as id1, 2 as id2
```

I think first step in getting the tests to pass should be rebasing against / merging in `main`. When I run locally, I see: `Running with dbt=0.21.0-rc1`.

core/dbt/include/global_project/macros/materializations/models/incremental/merge.sql
```diff
@@ -9,10 +9,19 @@
     {%- set sql_header = config.get('sql_header', none) -%}
 
     {% if unique_key %}
-        {% set unique_key_match %}
-            DBT_INTERNAL_SOURCE.{{ unique_key }} = DBT_INTERNAL_DEST.{{ unique_key }}
-        {% endset %}
-        {% do predicates.append(unique_key_match) %}
+        {% if unique_key is sequence and unique_key is not mapping and unique_key is not string %}
```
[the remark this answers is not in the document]
leahwicz: For the Spark change here, @gshank maybe you can pair with @McKnight-42 or @VersusFacit on making that change. 

@McKnight-42 @VersusFacit - this behavior difference is something good to document in the specific adapter pages if it isn't there yet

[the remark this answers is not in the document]
jtcohen6: I think we'll also want integration tests for Snowflake + BigQuery, to ensure this change actually achieves the desired result

gshank:
I'm not quite sure what the flow should be here. Should this PR be merged first and then the separate adapter changes be addressed?

leahwicz:
@gshank I'm not opposed to doing those changes in a separate PR as long as this change doesn't break anything w/o those other changes. If we do a separate PR, can you please open a tracking issue so we don't lose track of doing the work?

gshank:
@leah I think they would have to be separate PRs, because they're different repositories, right? You can't put changes from multiple repos into the same PR. I don't know if this change would break anything in the other repos, and I'm not sure how to test them against a dbt-core branch.

leahwicz:
> they're different repositories, right?

@gshank that is correct but I just didn't know if we needed to coordinate with the other repos. I don't want to merge this change if it causes the other repos to start failing tests and instead coordinate to merge all the PR's the same day if that makes sense.

McKnight-42 (approved):
Getting some context, and reading up LGTM

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ### Features
 - New Dockerfile to support specific db adapters and platforms.  See docker/README.md for details ([#4495](https://github.com/dbt-labs/dbt-core/issues/4495), [#4487](https://github.com/dbt-labs/dbt-core/pull/4487))
+- Allow unique_key to take a list ([#2479](https://github.com/dbt-labs/dbt-core/issues/2479), [#4618](https://github.com/dbt-labs/dbt-core/pull/4618))
```
jtcohen6, line 5: Let's make sure to add the contributor from the original PR to the changelog for v1.1:
```
- [@triedandtested-dev](https://github.com/triedandtested-dev) ([#4618](https://github.com/dbt-labs/dbt-core/pull/4618))
```

cdabel: By the way, I noticed that this list option is missing from the docs here:
https://docs.getdbt.com/reference/resource-configs/unique_key#use-a-combination-of-two-columns-as-a-unique-key

cdabel: By the way, I noticed that this list option is missing from the docs here:
https://docs.getdbt.com/reference/resource-configs/unique_key#use-a-combination-of-two-columns-as-a-unique-key

HeddeCrisp:
I encounter an error when running an incremental bigquery model with multiple unique keys. The error I am getting is:

`10:24:38  Database Error in model sample_model (models/ODL_derived/sample_model.sql)
10:24:38    Syntax error: Unexpected "[" at [546:33]. If this is a table identifier, escape the name with `, e.g. `table.name` rather than [table.name].
10:24:38    compiled Code at target/run/sample_dwh/models/ODL_derived/sample_model.sql` 

This is caused by the compiled code:

`[...] ) as DBT_INTERNAL_SOURCE
        on DBT_INTERNAL_SOURCE.['col1', 'col2', 'col3'] = DBT_INTERNAL_DEST.['col1', 'col2', 'col3']`

Config used is:

`      , name="sample_model"
        , materialized = 'incremental'
        , unique_key=['col1', 'col2', 'col3']
`

Am I doing something wrong or is this dbt code that fails with bigquery with tese settings?

VersusFacit:
Heya @HeddeCrisp ! It looks like you have the right syntax. Your error is familiar and strikes me as what used to happen before we merged this.

I did a little incremental model test with the following config:
```sql
{{
    config(
        materialized='incremental',
	unique_key=['id', 'first_name', 'last_name']
    )
}}
```

And my dbt version is:
```
Core:
  - installed: 1.4.5
  - latest:    1.4.5 - Up to date!

Plugins:
  - bigquery: 1.4.3 - Up to date!
```
 
So this to me implies there's not an error on our end with this feature but rather you might be using a dbt bigquery version that didn't have this feature active yet. What does `dbt --version` yield on your system?

nguyenann13:
Receiving an error when I pass a list
Parsing Error
  at path ['unique_key']: ['usr_key', 'usr_mbr_key', 'prgm_key', 'prmsn_key', 'resp_key', 'mbr_cmpy_key', 'usr_prgm_prmsn_key'] is not valid under any of the given schemas

{% snapshot snap_mkp_usr_prgm_prmsn %}

    {{
        config(
          target_schema='edl_workdb',
          strategy='timestamp',
          unique_key=['usr_key', 'usr_mbr_key', 'prgm_key', 'prmsn_key', 'resp_key', 'mbr_cmpy_key', 'usr_prgm_prmsn_key'],
          updated_at='upd_ts',
          invalidate_hard_deletes=True,
        )
    }}

    -- Filter out value analysis app as it's sunset and has duplicates
    select * from {{ source('hzn_mkp', 'mkp_usr_prmsn_app_t') }} where app_key <> 5

{% endsnapshot %}

CHANGELOG.md
dbeatty10: Good call @cdabel !

Just opened an issue for that here: https://github.com/dbt-labs/docs.getdbt.com/issues/4642

PR #4619, opened 2022-01-24
Guard against unnecessarily calling dump_graph in logging

### LINKED ISSUE #4569: [CT-39] Bug: ensure that dump_graph is not run for logging events in cachy.py unless cache logging is turned on

nathaniel-may:
resolves #4569

### Description

- Adds a new `Lazy` data type that can be reused anywhere in the code base (code quality passes `mypy --strict` without any ignores)
- Includes mashumaro serialization
- Has the offending cache events use the lazy type
- Tests that the existing `fire_event` function will work with this new lazy construction
- Tests that all relevant subclasses of `Cache` apply this lazy evaluation strategy to catch similar issues in the future

### The Challenge
Mashumaro does not support composing serialization strategies which makes serializing generic types, like `Lazy` tedious and error prone. Even if the wrapped type `T` has a mashumaro serialization strategy, it must be explicitly called by the serializer for `Lazy` itself. To be explicit, this means we would have to manually code mashumaro serialization strategies for all combinations of types that Lazy wraps in our code base e.g. `Lazy[str]`, `Lazy[int]`, `Lazy[Dict[str, List[str]]]`.

I have attempted to dig into Mashumaro's metaprogramming library to see if I can write a truly generic serialization for the `Lazy` type, but I haven't been able to figure it out.

### Questions
- We a few paths forward on this PR:
    1. Use this `Lazy` type to more easily communicate intention, and get good mypy type checking but introduce room for runtime errors when we forget to add a serialization strategy when it's going to be serialized like in the events module.
    2. Ditch `Lazy` so we don't have this gap to introduce new errors in serialization and instead pass unapplied methods with `# type: ignore` because mypy doesn't handle them well.
    3. Spend more time looking into mashumaro metaprogramming to potentially get the best of both worlds.

### Suggested Review Flow
1. Look at the two test cases. They should very clearly show the expected behavior. 
2. Read the implementation for lazy behavior which is found in `dbt.lazy.py::Lazy`.
3. Find the call sites for the updated cache events to see how it all fits together from beginning to end.

### Checklist


test/unit/test_events.py
```diff
@@ -5,18 +5,20 @@
 from dbt.events.functions import event_to_serializable_dict
 from dbt.events.types import *
 from dbt.events.test_types import *
-from dbt.events.base_types import Event
+from dbt.events.base_types import Event, InfoLevel
 from dbt.events.stubs import _CachedRelation, BaseRelation, _ReferenceKey, ParsedModelNode
 from importlib import reload
 import dbt.events.functions as event_funcs
 import dbt.flags as flags
+from dbt.lazy import Lazy
```
nathaniel-may, line 13: I'm getting a `ModuleNotFoundError: No module named 'dbt.lazy'` error when I run my tests. I found this pattern in other unit tests, what am I doing wrong here?

McKnight-42: what tests have dbt.lazy? if i do a search in finder I find no dbt.lazy in core

nathaniel-may: local virtual environment issue.

core/dbt/events/base_types.py
```diff
@@ -107,6 +107,11 @@ def asdict(cls, data: list) -> dict:
             # stringify all exceptions
             if isinstance(v, Exception) or isinstance(v, BaseException):
                 d[k] = str(v)
+            # use memoized version of all lazy values
+            elif k == '_f' and isinstance(v, Callable):  # type: ignore[arg-type]
+                d[k] = v()
+            elif k == 'memo' and v is None:
+                continue
```
nathaniel-may, line 114: I'm not happy about this. I'd much rather have a check for `type(v) == Lazy` but because of the asdict factories are only ever applied to _leaf values_ in a given nested dataclass I'm never given that opportunity. Just yet another reason why we need to swap out this serialization strategy.

emmyoop (commented):
This functionality seems worthwhile to mention in the `README`.

test/unit/test_events.py
```diff
@@ -429,4 +433,149 @@ def test_all_serializable(self):
...
+            else:
+                pass
+
+        # other exceptions are issues and should be thrown
+        except Exception as e:
+            raise e
+
+    # we should have at least 4 matching classes (raise this threshold if we add more)
+    expected = 4
+    x.assertTrue(len(matching_classes) >= expected, f"\nexpected at least {expected} matching classes.\nfound {len(matching_classes)}: {matching_classes}")
```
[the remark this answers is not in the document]
nathaniel-may: done

nathaniel-may:
Good call on the README, @emmyoop. I've added a new logging section now.

emmyoop (approved):

Fatal1ty:
Hi @nathaniel-may 

As the author of mashumaro, I would be happy to help you sort out the problem if you would explain what you want to achieve. If it's not difficult, could you please write a sample code explaining the problem? You can do it here or create a new issue in the mashumaro repo.

As far as I understand, you want to serialize generic types. Can [GenericSerializableType](https://github.com/Fatal1ty/mashumaro#genericserializabletype-interface) help with it?

nathaniel-may:
Hi, @Fatal1ty

That's very kind of you to offer to help here. Mashumaro does have several options for serializing generic types, but I haven't managed to figure out how to defer specifying the serialization for the concrete inner type to mashumaro.

Using the code in this PR, I currently have a `SerializationStrategy` registered with a config that returns a dict since that's the only concrete type we're using it with at this time:

```
class LazySerialization1(SerializationStrategy):
    def serialize(self, value) -> Dict[str, List[str]]:
        return value.force()
        
class Config(MashBaseConfig):
        serialization_strategy = {
            ...
            Lazy[Dict[str, List[str]]]: LazySerialization1()
        }
```

And this works great, but when I swap out the class to instead use a value of `Lazy[str]` mashumaro fails to serialize with the following exception: `mashumaro.exceptions.UnserializableField: Field "dump" of type Lazy[str] in DumpBeforeAddGraph is not serializable`.

What I want is a way for the above to work for any type `Lazy[T]` when `T` is either natively serializable (like `str`), or when `T` is a serializable type, or there exists a serialization strategy for `T` without needing to write code to handle serializing `T` again in the generic type's serializer. If something like the following worked that would be much nicer than what I've come up with:

_psudo-code:_
```
class LazySerialization(SerializationStrategy):
    def serialize(self, value):
        return value.force()
        
class Config(MashBaseConfig):
        serialization_strategy = {
            ...
            Lazy: LazySerialization()
        }
```

Fatal1ty:
@nathaniel-may thank you for the details!

I came up with the idea of using `DataClassDictMixin` with `SerializableType` since it's legit. The code should be written as follow:

```python
from datetime import date
from typing import Generic, TypeVar
from mashumaro import DataClassDictMixin
from mashumaro.types import SerializableType

T = TypeVar("T")

@dataclass
class Lazy(Generic[T], DataClassDictMixin, SerializableType):
    inner: T

    def _serialize(self) -> T:
        return self.to_dict()['inner']

    @classmethod
    def _deserialize(cls, value):
        return cls.from_dict({'inner': value})

@dataclass
class LazyTest(DataClassDictMixin):
    lazy: Lazy[date]

obj = LazyTest.from_dict({'lazy': '2022-01-31'})
assert obj.lazy == Lazy[date](date(2022, 1, 31))
print(obj.to_dict())
assert obj.to_dict() == {'lazy': '2022-01-31'}
```

But unfortunately it won't work as expected. I think it's a bug that could be fixed. I'll dig deeper and try to fix it.

nathaniel-may:
@gshank suggested `Lazy` live in `dbt.helper_types` which makes sense to me so I've moved it there.

nathaniel-may:
@Fatal1ty, since I've closed this PR if you'd like to continue our conversation in an issue or PR in the mashumaro repo, I'm happy to do so.

jtcohen6:
Nice work here!!

Is this a fix we're going to be able to backport to `1.0.latest`, for inclusion in v1.0.2 (given it resolves a significant performance regression)? Or does it intersect with / depend on https://github.com/dbt-labs/dbt-core/pull/4505 in a way that makes backporting too difficult?

nathaniel-may:
@jtcohen6 this definitely depends on #4505. If it's important to backport, I can make a different commit to 1.0.latest but the user-facing behavior won't be exactly the same. If I exclude the `dump` field from the structured logging output, but keep it in the string message, I could get this fixed over there too without mashumaro.

jtcohen6:
@nathaniel-may I'd be supportive of fixing the performance regression, even at the cost of losing that information in the structured logging output. To that end:

> If I exclude the `dump` field from the structured logging output, but keep it in the string message, I could get this fixed over there too without mashumaro.

This feels like the right balance of effort + impact

Fatal1ty:
> But unfortunately it won't work as expected. I think it's a bug that could be fixed. I'll dig deeper and try to fix it.

@nathaniel-may actually I missed a much simpler solution that works right now. You can wrap a value in the dataclass and use serialization hooks:
```python
from datetime import date
from typing import Generic, TypeVar
from mashumaro import DataClassDictMixin

T = TypeVar("T")


@dataclass
class Lazy(Generic[T], DataClassDictMixin):
    inner: T

    def __post_serialize__(self: T, d):
        return d.pop('inner')

    @classmethod
    def __pre_deserialize__(cls, d):
        return {'inner': d}

    # @classmethod
    # def __post_deserialize__(cls, obj: T):
    #     # if you want to get LazyTest(lazy=datetime.date(2022, 1, 31))
    #     # instead of LazyTest(lazy=Lazy(inner=datetime.date(2022, 1, 31)))
    #     # but serialization will be broken because date doesn't have to_dict
    #     return obj.inner

    class Config:
        serialization_strategy = {
            # ...
        }


@dataclass
class LazyTest(DataClassDictMixin):
    lazy: Lazy[date]


obj = LazyTest.from_dict({'lazy': '2022-01-31'})
print(obj)
assert obj.lazy == Lazy[date](date(2022, 1, 31))
print(obj.to_dict())
assert obj.to_dict() == {'lazy': '2022-01-31'}
```

PR #4620, opened 2022-01-25
Initial addition of CODEOWNERS file

leahwicz:
### Description

This file will help us automatically map files to the correct sub-teams for review. This file will be every growing and evolving so this is by no means a finished version. Just looking to get it started and then add to it as we go.

### Checklist

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -2,6 +2,10 @@
 # PRs will be automatically assigned for review to the associated 
 # team(s) or person(s) that touches any files that are mapped to them.
 #
+# A statement takes precedence over the statements above it so more general
+# assignments are found at the top with specific assignments being lower in
+# the ordering (i.e. catch all assignment should be the first item)
+#
 # Consult GitHub documentation for formatting guidelines:
 # https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners#example-of-a-codeowners-file
 
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -12,8 +12,25 @@
 # Changes to GitHub configurations including Actions
 /.github/ @leahwicz
 
-# Adapter / plugin code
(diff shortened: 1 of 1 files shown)


.github/CODEOWNERS
```diff
@@ -0,0 +1,22 @@
...
+# Changes to GitHub configurations including Actions
+/.github/ @leahwicz
+
+# Adapter / plugin code
+/plugins @dbt-labs/core-adapters
+
+# Perf regression testing framework
+# This excludes the test project files itself since those aren't specific 
+# framework changes (excluded by not setting an owner next to it- no owner)
+/performance @nathaniel-may
```
leahwicz, line 21: @nathaniel-may I put you down as the reviewer here for now until we ramp other people up here (which would be the goal). I hope that works for you but if not, let me know

nathaniel-may:
you can probably add the event module since it's pretty cleanly segmented:
/core/dbt/events - language team (def emily gerda me at least)

leahwicz:
@nathaniel-may I added both teams since in the log run, logging will fall under execution but for now the expertise lies with language

emmyoop (approved):
LGTM

nathaniel-may (approved):
looks like a good start to me!

VersusFacit (approved):
👍 Although the token-matcher in me is bothered that the `*` comes first since all matches will be against that and never fall through to your specific cases. 😂

nathaniel-may:
@VersusFacit good point. I have no idea what the semantics of line order are in this file 🤔

leahwicz:
@nathaniel-may @VersusFacit - you're right, it's weird but the last statement takes precedence over everything above it so that's why the `*` has to be at the top (according to the GitHub docs and example)

https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners#example-of-a-codeowners-file

leahwicz:
Maybe I'll add that as a comment

PR #4623, opened 2022-01-25
Set flags from args in lib module for dbt-server

racheldaniel:
resolves #TIGER-84 and Tiger-99

### Description

Changes are only in the `dbt.lib` model, which is the wrapper for dbt-server use cases. This updates config creation to pull in args passed to the server from client command flags, if any exist. This change is needed to enable use of several command flags that need to be applied prior to dbt task execution. This PR also defaults the profile to 'user' -- it looks like we may have attempted to do so before, but we had the field name wrong (`profile_name` vs `profile`)

### Checklist

### DIFF
--- a/core/dbt/lib.py
+++ b/core/dbt/lib.py
@@ -5,27 +5,24 @@ from dbt import flags
 from collections import namedtuple
 
 RuntimeArgs = namedtuple(
-    'RuntimeArgs', 'project_dir profiles_dir single_threaded profile_name'
+    'RuntimeArgs', 'project_dir profiles_dir single_threaded'
 )
 
-
-def get_dbt_config(project_dir, single_threaded=False):
+def get_dbt_config(project_dir, args=None, single_threaded=False):
     from dbt.config.runtime import RuntimeConfig
     import dbt.adapters.factory
     import dbt.events.functions
-
     if os.getenv('DBT_PROFILES_DIR'):
         profiles_dir = os.getenv('DBT_PROFILES_DIR')
     else:
         profiles_dir = os.path.expanduser("~/.dbt")
-
     # Construct a phony config
     config = RuntimeConfig.from_args(RuntimeArgs(
-        project_dir, profiles_dir, single_threaded, 'user'
+        project_dir, profiles_dir, single_threaded
     ))
     # Clear previously registered adapters--
     # this fixes cacheing behavior on the dbt-server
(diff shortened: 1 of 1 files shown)


jtcohen6:
After this is merged, do we want to backport it for inclusion in the next `v1.0.x` patch release? My sense is yes (cc @leahwicz)

core/dbt/lib.py
```diff
@@ -5,27 +5,28 @@
...
     else:
         profiles_dir = os.path.expanduser("~/.dbt")
 
-    # Construct a phony config
+    target = args.target if hasattr(args, 'target') else None
+
+    # Construct a phony config-- hardcoding user to match dbt-cloud for now
     config = RuntimeConfig.from_args(RuntimeArgs(
-        project_dir, profiles_dir, single_threaded, 'user'
+        project_dir, profiles_dir, single_threaded, 'user', target
```
drewbanin, line 25: do we want to continue to hard-code "user" here, or should we change this function to accept an argument for `profile` (or, possibly grab it out of `args.profile`?

racheldaniel: @drewbanin that's what I was doing in the previous commit on this branch, but I don't think that gets us the desired result. If I'm developing in cloud and I pass anything except `user` with the `--profile` flag, the run will fail. So we could keep this as a flag, but that would mean every single run they'd have to type `dbt-client run --profile user`.

drewbanin: ok - that makes sense - but I think fixing this problem here in the dbt Core repo might not be the best place for this logic. I think a more flexible approach might be to:
 - Allow callers of this function to provide an arbitrary profile name (as it sounds like you did in the previous commit)
 - Continue to allow users of the dbt-client to name their profiles whatever they want
 - Add logic to make sure that the name of the profile that is generated in dbt Cloud matches the name of the profile that the user has specified.

So, basically treat this as a dbt Cloud problem and fix it over there. Have you tried something like that already?

racheldaniel: Discussed offline-- we're going to hardcode the "user" profile in the server instead of dbt for now

drewbanin (approved):
This LGTM, but I have not tested out the code locally! Let me know if taking it for a spin would be helpful at all. Code looks great though! Thanks @racheldaniel :)

kwigley (approved):

PR #4631, opened 2022-01-27
fix comparision for new model/body

### LINKED ISSUE #4570: [CT-40] [Bug] Using state:modified.body returns NoneType object has no attribute raw_sql

### Current Behavior

I am trying to use the `state:modified.body` selector which is returning the following error:
```
'NoneType' object has no attribute 'raw_sql'
```

It appears to be struggling with identifying body changes when the model didn't previously exist as its not able to pull the `raw_sql` to reference. 

### Expected Behavior

The `state:modified.body` selector should return all models whose body changes OR didn't previously exist.

### Steps To Reproduce

1. Compile project
2. Add an entirely new model
3. Run `dbt ls --select state:modified.body` which yields this error message

```
[error] [MainThread]: Encountered an error:
'NoneType' object has no attribute 'raw_sql'
```

However,  if I repeat the steps, but modify the intersection to be just `state:modified`, it returns the expected models.

### Relevant log output

_No response_

### Environment

```markdown
- OS:  Darwin-20.6.0-x86_64-i386-64bit (Mac OS Big Sur 11.5.2)
- Python: 3.7.8
- dbt: 1.0.0
```

snowflake

### Additional Context

_No response_

ChenyuLInx:
resolves #4570

### Description
Selecting modified.body used to throw an error after adding a new model. This change fixed the behavior to return the newly added model boday

### Checklist

### DIFF
--- a/core/dbt/graph/selector_methods.py
+++ b/core/dbt/graph/selector_methods.py
@@ -488,9 +488,11 @@ class StateSelectorMethod(SelectorMethod):
         return self.check_macros_modified(new)
 
     @staticmethod
-    def check_modified_factory(compare_method: str) -> Callable:
+    def check_modified_factory(
+        compare_method: str
+    ) -> Callable[[Optional[SelectorTarget], SelectorTarget], bool]:
         # get a function that compares two selector target based on compare method provided
-        def check_modified_things(old: Optional[SelectorTarget], new: SelectorTarget,) -> bool:
+        def check_modified_things(old: Optional[SelectorTarget], new: SelectorTarget) -> bool:
             if hasattr(new, compare_method):
                 # when old body does not exist or old and new are not the same
                 return not old or not getattr(new, compare_method)(old)  # type: ignore
--- a/core/dbt/graph/selector_methods.py
+++ b/core/dbt/graph/selector_methods.py
@@ -489,7 +489,7 @@ class StateSelectorMethod(SelectorMethod):
 
     @staticmethod
     def check_modified_factory(compare_method: str) -> Callable:
-        
+        # get a function that compares two selector target based on compare method provided
         def check_modified_things(old: Optional[SelectorTarget], new: SelectorTarget,) -> bool:
             if hasattr(new, compare_method):
                 # when old body does not exist or old and new are not the same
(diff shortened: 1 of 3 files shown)


core/dbt/graph/selector_methods.py
```diff
@@ -478,42 +478,26 @@ def check_macros_modified(self, node):
             previous_macros = []
             return self.recursively_check_macros_modified(node, previous_macros)
 
-    def check_modified(self, old: Optional[SelectorTarget], new: SelectorTarget) -> bool:
+    # TODO check modifed_content and check_modified macro seems a bit redundent
```
ChenyuLInx, line 481: @gshank  do you know what's the main difference between these two functions? seems a little bit confusing here

core/dbt/graph/selector_methods.py
```diff
@@ -527,14 +513,21 @@ def search(
...
-            'modified': self.check_modified,
-            'modified.body': self.check_modified_body,
-            'modified.configs': self.check_modified_configs,
-            'modified.persisted_descriptions': self.check_modified_persisted_descriptions,
-            'modified.relation': self.check_modified_relation,
-            'modified.macros': self.check_modified_macros,
+            'modified':
+                self.check_modified_content,
+            'modified.body':
+                self.check_modified_factory('same_body'),
```
ChenyuLInx, line 522: Let me know if you think we should go change the underlying function names instead of using the factory here! they are used somewhere but I think it is not too much

core/dbt/graph/selector_methods.py
jtcohen6: I think the way you're doing this is fine. The `check_modified` method (`state:modified` for users) is meant to be the superset of all checks. Renaming this so it clearly maps to `same_contents` (the umbrella check) makes sense

core/dbt/graph/selector_methods.py
gshank: I like the factory. It avoids duplicating the code in the functions.

gshank (approved):
Looks good! I like the improved efficiency.

leahwicz:
@ChenyuLInx this PR needs backported if it hasn't yet

PR #4647, opened 2022-01-31
Change InvalidRefInTestNode level to DEBUG

### LINKED ISSUE #4594: [CT-71] Restore previous behavior when a test depends on a disabled (vs. missing) resource
_Prompted by Slack threads [here](https://getdbt.slack.com/archives/C02HM9AAXL4/p1642605259009400) + [here](https://getdbt.slack.com/archives/C02HM9AAXL4/p1641486649000200)_

Folks who've newly upgraded to v1.0 are seeing a _lot_ of warn-level (stdout) messages about tests defined on disabled models. This is what our internal analytics project looks like, and I don't think it's an uncommon experience:

Screenshot 2022-01-19 at 19 16 48

(There's many more below that)

It is desirable to warn about tests that have "missed" the correct model, such as due to a typo in a `ref` or yaml property (e.g. see https://github.com/dbt-labs/dbt-core/issues/4532). It's also a very common pattern to disable unused/irrelevant models in _packages_, and the result is a quite cluttered terminal. There are some useful warnings in there (missing node for patch), lost as needles in haystack-traces.

What I proposed in one of the threads linked above:

> If a test depends on a disabled model, log this as DEBUG-level (shows up in `logs/dbt.log`)
> If a test depends on a missing model, log this as WARN-level (shows up in standard CLI output)

I looked quickly at the code, and turns out this was actually our previous behavior:
https://github.com/dbt-labs/dbt-core/blob/08b2f16e10eb3420ecffee3b4d18ac2f45e601ea/core/dbt/parser/manifest.py#L790-L803

We've since replaced the first of those with a call to the `InvalidRefInTestNode` event, so this would just look like changing:
https://github.com/dbt-labs/dbt-core/blob/a588607ec6546de441812e215fbbf4e9ff20101b/core/dbt/events/types.py#L1186-L1192

To:
```python
@dataclass
class InvalidRefInTestNode(DebugLevel):
    msg: str
    code: str = "I051"

    def message(self) -> str:
        return self.msg
```

Voila, just the good stuff:

Screenshot 2022-01-19 at 19 19 35

jtcohen6:
**[v1.0.2]**

resolves #4594

### Description

Change `InvalidRefInTestNode` event to:
- Restore previous (pre-v1.0) behavior for this message's log level (DEBUG, not WARN) — thereby decluttering stdout for users of packages that enable/disable models as a common pattern
- Remove `ui.warning_tag` for consistency — why include the word `WARNING` when it's `DEBUG`?

### Checklist


emmyoop (approved):
Makes sense!

PR #4656, opened 2022-02-01
Modified Backport of #4619

### LINKED ISSUE #4569: [CT-39] Bug: ensure that dump_graph is not run for logging events in cachy.py unless cache logging is turned on

nathaniel-may:
resolves #4569

### Description

This commit defers the expensive function `dump_graph` to only be called in the event the value is going to be logged. Without this change, users always generate this expensive graph regardless of the value of the flag `--log-cache-events`.

This backport of #4619 is separate from the original commit so that we do not have to backport #4505 which introduced mashumaro as the internal serializer for the events module. In order to keep this backport simple, I implemented it by passing python functions as values and silencing mypy to get type checking to pass. All tests from the original commit have been backported and adapted to this strategy.

### Checklist


nathaniel-may:
Closing in favor of #4660

PR #4660, opened 2022-02-01
Alternative Modified Backport of #4619

### LINKED ISSUE #4569: [CT-39] Bug: ensure that dump_graph is not run for logging events in cachy.py unless cache logging is turned on

nathaniel-may:
resolves #4569

###Description

This commit defers the expensive function dump_graph to only be called in the event the value is going to be logged. Without this change, users always generate this expensive graph regardless of the value of the flag `--log-cache-events`. This is an alternative solution to the one proposed in #4656.

This backport of #4619 is separate from the original commit so that we do not have to backport #4505 which introduced mashumaro as the internal serializer for the events module. In order to keep this backport simple, I implemented it by adding a new function `fire_event_if` that only constructs the event and logs it if the condition is met. This keeps logging concerns within the logging module while exposing conditional logging at call sites. Tests for the original commit are not directly applicable anymore because we are not using `fire_event` anymore.

### Checklist


gshank (approved):

leahwicz (approved):
LGTM. My only thing would be could we add a test for this just to cover all the bases?

nathaniel-may:
@leahwicz I just added a simple unit test for `fire_event_if` which shows that if you use that function properly, it CAN have the behavior we want. I looked into writing an integration test to determine if it IS being used correctly, but I would need to add another event to log when `dump_graph` is being called since I don't see a way to access that state in a unit or integration test.

In the interest of not adding new behavior to a patch release, I think we shouldn't add the new log event for the test. Is the test I've added enough or should we look into other avenues?

leahwicz:
@nathaniel-may I think this is good enough. Just verifies the new method which I'm happy with. Thanks!

PR #4671, opened 2022-02-03
Black formatting TEST -- DO NOT MERGE

iknox-fa:
This is what black would do to our codebase if invoked for all files.
N.B.  line length (one of the very few settings black supports) was set to match our flake8 config at 99 chars per line.


.github/actions/latest-wrangler/main.py
```diff
@@ -35,12 +35,11 @@
...
 
     # TODO: verify package meta is "correct"
-    # https://github.com/dbt-labs/dbt-core/issues/4640 
+    # https://github.com/dbt-labs/dbt-core/issues/4640
 
     # map versions and tags
     version_tag_map = {
-        version["id"]: version["metadata"]["container"]["tags"]
-        for version in package_meta
+        version["id"]: version["metadata"]["container"]["tags"] for version in package_meta
```
iknox-fa, line 42: This is what I was talking about when it reformats my comprehensions.  In this case it does probably improve readability, but in more complex comprehensions having line breaks make logical separation of what's happening can be useful.. 

That said, if my comprehensions are complex enough to require visual clues about their processing, I should probably not use a comprehension.

:)

core/dbt/adapters/base/relation.py
```diff
@@ -173,40 +176,35 @@ def _render_iterator(
...
             identifier=source.identifier,
             quote_policy=quote_policy,
-            **kwargs
+            **kwargs,
         )
 
     @staticmethod
     def add_ephemeral_prefix(name: str):
-        return f'__dbt__cte__{name}'
+        return f"__dbt__cte__{name}"
```
iknox-fa, line 207: Anecdotally, I'd say that ~50% of changes are `'` -> `"`

core/dbt/adapters/base/relation.py
```diff
@@ -1,13 +1,16 @@
...
 from dataclasses import dataclass
-from typing import (
-    Optional, TypeVar, Any, Type, Dict, Union, Iterator, Tuple, Set
-)
+from typing import Optional, TypeVar, Any, Type, Dict, Union, Iterator, Tuple, Set
 
 from dbt.contracts.graph.compiled import CompiledNode
 from dbt.contracts.graph.parsed import ParsedSourceDefinition, ParsedNode
 from dbt.contracts.relation import (
-    RelationType, ComponentName, HasQuoting, FakeAPIObject, Policy, Path
```
iknox-fa, line 10: There are a lot of these as well.  I'm a big fan of this change as it's much easier to see list membership when scanning up/down vs left/right (for me anyway)

nathaniel-may:
do we have to change our flake8 configuration to be compatible with this as well?

core/dbt/adapters/base/impl.py
```diff
@@ -551,8 +517,7 @@ def list_relations_without_caching(
         :rtype: List[self.Relation]
         """
         raise NotImplementedException(
-            '`list_relations_without_caching` is not implemented for this '
-            'adapter!'
+            "`list_relations_without_caching` is not implemented for this " "adapter!"
```
gshank, line 520: This is kind of weird. I guess it works... but seems like something we should cleanup at some point.

core/dbt/adapters/base/relation.py
gshank: I don't really mind the double quotes, I guess they need to standardize on something. Is there a reason for preferring double to single quotes, or was it an arbitrary decision?

core/dbt/adapters/base/relation.py
gshank: I'm not such a big fan of this because in some cases it's going to make the import headers pretty long. But I shrug...

gshank:
I think we would be better off formatting all of core/dbt. I'd still like to request not reformatting the test directory, because that's going to be going away eventually anyway, and the changes are very dense there.

core/dbt/adapters/base/relation.py
```diff
@@ -159,9 +164,7 @@ def without_identifier(self) -> 'BaseRelation':
         """
         return self.include(identifier=False).replace_path(identifier=None)
 
-    def _render_iterator(
```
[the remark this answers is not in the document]
gshank: The line length was actually 79 when I started. We changed it to 99 at some point.

core/dbt/adapters/base/impl.py
nathaniel-may: I like that it's making lines fit the configured length when they can

kwigley:
If we would like to attempt to maintain the git history and blame with this, we could potentially try using https://github.com/AlayaCare/git-black for the initial reformatting of the codebase.

kwigley:
> do we have to change our flake8 configuration to be compatible with this as well?

yes! just tested the following..

in `tox.ini` we should add these settings
```ini
[flake8]
max-line-length = 99
extend-ignore = 
    # https://www.flake8rules.com/rules/E203.html - align with black formatting behavior
    E203,
    # https://www.flake8rules.com/rules/W504.html
    W504,
    # https://www.flake8rules.com/rules/E741.html
    E741
extend-exclude = 
    # ignore virtualenv dir
    env/
    # ignore test directory
    test/
    # ignore scripts
    scripts/
    # ignore scripts
    core/scripts/
    # ignore github actions
    .github/

```

and the command for the flake8 testenv in `tox.ini` should read 
```ini
commands = flake8
```
instead of
```ini
commands = flake8 --select=E,W,F --ignore=W504,E741 --max-line-length 99 \
  core/dbt \
  plugins/postgres/dbt
```

core/dbt/adapters/base/relation.py
iknox-fa: @gshank https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#strings

core/dbt/adapters/base/impl.py
iknox-fa: What's weird about it?  Looks like a straightforward application of 
`black uses single quotes` + `black honors line length` to me.  

The original code is a little weird (two quoted strings concat'd by the fact they follow each other), but that's not black's fault.

iknox-fa:
> do we have to change our flake8 configuration to be compatible with this as well?
We do, but it's trivial.  Basically we need to set the line-length to match and address [this bug](https://github.com/PyCQA/pycodestyle/issues/373) by ignoring E203 in the flake8 setup.

iknox-fa:
> If we would like to attempt to maintain the git history and blame with this, we could potentially try using https://github.com/AlayaCare/git-black for the initial reformatting of the codebase.

Ok, this is very cool if it works!  I'll give it a try soon.

iknox-fa:
> > do we have to change our flake8 configuration to be compatible with this as well?
> 
> yes! just tested the following..
> 
> in `tox.ini` we should add this config to the bottom, flake8 will look for this config regardless of where it is run (IDE, editor, cli, tox..)
> 
[SNIP]

I'm not 100% sold that we should be running Flake8 in tox at all-- moving it to pre-commit hooks and leaving tox to handle things where the environment actually matters would be a good idea IMO.

kwigley:
> I'm not 100% sold that we should be running Flake8 in tox at all-- moving it to pre-commit hooks and leaving tox to handle things where the environment actually matters would be a good idea IMO.

@iknox-fa 100% agree, I still think configuration for flake8 and black should live in config files. flake8 will look in the `tox.ini` file for any configuration, which is the example I provided, but this could also be a `.flake` or `setup.cfg` file in the project. It is only in tox because it has always been there and it is easy to use in CI (installs dependencies, runs tasks in parallel), but there is no reason to keep it there.

iknox-fa:
> > I'm not 100% sold that we should be running Flake8 in tox at all-- moving it to pre-commit hooks and leaving tox to handle things where the environment actually matters would be a good idea IMO.
> 
> @iknox-fa 100% agree, I still think configuration for flake8 and black should live in config files. flake8 will look in the `tox.ini` file for any configuration, which is the example I provided, but this could also be a `.flake` or `setup.cfg` file in the project. It is only in tox because it has always been there and it is easy to use in CI (installs dependencies, runs tasks in parallel), but there is no reason to keep it there.

That's fair-- `.flake` works for me.  Too bad we [can't use `pyproject.toml`](https://github.com/PyCQA/flake8/issues/234).  Oh python.

PR #4685, opened 2022-02-04
include directory README

McKnight-42:
### Description

Writing up a README for the include directory giving brief description of each folder or file in top level.

### Checklist


core/dbt/include/README.md
```diff
@@ -1 +1,10 @@
-# Include README
+# Include Module
+
+contains the a few key default `dbt-core` sub directories inlcuding.
+
+- `global_project` which defines the default implementations of jinja2 macros for `dbt-core ` which can be overwritten in each adapter repo to work more in line with those adapter plugins. to view adapter specific jinja2 changes please check the relevant adapter repo.
+
+- `starter_project` which produces the default init project inside the `dbt-core` project after doing the `make dev` command
```
emmyoop, line 7: The `starter_project` is what we use to generate a project from `dbt init`.  I think it's important to point out that cloud does *not* use this.  They use a tag of https://github.com/dbt-labs/dbt-starter-project.

How is `make dev` using this?

core/dbt/include/README.md
```diff
@@ -1 +1,10 @@
...
+
+- A `index.html` file used to generate the docs page after using the `generate docs` command in dbt.
```
emmyoop, line 9: Can we point out that `index.html` is actually built from the [dbt-docs](https://github.com/dbt-labs/dbt-docs) repo?  The `README` over there has details on how it's built.  And then that file is used when `docs generate` is run.

core/dbt/include/README.md
```diff
@@ -1 +1,10 @@
-# Include README
+# Include Module
+
+contains the a few key default `dbt-core` sub directories inlcuding.
+
+- `global_project` which defines the default implementations of jinja2 macros for `dbt-core ` which can be overwritten in each adapter repo to work more in line with those adapter plugins. to view adapter specific jinja2 changes please check the relevant adapter repo.
```
emmyoop, line 5: Probably worth pointing to what you're referring to in at least one of the adapter repos with a note that each adapter is different so someone looking at this at least know where to go look.

core/dbt/include/README.md
gshank: And maybe mention that the index.html is generated by hand prior to a release?

core/dbt/include/README.md
```diff
@@ -1 +1,10 @@
-# Include README
+# Include Module
+
+contains the a few key default `dbt-core` sub directories inlcuding.
```
emmyoop, line 3: There's a few typos here to fix.

core/dbt/include/README.md
McKnight-42: That's a good point will look for a macro that seems to be different from default in at least most of the adapters so it provides a varied look over.

core/dbt/include/README.md
McKnight-42: I might need a bit more context for that @gshank I haven't had to deal with the dbt-docs so to what degree to we have to regenerate the project? prior to release not sure if we have a write up on that anywhere.

gshank: Jeremy usually generated the file, but Emily did it last time. The file is created from the dbt-docs repo and copied here.

McKnight-42: ah, okay interesting, now i'm curious to why we have to do it each release? is it just in case any name or ui changes we make?

emmyoop: @gerda is it not generated only when `dbt-docs` changes?  It just needs to be done prior to a release when changes occur since the file gets committed to `dbt-core`.

gshank: We have sometimes changed attributes in core that dbt-docs depends on. So it needs to be generated for each release, to check that it works.

McKnight-42: ah that's intriguing, will highlight that and try to have a new version up today.

jtcohen6: To clarify this point: `dbt-docs` depends on `manifest.json` + `catalog.json`, which are contracted/versioned. If the schema of those artifacts changes (in new minor versions), we always want to test that `index.html` still works / is compatible. We only need to regenerate `index.html` if we've made changes to the `dbt-docs` source code (for compatibility with the new artifact schemas, to fix an unrelated bug, etc).

core/dbt/include/README.md
```diff
@@ -1 +1,15 @@
...
+
+## `global_project`
+Defines the default implementations of jinja2 macros for `dbt-core ` which can be overwritten in each adapter repo to work more in line with those adapter plugins. to view adapter specific jinja2 changes please check the relevant adapter repos [`adapter.sql` ](https://github.com/dbt-labs/dbt-bigquery/blob/main/dbt/include/bigquery/macros/adapters.sql) file in the `include` directory or in the [`impl.py`](https://github.com/dbt-labs/dbt-bigquery/blob/main/dbt/adapters/bigquery/impl.py) file for some ex. bigquery (truncate_relation).
+
+## `starter_project`
+Produces the default  project after running the `dbt init` command for CLI. `dbt-cloud` initializes the project by using [dbt-starter-project](https://github.com/dbt-labs/dbt-starter-project).
+
+
+# Files
+ - `index.html` a file built uising [dbt-docs](https://github.com/dbt-labs/dbt-docs) prior to new releases and replaced in the `dbt-core` directory; is used to generate the docs page after using the `generate docs` command in dbt.
```
emmyoop, line 15:
```suggestion
 - `index.html` a file built using [dbt-docs](https://github.com/dbt-labs/dbt-docs) prior to new releases and replaced in the `dbt-core` directory; is used to generate the docs page after using the `generate docs` command in dbt.
```

core/dbt/include/README.md
```diff
@@ -1 +1,15 @@
-# Include README
+# Include Module
+
+The Include module is reponsible for housing default macro definitions, starter project scaffold, and the html file used to generate the docs page.
+
+# Directories
+
+## `global_project`
+Defines the default implementations of jinja2 macros for `dbt-core ` which can be overwritten in each adapter repo to work more in line with those adapter plugins. to view adapter specific jinja2 changes please check the relevant adapter repos [`adapter.sql` ](https://github.com/dbt-labs/dbt-bigquery/blob/main/dbt/include/bigquery/macros/adapters.sql) file in the `include` directory or in the [`impl.py`](https://github.com/dbt-labs/dbt-bigquery/blob/main/dbt/adapters/bigquery/impl.py) file for some ex. bigquery (truncate_relation).
```
emmyoop, line 8:
```suggestion
Defines the default implementations of jinja2 macros for `dbt-core` which can be overwritten in each adapter repo to work more in line with those adapter plugins. To view adapter specific jinja2 changes please check the relevant adapter repo [`adapter.sql`](https://github.com/dbt-labs/dbt-bigquery/blob/main/dbt/include/bigquery/macros/adapters.sql) file in the `include` directory or the [`impl.py`](https://github.com/dbt-labs/dbt-bigquery/blob/main/dbt/adapters/bigquery/impl.py) file for some ex. BigQuery (truncate_relation).
```

core/dbt/include/README.md
```diff
@@ -1 +1,15 @@
...
+
+## `starter_project`
+Produces the default  project after running the `dbt init` command for CLI. `dbt-cloud` initializes the project by using [dbt-starter-project](https://github.com/dbt-labs/dbt-starter-project).
```
emmyoop, line 11:
```suggestion
Produces the default project after running the `dbt init` command for the CLI. `dbt-cloud` initializes the project by using [dbt-starter-project](https://github.com/dbt-labs/dbt-starter-project).
```

core/dbt/include/README.md
```diff
@@ -1 +1,15 @@
...
+
+
+# Files
+ - `index.html` a file built using [dbt-docs](https://github.com/dbt-labs/dbt-docs) prior to new releases and replaced in the `dbt-core` directory; is used to generate the docs page after using the `generate docs` command in dbt.
```
emmyoop, line 15:
```suggestion
 - `index.html` is a file generated from [dbt-docs](https://github.com/dbt-labs/dbt-docs) prior to new releases and replaced in the `dbt-core` directory.  It is used to generate the docs page after using the `generate docs` command in dbt.
```

emmyoop (changes requested):
A few tiny edits and then you'll need to generate a changelog entry with changie before merging in.

.changes/unreleased/Docs-20220328-094152.yaml
```diff
@@ -0,0 +1,7 @@
+kind: Docs
```
emmyoop, line 1: @jtcohen6 do we want `Docs` to just be documentation changes generally or strictly for `dbt-docs` changes?  

Apart from this PR - it might be worth updating the changie README with more details on each change type (ie, be sure to add details of what will break on a breaking change).

jtcohen6: just my opinions:
- original intent with `Docs` was "strictly for dbt-docs changes"
- generally I think a change like this can be `Skip changelog` — README, reflected in GitHub only, no real change to the functionality of installed `dbt-core`

emmyoop: I was going for a changelog entry for all the things!  But it makes sense that this is a good candidate for `Skip Changelog`.  @McKnight-42 can you make that happen?

emmyoop (approved):
Looks good!

PR #4689, opened 2022-02-05
Feature/dbt deps tarball

### LINKED ISSUE #4205: [Feature] add new dbt.deps type: url to internally hosted tarball
* [x]  I have searched the existing issues

#### Summary

From experience with dbt at a large org, there is no great way to distribute internal (non-public) dbt modules. The only currently supported solution is leveraging GitPackages + shipping ssh credentials to wherever `dbt deps` would be run. A new second option for a private module source, is suggested below, which relies on direct tarball links, and that is unpinned only. The implication is creating a source type that does not leverage all the pinned/unpinned internals, but the trade off is reasonable to grant users a clean direct install source for private packages that is not git. 

#### Full Description
dbt deps currently supports package installation from 3 options
- LocalPackage: install via [simlink](https://github.com/dbt-labs/dbt-core/blob/40ae6b6bc860a30fa383756b7cdb63709ce829a8/core/dbt/deps/local.py#L47). Audience: developers with local copy of dbt module to install. 
- GitPackage: install via [git checkout](https://github.com/dbt-labs/dbt-core/blob/40ae6b6bc860a30fa383756b7cdb63709ce829a8/core/dbt/deps/git.py#L113). Audience: end users, and production systems, with modules that can be posted publicly. Also supports private modules for end users/production systems with appropriate git credentials. 
- RegistryPackage: install via [tarball download and untar to modules destination](https://github.com/dbt-labs/dbt-core/blob/40ae6b6bc860a30fa383756b7cdb63709ce829a8/core/dbt/deps/registry.py#L61) from `https://hub.getdbt.com/`. Implemented via dedicated [client](https://github.com/dbt-labs/dbt-core/blob/main/core/dbt/clients/registry.py) due to API interactions with the Registry. Audience: end users, production systems, with modules that can be posted publicly. Great for beginners and power users alike. 

##### Whats missing? 
- Non-public modules hosted outside of GitHub. GitPackages, in production, require bundling the production systems with git credentials. 
- Wouldn't it be nice to be able to directly link to tarballs hosted on artifactory like services? That way there is no need to pass around a GitHub ssh key in production systems. 
- Wouldn't it be nice to be able to directly link to tarballs hosted on internal GCS/S3/Blob storage, instead of hitting git for installs?

##### Solution?
- Allow option to link to a tarball url. This allows for orgs to permission the tarball in a very granular way (either hosted on Artifactory like service, or cloud storage), instead the more broad git + ssh solution for non-public modules.
```
# example packages.yml
packages:
  - tarball: https://codeload.github.com/dbt-labs/dbt-utils/tar.gz/0.6.6
    # public package referenced here as example
    sha1: 549db5560efccbadc7ec7834c03cf13579199a66
    # optional. if specified, downloaded tar much match for install to proceed.
    subdirectory: dbt-utils-0.6.6
    # optional. same behavior as subdirectory arg with git packages
  ```

###### Why git is a poor solution/why a second option for private modules would benefit end users
 - Git is not brilliant at scoping access via ssh credentials.
 - From a security standpoint, bundling the git credentials on production systems is really really really really bad security practice. Yes one could dynamically load them in, no this is not a simple solution to orchestrate. 
 - For companies with enterprise GitHub setups, extra git clones on all CI/CD builds, is waste of finite resources. 
 - Larger companies already have Artifactory like services set up, being able to direct reference tarballs on these internal platforms a) solves all scoping issues without need to pass around an ssh key b) leverages caching on the Artifactory like services for free. 
 - Outside of using Artifactory like services as source, one could imagine instead a tarball hosted on S3/GCS,  where permissioning access is much easier via cloud IAM config. Easy to see where a setup like this would be preferable to handling access right through git ssh keys. 

### How I got here:
From experience with dbt at a large org, there is no great way to distribute internal (non-public) dbt modules. Gatekeeping for non-public repos relies on git credentials. But this is poor security practice (and not supported by our CI/CD infra, for good reason). Simply put, those reliant on exiting infra for managing builds may not have the luxury of issuing arbitrary git clone commands to external repos (as dbt deps does for GitPackage sources). 

- Currently, I bundle internal packages as part of our core dbt install image. All installs are then handled by doing `LocalPackage` symlinks to the packages on the core dbt image (along with some docker magic). It's working, but it's a hack, and being able to do straight install from a url would make me so much happier. 
 - Creating an internal alternative to `https://hub.getdbt.com/`. Redirect to internal service by leveraging ['DBT_PACKAGE_HUB_URL`](https://github.com/dbt-labs/dbt-core/blob/40ae6b6bc860a30fa383756b7cdb63709ce829a8/core/dbt/clients/registry.py#L9). Emulate current hub.getdbt.com API behavior in new internal only service, json object returned by API for a given package has a clear tarball source link. This would be a complicated way to do the same process as suggested here. Self hosted hub service remains an option for power users, but complex solution to simple problem to say the least. 
 - Submodules. Not bad, but very manual. Does not easily scale for multiple teams creating and providing packages for install. 
 
### Who will this benefit?
Any user looking to run `dbt deps` for internally hosted modules that wishes to not host the packages on git (or is not permitted). 

[Yes, PR in progress here](https://github.com/dbt-labs/dbt-core/pull/4220)

Sketch of planned work:
1) Add TarballPackage data class to [project.py](https://github.com/dbt-labs/dbt-core/blob/main/core/dbt/contracts/project.py). No plan to support pinned/unpinned since handling versioning is all but impossible given an arbitrary link to a tarball. Dataclass would look much like existing `LocalPackage`.
2) Add dbt.deps remoteTarball, TarballUnpinnedPackage as only class. Would look much like [local](https://github.com/dbt-labs/dbt-core/blob/main/core/dbt/deps/local.py#L47) in structure, but recycle most of `registry` [install design pattern](https://github.com/dbt-labs/dbt-core/blob/40ae6b6bc860a30fa383756b7cdb63709ce829a8/core/dbt/deps/registry.py#L68)
3) dbt.deps.resolver updated to support new `url` type, which maps newly created dbt.deps.tarball remoteTarball. 
4) [unit tests added](https://github.com/dbt-labs/dbt-core/blob/79aa13630125abbdbc9abb482399267f62a1d91e/test/unit/test_deps.py)

### Anything else?
First time dbt contributor, long time day to day dbt user. Looking forward to packaging this up and making my first contribution!

### PR DESCRIPTION
**Pull request by timle2**
resolves #4205

add new dbt.deps type: url to internally hosted tarball #420

Continued from
https://github.com/dbt-labs/dbt-core/pull/4220

*Revision 3 added Nov 6 2022*

Proposed solution for feature request [4205](https://github.com/dbt-labs/dbt-core/issues/4205)

### Description

Enable direct linking to tarball urls in `packages.yml`, for example:
```
# manufactured test, since you'd want to use hub to install these 
# public tarball used here as example only! 
# this would usually be a tarball hosted  on an internal network
packages:
  - tarball: https://codeload.github.com/dbt-labs/dbt-utils/tar.gz/0.6.5
    name: 'dbt_utils_065'

```
image
image

Rational:
- dbt projects being self hosted on larger enterprise environments often don't have a connection to the internet (dbt hubs won't work).
- dbt users on larger enterprise environments like to build internal private packages for non-public use (help out other dbt users in company with specific functionality)
-  git package install is not a good option at scale for larger enterprise environments
- internal file hosting service (such as internal artifactory service or internal cloud storage buckets) can be easily configured to host packages for install during deployment, so lets give dbt users a way to install from a direct tar file link

Sketching out doc changes here: 
https://github.com/timle2/docs.getdbt.com/blob/dbt-docs-tarball-package-updates/website/docs/docs/building-a-dbt-project/package-management.md#tar-files

### Checklist

- [x] I have [opened an issue to add/update docs](https://github.com/dbt-labs/docs.getdbt.com/issues/new/choose)~, or docs changes are not required/relevant for this PR~
    - https://github.com/dbt-labs/docs.getdbt.com/issues/2474

### CODE REVIEW

**Comment by timle2** [comment]
TODO: https://github.com/dbt-labs/dbt-core/pull/4220#discussion_r793784822

> We've updated how to we log to use the Events modules. You'll need to create a new log event for this, and everywhere else you log, and use `fire_event` to kick off writing the log. We have a [`README`](https://github.com/dbt-labs/dbt-core/blob/main/core/dbt/events/README.md) with some good info on our logging system.

**Comment by timle2** [comment]
TODO
https://github.com/dbt-labs/dbt-core/pull/4220#discussion_r793801205
> As a standard we don't use `assert` in our code other than for tests. They shadow the actual exceptions and throw an `AssertionError` instead. Please rewrite the assert statements to catch relevant exceptions (create new ones if needed!).
> 
> In this specific case, I do like the use `is_tarfile` to check the file! Can you think of a reason it would be worth retrying the download if the file is not a valid tarfile?

**Comment by emmyoop** [comment]
@cla-bot check

**Comment by emmyoop** [comment]
@timle2 thanks for opening up the PR with your new username!  I've double checked and you did indeed sign the CLA.  We have your username listed in the correct format.  Looks like you may need to configure with your email to get our bot to process it.

> Thank you for your pull request and welcome to our community. We could not parse the GitHub identity of the following contributors: **Tim Leonard**. This is most likely caused by a git client misconfiguration; please make sure to:
> 
> 1. check if your git client is configured with an email to sign commits `git config --list | grep email`
> 2. If not, set it up using `git config --global user.email email@example.com`
> 3. Make sure that the git commit email is configured in your GitHub account settings, see https://github.com/settings/emails

**Comment by timle2** [comment]
> TODO: [#4220 (comment)](https://github.com/dbt-labs/dbt-core/pull/4220#discussion_r793784822)
> 
> > We've updated how to we log to use the Events modules. You'll need to create a new log event for this, and everywhere else you log, and use `fire_event` to kick off writing the log. We have a [`README`](https://github.com/dbt-labs/dbt-core/blob/main/core/dbt/events/README.md) with some good info on our logging system.

✅ Switched all the logging over to the Events / fire_event. 

> TODO [#4220 (comment)](https://github.com/dbt-labs/dbt-core/pull/4220#discussion_r793801205)
> 
> > As a standard we don't use `assert` in our code other than for tests. They shadow the actual exceptions and throw an `AssertionError` instead. Please rewrite the assert statements to catch relevant exceptions (create new ones if needed!).
> > In this specific case, I do like the use `is_tarfile` to check the file! Can you think of a reason it would be worth retrying the download if the file is not a valid tarfile?

✅ Moved over to properly defined exceptions. For the most part leveraging dbt.exceptions.DependencyException. 

`Can you think of a reason it would be worth retrying the download if the file is not a valid tarfile?` No I think straight failure without retry makes the most sense. In almost all cases the user would want to know that the tarfile is invalid, with `dbt deps` failing, as opposed to hitting the server multiple times to confirm. A transmission error could result in a corrupted tar file, however this is such an edge case I'm not sure it's worth guarding against.

**Comment by emmyoop** [comment]
> `Can you think of a reason it would be worth retrying the download if the file is not a valid tarfile?` No I think straight failure without retry makes the most sense. In almost all cases the user would want to know that the tarfile is invalid, with `dbt deps` failing, as opposed to hitting the server multiple times to confirm. A transmission error could result in a corrupted tar file, however this is such an edge case I'm not sure it's worth guarding against.

I agree that this is rare, but it does happen.  The user will still find out the tar file is invalid if that is indeed the issue and not a transmission error we could recover from.  This will be rare but worth protecting against.  In dbt Cloud deps are installed every time a job runs so we want to ensure if there is a transmission error we can recover from with a simple retry, we do so as not to fail an entire job.

Unrelated: You'll want to pull in the latest on the main branch as we've added some auto formatting/checking and that is why you have a failing test.  You can read more about it [here](https://github.com/dbt-labs/dbt-core/blob/main/CONTRIBUTING.md#pre-commit).

File: core/dbt/events/types.py [hunk 1]
```diff
@@ -2384,6 +2384,30 @@ def message(self) -> str:
         return "Internal event buffer full. Earliest events will be dropped (FIFO)."
 
 
+@dataclass
+class Sha1ChecksumPasses(DebugLevel):
+    code: str = "Z049"
```
emmyoop, line 2389: Love that you added some log lines!  Let's label these as type 'M' instead of 'Z' since it's related to deps generation.

timle2: thanks for the direction here, made both of these changes

core/dbt/events/types.py
```diff
@@ -2384,6 +2384,30 @@ def message(self) -> str:
...
+
+    def message(self) -> str:
+        return f'sha1 checksum passes ({self.sha1})'
+
+
+@dataclass
+class UntarProjectRoot(DebugLevel):
+    subdirectory: str
+    tar_dir_name: str
+    code: str = "Z050"
```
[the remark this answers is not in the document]
timle2: thanks for the direction here, made these changes

core/dbt/deps/tarball.py
```diff
@@ -0,0 +1,238 @@
...
+        super().__init__(tarball)
+        self.sha1 = sha1
+        self.subdirectory = subdirectory
+        # self.tarfile_bin = self.get_tarfile()
+        self.get_tarfile()
+        ''' init tarfile in class, and use tarfile as cache.
+            This simplifies tempfile handling, and prevents multiple hits to
+              url, since we need the file for two seperate operations. One
+              for metadata fetch (in style of other packages) and once for
+              for install for install. '''
```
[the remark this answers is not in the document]
timle2: I've removed caching the tarfile as part of the TarballPinnedPackage, and now the url is hit twice (once to get metadata about the package, and a second time to download the package). It's only a minor changes to code, but if you think that's the right tradeoff, then no problem for me. 

For the more complex topic of reusing registry or git package code - 
I've spent quite a lot of time trying to recycle the registry code, prior to coming to the current solution. However there are some fundamental differences in how this is handled vs registry. Registry has all the lovely json metadata that coordinates the tar download, and naming (as well as enforcing tar file structure, and not allowing subdirectories. Nothing to recycle from registry client, since it's all loading and parsing the json. And the registry install is super simple, since registry api/json handles any unknowns. File source structure and destination name are taken care of, and there is no concept of package subdirectory.
Git provides a good 'cousin' example, in that it allows for subdirectories, and has no metadata source like registry. I've recycled the subdirectory scanning from the git code. However git uses clone and copy for the download and install, so none of the install routine can be copied here. 

Here's a *rough* sketch comparing registry install vs tarball install. Let me know where you have any ideas on streamlining the code between the two. My approach so far has been built on the premise that the registry metadata creates a very simple solution, and so the registry install routine is very straight forward, but the registry install routine overfits to scenario of not needing to 'discover' any metadata. Tarball, on the other hand, given added unknowns from not having metadata, becomes unique in having to discover the metadata from the file structure, and then in turn use that to unpack the package in an intelligent way. Additionally we are doing tarfile verification, sha1 verification, and size verification, which isn't happening for any of the registry (or git, or local) routine, so new code having to be added there. 
Untitled drawing

Phew! ok lots there, but my tldr is please let me know where you see opportunity to streamline/recycle from registry install routine. I've given this a lot of thought already, and do not see any obvious patterns, aside from requiring a third style of install (tarball) that is more of a hybrid of of parts of registry and Git install, than overly similar to either.

core/dbt/deps/tarball.py
```diff
@@ -0,0 +1,249 @@
...
+            if not tarfile.is_tarfile(filepath):
+                msg = f"{filepath} is not a valid tarfile."
+                raise DependencyException(msg)
+
+        def handle_local_tar(self):
+            # pass through local tarfile path primarily for testing
+            validate_checksum(self, self.tarball)
+            validate_tarfile(self, self.tarball)
+            tar = tarfile.open(self.tarball, "r:gz")
+            self.tarfile_bin = tar
```
emmyoop, line 89: This is duplicated by part of the code in `handle_remote_tar` (#L102-L105).  Combining this will also remove the need for the nested functions as they won't be called more than once.

core/dbt/clients/system.py
```diff
@@ -478,6 +478,24 @@ def untar_package(tar_path: str, dest_dir: str, rename_to: Optional[str] = None)
...
+def resolve_tar_dir_name(TarFile: tarfile.TarFile):
+    tarfile_dir_list = [x.name for x in TarFile.getmembers() if x.isdir()]
+    tar_dir_name = os.path.commonpath(tarfile_dir_list)
+    return convert_path(tar_dir_name)
+
+
+def untar_tarfile(
+    TarFile: tarfile.TarFile, dest_dir: str, rename_to: Optional[str] = None
+) -> None:
+    tar_dir_name = dbt.clients.system.resolve_tar_dir_name(TarFile)
```
emmyoop, line 490: You can update `untar_package` to use `resolve_tar_dir_name` which will remove the need for the new `untar_tarfile`.   You would just pass through the tar path instead of the TarFile object.

core/dbt/deps/tarball.py
```diff
@@ -0,0 +1,249 @@
...
+    def download_tar(self, download_url, tar_path):
+        """
+        Sometimes the download of the files fails and we want to retry.  Sometimes the
+        download appears successful but the file did not make it through as expected
+        (generally due to a github incident).  Either way we want to retry downloading
+        and untarring to see if we can get a success.  Call this within
+        `_connection_exception_retry`
+        """
+
+        system.download(download_url, tar_path)
```
emmyoop, line 138: This doesn't need to be a function since all it does it this single line.  Directly call `system.download` instead.

core/dbt/deps/tarball.py
timle2: Ok - updated accordingly, see 1dcd97bbc8176060d10c849ed584550ab3f410bd

core/dbt/clients/system.py
timle2: Based on your feedback I've removed everything I initially wrote around passing the TarFile object around, in preference for using a tar path. I've removed this `untar_tarfile` function, and moved over to untar_package() as you suggest.

core/dbt/deps/tarball.py
timle2: makes perfect sense, updated accordingly.

the-chris-mitchell:
This would be quite useful. Do you still plan to finish this?

timle2:
Yes this is something my team would benefit from as well, and I'm glad others see the value in an additional install method for deps. 
I haven't had time to get this past the finish line unfortunately. 
I welcome any collaborators/input/help whatever it takes to get this approved though! 
At minimum, I'll be able to finish this in Sept, but obviously if others are interested I'd love to get this done sooner.

emmyoop:
@timle2 it's great that you're still interested in getting this over the finish line, even if it has to wait until September.  Thanks so much!  I did want to drop in to say that we have done quite a bit of work around dbt deps since you worked on this issue last.  It would be helpful to peruse the changes we've made to (including adding caching!) to ensure your approach matches the pattern we're now using with other methods.

timle2:
@emmyoop 
Based on feedback from revision 1 and 2, I've stripped this PR way down, to the abosolute minumum required to support the tarball feature.
I've removed features - nice but not required - such as supporting subfolders like git packages, and sha1 checks.
Now we lean on existing code from RegistryPinnedPackage (install() and download_and_untar()). This slims down the PR tremendously!
However, RegistryPinnedPackage install() and download_and_untar() are not currently set up as functions that can be used across classes, and I'm not smart enought to set up some crazy class inheritance structure to get it done. So I've copy/pasted the fns from RegistryPinnedPackage to TarballPinnedPackage. I'd advocate for moving these fns to dbt.deps.base, or to a dbt.deps.common file, but I also don't feel that I'm the person to make that call - need some feedback from dbt maintainers here on how to move forward based dbt labs thoughts.

timle2:
Thank you  @the-chris-mitchell for sharing support on this. I've picked up 

> @timle2 it's great that you're still interested in getting this over the finish line, even if it has to wait until September. Thanks so much! I did want to drop in to say that we have done quite a bit of work around dbt deps since you worked on this issue last. It would be helpful to peruse the changes we've made to (including adding caching!) to ensure your approach matches the pattern we're now using with other methods.

done and done! totally rewrote from main branch :)

timle2:
Thank you @the-chris-mitchell  for sharing support on this. I've picked it back up and hoping to get across finish line in the next few weeks.

the-serious-programmer:
Would love to see this released! :) (Also see my comment here: https://github.com/dbt-labs/dbt-core/issues/4205#issuecomment-1293401553)

core/dbt/deps/tarball.py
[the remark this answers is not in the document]
timle2: I've just finished a total rewrite with this `If this behaved the same as the registry package more of the code could be shared to have a single path to follow.` in mind. So ready for a new review :)

core/dbt/deps/base.py
```diff
@@ -97,6 +99,48 @@ def get_installation_path(self, project, renderer):
     def get_subdirectory(self):
         return None
 
+    def _install(self, project, renderer):
+        """
+        verbatim from dbt.deps.registry.RegistryPinnedPackage
+            install() from RegistryPinnedPackage class
+            Q for dbt labs, is this an appropriate pattern to refactor against?
```
timle2, line 106: Questions for dbt labs, is this an appropriate pattern to refactor against?
Fns used upstream in RegistryPinnedPackage and TarballPinnedPackage via inheritance (originally from from dbt.deps.registry.RegistryPinnedPackage)

core/dbt/deps/registry.py
```diff
@@ -60,32 +57,7 @@ def _fetch_metadata(self, project, renderer) -> RegistryPackageMetadata:
         return RegistryPackageMetadata.from_dict(dct)
 
     def install(self, project, renderer):
-        metadata = self.fetch_metadata(project, renderer)
```
timle2, line 63: moved upstream to PinnedPackage class so they can be shared across TarballPinnedPackage and RegistryPinnedPackage via PinnedPackage inheritance.

timle2:
@emmyoop this is ready for another run against the git workflows for tests/approvals, can you please trigger? Previously failed due to formatting errors, which are fixed now (bf43d1f).

timle2:
@emmyoop something funny going on with GitHub actions today? Last git workflow run failed with the following:
image

dbeatty10:
@timle2 I think branches that start with `feature/` have a special meaning for us.

Can you rename it to something else (like `timle2/dbt-deps-tarball`) and see if it resolves the issue you are seeing with `pre-commit`?

timle2:
> @timle2 I think branches that start with `feature/` have a special meaning for us.
> 
> Can you rename it to something else (like `timle2/dbt-deps-tarball`) and see if it resolves the issue you are seeing with `pre-commit`?

Thanks for the help here. I've got a second PR with the altered branch name here:
https://github.com/dbt-labs/dbt-core/pull/6302

timle2:
> > @timle2 I think branches that start with `feature/` have a special meaning for us.
> > Can you rename it to something else (like `timle2/dbt-deps-tarball`) and see if it resolves the issue you are seeing with `pre-commit`?
> 
> Thanks for the help here. I've got a second PR with the altered branch name here, see if that works. #6302 No way for me to change the branch name here without closing the PR it seems.

Actually this PR working fine now, yay! Back to getting flake errors like I'm used to 😆

dbeatty10:
@timle2 Something for you to do:
1. Add this as the first line of your PR description:
    - `resolves #4205`
    - The ["resolves" keyword](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword) will link the PR to the issue and will automatically close the latter at the appropriate time.
    - Having it as the first line is just a convention we used and not strictly required (but appreciated!)

core/dbt/task/deps.py
```diff
@@ -37,10 +37,13 @@ def track_package_install(
         self, package_name: str, source_type: str, version: Optional[str]
     ) -> None:
         # Hub packages do not need to be hashed, as they are public
-        # Use the string 'local' for local package versions
+        # Use the string 'local' for local and 'tarball' package versions
```
emmyoop, line 40: This comment doesn't really add value, you can delete the entire line instead of modifying it.

core/dbt/deps/base.py
emmyoop: Sure is!  Everywhere else that inherits from `PinnedPackage` has a custom `install` method.

emmyoop (commented):
@timle2 this looks great! 🎉   Just had a few small comments.  

I wanted to give you some feedback now since I know it was overdue!  Before I officially approve and merge this in I wanted to do some manual testing on it.  I'll let you know if I find anything testing and if not I'll get this merged in once you finish a bit of cleanup.

core/dbt/deps/tarball.py
```diff
@@ -0,0 +1,249 @@
...
+            self.tarfile_bin = self.tarball
+        else:
+            raise ValueError('unknown tarball type defined')
+
+        if not self.get_tar_size() <= TARFILE_MAX_SIZE:
+            msg = ("Tarfile size is larger than limit "
+                   f"of {TARFILE_MAX_SIZE}.")
+            raise DependencyException(msg)
+
+            fire_event(Sha1ChecksumPasses())
```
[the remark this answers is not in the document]
timle2: This (and the sha1 checks) were removed as of [25ee2c6](https://github.com/dbt-labs/dbt-core/pull/4689/commits/25ee2c69712186dc3184fa966b66a318bd82e37b). There's no matches for `fire_event(Sha1ChecksumPasses())` in the current revision. Looks like this may just be based on an outdated snippet, but let me know if I've missed something!

core/dbt/deps/tarball.py
```diff
@@ -0,0 +1,90 @@
...
+
+    @property
+    def name(self):
+        return self.package
+
+    def get_version(self):
+        return self.version
+
+    def nice_version_name(self):
+        return "version {}".format(self.version)
```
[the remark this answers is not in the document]
timle2: ty! will keep this in mind going forward.

core/dbt/deps/base.py
timle2: ty for feedback, will close this comment and remove from code

core/dbt/task/deps.py
timle2: removed with 3d48f6ad612299db7d3d92464d2df1a518f991f7

timle2:
> @timle2 this looks great! 🎉 Just had a few small comments.
> 
> I wanted to give you some feedback now since I know it was overdue! Before I officially approve and merge this in I wanted to do some manual testing on it. I'll let you know if I find anything testing and if not I'll get this merged in once you finish a bit of cleanup.

🥳 Thanks @emmyoop! Thanks for the support through this PR, greatly appreciated. I've addressed all the comments, but please let me know if I've missed anything. 

Flake/Black passing on my local machine, so this should be good to re-run workflows for another round of validation.

dbeatty10:
I'll do some manual testing on this and report back here afterwards.

Mathyoub:
@dbeatty10 and I did some testing and the happy path works great! But, we would like to request some improvements or at least raise some observations for discussion. 

Because version is optional and can lead to misleading output we would recommend removing it altogether. In the case of a user specifying a different version than the actual tarball, we will see a package.yml that looks like this: 
```
packages:
  - tarball: https://codeload.github.com/dbt-labs/dbt-utils/tar.gz/0.6.5
    name: 'dbt_utils_065'
    version: 0.9.6
```
Produce an output like this:
```
23:43:57  Installing dbt-labs/dbt_utils
23:43:58    Installed from version 0.9.6
23:43:58    Up to date!
```
Where `Installed from version 0.9.6` is actually incorrect. Another clear example would be passing version but not specifying version and getting `Installed from version None`.

Another possibly improvement can be seen when a user give a name, no version, but tries something else like `revision:` or maybe a typo `versoin:` the output is not exactly user friendly and returns a stack trace that doesn't help the user much. Example: 
```
packages:
  - tarball: https://codeload.github.com/dbt-labs/dbt-utils/tar.gz/0.6.5
    name: 'dbt_utils_065'
    versoin: 0.9.6
```
Gives us an output: 
```
00:04:17  Encountered an error:
'version'
00:04:17  Traceback (most recent call last):
  File "/Users/dbeatty/projects/dbt-core-4689/core/dbt/main.py", line 135, in main
    results, succeeded = handle_and_check(args)
  File "/Users/dbeatty/projects/dbt-core-4689/core/dbt/main.py", line 198, in handle_and_check
    task, res = run_from_args(parsed)
  File "/Users/dbeatty/projects/dbt-core-4689/core/dbt/main.py", line 225, in run_from_args
    task = parsed.cls.from_args(args=parsed)
  File "/Users/dbeatty/projects/dbt-core-4689/core/dbt/task/deps.py", line 99, in from_args
    return super().from_args(args)
  File "/Users/dbeatty/projects/dbt-core-4689/core/dbt/task/base.py", line 109, in from_args
    config = cls.ConfigType.from_args(args)
  File "/Users/dbeatty/projects/dbt-core-4689/core/dbt/config/runtime.py", line 613, in from_args
    project, profile = cls.collect_parts(args)
  File "/Users/dbeatty/projects/dbt-core-4689/core/dbt/config/runtime.py", line 208, in collect_parts
    project = cls.collect_project(args=args, project_renderer=project_renderer)
  File "/Users/dbeatty/projects/dbt-core-4689/core/dbt/config/runtime.py", line 246, in collect_project
    project = partial.render(project_renderer)
  File "/Users/dbeatty/projects/dbt-core-4689/core/dbt/config/project.py", line 286, in render
    return self.create_project(rendered)
  File "/Users/dbeatty/projects/dbt-core-4689/core/dbt/config/project.py", line 408, in create_project
    packages = package_config_from_data(rendered.packages_dict)
  File "/Users/dbeatty/projects/dbt-core-4689/core/dbt/config/project.py", line 106, in package_config_from_data
    PackageConfig.validate(packages_data)
  File "/Users/dbeatty/projects/dbt-core-4689/core/dbt/contracts/project.py", line 103, in validate
    if not package["version"]:
KeyError: 'version'
```
This case could be handled similarly to how this resolves:
```
packages:
	- git: "https://github.com/dbt-labs/dbt-utils.git"
	  version: 1.0.0
```
```
16:40:47  Encountered an error while reading the project:
16:40:47  ERROR: Runtime Error
	The packages.yml file in this project is malformed. Please double check
	the contents of this file and fix any errors before retrying.

	You can find more information on the syntax for this file here:
	https://docs.getdbt.com/docs/package-management

	Validator Error:
	{'git': 'https://github.com/dbt-labs/dbt-utils.git', 'version': '1.0.0'} is not valid under any of the given schemas

Error encountered in /Users/matthewbeall/dbt-projects/propmgmt/dbt_project.yml
16:40:47  Encountered an error:
Runtime Error
	Could not run dbt
```
In this example revision is expected instead of version and gives a nice message about where and what the error is as well as reference for syntax.

We should also ensure [this doc](https://docs.getdbt.com/docs/build/packages#how-do-i-add-a-package-to-my-project)  is updated eventually.

emmyoop:
@Mathyoub thanks for the great testing and feedback!  I agree that `version` isn't necessary.  At a quick glance I don't see why the TarballPackage isn't hooked in to validation, but we definitely want to get the helpful validation error on invalid package definitions.

@timle2 I just want to add that you only need to open an issue for a new feature [here](https://github.com/dbt-labs/docs.getdbt.com/issues/new/choose) and a PM will get the docs written up.

However, adding info on `tarball.py` to the [README](https://github.com/dbt-labs/dbt-core/blob/main/core/dbt/deps/README.md) would be perfect.

dbeatty10:
> @timle2 I just want to add that you only need to open an issue for a new feature [here](https://github.com/dbt-labs/docs.getdbt.com/issues/new/choose) and a PM will get the docs written up.

I just created that issue [here](https://github.com/dbt-labs/docs.getdbt.com/issues/2474) so one less thing that needs to be done.

Updated the PR description accordingly.

timle2:
@Mathyoub Really appreciate the feedback here, and very much appreciate forcing me off the `happy path` 😄 

I'm submitting fixes to your two points with 7cb8a4ab5e3bb5c33bdfa72e28ea7214a8e3fd99. More context on each below: 

1) **No need for version** Totally agree. The registry install (now moved to `dbt.deps.base PinnedPackage _install`) expects a version entry, so that's why I included it originally.  
To address your comments I've opted to set the version to in the tarball class as "". I'm reluctant to remove `version` and `get_version` from the tarball class, since  there's a bunch of places dbt expects the pinned packaged class to return a usable .version or .version_name, so I'm inclined to find a solution that leaves it in the class (e.g. version = ""). For example `dbt.deps.base PinnedPackage _install`, or below
image


2) **Bad stack trace no good 100%** Great news, the changes from (1) fixed it!  
There is something brittle about using optional entries in core.dbt.contracts.project (and my shallow understanding of the full dbt code base).  
Somehow, the way I set it up, caused that `line 103, in validate \n\t if not package["version"]:` error, instead of the nicer `MALFORMED_PACKAGE_ERROR` you referenced.  
It's a moot point though: removing `version` (as in 1) -> removing the optional entry from `core.dbt.contracts.project TarballPackage` -> `MALFORMED_PACKAGE_ERROR` being properly called. So yeah, fixed (in the context of this PR at least)! 

**Now working examples of the cases you raised:**
*happy path*
Note: using the packages.yml tarball url now in tarball.get_version so that this message returns something informative to the user.
image

*duplicate project names in packages.yml*
image

*version removed*
Note: helpful error appears, is as expected.
image

*unexpected field added*
Note: helpful error appears, is as expected.
image

timle2:
>

timle2:
> @timle2 I just want to add that you only need to open an issue for a new feature [here](https://github.com/dbt-labs/docs.getdbt.com/issues/new/choose) and a PM will get the docs written up.
> 
> However, adding info on `tarball.py` to the [README](https://github.com/dbt-labs/dbt-core/blob/main/core/dbt/deps/README.md) would be perfect.

@emmyoop Thanks for flagging this! I've added/altered the readme with dc8ba0a30cc042df8d5dc9e166990cfba607ab75

core/dbt/deps/base.py
```diff
@@ -97,6 +99,34 @@ def get_installation_path(self, project, renderer):
     def get_subdirectory(self):
         return None
 
+    def _install(self, project, renderer):
+        metadata = self.fetch_metadata(project, renderer)
+
+        tar_name = "{}.{}.tar.gz".format(self.package, self.version)
```
iknox-fa, line 105: Minor detail:  We prefer using the more modern [f-string style python string formatting](https://peps.python.org/pep-0498/).

core/dbt/deps/base.py
```diff
@@ -97,6 +99,34 @@ def get_installation_path(self, project, renderer):
...
+        tar_path = os.path.realpath(os.path.join(get_downloads_path(), tar_name))
```
iknox-fa, line 106: Minor detail:  We're trying to avoid using `os.path` for new commits to dbt-core (preferring [Pathlib](https://docs.python.org/3/library/pathlib.html) instead).  THis is due to the many odd caveats when running in non-posix environments.

iknox-fa (approved):
LGTM other than the few mentioned minor details.

core/dbt/deps/base.py
timle2: Thanks Ian - I've changed it over to f-string with 8787ba41af585b954d18c24c2dccbf147035cd9f

core/dbt/deps/tarball.py
```diff
@@ -0,0 +1,74 @@
...
+
+    @property
+    def name(self):
+        return self.package
+
+    def get_version(self):
+        return self.version
+
+    def nice_version_name(self):
+        return f"Tarball (url: {self.tarball})"
```
Mathyoub, line 33: Can we have this be "tarball" instead of "Tarball"?

Mathyoub:
Changes look great @timle2! Just left one minor comment to lowercase "tarball" in the output!

core/dbt/deps/tarball.py
timle2: changed, thanks! 5410f1f6188bd87efebadc86c7d8cd9545aca744

timle2:
Tests failing with https://github.com/dbt-labs/dbt-core/pull/4689/commits/b06a66240a5d391b66577d298fc08fa43704f1d8 due to tests in test/unit/test_deps.py that still referenced a version input to tarball class. Fixed with https://github.com/dbt-labs/dbt-core/pull/4689/commits/e9f2121f9dc5cc2adc38082f635295beed321401. pytest passing now on local machine.

dbeatty10:
@timle2 we'll do some investigation and troubleshooting on our end for the two CI tests that aren't passing:
- Generate CLI API docs / check if generation needed
- Tests and Code Checks / integration test / python 3.8 / windows-latest

image
image

dbeatty10:
> @timle2 we'll do some investigation and troubleshooting on our end for the two CI tests that aren't passing:
> 
> * Generate CLI API docs / check if generation needed
> * Tests and Code Checks / integration test / python 3.8 / windows-latest

👍 After investigation, we are good from the perspective of CI tests.

### Justification
Confirmed with @emmyoop and @stu-k that `Generate CLI API docs` can be ignored for this PR. Alternatively, if you do fresh merge with `main`, it will skip this particular check in the future.

The windows-latest `Tests and Code Checks` was flaky and addressed by just manually re-running it. A future issue/PR combo with [these](https://github.com/dbt-labs/dbt-core/pull/6346/files#diff-5bab13c218ba16b33ff49936fb3b417cce1da20f705a4010d3e07d566100428c) updates should be able to address that flakiness.

timle2:
@emmyoop I think we are complete on the PR. Lmk if there is anything needed from me before Merge!

emmyoop (approved):
@timle2 thank you again for all your work (and persistence) on this!  This is such a great community contribution!  We'll be including it in the 1.4 release!

timle2:
> @timle2 thank you again for all your work (and persistence) on this! This is such a great community contribution! We'll be including it in the 1.4 release!

Amazing! Thanks to you as well for your patience and guiding me through all the first time contributor stuff. Looking forward to making some more (and much smoother) contributions in the future!

jtcohen6:
Nice work team !!

PR #4691, opened 2022-02-07
Initial pass at switching integration tests to pytest

### LINKED ISSUE #4690: [CT-169] Initial conversion of integration tests to pytest
We are converting our integration tests to pytest. This ticket is for the the initial creation of fixtures and utility functions that reproduce the functionality of the original test suite.

It will include the conversion of a handful of integration test directories.

gshank:
resolves #4690

### Description

This pull request is for preliminary review of the state of the integration test conversion to pytest.

### Checklist


tests/tables.py
```diff
@@ -0,0 +1,339 @@
...
+        self.unique_schema = unique_schema
+        self.default_database = database
+        # We need to get this from somewhere reasonable
+        self.quoting = {'database': True, 'schema': True, 'identifier': True}
+
+    def _assert_tables_equal_sql(self, relation_a, relation_b, columns=None):
+        if columns is None:
+            columns = self.get_relation_columns(relation_a)
+        column_names = [c[0] for c in columns]
+        sql = self.adapter.get_rows_different_sql(relation_a, relation_b, column_names)
```
jtcohen6, line 19: This is a funny method: It ultimately calls `COLUMNS_EQUAL_SQL`, a constant / Python f-string defined on the adapter class [here](https://github.com/dbt-labs/dbt-core/blob/2d0b975b6c2023cde219f0a045709a1fa5c6c840/core/dbt/adapters/base/impl.py#L1197-L1223), and reimplemented on other adapters ([e.g. spark](https://github.com/dbt-labs/dbt-spark/blob/63ebe042419a0feb5939c2eb1c010e7e52e8c802/dbt/adapters/spark/impl.py#L368-L400)). Why not call a macro, as other adapter Python methods do, and use our multiple dispatch pattern?

There are more examples of this further down, starting with `_assert_table_row_counts_equal`

tests/util.py
```diff
@@ -0,0 +1,115 @@
...
+from dbt.adapters.factory import get_adapter_by_type
+from dbt.events.test_types import IntegrationTestDebug
+from dbt.events.functions import fire_event
+
+
+# This is used in pytest tests to run dbt
+def run_dbt(args: List[str] = None, expect_pass=True):
+    # The logger will complain about already being initialized if
+    # we don't do this.
+    log_manager.reset_handlers()
```
jtcohen6, line 17: Some (all?) of this is holdover from "legacy" logging

tests/util.py
```diff
@@ -0,0 +1,115 @@
...
+    if os.path.exists(path):
+        with open(path, 'rb') as fp:
+            manifest_mp = fp.read()
+        manifest: Manifest = Manifest.from_msgpack(manifest_mp)
+        return manifest
+    else:
+        return None
+
+
+def run_sql_file(sql_path, unique_schema):
```
jtcohen6, line 39: My sense is that we've used this to:
- Supply fixture data - before the advent of `dbt seed`, and to avoid a dependency on that feature if unrelated to the current test
- Set up database state, when we need to test a specific scenario

Is this something we want to keep around / keep using? Are there alternative patterns to avoid splitting test setup across multiple files?

tests/functional/conftest.py
```diff
@@ -0,0 +1,268 @@
...
+    # in the test signature to a minimum. Also... database? ...?
+    return TestProjInfo(
+        project_root=project_root,
+        profiles_dir=profiles_root,
+        adapter=schema,
+        test_dir=request.fspath.dirname,
+        data_dir=data_dir,
+        test_schema=unique_schema,
+        # the following feels kind of fragile. TODO: better way of getting database
+        database=profiles_yml['test']['outputs']['default']['dbname'],
```
jtcohen6, line 267: This behavior is going to vary across adapters:
- Postgres cannot read from other databases
- Redshift cannot write to other databases (can sometimes read from others)
- BigQuery + Snowflake can read from + write to other databases/projects, it's important that we can test that functionality
- Spark/Databricks (currently) has no notion of a separate "database" level, it's just a synonym for `schema`

tests/functional/simple_copy/test_mixed_case_db.py
```diff
@@ -0,0 +1,50 @@
...
+
+def test_basic(project_root, project):
+
+    assert project.database == 'dbtMixedCase'
+
+    # Tests that a project with a single model works
+    results = run_dbt(['run'])
+    assert len(results) == 1
+    manifest = get_manifest(project_root)
+    assert 'model.test.model' in manifest.nodes
```
jtcohen6, line 48: This is neat!

tests/functional/simple_copy/test_copy_uppercase.py
```diff
@@ -0,0 +1,66 @@
+import pytest
+import os
+from tests.tables import TableComparison
+from tests.util import run_dbt
+
+from tests.functional.simple_copy.test_simple_copy import (
```
jtcohen6, line 6: This makes me think: What would it look like if an external plugin wanted to run `test_simple_copy`, exactly as written, with no changes? What would it look like if they wanted to run it, but needed to override a few of the fixtures?

tests/functional/simple_reference/test_invalid_reference.py
```diff
@@ -0,0 +1,28 @@
...
+    return {
+        'descendant.sql': descendant_sql,
+        'model.sql': model_sql,
+    }
+
+
+def test_undefined_value(project):
+    # Tests that a project with an invalid reference fails
+    with pytest.raises(CompilationException):
+        run_dbt(['compile'])
```
jtcohen6, line 28: - This test is much more modular than the others, and neatly maps to a functionality owned by the Language team. Is there some way we could tag this test with that team's ownership? Point back to the actual functionality it's testing?
- Related to the above: Would we want this to be `'parse'` or `'ls'`, rather than `'compile'`, since we'd expect this error to raise during parse time?

tests/functional/simple_reference/test_simple_reference.py
```diff
@@ -0,0 +1,238 @@
...
+        'view_using_ref.sql': view_using_ref_sql,
+    }
+
+
+@pytest.fixture
+def project_config_update():
+    return {
+        'vars': {
+            'test': {
+                'var_ref': '{{ ref("view_copy") }}',
```
jtcohen6, line 134: I don't see us using this anywhere below. I also wonder if we need to keep using dbt `--vars` for this kind of functionality? Could we instead update the pytest fixture (?) within the scope of a subsequent test? (I'm sure I'm misunderstanding fixture scope)

tests/functional/simple_copy/test_simple_copy.py
```diff
@@ -0,0 +1,217 @@
...
+        'advanced_incremental.sql': advanced_incremental_sql,
+        'compound_sort.sql': compound_sort_sql,
+        'disabled.sql': disabled_sql,
+        'empty.sql': empty_sql,
+        'get_and_ref.sql': get_and_ref_sql,
+        'incremental.sql': incremental_sql,
+        'interleaved_sort.sql': interleaved_sort_sql,
+        'materialized.sql': materialized_sql,
+        'schema.yml': schema_yml,
+        'view_model.sql': view_model_sql,
```
jtcohen6, line 142: Should we aim to consolidate these fixtures with the ones used in other tests? It would require more upfront work, but if we assembled a common set of standard fixtures (rather than recreating highly similar ones over and over again), then:
- each test can pick & choose just the ones it needs
- adapter plugins could reimplement specific fixtures (as needed) among the standard reused set, and then be able to get most tests "for free," rather than needing to reimplement similar fixtures in similar ways for each test

tests/CONVERTING.md
```diff
@@ -0,0 +1,37 @@
...
+  If you have teardown pieces in the fixture, do a 'yield' after the setup, and
+  the part after the 'yield' will be run at teardown.
+* 'run\_dbt', 'run\_sql', and 'get\_manifest' are provided by the tests/util.py file
+* You will probably want to make separate test files out of tests that use
+  substantially different projects. If they're only different by a file or two,
+  you could write out individual files instead and keep them in the same file.
+* You can also import file strings from other test cases
+* old: self.get\_models\_in\_schema, new: get\_tables\_in\_schema
+* somewhat easier way to get the legacy files into a more usable format:
+  ```tail -n +1 models/* > copy_models.out```
```
jtcohen6, line 29: Is the idea here that we can "print out" a set of pytest model fixtures into a functional set of dbt project files, to fully replicate end user experience of running dbt with a local file system? (Similar to what our integration tests do today)

jtcohen6 (commented):
This is really exciting!! Lots of comments as I read through the draft implementation.

As next steps, I'd be excited to:
- Try porting the two converted integration tests to an adapter plugin, and see how we'd need to reimplement fixtures/methods/etc
- Think about what other modular utilities we'd want, in a similar vein to `get_manifest`. I think @JCZuurmond's work on `pytest-dbt-core` could be generative here, especially [`macro_generator`](https://github.com/godatadriven/pytest-dbt-core/blob/29084e28e240cbf7964612675205ddef3b5c0af8/src/pytest_dbt_core/fixtures.py#L105-L131) to mock macro input + output. Could that do for "execution" code what `get_manifest` does for "parsing" code?

tests/util.py
gshank: Yes, this seems to be from the really early tests. It's a good idea to look at where it's used and see if we can do something different.

tests/functional/simple_reference/test_invalid_reference.py
gshank: Those sorts of changes sounds like an excellent idea. Not sure whether we want to do it when converting though, or as a post-conversion re-organization and cleanup.

tests/functional/simple_copy/test_copy_uppercase.py
gshank: I've been looking at options for sharing the fixtures, and what I found was using them as pytest plugins. I'm not fully up to speed on how that would work, but I think we would include the base conftest.py from the dbt-core repo (or a test fixture repo?), and the provide fixture overrides in a conftest.py in the adapter repos. Or something like that.

tests/functional/conftest.py
gshank: Yeah, this is one of the ragged edges. Some other ragged edges are getting the adapter in a good way, and the quoting config.

tests/tables.py
gshank: All of this code was extracted from test/integration/base.py. Quite a large chunk of that was dealing with the table comparisons, so I just copied, formatted and got it to work. I'm up for whatever better practices have evolved since that code was written, and at the least we would need to override this in adapter repos.

tests/functional/simple_copy/test_simple_copy.py
gshank: Ideally I think it would be nice to have some standard projects that can be reused. We have a lot of tests that create projects with only subtle or minimal differences. That can be done a directory basis, using a test directory conftest.py, with only the different parts overridden. We probably could do that for test_simple_copy.py and test_copy_upperbase, and just override the 'models' fixture for test_copy_upperbase. You can also make individual sets of project fixtures with different names and include them in a test as a plugin.

I think it will be easiest to do that after we've converted the tests though, since we'll then be able to more easily see the differences between test projects. The legacy tests with all the files in directories (often multiple) makes that sort of comparison harder.

tests/CONVERTING.md
gshank: The fixtures as currently implemented do write out a complete dbt project into the temporary project_root directory.  The 'tail' command is just a convenience to avoid copying and pasting files individually. I'm actually planning on opening a ticket for someone to create a script to write out the legacy files in a format that's usable for the pytest tests.

tests/functional/simple_copy/test_simple_copy.py
gshank: Though on second thought, if people see the same setup as they're going through converting, there's no reason to wait. "Oh, I saw this project before. Yank! It's now a plugin."

tests/functional/simple_reference/test_simple_reference.py
gshank: This is used in the 'view_using_ref.sql' file. The 'project_config_update' fixture is merged into the basic project_config in the fixtures code. So this var is actually in the dbt_project.yml.

core/dbt/tests/tables.py
```diff
@@ -0,0 +1,339 @@
+from dbt.tests.util import run_sql
+from dbt.context import providers
+from unittest.mock import patch
+from contextlib import contextmanager
+
+
+class TableComparison:
```
JCZuurmond, line 7: I would like to challenge this class a bit. The way I see it is that this "new way of testing" removes some of the "magic" of the "old way". The risk with using such a class is that magic is being introduced - in a different form.

JCZuurmond (commented):
Hi, super nice you are upgrading the testing framework! I added some comments, I am still thinking about how I would do this myself, view the comments as suggestions, probably some education is needed before I understand the intent of the changes.

It is great to remove all the `dbt` subprocess calls as this allows the user to start debugging sessions. 

Furthermore, it would be great to have a test suite which can be applied to the `core` and all the `adapters`. Especially, if it is written in such a way that adapters can easily extended it (for example with adapter specific settings) whilst reusing the test suite. 

Pytest recommends [two test layouts](https://docs.pytest.org/en/6.2.x/goodpractices.html#choosing-a-test-layout-import-rules): outside or inside the package. The most common set-up is having your tests outside of your package: this makes your package smaller. However, some packages don't do this (for example `numpy`). 

A benefit of having the tests inside your package is that you can run the test suite as a user. This could be helpful for the adapters:
1. Install `dbt-core`: `pip install dbt-core`
2. Run `dbt-core` test suite: `pytest --pyargs dbt`
If then the test suite is set-up to automatically pick up which adapter is installed, you can reuse that test suite for every adapter.

As mentioned in the comments, I don't think fixtures are usually shipped inside the package. If you want to reuse a fixture over different projects, I think the idea is to write a pytest plugin.

core/dbt/tests/tables.py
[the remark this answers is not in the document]
gshank: This is not "new way of testing". This is "old way of testing" copy-pasted to the new way so that we can convert the tests without so much pain. Ideally it would be converted to use macros and simplified a great deal. Eventually.

core/dbt/tests/util.py
```diff
@@ -0,0 +1,127 @@
...
+
+    finally:
+        stop_capture_stdout_logs()
+
+    return res, stdout
+
+
+# Used in test cases to get the manifest from the partial parsing file
+def get_manifest(project_root):
+    path = project_root.join("target", "partial_parse.msgpack")
```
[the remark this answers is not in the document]
gshank: The same function is working fine in our current CI. The target directory is a subdirectory of the temporary project_root directory that is created for every test.

core/dbt/tests/fixtures/project.py
```diff
@@ -0,0 +1,270 @@
...
+    return {}
+
+
+@pytest.fixture
+def snapshots():
+    return {}
+
+
+@pytest.fixture
+def tests():
```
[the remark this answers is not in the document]
gshank: Fixtures can be overridden in a number of places, like in individual test cases. We are still looking at how to connect up the current fixtures with the adapter testing.

tests/functional/basic/test_jaffle_shop.py
```diff
@@ -0,0 +1,15 @@
+import pytest
+from dbt.tests.util import run_dbt, get_manifest
+
+
+from tests.fixtures.jaffle_shop import models, seeds, project_config_update
```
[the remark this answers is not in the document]
gshank: You mean the 'models', 'seeds' and 'project_config_update'?  They are used. They override the base fixtures that come from the fixtures/project.py file.

core/dbt/tests/tables.py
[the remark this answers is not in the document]
JCZuurmond: Alright, let's forget about this class for now

core/dbt/tests/util.py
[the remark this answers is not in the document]
JCZuurmond: great!

tests/functional/basic/test_jaffle_shop.py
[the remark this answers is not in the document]
JCZuurmond: Flake gives shows the following warning:

> F401: 'tests.fixtures.jaffle_shop.seeds' imported but unused

I guess an `isort` or `black` would remove this line. 

Pytest recommends to [overwrite fixtures](https://docs.pytest.org/en/7.0.x/how-to/fixtures.html#overriding-fixtures-on-various-levels) by using `conftest`, within a test module or with direct test parameterization. By importing the fixtures it is _kinda_ like defining a fixture on test module level, but not according to the pytest overriding fixture way. And, flake does not like it.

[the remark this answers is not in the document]
gshank: Yeah, this isn't ideal from flake and mypy considerations. But the whole point here is to package up standard projects and use them in a single test. Putting them in a conftest.py defeats that because it would apply to all tests in the directory. If you can come up with a better way to import the fixtures that we need for a single project, I'd love to hear it.

core/dbt/tests/fixtures/project.py
```diff
@@ -0,0 +1,270 @@
+import os
```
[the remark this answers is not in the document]
gshank: This acts as a local plugin, and the idea here is that it could act as a plugin for the adapter tests too. Since the test fixtures and the dbt release are so tightly coupled, there would be overhead and coordination required to separate them out, and there aren't many advantages. Anybody using this file as a pytest plugin is also going to have to have dbt. Perhaps there will be considerations that make that necessary, but we're going to wait until we see how the adapter test rewrite plays out to make those decisions.

emmyoop (approved):
Looks good with the understanding that we'll make changes as needed as we learn more/establish patterns.

PR #4699, opened 2022-02-09
WIP: attempt to rewrite

ChenyuLInx:
resolves #

### Checklist


tests/functional/simple_selection/seed.sql
```diff
@@ -0,0 +1,119 @@
+create table {schema}.summary_expected (
```
ChenyuLInx, line 1: Seems like this file is duplicated across places

core/dbt/tests/tables.py
```diff
@@ -322,6 +322,8 @@ def _ilike(target, value):
 
 
 def get_tables_in_schema(schema):
+    # TODO the get_models_in_schema has a special case for snowflake, do we still need it
```
ChenyuLInx, line 325: Is this something we intentionally didn't do

tests/functional/simple_selection/test_graph_select.py
```diff
@@ -0,0 +1,132 @@
...
+# 4. created models become created tables, assertFalse might need to be updated
+# 5. assert_correct_schemas needs to be implemented
+
+# Question how to validate the tests actually correctly tested the intended case
+
+# idea of script to automate this:
+# 1. unittest2pytest can handle convert assert,
+# 2. probably need something like the steps described above to convert the sql files
+# 3. another pass to fix the functions and report remaining functions
+# run_sql_file means need to add create table function
```
ChenyuLInx, line 19: @gshank  what's your thoughts on this whole section

core/dbt/tests/tables.py
gshank: Yes. We're going to implement things in the adapter in a cleaner way (like overriding methods) instead of these hacky if tests.

tests/functional/simple_selection/seed.sql
gshank: Yes. We're going to be looking for opportunities to store these commonly repeated files elsewhere. There's a 'tests/fixtures' directory. Maybe a 'tests/data' directory?

gshank: Some of the seed.sql files might do different things, so we'll have to look at each one and see if it's a repeat.

tests/functional/simple_selection/test_graph_select.py
gshank: We don't need assertFalse. Just use 'assert' and a condition that evaluates how we need. We do not want to update the source('raw', 'seed') because that's a built-in dbt feature. If the 'seed' command is executed, the seed table should be there and that SQL should work. If unittest2pytest helps, that's great. Put something in the CONVERTING.md file describing it.

gshank: assert_correct_schema isn't used in any other tests that I can see, so putting it in this test case seems fine.

tests/functional/conftest.py
```diff
@@ -266,3 +267,9 @@ def project(
         # the following feels kind of fragile. TODO: better way of getting database
         database=profiles_yml['test']['outputs']['default']['dbname'],
     )
+
+
+@pytest.fixture
+def create_tables(data_dir, unique_schema):
```
gshank, line 273: Putting a 'create_tables' fixture into here is fine for now. Jeremy was suggesting that long-term we might want to look at using seeds for this, but that's a bunch more conversion work, so I'd wait until we see how many different files like this we have.

tests/functional/simple_selection/test_graph_select.py
gshank: Just fyi we generally should not usually have to update any schema references, because the original tests all used randomly generated schema names. This particular test did some config changes to the schema name in the SQL (a legitimate thing to test...) so it's special that way.

ChenyuLInx: For the source('raw', 'seed') part, I will get an error saying the table doesn't exist If I don't update it. What's the proper way to update it?

PR #4701, opened 2022-02-09
Add `--quiet` flag and `print` Jinja function

### LINKED ISSUE #3451: [CT-172] `--quiet` flag for run-operations
### Describe the feature
Currently, when you run `run-operation` you get the output of your macro _after_ a few lines of text, for example:

```
Running with dbt=0.19.1
[WARNING]: Configuration paths exist in your dbt_project.yml file which do not apply to any resources.
There are 2 unused configuration paths:
- seeds
- snapshots

{macro output here}
```

I'm proposing a `--quiet` flag for `run-operation` that would remove the `Running with dbt` and any other warnings, leaving just the output of the macro.

I've previously wrapped calls to `run-operation` in bash script which trims the first line from the output, however when warnings are triggered (like the example above) it's hard to tell in advance how many lines to trim.

### Who will this benefit?
Anyone who uses the codegen macro (and similar macros) outside of dbt cloud

Yes — I've poked around the logger code but couldn't see a straightforward way of doing this, but I'm open to trying with some input!

ehmartens:
resolves #3451 

### Description

This PR addresses the issue linked above following @emmyoop 's suggestions [here](https://github.com/dbt-labs/dbt-core/issues/3451#issuecomment-1031807152).

It creates a new global flag called `--quiet` that suppresses all non-exception output. Additionally it create a Jinja function called `print` whihc prints messages directly to stdout.

### Checklist

### DIFF
--- a/core/dbt/main.py
+++ b/core/dbt/main.py
@@ -1090,8 +1090,8 @@ def parse_args(args, cls=DBTArgumentParser):
         action='store_true',
         default=None,
         help='''
-        Supress all non-error logging during dbt execution. Output from
-        {{ print() }} macro are still displayed.
+        Suppress all non-error logging to stdout. Does not affect
+        {{ print() }} macro calls.
         '''
     )
 
--- a/core/dbt/main.py
+++ b/core/dbt/main.py
@@ -1084,6 +1084,17 @@ def parse_args(args, cls=DBTArgumentParser):
         '''
     )
 
+    p.add_argument(
+        '-q',
+        '--quiet',
+        action='store_true',
+        default=None,
+        help='''
+        Supress all non-error logging during dbt execution. Output from
+        {{ print() }} macro are still displayed.
+        '''
+    )
+
     subs = p.add_subparsers(title="Available sub-commands")
 
     base_subparser = _build_base_subparser()
--- a/test/integration/044_run_operations_tests/test_run_operations.py
+++ b/test/integration/044_run_operations_tests/test_run_operations.py
@@ -72,4 +72,5 @@ class TestOperations(DBTIntegrationTest):
 
     @use_profile('postgres')
     def test__postgres_print(self):
+        # Tests that calling the `print()` macro does not cause an exception
         self.run_operation('print_something')
--- a/test/integration/044_run_operations_tests/test_run_operations.py
+++ b/test/integration/044_run_operations_tests/test_run_operations.py
@@ -69,3 +69,7 @@ class TestOperations(DBTIntegrationTest):
     @use_profile('postgres')
     def test__postgres_access_graph(self):
         self.run_operation('log_graph')
+
+    @use_profile('postgres')
+    def test__postgres_print(self):
+        self.run_operation('print_something')
--- a/test/unit/test_flags.py
+++ b/test/unit/test_flags.py
@@ -115,7 +115,7 @@ class TestFlags(TestCase):
         os.environ['DBT_DEBUG'] = 'True'
         flags.set_from_args(self.args, self.user_config)
         self.assertEqual(flags.DEBUG, True)
-        os.environ['DBT_DEUBG'] = 'False'
+        os.environ['DBT_DEBUG'] = 'False'
         setattr(self.args, 'debug', True)
         flags.set_from_args(self.args, self.user_config)
         self.assertEqual(flags.DEBUG, True)
@@ -217,3 +217,10 @@ class TestFlags(TestCase):
         os.environ.pop('DBT_INDIRECT_SELECTION')
         delattr(self.args, 'indirect_selection')
         self.user_config.indirect_selection = None
+
+        # quiet
+        self.user_config.quiet = True
+        flags.set_from_args(self.args, self.user_config)
+        self.assertEqual(flags.QUIET, True)
+        # cleanup
+        self.user_config.quiet = None
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 - Added Support for Semantic Versioning ([#4644](https://github.com/dbt-labs/dbt-core/pull/4644))
 - New Dockerfile to support specific db adapters and platforms.  See docker/README.md for details ([#4495](https://github.com/dbt-labs/dbt-core/issues/4495), [#4487](https://github.com/dbt-labs/dbt-core/pull/4487))
 - Allow unique_key to take a list ([#2479](https://github.com/dbt-labs/dbt-core/issues/2479), [#4618](https://github.com/dbt-labs/dbt-core/pull/4618))
+- Add `--quiet` global flag and `print` Jinja function ([#3451](https://github.com/dbt-labs/dbt-core/issues/3451), [#4701](https://github.com/dbt-labs/dbt-core/pull/4701))
 
 ### Fixes
 - User wasn't asked for permission to overwite a profile entry when running init inside an existing project ([#4375](https://github.com/dbt-labs/dbt-core/issues/4375), [#4447](https://github.com/dbt-labs/dbt-core/pull/4447))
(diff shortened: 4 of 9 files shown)


core/dbt/main.py
```diff
@@ -1084,6 +1084,17 @@ def parse_args(args, cls=DBTArgumentParser):
         '''
     )
 
+    p.add_argument(
+        '-q',
```
nathaniel-may, line 1088: I'm fine with this, but I want to highlight the short argument for other reviewers to make sure we want that in addition to the long `--quiet`.

core/dbt/main.py
```diff
@@ -1084,6 +1084,17 @@ def parse_args(args, cls=DBTArgumentParser):
...
+        '--quiet',
+        action='store_true',
+        default=None,
+        help='''
+        Supress all non-error logging during dbt execution. Output from
+        {{ print() }} macro are still displayed.
```
nathaniel-may, line 1094: I just wanted to be clear about the stdout part, fixed a spelling mistake, and shifted wording around a bit. If anyone has a better way to phrase this I'm down for that too. 
```suggestion
        Suppress all non-error logging to stdout. Does not affect 
        {{ print() }} macro calls.
```

test/integration/044_run_operations_tests/test_run_operations.py
```diff
@@ -69,3 +69,7 @@ def test__postgres_select(self):
     @use_profile('postgres')
     def test__postgres_access_graph(self):
         self.run_operation('log_graph')
+
+    @use_profile('postgres')
+    def test__postgres_print(self):
```
nathaniel-may, line 74: I'd like a comment on this test describing precisely what we're testing. In this case I think it's just checking that calling the print macro doesn't blow up, which imo is good enough test coverage for me.

test/unit/test_flags.py
```diff
@@ -115,7 +115,7 @@ def test__flags(self):
         os.environ['DBT_DEBUG'] = 'True'
         flags.set_from_args(self.args, self.user_config)
         self.assertEqual(flags.DEBUG, True)
-        os.environ['DBT_DEUBG'] = 'False'
+        os.environ['DBT_DEBUG'] = 'False'
```
nathaniel-may, line 118: oh nice, good catch.

nathaniel-may (changes requested):
Thanks for your work here, @ehmartens! This looks like exactly the right implementation. I just left a few minor comments, and I'd like to get @emmyoop's eyes on it too before merging but otherwise this looks real good to me.

Blockers:
- Just the spelling fix in the help message for the new flag.

emmyoop (commented):
Agree with @nathaniel-may, this looks great!  Thanks for your work on this and opening up an issue over in `codegen`.

I cloned this locally to play with it and have one concern.  

I have this useless macro that I stuck into a `pre_hook` for a model for testing purposes.

```
{% macro do_it() %}

    {{log('********************This is a LOG line!', True)}}
    {{print('********************This is a PRINT line!')}}

{% endmacro %}
```
It works as expected. `dbt -quiet run` prints the `print()` statement and nothing else but the logs all still go to the file.  Great!

My concern is that when I don't use the `-quiet` flag the print statement still comes out.

```
(env) emily@Emily-Rockman basic-dbt % dbt run
19:12:19  Running with dbt=1.0.1
19:12:19  ********************This is a LOG line!
********************This is a PRINT line!
19:12:19  Found 5 models, 7 tests, 0 snapshots, 0 analyses, 354 macros, 0 operations, 1 seed file, 0 sources, 1 exposure, 0 metrics
19:12:19
19:12:19  Concurrency: 4 threads (target='dev')
19:12:19
19:12:19  1 of 5 START table model dbt_testing.even_id_table.............................. [RUN]
19:12:19  2 of 5 START view model dbt_testing.male_list_view.............................. [RUN]
19:12:19  3 of 5 START table model dbt_testing.my_first_dbt_model......................... [RUN]
19:12:19  ********************This is a LOG line!
********************This is a PRINT line!
19:12:20  2 of 5 OK created view model dbt_testing.male_list_view......................... [CREATE VIEW in 0.31s]
19:12:20  3 of 5 OK created table model dbt_testing.my_first_dbt_model.................... [SELECT 2 in 0.32s]
19:12:20  1 of 5 OK created table model dbt_testing.even_id_table......................... [SELECT 50 in 0.33s]
19:12:20  4 of 5 START view model dbt_testing.my_second_dbt_model......................... [RUN]
19:12:20  5 of 5 START view model dbt_testing.clean_table................................. [RUN]
19:12:20  4 of 5 OK created view model dbt_testing.my_second_dbt_model.................... [CREATE VIEW in 0.07s]
19:12:20  5 of 5 OK created view model dbt_testing.clean_table............................ [CREATE VIEW in 0.07s]
19:12:20
19:12:20  Finished running 3 view models, 2 table models in 0.61s.
19:12:20
19:12:20  Completed successfully
19:12:20
19:12:20  Done. PASS=5 WARN=0 ERROR=0 SKIP=0 TOTAL=5
```

This completely makes sense for how I said to implement it but I'm wondering if it's really desired behavior here.

jaypeedevlin:
@emmyoop Chiming in on the expected behavior her (I paired with Elizabeth on this work this morning) and correct me if I'm wrong:  To use the codegen example, if we replaced all the `log()`s with `print()`s, users would _have_ to use `--quiet` to get the output.  In my mind at least, that's probably not the behavior we want, but I will (of course) to defer to your collective judgement.

nathaniel-may:
@emmyoop thanks for the example output! Although it looks kind of awkward, I think this is the correct behavior for the `--quiet` flag. If we think there's a use case for turning off print statements, I might suggest another flag like `--no-print` so each user can run with the exact behavior they're looking for.

emmyoop:
@jaypeedevlin I _completely_  agree that the behavior when using the `--quiet` flag is right.

My concern is macros using `print()` will always have whatever they're printing come to stdout.  This is more of a concern for macros outside `codegen` or really any macros meant to be run with `run-operation`.  I was just surprised (though I should not have been) when I saw stdout muddied a bit with the print statement.  I think the way I tested the print statement is not the intended so maybe the right solution is just some docs pointing out the behavior.

core/dbt/main.py
```diff
@@ -1084,6 +1084,17 @@ def parse_args(args, cls=DBTArgumentParser):
...
+        '''
```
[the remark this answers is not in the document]
ehmartens: @emmyoop I updated this to the suggestion that @nathaniel-may made [here](https://github.com/dbt-labs/dbt-core/pull/4701#discussion_r802972271). Apologies if I should have waited!

core/dbt/main.py
ehmartens: Updated!

core/dbt/main.py
emmyoop: I like short arguments so I'm happy with this!

test/integration/044_run_operations_tests/test_run_operations.py
emmyoop: Agree that a comment here would be helpful.

ehmartens: Added!

nathaniel-may (approved):
Excellent work here. Thanks for contributing!

nhammad:
Currently, I get an output that looks like this:
```
dbt run-operation generate_model_yaml --args "{\"model_name\": "bookings"}"  > test.yml  
```
```
12:53:32  Running with dbt=1.0.1
12:53:34  version: 2

models:
  - name: bookings
    description: ""
    columns:
      - name: booking_id
        description: ""

      - name: masterclient_id
        description: ""
```
Can I use the --quite argument to get rid of the timestamps etc? I tried this:
```
dbt run-operation generate_model_yaml --quiet --args "{\"model_name\": \"$table\"}" > $target_dir/$table.yml
```
but:
```
dbt: error: unrecognized arguments: --quiet
```

Issue #4566, opened 2022-01-12
[CT-36] [Bug] Airflow v2.1, v2.2 constraints incompatible with dbt v1.0.0, v1.0.1 on codependency click.


twilly:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

NOTE: This is a minor bug regarding a common integration.

Airflow v2.1 and v2.2 transitively depends on click>=7.1,<9. Airflow [requires Flask v1.1](https://github.com/apache/airflow/blob/8f6c921e3cbcba0f46083ec6a55272ea70d8139f/setup.cfg#L105) and Flask v1.1.3+ further constrains click to >=5.1,<8.0.

dbt-core 1.0.0. and 1.0.1 depends on click>=8,<9 from commit https://github.com/dbt-labs/dbt-core/commit/11436fed45730cec392872cacfc511bbe7102ec3.

The combined click constraint of >=8,<8.0 is unsolvable, preventing the installation of Airflow with dbt v1.0.0 and v1.0.1.

### Expected Behavior

The behavior is correct; it's just a nightmare to deal with. IMHO, it can be better by supporting click v7 in dbt v1.0.

### Steps To Reproduce

Try installing the latest release of dbt-core and Flask together:

`$ pip install dbt-core==1.0.1 Flask==1.1.4`

### Relevant log output

```shell
ERROR: Cannot install dbt-core==1.0.1 and flask==1.1.4 because these package versions have conflicting dependencies.

The conflict is caused by:
    dbt-core 1.0.1 depends on click<9 and >=8
    flask 1.1.4 depends on click<8.0 and >=5.1
```
```


### Environment

```markdown
- Python: 3.9
- dbt: 1.0.0, 1.0.1
```

### Additional Context

Apache Airflow [provides constraint files of known-working configuration](https://airflow.apache.org/docs/apache-airflow/stable/installation/installing-from-pypi.html#constraints-files) and has published constraints that pin Flask to 1.1.4 (e.g. [v2.1, Py 3.9](https://github.com/apache/airflow/blob/6a0b26641f723d444298f07811e95d4369b1efe7/constraints-3.9.txt#L25)). More recent constraints with Airflow v2.2 pin Flask down to v1.1.2 ([v2.2, Py 3.9](https://github.com/apache/airflow/blob/62d490d4da17e35d4ddcd4ee38902a8a4e9bbfff/constraints-3.9.txt#L25)).

[...]

PR #4736, opened 2022-02-17
Updating CHANGELOG duplicate 1.0.2 section

leahwicz:
### Description

We had a duplicate `1.0.2` section in the CHANGELOG and didn't add the section for the RC yet so just fixing that up

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,14 +19,18 @@
 - Drop support for Python 3.7.0 + 3.7.1 ([#4584](https://github.com/dbt-labs/dbt-core/issues/4584), [#4585](https://github.com/dbt-labs/dbt-core/pull/4585), [#4643](https://github.com/dbt-labs/dbt-core/pull/4643))
 - Re-format codebase (except tests) using pre-commit hooks ([#3195](https://github.com/dbt-labs/dbt-core/issues/3195), [#4697](https://github.com/dbt-labs/dbt-core/pull/4697))
 - Add deps module README ([#4686](https://github.com/dbt-labs/dbt-core/pull/4686/))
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -19,36 +19,33 @@
 - Drop support for Python 3.7.0 + 3.7.1 ([#4584](https://github.com/dbt-labs/dbt-core/issues/4584), [#4585](https://github.com/dbt-labs/dbt-core/pull/4585), [#4643](https://github.com/dbt-labs/dbt-core/pull/4643))
 - Re-format codebase (except tests) using pre-commit hooks ([#3195](https://github.com/dbt-labs/dbt-core/issues/3195), [#4697](https://github.com/dbt-labs/dbt-core/pull/4697))
 - Add deps module README ([#4686](https://github.com/dbt-labs/dbt-core/pull/4686/))
+- Support click versions in the v7.x series ([#4681](https://github.com/dbt-labs/dbt-core/pull/4681))
```
leahwicz, line 22: This change was under the `1.0.2` section which I didn't think was right (it hasn't been backported) so I moved it up to be under `1.1.0`. @jtcohen6 just checking you didn't want this in `1.0.2`

jtcohen6: Thanks for the tag. I would like this in `1.0.3`. Reasoning originally stated in https://github.com/dbt-labs/dbt-core/issues/4566#issuecomment-1024375260:

> there's good cause to make it in a v1.0 patch release, so that folks can upgrade to v1.0 for use with Airflow, and don't need to wait until v1.1. The change feels pretty low-stakes since click is only used by dbt init.

Open to disagreement, however

PR #4758, opened 2022-02-21
Fix bug accessing target in deps and clean commands

### LINKED ISSUE #4752: [CT-259] [Bug]dbt deps and dbt clean stopped working on dbt-core-1.0.2

### Current Behavior

Hi,

Starting today I have started receving the folling error error when trying to run `dbt deps` as well as `dbt clean`:
`
Encountered an error:
cannot pickle 'Undefined' object
`
After some investigation I could find that:
* the process was working fine till 2022-02-17 and then stopped working on 2022-02-18 so on the very same date that dbt-core-1.0.2 was released, I am using dbt-snowflake==1.0.0 so DBT is trying to pull latest dbt-core version and I can see in the logs that it was fine till it was getting dbt-core-1.0.1 and with dbt-core-1.0.2 it's failing, easy to spot because some of the pipelines in Azure DevOps just stopped working without any code change
* the issue occur in both Azure Devops pipelines (Microsoft Hosted Agents) as well as on the local VM (Linux) for newly configured environment (python virtul env)
* the only difference I can see from python packages installation is the dbt-core version and the click version
* the very same error appears for both `dbt deps` and `dbt clean`
* it's failing when running `dbt deps` as well as when trying to specify the profile manually `dbt deps --profile-dir . --target CI`
* there was a change for both `dbt deps` and `dbt clean` in the following PR #4610 so most likely it might be related, but cannot really find an issue myself
* of course temporary workaround would be to force re-installation of previous dbt-core installation as someone proposed in a bit similar other issue
* I am using at this point local DBT packages (copy of archived packages), but if the same issue is for `dbt clean` so maybe it's not the issue
* installation of dbt-snowflake==1.0.0 is working fine, installation of dbt-core==1.0.2 itself is fine too, the 
* I am using profile.yml with user and pass stored in the profile.yml file
* dbt.log is empty, no error message, no progress regarding clean or deps, if I would run any other command even `dbt debug` the log is populated correctly
* dbt debug, dbt run, dbt test seems to be working still fine

I am happy to provide more logs and isight if needed and possible.
I will continue some more testing so I will keep the issue updated.

Anyone maybe have had the same issue?

Best regards,
SD 

### Expected Behavior

_No response_

### Steps To Reproduce

_No response_

### Relevant log output

_No response_

### Environment

```markdown
- OS: Ubuntu 20.04.3
- Python: 3.8.12
- dbt: 1.0.2
```

snowflake

### Additional Context

_No response_

gshank:
resolves #4752

### Description

After changes in #4610, direct access to target.<attribute> fails with cryptic message "cannot pickle undefined object", because an empty dictionary was returned for the target. This changes the behavior of deps and clean commands to return a dict subclass that returns None for attempts to access a dictionary key.

### Checklist

### DIFF
--- a/test/integration/006_simple_dependency_tests/test_simple_dependency.py
+++ b/test/integration/006_simple_dependency_tests/test_simple_dependency.py
@@ -257,6 +257,15 @@ class TestSimpleDependencyBadProfile(DBTIntegrationTest):
     def models(self):
         return "models"
 
+    @property
+    def project_config(self):
+        return {
+            'config-version': 2,
+            'models': {
+                '+any_config': "{{ target.name }}",
+            }
+        }
+
     def postgres_profile(self):
         # Need to set the environment variable here initially because
         # the unittest setup does a load_config.
--- a/test/integration/006_simple_dependency_tests/test_simple_dependency.py
+++ b/test/integration/006_simple_dependency_tests/test_simple_dependency.py
@@ -263,6 +263,7 @@ class TestSimpleDependencyBadProfile(DBTIntegrationTest):
             'config-version': 2,
             'models': {
                 '+any_config': "{{ target.name }}",
+                '+enabled': "{{ target.name in ['redshift', 'postgres'] | as_bool }}"
             }
         }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,5 @@
 ## dbt-core 1.1.0 (TBD)
 
-## dbt-core 1.0.3 (TBD)
-
-### Fixes
-- Fix bug causing empty node level meta, snapshot config errors ([#4459](https://github.com/dbt-labs/dbt-core/issues/4459), [#4726](https://github.com/dbt-labs/dbt-core/pull/4726))
-
 ### Features
 - Added Support for Semantic Versioning ([#4644](https://github.com/dbt-labs/dbt-core/pull/4644))
 - New Dockerfile to support specific db adapters and platforms.  See docker/README.md for details ([#4495](https://github.com/dbt-labs/dbt-core/issues/4495), [#4487](https://github.com/dbt-labs/dbt-core/pull/4487))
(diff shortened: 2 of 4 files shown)


leahwicz (approved):

test/integration/006_simple_dependency_tests/test_simple_dependency.py
```diff
@@ -257,6 +257,15 @@ def schema(self):
     def models(self):
         return "models"
 
+    @property
+    def project_config(self):
+        return {
+            'config-version': 2,
+            'models': {
+                '+any_config': "{{ target.name }}",
```
jtcohen6, line 265: This works for the most common cases that we should expect to see. It still doesn't work for one of the configs that we saw in the wild:
```yml
# dbt_project.yml
models:
  +enabled: "{{ target.name in ['redshift', 'postgres'] | as_bool }}"
```
That still raises the error:
```
17:08:37  Encountered an error:
'in <string>' requires string as left operand, not NoneType
```
Is there any way we could hope to handle that? I guess by returning `""` instead of `None` for all values?

gshank: Seems like it should be okay... I've pushed a change to return an empty string

jtcohen6: @gshank I see you made the change to use `""` instead of `None`. I think it's the right call. The only way I could think of this breaking is if:
- ~There's a specific node config that accepts `None`, but not `str`~ Confirmed that config-level validation doesn't happen in `deps`/`clean`, it's when we're parsing + validating actual nodes
- A user has defined an expression like:
```yml
models:
  # target.port is numeric when set, but it will be "" in deps/clean
  +query_tag: "{{ target.port + 1234 }}"
```
That doesn't work with `NoneType` either, so I think we're either at parity or strictly better off with a string:
```
can only concatenate str (not "int") to str
unsupported operand type(s) for +: 'NoneType' and 'int'
```

gshank: Here's where I miss Perl... :). It would have treated the empty string as 0 here.

PR #4773, opened 2022-02-23
Fix "dbt found two resources" error with multiple snapshot blocks in one file

### LINKED ISSUE #4771: [CT-273] [Bug] modifying snapshot files with multiple snapshots causes error (partial parsing)
If you modify a snapshot file with multiple snapshot blocks, and then do 'dbt run', you see the error "dbt found two resources with the name "date_orders_snapshot".  Running with "--no-partial-parse" will re-parse the entire project and then you can run again.

The partial parsing code does not handle multiple snapshot blocks.

gshank:
resolves #4771

### Description

Loop through nodes when processing mssat files, instead of assuming that there is only one.

### Checklist


emmyoop (approved):
LGTM

PR #4783, opened 2022-02-24
rewrite graph selection

### LINKED ISSUE #4729: [CT-235] Convert 007 integration test to use pytest
This is the first integration test we will target to move over to use pytest. All other tests will follow

ChenyuLInx:
resolves #4729

### Checklist


ChenyuLInx:
Should I also remove the original tests in this PR? @gshank

gshank:
Yes. Please remove the original test directory.

tests/functional/graph_selection/fixtures.py
```diff
@@ -0,0 +1,178 @@
...
+
+
+# seed function and create_tables would do similar things
+# if you want to use create tables, pass it as an argument to the test function
+# if you want to use seeds. do run_dbt(['seed']) in your test function
+@pytest.fixture
+def seeds(test_data_dir):
+    # Read seed file and return
+    path = os.path.join(test_data_dir, "seed-initial.csv")
+    with open(path, "rb") as fp:
```
gshank, line 169: It wasn't working on Windows to write files out in a generic way, so in another pull request I've changed all the reads and writes to explicitly be text utf-8. You'll need to change this after that pull request is merged.

tests/functional/graph_selection/test_graph_selection.py
```diff
@@ -0,0 +1,358 @@
...
+    with table_comp.get_connection():
+        exists = project.adapter.check_schema_exists(project.database, project.test_schema)
+        assert exists
+
+        schema = project.test_schema + "_and_then"
+        exists = project.adapter.check_schema_exists(project.database, schema)
+        assert not exists
+
+
+def test__postgres__specific_model(project):
```
gshank, line 35: The 'postgres' piece should be removed from all of the test names. This one, for example, should be 'test_specific_model'.

tests/functional/graph_selection/test_graph_selection.py
```diff
@@ -0,0 +1,358 @@
...
+    assert "users" in created_models
+    assert "alternative.users" in created_models
+    assert "users_rollup_dependency" not in created_models
+    table_comp = TableComparison(
+        adapter=project.adapter, unique_schema=project.test_schema, database=project.database
+    )
+    assert_correct_schemas(project, table_comp)
+
+
+def test__postgres__specific_model_and_children(project, create_tables):  # noqa
```
gshank, line 109: Was there a reason for switching most of the tests to use the 'seed' command and the seed file and not the handful which still have 'create_tables'?

gshank: Was there a reason for needing the 'noqa'?

gshank (changes requested):
Looking good overall!

tests/functional/graph_selection/test_graph_selection.py
ChenyuLInx: Create_tables is not used in this function so need noqa to get the pre-commit check pass.
I left this here to show that you can do seed in both ways, but if we prefer [this](https://github.com/dbt-labs/dbt-core/blob/8fd8dfcf74657ccc5314640b22cb2589eed8a824/tests/functional/basic/test_simple_reference.py#L140) I can switch all of it over

gshank (approved):
Looks good!

PR #4795, opened 2022-02-25
ct-237 test conversion 002_varchar_widening_tests

### LINKED ISSUE #4731: [CT-237] Convert 002 integration test to use pytest
@emmyoop I just started at [001](https://github.com/dbt-labs/dbt-core/tree/main/test/integration/001_simple_copy_tests) to do as the next test after the initial one @ChenyuLInx is doing (#4729). If there is a better next test to go to, feel free to open an issue for that one. I imagine you and @ChenyuLInx tag teaming on these conversions.

emmyoop:
resolves #4731 

### Description

Converts 002_varchar_widening_tests

### Checklist


tests/functional/database_change/test_varchar_widening.py
```diff
@@ -0,0 +1,60 @@
...
+
+
+@pytest.fixture
+def models():
+    return {"incremental.sql": incremental_sql, "materialized.sql": materialized_sql}
+
+
+def test_varchar_widening(project):
+    path = os.path.join(project.test_data_dir, "varchar10_seed.sql")
+    project.run_sql_file(path)
```
ChenyuLInx, line 41: This would work but I think we are moving everything to dbt seed, one example is [link](https://github.com/dbt-labs/dbt-core/blob/main/tests/functional/graph_selection/fixtures.py#L173)

emmyoop: This check needs to make changes directly to the database.  It's a slightly different use case.

gshank: Yeah, for this case the point is that the definition of the table in the database changes. At the very least the second sql file would have to be run, so it's probably more straightforward to have them both be sql.

ChenyuLInx: [seed.sql](https://github.com/dbt-labs/dbt-core/blob/0.21.latest/test/integration/007_graph_selection_tests/seed.sql#L10) this is the SQL file seed replaced, why it is a different use case?

gshank:
I think you can put this test in the 'basic' directory, since it's a pretty minor thing. I was thinking that all of the 'simple' tests that were first written could go in there.

ChenyuLInx (approved):
LGTM!

PR #4803, opened 2022-02-28
simplify and cleanup gha workflow

kwigley:
### Description
- merge main and integration test workflow
- run tests on every commit
- hard code OS and python version matrix for tests
  - run all python versions on ubuntu
  - run at least 1 python verion on other OS's
- cleanup tox file now that we only test one adapter in this repo

### Checklist


kwigley:
`Adapter Integration Tests / test-metadata` job is expected to fail, it is removed in this PR

gshank (approved):
I'm certainly no expert in this area, but it looks good to me...

.github/workflows/main.yml
```diff
@@ -110,80 +104,109 @@ jobs:
...
-          check-wheel-contents dist/*.whl --ignore W007,W008
+      - name: Run tests
+        run: tox
 
       - uses: actions/upload-artifact@v2
+        if: always()
         with:
-          name: dist
-          path: dist/
+          name: logs
```
leahwicz, line 162: Should we follow a similar naming convention like we do for unit tests in line 104?

kwigley: Oh, I see what you mean. This will clobber the files on each permutation of matrix. I'll update!

tox.ini
```diff
@@ -11,23 +11,14 @@ deps =
   -rdev-requirements.txt
   -reditable-requirements.txt
 
-[testenv:{integration,py37,py38,py39,py}-{postgres}]
+[testenv:{integration,py37-integration,py38-integration,py38-integration}]
```
leahwicz, line 14: Is the last one supposed to be py39 and not 2 py38s?

leahwicz (approved):
Thanks Kyle for the improvements here!

PR #4815, opened 2022-03-02
Convert tests in dbt-adapter-tests to use new pytest framework

### LINKED ISSUE #4811: [CT-312] Convert dbt-adapters-tests
Converting the https://github.com/dbt-labs/dbt-adapter-tests tests to from yaml

gshank:
resolves #4811

### Description

This takes the tests in the dbt-adapter-tests repository and converts them to use the new pytest framework. Subsequent issues and pull requests will do the work to connect the tests up to the adapter repositories and enable running them on multiple adapters.

### Checklist


core/dbt/tests/fixtures/project.py
```diff
@@ -310,38 +292,7 @@ def run_sql_file(self, sql_path):
 
     # run sql from a string, using adapter saved at test startup
     def run_sql(self, sql, fetch=None):
```
emmyoop, line 294: Can you rename this to differentiate it from the utils `run_sql`?

tests/functional/adapter/files.py
```diff
@@ -0,0 +1,208 @@
...
+"""
+
+base_materialized_var_sql = materialized_var + model_base
+base_table_sql = materialized_table + model_base
+base_view_sql = materialized_view + model_base
+base_ephemeral_sql = materialized_ephemeral + model_base
+ephemeral_with_cte_sql = materialized_ephemeral + model_ephemeral_with_cte
+ephemeral_view_sql = materialized_view + model_ephemeral
+ephemeral_table_sql = materialized_table + model_ephemeral
+incremental_sql = materialized_incremental + model_incremental
```
emmyoop, line 208: nit: Can the config vars be prefixed with `config`?  It took me a minute to see what was being combined here.  Naming with `config` helps point that out.  Plus if we do something like this in other tests where config varies (but not just by materializations) we can name the same way.

tests/functional/adapter/test_base.py
```diff
@@ -0,0 +1,99 @@
...
+        "table_model": "table",
+        "swappable": "view",
+    }
+    check_relation_types(project.adapter, expected)
+
+    # run_dbt changing materialized_var to incremental
+    results = run_dbt(["run", "-m", "swappable", "--vars", "materialized_var: incremental"])
+    assert len(results) == 1
+
+    # check relation types, swappable is table
```
emmyoop, line 92: All the comments are so so so nice.

emmyoop (commented):
Overall looks good to me.  Just a few names I'm concerned about.

core/dbt/tests/fixtures/project.py
gshank: I renamed the other one, since is the one that we're actually using in the code.

tests/functional/adapter/files.py
gshank: Yeah, good idea. These files were pretty confusing and adding config_ does help.

core/dbt/tests/adapter.py
```diff
@@ -0,0 +1,193 @@
...
+def update_rows(adapter, update_rows_config):
+    """
+    {
+      "name": "base",
+      "dst_col": "some_date"
+      "clause": {
+          "type": "add_timestamp",
+          "src_col": "some_date",
+       "where" "id > 10"
+    }
```
ChenyuLInx, line 147: This example is really helpful! I feel like we might want to do this in more places like adding the doc string. For me a lot of these things are new so it was harder to follow the function without an example arg.

core/dbt/tests/util.py
```diff
@@ -85,3 +90,59 @@ def read_file(*paths):
...
+# For updating yaml config files
+def update_config_file(updates, *paths):
+    current_yaml = read_file(*paths)
+    config = yaml.safe_load(current_yaml)
+    config.update(updates)
+    new_yaml = yaml.safe_dump(config)
+    write_file(new_yaml, *paths)
+
+
+def run_sql(sql, adapter, database, schema, fetch=None):
```
ChenyuLInx: how is this related to run_sql at [tables.py](https://github.com/dbt-labs/dbt-core/blob/8fd8dfcf74657ccc5314640b22cb2589eed8a824/core/dbt/tests/tables.py#L311)

tests/functional/adapter/files.py
```diff
@@ -0,0 +1,208 @@
+seeds_base_csv = """
```
ChenyuLInx, line 1: I really like this, I will add this to the auto conversion script

ChenyuLInx (commented):
Over all looks great! Maybe we want to start adding doc strings to function as a standard

core/dbt/tests/util.py
gshank: This one has been renamed to run_sql_with_adapter now. It does pretty much the same thing as in tables.py, but the plan is to get rid of tables.py when we can, so I'd rather leave refactoring it to that effort.

core/dbt/tests/adapter.py
gshank: I added a few more doc strings.

emmyoop (approved):
LGTM

PR #4816, opened 2022-03-02
Add bugged version tag value to finds.

VersusFacit:
## What's this?

A small change to an exception raised during yml property file parsing.

For a before-after demonstration, I put a value '2.0' for the `version:` tag in a property value.

Before: 
image
After: 
image

I'm not married to the wording here, but I thought verbosity was better since the definition of "version" is very cloudy to users.

## Why this?

From question threads in the dbt Slack. Here's two of many: https://getdbt.slack.com/archives/CBSQTAPLG/p1646223161801819

Users are running into issues with the `version:` tag in property files. The way docs and `dbt init` is setup seems to be confusing some to think `1.0` is the property value everywhere. This is partially because `config-version` and `version` tags; this triggers a hard to debug issue when users copy the `version:` tag from their dbt_project.yml which is a semantic version tag. `config-vers. They're having trouble locating the field causing the error. To recreate the issue:
1. have a property file
2. make version something other than an integer.

The error message without the tag ends up having users kind of guess what's wrong when we could very easily provide clearer guidance by including the value. I don't think much else is needed here, although this will extend into a dbt docs PR, sounds like. Even though we give the filepath in the bug, some users seem to think this means the version in `dbt_project.yml` or somesuch. Or they end up guessing what they need to fix.

### Checklist

### DIFF
--- a/core/dbt/parser/schemas.py
+++ b/core/dbt/parser/schemas.py
@@ -547,18 +547,22 @@ class SchemaParser(SimpleParser[GenericTestBlock, ParsedGenericTestNode]):
 
 def check_format_version(file_path, yaml_dct) -> None:
     if "version" not in yaml_dct:
-        raise_invalid_property_yml_version(file_path, "no version is specified")
+        raise_invalid_property_yml_version(
+            file_path,
+            "the yml property file {} is missing a version tag".format(file_path)
+        )
 
     version = yaml_dct["version"]
     # if it's not an integer, the version is malformed, or not
     # set. Either way, only 'version: 2' is supported.
     if not isinstance(version, int):
         raise_invalid_property_yml_version(
-            file_path, "the version must be an integer. {} is not an integer".format(version)
+            file_path,
+            "its 'version:' tag must be an integer value. {} is not an integer".format(version)
         )
     if version != 2:
         raise_invalid_property_yml_version(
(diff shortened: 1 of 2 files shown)


core/dbt/parser/schemas.py
```diff
@@ -544,15 +547,22 @@ def parse_file(self, block: FileBlock, dct: Dict = None) -> None:
...
+        )
 
     version = yaml_dct["version"]
     # if it's not an integer, the version is malformed, or not
     # set. Either way, only 'version: 2' is supported.
     if not isinstance(version, int):
-        raise_invalid_schema_yml_version(file_path, "the version is not an integer")
+        raise_invalid_property_yml_version(
+            file_path,
+            "its 'version:' tag must be an integer value. {} is not an integer".format(version),
```
nathaniel-may, line 560: Minor wording suggestion: give examples of the full line we expect like `version: 2`

nathaniel-may (approved):
Looks like a nice improvement to this exception's wording.

core/dbt/parser/schemas.py
VersusFacit: Alright, so I thought about this myself hahaha. At what point should the code just say "if it isn't equal to two, fail it." I can see where it's useful though from a word standpoint now.

PR #4826, opened 2022-03-04
pytest conversion test_selection, schema_tests, fail_fast, permission

ChenyuLInx:
resolves #
Some test in this is fully by [convert_test](https://github.com/ChenyuLInx/unittest2pytest/blob/dbt-pytest-all-dir/convert_test.py), by running
```
python convert_test.py --test_dir  ~/git/dbt-core/test/integration/066_test_selection_tests --output_dir ~/git/dbt-core/tests/functional/test_selection
```
then do a little bit clean up of unused import.
Some require a bit of manual work to clean up the functions, i.e. we didn't have dbt tests so people were implementing it in tests.

### Checklist


core/dbt/tests/fixtures/project.py
```diff
@@ -118,16 +128,22 @@ def project_config_update():
     return {}
 
 
+@pytest.fixture
+def model_path():
```
ChenyuLInx, line 132: This part is needed in some tests that have more complex file structure

core/dbt/tests/fixtures/project.py
```diff
@@ -197,7 +213,12 @@ def write_project_files(project_root, dir_name, file_dict):
 # Write files out from file_dict. Can be nested directories...
 def write_project_files_recursively(path, file_dict):
     for name, value in file_dict.items():
-        if name.endswith(".sql") or name.endswith(".csv") or name.endswith(".md"):
+        if (
+            name.endswith(".sql")
+            or name.endswith(".csv")
+            or name.endswith(".md")
+            or name.endswith(".SQL")
```
ChenyuLInx, line 220: Some tests require this

ChenyuLInx (commented):
@gshank @emmyoop this is one with multiple tests, do we want to keep 1 PR for a tests or we can combine multiple into one PR?

core/dbt/tests/fixtures/project.py
gshank: I don't think we support SQL (as opposed to sql). We don't read csv files in models directories.

core/dbt/tests/fixtures/project.py
gshank: The model dictionaries can be nested. i.e.:
`{ "my_model.sql": '....', "subdir": { "another_model.sql": "..."}}`

ChenyuLInx: Yea, but this is used to specify model files to use for a run. previous integration tests supported it [here](https://github.com/dbt-labs/dbt-core/blob/5e6be1660eb5672c9bfc3d79aa02cec275de7533/test/integration/base.py#L359)

core/dbt/tests/fixtures/project.py
ChenyuLInx: This is needed for a [test](https://github.com/dbt-labs/dbt-core/blob/main/test/integration/008_schema_tests_tests/case-sensitive-models/uppercase.SQL). We can also remove that test if we don't want to do this

tests/functional/fail_fast/test_fail_fast_run.py
```diff
@@ -0,0 +1,67 @@
...
+    def test_fail_fast_run(
+        self,
+        project,
+    ):
+        with pytest.raises(FailFastException):
+            run_dbt(["run", "--threads", "1", "--fail-fast"])
+            check_audit_table(project)
+
+
+class TestFailFastFromConfig(TestFastFailingDuringRun):
```
ChenyuLInx, line 51: The original test did this but I am not too sure why

core/dbt/tests/fixtures/project.py
gshank: Which tests in the old integration tests? I know that the old integration test framework did that, but it was because they had directories for multiple projects in the same test directory, and we don't need that anymore.

gshank: Ooops. Commented on wrong comment.

core/dbt/tests/fixtures/project.py
gshank: Which tests in the old integration tests? I know that the old integration test framework did that, but it was because they had directories for multiple projects in the same test directory, and we don't need that anymore.

tests/functional/fail_fast/test_fail_fast_run.py
gshank: If you look at the project config, there's an "on-run-start" config that creates the audit table at the beginning of a run, and a pre-hook that inserts into the table for every model in the project that's run. So it's looking to see how many models were run.

gshank:
I think it's fine to combine multiple tests into one PR.

core/dbt/tests/fixtures/project.py
ChenyuLInx: This [one ](https://github.com/dbt-labs/dbt-core/blob/5615691c329ca55b5782b8f98f75b36bb3a081d6/tests/functional/schema_tests/test_schema_v2_tests.py#L38). Basically every test class has it's own models

ChenyuLInx: I can refactor it to just one models artifact per class if we want to do that.

gshank: Yes, but we don't need that anymore. We just construct a models fixture with the right files. They did that because every test had to have a separate directory, otherwise tests would interfere with each other.

gshank: I think one models fixture per class is probably better.

ChenyuLInx: @gshank  this is resolved, anything else we would like to change about this PR?

tests/functional/schema_tests/fixtures.py
```diff
@@ -0,0 +1,1397 @@
...
+"""
+
+case_sensitive_models__uppercase_SQL = """
+select 1 as id
+
+"""
+
+case_sensitive_models__schema_yaml = """
+
+"""
```
emmyoop, line 165: Is this test testing for a yml and yaml file or is this just trash from the original test we can delete?  Based on the name it seems like trash.

ChenyuLInx: Did a search and looks like there's nothing using it so I removed it as suggested

core/dbt/tests/fixtures/project.py
gshank: Please remove the name.endswith(".csv") and the name.endswith(".SQL"). I'd rather not clutter up the code with special stuff for a one time test. You can achieve the same thing by using the "write_file" utility in dbt.tests.util, something like: write_file(file_contents, project.project_root, "models", "something.SQL")

gshank (changes requested):
Just the one change and it will be good to go.

core/dbt/tests/fixtures/project.py
ChenyuLInx: Removed `.SQL`, but we can't remove `.csv` as it is being used in seeds

emmyoop:
You need to add the `Skip Changelog` label to the PR to bypass the GHA since we all agreed these incremental conversions would not get CHANGELOG entries.

gshank (approved):
Looks good now, once you have the Changelog thing handled.

PR #4827, opened 2022-03-05
Add selector method capabilities to selectors

### LINKED ISSUE #4821: [CT-319] [Feature] Selection method based on a selection definition

### Describe the Feature

I would like to be able to use a selection definition within another selection definition.
I believe the technical feature request here would be that `selector` is included in the list of methods that can be applied to a selector definition.

```yaml
selectors:
  - name: default_for_dev
    definition: ...
  - name: default_for_prod
    definition: ...
  - name: my_ci_job
    definition:
        - union:
            - method: selector
              value: "{{ "default_for_dev" if target.name == 'dev' else "default_for_prod"}}"
            - exclude ...
```

The most immediate alternative here is to use anchors and aliases. However, the issue arrises where we want to _dynamically_ define the selector definition. Unfortunately you cannot create an alias variable, for instance `*{{ my_jinja_var }}` would not work.

Additionally tried creating variables at the start of the yaml but `Could not parse selector file data:`
```
variables:
  - foo: bar

selectors:
...
union
  - *$(foo)
```

### Who will this benefit?

- Anyone looking to create dynamic selector definitions i.e. if X use a previously defined selector, else use a different selector definition
- Anyone less familiar with anchors and aliases and wanting to implement DRY methods.

A specific example includes having two default definitions depending on the environment and wanting to run a seperate selector that leans on the default.
```yaml
selectors:
  - name: default_for_dev
    definition: ...
  - name: default_for_prod
    definition: ...
  - name: my_ci_job
    definition:
        - union:
            - method: selector
              value: "{{ "default_for_dev" if target.name == 'dev' else "default_for_prod"}}"
            - exclude ...
```

Yes

### Anything else?

_No response_

danieldiamond:
resolves #4821

### Description
Add selector method in order to be able to define a selector with another selector definition.

### To Do
- [ ] Prevent circular dependencies
- [ ] unit test to ensure circular dependency check works
- [x] unit test to confirm selector method has expected behaviour

### Checklist


danieldiamond:
Initial attempt, seems to be working but not such a fan of injecting the `selector_dict` through the class methods, particularly as we're updating it (hence the `deepcopy`). Ideally, this would be an instantiated class to lean on `self.selector_dict` or similar.

danieldiamond:
@jtcohen6 any capacity this month for the team to review?

gshank (changes requested):
This looks pretty good. I see that you have a checkbox unchecked for circular dependencies though. Do you have an idea of how to handle that?  It might be tricky to handle robustly, but it would be good to create a test case with circular dependencies and see what error occurs. Catching the error at a higher level would probably be fine (if possible). A python error for recursiveness would be not optimal...

gshank:
Oh, also we now use 'changie' for the changelog, so the CHANGELOG.md should not be changed from main. See: https://github.com/dbt-labs/dbt-core/blob/main/CONTRIBUTING.md#adding-changelog-entry

danieldiamond:
> I see that you have a checkbox unchecked for circular dependencies though. Do you have an idea of how to handle that?

I was hoping to get some guidance on this. Is this something you can help with @gshank ?
Something as simple as checking if the selector `key` already exists then error out as "already defined at root" or similar

gshank:
That part will probably be a bit tricky. I think we'd have to keep a list of already encountered selectors in this branch of the tree and check if this selector was encountered higher up. I haven't touched this code in a while, so I'd have to pull it apart a bit further to figure out where that would have to go. For recursive functions, I generally just pass along a list of names that I check...

danieldiamond:
@gshank any chance I can get some help to get this over the line? Please and thank you in advance!

gshank:
I won't have time today, but I'll try to find some time to take a look this week.

gshank:
I did some testing and it looks like we don't end up with recursion, we just get "Existing selector definition for test_inherited not found."  Which is pretty innocuous.

gshank (approved):

Issue #4789, opened 2022-02-25
[CT-289] Automate Changelog Generation on Release


emmyoop:
When releasing, it would be ideal to have the `CHANGELOG` auto generate for the release.  It makes the most sense to be part of the [version-bump.yml](https://github.com/dbt-labs/dbt-core/blob/main/.github/workflows/version-bump.yml) GHA so that the version bump and changelog all go in as a single PR.  Since the version bump PR does not include a CHANGELOG entry, the PR would need to have the `Skip Changelog` tag added to it.

This would involve the input of that action needing to be split out to be version (ex 1.0.5) and prerelease tag (ex rc1, b1, etc) since changie takes them as separate inputs and inserts a dash in the middle while the version bump does not use the dash.

Not doing this as part of the original release of changie so we can see how the files get generated and manually adjust as needed.

nathaniel-may:
Just a reminder to update the README with these changes once they're made.

jtcohen6:
> Posting the contents of the release file as opposed to a link would be better since the prereleases end up getting squashed down into a final release.

On second thought: We still could link to the release notes, using the release tag as the branch. E.g. for v1.0.5rc1: https://github.com/dbt-labs/dbt-core/blob/v1.0.5rc1/.changes/1.0.5-rc1.md

It makes sense to me that:
- changelog is by long-lived minor version branch (`1.0.latest`, `1.1.latest`), and includes updates as new backports / patches are released
- release notes are immutable, and tied to a specific release/tag

So this comes down to a cosmetic question: to show all, or to link elsewhere?

I do like having the release notes visible right from [the releases page](https://github.com/dbt-labs/dbt-core/releases)—but I also recognize it can add clutter, and it takes longer to scroll back to see older releases. (Maybe we do neat things with HTML toggles...?)

lostmygithubaccount:
@emmyoop @jtcohen6 is this already done? going to remove staleness

emmyoop:
Partially.  @leahwicz added the changelog generation into `version-bump.yml` [here](https://github.com/dbt-labs/dbt-core/blob/bbf4fc30a56e43734dfc94e43a8cf1d2ef36ec29/.github/workflows/version-bump.yml#L77-L88).

But we are not automatically posting release notes yet.

emmyoop:
This has been done for a while now!

PR #4841, opened 2022-03-08
Small updates to clarify change destinations

### LINKED ISSUE #4652: [CT-115] Automate Changelog
### Describe the Feature

Our current process for maintaining an accurate `CHANGELOG` is very manual and prone to conflicts.

### Current Painpoints

- causes merge conflicts when merging/backporting changelog file changes
- hard to know what section to add an entry in the changelog — e.g. what version of the project will this change make it into, coordinating between minor and patch versions, rc’s and betas
- strictly a manual process, it is the responsibility of the PR author to add a changelog entry
- there is no way to enforce formatting and syntax
- you have to create the PR in order to link the PR and issue to the changelog entry
- always asking “does this deserve a changelog entry?”
- we now have a GitHub action to backport PRs, it won’t be useful if there are always merge conflicts due to the changelog file 😥

Automating the process should resolve these issues and make it easier to add a new `CHANGELOG` entry, reduce all the conflicts we manually resolve and help the new Github action to back port PRs works much more reliably.

### Goals
- Automatically generate `CHANGELOG` entry from a few user inputs
- Keep current `CHANGELOG` format
- Stop all the back porting merge conflicts (and generally conflicts) caused by the `CHANGELOG` so that out Githib action for back porting will be more successful

Let's try out [changie](https://github.com/miniscruff/changie) to see what it's like.

@kwigley's [changes](https://github.com/kwigley/changes) is still a wip but we can check it out once it has had more added to it!

### Who will this benefit?

Everyone that contributes to `dbt-core`.  We all have `CHANGELOG` entires!

Tasks to complete this

- [x] add changie to project
- [ ] add github action to check if changelog entry exists - possibly run changie new when not
- [ ] add github action for release process
- [ ] ~~script to add contributors section (or upgrade changie if new release comes out)~~
- [x] script to handle prerelease collapse on final release
- [ ] update onboarding notion doc
- [ ] update contributing doc
- [ ] update release notion doc

emmyoop:
resolves #4652 

### Description

Since `changie new` is run on the main branch (where changes are made) there needs to be a way to separate what we backport and what is slated for the next major/minor release.

I came up with the following solutions.

1. Always have a folder called `backport` where backported changes go.  The problem with this is when we backport the change, the yaml file will end up in a backport folder on the `.latest` branch.   Not as clean as I would like.
2. Delete the yaml file off main after you backport.  *Really* don't like this one as it requires a separate commit after all your changes are done.
3. Always have a folder for the next major/minor version, `.changes/next`.  If something is to be backported to any `.latest` branch, leave the yaml file in the default folder (`.changes/unreleased`).  If it's to be slated for the next major/minor version that we have not cut a `./latest` branch for yet, the yaml file will live in this new folder.  When it's time to cut the new branch, delete all yaml files under `/unreleased` that have been backported and generate the batch off the `.changes/next` folder.  This could be automated along with the batch/merge automation in #4789. 

I went with 3 but am happy to hear feedback.

Still need to
[ ] ~~add logic to GHA to check if `/unreleased` or `/next` contains a new file~~
[ ] ~~see if this is something that could be native to changie~~

Edit:  Solution is to change where changelog entries land based on what version they're applying to.  No separate folders needed.

This PR is still needed to clean up what a release will look like for 1.1.0 since it's going to take some manual work to combine the original changelog with the new changie entries.

### Checklist


emmyoop:
@leahwicz ideally you would know before it's merged.  But if you change your mind it would require a new PR to move the file from one directory to another.

This doesn't create any more overhead than we have now since if you put your `CHANGELOG` entry (pre-changie process) into 1.1.0 but then realized you should backport it for 1.0.4, you need to manually update the `CHANGELOG` to reflect that.  Instead of updating the `CHANGELOG` now, you slide the yaml file around.  The benefit being no merge conflicts.

Since we need to keep backport and next changes separate, there's no way around this.

nathaniel-may:
@emmyoop and I just had a quick sync. The key question here is about what we want our changelogs to look like post 1.0. The simplest solution for engineering would be the following:
- only one directory for new changelog entries
- when commits are backported, they include the changelog entry and can be done entirely automatically because they live in the same directory on the target branch.
- this necessarily means that a change that goes into 1.1.0 and is backported onto 1.0.4 will show up _in both changelogs with identical entries_ which is a deviation from how we have historically done our changelog.

CC: @jtcohen6

emmyoop:
> * this necessarily means that a change that goes into 1.1.0 and is backported onto 1.0.4 will show up _in both changelogs with identical entries_ which is a deviation from how we have historically done our changelog.

Just to clarify this a touch more.  The identical changelog entry will end up under the 1.0.4 heading in the `1.0.latest` changelog and under the `1.1.0` heading in the `1.1.latest` changelog.

This also sounds like a better/cleaner solution to me if having the entries under different headings is acceptable.

jtcohen6:
This approach makes sense to me. I'm happy to have the same changelog entry show up in both versions. Given the nature of backporting, _we are actually making the change twice_. We should also expect the scope of the fix may actually be different, for minor vs. patch releases, in which case we might also edit the changie entry while backporting.

nathaniel-may (approved):

PR #4860, opened 2022-03-13
cache schema for selected models

### LINKED ISSUE #4688: [CT-168] Cache objects for selected resources only?
### Background

Should we cache everything dbt cares about? Or try to cache objects relevant only to resources that are selected to run? This is an increasingly important consideration as folks move to:
- Execute smaller sub-DAGs within ever-larger projects
- Manage their slice of a multi-package dbt deployment 

The intent of the relation cache is to speed up performance, after all. If it's not serving that purpose, in its current unadaptive form, then we should change its behavior.

### Details

We know the set of selected resources at the time when we populate the adapter cache. We only _create_ schemas (if they do not yet exist) for objects that are selected. Should we also limit the reach of our metadata queries, to only introspect the schemas we care about?

https://github.com/dbt-labs/dbt-core/blob/a588607ec6546de441812e215fbbf4e9ff20101b/core/dbt/task/run.py#L457-L458

Should we go one step further, and use filters/wildcards to only cache the _objects_ we care about? This seems necessary on `dbt-spark` (issue linked below.)

### Risks

- The cache has made us "greedy" when it comes to really simple metadata lookups. What if, over the course of running a few models, we end up spending more time in one-off lookup queries (due to cache misses) than we saved at the start, by skipping one batch metadata query that seemed irrelevant?
- `--defer` depends on being able to access cache information about resources that are not selected (context in #2946):

https://github.com/dbt-labs/dbt-core/blob/a588607ec6546de441812e215fbbf4e9ff20101b/core/dbt/contracts/graph/manifest.py#L952-L955)

### Questions
- If dbt misses something in its cache, are we liable to return _slow_ or _incorrect_ results? The answer significantly changes the stakes of such a change
- Would we want cache behavior to vary based on  `--defer` behavior? 

### Related
- https://github.com/dbt-labs/dbt-snowflake/issues/83
- https://github.com/dbt-labs/dbt-bigquery/issues/104
- https://github.com/dbt-labs/dbt-core/issues/2673
- https://github.com/dbt-labs/dbt-spark/issues/228

karunpoudel:
resolves #4688

### Description

Cache schema for selected models only instead of caching all schemas in a project when `--selected-schema-cache` option is provided in cli or config. This will help improve startup time when a project has many databases and schema but only few models are selected.

This behavior is similar to how schemas are created currently. Dbt only creates schema (if they don't exists) for selected models only rather than all schema defined in a project.

### Checklist

### DIFF
--- a/core/dbt/adapters/base/impl.py
+++ b/core/dbt/adapters/base/impl.py
@@ -337,11 +337,12 @@ class BaseAdapter(metaclass=AdapterMeta):
         # databases
         return info_schema_name_map
 
-    def _relations_cache_for_schemas(self, manifest: Manifest) -> None:
+    def _relations_cache_for_schemas(self, manifest: Manifest, cache_schemas: Set[BaseRelation] = None) -> None:
         """Populate the relations cache for the given schemas. Returns an
         iterable of the schemas populated, as strings.
         """
-        cache_schemas = self._get_cache_schemas(manifest)
+        if not cache_schemas:
+            cache_schemas = self._get_cache_schemas(manifest)
         with executor(self.config) as tpe:
             futures: List[Future[List[BaseRelation]]] = []
             for cache_schema in cache_schemas:
@@ -367,14 +368,15 @@ class BaseAdapter(metaclass=AdapterMeta):
             cache_update.add((relation.database, relation.schema))
         self.cache.update_schemas(cache_update)
 
-    def set_relations_cache(self, manifest: Manifest, clear: bool = False) -> None:
+    def set_relations_cache(self, manifest: Manifest, clear: bool = False,
+                            required_schemas: Set[BaseRelation] = None) -> None:
         """Run a query that gets a populated cache of the relations in the
         database and set the cache on this adapter.
         """
         with self.cache.lock:
             if clear:
                 self.cache.clear()
-            self._relations_cache_for_schemas(manifest)
+            self._relations_cache_for_schemas(manifest, required_schemas)
 
     @available
     def cache_added(self, relation: Optional[BaseRelation]) -> str:
--- a/core/dbt/main.py
+++ b/core/dbt/main.py
@@ -1084,6 +1084,27 @@ def parse_args(args, cls=DBTArgumentParser):
         """,
     )
 
+    schema_cache_flag = p.add_mutually_exclusive_group()
+    schema_cache_flag.add_argument(
+        '--selected-schema-cache',
+        action='store_const',
+        const=True,
+        default=None,
+        dest='selected_schema_cache',
+        help='''
+        Pre cache objects of schema relevant to selected resource only.
+        '''
+    )
+    schema_cache_flag.add_argument(
+        '--no-selected-schema-cache',
+        action='store_const',
+        const=False,
+        dest='selected_schema_cache',
+        help='''
+        Pre cache objects of all schema.
+        '''
+    )
+
     subs = p.add_subparsers(title="Available sub-commands")
 
     base_subparser = _build_base_subparser()
--- a/core/dbt/task/run.py
+++ b/core/dbt/task/run.py
@@ -436,8 +436,9 @@ class RunTask(CompileTask):
 
     def before_run(self, adapter, selected_uids: AbstractSet[str]):
         with adapter.connection_named("master"):
-            self.create_schemas(adapter, selected_uids)
-            self.populate_adapter_cache(adapter)
+            required_schemas = self.get_model_schemas(adapter, selected_uids)
+            self.create_schemas(adapter, required_schemas)
+            self.populate_adapter_cache(adapter, required_schemas)
             self.defer_to_manifest(adapter, selected_uids)
             self.safe_run_hooks(adapter, RunHookType.Start, {})
 
--- a/core/dbt/task/runnable.py
+++ b/core/dbt/task/runnable.py
@@ -390,8 +390,11 @@ class GraphRunnableTask(ManifestTask):
         for dep_node_id in self.graph.get_dependent_nodes(node_id):
             self._skipped_children[dep_node_id] = cause
 
-    def populate_adapter_cache(self, adapter):
-        adapter.set_relations_cache(self.manifest)
+    def populate_adapter_cache(self, adapter, required_schemas: Set[BaseRelation] = None):
+        if flags.SELECTED_SCHEMA_CACHE is True:
+            adapter.set_relations_cache(self.manifest, required_schemas=required_schemas)
+        else:
+            adapter.set_relations_cache(self.manifest)
 
     def before_hooks(self, adapter):
         pass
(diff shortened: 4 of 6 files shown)


core/dbt/task/runnable.py
```diff
@@ -390,8 +390,11 @@ def _mark_dependent_errors(self, node_id, result, cause):
         for dep_node_id in self.graph.get_dependent_nodes(node_id):
             self._skipped_children[dep_node_id] = cause
 
-    def populate_adapter_cache(self, adapter):
-        adapter.set_relations_cache(self.manifest)
+    def populate_adapter_cache(self, adapter, required_schemas: Set[BaseRelation] = None):
+        if flags.SELECTED_SCHEMA_CACHE is True:
```
ChenyuLInx, line 394: Feels like we can just remove this function and call `adapter.set_relations_cache` in places that call `populate_adapter_cache`

core/dbt/adapters/base/impl.py
```diff
@@ -337,11 +337,12 @@ def _get_catalog_schemas(self, manifest: Manifest) -> SchemaSearchMap:
         # databases
         return info_schema_name_map
 
-    def _relations_cache_for_schemas(self, manifest: Manifest) -> None:
+    def _relations_cache_for_schemas(self, manifest: Manifest, cache_schemas: Set[BaseRelation] = None) -> None:
```
ChenyuLInx, line 340: Since we are here, let's just remove this function and move all of the logic into `set_relations_cache`

core/dbt/task/run.py
```diff
@@ -436,8 +436,9 @@ def defer_to_manifest(self, adapter, selected_uids: AbstractSet[str]):
 
     def before_run(self, adapter, selected_uids: AbstractSet[str]):
         with adapter.connection_named("master"):
-            self.create_schemas(adapter, selected_uids)
-            self.populate_adapter_cache(adapter)
+            required_schemas = self.get_model_schemas(adapter, selected_uids)
```
ChenyuLInx, line 439: @jtcohen6 `get_model_schemas` would remove all of the node that is not in `selected_uids`, is there a case we need adapter cache to contain more nodes than `selected_nodes`?

core/dbt/main.py
```diff
@@ -1084,6 +1084,27 @@ def parse_args(args, cls=DBTArgumentParser):
         """,
     )
 
+    schema_cache_flag = p.add_mutually_exclusive_group()
+    schema_cache_flag.add_argument(
```
ChenyuLInx, line 1088: I am fairly new to all of the command arguments, would like to hear why you chose to add two arguments that would change the same `dest` vs the default behavior is not use this feature and a just one flag to enable the feature.

ChenyuLInx (commented):
Thanks for submitting the PR!! Added some questions and suggestions(some of them are more like making the codebase cleaner since we are here).
One other thing we should add is some tests to test the expected behavior. I understand this is might be tricky in terms of what use cases we want to test and how to properly test those. And I am not sure if we have an existing test for this kind of stuff. But would love to hear your thoughts and work with you on it.

core/dbt/main.py
karunpoudel: One could set `selected_schema_cache: true` in profile.yml as general config for their project but may choose to override that with `--no-selected-schema-cache` in cli for specific scenario.

core/dbt/task/runnable.py
karunpoudel: `populate_adapter_cache`  here in `GraphRunnableTask` has been inherited by `RunTask` where it is mostly used. Moving this logic to `before_run` function would duplicate the same logic in `before_run` function in `GraphRunnableTask` and `RunTask` class.
If there is room to reorganize stuffs around, might make sense to do by someone more experienced with these code then me. This is the first time I am touching dbt-core :-)

core/dbt/adapters/base/impl.py
jtcohen6: Because these methods are defined in the `adapters` module, these are potentially breaking changes to the adapter interface. E.g. the Postgres plugin reimplements this method (currently with two arguments), and would require updating:
https://github.com/dbt-labs/dbt-core/blob/5e0a7659174a9277e2013390f8615acc97464d40/plugins/postgres/dbt/adapters/postgres/impl.py#L129-L131

core/dbt/main.py
jtcohen6: Right on! The inheritance order for "global" configs is CLI flag > env var (`DBT_{config}`) > profile config

I would vote for naming this config in a way that's slightly more generic (not every database calls it "schema," not every database caches at the schema level). Perhaps `CACHE_SELECTED_ONLY`?

core/dbt/task/run.py
jtcohen6: @ChenyuLInx I'm not sure! That's _the_ critical question animating in this issue/PR. I will say, having this PR makes it a lot easier to test.

- **Will dbt return _incorrect_ results, if it fails to cache an unselected relation?** I don't think so, as long as caching happens consistently at the schema level. (That's an open question for us right now on Spark: https://github.com/dbt-labs/dbt-spark/issues/296)
- **How will this work for defer?** When `--defer` is enabled, it's necessary to call `adapter.get_relation` for _all_ unselected parents of selected models, to figure out whether to use a "dev" version or defer to the "prod" version. In my local testing, this still works! dbt will just record a cache miss, and run the query independently:
```
08:35:23.500589 [debug] [MainThread]: On "master": cache miss for schema "jerco.dbt_jcohen_schema_a", this is inefficient
```
It's likely that this will be less efficient. So, I think it's important that the behavior remains configurable. I'm not opposed to moving forward with this, in order to test the waters. We may someday wish to turn it on by default.

jtcohen6 (commented):
@karunpoudel Very cool! Thanks so much for taking the idea and running with it. I left a few comments

ChenyuLInx (approved):
Looks good to me! Thanks @karunpoudel  for making it happen! We added some pre-commit hook to format the code. I think if you follow the instruction [here](https://github.com/dbt-labs/dbt-core/pull/4860#discussion_r829792876) and push the changes then the failed GHA will pass.

karunpoudel:
@jtcohen6, @ChenyuLInx , I am thinking of adding test to check cached relation after supplying this new flag. Do you have existing integration test to verify that caching is working properly for project with multiple database/schema. I can add a new test to that.

ChenyuLInx:
> verify that caching is working properly for project with multiple database/schema
@karunpoudel  I don't think we do. Thanks for adding those! Let me know if there's anything I can help with! We haven't made the annocement yet but we are swithing to new testing framework, you can find examples [here](https://github.com/dbt-labs/dbt-core/tree/main/tests/functional). And here are some [initial docs](https://github.com/dbt-labs/docs.getdbt.com/pull/1263/files). This is not fully done so if you feel more comfortable using the original test framework that also works for me. But I highly recommand checking out the new ones.

jtcohen6:
I'd be happy to have this change as is, as an optional/experimental flag, included in v1.1.0-rc1. We already have a follow-on issue open to add more rigorous testing and tracking (https://github.com/dbt-labs/dbt-core/issues/4961), ahead of this being something we'd want to turn on by default.

core/dbt/main.py
```diff
@@ -1088,6 +1088,27 @@ def parse_args(args, cls=DBTArgumentParser):
...
+        "--cache_selected_only",
```
jtcohen6, line 1093: One thing I just caught: for consistency with other flags, this should probably be `--cache-selected-only` (hyphens not underscores)

jtcohen6:
Rebased + merged in https://github.com/dbt-labs/dbt-core/pull/5036

PR #4865, opened 2022-03-15
catch all requests exceptions to retry

### LINKED ISSUE #4849: [CT-352] Catch all Requests Exception on Deps Install and Retry
We are not retrying for all exceptions from Requests and it's causing more unexpected/seemingly unexplained failures.  We should try to mitigate these failures by retrying (up to 5 times).  The logic to retry already exists.  We just need to make sure we catch the right/entirety of exceptions.

1. Catch *all* exceptions from Requests by catching just `requests.exceptions.RequestException` since all exceptions that Requests explicitly raises inherit from [`requests.exceptions.RequestException`](https://docs.python-requests.org/en/latest/api/#requests.RequestException).  Note that the `ReadError` should remain.

2. Also fire a new event to track what exception we're retrying so that we can track where the failures are in the future.

https://github.com/dbt-labs/dbt-core/blob/ecfd77f1caa10b7b15ca01104bcf3a9b54f3bb89/core/dbt/utils.py#L609-L620

Related Issues: #4601, #4178

emmyoop:
resolves #4849

### Description

1. Catch all exceptions from Requests by catching just `requests.exceptions.RequestException` instead of individual requests exceptions since they all inherit from `RequestException`
2. Fire a new event to track the exception that we are catching for future tracking
3. Update relevant tests

### Checklist

### DIFF
--- a/.changes/unreleased/Fixes-20220315-105331.yaml
+++ b/.changes/unreleased/Fixes-20220315-105331.yaml
@@ -1,0 +1,8 @@
+kind: Fixes
+body: Catch all Requests Exceptions on deps install to attempt retries.  Also log
+  the exceptions hit.
+time: 2022-03-15T10:53:31.637963-05:00
+custom:
+  Author: emmyoop
+  Issue: "4849"
+  PR: "4865"


gshank (approved):
Looks good to me.

PR #4869, opened 2022-03-16
get_response type hint is AdapterResponse only.

### LINKED ISSUE #4499: [CT-297] `get_response` should always return `AdapterResponse`
Split off from #4433:

> In v1.1.0, let's enforce that the `get_response` abstract method must return an `AdapterResponse`

This will allow us to clear up some tech debt. At its very simplest, an `AdapterReponse` can just look like the one in `dbt-spark`:

```python
    @classmethod
    def get_response(cls, cursor) -> AdapterResponse:
        return AdapterResponse(
            _message='OK'
        )
```

Ideally, of course, an adapter has more interesting information to pass along (number of rows affected, bytes processed, relevant query/job ID, etc).

Let's do this for v1.1.0, with clear communication ahead of time for adapter maintainers, since this does constitute a breaking change.

VersusFacit:
resolves #4499 

A low-code PR which sets the intention for adapter method return types: Remove type `str` from return type hint's of `get_response`. Optionally, remove type `str` from return type hint's of `execute` method too!

### Description

We have two options to merge this PR. Choose your adventure 🗺️ :
1. only the first commit, which touches only `SQLConnectionManager.get_response` as mentioned in original ticket
2. the entire PR which changes the type signature of the `BaseConnectionManager.execute ... -> Union[str, AdapterResponse]` to mixin `get_response` changes

See Appendix for why we should do option `2`. Do share your opinion.
I surveyed Postgres, Redshift (inherits from Postgres), Snowflake (uses AdapterResponse), and Spark (uses AdapterResponse), this won't break dbt-labs adapters. 

This is a breaking change (in the spec) for anyone who continues to use `str`.

The issue asks to enforce this type, but Python is dynamic so all we can do is strongly suggest via those type hints. Plus, it's the Pythonic thing to trust the implementor after all.

### Appendix. Why should we merge changes to `BaseConnectionManager.execute` too?

Good question! By altering the spec for `get_response`, `execute` no longer needs this as a return type. It has this signature because it (historically) returns the response of `get_response` which will soon be changed. An adapter may bypass `get_response` and instead just implement over `BaseConnectionManager` ([Bigquery does this](https://github.com/dbt-labs/dbt-bigquery/blob/a426d3aaee7cb2b642fc0cbfe421d785c67529e3/dbt/adapters/bigquery/connections.py#L162)).

In a way, I'd argue, to truly change the underlying spec-question this issue seeks to address, we must also change execute.

For sanity's sake, I researched our adapters and confirmed that changing execute won't break our current repos:
▶️  dbt-core/core/dbt/adapters/base/connections.py:282 [`execute` abstract interface method]
✅  [Postgres](https://github.com/dbt-labs/dbt-core/blob/15077d087c04006f5f0a4351773c06022ce2a08d/core/dbt/adapters/sql/connections.py#L118): uses `PostgresConnectionManager` which inherits from `BaseConnectionManager`
✅ [Redshift](https://github.com/dbt-labs/dbt-redshift/blob/4b81b0be700e926faba4963388d00c232d6493fe/dbt/adapters/redshift/connections.py#L70): uses `PostgresConnectionManager` with no `execute` override
✅ [Snowflake](https://github.com/dbt-labs/dbt-snowflake/blob/032d2bb9203ee42b0863ee964162d97d1c0a51a0/dbt/adapters/snowflake/connections.py#L205): inherits from `SQLConnectionManager` with no override
✅ [Spark](https://github.com/dbt-labs/dbt-spark/blob/d7f1d38d0dcf272dc0e513db4eeada0c08c407f5/dbt/adapters/spark/connections.py#L277): inherits from `SQLConnectionManager` with no override
✅ [Bigquery](https://github.com/dbt-labs/dbt-bigquery/blob/a426d3aaee7cb2b642fc0cbfe421d785c67529e3/dbt/adapters/bigquery/connections.py#L405): implements its own `execute` which returns `AdapterResponse`, not a `Union`

### Checklist

### DIFF
--- a/.changes/unreleased/Breaking Changes-20220316-031311.yaml
+++ b/.changes/unreleased/Breaking Changes-20220316-031311.yaml
@@ -1,5 +1,5 @@
 kind: Breaking Changes
-body: get_response now only returns a connection.AdapterReponse in type hints
+body: get_response and execute now only return a connection.AdapterReponse in type hints
 time: 2022-03-16T03:13:11.656007-07:00
 custom:
   Author: versusfacit
--- a/.changes/unreleased/Breaking Changes-20220316-031311.yaml
+++ b/.changes/unreleased/Breaking Changes-20220316-031311.yaml
@@ -1,0 +1,7 @@
+kind: Breaking Changes
+body: get_response now only returns a connection.AdapterReponse in type hints
+time: 2022-03-16T03:13:11.656007-07:00
+custom:
+  Author: versusfacit
+  Issue: "4499"
+  PR: "4869"
--- a/core/dbt/adapters/base/connections.py
+++ b/core/dbt/adapters/base/connections.py
@@ -281,15 +281,15 @@ class BaseConnectionManager(metaclass=abc.ABCMeta):
     @abc.abstractmethod
     def execute(
         self, sql: str, auto_begin: bool = False, fetch: bool = False
-    ) -> Tuple[Union[str, AdapterResponse], agate.Table]:
+    ) -> Tuple[AdapterResponse, agate.Table]:
         """Execute the given SQL.
 
         :param str sql: The sql to execute.
         :param bool auto_begin: If set, and dbt is not currently inside a
             transaction, automatically begin one.
         :param bool fetch: If set, fetch results.
-        :return: A tuple of the status and the results (empty if fetch=False).
-        :rtype: Tuple[Union[str, AdapterResponse], agate.Table]
+        :return: A tuple of the query status and results (empty if fetch=False).
+        :rtype: Tuple[AdapterResponse, agate.Table]
         """
         raise dbt.exceptions.NotImplementedException(
             "`execute` is not implemented for this adapter!"
--- a/core/dbt/adapters/base/connections.py
+++ b/core/dbt/adapters/base/connections.py
@@ -4,7 +4,7 @@ import os
 # multiprocessing.RLock is a function returning this type
 from multiprocessing.synchronize import RLock
 from threading import get_ident
-from typing import Dict, Tuple, Hashable, Optional, ContextManager, List, Union
+from typing import Dict, Tuple, Hashable, Optional, ContextManager, List
 
 import agate
 
--- a/core/dbt/adapters/base/impl.py
+++ b/core/dbt/adapters/base/impl.py
@@ -221,7 +221,7 @@ class BaseAdapter(metaclass=AdapterMeta):
     @available.parse(lambda *a, **k: ("", empty_table()))
     def execute(
         self, sql: str, auto_begin: bool = False, fetch: bool = False
-    ) -> Tuple[Union[str, AdapterResponse], agate.Table]:
+    ) -> Tuple[AdapterResponse, agate.Table]:
         """Execute the given SQL. This is a thin wrapper around
         ConnectionManager.execute.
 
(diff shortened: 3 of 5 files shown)


nathaniel-may (commented):
Thanks for the detailed PR description!

This PR brings up an important ethos question by making this change in core:</br> _How do we want to leverage Python type hints?_

This PR is explicitly calling upon type hints to declare the input we expect. If the caller of the function misused the function, all bad behavior is their responsibility. As you rightfully call out, this is in line with the Python ethos and more generally dynamically typed languages.

Alternatively, type hints can be used to declare the shape of values that _could_ be passed, even unexpected values. This would allow mypy to help make sure we handle all the cases gracefully. This is the ethos statically typed-languages enforce.

Neither is wrong and we don't even have to be consistent about our usage across the code base. I'm honestly fine with merging either, I just thought it was worth pointing out to make sure we're making an active choice.

Alternatively, there is a middle ground option as well. In regards to this part of the description:

> The issue asks to enforce this type, but Python is dynamic so all we can do is strongly suggest via those type hints. Plus, it's the Pythonic thing to trust the implementor after all.

To help adapter maintainers, we could detect a string response at run time (or more-generally a non-AdapterResponse value) and raise a clear and helpful error message. Without this it might be very non-obvious that they need to update the return type. I haven't tried yet, but it might be worth seeing what it looks like when a string is returned with this change.

VersusFacit:
> Alternatively, there is a middle ground option as well. In regards to this part of the description:
> ...

Hey Nate! I looked into this and reached the following conclusion. `get_response` is abstract. We can't validate their implementation of either method.  We _could_ validate as you suggest on `def execute` in `SQLConnectionManager`'s implementation of `.execute` **with the understanding this only affects users who implement their adapters with this class which doesn't always hold true**, even for us. 

I'd erred on the conservative side with this PR (no validation step), but could pretty easily toss in an exception there, as I had sketched out doing so yesterday. Thoughts?

nathaniel-may:
@VersusFacit, I'm fine with this PR as is especially if adding an exception isn't particularly straight forward.

nathaniel-may (approved):

johannes-becker-otto:
Hi,
trying to understand the implications for the adapter we're using
In dbt-exasol it is still
```
    @classmethod
    def get_response(cls, cursor):
        return 'SUCCESS'
```
Will this break and if yes when and how?
All the best
Hannes

jtcohen6:
Hey @johannes-becker-otto, this changed in v1.1. It's a very very simple change; you or the `dbt-exasol` maintainers are welcome to check out https://github.com/dbt-labs/dbt-spark/pull/265 for an example.

The code snippet you pasted above should simply become:
```python
    @classmethod
    def get_response(cls, cursor) -> AdapterResponse:
        message = 'SUCCESS'
        return AdapterResponse(
            _message=message
        )
```

We called this out in [the v1.1 release notes](https://github.com/dbt-labs/dbt-core/releases/tag/v1.1.0#:~:text=Relevant%20to%20maintainers%20of%20adapter%20plugins%3A) and [migration guide](https://docs.getdbt.com/docs/guides/migration-guide/upgrading-to-v1.1#for-maintainers-of-adapter-plugins). We'd like all adapters to start returning an `AdapterResponse` object, or an object of an adapter-specific subclass (e.g. `BigQueryAdapterResponse`). This makes it easier for us, in `dbt-core`, to reason about the response object each query is returning.

johannes-becker-otto:
Thanks for the answer! (I should have read the release notes more closely but came to this info through unrelated an error about get_response...)
Cheers

PR #4878, opened 2022-03-16
Use cli_vars instead of context to create package and selector renderers

### LINKED ISSUE #4876: [CT-371] [Bug] CLI vars in packages.yml and selectors.yml don't work.
When we did the work to separate out rendering of secrets, a bug was introduced where instead of using cli_vars to construct the contexts for packages and selectors, we use the entire yaml context (which is also a dict). Because of this we get errors like: "Object of type method is not JSON serializable" and also vars are not found when rendering.

gshank:
resolves #4876

### Description

Fix errors rendering vars in packages.yml and selectors.yml.

### Checklist


emmyoop (approved):
Looks good!

PR #4883, opened 2022-03-16
fix printer task logic for warnings

### LINKED ISSUE #4814: [CT-316] [Bug] tests set with `severity: warn` creates a `level:error` row with `--log-format json`

### Current Behavior

I have a model with a test set to severity warn:
``` sql
-- test_trigger.sql
SELECT 1 AS numb
UNION ALL
SELECT 1
```
```yml
version: 2

models:
  - name: test_trigger
    columns:
      - name: numb
        tests:
          - unique:
              config:
                severity: warn
```

I run this as `dbt --log-format json test -m test_trigger`, which generates a bunch of logs. The two important are these:

```json
{
    "code": "Z021",
    "data": {
        "node_name": "unique_test_trigger_numb",
        "path": "models\\pbi\\test_trigger.yml",
        "resource_type": "test"
    },
    "invocation_id": "75c78961-8621-47f9-b082-f8b8efaf4203",
    "level": "warn",
    "log_version": 1,
    "msg": "\u001b[33mWarning in test unique_test_trigger_numb (models\\pbi\\test_trigger.yml)\u001b[0m",
    "node_info": {},
    "pid": 10548,
    "thread_name": "MainThread",
    "ts": "2022-03-02T11:58:14.824775Z",
    "type": "log_line"
}
```

```json
{
    "code": "Z024",
    "data": {
        "msg": "Got 1 result, configured to warn if != 0"
    },
    "invocation_id": "75c78961-8621-47f9-b082-f8b8efaf4203",
    "level": "error",
    "log_version": 1,
    "msg": "  Got 1 result, configured to warn if != 0",
    "node_info": {},
    "pid": 10548,
    "thread_name": "MainThread",
    "ts": "2022-03-02T11:58:14.824775Z",
    "type": "log_line"
}
```

The first line has `level:warn` as expected. But the second has `level:error`, which I do not expect.

### Expected Behavior

A test set with `severity:warn` should produce no log rows with `level:error`.

### Environment

```markdown
- OS: Windows 10
- Python: 3.9.6
- dbt: dbt-core=1.0.3, dbt-redshift=1.0.0
```

redshift

### Additional Context

_No response_

iknox-fa:
Resolves #4814 and CT-316
### Description
TL;DR-- When a dbt test has a severity level explicitly set to warn it may still trigger an error (the default) if it also contains a message

See 4814 for further details
### Checklist

- [X] I have added information about my change to be included in the [CHANGELOG](https://github.com/dbt-labs/dbt-core/blob/main/CONTRIBUTING.md#Adding-CHANGELOG-Entry).

### DIFF
--- a/core/dbt/task/printer.py
+++ b/core/dbt/task/printer.py
@@ -98,10 +98,10 @@ def print_run_result_error(result, newline: bool = True, is_warning: bool = Fals
                 )
             )
 
-        if result.message and not is_warning:
-            fire_event(RunResultError(msg=result.message))
-        else:
-            fire_event(RunResultErrorNoMessage(status=result.status))
+            if result.message:
+                fire_event(RunResultError(msg=result.message))
+            else:
+                fire_event(RunResultErrorNoMessage(status=result.status))
 
         if result.node.build_path is not None:
             with TextOnly():
--- a/core/dbt/task/printer.py
+++ b/core/dbt/task/printer.py
@@ -98,7 +98,7 @@ def print_run_result_error(result, newline: bool = True, is_warning: bool = Fals
                 )
             )
 
-        if result.message:
+        if result.message and not is_warning:
             fire_event(RunResultError(msg=result.message))
         else:
             fire_event(RunResultErrorNoMessage(status=result.status))
@@ -145,7 +145,6 @@ def print_run_end_messages(results, keyboard_interrupt: bool = False) -> None:
                 keyboard_interrupt=keyboard_interrupt,
             )
         )
-
         for error in errors:
             print_run_result_error(error, is_warning=False)
 
--- a/core/dbt/task/printer.py
+++ b/core/dbt/task/printer.py
@@ -145,6 +145,7 @@ def print_run_end_messages(results, keyboard_interrupt: bool = False) -> None:
                 keyboard_interrupt=keyboard_interrupt,
             )
         )
+
         for error in errors:
             print_run_result_error(error, is_warning=False)
 
--- a/.changes/unreleased/Fixes-20220316-174940.yaml
+++ b/.changes/unreleased/Fixes-20220316-174940.yaml
@@ -1,0 +1,8 @@
+kind: Fixes
+body: 'dbt tests with a set severity level of warn may still trigger error level log
+  events '
+time: 2022-03-16T17:49:40.236594-05:00
+custom:
+  Author: iknox-fa
+  Issue: "4814"
+  PR: "4883"


core/dbt/task/printer.py
```diff
@@ -98,10 +98,10 @@ def print_run_result_error(result, newline: bool = True, is_warning: bool = Fals
                 )
             )
 
-        if result.message:
-            fire_event(RunResultError(msg=result.message))
-        else:
-            fire_event(RunResultErrorNoMessage(status=result.status))
+            if result.message:
```
iknox-fa, line 101: @jtcohen6 @nathaniel-may Does this fix make sense in terms of desired behavior?  The logic isn't clear or documented about when things have messages and when things are warnings and what the intersection of the two should be so this is my ian-makes-assumptions solution.

jtcohen6: We are setting `result.message` when a test warns:
https://github.com/dbt-labs/dbt-core/blob/5e0a7659174a9277e2013390f8615acc97464d40/core/dbt/task/test.py#L194-L196

So the expectation is that we should keep firing that event, and printing it to the CLI:
```
08:14:33  1 of 1 START test unique_my_model_id............................................ [RUN]
08:14:33  1 of 1 WARN 1 unique_my_model_id................................................ [WARN 1 in 0.04s]
08:14:33
08:14:33  Finished running 1 test in 0.17s.
08:14:33
08:14:33  Completed with 1 warning:
08:14:33
08:14:33  Warning in test unique_my_model_id (models/tests.yml)
08:14:33    Got 1 result, configured to warn if != 0     <<< removed by the change in this PR
08:14:33
08:14:33    compiled SQL at target/compiled/testy/models/tests.yml/unique_my_model_id.sql
```

The difference is, that message should have a `WARN` log level, instead of an `ERROR` level.

**Simpler fix:** I imagine that might look like adding a new event type, and rearranging the logic here, so that `if is_warning and result.message`, we fire `RunResultWarningMessage`. (And rename event type Z024 from `RunResultError` to `RunResultErrorMessage`.)

**Better fix:** To turn these into real structured events, it sure seems like we should be passing in `num_errors` + `test.config.error_if`/`test.config.warn_if` as dataclass attributes, and synthesizing the warning/error messages in the events module, rather than constructing those strings in `task/test.py`.

nathaniel-may: Just spoke with @iknox-fa over zoom. This task was not created as a part of the structured logging effort, and is in dire need of a refactor. It also clearly calls out for the event module to handle "error on warning" behavior so I have made a tech debt ticket for it: #4914.

ChenyuLInx:
@iknox-fa  is this still something we would like to merge? If yes then maybe we can close [this](https://github.com/dbt-labs/dbt-core/pull/4828) community PR?

PR #4890, opened 2022-03-17
[HOTFIX] update dbt-extractor dependency

### LINKED ISSUE #4891: [CT-381] [HOTFIX] Homebrew Incident Resolution
### What Happened

dbt-core depends on the dbt-extractor package, and the dbt-extractor package depends on tree-sitter-jinja2. dbt-extractor specifies tree-sitter-jinja2 via a github link using the git protocol. Github security rules changed to require this link to use https which caused cargo to fail to build the dbt-extractor.

### Who Is Affected

Everyone attempting to build dbt-core from source after the github security rules took affect. This primarily affects homebrew users since homebrew builds dbt from source locally.

### Solution:
- release new dbt-extractor (0.4.1). The fix is already in main
    - dbt-labs/dbt-extractor#51
- release new dbt-core patch from branch [1.0.4-hotfix](https://github.com/dbt-labs/dbt-core/tree/1.0.4-hotfix) which depends on this new version and accepts all future patch releases so we can skip this step in the future. This branch is only the 3 necessary commits ahead of 1.0.3 to fix this incident.
    - main: #4890
    - backport is directly on branch [1.0.4-hotfix](https://github.com/dbt-labs/dbt-core/tree/1.0.4-hotfix) because of complications with running the bump-version workflow for a hotfix branch.
    
Getting the release out has been delayed due to complications with github actions due to an [ongoing GitHub incident](https://www.githubstatus.com/incidents/dcnvr6zym66r).

nathaniel-may:
resolves #4891

### Description

Github security rules changed over and the published version of the dbt extractor used a github link to resolve its dependency on tree-sitter-jinja2 that is causing it to fail for installations from source mainly affecting homebrew users. This in conjunction with the new release of dbt-extractor will fix the issue.

This must be backported to 1.0.latest and released to resolve the issue.

### DIFF
--- a/.changes/unreleased/Fixes-20220317-132418.yaml
+++ b/.changes/unreleased/Fixes-20220317-132418.yaml
@@ -1,0 +1,7 @@
+kind: Fixes
+body: depend on new dbt-extractor version with fixed github links
+time: 2022-03-17T13:24:18.419599-04:00
+custom:
+  Author: nathaniel-may
+  Issue: "4891"
+  PR: "4890"
--- a/core/setup.py
+++ b/core/setup.py
@@ -64,7 +64,7 @@ setup(
         "networkx>=2.3,<3",
         "packaging>=20.9,<22.0",
         "sqlparse>=0.2.3,<0.5",
-        "dbt-extractor==0.4.0",
+        "dbt-extractor~=0.4.1",
         "typing-extensions>=3.7.4,<4.2",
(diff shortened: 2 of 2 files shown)


stu-k (approved):

nathaniel-may:
will wait to merge till dbt-extractor is updated. which is why we see all these fails.

Issue #3348, opened 2021-05-12
Better node identification


iknox-fa:
### Describe the feature
As proven by the necessity to add a hash to a test node's `unique_id`, the current model of node identification could use some tweaking.

This was discussed in previous ticket comments:
https://github.com/fishtown-analytics/dbt/pull/3335#discussion_r631213752
https://github.com/fishtown-analytics/dbt/pull/3335#discussion_r629693671

### Describe alternatives you've considered
This isn't well fleshed out yet, but the ideas that come immediately to mind are:
- re-work the existing human readable unique_id so the naming scheme is actually unique
- replace the unique_id with a full hash and include enough tooling that we don't have to manage hashes by hand in tests, etc.

### Who will this benefit?
This will make development simpler and eliminate the need for [post-parsing checks like this](https://github.com/fishtown-analytics/dbt/blob/bfdf7f01b55a90ea8292e1c537b6cf44faa412ab/core/dbt/contracts/graph/manifest.py#L1262).

### Are you interested in contributing this feature?
![image](https://user-images.githubusercontent.com/81931810/118032134-eb25fb80-b32c-11eb-8052-e47a8c00ead6.png)

jtcohen6:
Thanks for this @iknox-fa! I'm tagging this `tests` because this is most relevant to schema/generic test nodes, for which the node `name` is auto-generated. (Every other node name is user-specified, often based on the file name.)

There was a great [thread over in Slack just now](https://getdbt.slack.com/archives/C01UM2N7814/p1620843008122800) about what might make sense from a user's point of view.

iknox-fa:
@jtcohen6 I see where you're coming from, but it's actually more about the `unique_id` than the `name`.  For that reason I'm thinking we should keep all nodes in-play here.

Right now we have three+ ways to ID a node: `unique_id` `unique_id` + hash, and `name`.  I'd love it if we could get that down to one human readable identifier, but I rather doubt it will stay human readable* as we add more to the node content.

If that is the case, uniform logic to handle `unique_id`s as a hash and `name` as a human readable label seems like a likely solution and would include all nodes.

*unless you want a Donaudampfschifffahrtselektrizitätenhauptbetriebswerkbauunterbeamtengesellschaft

leahwicz:
* Human readable vs hash? 
* If we do a hash we should be able to translate it back to the original name
* We need a design discussion on this

leahwicz:
Let's do a design doc on this first and then scope

[...]

Issue #4684, opened 2022-02-04
[CT-166] [Bug] accepted_values test causing  OSError: File name too long


treeforest456:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

We have 12 accepted values defined in the `accepted_values` test list. When running this, it gives this OSError
```
OSError: [Errno 36] File name too long: '/home/********/project/project-name/target/test.folder.source_accepted_values_very_long_accepted_values_strings.123456789.html'
```

### Expected Behavior

Maybe a better way to generate test file names? Or maybe document somewhere that there could be a limit of number of accepted values defined in the list?

### Environment

```markdown
- OS: macOS Big Sur, version 11.5.2
- Python: 3.9.5
- dbt: 1.0.0
```

### What database are you using dbt with?

snowflake

nathaniel-may:
Hi, @treeforest456! Thanks for the report, and I'm sorry to hear you're having issues running tests. We have run into issues with the way we internally uniquely identify test nodes before and it seems the file path that is too long for your file system is caused in part by the unique identifier being too large.

[I left a comment](https://github.com/dbt-labs/dbt-core/issues/3348#issuecomment-1031840405) on a previous issue to propose a long-term fix, and I wanted to point you to [a proposed short-term fix in this ticket](https://github.com/dbt-labs/dbt-core/issues/3626#issuecomment-886694083) which is identified as a good first issue for a community PR, and I believe would solve your issue as well.

Unfortunately, I cannot think of any workarounds for your particular issue that do not involve a code change to dbt.

Since this ticket specifically calls out the inability to write test results because of the filepath length, I do not believe this issue is a duplicate and will leave it open until resolved.

Issue #4102, opened 2021-10-20
[Feature] Allow multiple `unique` tests for the same column


ilmari-aalto:
### Is there an existing feature request for this?

- [X] I have searched the existing issues

### Describe the Feature

Enable more than one `unique` test for the same column.

I think there are usecases where you'd like to have several `unique` tests. For example, here's an example related to MRR. In this example, there can only be one conversion record per account, and separately, just one current record per account:
```yaml
      - name: account_id
        tests:
          # There should be just one is_new_biz record per account
          - unique:
              where: "is_new_biz"
          # There should be just one is_current record per account
          - unique:
              where: "is_current"
```

### Describe alternatives you've considered

* In some cases several tests could be combined using multiple conditions in `where`
* Writing a singular test would always work, but it's disconnected from the rest of the model configuration, and an extra effort
* Also `dbt_utils.unique_combination_of_columns` could work but the intention of that test would be less clear

### Who will this benefit?

Enabling more than one `unique` test per column would allow users combine existing tests in novel ways - users would have added flexibility with the same building blocks. A beginner dbt user will not even need to know of the possibility so it wouldn't need to confuse them.

### Anything else?

From the technical point of view, multiple `unique` tests are currently not possible because config is not part of the test name, and all tests get the same name. Perhaps a test order number (1, 2, ..), hash, or test config could be added to the test name to work past this. (Perhaps extended test naming would only kick in when more than one test of the same type exists for the same column.)

More discussion and examples in this [Slack thread](https://getdbt.slack.com/archives/C016X6ABVUK/p1634151813050300).

judahrand:
This is a rather annoying behaviour for us. We've just upgraded to 1.0 from 0.19 and had assumed that with the addition of the Test Config we'd be able to get rid of our custom recency test. However, due to this behaviour we cannot. Our use case looks like this:

```yaml
- recency:
    datepart: day
    field: date
    interval: 2
    where: "connection_type_code = 'tiktok'"
- recency:
    datepart: day
    field: date
    interval: 2
    where: "connection_type_code = 'twitter'"
- recency:
    datepart: day
    field: date
    interval: 2
    where: "connection_type_code = 'awin'"
- recency:
    datepart: day
    field: date
    interval: 2
    where: "connection_type_code = 'reddit'"
- recency:
    datepart: day
    field: date
    interval: 2
    where: "connection_type_code = 'pinterest'"
```

[...]

PR #4898, opened 2022-03-18
Custom names for generic tests

### LINKED ISSUE #4012: Significant Slowdown in dbt 0.21.0 on redshift
### Describe the bug
Since upgrading to 0.21.0 I have seen a significant slowdown in the speed with which DBT runs this includes both `dbt compile` and `dbt run`. I am running on a redshift data warehouse, though I don't know if the issue is redshift specific.

### Steps To Reproduce
Simply run `dbt run` or `dbt compile` with partial parse enabled or disabled the issue occurs either way.

### Expected behavior
I saw a 3-5x slowdown so i expected it to run significantly faster

image

image

### System information
- [ ] postgres
- [√] redshift
- [ ] bigquery
- [ ] snowflake

```
installed version: 0.21.0
   latest version: 0.21.0

Up to date!

Plugins:
  - bigquery: 0.21.0
  - snowflake: 0.21.0
  - redshift: 0.21.0
  - postgres: 0.21.0
```

Mac OS Big Sur 11.3.1
Python 3.7.5

jtcohen6:
resolves https://github.com/dbt-labs/dbt-core/issues/3348#issuecomment-1031840405
resolves #4012

The fuller issue is about rethinking how we construct `unique_id`. The comment, and this PR, seeks to address the narrower issue that we've seen come up most often in the community (and in https://github.com/dbt-labs/dbt-core/issues/4684, https://github.com/dbt-labs/dbt-core/issues/4102, etc).

### TODO

- [x] Do we want to append the model/source name to the test name? Or trust users to pick globally unique ones?
- [x] Do people like this proposed syntax?
- [x] Add integration tests, ideally using the new `pytest` framework :)

### Description

Support this:
```yml
models:
  - name: my_model
    columns:
      - name: id
        tests:
          - unique:
              name: my_cool_test
```
```
$ dbt test
...
07:43:51  1 of 1 START test my_cool_test.................................................. [RUN]
07:43:51  1 of 1 PASS my_cool_test........................................................ [PASS in 0.03s]
```

### Why this change?

See the linked comment for thorough explanation. Quick answers below:
- Better control over the name of the test as it appears in console output. Aliasing the database tables was nice, but insufficient
- Naming generic tests is a step toward documenting tests, until we can support actual test descriptions (#3249)
- Avoid collisions between generic tests that are identical except for differences in configuration, e.g.

What this PR doesn't do: Alter, in any way, the logic that dbt uses to synthesize generic test names if no custom name is provided. The proposal in https://github.com/dbt-labs/dbt-core/issues/3348#issuecomment-1031840405 does include a cleaner more consistent approach to hashed test names. I opted against it, to keep the changes here minimal and additive-only. We can discuss whether that change feels necessary.

### While we're here...

Also add support for defining tests as unified dictionaries, like other resource types, in addition to nested dictionaries with the test name as the top-level key. If done in this way, the name of the generic test (`unique`, `not_null`, etc) is defined as `test_name`.
```yml
models:
  - name: my_model
    columns:
      - name: id
        tests:
          # these are equivalent
          - accepted_values:
              name: only_allow_abc
              values: ['a', 'b', 'c']
              config:
                severity: warn
                where: 1=1
          - name: only_allow_abc
            test_name: accepted_values
            values: ['a', 'b', 'c']
            config:
              severity: warn
              where: 1=1
```

I like the second formulation a bit better when defining lots of custom properties and configurations. It's important to continue supporting the first spec; after all, most generic tests are defined with a single keyword (the test name alone).

### Checklist

### DIFF
--- a/.changes/unreleased/Features-20220318-085756.yaml
+++ b/.changes/unreleased/Features-20220318-085756.yaml
@@ -1,0 +1,7 @@
+kind: Features
+body: Support custom names for generic tests
+time: 2022-03-18T08:57:56.05584+01:00
+custom:
+  Author: jtcohen6
+  Issue: "3348"
+  PR: "4898"
--- a/core/dbt/exceptions.py
+++ b/core/dbt/exceptions.py
@@ -838,31 +838,46 @@ def raise_duplicate_macro_name(node_1, node_2, namespace) -> NoReturn:
 
 def raise_duplicate_resource_name(node_1, node_2):
     duped_name = node_1.name
+    node_type = NodeType(node_1.resource_type)
+    pluralized = (
+        node_type.pluralize()
+        if node_1.resource_type == node_2.resource_type
+        else "resources"  # still raise if ref() collision, e.g. model + seed
+    )
 
-    if node_1.resource_type in NodeType.refable():
-        get_func = 'ref("{}")'.format(duped_name)
-    elif node_1.resource_type == NodeType.Source:
+    action = "looking for"
+    # duplicate 'ref' targets
+    if node_type in NodeType.refable():
+        formatted_name = f'ref("{duped_name}")'
+    # duplicate sources
+    elif node_type == NodeType.Source:
         duped_name = node_1.get_full_source_name()
-        get_func = node_1.get_source_representation()
-    elif node_1.resource_type == NodeType.Documentation:
-        get_func = 'doc("{}")'.format(duped_name)
-    elif node_1.resource_type == NodeType.Test and "schema" in node_1.tags:
-        return
+        formatted_name = node_1.get_source_representation()
+    # duplicate docs blocks
+    elif node_type == NodeType.Documentation:
+        formatted_name = f'doc("{duped_name}")'
+    # duplicate generic tests
+    elif node_type == NodeType.Test and hasattr(node_1, "test_metadata"):
+        column_name = f'column "{node_1.column_name}" in ' if node_1.column_name else ""
+        # TODO: we need a better way of storing the target model/source/etc in generic tests
+        model_name = ".".join((node_1.refs or node_1.sources)[0])
+        duped_name = f'{node_1.name}" defined on {column_name}"{model_name}'
+        action = "running"
+        formatted_name = "tests"
+    # all other resource types
     else:
-        get_func = '"{}"'.format(duped_name)
+        formatted_name = duped_name
 
+    # should this be raise_parsing_error instead?
     raise_compiler_error(
-        'dbt found two resources with the name "{}". Since these resources '
-        "have the same name,\ndbt will be unable to find the correct resource "
-        "when {} is used. To fix this,\nchange the name of one of "
-        "these resources:\n- {} ({})\n- {} ({})".format(
-            duped_name,
-            get_func,
-            node_1.unique_id,
-            node_1.original_file_path,
-            node_2.unique_id,
-            node_2.original_file_path,
-        )
+        f'dbt found two {pluralized} with the name "{duped_name}". '
+        f"\n"
+        f"\nSince these resources have the same name, dbt will be unable to find the correct resource "
+        f"\nwhen {action} {formatted_name}. "
+        f"\n"
+        f"\nTo fix this, change the name of one of these resources: "
+        f"\n- {node_1.unique_id} ({node_1.original_file_path}) "
+        f"\n- {node_2.unique_id} ({node_2.original_file_path})"
     )
 
 
--- a/core/dbt/exceptions.py
+++ b/core/dbt/exceptions.py
@@ -859,8 +859,7 @@ def raise_duplicate_resource_name(node_1, node_2):
     # duplicate generic tests
     elif node_type == NodeType.Test and hasattr(node_1, "test_metadata"):
         column_name = f'column "{node_1.column_name}" in ' if node_1.column_name else ""
-        # TODO: we need a better way of storing the target model/source/etc in generic tests
-        model_name = ".".join((node_1.refs or node_1.sources)[0])
+        model_name = node_1.file_key_name
         duped_name = f'{node_1.name}" defined on {column_name}"{model_name}'
         action = "running"
         formatted_name = "tests"
@@ -870,14 +869,16 @@ def raise_duplicate_resource_name(node_1, node_2):
 
     # should this be raise_parsing_error instead?
     raise_compiler_error(
-        f'dbt found two {pluralized} with the name "{duped_name}". '
-        f"\n"
-        f"\nSince these resources have the same name, dbt will be unable to find the correct resource "
-        f"\nwhen {action} {formatted_name}. "
-        f"\n"
-        f"\nTo fix this, change the name of one of these resources: "
-        f"\n- {node_1.unique_id} ({node_1.original_file_path}) "
-        f"\n- {node_2.unique_id} ({node_2.original_file_path})"
+        f"""
+dbt found two {pluralized} with the name "{duped_name}".
+
+Since these resources have the same name, dbt will be unable to find the correct resource
+when {action} {formatted_name}.
+
+To fix this, change the name of one of these resources:
+- {node_1.unique_id} ({node_1.original_file_path})
+- {node_2.unique_id} ({node_2.original_file_path})
+    """.strip()
     )
 
 
--- a/core/dbt/parser/generic_test_builders.py
+++ b/core/dbt/parser/generic_test_builders.py
@@ -294,13 +294,23 @@ class TestBuilder(Generic[Testable]):
                 "test must be dict or str, got {} (value {})".format(type(test), test)
             )
 
-        test = list(test.items())
-        if len(test) != 1:
-            raise_parsing_error(
-                "test definition dictionary must have exactly one key, got"
-                " {} instead ({} keys)".format(test, len(test))
-            )
-        test_name, test_args = test[0]
+        # If the test is a dictionary with top-level keys, the test name is "test_name"
+        # and the rest are arguments
+        # {'name': 'my_favorite_test', 'test_name': 'unique', 'config': {'where': '1=1'}}
+        if "test_name" in test.keys():
+            test_name = test.pop("test_name")
+            test_args = test
+        # If the test is a nested dictionary with one top-level key, the test name
+        # is the dict name, and nested keys are arguments
+        # {'unique': {'name': 'my_favorite_test', 'config': {'where': '1=1'}}}
+        else:
+            test = list(test.items())
+            if len(test) != 1:
+                raise_parsing_error(
+                    "test definition dictionary must have exactly one key, got"
+                    " {} instead ({} keys)".format(test, len(test))
+                )
+            test_name, test_args = test[0]
 
         if not isinstance(test_args, dict):
             raise_parsing_error(
--- a/core/dbt/parser/generic_test_builders.py
+++ b/core/dbt/parser/generic_test_builders.py
@@ -273,6 +273,10 @@ class TestBuilder(Generic[Testable]):
         self.fqn_name: str = ""
 
         if "name" in self.args:
+            # TODO: Should we append the model name to the test name here?
+            # Or trust the user to have globally unique names within their project?
+            # Logic like this, with accounting for source table targets:
+            # generic_test_name = f"{self.args["name"]}_{target.name}"
             self.compiled_name = self.args["name"]
             self.fqn_name = self.args["name"]
             del self.args["name"]
--- a/core/dbt/parser/generic_test_builders.py
+++ b/core/dbt/parser/generic_test_builders.py
@@ -25,9 +25,12 @@ from dbt.exceptions import raise_compiler_error, raise_parsing_error
 from dbt.parser.search import FileBlock
 
 
-def get_nice_generic_test_name(
+def synthesize_generic_test_names(
     test_type: str, test_name: str, args: Dict[str, Any]
 ) -> Tuple[str, str]:
+    # Using the type, name, and arguments to this generic test, synthesize a (hopefully) unique name
+    # Returns a shorter version (hashed/truncated, for the compiled file)
+    # as well as the full name (for the unique_id + FQN)
     flat_args = []
     for arg_name in sorted(args):
         # the model is already embedded in the name, so skip it
@@ -263,13 +266,23 @@ class TestBuilder(Generic[Testable]):
         if self.namespace is not None:
             self.package_name = self.namespace
 
-        compiled_name, fqn_name = self.get_test_name()
-        self.compiled_name: str = compiled_name
-        self.fqn_name: str = fqn_name
+        # If the user has provided a custom name for this generic test, use it
+        # Then delete the "name" argument to avoid passing it into the test macro
+        # Otherwise, use an auto-generated name synthesized from test inputs
+        self.compiled_name: str = ""
+        self.fqn_name: str = ""
 
-        # use hashed name as alias if too long
-        if compiled_name != fqn_name and "alias" not in self.config:
-            self.config["alias"] = compiled_name
+        if "name" in self.args:
+            self.compiled_name = self.args["name"]
+            self.fqn_name = self.args["name"]
+            del self.args["name"]
+        else:
+            short_name, full_name = self.get_synthetic_test_names()
+            self.compiled_name = short_name
+            self.fqn_name = full_name
+            # use hashed name as alias if full name is too long
+            if short_name != full_name and "alias" not in self.config:
+                self.config["alias"] = short_name
 
     def _bad_type(self) -> TypeError:
         return TypeError('invalid target type "{}"'.format(type(self.target)))
@@ -401,7 +414,8 @@ class TestBuilder(Generic[Testable]):
             macro_name = "{}.{}".format(self.namespace, macro_name)
         return macro_name
 
-    def get_test_name(self) -> Tuple[str, str]:
+    def get_synthetic_test_names(self) -> Tuple[str, str]:
+        # Returns two names: shorter (for the compiled file), full (for the unique_id + FQN)
         if isinstance(self.target, UnparsedNodeUpdate):
             name = self.name
         elif isinstance(self.target, UnpatchedSourceDefinition):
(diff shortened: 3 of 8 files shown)


core/dbt/parser/generic_test_builders.py
```diff
@@ -25,9 +25,12 @@
 from dbt.parser.search import FileBlock
 
 
-def get_nice_generic_test_name(
+def synthesize_generic_test_names(
     test_type: str, test_name: str, args: Dict[str, Any]
 ) -> Tuple[str, str]:
+    # Using the type, name, and arguments to this generic test, synthesize a (hopefully) unique name
```
nathaniel-may, line 31: [nit] If we're going to use the word "hopefully" in this comment let's lay out exactly when this would not be unique.

core/dbt/parser/generic_test_builders.py
```diff
@@ -263,13 +266,27 @@ def __init__(
...
-        if compiled_name != fqn_name and "alias" not in self.config:
-            self.config["alias"] = compiled_name
+        # If the user has provided a custom name for this generic test, use it
+        # Then delete the "name" argument to avoid passing it into the test macro
+        # Otherwise, use an auto-generated name synthesized from test inputs
+        self.compiled_name: str = ""
+        self.fqn_name: str = ""
+
+        if "name" in self.args:
+            # TODO: Should we append the model name to the test name here?
```
nathaniel-may, line 276: [blocker] I believe the result of this name mangling strategy should include the model name so our test-uniqueness constraints can be mechanically verified.

core/dbt/parser/generic_test_builders.py
```diff
@@ -401,7 +428,8 @@ def macro_name(self) -> str:
             macro_name = "{}.{}".format(self.namespace, macro_name)
         return macro_name
 
-    def get_test_name(self) -> Tuple[str, str]:
+    def get_synthetic_test_names(self) -> Tuple[str, str]:
+        # Returns two names: shorter (for the compiled file), full (for the unique_id + FQN)
```
nathaniel-may, line 432: This type signature is a very easy-to-misuse one. good call adding a comment.

nathaniel-may (changes requested):
Looks pretty good to me! Thanks for the very clear description. I have one important thought on the keys we use to define this and a few minor notes. I'm also withholding approval till we get the integration tests in this PR.

[blocker] I don't like the distinction between `name` and `test_name`. I'd like to see the functionality of defining unique test names associated with exactly one key regardless of the test its being applied to.

> I opted against it, to keep the changes here minimal and additive-only.

I think this is a good approach. The advantage of modifying the internal hashing strategy is that a dbt version upgrade with zero code changes would fix the issue where long test names can't be run, and that can absolutely be done in another PR. We should probably open a clear ticket to that effect so the work doesn't get lost.

gshank:
One concern I have is that people may not realize that they've defined an 'only_allow_abc' test in a different file. I'm thinking that we should lookup the test name and see if a test already exists with that name, and if it does issue an error message. Otherwise it feels like the tests could silently overwrite each other (like they sometimes do today). Always suffixing the model name would eliminate a lot of duplicates, but it would still be possible to name two tests on the same model with the same test name.

jtcohen6:
@nathaniel-may @gshank Thank you both for the speedy review + feedback!

### Nomenclature

> [blocker] I don't like the distinction between `name` and `test_name`. I'd like to see the functionality of defining unique test names associated with exactly one key regardless of the test its being applied to.

Could you say a bit more about this?

I agree this nomenclature can be somewhat confusing:
- `name` is the name of this specific instance/case of a generic test (i.e. this test node)
- `test_name` is the name of the generic test (macro), of which this test (node) is a specific instance

I picked `test_name` for consistency with how it's called in our selection syntax: https://docs.getdbt.com/reference/node-selection/methods#the-test_name-method
```bash
# select all instances of "unique" tests in the project
$ dbt test --select test_name:unique
```

I'm not sure what you mean by "associated with exactly one key." If I define a test via just the test name (`unique`)—the most expedient and common type we see—I'll get an auto-generated ("synthetic") name for that specific test case/node, which factors in the model/column names. If I want to give that test case/node a custom name, that's an additional key.

### Internal hashing strat

> The advantage of modifying the internal hashing strategy is that a dbt version upgrade with zero code changes would fix the issue where long test names can't be run, and that can absolutely be done in another PR. We should probably open a clear ticket to that effect so the work doesn't get lost.

Agree! I think that ticket could be spun off #4684. FWIW our synthetic naming strategy is already meant to account for too-long test names, by hashing and truncating names longer than 64 characters. That's the motivation behind the `short_name, full_name` tuple situation. So that approach must be breaking down somewhere along the way.

The plus of such a fix would be, it works for users with zero code changes on their end. The downside would be, we potentially "reset" synthetic test names for lots of folks, causing a break in any metadata tracking. Very much worth it IMO, just something we'd want to include in a minor (not patch) release.

### How should uniqueness work?

> [blocker] I believe the result of this name mangling strategy should include the model name so our test-uniqueness constraints can be mechanically enforced if we want them to be.

> Always suffixing the model name would eliminate a lot of duplicates, but it would still be possible to name two tests on the same model with the same test name.

It turns out, this actually works:
```yml
version: 2

models:
  - name: model_a
    columns:
      - name: id
        tests:
          - unique:
              name: my_not_so_unique_test_name
  - name: model_b
    columns:
      - name: id
        tests:
          - unique:
              name: my_not_so_unique_test_name
```
```
$ dbt test
09:24:42  Running with dbt=1.0.1
09:24:43  Found 2 models, 2 tests, 0 snapshots, 0 analyses, 367 macros, 0 operations, 1 seed file, 0 sources, 0 exposures, 0 metrics
09:24:43
09:24:43  Concurrency: 5 threads (target='dev')
09:24:43
09:24:43  1 of 2 START test my_definitely_unique_test_name................................ [RUN]
09:24:43  2 of 2 START test my_definitely_unique_test_name................................ [RUN]
09:24:43  1 of 2 PASS my_definitely_unique_test_name...................................... [PASS in 0.04s]
09:24:43  2 of 2 PASS my_definitely_unique_test_name...................................... [PASS in 0.04s]
09:24:43
09:24:43  Finished running 2 tests in 0.17s.
09:24:43
09:24:43  Completed successfully
09:24:43
09:24:43  Done. PASS=2 WARN=0 ERROR=0 SKIP=0 TOTAL=2
```

Why? Because of the hash that we append to each test's `unique_id`, since #3335, which _does_ include the model name (and column name, and other arguments).
```
$ dbt ls -s my_not_so_unique_test_name --output json --output-keys name,unique_id
{"unique_id": "test.testy.my_not_so_unique_test_name.116ca7b894", "name": "my_not_so_unique_test_name"}
{"unique_id": "test.testy.my_not_so_unique_test_name.a6184f01da", "name": "my_not_so_unique_test_name"}
```

So the tests are disambiguated within dbt internally, just as they would be if the model name were prepended to the `name`. For the end user, it might be a bit confusing, since `dbt test --select my_not_so_unique_test_name` actually returns two tests... but that's about it.

The only way to get actual dupes is by reproducing the setup from #4102, where the same generic test is defined on the same model + column, and the only thing different is the config (not factored into the hash, for all the reasons we've discussed):
```yml
models:
  - name: model_a
    columns:
      - name: id
        tests:
          - unique:
              name: my_not_so_unique_test_name
              where: 1=1
          - unique:
              name: my_not_so_unique_test_name
              where: 1=2
```
```
Compilation Error
  dbt found two resources with the name "my_not_so_unique_test_name". Since these resources have the same name,
  dbt will be unable to find the correct resource when "my_not_so_unique_test_name" is used. To fix this,
  change the name of one of these resources:
  - test.testy.my_not_so_unique_test_name.116ca7b894 (models/one_file.yml)
  - test.testy.my_not_so_unique_test_name.116ca7b894 (models/one_file.yml)
```
The resolution there is to pick some actually unique test names. Those definitions are necessarily in the same file, defined on the same column, so they really cannot be more than a few lines apart from one another.

I'm inclined to prefer keeping it this way, to let users actually set the exact test names they want?

gshank:
I didn't actually realize that we would still be appending the hash to the unique id. I think that's fine then. Will need documenting of course :)

I do really like the idea of letting people name their tests.

nathaniel-may:
### RE: How should uniqueness work

I love this compilation error at the bottom of your comment. What a clear development experience. I definitely want to make sure we never silently override tests and it sounds like this would prevent that.

If we're going to apply the uniqueness constraint to {model, test_name, column pairs} instead of just {model, test_name} I think we need to display that in the error which is why I don't think hashing in name mangling is a good practice. If you have 1599 columns in your table, it should be simple to see where the conflict is if every column has a test with this name and one test accidentally got copy-pasted twice for a column.

to be concrete, I'd rather see:
```
...
  - test.testy.model_a.id.my_not_so_unique_test_name (models/one_file.yml)
  - test.testy.model_a.id.my_not_so_unique_test_name (models/one_file.yml)
```

### RE: Nomenclature

I think I totally misunderstood your proposal. My previous comment was under the assumption that sometimes you'd use `name` to apply your unique name to a test and sometimes you would use `test_name`. After re-reading your example more carefully it looks like you want to use `name` every time for that which sounds good to me.

### RE: Internal hashing strat

Yeah that's fair, the max length bit might just need to be looked into independently. I know windows enforces a max path length of 255 characters so with longer-named directory structures, a 64 character file could easily overflow that.

core/dbt/exceptions.py
```diff
@@ -838,31 +838,46 @@ def raise_duplicate_macro_name(node_1, node_2, namespace) -> NoReturn:
...
-        return
+        formatted_name = node_1.get_source_representation()
+    # duplicate docs blocks
+    elif node_type == NodeType.Documentation:
+        formatted_name = f'doc("{duped_name}")'
+    # duplicate generic tests
+    elif node_type == NodeType.Test and hasattr(node_1, "test_metadata"):
+        column_name = f'column "{node_1.column_name}" in ' if node_1.column_name else ""
+        # TODO: we need a better way of storing the target model/source/etc in generic tests
+        model_name = ".".join((node_1.refs or node_1.sources)[0])
```
gshank, line 863: I'm not sure if this would help, but I did add 'file_key_name' to ParsedGenericTestNode because I was having the same problem in partial parsing. It stores things like 'models.my_model' and 'sources.my_source'.

gshank (approved):

.changes/unreleased/Features-20220318-085756.yaml
```diff
@@ -0,0 +1,7 @@
+kind: Features
+body: Support custom names for generic tests
```
nathaniel-may, line 2: Is it just generic tests or all tests?

core/dbt/exceptions.py
```diff
@@ -838,31 +838,46 @@ def raise_duplicate_macro_name(node_1, node_2, namespace) -> NoReturn:
...
-            node_2.original_file_path,
-        )
+        f'dbt found two {pluralized} with the name "{duped_name}". '
+        f"\n"
+        f"\nSince these resources have the same name, dbt will be unable to find the correct resource "
+        f"\nwhen {action} {formatted_name}. "
+        f"\n"
+        f"\nTo fix this, change the name of one of these resources: "
+        f"\n- {node_1.unique_id} ({node_1.original_file_path}) "
+        f"\n- {node_2.unique_id} ({node_2.original_file_path})"
```
nathaniel-may, line 880: [nit] for readability, I'd rather see a multiline f-string instead of relying on c-style string concatenation with explicit newline characters. For example:

```
f'''line one of {message_noun}
line two of {message_noun}
etc.'''
```

core/dbt/parser/generic_test_builders.py
nathaniel-may: thanks for adding that!

core/dbt/exceptions.py
```diff
@@ -838,31 +838,46 @@ def raise_duplicate_macro_name(node_1, node_2, namespace) -> NoReturn:
...
-    elif node_1.resource_type == NodeType.Test and "schema" in node_1.tags:
-        return
+        formatted_name = node_1.get_source_representation()
+    # duplicate docs blocks
+    elif node_type == NodeType.Documentation:
+        formatted_name = f'doc("{duped_name}")'
+    # duplicate generic tests
+    elif node_type == NodeType.Test and hasattr(node_1, "test_metadata"):
+        column_name = f'column "{node_1.column_name}" in ' if node_1.column_name else ""
+        # TODO: we need a better way of storing the target model/source/etc in generic tests
```
nathaniel-may, line 862: Let's log a tech debt ticket and put the github issue number in the comment here, unless @gshank's suggestion is the solution.

core/dbt/parser/generic_test_builders.py
```diff
@@ -263,13 +267,25 @@ def __init__(
...
-        if compiled_name != fqn_name and "alias" not in self.config:
-            self.config["alias"] = compiled_name
+        # If the user has provided a custom name for this generic test, use it
+        # Then delete the "name" argument to avoid passing it into the test macro
+        # Otherwise, use an auto-generated name synthesized from test inputs
+        self.compiled_name: str = ""
+        self.fqn_name: str = ""
+
+        if "name" in self.args:
+            # Trust the user to have a unique name for this model + column combo
```
nathaniel-may, line 277: [nit] "trust the user" makes it sounds like we're relying on them to enforce uniqueness themselves. We aren't because we raise an error. I'd change this comment to something more like "assign the user-defined name here, but it will be checked for uniqueness later"

core/dbt/parser/generic_test_builders.py
```diff
@@ -281,13 +297,23 @@ def extract_test_args(test, name=None) -> Tuple[str, Dict[str, Any]]:
...
-        test = list(test.items())
-        if len(test) != 1:
-            raise_parsing_error(
-                "test definition dictionary must have exactly one key, got"
-                " {} instead ({} keys)".format(test, len(test))
-            )
-        test_name, test_args = test[0]
+        # If the test is a dictionary with top-level keys, the test name is "test_name"
+        # and the rest are arguments
+        # {'name': 'my_favorite_test', 'test_name': 'unique', 'config': {'where': '1=1'}}
```
nathaniel-may, line 302: `test_name` and test `name` are confusing, so this example really helps 👍

tests/functional/schema_tests/test_schema_v2_tests.py
```diff
@@ -658,6 +661,60 @@ def test_collision_test_names_get_hash(
         assert test_results[1].node.unique_id in expected_unique_ids
 
 
+class TestGenericTestsCollide:
+    @pytest.fixture(scope="class")
+    def models(self, dupe_tests_collide):  # noqa: F811
```
nathaniel-may, line 666: [nit] `# noqa: F811` is a little unexpected here since that's usually for import statements. Does it work without this comment?

tests/functional/schema_tests/test_schema_v2_tests.py
```diff
@@ -658,6 +661,60 @@ def test_collision_test_names_get_hash(
...
+            run_dbt()
+        assert "dbt found two tests with the name" in str(exc)
+
+
+class TestGenericTestsCustomNames:
+    @pytest.fixture(scope="class")
+    def models(self, custom_generic_test_names):  # noqa: F811
+        return custom_generic_test_names
+
+    def test_collision_test_names_get_hash(
```
nathaniel-may, line 684: [nit] let's add a comment to all the test methods describing why we need this test. (imagine the reader is considering deleting it)

nathaniel-may (approved):
Great work here! I left many nits, but no blockers.

Thanks for clarifying in the previous comment that the output line is the node's `unique_id`. I do think that is what should actually be changed in an ideal world, but the way we're doing it now works and has less far-reaching implications.

core/dbt/exceptions.py
jtcohen6: Great call! This is much less hacky, so I think it's the right way to go. I do wish the `file_key_name` for sources stored `sources.my_source.my_src_table`.

.changes/unreleased/Features-20220318-085756.yaml
jtcohen6: Singular tests take their name from the file in which they're defined (`tests/my_custom_test.sql`), just like models — so they've always required custom names

core/dbt/exceptions.py
jtcohen6: The downside of a multiline f-string is that indentation gets messy:
```
$ dbt test
10:16:41  Running with dbt=1.1.0-b1
10:16:41  Encountered an error:
Compilation Error
  dbt found two tests with the name "my_not_so_unique_column_level_test" defined on column "id" in "sources.my_src".
          Since these resources have the same name, dbt will be unable to find the correct resource
          when running tests.
          To fix this, change the name of one of these resources:

          - test.testy.my_not_so_unique_column_level_test.345c7a643d (models/one_file.yml)

          - test.testy.my_not_so_unique_column_level_test.345c7a643d (models/one_file.yml)
```
There is a way to get  the indentation right, but then  it looks a bit odd in the source code:
```python
    raise_compiler_error(f"""
dbt found two {pluralized} with the name "{duped_name}".

Since these resources have the same name, dbt will be unable to find the correct resource
when {action} {formatted_name}.

To fix this, change the name of one of these resources:
- {node_1.unique_id} ({node_1.original_file_path})
- {node_2.unique_id} ({node_2.original_file_path})
    """.strip())
```

core/dbt/exceptions.py
jtcohen6: Gerda's suggestion is definitely better than what I'm doing here, and I think it's enough for now—but I agree that this is a limitation in generic tests, which also prevents us from being able to do things like #4723

tests/functional/schema_tests/test_schema_v2_tests.py
jtcohen6: I copied this from the other test definitions, it looks like this is a `flake8` complaint:
```
tests/functional/schema_tests/test_schema_v2_tests.py:666:22: F811 redefinition of unused 'dupe_tests_collide' from line 6
```
Indeed because `dupe_tests_collide` is being imported at the top, from the separate file of fixtures (`tests.functional.schema_tests.fixtures`)

core/dbt/exceptions.py
jtcohen6: flake + black seem to be happy with the latter, so I think that's what I'll go with

nathaniel-may: yup. this is exactly the trade off, but I most commonly see the latter across languages that support "raw strings" so I'm inclined to do it here as well.

tests/functional/schema_tests/test_schema_v2_tests.py
nathaniel-may: interesting. that seems like a flake8 bug counting a reference as a redefinition. thanks for double checking!

core/dbt/exceptions.py
gshank: Yeah, I was thinking you might want that. It might be possible to update it to do that...

PR #4912, opened 2022-03-21
add compliation and cache tracking

### LINKED ISSUE #4625: [CT-94] Holistic tracking for perceived start-up performance
There are three time-consuming steps that contribute to perceived slowness at the start of `dbt run`:
1. **Parsing**: Reading files &rarr; creating the internal manifest. Scales with size of project, but helped by partial + static parsing, and with [fairly solid telemetry](https://github.com/dbt-labs/dbt-core/blob/3032594b26aaafaf1003db513645660d0e7c89dc/core/dbt/parser/manifest.py#L774) today.
2. **Graph compilation** (called [here](https://github.com/dbt-labs/dbt-core/blob/3032594b26aaafaf1003db513645660d0e7c89dc/core/dbt/task/runnable.py#L91)): Scales with the number of nodes in a project, limited by memory / networkx algorithm
3. **Adapter cache construction** (called [here](https://github.com/dbt-labs/dbt-core/blob/3032594b26aaafaf1003db513645660d0e7c89dc/core/dbt/task/runnable.py#L410): Once we have a database connection, run metadata queries (or call metadata endpoints) to populate the cache, saving repeated queries by subsequent materializations. Scales with number of objects in the database, and number of unique namespaces (databases/schemas) that need caching. Speed also varies _significantly_ across different databases; this is particularly slow on Spark/Databricks today (https://github.com/dbt-labs/dbt-spark/issues/228).

Not all commands complete all of these steps. For instance, `dbt ls` and `dbt run-operation` need only step 1. But all of these steps must run, each time and in sequence, for "graph runnable" tasks (`run`, `test`, `build`, etc)—and they contribute to the perceived delay between submitting the command, and seeing the first node's query hit the database.

While we don't have present appetite for top-to-bottom highly involved performance work, the sooner we can add telemetry, the better served we'll be when we do. Let's start by adding high-level telemetry for steps 2+3.

ChenyuLInx:
resolves #4625

### Description
Add tracking for compilation time and adapter cache population time. The compilation time and adapter cache time will be sent to snowplow events table as separate event under the action `runnable`. 

### Checklist

### DIFF
--- a/core/dbt/tracking.py
+++ b/core/dbt/tracking.py
@@ -44,6 +44,7 @@ LOAD_ALL_TIMING_SPEC = "iglu:com.dbt/load_all_timing/jsonschema/1-0-3"
 RESOURCE_COUNTS = "iglu:com.dbt/resource_counts/jsonschema/1-0-0"
 EXPERIMENTAL_PARSER = "iglu:com.dbt/experimental_parser/jsonschema/1-0-0"
 PARTIAL_PARSER = "iglu:com.dbt/partial_parser/jsonschema/1-0-1"
+RUNNABLE_TIMING = "iglu:com.dbt/runnable/jsonschema/1-0-0"
 DBT_INVOCATION_ENV = "DBT_INVOCATION_ENV"
 
 
@@ -407,6 +408,18 @@ def track_partial_parser(options):
         active_user,
         category="dbt",
         action="partial_parser",
+        label=get_invocation_id(),
+        context=context,
+    )
+
+
+def track_runnable_timing(options):
+    context = [SelfDescribingJson(RUNNABLE_TIMING, options)]
+    assert active_user is not None, "Cannot track runnable info when active user is None"
+    track(
+        active_user,
+        category="dbt",
+        action="runnable",
         label=get_invocation_id(),
         context=context,
     )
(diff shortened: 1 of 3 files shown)


core/dbt/tracking.py
```diff
@@ -412,6 +413,18 @@ def track_partial_parser(options):
     )
 
 
+def track_runnable_timing(options):
+    context = [SelfDescribingJson(RUNNABLE_TIMING, options)]
+    assert active_user is not None, "Cannot track runnable info when active user is None"
+    track(
+        active_user,
+        category="dbt",
+        action="runnable",
```
jtcohen6, line 422: Could we call this `runnable_timing` (or `runnable_task_startup_timing`) for better clarity/discoverability down the line?

jtcohen6 (approved):
@ChenyuLInx Nice work! Excited to have this in the beta, so we can start validating the data in advance of the final release.

One thing comes to mind: If we end up merging https://github.com/dbt-labs/dbt-core/pull/4860, it would _super_ valuable to track whether a user has opted into the flag/config added there (`cache_selected_only`). I don't think we have any general-purpose tracking for all "global" config values on the invocation.

Since that PR isn't yet merged, it doesn't really make sense to support it in this PR. We can follow up with a new field in the schema and in the tracking event calls.

PR #2850, opened 2020-10-26
Postgres: Prevent temp relation identifiers from being too long

elexisvenator:
related #2197 

### Description

The currently postgres `make_temp_relation` adds a 29 character suffix to the end of the temp relation identifier (9 from default suffix and 20 from timestamp).  This is a problem now that relations with more than 63 characters raise exceptions. 
The fix is to shorten the suffix and also trim the base_relation identifier so that the total length is always less than 63 characters.

An exception can also be raised if the default suffix is overridden with a value that is too long.

### Checklist
 - [x] I have signed the [CLA](https://docs.getdbt.com/docs/contributor-license-agreements)
 - [x] I have run this code in development and it appears to resolve the stated issue
 - [x] This PR includes tests, or tests are not required/relevant for this PR
 - [x] I have updated the `CHANGELOG.md` and added information about my change to the "dbt next" section.

PR #4921, opened 2022-03-22
Truncate relation names when appending a suffix

### LINKED ISSUE #2869: Relation name '*__dbt_tmp' is longer than 63 characters
### Describe the bug

My view name is pretty long, but it is less than 63 characters.

When dbt adds the `__dbt_tmp` suffix, it goes over the limit of 63 chars.

The following error is thrown by PostgreSQL:

>  Relation name 'foo__dbt_tmp' is longer than 63 characters

### Steps To Reproduce

1. Use a view name that is 63 characters long.
2. `dbt run`

### Expected behavior

To work anyways.

See above.

### System information

- [x] postgres
- [ ] redshift
- [ ] bigquery
- [ ] snowflake

Using Docker image `fishtownanalytics/dbt:0.18.1`.

Using Docker image `fishtownanalytics/dbt:0.18.1`.

Using Docker image `fishtownanalytics/dbt:0.18.1`.

### Additional context

Perhaps the temporary name can be truncated by the length of the suffix to make sure it fits within the limits?

epapineau:
Truncate relation names when appending a suffix that will result in len > 63 characters using make_temp_relation and make_backup_relation macros

resolves #2869

### Description

Suffixes are appended to temp and backup relation names. However, these may exceed the 63 character limit on Postgres which currently raises a compiler Error. This PR leverages existing `make_temp_relation` macro and adds `make_backup_relation` and `make_intermediate_relation` macros to truncate base relation name when generated relation name exceeds character length.

### Checklist


epapineau:
Hi, I'm running into an issue pushing further commits to this branch:
```
remote: Resolving deltas: 100% (11/11), completed with 10 local objects.
remote: error: GH006: Protected branch update failed for refs/heads/dev/epapineau.
remote: error: You're not authorized to push to this branch. Visit https://docs.github.com/articles/about-protected-branches/ for more information.
To github.com:dbt-labs/dbt-core.git
 ! [remote rejected]     dev/epapineau -> dev/epapineau (protected branch hook declined)
error: failed to push some refs to 'github.com:dbt-labs/dbt-core.git'
```

Did I configure something incorrectly?

core/dbt/include/global_project/macros/materializations/models/incremental/incremental.sql
```diff
@@ -6,11 +6,12 @@
   {% set target_relation = this.incorporate(type='table') %}
   {% set existing_relation = load_relation(this) %}
   {% set tmp_relation = make_temp_relation(target_relation) %}
+  {%- set tmp_identifier = tmp_relation['identifier'] -%}
```
jtcohen6, line 9: The use of `tmp_idenfier` below is to create the `intermediate_relation` below, which is actually the exact same as `tmp_relation`. We may be able to delete some lines of code, which is always a cool feeling

test/integration/017_runtime_materialization_tests/test_runtime_materialization.py
```diff
@@ -56,23 +55,23 @@ def test_postgres_full_refresh(self):
     @use_profile('postgres')
     def test_postgres_delete__dbt_tmp_relation(self):
         # This creates a __dbt_tmp view - make sure it doesn't interfere with the dbt run
-        self.run_sql_file("create_view__dbt_tmp.sql")
+        #self.run_sql_file("create_view__dbt_tmp.sql")
```
jtcohen6, line 58: Ok, I think I get what's going on here:
- During its materializations, dbt creates these temp + backup objects
- This test ensures that dbt first drops any temp/backup objects by the same name, if they already exist
- Previously, those temp + backup objects had predictable names/suffixes (`view__dbt_tmp`, `view__dbt_backup`)
- Thanks to the changes in this PR, those temp + backup objects have nondeterministic names like `view__dbt_tmp095351072872` and `view__dbt_backup095351077674`

That's better! It's much much less likely for these objects to accidentally collide with a preexisting object (almost impossible). So it makes this test _much_ trickier to define, at the same time, it makes the test significantly less important. I think we can delete this test case (`test_postgres_delete__dbt_tmp_relation`) and explain why.

jtcohen6 (commented):
_(putting on my "Team: Adapters" hat)_

@epapineau Thank you so much for the contribution! I've left a few comments on the organization of the code, and on the failing (no longer needed) test case.

GitHub logistics: I've added you to the CLA list. We do have branch protection rules set up for branches that start with `dev/`, because the base branch used to be called `dev/<version>`. Now it's just called `main`, and we can probably update that rule! I think you're not the last person who will instinctively create a branch prefixed `dev`.

test/integration/017_runtime_materialization_tests/test_runtime_materialization.py
VersusFacit: That as my thinking precisely when Elize and I talked. Glad you concur :)

core/dbt/include/global_project/macros/materializations/models/table/table.sql
```diff
@@ -22,6 +21,7 @@
       See ../view/view.sql for more information about this relation.
   */
   {%- set backup_relation_type = 'table' if old_relation is none else old_relation.type -%}
+  {%- set backup_identifier = make_backup_relation(target_relation)['identifier'] -%}
```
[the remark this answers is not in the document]
epapineau: Yes it can~ I added an arg to the `make_backup_relation` macro so we can still pass `backup_relation_type` and accomplish this in one line.

core/dbt/include/global_project/macros/materializations/models/incremental/incremental.sql
epapineau: Cleaned this up as well 🧹

plugins/postgres/dbt/include/postgres/macros/adapters.sql
```diff
@@ -155,6 +155,22 @@
                                   })) -%}
 {% endmacro %}
 
+{% macro postgres__make_backup_relation(base_relation, suffix) %}
```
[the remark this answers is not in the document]
epapineau: Using your feedback as a jumping off point, this is what came together:

- `make_temp_relation` is used in snapshots and incremental materializations requiring the relation returned to not have `schema` or `database` defined, specifically on Postgres, to avoid `cannot create temporary relation in non-temporary schema`. In order to accomplish [your other suggestions](https://github.com/dbt-labs/dbt-core/pull/4921#discussion_r832994173) in view and table I created a `make_intermediate_relation` macro that returns a relation with `schema` and `database` path keys intact. 
- `postgres__make_temp_relation`, `postgres__make_intermediate_relation`, and `postgres__make_backup_relation` all call a `postgres__make_relation_with_suffix` macro to avoid repeated code and then return the relation for their respective uses. 
- Added bonus of not removing `postgres__make_temp_relation` & alleviating your breaking changes consideration
- In the global project, `default__make_intermediate_relation` calls `default__make_temp_relation` since the non-temporary schema consideration does not apply on other data warehouses (as far as I understand at least - please let me know if this is incorrect).

Finally, I have a question that may affect the utility of `postgres__make_relation_with_suffix`, but they're more appropriate for [the testing thread](https://github.com/dbt-labs/dbt-core/pull/4921#discussion_r833019263).

test/integration/017_runtime_materialization_tests/test_runtime_materialization.py
epapineau: Triple checking that it's appropriate to have backup objects with nondeterministic names as proposed by this PR. I had originally removed that behavior [in the second commit](https://github.com/dbt-labs/dbt-core/pull/4921/commits/d985a8d7387cc98e8335a5c0950a4060e7fd6691) based on [this](https://github.com/dbt-labs/dbt-core/pull/2850#issuecomment-717672329) comment & the failing test, but after the suffix macro recommendations I added it back. Happy to go in either direction, just let me know what y'all think is best @jtcohen6 @VersusFacit and I'll either remove `test_postgres_delete__dbt_tmp_relation` or differentiate the behavior between backup/temp. Once that decision is made, I think the PR is ready for review 🥳

plugins/postgres/dbt/include/postgres/macros/adapters.sql
```diff
@@ -134,27 +134,39 @@
...
 {#
   Postgres tables have a maximum length off 63 characters, anything longer is silently truncated.
-  Temp relations add a lot of extra characters to the end of table namers to ensure uniqueness.
+  Temp and backup relations add a lot of extra characters to the end of table names to ensure uniqueness.
   To prevent this going over the character limit, the base_relation name is truncated to ensure
   that name + suffix + uniquestring is < 63 characters.
 #}
-{% macro postgres__make_temp_relation(base_relation, suffix) %}
+
+{% macro postgres__make_relation_with_suffix(base_relation, suffix) %}
```
jtcohen6, line 142: Ok, maybe this macro takes one more argument, `include_dtstring=true`, which controls whether or not the non-deterministic `dtstring` is included in the suffix.

If we care about avoiding a breaking change in the macro signature (for anyone who's overridden this macro locally), it could be:
```sql
{% macro postgres__make_relation_with_suffix(base_relation, suffix) %}
  ...
  {% if kwargs.get('include_dtstring', true %}
    {% set dtstring = dt.strftime("%H%M%S%f") %}
  {% else %}
    {% set dtstring = "" %}
  {% endif %}
{% endmacro %}
```

test/integration/017_runtime_materialization_tests/test_runtime_materialization.py
jtcohen6: @epapineau Appreciate the triple check! After a bit more thought, I think it might be _very slightly_ preferable to keep relation names non-deterministic _if_ they're creating permanent objects (views + non-temp tables). That way, we avoid cluttering the schema if the materialization fails during the alter-rename-swap-drop step. Deterministically named temp/backup objects get dropped the next time dbt runs that model/etc; actual temp tables get dropped automatically at the end of the session.

That said, failure in that step is _far_ less likely than in the `create table/view as` step, and it's a problem that exists today anyway, whenever users rename a model, and the "old" name of that model doesn't get dropped / cleaned up in the database.

If we did want to keep that distinction—between the temp identifiers used for true temp tables, and our very not-temporary "temporary"/"backup" relations—I think it would just look like one extra argument to these macros that disables the inclusion of `dtstring` in the suffix. That wouldn't be so hard, but it would require a little extra plumbing. It means we could keep this test in place.

What do you think?

epapineau: Alright, done~✨ PR is marked as ready for review.

epapineau:
Hello again~ Following up to see if this needs anything else for review? 🥲

jtcohen6:
@epapineau Thanks for the updated work here!

I started digging in, and found a few places to clarify + confirm this logic. I don't want to expand scope any more than we already have—just want to make sure we're doing the right things. I need a little more time to grok the deleted test, and to clean up my comments.

In the meantime, we _just_ converted the `017_runtime_materialization_tests` to our new testing framework. The test you're removing is now here:
https://github.com/dbt-labs/dbt-core/blob/ab0c3510ebe70a67fb669b88e6ea36aff2f46339/tests/functional/materializations/test_runtime_materialization.py#L179-L185

jtcohen6 (commented):
Great work @epapineau, and really great timing! This serves as a wonderful entrée to the work that the Core team is planning to take on for v1.2, around more ergonomic materializations—cleaning up and consolidating the logic in core, so that we can better streamline it across adapters.

Two actions:
- I believe that the deleted test should actually be passing, since it's checking for "intermediate" rather than "temp" behavior in the incremental materialization
- I opened a new PR against your branch that's slightly more aggressive about renaming things for consistency, and using more concise logic where possible. Take a look and let me know what you think! I know this goes far beyond the original remit, which was just around truncating table names for Postgres... but we might just manage to leave the room much tidier than we found it.

core/dbt/include/global_project/macros/materializations/models/incremental/incremental.sql
```diff
@@ -2,28 +2,25 @@
...
   -- will return None in that case. Otherwise, we get a relation that we can drop
   -- later, before we try to use this name for the current operation. This has to happen before
   -- BEGIN, in a separate transaction
-  {% set preexisting_intermediate_relation = adapter.get_relation(identifier=tmp_identifier,
-                                                                  schema=schema,
-                                                                  database=database) %}
-  {% set preexisting_backup_relation = adapter.get_relation(identifier=backup_identifier,
+  {% set preexisting_temp_relation = adapter.get_relation(identifier=temp_relation['identifier'],
+                                                          schema=schema,
+                                                          database=database) %}
```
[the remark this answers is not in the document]
epapineau: ✨  eureka ✨ 
I've implemented the requested change

test/integration/017_runtime_materialization_tests/test_runtime_materialization.py
```diff
@@ -71,14 +70,6 @@ def test_postgres_delete__dbt_tmp_relation(self):
...
         self.assertTablesEqual("seed", "view")
 
-        # Again, but against the incremental materialization
-        self.run_sql_file("create_incremental__dbt_tmp.sql")
-        results = self.run_dbt(['run', '--model', 'incremental', '--full-refresh'])
-        self.assertEqual(len(results), 1)
-
-        self.assertTableDoesNotExist('incremental__dbt_tmp')
-        self.assertTablesEqual("seed", "incremental")
-
```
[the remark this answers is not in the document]
epapineau: Test has restored, updated, and migrated to the new location

jtcohen6 (changes requested):
@epapineau This looks great to me! Thanks for rolling with the many changes and suggestions :)

Just one tiny thing: What is `tests/functional/simple_seed/data/tmp.csv`? Do we need that? I'm guessing it might be a holdover from a git pull/merge. As soon as that's gone, this is good to merge.

epapineau:
@jtcohen6 Great question! I do not know where that came from. It's been removed.

epapineau:
@jtcohen6 Okay resolved _now_ thanks for catching that

jtcohen6:
Going to rebase this PR against `main` to pull in the fix we merged yesterday for the flaky failing test

jtcohen6 (approved):
Amazing work @epapineau! Thanks so much for contributing, and for seeing this all the way through :)

fx86:
@jtcohen6 @epapineau this issue is still occurring with latest version of dbt-core. I am seeing the full table name followed by the suffix in the error log. Isn't this patch supposed to limit the table name ?

PR #4939, opened 2022-03-23
adapter_methods

ChenyuLInx:
### Description
More adapter test conversion

### Checklist


tests/adapter/dbt/tests/adapter/basic/test_adapter_methods.py
```diff
@@ -0,0 +1,101 @@
...
+            "name": "adapter_methods",
+        }
+
+    # snowflake need all tables in CAP name
+    @pytest.fixture(scope="class")
+    def equal_tables(self):
+        return ["model", "expected"]
+
+    def test_adapter_methods(self, project, equal_tables):
+        run_dbt(["compile"])  # trigger any compile-time issues
```
ChenyuLInx, line 90: @gshank  This works for snowflake, redshift, bigquery and here, it just doesn't work for spark(any type of connection method). the error got thrown was `Table or view not found: test16486637045872746973_test_basic.model`. Do you know what the issue could be?

gshank: Did you set the 'file_format' for models?

tests/adapter/dbt/tests/adapter/basic/test_adapter_methods.py
```diff
@@ -0,0 +1,101 @@
...
+        result = run_dbt()
+        assert len(result) == 3
+        table_comp = TableComparison(
+            adapter=project.adapter, unique_schema=project.test_schema, database=project.database
+        )
+
+        table_comp.assert_tables_equal(*equal_tables)
```
gshank, line 97: We should avoid using TableComparison. I'm removing it in my current branch. 'check_relations_equal' ought to work?

gshank (changes requested):

tests/adapter/dbt/tests/adapter/basic/test_adapter_methods.py
gshank: The 'basic' tests worked on spark, and I would think if there was an issue it would have shown up there. I notice that the 'adapter_test_methods' hasn't been copied to dbt-spark. Maybe there's a reason?

ChenyuLInx: Short story this doesn't work because when we do `dbt compile` we would actually try to run the macro defined [here](https://github.com/dbt-labs/dbt-core/pull/4939/files#diff-a98f96e4919c93a28aee46c131ac9f52bad92403794d7ed03a947c73cb24a96fR9), but in `dbt run` we don't since it is in a test. What's making spark different from others is that when describing a table that doesn't exist, the spark will throw an error and others will return an empty list. Here's the [internal thread](https://dbt-labs.slack.com/archives/C0131TY7EEA/p1649170666282589) about this issue

McKnight-42 (approved):
ran in local, passing for me and code looks great, LGTM

PR #4944, opened 2022-03-23
Cosmetic changelog/changie fixups

### LINKED ISSUE #4943: [CT-415] Reorder changie change kinds
I sense that the order here:

https://github.com/dbt-labs/dbt-core/blob/1c71bf414d36a48b425f6183308a888a7cd30306/.changie.yaml#L10-L16

Maps to the order in the generated changelog [here](https://github.com/dbt-labs/dbt-core/blob/v1.1.0b1/CHANGELOG.md#dbt-core-110-tbd).

As a user of dbt and reader of the changelog, I'd prefer the order to be:
- Breaking changes (side note: in these entries, we need to call out _who_ these might be breaking for)
- Features
- Fixes
- Docs
- Under the hood
- Dependencies

jtcohen6:
resolves #4943

- Reorder change types, both how they'll be ordered in the generated changelog (main concern), and how they'll be presented to contributors for interactive selection
- Reorder v1.1 release notes + changelog accordingly
- Add language to v1.1.0b1 "breaking change" to clarify exactly who this might be breaking for (sorta adapter plugin maintainers, really nobody)
- `Contributors:` &rarr; `### Contributors` (feels right)

**After merging:** Update [v1.1.0b1 release notes](https://github.com/dbt-labs/dbt-core/releases/tag/v1.1.0b1)

### DIFF
--- a/.changie.yaml
+++ b/.changie.yaml
@@ -8,11 +8,11 @@ versionFormat: '## dbt-core {{.Version}} - {{.Time.Format "January 02, 2006"}}'
 kindFormat: '### {{.Kind}}'
 changeFormat: '- {{.Body}} ([#{{.Custom.Issue}}](https://github.com/dbt-labs/dbt-core/issues/{{.Custom.Issue}}), [#{{.Custom.PR}}](https://github.com/dbt-labs/dbt-core/pull/{{.Custom.PR}}))'
 kinds:
-- label: Fixes
-- label: Features
-- label: Under the Hood
 - label: Breaking Changes
+- label: Features
+- label: Fixes
 - label: Docs
+- label: Under the Hood
 - label: Dependencies
 custom:
 - key: Author
--- a/.changie.yaml
+++ b/.changie.yaml
@@ -28,7 +28,7 @@ custom:
   type: int
   minLength: 4
 footerFormat: |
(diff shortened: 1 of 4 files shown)


.changie.yaml
```diff
@@ -28,7 +28,7 @@ custom:
   type: int
   minLength: 4
 footerFormat: |
-  Contributors:
+  ### Contributors
```
jtcohen6, line 31: This one is a no-look pass, I could be wildly off base

emmyoop: It's easy to test all your changes work as expected!

Run this to print out what the generated yaml file will be:

```changie batch 1.1.0 --dry-run --include '1.1.0'```

`--dry-run` just prints out what would be generated and doesn't actually create a file or delete any yaml and `--include` will include the yaml files in that directory.

Note: Two new changes have been merged in since the release (2 files exists in `.changes/unreleased`) so those will be erroneously included.  Since it's not creating the files it doesn't matter.  This command should be enough to eyeball that the file is generated like you expect.

jtcohen6: realized I needed to update my installed version of `changie`! after that, works like a charm

emmyoop (approved):
LGTM

Issue #4882, opened 2022-03-16
[Feature] <title>


McKnight-42:
### Is there an existing feature request for this?

- [X] I have searched the existing issues

### Describe the Feature

Next steps for work for https://github.com/dbt-labs/dbt-core/pull/4858

- Rewriting these tests to use the new pytest framework
- Make them truly cross-database compatible (I think by turning expected models into seeds)
- Add them to the inheritable adapter tests, and turn them on for Redshift + Snowflake + BigQuery + Spark. Delete the copy--pasted versions present in those repos :)

### Who will this benefit?

Tests will be much cleaner, and a lot less copy paste between repos

jtcohen6:
Anticipated process for this work:

1. Rewrite the test, added to this repo by #4858, using new pytest framework
2. Rewrite the logic to be database-agnostic, so the exact same code can run everywhere
    - [x] Less-database-specific error message — handled in #4858
    - [ ] Using seeds instead of SQL literals to load expected fixtures, to avoid database-specific data type casting
3. Move these tests into the "adapter testing zone" within this repo
4. For each plugin: Inherit the test cases from the core repo, and delete the copy-pasted code that's currently in that plugin.
    - For Snowflake, we need to test with both incremental strategies: `merge` (default) + `delete+insert`

jtcohen6:
@leahwicz As discussed yesterday, we're thinking that the group doing this work might split and do it in parallel with another "adapter zone" test conversion.

My vote would be for [`test_docs_generate`](https://github.com/dbt-labs/dbt-core/tree/main/test/integration/029_docs_generate_tests), which has proven especially obnoxious for us in the adapter plugins, requiring frequent copy-paste updates. I think we still want some testing for those capabilities in plugins, since catalog contents can vary. A nice exercise in pytest fixture overrides :)

PR #4958, opened 2022-03-25
init push up of converted unique_key tests

McKnight-42:
resolves https://github.com/dbt-labs/dbt-core/issues/4882

### Description
Converting the new unique_id_as_list tests to new pytest version in core and then in each adapter as needed.
### Checklist


jtcohen6:
@McKnight-42 Very glad to see an early draft PR!

There's a namespacing question here:
- Do we want this test to live in `tests/adapter/basic`, alongside the "baseline" dbt behavior? _You must be this tall to call yourself a dbt adapter_
- Or still within the "adapter zone," but in a way that signals it's more of an opt-in? `tests/adapter/optional`, `tests/adapter/advanced`, ...?

gshank:
That's a good question about where it should live. At some point we probably do want to put the required tests into one directory. Adapter tests will have to end up in the "adapter zone", which is in a directory in tests/adapter/dbt/tests/adapter.  That's where they have to be in order to be subclasses in the adapter repos. So right now I'd put them in tests/adapter/dbt/tests/adapter/incremental_unique_id. Then when we create override tests they can be in tests/functional/adapter/test_incremental_unique_id.py. We might want to reorganize them later.

McKnight-42:
> That's a good question about where it should live. At some point we probably do want to put the required tests into one directory. Adapter tests will have to end up in the "adapter zone", which is in a directory in tests/adapter/dbt/tests/adapter. That's where they have to be in order to be subclasses in the adapter repos. So right now I'd put them in tests/adapter/dbt/tests/adapter/incremental_unique_id. Then when we create override tests they can be in tests/functional/adapter/test_incremental_unique_id.py. We might want to reorganize them later.

Moved test up one directory to be on level with with basic. will keep looking into the errors that are popping up, have we picked the first adapter we want to convert in?

tests/adapter/dbt/tests/adapter/test_incremental_unique_id.py
```diff
@@ -0,0 +1,597 @@
...
+        )
+        test_case_fields = ResultHolder(
+            *self.setup_test(seed, incremental_model, update_sql_file, project),
+            opt_model_count=self.update_incremental_model(expected_model),
+            relation=incremental_model,
+        )
+
+        self.check_scenario_correctness(expected_fields, test_case_fields, project)
+
+    def test__bad_unique_key(self):
```
gshank, line 487: You don't have a project fixture on this test.

tests/adapter/dbt/tests/adapter/test_incremental_unique_id.py
```diff
@@ -0,0 +1,597 @@
...
+def seeds():
+    return {
+        "duplicate_insert.sql": seeds__duplicate_insert_sql,
+        "seed.csv": seeds__seed_csv,
+        "add_new_rows.sql": seeds__add_new_rows_sql,
+    }
+
+
+@pytest.fixture(scope="class")
+def project_files(
```
gshank, line 339: This fixture should be unncessary.

tests/adapter/dbt/tests/adapter/test_incremental_unique_id.py
McKnight-42: oh thank you wasn't sure if it would still need to take in the project as it doesn't seem to call on it since it just passes the fail case.

tests/adapter/dbt/tests/adapter/test_incremental_unique_id.py
```diff
@@ -0,0 +1,597 @@
...
+            expect_pass=True,
+        )
+        run_result = run_dbt(
+            ["run", "--select", incremental_model_name], expect_pass=False
+        ).results[0]
+
+        return run_result.status, run_result.message
+
+
+class TestIncrementalWithoutUniqueKey(IncrementalUniqueKeyBase):
```
gshank, line 433: None of these tests are actually importable, because they start with the name 'Test'.  Why aren't the tests in the base test class?

gshank (changes requested):
The tests need to be in the base test class so they can be importable.

tests/adapter/dbt/tests/adapter/test_incremental_unique_id.py
gshank: If you don't have a project fixture, you don't have a project to do dbt_run against.

McKnight-42: ah, seeing that its a unused arg threw me off but that makes sense, thanks for the context.

tests/adapter/dbt/tests/adapter/test_incremental_unique_id.py
McKnight-42: removed and tested locally

tests/adapter/dbt/tests/adapter/test_incremental_unique_id.py
McKnight-42: the original tests were separated out into classes that inherited the setup based on type of unique_key case it was testing. do we have a preferred way to reformat the test/class names?

McKnight-42: moved all tests to the base class and created a new test class based on the other examples we have.

jtcohen6: Ah, this conversation clarified for me why we're doing it like this:

https://github.com/dbt-labs/dbt-core/blob/5071b00baac3ff2a9bcce565d09661605de52654/tests/adapter/dbt/tests/adapter/basic/test_base.py#L109-L110

As I understand it:
- By default, `pytest` only runs tests if the method starts with `test_`, and the class starts with `Test`, in a file whose name starts with `test_`
- Our move has been to create a "base" class (_not_ named `Test`), with test methods defined on it that _do_ start with `test_*`
- Then, inherit that "base" class into an actual test class, whose name starts with `Test`, and either pass (no changes needed) or override/reimplement fixtures/methods from the base class

tests/adapter/dbt/tests/adapter/test_incremental_unique_id.py
```diff
@@ -0,0 +1,613 @@
...
+    config(
+        materialized='incremental',
+        unique_key=['state', 'county', 'city']
+    )
+}}
+
+select
+    state::varchar(2) as state,
+    county::varchar(12) as county,
+    city::varchar(12) as city,
```
jtcohen6, line 22: We'll want to be thoughtful about data types in order to get this running successfully on multiple databases. E.g. BigQuery doesn't support a data type named `varchar(length)`. Our options here:
- Override/reimplement all fixtures when we run this test in `dbt-bigquery` (ugh)
- Remove type casts here, try to handle all data type business via seeds instead. That may also require a different approach for `add_new_rows.sql`/`duplicate_insert.sql`

tests/adapter/dbt/tests/adapter/test_incremental_unique_id.py
gshank: Yes, that's correct :-). You *can* import a base class that starts with Test, the problem is that pytest would then run all of the tests in the base class plus all of those same tests in the subclass. Which means that if you need to override a test method or get failures, you'd get failures.

tests/adapter/dbt/tests/adapter/test_incremental_unique_id.py
gshank: These tests are also working on the adapter repos? Did you change the datatypes?

gshank: We could provide some of that information (with data_types) in a custom fixture. Avoiding that would be even better. There is actually an update_rows function in dbt.tests.util which came from dbt-adapter-tests and does work on all of our repos because it calls an adapter method.

Also, I think putting the test settings in variables and then putting them into the structures isn't the best practice. I'm gong to look at it a bit more to come up with recommendations.

test/integration/076_incremental_unique_id_test/test_incremental_unique_id.py
```diff
@@ -4,8 +4,8 @@
 from pathlib import Path
 
 
-TestResults = namedtuple(
-    'TestResults',
+ResultHolder = namedtuple(
```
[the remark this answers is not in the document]
McKnight-42: yes was holding off but we can delete it.

tests/adapter/dbt/tests/adapter/test_incremental_unique_id.py
jtcohen6: > These tests are also working on the adapter repos? Did you change the datatype

Yes, I believe there are slight modifications for each adapter repo

> There is actually an update_rows function in dbt.tests.util which came from dbt-adapter-tests and does work on all of our repos because it calls an adapter method.

I'm a bit skeptical of this method tbh (I know it required us to override a data type in Redshift for basic tests), and want to discuss alongside table comparison logic—but maybe it's the way to go!

Maybe the move should be:
- Once this test case is passing in this repo (for Postgres), we try running it (from this branch) in a branch of each plugin repo as well
- Update the test cases here, override them there, get them all passing for all plugins
- Fine-tune the framework so that the overrides feel minimal + appropriate
- Merge across the board (and delete all the previously duplicated test code)

gshank: That sounds like the right process.

The override in Redshift for a seed data column was because the test goes and updates the name by appending '_update'. Redshift created the seed table with a varchar datatype that wasn't big enough. The best solution might actually have been to change that update so the size didn't increase, but the 'update_rows' method was kind of fiddly.

tests/adapter/dbt/tests/adapter/test_incremental_unique_id.py
```diff
@@ -0,0 +1,584 @@
...
+            inc_test_model_count=1,
+            opt_model_count=opt_model_count,
+            relation=relation,
+        )
+
+    def fail_to_build_inc_missing_unique_key_column(self, incremental_model_name):
+        """should pass back error state when trying build an incremental
+        model whose unique key or keylist includes a column missing
+        from the incremental model"""
+        # seed_count = len(run_dbt(["seed", "--select", "seed", "--full-refresh"]))  # noqa:F841
```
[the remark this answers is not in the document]
McKnight-42: will do was testing how needed this was as the `seed_count` variable isn't actually being passed onto anything.

McKnight-42:
https://github.com/dbt-labs/dbt-redshift/pull/92 redshift seems to be passing tests locally as we would hope

tests/adapter/dbt/tests/adapter/test_incremental_unique_id.py
```diff
@@ -0,0 +1,584 @@
...
+            expect_pass=True,
+        )
+        run_result = run_dbt(
+            ["run", "--select", incremental_model_name], expect_pass=False
+        ).results[0]
+
+        return run_result.status, run_result.message
+
+    # no unique_key test
+    def test__no_unique_keys(self, project):
```
gshank, line 419: I think the factoring in the way these tests are done could be more straightforward and easier to read. I would recommend that you do something like:

```
  def test__no_unique_keys(self, project):
        """with no unique keys, seed and model should match"""

        expected_fields = self.get_expected_fields(relation="seed", seed_rows=8)
        test_case_fields = self.get_test_fields(
            project, seed="seed", incremental_model="no_unique_key", update_sql_file="add_new_rows"
        )
        self.check_scenario_correctness(project, expected_fields, test_case_fields)

```

gshank: Rename stub_expected_fields to 'get_expected_fields'. Rename 'setup_test' to 'get_test_fields'. Refactor get_test_fields to return the ResultHolder object and change the signature to use keywords, except for project, which should be first.

gshank:
Let's put 'test_incremental_unique_id.py' into an 'incremental' directory. Then when we convert the other incremental tests they will all be in the same directory. This will reduce the number of files that adapter repos will have to create to pull in adapter tests.

gshank (changes requested):

McKnight-42:
> Let's put 'test_incremental_unique_id.py' into an 'incremental' directory. Then when we convert the other incremental tests they will all be in the same directory. This will reduce the number of files that adapter repos will have to create to pull in adapter tests.

would this be within the dbt/tests/adapter area or would we want it on same level as functional and the other directories?

gshank:
Within the adapter zone, so tests/adapter/dbt/tests/adapter/incremental

ChenyuLInx:
> > Let's put 'test_incremental_unique_id.py' into an 'incremental' directory. Then when we convert the other incremental tests they will all be in the same directory. This will reduce the number of files that adapter repos will have to create to pull in adapter tests.
> 
> would this be within the dbt/tests/adapter area or would we want it on same level as functional and the other directories?

I believe we probably want the level like what Gerda suggested and for other repos, it would be just one file running all of the tests for 'incremental'.

gshank (approved):
Looks good!

McKnight-42:
@cla-bot[bot] check

Issue #4280, opened 2021-11-15
[CT-531] [Bug] More robust error messaging when keys are misspelled in yml files


saraleon1:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

A PR jobs started failing with no clear error message available to the end user. The only error surfaced was:

```
Error message:
Running with dbt=0.21.0
* Deprecation Warning: The “packages” argument of adapter.dispatch() has been
deprecated. Use the “macro_namespace” argument instead.
Raised during dispatch for: type_string
For more information, see:
https://docs.getdbt.com/reference/dbt-jinja-functions/dispatch
Found 347 models, 448 tests, 0 snapshots, 0 analyses, 701 macros, 2 operations, 21 seed files, 302 sources, 0 exposures
Found a seed (<removed>) >1MB in size at the same path, dbt cannot tell if it has changed: assuming they are the same
```

upon inspection of debug logs, the following log line was identified:

```
2021-11-04 08:50:01.342766 (MainThread): None
```

and upon inspection of more robust logging available to dbt cloud support, this stack trace was identified:

```
Traceback (most recent call last):
  File "/usr/src/app/sinter/clients/dbt.py", line 1147, in call
    result, success = dbt_main.handle_and_check(
  File "/usr/local/lib/python3.8/dist-packages/dbt/main.py", line 205, in handle_and_check
    task, res = run_from_args(parsed)
  File "/usr/local/lib/python3.8/dist-packages/dbt/main.py", line 258, in run_from_args
    results = task.run()
  File "/usr/local/lib/python3.8/dist-packages/dbt/task/runnable.py", line 433, in run
    self._runtime_initialize()
  File "/usr/local/lib/python3.8/dist-packages/dbt/task/runnable.py", line 155, in _runtime_initialize
    self.job_queue = self.get_graph_queue()
  File "/usr/local/lib/python3.8/dist-packages/dbt/task/runnable.py", line 146, in get_graph_queue
    return selector.get_graph_queue(spec)
  File "/usr/local/lib/python3.8/dist-packages/dbt/graph/selector.py", line 297, in get_graph_queue
    selected_nodes = self.get_selected(spec)
  File "/usr/local/lib/python3.8/dist-packages/dbt/graph/selector.py", line 278, in get_selected
    selected_nodes, indirect_only = self.select_nodes(spec)
[...]

Issue #2606, opened 2020-06-30
Feature Request: Strictness on arguments passed to config


smile-simon:
### Describe the feature
Have some sort of strictness in place for arguments passed to config. For example:
Currently, if you use `{{ config(materizalied='table') }}`, due to the typo in `materialized`, this model will be materialized as a `view` (default) and not as a `table`. This model will run successfully. No errors or warnings are thrown by DBT for this typo because DBT explicitly allows arbitrary arguments. 

This caused me some headache for a while because I couldn't understand why this model wouldn't materialize as a `table`.

### Describe alternatives you've considered
I have limited knowledge in this sort of stuff, but one solution I propose is to have an array/list somewhere of accepted values of arguments.
- If an argument is specified in the call, but is not in the array of accepted values, then an error/warning is thrown

### Additional context
I don't believe this is database specific. I use Snowflake but I'm not sure it's relevant. 

### Who will this benefit?
This will benefit most people as it will catch typo's that won't be caught by DBT.

### Are you interested in contributing this feature?
I'm still a beginner so I wouldn't want to participate. Maybe in the near future I'll have the knowledge to confidently contribute.

jtcohen6:
Hey @smile-simon, thanks for the writeup—this is a great idea. dbt does maintain lists of core and adapter-specific node configs. Giving developers a way to more quickly debug small typos feels like an obvious quality-of-life improvement.

One thing I could imagine is extending the functionality of the [`--strict` run mode](https://docs.getdbt.com/docs/running-a-dbt-project/command-line-interface/global-cli-flags/#strict) to raise CLI warnings for meaningless configs, similar to the warnings we raise when dbt finds YML resource configs for a source or model that is disabled or doesn't exist.

smile-simon:
Hmm, the custom materializations definitely complicates the idea. 

My thoughts here are:
If someone makes a custom materialization then it will live in the `macros` folder. For materializations specifically, the macro code is:

```
{%- materialization my_custom_materialization, default -%}
stuff
{%- endmaterialization -%}
```

Would it be possible to append the name of these macros (`my_custom_materialization`) to the list of "acceptable values" (along with the dbt list of core adapter-specific node configs)? So in this instance, we append `my_custom_materialization` to the list of acceptable values? Then we combine our ideas in our previous messages so that `--strict run` will raise a warning if a value is outside of the list.

[...]

PR #4960, opened 2022-03-25
initial pass at source config tests w/o overrides

### LINKED ISSUE #3662: [CT-201] Reconcile configs + properties for sources
Follow-up to #2401, #3616

```yml
# models/src_whatever.yml
version: 2
sources:
  - name: my_source
    config:
      enabled: ...
      quoting: {dict}
      freshness: {dict}
      loader: ...
      loaded_at_field: ...
      meta: {dict}
      tags: ...
    tables:
      - name: my_src_table
        tests: ... # not a config
        columns: ... # not a config
```

```yml
# dbt_project.yml
sources:
  project_name:
    subdirectory:
      +database: raw
      +loader: fivetran
      another_subdirectory:
        +enabled: false
```

```yml
sources:
  - name: my_source
    loaded_at_field: updated_at
```

```yml
sources:
  - name: my_source
    loaded_at_field: updated_at
    config:
      loaded_at_field: updated_at
```

### Notes

emmyoop:
resolves #3662 

### Description

This draft just has the first pass at new tests we would want to add.  It does not include anything for overrides.  Most tests will fail and are marked as such.

I'm looking for feedback on
1. Is the expected new behavior?
2. Are there any missing tests?

### Checklist

### DIFF
--- a/core/dbt/contracts/graph/model_config.py
+++ b/core/dbt/contracts/graph/model_config.py
@@ -335,6 +335,16 @@ class BaseConfig(AdditionalPropertiesAllowed, Replaceable):
 @dataclass
 class SourceConfig(BaseConfig):
     enabled: bool = True
+    quoting: Optional[Dict[str, Any]] = None
+    freshness: Optional[Dict[str, Any]] = None
+    loader: Optional[str] = None
+    loaded_at_field: Optional[
+        Any
+    ] = None  # TODO what type is this? docs say: "<column_name_or_expression>"
+    database: Optional[str] = None
+    schema: Optional[str] = None
+    meta: Optional[Dict[str, Any]] = None
+    tags: Optional[List[str]] = None
 
 
 @dataclass
--- a/core/dbt/contracts/graph/model_config.py
+++ b/core/dbt/contracts/graph/model_config.py
@@ -338,9 +338,8 @@ class SourceConfig(BaseConfig):
     quoting: Optional[Dict[str, Any]] = None
     freshness: Optional[Dict[str, Any]] = None
     loader: Optional[str] = None
-    loaded_at_field: Optional[
-        Any
-    ] = None  # TODO what type is this? docs say: "<column_name_or_expression>"
+    # TODO what type is this? docs say: "<column_name_or_expression>"
+    loaded_at_field: Optional[Any] = None
     database: Optional[str] = None
     schema: Optional[str] = None
     meta: Optional[Dict[str, Any]] = None
--- a/core/dbt/contracts/graph/model_config.py
+++ b/core/dbt/contracts/graph/model_config.py
@@ -335,15 +335,39 @@ class BaseConfig(AdditionalPropertiesAllowed, Replaceable):
 @dataclass
 class SourceConfig(BaseConfig):
     enabled: bool = True
-    quoting: Optional[Dict[str, Any]] = None
-    freshness: Optional[Dict[str, Any]] = None
-    loader: Optional[str] = None
+    quoting: Dict[str, Any] = field(
+        default_factory=dict,
+        metadata=MergeBehavior.Update.meta(),
+    )
+    freshness: Optional[Dict[str, Any]] = field(
+        default=None,
+        metadata=CompareBehavior.Exclude.meta(),
+    )
+    loader: Optional[str] = field(
+        default=None,
+        metadata=CompareBehavior.Exclude.meta(),
+    )
     # TODO what type is this? docs say: "<column_name_or_expression>"
-    loaded_at_field: Optional[Any] = None
-    database: Optional[str] = None
-    schema: Optional[str] = None
-    meta: Optional[Dict[str, Any]] = None
-    tags: Optional[List[str]] = None
+    loaded_at_field: Optional[str] = field(
+        default=None,
+        metadata=CompareBehavior.Exclude.meta(),
+    )
+    database: Optional[str] = field(
+        default=None,
+        metadata=CompareBehavior.Exclude.meta(),
+    )
+    schema: Optional[str] = field(
+        default=None,
+        metadata=CompareBehavior.Exclude.meta(),
+    )
+    meta: Dict[str, Any] = field(
+        default_factory=dict,
+        metadata=MergeBehavior.Update.meta(),
+    )
+    tags: Union[List[str], str] = field(
+        default_factory=list_str,
+        metadata=metas(ShowBehavior.Hide, MergeBehavior.Append, CompareBehavior.Exclude),
+    )
 
 
 @dataclass
--- a/tests/functional/sources/test_source_configs.py
+++ b/tests/functional/sources/test_source_configs.py
@@ -22,7 +22,6 @@ class SourceConfigTests:
             # loaded_at_field = some_column
             # database = custom_database
             # schema = custom_schema
-            # # identifier = "seed"  #this doesnt seems to be supported now?
             # meta = {'languages': ['python']}
             # tags = ["important_tag"]
         )
@@ -62,7 +61,7 @@ class TestSourceEnabledConfigProjectLevel(SourceConfigTests):
         }
 
     def test_enabled_source_config_dbt_project(self, project):
-        run_dbt(["compile"])
+        run_dbt(["parse"])
         manifest = get_manifest(project.project_root)
         assert "source.test.test_source.test_table" in manifest.sources
 
@@ -76,7 +75,7 @@ class TestSourceEnabledConfigProjectLevel(SourceConfigTests):
             }
         }
         update_config_file(new_enabled_config, project.project_root, "dbt_project.yml")
-        run_dbt(["compile"])
+        run_dbt(["parse"])
         manifest = get_manifest(project.project_root)
 
         assert (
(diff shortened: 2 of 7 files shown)


tests/functional/sources/test_source_configs.py
```diff
@@ -0,0 +1,594 @@
...
+            # freshness = FreshnessThreshold(
+            #     warn_after=Time(count=12, period=TimePeriod.hour),
+            #     error_after=Time(count=24, period=TimePeriod.hour),
+            #     filter=None
+            #     )
+            # loader = "a_loader"
+            # loaded_at_field = some_column
+            # database = custom_database
+            # schema = custom_schema
+            # # identifier = "seed"  #this doesnt seems to be supported now?
```
jtcohen6, line 25: Good callout. This is necessarily different by table, so it wouldn't make sense to configure for many tables at once.

tests/functional/sources/test_source_configs.py
```diff
@@ -0,0 +1,594 @@
...
+  - name: test_source
+    config:
+      enabled: False
+    tables:
+      - name: test_table
+      - name: disabled_test_table
+"""
+
+
+# Test enabled config at sources level in yml file
```
jtcohen6, line 100: 👍

tests/functional/sources/test_source_configs.py
```diff
@@ -0,0 +1,594 @@
...
+  - name: test_source
+    tables:
+      - name: test_table
+      - name: disabled_test_table
+        config:
+            enabled: False
+"""
+
+
+# Test enabled config at source table level in yaml file
```
jtcohen6, line 129: 👍

tests/functional/sources/test_source_configs.py
```diff
@@ -0,0 +1,594 @@
...
+                "test": {
+                    "test_source": {
+                        "enabled": True,
+                    },
+                }
+            }
+        }
+
+    def test_enabled_source_config_dbt_project(self, project):
+        run_dbt(["compile"])
```
jtcohen6, line 65: `compile` gets the job done; I'm also thinking this could just be `parse`, since we _just_ need the output of parsing. That's one way to draw the interface boundary, and denote this as Language behavior.

Do we want to develop consistent guidance around which commands we use, as a proxy for the functionality they're testing? Or are there other better ways to decorate tests, to denote their surface area?

tests/functional/sources/test_source_configs.py
```diff
@@ -0,0 +1,594 @@
...
+
+# Test all configs other than enabled in dbt_project.yml
+# expect fail - not implemented
+class TestAllConfigsProjectLevel(SourceConfigTests):
+    @pytest.fixture(scope="class")
+    def models(self):
+        return {"schema.yml": models__schema_yml}
+
+    @pytest.fixture(scope="class")
+    def project_config_update(self):
```
jtcohen6, line 154: 👍

tests/functional/sources/test_source_configs.py
```diff
@@ -0,0 +1,594 @@
...
+        meta: {'languages': ['python']}
+        tags: ["important_tag"]
+    tables:
+      - name: test_table
+      - name: other_test_table
+"""
+
+
+# Test configs other than enabled at sources level in yaml file
+# **currently passes since enabled is all that ends up in the
```
jtcohen6, line 218: Fair question: should we expect this to fail?

We let users set whichever configs they want in `dbt_project.yml`, regardless of whether they take any effect, since model/seed/snapshot/test configs might be used at runtime in the associated materialization. That does make it harder to raise error in the case of typos (https://github.com/dbt-labs/dbt-core/issues/2606, https://github.com/dbt-labs/dbt-core/issues/4280).

Sources aren't materialized... the closest thing I could think is, for `dbt source freshness`, if users want to pass totally custom configs for use in a custom `collect_freshness` macro. They _could_ achieve that with `config.meta`.

tests/functional/sources/test_source_configs.py
```diff
@@ -0,0 +1,594 @@
...
+            database: custom_database
+            schema: custom_schema
+            # identifier: "seed"  #this doesnt seems to be supported now?
+            meta: {'languages': ['python']}
+            tags: ["important_tag"]
+      - name: other_test_table
+"""
+
+
+# Test inheritence - set configs atproject, source, and source-table level - expect source-table level to win
```
jtcohen6, line 323: 👍

tests/functional/sources/test_source_configs.py
```diff
@@ -0,0 +1,594 @@
...
+        # identifier: "seed"  #this doesnt seems to be supported now?
+        meta: {'languages': ['python']}
+        tags: ["important_tag"]
+    tables:
+      - name: test_table
+      - name: other_test_table
+"""
+
+
+# Test inheritence - set configs at project and source level - expect source level to win
```
jtcohen6, line 386: 👍

tests/functional/sources/test_source_configs.py
```diff
@@ -0,0 +1,594 @@
...
+            database: custom_database
+            schema: custom_schema
+            # identifier: "seed"  #this doesnt seems to be supported now?
+            meta: {'languages': ['python']}
+            tags: ["important_tag"]
+      - name: other_test_table
+"""
+
+
+# Test inheritence - set configs at project and source-table level - expect source-table level to win
```
jtcohen6, line 442: 👍

tests/functional/sources/test_source_configs.py
```diff
@@ -0,0 +1,594 @@
...
+
+        assert config_other_test_table == expected_source_level_config
+
+
+all_configs_not_table_schema_yml = """version: 2
+
+sources:
+  - name: test_source
+    config:
+        enabled: True,
```
jtcohen6, line 364:
```suggestion
        enabled: True
```

tests/functional/sources/test_source_configs.py
```diff
@@ -0,0 +1,594 @@
...
+all_configs_everywhere__schema_yml = """version: 2
+
+sources:
+  - name: test_source
+    config:
+        tags: "source_level_important_tag",
+    tables:
+      - name: test_table
+        config:
+            enabled: True,
```
jtcohen6, line 303:
```suggestion
            enabled: True
```

tests/functional/sources/test_source_configs.py
```diff
@@ -0,0 +1,594 @@
...
+
+    def test_source_configs_as_properties(self, project):
+        run_dbt(["compile"])
+        manifest = get_manifest(project.project_root)
+        assert "source.test.test_source.test_table" in manifest.sources
+        config_test_table = manifest.sources.get("source.test.test_source.test_table").config
+
+        # this is new functionality - but it currently passes since SourceConfig is not updated
+        # and is commented out in the setup becuse it is not updated with new configs
+        # even when properties are defined at teh top level they should end up on the node.config
```
jtcohen6, line 517: > even when properties are defined at teh top level they should end up on the node.config

This should go both ways:
- When properties are defined as top-level keys (backwards compatibility for user code), they should end up in `node.config`
- Once `node.config` has been resolved, those configurations should be copied back as top-level properties (backwards compatibility for metadata consumers, including `dbt-docs` and the metadata API)

The downside is, more repetition in `metadata.json`, which is already a verbose file. I think it's worth it to ease backwards compatibility for metadata. Eventually, I could see our metadata interface including top-level node keys (well typed and validated), and excluding `node.config`, which is as `dict[Any]` as they come, and really more of an intermediate object for internal use.

tests/functional/sources/test_source_configs.py
```diff
@@ -0,0 +1,594 @@
...
+        # identifier: "seed"  #this doesnt seems to be supported now?
+        meta: {'languages': ['python']}
+        tags: ["important_tag"]
+    tables:
+      - name: test_table
+      - name: other_test_table
+"""
+
+
+# Raise an error when properties are set at top level and also as configs
```
jtcohen6, line 582: 👍

jtcohen6 (commented):
Test cases look good! Left some comments. TDD in dbt, who would've thought?

tests/functional/sources/test_source_configs.py
emmyoop: That makes a lot of sense.  `parse` it is!

I think the legacy (can I call them that yet?) tests use `run` way more often than is strictly needed.  I'm not sure what guidance we would develop short of thinking through what we really are testing.

We do have the option with pytest to create [custom markers](https://docs.pytest.org/en/6.2.x/example/markers.html) to mark what team a test applies to.  But I think it's important to define why we need to do this and what value it would add.

core/dbt/contracts/graph/model_config.py
```diff
@@ -335,6 +335,15 @@ def replace(self, **kwargs):
 @dataclass
 class SourceConfig(BaseConfig):
     enabled: bool = True
+    quoting: Optional[Dict[str, Any]] = None
+    freshness: Optional[Dict[str, Any]] = None
+    loader: Optional[str] = None
+    # TODO what type is this? docs say: "<column_name_or_expression>"
```
emmyoop, line 341: A string.  It's literally the name of the column that holds the `loaded_at` time.  It's also defined as a string [here](https://github.com/dbt-labs/dbt-core/blob/7f953a6d48a4db254b17fd3929d267e4b6069228/core/dbt/contracts/graph/parsed.py#L629)

emmyoop:
Closing in favor of #5003

Issue #4342, opened 2021-11-26
[Bug] dbt deps sometimes takes >90sec to complete


joellabes:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Over the last couple of weeks, I've noticed that `dbt deps` sometimes takes >90 seconds between `Running with dbt==0.21.0` and the first `Installing...` line.

It's sporadic, and when I re-run to try and capture it performance improves, which makes me suspect it's some sort of cold cache thing. 

### Expected Behavior

<img alt="Let's go faster" src="https://media.giphy.com/media/3ohuPuRnWLwIsCuHPa/giphy.gif">

### Steps To Reproduce

With this packages.yml file (from the internal-analyics project): 
```
packages:
  - local: '../dbt-utils'
  - package: dbt-labs/snowplow
    version: 0.13.0
  - package: dbt-labs/codegen
    version: 0.4.0
  - package: tailsdotcom/dbt_artifacts
    version: 0.5.0
  - package: fivetran/hubspot
    version: 0.4.0
  - package: fivetran/zendesk_source
    version: 0.4.0
  - package: dbt-labs/audit_helper
    version: 0.4.0
```

run `dbt deps`. I was on the VPN, but from memory I've experienced it off the VPN as well. 

Note that this combination will ultimately result in a duplicate package error due to codegen depending on utils as well, but I don't think that's relevant. 

### Relevant log output

```shell
https://www.loom.com/share/95a1c1f62c304673a4649d609aa98781
```

### Environment

```markdown
- OS: macOS 11.6.1
- Python: Python 3.8.10
- dbt: I've noticed this on 0.21.0 as well as various prerelease versions of 1.0
```

### What database are you using dbt with?

snowflake

jtcohen6:
@joellabes Thanks for opening! I've noticed this too.

I believe the significant delay comes from the fact that dbt needs to make _many_ hub API requests in order to "discover" available package versions:
```
2021-11-29 10:25:41.507359 (MainThread): Set downloads directory='/var/folders/7h/hj5_fw9j291c58hwfdvy5xbm0000gp/T/dbt-downloads-nlq7cw6p'
2021-11-29 10:25:41.507605 (MainThread): Making package registry request: GET https://hub.getdbt.com/api/v1/index.json
2021-11-29 10:25:41.786837 (MainThread): Response from registry: GET https://hub.getdbt.com/api/v1/index.json 200
[...]

PR #4982, opened 2022-03-30
[CT-352] catch and retry malformed json

### LINKED ISSUE #4849: [CT-352] Catch all Requests Exception on Deps Install and Retry

emmyoop:
resolves #4849, #4342 

### Description

There are still some exceptions when trying to install deps that fail without retrying.  This PR tries to catch the rest of them.

It also caches the HUB response to reduce calls.  See #4983 for some more detailed context.

### Checklist


core/dbt/clients/registry.py
```diff
@@ -31,18 +36,34 @@ def _get(path, registry_base_url=None):
...
+            "Request error: The response is not valid json", response=resp
         )
     return resp.json()
 
 
+def validJson(jsonData):
+    try:
+        # the quotes in the response were converted to single quotes which is not
+        # valid json.  json.dumps() will fix that.
+        json.loads(json.dumps(jsonData))
```
emmyoop, line 55: This feels bad.  Any suggestions on a better way to do this welcome!

emmyoop: The comment doesn't state this yet but the `api/v1/index.json` path also returns a list objects that causes it fail.

core/dbt/clients/registry.py
```diff
@@ -80,7 +98,12 @@ def package(name, registry_base_url=None):
 
 
 def package_version(name, version, registry_base_url=None):
-    return _get_with_retries("api/v1/{}/{}.json".format(name, version), registry_base_url)
+    # returns a dictionary of metadata for a single version of a package
+    expected_type = dict
+    expected_keys = ["name", "packages", "downloads"]
+    return _get_with_retries(
+        "api/v1/{}/{}.json".format(name, version), registry_base_url, expected_type, expected_keys
+    )
```
jtcohen6, line 106: We can avoid a lot of superfluous requests by making this a filter on the `package` request response, rather than a separate request. This is what I'm trying over in https://github.com/dbt-labs/dbt-core/pull/4983:

https://github.com/dbt-labs/dbt-core/blob/25d4c96d86bd04d9bf271e8e18d5a0b8ed8295e5/core/dbt/clients/registry.py#L85-L87

But we're not validating `["name", "packages", "downloads"]` as expected keys of the `package` response there, just the presence of a `versions` key (with a list of versions).

Maybe we should take a hybrid approach like:
- Grab the specific version info from the `package` response
- Check to ensure that the expected keys are present (`["name", "packages", "downloads"]`)
- If they're missing for some reason, fall back to making the current version-specific request, with retries: "api/v1/{}/{}.json"`

jtcohen6: > Check to ensure that the expected keys are present (["name", "packages", "downloads"])

Another way of doing this: `RegistryPackageMetadata.from_dict()` here, on the specific version response, instead of in `dbt.deps.registry._fetch_metadata`

core/dbt/clients/registry.py
```diff
@@ -12,55 +17,52 @@
...
-
-def packages(registry_base_url=None):
-    return _get_with_retries("api/v1/packages.json", registry_base_url)
+    # and https://github.com/dbt-labs/dbt-core/issues/4849
+    response = resp.json()
 
+    if not isinstance(response, dict):  # This will also catch Nonetype
+        error_msg = (
+            f"Request error: The response type of {type(response)} is not valid: {resp.text}"
+        )
```
nathaniel-may, line 51: let's put what we expected in this error message

core/dbt/clients/registry.py
```diff
@@ -12,55 +17,52 @@
...
+        raise requests.exceptions.ContentDecodingError(error_msg, response=resp)
 
-def package(name, registry_base_url=None):
-    response = _get_with_retries("api/v1/{}.json".format(name), registry_base_url)
+    expected_keys = ["name", "versions"]
```
nathaniel-may, line 54: since we seem to only use this as a set, maybe we could just create it as a set using `{}` instead of `[]`

core/dbt/clients/registry.py
```diff
@@ -12,55 +17,52 @@
...
+    expected_version_keys = ["name", "packages", "downloads"]
+    if not set(expected_keys).issubset(response) and not set(expected_version_keys).issubset(
+        response["versions"]
+    ):
+        error_msg = f"Request error: Expected the response to contain keys {expected_keys} but one or more are missing: {resp.text}"
```
nathaniel-may, line 59: [nit] we can use set diff to discover and explicitly report these keys. There's only 3 keys right now so if we don't want to bother I'm fine with that too.

core/dbt/clients/registry.py
```diff
@@ -12,55 +17,52 @@
...
+    get_fn = functools.partial(_get_cached, package_name, registry_base_url)
     return connection_exception_retry(get_fn, 5)
 
 
-def _get(path, registry_base_url=None):
-    url = _get_url(path, registry_base_url)
+def _get(package_name, registry_base_url=None):
+    url = _get_url(package_name, registry_base_url)
     fire_event(RegistryProgressMakingGETRequest(url=url))
     resp = requests.get(url, timeout=30)
```
nathaniel-may, line 36: can this line throw at all?

core/dbt/clients/registry.py
```diff
@@ -74,15 +76,58 @@ def package(name, registry_base_url=None):
...
+
+def get_available_versions(package_name):
+    # returns a list of all available versions of a package
+    response = package(package_name)
+    return list(response)
+
+
+def _get_index(registry_base_url=None):
+
+    url = _get_url("index", registry_base_url)
```
nathaniel-may, line 107: I like these functions and how they abstract over retries. Can we add type hints to them?

nathaniel-may (approved):
Makes sense to me! Added a bunch of optional suggestions as well.

core/dbt/clients/registry.py
emmyoop: Yes.  But we're catching all exceptions from requests in the retry logic.

nathaniel-may: lets add a comment to that effect then.

leahwicz (approved):

jtcohen6 (approved):
Nice work! This is a huge improvement

PR #4998, opened 2022-04-05
Creating ADR for versioning and branching strategy

leahwicz:
### Description
Creating an initial ADR to establish how we approach versioning and branching. I wanted to write this as a base to build the release process off of and make sure we have a shared understanding of how we are doing this and why.

### Checklist


docs/arch/adr-003-versioning-branching-strategy.md
```diff
@@ -0,0 +1,66 @@
+# Release Versioning / Branching Strategy
+
+## Context
+With `dbt` ever evolving and progressing, we continue to ship and release versions on a regular basis. As we release `dbt`, versioning and branching become key to identifying what changes users and conveying the expectations of those changes. We currently push all our changes to `main` which is the latest version of our code. We don't always want to release the latest version of our code though. We want to be able to isolate changes to go into particular releases based on urgency (i.e. bug fixes) and stability (i.e. new features). 
+
+We already have an established versioning strategy when it comes to major/minor/patch releases. The following is our approach as to what to expect from each release version:
+
+* Major: This signifies that breaking changes to users that they must react to are included in this release.
```
stu-k, line 8:
```suggestion
* Major: This signifies that breaking changes which require user action are included in this release.
```

docs/arch/adr-003-versioning-branching-strategy.md
```diff
@@ -0,0 +1,66 @@
...
+* Minor version: This signifies that new features are included in this release.
+* Patch version: This signifies that bug and security fixes are included in this release.
+
+We need a versioning strategy that allows us to to convey to users the confidence and stability of code they are installing. We want a versioning strategy that allows users to try out new changes and experiment with them to get early feedback. We want to make sure users can also get a preview of new features we are working on so that they can incorporate those changes as soon as possible to their own work.
```
stu-k, line 12:
```suggestion
We need a versioning strategy that allows us to to convey to users the confidence and stability of code they are installing. We want a versioning strategy that allows users to try out new changes, and experiment with them to get early feedback. We want to make sure users can also get a preview of new features we are working on, so they can incorporate these changes into their work as soon as possible. 
```

docs/arch/adr-003-versioning-branching-strategy.md
```diff
@@ -0,0 +1,66 @@
...
+
+* Major: This signifies that breaking changes to users that they must react to are included in this release.
+* Minor version: This signifies that new features are included in this release.
+* Patch version: This signifies that bug and security fixes are included in this release.
+
+We need a versioning strategy that allows us to to convey to users the confidence and stability of code they are installing. We want a versioning strategy that allows users to try out new changes and experiment with them to get early feedback. We want to make sure users can also get a preview of new features we are working on so that they can incorporate those changes as soon as possible to their own work. 
+
+We need a branching strategy to support the goals of the versioning strategy. We do not want to have to stop development to `main` as we await a release. We also don't want to have everything in `main` to go into every release. Because of this we can leverage branches to ensure what changes are going into what release and be deliberate about those changes.
```
stu-k, line 14:
```suggestion
We need a branching strategy to support the goals of the versioning strategy. We do not want to have to stop development on `main` as we await a release, nor do we want every change in `main` published in every release. Using branches, we can ensure which changes go into which release.
```

docs/arch/adr-003-versioning-branching-strategy.md
```diff
@@ -0,0 +1,66 @@
...
+
+We need a branching strategy to support the goals of the versioning strategy. We do not want to have to stop development to `main` as we await a release. We also don't want to have everything in `main` to go into every release. Because of this we can leverage branches to ensure what changes are going into what release and be deliberate about those changes. 
+
+## Requirements
+The following are requirements that must be met for releasing dbt:
+
+1. We must have major, minor, and patch releases.
+1. We must have release types that have the goal of garnering early feedback from the Community.
+1. We must have release types that aim to identify bugs and test for stability.
+1. We must have release types that are stable and production quality.
```
stu-k, line 22:
```suggestion
1. We must have release types that are stable and of production quality.
```

docs/arch/adr-003-versioning-branching-strategy.md
```diff
@@ -0,0 +1,66 @@
...
+1. We must have the ability to isolate changes from different versions based on the type of release.
+
+## Decisions
+
+### Version Types
+<details>
+<summary> Header Explanations: </summary>
+
+- Type: type of version that is occurring (i.e. a beta vs final)
```
stu-k, line 31:
```suggestion
- Type: the type of the version (e.g. beta, final)
```

docs/arch/adr-003-versioning-branching-strategy.md
```diff
@@ -0,0 +1,66 @@
...
+- Released?: does this type of version need to be released?
+- Branches: the branches where these version types are present (i.e. we will only find beta version numbers on the `main` branch)
```
stu-k, line 33:
```suggestion
- Branches: the branches where these version types are present (e.g. beta version numbers should only exist on the `main` branch)
```

docs/arch/adr-003-versioning-branching-strategy.md
```diff
@@ -0,0 +1,66 @@
...
+| RC | Yes | Release branch | Major / Minor / Patch | Pre-production |
+| Final | Yes | Release branch | Major / Minor / Patch | Production |
+
+### Branching Strategy
+Based off of the release version and expectations, the version type's branching strategy can be determined. 
+ * Alpha and Beta are experimental and therefore exist on `main` where all our changes reside. 
+ * Alpha is never released but instead signifies that development is happening but the changes during that period have never been released in any form.
+ * RC is moving to a stable state. Therefore, a release branch will be created for RC releases to start limiting the changes happening for that release.
+ * Final is a stable, tested version. Only verified changes will go into this release which is inherited from the RC so a release branch is also required.
+ * A release branch will exist for each unique major, minor version release and be named accordingly: `<major>.<minor>.latest` (ex `1.0.latest`)
```
stu-k, line 51:
```suggestion
 * Alpha and Beta versions are experimental, and therefore exist only on the `main` branch, where all our changes reside
 * Alpha versions are never released. They signify that development is underway, though they changes they contain have not been released in any form
 * An RC version denotes a more stable state. A release branch will be created for RC releases to limit the changes contained in a release
 * Final versions are stable and tested. Only verified changes will go into these releases which are inherited from the corresponding RC version
 * A release branch will exist for each unique major or minor version, to be named accordingly: `<major>.<minor>.latest` (e.g. `1.0.latest`)
```

docs/arch/adr-003-versioning-branching-strategy.md
```diff
@@ -0,0 +1,66 @@
...
+
+ ![Branching Strategy](images/ReleasingBranchStrategy.png)
+
+## Status
+In Progress
+
+## Consequences
+This doesn't drastically change our current versioning and branching strategy but instead documents what we have been doing.
+
+The only major change is we are now incorporating an Alpha version which we previously didn't use at all. The consequences of moving to use an Alpha are to make developing across multiple repos easier. We want `dbt-core` and adapters depend on minor versions. If we don't have a way to keep the `main` branches in sync with one another, then integration tests will start failing. It is also confusing when the `main` branch shows it's an RC or Final version when we do not release the version from there. This gives more clarity as to what versions live where.
```
stu-k, line 61:
```suggestion
This doesn't drastically change our current versioning or branching strategies, but documents our existing flows.

The only real change here is the addition of an Alpha release, the intent of which is to make developing across multiple repos easier. `dbt-core` and adapters should depend on minor versions. If we don't have a way to keep the `main` branches in sync with one another, integration tests will start to fail. It is also confusing when the `main` branch is marked as an RC or Final version when we never release from `main`. This change provides visibility around which versions live where.
```

docs/arch/adr-003-versioning-branching-strategy.md
```diff
@@ -0,0 +1,66 @@
...
+
+## Outside Scope
+The following are topics that are outside the scope of this document and will be addressed in their own ADR:
+* Hotfixes and how branching works for them
+* How and why we bump release versions right before a release.
```
stu-k, line 66:
```suggestion
* Hotfixes and their branching strategies
* How and why we bump release versions right before a release.
```

stu-k (approved):
Just pedantic comments about language. The suggestions in the ADR don't seem drastically different from our current process, so lgtm 👍

docs/arch/adr-003-versioning-branching-strategy.md
```diff
@@ -0,0 +1,66 @@
+# Release Versioning / Branching Strategy
+
+## Context
+With `dbt` ever evolving and progressing, we continue to ship and release versions on a regular basis. As we release `dbt`, versioning and branching become key to identifying what changes for users and conveying the expectations of those changes. We currently push all our changes to `main` which is the latest version of our code. We don't always want to release the latest version of our code though. We want to be able to isolate changes to go into particular releases based on urgency (e.g. bug fixes) and stability (e.g. new features).
```
nathaniel-may, line 4: > As we release `dbt`, versioning and branching become key to identifying what changes for users and conveying the expectations of those changes.

This sentence uses "changes" as both a noun and a verb in the same sentence and confused me. Maybe use "is different" for the verb?

docs/arch/adr-003-versioning-branching-strategy.md
```diff
@@ -0,0 +1,66 @@
+# Release Versioning / Branching Strategy
+
+## Context
+With `dbt` ever evolving and progressing, we continue to ship and release versions on a regular basis. As we release `dbt`, versioning and branching become key to identifying what changes for users and conveying the expectations of those changes. We currently push all our changes to `main` which is the latest version of our code. We don't always want to release the latest version of our code though. We want to be able to isolate changes to go into particular releases based on urgency (e.g. bug fixes) and stability (e.g. new features). 
+
+We already have an established versioning strategy when it comes to major/minor/patch releases. The following is our approach as to what to expect from each release version:
+
+* Major: This signifies that breaking changes which require user action are included in this release.
+* Minor version: This signifies that new features are included in this release.
```
nathaniel-may, line 9: are we making the guarantee that there will be no breaking changes here either?

docs/arch/adr-003-versioning-branching-strategy.md
```diff
@@ -0,0 +1,66 @@
...
+We need a versioning strategy that allows us to to convey to users the confidence and stability of code they are installing. We want a versioning strategy that allows users to try out new changes, and experiment with them to get early feedback. We want to make sure users can also get a preview of new features we are working on, so they can incorporate these changes into their work as soon as possible. 
+
+We need a branching strategy to support the goals of the versioning strategy. We do not want to have to stop development on `main` as we await a release, nor do we want every change in `main` published in every release. Using branches, we can decide which changes go into which release.
+
+## Requirements
+The following are requirements that must be met for releasing dbt:
+
+1. We must have major, minor, and patch releases.
+1. We must have release types that have the goal of garnering early feedback from the Community.
+1. We must have release types that aim to identify bugs and test for stability.
```
nathaniel-may, line 21: What is a release type

docs/arch/adr-003-versioning-branching-strategy.md
```diff
@@ -0,0 +1,66 @@
...
+
+### Branching Strategy
+Based off of the release version and expectations, the version type's branching strategy can be determined. 
+ * Alpha and Beta versions are experimental, and therefore exist only on the `main` branch, where all our changes reside
+ * Alpha versions are never released. They signify that development is underway, though the changes they contain have not been released in any form
+ * An RC version denotes a more stable state. A release branch will be created for RC releases to limit the changes a release contains
+ * Final versions are stable and tested. Only verified changes will go into these releases, which are inherited from the corresponding RC version
+ * A release branch will exist for each unique major or minor version, to be named accordingly: `<major>.<minor>.latest` (e.g. `1.0.latest`)
+
+ ![Branching Strategy](images/ReleasingBranchStrategy.png)
```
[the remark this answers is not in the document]
nathaniel-may: to this point, maybe this doc can explicitly describe what the plan is for a commit we do _not_ want to go out in the next release is?

nathaniel-may (changes requested):
This does a good job documenting the way we expect everything to go, but I think it leaves out some edge cases:
- developing a feature before we release 1.1 that's slated for 1.2
- when developing larger features how we decide between waiting in a feature branch vs merging behind a feature flag
- backporting to diverging sections of the codebase
- etc

These are from the perspective of "which branch does this commit belong in" rather than "how do branches map to our releases." I think both are important to describe our strategy more completely.

emmyoop (commented):
I agree with most of what's in this.  But I agree with @nathaniel-may that more clarity around the branching strategy for those edge cases will be useful to document here.

docs/arch/adr-003-versioning-branching-strategy.md
```diff
@@ -0,0 +1,66 @@
...
+* Major: This signifies that breaking changes which require user action are included in this release.
+* Minor version: This signifies that new features are included in this release.
+* Patch version: This signifies that bug and security fixes are included in this release.
+
+We need a versioning strategy that allows us to to convey to users the confidence and stability of code they are installing. We want a versioning strategy that allows users to try out new changes, and experiment with them to get early feedback. We want to make sure users can also get a preview of new features we are working on, so they can incorporate these changes into their work as soon as possible. 
+
+We need a branching strategy to support the goals of the versioning strategy. We do not want to have to stop development on `main` as we await a release, nor do we want every change in `main` published in every release. Using branches, we can decide which changes go into which release.
+
+## Requirements
+The following are requirements that must be met for releasing dbt:
```
iknox-fa, line 17: I think we need to make further distinctions here for the target of the release (pypi, docker, github, etc) because not all of them fit into this list (docker images specifically)

iknox-fa: Edit:  Is target what you meant by `type`?

docs/arch/adr-003-versioning-branching-strategy.md
```diff
@@ -0,0 +1,66 @@
...
+| RC | Yes | Release branch | Major / Minor / Patch | Pre-production |
+| Final | Yes | Release branch | Major / Minor / Patch | Production |
+
+### Branching Strategy
+Based off of the release version and expectations, the version type's branching strategy can be determined. 
+ * Alpha and Beta versions are experimental, and therefore exist only on the `main` branch, where all our changes reside
+ * Alpha versions are never released. They signify that development is underway, though the changes they contain have not been released in any form
+ * An RC version denotes a more stable state. A release branch will be created for RC releases to limit the changes a release contains
+ * Final versions are stable and tested. Only verified changes will go into these releases, which are inherited from the corresponding RC version
+ * A release branch will exist for each unique major or minor version, to be named accordingly: `<major>.<minor>.latest` (e.g. `1.0.latest`)
```
iknox-fa, line 51: I think we should consider creating a release branch for every release down to the patch level.  There have been at least few cases where that would have made life much easier.

nathaniel-may: I have also liked working on projects that have a 1:1 relationship between releases and branches.

emmyoop: @iknox-fa can you explain what we gain by doing it this way?  It seems to me it adds complexity of what branch to merge to.  It also adds complexity to the backport action and our changelog generation.

nathaniel-may: @emmyoop the way I've done it, you cut a branch for the release then never touch it again. No more figuring out what commit a release was released from and rolling back your local git to investigate bugs.

docs/arch/adr-003-versioning-branching-strategy.md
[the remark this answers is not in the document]
leahwicz: Historically, everything in `main` has gone into the release branch. We have never started (not saying this can't happen in the future) working on something that was slated for the next release before the release branch was cut for the upcoming release. In most cases, I would think this would most likely happen for a large feature and in this case we should leverage feature branches which I can add to the doc to make it clear on how we should handle something like this if it would arise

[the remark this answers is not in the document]
nathaniel-may: yeah that works. just something we can have to start with and iterate on.

leahwicz:
> * when developing larger features how we decide between waiting in a feature branch vs merging behind a feature flag

We don't have a way to do feature flags in the sense where we can easily enable/disable them like in Cloud services but we have used flags for things such as the experimental parser where we want people to try things out. I think that is slightly different where we are intentionally adding a feature/code to a release (albeit not defaulted on) and not that we have started developing a feature for a further in the future release. I can document that as well in the doc. Feature flagging should be used to gain traction on a new feature. Feature branch means the feature is not ready in any sort of shape to ship or include in the upcoming release

leahwicz:
> * backporting to diverging sections of the codebase

@nathaniel-may what more would you like to see on this topic? I just want to make sure I'm addressing the right thing

docs/arch/adr-003-versioning-branching-strategy.md
leahwicz: In minor releases? Correct, no breaking changes. CC @jtcohen6 to double confirm 😄 b/c I know we did put a breaking changelog in the other day...

docs/arch/adr-003-versioning-branching-strategy.md
leahwicz: I was struggling for a good word here. I can give examples (beta, rc, etc.). Any thought on a better term? I didn't want to use version since we are using version in the context of major, minor, and patch.

docs/arch/adr-003-versioning-branching-strategy.md
leahwicz: @iknox-fa or @nathaniel-may have you then just created patch branches off of the initial minor version branch? So for example: 
For the beta, I create the `1.1.0` branch off of `main`
Then for `1.1.1` patch, I create the `1.1.1` branch off of the `1.1.0` branch?
What about for RCs? Do they get their own branch as well?

nathaniel-may: @leahwicz exactly like that. I'm used to all releases having a branch including RCs.

nathaniel-may:
> - backporting to diverging sections of the codebase

@leahwicz when you have to write new code in order for a change to be backported, there are going to be multiple ways to do it. I think in our case we generally want to minimize the surface area for additional risk even if the code is less maintainable since patch releases should never introduce _new_ bugs. When would a change introduce enough of that risk it shouldn't be backported?

docs/arch/adr-003-versioning-branching-strategy.md
nathaniel-may: release "stage" maybe?

leahwicz:
> When would a change introduce enough of that risk it shouldn't be backported?

@nathaniel-may I see this being outside the scope of this doc (I can add it to that section) and as a follow on doc to this. I would prefer to not try and answer ALL the questions in this single doc and instead have follow on ones that address specific well defined questions such as this one. We know backporting will be a thing (unless this doc pivots from that approach) so a "How / when to backport" doc can be its own standalone write-up

docs/arch/adr-003-versioning-branching-strategy.md
leahwicz: Ok updating to that. Thanks for the idea!

docs/arch/adr-003-versioning-branching-strategy.md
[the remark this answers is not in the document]
leahwicz: I added a section about this so let me know if that helps clarify things

docs/arch/adr-003-versioning-branching-strategy.md
jtcohen6: No breaking changes for project code / end users of dbt!!!

We reserve the right to make breaking changes ONLY for:
- the adapter interface (Python methods), relevant to maintainers of adapter plugins
- metadata interfaces — new versions of JSON artifact versions, new version of structured logging

We should still aim to avoid those changes whenever possible. When they do happen, we should communicate them proactively and clearly, and call out _exactly_ who they're for. Probably adapter plugin maintainers. E.g.
https://github.com/dbt-labs/dbt-core/blob/9c233f27ab280f36cc5b3007d0b85a0f9d3ad13a/.changes/1.1.0/Breaking%20Changes-20220316-031311.yaml#L3
https://github.com/dbt-labs/dbt-core/blob/9c233f27ab280f36cc5b3007d0b85a0f9d3ad13a/.changes/1.1.0/Breaking%20Changes-20220412-152450.yaml#L2

docs/arch/adr-003-versioning-branching-strategy.md
```diff
@@ -52,6 +56,18 @@ Based off of the release version and expectations, the version type's branching
 
  ![Branching Strategy](images/ReleasingBranchStrategy.png)
 
+### Changes for Specific Release Versions
+There will be times that we will want to target changes to go into specific release versions. Below outlines 2 different scenarios in which this can occur and how we will approach them.
+
+1. If a release branch already exists: This is the most common scenario where we want to add a change to an upcoming release and the release branch has already been cut. In this instance, developers will commit their changes to `main` first. Those changes (or some form of them) will then be applied to the release branch in question so that they can be included in the specific release version. 
+\* Disclaimer: the how and when to apply changes to a release branch will be covered in another upcoming ADR.
+
+1. If a release branch is not yet cut: This is the scenario where we start to develop a feature for a future version before we have cut a release branch for the upcoming version (eg. we want to work on a feature for `1.2.0` but we haven't cut the release branch for `1.1.0` yet).
```
nathaniel-may, line 65: all the number bullets are "1"

docs/arch/adr-003-versioning-branching-strategy.md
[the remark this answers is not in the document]
nathaniel-may: it does! thanks.

docs/arch/adr-003-versioning-branching-strategy.md
leahwicz: @iknox-fa @nathaniel-may we tag the commit that is released. Are tags not a good way to map what we released to what version? Relying solely on a branch for having only the commits we released actually scare me since in theory you can add more commits to them after a release (not that it would be intentional) but a tag is a definitive end commit.

nathaniel-may: tags do work for this mapping (I wrote more about it in our slack thread). But the same worry exists with our current strategy because we could edit the history of the branch. It might be a little harder to do by accident, but we already live with that possibility. In both situations we would have branch protections to limit the surface area of that risk.

docs/arch/adr-003-versioning-branching-strategy.md
dbeatty10: [PEP 440](https://peps.python.org/pep-0440/#pre-releases) uses the term "phase" to describe options like beta, rc, etc within the pre-release "segment".

nathaniel-may: I also like "phase."

docs/arch/adr-003-versioning-branching-strategy.md
leahwicz: In markdown, you just use `1.` everywhere and when markdown renders, it uses the correct numbers. If you look at this in the rich diff view for the PR, it shows correctly for me. Let me know if that isn't the case for you

docs/arch/adr-003-versioning-branching-strategy.md
leahwicz: @iknox-fa I'm changing type to phase. I used the wrong term.

Also can you elaborate on what you would like to see in requirements for the release targets? Do you mean that Docker always releases from the `main` branch? I was thinking of creating a new doc specifically for the release targets but do you think that will impact the statements made in the requirements here?

dbeatty10 (approved):
Thank you for your leadership and thoughtfulness on this @leahwicz !

This looks good to me, and as-of last week we intend to align with this ADR for dbt-utils doing forward.

Will definitely be following the development of the ADRs for the topics that are outside scope:
* Hotfixes and their branching strategies
* How and why we bump release versions right before a release.
* How and when do we backport changes to release branches

docs/arch/adr-003-versioning-branching-strategy.md
```diff
@@ -0,0 +1,108 @@
...
+The following are requirements that must be met for releasing dbt:
+
+1. We must have major, minor, and patch releases.
+1. We must have release phases that have the goal of garnering early feedback from the Community.
+1. We must have release phases that aim to identify bugs and test for stability.
+1. We must have release phases that are stable and of production quality.
+1. We must have the ability to isolate changes from different versions and phases to release.
+
+## Branching Strategy Options
+The 2 branching strategies we are considering using going forward are:
```
nathaniel-may, line 30: I think this should be more like "alternatives we considered" since we're continuing with our current strategy after talking about it.

docs/arch/adr-003-versioning-branching-strategy.md
```diff
@@ -0,0 +1,108 @@
...
+* Confusing for backporting. We will need to keep creating new labels and making sure we remove the old labels because we don't want to accidently backport to the wrong branch.
+
+### Minor release branches (current strategy)
+#### Pros:
+* Adapter CI testing will continue to work with this strategy and makes it less complicated overall. Adapters and `dbt-core` need to be on the same minor version but can be on different patch versions. The current CI workflows verify that the changes slated for a new `1.1` patch are compatible across all repos by checking out the `1.1.latest` branch of each repo. The 1:1 mapping of adapter branches to `dbt-core` branches makes this easy.
+* Easy backporting. There is no need to figure out the exact branch that you need to backport to and accidently backporting to an older, stale branch.
+#### Cons:
+* Branches contain more than 1 release which is confusing. Without a 1:1 mapping between branch and release, you must rely on tags instead to know where in the branch the release was cut.
+* Releasing from the same branch multiple times can lead to confusion as to which version we are releasing and actually stops us from releasing the wrong version.
```
nathaniel-may, line 48: > and actually stops us from releasing the wrong version.

I think I know what you're getting at, but the wording is confusing to me here. Maybe it should be a separate sentence.

docs/arch/adr-003-versioning-branching-strategy.md
```diff
@@ -0,0 +1,108 @@
...
+| Alpha | No | `main` | Major / Minor | Experimental |
+| Beta | Yes | `main` | Major / Minor | Experimental |
+| RC | Yes | Release branch | Major / Minor / Patch | Pre-production |
+| Final | Yes | Release branch | Major / Minor / Patch | Production |
+
+### Branching Strategy
+Based off of the release version and expectations, the version phase's branching strategy can be determined. 
+ * Alpha and Beta versions are experimental, and therefore exist only on the `main` branch, where all our changes reside
+ * Alpha versions are never released. They signify that development is underway, though the changes they contain have not been released in any form
+ * An RC version denotes a more stable state. A release branch will be created for RC releases to limit the changes a release contains
```
nathaniel-may, line 74: we should probably use the term "release candidate" here even though it's implied.

docs/arch/adr-003-versioning-branching-strategy.md
leahwicz: What about "Branching Strategies Considered"? In ADR's I've seen, it's more about listing out the options in the first half of the doc and then the decisions made and why in the second half. I like having the 2 strategies juxtaposed here to see the pros and cons. I do call out the current strategy so we remember going forward that we kept with the existing position but this time we knows the reason why it was chosen

docs/arch/adr-003-versioning-branching-strategy.md
leahwicz: I see what you mean about the wording. I took another go at clarifying that

leahwicz:
> Also all the number bullets are "1." instead of increasing in value.

@nathaniel-may for markdown you can use `1.` for all list entries and then when it's rendered it turns into the correct numbers. It's nice for if you add or subtract from the list so you don't need to renumber everything. In the rendered view, the numbering looks right to me but if it doesn't for you, let me know

docs/arch/adr-003-versioning-branching-strategy.md
```diff
@@ -0,0 +1,108 @@
...
+\* Disclaimer: the how and when to apply changes to a release branch will be covered in another upcoming ADR.
+
+1. If a release branch is not yet cut: This is the scenario where we start to develop a feature for a future version before we have cut a release branch for the upcoming version (eg. we want to work on a feature for `1.2.0` but we haven't cut the release branch for `1.1.0` yet). 
+    1. Is this dead code? If so, then commiting the changes to `main` is acceptable. There must be 0 risk of users hitting this so if unsure, please uses the alternative option.
+    1. Use a feature branch until the release branch is cut. Then when it makes sense, merge the feature branch into `main` and continue development.
+
+   ** Sometimes feature flags are used by teams to also achieve this goal. In our case, feature flags aren't controlled by the team and instead dependent of a user to toggle them on/off. Core uses feature flags instead for offering experimental functionality for users to try and test out. This is different from hiding features from going into certain releases.
+
+## Status
+In Progress
```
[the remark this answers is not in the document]
leahwicz: I always save this for the last commit :-)

docs/arch/adr-003-versioning-branching-strategy.md
leahwicz: I added another requirement bullet point to define this. @jtcohen6 would you agree with my last point of what "released" means to us?

docs/arch/adr-003-versioning-branching-strategy.md
nathaniel-may: the specifics make it much clearer. thanks!

nathaniel-may (approved):

PR #5000, opened 2022-04-06
add DO_NOT_TRACK environment variable support

### LINKED ISSUE #3540: Support console env var DO_NOT_TRACK
### Describe the feature
If the console environment variable `DO_NOT_TRACK` is set (to any value), do not send anonymous usage stats.  The current configuration equivalent is in `~/.dbt/profiles.yml`
```
config:
  send_anonymous_usage_stats: False
```
The website promoting and documenting the standard is: https://consoledonottrack.com/

The current yaml config: send_anonymous_usage_stats serves the same purpose, but is specific to DBT, where `DO_NOT_TRACK` is intended to be universal.

### Who will this benefit?
Users who would like to avoid sending usage statistics by default would no longer need to:
* discover dbt sends usage statistics
* read documentation to find how to opt-out

Yes, I would be more than willing to implement the feature if the issue is accepted.

agoblet:
resolves #3540

### Description

To be compliant with the [Console Do Not Track](https://consoledonottrack.com) initiative, this PR adds support for the `DO_NOT_TRACK` environment variable. `DO_NOT_TRACK` is a generic version of the dbt-specific `SEND_ANONYMOUS_USAGE_STATS` variable. Tracking will be disabled if any of these 2 variables specifies that the user does not wish to be tracked.

### Checklist

### DIFF
--- a/.changes/unreleased/Features-20220406-080328.yaml
+++ b/.changes/unreleased/Features-20220406-080328.yaml
@@ -1,5 +1,5 @@
 kind: Features
-body: add support for the DO_NOT_TRACK environment variable
+body: Support the DO_NOT_TRACK environment variable from the consoledonottrack.com initiative
 time: 2022-04-06T08:03:28.877054+02:00
 custom:
   Author: agoblet
--- a/.changes/unreleased/Features-20220406-080328.yaml
+++ b/.changes/unreleased/Features-20220406-080328.yaml
@@ -1,0 +1,7 @@
+kind: Features
+body: add support for the DO_NOT_TRACK environment variable
+time: 2022-04-06T08:03:28.877054+02:00
+custom:
+  Author: agoblet
+  Issue: "3540"
+  PR: "5000"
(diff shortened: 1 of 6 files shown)


gshank (changes requested):
I don't think that we want two separate flags, SEND_ANONYMOUS_USAGE_STATS and DO_NOT_TRACK. We only want one flag, SEND_ANONYMOUS_USAGE_STATS, but one that can also be set from a DO_NOT_TRACK env var.

agoblet:
> I don't think that we want two separate flags, SEND_ANONYMOUS_USAGE_STATS and DO_NOT_TRACK. We only want one flag, SEND_ANONYMOUS_USAGE_STATS, but one that can also be set from a DO_NOT_TRACK env var.

Thanks for the review @gshank. Having 2 separate flags for the same feature adds complexity to the feature itself. I changed the code to your proposed solution. the `DO_NOT_TRACK` will override `SEND_ANONYMOUS_USAGE_STATS ` when set. This makes the flags module a bit more complex, since we did not have to deal with overriding flags before. It keeps the tracking module clean though.

PR #5001, opened 2022-04-06
Add `selected_resources` to the Jinja context

### LINKED ISSUE #3471: Jinja context variable for selected resources
### Describe the feature

A Jinja context variable that includes the list of resources (by `unique_id`) selected in the current invocation:
```
{% if 'model.package_name.my_model' in selected_resources %}
  ...
{% endif %}
```

This would have similar caveats to `graph` and `execute`, because it wouldn't be populated at parse time, only at compile/execute/runtime.

Bonus points if it includes DAG information, i.e. the order in which resources will be queued for execution.

Use cases:
- An `on-run-start` hook that wants to know all resources about to execute. This is available in the `on-run-end` context, in the form of the [`Results`](https://docs.getdbt.com/reference/dbt-jinja-functions/on-run-end-context#results) object.
- Use standard node selection syntax as the input to a macro. E.g. run [`audit_helper.compare_relations()`](https://github.com/fishtown-analytics/dbt-audit-helper/tree/0.3.0/#compare_relations-source) for each resource selected/built by a CI job by looping over `selected_resources` in a `run-operation`. (Could we do a half-decent version of this with `Results` today in an `on-run-end` hook?)

- Adding a property to manifest nodes: `selected: true|false`. Then users would access the same `graph` context variable and use that property. I understand that some users already do something similar, e.g. inspect the `tags` on each node, to emulate selection criteria. We could also consider `selection_criteria`, detailing _why_ a given node has been selected. This could be included in the `list` output and a helpful tool for visualizing selection groupings.

- Make the selection criteria directly available, via `flags.args.models` and `flags.args.exclude`: https://github.com/fishtown-analytics/dbt/issues/2253#issuecomment-606637983

### Additional context

Further down the road:
- The `run-operation` task should accept `--select` and `--selector`
- Custom tasks (https://github.com/fishtown-analytics/dbt/issues/2381) that can run nodes in parallel using threads, rather than in sequence

### Who will this benefit?

This comes up in a surprising number of more-complex use cases.

b-per:
resolves #3471

### Description

I don't know if I followed the correct approach but this code is adding a variable called `selected_resources` to the Jinja context. This is a list of all models/tests/snapshots etc... captured by the selectors for the dbt run. I checked how `flags` were added to the context and did the same for `selected_resources`, hence why it is under the `core/dbt` folder.

The issue #3471 also mentions making this available for `run-operation` but this will be taken care of in another issue.

My testing approach is to use a `on-run-start` hook to compare the value of `selected_resources` with the expected one.

If this is merged the docs will need to be updated to mention the new variable and highlight that similar to `graph` is will only contain some information at `execute` time.

### Checklist

### DIFF
--- a/.changes/unreleased/Features-20220404-190439.yaml
+++ b/.changes/unreleased/Features-20220404-190439.yaml
@@ -1,0 +1,9 @@
+kind: Features
+body: Add a variable called selected_resources in the Jinja context containing a list
+  of all the resources matching the nodes for  the --select, --exclude and/or --selector
+  parameters.
+time: 2022-04-04T19:04:39.347479+02:00
+custom:
+  Author: b-per
+  Issue: "3471"
+  PR: "5001"
--- a/core/dbt/context/base.py
+++ b/core/dbt/context/base.py
@@ -4,6 +4,7 @@ from typing import Any, Dict, NoReturn, Optional, Mapping
 
 from dbt import flags
 from dbt import tracking
+from dbt import selected_resources
 from dbt.clients.jinja import get_rendered
 from dbt.clients.yaml_helper import yaml, safe_load, SafeLoader, Loader, Dumper  # noqa: F401
 from dbt.contracts.graph.compiled import CompiledResource
@@ -555,6 +556,15 @@ class BaseContext(metaclass=ContextMeta):
         TODO: Replace with object that provides read-only access to flag values
         """
         return flags
+
+    @contextproperty
+    def selected_resources(self) -> Any:
+        """The `selected_resources` variable contains a list of the resources
+        selected based on the parameters provided to the dbt command.
+        Currently, is not populated for the command `run-operation` that
+        doesn't support `--select`.
+        """
+        return selected_resources.SELECTED_RESOURCES
 
     @contextmember
     @staticmethod
--- a/core/dbt/context/base.py
+++ b/core/dbt/context/base.py
@@ -4,7 +4,6 @@ from typing import Any, Dict, NoReturn, Optional, Mapping
 
 from dbt import flags
 from dbt import tracking
-from dbt import selected_resources
 from dbt.clients.jinja import get_rendered
 from dbt.clients.yaml_helper import yaml, safe_load, SafeLoader, Loader, Dumper  # noqa: F401
 from dbt.contracts.graph.compiled import CompiledResource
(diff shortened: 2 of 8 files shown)


core/dbt/context/base.py
```diff
@@ -556,6 +557,15 @@ def flags(self) -> Any:
         """
         return flags
 
+    @contextproperty
+    def selected_resources(self) -> Any:
```
gshank, line 561: I'm pretty sure that the base context is the wrong context for this functionality. We don't need this in yaml rendering and other non-sql rendering. It probably ought to go in a provider context. Is there some reason that you put it in base?

gshank (changes requested):

core/dbt/context/base.py
b-per: I have just moved it to the ProviderContext

gshank (changes requested):
Looks like you need to update the tests. In test_context.py you need to add 'selected_resources' to REQUIRED_MACRO_KEYS.

tests/functional/selected_resources/fixtures.py
```diff
@@ -0,0 +1,35 @@
+on_run_start_macro_assert_selected_models_expected_list = """
+{% macro assert_selected_models_expected_list(expected_list) %}
```
ChenyuLInx, line 2: Really like the way you use this macro to test the functionality!

tests/functional/selected_resources/test_selected_resources.py
```diff
@@ -0,0 +1,81 @@
...
+        "on-run-start": "{{ assert_selected_models_expected_list(var('expected_list',[])) }}",
+    }
+
+
+def test_selected_resources_build(project):
+    results = run_dbt(
+        [
+            "build",
+            "--select",
+            "model1+",
```
ChenyuLInx, line 42: Should we do a `model2+` as one of the test cases? since the result of this selection is the same as the no select specified test below.

ChenyuLInx (commented):
LGTM with one nitpicking item for tests. I feel like this is ready to go once other tests are fixed.

tests/functional/selected_resources/test_selected_resources.py
b-per: I just modified a test to be `model2+`

It was failing at first (because `model1` was not built) so I moved all the tests to a Class that uses the fixture `build_all`, meaning that a full `dbt build` is run before each test (which allows us to select nodes that are depending on other ones)

.changes/unreleased/Features-20220404-190439.yaml
```diff
@@ -0,0 +1,9 @@
+kind: Features
+body: Add a variable called selected_resources in the Jinja context containing a list
```
ChenyuLInx, line 2: @gshank could there be situations where `selected_resources` is referenced before it got set by `get_graph_queue `? I think there isn't but want to double-check.

ChenyuLInx (approved):
LGTM! Left one question but not a blocker

.changes/unreleased/Features-20220404-190439.yaml
gshank: I think the selected_resources will be empty unless it's execution time, which should be okay.  I suppose the list could be really big for a run command that executes the whole project, but I think that should be okay too; it doesn't do much except reference the list of nodes.

Issue #3576, opened 2021-07-15
Test modifiers do not store unrendered configs


jtcohen6:
### Describe the bug
Similar to #2744, but for tests. We're not properly storing the un-rendered version of test configs supplied as modifiers, so any env-aware logic in them will trigger `state:modified`. We actually _do_ have access to `node.unrendered_config`, we just need to pass the original values before they're rendered here:
https://github.com/dbt-labs/dbt/blob/f460d275baa4df8cf8713f27e198a3577d6ea71c/core/dbt/parser/schema_test_builders.py#L228-L233

This is true for both generic-generic tests, and built-in shortcut generic tests (`unique` + `not_null`).

### Steps To Reproduce

```yml
version: 2
models:
  - name: my_model
    columns: 
      - name: id
        tests:
          - not_null:
              limit: "{{ ('50' if target.name == 'dev' else '500') | as_number }}"
          - accepted_values:
              values: [1]
              limit: "{{ ('50' if target.name == 'dev' else '500') | as_number }}"
```
```bash
$ dbt run --target prod
$ dbt test --target prod
$ mv target/manifest.json .state
$ dbt ls -s state:modified --state .state
```
```
my_project.schema_test.accepted_values_my_model_id__1
my_project.schema_test.not_null_my_model_id
```

Issue #2744, opened 2020-09-09
Store the unresolved form of source database representations


jtcohen6:
_extension of #2713_

### Describe the feature

Let's say a project has a source defined like:
```yml
sources:
  - name: my_postgres_db
    database: "{{ 'raw' if target.name == 'prod' else 'raw_sampled' }}"
```

Today, if running in `dev` and comparing to a manifest generated by a `prod` target, `state:modified+` will include _all_ models that depend on `source:my_postgres_db` because the `database` differs. Ideally, we'd be able to tell that the unresolved `database` (Jinja statement) is identical.

I'm most interested in **sources** and **database representations** (`database`, `schema`, `identifier`). If it made sense, though, we could broaden this issue to the matter of storing unresolved:
* database representations
* resource properties (will the distinction between node configs and resource properties even make sense in a post-#2401 world?)

### Describe alternatives you've considered
- The proposal for `state:modified` subselectors gives decent user-side answers by letting people "switch off" the intersection of `state:modified.database_representations` and specific sources.

### Who will this benefit?
- Users who develop, test, and deploy against genuinely different source datasets. While this isn't something we recommend in the general case, it's the norm at organizations of a certain size + maturity.

fabrice-etanchaud:
You are totally right, @jtcohen6 . That 's exactly our use case, here @maif-vie (life insurance). We have a kind of environment notion (we call that a 'track') that groups data sources for a given project/feature. I ended up adding a extra credential parameter called 'environment' to use as a prefix for source's databases. That way I can configure separately the source version and the target database. This enhancement would be of great help here.
Best regards from french west coast ;-)

jtcohen6:
- This will be improved by the introduction of sub-selectors (#2704)
- Let's revisit after #2401, since that may have implications for how source `database`/`schema`/`identifier` are stored on the source node (`node.config` vs. node-level properties)

katieclaiborne:
Would it be possible to revive this issue?

Our (BigQuery) project defines sources that are [implicitly different across environments](https://github.com/dbt-labs/dbt-core/issues/2744#issuecomment-811046508).

We're using `state:modified` sub-selectors as a workaround, but because this is the general pattern for our sources, creating a YAML selector with the intersection of each one is unsustainable.

wjhrdy:
> Would it be possible to revive this issue?
> 
[...]

PR #5003, opened 2022-04-06
[Feature] CT-201 source config WIP

### LINKED ISSUE #3662: [CT-201] Reconcile configs + properties for sources

emmyoop:
resolves #3662

### Description

Draft PR for feature branch.

Some context over in #4960 

### Checklist

### DIFF
--- a/core/dbt/contracts/graph/model_config.py
+++ b/core/dbt/contracts/graph/model_config.py
@@ -335,6 +335,16 @@ class BaseConfig(AdditionalPropertiesAllowed, Replaceable):
 @dataclass
 class SourceConfig(BaseConfig):
     enabled: bool = True
+    quoting: Optional[Dict[str, Any]] = None
+    freshness: Optional[Dict[str, Any]] = None
+    loader: Optional[str] = None
+    loaded_at_field: Optional[
+        Any
+    ] = None  # TODO what type is this? docs say: "<column_name_or_expression>"
+    database: Optional[str] = None
+    schema: Optional[str] = None
+    meta: Optional[Dict[str, Any]] = None
+    tags: Optional[List[str]] = None
 
 
 @dataclass
--- a/core/dbt/contracts/graph/model_config.py
+++ b/core/dbt/contracts/graph/model_config.py
@@ -338,9 +338,8 @@ class SourceConfig(BaseConfig):
     quoting: Optional[Dict[str, Any]] = None
     freshness: Optional[Dict[str, Any]] = None
     loader: Optional[str] = None
-    loaded_at_field: Optional[
-        Any
-    ] = None  # TODO what type is this? docs say: "<column_name_or_expression>"
+    # TODO what type is this? docs say: "<column_name_or_expression>"
+    loaded_at_field: Optional[Any] = None
     database: Optional[str] = None
     schema: Optional[str] = None
     meta: Optional[Dict[str, Any]] = None
--- a/core/dbt/contracts/graph/model_config.py
+++ b/core/dbt/contracts/graph/model_config.py
@@ -335,15 +335,39 @@ class BaseConfig(AdditionalPropertiesAllowed, Replaceable):
 @dataclass
 class SourceConfig(BaseConfig):
     enabled: bool = True
-    quoting: Optional[Dict[str, Any]] = None
-    freshness: Optional[Dict[str, Any]] = None
-    loader: Optional[str] = None
+    quoting: Dict[str, Any] = field(
+        default_factory=dict,
+        metadata=MergeBehavior.Update.meta(),
+    )
+    freshness: Optional[Dict[str, Any]] = field(
+        default=None,
+        metadata=CompareBehavior.Exclude.meta(),
+    )
+    loader: Optional[str] = field(
+        default=None,
+        metadata=CompareBehavior.Exclude.meta(),
+    )
     # TODO what type is this? docs say: "<column_name_or_expression>"
-    loaded_at_field: Optional[Any] = None
-    database: Optional[str] = None
-    schema: Optional[str] = None
-    meta: Optional[Dict[str, Any]] = None
-    tags: Optional[List[str]] = None
+    loaded_at_field: Optional[str] = field(
+        default=None,
+        metadata=CompareBehavior.Exclude.meta(),
+    )
+    database: Optional[str] = field(
+        default=None,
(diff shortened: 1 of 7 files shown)


core/dbt/parser/sources.py
```diff
@@ -270,10 +268,11 @@ def _generate_source_config(self, fqn: List[str], rendered: bool, project_name:
 
         return generator.calculate_node_config(
             config_call_dict={},
-            fqn=fqn,
+            fqn=target.fqn,
             resource_type=NodeType.Source,
-            project_name=project_name,
+            project_name=target.package_name,
             base=False,
+            patch_config_dict=target.source.config,
```
jtcohen6, line 275: We need access to both:
- `target.source.config` = config defined at the source level
- `target.table.config` = config defined at the source table level

I think there are two options here:
- Merge those configs together (preferring table-level over source-level), before passing into `patch_config_dict`
- Modify `calculate_node_config` to accept multiple "levels" of patches, so that all calculation happens in that same method: source table > source > project-level config

core/dbt/parser/manifest.py
```diff
@@ -465,6 +465,8 @@ def parse_project(
                     else:
                         dct = block.file.dict_from_yaml
                     parser.parse_file(block, dct=dct)
+                    # Came out of here with UnpatchedSourceDefinition containing configs at the source level
+                    # and not configs at the table level (as expected)
```
jtcohen6, line 469: See comment re: `patch_config_dict` passed into `calculate_node_config`

core/dbt/contracts/graph/model_config.py
```diff
@@ -335,6 +335,39 @@ def replace(self, **kwargs):
...
+    )
+    freshness: Optional[Dict[str, Any]] = field(
+        default=None,
+        metadata=CompareBehavior.Exclude.meta(),
+    )
+    loader: Optional[str] = field(
+        default=None,
+        metadata=CompareBehavior.Exclude.meta(),
+    )
+    # TODO what type is this? docs say: "<column_name_or_expression>"
```
jtcohen6, line 350: String is right.

Food for thought: These dataclass attributes are duplicative with the ones already defined:
- in [UnparsedSourceTableDefinition](https://github.com/dbt-labs/dbt-core/blob/899b0ef2241d97d97df5270c20b67be02ec7ca61/core/dbt/contracts/graph/unparsed.py#L243-L256)
- in [ParsedSourceDefinition](https://github.com/dbt-labs/dbt-core/blob/899b0ef2241d97d97df5270c20b67be02ec7ca61/core/dbt/contracts/graph/parsed.py#L627-L641)

Given our aim here is backwards compatibility, that duplication may be unavoidable

core/dbt/contracts/graph/model_config.py
```diff
@@ -335,6 +335,39 @@ def replace(self, **kwargs):
 @dataclass
 class SourceConfig(BaseConfig):
     enabled: bool = True
+    quoting: Dict[str, Any] = field(
```
jtcohen6, line 338: Should we move the addition of these new configs into a separate PR? If the goal is just to support setting `enabled` as a config on the source/table levels

core/dbt/parser/sources.py
```diff
@@ -137,15 +137,13 @@ def parse_source(self, target: UnpatchedSourceDefinition) -> ParsedSourceDefinit
...
         config = self._generate_source_config(
-            fqn=target.fqn,
+            target=target,
             rendered=True,
-            project_name=target.package_name,
         )
 
         unrendered_config = self._generate_source_config(
-            fqn=target.fqn,
+            target=target,
```
jtcohen6, line 145: Background context on `rendered: bool`, `UnrenderedConfigGenerator`, and `unrendered_config`, since these are pretty confusing!

This exists to limit "false positives" during state comparison, i.e. to power `state:modified` selection in Slim CI runs ("just build what's changed"): https://docs.getdbt.com/reference/node-selection/state-comparison-caveats#false-positives

The comparison happens in the `same_config` method here:
https://github.com/dbt-labs/dbt-core/blob/899b0ef2241d97d97df5270c20b67be02ec7ca61/core/dbt/contracts/graph/parsed.py#L317-L321

Idea being, it's not uncommon for users to have environment-based configuration like:
```yml
sources:
  - name: my_source
    config:
      database: "{{ 'prod_raw' if target.name == 'prod' else 'sampled_dev' }}"
      schema: "{{ env_var('DBT_SOURCE_SCHEMA') }}"
      # etc
```
The rendered versions of those configs will be different based on the values of env vars or `target` values. The thing we _actually_ want to detect changes in is the raw un-rendered Jinja expression. If that's changed, it's because someone edited the thing.

This works for configs set in `dbt_project.yml`, which are stored as unrendered Jinja expressions, and passed twice into `calculate_node_config` (once with rendering, once without).

However, we've _already_ rendered all fields in (non-project) yaml files, during initial parsing:
https://github.com/dbt-labs/dbt-core/blob/899b0ef2241d97d97df5270c20b67be02ec7ca61/core/dbt/parser/schemas.py#L633-L635

So we're passing already-rendered Jinja expressions into `calculate_node_config`, and we never get access to their unrendered forms. We'd need a subtle reordering of parsing/rendering to make it work as intended.

Related: https://github.com/dbt-labs/dbt-core/issues/2744, https://github.com/dbt-labs/dbt-core/issues/3576

jtcohen6 (commented):
Getting close! I left a few comments that may provide helpful (or least historical) context

nathaniel-may:
@jtcohen6 I definitely requested your review on the wrong PR. This one will be for the full feature. The enabled flag work is going on in #5008. Thank you for all the context here!

PR #5008, opened 2022-04-07
Add enabled as a source config

### LINKED ISSUE #3662: [CT-201] Reconcile configs + properties for sources

emmyoop:
resolves #3662

### Description

This only adds enabled as a config.  Tests have been slimmed down to show what is expected to work.  Since you can't set enabled as a property currently, we don't need to deal with overrides.  `enabled` as a config is also highly valuable to the users of dbt.

The rest of the configs will be added in #5003.

### Checklist


jtcohen6 (commented):
Whoops, I left my comments in https://github.com/dbt-labs/dbt-core/pull/5003 instead of here. Some of the context there may still be helpful/interesting to you, specifically the piece about `unrendered_config`.

In any case, I think you're on the right track here! Looks good in local testing. I'd be very happy about shipping this as a standalone piece of net-new functionality in v1.1.

core/dbt/parser/sources.py
```diff
@@ -261,19 +259,27 @@ def parse_source_test(
...
             generator = ContextConfigGenerator(self.root_project)
         else:
             generator = UnrenderedConfigGenerator(self.root_project)
 
+        # configs with precendence set
+        precedence_configs = dict()
+        # first apply source configs
+        precedence_configs.update(target.source.config)
+        # then overrite anything that is defined on source tables
+        precedence_configs.update(target.table.config)
```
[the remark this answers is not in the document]
nathaniel-may: yeah this is the part that I really wasn't sure was a good idea. Is there a `merge_config_dict` function of sorts that will handle some of this?

nathaniel-may:
blocked on #5015

nathaniel-may:
looking into the one remaining failure.

emmyoop (commented):
This looks good to me as a first step to the end goal.  I'm not going to approve it since I wrote some of it too!  I think it would be great if @gshank could take a look.

core/dbt/contracts/graph/unparsed.py
```diff
@@ -242,6 +242,7 @@ class Quoting(dbtClassMixin, Mergeable):
 
 @dataclass
 class UnparsedSourceTableDefinition(HasColumnTests, HasTests):
+    config: Dict[str, Any] = field(default_factory=dict)
```
gshank, line 245: There is a HasConfig mixin class which some of the other Unparsed nodes use. If it's possible to use that (sometimes you run into precedence issues with fields without defaults) it probably should for consistency.

core/dbt/parser/sources.py
```diff
@@ -261,19 +259,29 @@ def parse_source_test(
...
         else:
             generator = UnrenderedConfigGenerator(self.root_project)
 
+        # configs with precendence set
+        precedence_configs = dict()
+        # first apply source configs
+        precedence_configs.update(target.source.config)
+        # then overrite anything that is defined on source tables
+        # this is not quite complex enough for configs that can be set as top-level node keys, but
+        # it works while source configs can only include `enabled`.
```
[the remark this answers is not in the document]
gshank: In the other nodes configs are merged by the code in the BaseConfig object, in the 'update_from' method, which handles merging things like tags and meta, etc. (MergeBehavior, etc). Which is called by 'calculate_node_config'.  I imagine you'll need to use some of that code to handle merging the Source configs with the SourceTable configs?

gshank (approved):
Looks good. I left a few comments. The merging of configs will have to be handled eventually but that can happen in the next phase.

core/dbt/contracts/graph/unparsed.py
emmyoop: We ran into issues with defaults when we tried to use it.

gshank: I'm not surprised. Getting the default and non-default configs lined up is a headache.

nathaniel-may:
Also not approving since I wrote some of this as well. Thanks for the review, @gshank!

PR #5012, opened 2022-04-08
make graph_selection tests just checking selection

ChenyuLInx:
### Description
Fix graph tests since `expect_pass` is enabled for `run_dbt`. This change makes most of the tests in graph selection tests only check whether the selection is correct but not check whether the model run was successful. This makes the tests run faster(since we no longer need to do seed). In the future, if selection logic is moved out we could also just make this into unit tests.

### Checklist

### DIFF
--- a/tests/functional/graph_selection/test_graph_selection.py
+++ b/tests/functional/graph_selection/test_graph_selection.py
@@ -2,7 +2,7 @@ import os
 import json
 import pytest
 
-from dbt.tests.util import run_dbt
+from dbt.tests.util import run_dbt, check_result_nodes_by_name
 from tests.functional.graph_selection.fixtures import SelectionFixtures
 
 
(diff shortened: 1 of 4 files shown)


stu-k (commented):
Nothing stands out, will leave for those more familiar with pytest and graph selection to approve 👍

gshank (approved):
This makes a lot of sense. All of the checking that tables actually exist really belongs to different functionality, not selection. And it was making these tests really slow.

There is a "check_result_nodes_by_name' which might be helpful for some of these tests. It uses "set" because sometimes the results come out in a different order. The order of the results isn't guaranteed.

ChenyuLInx:
> This makes a lot of sense. All of the checking that tables actually exist really belongs to different functionality, not selection. And it was making these tests really slow.
> 
> There is a "check_result_nodes_by_name' which might be helpful for some of these tests. It uses "set" because sometimes the results come out in a different order. The order of the results isn't guaranteed.

Ha, I actually did all of them using set. But I will update it using `check_result_nodes_by_name` to promote implemented function.

tests/functional/graph_selection/test_graph_selection.py
```diff
@@ -32,41 +32,23 @@ def clear_schema(project):
...
-        assert len(results) == 1
-
-        check_relations_equal(project.adapter, ["seed", "users"])
-
-        created_tables = project.get_tables_in_schema()
-        assert "users_rollup" not in created_tables
-        assert "alternative.users" not in created_tables
-        assert "base_users" not in created_tables
-        assert "emails" not in created_tables
+        results = run_dbt(["run", "--select", "users"], expect_pass=False)
```
jtcohen6, line 43: Do we want this to just be `list`, rather than `run` + expect fail?

(feel free to ignore me and ship as is)

ChenyuLInx: That's a great idea! updating

ChenyuLInx: Actually `ls` and `run` doesn't seems to select the exact same items, it would include source, exposure etc. It would be much bigger change including change the actual expected results to make the update. I think we can leave it as an item if we plan to touch more things related to selector

jtcohen6: Oh good call. You can limit this to models (same as `run`) with `dbt ls --resource-type model`. But very happy with leaving it as is

ChenyuLInx: Ops, this poped up after I hit the merge button. This is good to know!

PR #5013, opened 2022-04-08
Ct 355/migrate simple seed

VersusFacit:
Migrates most (soon all) of 005 simple seed.

### Description
~~Using my best judgement of the test framework, I've gotten these tests working locally! There are two main TODOs where I'm sure there is a better way to accomplish things than the ideas I was able to come up with. I'll mark with inline comments. Please do suggest things so I can get them ship-shape.~~

### Updated:
I've figured out how to get all the tests working per the issues stated above.

### Checklist


tests/functional/simple_seed/test_seed.py
```diff
@@ -1,19 +1,290 @@
...
+            "seeds": {
+                "schema": "custom_schema",
+                "quote_columns": False,
+            },
+        }
+
+    def test_postgres_simple_seed_with_schema(self, project):
+        results = run_dbt(["seed"])
+        assert len(results) == 1
+        # TODO: 1/4 custom schema work; but unsure how to provide these lines with custom schema on seed_actual
```
VersusFacit, line 141: So, I did my best and couldn't quite figure out the right way to tell the test framework a custom schema for a table. Suggestions?

tests/functional/simple_seed/test_seed.py
```diff
@@ -1,19 +1,290 @@
...
+            },
+        }
+
+    def test_postgres_simple_seed_with_disabled(self, project):
+        results = run_dbt(["seed"])
+        len(results) == 2
+        check_table_does_exist(project.adapter, "seed_enabled")
+        check_table_does_not_exist(project.adapter, "seed_disabled")
+        check_table_does_exist(project.adapter, "seed_tricky")
+        # TODO: 1/3 is there a better way to enforce cleanup
```
VersusFacit, line 188: So this is required cleanup. I'm guessing there is a better way to reset the database after every test_ when using the `def seeds` fixture in a test class.

tests/functional/simple_seed/test_seed.py
ChenyuLInx: Looks like we can define the schema for seed in dbt_project.yml [here](https://docs.getdbt.com/reference/seed-configs#general-configurations-1). You can overwrite it by overwriting [this fixture](https://github.com/dbt-labs/dbt-core/blob/main/core/dbt/tests/fixtures/project.py#L134) in your class for now and include the seed definitions. But if you feel like that is something needed in other places also, you can also make an update to that `dbt_profile_data` to make it require another fixture returns seed configurations and add it to the dbt_profile.

tests/functional/simple_seed/test_seed.py
ChenyuLInx: You can have your setup fixture do it by doing the setup, yield something and do the drop table after yield. [Examples](https://docs.pytest.org/en/6.2.x/fixture.html#yield-fixtures-recommended)

ChenyuLInx (commented):
This looks great!!!

tests/functional/simple_seed/test_seed.py
```diff
@@ -1,19 +1,290 @@
...
+        assert len(run_result) == 1
+        check_relations_equal(project.adapter, ["seed_actual", "seed_expected"])
+        if exists:
+            check_table_does_exist(project.adapter, "models__downstream_from_seed_actual")
+        else:
+            check_table_does_not_exist(project.adapter, "models__downstream_from_seed_actual")
+
+
+class TestBasicSeedTests(SeedTestBase):
+    def test_postgres_simple_seed(self, project):
```
ChenyuLInx, line 77: We should remove Postgres in the test name

tests/functional/simple_seed/test_seed.py
```diff
@@ -1,19 +1,290 @@
...
+
+    def test_postgres_simple_seed_with_drop_and_schema(self):
+        results = run_dbt(["seed"])
+        assert len(results) == 1
+        # TODO: 3/4
+        # check_relations_equal(project.adapter, ["seed_actual", "seed_expected"])
+
+        # this should drop the seed table, then re-create
+        results = run_dbt(["seed", "--full-refresh"])
+        results = run_dbt(["seed"])
```
ChenyuLInx, line 158: Why do we want to run it twice here?

tests/functional/simple_seed/test_seed.py
```diff
@@ -1,19 +1,290 @@
...
+        # TODO: 4/4
+        # check_relations_equal(project.adapter, ["seed_actual", "seed_expected"])
+
+
+class TestSimpleSeedEnabledViaConfig(object):
```
ChenyuLInx, line 163: @gshank @VersusFacit How do you guys feel about moving this to selection(graph selection?) tests? since this is mainly testing the selection functionality

tests/functional/simple_seed/test_seed.py
```diff
@@ -1,19 +1,290 @@
...
+
+class TestSimpleSeedWithBOM(SeedConfigBase):
+    @pytest.fixture(scope="class", autouse=True)
+    def setUp(self, project):
+        """Create table for ensuring seeds and models used in tests build correctly"""
+        project.run_sql_file(project.test_data_dir / Path("seed_expected.sql"))
+
+    @pytest.fixture(scope="class")
+    def seeds(self):
+        dir_path = os.path.dirname(os.path.realpath(__file__))
```
ChenyuLInx, line 241: I think you can use 'test_data_dir' [fixture](https://github.com/dbt-labs/dbt-core/blob/899b0ef2241d97d97df5270c20b67be02ec7ca61/tests/functional/basic/test_copy_uppercase.py#L62) for this.

tests/functional/simple_seed/test_seed.py
```diff
@@ -1,19 +1,290 @@
...
+class SeedTestBase(SeedConfigBase):
+    @pytest.fixture(scope="class", autouse=True)
+    def setUp(self, project):
+        """Create table for ensuring seeds and models used in tests build correctly"""
+        project.run_sql_file(project.test_data_dir / Path("seed_expected.sql"))
+
+    @pytest.fixture(scope="class")
+    def seeds(self):
+        dir_path = os.path.dirname(os.path.realpath(__file__))
+        seed_actual_csv = read_file(dir_path, "seeds", "seed_actual.csv")
```
gshank, line 46: Use project.test_data_dir, or test_data_dir, and put the seed in a 'data' directory.

tests/functional/simple_seed/test_seed.py
gshank: Yes, please remove all the 'postgres' parts of the test names. Remember these tests will be run in all the other adapters with the same test name that's defined here.

tests/functional/simple_seed/test_seed.py
```diff
@@ -1,19 +1,290 @@
...
+    def test_postgres_simple_seed_full_refresh_flag(self, project):
+        """Drop the seed_actual table and re-create. Verifies correct behavior by the absence of the
+        model which depends on seed_actual."""
+        self._build_relations_for_test(project)
+        self._check_relation_end_state(
+            run_result=run_dbt(["seed", "--full-refresh"]), project=project, exists=False
+        )
+
+
+class TestSeedConfigFullRefreshOn(SeedTestBase):
```
gshank, line 91: Just a reminder that if these tests are moved into the adapter zone, they'll all have to have names that don't start with 'Test' and a simple class wrapper that does start with 'Test'. Doesn't have to be done now though.

tests/functional/simple_seed/test_seed.py
```diff
@@ -1,19 +1,290 @@
...
+    @pytest.fixture(scope="class")
+    def project_config_update(self):
+        return {
+            "config-version": 2,
```
gshank, line 95: You don't need to have "config-version" in the project_config_update. It's in the base config.

tests/functional/simple_seed/test_seed.py
gshank: We do now have project.drop_test_schema(). I haven't tested it for this use case, but it should work.

tests/functional/simple_seed/test_seed.py
```diff
@@ -1,19 +1,290 @@
...
+
+    @pytest.fixture(scope="class")
+    def seeds(self):
+        dir_path = os.path.dirname(os.path.realpath(__file__))
+        seed_bom = read_file(dir_path, "data", "seed_bom.csv")
+        return {"seed_bom.csv": seed_bom}
+
+    def test_postgres_simple_seed(self, project):
+        # first make sure nobody "fixed" the file by accident
+        seed_path = project.test_data_dir / Path("seed_bom.csv")
```
gshank, line 247: Did the 'read_file' utility not work?

gshank (changes requested):
Looks good, but needs a few minor changes.

tests/functional/simple_seed/test_seed.py
VersusFacit: Bah, I asked this question terribly. That's on me. What I meant to ask was how do I specify a custom schema for `check_relations_equal`. Lo and behold, dive deep enough into our test frame and you find you can specify a custom schema. Problem solved! 👍

tests/functional/simple_seed/test_seed.py
VersusFacit: I tried to get the `project.drop_test_schema()` to work but no dice. Digging into the code didn't reveal any answers. Opted for this instead:
```python
182     @pytest.fixture(scope="function")
183     def clear_test_schema(self, project):
184         yield
185         project.run_sql(f"drop schema if exists {project.test_schema} cascade")
186
187     def test_postgres_simple_seed_with_disabled(self, clear_test_schema, project):
```

tests/functional/simple_seed/test_seed.py
VersusFacit: Oops, just an oversight. Draft PR things. good catch.

tests/functional/simple_seed/test_seed.py
VersusFacit: Streamlined several tests accordingly!

tests/functional/simple_seed/test_seed.py
VersusFacit: It was my thought you had to keep it as is because you need a file object and not a str (which is what read_file returns, when I tested the utility you suggest here).

tests/functional/simple_seed/test_seed.py
VersusFacit: I could weigh in better if I understood the primacy of the selection feature. Do we consider it a subfeature of seed or do we consider seed a command which mixes in the selection behavior. I feel like that reasoning best informs whether it should be moved to a selection test or not.

gshank (changes requested):
I have no idea what's up with the postgres error on Windows:
[DEBUG] configured_file: 22:21:45.398957 [debug] [MainThread]: On _test: /* {"app": "dbt", "dbt_version": "1.1.0b1", "profile_name": "test", "target_name": "default", "connection_name": "_test"} */
with diff_count as (
    SELECT
        1 as id,
        COUNT(*) as num_missing FROM (
            (SELECT "birthday", "email", "first_name", "ip_address", "seed_id" FROM "dbt"."test16495429020601287135_test_seed"."seed_expected" EXCEPT
             SELECT "birthday", "email", "first_name", "ip_address", "seed_id" FROM "dbt"."test16495429020601287135_test_seed"."seed_bom")
             UNION ALL
            (SELECT "birthday", "email", "first_name", "ip_address", "seed_id" FROM "dbt"."test16495429020601287135_test_seed"."seed_bom" EXCEPT
             SELECT "birthday", "email", "first_name", "ip_address", "seed_id" FROM "dbt"."test16495429020601287135_test_seed"."seed_expected")
        ) as a
), table_a as (
    SELECT COUNT(*) as num_rows FROM "dbt"."test16495429020601287135_test_seed"."seed_expected"
), table_b as (
    SELECT COUNT(*) as num_rows FROM "dbt"."test16495429020601287135_test_seed"."seed_bom"
), row_count_diff as (
    select
        1 as id,
        table_a.num_rows - table_b.num_rows as difference
    from table_a, table_b
)
select
    row_count_diff.difference as row_count_difference,
    diff_count.num_missing as num_mismatched
from row_count_diff
join diff_count using (id)
[DEBUG] configured_file: 22:21:45.398957 [debug] [MainThread]: Postgres adapter: Postgres error: column "seed_id" does not exist
LINE 7: ... "birthday", "email", "first_name", "ip_address", "seed_id" ...
                                                             ^
HINT:  There is a column named "seed_id" in table "*SELECT* 1", but it cannot be referenced from this part of the query.

tests/functional/simple_seed/test_seed.py
ChenyuLInx: Awesome!

gshank:
It looks like the current failure is in pytest-csv and is due to codepage 1252. So it could be that that pytest plugin isn't compatible with unicode csv files. But I have no idea how the previous error got turned into this error...

tests/functional/simple_seed/test_seed.py
```diff
@@ -1,19 +1,278 @@
...
+    def setUp(self, project):
+        """Create table for ensuring seeds and models used in tests build correctly"""
+        project.run_sql_file(project.test_data_dir / Path("seed_expected.sql"))
+
+    def test_simple_seed(self, project):
+
+        shutil.copyfile(
+            project.test_data_dir / Path("seed_bom.csv"),
+            project.project_root / Path("seeds") / Path("seed_bom.csv"),
+        )
```
ChenyuLInx, line 236: @gshank  this is currently being used to copy that file over to test dir. I wonder which part actually uses `pytest_csv`

gshank: It looks like pytest-csv is for creating test reports in csv format. So I have no idea how it got involved, but I saw it in the error log.

VersusFacit:
At last! On remembering some scattered notes about unicode I studied some 2ish months ago, I got the idea that the encoding for the unicode file might be cp1252 on Windows. Using the system codepoints and manually enforcing the encoding resolves this (Gerda suggested this yesterday I see and I wish I had seen it sooner!)

The insight came from comparing what the default behavior of open was on Windows vs. Unix-systems. I need to move some more things around and this PR is a mess. I'll reopen shortly for a much cleaner experience. :)

PR #5029, opened 2022-04-11
Update version output and logic

### LINKED ISSUE #4724: [CT-229] Adapter compatible message rule update
For adapter compatible message, there are two rules we missed in the previous commit:
- plugins can put out patch releases before dbt-core puts out a patch release, so `dbt-core==1.0.1` and `dbt-synapse==1.0.2` can be both available
- `dbt-core` + plugins are compatible at the minor-version level.

Here's the [original thread](https://github.com/dbt-labs/dbt-core/pull/4565#pullrequestreview-872805134), #4565

stu-k:
resolves #4724

### Description

Most notably, updated the logic for displaying plugin information as follows:
- If the plugin has neither the same major nor the same minor as the installed core version, it is considered incompatible
- Then, if the latest version of the plugin can not be found, a "no version available" message will be displayed
- Then, the plugin is compared with its latest version to display an appropriate message based on if it is ahead, matches, is behind the latest version

The tests capture the intended code paths, please review them for more information about outputs.

### Checklist

- [X] I have added information about my change to be included in the [CHANGELOG](https://github.com/dbt-labs/dbt-core/blob/main/CONTRIBUTING.md#Adding-CHANGELOG-Entry).


core/dbt/version.py
```diff
@@ -27,81 +51,175 @@ def get_latest_version(version_url: str = PYPI_VERSION_URL):
...
+
+def _get_plugin_msg_info(
+    name: str, version_s: str, core: Type[dbt.semver.VersionSpecifier]
+) -> Tuple[str, str]:
+    plugin = dbt.semver.VersionSpecifier.from_version_string(version_s)
+    latest_plugin = get_latest_version(version_url=get_package_pypi_url(name))
+
+    update_msg = ""
+
+    if plugin.major != core.major and plugin.minor != core.minor:
```
jtcohen6, line 131: [Blocking] This wants to be OR, or NOT AND, since a mismatch in either should count as incompatible. Either of these should do the trick:
```suggestion
    if plugin.major != core.major or plugin.minor != core.minor:
```
```suggestion
    if not (plugin.major == core.major and plugin.minor == core.minor):
```

core/dbt/version.py
```diff
@@ -27,81 +51,175 @@ def get_latest_version(version_url: str = PYPI_VERSION_URL):
...
+        compatibility_msg = red("Not compatible!")
+        update_msg = (
+            f"  Your version of dbt-{name} is not compatible with core!\n"
```
jtcohen6, line 134: [Not blocking, cosmetic only] This feels a little clearer to me:
```suggestion
            f"  The installed versions of dbt-core and dbt-{name} are not compatible.\n"
```

core/dbt/version.py
```diff
@@ -27,81 +51,175 @@ def get_latest_version(version_url: str = PYPI_VERSION_URL):
...
-def get_package_pypi_url(package_name: str) -> str:
-    return f"https://pypi.org/pypi/dbt-{package_name}/json"
+def _get_core_msg_lines(installed, latest) -> Tuple[List[List[str]], str]:
+    installed_s = installed.to_version_string(skip_matcher=True)
+    installed_line = ["installed", installed_s, ""]
+    update_info = ""
+
+    if latest is None:
+        update_info = (
+            "The latest version of dbt could not be determined!\n"
```
jtcohen6, line 69: [Not blocking, cosmetic only]
```suggestion
            "The latest version of dbt-core could not be determined.\n"
```

core/dbt/version.py
```diff
@@ -27,81 +51,175 @@ def get_latest_version(version_url: str = PYPI_VERSION_URL):
...
+            "  You can find instructions for upgrading here:\n"
+            "  https://docs.getdbt.com/dbt-cli/install/overview"
         )
+        return (compatibility_msg, update_msg)
 
-    else:
-        return (
-            "{}Your version of dbt is out of date! "
+    if not latest_plugin:
+        compatibility_msg = yellow("No PYPI version available")
```
jtcohen6, line 141: [Not blocking, cosmetic only] More like the `dbt-core` message above:
```suggestion
        msg = f"The latest version of dbt-{name} could not be determined from {get_package_pypi_url(name)}"
        compatibility_msg = yellow(msg)
```

core/dbt/version.py
```diff
@@ -27,81 +51,175 @@ def get_latest_version(version_url: str = PYPI_VERSION_URL):
...
+    latest_line = ["latest", latest_s, green("Up to date!")]
 
-def get_version_information():
-    flags.USE_COLORS = True if not flags.USE_COLORS else None
+    if installed > latest:
+        latest_line[2] = green("Ahead of latest version!")
+    elif installed < latest:
+        latest_line[2] = yellow("Update available!")
+        update_info = (
+            "Your version of dbt is out of date! "
```
jtcohen6, line 83: [Not blocking, cosmetic only]
```suggestion
            "Your version of dbt-core is out of date! "
```

core/dbt/version.py
```diff
@@ -27,81 +51,175 @@ def get_latest_version(version_url: str = PYPI_VERSION_URL):
...
+    for name, version_s in _get_dbt_plugins_info():
+        compatability_msg, update_msg = _get_plugin_msg_info(name, version_s, installed)
+        plugins.append([name, version_s, compatability_msg])
+        plugin_msgs.append(update_msg)
 
-    elif installed > latest:
-        return "{}Your version of dbt is ahead of the latest " "release!\n\n{}".format(
-            version_msg, plugin_version_msg
+    for i, plugin in enumerate(_pad_lines(plugins, seperator=":")):
+        msg_lines.append(_format_single_plugin(plugin, plugin_msgs[i]))
```
jtcohen6, line 119: [More significant feedback, shouldn't block initial merge]

I see that `_format_single_plugin` combines both the `msg` (short one-liners) and the `update_msg` (longer, more text). Was your thinking here that all lengthier "update" messages should actually go at the bottom, alongside the core update info?

To achieve that, I'd think about factoring the code like:
- `_format_single_plugin` accepts `plugin` and returns just that plugin's one-liner
- This method (`_get_plugins_msg`) returns each plugin's one-liner, and update messages (if required)
- `get_version_information()` combines all messages to create the following structure:
```
Core:
  - installed: ...
  - latest: ...

Plugins:
  - <one liner for each plugin>

<More text here if update message needed>
```

I gave that a quick go, but found the update messages get redundant quickly if you have multiple plugins installed:
```
$ dbt --version
Core:
  - installed: 1.0.2
  - latest:    1.0.4 - Update available!

Plugins:
  - postgres:  1.0.3   - Update available!
  - redshift:  1.1.0b1 - Not compatible!
  - snowflake: 1.1.0b1 - Not compatible!
  - bigquery:  1.1.0b1 - Not compatible!
  - spark:     1.1.0b1 - Not compatible!

  Your version of dbt-postgres is out of date! You can find instructions for upgrading here:
  https://docs.getdbt.com/dbt-cli/install/overview
  Your version of dbt-redshift is not compatible with core!
  https://docs.getdbt.com/dbt-cli/install/overview
  Your version of dbt-snowflake is not compatible with core!
  https://docs.getdbt.com/dbt-cli/install/overview
  Your version of dbt-bigquery is not compatible with core!
  https://docs.getdbt.com/dbt-cli/install/overview
  Your version of dbt-spark is not compatible with core!
  https://docs.getdbt.com/dbt-cli/install/overview
```

At which point, I'd almost rather just make it say:
```
$ dbt --version
Core:
  - installed: 1.0.2
  - latest:    1.0.4 - Update available!

Plugins:
  - postgres:  1.0.3   - Update available!
  - redshift:  1.1.0b1 - Not compatible!
  - snowflake: 1.1.0b1 - Not compatible!
  - bigquery:  1.1.0b1 - Not compatible!
  - spark:     1.1.0b1 - Not compatible!

Your installed version of dbt-core is out of date.
```

jtcohen6 (commented):
Nice work @stu-k, this represents a meaningful improvement and refactor of the code! I left a handful of comments. Only one is a true blocker (`Major.Minor` compatibility comparison). The rest are smaller + cosmetic only.

There's one bigger question about the right way to present upgrade info when it's relevant to core + many plugins. I think it would be worth a more generic message that prints `https://docs.getdbt.com/dbt-cli/install/overview` once, rather than once for each.

core/dbt/version.py
stu-k: I think it's a bit overkill to have the name of the version twice, as well as having the URL on the same line. This change would also make this the longest message at the end of a line.
```
Plugins:
  - redshift:         1.0.0 - Update available!
  - bigquery:       1.0.0 - Not compatible!
  - postgres:       1.0.0 - Up to date!
  - snowflake:     1.0.0 - Your version of dbt-snowflake could not be determined from https://pypi.org/pypi/dbt-snowflake/json
```
This seems a bit out of place.

core/dbt/version.py
stu-k: Is it necessary to call out `dbt-core` instead of `dbt`? Do users call it `dbt-core` regularly?

core/dbt/version.py
stu-k: Similar to my question below, but how often do users call `dbt` `dbt-core`?

core/dbt/version.py
stu-k: Could this be something like the following, so the user has a bit more context for which message belongs to which section?
```
$ dbt --version
Core:
  - installed: 1.0.2
  - latest:    1.0.4 - Update available!

  Your installed version of dbt is out of date.

Plugins:
  - postgres:  1.0.3   - Update available!
  - redshift:  1.1.0b1 - Not compatible!
  - snowflake: 1.1.0b1 - Not compatible!
  - bigquery:  1.1.0b1 - Not compatible!
  - spark:     1.1.0b1 - Not compatible!

  At least one plugin is out of date or incompatible with dbt.
```

core/dbt/version.py
jtcohen6: Good question!
- People call this `dbt-core` to distinguish it from the dbt adapter plugin (`dbt-postgres`, `dbt-redshift`, ...)
- For trademark purposes, there exists an important distinction between "dbt" (Core + Cloud) and "dbt Core" (hence why we renamed this repo)

stu-k: That makes sense!

core/dbt/version.py
stu-k: Disregard!

core/dbt/version.py
jtcohen6: Good point — in context it doesn't look quite right. My main interest here was the PyPi link, since we're just naively assuming that the PyPi link is `f"https://pypi.org/pypi/dbt-{name}/json"`. What do you think of:
```suggestion
        compatibility_msg = yellow(f"Could not get latest version from {get_package_pypi_url(name)}")
```

jtcohen6: Not really — it just lets us show our cards, since we're being naive about where we're checking for the link :)

`"No PYPI version available"` feels, to me, like we're being a bit smarter / more thorough than we really are. I'd be fine with `Could not determine latest version`

core/dbt/version.py
jtcohen6: I like tha!

core/dbt/version.py
```diff
@@ -27,81 +51,175 @@ def get_latest_version(version_url: str = PYPI_VERSION_URL):
     return dbt.semver.VersionSpecifier.from_version_string(version_string)
 
 
-def get_installed_version():
-    return dbt.semver.VersionSpecifier.from_version_string(__version__)
+def _get_core_msg(installed, latest) -> Tuple[str, str]:
+    core_msg_lines, update_info = _get_core_msg_lines(installed, latest)
+    return (
+        _format_core_msg(core_msg_lines),
```
ChenyuLInx, line 57: Looks like this would just call the other two functions, should we consider consolidating three functions into one and just use comments to separate it into pieces?

ChenyuLInx (commented):
This looks great! Very solid refactor of the code! Left one nitpick that's only my personal preference. Can also leave it out if you feel like the current way is easier to read!

core/dbt/version.py
stu-k: Totally! I think this was stripped down during refactoring, and now serves no real purpose.

ChenyuLInx:
Looks like there's a mypy error, otherwise looks good!

ChenyuLInx:
> @ChenyuLInx That error doesn't have to do with anything I added or changed.
> The same line exists on `main` https://github.com/dbt-labs/dbt-core/blob/main/core/dbt/config/project.py#L194

Yes, looks like this line is not even touched. I am not sure why mypy complains about it either. But when I run mypy on main it doesn't have any error. We will have to fix it or ignore this to get the workflow pass

ChenyuLInx (approved):

PR #5034, opened 2022-04-12
CT-476 convert deprecation tests

### LINKED ISSUE #5030: [CT-476] Convert 012_deprecation_tests
Update tests related to deps.

012_deprecation_tests

Create a new test that will mock the response to be valid/invalid so we can test the retry and caching logic.

emmyoop:
resolves #5030

### Description

Converts 012_deprecation_tests

### Checklist


gshank:
The project fixture when it runs the drop_schema after the yield expects there to be an installed adapter and there isn't after the dbt deps command fails, it looks like.  Probably the easiest way to fix it is to add a try/except around the drop_schema after the yield in the project fixture.

It would be nice to have a better way of dealing with that, but it would take a bit of work.

core/dbt/tests/fixtures/project.py
```diff
@@ -454,5 +454,9 @@ def project(
 
     yield project
 
-    project.drop_test_schema()
+    # Some commands will not have an installed adapter when running and will raise a KeyError here.  Just pass for now.
+    try:
+        project.drop_test_schema()
+    except KeyError:
```
ChenyuLInx, line 460: There was another case where if you run dbt with another target the adapter will be switched and raise an error here. @gshank I feel like a better solution might be
- keep track of all the adapters we used in the test and loop over all of them to drop the schema it was using

gshank: The reason that I switched to using the most recent adapter was that if we use an adapter that's not the latest we have to mock.patch the providers in order to avoid errors running sql, and we have to do it in every place that we connect to the database (this would be the case for any "saved" adapter that's not the latest).  Any adapter object should be able to drop the schema (if one exists) and it has the normal permissions.   The problem in the previous test that we dropped was that it was hacking the permissions in a kind of stupid way. The problem here is that there is no adapter -- which I didn't realize would be the case for deps, but makes sense because we now don't load the profile.

The more complete solution would be to check that an adapter exists and if it doesn't load one, but that seemed like too much work for this particular pull request.

gshank: Opened ticket #5041

gshank: Note: normally this would be easy to fix by running run_dbt() at the end of the test, but because this test fails when loading dependencies, that won't work.

ChenyuLInx: Right, I always forgot the persmission part. Then just check existance of adatper sounds perfect!

core/dbt/tests/fixtures/project.py
```diff
@@ -454,5 +454,13 @@ def project(
...
 
-    project.drop_test_schema()
+    # deps, debug and clean commands will not have an installed adapter when running and will raise
+    # a KeyError here.  Just pass for now.
+    # See https://github.com/dbt-labs/dbt-core/issues/5041
+    # The debug command also results in an AttributeError since `QueryCommentedProfile` doesn't have
+    # a `load_dependencies` method.
+    try:
+        project.drop_test_schema()
+    except (KeyError, AttributeError):
```
ChenyuLInx, line 464: @gshank I remember you had some changes for drop_test_schema in one of your PR, is that merged in? Is this still needed if your changes are merged?

gshank: I changed create_schema and drop_schema to allow multiple schemas, but I think this try/except is still needed for this test. It would be good to figure out a more complete way of handling this, but I think this is okay for now.

ChenyuLInx (approved):
Looks good to me!

Issue #4961, opened 2022-03-25
[CT-429] Add tracking and more tests for cache only selected nodes


ChenyuLInx:
We would want to capture the config value added in #4860.
When we have more understanding of how this flag changes behavior we should also start to add some tests around intended behavior for future reference.

PR #5036, opened 2022-04-12
Add experimental `cache_selected_only` config

### LINKED ISSUE #4688: [CT-168] Cache objects for selected resources only?

jtcohen6:
resolves #4688
rebased version of #4860
more tracking + testing to come in https://github.com/dbt-labs/dbt-core/issues/4961

Only other changes:
- `--cache-selected-only` (kebab case) instead of `--cached_selected_only` (snake case)
- Add a breaking change note for the internal adapter method (Python code) with a change in its signature. Among our plugins, only `dbt-postgres` reimplements this method; all the others inherit the base version.

I'd love to include this as "experimental" functionality in v1.1 if possible. We would document it as such: https://github.com/dbt-labs/docs.getdbt.com/pull/1327

### Checklist

### DIFF
--- a/.changes/unreleased/Breaking Changes-20220412-152450.yaml
+++ b/.changes/unreleased/Breaking Changes-20220412-152450.yaml
@@ -1,0 +1,9 @@
+kind: Breaking Changes
+body: 'For adapter plugin maintainers only: Internal adapter method _relations_cache_for_schemas
+  takes an additional argument, cache_schemas, for use with experimental CACHE_SELECTED_ONLY
+  config'
+time: 2022-04-12T15:24:50.159572+02:00
+custom:
+  Author: karunpoudel
+  Issue: "4688"
+  PR: "4860"
--- a/.changes/unreleased/Breaking Changes-20220412-152450.yaml
+++ b/.changes/unreleased/Breaking Changes-20220412-152450.yaml
@@ -1,6 +1,6 @@
 kind: Breaking Changes
-body: 'For adapter plugin maintainers only: Internal adapter method _relations_cache_for_schemas
-  takes an additional argument, cache_schemas, for use with experimental CACHE_SELECTED_ONLY
+body: 'For adapter plugin maintainers only: Internal adapter methods `set_relations_cache` + `_relations_cache_for_schemas`
+  each take an additional argument, for use with experimental `CACHE_SELECTED_ONLY`
   config'
 time: 2022-04-12T15:24:50.159572+02:00
 custom:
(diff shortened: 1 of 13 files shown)


gshank (approved):
Looks good! I see there's only a test for the flag setting not the caching though.

gshank:
Changing an existing test is perfectly legit :) They ought to behave the same...

ChenyuLInx (commented):
Late to the party but this looks great!

rmargarint-nydig:
I realized this is a comment on an old PR, but I was experimenting today with `cache_selected_only` to speed up start up time on our test execution pipelines which run as part of Airflow orchestration. It seems that `cache_selected_only` only seems to affect `dbt run` but not `dbt test --select "test_name"`? From what I've traced, it seems `get_model_schemas(...)` will return an empty set when one of the selected uids is a test resource, which will result in the entire set of schemas being pulled into the cache. Is this expected?

abecquet77:
I'm having the same issue. `cache_selected_only` saves us a lot of time in Airflow with Cosmos. 
But `cache_selected_only` doesn't work with `dbt test`.

Issue #4745, opened 2022-02-18
[CT-255] [Bug] ImportError: cannot import name soft_unicode from markupsafe


mailtosumitrai:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Markup recently released its new version 2.1.0 which is essential for Jinja2 package. Jinja has already released version 3.x.x and dbt still uses version 2.x.x and their [developers won't support 2.x.x for some time](https://github.com/pallets/jinja/issues/1585). Now the command `dbt deps` throws an error.

### Expected Behavior

While running `dbt deps` the packages should install instead of throwing errors.

### Relevant log output

```shell
Traceback (most recent call last):
  File "/opt/hostedtoolcache/Python/3.8.12/x64/bin/dbt", line 5, in <module>
    from dbt.main import main
  File "/opt/hostedtoolcache/Python/3.8.12/x64/lib/python3.8/site-packages/dbt/main.py", line 11, in <module>
    import dbt.version
  File "/opt/hostedtoolcache/Python/3.8.12/x64/lib/python3.8/site-packages/dbt/version.py", line 11, in <module>
    import dbt.semver
  File "/opt/hostedtoolcache/Python/3.8.12/x64/lib/python3.8/site-packages/dbt/semver.py", line 8, in <module>
    import dbt.utils
  File "/opt/hostedtoolcache/Python/3.8.12/x64/lib/python3.8/site-packages/dbt/utils.py", line 9, in <module>
    import jinja2
  File "/opt/hostedtoolcache/Python/3.8.12/x64/lib/python3.8/site-packages/jinja2/__init__.py", line 12, in <module>
    from .environment import Environment
  File "/opt/hostedtoolcache/Python/3.8.12/x64/lib/python3.8/site-packages/jinja2/environment.py", line 25, in <module>
    from .defaults import BLOCK_END_STRING
  File "/opt/hostedtoolcache/Python/3.8.12/x64/lib/python3.8/site-packages/jinja2/defaults.py", line 3, in <module>
    from .filters import FILTERS as DEFAULT_FILTERS  # noqa: F401
  File "/opt/hostedtoolcache/Python/3.8.12/x64/lib/python3.8/site-packages/jinja2/filters.py", line 13, in <module>
    from markupsafe import soft_unicode
[...]

PR #5039, opened 2022-04-12
Flexibilize MarkupSafe pinned version

adamantike:
### Description

The current `MarkupSafe` pinned version has been added in #4746 as a temporary fix for #4745.

However, the current restrictive approach isn't compatible with other libraries that could require an even older version of `MarkupSafe`, like Airflow `2.2.2` [0], which requires `markupsafe>=1.1.1, <2.0`.

To avoid that issue, we can allow a greater range of supported `MarkupSafe` versions. Considering the direct dependency `dbt-core` has is `Jinja2==2.11.3`, we can use its pinning as the lower bound, which is `MarkupSafe>=0.23` [1].

This fix should be also backported this to `1.0.latest` for inclusion in the next v1.0 patch.

[0] https://github.com/adamantike/airflow/blob/2.2.2/setup.cfg#L125
[1] https://github.com/pallets/jinja/blob/2.11.3/setup.py#L53

### Checklist

### DIFF
--- a/.changes/unreleased/Fixes-20220413-151735.yaml
+++ b/.changes/unreleased/Fixes-20220413-151735.yaml
@@ -1,0 +1,7 @@
+kind: Fixes
+body: Relax minimum supported version of MarkupSafe
+time: 2022-04-13T15:17:35.055274145-03:00
+custom:
+  Author: adamantike
+  Issue: "4745"
+  PR: "5039"
--- a/core/setup.py
+++ b/core/setup.py
@@ -52,7 +52,7 @@ setup(
     ],
     install_requires=[
         "Jinja2==2.11.3",
-        "MarkupSafe==2.0.1",
+        "MarkupSafe>=0.23,<2.1",
         "agate>=1.6,<1.6.4",
         "click>=7.0,<9",
         "colorama>=0.3.9,<0.4.5",


adamantike:
Not sure if a change like this one would require a changelog entry. Open to suggestions!

iknox-fa:
@adamantike Thanks for your PR.. a changelog entry for this would be great!

iknox-fa (approved):
LGTM

adamantike:
Thanks for the quick reply! So, to validate before adding it to the changelog, would this be applicable to be backported to `1.0.x`? 

In case it is, should I add the changelog entry for that here too, or do you take care of that when cherry-picking the change?

ChenyuLInx:
@adamantike we changed our changelog system to [changie](https://github.com/dbt-labs/dbt-core/blob/main/CONTRIBUTING.md#adding-changelog-entry) so all you need to do is follow the instruction in the link and the actual `changelog.md` will be generated during release. 
@iknox-fa is there a reason we are fixing the upper bound version of `MarkUpSafe` we can use? I think the `install_requires` is usually only used for lower bound unless there's a good reason to cap the version.

adamantike:
Thanks, I have added the changelog entry after following the `changie` instructions.

Regarding why we need to provide an upper bound, it's because of this discussion: https://github.com/pallets/jinja/discussions/1602. `jinja` didn't fix it in their side because version 2.x isn't supported anymore, so `dbt-core` requires this pinning until the `jinja2` dependency can be upgraded to version 3.x.

ChenyuLInx:
@adamantike  Thanks for contributing!!!

adamantike:
Thanks for the quick resolution!

With this PR merged, is there a formal process to ask for this fix to be backported to `1.0.latest`?

stu-k:
@adamantike Thanks again for the contribution! Our team will handle backporting your change!

PR #5057, opened 2022-04-13
feat: skip project name prompt with --skip-profile-setup flag

### LINKED ISSUE #5056: [CT-490] [Feature] init: respect --skip-profile-setup when project name is invalid

### Describe the Feature

If someone is using dbt in a non-interactive manner, you'd think the `-s` flag would ensure the script isn't stuck in place waiting for some input from the user.

Reproduction on dbt 1.0.4
```
dbt init invalid-project-name -s
```
This will result in dbt showing a prompt to the user for a new project name. Whereas a more sensible thing to do is print an error message and exit.

Ideally, in all cases, when `init` is run with `-s`/`--skip-profile-setup`, we wouldn't prompt the user with any.

Or perhaps there should be a separate flag, for example `--no-interactive`, that skips all interactions and exits if unable proceed?

There's really no other options besides making sure you enter the right options/args and run with the correct dbt config to ensure the init command doesn't trigger a prompt.

### Who will this benefit?

Anyone using dbt in a non-interactive manner.

Yes

### Anything else?

Related to https://github.com/dbt-labs/dbt-core/issues/4568

varun-dc:
resolves #5056

### Description
Check if the `--skip-profile-setup` flag is set, if so, no prompt is triggered upon an invalid project name.

### Checklist

### DIFF
--- a/core/dbt/events/types.py
+++ b/core/dbt/events/types.py
@@ -2296,6 +2296,15 @@ class DepsSetDownloadDirectory(DebugLevel):
 
 
 @dataclass
+class ProjectInvalidName(InfoLevel):
+    name: str
+    code: str = "A028"
+
+    def message(self) -> str:
+        return f"{self.name} is not a valid project name."
+
+
+@dataclass
 class EnsureGitInstalled(ErrorLevel):
     code: str = "Z036"
 
--- a/core/dbt/task/init.py
+++ b/core/dbt/task/init.py
@@ -3,7 +3,7 @@ import os
 from pathlib import Path
 import re
 import shutil
-from typing import Optional
+from typing import Optional, Tuple
 
 import yaml
 import click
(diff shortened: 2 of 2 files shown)


core/dbt/events/types.py
```diff
@@ -2295,6 +2295,15 @@ def message(self) -> str:
         return f"Set downloads directory='{self.path}'"
 
 
+@dataclass
+class ProjectInvalidName(InfoLevel):
```
stu-k, line 2299: If this event is only going to be fired in cases like `dbt init invalid-project-name -s`, this might be better suited as an `ErrorLevel` event.

stu-k, line 2299: If this event is only going to be fired in cases like `dbt init invalid-project-name -s`, this might be better suited as an `ErrorLevel` event.

PR #5058, opened 2022-04-13
Convert docs_generate_tests to new framework

### LINKED ISSUE #5035: [CT-478] Convert 029_docs_generate tests to new framework.
Convert the docs generate tests to new framework and move into adapter zone.

gshank:
resolves #5035

### Description

Convert the 029_docs_generate_tests to new pytest framework.

Pull request opened for initial review, but will also be splitting up these tests into adapter zone and non-adapter-zone tests.

### Checklist


gshank:
I am going to split this test into two parts, a 'docs_generate' part in the adapter zone, and an 'artifacts' tests in dbt-core standard tests.

ChenyuLInx:
> I am going to split this test into two parts, a 'docs_generate' part in the adapter zone, and an 'artifacts' tests in dbt-core standard tests.

Is the docs generate part going to be mainly related to catalog generate?

gshank:
Yes, I'll remove the manifest and run_results varification from the adapter zone test, and remove the catalog verification from the artifacts tests. In the artifacts tests I will try to switch to compile instead of 'docs generate".

ChenyuLInx (approved):
This looks great! Love that adapter one only have catalog related

core/dbt/tests/fixtures/project.py
```diff
@@ -120,6 +120,11 @@ def dbt_profile_target():
     }
 
 
+@pytest.fixture(scope="class")
+def profile_user(dbt_profile_target):
```
ChenyuLInx, line 124: Which adapter need this change and why?

core/dbt/tests/util.py
```diff
@@ -430,3 +435,46 @@ def check_table_does_not_exist(adapter, name):
...
+
+    def __eq__(self, other):
+        return isinstance(other, int)
+
+
+class AnyString:
+    """Any string. Use this in assert calls"""
+
+    def __eq__(self, other):
+        return isinstance(other, str)
```
ChenyuLInx, line 461: The three classes here looks very similar, why we are doing it as 3 individual instead of having one that you can maybe pass in a specific type? Also, think we can achieve `AnyString` by not passing in contains for `AnyStringWith`?

core/dbt/tests/fixtures/project.py
gshank: We need a way to get the "owner" in the catalog. It's different for different adapters. Some of them use the "user" in the profile, some use other things. So this is special fixture to help with that. We didn't have a standard way to get the profile's user before.

core/dbt/tests/util.py
gshank: They're separate classes because they were copied from existing classes in the integration test suite. You do have a point that a single class that takes a type would work here.

PR #5065, opened 2022-04-14
add new GHA for dependabot PRs

emmyoop:
resolves #5050

### Description

- [Action that generated changelog file](https://github.com/dbt-labs/dbt-core/runs/6024931113?check_suite_focus=true)
- I've verified this works when `dependencies` label is added before and after PR creation.  That seems sufficient for these purposes.
- This action checks that a changelog file does not already exist

- Note: I'm leaving the autogenerated changelog yaml so the outputs can be viewed.  It needs to be deleted before merge.  Or perhaps converted into its own changelog `Under the Hood`.

Open Questions
1. Should this be backported?  Probably not necessary since we wouldn't make a change directly to the `.latest` branch so the changelog file itself would be backported on a dependency update.  Can still backport for safety though.  It won't break anything.
2. If changie is going to be implemented on other repos should this (and [`changelog_check`](https://github.com/dbt-labs/dbt-core/blob/main/.github/workflows/changelog-check.yml)) be written as reusable actions?  This could be tweaked to take in the kind (Dependency is this case) and label and be able to autogenerate a changie file for maximum reuse.  Although I'm not sure that would work when the branch to commit to is actually a contributors fork?
3. 
### Checklist


.github/workflows/dependency_changelog.yml
```diff
@@ -0,0 +1,111 @@
...
+name: Dependency Changelog
+
+on:
+  pull_request:
+    # catch when the PR is opened with the label or when the label is added
+    types: [opened, labeled]
+
+permissions:
+  contents: write
+  pull-requests: read
```
emmyoop, line 31: Since it's committing to the branch I think this permission setting is least privilege.

.github/workflows/dependency_changelog.yml
```diff
@@ -0,0 +1,111 @@
...
+      # The action to update the PR expects the author to be in the format of name <email>.
+      # Grab the name and email from the last commit to this PR and use it to commit the new
+      # changelog file
+      if: steps.filter.outputs.changelog == 'false'
+      id: user
+      run: |
+        USERNAME="$(git log -n 1 --pretty=format:%an)"
+        USEREMAIL="$(git log -n 1 --pretty=format:%ae)"
+        echo "::set-output name=USERNAME::$USERNAME"
+        echo "::set-output name=USEREMAIL::$USEREMAIL"
```
emmyoop, line 91: Would _love_ to hear this is not the best way to get the name/email but it's the best I could come up with!

emmyoop:
clore/reopen to retrigger tests that are hanging

.github/workflows/dependency_changelog.yml
emmyoop: @leahwicz  it makes sense to replace this with a bot as we discussed offline. [dbtBot](https://github.com/dbt-labs/dbtBot) is the only one I could find but it doesn't seem right by the README?

jtcohen6: There's also @FishtownBuildBot

leahwicz: Let's use https://github.com/FishtownBuildBot. We use it for some commits for version bumping now that I look
https://github.com/dbt-labs/dbt-core/blob/080dd41876aaf5d348f75db402b2efbb074be4a2/.github/workflows/version-bump.yml#L95

.github/workflows/dependency_changelog.yml
```diff
@@ -0,0 +1,100 @@
...
+permissions:
+  contents: write
+  pull-requests: read
+
+jobs:
+  dependency_changelog:
+    if: ${{ github.event.label.name == 'dependencies' }}
+    runs-on: ubuntu-latest
+
+    steps:
```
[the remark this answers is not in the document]
emmyoop: That helps quite a bit with readability!

.github/workflows/dependency_changelog.yml
leahwicz: I think this is correct

leahwicz (approved):
LGTM!

emmyoop:
@leahwicz I've reworked this a bit since you approved it and would appreciate your eyes on it one more time.  Most of the changes are adding more comments as to why I'm doing things and adding in using the PAT.

.github/workflows/dependency-changelog.yml
```diff
@@ -0,0 +1,114 @@
...
+    # catch when the PR is opened with the label or when the label is added
+    types: [opened, labeled]
+
+permissions:
+  contents: write
+  pull-requests: read
+
+jobs:
+  dependency_changelog:
+    if: ${{ github.event.label.name == 'dependencies' }}
```
leahwicz, line 35: If a PR is opened with the label, will this `github.event.label.name ` have a value since the event is a PR open event and not a labeled event? 

If this becomes a snag, I'm ok with this only triggering on labeling right now since for dependencies, these are all auto-opened for us

leahwicz (approved):
LGTM 👍 !

.github/workflows/dependency-changelog.yml
emmyoop: Yes.  It seems that `github.event...` is not always an actual event.  Similar to how `github.event.pull_request.merged` is not an event but an attribute.  Not intuitive.

I did specifically test the case of adding the label before the PR is created and it triggers the workflows as expected and creates a changelog file.

PR #5069, opened 2022-04-14
move deprecation check outside package caching

### LINKED ISSUE #5068: [CT-494] Move the deprecation check for dependancies outside the caching logic
Move the deprecation check for dependancies outside the caching logic.  The deprecations don't actually get cached, just the response so it's not really valid.

Relevant code:
https://github.com/dbt-labs/dbt-core/blob/080dd41876aaf5d348f75db402b2efbb074be4a2/core/dbt/clients/registry.py#L92-L109

emmyoop:
resolves #5068

### Description

Check deprecation of packages outside of package caching.  This is not actually causing error today.  However, the deprecation doesn't get cached so it should live outside that logic.  It's also preventing tests from working as expected so a real fix here allows for better tests.

### Checklist


gshank (approved):
Looks good.

PR #5082, opened 2022-04-18
Updating backport action to latest

leahwicz:
### Description
The backport Action made several changes including security updates and our ask to give a branch name. I'm bumping the version of the action to pull in those changes and also updating our own workflow file to their recommended example which is more security conscious. 

Their example:
https://github.com/tibdex/backport/blob/main/.github/workflows/backport.yml

TBD: I can delete our internal fork now since we have the workaround and fix in place

### Checklist

### DIFF
--- a/.github/workflows/backport.yml
+++ b/.github/workflows/backport.yml
@@ -35,13 +35,7 @@ jobs:
     # See https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target.
     if: >
       github.event.pull_request.merged
-      && (
-        github.event.action == 'closed'
-        || (
-          github.event.action == 'labeled'
-          && contains(github.event.label.name, 'backport')
-        )
-      )
+      && contains(github.event.pull_request.labels, 'backport')
     steps:
       - uses: tibdex/backport@v2.0.2
         with:
--- a/.github/workflows/backport.yml
+++ b/.github/workflows/backport.yml
@@ -16,15 +16,16 @@
 # Once a PR is "Squash and merge"'d and it has been correctly labeled
 # according to the naming convention.
 
-# This workflow file is copied straight from the given example:
-# https://github.com/tibdex/backport/blob/main/.github/workflows/backport.yml
-
 name: Backport
 on:
-  pull_request_target:
+  pull_request:
     types:
       - closed
       - labeled
+
+permissions:
+  contents: write
+  pull-requests: read
 
 jobs:
   backport:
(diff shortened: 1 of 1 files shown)


.github/workflows/backport.yml
```diff
@@ -16,19 +16,32 @@
...
+    runs-on: ubuntu-latest
+    # Only react to merged PRs for security reasons.
+    # See https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target.
+    if: >
+      github.event.pull_request.merged
+      && (
+        github.event.action == 'closed'
+        || (
+          github.event.action == 'labeled'
+          && contains(github.event.label.name, 'backport')
```
emmyoop, line 41: Is `github.event.pull_request.merged` not an event that happens when we merge?  So if we add a label after a merge this will never trigger since we would never pass this conditional.

For simplicity, we could just check here if the PR is merged (and closed?) and let the backport action handle the rest.  This is not currently checking all the logic as is since it would currently run (but not backport) if merged and closed but not labeled.

leahwicz: `github.event.pull_request.merged` is an attribute on the PR which has triggered the event and not the event itself. The event follows the format `github.event.action` to tell you what the actual event triggered the workflow. I verified running some of my own tests and the `merged` attribute prints out true/false according to the merge state of the PR and not based off the trigger. Checking that `merged` is true seems like the correct thing here then

We then check 1) if merged 2) if the action was a PR being closed or a backport label being applied which seems correct to me. Does that make sense to you? The checking of the label is extra like you said just to save on running time since the action checks as well

emmyoop: > `github.event.pull_request.merged` is an attribute on the PR which has triggered the event and not the event itself

This makes sense then!  Good to know.

As for the rest, we're already only triggering this action on `closed` or `labeled` per line 26/27 so this would be a bit simpler and catch all cases:
```
if: >
      github.event.pull_request.merged
      && contains(github.event.label.name, 'backport')
```

leahwicz: That makes sense to me. I'm guessing the Actions author didn't know you could check like this

leahwicz: I made the change to only trigger on label changes and not the PR closing. It makes it simple. You can only add the label to backport AFTER the PR is closed. That let's us know exactly what is being backported at the time we make the decision. That makes it so we don't assume we want to backport early and add the label but then only realize after a change grows that it may not be a good candidate to backport

emmyoop (approved):
Looks good!

PR #5107, opened 2022-04-19
Feature: Add set/zip function to contexts

### LINKED ISSUE #2345: Make zip function available in Jinja
Would like to use zip to loop over some results in a Jinja template

```
{% for j,k in zip(a, b) %
```

Currently zip is not available, but would make my code a much cleaner.

jeremyyeo:
resolves #2345 

### Description

Adding the `set` and `zip` functions as prescribed in the issue linked above.

### Checklist


jeremyyeo:
@jtcohen6 any thoughts on adding any of the others mentioned in the issue (i.e. `any()`, `all()`)? 

Also had some flake8 fails on:

```
flake8...................................................................Failed
- hook id: flake8
- exit code: 1

core/dbt/context/base.py:8:1: F401 'dbt.clients.yaml_helper.SafeLoader' imported but unused
core/dbt/context/base.py:8:1: F401 'dbt.clients.yaml_helper.Loader' imported but unused
core/dbt/context/base.py:8:1: F401 'dbt.clients.yaml_helper.Dumper' imported but unused
```

in my pre-commit, wasn't sure if it was okay to remove or not so I just committed with `--no-verify`.

core/dbt/context/base.py
```diff
@@ -457,6 +457,48 @@ def toyaml(
...
+
+        Usage:
+            {% set my_list = [1, 2, 2, 3] %}
+            {% set my_set = set(my_list) %}
+            {% do log(my_set) %}  {# {1, 2, 3} #}
+        """
+        try:
+            return set(value)
+        except TypeError:
+            return default
```
emmyoop, line 478: We've been having some internal discussion around the try/except pattern you see in this file.  We've decided to maintain the current state but also add new partner methods that would allow for failure instead of passing the default back.  Something that would need to be opted in to.  #5167 has some more details of the pattern we would like to follow.

Please add in another function called `try_set` that will fail on a `TypeError`.  Same applies for `zip` below.

emmyoop (changes requested):
Looks good so far.  Sorry it took a bit to review!  We've been having some conversations around the best pattern for functions that we add.

gshank:
Did you want to try creating the "try_set" methods that @emmyoop asked for?

jeremyyeo:
> Did you want to try creating the "try_set" methods that @emmyoop asked for?

Hey @gshank yep keen to do that but kind of waiting to see if we will be adding a new special error type to raise (#5167). You reckon I should just proceed with a generic `TypeError` for now?

```python
def try_set(value):
    try:
        return set(value)
    except TypeError:
        raise TypeError
```

gshank:
It looks like we mainly raise CompilationExceptions in the contexts, partly because Jinja catches the exceptions and we don't have code to distinguish between parse-time and run-time for those errors. For now, just make it a CompilationException? We plan to do some work on organizing exceptions at some point and we can clean it up then.

jeremyyeo:
@gshank / @emmyoop what do you think of the `try_` methods "skipping" the raise if the user was the one that set up a default return value? Would make it behave similar to other methods, e.g:

```
var('foo')  
> error if variable 'foo' not set.

var('foo', 'bar') 
> var 'foo' set to 'bar' if var 'foo' not provided.
```

emmyoop:
@jeremyyeo the `try` methods shouldn't allow a default.  The intention of their use is for when the value really should always be the expected type.  ie. if it's not iterable, etc it should be a hard stop.  If the project wants to allow a default then the non-`try` version of the method should be used.  We plan to have both versions of these methods where appropriate so the most appropriate one can be used and we don't break any projects going forward.

gshank:
Looks like this is getting really close :-). If you can just add a couple of simple tests we can get this wrapped up. Probably a test in tests/functional/context_methods would be good; just something simple like a model that uses the functions.

tests/functional/context_methods/test_builtin_functions.py
```diff
@@ -0,0 +1,55 @@
...
+
+from dbt.tests.util import run_dbt, write_file
+from dbt.exceptions import CompilationException
+
+
+tests__set_sql = """
+{% set set_result = set([1, 2, 2, 3, "foo", False]) %}
+{% set try_set_result = try_set([1, 2, 2, 3, "foo", False]) %}
+
+{% set simple_set = (set_result == try_set_result == set((1, 2, 3, "foo", False))) %}
```
emmyoop, line 11: This check is a bit off since you're using the new method to check the new method.  

Instead of using tests I would use macros that log the values of `set_result` and`try_set_result`.  Something like:

```sql
{% macro validate_set() %}
    {% set set_result = set([1, 2, 2, 3, 'foo', False]) %}
    {{ log("set_result: " ~ set_result) }}
    {% set try_set_result = try_set([1, 2, 2, 3, 'foo', False]) %}
    {{ log("try_set_result: " ~ try_set_result) }}
{% endmacro %}
```

Then in the test you can capture the logs from `run-operation` with the `--debug` flag for the macro and assert the logs show up:

```python
def test_builtin_set_function(self, project):
        _, log_output = run_dbt_and_capture(["--debug", "run-operation", "validate_set"])

        expected_set = {False, 1, 2, 3, 'foo'}
        assert f"set_result: {expected_set}" in log_output
        assert f"try_set_result: {expected_set}" in log_output
```

Also do the same for zip for consistency.

emmyoop (changes requested):
Just some test changes but otherwise is looking great!

emmyoop (approved):
This looks great @jeremyyeo!  Thanks for working with us on the few iterations as we move towards a better way to implement these methods.

I'll get it merged in!

emmyoop:
@jtcohen6 should this be back ported to `1.1.latest`?

leahwicz:
@emmyoop I would wait for this to go out in v1.2 since it's net new functionality. @jtcohen6 let me know if you feel differently

skwaugh:
Any chance of this making it into 1.2?

jtcohen6:
@skwaugh This will be in v1.2!

@jeremyyeo @emmyoop I'm working on docs for this right now (https://github.com/dbt-labs/docs.getdbt.com/issues/1635), and it looks like some of the docstrings need updating. (E.g. all methods include `:param default: A default value to return if `*args` is not iterable`, including the ones that do not accept a `default` argument.)

One other note: To my mind, the implementation of `set` and `try_set` here feels exactly backwards. A big inspiration for the `try_` nomenclature is Snowflake's [`try_cast`](https://docs.snowflake.com/en/sql-reference/functions/try_cast.html) and BigQuery's [`safe_cast`](https://cloud.google.com/bigquery/docs/reference/standard-sql/functions-and-operators#safe_casting), as opposed to just `cast`. The idea there is that `cast` raises an error if the provided column expression / literal value cannot be converted or coerced to the desired type, whereas `try_cast`/`safe_cast` return a default value (`null`) instead.

> When using CAST, a query can fail if BigQuery is unable to perform the cast. If you want to protect your queries from these types of errors, you can use [SAFE_CAST](https://cloud.google.com/bigquery/docs/reference/standard-sql/functions-and-operators#safe_casting).

As a dbt user, and as someone familiar with those SQL methods, that's what I was expecting those methods to do, too.

jeremyyeo:
> and it looks like some of the docstrings need updating.

Ah yes - thanks for catching that :)

> To my mind, the implementation of set and try_set here feels exactly backwards.

Yeah I think so too. 

However, at the same time, I'm pretty sure we were trying to align these new functions (in terms of whether `try_` "returns default upon error" or "error upon error") with the behaviour of existing context methods like `toyaml`, `tojson`:

https://github.com/dbt-labs/dbt-core/blob/3594ced4da797b2eabf54da8884c22385880cec7/core/dbt/context/base.py#L463-L465

Which has no `try_` equivalent at the moment. I suppose we could:

1. Swap `try_zip/try_set` with `_zip/_set` but then this will deviate a little bit with the other existing methods `toyaml`, etc.
2. Swap `try_zip/try_set` with `_zip/_set` AND add `try_` equivalents of the existing methods too (`try_toyaml`)?

----

> I'm also struggling to come up with a working example where zip fails (raising a TypeError)

Non-iters seem to raise TypeError in Python...

image

I think it's because:

```
{% set my_list_a = 1234 %}
{% set my_list_b = ['alice', 'bob', 'another'] %}
{% set my_zip = zip(my_list_a, my_list_b) %}  -- Here the TypeError is raised and value returned is None
{% set into_list = my_zip | list %} -- Effectively we're doing `list(None)`
```

And `list(None)` does result in `'NoneType' object is not iterable`.

image

jtcohen6:
@jeremyyeo Thanks for the quick response! And good call on the `| list` filter being the culprit here. I was able to get this example working:

```
{% set my_list_a = 1234 %}
{% set my_list_b = ['alice', 'bob', 'another'] %}
{% set my_zip = zip(my_list_a, my_list_b, default = []) | list %}
{% do log(my_zip, info = true) %} {#  #}
```

I'm catching myself back up on these threads from a few months ago, and just dropped a few thoughts in https://github.com/dbt-labs/dbt-core/issues/5167#issuecomment-1184158346. I recall agreeing at the time that:
- This would necessarily feel inconsistent with `toyaml`/`fromyaml`/`tojson`/`fromjson` (we don't love how those were implemented, too much in one method)
- At the risk of some interim inconsistency, it's worth doing the right thing going forward. Later on, we can scope and schedule work to bring the others into line.

My strong preference is to get this right going forward, by giving the right functionality to the right name. Two options for us right now (while we're in the v1.2 RC period):
1. Switch the behavior of `set` + `try_set`, `zip` + `try_zip`
2. Rename `try_*` methods to indicate the opposite behavior, e.g. `strict_`

@emmyoop @nathaniel-may open to your thoughts as well!

jtcohen6:
@nathaniel-may and I had a chance to discuss live. We're definitely lacking a larger "language design" discussion and spec which would make it easier for community members to contribute, and for us to review, these types of changes with confidence. If he's interested, I think @dbeatty10 might be able to help with this :)

We don't have the bandwidth to do that in the immediate term, but it's definitely worth doing. In the meantime, let's move forward with one of the short-term fixes identified above, to include in v1.2.0-rc2, just to avoid shipping functionality that does the exact opposite of what (I suspect) many users will expect.

emmyoop:
@jtcohen6 I'd lean toward option 2 to avoid breaking users until we've figured out the standard we want to live by.  `strict` is better than `try` 💯 .  And while `try` has meaning in Snowflake, does it for other adapters?  I don't love using `try` if it only has meaning in Snowflake.  I'd rather move away from it entirely.

jtcohen6:
@emmyoop Okay, I buy it! If we choose to adopt it as a consistent pattern, it will also be much easier to add in `strict_toyaml` et al, since it requires no change to the current behavior of `toyaml`.

@jeremyyeo I can pick this up, unless you have the motivation and capacity to see it through. We'd just need to get the change merged + backported by next Tuesday.

nathaniel-may:
minor note: would you consider `to_yaml_strict` so in an alphabetized docs page they show up next to each other?

Clarification question: If I see a function with `strict` in its name, what exactly do I expect from its behavior?

jtcohen6:
> minor note: would you consider `toyaml_strict` so in an alphabetized docs page they show up next to each other?

👍 good call! all about it

> Clarification question: If I see a function with `strict` in its name, what exactly do I expect from its behavior?

As I understand it, "strict" methods:
- should raise a standard exception if the value cannot be converted to a set / yaml / json, zipped, etc
- should not not allow for a `default` value to gracefully fail over to

Whereas the "non-strict" versions of those methods return `None` if the conversion/serialization fails (or a configurable default instead of `None`).

IIRC, we had previously discussed `strict: bool = False` as an additional optional argument of these methods, but decided it would be better to implement separate methods / variants.

jeremyyeo:
@jtcohen6 Took a crack at that https://github.com/dbt-labs/dbt-core/compare/tags/v1.2.0rc1...rename-strict-mode

Should we do `toyaml` and `tojson` at this time too?

jtcohen6:
@jeremyyeo That branch looks great! Want to open a PR for it?

I just opened a dedicated issue for this switch: https://github.com/dbt-labs/dbt-core/issues/5475

The timeline to do `fromjson`/`tojson`/`fromyaml`/`toyaml` is not as pressing. The reason to rename `try_set` and `try_zip` right now is that they've already been included in an RC, and not yet in a final release, so we only have one week to correct it before we've made a commitment to backwards compatibility. That said, we should definitely make sure the work is tracked (in either #5475 or #5167), clearly defined, and (probably) open for community contribution.

PR #5134, opened 2022-04-21
Refactor `build` test to new framework

### LINKED ISSUE #5128: [CT-538] 069_build_tests

iknox-fa:
resolves #5128 (CT-538)

### Description
Converts `build` test to new framework

### Checklist


tests/functional/build/test_build.py
```diff
@@ -0,0 +1,192 @@
...
+    seeds,
+    models_circular_relationship,
+    models_simple_blocking,
+    test_files,
+    models_interdependent,
+    project_files,
+)
+
+
+# from test.integration.base import DBTIntegrationTest, use_profile, normalize
```
stu-k, line 20: Can we remove this comment?

stu-k (approved):
Nothing stands out, all previous models and snapshots accounted for 👍

tests/functional/build/test_build.py
```diff
@@ -0,0 +1,192 @@
...
+        return {
+            "model_3.sql": models_failing["model_3.sql"],
+            "model_2.sql": models_failing["model_2.sql"],
+            "test.yml": models_failing["test.yml"],
+            "model_0.sql": models_failing["model_0.sql"],
+            "model_1.sql": models_failing["model_1.sql"],
+            "model_99.sql": models_failing["model_99.sql"],
+        }
+
+    def test__build_sad_path(
```
emmyoop, line 72: All the test names should be single underscore ie `test_build_sad_path` to be consistent with all other tests.

tests/functional/build/test_build.py
```diff
@@ -0,0 +1,192 @@
...
+)
+
+
+# from test.integration.base import DBTIntegrationTest, use_profile, normalize
+from test.integration.base import normalize
+import yaml
+import os
+
+
+class TestBuildBase:
```
emmyoop, line 26: Only the classes that contain tests should start with `Test`.  Maybe just name this `BuildBase`?

tests/functional/build/test_build.py
```diff
@@ -0,0 +1,192 @@
...
+    @pytest.fixture(scope="class")
+    def project_config_update(self):
+        return {
+            "config-version": 2,
+            "snapshot-paths": ["snapshots"],
+            "seed-paths": ["seeds"],
+            "seeds": {
+                "quote_columns": False,
+            },
+        }
```
emmyoop, line 36: This isn't necessary unless you need to change from the default config.  Since we no longer use the complex mult-folder structure to store files for testing, most of the time this doesn't need to be touched at all.  Since the default for quote_columns is False, you can remove this entirely.  This is true in all other tests that I can tell.  Some of them are pointing to use non-default directories but I don't think it's strictly necessary or what we're actually testing here.

tests/functional/build/test_build.py
```diff
@@ -0,0 +1,192 @@
...
+    def test__interdependent_models(self, project, test_data_dir):
+        # check that basic build works
+        copy_file(test_data_dir, "model_b.sql", project.project_root + "/models", ["model_b.sql"])
+        results = self.build()
+        assert len(results) == 16
+
+        # return null from model_b
+        copy_file(
+            test_data_dir, "model_b_null.sql", project.project_root + "/models", ["model_b.sql"]
+        )
```
emmyoop, line 187: Why does this file and the `model_b.sql` file exist in the `/data` and in `fixtures`?

emmyoop (changes requested):

tests/functional/build/fixtures.py
```diff
@@ -0,0 +1,474 @@
...
+select * from {{ ref('model_b') }}
+
+"""
+
+models_interdependent__model_a_sql = """
+select 1 as id
+
+"""
+
+
```
emmyoop, line 368: Everything below this comment should be defined in the test file.  It makes it much more clear what's happening in the tests that way.  The project fixture takes care of writing all the files to the project.

emmyoop: I've been putting the actual model strings into the test files most of the time.  I tend to only have strings in this file that are extremely long or shared across multiple files.  If it makes sense to do so i also split up the single giant test file into multiple files to make it easier to find and organize tests.  That doesn't always make sense though, it's a different call for each set of tests, not a rule.

tests/functional/build/test_build.py
iknox-fa: Whoops.  Good catch

tests/functional/build/test_build.py
iknox-fa: Sure.  @ChenyuLInx this came from the conversion script so it's probably worth updating.

tests/functional/build/test_build.py
iknox-fa: Sure thing.  @ChenyuLInx , same as above.

tests/functional/build/test_build.py
iknox-fa: Because I forgot to remove them from fixtures :)
updated

tests/functional/build/test_build.py
iknox-fa: Sure, another conversion script thing here @ChenyuLInx

tests/functional/build/fixtures.py
iknox-fa: > Everything below this comment should be defined in the test file. It makes it much more clear what's happening in the tests that way. 

I'm not sure I agree with that assessment.  IMO it's easier to look at the test logic in one place (the test file) and the underlying project data in another file (the fixtures file).  That said, I don't have strong feelings here so I'm happy to move it if it's more consistent.

tests/functional/build/test_build.py
ChenyuLInx: We could try to find the class that doesn't have a test function and change its name but feels like this might be tricky and going over the tests before merging could catch it easily.

tests/functional/build/test_build.py
ChenyuLInx: We could add a process function [here](https://github.com/ChenyuLInx/unittest2pytest/blob/dbt-pytest-all-dir/unittest2pytest/fixes/fix_functions.py#L130) to update that node? But manipulating and checking the dict structure with the framework is not super easy so I vote for scan through case by case, I will document this in the conversion guide.

tests/functional/build/test_build.py
iknox-fa: @ChenyuLInx Yeah that makes sense, maybe just add it as a thing to check for in the docs?  /shrug

tests/functional/build/test_build.py
```diff
@@ -0,0 +1,296 @@
...
+    models_interdependent,
+):
+    write_project_files(project_root, "snapshots", snapshots)
+    write_project_files(project_root, "tests-failing", tests_failing)
+    write_project_files(project_root, "models", models)
+    write_project_files(project_root, "models-failing", models_failing)
+    write_project_files(project_root, "seeds", seeds)
+    write_project_files(project_root, "models-circular-relationship", models_circular_relationship)
+    write_project_files(project_root, "models-simple-blocking", models_simple_blocking)
+    write_project_files(project_root, "models-interdependent", models_interdependent)
```
emmyoop, line 138: None of this is needed.  Just define the models/snapshots/etc in the class.  The project fixture takes care of the `write_project_files()` function.  I'll add an example below of what I mean.

tests/functional/build/test_build.py
```diff
@@ -0,0 +1,296 @@
...
+    @pytest.fixture(scope="class")
+    def models(self, models_failing):  # noqa: F811
+        return {
+            "model_3.sql": models_failing["model_3.sql"],
+            "model_2.sql": models_failing["model_2.sql"],
+            "test.yml": models_failing["test.yml"],
+            "model_0.sql": models_failing["model_0.sql"],
+            "model_1.sql": models_failing["model_1.sql"],
+            "model_99.sql": models_failing["model_99.sql"],
+        }
```
emmyoop, line 174: Here's the example of how to define the models without everything at the top of this file.

```
def models(self):
    return {
        "model_3.sql": models_failing__model_3_sql,
        "model_2.sql": models_failing__model_2_sql,
        "test.yml": models_failing__test_yml,
        "model_0.sql": models_failing__model_0_sql,
        "model_1.sql": models_failing__model_1_sql,
        "model_99.sql": models_failing__model_99_sql,
    }
```

emmyoop (changes requested):
Just a bit more cleanup to reduce repetition.

tests/functional/build/test_build.py
ChenyuLInx: the reason we were having different folders for different model was that there could be only one set of files for the whole tests so we put everything in different folders and use `model_path` to point to it. In the new test framework, we write a project repo to disk for every test class, so you can define the `models` fixture that only containes files needed for current class, and not need the `model_path` any more.
https://github.com/dbt-labs/dbt-core/blob/ab0c3510ebe70a67fb669b88e6ea36aff2f46339/tests/functional/schema_tests/test_schema_v2_tests.py#L41-L43

ChenyuLInx:
Agreed with Emily's comment, just some clean up then this is good to go!

ChenyuLInx (approved):
@iknox-fa Looks like one conflict to resolve otherwise looks great to me!

tests/functional/artifacts/data/state/v6/manifest.json
```diff
@@ -1 +1 @@
...u can click the blue icon on the bottom-right corner of the page to view the lineage graph of your models.\n\nOn model pages, you'll see the immediate parents and children of the model you're exploring. By clicking the `Expand`\nbutton at the top-right of this lineage pane, you'll be able to see all of the models that are used to build,\nor are built from, the model you're exploring.\n\nOnce expanded, you'll be able to use the `--select` and `--exclude` model selection syntax to filter the\nmodels in the graph. For more information on model selection, check out the [dbt docs](https://docs.getdbt.com/docs/model-selection-syntax).\n\nNote that you can also right-click on models to interactively filter and explore the graph.\n\n---\n\n### More information\n\n- [What is dbt](https://docs.getdbt.com/docs/introduction)?\n- Read the [dbt viewpoint](https://docs.getdbt.com/docs/viewpoint)\n- [Installation](https://docs.getdbt.com/docs/installation)\n- Join the [dbt Community](https://www.getdbt.com/community/) for questions and discussion"}}, "exposures": {}, "metrics": {}, "selectors": {}, "disabled": {}, "parent_map": {"model.test.my_model": []}, "child_map": {"model.test.my_model": []}}
```
emmyoop, line 1: Why did this file get modified?

tests/functional/build/fixtures.py
emmyoop: I hear you.  I'm going to start a thread so we can decide on a standard here as a team.

emmyoop (changes requested):
I don't think the new manifest should be here?

We also have a pending conversation in our slack about how to approach pulling in files.

tests/functional/artifacts/data/state/v6/manifest.json
iknox-fa: Because of [this bug](https://github.com/dbt-labs/dbt-core/issues/5584), which I erroneously assumed was a feature and committed the changed file.
Fixed now.

jtcohen6:
@emmyoop @iknox-fa Is the only remaining blocker for this PR a question about how fixture data should be formatted / located in functional tests?

tests/functional/build/fixtures.py
```diff
@@ -0,0 +1,430 @@
...
+    }
+
+
+@pytest.fixture(scope="class")
+def models_interdependent():
+    return {
+        "schema.yml": models_interdependent__schema_yml,
+        "model_c.sql": models_interdependent__model_c_sql,
+        "model_a.sql": models_interdependent__model_a_sql,
+    }
```
emmyoop, line 430: You should completely remove all of this, then import the specific strings you're using instead of the fixtures in the test file.

I was not in the team meeting but my understanding was if there are more than 50 lines to define, we define them in this file but _just_ the strings.  It is the pattern I've seen with most other conversions we have done.

tests/functional/build/test_build.py
emmyoop: When you import the strings this would be how you define the models fixture and can then remove the `# noqa`.

emmyoop (changes requested):
There's a bit more missing with the standardization around how we format and import.

Issue #4887, opened 2022-03-17
[CT-378] Create tests for vars in packages.yml and selectors.yml


gshank:
We fixed a bug in $4878 with the use of vars in package.yml and selectors.yml. For ease of backporting and because we're in the middle of a test overhaul, new tests were not added in that ticket/pull request. This ticket is to add additional tests.

gshank:
Also convert the tests.

PR #5136, opened 2022-04-22
Use yaml renderer (with target context) for rendering selectors

### LINKED ISSUE #5131: [CT-540] [Bug] 'target' suddenly not available at compile time in selectors.yml

### Current Behavior

After upgrading to `dbt v1.0.5`, I get the following error when compiling my model:
```
Compilation Error
  Could not render {{ target.name == 'dev' | as_bool }}: 'target' is undefined
```

### Expected Behavior

The project should compile without error, as it did in earlier post-v1 releases.

### Steps To Reproduce

1. Add a selector like this to `selectors.yml`
```yml
selectors:
  - name: dev_defer_snapshots
    definition:
      method: fqn
      value: '*'
      exclude:
        - method: config.materialized
          value: snapshot
```
2. Compile the project with `dbt v1.0.5`
3. Note the error
4. Comment out the line containing `default: "{{ target.name == 'dev' | as_bool }}"`
5. Compile the project
6. Note no error

### Relevant log output

_No response_

### Environment

```markdown
- OS: MacOS 12.3.1
- Python: 3.9.10
- dbt: 1.0.5
```

redshift

### Additional Context

_No response_

gshank:
resolves #5131

### Description

When we set up "secret" contexts for profiles.yml and packages.yml, we also made the context for rendering selectors.yml into a secret context (which does not include the target), when it should have stayed as a target context.

### Checklist


gshank:
Note: I haven't created tests yet, but wanted to get this out there for inspection and running tests.

gshank:
Note: this will be backported, but without the test commits.

core/dbt/config/renderer.py
```diff
@@ -114,12 +114,10 @@ def __init__(
     def name(self):
         "Project config"
 
+    # Uses SecretRenderer
```
jtcohen6, line 117: 👍

core/dbt/config/utils.py
```diff
@@ -62,6 +62,8 @@ def get_project_config(
             user_config = read_user_config(flags.PROFILES_DIR)
         # Update flags
         flags.set_from_args(args, user_config)
+        if cli_vars is None:
+            cli_vars = {}
```
jtcohen6, line 66: Is this just moving the code from `SecretRenderer` here?

This is the only part that scares me, insofar as it touches code that's more widely used than just `selectors.yml`. It's hard to imagine how this could cause issues, though.

core/dbt/config/selectors.py
```diff
@@ -3,7 +3,7 @@
 from dbt.clients.yaml_helper import yaml, Loader, Dumper, load_yaml_text  # noqa: F401
 from dbt.dataclass_schema import ValidationError
 
-from .renderer import SelectorRenderer
+from .renderer import BaseRenderer
```
jtcohen6, line 6: Makes sense to just use `BaseRenderer`, since the context available to `selectors.yml` is the same as every other yaml files

jtcohen6 (commented):
This PR feels like a lot of changes for something we ultimately need to backport to both `1.1.latest` + `1.0.latest`. After looking through it, most of the changes are:
- Converting relevant test cases + adding new ones — resolving https://github.com/dbt-labs/dbt-core/issues/4887, as well as securing the regression case
- Relevant only to rendering in `selectors.yml`, which significantly limits the "blast radius" of this change

So, after looking through it, and a bit of local testing, I'm much more comfortable with the change. I'll leave final code review to the Language team. We should aim to merge this ASAP so we can backport and include it in v1.0.6-rc1.

gshank:
Last time I made the separate ticket to create tests and it didn't actually happen. I really didn't want to leave this without tests this time, so I included them (the conversion was ready to go).  I'm going to cherry-pick the non-test commit for backports.

core/dbt/config/utils.py
jtcohen6: Got it! Thanks for the explanation

emmyoop (approved):
Looks good.  Agree with @jtcohen6 that this seemed like a bigger change than it is because of the tests.

PR #5139, opened 2022-04-24
migrate 009_data_tests to new test framework

VersusFacit:
### Description

Migrates 009_data_tests over to the new framework. I've also taken the opportunity to rename the test to match updated language. The old language has been deprecated in favor of easier to understand terms (generic, singular, etc.).

This test currently exists in snowflake, but there's no functional difference between its version and the version here in core. I think that should be removed and this test should exist as a core-only test.

For matching snowflake PR which merely needs a simple OK if we agree on its removal, [see here](https://github.com/dbt-labs/dbt-snowflake/pull/144).

### Checklist


tests/functional/custom_singular_tests/test_custom_singular_tests.py
```diff
@@ -0,0 +1,130 @@
...
+        args = FakeArgs()
+
+        kwargs = {
+            "profile": None,
+            "profiles_dir": profiles_dir,
+            "target": None,
+        }
+
+        config = RuntimeConfig.from_args(ArgsForTest(kwargs))
+        test_task = TestTask(args, config)
```
gshank, line 108: It's not clear to me why we're doing test_task.run here instead of just running tests in the usual way. If you've figured that out in the process of converting these tests, could you add a comment?

gshank (approved):
Looks good. Would like a couple of lines of comments if you can.

tests/functional/custom_singular_tests/test_custom_singular_tests.py
VersusFacit: Good callout. There was a reason that no longer applies. In my early stages of converting tests, dbt was not invoking all 5 tests, only 2. Any nodes with a `ref` to the `table_copy` model were skipping. I'm not exactly sure why in retrospect -- my guess is that I elided over the table_copy model. At the time, however, I was thinking a failure during test run might be causing an early abort. Hence, I thought you needed the TestTask to ensure all 5 were invoked even though two would fail. This of course was all incorrect!

I have since verified that a simple `run_dbt(['test'])` will cover all 5 tests and streamlined the code to reflect better code practices.

PR #5146, opened 2022-04-25
Fix: add warning on duplicated yaml keys

### LINKED ISSUE #5114: [CT-520] [Enhancement] Schema files with repeated top level keys (e.g. `models`) results in only the last key being parsed

### Current Behavior

If you have a schema file with repeated top level keys:

```yml
# models/schema.yml
version: 2
models:
  - name: my_model_a
    description: "This is a"
    columns:
      - name: user_name
        tests:
          - not_null
models:
  - name: my_model_b
    description: "This is b"
    columns:
      - name: user_name
        tests:
          - not_null
```

Only the last `models` key will apply to the project when you execute dbt tasks that depend on them (like `test`, `docs generate`). We should raise an exception for such cases.

### Expected Behavior

Perhaps we should raise an error in this case instead of silently just including the last `models` key of the schema file in our project's manifest.

### Steps To Reproduce

1. Add the above `schema.yml` file to your project along with 2 simple models:

```sql
-- models/my_model_a.sql
select 1 as user_name
-- models/my_model_b.sql
select 1 as user_name
```

2. Do a `dbt run` to build our models.
3. Do a `dbt test` and observe that we only tested `my_model_b`.
4. Do a `dbt docs generate && dbt docs serve` and observe that only `my_model_b` has a description (`my_model_a` has no description).

### Relevant log output

```shell
$ dbt test
04:02:08  Running with dbt=1.0.4
04:02:09  Found 2 models, 1 test, 0 snapshots, 0 analyses, 165 macros, 0 operations, 0 seed files, 0 sources, 0 exposures, 0 metrics
04:02:09  
04:02:09  Concurrency: 1 threads (target='dev')
04:02:09  
04:02:09  1 of 1 START test not_null_my_model_b_user_name................................. [RUN]
04:02:09  1 of 1 PASS not_null_my_model_b_user_name....................................... [PASS in 0.07s]
04:02:09  
04:02:09  Finished running 1 test in 0.24s.
04:02:09  
04:02:09  Completed successfully
04:02:09  
04:02:09  Done. PASS=1 WARN=0 ERROR=0 SKIP=0 TOTAL=1
```

### Environment

```markdown
- OS: macOS
- Python: 3.9.7
- dbt: 1.0.4
```

postgres

### Additional Context

image

jeremyyeo:
resolves #5114 

### Description

When we have a duplicate top level key in our schema yml files, we should see the following warning:

```
$ dbt run
08:01:05  Running with dbt=1.2.0-a1
08:01:05  [WARNING]: Compilation Error
  Duplicate 'models' top level key found in yaml file models/schema.yml.
08:01:05  [WARNING]: Compilation Error
  Duplicate 'models' top level key found in yaml file models/sub/schema.yml.
08:01:05  Found 2 models, 0 tests, 0 snapshots, 0 analyses, 167 macros, 0 operations, 0 seed files, 0 sources, 0 exposures, 0 metrics
08:01:05  
08:01:06  Concurrency: 1 threads (target='dev')
08:01:06  
08:01:06  1 of 2 START table model public.my_model_a ..................................... [RUN]
08:01:06  1 of 2 OK created table model public.my_model_a ................................ [SELECT 1 in 0.10s]
```

----

1. In `core/dbt/exceptions.py`, `DuplicateYamlKeyException` is a no-op of `CompilationException` - is this okay?

2. For the tests, because we warn instead of error, I'm using `run_dbt_and_capture()` from `dbt.tests.utils` and asserting that stdout has the right string - not too sure if this is the right way to do this. I also tried something along the lines of:

```python
from dbt.exceptions import DuplicateYamlKeyException

...

def test_duplicate_key_in_yaml(project):
    with pytest.raises(DuplicateYamlKeyException):
        run_dbt(["run"])
```

Which didn't quite work because we print a warning instead of raising an error. (Ftr, I just cribbed what was in `dbt/tests/functional/basic/tests_basic.py` for this one).

### Checklist

### DIFF
--- a/core/dbt/clients/yaml_helper.py
+++ b/core/dbt/clients/yaml_helper.py
@@ -22,6 +22,13 @@ Raw Error:
 
 
 class UniqueKeyLoader(SafeLoader):
+    """A subclass that checks for unique yaml mapping nodes.
+
+    This class extends `SafeLoader` from the `yaml` library to check for
+    unique top level keys (mapping nodes). See issue (https://github.com/yaml/pyyaml/issues/165)
+    and solution (https://gist.github.com/pypt/94d747fe5180851196eb?permalink_comment_id=4015118).
+    """
+
     def construct_mapping(self, node, deep=False):
         mapping = set()
         for key_node, value_node in node.value:
@@ -77,5 +84,5 @@ def load_yaml_text(contents, path=None):
         raise dbt.exceptions.ValidationException(error)
     except dbt.exceptions.DuplicateYamlKeyException as e:
         # TODO: We may want to raise an exception instead of a warning in the future.
-        msg = f"{e} {path.searched_path}/{path.relative_path}."
-        dbt.exceptions.warn_or_error(msg, log_fmt=warning_tag("{}"))
+        e.msg = f"{e} {path.searched_path}/{path.relative_path}."
+        dbt.exceptions.warn_or_raise(e, log_fmt=warning_tag("{}"))
--- a/core/dbt/clients/yaml_helper.py
+++ b/core/dbt/clients/yaml_helper.py
@@ -8,6 +8,7 @@ try:
 except ImportError:
     from yaml import Loader, SafeLoader, Dumper  # type: ignore  # noqa: F401
 
+from dbt.ui import warning_tag
 
 YAML_ERROR_MESSAGE = """
 Syntax error near line {line_number}
@@ -18,6 +19,19 @@ Raw Error:
 ------------------------------
 {raw_error}
 """.strip()
+
+
+class UniqueKeyLoader(SafeLoader):
+    def construct_mapping(self, node, deep=False):
+        mapping = set()
+        for key_node, value_node in node.value:
+            key = self.construct_object(key_node, deep=deep)
+            if key in mapping:
+                raise dbt.exceptions.DuplicateYamlKeyException(
+                    f"Duplicate {key!r} key found in yaml file"
+                )
+            mapping.add(key)
+        return super().construct_mapping(node, deep)
 
 
 def line_no(i, line, width=3):
@@ -48,10 +62,10 @@ def contextualized_yaml_error(raw_contents, error):
 
 
 def safe_load(contents) -> Optional[Dict[str, Any]]:
-    return yaml.load(contents, Loader=SafeLoader)
+    return yaml.load(contents, Loader=UniqueKeyLoader)
 
 
-def load_yaml_text(contents):
+def load_yaml_text(contents, path=None):
     try:
         return safe_load(contents)
     except (yaml.scanner.ScannerError, yaml.YAMLError) as e:
@@ -61,3 +75,7 @@ def load_yaml_text(contents):
             error = str(e)
 
         raise dbt.exceptions.ValidationException(error)
+    except dbt.exceptions.DuplicateYamlKeyException as e:
+        # TODO: We may want to raise an exception instead of a warning in the future.
+        msg = f"{e} {path.searched_path}/{path.relative_path}."
+        dbt.exceptions.warn_or_error(msg, log_fmt=warning_tag("{}"))
--- a/test/integration/006_simple_dependency_tests/local_dependency/models/schema.yml
+++ b/test/integration/006_simple_dependency_tests/local_dependency/models/schema.yml
@@ -4,7 +4,6 @@ sources:
     schema: invalid_schema
     tables:
       - name: my_table
-sources:
   - name: seed_source
     schema: "{{ var('schema_override', target.schema) }}"
     tables:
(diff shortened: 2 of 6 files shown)


jeremyyeo:
Integration test from days of yonder fails led me to a yaml test file (https://github.com/dbt-labs/dbt-core/blob/main/test/integration/006_simple_dependency_tests/local_dependency/models/schema.yml) with actual duplicated top level keys - it looks like the "mapping node should be a set" [^1] so curious if that particular test worked properly before 🤔

[^1]: https://stackoverflow.com/questions/47668308/duplicate-key-in-yaml-configuaration-file

tests/functional/duplications/test_basic_duplications.py
```diff
@@ -0,0 +1,27 @@
...
+def models():
+    return {
+        "my_model.sql": my_model_sql,
+        "schema.yml": duplicate_key_schema__schema_yml,
+    }
+
+
+def test_duplicate_key_in_yaml(project):
+    results, stdout = run_dbt_and_capture(["run"])
+    assert "Duplicate 'models' key found in yaml file models/schema.yml" in stdout
```
emmyoop, line 27: You can add an additional test that sends the `--warn-error` flag through and have to catch.  Something like this:

```
def test_duplicate_key_in_yaml_error(project):
    with pytest.raises(DuplicateYamlKeyException):
        run_dbt(["--warn-error", "run"])
```

core/dbt/clients/yaml_helper.py
```diff
@@ -20,6 +21,19 @@
 """.strip()
 
 
+class UniqueKeyLoader(SafeLoader):
```
emmyoop, line 24: Can you link to where you found this solution?  It gives some great contect as to why we're doing this which will be 100% helpful down the road when we ask "but why".

core/dbt/clients/yaml_helper.py
```diff
@@ -61,3 +75,7 @@ def load_yaml_text(contents):
             error = str(e)
 
         raise dbt.exceptions.ValidationException(error)
+    except dbt.exceptions.DuplicateYamlKeyException as e:
+        # TODO: We may want to raise an exception instead of a warning in the future.
+        msg = f"{e} {path.searched_path}/{path.relative_path}."
+        dbt.exceptions.warn_or_error(msg, log_fmt=warning_tag("{}"))
```
emmyoop, line 81: You should use `warn_or_raise` here instead of `warn_or_error`.  That way the exception you throw above bubbles up instead of getting overlaid by the `CompilationException` by `warn_or_error`.  To get the message passed through as `DuplicateYamlKeyException` you can modify the message.  
```
e.msg = f"{e} {path.searched_path}/{path.relative_path}."
dbt.exceptions.warn_or_raise(e, log_fmt=warning_tag("{}"))
```

emmyoop (changes requested):
I had a few comments about structure that I wanted to get back to you on.  I'll take a look at the failing legacy tests and see what's happening there next.

tests/functional/duplications/test_basic_duplications.py
jeremyyeo: Thanks @emmyoop - forgot about that flag :)

core/dbt/clients/yaml_helper.py
jeremyyeo: With the new changes I see:

```
$ dbt run                                 
10:59:42  Running with dbt=1.2.0-a1
10:59:42  [WARNING]: Compilation Error
  Compilation Error
    Duplicate 'models' key found in yaml file models/schema.yml.
10:59:43  Found 2 models, 0 tests, 0 snapshots, 0 analyses, 167 macros, 0 operations, 0 seed files, 0 sources, 0 exposures, 0 metrics
10:59:43  
10:59:43  Concurrency: 1 threads (target='dev')
...
```

Wondering if that was what you expected here?

test/integration/006_simple_dependency_tests/local_dependency/models/schema.yml
```diff
@@ -4,7 +4,6 @@ sources:
     schema: invalid_schema
     tables:
       - name: my_table
-sources:
```
jeremyyeo, line 7: @emmyoop surprisingly fixing the duplicate top level keys in this yaml file in the legacy tests causes it to pass so now really curious what this was doing previously... makes me wonder if we had left 2 `sources` keys in this file by mistake the last time around but test was passing due to pyaml default of only reading the last `source` key of this file.

emmyoop (approved):

test/integration/006_simple_dependency_tests/local_dependency/models/schema.yml
emmyoop: I'm not seeing anything in the tests that indicate we did it on purpose so you may be right.  These tests are going to be rewritten and we'll work through some of the weirdness in them as we go.

PR #5171, opened 2022-04-26
Update mypy to latest and turn on everywhere

iknox-fa:
resolves #

### Checklist


core/dbt/adapters/factory.py
```diff
@@ -140,8 +140,6 @@ def get_adapter_plugins(self, name: Optional[str]) -> List[AdapterPlugin]:
                 raise InternalException(f"No plugin found for {plugin_name}") from None
             plugins.append(plugin)
             seen.add(plugin_name)
-            if plugin.dependencies is None:
```
iknox-fa, line 143: @VersusFacit This works right?  From what [I can see here](https://github.com/dbt-labs/dbt-core/blob/4e57c51c7abc5dd0134e14a3df27842c5fd57b65/core/dbt/adapters/base/plugin.py#L39) `plugin.dependencies` will never be `None`.

core/dbt/graph/selector_methods.py
```diff
@@ -540,7 +540,7 @@ def search(self, included_nodes: Set[UniqueId], selector: str) -> Iterator[Uniqu
             )
 
         current_state_sources = {
-            result.unique_id: getattr(result, "max_loaded_at", None)
```
iknox-fa, line 543: @sungchun12 This should work right?  Mypy was complaining that line 571 is potentially an invalid statement given you can't do greater-than / less-than comparisons on `None`

.changes/unreleased/Under the Hood-20220427-112127.yaml
```diff
@@ -0,0 +1,7 @@
+kind: Under the Hood
+body: Mypy -> 0.942 + fixed import logic to allow for full mypy coverage
+time: 2022-04-27T11:21:27.499359-05:00
+custom:
+  Author: iknox-fa
+  Issue: "4805, 4089"
```
iknox-fa, line 6: @emmyoop Can I put two issues in this field?  This PR does legit address them both but I don't know if changie knows how to handle it...

core/scripts/upgrade_dbt_schema_tests_v1_to_v2.py
```diff
@@ -395,7 +396,7 @@ class CustomSortedColumnsSchema(CustomSortedSchema):
 
 
 for cls in (CustomSortedRootSchema, CustomSortedModelsSchema, CustomSortedColumnsSchema):
-    yaml.add_representer(cls, cls.representer)
+    yaml.add_representer(cls, cls.representer)  # type: ignore [attr-defined]
```
iknox-fa, line 399: @nathaniel-may You might have a solution to this one.. but for the life of me I couldn't sort out why mypy thinks `cls.representer` was undefined.  It's clearly in all three classes above.

.changes/unreleased/Under the Hood-20220427-112127.yaml
nathaniel-may: if not you can always do two change log entries. that's been done before.

core/scripts/upgrade_dbt_schema_tests_v1_to_v2.py
iknox-fa: So why does it find `ABCMeta` as the common type bound? Shouldn't it be `CustomSortedSchema`?  All items in the tuple are subclasses of it...

Either way, I'll just leave it ass an ignore for now.

core/dbt/graph/selector_methods.py
sungchun12: Yes this should work. As a fail-safe, I force a condition for it to have the attribute in the first place on line 545 for example

```python
if hasattr(result, "max_loaded_at")
        }
```

core/scripts/upgrade_dbt_schema_tests_v1_to_v2.py
nathaniel-may: oops I'm totally wrong here. The tuple is of type `Tuple3[Type[CSRS], Type[CSMS], Type[CSCS]]` since those are all classes not class instances.

I'd expect the following to type check, but it doesn't.
```python
from typing import Iterator, Type, List
boop: Iterator[Type[CustomSortedSchema]] = (CustomSortedRootSchema, CustomSortedModelsSchema, CustomSortedColumnsSchema).__iter__()
```
```
error: Incompatible types in assignment (expression has type "Iterator[ABCMeta]", variable has type "Iterator[Type[CustomSortedSchema]]")
```

I'm actually not sure why it's not unifying those types correctly.

In any case doing it on 3 separate lines will fix it since the problem is definitely with the conversion to an iterator.

.changes/unreleased/Under the Hood-20220427-112127.yaml
emmyoop: In the current state you cannot link multiple issues.  If you see this as valuable, we can definitely add it though.  We control how the fields end up being generated.  I recently made an update to allow multiple contributor names.

core/scripts/upgrade_dbt_schema_tests_v1_to_v2.py
nathaniel-may: Tracked it down to the point where it works fine if the parent class doesn't inherit from `dict` it works. I'll probably dig farther than this warrants for my own understanding.

gshank: Do we even need this script? It's pretty old and we've been on v2 of schema files for years. Maybe it should just be removed?

jtcohen6: yeah, let's :knife: it

the link to this script is pointing to the long-unused `development` branch anyway: https://discourse.getdbt.com/t/migrating-to-schema-yml-v2/111

.github/workflows/main.yml
```diff
@@ -52,9 +52,10 @@ jobs:
           pip --version
           pip install pre-commit
           pre-commit --version
-          pip install mypy==0.782
+          pip install mypy==0.942
           mypy --version
           pip install -r editable-requirements.txt
+          pip install -r dev-requirements.txt
```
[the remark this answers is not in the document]
kwigley: I think we want to use `requirements.txt` here
```
pip install -r requirements.txt
pip install -r dev-requirements.txt
```
requirement.txt -> installs dbt-core and dbt-postgres (essentially same as `pip install dbt-core dbt-postgres`)
editable-requirement.txt -> installs dbt-core and dbt-postgres in editable mode (https://pip.pypa.io/en/stable/cli/pip_install/#editable-installs)
dev-requirements.txt -> all the development dependencies

My understanding is that mypy needs all dbt-core and dbt-postgres depenedencies installed in order to perform type checking. Just installing `dev-requirements.txt` doesn't install dbt-core and dbt-postgres

stu-k (approved):
Nothing stands out 👍

.github/workflows/main.yml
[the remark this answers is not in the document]
iknox-fa: ...annd updated.

sungchun12 (approved):

leahwicz (approved):

PR #5172, opened 2022-04-26
Add percentage support for error_if, warn_if

### LINKED ISSUE #4723: [Feature] Allow tests to warn/fail based on percentage

### Describe the Feature

Add ability to support warn/fail threshold being a percentage instead of just fixed number. We will likely need to add the definition of the total number to calculate percentage with. Original request #4334
```models:
  - name: my_view
    columns:
      - name: id
        tests:
          - unique:
              config:
                warn_if: "> 10%"
                fail_if: "> 20%"
```

_No response_

### Who will this benefit?

_No response_

_No response_

### Anything else?

_No response_

jaypeedevlin:
resolves #4723, in collaboration with @ehmartens

### Description

Adds support for using percentages in the `error_if` and `warn_if` configurations.

We'd love some feedback on the following:

* Whether the method is too 'hacky' and/or if there are alternate ways of going about this.
* Where to add tests — we took a look but couldn't find exactly where the configs are currently tested, and additionally noted that the tests are currently moving between two frameworks

### Checklist


core/dbt/include/global_project/macros/materializations/tests/helpers.sql
```diff
@@ -3,12 +3,27 @@
 {%- endmacro %}
 
 {% macro default__get_test_sql(main_sql, fail_calc, warn_if, error_if, limit) -%}
+    {% set model_name = model.file_key_name.split('.')[1] %}
```
jtcohen6, line 6: Two thoughts:
- For tests defined on models, this should work. For tests defined on sources, `file_key_name` is `sources.source_name` (and missing the name of the specific source table), so I don't think it would work as currently constituted. We could update `file_key_name` to include `sources.source_name.source_table`.
- If users have defined a `where` config, and we're calculating `%`, I think we would want to filter both the numerator and denominator by that `where` config. You could reproduce that logic here (`config.get("where")`), but all it all it seems preferable if there were some where to access the rendered version of `{{ model }}` (from the generic test query) or `{{ test_metadata.kwargs.model }}` (from the test node), since that's really the thing we want.

core/dbt/include/global_project/macros/materializations/tests/helpers.sql
```diff
@@ -3,12 +3,27 @@
...
+      {{ fail_calc }} {{ error_pct_division }} {{ error_if }} as should_error
     from (
       {{ main_sql }}
       {{ "limit " ~ limit if limit != none }}
     ) dbt_internal_test
 {%- endmacro %}
+
+
+{% macro get_pct_division(threshold, model_name) %}
+  {% if threshold[-1] == '%' %}
```
jtcohen6, line 22: Could you imagine someone wanting to define a `%` failure threshold for a singular test (as opposed to generic tests)? I can't think of one myself

jtcohen6 (commented):
Thanks for taking a first crack at this @jaypeedevlin @ehmartens!

I'm going to tag in the `Language` team to help here, since I think the real question is actually around the metadata that generic tests can make available within their materializations. Ideally, we'd have access to the rendered form of `"{{ get_where_subquery(<ref_or_source>) }}"`, i.e. that which gets passed in as the `{{ model }}` argument to generic test definitions.

core/dbt/include/global_project/macros/materializations/tests/helpers.sql
jaypeedevlin: We took a list at that this morning, I think I probably could think of someone wanting to do this (in a rare case), but it wasn't clear to me why the current approach wouldn't work (given that thresholds should work in the same way).  Could you help me understand how that might work?

core/dbt/include/global_project/macros/materializations/tests/helpers.sql
jaypeedevlin: Great point on the 'where' @jtcohen6, we totally overlooked that!  We've added that in, but not pushed the change yet, while we wait to hear from the language team on how we might approach getting the non-model resources appropriately.

gshank:
I played with some code in #5214. It doesn't handle the rendering bit yet, but thought I'd get feedback on whether this is what we're thinking of.

jaypeedevlin:
Hey @gshank, are you able to provide any guidance on how to test your changes in #5214 in conjunction with our current branch?

gshank:
I took another look at this, and we're creating the same strings in the generic_test_builder, and rendering them in 'add_rendered_test_kwargs'. Then at compilation time they are added to the context with the top-level key "_dbt_generic_test_kwargs", and the key 'model'. So I'm wondering if we can use that instead making a new function.

That rendered 'model' string is used by 'build_raw_sql' in the generic_test_builder which builds the test sql.

jtcohen6:
@gshank Really good point! This "just works" for me if I start storing `_dbt_generic_test_kwargs` on the test node (which seems easier than trying to pass it through the different Jinja contexts).

https://github.com/dbt-labs/dbt-core/commit/f0dc98a7802f7fc3a8c86bd98a7f069db2164faf

Granted, I'm taking a pretty simple approach here, and just overwriting the raw `test_metadata.kwargs` with the compiled versions. Is there a reason we shouldn't do that? Doesn't it feel... better? We're not doing anything with the raw versions, not even using it for the [`same_contents` comparison method](https://github.com/dbt-labs/dbt-core/blob/7532420eef589c241d6da6bdea9de5e2204e95c8/core/dbt/contracts/graph/compiled.py#L133-L137) that powers `state:modified`.

```yml
version: 2
models:
  - name: my_model
    columns:
      - name: id
        tests:
          - unique:
              warn_if: "=5%"
          - not_null:
              config:
                where: 1=1
                error_if: ">5%"
```

At runtime (using BigQuery):
```
  select
      count(*) as failures,
      count(*)  != 0 as should_warn,
      count(*) / (select count(*) from (select * from `dbt-test-env`.`dbt_jcohen`.`my_model` where 1=1) dbt_subquery) * 100 >5 as should_error
...
    select
      count(*) as failures,
      count(*) / (select count(*) from `dbt-test-env`.`dbt_jcohen`.`my_model`) * 100 =5 as should_warn,
      count(*)  != 0 as should_error
```

core/dbt/include/global_project/macros/materializations/tests/helpers.sql
jtcohen6: In the case where a singular test depends on multiple nodes, I'm not sure how we can tell which node should represent the denominator?

I also think, in a singular test, you're writing all the SQL yourself, it's pretty trivial to add a `/ (select count(*) from total * 100` and define `warn_if`/`error_if` with percents-as-integers.

jaypeedevlin:
I'm going to close this PR for now, since we're not actively working on it.

gracie2339:
Hi All, I found a workaround with this by writing a generic test in dbt and then referencing it in the yml files like any other ordinary relationships test:

This one has a 1% error threshold and works for BigQuery. Feel free to change as you see fit.

```
{% test relationships_with_1_pct_error_threshold(model, column_name, field, to) %}

with parent as (
    select
        {{ field }} as id
    from {{ to }}
),

child as (
    select 
        {{ column_name }} as id
    from {{ model }}
),

error_threshold as (
    select
        cast(count(distinct id) * .01 as int) as num
    from parent
)

select distinct id
from child 
where id is not null
    and id not in (select id from parent)
    and (
        select count(distinct id) 
        from child
        where id is not null
            and id not in (select id from parent)
        ) > (select num from error_threshold)

{% endtest %}
```

PR #5194, opened 2022-04-29
update label check for dependabot action

emmyoop:
Fix small bug where action to add changelog file never kicked off if the label was added before PR creation.  It still worked if you added label after creation.

test to add label [after PR creation](https://github.com/emmyoop/action_testing/pull/11)

test to add label [before PR creation](https://github.com/emmyoop/action_testing/pull/10)

This should eventually be moved over to shared actions since we will need it across multiple repos.

### Checklist

- [x] ~This PR includes tests, or~ tests are not required/relevant for this PR
- [x] ~I have [opened an issue to add/update docs](https://github.com/dbt-labs/docs.getdbt.com/issues/new/choose), or~ docs changes are not required/relevant for this PR
- [ ] ~I have run `changie new` to [create a changelog entry](https://github.com/dbt-labs/dbt-core/blob/main/CONTRIBUTING.md#Adding-CHANGELOG-Entry)~

### DIFF
--- a/.github/workflows/dependency-changelog.yml
+++ b/.github/workflows/dependency-changelog.yml
@@ -32,7 +32,7 @@ permissions:
 
 jobs:
   dependency_changelog:
-    if: ${{ github.event.label.name == 'dependencies' }}
+    if: "contains(github.event.pull_request.labels.*.name, 'dependencies')"
     runs-on: ubuntu-latest
 
     steps:


.github/workflows/dependency-changelog.yml
```diff
@@ -32,7 +32,7 @@ permissions:
 
 jobs:
   dependency_changelog:
-    if: ${{ github.event.label.name == 'dependencies' }}
```
leahwicz, line 35: In theory this would also kick-off if another label was added after the `dependencies` label, but nothing would happen since the changelog file would already exist, correct?

emmyoop: @leahwicz, correct.  There's not a way I know of to only run if a specific label is added.  So this will run, then it hits a check for if the changelog exists:

https://github.com/dbt-labs/dbt-core/blob/efd1585966166c5e34735328eada291537163120/.github/workflows/dependency-changelog.yml#L61-L75

And should the pass on adding another one.

I could add another `if` to check if the action was `labeled` and the label was dependencies.  IT would still trigger the workflow but would fail sooner.  Something like this (syntax may be wrong):

```
jobs:
  dependency_changelog:
    if: |
      (${{ github.event == 'opened' }} && "contains(github.event.pull_request.labels.*.name, 'dependencies')")
      || (${{ github.event == 'labeled' }} && ${{ github.event.label.name == 'dependencies' }})
```

leahwicz: Yeah what you have is fine then since it does check for the changelog existing so the danger of the workflow running multiple times is accounted for. 🚢 it!

PR #1983, opened 2019-12-05
Restore drop_schema to adapter in jinja context (#1980)

beckjake:
Fixes #1980 

I can't remember if we removed this as an adapter method intentionally or accidentally, but it was a mistake either way.

As part of this I fixed up the cache's behavior when you do things to schemas (dropping a schema now cascades into dropping all its contents, as it should). There was an especially tricky bug with renaming relations where we effectively dropped the old relation's schema(!!!)

I added a test, it is basically a big pile of things you should never do, but it tests this functionality well enough. I made the test for each db as this seems like something easy to break on a per-db basis.

PR #5198, opened 2022-04-29
propseal for modification to drop_test_schema

McKnight-42:
resolves # https://github.com/dbt-labs/dbt-redshift/issues/110

### Description
Test files where getting locked due to transactions leaving them in database after test ended leading to error:
```
psycopg2.errors.InternalError_: 1040
DETAIL:  Maximum tables limit exceeded. The maximum number of tables per cluster is 9900 for this instance type. The limit includes permanent and temporary tables. (pid:25694)
```
due to database reaching limit on tables.

### Checklist

### DIFF
--- a/core/dbt/tests/fixtures/project.py
+++ b/core/dbt/tests/fixtures/project.py
@@ -404,11 +404,16 @@ class TestProjInfo:
 
     # Drop the unique test schema, usually called in test cleanup
     def drop_test_schema(self):
-        with get_connection(self.adapter):
-            for schema_name in self.created_schemas:
-                relation = self.adapter.Relation.create(database=self.database, schema=schema_name)
-                self.adapter.drop_schema(relation)
-            self.created_schemas = []
+        for schema_name in self.created_schemas:
+            relation = self.adapter.Relation.create(database=self.database, schema=schema_name)
+            schema = self.adapter.quote_as_configured(relation.schema, "schema")
+            self.run_sql(f"drop schema if exists {schema} cascade")
+        self.created_schemas = []
+        # with get_connection(self.adapter):
+        #     for schema_name in self.created_schemas:
+        #         relation = self.adapter.Relation.create(database=self.database, schema=schema_name)
+        #         self.adapter.drop_schema(relation)
+        #     self.created_schemas = []
 
     # This return a dictionary of table names to 'view' or 'table' values.
     def get_tables_in_schema(self):
--- a/core/dbt/tests/fixtures/project.py
+++ b/core/dbt/tests/fixtures/project.py
@@ -404,16 +404,11 @@ class TestProjInfo:
 
     # Drop the unique test schema, usually called in test cleanup
     def drop_test_schema(self):
-        for schema_name in self.created_schemas:
-            relation = self.adapter.Relation.create(database=self.database, schema=schema_name)
-            schema = self.adapter.quote_as_configured(relation.schema, "schema")
-            self.run_sql(f"drop schema if exists {schema} cascade")
-        self.created_schemas = []
-        # with get_connection(self.adapter):
-        #     for schema_name in self.created_schemas:
-        #         relation = self.adapter.Relation.create(database=self.database, schema=schema_name)
-        #         self.adapter.drop_schema(relation)
-        #     self.created_schemas = []
+        with get_connection(self.adapter):
+            for schema_name in self.created_schemas:
+                relation = self.adapter.Relation.create(database=self.database, schema=schema_name)
+                self.adapter.drop_schema(relation)
+            self.created_schemas = []
 
     # This return a dictionary of table names to 'view' or 'table' values.
     def get_tables_in_schema(self):
--- a/.changes/unreleased/Fixes-20220429-160742.yaml
+++ b/.changes/unreleased/Fixes-20220429-160742.yaml
@@ -1,0 +1,8 @@
+kind: Fixes
+body: Modifying the drop_test_schema to work better with Redshift issues around locked
+  tables and current transactions
+time: 2022-04-29T16:07:42.750046-05:00
+custom:
+  Author: Mcknight-42
+  Issue: "5200"
+  PR: "5198"
--- a/core/dbt/adapters/sql/impl.py
+++ b/core/dbt/adapters/sql/impl.py
@@ -171,6 +171,7 @@ class SQLAdapter(BaseAdapter):
             "relation": relation,
         }
         self.execute_macro(DROP_SCHEMA_MACRO_NAME, kwargs=kwargs)
+        self.commit_if_has_connection()
         # we can update the cache here
         self.cache.drop_schema(relation.database, relation.schema)
 


McKnight-42:
@gshank, @jtcohen6  was wondering if this change seems reasonable or if it posses any known problems?

gshank:
The reason I used adapter.drop_schema was in theory to use more adapter specific macros/methods. Do you know why this works and adapter.drop_schema doesn't?

McKnight-42:
for redshift I think the drop_schema method attaches the cascade to it which doesn't work if a table is locked due to a current transaction.

core/dbt/tests/fixtures/project.py
```diff
@@ -404,11 +404,16 @@ def create_test_schema(self, schema_name=None):
...
     def drop_test_schema(self):
-        with get_connection(self.adapter):
-            for schema_name in self.created_schemas:
-                relation = self.adapter.Relation.create(database=self.database, schema=schema_name)
-                self.adapter.drop_schema(relation)
-            self.created_schemas = []
+        for schema_name in self.created_schemas:
+            relation = self.adapter.Relation.create(database=self.database, schema=schema_name)
+            schema = self.adapter.quote_as_configured(relation.schema, "schema")
+            self.run_sql(f"drop schema if exists {schema} cascade")
```
jtcohen6, line 410: I'd definitely prefer us _not_ going back to hard-coded `run_sql` here. This will definitely cause issues on other adapters.

Do we know _why_ we're running `drop schema` at a point in time when there are still open transactions against objects in the schema? In theory, every test cases uses its own dedicated schema, so by the time we're running `drop_test_schema`, we should be totally done with it

McKnight-42: @jtcohen6 from what I can tell it seems like a lock gets put on test_basic and Incremental_unique_id because they both reference the same tables between tests e.g `snapshot_expected`, and `duplicate_insert` and wonder if test hasn't had enough time to tear down ref to that table before next test kicks off especially if in same class so it locks as to not have two people updating same. table. as mentioned in https://docs.aws.amazon.com/redshift/latest/dg/r_STV_LOCKS.html.

jtcohen6: > the same tables

These should be in different schemas, though—a dedicated schema for each test case, for all the resources required for that test—right?

gshank: Yes, every tests should get a randomly generated unique schema. So in theory different tests should not be locking each other's tables. And drop_test_schema should only be executed at the end of the test.

gshank:
It's been a while since I had to interact with Redshift, but I do vaguely remember that dropping tables in Redshift didn't necessary give back everything... you had to recover the space, etc, by .... vacuuming? It's been a while. I'm wondering if this is a Redshift management thing anyway.

gshank:
Also the "drop_schema" default implementation (which I believe is used by Redshift) is: drop schema if exists {{ relation.without_identifier() }} cascade. So that code should pretty much be doing the same thing as what's happening here.

McKnight-42:
today i've continued playing around with what tests do leave files after running, non of the older integration tests have today, its only the 11 files generated by the new functional tests, so could this be an issue with how the fixtures work by chance? the initiating models/seeds staying open technically being a active transaction against the tests?

McKnight-42:
> Also the "drop_schema" default implementation (which I believe is used by Redshift) is: drop schema if exists {{ relation.without_identifier() }} cascade. So that code should pretty much be doing the same thing as what's happening here.

gshank:
One thing that is different is that the older test framework used the initial adapter that was created by test/integration/base.py, which meant that we had to patch providers. In order to avoid that patching, the new tests now use the latest adapter. I wouldn't *think* that would make a difference. It's pretty much the same thing that you get when you run separate CLI dbt commands (a "new" adapter object every time).

gshank:
By "leave files after running", you actually mean "leave schemas after running"?

gshank:
Also, what query are you running to see what tables are left over, and what are the schema/table names?

McKnight-42:
> Also, what query are you running to see what tables are left over, and what are the schema/table names?

General process to look this over has been to move breakpoints throughtut the process even tried to debug in the macro which is interesting.

then in postico or whatever you need to sign in as super user to see lock tables and other relevant information 
various commands run:
to see if you can drop the talbe
`--drop schema if exists "test16512569692993986231_test_basic" cascade;`
to look at current locked files
`select table_id, last_update, lock_owner, lock_owner_pid from stv_locks;`
see open locked. tables
`select "table" from svv_table_info where table_id = <Insert a table_id from above command>;`
to get more information about table like what type of lock is on table.
`select * from svv_transactions;`

kwigley:
Something wonky is going here when calling `self.adapter.drop_schema(relation)` for redshift specifically that is difficult to debug. I have a feeling it is related to this bit of code https://github.com/dbt-labs/dbt-redshift/blob/6543b82bf5943e2bc272880984380205437d49e6/dbt/adapters/redshift/impl.py#L36-L52. While this PR isn't a solution, it works properly for dropping schemas created by tests in `dbt-redshift`. I'm still working to understand why though..

gshank:
So it sounds like the problem is with the drop_schema/drop_relation redshift code, not the test_drop_schema per se.

gshank:
The base implementation of "drop_relation" in core/dbt/adapters/sql/impl.py does "self.cache_dropped(relation)", but the redshift version doesn't.

McKnight-42:
> The base implementation of "drop_relation" in core/dbt/adapters/sql/impl.py does "self.cache_dropped(relation)", but the redshift version calls the super version... so the only thing different is the fresh_transaction bit?

Yeah which applies the exclusivelock and starts a new transaction. but is supposed to drop it to allow us to use `drop_relation` from core

gshank:
It looks like adapter.drop_schema doesn't work in postgres either.

gshank:
I looked at the logs and the 'drop_schema' call was getting rolled back every time. The 'drop_schema' method in core/dbt/adapters/sql/impl.py did not have a commit. I added "self.commit_if_has_connection()" after "self.execute_macro(DROP_SCHEMA_MACRO_NAME, kwargs=kwargs)" and it worked.

McKnight-42:
I just tried the addition and ran against redshift I still had tests remain, will play around with it some more tonight. thanks for having eyes on this everyone!

jtcohen6:
If adding `commit_if_has_connection()` fixes this, I'm happy with doing that.

@gshank Thanks for finding and linking the context from https://github.com/dbt-labs/dbt-core/pull/1983. I have to disagree with Drew from 2019 on these points, though:

> these sprinkled-in transactional statements make it super hard to reason about what dbt is doing in different parts of the codebase.
> it's the destructive nature of this method that gave me pause

If the `commit` is necessary, we should have it. It should be sprinkled through the logs wherever it's needed. As far as the "destructive" nature of `drop schema`: dbt itself doesn't actually call `drop_schema` _anywhere_, in the course of its standard operations. It gets called in two highly specific places:
- While running functional tests on `dbt-core` (this issue)
- By dbt Cloud, after a PR is merged, to drop the "PR schema" used by the CI check built from that PR. I just checked into this part of the dbt Cloud codebase, and guess what's there?
```python
            adapter.drop_schema(database, schema)
            adapter.commit_if_has_connection()
```

Of course, users may also call it for their own purposes. If they do, I think they'd expect it to work :)

gshank (approved):

McKnight-42:
@gshank do we still want to add a test before merging?

Issue #4988, opened 2022-04-03
Make it easy to run integration tests in an adapter plugin against a dbt-core PR


jtcohen6:
I dream of the ability to:
- Tag a PR in this repo with `test redshift`, `test snowflake`, `test bigquery`, `test spark`
- Have that trigger GHA to run adapter integration tests, _in those respective repos_, with links to each running test

This makes it easy to ensure that a PR changing adapter-facing functionality, such as #4986, does not include breaking changes for our adapter plugins. The alternative is to find out a day later, when the nightly integration tests have failed for that plugin.

### challenge

This would require the GHA for adapter integration tests to take an optional configuration, for the branch of `dbt-core` that they install in `dev_requirements.txt`. The configuration of that branch is already a problem for us, which requires manual intervention whenever we cut a minor version `.latest` branch: https://github.com/dbt-labs/dbt-snowflake/pull/128, https://github.com/dbt-labs/dbt-redshift/pull/89

Would it be possible to set this up so that:
- **In local dev**, the version of `dbt-core` installed by `dev_requirements.txt` is always `main` (i.e. local development). If you're making local changes to `dbt-core` as well, this shouldn't even take effect, since you already have your working local version of `dbt-core` installed in the same virtualenv
- **In CI**, the version installed can be set by supplied configuration (arbitrary branch/commit of `dbt-core`, as above). If not set, it should use the "target" PR branch (`main`, `1.0.latest`, `1.1.latest`, ...) — just requires us to be consistent in our branch naming across `dbt-core` + plugins

jtcohen6:
Probably a cross-repo GHA workflow, which dynamically updates `dev-requirements.txt`.

Halfway step here: Trigger on merge to `main`. (Then we skip the need for dynamic updates to `dev-requirements.txt`.)

@colin-rogers-dbt's idea: Always run against `dbt-postgres`, and then one of the "fun" ones chosen at random, rather than all adapters every time. `test random`?

Honestly would just be very helpful to have this as a manually triggered GHA, too! Rather than the manual PR process today...

emmyoop:
The adapters are already going to need a script to automate updating `dev-requirements.txt` for cutting the `*.latest` branches per https://github.com/dbt-labs/actions/issues/76.  That's a lot of the lift to achieve this honestly.    Being able to kick off an adapter test before merging to main by adding a label to a core PR would be amazing.

colin-rogers-dbt:
[...]

PR #4893, opened 2022-03-17
Fix adding new cols to check_cols in snapshots

GtheSheep:
resolves #3146 

Let me know if I need to open a new issue for this/ this can't be re-opened?
Also, does this require a test? Will add if so

### Description

Adds new columns to a snapshot when a new entry is added to the `check_cols` arg with the `check` strategy

### Checklist

- [x] I have signed the [CLA](https://docs.getdbt.com/docs/contributor-license-agreements)
- [x] I have run this code in development and it appears to resolve the stated issue
- [x] This PR includes tests, or tests are not required/relevant for this PR
- [x] I have added information about my change to be included in the [CHANGELOG](https://github.com/dbt-labs/dbt-core/blob/main/CONTRIBUTING.md#Adding-CHANGELOG-Entry).

PR #5223, opened 2022-05-09
Fix: column comparison logic for check-strategy snapshots

### LINKED ISSUE #5222: [CT-619] [Bug] Column comparison in snapshot_check_all_get_existing_columns
https://github.com/dbt-labs/dbt-core/pull/4893 added a `snapshot_check_all_get_existing_columns` macro, with more complex logic to handle column schema evolution, in cases where a user has simultaneously added a new column to the snapshot query _and_ added it to an explicit `check_cols` config (i.e. not `"all"`).

https://github.com/dbt-labs/dbt-core/blob/ab0c3510ebe70a67fb669b88e6ea36aff2f46339/core/dbt/include/global_project/macros/materializations/snapshots/strategies.sql#L106-L134

However, this logic depends on comparison between user-provided `check_cols` config to the columns accessed from the database. This doesn't account for quoting/casing differences, which are just about guaranteed on Snowflake. The macro always finds that `column_added`, so snapshots with `check` strategy + explicit `check_cols` config always return `where TRUE` for their `row_changed_expr`:

https://github.com/dbt-labs/dbt-core/blob/ab0c3510ebe70a67fb669b88e6ea36aff2f46339/core/dbt/include/global_project/macros/materializations/snapshots/strategies.sql#L147-L151

Hence, these failing tests: https://github.com/dbt-labs/dbt-snowflake/issues/154

We have two options for refactoring this macro:
1. Do a case-and-quote-insensitive comparison between `check_cols_config` and `existing_cols`
2. Query `check_cols_config` from the database, in all cases, to ensure that casing/quoting will line up exactly

Personally, I'd opt for the second. It's better to run an extra query than to risk snapshot failure (very bad!). I'm aware that the change in #4893 has us running at least 2 extra queries, relative to the previous behavior for explicit `check_cols`. I don't think it's worth an opt-in config: the principle of the thing is that _snapshots shouldn't fail_, even if it requires a lot of metadata queries along the way. Worst case, this is user-space code, someone could override `snapshot_check_all_get_existing_columns` for a more concise (if risky) approach.

jtcohen6:
resolves #5222
resolves https://github.com/dbt-labs/dbt-snowflake/issues/154

### Description

The logic in `snapshot_check_all_get_existing_columns` added by https://github.com/dbt-labs/dbt-core/pull/4893 doesn't account for cases where the user-configured `check_cols` doesn't match the database's column quoting/casing.

This is almost always true on Snowflake, which is where we've been seeing test failures. The snapshot doesn't fail explicitly, but it always considers the column schema to have changed, so it adds new records _every single time it runs_.

I added a new test case, on top of the one added in #4893. It fails without the changes to the macro in this PR, and passes with them.

For a longer-term resolution, and to ensure this doesn't happen again, let's:
- Move `tests/functional/simple_snapshot` into the "adapter zone" and get it running on Snowflake
- Create a mechanism to proactively test `dbt-core` changes against one/all adapter(s): https://github.com/dbt-labs/dbt-core/issues/4988

That's out of scope for the current PR, which is just trying to fix the nightly tests in `dbt-snowflake`, and preclude an unintended regression in v1.2.

### Checklist

### DIFF
--- a/.changes/unreleased/Fixes-20220509-130021.yaml
+++ b/.changes/unreleased/Fixes-20220509-130021.yaml
@@ -1,0 +1,8 @@
+kind: Fixes
+body: Fix column comparison in snapshot_check_all_get_existing_columns for check-strategy
+  snapshots with explicit check_cols defined
+time: 2022-05-09T13:00:21.649028+02:00
+custom:
+  Author: jtcohen6
+  Issue: "5222"
+  PR: "5223"
--- a/core/dbt/include/global_project/macros/materializations/snapshots/strategies.sql
+++ b/core/dbt/include/global_project/macros/materializations/snapshots/strategies.sql
@@ -104,22 +104,30 @@
 
 
 {% macro snapshot_check_all_get_existing_columns(node, target_exists, check_cols_config) -%}
+    {%- if not target_exists -%}
+        {#-- no table yet -> return whatever the query does --#}
+        {{ return((false, query_columns)) }}
+    {%- endif -%}
+
+    {#-- handle any schema changes --#}
+    {%- set target_relation = adapter.get_relation(database=node.database, schema=node.schema, identifier=node.alias) -%}
+
     {% if check_cols_config == 'all' %}
         {%- set query_columns = get_columns_in_query(node['compiled_sql']) -%}
+
     {% elif check_cols_config is iterable and (check_cols_config | length) > 0 %}
-        {% set query_columns = check_cols_config %}
+        {#-- query for proper casing/quoting, to support comparison below --#}
+        {%- set select_check_cols_from_target -%}
+          select {{ check_cols_config | join(', ') }} from ({{ node['compiled_sql'] }}) subq
+        {%- endset -%}
+        {% set query_columns = get_columns_in_query(select_check_cols_from_target) %}
+
     {% else %}
         {% do exceptions.raise_compiler_error("Invalid value for 'check_cols': " ~ check_cols_config) %}
     {% endif %}
-    {%- if not target_exists -%}
-        {# no table yet -> return whatever the query does #}
-        {{ return([false, query_columns]) }}
-    {%- endif -%}
-    {# handle any schema changes #}
-    {%- set target_table = node.get('alias', node.get('name')) -%}
-    {%- set target_relation = adapter.get_relation(database=node.database, schema=node.schema, identifier=target_table) -%}
-    {%- set existing_cols = get_columns_in_query('select * from ' ~ target_relation) -%}
-    {%- set ns = namespace() -%} {# handle for-loop scoping with a namespace #}
+
+    {%- set existing_cols = get_columns_in_relation(target_relation) | map(attribute = 'name') | list -%}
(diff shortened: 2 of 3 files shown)


jtcohen6:
@leahwicz I believe so. I'll kick off a `dbt-snowflake` PR to run against this branch to confirm.

ueshin:
Hi @jtcohen6, seems like this commit broke `dbt-databrick`'s integration tests and I guess `dbt-spark`'s, too.

```
[gw0] [100%] FAILED tests/functional/adapter/test_basic.py::TestSnapshotCheckColsDatabricks::test_snapshot_check_cols

==================================================================================================== FAILURES =====================================================================================================
____________________________________________________________________________ TestSnapshotCheckColsDatabricks.test_snapshot_check_cols _____________________________________________________________________________
[gw0] darwin -- Python 3.9.11 /.../bin/python

self = <test_basic.TestSnapshotCheckColsDatabricks object at 0x7fbb684f9e20>, project = <dbt.tests.fixtures.project.TestProjInfo object at 0x7fbbb8a081f0>

    def test_snapshot_check_cols(self, project):
        # seed command
        results = run_dbt(["seed"])
        assert len(results) == 2

        # snapshot command
        results = run_dbt(["snapshot"])
        for result in results:
            assert result.status == "success"

        # check rowcounts for all snapshots
        check_relation_rows(project, "cc_all_snapshot", 10)
        check_relation_rows(project, "cc_name_snapshot", 10)
        check_relation_rows(project, "cc_date_snapshot", 10)

        relation = relation_from_name(project.adapter, "cc_all_snapshot")
        result = project.run_sql(f"select * from {relation}", fetch="all")

        # point at the "added" seed so the snapshot sees 10 new rows
        results = run_dbt(["--no-partial-parse", "snapshot", "--vars", "seed_name: added"])
        for result in results:
            assert result.status == "success"

        # check rowcounts for all snapshots
>       check_relation_rows(project, "cc_all_snapshot", 20)

/.../dbt-core/tests/adapter/dbt/tests/adapter/basic/test_snapshot_check_cols.py:62:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

project = <dbt.tests.fixtures.project.TestProjInfo object at 0x7fbbb8a081f0>, snapshot_name = 'cc_all_snapshot', count = 20

    def check_relation_rows(project, snapshot_name, count):
        relation = relation_from_name(project.adapter, snapshot_name)
        result = project.run_sql(f"select count(*) as num_rows from {relation}", fetch="one")
>       assert result[0] == count
E       assert 30 == 20

/.../dbt-core/tests/adapter/dbt/tests/adapter/basic/test_snapshot_check_cols.py:15: AssertionError
```

PR #5241, opened 2022-05-13
Add support for File Selectors and add file selectors to the default method selector list

### LINKED ISSUE #5240: [CT-640] File selectors and support for file selectors as a default method

### Describe the Feature

`dbt should just know that when i say: 
    dbt run -m my_model.sql 
i mean: 
    dbt run -m my_model
`

https://twitter.com/pedram_navid/status/1519160872588898304

Seemed like a pretty straightforward paper-cut feature request.

### Who will this benefit?

Pedram, apparently

yes!

### Anything else?

_No response_

jwills:
resolves this small gripe: "To this day, I can’t give dbt a file name and hope it figures out what I mean, but instead I still have to remove the .sql at the end."

https://pedram.substack.com/p/we-need-to-talk-about-dbt

and thus fix #5240 

### Description

I'm adding a `FileSelectorMethod` that uses the name of a .sql file as the way to find model(s) to run via a selector and I added a check to the `default_method` code that does an automatic determination of which selector method to use that will prefer the `File` selector to the `FQN` selector for selector values that end with the string ".sql".

### Checklist

### DIFF
--- a/.changes/unreleased/Features-20220512-215748.yaml
+++ b/.changes/unreleased/Features-20220512-215748.yaml
@@ -1,0 +1,7 @@
+kind: Features
+body: Adds file selectors and support for file selectors in the default method selector
+time: 2022-05-12T21:57:48.289674-07:00
+custom:
+  Author: jwills
+  Issue: "5240"
+  PR: "5241"
--- a/core/dbt/graph/selector_methods.py
+++ b/core/dbt/graph/selector_methods.py
@@ -39,6 +39,7 @@ class MethodName(StrEnum):
     Tag = "tag"
     Source = "source"
     Path = "path"
+    File = "file"
     Package = "package"
     Config = "config"
     TestName = "test_name"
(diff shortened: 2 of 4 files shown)


jwills:
Hi! So this is a, uhh, emotionally fraught (?!) pull request, but I don't want to let that get in the way of, ya know, writing good software.

The other way I could have implemented this feature was by removing the last few characters of the input selector if they looked like a file extension (.sql to start) and then falling back to the FQN method. I didn't have a strong preference between what I did here and that option; I'm not super-familiar with the code I'm changing here and so I went with the new file selector impl on the principle that I should try not to muck with user input too much, but am more than happy to take this code in a different direction if there are other considerations here that I am unaware of. 🙇

ChenyuLInx (approved):
@jwills Thanks for submitting this so fast!!! And take me to different parts of the codebase! I think the current approach looks great!
If we take the other route you mentioned, we will run into some unintended behavior. For example, in my `models` dir I have, 

> test.sql
test/inside.sql

If I do `dbt ls -s test.sql` it will actually show both.
Maybe there's a path to refactor but I think what we currently get looks great!

PR #5264, opened 2022-05-17
require lower version of networkx in py37

### LINKED ISSUE #5254: [CT-650] [Bug] python 3.7 version incompatible with networkx

### Current Behavior

Networkx is pinned across several minor versions at the moment

https://github.com/dbt-labs/dbt-core/blob/72c17c446440afe7d4b6501ed364effe8e14c9f2/core/setup.py#L64

Networkx minor versions (or maybe even patches) can introduce python version breaking changes for DBT's implementation. For example 2.3 was for python 3.5, 3.6, 3.7, while 2.8 is listed as being >3.8.

The current version of networkx (2.8) worked fine with 3.7 for DBT until a recent release
https://github.com/networkx/networkx/releases/tag/networkx-2.8.1rc1

this prerelease seems to be installed at the moment. This introduced use of `functools.cached_property` in `networkx.graph` which has made many versions of DBT require python 3.8, despite that not being clear.

For example, we are using 1.0.1 which is now unstable for python 3.7 users.

### Expected Behavior

I'd expect this to be more stringently pinned to ensure that old versions of DBT aren't subject to versions changing in other modules.

### Steps To Reproduce

_No response_

### Relevant log output

_No response_

### Environment

```markdown
- OS: Mac
- Python: 3.7
- dbt: 1.0.1, true for everything until 1.1.0 AFAIK
```

snowflake

### Additional Context

_No response_

ChenyuLInx:
resolves #5254 

# Description

This will install the networkx with version lower than 2.8.1 for python 3.7


core/setup.py
```diff
@@ -28,7 +28,33 @@
...
+    "typing-extensions>=3.7.4",
+    "werkzeug>=1,<3",
+    # the following are all to match snowflake-connector-python
+    "requests<3.0.0",
+    "idna>=2.5,<4",
+    "cffi>=1.9,<2.0.0",
+]
 
+# we check python version > 3.7.2 above so this will make sure we are using python 3.7
+if sys.version_info < (3, 8):
```
jtcohen6, line 54: I _think_ this only gets checked at "build" time, so for people installing a prebuilt wheel, this will reflect the `sys` of the program that built the wheel, not the person installing it. I think we need the actual [environment marker](https://peps.python.org/pep-0508/#environment-markers) of `;python_version` in the `install_requires`, akin to what we do for `hologram` [here](https://github.com/dbt-labs/hologram/blob/6735a19757787cd4bdfa135a199547c71a5ce80e/setup.py#L6)

(I could be way off base here, I'm just doing my best KWigs impression)

jtcohen6 (commented):
We'll need to backport this to `1.1.latest` + `1.0.latest` for patch releases as well

stu-k (approved):

leahwicz:
@ChenyuLInx do you know if we still need this PR now after @iknox-fa fix #5326?

ChenyuLInx:
> @ChenyuLInx do you know if we still need this PR now after @iknox-fa fix #5326?

Oh, I missed this one, yes this is separate from what Ian's change is. I am going to resolve the conflict and merge it

Issue #480, opened 2017-07-05
Create custom jinja filters


drewbanin:
From [Haney](https://getdbt.slack.com/archives/C2JRTGLLS/p1499283062166315?thread_ts=1498596548.340340&cid=C2JRTGLLS)

Currently Jinja doesn't support (or at least I can't find support for) many `set` operations. So, I'm using `adapter.get_columns_in_table` but I want to "ignore" a list of columns. Jinja doesn't seem to support things like `set(a) - set(b)` but if there was a way to supply custom filter functions that were passed down to Jinja I could have handed it a custom function to do something like {{ set_a|not_in(set_b) }}

josegonzalez:
+1 This would be pretty good. An example use case is loading a `profile` with a DSN as an environment variable, and then a filter that retrieves a specific part of that DSN. For example:

```
dbt:
  outputs:
    warehouse:
      type: "{{ dsn_env_var('REDSHIFT_URL', 'scheme') }}"
      threads: 4
      host: "{{ dsn_env_var('REDSHIFT_URL', 'host') }}"
      port: {{ dsn_env_var('REDSHIFT_URL', 'port') }}
      user: "{{ dsn_env_var('REDSHIFT_URL', 'user') }}"
      pass: "{{ dsn_env_var('REDSHIFT_URL', 'password') }}"
      dbname: "{{ dsn_env_var('REDSHIFT_URL', 'database') }}"
  target: warehouse
```

drewbanin:
It may be compelling to revisit this issue in the future when dbt has a more well-defined Python API. Until then, there isn't a good way to write Python code (to implement filters) in the context of a dbt project.

gerbenoostra:
@drewbanin Writing python in a dbt project might make it all more complex. 
Wouldn't it be more straightforward to include a python package in the python environment, just as custom adapters are currently possible?

giovanni-girelli-sdg:
@drewbanin Will the release of dbt 1.0.0 mean that the python API is solid enough to look at this? I think custom filters (but also macros written in python) would allow for some really neat plugins.

PR #5274, opened 2022-05-18
draft: allow users to augment their jinja context with python objects

mistercrunch:
This is a bit of a proposal and proof of concept at the same time.
It gives a handle to people over their jinja context, allowing them
to inject python objects in there under a `extra_jinja_context`
namespace.

## how?

```bash
export PYTHONPATH=~/.dbt
```

Assuming a file `~/.dbt/dbt_config.py`
```python
# anything added into `extra_jinja_context` becomes available
# in the jinja context under a namespace of the same name
extra_jinja_context = {
    'hello': 'world',
    'test': lambda x: x,
}
```

assuming a model `test.sql`
```sql
SELECT '{{ extra_jinja_context.hello }}' AS test
UNION ALL
SELECT '{{ extra_jinja_context.test('print me') }}' AS test
```

## why?

Writing Jinja is ok when you're writing a lot of SQL with a bit of logic
in it, but highly suboptimal when writing complex logic with a bit of
string outputs. Clearly python is superior to Jinja in many ways:

* access to the full python standard lib
* access to external/powerful libs
* Turing complete, object oriented
* testable using sane mechanisms

## use cases
Use cases are limitless. Bind, hook, trigger, generate SQL.
In order of legitimacy:

* string processing, what you'd typically use Jinja macros for, but
  maybe you prefer python functions over jinja macros. Our main use
  case is creating a `generate_incremental_load_date_bounds()` where
  we'll look at `vars` to offer different loading modes (catchup, date
  range, date list, offsets, ...)
* custom logging similar to
  [this](https://github.com/dbt-labs/dbt-event-logging) but much more
  flexible, we happen to use BigQuery and this particular approach
  doesn't work for BigQuery
* custom logic for hooks
* trigger external things (webhooks!)
* ...

## need more thinking / conversation

* I'm introducing `dbt_config.py`, a new place where python logic can be
  injected into dbt-core, it can be powerful, but can lead to
  environment issues / complexity. People need to not abuse this file.
  No `import pandas as pd` in there please! Where should this live?
  `~/.dbt/`?
* Seems like it'd be great to do this at the project level too, so a
  project can be packaged with the extra jinja context it needs to
  operate. `macros/my_macros.py` anyone!? Conceptually that makes a dbt
  project also a python app in some ways and that may not be ideal.
* interoperability: people sharing projects may need to also share
  dbt_config.py and put it in their pythonpath. It's not that outrageous
  but raises the complexity of the project / env setup. This kind of
  complexity already exist with stuff like `profiles.yml`

# TODO
* agree this is useful
* agree on an approach
* write tests
* write docs
* hand off?

### DIFF
--- a/core/dbt/context/base.py
+++ b/core/dbt/context/base.py
@@ -17,6 +17,7 @@ from dbt.exceptions import (
 from dbt.logger import SECRET_ENV_PREFIX
 from dbt.events.functions import fire_event, get_invocation_id
 from dbt.events.types import MacroEventInfo, MacroEventDebug
+from dbt.utils import get_dbt_config_module
 from dbt.version import __version__ as dbt_version
 
 # These modules are added to the context. Consider alternative
@@ -651,6 +652,22 @@ class BaseContext(metaclass=ContextMeta):
             {{ dt_local }}
         """  # noqa
         return get_context_modules()
+
+    @contextproperty
+    def extra_jinja_context(self) -> Any:
+        """
+        Looks for `extra_jinja_context` in a dbt_config in your
+        PYTHONPATH and exposes it in your jinja context
+        """
+        extra_jinja_context = {}
+        try:
+            dbt_config = get_dbt_config_module()
+            if dbt_config and hasattr(dbt_config, "extra_jinja_context"):
+                extra_jinja_context = dbt_config.extra_jinja_context
+        except Exception:
+            # do nothin'!
+            pass
+        return extra_jinja_context
 
     @contextproperty
     def flags(self) -> Any:
--- a/core/dbt/utils.py
+++ b/core/dbt/utils.py
@@ -20,6 +20,7 @@ from dbt.events.functions import fire_event
 from dbt.events.types import RetryExternalCall, RecordRetryException
 from dbt import flags
 from enum import Enum
+from importlib import import_module
 from typing_extensions import Protocol
 from typing import (
     Tuple,
@@ -659,3 +660,19 @@ def args_to_dict(args):
             var_args[key] = str(var_args[key])
         dict_args[key] = var_args[key]
     return dict_args
+
+
+def get_dbt_config_module():
+    """
+    Looks for a module `dbt_config` somewhere in your PYTHONPATH
+    and returns it. This allows for things like the `extra_jinja_context`
+    configuration hook that can contain python callables or objects, and
+    gets exposed in your jinja context.
+    """
+    dbt_config = None
+    try:
+        dbt_config = import_module("dbt_config")
+    except Exception:
+        # do nothin'!
+        pass
+    return dbt_config


jtcohen6:
@mistercrunch Thanks for the PR, and for the thoughtful accompanying writeup!

I like this idea better than giving people the ability to write [custom Jinja filters](https://jinja.palletsprojects.com/en/3.0.x/api/#writing-filters), or hooking into other Jinja-specific "extensions." There's a _very_ old issue proposing just that: https://github.com/dbt-labs/dbt-core/issues/480. I find myself agreeing with the two most-recent comments there: This feels like a case for plugins!

I'll leave some more detailed thoughts below. I'm indebted to @jwills @gshank @nathaniel-may for talking through this with me yesterday. Maybe this PR actually wants to be a discussion?

### Why not?

> it can be powerful, but can lead to environment issues / complexity. People need to not abuse this file. No `import pandas as pd` in there please!

This is the name of the game. The limitation of Jinja is one of the best things about it :)

At definite risk of mixing my mythological metaphors, custom Python code _within_ the dbt project environment opens a Pandora's box of anti-pattern practices, and books us a one-way passage over the river ~Styx~ Pipx to dependency hell. That risk is present to some degree in [dbt Python models](https://github.com/dbt-labs/dbt-core/discussions/5261#:~:text=Packages%3A%20Python%20models%20can%20import%20PyPI%20packages%20available%20on%20the%20data%20platform.%20dbt%20offers%20product%20functionality%20or%20clear%20recommendations%20on%20how%20to%20configure%20those%20packages%20on%20the%20platform), too, but those dependencies will be installed on the warehouse, in per-model or per-cluster sandboxes. They will _not_ be installed "globally" in the dbt environment, which is the terrifying part. We already [fight enough](https://github.com/dbt-labs/dbt-core/labels/dependencies) with [our dependencies](https://github.com/dbt-labs/dbt-core/blob/e50678c91424142aa1a4995ee1d358dac6a6fdf8/core/setup.py#L53-L73) as is. Imagine if we had to know every single PyPI package that everyone depended on at any point in time, before we felt confident upgrading any?

[One other quick note about Python models: It will of course be possible to write and use `generate_date_filter_clause()` as a Python function, insofar as it's Python code that runs remotely in the warehouse. In some cases, we think it will be desirable for a Python model (?) to actually register its result as a UDF in the database, which can be called from SQL models—but this makes more intuitive sense for `ml_predict()` or `business_minutes_between()` than it does for fancier SQL templating. The risk is much lower visibility for the end user, since dbt isn't "compiling" that more-concise function down to some more-verbose "actual" SQL, as it does with macros.]

### So what then?

Here's my take, with which you're more than welcome to disagree: I think what you're describing is actually a dbt plugin, written in Python, which has the ability to register and expose some of its methods in the user's Jinja context. That plugin would be installed alongside `dbt-core`, and discovered by the `dbt-core` application by namespacing its modules in a well-understood way.

We have a pattern for that today, in the form of **adapter plugins.** We've understood for a long time that the work of translating between dbt <> Another Analytical Database is a _much_ bigger lift than translating business/transformation logic into Jinja-SQL. Adapter authors need the ability to write and test their functionality in a real programming language, and in many cases to access data platform APIs that simply aren't exposed via SQL. (BigQuery is the most prominent example.)

It's also a recognition that the "adapter maintainer" persona is a very different one from "dbt user" / project code writer. There are many fewer adapter maintainers; we (as `dbt-core` maintainers) have a much closer relationship with them; they're opt into a higher maintenance burden, to do a more complex thing; and they understand that they're creating a _product on top of dbt_, with all the thinking about longer-term repercussions that entails, not just hacking a thing that needs to work right now.

Concretely, we allow adapter maintainers to write whatever Python code they need, and to register some of them as class methods and members on the `adapter` object—which end users experience call as "macros" from the `adapter` namespace. To your point about "string processing," for instance: as `dbt-bigquery` maintainers, we wrote the logic for validating user-provided `partition_by` config, and it rendering it as an appropriate string to BigQuery's specifications, [in Python here](https://github.com/dbt-labs/dbt-bigquery/blob/824052240a14d78b74f7e33a17dc8529d6326199/dbt/adapters/bigquery/impl.py#L44-L78). We then call it from within the Jinja context [here](https://github.com/dbt-labs/dbt-bigquery/blob/824052240a14d78b74f7e33a17dc8529d6326199/dbt/include/bigquery/macros/adapters.sql#L49) and [here](https://github.com/dbt-labs/dbt-bigquery/blob/824052240a14d78b74f7e33a17dc8529d6326199/dbt/include/bigquery/macros/adapters.sql#L6).

### How could this be better?

Down the line, I could see a generalized pattern for supporting dbt plugins that don't actually want to be database adapters. They could register custom namespaces, instead of over-loading the `adapter` namespace. Perhaps there could be a `dbt-logging` plugin, which makes available methods like `logging.fire_event()` in the Jinja context—although I'll note that we've meaningfully reinvested in the [core eventing and structured logging interface](https://github.com/dbt-labs/dbt-core/blob/main/core/dbt/events/README.md) for real-time metadata. We're still building the companion tooling for it, but we definitely recommend it instead of [this package](https://github.com/dbt-labs/dbt-event-logging).) Such a plugin _almost_ exists today for [dbt + Openlineage](https://openlineage.io/integration/dbt/), but `openlineage-dbt` has to be a wrapper around the `dbt-core` CLI, rather than a plugin registered within it.) Not today, but eventually, we'd want to support totally custom plugins in dbt Cloud, whether written by technology partners or community members or in-house dbt pros.

In the meantime: What do you think of forking of `dbt-bigquery`? Or an extension off it, in the same way that [`dbt-materialize` is an extension of `dbt-postgres`](https://github.com/MaterializeInc/materialize/blob/790263c601c360f85f6a19fb0328fc9b28a28f21/misc/dbt-materialize/dbt/adapters/materialize/impl.py#L22)? That requires a conscious choice, and a separation between Special Methods and Actual dbt Project Code. If you write methods that solve for a common problem, one which you think other folks would benefit from, a PR upstream to the main `dbt-bigquery` repo is always on the table.

gshank:
I'm going to close this pr since we're not going to merge these changes. Please continue the conversation if you're so inclined.

kdazzle:
Hey @mistercrunch - did you ever continue with something like this? I just proposed a similar thing (#8000) and agree that it would be very useful to have this sort of flexibility within dbt.

PR #5292, opened 2022-05-23
Fix postgres handling for unlimited varchars

### LINKED ISSUE #5238: [CT-636] [Bug] Postgres unlimited varchar default to varchar(256)

### Current Behavior

When using postgres adapter with some varchar typed columns (like when using dbt_utils.type_string()), if you use incremental materialization with sync_new_columns or append_new_columns, or actually anything that will make dbt generate DDL statements, varchars max size defaults to 256 (handler in base/Column.py behaviour when char_size is None). This makes schema evolution impossible when having textual columns defined as {{ dbt_utils.type_string() }}

### Expected Behavior

The original varchar type with no limit should remain as is and not be assigned a 256 default limit, because we don't know the size limit on a varchar without explicit limits. The schema evolution should be possible, with new columns not created as varchar(256)

### Steps To Reproduce

1. Create an incremental model with on_schema_change: sync_all_columns
2. run the model for the first time
3. add a column whose type is {{ dbt_utils.type_string() }} and whose length exceeds 256 characters
4. re-run the model. It will fail because it adds a varchar(256) column

### Relevant log output

```shell
Debug logs for the schema update : 
 "database"."schema"."table":
         Schema changed: True
         Source columns not in target: [<Column col1 (character varying(256))>, <Column col2 (character varying(256))>, <Column col3 (character varying(256))>]
         Target columns not in source: [<Column col4 (character varying(256))>]
         New column types: []
(Note that all these columns are typed are varchar initially not varchar(256))
Postgres columns data check : 
```
```sql
select
          column_name,
          data_type,
          character_maximum_length,
          numeric_precision,
          numeric_scale

      from "production-data-warehouse".INFORMATION_SCHEMA.columns
      where table_name = '{{table}}'
        
        and table_schema = '{{schema}}'
        
      order by ordinal_position
```
```shell
Error: 
`value too long for type character varying(256)`
```

### Environment

```markdown
- OS: Any
- Python: 3.9-slim
- dbt: 1.0.0
```

postgres

### Additional Context

I already found out why it worked like that, and if validated will be happy to submit a PR (linked to the fact that the way postgres columns are retrieved have NULL info regarding varchar columns max_length, and thus when using it it falls back in default Column handler)
Just wondered if this shall be fixed in the PostgresColumn class (that already handles text columns this way), or if we shall just update the dbt_utils.string_type macro to use text instead of varchar for postgres

shrodingers:
resolves #5238 

### Description

Handles `varchar` the same way as `text` for postgres columns (since unlimited `varchar` on postgres is not implicitely limited to 256 characters as on redshift)

### Checklist

### DIFF
--- a/.changes/unreleased/Fixes-20220523-103843.yaml
+++ b/.changes/unreleased/Fixes-20220523-103843.yaml
@@ -4,5 +4,5 @@ body: Remove the default 256 characters limit on postgres character varying type
 time: 2022-05-23T10:38:43.392232+02:00
 custom:
   Author: shrodingers
-  Issue: "636"
+  Issue: "5238"
   PR: "5292"
--- a/.changes/unreleased/Fixes-20220523-103843.yaml
+++ b/.changes/unreleased/Fixes-20220523-103843.yaml
@@ -1,0 +1,8 @@
+kind: Fixes
+body: Remove the default 256 characters limit on postgres character varying type when
+  no limitation is set
+time: 2022-05-23T10:38:43.392232+02:00
+custom:
+  Author: shrodingers
+  Issue: "636"
+  PR: "5292"
--- a/plugins/postgres/dbt/adapters/postgres/relation.py
+++ b/plugins/postgres/dbt/adapters/postgres/relation.py
@@ -26,7 +26,7 @@ class PostgresRelation(BaseRelation):
 class PostgresColumn(Column):
     @property
     def data_type(self):
-        # on postgres, do not convert 'text' to 'varchar()'
-        if self.dtype.lower() == "text":
+        # on postgres, do not convert 'text' or 'varchar' to 'varchar()'
+        if self.dtype.lower() == "text" or (self.dtype.lower() == "varchar" and self.char_size is None):
             return self.dtype
         return super().data_type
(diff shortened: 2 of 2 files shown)


emmyoop:
Close/reopen to retrigger snyk

.changes/unreleased/Fixes-20220523-103843.yaml
```diff
@@ -0,0 +1,8 @@
+kind: Fixes
+body: Remove the default 256 characters limit on postgres character varying type when
+  no limitation is set
+time: 2022-05-23T10:38:43.392232+02:00
+custom:
+  Author: shrodingers
+  Issue: "636"
```
emmyoop, line 7:
```suggestion
  Issue: "5238"
```
This should be the GitHub issue number.  636 is the issue number in our Jira instance.

emmyoop (commented):
Looks good but I'm going to let @McKnight-42 have the final approval.

McKnight-42 (approved):
@shrodingers Sorry for how long it's taken to get eyes on this, seems like a super reasonable! great conversation in the issue between you and @jtcohen6 just the minor fix @emmyoop pointed out for changelog and this should be good to merge!

shrodingers:
Thanks @emmyoop, @McKnight-42 and @jtcohen6 ! 
Just accepted the changelog update (sorry for the late reply was on holidays)

McKnight-42:
@shrodingers great work on this!

PR #5265, opened 2022-05-18
Initialize lift + shift for cross-db macros

jtcohen6:
a step on the way toward resolving #4813

### Description

- Move dispatched + `default__` macros into global project, `macros/utils`
- Move Postgres-specific implementations into `dbt-postgres`
- Move test cases into `dbt-tests-adapter`, with options to override the `macro_namespace`. This allows us to test for backwards compatibility when people call the `dbt_utils`-namespaced versions.

### Checklist

- [x] I have read [the contributing guide](https://github.com/dbt-labs/dbt-core/blob/main/CONTRIBUTING.md) and understand what's expected of me
- [x] I have signed the [CLA](https://docs.getdbt.com/docs/contributor-license-agreements)
- [x] I have run this code in development and it appears to resolve the stated issue
- [x] This PR includes tests, or tests are not required/relevant for this PR
- [ ] I have [opened an issue to add/update docs](https://github.com/dbt-labs/docs.getdbt.com/issues/new/choose), or docs changes are not required/relevant for this PR — TK
- [ ] I have run `changie new` to [create a changelog entry](https://github.com/dbt-labs/dbt-core/blob/main/CONTRIBUTING.md#Adding-CHANGELOG-Entry)

PR #5298, opened 2022-05-25
Lift + shift for cross-db macros

dbeatty10:
Continues #5265

### DIFF
--- a/.changes/unreleased/Features-20220518-114604.yaml
+++ b/.changes/unreleased/Features-20220518-114604.yaml
@@ -1,0 +1,7 @@
+kind: Features
+body: Move cross-db macros from dbt-utils into dbt-core global project
+time: 2022-05-18T11:46:04.557104+02:00
+custom:
+  Author: jtcohen6
+  Issue: "4813"
+  PR: "5265"
--- a/core/dbt/include/global_project/macros/utils/cast_bool_to_text.sql
+++ b/core/dbt/include/global_project/macros/utils/cast_bool_to_text.sql
@@ -1,0 +1,7 @@
+{% macro cast_bool_to_text(field) %}
+  {{ adapter.dispatch('cast_bool_to_text', 'dbt') (field) }}
+{% endmacro %}
+
+{% macro default__cast_bool_to_text(field) %}
+    cast({{ field }} as {{ dbt_utils.type_string() }})
+{% endmacro %}
--- a/core/dbt/include/global_project/macros/utils/cast_bool_to_text.sql
+++ b/core/dbt/include/global_project/macros/utils/cast_bool_to_text.sql
@@ -3,5 +3,5 @@
 {% endmacro %}
 
 {% macro default__cast_bool_to_text(field) %}
-    cast({{ field }} as {{ dbt_utils.type_string() }})
+    cast({{ field }} as {{ dbt.type_string() }})
 {% endmacro %}
--- a/core/dbt/include/global_project/macros/utils/listagg.sql
+++ b/core/dbt/include/global_project/macros/utils/listagg.sql
@@ -1,0 +1,30 @@
+{% macro listagg(measure, delimiter_text="','", order_by_clause=none, limit_num=none) -%}
+    {{ return(adapter.dispatch('listagg', 'dbt') (measure, delimiter_text, order_by_clause, limit_num)) }}
+{%- endmacro %}
+
+{% macro default__listagg(measure, delimiter_text, order_by_clause, limit_num) -%}
+
+    {% if limit_num -%}
+    array_to_string(
+        array_slice(
+            array_agg(
+                {{ measure }}
+            ){% if order_by_clause -%}
+            within group ({{ order_by_clause }})
+            {%- endif %}
+            ,0
+            ,{{ limit_num }}
+        ),
+        {{ delimiter_text }}
+        )
+    {%- else %}
+    listagg(
+        {{ measure }},
+        {{ delimiter_text }}
+        )
+        {% if order_by_clause -%}
+        within group ({{ order_by_clause }})
+        {%- endif %}
+    {%- endif %}
+
+{%- endmacro %}
--- a/core/dbt/include/global_project/macros/utils/any_value.sql
+++ b/core/dbt/include/global_project/macros/utils/any_value.sql
@@ -1,0 +1,9 @@
+{% macro any_value(expression) -%}
+    {{ return(adapter.dispatch('any_value', 'dbt') (expression)) }}
+{% endmacro %}
+
+{% macro default__any_value(expression) -%}
+    
+    any_value({{ expression }})
+    
+{%- endmacro %}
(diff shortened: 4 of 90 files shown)


.changes/unreleased/Features-20220518-114604.yaml
```diff
@@ -0,0 +1,7 @@
+kind: Features
+body: Move cross-db macros from dbt-utils into dbt-core global project
+time: 2022-05-18T11:46:04.557104+02:00
+custom:
+  Author: jtcohen6
```
jtcohen6, line 5:
```suggestion
  Author: jtcohen6 dbeatty10
```

jtcohen6 (commented):
@dbeatty10 Great work here!!

As discussed the other day, let's kick out the `current_timestamp` + `type_*` macros/tests for now. We have other issues to track that work: https://github.com/dbt-labs/dbt-utils/issues/599, https://github.com/dbt-labs/dbt-utils/issues/598

The rest of this looks good to move forward. This is the first PR we'll need to merge, and then we can move forward with the rest.

ChenyuLInx (commented):
There seems to be a lot of tests that didn't got implemented, do we have anything tracking those?

Otherwise this is mainly moving things in dbt-utils to core right?

jtcohen6:
> There seems to be a lot of tests that didn't got implemented, do we have anything tracking those?

Yes! Let's kick those out of this PR. Those tests will be added in https://github.com/dbt-labs/dbt-utils/pull/586 instead.

> Otherwise this is mainly moving things in dbt-utils to core right?

✅

ChenyuLInx (approved):

jtcohen6 (approved):
Let's ship it! And then, let's ship the rest of it!

core/dbt/include/global_project/macros/utils/cast_bool_to_text.sql
```diff
@@ -0,0 +1,7 @@
+{% macro cast_bool_to_text(field) %}
+  {{ adapter.dispatch('cast_bool_to_text', 'dbt') (field) }}
+{% endmacro %}
+
+{% macro default__cast_bool_to_text(field) %}
+    cast({{ field }} as {{ api.Column.translate_type('string') }})
+{% endmacro %}
```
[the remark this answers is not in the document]
dbeatty10: Yep, I think I just took the `type_*` macro behavior based on `api.Column.translate_type()` from your dbt-utils branch.

If/when we push a version of the the `type_*` macro into core, we have the option of replacing these references with that macro instead.

core/dbt/include/global_project/macros/utils/listagg.sql
```diff
@@ -0,0 +1,30 @@
+{% macro listagg(measure, delimiter_text="','", order_by_clause=none, limit_num=none) -%}
```
jtcohen6, line 1: In my work on https://github.com/dbt-labs/dbt-spark/pull/359, I've realized that `order_by_clause` is really not supported on the SparkSQL analogue for `listagg`. I don't think that's a reason to exclude it from the standard / default implementation, since so many other SQL dialects support it so handily, but:
- We _might_ be able to support it, via some array sorting, if it were expressed as `order_by=<column_name>`, instead of expecting a SQL `order by column_name` expression
- I'm curious what you think about the approach I took [here](https://github.com/dbt-labs/dbt-spark/pull/359/files#diff-767481c1a1c644d78d4df7e35bb740cbe0b8fb65ce3eeadeb71f93300058037eR3-R5) — basically, raise a warning if the argument is passed, then ignore it

dbeatty10: The approach you took with ignoring the `order_by_clause` parameter and raising a warning is completely reasonable.

Definitely worth a look if we can replace a user-supplied value like `order by column_name expression` with `order_by=<column_name>` instead. I'll take a peek and report back.

dbeatty10: Overall, I think we should defer this work to a separate PR (if we do in fact take it on).

Before taking it on, here are the main things for us to answer:
1. Do all databases operate correctly with an expression like `order_by_clause="order by column_name"`?
2. Would it be a breaking change? If so, is there any path forward that is non-breaking?

## Findings for question 1
Summary: Yes, it looks like they all support `order_by_clause="order by column_name"`

Details:
- [Here](https://github.com/dbt-labs/dbt-utils/blob/ac072a3c4b78d43a1c013e7de8b8fa6e290b544e/macros/cross_db_utils/listagg.sql) are current implementations across databases.
    - Looks like our supported adapters each needs a different implementation, possibly because it was a relative late-comer first appearing in SQL:2016. (Fascinating discussion on `listagg` [here](https://modern-sql.com/feature/listagg))
- The pytests [here](https://github.com/dbt-labs/dbt-core/blob/1001e42bdd686c85b4191d943d5089d9af333d79/tests/adapter/dbt/tests/adapter/utils/fixture_listagg.py) include expressions like `"order by order_col"` (exactly as you expected)
- The pytests use positional arguments exclusively and don't currently cover named keyword arguments.
    - 🟢 Covering the positional arguments is most crucial since messing with order of those can affect consumers relying on a particular order.
    - 🟡 We might want to add keyword testing for good measure though.
    
## Findings for question 2
It _would_ be a breaking change if we drop `order_by_clause="order by column_name"` and replace it with something like `order_by="column_name"`.

Options:
1. Accept the breaking change
2. Reject the breaking change
3. Introduce the new parameter and deprecate the old parameter in a non-breaking fashion.

The last option would look something like this:
- New signature:`{% macro listagg(measure, delimiter_text="','", order_by_clause=none, limit_num=none, order_by=none) -%}`
    - Doing it this way prevents unexpected errors for any consumers relying upon positional arguments
- Raise an error (or warn) if both `order_by_clause` and `order_by` are provided
- When only `order_by_clause` is provided:
    - This is the "base case", and all the current code is relevant
    - Provide a deprecation warning that this parameter will be removed in a future release.
- When only `order_by_clause` is provided:
    - For non-spark adapters, convert `order_by` into the relevant `order_by_clause` value. Proceed to the code in the base case above.
    - For spark, use `order_by` according to your plans 😈 
    
When doing the final deprecation in the future, we'd probably want to leave `order_by_clause=none` as a non-functional vestigial parameter to preserve the positions of each keyword parameter.

jtcohen6: @dbeatty10 Thanks for going deep on this—much deeper than I did!

I did some more poking, and the amount of `struct` + `array_sort` nonsense that would be required even just to get `order_by=column_name` working is nontrivial. More effort than it's worth until we have a proven need. It's an optional argument, and feels like a nice-to-have.

I'm quite happy to leave this in place, and document / warn that `order_by_clause` is not supported on SparkSQL. It does seem to be a (recent) standard, and most other databases support it in exactly this form (`order_by_clause`). Perhaps we could convince the fine folks developing Apache Spark to add support for it, someday? :)

jtcohen6:
🎉 🎉 🎉

Issue #5286, opened 2022-05-20
[CT-677] [Bug] Downstream models do not run correctly after NetworkX 2.8.1


iknox-fa:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Any 1.0.x version of dbt installed after NetworkX 2.8.1 was released does not correctly run downstream models when using selection syntax.  For some reason this does not seem to effect 1.1

### Expected Behavior

All downstream models are run

### Steps To Reproduce

Install dbt-<whatever> @ v1.0.x
Create a project with a `model_foo` and downstream `model_bar`
Run dbt with selection `dbt run --select +model_foo+`
Observe that only `model_foo` is run

### Environment

```markdown
- OS: MacOS
- Python: 3.10
- dbt: 1.0.6
```

### What database are you using dbt with?

postgres

barberscott:
https://github.com/networkx/networkx/milestone/21?closed=1 -- looks like 2.8.2 of networkx is incoming...

barberscott:
2.8.1 broke, but 2.8.2 did not fix whatever is now wrong with the graph stuff and newtorkx.

iknox-fa:
After quite a few hours of investigation, this is proving a VERY difficult bug to track down.
Here's a synopsis of the work so far.

### We're doing this wrong
The issue can be traced back to our use of `nx.single_source_shortest_path_length` in dbt-core <`1.1.x` [here](https://github.com/dbt-labs/dbt-core/blob/6a772f185d4617461a64e9efee59270396a06cdb/core/dbt/graph/graph.py#L35) and [here](https://github.com/dbt-labs/dbt-core/blob/6a772f185d4617461a64e9efee59270396a06cdb/core/dbt/graph/graph.py#L47).

Specifically, it's used to return a list of a given node's descendants, and by reversing the graph first, it's ancestors.  However, that isn't really what that function is [supposed to do](https://networkx.org/documentation/stable/reference/algorithms/generated/networkx.algorithms.shortest_paths.unweighted.single_source_shortest_path_length.html)!  Using the DAG-specific `nx.ancestors` and `nx.descendants` works correctly in all versions and should have probably been what we did in the first place.  The only downside to this approach is that `nx.desc` / `nx.ances` don't offer a max depth option but it's a simple enough thing to implement.

### There's still a bug, even when we do it the "right way"
[...]

PR #5326, opened 2022-06-01
New ancestor/descendant algo

iknox-fa:
resolves #5286

### Description
Updates ancestor/descendant algos to something more sane that also seems to correct an as-yet-undetermined root cause bug in either networkx or in the internal graph construction.

### Checklist

- [x] I have read [the contributing guide](https://github.com/dbt-labs/dbt-core/blob/main/CONTRIBUTING.md) and understand what's expected of me
- [x] I have signed the [CLA](https://docs.getdbt.com/docs/contributor-license-agreements)
- [x] I have run this code in development and it appears to resolve the stated issue
- [x] This PR includes tests, or tests are not required/relevant for this PR
- [x] I have [opened an issue to add/update docs](https://github.com/dbt-labs/docs.getdbt.com/issues/new/choose), or docs changes are not required/relevant for this PR
- [x] I have run `changie new` to [create a changelog entry](https://github.com/dbt-labs/dbt-core/blob/main/CONTRIBUTING.md#Adding-CHANGELOG-Entry)

PR #5335, opened 2022-06-06
Backports: networkx, context cleanup

jtcohen6:
Backports:
- `networkx` tight pin for remaining v1.1 patches (https://github.com/dbt-labs/dbt-core/issues/5286#issuecomment-1146262827)
- docs + code cleanup in https://github.com/dbt-labs/dbt-core/pull/5334

Let's include both of these in v1.1.1-rc2

### DIFF
--- a/.changes/unreleased/Dependencies-20220606-234258.yaml
+++ b/.changes/unreleased/Dependencies-20220606-234258.yaml
@@ -1,0 +1,7 @@
+kind: Dependencies
+body: Pin networkx to <2.8.4 for v1.1 patches
+time: 2022-06-06T23:42:58.406625+02:00
+custom:
+  Author: jtcohen6
+  Issue: "5286"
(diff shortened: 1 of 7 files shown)


gshank:
I thought 1.1 didn't have the networkx problem?

jtcohen6:
@gshank It didn't, but we're still using a less-than-ideal method (relative to new approach proposed in https://github.com/dbt-labs/dbt-core/pull/5326).

Per https://github.com/dbt-labs/dbt-core/issues/5286#issuecomment-1146262827:

> I like pinning 1.1 and not backporting since it's a change that introduces risk. We can add that to the RC2 that we are going to push out soon

gshank:
Makes sense.

stu-k (approved):

gshank (approved):

PR #5344, opened 2022-06-07
CT-729 Include schema model config in unrendered config

### LINKED ISSUE #5338: [CT-729] [Bug] unrendered_config in manifest.json doesn't update correctly from model properties yaml

### Current Behavior

Config changes from model properties yaml are not correctly propagated to the manifest.json, thus not showing up as a state change when using the `state:modified` selector.

### Expected Behavior

I expect the config changes from the model properties yaml to be correctly propagated to the `manifest.json` - "unrendered_config", just like it does from the config changes made in `dbt_project.yml`.

This will fix the state:modified selector to correctly detect my config changes made in the model properties yaml files.

### Steps To Reproduce

1. Clone [the jaffle shop example](https://github.com/dbt-labs/jaffle_shop ), and set up the profiles.yml to get dbt working.
2. Run `dbt compile`. Look for `customers` model in `target/manifest.json` - has `"unrendered_config": {"materialized": "table"}`
3. Copy `target/manifest.json` to top directory. 
4. Go to `models/schemas.yaml` and add the config block for customers model.
```
- name: customers
   config:
     materialized: view
```
5.  Run `dbt compile`. Look for `customers` model in `target/manifest.json` - still has `"unrendered_config": {"materialized": "table"}`. **The materialization config change hasn't been propagated**
5-1. Side-effect (which is the real problem for me): Run `dbt ls -s state:modified --state .` 
```
The selection criterion 'state:modified' does not match any nodes
No nodes selected!
```
6. Go to `dbt_project.yml`, and update the model config as below:
```
models:
  jaffle_shop:
      materialized: table
      customers:
        materialized: view
      staging:
        materialized: view
```
7. Run `dbt compile`. Look for `customers` model in `target/manifest.json` - now the updated config is shown here: `"unrendered_config": {"materialized": "table"}`.
7-1. Run `dbt ls -s state:modified --state .` 
```
jaffle_shop.customers
jaffle_shop.not_null_customers_customer_id
jaffle_shop.relationships_orders_customer_id__customer_id__ref_customers_
jaffle_shop.unique_customers_customer_id
```
I expect the same result from step 5-1.

### Relevant log output

_No response_

### Environment

```markdown
- OS: MacOS 12.1
- Python: 3.8.9
- dbt: 1.1.0 (dbt-bigquery 1.1.0)
```

bigquery

### Additional Context

Similar issue is reported for the jinja config macro:
https://github.com/dbt-labs/dbt-core/issues/3680
But the fix for the jinja config macro won't necessarily fix this issue.

gshank:
resolves #5338

### Description

Unrendered_config originally only included config from models. When we added config for models in schema files, we didn't include that in unrendered_config. Unrendered_config still doesn't include model config from dbt_project.yml, since the code explicitly avoided that. 

### Checklist

### DIFF
--- a/.changes/unreleased/Fixes-20220607-123058.yaml
+++ b/.changes/unreleased/Fixes-20220607-123058.yaml
@@ -1,0 +1,7 @@
+kind: Fixes
+body: Include schema file config in unrendered_config
+time: 2022-06-07T12:30:58.535207-04:00
+custom:
+  Author: gshank
+  Issue: "5338"
+  PR: "5344"
--- a/core/dbt/parser/base.py
+++ b/core/dbt/parser/base.py
@@ -309,7 +309,9 @@ class ConfiguredParser(
 
         # unrendered_config is used to compare the original database/schema/alias
         # values and to handle 'same_config' and 'same_contents' calls
-        parsed_node.unrendered_config = config.build_config_dict(rendered=False)
+        parsed_node.unrendered_config = config.build_config_dict(
+            rendered=False, patch_config_dict=patch_config_dict
+        )
 
         parsed_node.config_call_dict = config._config_call_dict
 
--- a/tests/functional/artifacts/expected_manifest.py
+++ b/tests/functional/artifacts/expected_manifest.py
@@ -205,9 +205,13 @@ def expected_seeded_manifest(project, model_database=None, quote_model=False):
         schema="test", docs={"node_color": None, "show": False}
     )
 
-    unrendered_model_config = get_unrendered_model_config(materialized="view")
+    unrendered_model_config = get_unrendered_model_config(
+        materialized="view", docs={"show": False}
+    )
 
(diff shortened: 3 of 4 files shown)


gshank:
@jtcohen6  I didn't find any actual documentation about what unrendered_config is *supposed* to contain. The code goes to some trouble to *not* include the config from dbt_project.yml.  I'm not exactly clear on why that is. Could you cast some light on it?

stu-k (approved):

jtcohen6:
@gshank Did we decide if the behavior change here was the right one? Or better than the status quo, at least? I remember discussing that we don't have any way to access the _unrendered_ form of patch (`.yml`) configs for the model, but that is the most correct set of information to include in the model's `unrendered_config`.

gshank:
My recollection is that not including the dbt_project configs when re-rendering was actually a bug.

PR #5345, opened 2022-06-07
python model adr

ChenyuLInx:
These are the thoughts we have around the python model right now! Comments are welcomed!

### DIFF
--- a/docs/arch/adr-004-python-model-arch.md
+++ b/docs/arch/adr-004-python-model-arch.md
@@ -1,0 +1,60 @@
+# Python Model Arch
+
+## Context
+We are thinking of supporting `python` ([roadmap](https://github.com/dbt-labs/dbt-core/blob/main/docs/roadmap/2022-05-dbt-a-core-story.md#scene-3-python-language-dbt-models), [discussion](https://github.com/dbt-labs/dbt-core/discussions/5261)) as a language other than SQL in dbt-core. This would allow users to express transformation logic that is tricky to do in SQL and have more libraries available to them.
+
+### Options
+
+#### Where to run the code
+- running it locally where we run dbt core.
+- running it in the cloud providers' environment.
+
+#### What are the guardrails dbt would enforce for the python model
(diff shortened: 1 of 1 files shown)


docs/arch/adr-004-python-model-arch.md
```diff
@@ -0,0 +1,60 @@
...
+    def ref(*args):
+    refs = {"my_sql_model": "DBT_TEST.DBT_SOMESCHEMA.my_sql_model"}
+    key = ".".join(args)
+    return load_df_function(refs[key])
+    ```
+
+1. functional tests for the python model, these tests are expected to be inherited in the adapter code to make sure intended functions are met.
+1. Generalizing the names of properties (`sql`, `raw_sql`, `compiled_sql`) for a future where it's not all SQL.
+
+This means adapters will need to implement/modify computing engine specific functions including but not limited to
```
leahwicz, line 45: So we are saying these items listed all will include specific logic per data warehouse and that is why they need to live in Adapters code?

leahwicz: Yeah maybe be a little more explicit that the logic in Core is universal and the Adapter logic is very warehouse specific. The `This means adapters...` sentence sounds like whatever we didn't put in Core we are just putting in Adapters instead of there being concrete reasons

docs/arch/adr-004-python-model-arch.md
```diff
@@ -0,0 +1,60 @@
...
+- running it locally where we run dbt core.
+- running it in the cloud providers' environment.
+
+#### What are the guardrails dbt would enforce for the python model
+- None, users can write whatever code they like.
+- focusing on data transformation logic where each python model should have a model function that returns a database object for dbt to materialize.
+
+#### What logic should live in dbt-core
+
+#### Are we going to allow writing macros in python
```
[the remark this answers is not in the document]
leahwicz: I like that. At least it gives a little more commentary as to what we are weighing for this question

PR #5346, opened 2022-06-08
Declare compatibility for previous artifact versions

### LINKED ISSUE #5213: [CT-599] [Bug] Upgrading from 1.0 to 1.1 breaks defer/state

### Current Behavior

After the merge of https://github.com/dbt-labs/dbt-core/pull/5032 manifest version was bumped to v5, but this change was not made backwards compatible like in other version (0.19 -> 0.20 -> 0.21 -> 1.0), and its a breaking change for a `minor` release.

This causes

```
Expected a schema version of "https://schemas.getdbt.com/dbt/manifest/v5.json" in /root/dbt_repo/manifest.json, but found "https://schemas.getdbt.com/dbt/manifest/v4.json". Are you running with a different version of dbt? 
```

### Expected Behavior

DBT version 1.1 is able to run with `--defer` or old state reference to v4.

### Steps To Reproduce

1. Install dbt 1.0
2. In some sample repository/project run `dbt ls`
3. Move `target/` to `old/`
4. Install dbt 1.1
5. Run `dbt run --defer --state old/`

### Relevant log output

```shell
Expected a schema version of "https://schemas.getdbt.com/dbt/manifest/v5.json" in /root/dbt_repo/manifest.json, but found "https://schemas.getdbt.com/dbt/manifest/v4.json". Are you running with a different version of dbt? 
```
```


### Environment

N/A

snowflake

### Additional Context

N/A

jtcohen6:
resolves #5213

### Description

We've been very aggressive about bumping artifact versions _any_ time the artifact schema changes. For `manifest.json`, that's meant a bump for every minor version thus far, even when the schema changes are very minor (hah).

Sometimes, those changes do make it impossible for the latest version of dbt to deserialize an older manifest. Sometimes, they don't, and dbt can use the previous manifest version happily. These cases deserve different treatment, but today, we raise a version mismatch error in all cases.

This PR:
- Develops the capability to differentiate between compatible and incompatible manifest versions, via a new base artifact class method `is_compatible_version`
- Adds manifest `v4` (produced by v1.0) to the set of "compatible" versions
- Adds a functional test case

Very open to feedback on the implementation and nomenclature. If it's something we can feel good about, I'd really like to backport it to `1.1.latest` for inclusion in an upcoming v1.1 patch (probably v1.1.2), since this will ease the migration path for v1.0 &rarr; v1.1 for anyone who uses `state:modified` / the "Slim CI" feature.

### Checklist


gshank (approved):
This looks like a nice neat solution to me.

ChenyuLInx (approved):
The changes look good to me!
One question that doesn't affect the merge of this PR: is it possible to automatically check compatibility by loading the manifest so that we don't need to always define it? But feels like that would have more drawbacks then benefits

jtcohen6:
> is it possible to automatically check compatibility by loading the manifest so that we don't need to always define it? But feels like that would have more drawbacks then benefits

Yeah, I'd rather just keep it simple here, and test the end-user experience as closely as possible by including the actual contents of an actual manifest. It's not a guarantee that this will work in all cases, since this manifest is for a very simple project, but I hope that it would catch most errors that would arise during deserialization. It's also possible that the version upgrade has changed underlying dbt behavior, such as a core built-in macro, thereby triggering many many resources to rerun during "Slim CI."

I'm going to hold off on merging and backporting this PR until we've cut the final release of v1.1.1 (currently available as RC2).

Issue #5331, opened 2022-06-03
[CT-722] `DuplicateYamlKeyException` needs to be warning, not (broken) exception


jtcohen6:
Net-new bug in v1.2. See also https://github.com/dbt-labs/dbt-core/issues/5268, which I believe is caused by the same root change (https://github.com/dbt-labs/dbt-core/pull/5146).

https://github.com/dbt-labs/dbt-core/blob/eea872c31994c90e995405c74496c1ce20e07025/core/dbt/clients/yaml_helper.py#L75-L88

If `path=None`, it has no `searched_path` / `relative_path`. If I change that to 

### Reproduction case

```yml
# dbt_project.yml
vars:
  one_var: foo
  one_var: bar
```
```
$ dbt parse
17:20:39  Encountered an error:
'NoneType' object has no attribute 'searched_path'
```

If I change the code linked above to:
```python
     except dbt.exceptions.DuplicateYamlKeyException as e: 
         # TODO: We may want to raise an exception instead of a warning in the future. 
         if path:
             e.msg = f"{e} {path.searched_path}/{path.relative_path}." 
         dbt.exceptions.warn_or_raise(e, log_fmt=warning_tag("{}")) 
```

```
$ dbt parse
17:21:36  [WARNING]: Compilation Error
  Duplicate 'one_var' key found in yaml file
17:21:36  Encountered an error while reading the project:
17:21:36    ERROR: Runtime Error
  dbt_project.yml does not parse to a dictionary
17:21:36  Encountered an error:
Runtime Error
  Could not run dbt
```

This is still a problem. This parses successfully in v1.1. We MUST make this a warning, not an error.

We can do that by using `SafeLoader` instead of `UniqueKeyLoader`, and still returning the loaded result (if not `--warn-error`):
```python
    except dbt.exceptions.DuplicateYamlKeyException as e:
        if path:
            # TODO: We may want to raise an exception instead of a warning in the future.
            e.msg = f"{e} {path.searched_path}/{path.relative_path}."
        dbt.exceptions.warn_or_raise(e, log_fmt=warning_tag("{}"))
        return yaml.load(contents, Loader=SafeLoader)
```

PR #5347, opened 2022-06-08
Fix YAML parsing with anchors and duplicate keys in dbt_project.yml file

### LINKED ISSUE #5268: [CT-663] YAML anchor expansion failure in 1.2.0a1 / `main`
Proactively identified regression in `1.2.0a1`. If I had to guess, it might be related to https://github.com/dbt-labs/dbt-core/pull/5146?

Here's a profile that uses [YAML anchors](https://support.atlassian.com/bitbucket-cloud/docs/yaml-anchors/) to DRY up duplicated code. This is valid YAML, per ["online YAML parser"](https://yaml-online-parser.appspot.com/?yaml=%23+profiles.yml%0Agarage-postgres%3A%0A++outputs%3A%0A++++dev%3A+%26garage-pg%0A++++++database%3A+jerco%0A++++++host%3A+localhost%0A++++++pass%3A+nopass%0A++++++port%3A+5432%0A++++++schema%3A+dbt_jcohen%0A++++++threads%3A+5%0A++++++type%3A+postgres%0A++++++user%3A+jerco%0A++++prod%3A%0A++++++%3C%3C%3A+*garage-pg%0A++++++schema%3A+analytics%0A++target%3A+dev&type=json):
```yml
# profiles.yml
garage-postgres:
  outputs:
    dev: &garage-pg
      database: jerco
      host: localhost
      pass: nopass
      port: 5432
      schema: dbt_jcohen
      threads: 5
      type: postgres
      user: jerco
    prod:
      <<: *garage-pg
      schema: analytics
  target: dev
```
When running v1.1.0, dbt is perfectly happy to expand the YAML anchor:
```
$ dbt debug
11:02:25  Running with dbt=1.1.0
dbt version: 1.1.0
python version: 3.9.12
python path: /Users/jerco/dev/scratch/testy/env/bin/python3.9
os info: macOS-12.3.1-x86_64-i386-64bit
Using profiles.yml file at /Users/jerco/.dbt/profiles.yml
Using dbt_project.yml file at /Users/jerco/dev/scratch/testy/dbt_project.yml

Configuration:
  dbt_project.yml file [OK found and valid]

Required dependencies:
 - git [OK found]

Connection:
  host: localhost
  port: 5432
  user: jerco
  database: jerco
  schema: dbt_jcohen
  search_path: None
  keepalives_idle: 0
  sslmode: None
  Connection test: [OK connection ok]

All checks passed!
```
Running `dbt-core` installed from `main`, it's not:
```
$ dbt debug
11:02:38  Running with dbt=1.2.0-a1
dbt version: 1.2.0-a1
python version: 3.9.12
python path: /Users/jerco/dev/product/dbt-core/env/bin/python3.9
os info: macOS-12.3.1-x86_64-i386-64bit
Using profiles.yml file at /Users/jerco/.dbt/profiles.yml
Using dbt_project.yml file at /Users/jerco/dev/scratch/testy/dbt_project.yml

11:02:39  Encountered an error:
Runtime Error

  dbt encountered an error while trying to read your profiles.yml file.

  Runtime Error
    Syntax error near line 55
    ------------------------------
    52 |       type: postgres
    53 |       user: jerco
    54 |     prod:
    55 |       <<: *garage-pg
    56 |       schema: analytics
    57 |   target: dev
    58 | sandbox-redshift:

    Raw Error:
    ------------------------------
    could not determine a constructor for the tag 'tag:yaml.org,2002:merge'
      in "<unicode string>", line 55, column 7
```

jeremyyeo:
resolves #5268, #5331 

### Description

Just pondering on a fix for the linked issues.

1. ~~Seems like `safe_load` is used in multiple places - so just revert to previous behaviour by default + option to use the newly introduced `UniqueKeyLoader`.~~
2. Looks like a call to `flatten_mapping` sorts out the YAML anchors.

This seems to work for:

```yaml
# ~/.dbt/profiles.yml

postgres:
  outputs:
    dev: &profile
      type: postgres
      host: localhost
      user: root
      password: password
      schema: public
      database: postgres
      port: 5432
      threads: 8
    prod:
      <<: *profile
    uat: *profile
  target: dev
```

and

```yaml
# dbt_project.yml

...
vars:
  foo: bar
  foo: bar
...
```

### Checklist

### DIFF
--- a/core/dbt/clients/yaml_helper.py
+++ b/core/dbt/clients/yaml_helper.py
@@ -31,6 +31,7 @@ class UniqueKeyLoader(SafeLoader):
 
     def construct_mapping(self, node, deep=False):
         mapping = set()
+        self.flatten_mapping(node)  # This processes yaml anchors / merge keys (<<).
         for key_node, value_node in node.value:
             key = self.construct_object(key_node, deep=deep)
             if key in mapping:
@@ -68,13 +69,16 @@ def contextualized_yaml_error(raw_contents, error):
     )
 
 
-def safe_load(contents) -> Optional[Dict[str, Any]]:
-    return yaml.load(contents, Loader=UniqueKeyLoader)
+def safe_load(contents, unique=False) -> Optional[Dict[str, Any]]:
+    if unique:
+        return yaml.load(contents, Loader=UniqueKeyLoader)
+    else:
+        return yaml.load(contents, Loader=SafeLoader)
 
 
 def load_yaml_text(contents, path=None):
     try:
-        return safe_load(contents)
+        return safe_load(contents, unique=True)
     except (yaml.scanner.ScannerError, yaml.YAMLError) as e:
         if hasattr(e, "problem_mark"):
             error = contextualized_yaml_error(contents, e)
@@ -84,5 +88,7 @@ def load_yaml_text(contents, path=None):
         raise dbt.exceptions.ValidationException(error)
     except dbt.exceptions.DuplicateYamlKeyException as e:
         # TODO: We may want to raise an exception instead of a warning in the future.
-        e.msg = f"{e} {path.searched_path}/{path.relative_path}."
+        if path:
+            e.msg = f"{e} {path.searched_path}/{path.relative_path}."
         dbt.exceptions.warn_or_raise(e, log_fmt=warning_tag("{}"))
+        return safe_load(contents)
--- a/core/dbt/clients/yaml_helper.py
+++ b/core/dbt/clients/yaml_helper.py
@@ -35,8 +35,9 @@ class UniqueKeyLoader(SafeLoader):
         for key_node, value_node in node.value:
             key = self.construct_object(key_node, deep=deep)
             if key in mapping:
-                raise dbt.exceptions.DuplicateYamlKeyException(
-                    f"Duplicate {key!r} key found in yaml file"
+                msg = f"Duplicate {key!r} key found in yaml file"
+                dbt.exceptions.warn_or_raise(
+                    dbt.exceptions.DuplicateYamlKeyException(msg), log_fmt=warning_tag("{}")
                 )
             mapping.add(key)
         return super().construct_mapping(node, deep)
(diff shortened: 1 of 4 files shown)


core/dbt/clients/yaml_helper.py
```diff
@@ -68,13 +69,16 @@ def contextualized_yaml_error(raw_contents, error):
     )
 
 
-def safe_load(contents) -> Optional[Dict[str, Any]]:
-    return yaml.load(contents, Loader=UniqueKeyLoader)
+def safe_load(contents, unique=False) -> Optional[Dict[str, Any]]:
+    if unique:
+        return yaml.load(contents, Loader=UniqueKeyLoader)
+    else:
+        return yaml.load(contents, Loader=SafeLoader)
```
jtcohen6, line 76: Question for @dbt-labs/core-language on how to factor this code: Should we opt for one method with a boolean argument like this, or for two different methods: `load_and_validate_unique_keys` + `safe_load`?

jeremyyeo:
Not sure what's up with the unit test but standard run with dupe vars keys worked okay

image

jtcohen6:
> Perhaps it's not such a bad idea to error as it was doing in the first place?

This wasn't an error in previous versions, so we can't raise an error now. The implicit behavior is to "take the last"; that's what it used to do, and what it will keep doing. It's a huge net improvement that we're raising a warning, so that users can become aware of the duplication/discrepancy and fix it themselves!

One other small aesthetic thing. It's a bit confusing to see:
```
06:12:47  [WARNING]: Compilation Error
```
Rather than raising an exception, which we then catch and raise as a warning within `load_yaml_text`, could `construct_mapping` just use `dbt.exceptions.warn_or_raise` to raise the warning message directly? I'm not sure exactly how to make that interact with the `try`/`except`, but we should really be falling back to real `SafeLoader` in all cases, and just using the `UniqueKeyLoader` for validation/warning.

core/dbt/clients/yaml_helper.py
gshank: Whether I prefer a 'unique' argument or a separate method name depends on why we're only actually using the 'unique' arg in 'load_yaml_text', but we call 'safe_load' in multiple other places. Do those other places not need the duplicate key validation? Why is it only called there?

jeremyyeo: That's a good point - I think we should be validating duplicate key everywhere.. so I reverted this to what it was before.

jeremyyeo:
1. Use `warn_or_raise` in `construct_mapping` instead. `warn_or_raise` calls the exception - not sure if this is the right way to do this since I couldn't find any other use of `warn_or_raise`.
2. Tidied up confusing `[WARNING]: Compilation Error` print outs.
3. Not too sure how to get the path of the file to be included in the warning message as before (when we had the `path` in the `load_yaml_text` function).

jtcohen6:
Is it fair to close this PR for now, since we ended up reverting the change in https://github.com/dbt-labs/dbt-core/pull/5146?

PR #5350, opened 2022-06-09
fix build test logic

### LINKED ISSUE #5277: [CT-671] [Bug] Failing singular test with two dependent models a&b doesn't block a downstream model c unless it selects from both a&b

### Current Behavior

- Singular `test_ab` selects from `model_a` and `model_b`.
- `model_c` selects from `model_a` only
- when running `dbt build -s a+ b+`, `test_ab` doesn't run before `model_c`

### Expected Behavior

A singular test with multiple dependencies should run before any models downstream of those dependencies.

### Steps To Reproduce

```sql
-- models/model_a.sql
select 1 as foo
```

```sql
-- models/model_b.sql
select 1 as foo
```

```sql
-- tests/test_ab.sql
select 1 as foo
-- {{ ref('model_a') }}
-- {{ ref('model_b') }}
```

```sql
-- models/model_c.sql
select 1 as foo
-- {{ ref('model_a') }}
```

```
$ dbt build -s a+ b+
03:44:55  Running with dbt=1.1.0
03:44:56  Found 3 models, 1 test, 0 snapshots, 0 analyses, 0 macros, 0 operations, 0 seed files, 0 sources, 0 exposures, 0 metrics
03:44:56  
03:45:04  Concurrency: 8 threads (target='dev')
03:45:04  
03:45:04  1 of 4 START view model schema.model_a ............................. [RUN]
03:45:04  2 of 4 START view model schema.model_b ............................. [RUN]
03:45:08  1 of 4 OK created view model schema.model_a ........................ [SUCCESS 1 in 3.52s]
03:45:08  3 of 4 START view model schema.model_c ............................. [RUN]
03:45:10  2 of 4 OK created view model schema.model_b ........................ [SUCCESS 1 in 5.22s]
03:45:10  4 of 4 START test test_ab ...................................................... [RUN]
03:45:11  3 of 4 OK created view models schema.model_c ........................ [SUCCESS 1 in 3.17s]
03:45:12  4 of 4 FAIL 1 test_ab .......................................................... [FAIL 1 in 2.35s]
03:45:12  
03:45:12  Finished running 3 view models, 1 test in 15.79s.
03:45:12  
03:45:12  Completed with 1 error and 0 warnings:
03:45:12  
03:45:12  Failure in test test_ab (tests/test_ab.sql)
03:45:12    Got 1 result, configured to fail if != 0
03:45:12  
03:45:12    compiled SQL at target/compiled/project_name/tests/test_ab.sql
03:45:12  
03:45:12  Done. PASS=3 WARN=0 ERROR=1 SKIP=0 TOTAL=4
```

Either dropping `test_ab`s dependency on `model_b` or adding a dependency from `model_c` to `model_b` causes the test to block `model_c` as expected.

### Relevant log output

_No response_

### Environment

```markdown
- OS: MacOS 12.3.1
- Python: 3.8.10
- dbt: 1.1.0
```

snowflake

### Additional Context

Reproduced from someone I was helping in [this slack thread](https://getdbt.slack.com/archives/C50NEBJGG/p1652987147658199)

ChenyuLInx:
resolves #5277

### Description
Keeping original case and also fix situation in #5277 

### Checklist

### DIFF
--- a/core/dbt/compilation.py
+++ b/core/dbt/compilation.py
@@ -410,14 +410,19 @@ class Compiler:
         for metric in manifest.metrics.values():
             self.link_node(linker, metric, manifest)
 
-        cycle = linker.find_cycles()
-
-        if cycle:
-            raise RuntimeError("Found a cycle: {}".format(cycle))
+        self.check_cycles(linker)
 
         if add_test_edges:
             manifest.build_parent_and_child_maps()
             self.add_test_edges(linker, manifest)
+
+        self.check_cycles(linker)
+
+    @staticmethod
+    def check_cycles(linker: Linker):
+        cycle = linker.find_cycles()
+        if cycle:
+            raise RuntimeError("Found a cycle: {}".format(cycle))
 
     def add_test_edges(self, linker: Linker, manifest: Manifest) -> None:
         """This method adds additional edges to the DAG. For a given non-test
@@ -465,10 +470,16 @@ class Compiler:
                     # node(s) it depends on.
                     test_depends_on = set(manifest.nodes[upstream_test].depends_on_nodes)
 
+                    all_depends_on_node = set()
+                    for depends_on_node in test_depends_on:
+                        all_depends_on_node.update(
+                            nx.traversal.bfs_tree(linker.graph, depends_on_node, reverse=True)
+                        )
+
                     # If the set of nodes that an upstream test depends on
                     # is a subset of all upstream nodes of the current node,
                     # add an edge from the upstream test to the current node.
-                    if test_depends_on.issubset(upstream_nodes):
+                    if node_id not in all_depends_on_node:
                         linker.graph.add_edge(upstream_test, node_id)
 
     def compile(self, manifest: Manifest, write=True, add_test_edges=False) -> Graph:


core/dbt/compilation.py
```diff
@@ -465,10 +470,16 @@ def add_test_edges(self, linker: Linker, manifest: Manifest) -> None:
                     # node(s) it depends on.
                     test_depends_on = set(manifest.nodes[upstream_test].depends_on_nodes)
 
+                    all_depends_on_node = set()
+                    for depends_on_node in test_depends_on:
```
ChenyuLInx, line 474: This might be expensive to run

core/dbt/compilation.py
```diff
@@ -465,10 +470,16 @@ def add_test_edges(self, linker: Linker, manifest: Manifest) -> None:
...
+                        all_depends_on_node.update(
+                            nx.traversal.bfs_tree(linker.graph, depends_on_node, reverse=True)
+                        )
+
                     # If the set of nodes that an upstream test depends on
```
ChenyuLInx, line 479: update comment

core/dbt/compilation.py
```diff
@@ -465,10 +470,16 @@ def add_test_edges(self, linker: Linker, manifest: Manifest) -> None:
                     # node(s) it depends on.
```
ChenyuLInx, line 470: update comment

core/dbt/compilation.py
jtcohen6: Good callout. This is worth testing with a very big project (e.g. [this one](https://github.com/dbt-labs/dbt-core/tree/main/performance/projects/01_2000_simple_models)). `dbt build` is already much slower during graph compilation than other tasks, because it alone adds extra test edges (this method).

jtcohen6: I just tested locally using that perf project, and found that this method takes:
- 3 seconds to run on current `main`
- multiple minutes with the change on this branch

gshank: When we had a similar performance issue earlier (also looping over everything for every node) we switched to using the pre-built dag (parent_map, child_map). Maybe we could build a specialized dag including tests?

ChenyuLInx:
```python
                ### Solution 2
                # Get all tests that depend on any upstream nodes.
                # this has issue with out test big project and will generate loop
                upstream_tests = []
                for upstream_node in upstream_nodes:
                    upstream_tests += _get_tests_for_node(manifest, upstream_node)

                for upstream_test in upstream_tests:

                    test_depends_on = manifest.nodes[upstream_test].depends_on_nodes
                    down_stream_nodes = set([node for node in manifest.child_map[node_id]])
                    # if node_id not in test_depends_on:
                    if not any(x in down_stream_nodes for x in test_depends_on) and node_id not in test_depends_on:
                        linker.graph.add_edge(upstream_test, node_id)
                    # try:
                    #     self.check_cycles(linker)
                    # except:
                    #     breakpoint()
                    # if test_depends_on.issubset(upstream_nodes):
                    #     linker.graph.add_edge(upstream_test, node_id)
```

There are situations described in the ticket that will cause this to generate cycle in the end.

core/dbt/compilation.py
ChenyuLInx: Ha mine doesn't even finish on it. And switching to using the (parent_map, child_map) is still slow. I tried out limiting the bfs search depth to be 1 and found that speeds things up a lot on big projects and still is correct.
Though in the original issue there is still a situation I think makes leaving current behavior desirable.

PR #5359, opened 2022-06-10
Initial refactoring of incremental materialization

### LINKED ISSUE #5245: [CT-646] [Proposal] Streamline Incremental Strategies
When defining an incremental model, there are several incremental strategies that could be used to execute the incremental model, (namely: "append," "delete+insert," "merge," and "insert_overwrite.") but not all of them are supported by every warehouse. Working with these incremental strategies as an adapter maintainer and as an advanced user is difficult. This ticket aims to improve the experience for both of these groups of people.

Functional Requirements
1. Adapter maintainers need to be able to specify which incremental strategies are supported by the warehouse and be able to supply the correct sql statement to execute if the default is not sufficient.
2. Advanced users that wish to specify a custom incremental strategy must be able to do so.

Behavior Today
1. To specify which incremental strategies are supported by the warehouse, the adapter maintainer must write a macro that returns a list of strings that name the supported strategies, and include the jinja macro `adapter__get_merge_sql` that takes a fixed set of parameters and returns the materialization sql. This macro is often copy-pasted from [core](https://github.com/dbt-labs/dbt-core/blob/ea5a9da71ef71cc13414cd8c2f8097448add5edb/core/dbt/include/global_project/macros/materializations/models/incremental/merge.sql#L5-L48), and is error prone when editing.
2. Advanced dbt users who wish to specify a custom incremental strategy must override the same boiler plate jinja macro by copy pasting it into their dbt project.
3. When an unsupported incremental strategy is specified, the user is notified via an exception at during the first `dbt run`.

Desired Behavior
1. To specify which incremental strategies are supported by the warehouse, the adapter maintainer must write a macro that returns a list of strings that name the supported strategies. If the warehouse only supports a subset of the default incremental strategies and the default implementation in the global project produces sql that works on the target warehouse, no additional work needs to be done. If any of the default behavior needs to be overridden, it can be done by writing a macro that conforms to the naming convention "get_incremental_NAME_sql" where "NAME" is the string used to configure the incremental strategy (e.g. "get incremental_append_sql"). If the warehouse has an additional incremental strategy, defining a new macro using the same "get_incremental_NAME_sql" convention will make it available to users of the adapter.
2. Advanced dbt users who wish to specify a custom incremental strategy will only need to create a macro that conforms to the naming convention "get_incremental_NAME_sql" that produces the correct sql for the target warehouse.
3. When an unsupported incremental strategy is designated, raise an error to users before runtime.
4. Existing projects will experience no breaking changes when they upgrade to the version of dbt that includes this new behavior.

Implementation
- [ ] For backwards compatibility, create new macros in `core/dbt/include/global_project/macros/materializations/models/incremental` that follow the naming convention for the existing incremental strategies and pass through to the existing current macros. The new macros will take a dictionary, and pass the key-value pairs explicitly to the wrapped macro call to return the sql string that executes the materialization. Concretely:
    - [ ] create a new macro `get_insert_into_sql() -> str` that returns `"insert into"` 
    - [ ] create a new macro `get_incremental_append_sql(dict)` that wraps `get_insert_into_sql`
    - [ ] create a new macro `get_incremental_delete_insert_sql(dict)` that wraps `get_delete_insert_merge_sql`
    - [ ] create a new macro `get_incremental_merge_sql(dict)` that wraps `get_merge_sql`
    - [ ] create a new macro `get_incremental_insert_overwrite_sql(dict)` that wraps `get_insert_overwrite_sql`
    - [ ] create a new macro `get_incremental_default_sql(dict)` that wraps `get_incremental_append_sql`. This is where adapters can override the default incremental strategy when append isn't appropriate.
    - Note: this dictionary will always pass the same key-value pairs from core. The dictionary value allows core to add new key-value pairs in the future without breaking existing adapters. It is very possible that a custom incremental strategy may require different parameters than the ones core provides. These can be added by users via context-global i.e. `config.get('...')`.
- [ ] Add `incremental_strategy` to `NodeConfig` with a default value of `"default"`.
- [ ] Create a new Python method on `core/dbt/adapters/sql/imple.py::AdapterConfig` called `validate_incremental_strategy` that takes no parameters and returns a list of strings. Adapter maintainers are expected to implement this so that the returned strings name the supported incremental strategies. This could have been a macro, but it's much easier for core to call this as a python function to validate incremental strategies in projects. (TODO what if an adapter supports a different set of incremental strategies for sql and python models? where would they indicate that?)
- [ ] Write a new Python function `get_incremental_strategy_macro` that takes in a string and returns the macro from the context. (e.g. "insert_overwrite" -> callable get_incremental_insert_overwrite_sql). This function will call `validate_incremental_strategy` and throw an exception if the requested strategy is not returned from `validate_incremental_strategy` or if it does not exist in the context. This function will need to be called in the [materialization macro for models](https://github.com/dbt-labs/dbt-core/blob/main/core/dbt/include/global_project/macros/materializations/models/incremental/incremental.sql)
- [] create a "functional" test for each incremental strategy in `tests/adapter` including tests for several configs for each strategy. This will allow each adapter to import the tests it needs as well as make their own as needed. (Thanks for pointing this requirement out, @dataders)

Docs
This will require a significant change to the documentation on incremental materializations.

gshank:
resolves #5245

### Description

This is not ready for final review, but I'm putting it out here for discussion and commenting.

### Checklist


ChenyuLInx (commented):
This change looks good to me. Love the fact that we are breaking things down into smaller parts for folks to implement and adding more guardrails.
With this kind of change, do we think we can actually try to have smaller tests focusing on certain macros vs having one functional test testing the whole function?

core/dbt/include/global_project/macros/materializations/models/incremental/strategies.sql
```diff
@@ -0,0 +1,83 @@
+{% macro get_incremental_append_sql(arg_dict) %}
+
+  {{ return(adapter.dispatch('get_incremental_append_sql', 'dbt')(arg_dict)) }}
+
+{% endmacro %}
+
+
+{% macro default__get_incremental_append_sql(arg_dict) %}
+
+  {% do return(get_insert_into_sql(arg_dict["target_relation"], arg_dict["temp_relation"], arg_dict["dest_columns"])) %}
```
jtcohen6, line 10: I didn't get this at first, but makes sense that we need to "unpack" the arguments to preserve the existing signature of the `get_insert_into_sql` macro (for backwards compatibility, for anyone who's overridden/reimplemented it).

core/dbt/include/global_project/macros/materializations/models/incremental/strategies.sql
```diff
@@ -0,0 +1,83 @@
...
+  {% if arg_dict["unique_key"] %}
+    {% do return(get_incremental_delete_insert_sql(arg_dict)) %}
+  {% else %}
+    {% do return(get_incremental_append_sql(arg_dict)) %}
+  {% endif %}
+
+{% endmacro %}
+
+
+{% macro get_insert_into_sql(target_relation, temp_relation, dest_columns) %}
```
jtcohen6, line 73: Explaining to myself out loud: No need to dispatch this one, since it's really just here for backwards compatibility, and we'll be asking users / adapter maintainers to implement `<adapter>__get_incremental_append_sql` instead. (That one is dispatched.)

We will need to remove `get_insert_into_sql` from `dbt-spark`.

core/dbt/include/global_project/macros/materializations/models/incremental/strategies.sql
```diff
@@ -0,0 +1,83 @@
...
+
+  {{ return(adapter.dispatch('get_incremental_merge_sql', 'dbt')(arg_dict)) }}
+
+{% endmacro %}
+
+{% macro default__get_incremental_merge_sql(arg_dict) %}
+
+  {% do return(get_merge_sql(arg_dict["target_relation"], arg_dict["temp_relation"], arg_dict["unique_key"], arg_dict["dest_columns"])) %}
+
+{% endmacro %}
```
jtcohen6, line 40: I'm not sure exactly how we want to factor this going forward when add new capabilities to these macros, leveraging the extensibility of `arg_dict`. E.g. If we wanted to support `predicates` as a user-provided config for the `merge` strategy, which templates out additional conditions for the resulting `MERGE` statement —  would we modify the signature of `get_merge_sql`? Pass in optional `**kwargs`, so that it works whether or not the `predicates` argument is provided?

core/dbt/adapters/base/impl.py
```diff
@@ -1125,6 +1125,37 @@ def get_rows_different_sql(
...
+        from dbt.context.providers import generate_runtime_macro_context
+
+        macro_context = generate_runtime_macro_context(
+            macro=macro,
+            config=self.config,
+            manifest=manifest,  # type: ignore[arg-type]
+            package_name=self.config.project_name,
+        )
+        macro_function = MacroGenerator(macro, macro_context)
+        return macro_function
```
jtcohen6, line 1167: This code feels rough for us, but on the bright side, we're the only ones that have to deal with it. It's nicely cordoned off from user space. Up to us to decide whether this approach feels cleaner than the macro discovery/lookup/"dispatch" that happens for snapshots today ([here](https://github.com/dbt-labs/dbt-core/blob/5b1bc72ae11a36031b88d4d26a4ead3be87b3e50/core/dbt/include/global_project/macros/materializations/snapshots/strategies.sql#L1-L32)).

core/dbt/adapters/base/impl.py
```diff
@@ -1125,6 +1125,37 @@ def get_rows_different_sql(
 
         return sql
 
+    def validate_incremental_strategy(self):
+        return ["append", "delete+insert"]
```
jtcohen6: Lots of thoughts here — thanks for pushing forward the conversation!

- I think this should be `append` by default—the truly common denominator—but then Postgres (`plugins/postgres/dbt/adapters/posrtgres/impl.py`) should reimplement as both `append` + `delete+insert`
- Where does this validation get called? Within `get_incremental_strategy_macro`?
- This works great for adapter maintainers to override. As an average user, though, how would I add my new strategy `silly`? Should our presumption be that, if I've defined a macro named `get_incremental_silly_sql` (i.e. in a non-internal package), I want to use it? That would work, but feels like tricky implicit behavior

core/dbt/include/global_project/macros/materializations/models/incremental/strategies.sql
```diff
@@ -0,0 +1,83 @@
...
+
+{% macro default__get_incremental_default_sql(arg_dict) %}
+
+  {% if arg_dict["unique_key"] %}
+    {% do return(get_incremental_delete_insert_sql(arg_dict)) %}
+  {% else %}
+    {% do return(get_incremental_append_sql(arg_dict)) %}
+  {% endif %}
+
+{% endmacro %}
```
jtcohen6, line 70: This is clever. I like it!

- We want similar behavior for `dbt-snowflake` + `dbt-bigquery`, though using `merge` instead of `delete+insert` if a `unique_key` is specified
- I think this should be the behavior for `dbt-postgres` (+ `dbt-redshift`), but I'm not sure it should be the true default (global project) behavior.
- I think `append` is more realistic as universal default. That's also the default we want for `dbt-spark`

tests/functional/artifacts/expected_manifest.py
```diff
@@ -31,6 +31,7 @@ def get_rendered_model_config(**updates):
         "meta": {},
         "unique_key": None,
         "grants": {},
+        "incremental_strategy": None,
```
jtcohen6, line 34: Should this be `None` by default, or should it be `"default"`? Any reason to prefer that logic in the adapter method (`get_incremental_strategy_macro`) versus here?

core/dbt/include/global_project/macros/materializations/models/incremental/merge.sql
```diff
@@ -26,7 +26,7 @@
         {% do predicates.append('FALSE') %}
     {% endif %}
 
-    {{ sql_header if sql_header is not none }}
+    {# sql_header if sql_header is not none #}
```
jtcohen6, line 29: (I assume this is commented out while we're working through how to plumb args/configs through)

core/dbt/include/global_project/macros/materializations/models/incremental/merge.sql
```diff
@@ -86,16 +86,16 @@
...
 {%- endmacro %}
 
-{% macro default__get_insert_overwrite_merge_sql(target, source, dest_columns, predicates, include_sql_header) -%}
+{% macro default__get_insert_overwrite_merge_sql(target, source, dest_columns, predicates) -%}
     {%- set predicates = [] if predicates is none else [] + predicates -%}
     {%- set dest_cols_csv = get_quoted_csv(dest_columns | map(attribute="name")) -%}
     {%- set sql_header = config.get('sql_header', none) -%}
 
-    {{ sql_header if sql_header is not none and include_sql_header }}
+    {{ sql_header if sql_header is not none }}
```
jtcohen6, line 98: `include_sql_header` is indeed very very awkward. The reason we need it is that `dbt-bigquery` wants the `sql_header` in slightly different places for ["static"](https://github.com/dbt-labs/dbt-bigquery/blob/1b85fc50287ac9afd81fd92e715dffa5f0757379/dbt/include/bigquery/macros/materializations/incremental.sql#L49) versus ["dynamic"](https://github.com/dbt-labs/dbt-bigquery/blob/1b85fc50287ac9afd81fd92e715dffa5f0757379/dbt/include/bigquery/macros/materializations/incremental.sql#L83-L88) `insert_overwrite`, though I'm having a hard time remembering exactly why that is.

jtcohen6 (commented):
Great work @gshank! Thanks for throwing up the draft PR, to push forward the conversation. I'm excited by the progress here. I left a bunch of comments, some of which are just "oh I get this now," others being open questions that we should look to answer before merging.

core/dbt/include/global_project/macros/materializations/models/incremental/strategies.sql
gshank: Yes. I am thinking of switching to using keyword arguments instead of positional though.

core/dbt/include/global_project/macros/materializations/models/incremental/strategies.sql
gshank: I've been wondering about this too. It almost feels easier for people to add new extra config that can be pulled out in the user incremental strategy. Because the creation of that arg_dict happens in the materialization macro, so new arguments would require modifying the materialization macro, which seems like the wrong direction.  I did include predicates (from an extra 'incremental_predicates' config) in the arg_dict.

tests/functional/artifacts/expected_manifest.py
gshank: If we make the default 'default' here it shows up in every node, even those that aren't incremental. That felt like a waste to me.  If we can come up with a way to make configs more pluggable it would be less of an issue.

core/dbt/adapters/base/impl.py
jtcohen6: If it only validates the "builtin"s, and trusts the user for anything else, I'm cool with that! As long as there's a clear error message if they've specified `incremental_strategy: 'silly'`, and we can't find a macro named `get_incremental_silly_sql`

core/dbt/include/global_project/macros/materializations/models/incremental/strategies.sql
jtcohen6: I 100% agree that should keep being the behavior for `dbt-postgres`. I'm wondering if what's above should be `default__get_incremental_default_sql` — or if that logic should live in `postgres__get_incremental_default_sql`, with the "true" default just:
```sql
{% macro default__get_incremental_default_sql(arg_dict) %}
    {% do return(get_incremental_append_sql(arg_dict)) %}
{% endmacro %}
```
Reason being: I'd like to see us move into a world where `default` != `postgres`, at least not exactly, even if PostgreSQL is still the reference implementation for most desirable behaviors.

Many many analytical databases don't support `delete+insert`, and I'm not sure we should take it as the high water-mark to which they should aspire. An adapter that _just_ supports `append` is pretty uninspiring, but it is the thing we test for today in the "basic" incremental test: [`dbt/tests/adapter/basic/test_incremental.py`](https://github.com/dbt-labs/dbt-core/blob/main/tests/adapter/dbt/tests/adapter/basic/test_incremental.py)

core/dbt/include/global_project/macros/materializations/models/incremental/merge.sql
gshank: Yes.

core/dbt/include/global_project/macros/materializations/models/incremental/merge.sql
jtcohen6: Yup. Let me do a little digging, and try to follow up with an explanation of why it has to be that way :)

core/dbt/include/global_project/macros/materializations/models/incremental/strategies.sql
gshank: We could do a "by convention" thing where we pull in all configs that start with 'incremental_strategy' or something. It does feel a bit hacky, but I haven't come up with a solution that's better yet.

gshank: Or have an incremental_strategy dictionary with extra args

core/dbt/include/global_project/macros/materializations/models/incremental/merge.sql
gshank: @jtcohen6 Any more info on the need for the sql header in BigQuery?

gshank:
@jtcohen6 There's a difference in the way the config.get('...', default='...') works when the config key exists (built-in or "extra") and if it doesn't. The materialization code is doing config.get('incremental_strategy', default='merge') but it only sets incremental_strategy to 'merge' if the config key (and the extra key) doesn't exist. So we're not actually getting the default set from the macro config.get calls because we now have an 'incremental_strategy' config attribute.

I almost want to check for a None value to set a default, but that's a pretty big change, so I'm guessing that's not a good idea. We could change the lines to {%- set strategy = config.get("incremental_strategy") or "merge" %}

In the new code in 'get_incremental_strategy_macro' I set the strategy to 'default' if it's none, but if we're only doing the minimal changes to the adapters (not including using that bit) we'll have to do something else.

gshank:
I'm wondering if putting the args in the strategy_arg_dict actually serves any real purpose? It's not clear to me what it gives us that simply passing in the args doesn't. If users have a custom materialization, unless they copy-and-paste the whole incremental materialization, I assume they will be pulling the in any new args from config in their custom code, right?

It's still not clear to me why we want the sql_header in bigquery and only in bigquery. As far as I can tell, none of the other adapters include the sql_header, since "include_sql_header" is set to false everywhere except that one place in bigquery.

I think that in order to use the default__get_incremental_default_sql in the previous comment, we'd have to pretty much implement all the changes in this pull request.

Redshift works fine with no changes on top of dbt-core. Snowflake works with just the tweak to the config.get of the incremental_strategy. Spark works with the tweak to the config.get call. BigQuery works, except that for testing purposes I removed the sql_header thing and everything else works fine except for the test that explicitly checks for the sql header.

gshank:
For BigQuery, maybe we could set an extra config? It would be nice to not have to pass around an extra parameter just for that one case.

As far as using default__get_incremental_default_sql, the code just isn't set up such that we can just insert that one bit. I look at putting it in and it just doesn't fit. We'd need to pull in almost all of the new macros. I can try doing that... What the timeframe here?

plugins/postgres/dbt/include/postgres/macros/materializations/incremental_strategies.sql
```diff
@@ -0,0 +1,9 @@
+{% macro postgres__get_incremental_default_sql(arg_dict) %}
+
+  {% if arg_dict["unique_key"] %}
+    {% do return(get_incremental_delete_insert_sql(arg_dict)) %}
+  {% else %}
+    {% do return(get_incremental_append_sql(arg_dict)) %}
+  {% endif %}
+
+{% endmacro %}
```
jtcohen6, line 9: Love to see this!!

core/dbt/include/global_project/macros/materializations/models/incremental/strategies.sql
```diff
@@ -0,0 +1,79 @@
...
+
+  {{ return(adapter.dispatch('get_incremental_default_sql', 'dbt')(arg_dict)) }}
+
+{% endmacro %}
+
+{% macro default__get_incremental_default_sql(arg_dict) %}
+
+  {% do return(get_incremental_append_sql(arg_dict)) %}
+
+{% endmacro %}
```
jtcohen6, line 66: This represents an implicit breaking change for maintainers of existing adapter plugins who use the default incremental materialization:
- Previously, the (implicit) default was `delete+insert`
- Now, the explicit default is `append`

I think it's a good change! It's just one we'll want to document very clearly

(cc @dataders)

core/dbt/include/global_project/macros/materializations/models/incremental/merge.sql
```diff
@@ -86,14 +86,16 @@
...
+{% macro get_insert_overwrite_merge_sql(target, source, dest_columns, predicates) -%}
+  {{ adapter.dispatch('get_insert_overwrite_merge_sql', 'dbt')(target, source, dest_columns, predicates) }}
 {%- endmacro %}
 
-{% macro default__get_insert_overwrite_merge_sql(target, source, dest_columns, predicates, include_sql_header) -%}
+{% macro default__get_insert_overwrite_merge_sql(target, source, dest_columns, predicates) -%}
     {%- set predicates = [] if predicates is none else [] + predicates -%}
     {%- set dest_cols_csv = get_quoted_csv(dest_columns | map(attribute="name")) -%}
     {%- set sql_header = config.get('sql_header', none) -%}
+    {# The following is only true in BigQuery #}
```
jtcohen6, line 97: I like the way you did this!

Maybe even more explicit in the code annotation:
```
    {#-- The only time include_sql_header = True: --#}
    {#-- BigQuery + insert_overwrite strategy + "static" partitions config --#}
```

core/dbt/adapters/base/impl.py
```diff
@@ -1162,6 +1162,47 @@ def get_rows_different_sql(
...
         return sql
 
+    def valid_incremental_strategies(self):
+        return ["append", "delete+insert"]
+
+    @available.parse_none
+    def get_incremental_strategy_macro(self, strategy: str):
+        # Construct macro_name from strategy name
+        if strategy is None:
+            strategy = "default"
```
jtcohen6, line 1172: 👍

core/dbt/adapters/base/impl.py
```diff
@@ -1162,6 +1162,47 @@ def get_rows_different_sql(
...
+
+        # validate strategies for this adapter
+        valid_strategies = self.valid_incremental_strategies()
+        valid_strategies.append("default")
+        builtin_strategies = ["append", "delete+insert", "merge", "insert_overwrite"]
+        if strategy in builtin_strategies and strategy not in valid_strategies:
```
jtcohen6, line 1178:
```python
if strategy in builtin_strategies
```
👍 I see, this is how we avoid validating custom strategies defined by _end users_, e.g. `silly` + `get_silly_incremental_sql`

What do you think about moving `builtin_strategies` into its own method? That way, if an external adapter plugin maintainer defines their own strategy (e.g. `special_athena_upsert`), they can update both methods (`valid_incremental_strategies` + `builtin_incremental_strategies`), and take advantage of this validation — without having to copy-paste-edit all of `get_incremental_strategy_macro` just to change one line.

core/dbt/adapters/base/impl.py
```diff
@@ -1162,6 +1162,47 @@ def get_rows_different_sql(
 
         return sql
 
+    def valid_incremental_strategies(self):
+        return ["append", "delete+insert"]
```
jtcohen6, line 1166: I _think_ this should just be `"append"` by default, since that's "true neutral".

In Postgres, we should override it to what's here:
```python
    def valid_incremental_strategies(self):
        return ["append", "delete+insert"]
```
Let's also add a docstring to the effect of: "The set of standard builtin strategies which this adapter supports out-of-the-box. Not used to validate custom strategies defined by end users."

(cc @dataders)

core/dbt/adapters/base/impl.py
```diff
@@ -1162,6 +1162,47 @@ def get_rows_different_sql(
...
+                'dbt could not find an incremental strategy macro with the name "{}" in {}'.format(
+                    macro_name, self.config.project_name
+                )
+            )
+
+        from dbt.context.providers import generate_runtime_macro_context
+
+        macro_context = generate_runtime_macro_context(
+            macro=macro,
+            config=self.config,
```
jtcohen6, line 1199: [BLOCKING] Of all my comments, I think this is the only one we _must_ address before merging.

I believe this is causing the test failure in https://github.com/dbt-labs/dbt-snowflake/pull/196, because `config.get('merge_update_columns')` here is returning `None`:

https://github.com/dbt-labs/dbt-core/blob/e9e811eb99e736c7028f4d52db4b9da8c860e17a/core/dbt/include/global_project/macros/materializations/models/incremental/merge.sql#L8

Why? The `config` is different! I think this line, in `generate_runtime_macro_context`, may have the effect of setting a _different_ `config`. I dropped [our Jinja debugger](https://docs.getdbt.com/reference/dbt-jinja-functions/debug-method) into that line above, and saw this:
```
ipdb> l_1_config
<dbt.context.providers.RuntimeConfigObject object at 0x13b10c4f0>
ipdb> l_1_config.get("merge_update_columns")
ipdb> l_1_config.__dict__
{'model': ParsedMacro(unique_id='macro.dbt.get_incremental_default_sql', package_name='dbt', root_path='/Users/jerco/dev/product/dbt-core/core/dbt/include/global_project', path='macros/materializations/models/incremental/strategies.sql', original_file_path='macros/materializations/models/incremental/strategies.sql', name='get_incremental_default_sql', macro_sql="{% macro get_incremental_default_sql(arg_dict) %}\n\n  {{ return(adapter.dispatch('get_incremental_default_sql', 'dbt')(arg_dict)) }}\n\n{% endmacro %}", resource_type=<NodeType.Macro: 'macro'>, tags=[], depends_on=MacroDependsOn(macros=['macro.dbt_snowflake.snowflake__get_incremental_default_sql']), description='', meta={}, docs=Docs(show=True), patch_path=None, arguments=[], created_at=1658322950.229271)}
```

Whereas this is the value of `config` within the incremental materialization:
```
ipdb> l_1_config
<dbt.context.providers.RuntimeConfigObject object at 0x1382a6d30>
ipdb> l_1_config.get("merge_update_columns")
['email', 'ip_address']
ipdb> l_1_config.__dict__
{'model': CompiledModelNode(raw_sql='{{\n    config(\n        materialized = "incremental",\n        unique_key = "id",\n        merge_update_columns = ["email", "ip_address"]\n    )\n}}\n\n\nselect *\nfrom {{ ref(\'seed\') }}\n\n{% if is_incremental() %}\n\n    where load_date > (select max(load_date) from {{this}})\n\n{% endif %}', compiled=True, database='DBT_TEST', schema='TEST16583249731264382551_SIMPLE_COPY', fqn=['test', 'incremental_update_cols'], unique_id='model.test.incremental_update_cols', package_name='test', root_path='/private/var/folders/7h/hj5_fw9j291c58hwfdvy5xbm0000gp/T/dbt-int-test-frmfwd6h', path='incremental_update_cols.sql', original_file_path='models-merge-update/incremental_update_cols.sql', name='incremental_update_cols', resource_type=<NodeType.Model: 'model'>, alias='incremental_update_cols', checksum=FileHash(name='sha256', checksum='e16fc1080522481ea1692289e54b2c30a243084f9e5019e5f6fc05d05c83800b'), config=NodeConfig(_extra={'merge_update_columns': ['email', 'ip_address']}, enabled=True, alias=None, schema=None, database=None, tags=[], meta={}, materialized='incremental', incremental_strategy=None, persist_docs={}, post_hook=[], pre_hook=[], quoting={}, column_types={}, full_refresh=None, unique_key='id', on_schema_change='ignore', grants={}), _event_status={'node_status': <RunningStatus.Executing: 'executing'>}, tags=[], refs=[['seed']], sources=[], metrics=[], depends_on=DependsOn(macros=['macro.dbt.is_incremental', 'macro.dbt_snowflake.set_query_tag', 'macro.dbt.should_full_refresh', 'macro.dbt.load_relation', 'macro.dbt.make_temp_relation', 'macro.dbt.incremental_validate_on_schema_change', 'macro.dbt.run_hooks', 'macro.dbt.create_table_as', 'macro.dbt.run_query', 'macro.dbt.process_schema_changes'], nodes=['seed.test.seed']), description='', columns={}, meta={}, docs=Docs(show=True), patch_path=None, compiled_path='target/compiled/test/models-merge-update/incremental_update_cols.sql', build_path=None, deferred=False, unrendered_config={'materialized': 'incremental', 'unique_key': 'id', 'merge_update_columns': ['email', 'ip_address']}, created_at=1658324991.3003168, config_call_dict={'materialized': 'incremental', 'unique_key': 'id', 'merge_update_columns': ['email', 'ip_address']}, compiled_sql='\n\n\nselect *\nfrom DBT_TEST.TEST16583249731264382551_SIMPLE_COPY.seed\n\n\n\n    where load_date > (select max(load_date) from DBT_TEST.TEST16583249731264382551_SIMPLE_COPY.incremental_update_cols)\n\n', extra_ctes_injected=True, extra_ctes=[], relation_name='DBT_TEST.TEST16583249731264382551_SIMPLE_COPY.incremental_update_cols', _pre_injected_sql=None)}
```

Two options come to mind:
1. Pass the `config` in as a second argument to `adapter.get_incremental_strategy_macro()`, so that it can also be passed in here as well (instead of using `self.config`). Or somehow find a way to use the same model context from the materialization, to resolve these macros. This feels finicky to me, and I wasn't able to get it working after a very quick attempt, but it might be doable.
2. Pass in `merge_update_columns`, and `sql_header`, and any other "grab-from-global-state" configs as a keyword arguments to the macro (`get_merge_sql`). That's actually the pattern we want, ultimately! I'm just not sure if there's a way to do this while preserving backwards compatibility for the macro signature. It also removes the ability for users to define a new config in the model, and _just_ override the one macro they need (`get_merge_sql`), without copy-paste-editing the whole incremental materialization.

jtcohen6 (changes requested):
Feels like we're very close! One blocking/functional comment, and a few nit picks around the edges. In all, though, I'm excited to get this foundation laid, so we can reap the fruits of it in adapter repos. Thanks as well for pushing that work a bit further in `dbt-snowflake`, as demonstration.

Looking forward to talking through it all in a few minutes

core/dbt/include/global_project/macros/materializations/models/incremental/merge.sql
gshank: Done!

core/dbt/adapters/base/impl.py
gshank: I've added a builtin_incremental_strategies method.

core/dbt/adapters/base/impl.py
gshank: Done!

core/dbt/adapters/base/impl.py
gshank: I'm passing in the context on the call to get_incremental_strategy_macro. The python code can just return the callable macro from the context.

jtcohen6: @gshank It works!! Nice one :)

core/dbt/adapters/base/impl.py
jtcohen6: Thanks for adding the docstring! Going to be super helpful to maintainers of other adapters.

Could we update this to just `"append"` (default), and then override to `["append", "delete+insert"]` for Postgres?

core/dbt/include/global_project/macros/materializations/models/incremental/strategies.sql
jtcohen6: Opened a docs issue: https://github.com/dbt-labs/docs.getdbt.com/issues/1761

jtcohen6 (commented):
Nice work resolving that `dbt-snowflake` failure! And thanks for addressing most of my outstanding comments on the PR.

I see there's one remaining failure in `dbt-bigquery` tests, I think related to the matter of `include_sql_header`. Pending that resolution, this is good to go from a functional perspective / my point of view. Another engineer should still take a look for code review.

gshank:
I didn't find any good way to get include_sql_header without a param, so a put the param back. You can't set config values at runtime...

core/dbt/adapters/base/impl.py
gshank: Actually done now...

core/dbt/include/global_project/macros/materializations/models/incremental/strategies.sql
ChenyuLInx: The changes in `incremental.sql` made me wonder the same thing about default behavior.

ChenyuLInx (approved):
LGTM! There's no impact for python model's incremental change as we just reuse the same logic here

jtcohen6 (approved):
Nice work wrapping this up!

core/dbt/contracts/graph/model_config.py
```diff
@@ -433,6 +433,7 @@ class NodeConfig(NodeAndTestConfig):
     # Note: if any new fields are added with MergeBehavior, also update the
     # 'mergebehavior' dictionary
     materialized: str = "view"
+    incremental_strategy: Optional[str] = None
```
dataders, line 436: @gshank  @nathaniel-may does this now mean that the value of `incremental_strategy` is now in the manifest.json when it wasn't before? or does this add it to the python context such that it is accessible as an attribute of `model` as in `model.incremental_strategy`?

gshank: It means that it should be in the manifest.json. It's accessible like other config keys, but the behavior is a bit different for builtin attributes than for adhoc attributes, in that setting defaults doesn't work the same way.

core/dbt/include/global_project/macros/materializations/models/incremental/strategies.sql
```diff
@@ -0,0 +1,79 @@
...
+
+{% macro default__get_incremental_append_sql(arg_dict) %}
+
+  {% do return(get_insert_into_sql(arg_dict["target_relation"], arg_dict["temp_relation"], arg_dict["dest_columns"])) %}
+
+{% endmacro %}
+
+
+{# snowflake #}
+{% macro get_incremental_delete_insert_sql(arg_dict) %}
```
gvashishtha, line 16: @gshank so if I want to override the "delete+insert" strategy on Postgres, what do I do? Tried creating a macro with name `get_incremental_delete_insert_sql` but that doesn't seem to work

PR #5369, opened 2022-06-13
Add Grant SQL Macros

### LINKED ISSUE #5263: [CT-660] [Feature] Add Grant SQL to Global Project
### Description

This is the second in a two ticket series. #5189 is the first ticket. They will be merged sequentially, but both are required for this feature to be exposed to users.

Today users often configure a `post_hook` to grant permissions on models, seeds, and snapshots:
```
{{ config(post_hook = 'grant select on {{ this }} to role reporter') }}
```

These two tickets aim to make it easier to specify grants allowing it to be configured directly both in dbt_project.yml as well as in source files:
```
# dbt_project.yml
models:
  export:
    +grants:
      select: ['reporter', 'bi']
```
```
-- SQL
{{ config(grants = {'select': ['other_user']}) }}
```

These grant configs will not necessarily look the same for each and every warehouse. The logic to generate the sql from these configs can be overriden by adapters.

Some complexity gets added when you consider that you cannot simply revoke all privileges from a table and apply the configured ones. This is because revoking all may also revoke dbt's ownership of the table, and because some warehouses require explicitly naming the grants to be removed. This implementation reads the grants first, then constructs a minimal revoke and grant statement from the resulting diff.

### Implementation
- [ ] create a new macro in the global project: `get_show_grant_sql(relation: Relation) -> str` that retrieves grant information for that model. The default implementation should return a string in the form `"show grants on table dbt_jcohen.whatever"`
- [ ] create a new macro in the global project: `get_grant_sql(relation: Relation, grant_config: dict) -> str` that creates the warehouse-specific sql from grant portion of the config. Since it will be common, the default implementation should return `grant <privilege> on <relation.type> <relation> to <recipients>`. Warehouses that deviate from this can override. Use dispatch pattern ([dbt docs](https://docs.getdbt.com/reference/dbt-jinja-functions/dispatch)) with a `default__` to enable this. The macro signature here is designed to accept different shapes of grant configs for each adapter to use whatever best fits the warehouse's permissions system.
- [ ] create a new macro called `get_revoke_sql` that takes a relation, and a grant config dict and returns a string in the form `f"revoke {grant_config.privilege} on {relation} from {grant_config.recipients}"`. The grant config will contain multiple priv-recipient mappings.
- [ ] create a new macro in the global project: `apply_grants` which takes 3 parameters: `revoke: Bool`, `relation: Relation`, `grant_config: dict` and returns `None`. It calls `get_show_grant_sql` to determine what grants are currently applied, and uses grant config to determine what grants need to be revoked, and which grants need to be granted. If the `revoke` param is `True`, `get_revoke_sql` is called, then `get_grant_sql` with a new dictionary representing the diff to apply the grants. (see `persist_docs` for an example of similar implementation). This should be overridable by adapters so use the dispatch pattern ([dbt docs](https://docs.getdbt.com/reference/dbt-jinja-functions/dispatch)) with a `default__` to enable this. For a complete example see @jtcohen6's [comment](https://github.com/dbt-labs/dbt-core/issues/5263#issuecomment-1145773353) below.
- [ ] add a call to `apply_grants(..., revoke=True)` in all materializations. This includes incremental models, seeds, and snapshots even though it will usually not be necessary. Users can override if they come across a special case where they need to.
- [ ] #5437
  - [ ] test that projects can define grants in 1) the dbt_project.yml, 2) models, seeds, and snapshots, and 3) both the dbt_project.yml and models, seeds, and snapshots and those permissions can be read back directly from the postgres warehouse after a run.
  - [ ] test two subsequent runs (dbt build && dbt build) with: same grants; additional grants in second run; fewer grants in second run. Confirm that the final set of grants matches the final configured set of grants in all cases.
- [ ] test a model that has preexisting grants, and does not have grants configured. dbt should leave it be, and not run any show/revoke/grant statements.

### Adapters 
Links for your convenience. These do not need to be completed to close out the ticket.
- Snowflake: dbt-labs/dbt-snowflake#168
- BigQuery: dbt-labs/dbt-bigquery#198
- Spark: dbt-labs/dbt-spark#366
- Postgres: #5329
- Redshift: dbt-labs/dbt-redshift#128

McKnight-42:
Resolves #5263 

Description:
Allow users to define grants as a reasonable default in dbt_project.yml and within each model sql or yml file as a config.

### Checklist


core/dbt/include/global_project/macros/adapters/apply_grants.sql
```diff
@@ -0,0 +1,33 @@
+{% macro get_show_grant_sql(relation) %}
+{{ return(adapter.dispatch('get_show_grant_sql', 'dbt')(relation)) }}
+{% endmacro %}
+
+{% macro default__get_show_grant_sql(relation) %}
+{{ return('show grants on table'(relation.schema)) }}
```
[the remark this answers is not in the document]
jtcohen6: Additionally, there's some post-processing logic that needs to happen during/after this query, to accomplish two things:
- Conform to a standardized format
- Ignore grants to the _current user_ (dbt's user), since we don't actually want to compare / revoke these! This this will include `ownership`, etc.

For the purposes of this PR, we need to figure out **what that standardized format needs to look like**. The information we need is: WHO (user/group) has WHICH PRIVILEGE on THIS RELATION, so that we can calculate diffs and template out the appropriate `revoke` / `grant` statements.

---

This is what `show grants` returns on Snowflake:
```
| created_on                     | privilege | granted_on | name                       | granted_to | grantee_name | grant_option | granted_by |
|--------------------------------|-----------|------------|----------------------------|------------|--------------|--------------|------------|
| 2022-06-02 01:52:28.935 -0700  | OWNERSHIP | TABLE      | AD_HOC.DBT_JCOHEN.MY_TABLE | ROLE       | AD_HOC       | true         | AD_HOC     |
```

This is what `show grants` returns on Databricks:
```
| Principal          | ActionType    | ObjectType   | ObjectKey           |
|--------------------|---------------|--------------|---------------------|
| jeremy@dbtlabs.com | OWN           | TABLE        | `airbnb`.`calendar` |
```

This is what the analogous query returns on Postgres:
```
grantor | grantee | table_catalog | table_schema | table_name | privilege_type | is_grantable | with_hierarchy
---------+---------+---------------+--------------+------------+----------------+--------------+----------------
 jerco   | other   | jerco         | dbt_jcohen   | whatever   | INSERT         | NO           | NO
 jerco   | other   | jerco         | dbt_jcohen   | whatever   | SELECT         | NO           | YES
 jerco   | another | jerco         | dbt_jcohen   | whatever   | INSERT         | NO           | NO
 jerco   | another | jerco         | dbt_jcohen   | whatever   | SELECT         | NO           | YES
```

And on BigQuery:
```
| object_catalog | object_schema | object_name | object_type | privilege_type            | grantee                  |
|----------------|---------------|-------------|-------------|---------------------------|--------------------------|
| dbt-test-env   | dbt_jcohen    | my_table    | TABLE       | roles/bigquery.dataViewer | user:jeremy@example.com  |
| dbt-test-env   | dbt_jcohen    | my_table    | TABLE       | roles/custom.whatEver     | user:another@example.com |
```

([The Redshift version of this query will be much messier](https://stackoverflow.com/questions/18741334/how-do-i-view-grants-on-redshift), unfortunately)

---

The nice thing on Postgres/Redshift/BigQuery is that, because it's a real `select` query, we can do all that processing right in the SQL query: just the columns we need, renamed to conformed standard, with current user's grants filtered out. For databases that support `show grants`, we need to do that post-processing in Jinja/Python instead. If that processing proves very difficult, Snowflake does give us the option of querying an `information_schema` view instead (https://docs.snowflake.com/en/sql-reference/info-schema/object_privileges.html). There are some performance costs to using `information_schema` instead of `show` on Snowflake, which we can discuss if/when we get there.

jtcohen6 (commented):
Great start on this @McKnight-42! Thanks for opening the draft PR so early, so that we can keep pushing forward on this.

In my comments, I tried to provide relevant information that will help us decide if we're implementing the right standard, by initializing this default behavior in `dbt-core`. Much of the database-specific nuance will come out in the per-adapter implementations, but we always need to first make sure that we're putting the right core constructs in place.

core/dbt/include/global_project/macros/adapters/apply_grants.sql
```diff
@@ -0,0 +1,33 @@
...
+
+
+{% macro apply_grants(revoke, relation, grant_config) %}
+{{ return(adapter.dispatch('apply_grant', 'dbt')(revoke, relation, grant_config)) }}
+{% endmacro %}
+
+{% macro default__apply_grants(revoke, relation, grant_config) %}
+{{ get_show_grant_sql() }}
+{% return %}
+{% endmacro %}
```
jtcohen6, line 33: If `grant_config` is not specified by the user, `apply_grants` should be a no-op. We should take that to mean that the user does not want dbt running `grant` or `revoke` on their behalf.

core/dbt/include/global_project/macros/adapters/apply_grants.sql
```diff
@@ -0,0 +1,33 @@
...
+{{ return(relation) }}
+{% endmacro %}
+
+{% macro get_revoke_sql(relation, grant_config) %}
+{{ return(adapter.dispatch('get_revoke_sql', 'dbt')(relation, grant_config)) }}
+{% endmacro %}
+
+{% macro default__get_revoke_sql(relation, grant_config) %}
+return
+{% endmacro %}
```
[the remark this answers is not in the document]
McKnight-42: Very much a work in progress, updated some, love all feedback thank you so much.

core/dbt/include/global_project/macros/adapters/apply_grants.sql
```diff
@@ -0,0 +1,48 @@
+{% macro get_show_grant_sql(relation) %}
+{{ return(adapter.dispatch("get_show_grant_sql", "dbt")(relation)) }}
+{% endmacro %}
+
+{% macro default__get_show_grant_sql(relation) %}
+show grants on {{ relation.type }} {{ relation }}
+{% endmacro %}
```
jtcohen6, line 7: Nice! What information do we need this macro to return?  I think it's just:
- privileges being granted (`select`, `usage`, etc)
- recipients (also called "principals" or "grantees")

That is, for _this_ relation, excluding privileges held by the current user/role/principal. I'm also realizing, we should probably limit to just the grants given _by_ this user/role/principal, to guarantee that any subsequent `revoke` statements will succeed.

As previously mentioned, `show grants` feels like a good standard/default. Assuming we're starting by testing with Postgres, we'll need to implement a PostgreSQL version, which looks a little different:
```sql
{% macro postgres__get_show_grant_sql(relation) %}

    select
    
        -- we'll need to standardize these names across databases
        -- which ones do we want to pick?
        
        grantee,          -- or 'recipient'? or 'principal'?
        privilege_type    -- or 'privilege'?
    
    from information_schema.role_table_grants
    where grantor = current_role
      and grantee != current_role
      and table_schema = '{{ relation.schema }}'
      and table_name = '{{ relation.identifier }}'
        
{% endmacro %}
```

jtcohen6 (commented):
nice progress!

core/dbt/include/global_project/macros/adapters/apply_grants.sql
```diff
@@ -0,0 +1,50 @@
...
+{{ return(adapter.dispatch("apply_grants", "dbt")(relation, grant_config, should_revoke)) }}
+{% endmacro %}
+
+{% macro default__apply_grants(relation, grant_config, should_revoke=True) %}
+{% if grant_config %}
+    {% call statement('grants') %}
+        {% if should_revoke %}
+            {% set current_grants =  run_query(get_show_grant_sql(relation)) %}
+             {{ log(current_grants, info = true) }}
+            {% set diff_grants = { k: current_grants[k] for k in set(current_grants) - set(grant_config) } %}
```
McKnight-42, line 42: So I know inline python isn't really allowed in jinja without doing some extending or building out a new block of some form to allow it. so beyond building out a new single use function in impl or something like that is there a better way to try and grab the diffs?

jtcohen6: Yup, our options here are:
- writing this in Jinja, using `set` methods as you're trying here — I bet it's possible!
- adding a new Python context method to handle this logic for us (which could be defined as a [context member](https://github.com/dbt-labs/dbt-core/tree/main/core/dbt/context), or as a class method on the adapter)

McKnight-42: oh great, will continue playing around with trying to jinjafy the base python i'm trying to run was really digging into the postgres version of persist_docs as you mentioned

McKnight-42:
NOTE TO SELF: create postgres version of get_show_grant sql https://linuxhint.com/check-postgres-user-privileges/

McKnight-42:
NOTE: Working on `diff_grants` variable not be passed all the way through, fixed issue where we were turning the object into a table which made it incompatible for `diff_of_two_dicts` context.

plugins/postgres/dbt/include/postgres/macros/adapters.sql
```diff
@@ -202,3 +202,15 @@
...
   {% endfor %}
 {% endmacro %}
+
+{% macro postgres__get_show_grant_sql(relation) %}
+ {% call statement('get_show_grant_sql', fetch_result=True) -%}
+  select grantee
+  from information_schema.role_table_grants
+      {{ log(grantee, info=True) }}
+      where grantor = current_role
+        and grantee != current_role
```
dataders, line 212: @jtcohen6 where does `current_role` come from here?

jtcohen6: @dataders This is a value that PostgreSQL makes available to us, in all queries. In other databases, we may need to use dbt's own knowledge of the user/role (`{{ target.user }}` or `{{ target.role }}`) here instead.

core/dbt/include/global_project/macros/adapters/apply_grants.sql
```diff
@@ -0,0 +1,52 @@
...
+{% macro get_grant_sql(relation, grant_config) %}
+{{ return(adapter.dispatch('get_grant_sql', 'dbt')(relation, grant_config)) }}
+{% endmacro %}
+
+{% macro default__get_grant_sql(relation, grant_config) %}
+    {% for privilege in grant_config.keys() %}
+     {{ log('privilege: ' ~ privilege) }}
+        {% set grantees = grant_config[privilege] %}
+        {% for grantee in grantees %}
+            grant {{ privilege }} on {{ relation.type }} {{ relation }} to {{ grantee }};
```
jtcohen6, line 18: This works totally fine, just an ergonomic piece: It's possible to `grant <privilege> on <object> to first_grantee, second_grantee, third_grantee, ...`. It's slightly neater and avoids an extra looping step for us:
```sql
        {% set grantees = grant_config[privilege] %}
        grant {{ privilege }} on {{ relation.type }} {{ relation }} to {{ grantees | join(', ') }};
```

core/dbt/include/global_project/macros/adapters/apply_grants.sql
```diff
@@ -0,0 +1,52 @@
...
+{% endmacro %}
+
+{% macro get_revoke_sql(relation, grant_config) %}
+{{ return(adapter.dispatch("get_revoke_sql", "dbt")(relation, grant_config)) }}
+{% endmacro %}
+
+{% macro default__get_revoke_sql(relation, grant_config) %}
+    {% for privilege in grant_config.keys() %}
+        {% set grantees = grant_config[privilege] %}
+        {% for grantee in grantees if grantee !=  target.user %}
```
jtcohen6, line 30: Ah! I don't know if this is the spot where we want to filter out `grantee != target.user`. I would rather have that check during the `show grants` / post-processing, so that we don't even consider the current user's grants to begin with.

As in `get_grant_sql` above, we can also avoid the extra loop step and just join `grantees` together as a comma-separated list.

core/dbt/include/global_project/macros/materializations/models/incremental/incremental.sql
```diff
@@ -59,6 +61,8 @@
       {% do to_drop.append(backup_relation) %}
   {% endif %}
 
+  {% do apply_grants(target_relation, grant_config, should_revoke=True) %}
```
jtcohen6, line 64: As we discussed yesterday, `should_revoke=True` will always work. For maximal efficiency, it should be conditional based on:
- Are we running this incremental model from scratch? (`existing_relation is none or full_refresh_mode`)
- Is this database one that carries over grants when fully replacing an object? (adapter-specific method to tell us `True` or `False`)

core/dbt/include/global_project/macros/adapters/apply_grants.sql
```diff
@@ -0,0 +1,52 @@
...
+            revoke {{ privilege }} on {{ relation.type }} {{ relation }} from {{ grantee }}
+        {% endfor %}
+    {% endfor %}
+{% endmacro %}
+
+{% macro apply_grants(relation, grant_config, should_revoke) %}
+{{ return(adapter.dispatch("apply_grants", "dbt")(relation, grant_config, should_revoke=True)) }}
```
jtcohen6, line 37: We need to pass the actual argument value through, rather than always setting this to `True`:
```suggestion
{{ return(adapter.dispatch("apply_grants", "dbt")(relation, grant_config, should_revoke)) }}
```
(This doesn't set the default, it overrides the value being passed in!)

core/dbt/context/base.py
```diff
@@ -657,6 +657,16 @@ def print(msg: str) -> str:
...
         return ""
 
+    @contextmember
+    @staticmethod
+    def diff_of_two_dicts(grant_config, current_grants):
+        """compares list of current grants and any changes made to grants on a model in a
+        current run and returns the diff as a new object
+        """
+
+        diff_dict = {k: grant_config[k] for k in set(grant_config) - set(current_grants)}
```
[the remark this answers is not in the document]
jtcohen6: Here's the super janky pseudo-code way in which I first tried to express that logic:
```python
    def diff_of_two_dicts(grant_config, current_grants):
        """compares list of current grants and any changes made to grants on a model in a
        current run and returns the diff as a new object
        """
        needs_revoking = {}
        needs_granting = {}
        
        # first: what needs revoking?
        # think about this as "current_grants MINUS grant_config"
        # for each privilege type currently granted on the current object
        for k in current_grants:
            # do we want to be granting that privilege type *at all*?
            if k in grant_config:
                # if so, is it granted to more people than it should be?
                diff = list(set(current_grants[k]) - set(grant_config[k]))
                if diff:
                    needs_revoking.update({k: diff})
            else:
                # if not, let's revoke everyone who has that privilege type
                needs_revoking.update({k: current_grants[k]})
        
        # the same exact thing, now in reverse
        # what needs granting? "grant_config MINUS current_grants"
        for k in grant_config:
            if k in current_grants:
                diff = list(set(grant_config[k]) - set(current_grants[k]))
                if diff:
                    needs_granting.update({k: diff})
            else:
                needs_granting.update({k: grant_config[k]})
        
        return needs_revoking, needs_granting
```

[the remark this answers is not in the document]
jtcohen6: Now, after writing all that, and realizing that the name of the method is `diff_of_two_dicts` (much more generic), I then realized that we could generalize this to the actual diff of two dicts!
```python
    def diff_of_two_dicts(dict_a, dict_b):
        """compares list of current grants and any changes made to grants on a model in a
        current run and returns the diff as a new object
        """
        dict_diff = {}
        for k in dict_a:
            if k in dict_b:
                diff = list(set(dict_a[k]) - set(dict_b[k]))
                if diff:
                    dict_diff.update({k: diff})
            else:
                dict_diff.update({k: dict_a[k]})
        return dict_diff
```
Then, we'd call it twice down in the actual `apply_grants` macro: one to get **current minus configured** (= needs revoking), and again to get **configured minus current** (= needs granting).

jtcohen6 (commented):
@McKnight-42 Amazing progress!! I feel like we're getting really really close. I think it's just about time to start writing some functional tests, so we can automate the pieces that  :)

I was poking around some of the code in order to really understand how the pieces fit together. (For better or worse, that's the only real way I can do that.) I'm linking here ONLY if this is helpful, or illustrative of some of the comments I've left below. Please do not feel in any way constrained by any of this code: https://github.com/dbt-labs/dbt-core/compare/ct-660-grant-sql...jerco/grant-experimentation

plugins/postgres/dbt/include/postgres/macros/adapters.sql
```diff
@@ -202,3 +202,15 @@
     comment on column {{ relation }}.{{ adapter.quote(column_name) if column_dict[column_name]['quote'] else column_name }} is {{ escaped_comment }};
   {% endfor %}
 {% endmacro %}
+
+{% macro postgres__get_show_grant_sql(relation) %}
+ {% call statement('get_show_grant_sql', fetch_result=True) -%}
+  select grantee
```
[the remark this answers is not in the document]
McKnight-42: So correct adding in. btw going over your experimental code and really fantastic work how quickly you were able to figure out the agate.Table conversion and things like that!

core/dbt/include/global_project/macros/adapters/apply_grants.sql
McKnight-42: oh okay thats interesting when we were first testing out it seemed like it was replacing the True we set via default macro call to a None value so good to know.

gshank:
Right now we're retrieving the 'grants_config' dictionary in the materialization macros and passing it in to 'apply_grants', but we never do anything else with it prior to the call. So I'm wondering if it wouldn't be better to move retrieving the grants_config dictionary into the 'apply_grants' call, to simplify the materialization setup?

jtcohen6:
@gshank Super fair point. As a question of coding style, we've discussed (though haven't written down clearly) a preference for explicitly passing in arguments from the materialization into macros called by the materialization, rather than repeatedly accessing `config` as "global state" within nested macros.

At the same time, I'm aware that we do that _all over the place_ today, and we're not consistent about it one way or the other. For example, `persist_docs` is a relatively more recent feature, and we access `config` directly within the `persist_docs` macro (rather than within the materialization):

https://github.com/dbt-labs/dbt-core/blob/60e491b3c178ee8112dbe1311bcc69e5c778ccfa/core/dbt/include/global_project/macros/adapters/persist_docs.sql#L25-L33

core/dbt/include/global_project/macros/adapters/apply_grants.sql
```diff
@@ -0,0 +1,52 @@
...
+{% endmacro %}
+
+{% macro default__apply_grants(relation, grant_config, should_revoke=True) %}
+    {% if grant_config %}
+        {% call statement('grants') %}
+            {% if should_revoke %}
+                {% set current_grants =  run_query(get_show_grant_sql(relation)) %}
+                {{ log(current_grants.print_table()) }}
+                {% set diff_grants = diff_of_two_dicts(grant_config, current_grants) %}
```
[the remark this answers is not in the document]
emmyoop: @McKnight-42 here's the reason for using `.lower()` in `standardize_grants_dict`.  It's worth a comment in the code for _why_ we're adding the `.lower()`.

McKnight-42:
7/5/22 EOD report: models building on **snowflake**, and **postgres**; possible issue with how `grants_table` is getting configured leading to needs_revoking` and `needs_granting` to possibly be wrong? `needs_granting` remains roughly same as `grant_config` while `needs_revoking` stays an empty `{}`. double checked apply_grants that it's not reseting by hitting the else statement in either.

plugins/postgres/dbt/adapters/postgres/impl.py
```diff
@@ -85,6 +86,18 @@ def verify_database(self, database):
     def parse_index(self, raw_index: Any) -> Optional[PostgresIndexConfig]:
         return PostgresIndexConfig.parse(raw_index)
 
+    @available
+    def standardize_grants_dict(self, grants_table: agate.Table) -> dict:
+        grants_dict = {}
+        for row in grants_table:
+            grantee = row["grantee"]
+            privilege = row["privilege"]
```
jtcohen6, line 94: This is currently raising a `KeyError`:
```
11:53:03  Unhandled error while executing model.testy.incremental
'privilege'
```
The reason? This column is called `privilege_type` in your `postgres__get_show_grant_sql` query below. You can either rename it to use `privilege_type` here, or add a column alias in the query below (`privilege_type as privilege`).

core/dbt/include/global_project/macros/adapters/apply_grants.sql
```diff
@@ -0,0 +1,70 @@
...
+{% macro get_revoke_sql(relation, grant_config) %}
+{{ return(adapter.dispatch("get_revoke_sql", "dbt")(relation, grant_config)) }}
+{% endmacro %}
+
+{% macro default__get_revoke_sql(relation, grant_config) %}
+    {% for privilege in grant_config.keys() %}
+        {% set grantees = [] %}
+        {% set all_grantees = grant_config[privilege] %}
+        {% for grantee in all_grantees %}
+            {% if grantee != target.user %}
```
jtcohen6, line 33: IMO the `target.user` filter would be better in post-processing / standardization, rather than here, since this behavior will differ by adapter (= whether the filter can happen in SQL or not).

To make that possible, I think we could either:
1. Reach deep into the `adapter` object to find `credentials.user` or `credentials.role`
2. Add an argument to `standardize_grants_dict` which accepts `target.user` (default) / `target.role` (Snowflake)
3. Run a separate query that returns `current_user` / `current_role`

1 is gross, but it does nicely account for this being a potentially different property on different databases

core/dbt/include/global_project/macros/adapters/apply_grants.sql
```diff
@@ -0,0 +1,70 @@
...
+
+{% macro default__get_show_grant_sql(relation) %}
+show grants on {{ relation.type }} {{ relation }}
+{% endmacro %}
+
+{% macro get_grant_sql(relation, grant_config) %}
+{{ return(adapter.dispatch('get_grant_sql', 'dbt')(relation, grant_config)) }}
+{% endmacro %}
+
+{% macro default__get_grant_sql(relation, grant_config) %}
```
jtcohen6, line 13: **Nit**

For `default__get_grant_sql` + `default__get_revoke_sql`: Let's switch out most of the `{% %}` for `{%- -%}`, to avoid tons of whitespace in the logs. This:
```
On model.testy.view: /* {"app": "dbt", "dbt_version": "1.2.0b1", "profile_name": "garage-postgres", "target_name": "dev", "node_id": "model.testy.view"} */

                    
    

                    
    
        
        
            
                grant select on table "jerco"."dbt_jcohen"."view" to other;
            
                grant select on view "jerco"."dbt_jcohen"."view" to another;
            
        
    

                
```
Should become (at least a little better):
```
/* {"app": "dbt", "dbt_version": "1.2.0b1", "profile_name": "garage-postgres", "target_name": "dev", "node_id": "model.testy.table"} */

                    
                    
                grant select on table "jerco"."dbt_jcohen"."table" to other;
            
                grant select on table "jerco"."dbt_jcohen"."table" to another;
            
                
```

plugins/postgres/dbt/include/postgres/macros/adapters.sql
```diff
@@ -202,3 +202,13 @@
     comment on column {{ relation }}.{{ adapter.quote(column_name) if column_dict[column_name]['quote'] else column_name }} is {{ escaped_comment }};
   {% endfor %}
 {% endmacro %}
+
```
jtcohen6, line 205: Funny story: Postgres/Redshift use `grant {privilege} on table {relation}`, _regardless of whether `{relation}` is a table or a view_.

So Postgres will raise an error if we try to do `grant {privilege} on view {view_name}`. It must be `grant {privilege} on table {view_name}`.

Given that this is a wonkiness of PostgreSQL, I think the right move is to reimplement `postgres__get_grant_sql`. Notice that this is always `table`, instead of `{{ relation.type }}`:
```sql
{%- macro postgres__get_grant_sql(relation, grant_config) -%}
    {%- for privilege in grant_config.keys() -%}
        {%- set grantees = grant_config[privilege] -%}
        {%- if grantees -%}
            {%- for grantee in grantees %}
                grant {{ privilege }} on table {{ relation }} to {{ grantee}};
            {% endfor -%}
        {%- endif -%}
    {%- endfor -%}
{%- endmacro -%}
```

jtcohen6 (changes requested):
Feels like we're very very close!! Let me know if it would be helpful to talk through this live.

MUST fix (currently raising errors):
- [x] Postgres views: `grant on table` always
- [x] Postgres: standardize name for `privilege`/`privilege_type`

SHOULD fix (inefficient but correct):
- [x] Case-insensitive comparison within `diff_of_two_dicts`
- [x] `should_revoke` + `do_i_carry_over_grants_when_an_object_is_replaced` - Defined per adapter or per materialization? No need to revoke when replacing on Postgres + Redshift (always), Snowflake IFF `copy_grants` is enabled
- [x] Move current user/role filter to post-processing / standardization, rather than within `get_revoke_sql`, since it will likely differ across adapters

COULD fix (aesthetic):
- [x] Whitespace in templated SQL

plugins/postgres/dbt/include/postgres/macros/adapters.sql
```diff
@@ -202,3 +202,13 @@
...
+{% macro postgres__get_show_grant_sql(relation) %}
+  select grantee, privilege_type
+  from information_schema.role_table_grants
+      {{ log(grantee, info=True) }}
```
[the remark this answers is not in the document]
McKnight-42: done

plugins/postgres/dbt/adapters/postgres/impl.py
```diff
@@ -85,6 +86,18 @@ def verify_database(self, database):
     def parse_index(self, raw_index: Any) -> Optional[PostgresIndexConfig]:
         return PostgresIndexConfig.parse(raw_index)
 
+    @available
+    def standardize_grants_dict(self, grants_table: agate.Table) -> dict:
```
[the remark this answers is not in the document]
McKnight-42: done

plugins/postgres/dbt/adapters/postgres/impl.py
McKnight-42: [standardize_grants_dict](https://github.com/dbt-labs/dbt-core/pull/5369/files#diff-4c547e7dad44fb23a74855d32871c8d60d47382ceba0a5bf64273967256bdd08R91-R100) changes made for **postgres**

jtcohen6:
Follow-ups from pairing today: https://github.com/dbt-labs/dbt-core/compare/ct-660-grant-sql...jerco/grants-over-finish-line

That branch includes some of the SHOULDs + COULDs noted above:
- case-insensitive comparisons
- whitespace cleanup

The trickiest bit here is definitely the `should_revoke` conversation, and that's what we tackled head-on during our pairing session.

We ended up developing an approach with two macros. For now, they have silly long names, but it's important to get the concept right first. We can come up with more concise names later. They are:
1. `do_we_need_to_show_and_revoke_grants()` <-- called within the **materialization**, takes two inputs: Does the object already exist in the database? Are we fully replacing that object (a.k.a. "full refresh mode")?
2. `are_grants_copied_over_when_replaced()` <-- defined **per adapter**. By default, `true` (play it safe!); on Redshift/Postgres, always `false`; on Snowflake, it depends on the value of the `copy_grants` config.

The appeal is that the `apply_grants` call site within most materializations can look (almost) exactly the same everywhere:
```sql
  {% set should_revoke = do_we_need_to_show_and_revoke_grants(existing_relation, full_refresh_mode=...) %}
  {% do apply_grants(target_relation, grant_config, should_revoke=should_revoke) %}
```
Where:
- `existing_relation` (sometimes called `old_relation`) is the cache lookup that answers the question, "Does the object already exist in the database?"
- `full_refresh_mode` answers the question, "Are we fully replacing that preexisting object?" This is always `True` for the table + view materializations, always `False` for snapshots, and conditionally set for incrementals + seeds.

That logic does _not_ need to differ across adapters, because that piece is handled within the `are_grants_copied_over_when_replaced()` macro. So, once finalized here, it will just need updating in our adapter PRs as well.

core/dbt/include/global_project/macros/materializations/models/view/create_or_replace_view.sql
```diff
@@ -34,6 +34,7 @@
     {{ get_create_view_as_sql(target_relation, sql) }}
   {%- endcall %}
 
+  {% do apply_grants(target_relation, grant_config, should_revoke=True) %}
```
jtcohen6, line 37: Similar logic to the default `view` materialization for this:
```suggestion
  {% set should_revoke = do_we_need_to_show_and_revoke_grants(exists_as_view, full_refresh_mode=True) %}
  {% do apply_grants(target_relation, grant_config, should_revoke=should_revoke) %}
```

core/dbt/adapters/base/impl.py
```diff
@@ -1072,6 +1093,11 @@ def get_compiler(self):
 
         return Compiler(self.config)
 
+    # used in apply_grants -- True is a safe default
+    @available
+    def do_i_carry_over_grants_when_an_object_is_replaced(self) -> bool:
+        return True
+
```
jtcohen6, line 1100: Remove since now unused. We've opted for macros instead.
```suggestion
```

core/dbt/include/global_project/macros/adapters/apply_grants.sql
```diff
@@ -0,0 +1,93 @@
+{% macro are_grants_copied_over_when_replaced() %}
```
jtcohen6, line 1: TODO before merge: pick shorter/clearer names for
- `are_grants_copied_over_when_replaced`
- `do_we_need_to_show_and_revoke_grants`

jtcohen6 (commented):
The functional test added in #5230, [`test_model_grant_config`](https://github.com/dbt-labs/dbt-core/blob/f304b4b2dae343ed5359a0169d8e5ea32c2a1c62/tests/functional/configs/test_grant_configs.py#L59), is now failing on Postgres. Why? Because we're actually applying the grant during `dbt run` — **a very good thing!!**

```
09:05:50.973415 [debug] [Thread-1  ]: Using postgres connection "model.test.my_model"
09:05:50.973778 [debug] [Thread-1  ]: On model.test.my_model: /* {"app": "dbt", "dbt_version": "1.2.0b1", "profile_name": "test", "target_name": "default", "node_id": "model.test.my_model"} */



                grant my_select on table "dbt"."test16571847508452268512_test_grant_configs"."my_model" to reporter;

                grant my_select on table "dbt"."test16571847508452268512_test_grant_configs"."my_model" to bi;


09:05:50.975758 [debug] [Thread-1  ]: Postgres adapter: Postgres error: role "reporter" does not exist

09:05:50.976081 [debug] [Thread-1  ]: On model.test.my_model: ROLLBACK
09:05:50.976613 [debug] [Thread-1  ]: finished collecting timing info
09:05:50.976939 [debug] [Thread-1  ]: On model.test.my_model: Close
09:05:50.977378 [debug] [Thread-1  ]: Database Error in model my_model (models/my_model.sql)
  role "reporter" does not exist
  compiled SQL at target/run/test/models/my_model.sql
```
We have two options here:
1. Keep the existing functional test, `test_model_grant_config`, _just_ for checking configs, and not for validating actual runtime behavior. We could do that very simply by swapping out all instances of `run_dbt(["run"])` for `run_dbt(["list"])`. I tried that locally, and the test passes again.
2. Create `reporter`, `bi`, `someone`, `other_user` as users while setting up our Postgres testing instance.

My instinct is to do (1). Open to your + @gshank's thoughts here. I know Gerda is working on separate functional tests to validate actual runtime behavior, for which we will need to add new [Postgres setup steps](https://github.com/dbt-labs/dbt-core/blob/main/test/setup_db.sh#L41-L51) to include users/groups who can actually receive our grants in tests.

core/dbt/adapters/base/impl.py
```diff
@@ -538,6 +539,26 @@ def list_relations_without_caching(self, schema_relation: BaseRelation) -> List[
             "`list_relations_without_caching` is not implemented for this " "adapter!"
         )
 
+    ###
+    # Abstract methods about grants
+    ###
+    @abc.abstractmethod
```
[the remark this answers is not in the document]
jtcohen6: Update: looks like @gshank went for option (1) in https://github.com/dbt-labs/dbt-core/pull/5447, and I think that's the right move!

core/dbt/include/global_project/macros/materializations/models/view/create_or_replace_view.sql
McKnight-42: done

core/dbt/adapters/base/impl.py
McKnight-42: done

core/dbt/include/global_project/macros/adapters/apply_grants.sql
McKnight-42: @jtcohen6 `copy_grants` seems very reasonable.

if keeping the name same across the board seems like the best plan, i'd personally err towards what the goal of that input will be which would be should we revoke this thing, if we wanted to keep names as similar as possible but possibly make it easier to point out logs wise where we are in future debugging, maybe something like `should_revoke_bool` which is what the output thats assigned to the `should_revoke` variable that becomes our argument for apply_grants.

jtcohen6: > goal of that input will be which would be should we revoke this thing

Agreed — I'm thinking that the goal here is less _should we revoke_, and actually _should we calculate diffs between current + desired, then revoke &/or grant as need be_.

Whatever we pick, I think we should use the same name for this macro, and for the argument to `apply_grants` (currently named `should_revoke`).

plugins/postgres/dbt/include/postgres/macros/adapters.sql
```diff
@@ -202,3 +202,16 @@
     comment on column {{ relation }}.{{ adapter.quote(column_name) if column_dict[column_name]['quote'] else column_name }} is {{ escaped_comment }};
   {% endfor %}
 {% endmacro %}
+
+{%- macro postgres__get_show_grant_sql(relation) -%}
+  select grantee, privilege_type
+  from information_schema.role_table_grants
+      where grantor = current_role
```
jtcohen6, line 209: Not a TODO, just an open question: Should we / do we need to limit ourselves to just the grants provided by the `current_role`? 

To copy-paste what I just said in `dbt-snowflake`:

> The original motivation for this was: If someone else has granted a privilege on this model, and then dbt tries to revoke it, the revocation may fail. But if dbt's user/role owns the table, it should be able to revoke any access on it... right?

By removing this logic, we'd be making a strong claim, that ALL the grants applied on this object, MUST be applied by dbt, in the `grants` config.

No action needed right now. Just wanted to leave the comment for future reference.

core/dbt/include/global_project/macros/adapters/apply_grants.sql
```diff
@@ -0,0 +1,87 @@
...
+
+{% macro default__apply_grants(relation, grant_config, should_revoke=True) %}
+    {% if grant_config %}
+            {% if should_revoke %}
+                {% set current_grants_table = run_query(get_show_grant_sql(relation)) %}
+                {% set current_grants_dict = adapter.standardize_grants_dict(current_grants_table) %}
+                {% set needs_granting = diff_of_two_dicts(grant_config, current_grants_dict) %}
+                {% set needs_revoking = diff_of_two_dicts(current_grants_dict, grant_config) %}
+                {% if not (needs_granting or needs_revoking) %}
+                    {{ log('All grants are in place, no revocation or granting needed.')}}
```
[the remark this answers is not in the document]
McKnight-42: Have swapped  will push up,over to having the log for now, is there something that would be more preferable? just not acknowledging this use case or should we always let the user know or have ability to check against in logs?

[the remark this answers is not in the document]
jtcohen6: Yeah, I actually like leaving it in! Shows that dbt has done some good work, calculated diffs and found none ;)

Adding the `On {relation}` bit ensures that they know which model this statement is relevant to

jtcohen6:
@McKnight-42 I've gone back through the threads above, and I think all the blocking TODOs have been resolved! Nice work!

The tests are still failing, and the fixes for those tests are in #5447. Let's either:
- pull the fixes into this branch, so we can merge this PR into `main`
- wait to merge #5447 (into this branch), before merging anything into `main`

core/dbt/include/global_project/macros/adapters/apply_grants.sql
McKnight-42: currently using the original `should_revoke` and `copy_grants` names till we get some other ideas, `calculate_diffs` is a very good option as well more than willing to change it later throughout as needed

jtcohen6: Happy with the consistency for now!

core/dbt/include/global_project/macros/adapters/apply_grants.sql
```diff
@@ -0,0 +1,87 @@
...
+    {{ return(adapter.dispatch('get_grant_sql', 'dbt')(relation, grant_config)) }}
+{% endmacro %}
+
+{%- macro default__get_grant_sql(relation, grant_config) -%}
+    {%- for privilege in grant_config.keys() -%}
+        {%- set grantees = grant_config[privilege] -%}
+        {%- if grantees -%}
+            {%- for grantee in grantees -%}
+                grant {{ privilege }} on {{ relation }} to {{ grantee }};
+            {%- endfor -%}
```
jtcohen6, line 42: Caught a bug. This doesn't properly handle multiple `grantees` (whitespace). We can more efficiently provide a list of multiple grantees by joining them together as a comma-separated list: `grant ... to grantee_a, grantee_b, ...`

```suggestion
            grant {{ privilege }} on {{ relation }} to {{ grantees | join(', ' }};
```

We need testing for multiple grantees!

McKnight-42: made changes for both revoke and grant statements trying to test it out now! locally with what we currently have

McKnight-42: change made in more grant adapter tests pr

core/dbt/context/base.py
```diff
@@ -657,6 +657,30 @@ def print(msg: str) -> str:
...
+        dict_b = {k.lower(): v for k, v in dict_b.items()}
+        for k in dict_a:
+            if k in dict_b:
+                a_lowered = map(lambda x: x.lower(), dict_a[k])
+                b_lowered = map(lambda x: x.lower(), dict_b[k])
+                diff = list(set(a_lowered) - set(b_lowered))
+                if diff:
+                    dict_diff.update({k: diff})
+            else:
+                dict_diff.update({k: dict_a[k]})
```
dataders, line 681: have we tested the performance on this when the grants dict config is huge (and very different from current grants)? might not suck once, but might if they have this same complicated diff on hundreds of models?

plugins/postgres/dbt/include/postgres/macros/adapters.sql
```diff
@@ -202,3 +202,16 @@
...
+        and grantee != current_role
+        and table_schema = '{{ relation.schema }}'
+        and table_name = '{{ relation.identifier }}'
+{%- endmacro -%}
+
+{% macro postgres__are_grants_copied_over_when_replaced() %}
+    {{ return(False) }}
+{% endmacro %}
```
dataders, line 217: is this used anywhere?

core/dbt/include/global_project/macros/materializations/models/incremental/incremental.sql
```diff
@@ -20,6 +20,8 @@
   -- BEGIN, in a separate transaction
   {%- set preexisting_intermediate_relation = load_cached_relation(intermediate_relation)-%}
   {%- set preexisting_backup_relation = load_cached_relation(backup_relation) -%}
+   -- grab current tables grants config for comparision later on
```
ChenyuLInx, line 23: Not something needed to do before merging, but we added the same change pretty much for all materialization methods. Should we consider refactoring the code so we can only do it in one place?

plugins/postgres/dbt/include/postgres/macros/adapters.sql
jtcohen6: @McKnight-42 I think this needs to be renamed to `postgres__copy_grants`! Since we've now renamed `are_grants_copied_over_when_replaced` to `copy_grants`

McKnight-42: ranemed in this pr and pushed up --Done

core/dbt/context/base.py
jtcohen6: This code can + should absolutely be refactored for more Pythonic comprehension + better performance / lower memory footprint. I'm not the right person to do that :)

> hundreds of models?

FWIW this code will be executed in parallel `--threads`, assuming the user is making use of them

dbeatty10 (approved):
👍 
Looks good to me. Only thing we should do first is apply the white space fixes.

My opinion is to ship this now so that we can unblock merging for everything else downstream.

Separately, we can break out any optimizations or refactoring separately to incorporate @nathaniel-may and @ChenyuLInx's ideas.

jtcohen6 (approved):

jtcohen6:
What Doug said! Those are good changes, and we can pick them up in subsequent refactors. Let's unblock the adapter PRs in the meantime.

core/dbt/include/global_project/macros/materializations/snapshots/snapshot.sql
```diff
@@ -5,6 +5,8 @@
 
   {%- set strategy_name = config.get('strategy') -%}
   {%- set unique_key = config.get('unique_key') %}
+  -- grab current tables grants config for comparision later on
+  {%- set  grant_config = config.get('grants') -%}
```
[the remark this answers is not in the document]
McKnight-42: spacing fixed pushing up

PR #5370, opened 2022-06-13
feature/deps-regex-semver

### LINKED ISSUE #5201: [CT-589] [Bug] `dbt deps` does not precisely match the regex for SemVer

### TL;DR
There's probably not a very impactful issue here -- opening up this issue at risk of being very nit picky.

### Current Behavior

[This regular expression](https://github.com/dbt-labs/dbt-core/blob/28c44a9be733491314be692e93f6f15b11a91be9/core/dbt/semver.py#L70) is used within `dbt deps` to determine valid installable versions. It might also be utilized during `dbt --version` to [determine if the installed adapters are up-to-date](https://docs.getdbt.com/docs/core-versions#how-we-version-adapter-plugins) with the latest versions available.

If I did the string substitution correctly, it produces these results for a range of input strings:
https://regex101.com/r/YyaxyD/1

### Expected Behavior

There is a [suggested regular expression](https://semver.org/#is-there-a-suggested-regular-expression-regex-to-check-a-semver-string)  (RegEx) to check a SemVer string.

It produces these results:
https://regex101.com/r/Ly7O1x/3/

### Steps To Reproduce

1. Compile the regex [here](https://github.com/dbt-labs/dbt-core/blob/28c44a9be733491314be692e93f6f15b11a91be9/core/dbt/semver.py#L70) (manually or via Python)
    - `^(?P<major>0|[1-9]\d*)\.(?P<minor>0|[1-9]\d*)\.(?P<patch>0|[1-9]\d*)(?:-(?P<prerelease>(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\+(?P<buildmetadata>[0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?$`
2. Drop it into [regex101.com](https://regex101.com/)
3. Re-use the test cases from [here](https://regex101.com/r/Ly7O1x/3/)
4. Browse through the output of valid semantic versions (for false negatives) and invalid semantic versions (for false positives)

### Relevant log output

image

image

### Why it's not that impactful

There's two cases of errors ([false negatives and false positives](https://en.wikipedia.org/wiki/Type_I_and_type_II_errors)). The former is more impactful to users; there are very few known _theoretical_ cases (and zero known _actual_ cases in the wild).

#### False negative - a valid semantic version doesn't match the regex
This is the case that would impact users the most.
There are only a couple of these, and it's easy to change to one that is recognized:
- 1.0.0-alpha.0valid
- 1.0.0-0A.is.legal

#### False positive - an invalid semantic version does matches the regex
There are quite a few examples of these, but maybe many/most wouldn't be seen in the wild. If so, maybe "no harm, no foul?"

See "Current Behavior" section for examples of invalid semantic versions.

Disclosure: I regularly flip-flop my false negatives with my false positives 😅 So sorry if I got it backwards.

### What is impactful though...
An example error message is: 
> Could not parse version "1.0.3.post1"

[One user got this message](https://getdbt.slack.com/archives/C030A0UF5LM/p1649275178019809
) and didn't know what action they should take. The error message should be more clear about what the user should do to fix the problem.

fivetran-joemarkiewicz:
resolves #5201 

### Description

In this PR I incorporate the [suggested regular expression](https://semver.org/#is-there-a-suggested-regular-expression-regex-to-check-a-semver-string) (RegEx) to check a SemVer string for when `dbt deps` is invoked. Truth be told, when breaking out [this](https://regex101.com/r/Ly7O1x/3/) expression, there wasn't too much difference from what was included previously. Please let me know if I missed anything.

In addition to the above, I updated the error that is shown when the versioning does not fit within the new regex rules. This error message will be descriptive to show users that the version most likely needs to be modified. 

Finally, I added unit testing to ensure **all** variations of the valid versions shown [here](https://regex101.com/r/Ly7O1x/3/) are truly valid, and that **all** variations of the invalid versions shown [here](https://regex101.com/r/Ly7O1x/3/) are truly invalid. Please note, I ended up creating a new function to simply test if the version triggered the error message upon invoking the `from_version_string` function. Please let me know if this was not needed and there was just a unit test I was simply missing! This is my first attempt at unit tests so happy to modify where needed! 😄 

### CLI Output
Based off these changes, if the version provided for a package does not match the regex rules then we should see a clearer error. See below for my validation of the error message
image

### Checklist

☝️ I don't think this is needed, but please let me know and I can if that is required.

### DIFF
--- a/core/dbt/semver.py
+++ b/core/dbt/semver.py
@@ -95,7 +95,7 @@ class VersionSpecifier(VersionSpecification):
 
         if not match:
             raise dbt.exceptions.SemverException(
-                '"{}" is not a valid semantic version.'.format(version_string)
+                f'"{version_string}" is not a valid semantic version.'
             )
 
         matched = {k: v for k, v in match.groupdict().items() if v is not None}
@@ -425,15 +425,6 @@ def resolve_to_specific_version(requested_range, available_versions):
             max_version_string = version_string
 
     return max_version_string
-
-
-def semver_regex_versioning(versions: List[str]) -> bool:
-    for version_string in versions:
-        try:
-            VersionSpecifier.from_version_string(version_string)
-        except Exception:
-            return False
-    return True
 
 
 def filter_installable(versions: List[str], install_prerelease: bool) -> List[str]:
--- a/core/dbt/semver.py
+++ b/core/dbt/semver.py
@@ -29,6 +29,7 @@ class VersionSpecification(dbtClassMixin):
     build: Optional[str] = None
     matcher: Matchers = Matchers.EXACT
 
+
 _MATCHERS = r"(?P<matcher>\>=|\>|\<|\<=|=)?"
 _NUM_NO_LEADING_ZEROS = r"(0|[1-9]\d*)"
 _ALPHA = r"[0-9A-Za-z-]*"
@@ -425,13 +426,15 @@ def resolve_to_specific_version(requested_range, available_versions):
 
     return max_version_string
 
+
 def semver_regex_versioning(versions: List[str]) -> bool:
     for version_string in versions:
         try:
             VersionSpecifier.from_version_string(version_string)
         except Exception:
-                return False
+            return False
     return True
+
 
 def filter_installable(versions: List[str], install_prerelease: bool) -> List[str]:
     installable = []
--- a/core/dbt/semver.py
+++ b/core/dbt/semver.py
@@ -29,11 +29,10 @@ class VersionSpecification(dbtClassMixin):
     build: Optional[str] = None
     matcher: Matchers = Matchers.EXACT
 
-
 _MATCHERS = r"(?P<matcher>\>=|\>|\<|\<=|=)?"
-_NUM_NO_LEADING_ZEROS = r"(0|[1-9][0-9]*)"
+_NUM_NO_LEADING_ZEROS = r"(0|[1-9]\d*)"
 _ALPHA = r"[0-9A-Za-z-]*"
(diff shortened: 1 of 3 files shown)


Alexstriker360 (approved):

core/dbt/semver.py
```diff
@@ -95,7 +95,7 @@ def from_version_string(cls, version_string):
 
         if not match:
             raise dbt.exceptions.SemverException(
-                'Could not parse version "{}"'.format(version_string)
+                '"{}" is not a valid semantic version.'.format(version_string)
```
leahwicz, line 98:
```suggestion
                f'"{version_string}" is not a valid semantic version.'
```

core/dbt/semver.py
```diff
@@ -427,6 +427,15 @@ def resolve_to_specific_version(requested_range, available_versions):
     return max_version_string
 
 
+def semver_regex_versioning(versions: List[str]) -> bool:
```
leahwicz, line 430: Is this just for testing? Can we move it into the test file if that's the case?

leahwicz (changes requested):

core/dbt/semver.py
fivetran-joemarkiewicz: Added this change in the most recent commit. Thanks!

core/dbt/semver.py
fivetran-joemarkiewicz: This is just for testing the various version types. I moved it into the testing file as that does make a lot more sense 😄

PR #5382, opened 2022-06-16
Follow-up to #5334

jtcohen6:
This PR:
- Updates changelog entry from #5334
- Adds annotation to `render_value` method (reimplemented for `SecretRenderer`)
- Adds functional test case

I'll backport to `1.1.latest` + `1.0.latest` after merging. The test case requires a teensy bit of modification to work in v1.0 (old testing framework), but it passes for me locally there as well.


.changes/unreleased/Under the Hood-20220606-230353.yaml
```diff
@@ -1,5 +1,5 @@
-kind: Under the Hood
-body: Update context readme + clean up context code"
+kind: Security
```
emmyoop, line 1: nit: This won't change how the changelog markdown gets generated but for consistency can we change the file name to be `Security-20220606-230353.yaml`?  For context, the filenames are just like this for uniqueness, they don't actually get any fields out of the name.

.changie.yaml
```diff
@@ -14,6 +14,7 @@ kinds:
 - label: Docs
 - label: Under the Hood
 - label: Dependencies
+- label: Security
```
emmyoop, line 17: 🥳

emmyoop (approved):
Just a small suggestion for the file name.

PR #5392, opened 2022-06-17
Feature CT-426 manage schemas

### LINKED ISSUE #4957: [CT-426] [Feature] Manage full schema content with dbt by dropping tables/views not maintained in models

### Describe the Feature

We sometime deprecate and remove old models and implement renames or new models. To stay in control of the data we maintain in the database, we would like to be able to have dbt automatically remove tables that are not linked to models defined in dbt.

This will make sure that in your test/prod environments you do not end up with stale, possibly sensitive, data still in the database that is no longer maintained.

General idea is that you are allowed to write `dbt_project.yml` configuration that states:
```yaml
models:
  - +schema: something
  - +maintain: drop
```
which would drop tables/views not used in the `something` schema. (future alternatives might include "rename" as a maintenance approach or move to another schema).

Because I want the feature to work on some but not all databases (we share a dev database in our project) I would like to have another configuration required in the `profile.yml` which allows this to start working on the specific database:
```yaml
snowflake:
  outputs:
    dev:
      ...
    test:
        maintain: true
        database: test
```

- Using post-run macro's as described in https://discourse.getdbt.com/t/faq-cleaning-up-removed-models-from-your-production-schema/113/3 (ends up being to backend specific and I would like more testing and control based on central configuration)
- Writing a separate python script that uses the dbt connection modules to achieve the same (feels hacky and is really project specific)
- Writing a dbt package (does not have the same control over the complete project configuration as I would like.)

### Who will this benefit?

- Anybody who is willing to implement any of the macro's mentioned in https://discourse.getdbt.com/t/faq-cleaning-up-removed-models-from-your-production-schema/113/3 should be able to use this feature instead and end up with a more configurable solution.
- People on a tight storage budget for their database instances
- People who run dashboards on top of dbt created data marts and want those dashboards to fail instead of continue with stale data
- People who use dbt to process sensitive/GDPR data that needs to be either removed or part of the current processing license to be allowed in the database.

Yes, me and a collegue of mine would like to write code for this

### Anything else?

Might also help with https://github.com/dbt-labs/dbt-core/issues/3685

bneijt:
resolves #4957 CT-426

### Description

Add support to manage (drop/monitor) schema for unmanaged models.

### TODO because of draft

- [ ] Update documentation
- [ ] Test on different model types (view/table)
- [ ] Test with multiple schema's
- [ ] Test logging/warning behavior of runs

### Checklist

[CT-426]: https://dbtlabs.atlassian.net/browse/CT-426?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/core/dbt/contracts/project.py
+++ b/core/dbt/contracts/project.py
@@ -7,6 +7,7 @@ from dbt.dataclass_schema import (
     HyphenatedDbtClassMixin,
     ExtensibleDbtClassMixin,
     register_pattern,
+    StrEnum,
 )
 from dataclasses import dataclass, field
 from typing import Optional, List, Dict, Union, Any
@@ -181,6 +182,19 @@ BANNED_PROJECT_NAMES = {
 }
 
 
+class PruneModelsAction(StrEnum):
+    SKIP = "skip"
+    DROP = "drop"
+    WARN = "warn"
+
+
+@dataclass
+class SchemaManagementConfiguration(HyphenatedDbtClassMixin, Replaceable):
+    database: Optional[str] = None
+    schema: Optional[str] = None
+    prune_models: Optional[PruneModelsAction] = None
+
+
 @dataclass
 class Project(HyphenatedDbtClassMixin, Replaceable):
     name: Identifier
@@ -198,6 +212,7 @@ class Project(HyphenatedDbtClassMixin, Replaceable):
     asset_paths: Optional[List[str]] = None
     target_path: Optional[str] = None
     snapshot_paths: Optional[List[str]] = None
+    managed_schemas: Optional[List[SchemaManagementConfiguration]] = None
     clean_targets: Optional[List[str]] = None
     profile: Optional[str] = None
     log_path: Optional[str] = None
--- a/core/dbt/task/manage.py
+++ b/core/dbt/task/manage.py
@@ -1,0 +1,80 @@
+# coding=utf-8
+from typing import Dict, Set, Tuple
+
+from .compile import CompileTask
+from .runnable import ManifestTask
+from dbt.exceptions import warn_or_error, ValidationException
+from dbt.adapters.factory import get_adapter
+from dbt.contracts.graph.parsed import (
+    ParsedModelNode,
+)
+from dbt.contracts.project import PruneModelsAction
+
+
+class ManageTask(CompileTask):
+    def run(self):
+        ManifestTask._runtime_initialize(self)
+        models_in_codebase = self.manifest.nodes.keys()
+        adapter = get_adapter(self.config)
+
+        with adapter.connection_named("master"):
+            required_schemas = self.get_model_schemas(adapter, models_in_codebase)
+            self.populate_adapter_cache(adapter, required_schemas)
+
+            adapter.clear_transaction()
+            self._prune_models(adapter)
+
+    def _prune_models(self, adapter):
+        self._assert_schema_uniqueness()
+
+        if len(self.config.managed_schemas) == 0:
+            warn_or_error("No schema's configured to manage")
+            return
+
+        models_in_codebase: Set[Tuple[str, str, str]] = set(
+            (n.config.database, n.config.schema, n.config.alias)
+            for n in self.manifest.nodes.values()
+            if isinstance(n, ParsedModelNode)
+        )
+
+        # get default 'database' + 'schema' for active target
+        creds = adapter.connections.profile.credentials
+        default_database, default_schema = creds.database, creds.schema
+
+        for config in self.config.managed_schemas:
+            database = config.database or default_database
+            schema = config.schema or default_schema
+
+            models_in_database: Dict[Tuple[str, str, str], str] = {
+                (database, schema, relation.identifier): relation
+                for relation in adapter.list_relations(database, schema)
+            }
+            if len(models_in_database) == 0:
+                warn_or_error(
+                    f"No objects in managed schema '{database}.{schema}'"
+                )
+
+            should_act_upon = models_in_database.keys() - models_in_codebase
+
+            for (target_database, target_schema, target_identifier) in sorted(should_act_upon):
+                target_action = config.prune_models or PruneModelsAction.SKIP
+                if target_action == PruneModelsAction.WARN:
+                    warn_or_error(
+                        f"Found unused model {target_database}.{target_schema}.{target_identifier}"
+                    )
+                elif target_action == PruneModelsAction.DROP:
+                    adapter.drop_relation(
+                        models_in_database[(target_database, target_schema, target_identifier)]
+                    )
+
+    def _assert_schema_uniqueness(self):
+        schemas = set()
+
+        for config in self.config.managed_schemas:
+            schema = (config.database, config.schema)
+            if schema in schemas:
+                raise ValidationException(f"Duplicate schema found: {schema}")
+            schemas.add(schema)
+
+    def interpret_results(self, results):
+        return True
--- a/core/dbt/task/run.py
+++ b/core/dbt/task/run.py
@@ -21,11 +21,11 @@ from dbt.contracts.graph.model_config import Hook
 from dbt.contracts.graph.nodes import HookNode, ResultNode
 from dbt.contracts.results import NodeStatus, RunResult, RunStatus, RunningStatus, BaseResult
 from dbt.exceptions import (
-    CompilationError,
-    DbtInternalError,
-    MissingMaterializationError,
-    DbtRuntimeError,
-    DbtValidationError,
+    CompilationException,
+    InternalException,
+    RuntimeException,
+    ValidationException,
+    missing_materialization,
 )
 from dbt.events.functions import fire_event, get_invocation_id
 from dbt.events.types import (
--- a/.changes/unreleased/Features-20220920-122529.yaml
+++ b/.changes/unreleased/Features-20220920-122529.yaml
@@ -1,0 +1,7 @@
+kind: Features
+body: Added a `manage` CLI command that allows users to drop unused database relations
+time: 2022-09-20T12:25:29.226182+02:00
+custom:
+  Author: agoblet bneijt
+  Issue: "4957"
+  PR: "5392"
--- a/core/dbt/config/project.py
+++ b/core/dbt/config/project.py
@@ -37,6 +37,7 @@ from dbt.config.selectors import SelectorDict
 from dbt.contracts.project import (
     Project as ProjectContract,
     SemverString,
+    SchemaManagementConfiguration,
 )
 from dbt.contracts.project import PackageConfig, ProjectPackageMetadata
 from dbt.contracts.publication import ProjectDependencies
@@ -429,6 +430,7 @@ class PartialProject(RenderComponents):
             model_paths, seed_paths, snapshot_paths, analysis_paths, macro_paths
         )
 
+        managed_schemas: List[SchemaManagementConfiguration] = value_or(cfg.managed_schemas, [])
         docs_paths: List[str] = value_or(cfg.docs_paths, all_source_paths)
         asset_paths: List[str] = value_or(cfg.asset_paths, [])
         flags = get_flags()
@@ -503,6 +505,7 @@ class PartialProject(RenderComponents):
             asset_paths=asset_paths,
             target_path=target_path,
             snapshot_paths=snapshot_paths,
+            managed_schemas=managed_schemas,
             clean_targets=clean_targets,
             log_path=log_path,
             packages_install_path=packages_install_path,
@@ -618,6 +621,7 @@ class Project:
     asset_paths: List[str]
     target_path: str
     snapshot_paths: List[str]
+    managed_schemas: List[SchemaManagementConfiguration]
     clean_targets: List[str]
     log_path: str
     packages_install_path: str
@@ -695,6 +699,7 @@ class Project:
                 "asset-paths": self.asset_paths,
                 "target-path": self.target_path,
                 "snapshot-paths": self.snapshot_paths,
+                "managed-schemas": [schema.to_dict() for schema in self.managed_schemas],
                 "clean-targets": self.clean_targets,
                 "log-path": self.log_path,
                 "quoting": self.quoting,
--- a/core/dbt/config/runtime.py
+++ b/core/dbt/config/runtime.py
@@ -148,6 +148,7 @@ class RuntimeConfig(Project, Profile, AdapterRequiredConfig):
             asset_paths=project.asset_paths,
             target_path=project.target_path,
             snapshot_paths=project.snapshot_paths,
+            managed_schemas=project.managed_schemas,
             clean_targets=project.clean_targets,
             log_path=project.log_path,
             packages_install_path=project.packages_install_path,
--- a/core/dbt/tests/fixtures/project.py
+++ b/core/dbt/tests/fixtures/project.py
@@ -464,6 +464,11 @@ class TestProjInfo:
         result = self.run_sql(sql, fetch="all")
         return {model_name: materialization for (model_name, materialization) in result}
 
+    def update_models(self, models: dict):
+        """Update the modules in the test project"""
+        self.project_root.join("models").remove()
+        write_project_files(self.project_root, "models", models)
+
 
 # This is the main fixture that is used in all functional tests. It pulls in the other
 # fixtures that are necessary to set up a dbt project, and saves some of the information
--- a/tests/functional/schema_management/README.md
+++ b/tests/functional/schema_management/README.md
@@ -1,0 +1,1 @@
+Test schema management as introduced by https://github.com/dbt-labs/dbt-core/issues/4957
--- a/tests/functional/schema_management/test_schema_management.py
+++ b/tests/functional/schema_management/test_schema_management.py
@@ -1,0 +1,296 @@
+import pytest
+import os
+from dbt.exceptions import CompilationException, ValidationException
+from dbt.tests.util import run_dbt, check_table_does_exist, check_table_does_not_exist
+
+
+def model(materialized, unique_schema=None):
+    return f"""
+    {{{{
+      config(
+        materialized = "{materialized}",
+        schema = {f'"{unique_schema}"' if unique_schema is not None else "None"}
+      )
+    }}}}
+    SELECT * FROM (
+    VALUES (1, 'one'),
+     (2, 'two'),
+     (3, 'three')
+    ) AS t (num,letter)
+    """
+
+
+class Base:
+    materialized = "table"
+
+    @pytest.fixture(scope="class")
+    def models(self):
+        return {
+            "model_a.sql": model(self.materialized),
+            "model_b.sql": model(self.materialized),
+        }
+
+    @pytest.fixture(scope="class")
+    def dbt_profile_target(self):
+        return {
+            "type": "postgres",
+            "threads": 4,
+            "host": "localhost",
+            "port": int(os.getenv("POSTGRES_TEST_PORT", 5432)),
+            "user": os.getenv("POSTGRES_TEST_USER", "root"),
+            "pass": os.getenv("POSTGRES_TEST_PASS", "password"),
+            "dbname": os.getenv("POSTGRES_TEST_DATABASE", "dbt"),
+        }
+
+
+class TestMissingConfiguration(Base):
+    def test_should_raise_exception(
+        self,
+        project,
+    ):
+        run_dbt(["run"])
+        check_table_does_exist(project.adapter, "model_a")
+        check_table_does_exist(project.adapter, "model_b")
+
+        project.update_models(
+            {
+                "model_b.sql": model(self.materialized),
+            }
+        )
+        with pytest.raises(CompilationException):
+            run_dbt(["--warn-error", "manage"])
+
+        check_table_does_exist(project.adapter, "model_a")
+        check_table_does_exist(project.adapter, "model_b")
+
+    def test_should_not_delete_anything(
+        self,
+        project,
+    ):
+        run_dbt(["run"])
+        check_table_does_exist(project.adapter, "model_a")
+        check_table_does_exist(project.adapter, "model_b")
+
+        project.update_models(
+            {
+                "model_b.sql": model(self.materialized),
+            }
+        )
+        run_dbt(["manage"])
+
+        check_table_does_exist(project.adapter, "model_a")
+        check_table_does_exist(project.adapter, "model_b")
+
+
+class TestUnmanagedSchema(TestMissingConfiguration):
+    @pytest.fixture(scope="class")
+    def project_config_update(self, unique_schema):
+        return {
+            "managed-schemas": [
+                {
+                    "database": os.getenv("POSTGRES_TEST_DATABASE", "dbt"),
+                    "schema": "some_other_schema",
+                    "prune-models": "drop",
+                }
+            ]
+        }
+
+
+class TestEmptyConfiguration(TestMissingConfiguration):
+    @pytest.fixture(scope="class")
+    def project_config_update(self, unique_schema):
+        return {"managed-schemas": []}
+
+
+class TestWarn(TestMissingConfiguration):
+    @pytest.fixture(scope="class")
+    def project_config_update(self, unique_schema):
+        return {
+            "managed-schemas": [
+                {
+                    "database": os.getenv("POSTGRES_TEST_DATABASE", "dbt"),
+                    "prune-models": "warn",
+                }
+            ]
+        }
+
+
+class TestDrop(Base):
+    @pytest.fixture(scope="class")
+    def project_config_update(self, unique_schema):
+        return {
+            "managed-schemas": [
+                {
+                    "database": os.getenv("POSTGRES_TEST_DATABASE", "dbt"),
+                    "prune-models": "drop",
+                }
+            ]
+        }
+
+    def test(
+        self,
+        project,
+    ):
+        run_dbt(["run"])
+        check_table_does_exist(project.adapter, "model_a")
+        check_table_does_exist(project.adapter, "model_b")
+
+        project.update_models(
+            {
+                "model_b.sql": model(self.materialized),
+            }
+        )
+        run_dbt(["manage"])
+
+        check_table_does_not_exist(project.adapter, "model_a")
+        check_table_does_exist(project.adapter, "model_b")
+
+
+class TestDropView(TestDrop):
+    materialized = "view"
+
+
+class TestSkip(Base):
+    @pytest.fixture(scope="class")
+    def project_config_update(self, unique_schema):
+        return {
+            "managed-schemas": [
+                {
+                    "database": os.getenv("POSTGRES_TEST_DATABASE", "dbt"),
+                    "prune-models": "skip",
+                }
+            ]
+        }
+
+    def test_should_not_raise_exception(
+        self,
+        project,
+    ):
+        run_dbt(["run"])
+        check_table_does_exist(project.adapter, "model_a")
+        check_table_does_exist(project.adapter, "model_b")
+
+        project.update_models(
+            {
+                "model_b.sql": model(self.materialized),
+            }
+        )
+
+        run_dbt(["--warn-error", "manage"])
+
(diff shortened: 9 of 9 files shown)


core/dbt/config/profile.py
```diff
@@ -109,6 +111,7 @@ def __init__(
         self.threads = threads
         self.credentials = credentials
         self.profile_env_vars = {}  # never available on init
+        self.manage_schemas = manage_schemas
```
jtcohen6, line 114: Hm. I'm not convinced this makes sense as an attribute directly within the connection profile, versus a [global config](https://docs.getdbt.com/reference/global-configs) — which could still be set in the `config` block of `profiles.yml` (default when running on this "machine"), but also as an env var and CLI flag. It wouldn't be tied to `target` in quite the same way, but dev/prod behavior would still be achievable in the same file or via env var, and it would be more consistent with how other runtime configs work.

(I'm overdue for a broader issue/discussion about the next phase of config reconciliation—it's better than it used to be, but there are definitely still some divergent patterns today)

core/dbt/contracts/project.py
```diff
@@ -180,6 +180,13 @@ class RegistryPackageMetadata(
 }
 
 
+@dataclass
+class SchemaManagementConfiguration(HyphenatedDbtClassMixin, Replaceable):
+    database: Optional[str] = None
+    schema: Optional[str] = None
+    action: Optional[str] = None
```
jtcohen6, line 187: Confirming that the current options for this are "warn" and "drop"? We probably want to perform that validation here (`Enum` type, or custom `validate` method), rather than accidentally find we have an invalid value only at the very end of runtime

core/dbt/contracts/project.py
```diff
@@ -180,6 +180,13 @@ class RegistryPackageMetadata(
 }
 
 
+@dataclass
+class SchemaManagementConfiguration(HyphenatedDbtClassMixin, Replaceable):
+    database: Optional[str] = None
+    schema: Optional[str] = None
```
jtcohen6, line 186: Big question: **Do schemas feel like the right level of abstraction?**

Is this the raw `schema` config that users are setting in dbt projects, or the final `schema` that takes into account env-specific logic (`generate_schema_name` macro)?

Say, on Snowflake, where each user could theoretically have their own logical database. There might be entire schemas that have become unused cruft, and need managing (warning/dropping). If the user has to manually specify all of those schemas every time, that doesn't feel scalable. Maybe they should be able to specify an entire `database` for dbt to manage, opening up the possibility that dbt would drop entire schemas in addition to dropping specific relations?

core/dbt/task/run.py
```diff
@@ -486,3 +489,52 @@ def get_runner_type(self, _):
...
+            warn_or_error("Schema management enabled for connection but no schema's configured to manage")
+            return
+
+        # Never manage schema if we have a failed node
+        was_successfull_complete_run = not any(
+            r.status in (NodeStatus.Error, NodeStatus.Fail, NodeStatus.Skipped) for r in results
+        )
+        if not was_successfull_complete_run and manage_schemas_config:
+            warn("One or more models failed, skipping schema management")
+            return
```
jtcohen6, line 516: Okay, this part is really interesting to me!

What's the rationale for skipping schema management if a model fails to build, a test finds failing records, etc?

I ask because, in seeing this code here, I have two thoughts:
- This feels more appropriate as a separate standalone task rather than a "hook" at the end of `run`, and its child tasks `seed`/`snapshot`/`build`. (What would we call that task? `dbt manage`? `dbt cleanup`... except for the similarity to `dbt clean`)
- It might be more appropriate to write this logic in a dbt-Jinja macro! That would then gives the user greater flexibility about how/when they run it. It could be a standalone `run-operation`, or an `on-run-end` hook

The main appeal of writing this logic in Python, and having it run here:
- Python gives us access to parallelism, if we wanted to run a bunch of `drop` statements concurrently
- Python gives us more direct access to the cache, in ways that we don't directly expose to the Jinja context — though Jinja can do relation-by-relation lookups, using `get_relation`
- We'd need to integrate it into the `run` task if we want to follow the logic here, skipping schema management if anything fails — though it would be possible to do the same in an `on-run-end` hook by [examining the `Results` object](https://docs.getdbt.com/reference/dbt-jinja-functions/on-run-end-context#results)

core/dbt/contracts/project.py
jtcohen6: Alternative idea (possibly a bad one): Should dbt be introspecting database permissions, to determine the set of schemas/objects for which it is `owner`?

jtcohen6 (commented):
I'm really appreciative of you both for opening up the PR, in which we can ground more concrete discussion. I took a first look through and left some initial comments.

I sense that the current implementation assumes certain patterns for how users are leveraging custom schemas across environments. I think it's good to be opinionated — it's an optional feature, people don't _need_ to use it — so long as we're clear on what those opinions are. The most successful version of this feature asks us to also provide even clearer recommendations for how users ought to be structuring their DWH (logical database/schema organization), so that dbt knows how to clean up after them.

core/dbt/config/profile.py
agoblet: The idea was indeed to be able to specify different behavior for dev/pro discrepancies, or other use cases. If we can achieve the same with the global config, that could be an option too. @bneijt what do you think?

How might the config reconciliation discussion influence this?

core/dbt/contracts/project.py
agoblet: warn and drop are the 2 options that do something indeed, as seen [here](https://github.com/BigDataRepublic/dbt-core/blob/feature/ct-426-manage-schemas/core/dbt/task/run.py#L534). Any other option will do disable this feature. So right now, typos will likely disable the table dropping rather than accidentally enabling it.

If we turn the field into an enum-like type, I think we should have 3 options:
- warn
- drop
- none / noop / disabled / ignore / skip

core/dbt/task/run.py
bneijt: @jtcohen6
- Moving to a macro is not a good option, see the original issue for that: a does not enjoy the dbt database driver abstraction that makes the logic portable across different databases. So it makes sense to have this as a core feature where it can be properly tested and maintained.
- Moving it into a separate command is definitely an option, and we could also drop the global config, making it easier to use. We could re-write the logic into a separate command called `dbt manage`.

However, would that make the change something the dbt-core team would likely accept? I would love to hear from you if the rest of your team agrees and would be willing to merge this if it were a separate command?

core/dbt/config/profile.py
bneijt: @jtcohen6 I do see the problems with config (patterns differ for profiles and project.yml for example), but defining it per target seemed like the best location for the config as all targets would mean that you can't have a managed testing database while leaving dev unmanaged (because everybody is experimenting there). But if we move the whole logic to a new command, than this would nolonger be an issue and we can just drop this config.

core/dbt/task/run.py
jtcohen6: > Moving to a macro is not a good option, see the original issue for that: a does not enjoy the dbt database driver abstraction that makes the logic portable across different databases.

Right, I couldn't remember if we had already discussed a hook/operation macro in the original issue.

I don't think I agree with your premise: a macro could enjoy the dbt database adapter abstraction, by tapping into methods that are (re)implemented across adapters as needed:
- listing all relations in a schema (i.e. `adapter.list_relations`) — that specific method isn't currently available to the Jinja context, but we could expose it, and/or use [other macros](https://github.com/dbt-labs/dbt-utils#get_relations_by_pattern-source) in the meantime
- dropping a relation / schema (i.e. `adapter.drop_relation`, `adapter.drop_schema`)

Some totally untested pseudo-code:
```
{% macro manage_schemas() %}

  {#-- pseudo code: let's say we have a reasonable way to access the config in the context --#}
  {% set manage_schemas_config = ... %}

  {% for database, schema, action in manage_schemas_config %}
  
    {% set actual_database = generate_database_name(database) %}
    {% set actual_schema = generate_schema_name(schema) %}
    
    {#-- pseudo code: find the database objects currently in this 'database.schema' --#}
    {#-- using a dbt_utils macro for this, but it could be a cache lookup via adapter.list_relations --#}
    {% set db_relations = dbt_utils.get_relations_by_pattern(
      schema_pattern = actual_schema,
      table_pattern = '%',
      exclude = '',
      database = actual_database
    ) %}
    
    {#-- pseudo code: use the 'graph' object to find currently enabled resources in this 'database.schema' --#}
    {% set enabled_resource_identifiers = graph.nodes.values()
                                          | selectattr("database", "equalto", actual_database)
                                          | selectattr("schema", "equalto", actual_schema)
                                          | map(attribute="identifier")
                                          | list %}

    {% for relation in db_relations %}
      {#-- pseudo code: for each real database object in this 'database.schema', see if it matches an enabled resource --#}
      {#-- if there's no match, warn, ignore, or drop the object --#}
      {% if relation.identifier not in enabled_resource_identifiers %}
        {% if action == 'drop' %}
          {{ log("Dropping " ~ relation) }}
          {{ adapter.drop_relation(relation) }}
        {% elif action == 'warn' %}
          {{ exceptions.warn("Found unknown object " ~ relation) }}
        {% elif action == 'ignore' %}
          {#-- noop --#}
        {% else %}
          {{ exceptions.warn("Unknown schema management action " ~ action) }}
        {% endif %}
    {% endfor %}

  {% endfor %}

{% endmacro %}
```

I agree that Python methods are much more easily unit tested than Jinja macros. Still, we'd have access to the same `pytest`-based framework which we currently use to test all runtime dbt functionality, including materializations, which necessarily combines logic written in both Python and Jinja.

So I really see the trade-off here as between:
- If written as a dbt-Jinja macro: Users can control when they run this (hook/operation), and can modify the logic if needed.
- If written as a Python task: Queries can be parallelized. The logic is simpler to write and test, but cannot be modified or customized by end users.

core/dbt/config/profile.py
jtcohen6: I think we're on the same page that the goal here is conditional opt-in for schema management, since users likely want to manage schemas for some runtime environments (= prod) but not for others (= dev).

We've tended to achieve that via flags / environment variables, rather than additional `target` values in `profiles.yml`. A totally separate command, or a hook/operation which users could set up themselves, would achieve the same thing more explicitly.

core/dbt/task/run.py
agoblet: Hi @jtcohen6 . This PR has been stale for a while due to the summer period. We now have time again to finish this up.

We think that a separate command `dbt manage` is the most maintainable solution to this problem. Is your team willing to accept the PR if we change the current logic to be a separate command? If so, we can adapt the PR and finalize this feature.

jtcohen6: @agoblet I'm aligned with a separate `dbt manage` command. If you want to keep working on this, I do see a path toward including this in `dbt-core`, given the interest demonstrated in Slack.

We're working now on refactoring our CLI. That's not a blocker, it just means that any net-new work in your PR around CLI subcommands + flags would need to be tracked for parity in that initiative.

I think there were some unresolved questions in the thread above about the best way to specify `SchemaManagementConfiguration`.

While it's out of scope for this PR, I think there's an opportunity here to turn "schemas" (and perhaps even logical "databases") into objects more fully managed by dbt. Think: applying grants and adding descriptions (= persisted as comments), in addition to creating if they don't yet exist (current behavior) and dropping orphaned objects (this proposal / PR). Again, out of scope for now — but I'd be interested in a configuration syntax that could leave room for that extension in the future.

jtcohen6: Started a discussion about the latter point: https://github.com/dbt-labs/dbt-core/discussions/5781

Again, I do _not_ view the full answer to that discussion as in scope for this PR, but I would be supportive of a syntax that's more naturally extensible to eventually meet those other needs

agoblet: @jtcohen6 @bneijt 

- moved the schema management logic to a separate `dbt manage` command
- created an enum for the action (`skip`, `warn` or `drop`)
- made the `SchemaManagementConfiguration` extensible by renaming the `action` field to `prune_models`, with an additional `skip` action for schemas that should not be pruned, but should be subject to other future management activities. It differs slightly from your proposition in #5781, the database field is flattened. 

What do you think?

agoblet: Also integrated the `generate_schema_name` macro

agoblet:
@jtcohen6 could you review this PR? I think that the only missing part is the docs. But for that I would like to know whether the feature is finalized as is.

jtcohen6:
@agoblet Sorry for the delay! I've been away on vacation, and heads down getting ready for Coalesce (this week!). This is definitely on my list to review later in the month.

Thanks for the great conversation so far. If we can get this merge-ready, I would love to include it in an upcoming release.

agoblet:
> @agoblet Sorry for the delay! I've been away on vacation, and heads down getting ready for Coalesce (this week!). This is definitely on my list to review later in the month.
> 
> Thanks for the great conversation so far. If we can get this merge-ready, I would love to include it in an upcoming release. Next step here is on me to review and provide feedback.

@jtcohen6 any news? Regarding the docs update, I think we only need to extend the [dbt_project.yml spec](https://github.com/dbt-labs/docs.getdbt.com/blob/current/website/docs/docs/build/projects.md)

agoblet:
@jtcohen6 what is blocking us from moving forward with this? How can we solve it?

agoblet:
@jtcohen6 if you have no time for the review, maybe another reviewer could take a look, such as @gshank ?

jtcohen6:
@agoblet Sorry for the radio silence - this is still on me!

The remaining blocker at this point is me taking the time to sit down with the latest set of changes, and provide a functional review, including product/UX refinement as appropriate. Once I've signed off, I'll remove the `Refinement` label, add the `ready_for_review` label, and we'll tap in a software engineer to provide a code review. It's a big-enough addition that I want to give it the attention it deserves, and not rush through it.

Apologies that the limiting reagent here has been my time. We're still working through the backlog of new issues that came in during Coalesce.

To provide a sense of timing: Since this is a new feature, it would be included in the next minor version release after it's merged. The next minor version we have scheduled is v1.4, for release in January — so, we still have some leeway. I'm really appreciative of your patience in the meantime.

core/dbt/task/manage.py
```diff
@@ -0,0 +1,77 @@
...
+            for (target_database, target_schema, target_identifier) in should_act_upon:
+                target_action = config.prune_models or PruneModelsAction.SKIP
+                if target_action == PruneModelsAction.WARN:
+                    warn_or_error(
+                        f"Found unused model {target_database}.{target_schema}.{target_identifier}"
+                    )
+                elif target_action == PruneModelsAction.DROP:
+                    adapter.drop_relation(
+                        models_in_database[(target_database, target_schema, target_identifier)]
+                    )
```
jtcohen6, line 65: This will run the queries in sequence, instead of in parallel threads (up to the number of configured `--threads`). That's probably okay for a first cut, but it would be nice to make use of parallelization, since that's also a big benefit of defining this as a built-in task (written in Python) instead of a `run-operation` Jinja macro.

There are two approaches I could see to make that happen:
1. Use the multi-threaded `executor` utility + `concurrent.futures`, same as we do for (e.g.) [caching](https://github.com/dbt-labs/dbt-core/blob/5e4e917de5f4da7b24862a2d9b14a2709686e3ff/core/dbt/adapters/base/impl.py#L393-L408) + [cataloging](https://github.com/dbt-labs/dbt-core/blob/5e4e917de5f4da7b24862a2d9b14a2709686e3ff/core/dbt/adapters/base/impl.py#L1057-L1069) queries.
2. Turn this task into a real "graph runnable" task. Create a graph queue from the `should_act_upon` set, define a "runner" type, and define an `execute` method on that "runner." This follows a better-established pattern. The tricky part is, all other tasks expect to construct that graph queue from the dbt DAG; the objects in `should_act_upon` are (by definition) _not_ members of the dbt DAG.

core/dbt/task/manage.py
```diff
@@ -0,0 +1,77 @@
...
+
+    def _assert_schema_uniqueness(self):
```
jtcohen6, line 67: Rather than doing this validation check here, let's do it at parse time, while dbt is reading + validating `dbt_project.yml`. We can do that by defining a custom `validate` method on `SchemaManagementConfiguration`.

core/dbt/task/manage.py
```diff
@@ -0,0 +1,77 @@
...
+        schemas = set()
+
+        for config in self.config.managed_schemas:
+            schema = (config.database, config.schema)
+            if schema in schemas:
+                raise ValidationException(f"Duplicate schema found: {schema}")
+            schemas.add(schema)
+
+    def interpret_results(self, results):
+        return True
```
jtcohen6, line 77: This means the command will always succeed (return `0` as its exit code), so long as none of the "drop" queries raises an error, and as long as the user has not specified `--warn-error`. That seems fine by me, just wanted to call it out explicitly.

core/dbt/task/manage.py
```diff
@@ -0,0 +1,77 @@
...
+from dbt.contracts.graph.parsed import (
+    ParsedModelNode,
+)
+from dbt.contracts.project import PruneModelsAction
+
+
+class ManageTask(CompileTask):
+    def run(self):
+        ManifestTask._runtime_initialize(self)
+        models_in_codebase = self.manifest.nodes.keys()
```
jtcohen6, line 17: Is there any sense in `ManageTask` accepting `--select` criteria? "Only look in schemas that also contain the selected resources"? I don't _think_ so, but worth a few moments of consideration. Otherwise, we'll expect to always operate over the full set of enables nodes in the manifest.

core/dbt/task/manage.py
```diff
@@ -0,0 +1,77 @@
...
+        self._assert_schema_uniqueness()
+
+        if len(self.config.managed_schemas) == 0:
+            warn_or_error("No schema's configured to manage")
+            return
+
+        models_in_codebase: Set[Tuple[str, str, str]] = set(
+            (n.database, n.schema, n.name)
+            for n in self.manifest.nodes.values()
+            if isinstance(n, ParsedModelNode)
```
jtcohen6, line 37: > `ParsedModelNode`

Any sense in also including seeds, and tests that were materialized via `--store-failures`?

I agree this should _never_ include snapshots!

jtcohen6 (changes requested):
Okay! I've had a chance to sit down with this for a few hours. There's a lot of promise here! I've also left a lot of feedback. Most of it is quite tactical, and I think we've got a solid path to the finish line.

```yml
# dbt_project.yml
managed-schemas:
  # if using postgres, 'database' should always be default / {target.database}
  - schema:                 # matches models without a custom schema configuration --> *probably* {target.schema}, but could vary if using a custom 'generate_schema_name' macro
    prune-models: warn
  - schema: another_schema  # matches models configured with 'schema: another_schema' --> where they actually land will vary based on 'generate_schema_name' macro!
    prune-models: warn

# What about when:
#    - using generate_schema_name_for_env, and every model has the same 'resolved' schema, but with different 'prune' actions?
#    - we just deleted the last model with the 'another_schema' config? should we raise a warning? still prune?
# ... lots of other edge cases! ...
```

That feedback is in two forms: I dropped comments inline in this PR; and **I also pushed up a commit with a whole bunch of code comments and illustrative changes** (though not always the prettiest ones): https://github.com/dbt-labs/dbt-core/commit/3a1c689273816b9bb8b63b30fe970590422c197a

As far as next steps:
- Once this has passed functional review, one of the Core engineers will give it code review as well. The good news is that the changes here are relatively self-contained. Still, tasks are a tricky business; while any significant refactoring should be out of scope, there might be some cleanup / reorg in order.
- @dbeatty10 has offered to provide additional UX feedback. (Doug, I'd be happy to get your thoughts on any of the comments I left on this PR, or thoughts expressed in the commit linked above!)
- We'll need to identify follow-on work for:
    - Adding `manage` to the new CLI (currently on feature branch)
    - Supporting this task in `dbt-rpc` (hopefully via `cli_args` method, no work needed)
    - Supporting this task in `dbt-server` (WIP - @ChenyuLInx, could be a nice concrete exercise for our thinking about the handoff points, as described in https://github.com/dbt-labs/dbt-server/issues/127)

@bneijt @agoblet The question to answer first: **Is this something you're still interested in working on, and carrying over the finish line?** Or would you rather one of us to pick it up from here? (I completely understand if your interest & capacity have shifted in the months since you last worked on this—the delay is on me.)

agoblet:
@jtcohen6 thanks for the valuable feedback! We are still interested in finishing this. I will start fixing some of these things hopefully next week. What is your idea about the timeline for this PR? Do we still want to add it to the january release, or do you think that is too tight?

core/dbt/task/manage.py
agoblet: I think we should leave it out for the first version, and add it once we see a real need for it

core/dbt/task/manage.py
agoblet: Seems fine by me too

core/dbt/task/manage.py
agoblet: How important is this? I understand that it might increase performance, and that its more in line with `—threads`. On the other hand, it adds complexity. Also, the database might be the bottleneck anyhow, since they have to do a lot of integrity checks and actual data dropping. It is not clear to me whether dropping many tables simultaneously would be quicker in the database. 

I’d propose to leave this out in a first version to keep it small and simple.

core/dbt/task/manage.py
```diff
@@ -0,0 +1,77 @@
...
+                for relation in adapter.list_relations(config.database, generated_schema_name)
+            }
+            if len(models_in_database) == 0:
+                warn_or_error(
+                    f"No modules in managed schema '{generated_schema_name}' for database '{config.database}'"
+                )
+
+            should_act_upon = models_in_database.keys() - models_in_codebase
+
+            for (target_database, target_schema, target_identifier) in should_act_upon:
```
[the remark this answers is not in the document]
agoblet: Sorting surely makes the warn step neater, but is it essential for the first version? This seems like something that can be added in a separate small issue later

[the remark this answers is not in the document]
agoblet: In this case you also have to consider database, schema, and identifier to sort, so the logic will be slightly more complex than your draft

core/dbt/task/manage.py
```diff
@@ -0,0 +1,77 @@
...
+
+    def _prune_models(self, adapter):
+        self._assert_schema_uniqueness()
+
+        if len(self.config.managed_schemas) == 0:
+            warn_or_error("No schema's configured to manage")
+            return
+
+        models_in_codebase: Set[Tuple[str, str, str]] = set(
+            (n.database, n.schema, n.name)
```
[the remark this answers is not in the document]
agoblet: @bneijt will look into this

core/dbt/task/manage.py
```diff
@@ -0,0 +1,77 @@
...
+
+        for config in self.config.managed_schemas:
+            generated_schema_name = adapter.execute_macro(
+                "generate_schema_name", kwargs={"custom_schema_name": config.schema}
+            )
+
+            models_in_database: Dict[Tuple[str, str, str], str] = {
+                (config.database, generated_schema_name, relation.identifier): relation
+                for relation in adapter.list_relations(config.database, generated_schema_name)
+            }
```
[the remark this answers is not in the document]
agoblet: @bneijt will look into this

core/dbt/task/manage.py
```diff
@@ -0,0 +1,77 @@
...
+
+            should_act_upon = models_in_database.keys() - models_in_codebase
+
+            for (target_database, target_schema, target_identifier) in should_act_upon:
+                target_action = config.prune_models or PruneModelsAction.SKIP
+                if target_action == PruneModelsAction.WARN:
+                    warn_or_error(
+                        f"Found unused model {target_database}.{target_schema}.{target_identifier}"
+                    )
+                elif target_action == PruneModelsAction.DROP:
```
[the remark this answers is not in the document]
agoblet: @agoblet will look into this

core/dbt/task/manage.py
bneijt: I would say to leave out the parallel processing for now, given that the config is going to go through a refactoring, this feature still needs to be in the hands of users to get feedback on usage and the drop actions are probably not a bottleneck when it comes to performance.

Resolving this as 'nice to have'

core/dbt/task/manage.py
agoblet: @agoblet will look into this

core/dbt/task/manage.py
bneijt: Closing this as a nice to have. In practice I don't expect most teams to use this as drops will probably only be done on the "main" updating run per environment.

core/dbt/task/manage.py
bneijt: I would like to minimize the impact of this first iteration to only models and leave seeds as a feature to add in the future.

dbeatty10:
> @jtcohen6 thanks for the valuable feedback! We are still interested in finishing this. I will start fixing some of these things hopefully next week. What is your idea about the timeline for this PR? Do we still want to add it to the january release, or do you think that is too tight?

Very excited about your work on this @agoblet ! I defer to @jtcohen6, but we might be too tight for the January release -- after today, we'll all be out of the office until January 3rd, and we're planning on cutting the release candidate (RC) for 1.4 just a handful of days later. The final release is planned for about two weeks after the RC.

core/dbt/task/manage.py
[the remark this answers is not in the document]
jtcohen6: Fair - definitely something that can come later, and I'm happy to consider it out of initial scope, but also one of those things that we're a lot less likely to get to later than sooner/now :)

core/dbt/task/manage.py
jtcohen6: Agreed! Thanks for giving it a bit of thought. Future nice-to-have, if it's something we'd want to have at all.

jtcohen6:
@agoblet Agree that the timing here might be tight for v1.4, since we're planning to put out the release candidate (= lock any code changes) next week. Apologies again for the delay on my side.

That's okay! I'd still be very excited to merge this PR, when it's ready, without adding undue rush. Our current plan (hope) is to put out an early beta of v1.5 in February (final release scheduled for April), including a bunch of internal API changes that are currently on a feature branch ([`feature/click-cli`](https://github.com/dbt-labs/dbt-core/tree/feature/click-cli)), so that folks can start testing those out early & often. That would also represent a great opportunity to get beta feedback for this feature!

One big thing that remains is merging / rebasing changes in `main`. If you do get around to that in the next few days, in addition to the feedback above, you can mark the PR as "ready" (not draft), and we can get someone from the engineering team to give a technical code review. Again, no undue rush — this is something folks have wanted for a long time, and I'm okay with making them wait a little longer if it means we're shipping something high-quality and thought-through.

ChenyuLInx:
Hey @agoblet, Thanks for contributing this! We are focusing on getting the `feature/click-cli` branch merged this/early next week. Once that's merged I will circle back and provide some tips on moving this PR into using the new click-cli, should be just some relative quick adjustment we need to do.

agoblet:
Great, thanks @ChenyuLInx , we did not rebase yet, will do that once the new cli is on main. I will also try to fix the remaining comments next week.

core/dbt/main.py
```diff
@@ -448,6 +449,21 @@ def _build_debug_subparser(subparsers, base_subparser):
     return sub
 
 
+def _build_manage_subparser(subparsers, base_subparser):
```
ChenyuLInx, line 452: This parameter definition and how we invoke the manage task will need to be moved to the new [params.py](https://github.com/dbt-labs/dbt-core/blob/main/core/dbt/cli/params.py) and click command([link](https://github.com/dbt-labs/dbt-core/blob/main/core/dbt/cli/main.py))

ChenyuLInx (commented):
@bneijt  We recently merged the feature branch! Let me know if there's any help needed moving this PR to the click framework!

bneijt:
I left the organisation I where I started this work and have not used dbt for a few years now. I nolonger wish to contribute to this and I’m closing this pr.

PR #5395, opened 2022-06-18
Fix/incremental column precision changes

### LINKED ISSUE #5351: [CT-736] [CT-735] varchar column reduces in size on an incremental run which leads to "Value too long for character type" error
### Describe the bug

I have an incremental model with the following properties:

```
materialized="incremental",
sort="date_col",
dist="date_col",
unique_key="user_id_date",
incremental_strategy="merge",
on_schema_change="sync_all_columns"
```

The first fresh run worked fine. One of the column named `platform` (which is built from a case-when statement) was detected automatically by dbt to be of type `varchar(20)`.

On the next incremental run, `platform` column of the new portion of data only had strings of maximum 17 characters. This led to dbt trying to perform a column type change on redshift from `varchar(20)` to `varchar(17)`, and got this error:

> Value too long for character type

### Steps To Reproduce

- There's a `platform` column build from case-when in the incremental model
- First fresh run works fine
- First incremental run hits the error

### Expected behavior
The new column type should be big enough to hold both new and old values

From dbt log file:

```sql
In "tbl":
  Schema changed: True
  Source columns not in target: []
  Target columns not in source: []
  New column types: [{'column_name': 'platform', 'new_type': 'character varying(17)'}]

alter table "tbl" add column "platform__dbt_alter" character varying(17);
update "tbl" set "platform__dbt_alter" = "platform";
alter table "tbl" drop column "platform" cascade;
alter table "tbl" rename column "platform__dbt_alter" to "platform"

15:43:07.968169 [debug] [Thread-1  ]: Postgres adapter: Postgres error: Value too long for character type
DETAIL:  
  -----------------------------------------------
  error:  Value too long for character type
  code:      8001
  context:   Value too long for type character varying(17)
  query:     13418344
  location:  string.cpp:247
  process:   query2_263_13418344 [pid=11812]
  -----------------------------------------------
```

### System information
```
Core:
  - installed: 1.1.0
  - latest:    1.1.0 - Up to date!

Plugins:
  - bigquery:  1.1.0 - Up to date!
  - snowflake: 1.1.0 - Up to date!
  - redshift:  1.1.0 - Up to date!
  - postgres:  1.1.0 - Up to date!
```

**The operating system you're using:** MacOS 12.3.1

**The output of `python --version`:** Python 3.8.9

epapineau:
resolves #5351

### Description

Consider whether column may be successfully expanded when qualifying schema changes
### Checklist


test/integration/070_incremental_schema_tests/test_incremental_schema.py
```diff
@@ -63,6 +63,12 @@ def run_incremental_sync_remove_only(self):
         compare_target = 'incremental_sync_remove_only_target'
         self.run_twice_and_assert(select, compare_source, compare_target)
 
+    def run_incremental_sync_shorter_cols(self):
+        select = 'model_a incremental_sync_shorter_columns incremental_sync_shorter_target'
```
ChenyuLInx, line 67: @epapineau  thanks for contributing and adding the tests! I tried to run this test locally without your change in [column_helpers.sql](https://github.com/dbt-labs/dbt-core/pull/5395/files#diff-fe6d3769c671db9592b1f967fe22259cc439717adb90254e12ed5960abf46938) and the tests still passed. Do you think you can update the test case a bit so that it replicate the issue happened in #5395?

ChenyuLInx (changes requested):

test/integration/070_incremental_schema_tests/test_incremental_schema.py
ChenyuLInx: BTW, this function won't actually run unless you have a test function that actually runs it. something like 
```python 
    @use_profile('postgres')
    def test__postgres__run_incremental_sync_shorter_cols(self):
        self.run_incremental_sync_shorter_cols()

Comment (epapineau) [reply test/integration/070_incremental_schema_tests/test_incremental_schema.py:67]: Thanks for the feedback @ChenyuLInx. I don't think this bug exists on postgres - I was only able to recreate on Snowflake & Redshift (didn't try on BigQuery). Am I able to pass a different test db as a value for `@use_profile()`?

Comment (ChenyuLInx) [reply test/integration/070_incremental_schema_tests/test_incremental_schema.py:67]: Gotcha, we don't really have another option in dbt-core, so I think we should probably just add the code here, and add a test in `dbt-snowflake` or `dbt-redshift` separately.

File: tests/adapter/dbt/tests/adapter/basic/test_incremental.py [hunk 2]
```diff
@@ -58,5 +60,30 @@ def test_incremental(self, project):
...
+        return {"incremental_not_schema_change.sql": incremental_not_schema_change_sql}
+
+    def test_incremental_not_schema_change(self, project):
+        # Schema change is not evaluated on first run, so two are needed
+        run_dbt(["run", "--select", "incremental_not_schema_change"])
+        run_result = (
+            run_dbt(["run", "--select", "incremental_not_schema_change"]).results[0].status
+        )
+
+        print(run_result)
```
ChenyuLInx, line 79: This can be removed before merge

tests/adapter/dbt/tests/adapter/basic/test_incremental.py
```diff
@@ -58,5 +60,30 @@ def test_incremental(self, project):
...
+
+        assert run_result == RunStatus.Success
+
+
 class Testincremental(BaseIncremental):
     pass
+
+
+class TestBaseIncrementalNotSchemaChange(BaseIncrementalNotSchemaChange):
```
ChenyuLInx, line 88: Great!! Can we also add this test in [dbt-snowflake](https://github.com/dbt-labs/dbt-snowflake/blob/main/tests/functional/adapter/test_incremental_unique_id.py) and [dbt-redshift](https://github.com/dbt-labs/dbt-redshift/blob/main/tests/functional/adapter/test_incremental_unique_id.py)?

ChenyuLInx (approved):
@epapineau  Thanks for moving the test into an adapter test!!

This all looks good! Just want to double-check, running the test in dbt-snowflake/redshift before your fix would cause the test to fail right?

tests/adapter/dbt/tests/adapter/basic/test_incremental.py
epapineau: Oops! Removed.

epapineau:
@leahwicz Sorry for the delay in responding, I'm just getting caught back up after two weeks of PTO. Let me confirm the test fails as @ChenyuLInx asked and then this is ready to merge

ChenyuLInx:
Paired with @epapineau today and adjusted the test to make sure it will fail on snowflake before applying the fix

PR #5421, opened 2022-06-29
Draft for python model

ChenyuLInx:
This is the initial change for #5261

This PR would enable ref, source, and config for the python model
We added checks for python model that enforce definition of a model, args to pass in to the model, restriction on only return one dataframe.

Also super happy to hear any feedback and what you think we missed.

TODOs before this merge to main:
- [x] #5510
- [x] #5511
- [x] #5512
- [x] We have extra [traceback.print_exc()](https://github.com/dbt-labs/dbt-core/blob/c521fa6b747827c26c1a5b5a7a2b0ea5bc284cc1/core/dbt/main.py#L145) added recently, do we want to make a flag for it? Gerda fixed it in #5560
- [x] add check for materialization: this can be done after beta release, doing it in #5569


core/dbt/include/global_project/macros/etc/statement.sql
```diff
@@ -1,13 +1,27 @@
...
-      {{ log('Writing runtime SQL for node "{}"'.format(model['unique_id'])) }}
-      {{ write(sql) }}
+      {{ log('Writing runtime {} for node "{}"'.format(language, model['unique_id'])) }}
+      {{ write(compiled_code) }}
+    {%- endif -%}
+    {%- if language == 'sql'-%}
+      {%- set res, table = adapter.execute(compiled_code, auto_begin=auto_begin, fetch=fetch_result) -%}
+    {%- elif language == 'python' -%}
+      {%- set res = adapter.submit_python_job(model, compiled_code) -%}
+      {#-- TODO: What should table be for python models? --#}
```
chamini2, line 19: Question: Doesn't a Python job also generate a table on the data warehouse?

jtcohen6: Yes! This `table` refers to metadata returned when the query completes, which dbt uses to create [its "result" object](https://docs.getdbt.com/reference/artifacts/run-results-json). Depending on how we submit and return the query, that table can include information like status code and number of rows created/updated. It's orthogonal to the actual table (relational object) produced by the query in the database.

core/dbt/include/global_project/macros/python_model/python.sql
```diff
@@ -0,0 +1,82 @@
...
+
+    @staticmethod
+    def get(key):
+        return config_dict.get(key)
+
+class this:
+    """dbt.this() or dbt.this.identifier"""
+    database = '{{ this.database }}'
+    schema = '{{ this.schema }}'
+    identifier = '{{ this.identifier }}'
```
ueshin, line 66: Can we check the `this.database`, etc. are `None` or not and set `None` if it's `None` instead of always setting strings?

ChenyuLInx: I think we can probably do something like that, this, for now, is only going to affect folks using `dbt.this.database` in python script so we will probably leave it at the end to polish

core/dbt/include/global_project/macros/etc/statement.sql
ChenyuLInx: @chamini2 I tool a swing at the 3 adapters we added python support for and realized that we don't have much info to give for now other than status. Snowflake will only return row affected as 1 for stored procedure, for databricks and dataproc on gcp there seems to be no such info right now.  We will probably circle back later on to see if things changes

core/dbt/contracts/graph/manifest.py
```diff
@@ -1157,7 +1157,7 @@ def __init__(self, macros):
 
 
 @dataclass
-@schema_version("manifest", 6)
+@schema_version("manifest", 7)
```
ChenyuLInx, line 1160: All manifest versions here need to be double check/updated before merging in

core/dbt/contracts/util.py
```diff
@@ -234,9 +255,11 @@ def read_and_check_versions(cls, path: str):
                 # cls.dbt_schema_version is a SchemaVersion object
                 if not cls.is_compatible_version(previous_schema_version):
                     raise IncompatibleSchemaException(
-                        expected=str(cls.dbt_schema_version), found=previous_schema_version
+                        expected=str(cls.dbt_schema_version),
+                        found=previous_schema_version,
                     )
-
+        if get_manifest_schema_version(data) <= 6:
```
ChenyuLInx, line 261: Logic to load existing manifest pre update, double check version before merge/release

core/dbt/include/global_project/macros/etc/statement.sql
```diff
@@ -1,13 +1,27 @@
-{% macro statement(name=None, fetch_result=False, auto_begin=True) -%}
+{#--
+TODO:  Only certian adapters need to support python models so this should probably be
+moved to an override I.E. `spark__statement`  but for some reason it wasn't picking up that change
```
gshank, line 3: The macro override naming method (spark__statement) only works for macros which are called with adapter.dispatch. For macros called directly, you can just redefine them. (macro statement)

core/dbt/main.py
```diff
@@ -141,11 +139,10 @@ def main(args=None):
         except SystemExit as e:
             exit_code = e.code
 
-        except BaseException as e:
-            traceback.print_exc()
-            fire_event(MainEncounteredError(e=str(e)))
-            fire_event(MainStackTrace(stack_trace=traceback.format_exc()))
-            exit_code = ExitCodes.UnhandledError.value
+        # except BaseException as e:
```
gshank, line 142: Why is this commented out?

core/dbt/include/global_project/macros/etc/statement.sql
gshank: Though maybe not in adapters... redefining macros in a project works...

We probably need some other way to check whether python models are supported for this adapter. Possibly a method in the adapter class in impl.py?

gshank (commented):
It's looking good!

core/dbt/compilation.py
```diff
@@ -435,7 +454,7 @@ def add_test_edges(self, linker: Linker, manifest: Manifest) -> None:
         # test1
         #
         # Produce the following graph:
-        # model1 --> model2 --> model3
+        # model1 --> model2 --> model
```
lostmygithubaccount, line 457: errant removal of model3?

core/dbt/parser/models.py
```diff
@@ -25,6 +26,109 @@
...
+from dbt.exceptions import ParsingException, validator_error_message, UndefinedMacroException
+
+
+class PythonValidationVisitor(ast.NodeVisitor):
+    def __init__(self):
+        super().__init__()
+        self.dbt_errors = []
+
+    def visit_FunctionDef(self, node: ast.FunctionDef) -> None:
+        if node.name == "model":
```
ChenyuLInx, line 41: Add check to validate there's a model function

core/dbt/main.py
ChenyuLInx: Probably by error, will fix it

core/dbt/include/global_project/macros/etc/statement.sql
ChenyuLInx: Gotcha, yes, I think we are going to go with Jeremy's suggested way of updating materialization to specify supported language, and default to `sql` to make restrict this

gshank (approved):
Looks good!

core/dbt/parser/models.py
ChenyuLInx: Added more checks in latest comment

stu-k (approved):
Comments addressed, new additions look good 👍

PR #5425, opened 2022-06-29
Decision Record (ADR) for `profiles.yml` directory search order

dbeatty10:
### Description
This PR is for an ADR. See #5411 and the ADR itself for a full description.

### Checklist

- [x] ~I have run this code in development and it appears to resolve the stated issue~
- [x] ~I have run `changie new` to [create a changelog entry](https://github.com/dbt-labs/dbt-core/blob/main/CONTRIBUTING.md#Adding-CHANGELOG-Entry)~

### DIFF
--- a/docs/arch/adr-005-profiles-directories.md
+++ b/docs/arch/adr-005-profiles-directories.md
@@ -25,9 +25,6 @@ Search order for `dbt_project.yml`:
 1. `DBT_PROJECT_DIR` environment variable (**NEW**)
 1. current working directory
 
-
-
-
 ### General design requirements
 There's two necessary pieces for dbt to use a profile to connect to a target database:
 - property definitions for the desired target database
@@ -71,10 +68,6 @@ This requires doing all of the following for local development:
 ##### Cons
 - Still need to configure the `profiles.yml` somehow (`~/.dbt/` or `DBT_PROJECT_DIR` or `--profiles-dir`)
     - There is _no way_ to clone a repo and have it "just work"
-
-
-
-
 
 ### Options
 
@@ -147,11 +140,6 @@ The most straight-forward solution to this currently is to just set the `DBT_PRO
 - They are isolated from each other
 ##### Cons
 - The end user needs to install Docker for their host operating system
-
-
-
-
-
 
 ## Decision
 Default to current working directory for `profiles.yml`. Fall back to `~/.dbt/`.
--- a/docs/arch/adr-005-profiles-directories.md
+++ b/docs/arch/adr-005-profiles-directories.md
@@ -1,0 +1,187 @@
+# Search order for `profiles.yml`
+
+## Context
+dbt has a solid foundation of [convention over configuration](https://en.wikipedia.org/wiki/Convention_over_configuration) (CoC), and this proposal would lean into this further.
+
+### Current behavior
+dbt needs a `profiles.yml` configuration file for database connection info. I believe the current search order is:
+1. `--profiles-dir` option
+1. `DBT_PROFILES_DIR` environment variable
+1. `~/.dbt/` directory
+
+dbt also needs a `dbt_project.yml`. The current order of precedence is:
+1. `--project-dir` option
+1. current working directory
+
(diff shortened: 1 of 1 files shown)


docs/arch/adr-005-profiles-directories.md
```diff
@@ -0,0 +1,175 @@
...
+1. `--project-dir` option
+1. `DBT_PROJECT_DIR` environment variable (**NEW**)
+1. current working directory
+
+### General design requirements
+There's two necessary pieces for dbt to use a profile to connect to a target database:
+- property definitions for the desired target database
+- secrets to plug into the definition slots
+
+There are two main design requirements in terms of discoverability and accessibility:
```
leahwicz, line 33: Discoverable/Accessible in what sense? Like inside your project?

docs/arch/adr-005-profiles-directories.md
```diff
@@ -0,0 +1,175 @@
...
+1. property definitions are easily discoverable (since they are not sensitive)
+1. secrets are non-discoverable and access is restricted per security policies
+
+#### Approach 1
```
leahwicz, line 37:
```suggestion
#### Approach 1 (current approach)
```

leahwicz:
@dbeatty10 thanks so much for making this! 

I get a little lost going through this so let me list what I followed and then where it got blurry for me.

1. Initially, you describe the current behavior for how we look for certain yml files and then you propose a new search order. - I followed
2. Then list approaches to how to store properties and secret - I followed the approaches but got confused as to if these were current approaches people used today or if these were approaches you were proposing us to implement. I _think_ these are current ways we can approach things today so maybe that could be clearer
3. Options section- from my perspective, you already have a solution because you listed it as the desired outcome above so it doesn't seem like the other outcomes are valid at this point. Does that make sense? Maybe move the desired outcome section down to the Decision section?

If you want, we can meet and chat about this if that's easier and this feedback isn't very clear.

dbeatty10:
@leahwicz Your feedback is great! Thank you for outlining where it was clear and where it was unclear.

In my mind, there's five main topics to cover:
1. Current search order
1. Proposed search order
1. Current options for connection properties and connection secrets (two approaches)*
1. Listing of alternatives to the proposal (to proactively answer the question, "did you think about X instead?")
1. Pros and cons / tradeoffs of each alternative that was seriously considered

In my attempt to cover all of those, I think it got a bit confusing. Thanks to your feedback, I will try to reorganize for readability.

Probably the most confusing part was that while explaining the pros and cons of the two ways that `profiles.yml` can be used, I made it sound like these were options under consideration (when in fact, that was settled long, long ago).

*I wonder if it would be worth creating a stand-alone post-hoc ADR for the two ways that `profiles.yml` can be used (plain-text secrets vs. exclusively environment variables rather than plain-text secrets)?

docs/arch/adr-005-profiles-directories.md
dbeatty10: I intended to describe that both approach 1 and 2 are possible currently, and each are probably used frequently.

1 would be more likely to occur in a local environment.
2 is probably used most often in continuous integration (CI) environments.

I will re-work this section for clarity.

What I probably should have used as a guiding light instead:
- Categorizing as **Approach A** or **Approach B** entirely depends on if _any_ secrets are in the `profiles.yml` file or not.

Whether there are any secrets depends whether it can be stored in VCS or not. Presence in VCS (or not) will then affect where the file is located within the file system.

docs/arch/adr-005-profiles-directories.md
dbeatty10: Yep, "discoverable/accessible" was intended to correlate with being "safe to store in GitHub".

gshank (approved):

jtcohen6:
@dbeatty10 Is there any outstanding work here, or is it ready to be merged in? I'd love to have it in the repo, given the questions we've got from folks since the v1.3 release, which have largely been answered by the level of care and consideration you put into writing this up.

PR #5426, opened 2022-06-30
Proposed updates to issue "wayfinder"

jtcohen6:
Screenshot 2022-06-30 at 12 18 36

## What problem(s) are we trying to solve?

- Community members opening `dbt-core` issues, when it's not the appropriate or most expedient channel for getting the help they need, or the response they're after
- High-priority regressions getting lost in the shuffle, because they appear at first like any other bug report

### Issue templates
- Add a new "Regression" issue type/template, for high-priority bugs that regress from functionality in older versions of `dbt-core`
- Fix the `checkboxes` structure, and add some more "checks." In particular, ask users whether their feature request should actually be an `Idea`-type discussion. Very open to thoughts on the language here.

### Wayfinder
- Fix the link to the "getting help" docs guide (broken pending https://github.com/dbt-labs/docs.getdbt.com/pull/1643)
- Add a link to Discussions
- Add an explicit external link for dbt Cloud support
- Remove the link to StackOverflow, since this isn't really what we're recommending. Should we link folks to create discussions in this repo / docs repo instead? I'm still not totally clear on what our recommended path is, per ["getting help"](https://docs.getdbt.com/guides/legacy/getting-help) + ["OSS expectations"](https://docs.getdbt.com/docs/contributing/oss-expectations)

### todo?

We discussed adding a dedicated issue template for "Unable to Install" cases as well, to create a dedicated stream for those separate from Bug Report. I haven't done that yet, but I can. I think it could also reasonably fall under the purview of "Regression" (= a thing that used to work and no longer does). We'll want to make clear that the priority order is `pip` >>> everything else, and that Homebrew installation issues should actually be reported in the `homebrew-dbt` repo.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug-report.yml
+++ b/.github/ISSUE_TEMPLATE/bug-report.yml
@@ -11,12 +11,16 @@ body:
     attributes:
       label: Is this a new bug in dbt-core?
       description: >
-        Issues aren't the right place for requesting help or troubleshooting code in your own dbt project.
-        Bugs in adapter plugins should be opened in that adapter's repository.
-        Bugs in dbt Cloud should be reported to [support](mailto:support@getdbt.com).
+        In other words, is this an error, flaw, failure or fault in our software?
+        
+        If this is a bug that broke existing functionality that used to work, please open a regression issue.
+        If this is a bug in an adapter plugin, please open an issue in the adapter's repository.
+        If this is a bug experienced while using dbt Cloud, please report to [support](mailto:support@getdbt.com).
+        If this is a request for help or troubleshooting code in your own dbt project, please join our [dbt Community Slack](https://www.getdbt.com/community/join-the-community/) or open a [Discussion question](https://github.com/dbt-labs/docs.getdbt.com/discussions).
+        
         Please search to see if an issue already exists for the bug you encountered.
       options:
-        - label: I believe this is a bug in dbt-core
+        - label: I believe this is a new bug in dbt-core
           required: true
         - label: I have searched the existing issues, and I could not find an existing issue for this bug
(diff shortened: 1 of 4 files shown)


.github/ISSUE_TEMPLATE/regression-report.yml
```diff
@@ -0,0 +1,91 @@
...
+  - type: textarea
+    id: logs
+    attributes:
+      label: Relevant log output
+      description: |
+        If applicable, log output to help explain your problem.
+      render: shell
+    validations:
+      required: false
+  - type: textarea
```
leahwicz, line 53: What if we ask for what version this functionality worked for them and then what version it broke in? Narrows down the window of possibility of where the bug was introduced for us

leahwicz:
@barberscott I added you on this just to get your perspective on the changes we are making to the template since we talked about it last week

.github/ISSUE_TEMPLATE/bug-report.yml
```diff
@@ -9,23 +9,33 @@ body:
...
+          required: true
+        - label: I have searched the existing issues, and I could not find an existing issue for this bug
           required: true
   - type: textarea
     attributes:
       label: Current Behavior
       description: A concise description of what you're experiencing.
     validations:
-      required: false
+      required: true
```
leahwicz, line 32: @jtcohen6 I made some more of these fields required since they are important to our debugging process

jtcohen6 (commented):
lgtm! can't approve since I started the PR :)

@dbeatty10 Would you be interested in taking a look over these changes, too?

dbeatty10 (approved):
Looks good to me 👍

PR #5451, opened 2022-07-07
Proper internal representation

ChenyuLInx:
resolves #
This change along with https://github.com/dbt-labs/hologram/pull/49 would make sure we switch out the term raw_sql in the codebase.

Expected behavior for this PR:
1. all raw_sql  and compiled_sql  are swapped out with raw_code  and compiled_code  in dbt’s code
1. older manifest.json will pass schema validation
1. for the new manifest.json, raw_sql  is still used there for the SQL model, for the python model it is going to be raw_code

TODOs:
- [x] this causes `local variable 'context' referenced before assignment` now
- [ ] handle the newly added language in jsonschema customization
- [ ] change config to be model language specific


gshank (approved):
Nice job making a somewhat disruptive change as neat as possible!

core/dbt/contracts/graph/unparsed.py
```diff
@@ -26,26 +27,52 @@ def file_id(self):
 
 
 @dataclass
-class HasSQL:
-    raw_sql: str
+class HasCode(dbtClassMixin):
```
jtcohen6, line 30: Ah, this is clever!

core/dbt/parser/manifest.py
```diff
@@ -644,6 +644,7 @@ def read_manifest_for_partial_parse(self) -> Optional[Manifest]:
             try:
                 with open(path, "rb") as fp:
                     manifest_mp = fp.read()
+                breakpoint()
```
jtcohen6, line 647: will need to be removed (I assume you're using to debug failing tests)

jtcohen6 (commented):
> 1. all raw_sql and compiled_sql are swapped out with raw_code and compiled_code in dbt’s code

✅ 

> 2. older manifest.json will pass schema validation

Nice! Glad we have a test for this now (`tests/functional/artifacts/test_previous_version_state.py`)

> 3. for the new manifest.json, raw_sql is still used there for the SQL model, for the python model it is going to be raw_code

I left my thoughts on this in the doc you shared. This will require updates in other codebases (`dbt-docs`, `dbt-rpc`, Cloud metadata API, probably Cloud IDE as well) no matter what, whether to handle a breaking change, or in order to add support for Python models.

The two options are:
1. Provide a clean break, and totally consistent behavior between SQL + Python models
2. Preserve backwards compatibility for SQL models, but require those consumers to look at both `*_sql` + `*_code` to support both SQL + Python models

I have a slight preference is for 1. Happy to continue the conversation in the doc, and keep this PR for actual implementation details.

test/unit/utils.py
```diff
@@ -194,12 +194,19 @@ def assert_from_dict(obj, dct, cls=None):
     if cls is None:
         cls = obj.__class__
     cls.validate(dct)
-    obj_from_dict = cls.from_dict(dct)
+    try:
+        obj_from_dict = cls.from_dict(dct)
+    except:
+        breakpoint()
```
ChenyuLInx, line 200: This should be removed

test/unit/utils.py
```diff
@@ -194,12 +194,19 @@ def assert_from_dict(obj, dct, cls=None):
...
         obj_from_dict.created_at = 1
         obj.created_at = 1
-    assert obj_from_dict == obj
-
+    try:
+        assert obj_from_dict == obj
+    except:
+        # print(obj_from_dict.keys())
+        # print(obj_from_dict.keys())
+        breakpoint()
```
ChenyuLInx, line 209: This should be removed

ChenyuLInx:
Closing since we did the other approach #5499

PR #5465, opened 2022-07-12
naive jinja2 v3 upgrade

### LINKED ISSUE #4748: Upgrade to Jinja3?
Picking up from https://github.com/dbt-labs/dbt-core/issues/4745#issuecomment-1044354226:

> We've hesitated to do this, given the risk of breaking changes to user/project code—but we need to weigh that against the risk of running into this type of dependency/installation issue, given that Jinja2 is no longer officially supported and won't be patched accordingly.

Let's take the time to look through the [set of (breaking) changes](https://jinja.palletsprojects.com/en/3.0.x/changes/#version-3-0-0), and predict the impact on user/project code. Based on that analysis, we should move ahead with an upgrade, or consider our alternatives. Forking/vendoring `Jinja2` is far from ideal, but it is an option on the table.

emmyoop:
resolves #4748

### Description

Just upgrade and see what happens.

### Checklist


leahwicz (approved):

jtcohen6:
Could we also now remove our upper bound on `MarkupSafe`, which we put in place earlier this year due to incompatibility with Jinja2?

PR #5499, opened 2022-07-20
Proper internal representation option 1

ChenyuLInx:
resolves #

### Checklist


core/dbt/include/global_project/macros/materializations/snapshots/snapshot.sql
```diff
@@ -26,7 +26,7 @@
 
   {% if not target_relation_exists %}
 
-      {% set build_sql = build_snapshot_table(strategy, model['compiled_sql']) %}
+      {% set build_sql = build_snapshot_table(strategy, model['compiled_code']) %}
```
ChenyuLInx, line 29: @jtcohen6 I am not sure whether this is used in other places but sort of a breaking change we got here

ChenyuLInx: we can fix adapters we maintain if there's a issue but hard to do for third party ones

jtcohen6: @ChenyuLInx Good find. This is pretty uncommon behavior, but it's true that we have [the same code](https://github.com/dbt-labs/dbt-spark/blob/244742ac0ee35db9774d9830a96d5796bf738448/dbt/include/spark/macros/materializations/snapshot.sql#L120) in `dbt-spark` right now.

Could we preserve `raw_sql` + `compiled_sql` as node attributes / aliases for their `_code` counterparts, without including them in `manifest.json` output?

Otherwise, we'll just need to call this out as a breaking change. It should be one with limited blast radius.

ChenyuLInx: Yea, we can do that since we just need to do some custom logic in the provider to do it when generating the context. And maybe put some depreciation warnings to remove it in 2.0?

jtcohen6: > And maybe put some depreciation warnings to remove it in 2.0?

Good call. Happy to make that piece a follow-on ticket, rather than expand the scope of this PR. Could be a nice exercise for someone on the team who hasn't yet gotten to touch our [`deprecations` module](https://github.com/dbt-labs/dbt-core/blob/main/core/dbt/deprecations.py)

jtcohen6:
This PR (or a follow-up PR) should also bump the manifest version to v7. This is decidedly a change to the schema.

core/dbt/include/global_project/macros/materializations/snapshots/snapshot.sql
ChenyuLInx: This has been updated and #5558 is created  to track it

core/dbt/contracts/util.py
```diff
@@ -234,9 +255,11 @@ def read_and_check_versions(cls, path: str):
                 # cls.dbt_schema_version is a SchemaVersion object
                 if not cls.is_compatible_version(previous_schema_version):
                     raise IncompatibleSchemaException(
-                        expected=str(cls.dbt_schema_version), found=previous_schema_version
+                        expected=str(cls.dbt_schema_version),
+                        found=previous_schema_version,
                     )
-
+        if get_manifest_schema_version(data) <= 6:
+            data = upgrade_manifest_json(data)
```
ChenyuLInx, line 262: How we handle the migration of older manifest, the number might need to be updated before final merge to main

ChenyuLInx:
@jtcohen6 I am merging this into `feature/python-model-v1` branch first given that all tests passed, and we will have a final review going on over there.

Issue #48, opened 2016-07-12
"incremental" deploy


drewbanin:
only update views/tables which have changed
for developmental debugging, not for production

jthandy:
Rather than detecting changes, it may be easier to just have a --models flag that you can send a list to and it will just deploy those models and their dependencies.

This would be very useful in another situation as well. With a client, I'm building a bunch of models that are views and stacking them on top of one another. Then, in a final model, I'm taking the results of the final view and storing them to disk. This dramatically speeds up analytical query performance. I would ideally like to be able to say `dbt run --model transactions` or something similar to only build this final one. I could imagine wanting to do this in production on an ongoing basis so as to avoid tearing down all of the underlying views every hour in a production environment.

drewbanin:
deployed!

PR #5507, opened 2022-07-21
Remove markupsafe pin

### LINKED ISSUE #5506: [CT-876] Could we also now remove our upper bound on `MarkupSafe`, which we put in place earlier this year due to incompatibility with Jinja2?
Remove our upper bound on `MarkupSafe`, which we put in place earlier this year due to incompatibility with Jinja2(#4745).  Also bump minimum requirement to match [Jinja2's requirements](https://github.com/pallets/jinja/blob/1c4066a4fad5aaeb2ac55809d1d38477cd23a0f6/setup.py#L6).

emmyoop:
resolves #5506 

### Description

Remove pin of MarkUpSafe after Jinja2 upgrade.

### Checklist

### DIFF
--- a/core/setup.py
+++ b/core/setup.py
@@ -49,7 +49,7 @@ setup(
     },
     install_requires=[
         "Jinja2==3.1.2",
-        "MarkupSafe>=0.23,<2.1",
+        "MarkupSafe>=2.0",
         "agate>=1.6,<1.6.4",
         "click>=7.0,<9",
         "colorama>=0.3.9,<0.4.6",
--- a/core/setup.py
+++ b/core/setup.py
@@ -49,7 +49,6 @@ setup(
     },
     install_requires=[
         "Jinja2==3.1.2",
-        "MarkupSafe>=2.0",
         "agate>=1.6,<1.6.4",
         "click>=7.0,<9",
         "colorama>=0.3.9,<0.4.6",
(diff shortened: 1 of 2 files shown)


core/setup.py
```diff
@@ -49,7 +49,7 @@
     },
     install_requires=[
         "Jinja2==3.1.2",
-        "MarkupSafe>=0.23,<2.1",
+        "MarkupSafe>=2.0",
```
jtcohen6, line 52: Any sense in just removing it entirely? I don't think we had it here until we needed to add the upper boundary earlier this year (https://github.com/dbt-labs/dbt-core/pull/4746)

emmyoop: Removing the pin entirely is the right move.  I wasn't aware of the history there.

It looks like dbt-labs/dbt-server([#48](https://github.com/dbt-labs/dbt-server/pull/48)) and dbt-labs/dbt-cloud ([#4928](https://github.com/dbt-labs/dbt-cloud/pull/4928)) both pinned it as well and can not remove their pins.

jtcohen6 (approved):

MichaelTiemannOSC:
This missed the 1.2.1 dbt-core cutoff.  Would be so nice if it could land in a near-term 1.2.2.

emmyoop:
@MichaelTiemannOSC unfortunately we will not be able to remove this pin for dbt versions previous to 1.3.0.  

In dbt-core 1.3.0 we were able to upgrade Jinja2 to version 3 which is what allowed the removal of the Markupsafe pin. You can see more information on the dependency conflicts between markupsafe and Jinja2 in the [related issue](https://github.com/dbt-labs/dbt-core/issues/4745).

Issue #5325, opened 2022-06-01
[CT-711] [Draft] Add Structured Logs to Materializations


nathaniel-may:
### Motivation
Today, run_results.json is a static artifact written at the end of a run which contains a summary of the objects that `dbt run` created and updated along with an incomplete view of the sql from each materialization. This information should be more granular and be visible during a run. Structured logging is a formal interface for programmatically consuming metadata about a dbt run, and is where this more granular information will be streamed to outside consumers.

### Context
The materialization macro and the adapter functions it calls are natural places to put these log lines. However, many adapters override materialiations directly (i.e. snowflake, bigquery, spark), and others override key adapter functions such as `adapter.execute` (i.e. bigquery) which would require these and future adapter maintainers to remember to copy our log lines exactly the same. Requiring adapter maintainers to take direct action to maintain the reliability of the structured logging interface is something we'd like to avoid, which is why this ticket puts these logs lines before and after the call to execute in statement.sql ([source](https://github.com/dbt-labs/dbt-core/blob/7f953a6d48a4db254b17fd3929d267e4b6069228/core/dbt/include/global_project/macros/etc/statement.sql#L10)). Since statement.sql is not a Python file, structured logging must be accessible to jinja and today it is not.

### Implementation
- [ ] Create a new function in the jinja context `log_event` like `log` ([source](https://github.com/dbt-labs/dbt-core/blob/75f3e8cb749c7b52d226844394ecb34c1e0070a4/core/dbt/context/base.py#L576-L593)) so that instead of taking a string msg, it takes a string `event_name` and a list of values of type `List[Any]`. This will fire an event with the same name as the string passed with the parameters in the list. (Using `Event.__subclasses__()`, and the splat operator for lists (i.e. `*[foo, bar]`) may help here). Mypy will not be able to catch any type mismatches here since the event constructors aren't passed concrete values till runtime. Since these log lines will run with every materialization, tests should catch any issues with these log lines.
[...]

PR #5510, opened 2022-07-21
A decorator for log code execution

ChenyuLInx:
added a decorator for submit_python_job function to log events


core/dbt/adapters/base/impl.py
```diff
@@ -1134,6 +1134,14 @@ def get_rows_different_sql(
 
         return sql
 
+    @available.parse_none
+    def submit_python_job(self, parsed_model: dict, compiled_code: str):
+        raise NotImplementedException("`submit_python_job` is not implemented for this adapter!")
+
+    def valid_python_model_materialization(self):
+        """The valid materialization an adapter should have for python model"""
+        return ["table"]
```
ChenyuLInx, line 1143: Maybe we shouldn't have a default here, but then this makes testing harder

core/dbt/adapters/base/impl.py
```diff
@@ -1185,3 +1194,24 @@ def catch_as_completed(
             # exc is not None, derives from Exception, and isn't ctrl+c
             exceptions.append(exc)
     return merge_tables(tables), exceptions
+
+
+def log_code_execution(code_execution_function):
```
ChenyuLInx, line 1199: I tested this decorator in dbt-spark and it works fine, any suggestions on adding a test for it? Maybe just a unittest with a lot of mocking?

core/dbt/adapters/base/impl.py
jtcohen6: Hmm. Would a check like this one prohibit users from defining their own custom materializations that support Python (not named `table`)? Given all we've been discussing about guardrails, this idea isn't so wild to me — but it does go against the flexibility we've allowed user-space code thus far.

I feel like the right pattern here might look like:
- A mapping, stored on the adapter, of `{language: [supported, materializations]}`. (For `sql`, that can just be "all" of them.)
- Each materialization declaring which languages it supports — either as information we can parse statically (tricky, perhaps ideal), _or_ a validation check at runtime — very easy, and still better than status quo, where dbt tries to wrap Python code in `create view as`

That would require a teensy bit of net-new code in the `view` materialization, for those adapters which support Python model code for `table` + `incremental`

core/dbt/task/runnable.py
```diff
@@ -79,13 +82,31 @@ def write_manifest(self):
...
+            ):
+                if (
+                    node.config.materialized
+                    not in self.adapter.valid_python_model_materialization()
+                ):
+                    raise_compiler_error(
+                        f"'{node.config.materialized}' is not a supported materialization of python model for the current adapter,\n"
+                        f"current adapter support {self.adapter.valid_python_model_materialization()}",
+                        node,
+                    )
```
jtcohen6, line 104: I like the idea of the decorator! Makes it easy for adapter maintainers to do the right thing, while we still have to count on them to do it. (Fuller conversation about this in https://github.com/dbt-labs/dbt-core/issues/5325#issuecomment-1191804571)

It feels like it would require a _lot_ of mocking to test, which intimidates me, but maybe doesn't intimidate you. My personal approach would be a functional test in `dbt-spark` + `dbt-bigquery` to validate that the Python code shows up in debug-level logs (e.g. using [`run_dbt_and_capture`](https://github.com/dbt-labs/dbt-core/blob/2548ba99369a38fa8c15050e00e4c5f7152d27d6/core/dbt/tests/util.py#L80))

core/dbt/adapters/base/impl.py
ChenyuLInx: I see, my original idea was that this is a function to overwrite in adapter's `impl.py`.(I did that for dbt-snowflake and dbt-spark). And since we are checking it at parsing time we won't need to actually run any model before an error is raised. 

But this does prohibit users who can't modify the python code in adapter from adding new materialization.

Happy to move this to a simple runtime check.

core/dbt/task/runnable.py
gshank: I vote for checking the logs too. Not a perfect option, but lots of mocking means that what you're testing isn't very real anyway.

ChenyuLInx: Sounds good! I will get the check for materialization into a separate PR, do some test in the adapters(as we don't really have a python job runner here, maybe we should switch out test db to duckdb sometime haha).

gshank: Can duckdb run python models?

ChenyuLInx: Not yet, we can add it and it is a embedded database, which means we are going to execute user code(given user here is just our devs I think that's fine)

ChenyuLInx:
I removed the part for validate materialization and only kept the decorator change and a minor test adjustment

PR #5512, opened 2022-07-22
add model_language to run

ChenyuLInx:
add language to model run log

### DIFF
--- a/core/dbt/task/run.py
+++ b/core/dbt/task/run.py
@@ -170,9 +170,8 @@ class ModelRunner(CompileRunner):
         return str(relation)
 
     def describe_node(self):
-        return "{} model {}".format(
-            self.node.get_materialization(), self.get_node_representation()
-        )
+        # TODO CL 'language' will be moved to node level when we change representation
+        return f"{self.node.config.get('language', 'sql')} {self.node.get_materialization()} model {self.get_node_representation()}"
 
     def print_start_line(self):
         fire_event(


core/dbt/task/run.py
```diff
@@ -170,9 +170,8 @@ def get_node_representation(self):
         return str(relation)
 
     def describe_node(self):
-        return "{} model {}".format(
-            self.node.get_materialization(), self.get_node_representation()
-        )
+        # TODO CL 'language' will be moved to node level when we change representation
```
jtcohen6, line 173: 👍

core/dbt/task/run.py
```diff
@@ -170,9 +170,8 @@ def get_node_representation(self):
...
+        return f"{self.node.config.get('language', 'sql')} {self.node.get_materialization()} model {self.get_node_representation()}"
```
jtcohen6, line 174: Should we skip if `sql` (implied), and include only if `python`? Especially since there will be many many adapters that don't support Python in the first cut.

I do like this as a way to cue people into the new capabilities, though. Multi-lingual dbt, it's here to be

jtcohen6 (commented):
I think this is cool :)

```$ dbt run -s marts.marketing.in_quarter_magic
10:37:26  Running with dbt=1.3.0-a1
10:37:42  Found ...
10:37:42
10:37:46  Concurrency: 8 threads (target='dev')
10:37:46
10:37:46  1 of 2 START sql view model dbt_jcohen.all_closed_won_accounts ................. [RUN]
10:37:49  1 of 2 OK created sql view model dbt_jcohen.all_closed_won_accounts ............ [SUCCESS 1 in 2.64s]
10:37:49  2 of 2 START python table model dbt_jcohen.in_quarter_magic_factors ............ [RUN]
10:40:18  2 of 2 OK created python table model dbt_jcohen.in_quarter_magic_factors ....... [SUCCESS 1 in 149.95s]
...
10:40:32  Finished running 1 view model, 1 table model, 2 hooks in 0 hours 2 minutes and 49.80 seconds (169.80s).
```

(cc @lostmygithubaccount @brian-gillet)

core/dbt/task/run.py
ChenyuLInx: Either way works for me! Sounds like a product call!

lostmygithubaccount: I would lean toward always indicating the language (since longer term it will probably head that way anyway?), but it's not a strongly held opinion!

jtcohen6: always it is :)

core/dbt/task/run.py
gshank: Yes. Config is usually things the user can set, and I think 'language' is not.

ChenyuLInx: Yea, I have a PR with this coming, the internal representation one, moving it to node level. Currently that PR is on hold for team dependency reason

gshank (approved):
Run out put looks good!

PR #5560, opened 2022-07-26
Fix handling of top-level exceptions

### LINKED ISSUE #4357: [CT-461] Capture exceptions at top level of threads and cli

### Describe the Feature

discovered while coding up #4326. Somewhere in [here](https://github.com/dbt-labs/dbt-core/blob/a9b4316346e3c25b82fc59fa1f090e14b57a65db/core/dbt/adapters/base/impl.py#L191-L202) when an exception is raised, the thread just hangs and you have to step through a debugger to read the exception message. Exceptions should bubble all the way to the top even in threaded environments.

you can recreate this by making `fields_to_json` throw.

### LINKED ISSUE #5564: [CT-943] Captured top-level exception do not store into log or display stacktrace.
Because of the flags associated with the MainEncounteredErroor and MainStackTrace events the stack trace was neither sent to stdout nor written to a file, and neither event was written to the log.

gshank:
resolves #5564

### Description

Errors that were caught as BaseExceptions in the main function were not displaying or storing the stack trace because it was flagged as debug level and also flagged as not to be written to the file. I have removed the "NoFile" flag from the events that handle top level exceptions and left it only for integration tests and "clean" command events.

A different pr will be used to address catching exception in threads.

### Checklist

### DIFF
--- a/core/dbt/main.py
+++ b/core/dbt/main.py
@@ -142,7 +142,6 @@ def main(args=None):
             exit_code = e.code
 
         except BaseException as e:
-            traceback.print_exc()
             fire_event(MainEncounteredError(e=str(e)))
             fire_event(MainStackTrace(stack_trace=traceback.format_exc()))
             exit_code = ExitCodes.UnhandledError.value
--- a/.changes/unreleased/Fixes-20220726-113636.yaml
+++ b/.changes/unreleased/Fixes-20220726-113636.yaml
@@ -1,0 +1,7 @@
+kind: Fixes
+body: Fix handling of top-level exceptions
+time: 2022-07-26T11:36:36.824979-04:00
+custom:
+  Author: gshank
+  Issue: "4357"
+  PR: "5560"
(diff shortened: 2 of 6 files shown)


core/dbt/main.py
```diff
@@ -142,7 +142,6 @@ def main(args=None):
             exit_code = e.code
 
         except BaseException as e:
-            traceback.print_exc()
```
jtcohen6, line 145: Given that this was an incidental inclusion in v1.2.0, I'd also be in favor of a PR that removed this line alone, and backporting that PR to `1.2.latest` for inclusion in a v1.2.1 patch release

gshank: I'm actually looking at two different problems, really, the lack of a stack trace for top level exceptions and catching exceptions in threads. I could make the current code the "lack of a stack trace" fix and backport it?

jtcohen6: I like that idea!

nathaniel-may: good plan.

ChenyuLInx (approved):
Looks good to me!
No blocker but do we want to make this available under a flag so engineers/users can access more trace when they need to?

PR #5573, opened 2022-07-28
[Backport 1.1.latest] Add changelog and whitespace fix to version bump Action

leahwicz:
Backport 2547e4f55ebc9c595506fbc99d013c2763abe2e0 from #5563.

### DIFF
--- a/.github/workflows/version-bump.yml
+++ b/.github/workflows/version-bump.yml
@@ -1,18 +1,15 @@
 # **what?**
-# This workflow will take a version number and a dry run flag. With that
+# This workflow will take the new version number to bump to. With that
 # it will run versionbump to update the version number everywhere in the
-# code base and then generate an update Docker requirements file. If this
-# is a dry run, a draft PR will open with the changes. If this isn't a dry
-# run, the changes will be committed to the branch this is run on.
+# code base and then run changie to create the corresponding changelog.
+# A PR will be created with the changes that can be reviewed before committing.
 
 # **why?**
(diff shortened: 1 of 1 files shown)


emmyoop (approved):
Just like main!

.github/workflows/version-bump.yml
```diff
@@ -1,54 +1,37 @@
...
-       required: true
-       default: 'true'
-  repository_dispatch:
-    types: [version-bump]
 
 jobs:
   bump:
     runs-on: ubuntu-latest
     steps:
+      - name: "[DEBUG] Print Variables"
```
ChenyuLInx, line 27: Is this still needed?

ChenyuLInx (approved):
Looks good with one step I think we might remove?
Or we can leave it since then it just surface out more info instead of need to rerun

.github/workflows/version-bump.yml
leahwicz: Yes, I see how this could look like a remnant of testing so maybe we rename this in `main` but it just allows us to easily see what the input is since GHA doesn't show you that automatically

emmyoop: @ChenyuLInx this is in the GHA standards [README](https://github.com/dbt-labs/dbt-core/tree/main/.github).  I am open to renaming this step, but the step itself is valuable.

ChenyuLInx: Makes total sense!! I think I left that comment with the first impression of this is something being used to debug during developing

PR #5589, opened 2022-08-01
Store when default env vars are used in manifest

### LINKED ISSUE #5155: [CT-559] [Bug] Partial parser does not recognize default values in envvars

### Current Behavior

The partial parser does not support default values for envvars. As a result, it will consider all sources deleted since the current value is `None` even if the default value and previous value are the same and unchanged.

https://github.com/dbt-labs/dbt-core/blob/220d8b888c44d43fe54539684f397a05dacf0f02/core/dbt/parser/partial.py#L959

### Expected Behavior

Since nothing has been changed in the files, I would expect the partial parser to detect the default value matches the previous value and not reparse all the files.

### Steps To Reproduce

Include an envvar like this in a source YAML file:

```
schema: '{{ env_var("OUR_SCHEMA_ENV_VAR", "default_value") }}'
```

### Relevant log output

_No response_

### Environment

```markdown
- OS: Ubuntu 20.04
- Python: 3.9.12
- dbt: 1.0.4
```

_No response_

### Additional Context

_No response_

emmyoop:
resolves #5155 

### Checklist


.changes/unreleased/Under the Hood-20220802-112936.yaml
```diff
@@ -0,0 +1,8 @@
+kind: Under the Hood
+body: Save use of defauly env vars to manifest to enable partial parsing in those
```
gshank, line 2: Typo: "defauly"

core/dbt/context/providers.py
```diff
@@ -1535,7 +1542,14 @@ def env_var(self, var: str, default: Optional[str] = None) -> str:
         if return_value is not None:
             # Save the env_var value in the manifest and the var name in the source_file
             if self.model:
-                self.manifest.env_vars[var] = return_value
+                # self.manifest.env_vars[var] = return_value
```
gshank, line 1545: You didn't use the same syntax here as the other places... Are you intending to change it?

core/dbt/logger.py
```diff
@@ -32,6 +32,7 @@
 )
 
 SECRET_ENV_PREFIX = "DBT_ENV_SECRET_"
+DEFAULT_ENV_PLACEHOLDER = "DBT_DEFAULT_PLACEHOLDER"
```
gshank, line 35: When it was just the SECRET_ENV_PREFIX, it felt kind of okay to have it in the logger file. But I'm not sure it does now. Can you think of a better place to put them? Particularly since eventually we do intend to get rid of logger.py...

gshank (approved):
Looks good, though I did have a couple of minor questions...

PR #5607, opened 2022-08-03
Adding Metric Helper Functions

### LINKED ISSUE #5567: [CT-948] [Feature] Adding Metric Tree Parsing Helper Functions to dbt Core

### Describe the Feature

Right now, the `dbt_metrics` package uses a macro called `get_metric_tree` to traverse the graph in reverse order and get the following information for each metric that is used:
- `full_set`: the full list of metrics that are used in the macro call.
- `parent_set`: the list of metrics upstream of the metrics used in the macro call. NOT inclusive of expression metrics - this list is only comprised of base metrics
- `expression_set`: the list of expression metrics used in the macro call. Inclusive of ones listed or upstrea
- `base_set`: the list of metrics listed in the macro call. 
- `ordered_expression_set`: the list of expression metrics + their depth. Starting at 999 and moving donwards to limit to 1k. Ex:
  - metric a: 999
  - metric b (downstream of a): 998

### Helper Functions
This is a list of helper functions that I believe would satisfy the same needs that we currently use the metric tree for:
1. **`get_full_metric_list`**: 
    - **Function**: This function would traverse through the dag in the opposite direction of 1 or more provided metrics and return a list of all metrics upstream of them. **Inclusive of metrics listed in input**
    - **Input**: 1+ metrics via the format `metric('metric_name')`
2. **`get_base_metric_list`**: 
    - **Function**: This function would traverse through the dag in the opposite direction of 1 or more provided metrics and return a list of all base (non-expression) metrics upstream. **Inclusive of metrics listed in input**
    - **Input**: 1+ metrics via the format `metric('metric_name')`
3. **`get_expression_metric_list`**: 
    - **Function**: This function would traverse through the dag in the opposite direction of 1 or more provided metrics and return a list of all expression metrics upstream of them. **Inclusive of metrics listed in input**
    - **Input**: 1+ metrics via the format `metric('metric_name')`
4. **`get_ordered_expression_metric_list`**: 
    - **Function**: This function would traverse through the dag in the opposite direction of 1 or more provided metrics and return a list of all expression metrics upstream of them. It would also contain a numeric representation of the depth of the expression metric  **Inclusive of metrics listed in input**
    - **Input**: 1+ metrics via the format `metric('metric_name')`
    - **Output**: A list of dictionaries like: `['metric_a':1, 'metric_b':2]`

I believe we can get away from having a `get_base_metric_list` because we can have that just be the input from the macro now that we don't need to iterate recursively over a dictionary like we're doing now. Additionally we might be able to get away with just the second expression metric list .`

The alternative is the current implementation in the `dbt_metrics` package.

### Who will this benefit?

Anyone using the `dbt_metrics` package (ie anyone using the Semantic Layer) and the developer/maintainers of the package!

Yes! I'd love to try and write some of these helper functions.

### Anything else?

[This is the link to the jinja code that currently accomplishes similar functionality.](https://github.com/dbt-labs/dbt_metrics/blob/main/macros/variables/get_metric_tree.sql)

callum-mcdata:
Resolves #5567 

### Description

This PR adds helper functions to `ResolvedMetricReference` to remove DAG parsing from `dbt_metrics` that we implemented in Jinja 🤮 . It is slightly different from the functions named in the issue in that the naming convention was slightly changed to match what appeared to be the format of making helper functions sound like attributes of the node.

The functions are:
- `parent_metric_names`: returns a list of the all metrics upstream
- `reverse_dag_parsing`: function used in other helper function that reverse iterates through the dag and returns a list of dictionaries with the expression metric name and their depth. If an expression metric is built on top of two others, the list would have 3 values returned, where the metric in question has value 1 then +1 for each upstream metric.
- `full_metric_dependency`: returns a unique list of all metrics upstream by converting the result of `parent_metric_names` to a set and then back to a list.
- `base_metric_dependency`: returns a unique list of all non-expression metrics upstream
- `derived_metric_dependency`: returns a unique list of all expression metrics upstream
- `derived_metric_dependency_depth`: returns a list of dictionaries by building off of `reverse_dag_parsing` and setting the base value at 1. 

### Checklist

I have not added tests or run changie new as these are draft changes.


gshank:
This looks really good, and the names look fine to me. I assume that a test of these functions would be special metric SQL?

callum-mcdata:
Hmmm. Honestly a test of this functionality could probably be compiling a few metric and ensuring that the results returned from these functions matches the output we expect. Let me see if I can take a stab of this in a pytest

callum-mcdata:
@gshank after some help from @ChenyuLInx around parsing classes, we were able to write a test that ensures the helper functions return what we would expect to return. If you give me the green light I'll add the changelog information and ensure that it passes all tests.

gshank (commented):
Looks good! You'll have to run pre-commit on it to fix the whitespace and black formatting.

callum-mcdata:
@gshank I've run `change` and `pre-commit` and believe that it passes all tests. Let me know if there is anything else you see that should be changed!

ChenyuLInx (approved):
Looks good to me!
I have one question but not blocking: with this getting in, I am assuming we are getting rid of some logic in Macros? should we do it also in this PR?

callum-mcdata:
@ChenyuLInx once this merges I'll create a new version of dbt_metrics that builds off these helper functions. The benefit is that there isn't any functionality loss in this that would introduce downtime - I'd just be removing functionality from dbt_metrics and pulling the same logic from elsewhere. So its not time sensitive

PR #5627, opened 2022-08-08
Use sys.exit instead of exit

varun-dc:
resolves https://github.com/dbt-labs/dbt-core/issues/5621

### Description

`exit` is implicitly imported, but it's not recommended to use it in programs (see linked issue for more details).
Changes the usage to use `sys.exit` instead. 

### Checklist

- [x] I have read [the contributing guide](https://github.com/dbt-labs/dbt-core/blob/main/CONTRIBUTING.md) and understand what's expected of me
- [x] I have signed the [CLA](https://docs.getdbt.com/docs/contributor-license-agreements)
- [x] I have run this code in development and it appears to resolve the stated issue
- [x] This PR includes tests, or tests are not required/relevant for this PR
- [x] I have [opened an issue to add/update docs](https://github.com/dbt-labs/docs.getdbt.com/issues/new/choose), or docs changes are not required/relevant for this PR
- [x] I have run `changie new` to [create a changelog entry](https://github.com/dbt-labs/dbt-core/blob/main/CONTRIBUTING.md#Adding-CHANGELOG-Entry)

PR #5640, opened 2022-08-10
[Backport] #5627 to 1.2.latest

jtcohen6:
I think the automated backport action does not work with PRs from forks: https://github.com/dbt-labs/dbt-core/actions/runs/2834765275


emmyoop:
Aw, you're right.  It's running in the context of the fork so the token does not have permission to commit directly to a branch in dbt-labs.  it has read only permission:

```
GITHUB_TOKEN Permissions
  Contents: read
  Metadata: read
  PullRequests: read
```

Not a blocker but I'll create a tech debt ticket to see if we can fix this.

emmyoop (approved):
Looks good!

PR #4303, opened 2021-11-17
adjust level length for text only

nathaniel-may:
### Description

level lengths need to line up in stdout but explicitly shouldn't for json outputs.

### Checklist

- [x] I have signed the [CLA](https://docs.getdbt.com/docs/contributor-license-agreements)
- [ ] I have run this code in development and it appears to resolve the stated issue
- [ ] This PR includes tests, or tests are not required/relevant for this PR
- [ ] I have updated the `CHANGELOG.md` and added information about my change

PR #5643, opened 2022-08-10
First test proto files

### LINKED ISSUE #5610: [CT-986] [SPIKE] Investigate options for creating proto files for dbt-core events
When working with Python and Protobuf, the normal relationship is that .proto files are defined, converted to Python classes by the proto compiler, and then integrated into Python code by importing the compiled proto classes. For simple logging events this will probably work fine. 

But in some cases we are going to want to provide a lot more data, up to and including an almostly complete serialization of node contents. Defining nodes *first* in proto files would be really difficult, particularly considering the somewhat complex class hierarchy for various node and node config classes. We will need to decide how to handle these more complicated classes, whether by some kind of field metadata and generating proto classes, or parallel but tested creation of both proto files and our existing Python classes, or creating utility functions that generate a proto based object from a node object.

In addition, we are going to need to have the proto compiler available at development time, but it should never be required at run time or for the installation of dbt core as an application. So we need to specify it as a dev requirement.

Part of this spike is also to come up with policies for handling the event codes, the proto schemas, the naming of the relevant fields, possible simplifying of the existing schemas (i.e. do we need an event "type"...). In addition we will have to come up with a standard way of importing the proto based event classes and instantiating them in our Python code. For example, proto definitions cannot have defaults, so the actual code can't be specified in a proto file. Do we subclass every proto generated class? Where/how do we set the codes for the events? And other issues like that.

The goal of this spike is that at the end of it we have a path forward so that we can start converting all of the existing events.

gshank:
resolves #5610

### Description

Spike/exploration of creating proto files for our logging events.  Discussion pull request. 

### Checklist


gshank:
This example of some proto classes uses the "code" identifier as the name of the matching proto class.

I've put all of the "base" event fields inline, which is a little duplicative, but feels simpler because it's fewer changes from our current arrangement.  Because you can't inherit in protobufs it would either have to be this way or to embed an Event message in all of the event types. I'm open to arguments though.

There are lots of warnings to NOT subclass proto generated Python classes, which makes things a little difficult. That means that we can't put any default or generated fields into the classes, or any other code, including our standard serializing code. After thinking about this and going around and around, I'm inclined to have matching Python event classes and proto messages. This isn't ideal, and I'm certainly open to arguments, but that's what I have in this initial attempt.

I think I'm going to reorganize the existing base Event class to make all of the "standard" event fields automatically generated (if possible), mostly for ease of testing, but also because I think it would be a simplification of the code in dbt/events/functions.py.

Consider this an invitation to discuss!

gshank:
This moves the individual attributes of an event class up to the top level, instead of embedding them in the dictionary under "data".  If we left those attributes under "data" we'd have to have yet another proto class to represent the data portion, which is more complicated.

I also removed "type" because we have no use case for it yet, and "log_version" because I think the proto definitions should be handling that with field numbers.

gshank:
I opened #5652 because we might want to fix that before this goes live. The logging types with ShowException in them do not show exc_info except for new style stdout logs.

gshank:
Does it work to use the code as the message name?  The Python event class names are more human understandable. We could make an index of codes and names...

core/dbt/events/functions.py
```diff
@@ -188,7 +171,9 @@ def create_debug_text_log_line(e: T_Event) -> str:
     color_tag: str = reset_color()
     ts: str = get_ts().strftime("%H:%M:%S.%f")
     scrubbed_msg: str = scrub_secrets(e.message(), env_secrets())
-    level: str = e.level_tag() if len(e.level_tag()) == 5 else f"{e.level_tag()} "
+    # What's the point of this?
+    # level: str = e.level_tag() if len(e.level_tag()) == 5 else f"{e.level_tag()} "
```
emmyoop, line 175: This is literally to make the text output line up.  The `else` has an extra space in it.  See https://github.com/dbt-labs/dbt-core/pull/4303.  Probably should have a comment as to what it's doing.

gshank: Ah. Yeah, I did miss the extra space.

gshank:
Question: Would it be better to put the standard fields in an event_info dictionary or in the top-level of the message?
```
message A001 {
    string code = 1;
    string msg = 2;
    string level = 3;
    string invocation_id = 4;
    int32 pid = 5;
    string thread_name = 6;
    google.protobuf.Timestamp ts = 7;
    string version = 10;
}

#============================

message Event {
    string code = 1;
    string msg = 2;
    string level = 3;
    string invocation_id = 4;
    int32 pid = 5;
    string thread_name = 6;
    google.protobuf.Timestamp ts = 7;
}

message A001 {
    event Event = 1
    string version = 2;
}

```

emmyoop:
> Question: Would it be better to put the standard fields in an event_info dictionary or in the top-level of the message?

I believe if we put the events in an event level dictionary it will mean the current event format will change - the  fields would be nested under `event`.  I guess we're already doing that though since we're not nesting the `data` attributes under a `data` message.

```
message Event {
    string code = 1;
    string msg = 2;
    string level = 3;
    string invocation_id = 4;
    int32 pid = 5;
    string thread_name = 6;
    google.protobuf.Timestamp ts = 7;
}

message A001 {
    event Event = 1
    message Data {
        string version = 1;
    }
    data Data = 2;
}
```

I do prefer this kind of format to reduce repetition if we're okay with the current log structure changing.

emmyoop:
> Does it work to use the code as the message name?

Why does it need to be defined with the code as the message name?  I understand we can't have defaults with proto definitions but why do we need them if we're still generating the events in the old way and the proto definitions are just a method of enforcement and a way for other systems to ingest our events?

gshank:
There has to be a way of connecting a json record with the appropriate proto definition. So if we don't do that with the code, we'd have to provide some other field with the name of the proto definition.

jtcohen6:
@gshank I'm assuming it's safe to close this one — please reopen if I'm mistaken!

PR #5651, opened 2022-08-12
Check dbt-core version requirements when installing Hub packages

### LINKED ISSUE #5648: [CT-1035] [Feature] dbt deps should take require-dbt-version into account when resolving dependencies

### Describe the feature

Because `dbt deps` doesn't take require-dbt-version into account when resolving dependencies, package vendors have to ship many more versions of packages to thread the needles of compatibility themselves.

👇 This is a lazy example because I couldn't be bothered installing an old version of dbt Core, but the _actual_ use case is an old Core installation trying to get the newest utils version. 

```yaml
# packages.yml
packages:
  - package: dbt-labs/dbt_utils
    version: "<0.8.0"
```

Running `dbt deps` with the above code on dbt Core 1.2.0 works fine, even though utils 0.7.6 has a `require-dbt-version` of `require-dbt-version: [">=0.20.0", "<1.1.0"]`. At runtime, we get an error: 
```
  Failed to read package: Runtime Error
    This version of dbt is not supported with the 'dbt_utils' package.
      Installed version of dbt: =1.2.0
      Required version of dbt for 'dbt_utils': ['>=0.20.0', '<1.1.0']
    Check for a different version of the 'dbt_utils' package, or run dbt again with --no-version-check
```

-------
(Please assume that dbt utils 0.9.0 is already out and requires dbt Core 1.2 or above)

As the maintainer of a package like dbt-audit-helper, I'd like to be able to change my [utils dependency](https://github.com/dbt-labs/dbt-audit-helper/blob/main/packages.yml) from `[">=0.8.0", "<0.9.0"]` to `[">=0.8.0", "<1.0.0"]`. Then, I would like people whose `packages.yml` looks like this: 
```yml
# packages.yml
packages:
  - package: dbt-labs/audit_helper
    version: [">=0.5.0", "<0.6.0"]
 ```
to get the correct version of dbt utils for their Core version: utils 0.8.latest for Core 1.0 or 1.2, and 0.9.latest for 1.2 and above.

Releasing a new version of audit helper which only allows utils 0.9.x. As an end user this is a problem because it means that every package that I depend on has to update at the same time, or I am trapped on an older version for longer. 

Doing a lot more backwards compatibility work in dbt utils so that 0.9 can also work with any Core 1.x version. This would solve the most pressing issue but I imagine that similar issues will arise in the future, especially as we get to utils 1.0 and beyond which might need to start taking advantage of new functionality added in minor versions of Core, but we want to have wide open dependency ranges to minimise thrash in the ecosystem. 

### Who will this benefit?

Package installers and vendors 

image
### Anything else?

_No response_

### PR DESCRIPTION
**Pull request by jtcohen6**
Resolves #5648

Depends on https://github.com/dbt-labs/hubcap/pull/146
For testing, check out https://github.com/dbt-labs/hub.getdbt.com/pull/1770

~As mentioned in the Hubcap PR above, I'm reusing the existing (and always empty) `works_with` field from the Hub API to represent the package's `require-dbt-version`. I'm not sure it's an old field being used for something in very old versions of `dbt-core`.~ This PR leverages a new API field added in the Hubcap PR above: `require_dbt_version`.

### Description

- Replace `registry.get_available_versions` with `registry.get_compatible_versions`, which only returns package versions whose `require-dbt-version` config is compatible with the currently installed version of `dbt-core`
- If a user has disabled `VERSION_CHECK` (e.g. `dbt --no-version-check deps`), then dbt will _not_ check for version compatibility
- Update unit tests

### Checklist

- [x] ~I have [opened an issue to add/update docs](https://github.com/dbt-labs/docs.getdbt.com/issues/new/choose), or~ docs changes are not required/relevant for this PR

### DIFF
--- a/core/dbt/deps/registry.py
+++ b/core/dbt/deps/registry.py
@@ -145,6 +145,7 @@ class RegistryUnpinnedPackage(RegistryPackageMixin, UnpinnedPackage[RegistryPinn
         else:
             target = None
         if not target:
+            # raise an exception if no installable target version is found
             package_version_not_found(self.package, range_, installable, should_version_check)
         latest_compatible = installable[-1]
         return RegistryPinnedPackage(
--- a/core/dbt/deps/registry.py
+++ b/core/dbt/deps/registry.py
@@ -136,7 +136,7 @@ class RegistryUnpinnedPackage(RegistryPackageMixin, UnpinnedPackage[RegistryPinn
         installable = semver.filter_installable(
             compatible_versions, self.install_prerelease or prerelease_version_specified
         )
-        available_latest = installable[-1]
+        latest_compatible = installable[-1]
 
         # for now, pick a version and then recurse. later on,
         # we'll probably want to traverse multiple options
@@ -146,5 +146,5 @@ class RegistryUnpinnedPackage(RegistryPackageMixin, UnpinnedPackage[RegistryPinn
         if not target:
             package_version_not_found(self.package, range_, installable)
         return RegistryPinnedPackage(
-            package=self.package, version=target, version_latest=available_latest
+            package=self.package, version=target, version_latest=latest_compatible
         )
--- a/core/dbt/deps/registry.py
+++ b/core/dbt/deps/registry.py
@@ -3,6 +3,8 @@ import functools
 from typing import List
 
 from dbt import semver
+from dbt import flags
+from dbt.version import get_installed_version
 from dbt.clients import registry, system
 from dbt.contracts.project import (
     RegistryPackageMetadata,
@@ -125,10 +127,14 @@ class RegistryUnpinnedPackage(RegistryPackageMixin, UnpinnedPackage[RegistryPinn
             new_msg = "Version error for package {}: {}".format(self.name, e)
             raise DependencyException(new_msg) from e
 
-        available = registry.get_available_versions(self.package)
+        should_version_check = bool(flags.VERSION_CHECK)
+        dbt_version = get_installed_version()
+        compatible_versions = registry.get_compatible_versions(
+            self.package, dbt_version, should_version_check
+        )
         prerelease_version_specified = any(bool(version.prerelease) for version in self.versions)
         installable = semver.filter_installable(
-            available, self.install_prerelease or prerelease_version_specified
+            compatible_versions, self.install_prerelease or prerelease_version_specified
         )
         available_latest = installable[-1]
 
--- a/core/dbt/deps/registry.py
+++ b/core/dbt/deps/registry.py
@@ -136,15 +136,17 @@ class RegistryUnpinnedPackage(RegistryPackageMixin, UnpinnedPackage[RegistryPinn
         installable = semver.filter_installable(
             compatible_versions, self.install_prerelease or prerelease_version_specified
         )
-        latest_compatible = installable[-1]
-
-        # for now, pick a version and then recurse. later on,
-        # we'll probably want to traverse multiple options
-        # so we can match packages. not going to make a difference
-        # right now.
-        target = semver.resolve_to_specific_version(range_, installable)
+        if installable:
+            # for now, pick a version and then recurse. later on,
+            # we'll probably want to traverse multiple options
+            # so we can match packages. not going to make a difference
+            # right now.
+            target = semver.resolve_to_specific_version(range_, installable)
+        else:
+            target = None
         if not target:
-            package_version_not_found(self.package, range_, installable)
+            package_version_not_found(self.package, range_, installable, should_version_check)
+        latest_compatible = installable[-1]
         return RegistryPinnedPackage(
             package=self.package, version=target, version_latest=latest_compatible
         )
--- a/core/dbt/exceptions.py
+++ b/core/dbt/exceptions.py
@@ -778,9 +778,10 @@ def package_not_found(package_name):
 
 def package_version_not_found(package_name, version_range, available_versions):
     base_msg = (
-        "Could not find a matching version for package {}\n"
+        "Could not find a matching compatible version for package {}\n"
         "  Requested range: {}\n"
-        "  Available versions: {}"
+        "  Compatible versions: {}\n"
+        "  (Not shown: versions incompatible with installed version of dbt-core)"
     )
     raise_dependency_error(base_msg.format(package_name, version_range, available_versions))
 
--- a/core/dbt/exceptions.py
+++ b/core/dbt/exceptions.py
@@ -787,8 +787,8 @@ def package_version_not_found(
     addendum = (
         (
             "\n"
-            "  Not shown: package versions incompatible with installed version of dbt-core"
-            "  Update your requested range, or run 'dbt --no-version-check deps'"
+            "  Not shown: package versions incompatible with installed version of dbt-core\n"
+            "  To include them, run 'dbt --no-version-check deps'"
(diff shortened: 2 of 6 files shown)

### CODE REVIEW

File: core/dbt/exceptions.py [hunk 1]
```diff
@@ -778,9 +778,10 @@ def package_not_found(package_name):
 
 def package_version_not_found(package_name, version_range, available_versions):
     base_msg = (
-        "Could not find a matching version for package {}\n"
+        "Could not find a matching compatible version for package {}\n"
         "  Requested range: {}\n"
-        "  Available versions: {}"
+        "  Compatible versions: {}\n"
+        "  (Not shown: versions incompatible with installed version of dbt-core)"
```
joellabes, line 784: Should we recommend using the no-version-check flag here, like we do when a dbt project fails to run due to mismatching versions?

jtcohen6: good idea!

dbeatty10 (commented):
There are two main cases in regards to pre-release versions of dbt-core:
- excluded, because it's not within the range specified in `require_dbt_version`.
- included, because it is within the range specified in `require_dbt_version`.

It looks like the first case is covered in the tests, but I couldn't tell if the second case is covered or not. Could you shed light if both are covered or not?

core/dbt/deps/registry.py
```diff
@@ -125,12 +127,16 @@ def resolved(self) -> RegistryPinnedPackage:
...
+        compatible_versions = registry.get_compatible_versions(
+            self.package, dbt_version, should_version_check
+        )
         prerelease_version_specified = any(bool(version.prerelease) for version in self.versions)
         installable = semver.filter_installable(
-            available, self.install_prerelease or prerelease_version_specified
+            compatible_versions, self.install_prerelease or prerelease_version_specified
         )
-        available_latest = installable[-1]
+        latest_compatible = installable[-1]
```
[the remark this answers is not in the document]
jtcohen6: Good call! We should be sure to raise a good exception before we try to access an element. I'll rearrange the logic accordingly

jtcohen6:
Prereleases! A great callout.

We're using the exact same `dbt.semver.versions_compatible` method here, that we're already using to check `require-dbt-version` [here](https://github.com/dbt-labs/dbt-core/blob/d80de8231696b780123fe825436fc25be601d936/core/dbt/config/project.py#L193-L208). That already contains handling for how to perform version comparison with prereleases.

As you note, it's possible for a pre-release version of `dbt-core` to be included or excluded by the `require-dbt-version` config defined in a package. The current logic performs two checks:
- First, are there any package versions incompatible with the installed version of `dbt-core`? If so, remove them from the `compatible_versions` list.
- Second, is the user willing to install a prerelease version? If not, remove any prerelease versions from the `compatible_versions` list to produce the `installable` list.

If there are no compatible + desired versions remaining in `installable`, we'll raise an exception saying as much.

dbeatty10:
> As you note, it's possible for a pre-release version of `dbt-core` to be included or excluded by the `require-dbt-version` config defined in a package. The current logic performs two checks:
> 
> * First, are there any package versions incompatible with the installed version of `dbt-core`? If so, remove them from the `compatible_versions` list.
> * Second, is the user willing to install a prerelease version? If not, remove any prerelease versions from the `compatible_versions` list to produce the `installable` list.

It sounds like there are two different contexts to consider prereleases (which of course makes it harder for me to think clearly, 😅 ):
1. Prereleases of `dbt-core`
2. Prereleases of a dbt package

I'm curious if the following case is represented in the continuous integration testing or not:
- user has a prerelease version of `dbt-core`
- the package has at least one version that is compatible with that particular prerelease version of `dbt-core`

### Also
Separately (but related), do we have functional tests that cover all these relevant permutations?
- installed version of `dbt-core` is a prerelease -- (2) options: yes/no
- there exists a package prerelease compatible with the installed version of `dbt-core` -- (2) options: yes/no
- user is willing to install a prerelease version of the package -- (2) options: yes/no

This is 8 different permutations (if I'm doing the math correctly). Potentially, there are some permutations that are not relevant for some reason -- I didn't expand it out fully and consider each.

jtcohen6:
@dbeatty10 You're right to push on this! There may very well be an edge case we're missing.

There is definitely a distinction in this functionality between prerelease and final releases of dbt packages. That's already pretty well tested for.

For the purposes of this functionality, I don't actually think there's any distinction between running with a prerelease versus final release of `dbt-core`. The same method — `dbt.semver.versions_compatible(current_dbt_version, require_dbt_version_range)` — will return `True` or `False` for the current installed version of `dbt-core`, whether it's a prerelease or final release.

So for instance:
- There's a new prerelease of `dbt-utils` with `require-dbt-version: [">=1.3.0b1"]`. Let's call it version `0.9.0a1`
- You're using `dbt-core==1.3.0b1` locally
    - You have `install-prerelease: true`, or a `version: 0.9.0a1` spec that allows prereleases &rarr; you'd install that version
    - You have `install-prerelease: false` in `packages.yml` &rarr; you won't install `0.9.0a1`, instead you'll install the most recent compatible non-prerelease of `dbt-utils`
- You're using `dbt-core==1.2.1rc1` locally
    - You will never install `0.9.0a1` because it is not compatible

test/unit/test_deps.py
```diff
@@ -596,3 +643,14 @@ def test_dependency_resolution(self):
...
+    def test_dependency_resolution_allow_prerelease(self):
+        package_config = PackageConfig.from_dict({
+            'packages': [
+                {'package': 'dbt-labs-test/a', 'version': '>0.1.2', 'install_prerelease': True},
+                {'package': 'dbt-labs-test/b', 'version': '0.2.1'},
+            ],
+        })
+        resolved = resolve_packages(package_config.packages, mock.MagicMock(project_name='test'))
+        self.assertEqual(resolved[0].name, 'dbt-labs-test/a')
+        self.assertEqual(resolved[0].version, '0.1.4a1')
```
dbeatty10, line 656: 🤩

core/dbt/deps/registry.py
```diff
@@ -125,20 +127,26 @@ def resolved(self) -> RegistryPinnedPackage:
...
+            # we'll probably want to traverse multiple options
+            # so we can match packages. not going to make a difference
+            # right now.
+            target = semver.resolve_to_specific_version(range_, installable)
+        else:
+            target = None
         if not target:
-            package_version_not_found(self.package, range_, installable)
+            package_version_not_found(self.package, range_, installable, should_version_check)
+        latest_compatible = installable[-1]
```
dbeatty10, line 149: I might be misreading this, but it looks like this will throw an error if `installable` is empty or `None` when it tries to get the last item via `installable[-1]`.

Do we have a test case for when there are no installable versions available after the semver filter?

jtcohen6: Yeah! If `installable` is `None` or `[]`, then we'll set `target = None` on line 146. That would cause us to raise the `package_version_not_found` exception, rather than proceed to get the indexed element

dbeatty10, line 149: Ah, I see now! Definitely a bit tricky. Thanks for the explanation.

Maybe add a quick comment like:
```suggestion
            # raise an exception if no installable target version is found
            package_version_not_found(self.package, range_, installable, should_version_check)
        latest_compatible = installable[-1]
```

dbeatty10 (approved):
@jtcohen6 Ready to rock 🎸

PR #5659, opened 2022-08-16
Print more information on log line interop test failures

### LINKED ISSUE #5658: [CT-1042] Improve unhelpful log line checks
It was discovered in #5654 which failed the log line interop tests during development that the tests don't print out the line that failed the test which makes debugging these failures near impossible. More information about the failure should be printed during testing.

nathaniel-may:
resolves #5658

### Description

Adds more information to std out when these log tests fail. Also refactors the code to be higher quality.

### Reviewers

This has a few whitespace changes. My non-whitespace changes start at 167 of the original. You can start at `test_deserialize_serialize_is_unchanged` which is the test. This calls the helper functions that have been modified.

The biggest risk of this change is that I could have weakened the test so it would pass in more situations. I, of course, don't suspect this is the case but it should be verified by the reviewer.

### Checklist

### DIFF
--- a/.changes/unreleased/Under the Hood-20220816-122032.yaml
+++ b/.changes/unreleased/Under the Hood-20220816-122032.yaml
@@ -1,0 +1,7 @@
+kind: Under the Hood
+body: add more information to log line interop test failures
+time: 2022-08-16T12:20:32.119588+01:00
+custom:
+  Author: nathaniel-may
+  Issue: "5658"
+  PR: "5659"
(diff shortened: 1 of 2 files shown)


McKnight-42 (approved):
The Rust looks fine from what I know of rust, logic seems fair, curious about the possible weakening though.

VersusFacit (approved):

nathaniel-may:
@McKnight-42 the risk with changing tests like this is that they could be accidentally modified to fail less often. I've frequently found tests in old code bases that could _never fail_ which are obviously not good tests. I don't think this modification falls into that category, but I wanted to call it out as something for reviewers to look for since we're modifying a test.

PR #5697, opened 2022-08-22
FIX: update connections.py to support connection_timeout parameter

danielkh:
resolves #

### Description

There is an issue with PostgreSQL connection timeout with unstable network connection. After some investigation I've found that it's hardcoded to set connection timeout to 10s and the connect_timeout parameter of profile is not applied. So, this PR will fix it by replacing the hardcoded value with an optional connect_timeout field of the PostgresCredentials dataclass with default value to 10 seconds (current behavior). Summary, after this patch applied there are two options: 
- specify the connect_timeout parameter in profile and change the current behavior to the desirable;
- do not specify anything and keep the current behavior.

### Checklist

### DIFF
--- a/plugins/postgres/dbt/adapters/postgres/connections.py
+++ b/plugins/postgres/dbt/adapters/postgres/connections.py
@@ -19,6 +19,7 @@ class PostgresCredentials(Credentials):
     role: Optional[str]
     port: Port
     password: str  # on postgres the password is mandatory
+    connect_timeout: Optional[int] = 10
     search_path: Optional[str] = None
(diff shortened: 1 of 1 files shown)


jtcohen6:
@danielkh Thanks for the PR! Would you be able to open a bug report (issue) so we can track this (the description you've provided above is fine), and sign the CLA so we could accept your code contribution?

danielkh:
I've created an issue: https://github.com/dbt-labs/dbt-core/issues/5698
Regarding CLA - I've signed it - but don't know how to refresh the CLA check status.

PR #5699, opened 2022-08-22
[CT-1050] convert 020_ephemeral_test 

### LINKED ISSUE #5666: [CT-1050] 020_ephemeral_tests

McKnight-42:
resolves #5666 

### Description
conversion of integration test to functional test
### Checklist


tests/functional/ephemeral/test_ephemeral.py
```diff
@@ -0,0 +1,207 @@
...
+        sql_file = "".join(sql_file.split())
+        expected_sql = "".join(expected_sql.split())
+        assert sql_file == expected_sql
+
+
+class TestEphemeralErrorHandling:
+    def test_ephemeral_error_handling(self, project):
+        results = run_dbt(["run", "--models", "dependent_errors"], expect_pass=False)
+        assert len(results) == 1
+        assert results[0].status == "skipped"
```
McKnight-42, line 206: @ChenyuLInx any idea why this might showing incorrect status of `error`  in the `RunResult` instead of still showing a `skipped`?  Was able to backtrack somewhat to [`base.py`](https://github.com/dbt-labs/dbt-core/blob/main/core/dbt/task/base.py#L267) but this looks like its more around manifest creation. did we change the status's we accept in conversion?

ChenyuLInx: So this status is still `skipped` when running this in previous tests? 

Trying to understand what's going on here, `dependent_errors` model ref an ephemeral model that has issue? So it should be skipped vs error?

McKnight-42: hey @ChenyuLInx  thanks for taking a look. yes, it was skipped, I think I just solved this actually with a small change in how I was ref the models being used  for the class for testing. testing now in ci/cd.

McKnight-42: from what I can tell since we are getting an error from an invalid non existent method on the adapter via the `ephemeral_errors__base__base_copy_sql ` model I think we move it to skipped over raising an error as there would be no real clean error.

tests/functional/ephemeral/test_ephemeral.py
```diff
@@ -0,0 +1,216 @@
...
+    def test_ephemeral_nested(self, project):
+        results = run_dbt(["run"])
+        assert len(results) == 2
+        assert os.path.exists("./target/run/test/models/root_view.sql")
+        with open("./target/run/test/models/root_view.sql", "r") as fp:
+            sql_file = fp.read()
+
+        sql_file = re.sub(r"\d+", "", sql_file)
+        expected_sql = (
+            'create view "dbt"."test_test_ephemeral"."root_view__dbt_tmp" as ('
```
McKnight-42, line 187: is renaming the hardcoded value for `test_ephemeral_` -> `test_test_ephemeral` the best way to handle file name changes here or is there something else we might prefer too?

McKnight-42:
Do we want to move this into the `adapter` section to reuse in `dbt-snowflake`? only noticable difference in tests is that the ref of model/seed file names there are capitalized.

emmyoop (approved):
Looks good.

core/dbt/tests/util.py
```diff
@@ -215,6 +215,11 @@ class TestProcessingException(Exception):
 def run_sql_with_adapter(adapter, sql, fetch=None):
     if sql.strip() == "":
         return
+
```
McKnight-42, line 218: Should this be its own small little pr?

core/dbt/tests/util.py
```diff
@@ -215,6 +215,11 @@ class TestProcessingException(Exception):
...
+    # make data type change if warehouse type is snowflake
+    if adapter.acquire_connection().type == "snowflake":
+        sql = sql.replace("BIGSERIAL", "BIGINT AUTOINCREMENT")
+
```
McKnight-42, line 222: Should this.
1. be its own small little pr? needed to make at least a few tests pass in snowflake in the new method
2. does this seem like the best method for grabbing the type in new testing format? couldn't iterate or call on a bunch of other methods or the object itself so want to make sure I didn't miss something

gshank: We'd like to avoid having adapter-specific code in the testing utilities, so this is probably not the way to to go. Do you really need to specify the type of that column? Some of the older tests using a seed.sql file, when we could actually switch to a csv file, and which will allow us to not specify the type of the 'id' column in the test.

If that doesn't work, then we can create a Snowflake override of the models fixture in the Snowflake tests.

gshank (changes requested):
Let's come up with a different way to fix the Snowflake tests.

ChenyuLInx (changes requested):

core/dbt/tests/util.py
```diff
@@ -215,6 +215,11 @@ class TestProcessingException(Exception):
 def run_sql_with_adapter(adapter, sql, fetch=None):
     if sql.strip() == "":
         return
+
+    # make data type change if warehouse type is snowflake
+    if adapter.acquire_connection().type == "snowflake":
```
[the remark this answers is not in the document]
McKnight-42: can move back to that format, would this change the location the files would need to be kept? like pulled out of adapter level into just functional level?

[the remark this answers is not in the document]
ChenyuLInx: I think adapter can just copy paste this one file and do the adjustment, or import it and do the replace in the adapter

ChenyuLInx (approved):
LGTM

core/dbt/tests/util.py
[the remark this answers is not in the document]
McKnight-42: just swapped them to csv files, and got these tests working that way, was able to remove the adapter specific logic from utils.py. so both this and snowflake version are passing, the [concurrency](https://github.com/dbt-labs/dbt-core/pull/5753) conversion is passing on core following same method but currently throwing a random syntax error on snowflake so think it might be good if all looks good.

gshank (approved):

PR #5715, opened 2022-08-25
Test priority order of profiles directory configuration

### LINKED ISSUE #5714: [CT-1086] [Feature] Test priority order of profiles directory configuration

### Describe the feature

Currently, there isn't any functionality testing that verifies the following priority order to search for `profiles.yml`:
1. `--profiles-dir` command-line argument
1. `DBT_PROFILES_DIR` environment variable
1. HOME directory of the user (i.e. `~/.dbt/`

N/A

### Who will this benefit?

This will help verify that any future code changes don't affect the expected behavior.

It will also provide a natural location in the testing code to add any behavior changes (like #5411).

Yes, will submit a pull request

### Anything else?

_No response_

dbeatty10:
resolves #5714

### Description

Verifies the following reverse priority order to search for `profiles.yml`:
1. HOME directory of the user (i.e. `~/.dbt/`)
1. `DBT_PROFILES_DIR` environment variable
1. `--profiles-dir` command-line argument

### Checklist

- [X] ~I have [opened an issue to add/update docs](https://github.com/dbt-labs/docs.getdbt.com/issues/new/choose), or~ docs changes are not required/relevant for this PR
- [ ] ~I have run `changie new` to [create a changelog entry](https://github.com/dbt-labs/dbt-core/blob/main/CONTRIBUTING.md#Adding-CHANGELOG-Entry)~

### DIFF
--- a/core/dbt/tests/util.py
+++ b/core/dbt/tests/util.py
@@ -1,4 +1,3 @@
-import io
 import os
 import shutil
 import yaml
@@ -6,7 +5,7 @@ import json
 import warnings
 from datetime import datetime
 from typing import List
-from contextlib import contextmanager, redirect_stdout
+from contextlib import contextmanager
 
 from dbt.main import handle_and_check
 from dbt.logger import log_manager
@@ -18,7 +17,6 @@ from dbt.events.test_types import IntegrationTestDebug
 # Test utilities
 #   run_dbt
 #   run_dbt_and_capture
-#   run_dbt_and_capture_stdout
 #   get_manifest
 #   copy_file
 #   rm_file
@@ -95,16 +93,6 @@ def run_dbt_and_capture(args: List[str] = None, expect_pass=True):
     # make checks for strings in the logs fail, so remove those.
     if '{"code":' in stdout:
         stdout = stdout.replace("\\", "")
-
-    return res, stdout
-
-
-# Use this if you need to capture the standard out in a test
-def run_dbt_and_capture_stdout(args: List[str] = None, expect_pass=True):
-    stringbuf = io.StringIO()
-    with redirect_stdout(stringbuf):
-        res = run_dbt(args, expect_pass=expect_pass)
-    stdout = stringbuf.getvalue()
 
     return res, stdout
 
--- a/core/dbt/tests/util.py
+++ b/core/dbt/tests/util.py
@@ -1,3 +1,4 @@
+import io
 import os
 import shutil
 import yaml
@@ -5,7 +6,7 @@ import json
 import warnings
 from datetime import datetime
 from typing import List
-from contextlib import contextmanager
+from contextlib import contextmanager, redirect_stdout
 
 from dbt.main import handle_and_check
 from dbt.logger import log_manager
@@ -17,6 +18,7 @@ from dbt.events.test_types import IntegrationTestDebug
 # Test utilities
 #   run_dbt
 #   run_dbt_and_capture
+#   run_dbt_and_capture_stdout
 #   get_manifest
 #   copy_file
 #   rm_file
@@ -90,6 +92,16 @@ def run_dbt_and_capture(args: List[str] = None, expect_pass=True):
     # make checks for strings in the logs fail, so remove those.
     if '{"code":' in stdout:
         stdout = stdout.replace("\\", "")
+
+    return res, stdout
+
+
+# Use this if you need to capture the standard out in a test
+def run_dbt_and_capture_stdout(args: List[str] = None, expect_pass=True):
+    stringbuf = io.StringIO()
+    with redirect_stdout(stringbuf):
+        res = run_dbt(args, expect_pass=expect_pass)
+    stdout = stringbuf.getvalue()
 
     return res, stdout
 
--- a/core/dbt/tests/util.py
+++ b/core/dbt/tests/util.py
@@ -71,7 +71,10 @@ def run_dbt(args: List[str] = None, expect_pass=True):
 
     print("\n\nInvoking dbt with {}".format(args))
     res, success = handle_and_check(args)
-    assert success == expect_pass, "dbt exit state did not match expected"
+
+    if expect_pass is not None:
+        assert success == expect_pass, "dbt exit state did not match expected"
+
     return res
 
 
--- a/tests/functional/profiles/test_profile_dir.py
+++ b/tests/functional/profiles/test_profile_dir.py
@@ -1,11 +1,13 @@
+import io
 import os
 import pytest
 import yaml
-from contextlib import contextmanager
+from contextlib import contextmanager, redirect_stdout
 from pathlib import Path
+from typing import List
 
 import dbt.flags as flags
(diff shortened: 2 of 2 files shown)


ChenyuLInx:
Looks good! You will need to setup [pre-commit](https://github.com/dbt-labs/dbt-core/blob/main/CONTRIBUTING.md#pre-commit) so that you commit are passing the code quality checks.

[Here's the failed check](https://github.com/dbt-labs/dbt-core/runs/8025192058?check_suite_focus=true). The command it run was `pre-commit run --all-files --show-diff-on-failure`

dbeatty10:
@ChenyuLInx it is passing all the automated checks now.

core/dbt/tests/util.py
```diff
@@ -69,7 +71,10 @@ def run_dbt(args: List[str] = None, expect_pass=True):
 
     print("\n\nInvoking dbt with {}".format(args))
     res, success = handle_and_check(args)
-    assert success == expect_pass, "dbt exit state did not match expected"
+
+    if expect_pass is not None:
```
gshank, line 75: I don't see the reason for this change. "expect_pass" should always be True or False.

core/dbt/tests/util.py
```diff
@@ -94,6 +99,16 @@ def run_dbt_and_capture(args: List[str] = None, expect_pass=True):
     return res, stdout
 
 
+# Use this if you need to capture the standard out in a test
```
gshank, line 102: Why duplicate the 'run_dbt_and_capture' functions? What does this do different?

tests/functional/profiles/test_profile_dir.py
```diff
@@ -0,0 +1,159 @@
...
+
+
+@pytest.fixture(scope="class")
+def profiles_yml(profiles_root, dbt_profile_data):
+    write_file(yaml.safe_dump(dbt_profile_data), profiles_root, "profiles.yml")
+    return dbt_profile_data
+
+
+@pytest.fixture(scope="class")
+def profiles_home_root():
```
gshank, line 18: Have you tested what happens if there is no profiles.yml at this location? Or the ~/.dbt directory doesn't exist? Because we can't depend on there being such a file in a test environment.

core/dbt/tests/util.py
dbeatty10: `dbt debug` contains `print()` statements which weren't showing up in the output for `run_dbt_and_capture`.

In my hands-on experiments, `run_dbt_and_capture` doesn't actually capture standard out -- I think it captures content streamed to the logs.

So I created a variant that includes the output of `print()` statements.

core/dbt/tests/util.py
dbeatty10: If `expect_pass` is a binary option, then the exit code of the `dbt <subcommand> <options>` will preempt everything else and take priority over subsequent assertions. Having an option to `run_dbt` that doesn't `assert` allows the user to take control of the assertion process independent of the exit code.

Why this matters in this case:
- `dbt debug` will have a non-zero exit code in certain situations (like if a profile is not found at a particular location).

In these tests, I want to get the output of `dbt debug` _regardless_ of the exit code and test the content.

tests/functional/profiles/test_profile_dir.py
dbeatty10: I did for `profiles.yml`! I didn't check how `dbt debug` behaves if `~/.dbt` doesn't exist, but that's something I can follow-up on.

The change I made to `run_dbt` to allow `expect_pass=None` enabled the tests to run `dbt debug` and capture the output, even when there are no `profiles.yml` files.

The beautify of these tests is that they merely confirm that it is looking in the expected directories for `profiles.yml` -- it is irrelevant to these tests if the `profiles.yml` actually exist within the user-specified directory or not. Other tests should verify the correct behavior when `profiles.yml` does/doesn't exist.

core/dbt/tests/util.py
gshank: We don't need two different variations of this function. Looking at the original one, it looks like it certainly intended to capture the stdout logs, if you look at the 'capture_stdout_logs' in core/dbt/events/functions.py. 

If there's something that it's missing, we should fix it rather than create an almost duplicate. What do you mean by "print()" statements? We shouldn't have any actual print statements to capture, should we?

dbeatty10: I used `git grep "print(" ./` to find instances of `print()`.

`dbt debug` uses `print()` statements. There are a some other places too, but I didn't examine them in detail.

Do you know how to modify `capture_stdout_logs` to capture `print()` statements? Or some other alternative to avoid functions that are near duplicates? I saw how `capture_stdout_logs` captures logging, but didn't see any obvious way for it to capture standard out also.

core/dbt/tests/util.py
dbeatty10: ### Overview
`pytest` recommends thinking of tests being composed of [four steps](https://docs.pytest.org/en/latest/explanation/anatomy.html#test-anatomy):
1. Arrange
1. Act
1. Assert
1. Cleanup

Currently, the implementation of `dbt_run` combines both "Act" and "Assert". I want the _option_ for it to only "Act" and leave the "Assert" for a separate and independent step. This also seems like it would align with the anatomy of a test that `pytest` is promoting.

### Why
This pull request is a prelude to testing any implementation of https://github.com/dbt-labs/dbt-core/issues/5411.

As-is, `expect_pass=False` _might_ work for the ~9 different `dbt_run`s that end up being executed. But an implementation of #5411 will introduce another ~3 executions of `dbt_run`s (actually doing `dbt debug`) which might actually have a zero exit codes, hence those specific cases would need to be `expect_pass=True`.

It's greatly simplified if there is the option to separate "Assert" from "Act" within `dbt_run`.

Additionally, this would give more flexibility to other test implementers (while retaining the current default value of `expect_pass=True`).

Are there downsides or risks you see in my proposed change?

gshank: I don't really object to this change. It mostly works this way because that's the way it was in test/integration/base.py and it's just convenient to not have to do an assert for almost every dbt_run.

dbeatty10:
@gshank is the conversation regarding `run_dbt_and_capture_stdout` the only outstanding piece for this review? Or are there other portions as well?

In regards to `run_dbt_and_capture_stdout`, do you see how it can be combined with `run_dbt_and_capture`? I didn't see any way that was obvious to me, and that's why I created a separate function.

gshank:
That's the only issue. I'd really rather not have a separate method in dbt.tests.util, but if you want to move it into the test case, that would be fine. Unfortunately I don't have time right now to investigate how to do it in the same method. Ideally we wouldn't have any print statements at all to deal with. Moving those print statements to logging statements might be the way to go, but I don't know if there are complications there.

dbeatty10:
@gshank Per your recommendation, moved `run_dbt_and_capture_stdout` into the test case.

dbeatty10:
@gshank Per your recommendation, moved `run_dbt_and_capture_stdout` into the test case.

gshank (approved):
Looks good!

PR #5726, opened 2022-08-30
ADR for why we're using betterproto for protobuf

gshank:
### Description

ADR document on why we're using betterproto


leahwicz (approved):
LGTM

McKnight-42 (approved):
Looks good to me

PR #5746, opened 2022-08-31
[Postgresql] Postgresql index creation improvement

### LINKED ISSUE #4567: [CT-37] [Bug] <Postgresql> Postgresql index creation improvement

### Current Behavior

The current behavior according to [this code](https://github.com/dbt-labs/dbt-core/blob/e943b9fc842535e958ef4fd0b8703adc91556bc6/core/dbt/include/global_project/macros/materializations/models/table/table.sql#L44-L47) is
- Create the `intermediate_relation` 
- Swap the tables with `target_relation` using a 3rd table
- Add index on the 3rd table

Switching steps 2 and 3 would help in avoiding locking the table.

We use postgresql as our transactional reporting layer against big tables. Locking them during index creation sometimes creates problems when our reporting tool tries to query the db at the same time.

### Expected Behavior

Moving [line](https://github.com/dbt-labs/dbt-core/blob/e943b9fc842535e958ef4fd0b8703adc91556bc6/core/dbt/include/global_project/macros/materializations/models/table/table.sql#L56) above this [line](https://github.com/dbt-labs/dbt-core/blob/e943b9fc842535e958ef4fd0b8703adc91556bc6/core/dbt/include/global_project/macros/materializations/models/table/table.sql#L49) will avoid locking on the table in turn avoiding slower queries which the dbt run is happening.

### Steps To Reproduce

Postgresql index creation without concurrently puts a shared lock on the table, even if its a transaction. A easy way to test this could be
```sql
begin;
Lock {table_name} in share mode;
select pg_sleep(30);
commit;

select * from {table_name};
```

### Relevant log output

_No response_

### Environment

```markdown
- OS: debian slim bullseye
- Python: Python 3.9.2
- dbt:
installed version: 0.20.1
   latest version: 1.0.0

Plugins:
  - postgres: 0.20.1
```

postgres

### Additional Context

_No response_

Goodkat:
resolves #4567

### Description
Using Postgresql as a transactional reporting layer against big tables requires index creation. Locking the big tables during index creation sometimes creates problems when reporting tool tries to query the db at the same time. The solution is building the index on the empty table before populating it so the initial index build with the shared lock is quicker.

### Checklist

### DIFF
--- a/core/dbt/include/global_project/macros/materializations/models/table/table.sql
+++ b/core/dbt/include/global_project/macros/materializations/models/table/table.sql
@@ -38,7 +38,7 @@
       {{ adapter.rename_relation(existing_relation, backup_relation) }}
   {% endif %}
 
-  {{ adapter.rename_relation(intermediate_relation, target_relation) }}  
+  {{ adapter.rename_relation(intermediate_relation, target_relation) }}
 
   {{ run_hooks(post_hooks, inside_transaction=True) }}
 
--- a/core/dbt/include/global_project/macros/materializations/models/table/table.sql
+++ b/core/dbt/include/global_project/macros/materializations/models/table/table.sql
@@ -31,14 +31,14 @@
     {{ get_create_table_as_sql(False, intermediate_relation, sql) }}
   {%- endcall %}
 
+  {% do create_indexes(target_relation) %}
+
   -- cleanup
   {% if existing_relation is not none %}
       {{ adapter.rename_relation(existing_relation, backup_relation) }}
   {% endif %}
 
-  {{ adapter.rename_relation(intermediate_relation, target_relation) }}
-
-  {% do create_indexes(target_relation) %}
+  {{ adapter.rename_relation(intermediate_relation, target_relation) }}  
 
   {{ run_hooks(post_hooks, inside_transaction=True) }}
 
--- a/.changes/unreleased/Fixes-20220831-092545.yaml
+++ b/.changes/unreleased/Fixes-20220831-092545.yaml
@@ -1,0 +1,7 @@
+kind: Fixes
+body: '[CT-37] [Bug] <Postgresql> Postgresql index creation improvement'
+time: 2022-08-31T09:25:45.300953721+02:00
+custom:
+  Author: Goodkat
+  Issue: "4567"
+  PR: "5746"


core/dbt/include/global_project/macros/materializations/models/table/table.sql
```diff
@@ -31,14 +31,14 @@
     {{ get_create_table_as_sql(False, intermediate_relation, sql) }}
   {%- endcall %}
 
+  {% do create_indexes(target_relation) %}
```
jtcohen6, line 34: I think this will need to be
```suggestion
    {% do create_indexes(intermediate_relation) %}
```
since the table has not yet been renamed. I think that might get the failing test passing (haven't been able to check locally).

jtcohen6 (commented):
Thanks @Goodkat! Looks like there's one failing test. The change itself makes sense. We'll get this queued up for review

Goodkat:
@jtcohen6 thanks for the review. All tests should be fine now.

Goodkat:
Can anyone review the last changes?

leahwicz:
@nathaniel-may could we prioritize getting some eyes on this when possible?

jtcohen6:
@colin-rogers-dbt Sorry I missed your comment a few months ago! You raise a valid concern, but I'm not sure I follow how it applies in this case. Looking at the code change, we'd still be creating + applying the indexes after the "new" table has been created — we'd just be doing it _before_ the alter-rename-swap with the _existing_ table, and thereby minimizing the lock/downtime after the "new" table has taken its place.

Worth calling out that, if this change really is desirable, we'd want a similar one within the `incremental` materialization, in the case where (a) `--full-refresh` and (b) the model table already exists.

colin-rogers-dbt:
@jtcohen6 - my comment wasn't super clear, my concern is more that based on some cursory reading the optimal behavior depends on external factors like the postgres, index complexity and the table size itself so we probably want this to be configurable

jtcohen6:
@colin-rogers-dbt Gotcha. I'm not opposed to making it configurable, if we think the performance boost will be case-by-case, or come with other tradeoffs.

Fleid:
Sorry @Goodkat, I know this is frustrating. 

I'm looking at our indecision as a symptom of our lack of confidence in the current approach. We're not positive the changes you propose will help across the board, for everybody using `dbt-postgres` today. That's why I'm not feeling comfortable about merging those changes.

So should we get back to the drawing board here?
I like the idea of making the behavior configurable, with potentially a setting allowing the use of concurrent index rebuilding?

Something like a [model level configuration](https://docs.getdbt.com/reference/resource-configs/postgres-configs#indexes):

```SQL
{{ config(
    materialized = 'table',
    indexes=[
      {'columns': ['column_a'], 'type': 'hash'},
      {'columns': ['column_a', 'column_b'], 'unique': True},
    ],
    indexing='before'
)}}

select ...
```

The options would be:
```YAML
    indexing: before|concurrent|after
```
with:
  - `before`: the change you're proposing
  - `concurrently`: `before` but with a `CREATE INDEX CONCURRENTLY`
  - `after`: the current behavior, default option

Thoughts @jtcohen6 @colin-rogers-dbt ?

Jordanjc3:
I'd be interested in seeing this PR re-opened, the option to specify `CREATE INDEX CONCURRENTLY` would be particularly helpful. We're running PostgreSQL `13.12` with `dbt-core` `1.6.7` and are observing index creation blocking reads on the target table. 

To re-produce this you can: 

1. Create a `table` with some indexes 
```sql
{{ config(
    materialized="table",
    indexes=[
        {
                "columns": ["id"],
                "unique": true,
                "type": "btree"
        }

select * from my_test_model
```
2. Issue a `dbt run --threads 1`.
3. While that model is running, continually execute the following query to find the lock coming from the `create index` command:
```sql
SELECT
  a.pid,
  a.query,
  a.state,
  l.locktype,
  l.relation::regclass,
  l.mode,
  l.granted
FROM
  pg_locks l
JOIN
  pg_stat_activity a
ON
  l.pid = a.pid
WHERE
  a.datname = current_database()
  AND mode = 'AccessExclusiveLock'
  AND a.pid <> pg_backend_pid()
  AND query LIKE '%index%'
```
One relevant note, I needed to test this with a large enough dataset so that the `create index` blocking command was noticeable. Further, if you try to read from the table while the `dbt run` is executing, you'll eventually observe the blocking behavior once it begins to create the index, which is consistent with the behavior of `AccessExclusiveLock` ([see PostgreSQL docs](https://www.postgresql.org/docs/current/explicit-locking.html)).

PR #5654, opened 2022-08-15
Ct 866/migrate hook tests

VersusFacit:
### Description

Migrates test 14 to new framework.

### Checklist

- [X] I have read [the contributing guide](https://github.com/dbt-labs/dbt-core/blob/main/CONTRIBUTING.md) and understand what's expected of me
- [X] I have signed the [CLA](https://docs.getdbt.com/docs/contributor-license-agreements)
- [X] I have run this code in development and it appears to resolve the stated issue
- [X] This PR includes tests, or tests are not required/relevant for this PR
- [X] I have [opened an issue to add/update docs](https://github.com/dbt-labs/docs.getdbt.com/issues/new/choose), or docs changes are not required/relevant for this PR

PR #5760, opened 2022-09-05
Ct 866/migrate hook tests fixed

VersusFacit:
### Description

This was supposed to a simple, humble test conversion. Instead, I fixed a weird timezone error that intermittently arises on snapshots tests. And I finally found the original root of the weird structured logging problem. 

### So what the heck happened in this PR

There was...a lot behind these code changes.

Without the presence of certain integration tests (but not others -- this I was never able to solve), DBT_LOG_FORMAT doesn't get set to json for pre-integration-test class instantiations. It shouldn't be NULL, but flags.py doesn't get invoked for certain integration test events....unless you do a secondary check.

You can see this happen when removing 014 or 015 and I presume many many others. When I removed test014, the structured log test broke because the internal FLAG was not being set that suppresses a handful of debug connection messages on each thread.

With this PR's ancillary check in place, the test (thank goodness) passes the structured logging check--that garbage data is no longer coughed up.

PRs #5705 #5654 were the poor PRs that I subjected extensive CI testing to when coaxing out exactly what was going on and how our CI differs from our local testing environments.

### Checklist


core/dbt/events/functions.py
```diff
@@ -57,7 +57,11 @@ def setup_event_logger(log_path, level_override=None):
     EVENT_HISTORY = deque(maxlen=flags.EVENT_BUFFER_SIZE)  # type: ignore
 
     make_log_dir_if_missing(log_path)
-    this.format_json = flags.LOG_FORMAT == "json"
+    # Flags is not set in the integration test suite but is by the environment.
+    # DBT_LOG_FORMAT is used in our test suite so checking for it's presence is key.
+    this.format_json = (
+        flags.LOG_FORMAT == "json" or os.environ.get("DBT_LOG_FORMAT", "json") == "json"
```
VersusFacit, line 63: When flags.py doesn't set it (rare but it happens) we must do an ancillary check or else the logging will get messed up by non-json messages referring to a _test database -- some of these are duplicates of subsequent json messages and some appear not to be  (hard to verify)

tests/functional/hooks/fixtures.py
```diff
@@ -0,0 +1,333 @@
...
+models__hooks = """
+select 1 as id
+"""
+
+models__hooks_configured = """
+{{
+    config({
+        "pre_hook": "\
+            insert into {{this.schema}}.on_model_hook (\
+                \\"state\\",\
```
VersusFacit, line 50: yep, this much escaping is ~~the~~ an answer

tests/functional/simple_snapshot/test_hard_delete_snapshot.py
```diff
@@ -24,6 +36,12 @@ def datetime_snapshot():
     return begin_snapshot_datetime
 
 
+@pytest.fixture(scope="class", autouse=True)
+def setUp(project):
```
VersusFacit, line 40: was messing around in here and left this change as it better reflects our current best practices

tests/functional/simple_snapshot/test_hard_delete_snapshot.py
```diff
@@ -79,7 +94,7 @@ def test_snapshot_hard_delete(project):
     for result in snapshotted[10:]:
         # result is a tuple, the dbt_valid_to column is the latest
         assert isinstance(result[-1], datetime)
-        assert result[-1].replace(tzinfo=pytz.UTC) >= invalidated_snapshot_datetime
+        assert result[-1].replace(tzinfo=currenttz()) >= invalidated_snapshot_datetime
```
VersusFacit, line 97: this fixes an intermittent error that will show up in mine (and others, judging from slack threads) local test invocations. I ran into, and fixed, this as part of getting my local CI to precisely match our remote CI

.changes/unreleased/Under the Hood-20220905-001749.yaml
```diff
@@ -0,0 +1,7 @@
+kind: Under the Hood
+body: Migrate integration test 014 but also make a few testing suite fixes
```
nathaniel-may, line 2: let's expand on "fixes" here to be more specific in case we need to find this commit again.

core/dbt/events/functions.py
```diff
@@ -57,7 +57,11 @@ def setup_event_logger(log_path, level_override=None):
     EVENT_HISTORY = deque(maxlen=flags.EVENT_BUFFER_SIZE)  # type: ignore
 
     make_log_dir_if_missing(log_path)
-    this.format_json = flags.LOG_FORMAT == "json"
+    # Flags is not set in the integration test suite but is by the environment.
+    # DBT_LOG_FORMAT is used in our test suite so checking for it's presence is key.
```
nathaniel-may, line 61: I'm a little confused by this comment.

tests/functional/hooks/fixtures.py
nathaniel-may: hurts my soul but it's jinja so whatever.

tests/functional/simple_snapshot/test_hard_delete_snapshot.py
```diff
@@ -14,6 +18,14 @@
 # These tests uses the same seed data, containing 20 records of which we hard delete the last 10.
 # These deleted records set the dbt_valid_to to time the snapshot was ran.
 
+# Using replace on a timestamp won't account for hour differences unless given the local timezone.
+# We can force python as utc but not postgres fields which need to be handled as local timestamps.
+def currenttz():
+    if time.daylight:
+        return timezone(timedelta(seconds=-time.altzone), time.tzname[1])
+    else:
+        return timezone(timedelta(seconds=-time.timezone), time.tzname[0])
```
nathaniel-may, line 27: I suspect there might be a better way to do this, but I'm find merging this if it's not obvious.

nathaniel-may (approved):
Great job finding these two VERY sneaky issues. I'm so happy we can finally put this to rest.

core/dbt/events/functions.py
gshank, line 63: I don't think that this is correct. If the log_format isn't set by env_var, user_config or cli option, the default is None. This is basically changing the default to be 'json'.  If the log_format isn't set and it IS set in an env_var, it means that flags.set_from_args has not been called someplace where it ought to be.

gshank (changes requested):
I'm glad you got this working, but I don't think the solution is right yet. I think we need to be calling flags.set_from_args someplace.

tests/functional/simple_snapshot/test_hard_delete_snapshot.py
VersusFacit: Honestly, it's hard to find an answer in the std lib. I went with this to avoid adding a library. Not tied to any which way though.

core/dbt/events/functions.py
VersusFacit: > This is basically changing the default to be 'json'.
Yep, that was to get this working, and I think I mentioned that in Slack.
> If the log_format isn't set and it IS set in an env_var, it means that flags.set_from_args has not been called someplace where it ought to be.
See, this is exactly where I was hoping someone more familiar would know the answer. Let me see about getting that routine called and reverting this line.

core/dbt/events/functions.py
VersusFacit: Well that's because it was a placeholder of sorts since the solution I originally had was a jackhammer for a nail. Revised!

core/dbt/events/functions.py
VersusFacit: Think I've got something better now. Best part is once the base.py is removed, this patch will go with it, i.e. the fix is scoped to integration tests

gshank (approved):
In theory, there might be a better place to put the flags.set_from_args call (including the user_config and args params), but integration/base.py is pretty convoluted and as you say is going to go away. So I'm fine with this as a way to move forward :)

PR #5780, opened 2022-09-07
Ct 1049/migrate 016 macro tests

VersusFacit:
### Description

Test conversion hot off the presses.

### Checklist


tests/functional/macros/test_macros.py
```diff
@@ -0,0 +1,337 @@
...
+    def macros(self):
+        return {"macros.sql": macros__override_get_columns_macros}
+
+    def test_overrides(self, project):
+        # the first time, the model doesn't exist
+        run_dbt()
+        run_dbt()
+
+
+class TestMacroOverridePackage(object):
```
VersusFacit, line 239: This test involves an adapter repo but I'm sure it's an "adapter" test that should be factored out per se. Would this be best as a postgres test? Aren't those technically just core tests? I'm not seeing the reason for doing something special here.

tests/functional/macros/test_macros.py
```diff
@@ -0,0 +1,337 @@
...
+"""
+
+macros__deprecated_adapter_macro = """
+{% macro some_macro(arg1, arg2) -%}
+    {{ adapter_macro('some_macro', arg1, arg2) }}
+{%- endmacro %}
+"""
+
+
+class TestMacros(object):
```
emmyoop, line 146:
```suggestion
class TestMacros:
```

In python 3 the `object` parameter is not needed.  Here's a [random SO](https://stackoverflow.com/questions/7375595/class-with-object-as-a-parameter) post about it.

emmyoop (changes requested):
Had a few questions and some changes.

tests/functional/macros/test_macros.py
```diff
@@ -0,0 +1,337 @@
...
+
+
+class TestAdapterMacroNoDestination(object):
+    @pytest.fixture(scope="class")
+    def models(self):
+        return {"model.sql": models__with_undefined_macro}
+
+    @pytest.fixture(scope="class")
+    def macros(self):
+        return {"my_macros.sql": macros__no_default_macros}
```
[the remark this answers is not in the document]
VersusFacit: It is! Remove it and the test fails because of an embedded reference. Put it back, it passes. Verified before writing this here.

tests/functional/macros/test_macros.py
```diff
@@ -0,0 +1,337 @@
...
+        return {"macros.sql": macros__package_override_get_columns_macros}
+
+    @pytest.fixture(scope="class")
+    def project_config_update(self):
+        return {
+            "dispatch": [{"macro_namespace": "dbt", "search_order": ["dbt"]}],
+        }
+
+    def test_overrides(self, project):
+        # the first time, the model doesn't exist
```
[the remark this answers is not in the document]
VersusFacit: Pretty sure this is due to:

> {% if execute and result != 'a string' %}
>   {% do exceptions.raise_compiler_error('overriding get_columns_in_relation failed') %}
> {% endif %}

The model this test calls will build and then because it now exists, this exception will throw. My understanding is we want _that_ error to be caught. `execute` makes this sketchy since that's a complicated flag.

tests/functional/macros/test_macros.py
```diff
@@ -0,0 +1,337 @@
...
+{% macro postgres__get_columns_in_relation(relation) %}
+    {{ return('a string') }}
+{% endmacro %}
+"""
+
+macros__deprecated_adapter_macro = """
+{% macro some_macro(arg1, arg2) -%}
+    {{ adapter_macro('some_macro', arg1, arg2) }}
+{%- endmacro %}
+"""
```
emmyoop, line 143: A new standard was defined that when there are more than 50 lines of these string definitions, they need to live in `fixtures.py` and then get imported.

emmyoop (changes requested):
One more small change due to a change in standards since this was first written.

emmyoop (approved):
LGTM!

PR #5782, opened 2022-09-07
Add invocation_args_dict to ProviderContext

### LINKED ISSUE #5524: [CT-898] [Feature] Make entire dbt invocation command available to Jinja context

### Describe the Feature

Right now `flags.WHICH` can be used to return the top-level command, but it would be great to surface the entire command string of args, flags, vars etc. The use case here is the dbt_artifacts package's https://github.com/brooklyn-data/dbt_artifacts/blob/main/models/fct_dbt__invocations.sql model.

Even better would be to also make each section of the command accessible separately, such as the selector, any args, flags etc.

_No response_

### Who will this benefit?

_No response_

_No response_

### Anything else?

_No response_

jared-rimmer:
resolves #5524 

### Description

- Added a `invocation_args_dict` contextproperty to the ProviderContext class.
- Added a test in `test_context.py` using a MacroContext
- Added a test in `test_builtin_functions` that prints {{invocation_args_dict}} to the logs. Again using a MacroContext

### Checklist

### DIFF
--- a/core/dbt/context/providers.py
+++ b/core/dbt/context/providers.py
@@ -63,7 +63,7 @@ from dbt.exceptions import (
 from dbt.config import IsFQNResource
 from dbt.node_types import NodeType, ModelLanguage
 
-from dbt.utils import merge, AttrDict, MultiDict
+from dbt.utils import merge, AttrDict, MultiDict, args_to_dict
 
 from dbt import selected_resources
 
@@ -709,6 +709,10 @@ class ProviderContext(ManifestContext):
             internal_packages,
             self.model,
         )
+
+    @contextproperty
+    def invocation_args_to_dict(self):
+        return args_to_dict(self.config.args)
 
     @contextproperty
     def _sql_results(self) -> Dict[str, AttrDict]:
--- a/core/dbt/context/providers.py
+++ b/core/dbt/context/providers.py
@@ -711,7 +711,7 @@ class ProviderContext(ManifestContext):
         )
 
     @contextproperty
-    def invocation_args_to_dict(self):
+    def invocation_args_dict(self):
         return args_to_dict(self.config.args)
 
     @contextproperty
--- a/test/unit/test_context.py
+++ b/test/unit/test_context.py
@@ -238,7 +238,7 @@ REQUIRED_MACRO_KEYS = REQUIRED_QUERY_HEADER_KEYS | {
     "sql_now",
     "adapter_macro",
     "selected_resources",
-    "invocation_args_to_dict"
+    "invocation_args_dict",
 }
 REQUIRED_MODEL_KEYS = REQUIRED_MACRO_KEYS | {"this", "compiled_code",}
 MAYBE_KEYS = frozenset({"debug"})
@@ -429,11 +429,11 @@ def test_invocation_args_to_dict_in_macro_runtime_context(
     )
 
     # Comes from dbt/flags.py as they are the only values set that aren't None at default
-    assert ctx["invocation_args_to_dict"]["event_buffer_size"] == 100000
-    assert ctx["invocation_args_to_dict"]["printer_width"] == 80
+    assert ctx["invocation_args_dict"]["event_buffer_size"] == 100000
+    assert ctx["invocation_args_dict"]["printer_width"] == 80
 
     # Comes from unit/utils.py config_from_parts_or_dicts method
-    assert ctx["invocation_args_to_dict"]["profile_dir"] == "/dev/null"
+    assert ctx["invocation_args_dict"]["profile_dir"] == "/dev/null"
 
 def test_model_parse_context(config_postgres, manifest_fx, get_adapter, get_include_paths):
     ctx = providers.generate_parser_model_context(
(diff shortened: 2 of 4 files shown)


core/dbt/context/providers.py
```diff
@@ -710,6 +710,10 @@ def _get_namespace_builder(self):
             self.model,
         )
 
+    @contextproperty
+    def invocation_args_to_dict(self):
```
jtcohen6, line 714: Nit: I slightly prefer
```suggestion
    def invocation_args_dict(self):
```

test/unit/test_context.py
```diff
@@ -417,6 +418,22 @@ def test_macro_runtime_context(config_postgres, manifest_fx, get_adapter, get_in
     )
     assert_has_keys(REQUIRED_MACRO_KEYS, MAYBE_KEYS, ctx)
 
+def test_invocation_args_to_dict_in_macro_runtime_context(
```
jtcohen6, line 421: Nice test!

jared-rimmer:
@gshank the Structured Logging Schema check failed but the integration tests pass when I run them locally.

I was looking into using --log-format-json in the test itself:

```
    def test_builtin_invocation_args_to_dict_function(self, project):
        _, log_output = run_dbt_and_capture(
            [
                "--debug",
                "--log-format=json",
                "run-operation",
                "validate_invocation",
                "--args",
                "{my_variable: test_variable}",
            ]
        )
        # Result is checked in two parts because profiles_dir is unique each test run
        parsed_logs = []
        for line in log_output.split("\n"):
            try:
                log = json.loads(line)
            except ValueError:
                continue

            parsed_logs.append(log)

```
And then running the assert on the parsed_logs list of dicts.

What are your thoughts on this?

tests/functional/context_methods/test_builtin_functions.py
```diff
@@ -55,6 +62,24 @@ def test_builtin_zip_function(self, project):
...
+                "{my_variable: test_variable}",
+            ]
+        )
+        # Result is checked in two parts because profiles_dir is unique each test run
+
+        result = "invocation_result: {'debug': True, 'write_json': True, 'use_colors': True, 'printer_width': 80, 'version_check': True, 'partial_parse': True, 'static_parser': True, 'profiles_dir': "
+        assert result in log_output
+
+        second_result = "'send_anonymous_usage_stats': False, 'event_buffer_size': 100000, 'quiet': False, 'no_print': False, 'macro': 'validate_invocation', 'args': '{my_variable: test_variable}', 'which': 'run-operation', 'rpc_method': 'run-operation', 'indirect_selection': 'eager'}"
+        assert second_result in log_output
```
jtcohen6, line 81: This fails in our "structured logging" test, which has all log output in JSON format. Two options to get this passing:
- Add `--log-format`, `json` to the command arguments in `run_dbt_and_capture` above, and then adjust the expected `result` to account for `\` escapes when logs are JSON formatted
- Instead of logging out the dict, and then inspecting the log outputs, you could write the anticipated `result` dict in Jinja and perform a comparison there. Something like:
```
{% macro validate_invocation(my_variable) %}
    {% set expected_dict = {'debug': True, 'write_json': True, ...} %}
    {% if expected_dict != invocation_args_dict %}
        {% do exceptions.raise_compiler_error("Assertion failed") %}
    {% endif %}
{% endmacro %}
```

jtcohen6 (commented):
Nice work @jared-rimmer! Just a teensy comment about naming, and one about fixing the failing test. Could you also [create a changelog entry](https://github.com/dbt-labs/dbt-core/blob/main/CONTRIBUTING.md#adding-a-changelog-entry)?

tests/functional/context_methods/test_builtin_functions.py
jared-rimmer: Hey @jtcohen6 thanks for the feedback. I mentioned this same issue but on the PR rather than the code as I'm trying to work through it.

I think parsing the JSON logs is perhaps easier but makes the test harder to understand as it needs to iterate through a list of dict logs to find the matching string.

The issue with doing it in the macro is that the profiles_dir value in the invocation dict isn't static so it can't be a direct dictionary comparison

gshank: 'run_dbt_and_capture' replaces double blackslashes with an empty string, for easier log checking

tests/functional/context_methods/test_builtin_functions.py
```diff
@@ -55,6 +63,46 @@ def test_builtin_zip_function(self, project):
...
+        )
+
+        assert result
+
+        # Result is checked in two parts because profiles_dir is unique each test run
+        expected = "invocation_result: {'debug': True, 'log_format': 'json', 'write_json': True, 'use_colors': True, 'printer_width': 80, 'version_check': True, 'partial_parse': True, 'static_parser': True, 'profiles_dir': "
+        assert expected in str(result)
+
+        expected = "'send_anonymous_usage_stats': False, 'event_buffer_size': 100000, 'quiet': False, 'no_print': False, 'macro': 'validate_invocation', 'args': '{my_variable: test_variable}', 'which': 'run-operation', 'rpc_method': 'run-operation', 'indirect_selection': 'eager'}"
+        assert expected in str(result)
```
jared-rimmer, line 104: I opted to add `--log-format=json` to the dbt method and parse the json logs then implement the asserts on this within the test rather than having it in the macro jinja.

Based on similar logic I found in [test_query_comments](https://github.com/dbt-labs/dbt-core/blob/main/test/integration/051_query_comments_tests/test_query_comments.py#L55-L72)

PR #5784, opened 2022-09-07
support for jinja-sql files

pepelovesvim:
`dbt` should load in `.jinja-sql` files so we can have independent language servers for both

resolves #

### Checklist

### DIFF
--- a/core/dbt/parser/read_files.py
+++ b/core/dbt/parser/read_files.py
@@ -148,23 +148,23 @@ def read_files(project, files, parser_files, saved_files):
     project_files = {}
 
     project_files["MacroParser"] = read_files_for_parser(
-        project, files, project.macro_paths, [".sql"], ParseFileType.Macro, saved_files
+        project, files, project.macro_paths, [".sql", ".jinja-sql"], ParseFileType.Macro, saved_files
     )
 
     project_files["ModelParser"] = read_files_for_parser(
-        project, files, project.model_paths, [".sql", ".py"], ParseFileType.Model, saved_files
+        project, files, project.model_paths, [".sql", ".jinja-sql", ".py"], ParseFileType.Model, saved_files
     )
 
     project_files["SnapshotParser"] = read_files_for_parser(
-        project, files, project.snapshot_paths, [".sql"], ParseFileType.Snapshot, saved_files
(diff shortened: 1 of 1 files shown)


jtcohen6:
Prior attempt: https://github.com/dbt-labs/dbt-core/pull/3498

And discussion about exactly which file extension to add support for: https://github.com/dbt-labs/dbt-core/issues/3484

We ran into some thorny errors in unit tests that were more difficult to debug than we had capacity at the time.

pepelovesvim:
interesting i thought it would be an easy add. do you have any prior progress on what could be causing failed unit tests?

jtcohen6:
It's been a while since that PR, but it was trickier business than we anticipated. If I recall correctly, I _think_ one of the tricks was around `.sql.jinja` containing a dot (`.`), and dbt supports model names containing dots (`.`), so there was internal confusion about how to treat `some.model.sql.jinja`.

The reason we went for that file extension:

> the Jinja maintainers, who previously did not endorse an official jinja file extension—preferring to just include Jinja in .html, .md, .xml, whatever—have since thrown their weight behind `.jinja` ([docs](https://jinja.palletsprojects.com/en/2.11.x/templates/#template-file-extension), [StackOverflow discussion](https://stackoverflow.com/questions/29590931/is-there-an-idiomatic-file-extension-for-jinja-templates)):
>> Adding a `.jinja` extension, like `user.html.jinja` may make it easier for some IDEs or editor plugins, but is not required.

It's possible that `.jinja-sql` wouldn't run into the same set of issues

pepelovesvim:
@jtcohen6 the reason I use `.jinja-sql` is because of this [vscode extension](https://marketplace.visualstudio.com/items?itemName=samuelcolvin.jinjahtml)](https://marketplace.visualstudio.com/items?itemName=samuelcolvin.jinjahtml). vscode (and I would assume most decent IDEs) has a file association feature. perhaps a better feature would be allowing users to define file extensions _within dbt_ that associates to `.sql`, thus allowing users with an IDE of their choosing on what file extension is considered a jinja file type.

jtcohen6:
Got it - yeah it looks like the default file associations for that VSCode extension are the ones recommended:

> Jinja SQL
> Jinja SQL templates: system name `jinja-sql`.
> Default file associations: `.sql.j2`, `.sql.jinja` and `.sql.jinja2`

PR #5786, opened 2022-09-07
Call build_flat_graph in merge_from_artifact

### LINKED ISSUE #5525: [CT-899] Update manifest flat_graph after deferral, to support dbt-metrics compilation
Metrics + deferral don't work perfectly together, because of how the `dbt-metrics` package is doing the "lookup" for a metric's model parent.

I can envision two ways to solve this problem:
1. Use a "real" `ref` in the `dbt-metrics` package [here](https://github.com/dbt-labs/dbt_metrics/blob/main/macros/graph_parsing/get_model_relation.sql), instead of re-constructing the model relation via `graph` lookup. I'd certainly feel better about it, if we did it this way :)
2. Rebuild the `flat_graph` after deferral. This is actually very simple, and could be a good way to go regardless. It will also enable users who are extracting metadata from the `graph` variable (e.g. `dbt_artifacts` package rewrite: https://github.com/brooklyn-data/dbt_artifacts/pull/132). There are two places we could do this: https://github.com/dbt-labs/dbt-core/commit/b26c7ca3efb277b9fefe778e92c2c488ef816d7b

Those aren't mutually exclusive. Both could be good ideas in their own right.

### Reproduction case

```sql
-- models/model_a.sql
select 1 as id, current_timestamp as ts

```
```yml
# models/metric.yml
version: 2

metrics:
  - name: some_metric
    label: Some Metric
    model: ref('model_a')

    type: count
    sql: id # superfluous here, but shown as an example

    timestamp: ts
    time_grains: [day]
```
```sql
-- models/model_b.sql

-- {{ metric('some_metric') }}

{% if execute %}
  {% set model_ref_node = graph.nodes.values() | selectattr('name', 'equalto', 'model_a') | first %}
  {% set relation = api.Relation.create(
      database = model_ref_node.database,
      schema = model_ref_node.schema,
      identifier = model_ref_node.alias
  )
  %}
{% else %}
  {% set relation = "" %}
{% endif %}

-- this one is a real ref
union all
-- this one is synthesized via 'graph' var
select * from {{ relation }}
```

1. Run into a "prod" schema: `dbt run --target prod`
2. Move the "prod" manifest (`target/manifest.json`) into a new folder, e.g. `state/`
3. Change the metric definition (e.g. `sql: id` &rarr; `sql: user_id`)
4. `dbt ls -s state:modified+ --state state/` to confirm that `some_metric` + `model_b` are both selected, and `model_a` is not selected
5. Ensure that dev schema is empty (e.g. `drop schema dbt_jcohen cascade`)
6. `dbt run -s state:modified+ --state state/ --defer`
7. Check logs / compiled SQL:
```sql
-- this one is a real ref
select * from "jerco"."analytics"."model_a"
union all
-- this one is synthesized via 'graph' var
select * from "jerco"."dbt_jcohen"."model_a"
```

### Resolution

Following proposal 2 above, after either of the additions in https://github.com/dbt-labs/dbt-core/commit/b26c7ca3efb277b9fefe778e92c2c488ef816d7b, this works:
```sql
-- this one is a real ref
select * from "jerco"."analytics"."model_a"
union all
-- this one is synthesized via 'graph' var
select * from "jerco"."analytics"."model_a"
```

### Any concerns?

https://github.com/dbt-labs/dbt-core/blob/1071a4681df91633301fdf23e34de819b66fbeb7/core/dbt/contracts/graph/manifest.py#L690-L701

This does require us to rebuild `flat_graph` after parsing is over, at the beginning of execution, since that's where deferral happens. Deferral is resolved before any individual nodes are run in threads, though, so I don't believe we'd need to worry about thread safety.

gshank:
resolves #5525

### Description

Rebuild flat_graph after manifest modified in 'merge_from_artifact' when deferring.

### Checklist

### DIFF
--- a/.changes/unreleased/Features-20220907-154639.yaml
+++ b/.changes/unreleased/Features-20220907-154639.yaml
@@ -1,0 +1,7 @@
+kind: Features
+body: Support dbt-metrics compilation by rebuilding flat_graph
+time: 2022-09-07T15:46:39.937326-04:00
+custom:
+  Author: gshank
+  Issue: "5525"
+  PR: "5786"
--- a/.changes/unreleased/Under the Hood-20220908-145644.yaml
+++ b/.changes/unreleased/Under the Hood-20220908-145644.yaml
@@ -1,6 +1,6 @@
-kind: Features
+kind: Under the Hood
 body: Support dbt-metrics compilation by rebuilding flat_graph
-time: 2022-09-07T15:46:39.937326-04:00
+time: 2022-09-08T14:56:44.173322-04:00
 custom:
   Author: gshank
   Issue: "5525"
--- a/core/dbt/contracts/graph/manifest.py
+++ b/core/dbt/contracts/graph/manifest.py
@@ -1018,6 +1018,10 @@ class Manifest(MacroMethods, DataClassMessagePackMixin, dbtClassMixin):
                 merged.add(unique_id)
                 self.nodes[unique_id] = node.replace(deferred=True)
 
+        # Rebuild the flat_graph, which powers the 'graph' context variable,
+        # now that we've deferred some nodes
+        self.build_flat_graph()
+
         # log up to 5 items
         sample = list(islice(merged, 5))
         fire_event(MergedFromState(nbr_merged=len(merged), sample=sample))
--- a/tests/functional/metrics/test_metric_deferral.py
+++ b/tests/functional/metrics/test_metric_deferral.py
@@ -1,0 +1,131 @@
+import os
+import pytest
+from dbt.tests.util import run_dbt, copy_file, write_file
+from pathlib import Path
+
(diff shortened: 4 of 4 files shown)


.changes/unreleased/Features-20220907-154639.yaml
```diff
@@ -0,0 +1,7 @@
+kind: Features
```
emmyoop, line 1: nit:  This seems like it's `Under the Hood` instead of a `Feature`.

emmyoop (approved):
One nit but otherwise seems good.

.changes/unreleased/Features-20220907-154639.yaml
gshank: Changed.

PR #5789, opened 2022-09-08
Complete CLI modeling for Click

### LINKED ISSUE #5526: [CT-900] Complete CLI modeling (Top level API)
**User story:**
As developer of dbt or a developer who would like to use dbt as a library, I would like to have an easily usable interface that models the user interactions at the CLI level (`dbt compile`, `dbt run`, etc).

**Acceptance Criteria:**
Extending the work done in the RAT, `dbt-core` has a new cli interface based on the click cli framework:
* It supports all regular user-facing use cases
* It fully and accurately describes each task and it's inputs (note, not the case today!)
This work stops at the click.command that represents the task, or to be more explicit this work is only to _model_ the existing cli not to fully implement successful task running.

Note: This should be on a feature branch for the time being.

iknox-fa:
resolves #5526

### Description
New Click CLI that models the existing cli

### Checklist


core/setup.py
```diff
@@ -43,9 +43,7 @@
     include_package_data=True,
     test_suite="test",
     entry_points={
-        "console_scripts": [
-            "dbt = dbt.main:main",
-        ],
+        "console_scripts": ["dbt = dbt.main:main", "dbt-cli = dbt.cli.main:cli"],
```
iknox-fa, line 46: I'm not sure if it's kosher to include a second  cli entry point given it's not complete yet-- @jtcohen6 do you have feelings on this?

nathaniel-may: jc- how would `dbt-cli` get called?

iknox-fa: You would type it on the command line.  Each entry to `console_scripts` gets registered as a thing-you-can-run-from-the-command-line`.
https://setuptools.pypa.io/en/latest/userguide/entry_point.html

jtcohen6: What would the alternative be, in order to test out the new CLI? Is there a way to invoke it via module (`python3 -m dbt.cli.main`) instead?

- If we're planning to merge this before v1.3.0-rc1, I'd prefer _not_ to add a new incomplete CLI entry point
- If we're waiting to merge this until after we've cut v1.3.0-rc1, I think it's fine to have it in `main` (alpha state) while we work on completing the new CLI

iknox-fa: > Is there a way to invoke it via module (python3 -m dbt.cli.main) instead?
It's not set up as a runnable module right now, but that's just a two-line edit.

Why don't I go ahead and make it runnable as a module and remove the direct entrypoint for now since I'm pretty sure we're merging this pre-1.3.0-rc1.

iknox-fa: Updated.

PR #5804, opened 2022-09-09
Fix macro namespace search packages

### LINKED ISSUE #5801: [CT-1146] [Bug] providers.py raises unhelpful error if dispatch provides package name that doesn't exist

### Current Behavior

I want to preface this by saying I am not surprised that no one has run into this bug before. It takes a special kind of person to make changes to every macro reference in their package and then forget to update the package name in the dispatch line. That being said, here is the bug.

If the adapter.dispatch provides the name of a package that doesn't exist, the error handling build into providers.py doesn't account for this pattern.

Example: The namespace provided `metrics` **exists** so it doesn't go into the first path. In the second path, it looks for the namespace in `get_macro_search_order` but that can't complete because the namespace doesn't exist in the `dbt_projects` path. The end result is returning a None object.  

```
        if namespace is None:
            search_packages = [None]
        elif isinstance(namespace, str):
            search_packages = self._adapter.config.get_macro_search_order(namespace)
            if not search_packages and namespace in self._adapter.config.dependencies:
                search_packages = [self.config.project_name, namespace]
```

This None object then can't be iterated in line 167:
```
        for package_name in search_packages:
```

This returns an unhelpful error message shown in relevant log output.

#### Potential Fix?
@dbeatty10 and I were able to solve this problem by adding another path after line 154:

```
            if not search_packages and namespace in self._adapter.config.dependencies:
                search_packages = [self.config.project_name, namespace]
            else:
                search_packages = [None]
```

This allows the search_packages variable to be iterated over with and helps raise a much more helpful error message around what is going wrong. 

### Expected Behavior

This error message

```
dbt.exceptions.CompilationException: Compilation Error in model develop_metric (models/metric_testing_models/develop_metric.sql)
  In dispatch: No macro named 'develop' found
      Searched for: 'postgres__develop', 'default__develop'
  
  > in macro develop (macros/develop.sql)
  > called by model develop_metric (models/metric_testing_models/develop_metric.sql)
```

### Steps To Reproduce

Loom explaining and showing issue: https://www.loom.com/share/347a5dc783cc45f2be5c8d7eede01622

### Relevant log output

```shell
File "/Users/callummccann/repos/personal_dotfiles/poetry_environments/dbt-testing-metrics/.venv/lib/python3.10/site-packages/dbt/context/providers.py", line 171, in dispatch
    for package_name in search_packages:
TypeError: 'NoneType' object is not iterable
```

### Environment

```markdown
- OS: MacOS
- Python: 3.10.6
- dbt: 1.2.1
```

postgres

### Additional Context

Happy to submit the fix if the team thinks that the proposed fix above matches what the team thinks it is good code!

dbeatty10:
resolves #5801

### Description

Make sure that `search_packages` is a list with _at least_ one entry. `[None]` is the desired base case.

### Tradeoffs

Refactored logic into a new method named `_get_search_packages`. The benefit is that the `dispatch` method is shorter by 13 lines and is easier to read. The downside is that the PR diff is longer, so harder to determine as a code reviewer of what changed and what didn't.

To aid the reviewer, here's a series of diffs to show what really changed:
1. https://github.com/dbt-labs/dbt-core/pull/5804/commits/1a94a150d3a2a2aca23ed61634bed32f4e0becb4 - just move the code as-is into a new method
2. https://github.com/dbt-labs/dbt-core/pull/5804/commits/0f57cbbe13ab3237fc168ef098953f5c642d7813 - rename one variable
3. https://github.com/dbt-labs/dbt-core/pull/5804/commits/a393c4419dd40d61cadc4f086ff7f5d0f54915fb - actually fix the bug

### Checklist


dbeatty10:
@gshank functional testing this would require testing the two-argument version of `dispatch` (to cover the 2nd, optional `macro_namespace` argument).

I didn't see a good place that was already testing that optional argument to utilize as a pattern.

Do you know of an example? Or do you know anyone that could implement the testing part of this PR?

gshank:
Yes, there seem to be some two-argument adapter.dispatch calls in tests/functional/schema_tests.

dbeatty10:
@gshank  Would you be willing to give this a look? This bug fix has tests cases now 🎉

As expected [you were right](https://github.com/dbt-labs/dbt-core/issues/5801#issuecomment-1242229418) when you said this: 
> I think the fix is fine. Please submit a PR. Creating the test will probably be a lot more work than the fix :)

 😂

gshank (approved):

PR #5815, opened 2022-09-12
Add config to disable metrics/exposures

### LINKED ISSUE #5422: [CT-788] Support disabling `metrics` + `exposures`
Pretty simple:
```yml
# models/specific_file.yml

metrics:
  - name: disabled_metric
    config:
      enabled: false
  
exposures:
  - name: disabled_exposure
    config:
      enabled: false
```
```yml
# dbt_project.yml

metrics:
  +enabled: false

exposures:
  +enabled: false
```

Expectations:
- Moved to `disabled` section of `manifest.json`
- Needs to work with partial parsing
- If a disabled metric is referenced via `metric()`, it should raise an error at the end of parsing (just as `ref()` or `source()` do)

Grouping these together because I believe it would require touching the same code paths!

emmyoop:
resolves #5422

### Description

Support disabling metrics and exposures via config.

### Checklist


core/dbt/parser/schemas.py
```diff
@@ -1027,14 +1084,33 @@ def __init__(self, schema_parser: SchemaParser, yaml: YamlBlock):
         self.schema_parser = schema_parser
         self.yaml = yaml
 
-    def parse_metric(self, unparsed: UnparsedMetric) -> ParsedMetric:
+    def parse_metric(self, unparsed: UnparsedMetric) -> None:
```
emmyoop, line 1087: This seems wrong but if the metric is disabled there's nothing to return.  Or should I be building a metric patch?

core/dbt/contracts/graph/manifest.py
```diff
@@ -616,7 +616,8 @@ class Manifest(MacroMethods, DataClassMessagePackMixin, dbtClassMixin):
     flat_graph: Dict[str, Any] = field(default_factory=dict)
     state_check: ManifestStateCheck = field(default_factory=ManifestStateCheck)
     source_patches: MutableMapping[SourceKey, SourcePatch] = field(default_factory=dict)
-    disabled: MutableMapping[str, List[CompileResultNode]] = field(default_factory=dict)
+    # TODO: not sure this is the right move here?
+    disabled: MutableMapping[str, List[GraphMemberNode]] = field(default_factory=dict)
```
emmyoop, line 620: This also seems off but exposures/metrics are not compiled.

gshank: Exposures aren't compiled but metrics are.  So "GraphMemberNode" isn't quite accurate.

core/dbt/parser/schemas.py
gshank: The whole 'patch' thing is intended for nodes that are built from other files (such as SQL files) but that have config or other attributes changes in a schema file. 

I think what we should do is change the way that 'parse' works for metrics and exposures. The way this is working right now the node is added to disabled in the 'parse_metric' and 'parse_exposure' methods, but not added to the manifest until the node is returned in the 'read_files' method. I think we should stop returning anything from parse for metrics and exposures and move adding the node to the manifest into 'parse_metric' and 'parse_exposure'. If you look at the various parser returns in 'read_files' we already have a number of parsers which don't do anything with the 'parse' method return. Mostly it's used for handling tests.

gshank (changes requested):
This config stuff really does get complicated. This looks pretty good to me, with the exception that I think we should change what's returned from 'parse' for metrics and exposures as discussed in that comment.

core/dbt/contracts/graph/manifest.py
gshank: Well, it's IS true that it "participates" in the graph. So its fine.

gshank (approved):
Looks good!

callum-mcdata:
Just tested this on my local and can confirm that it works as expected with metrics. Adding the same metric to two different folders resulted in a compilation error. When adding the below code, the definition from one of the folder paths was disabled and thus didn't get picked up. 

``` yml
metrics:
  dbt_metrics_integration_tests:
    testing_disabling__overridden:
      +enabled: false
    testing_disabling:
      +enabled: true
```

PR #5822, opened 2022-09-13
restrict python submission

### LINKED ISSUE #5596: [CT-972] [Feature] Block usage of `submit_python_job` outside of materialization logic

### Describe the feature

Throw a clear error during parsing(or later time in dbt run) if `adapter.submit_python_job` is being used in places other than the materialization logic.

## Relative information:
`adapter.submit_python_job` is currently being called in macro [statement](https://github.com/dbt-labs/dbt-core/blob/a7ff003d4f74763f126e77a0fc9c55e7a3311b3d/core/dbt/include/global_project/macros/etc/statement.sql#L15).

## Ideas:
- when we parse macro we can look up the macro tree and check that for all macro nodes that isn't `materialization` -> `statement`, we run regex on it to make sure `adapter.submit_python_job` is not there. This could be very costy
- somehow update how we provide context to macros and only provide `adapter.submit_python_job` in desired situation. This means we will have to remove it from the context sometime during jinjia compilation, this could be complex

## related links
- [where we parse all of the jinja macros](https://github.com/dbt-labs/dbt-core/blob/main/core/dbt/parser/macros.py#L69)
- [check for banned macro](https://github.com/dbt-labs/dbt-core/blob/0a0f5778aff7c780ec1b2ad7b2edc26aad6bffe9/core/dbt/parser/models.py#L354)

ChenyuLInx:
resolves #5596

### Description
Only allow `submit_python_job` being used in statement macro within materializition logic.
TODO
- [ ] make adapter.submit_python_job in dbt snowflake not available in jinja

### Checklist


core/dbt/clients/jinja.py
```diff
@@ -285,6 +292,13 @@ def __init__(
         self.node = node
         self.stack = stack
 
+        if self.context and "adapter" in self.context:
+            if "materialization" in self.macro.unique_id:
+                self.context["special_functions"] = self.context["adapter"].submit_python_job
```
ChenyuLInx, line 297: @gshank  This way user actually still can use this submit_python_job function  via `{{special_function(model, python_code)}}` in macro, is there a way to actually have this in context but not allow jinja to access it?

Other ways I have been trying to do is save it somewhere else individually, but haven't found a good place, the `MacroGenerater` object for `materialization macro` and `statement macro` are actually two different objects, with only context being passed from one to another(node, stack are both None for `materialization macro`)

ChenyuLInx: Also tag @jtcohen6 to see if you have any thoughts

core/dbt/clients/jinja.py
```diff
@@ -301,6 +319,15 @@ def exception_handler(self) -> Iterator[None]:
             e.stack.append(self.macro)
             raise e
 
+    def _is_statement_under_materailization(self) -> bool:
+        return bool(
+            self.macro.unique_id == "macro.dbt.statement"
```
ChenyuLInx, line 324: @jtcohen6 with this user can't actually access submit_python_job in custom `statement` macro. Any concern with that?

ChenyuLInx:
Don't have a test since we can't run python model in core now. but manually checked locally and made sure `materialization-> statement` is the only place that you can properly access python submit job function. You can also access it in `materialization` root level with some strange and clearly not good looking syntax. And not in all other places.

ChenyuLInx:
Also skipping changelog since I don't think we want to tell everyone about it

core/dbt/clients/jinja.py
```diff
@@ -301,6 +319,15 @@ def exception_handler(self) -> Iterator[None]:
             e.stack.append(self.macro)
             raise e
 
+    def _is_statement_under_materailization(self) -> bool:
```
jtcohen6, line 322:
```suggestion
```

core/dbt/clients/jinja.py
jtcohen6: @ChenyuLInx I'm aligned. I'm not familiar with any reason or legitimate use case for an end user to override the `statement` macro — this should really be treated as internal to `dbt-core`.

If we want to loosen this restriction in the future, we can. (E.g. by changing this conditional to check for any macro named `statement`, rather than the `unique_id` containing project name `dbt`.)

core/dbt/clients/jinja.py
```diff
@@ -272,6 +275,13 @@ def pop(self, name):
             raise InternalException(f"popped {got}, expected {name}")
 
 
+def raise_error_func(func_name: str) -> Callable:
+    def raise_error(*args, **kwargs):
+        raise InternalException(f"{func_name} is not intended to be called here.")
```
jtcohen6, line 280: Will this include the name of the macro / node calling the unsupported function?

jtcohen6 (commented):
Just to make sure I understand, the idea here is: `submit_python_job` can only be called from within a `statement`, within a materialization (macro containing the name `materialization`). Otherwise, the user should see an exception like:
```
submit_python_job is not intended to be called here.
```

I haven't been able to see the desired behavior when running locally. I tried checking out the branch, and then defining a model like:
```sql
-- models/my_model.sql
{% set compiled_code %}

def main(session):
    pass

{% endset %}

{% set result = adapter.submit_python_job(model, compiled_code) %}

-- actual view definition
select '{{ result }}' as result
```
This seems to actually work — it doesn't get blocked

core/dbt/clients/jinja.py
ChenyuLInx: Not really, but I can add it!

ChenyuLInx: Added for usage in models, for hooks, operation, test we can't really tell(they use `MacroGenerator`) but we don't get info about which function actually calls it unless we do some really aggressive tracking. So I just added a generic message indication it is in one of those places. Let me know if we want to be very specific.
I also included the full name `adapter.submit_python_job` in the message so shouldn't be hard to do a search

core/dbt/clients/jinja.py
ChenyuLInx: Fixed

ChenyuLInx:
@jtcohen6 fixed the issue you mentioned by also remove that function for compilation time.

core/dbt/clients/jinja.py
jtcohen6: Really solid UX improvement - thank you!

gshank (changes requested):
I'm kind of concerned with this approach. I think doing this in the MacroGenerator is almost certainly the wrong place.  That code would execute for every single macro that's built. The adapter that's linked in the contexts comes from the global adapter? Why doesn't the last created MacroGenerator determine which monkeypatched method is active? 

I think the MacroNamespaceBuilder is a more likely place to put this -- you can modify the context before it's used to create the MacroGenerator object. But I would prefer finding some way to just pass along information on the 'submit_python_job' call that would allow us to determine the validity in the python code. What is it that we need to know? The call stack?

ChenyuLInx:
@gshank  the `MacroNamespaceBuilder`  sounds promising. One question, for the materialization macro does that also go through `MacroNamespaceBuilder` somehow? I feel like we need to somehow pass in the call stack of the macros if we want to check it at that function, not sure how that's going to look like. Scheduled a meeting tomorrow to talk this live

jtcohen6:
Really appreciate you taking a look here @gshank! If we're going to do this, it ought to be a surgical change in the right spot.

The high-level goal here is to restrict `adapter.submit_python_job` to only be callable within a `statement` of a materialization. We want to prevent users from firing up Python jobs willy-nilly (in post-hooks, in tests, wherever). We _may_ want to support those as patterns in the future. So we're trying to put guardrails in place for now, with the possibility of removing them later, without adding a ton of cruft to the codebase in the process.

gshank:
So maybe that second suggestion wouldn't work. It looks like the materialization macro is the wrapping thing that's called to execute models, so it's passed the model_context used by everything else. It wouldn't be very clean to have two contexts at that point...

core/dbt/clients/jinja.py
```diff
@@ -305,13 +306,13 @@ def exception_handler(self) -> Iterator[None]:
...
-        if self.stack is None or self.node is None:
+        if self.stack is None:
             yield
         else:
             unique_id = self.macro.unique_id
             depth = self.stack.depth
-            # only mark depth=0 as a dependency
-            if depth == 0:
+            # only mark depth=0 as a dependency, when creating this dependency we don't pass in stack
+            if depth == 0 and self.node:
```
ChenyuLInx, line 315: @gshank  the depth is being kept as 0 because of the comment in I updated

gshank (approved):
This looks so much cleaner!

core/dbt/context/providers.py
```diff
@@ -1249,6 +1249,19 @@ def selected_resources(self) -> List[str]:
...
+        # Check macro_stack and that the unique id is for a materialization macro
+        if not (
+            self.context_macro_stack.depth == 2
+            and self.context_macro_stack.call_stack[1] == "macro.dbt.statement"
+            and "materialization" in self.context_macro_stack.call_stack[0]
+        ):
+            raise RuntimeException(
+                f"submit_python_job is not intended to be called here, at model {parsed_model['alias']}, with macro call_stack {self.context_macro_stack.call_stack}."
+            )
+        return self.adapter.submit_python_job(parsed_model, compiled_code)
```
jtcohen6, line 1263: QQ: Will `adapter.submit_python_job` still be callable (as a classmethod) from within the Jinja context? Would we need to decorate it as "unavailable"?

jtcohen6 (commented):
I agree - looks much cleaner!

core/dbt/context/providers.py
ChenyuLInx: It is no longer available from within Jinja context, we make functions available using `@available` decorator, and I removed that decorator for `submit_python_job` function. Also have a snowflake PR opened for it(given that snowflake actually submit things slightly different). Gonna also sync with the dbt-databricks maintainer to make sure they don't add that `@available` decorator anymore

ChenyuLInx: https://github.com/dbt-labs/dbt-core/pull/5822/files#diff-aa1e9e602a0f1e2930b465b30a35a28778fbe131da4fbf2c4e88dd52ca44f140L1223

PR #5775, opened 2022-09-06
Renaming Attributes In Metric Spec

callum-mcdata:
resolves #5774 

### Description

- Renaming `sql` to `expression`
- Renaming `type` to `calculation_method`
- Renaming `expression` type metrics to `derived`

### Checklist

- [X] I have read [the contributing guide](https://github.com/dbt-labs/dbt-core/blob/main/CONTRIBUTING.md) and understand what's expected of me
- [X] I have signed the [CLA](https://docs.getdbt.com/docs/contributor-license-agreements)
- [ ] I have run this code in development and it appears to resolve the stated issue
- [X] This PR includes tests, or tests are not required/relevant for this PR
- [ ] I have [opened an issue to add/update docs](https://github.com/dbt-labs/docs.getdbt.com/issues/new/choose), or docs changes are not required/relevant for this PR
- [x] I have run `changie new` to [create a changelog entry](https://github.com/dbt-labs/dbt-core/blob/main/CONTRIBUTING.md#Adding-CHANGELOG-Entry)

PR #5825, opened 2022-09-13
Back/fwd compatibility for renamed `metrics` attributes

### LINKED ISSUE #5807: [CT-1148] [Bug] Follow-up to metrics field renaming

### Current Behavior

The field renaming in https://github.com/dbt-labs/dbt-core/pull/5775 means that a user upgrading from v1.2 &rarr; v1.3 will see a parsing error immediately after upgrading, if they have `metrics` already defined in their project:
```
$ dbt parse
dbt.exceptions.ParsingException: Parsing Error
  Invalid metrics config given in FilePath(searched_path='models', relative_path='metric.yml', modification_time=1662969785.677178, project_root='/Users/jerco/dev/scratch/testy') @ metrics: {'name': 'new_customers', 'label': 'New Customers', 'model': "ref('dim_customers')", 'description': 'The number of paid customers using the product', 'type': 'count', 'sql': 'user_id', 'timestamp': 'signup_date', 'time_grains': ['day', 'week', 'month'], 'dimensions': ['plan', 'country'], 'filters': [{'field': 'is_paying', 'operator': 'is', 'value': 'true'}, {'field': 'lifetime_value', 'operator': '>=', 'value': '100'}, {'field': 'company_name', 'operator': '!=', 'value': "'Acme, Inc'"}, {'field': 'signup_date', 'operator': '>=', 'value': "'2020-01-01'"}], 'meta': {'team': 'Finance'}} - at path []: 'calculation_method' is a required property
```

This also breaks backwards compatibility for users of "Slim CI":
```
$ dbt ls -s state:modified --state state/
08:07:02  Encountered an error:
Field "metrics" of type Mapping[str, ParsedMetric] in WritableManifest has invalid value {'metric.my_project.new_customers': {'fqn': ['my_project', 'new_customers'], 'unique_id': 'metric.my_project.new_customers', 'package_name': 'my_project', 'root_path': '/Users/jerco/dev/scratch/testy', 'path': 'metric.yml', 'original_file_path': 'models/metric.yml', 'name': 'new_customers', 'description': 'The number of paid customers using the product', 'label': 'New Customers', 'type': 'count', 'sql': 'user_id', 'timestamp': 'signup_date', 'filters': [{'field': 'is_paying', 'operator': 'is', 'value': 'true'}, {'field': 'lifetime_value', 'operator': '>=', 'value': '100'}, {'field': 'company_name', 'operator': '!=', 'value': "'Acme, Inc'"}, {'field': 'signup_date', 'operator': '>=', 'value': "'2020-01-01'"}], 'time_grains': ['day', 'week', 'month'], 'dimensions': ['plan', 'country'], 'model': "ref('dim_customers')", 'model_unique_id': None, 'resource_type': 'metric', 'meta': {'team': 'Finance'}, 'tags': [], 'sources': [], 'depends_on': {'macros': [], 'nodes': ['model.my_project.dim_customers']}, 'refs': [['dim_customers']], 'metrics': [], 'created_at': 1662969788.598332}}
08:07:02  Traceback (most recent call last):
  File "<string>", line 73, in from_dict
  File "<string>", line 73, in <dictcomp>
  File "<string>", line 76, in from_dict
mashumaro.exceptions.MissingField: Field "calculation_method" of type str is missing in ParsedMetric instance
```

Metrics are still an officially experimental feature, so we have reserved the right to do this — but we should aim for a smoother upgrade experience if possible.

### Expected Behavior

- Upgrading raises a deprecation warning, but does not immediately break. Specifically, if I have a metric definition containing attributes named `sql` and `type`, or `type: expression`, dbt should implicitly rename those and store those attributes with the new names/values instead.
- Compatibility for `--state` comparison to older manifests, if possible. Could we extend the logic in [`upgrade_manifest_json`](https://github.com/dbt-labs/dbt-core/blob/b26280d1cf1af1c715753f73eb3b46f6eaa717f5/core/dbt/contracts/util.py#L210-L217) to that effect? Otherwise, we should declare that v1.3 breaks forwards compatibility for manifests from v1.0-1.2.

### Steps To Reproduce

1. Create a metric [as documented](https://docs.getdbt.com/docs/building-a-dbt-project/metrics)
2. Using dbt-core==1.2, generate a manifest (e.g. via `dbt ls`), and save it somewhere locally (`mv target/manifest.json state/`)
3. Upgrade to dbt-core==1.3
4. Run `dbt parse`
5. Run `dbt ls -s state:modified --state ...` (where `...` is relative path to the previous manifest)

### Relevant log output

_No response_

### Environment

```markdown
- OS: macOS 12.4
- Python: 3.9.13
- dbt: 1.2 -> 1.3
```

_No response_

### Additional Context

_No response_

jtcohen6:
resolves #5807
follow-up to #5775

### Description

Define a method, `rename_metric_attr`, that switches the old metric attribute names + values (`sql`, `type`, `type: expression`) to the new names + values (`expression`, `calculation_method`, `calculation_method: derived`). This is called in two places:
- During parsing, if the old names are present, raise a deprecation warning and then handle the rename
- When loading an old manifest (`--state`), just handle the rename

I also removed some annotations within the metrics parsing code that I believe to be outdated

_Pedantic question, not very important: We tend to call these things "properties" within dbt, but am I right in thinking the Pythonic name for them is dataclass **attributes**?_

### Checklist

- ~I have [opened an issue to add/update docs](https://github.com/dbt-labs/docs.getdbt.com/issues/new/choose), or docs changes are not required/relevant for this PR~


core/dbt/contracts/util.py
```diff
@@ -207,13 +208,45 @@ def get_manifest_schema_version(dct: dict) -> int:
...
 
 
+# we renamed these properties in v1.3
+# this method allows us to be nice to the early adopters
+def rename_metric_attr(data: dict, raise_deprecation_warning: bool = False) -> dict:
+    if raise_deprecation_warning and "sql" in data.keys() or "type" in data.keys():
+        deprecations.warn("metric-attr-renamed")
+    duplicated_attribute_msg = """
+Cannot define both '{}' (old name) and '{}' (new name).
+These are the same attribute. Remove the first and leave the second.
```
jtcohen6, line 218: @callum-mcdata Very open to your wordsmithing on the exception here...

core/dbt/deprecations.py
```diff
@@ -87,6 +87,16 @@ class AdapterDeprecationWarning(DBTDeprecation):
...
 
 
+class MetricAttributesRenamed(DBTDeprecation):
+    _name = "metric-attr-renamed"
+    _description = """\
+dbt-core v1.3 renamed attributes for metrics:
+\n  'sql'              -> 'expression'
+\n  'type'             -> 'calculation_method'
+\n  'type: expression' -> 'calculation_method: derived'
+    """
```
jtcohen6, line 97: ...and the deprecation message here

callum-mcdata: Is it possible to include the specific metric name in the deprecation message? Something like:

    _description = """\
dbt-core v1.3 renamed attributes for metrics:
\n  'sql'              -> 'expression'
\n  'type'             -> 'calculation_method'
\n  'type: expression' -> 'calculation_method: derived'
\n 
\n. The now old metric parameters will be fully deprecated in version 1.4.0. Please remove them from the metric definition of metric {{metric_name}} 
"""

core/dbt/contracts/util.py
```diff
@@ -207,13 +208,45 @@ def get_manifest_schema_version(dct: dict) -> int:
...
+"""
+    if "sql" in data.keys():
+        if "expression" in data.keys():
+            raise ValidationError(duplicated_attribute_msg.format("sql", "expression"))
+        else:
+            data["expression"] = data.pop("sql")
+    if "type" in data.keys():
```
callum-mcdata, line 225: My testing brought my attention to a small issue that people might run into where they might mismatch aspects of the new with the old. Ex:

> calculation_method: expression

Would it be worthwhile to handle that? In my PR I used the following after line 233:

```
    if data["calculation_method"] == "expression":
        calculation_method = "derived"
        data["calculation_method"] = calculation_method
```

core/dbt/contracts/util.py
jtcohen6: Love it. I'm just going to change `parameter` &rarr; `property` because that seems most consistent with what we call these things in our docs today

core/dbt/deprecations.py
jtcohen6: yes for sure!

emmyoop (commented):
Why did the `manifest.json` files get updated?

jtcohen6:
@emmyoop I [added a metric to the project](https://github.com/dbt-labs/dbt-core/pull/5825/files#diff-0daec96b3eabe04477b4cd53a1c3ad5ce9baa15d0256b968ffaf443ce0d05e5bR13-R26) used in `TestPreviousVersionState`, and re-generated the v4-6 manifests (dbt-core v1.0-1.2) using the new project. The test confirms that dbt can successfully read in older `manifest.json`, containing the old metric properties names, and detect zero changes in `state:modified`.

emmyoop:
Oh, so we just didn't have metrics defined in the previous project we were using to generate the manifest?  Makes sense.

callum-mcdata (approved):
This looks good to me! Only one small comment but that's mainly for my own understanding

core/dbt/contracts/graph/unparsed.py
```diff
@@ -1,5 +1,10 @@
 from dbt.node_types import NodeType
-from dbt.contracts.util import AdditionalPropertiesMixin, Mergeable, Replaceable
+from dbt.contracts.util import (
+    AdditionalPropertiesMixin,
+    Mergeable,
+    Replaceable,
+    rename_metric_attr,
```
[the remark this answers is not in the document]
jtcohen6: > apart from data class vs function

I think that's it!

jtcohen6:
Thanks for the functional review @callum-mcdata!

@emmyoop I'll wait to merge till you have a chance to review the code

core/dbt/contracts/util.py
```diff
@@ -207,13 +208,54 @@ def get_manifest_schema_version(dct: dict) -> int:
...
             node_content["raw_code"] = node_content.pop("raw_sql")
         if "compiled_sql" in node_content:
             node_content["compiled_code"] = node_content.pop("compiled_sql")
         node_content["language"] = "sql"
+    for metric_content in manifest.get("metrics", {}).values():
+        # handle attr renames + value translation ("expression" -> "derived")
+        metric_content = rename_metric_attr(metric_content)
+        # mashumaro.exceptions.MissingField: Field "window" of type Optional[MetricTime] is missing in ParsedMetric instance
+        if "window" not in metric_content:
+            metric_content["window"] = None
```
emmyoop, line 258: Instead of this you should change the `ParsedMetric` definition to default to None for window, same as it already does for `UnparsedMetric`.  I would expect that to resolve this.

While looking at that bit of code, I noticed `ParsedMetric` still has `timestamp` as optional and I know @callum-mcdata updated the `UnparsedMetric` definition to _not_ be optional in a recent PR.  `timestamp` should probably match.

emmyoop (changes requested):
Nice temporary help for early adopters!  Just a few suggestions.

core/dbt/deprecations.py
```diff
@@ -87,6 +87,18 @@ class AdapterDeprecationWarning(DBTDeprecation):
...
+class MetricAttributesRenamed(DBTDeprecation):
+    _name = "metric-attr-renamed"
+    _description = """\
+dbt-core v1.3 renamed attributes for metrics:
+\n  'sql'              -> 'expression'
+\n  'type'             -> 'calculation_method'
+\n  'type: expression' -> 'calculation_method: derived'
+\nThe old metric parameter names will be fully deprecated in v1.4.
+\nPlease remove them from the metric definition of metric '{metric_name}'
+"""
```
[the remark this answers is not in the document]
callum-mcdata: Issue created here: https://github.com/dbt-labs/dbt-core/issues/5849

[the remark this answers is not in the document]
callum-mcdata: @emmyoop the link to this GH issue has been added to the deprecation warning!

core/dbt/contracts/util.py
callum-mcdata: @emmyoop you were totally right! I was able to remove this component by setting the default value of window to be None in ParsedMetric. I also fixed the timestamp field to reflect UnparsedMetric like you suggested - must have missed that on my last PR!

core/dbt/deprecations.py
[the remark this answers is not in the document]
emmyoop: @callum-mcdata I actually meant a comment in the code, not as part of the deprecation message.  Sorry if that was unclear!  Unless you think users will find the link to the issue useful?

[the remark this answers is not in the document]
callum-mcdata: 🤦 woops! Although now that I think about it, if we include the link to the issue then anyone who feels very strongly that this behavior should be retained could go to it easily and let us know there. That way the information isn't spread over multiple channels..... so actually I do think it would be useful!

emmyoop (approved):

jtcohen6:
Thanks for the review @emmyoop, and for getting this over the finish line @callum-mcdata!

PR #3969, opened 2021-09-29
Move redshift, snowflake, bigquery plugins

jtcohen6:
> It seems such a waste of time
> If that's what it's all about
> Mama if that's movin' up
> Then I'm movin' out
> I'm movin' out

These are private for now, while we set up automated tests and do some repo housekeeping:

- https://github.com/dbt-labs/dbt-redshift
- https://github.com/dbt-labs/dbt-snowflake
- https://github.com/dbt-labs/dbt-bigquery

PR #174, opened 2016-10-01
Feature/first class pg

drewbanin:
Previously, the only target`type` option was `"redshift"`. Now, this `type` field can be set to `"postgres"` instead. Everything works almost exactly the same, but this differentiation is used for injecting variables into the compilation context.

In this branch, one new context variable is added, `sql_now`. In Redshift, this expands to `getdate()` and in Postgres, it expands to `clock_timestamp()`. While `now()` is more canonical (I think?), `now()` reflects the time of the start of the transaction, whereas `clock_timestamp() reflects the wall-clock time when the statement is executed. The primary use case for `sql_now` is for pre-and-post-hook audit records, we definitely want wall-clock time.

```yml
# profiles.yml
user:
  outputs:
    dev:
      type: postgres
      host: localhost
      user: Drew
....
```

An example of `sql_where` (and macros):
```sql
# audit_macros.sql
{% set audit_table = ref('audit') %}

{% macro log_run_with_status(status) %}

insert into {{ audit_table }}
    (invocation_id, run_started_at, run_at, model, status)
values
    ('{{ invocation_id }}', '{{ run_started_at }}', {{ sql_now }}, '{{ this.name }}', '{{ status }}')
--                                                        ^
--                                                        |
--                                      sql_now ----------+

{% endmacro %}
```

Which compiles to:

```sql
insert into "analytics"."audit"
    (invocation_id, run_started_at, run_at, model, status)
values
    ('{{ invocation_id }}', '{{ run_started_at }}', clock_timestamp(), 'accounts', 'started');
```

In sum, this is a pretty minor change, but it has pretty big ramifications for how we think about packages. Curious to hear your thoughts, @jthandy

PR #5838, opened 2022-09-14
Consolidate date macros into timestamps.sql

### LINKED ISSUE #5521: [CT-895] [Spike] Consolidate current_timestamp & associates
Split out from https://github.com/dbt-labs/dbt-utils/pull/577 + https://github.com/dbt-labs/dbt-utils/pull/597, resolves https://github.com/dbt-labs/dbt-utils/issues/339

### Describe the feature
Rewrite `current_timestamp` + `current_timestamp_utc` to be built out of core/plugin functionality (e.g. `date_function`) and a `convert_timezone` macro.

Currently in `dbt-core`, there are _three_ methods/macros all purporting to do the same thing. I've linked the examples from Postgres.
- [`date_function`](https://github.com/dbt-labs/dbt-core/blob/75f3e8cb749c7b52d226844394ecb34c1e0070a4/plugins/postgres/dbt/adapters/postgres/impl.py#L66-L68) (Python classmethod)
- [`snapshot_get_time`](https://github.com/dbt-labs/dbt-core/blob/75f3e8cb749c7b52d226844394ecb34c1e0070a4/plugins/postgres/dbt/include/postgres/macros/adapters.sql#L131-L133) (macro used for snapshots)
- [`current_timestamp`](https://github.com/dbt-labs/dbt-core/blob/75f3e8cb749c7b52d226844394ecb34c1e0070a4/plugins/postgres/dbt/include/postgres/macros/adapters.sql#L121-L123) (macro used nowhere, except as [_the default implementation_ of `snapshot_get_time`](https://github.com/dbt-labs/dbt-core/blob/75f3e8cb749c7b52d226844394ecb34c1e0070a4/core/dbt/include/global_project/macros/materializations/snapshots/strategies.sql#L50-L59) — but often the two are different, mostly for data type reasons)

Then, there are the two versions currently living in `dbt-utils`:
- [`dbt_utils.current_timestamp`](https://github.com/dbt-labs/dbt-utils/blob/44eeb1978c1c720baca2e2cfe07671e94db02c7c/macros/cross_db_utils/current_timestamp.sql#L5)
- [`dbt_utils.current_timestamp_in_utc`](https://github.com/dbt-labs/dbt-utils/blob/44eeb1978c1c720baca2e2cfe07671e94db02c7c/macros/cross_db_utils/current_timestamp.sql#L31)

### Things we need to figure out

- Are the different implementations here functionally identical? (Doug's initial finding was "yes": https://github.com/dbt-labs/dbt-utils/pull/597#issuecomment-1142239494)
- If the data types are different, why?
- In what cases would their functional result differ? I'm imagining a case where Snowflake parameters [`TIMEZONE`](https://docs.snowflake.com/en/sql-reference/parameters.html#timezone) and [`TIMESTAMP_TYPE_MAPPING`](https://docs.snowflake.com/en/sql-reference/parameters.html#timestamp-type-mapping) are both set

### Ideal outcome

When I close my eyes and imagine how this could be really good:
1. We have _one_ method/macro, defined in `dbt-core` and implemented by each adapter plugin, that provides the SQL for getting the current timestamp. Whenever we need the current timestamp in a different data type, we use that `current_timestamp` macro plus a data type conversion function (see dbt-labs/dbt-utils#586).
2. We have a `convert_timezone` macro, defined in `dbt-core` and implemented by each adapter plugin. @clausherther [has already written one](https://github.com/calogica/dbt-date/blob/main/macros/calendar_date/convert_timezone.sql) for the `dbt-date` package—Claus, would you have any interest in kicking it upstream? :)
3. It's trivial to write a `current_timestamp_in_utc` macro that builds on top of 1+2. Whether that lives in `dbt-core`, `dbt-utils`, or somewhere else doesn't matter so much to me. Wherever it lives, it should look a lot like [`dbt_date.now(tz=None)`](https://github.com/calogica/dbt-date/blob/main/macros/calendar_date/now.sql). We'd leave the `dbt_utils` one in place for backwards compatibility.

Leaving all of it where it is :)

### Additional context

Timezones are no fun

This stuff is pretty tricky to test. [We're not _really_ testing it currently](https://github.com/dbt-labs/dbt-utils/blob/44eeb1978c1c720baca2e2cfe07671e94db02c7c/tests/functional/cross_db_utils/fixture_current_timestamp.py#L4). I had an idea in https://github.com/dbt-labs/dbt-utils/pull/577#discussion_r869149492, but it's not a great one. We are indirectly testing our use of `snapshot_get_time` in `dbt-core` functional tests (specifically [`test_hard_delete_snapshot`](https://github.com/dbt-labs/dbt-core/blob/main/tests/functional/simple_snapshot/test_hard_delete_snapshot.py).

### Who will this benefit?

- Users who want to call well-named macros, and expect them to do the thing they're meant to
- Claus, as maintainer of the `dbt-date` package (I hope!)

colin-rogers-dbt:
resolves #5521

related PRs:
* https://github.com/dbt-labs/dbt-snowflake/pull/273
* https://github.com/dbt-labs/dbt-redshift/pull/191
* https://github.com/dbt-labs/dbt-spark/pull/480
* https://github.com/dbt-labs/dbt-bigquery/pull/323

### Checklist


gshank (approved):
Looks good.

McKnight-42 (approved):
Looks good!

core/dbt/include/global_project/macros/adapters/timestamps.sql
```diff
@@ -0,0 +1,44 @@
...
+{% macro default__current_timestamp() %}
+    current_timestamp
+{% endmacro %}
+
+{%- macro current_timestamp_in_utc() -%}
+    {{ adapter.dispatch('current_timestamp_in_utc', 'dbt')() }}
+{%- endmacro -%}
+
+{% macro default__current_timestamp_in_utc() %}
+    {{ default__current_timestamp() }} at time zone 'utc'
```
jtcohen6, line 14: IMO `default__current_timestamp_in_utc` should just be `default__current_timestamp` + `default__convert_timezone`

Even though the end result is basically the same on Postgres, it avoids the need to reimplement `current_timestamp_in_utc` if an adapter plugin has successfully reimplemented `current_timestamp` + `convert_timezone`

core/dbt/include/global_project/macros/adapters/timestamps.sql
```diff
@@ -0,0 +1,44 @@
...
+{%- macro convert_timezone(timestamp,  target_tz, source_tz) -%}
+    {%- if not target_tz is string -%}
+        {{ exceptions.raise_compiler_error("'target_tz' must be a string") }}
+    {%- else -%}
+        {{ adapter.dispatch('convert_timezone', 'dbt') (column, target_tz, source_tz) }}
+    {%- endif -%}
+
+{%- endmacro -%}
+
+{%- macro default__convert_timezone(column, target_tz, source_tz) -%}
```
jtcohen6, line 34: This takes Postgres's syntax as the default, which is often a fair play. In this case, the majority of other databases support `convert_timezone(source_tz, target_tz, column)` — so it probably makes more sense to have that one by our default in `dbt-core`, and reimplement just for `postgres__`.

_yes, `source_tz` is optional, doesn't it just delight you that the _optional_ argument is the first positional one??_

core/dbt/include/global_project/macros/adapters/timestamps.sql
```diff
@@ -0,0 +1,44 @@
...
+    {%- if not source_tz -%}
+        {{ column }} at time zone '{{ target_tz }}'
+    {%- else -%}
+        {{ column }} at time zone '{{ source_tz }}' at time zone '{{ target_tz }}'
+    {%- endif -%}
+{%- endmacro -%}
+
+{%- macro now(tz=None) -%}
+{{ convert_timezone(current_timestamp_in_utc(), tz) }}
+{%- endmacro -%}
```
jtcohen6, line 44: Is the idea here to offer an alias for `{{ convert_timezone(tz, current_timestamp_in_utc()) }}`, in keeping with Postgres-y syntax?

jtcohen6 (changes requested):
Very very happy to have these macros living in one place! Even if inconsistency must remain, to support backwards compatibility, we should aim to concentrate and cordon off that inconsistency as much as possible, highlight the contradictions, and provide clear guidance for net-new projects going forward.

---

I like the idea of trying to also incorporate the changes I recommended in https://github.com/dbt-labs/dbt-core/issues/5521#issuecomment-1246795511, which would enable the full removal of `dbt_utils.current_timestamp` by dbt-utils v1.0.

---

What do you think about deprecating [`date_function`](https://github.com/dbt-labs/dbt-core/blob/e00eb9aa3a173ecf40f0e0190fb496a64d3ca876/plugins/postgres/dbt/adapters/postgres/impl.py#L66-L68) + [`sql_now()`](https://github.com/dbt-labs/dbt-core/blob/e00eb9aa3a173ecf40f0e0190fb496a64d3ca876/core/dbt/context/providers.py#L1183-L1185)? That's the last tine in our fork of inconsistency.

Specifically, we could:
- Stop making the [base `date_function`](https://github.com/dbt-labs/dbt-core/blob/e00eb9aa3a173ecf40f0e0190fb496a64d3ca876/core/dbt/adapters/base/impl.py#L483-L487) an abstract method that every adapter must implement
- Add a [deprecation warning](https://github.com/dbt-labs/dbt-core/blob/main/core/dbt/deprecations.py) for anyone who calls `sql_now` in their project code. The deprecation message can say something like: "`{{ sql_now }}` and `{{ adapter.date_function }}` are deprecated, use `{{ current_timestamp() }}` or `{{ now() }}` instead." The `sql_now` function is very very old (https://github.com/dbt-labs/dbt-core/pull/174), I don't know of anyone who actually uses it, but we're better safe than sorry. Our `deprecations` module shows a warning to the user, and it also fires an anonymous tracking event that can confirm for us whether anyone is actually using the thing.

---

Lastly, we'll want companion PRs in all our adapter plugins, and to test each of them thoroughly. I don't think our existing suite of functional tests is sensitive enough to catch subtly breaking changes in these macros.

@dbeatty10 - do you have thoughts on how we might write more sensitive automated tests? How did you pursue the work of comparing `dbt.current_timestamp()` and `dbt_utils.current_timestamp()` in https://github.com/dbt-labs/dbt-utils/pull/597#issuecomment-1231074577?

core/dbt/include/global_project/macros/adapters/timestamps.sql
dbeatty10: I couldn't find `NOW` within the SQL standard, and implementation of it across a wide variety of databases is hit-and-miss.

Summary across the database adapters maintained by dbt Labs:
- It doesn't look like either Snowflake or BigQuery implement a `now` function.
- Redshift [deprecated their implementation](https://docs.aws.amazon.com/redshift/latest/dg/c_SQL_functions_leader_node_only.html#:~:text=The%20following%20leader%2Dnode%20only%20functions%20are%20deprecated) of `now`.
- Postgres has a built-in function named `now()` which is an alias for [`transaction_timestamp`](https://www.postgresql.org/docs/current/functions-datetime.html#FUNCTIONS-DATETIME-CURRENT:~:text=now%20(%20)%20%E2%86%92%20timestamp%20with%20time%20zone). It returns the "current date and time (start of current transaction)". The return type is equivalent to Snowflake's `timestamp_ltz`.
- In Spark and Databricks, `now` is an alias for `current_timestamp` (with a return type also equivalent to Snowflake's `timestamp_ltz`).

core/dbt/include/global_project/macros/adapters/timestamps.sql
colin-rogers-dbt: gotcha, will update, let's definitely avoid for the future

core/dbt/include/global_project/macros/adapters/timestamps.sql
colin-rogers-dbt: I think it's nice syntactic sugar/short hand but not necessary

tests/functional/timestamps/test_timestamps.py
```diff
@@ -0,0 +1,25 @@
...
+        results = run_dbt(["run"])
+        assert len(results) == 1
+        check_relation_has_expected_schema(
+            project.adapter,
+            relation_name="get_current_timestamp",
+            expected_schema={
+                "current_timestamp": "timestamp with time zone",
+                "current_timestamp_in_utc": "timestamp without time zone",
+            },
+        )
```
dbeatty10, line 25: This looks like a common way that schema comparisons are accomplished currently within pytests:
https://github.com/dbt-labs/dbt-core/blob/3bfce2bac91a89a2886785e2fb31315e41830239/tests/adapter/dbt/tests/adapter/utils/data_types/base_data_type_macro.py#L26-L34

It uses this function:
https://github.com/dbt-labs/dbt-core/blob/0bd6df0d1b1e4e382c413929403f7972108a5b8a/core/dbt/tests/util.py#L447-L453

dbeatty10 (commented):
@colin-rogers-dbt I was working on some pytests in a totally separate PR and saw the code that Jerco mentioned to use yesterday that does schema comparisons. Linked inline.

tests/functional/timestamps/test_timestamps.py
colin-rogers-dbt: yea I saw this but it means I need to create a comparison table which feels unnecessary and less explicit

core/dbt/include/global_project/macros/adapters/timestamps.sql
```diff
@@ -0,0 +1,73 @@
...
+
+{% macro current_timestamp_backcompat() %}
+    {{ return(adapter.dispatch('current_timestamp_backcompat', 'dbt')()) }}
+{% endmacro %}
+
+{% macro default__current_timestamp_backcompat() %}
+    {{ return(adapter.dispatch('current_timestamp', 'dbt')()) }}
+{% endmacro %}
+
+-- TODO move to dbt-postgres
```
jtcohen6, line 64: move to `dbt-postgres` (the file below) before merge

core/dbt/include/global_project/macros/adapters/timestamps.sql
```diff
@@ -0,0 +1,73 @@
...
+{% macro postgres__current_timestamp_backcompat() %}
+    current_timestamp::{{ type_timestamp() }}
+    
+{% endmacro %}
+
+-- TODO move to dbt-snowflake
```
jtcohen6, line 70: move to `dbt-snowflake` (new PR in other repo) before merge

jtcohen6 (commented):
Very very happy to have this in for v1.3, and smooth the path for dbt-utils v1.0!

tests/adapter/dbt/tests/adapter/basic/test_timestamps.py
```diff
@@ -0,0 +1,25 @@
+import pytest
+from dbt.tests.util import check_relation_has_expected_schema, run_dbt
```
colin-rogers-dbt, line 2: @jtcohen6 migrate tests like this?

jtcohen6: yup!

by convention, we have reserved `adapter/basic` as, the basic tests to start with when developing/testing a new adapter: https://docs.getdbt.com/docs/contributing/adapter-development/4-testing-a-new-adapter#getting-started-running-basic-tests

I think this specific test might want to live in [`dbt.tests.adapter.utils.data_types`](https://github.com/dbt-labs/dbt-core/tree/main/tests/adapter/dbt/tests/adapter/utils/data_types)

core/dbt/include/global_project/macros/adapters/timestamps.sql
colin-rogers-dbt: if we do it before merge we can't add the functional testing defined in this merge: https://github.com/dbt-labs/dbt-snowflake/pull/273

core/dbt/include/global_project/macros/adapters/timestamps.sql
colin-rogers-dbt: moved

core/dbt/include/global_project/macros/adapters/timestamps.sql
jtcohen6: looks like the macro is already over there! that's the only place the `snowflake__` version needs to live. (or maybe I'm not following what you mean)

tests/adapter/dbt/tests/adapter/basic/test_timestamps.py
colin-rogers-dbt: I thought those were for data type casting and I didn't want it to get confused with the existing type_timestamp tests

core/dbt/include/global_project/macros/adapters/timestamps.sql
```diff
@@ -0,0 +1,73 @@
...
+
+{%- macro snapshot_get_time() -%}
+    {{ adapter.dispatch('snapshot_get_time', 'dbt')() }}
+{%- endmacro -%}
+
+{% macro default__snapshot_get_time() %}
+    {{ current_timestamp() }}
+{% endmacro %}
+
+{%- macro convert_timezone(source_tz, target_tz, timestamp) -%}
```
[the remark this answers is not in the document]
colin-rogers-dbt: good catch! I think `source_timestamp` is a solid choice

tests/adapter/dbt/tests/adapter/basic/test_timestamps.py
jtcohen6: ah you're right! maybe just a new subfolder under [`utils`](https://github.com/dbt-labs/dbt-core/tree/main/tests/adapter/dbt/tests/adapter/utils)

colin-rogers-dbt: moved to `utils`

core/dbt/include/global_project/macros/adapters/timestamps.sql
colin-rogers-dbt: just noticed it's the same as what's in there! Also I think the order is:
1. Merge this
2. Add adapters timestamp macros
3. remove timestamp code from dbt-utils
I *think* we can do steps 2 & 3 concurrently

dbeatty10: Close. Here's the workflow (start to finish) that we typically use for changes across core + adapters:
1. Open PR for dbt-core (like you did [here](https://github.com/dbt-labs/dbt-core/pull/5838) already)
1. Open PRs for each of the adapters (like you did [here](https://github.com/dbt-labs/dbt-snowflake/pull/273/files), etc already)
1. _Temporarily_ update the `dev-requirements.txt` file within each adapter to point to your PR in Step 1, like [this](https://github.com/dbt-labs/dbt-bigquery/pull/308/commits/7f795796abe4e15b8627d6cf086accce772b5685).
1. Once all PRs are passing CI, reviewed, and approved, then merge the dbt-core PR
1. Undo the temporary update to `dev-requirements.txt` in each of the adapter PRs (so that they point at the `main` branch of dbt-core again) like [this](https://github.com/dbt-labs/dbt-bigquery/pull/313/commits/27d28176b0148263cb05b41be2230ffc43f42700)
1. Once CI passes, then all the adapter PRs can be merged as well.

dbeatty10: Here's the full life-cycle of a similar adapter PR for reference:
https://github.com/dbt-labs/dbt-bigquery/pull/313/commits

And the associated dbt-core PR:
https://github.com/dbt-labs/dbt-core/pull/5875/commits

dbeatty10: It would probably be handy if we write out this workflow in some centralized place for easier reference.

colin-rogers-dbt: that's a cool process, however I'm not sure we need to do that here since we haven't removed these macros from dbt-utils yet

dbeatty10: We've typically coordinated the changes in dbt-core and each dbt adapter in concert (as described above).

The alternative would be to merge a dbt-core in isolation and then do each adapter PR separately afterwards.

⚠️ A situation that can arise when merging an isolated feature to dbt-core:
- the implementation might need to change based on what we learn  and experience within the adapter PRs

The advantage of coordinating the adapter+core PRs is preventing those types of situations.

### In the days of yore
Historically, our adapter code was in the same repo as core (in the [plugins](https://github.com/dbt-labs/dbt-core/tree/main/plugins) directory). In those days (not yet even one year ago!), you'd just do a single PR 'cause everything was in the same repo.

The adapters moved out to their own repos in https://github.com/dbt-labs/dbt-core/pull/3969:
image

(There's an Easter egg in #3969 for Billy Joel fans (like me!))

.changie.yaml
```diff
@@ -5,15 +5,15 @@ versionHeaderPath: ""
 changelogPath: CHANGELOG.md
 versionExt: md
 versionFormat: '## dbt-core {{.Version}} - {{.Time.Format "January 02, 2006"}}'
-kindFormat: '### {{.Kind}}'
-changeFormat: '- {{.Body}} ([#{{.Custom.Issue}}](https://github.com/dbt-labs/dbt-core/issues/{{.Custom.Issue}}), [#{{.Custom.PR}}](https://github.com/dbt-labs/dbt-core/pull/{{.Custom.PR}}))'
+kindFormat: "### {{.Kind}}"
+changeFormat: "- {{.Body}} ([#{{.Custom.Issue}}](https://github.com/dbt-labs/dbt-core/issues/{{.Custom.Issue}}), [#{{.Custom.PR}}](https://github.com/dbt-labs/dbt-core/pull/{{.Custom.PR}}))"
```
dbeatty10, line 9: Can we merge the updates to `.changie.yml` in a separate PR?

dbeatty10 (changes requested):
Is there a reason we don't need a separate PR opened for each adapter for this change?

There seem to be a couple primary changes within dbt-snowflake [here](https://github.com/dbt-labs/dbt-snowflake/pull/273/files), and I'd expect similar changes within _all_ of the adapters unless there is a particular reason they are left out.

With a new test class like `TestCurrentTimestamps`, we need to explicitly add it within each adapter repo and run it through CI.

Here's the workflow (start to finish) that we typically use for changes in core that affect adapters:
1. Open PR for dbt-core (like you did [here](https://github.com/dbt-labs/dbt-core/pull/5838) already)
1. Open PRs for each of the adapters (like you did [here](https://github.com/dbt-labs/dbt-snowflake/pull/273/files), etc already)
    - dbt-snowflake
    - dbt-redshift
    - dbt-bigquery
    - dbt-spark
1. _Temporarily_ update the `dev-requirements.txt` file within each adapter to point to your PR in Step 1, like [this](https://github.com/dbt-labs/dbt-bigquery/pull/308/commits/7f795796abe4e15b8627d6cf086accce772b5685).
1. Once all PRs are passing CI, reviewed, and approved, then merge the dbt-core PR
1. Undo the temporary update to `dev-requirements.txt` in each of the adapter PRs (so that they point at the `main` branch of dbt-core again) like [this](https://github.com/dbt-labs/dbt-bigquery/pull/313/commits/27d28176b0148263cb05b41be2230ffc43f42700)
1. Once CI passes, then all the adapter PRs can be merged as well.

.changie.yaml
colin-rogers-dbt: looks like some extra formatting changes crept in here, removed those and just kept adding my user name

plugins/postgres/dbt/include/postgres/macros/timestamps.sql
```diff
@@ -0,0 +1,17 @@
...
+{%- endmacro %}
+
+{% macro postgres__snapshot_string_as_time(timestamp) -%}
+    {%- set result = "'" ~ timestamp ~ "'::timestamp without time zone" -%}
+    {{ return(result) }}
+{%- endmacro %}
+
+{% macro postgres__snapshot_get_time() -%}
+    {{ current_timestamp() }} :: TIMESTAMP without TIME ZONE
+{%- endmacro %}
```
dbeatty10, line 12: Nit:
Is there a reason why we don't want to preserve the casing and spacing of the original implementation?
```suggestion
{% macro postgres__snapshot_get_time() -%}
  {{ current_timestamp() }}::timestamp without time zone
{%- endmacro %}
```

tests/adapter/dbt/tests/adapter/utils/test_timestamps.py
```diff
@@ -0,0 +1,29 @@
+import pytest
+from dbt.tests.util import check_relation_has_expected_schema, run_dbt
+
+_MODEL_CURRENT_TIMESTAMP = """
+SELECT {{current_timestamp()}} as current_timestamp,
+       {{current_timestamp_in_utc()}} as current_timestamp_in_utc
+"""
```
dbeatty10, line 7: We typically [lowercase `select`](https://github.com/dbt-labs/corp/blob/4b553082bf0fb8e285934195331925461bef4784/dbt_style_guide.md?plain=1#L100) and use [spaces inside Jinja delimiters](https://github.com/dbt-labs/corp/blob/4b553082bf0fb8e285934195331925461bef4784/dbt_style_guide.md?plain=1#L236).
```suggestion
_MODEL_CURRENT_TIMESTAMP = """
SELECT {{ current_timestamp() }} as current_timestamp,
       {{ current_timestamp_in_utc() }} as current_timestamp_in_utc
"""
```

dbeatty10 (changes requested):
What are your thoughts or plans for testing the following new macros?
- `convert_timezone`
- `current_timestamp_in_utc`

Of course we always want to do testing on any new macros, but it has enhanced relevance here because the `at time zone` syntax in Postgres and the `convert_timezone()` function in Snowflake have complicated behavior (e.g., the output data type changes depending on the data type of the inputs):
- https://docs.snowflake.com/en/sql-reference/functions/convert_timezone.html
- https://www.enterprisedb.com/postgres-tutorials/postgres-time-zone-explained

Could you implement a test that confirms that `current_timestamp_in_utc` yields the expected _value_ as well as the expected data type? I'd suggest something like [this](https://github.com/dbt-labs/dbt-core/pull/5935/files#diff-2ac0a701cbb34cffba92f797869e76e9ef01913841d006d074eb0248c2794268R39-R46).

The type of thing I'm worried about is if a user has their database timezone set to some non-UTC timezone and somehow `current_timestamp_in_utc` accidentally reports a timezone-naive local time that isn't actually in alignment with UTC.

---
Also:

Do you know if the data type (or value) of `current_timestamp_in_utc` will be different than the implementation within dbt-utils for any adapters? (I didn't check those implementations, so I don't know.)

colin-rogers-dbt:
> Could you implement a test that confirms that `current_timestamp_in_utc` yields the expected _value_ as well as the expected data type? I'd suggest something like [this](https://github.com/dbt-labs/dbt-core/pull/5935/files#diff-2ac0a701cbb34cffba92f797869e76e9ef01913841d006d074eb0248c2794268R39-R46).

While we certainly could I do not see a compelling case for doing so as we already have added more testing for this than other macros, if a particular adapter has this as a concern it wouldn't be difficult to add this in the adapter specific implementation.

> The type of thing I'm worried about is if a user has their database timezone set to some non-UTC timezone and somehow `current_timestamp_in_utc` accidentally reports a timezone-naive local time that isn't actually in alignment with UTC.

This is a valid concern for redshift where the established behavior for _utc relies on it's default behavior but adding a test for utc won't actually catch this. 

> Do you know if the data type (or value) of `current_timestamp_in_utc` will be different than the implementation within dbt-utils for any adapters? (I didn't check those implementations, so I don't know.)

It will not, I'm specifically checking for that in the adapter level tests.

colin-rogers-dbt:
Received offline feedback concerning the need to include a "backcompat" macro for `current_timestamp_in_utc` as we may be diverging from dbt-utils versions in a breaking manner. 
Digging in there are two adapter specific _utc implementations (that don't just use current_timestamp):
* Snowflake
* Postgres 
* Redshift 
Looking at the three we aren't changing the datatype (and checking the values)

core/dbt/include/global_project/macros/adapters/timestamps.sql
```diff
@@ -0,0 +1,62 @@
...
+{% endmacro %}
+
+{%- macro current_timestamp_in_utc() -%}
+    {{ adapter.dispatch('current_timestamp_in_utc', 'dbt')() }}
+{%- endmacro -%}
+
+{% macro default__current_timestamp_in_utc() %}
+    {{ convert_timezone(target_tz="UTC",
+        source_timestamp=current_timestamp())}}
+{% endmacro %}
```
dbeatty10, line 16: @jtcohen6

From a product perspective, what do you want the return type to be for this version of `current_timestamp_in_utc`?

Let's suppose for simplicity that all data platforms have a timestamp data types with equivalent semantics to those defined in the SQL standard \*:
- `timestamp without time zone`
- `timestamp with time zone`

Then there are three options for the data type produced by `current_timestamp_in_utc`:
1. `timestamp without time zone` for all adapters
1. `timestamp with time zone` for all adapters
1. a mix of return types across adapters -- some with one type and some with the other

---

\* Although extremely uncommon, there are some platforms that only support a single timestamp data type (e.g., Spark and Databricks).

jtcohen6: Opened a new issue to answer this thread, and reflect our discussion earlier today: https://github.com/dbt-labs/dbt-core/issues/5969

jtcohen6:
@colin-rogers-dbt @dbeatty10 Thanks for all the great discussion in here!

I had a chance to chat with Doug earlier today, and hear his perspective on the complexity that he's been going deep on, and simmers under the surface. There's quite a lot!

To simplify matters, here are my proposed next steps:
- Split this PR into two.
- The first PR should _just_ provide backward compatible implementations for `dbt_utils.current_timestamp` and `dbt_utils.current_timestamp_in_utc` — via `dbt.current_timestamp_backcompat` and `dbt.current_timestamp_in_utc_backcompat`, respectively — so that we can unblock dbt-utils v1.0. **Exit criteria:** The return types and values of these back-compat macros match **exactly** the return types and values of the deprecated `dbt_utils` macros. (I'd also be happy to see the reorganization / centralization of our existing macros.)
- The second PR should add in `dbt.convert_timezone` + `dbt.current_timestamp_in_utc`, as net-new macros that do not exist currently. I've opened a new issue outlining the spec for both: https://github.com/dbt-labs/dbt-core/issues/5969
- For future (tech debt): https://github.com/dbt-labs/dbt-core/issues/5967

In terms of timing, we only _need_ the first PR for inclusion in v1.3, to unblock dbt-utils v1.0. The second PR could be merged for inclusion in v1.4 instead.

tests/adapter/dbt/tests/adapter/utils/test_timestamps.py
```diff
@@ -0,0 +1,31 @@
...
+from dbt.tests.util import check_relation_has_expected_schema, run_dbt
+
+_MODEL_CURRENT_TIMESTAMP = """
+select {{ current_timestamp() }} as current_timestamp,
+       {{ current_timestamp_in_utc_backcompat() }} as current_timestamp_in_utc_backcompat,
+       {{ current_timestamp_backcompat() }} as current_timestamp_backcompat
+"""
+
+
+class TestCurrentTimestamps:
```
dbeatty10, line 11: The patten I'm used to seeing is like [this](https://github.com/dbt-labs/dbt-core/blob/c2ab2971b073245366f1ccd8a9e1cbe198f18865/tests/adapter/dbt/tests/adapter/grants/test_invalid_grants.py).

```
import pytest
from dbt.tests.adapter.category.base_thing import BaseThing

class BaseSpecificThing(BaseThing):
    <a bunch of code>

class TestSpecificThing(BaseSpecificThing):
    pass
```

---

Is there a purpose for separating these two files that I'm overlooking?
- `tests/adapter/dbt/tests/adapter/utils/test_timestamps.py`
- `tests/functional/timestamps/test_timestamps.py` 

Or could they be a single file like the pattern above?

---

Assuming they can be a single file, then a three-part suggestion:
1. Make the update suggested below
1. Add the trivial wrapper to the end of this file
1. Remove the other wrapper file

### Implementation details
1. Suggested update
```suggestion
class BaseCurrentTimestamps:
```

2. Add this at the end of the file:
```
class TestCurrentTimestamps(BaseCurrentTimestamps):
    pass
```

3. If you do this, then you can drop the `tests/functional/timestamps/test_timestamps.py` file altogether.

test/unit/test_macro_calls.py
```diff
@@ -23,7 +23,7 @@ def setUp(self):
                     *
                     from {{ model }} )
                 {% endmacro %}""",
-            "{% macro test_my_test(model) %} select {{ dbt_utils.current_timestamp() }} {% endmacro %}",
+            "{% macro test_my_test(model) %} select {{ current_timestamp_backcompat() }} {% endmacro %}",
```
dbeatty10, line 26: Good catch.

For completeness, probably want to update references to `dbt_utils.current_timestamp` here too (although I don't know what the appropriate content would be):
- `core/dbt/clients/jinja_static.py`

core/dbt/include/global_project/macros/adapters/timestamps.sql
```diff
@@ -0,0 +1,43 @@
+{%- macro current_timestamp() -%}
+    {{ adapter.dispatch('current_timestamp', 'dbt')() }}
+{%- endmacro -%}
+
+{% macro default__current_timestamp() %}
+    current_timestamp
+{% endmacro %}
```
dbeatty10, line 7: `current_timestamp` as the default SQL for `{{ dbt.current_timestamp() }}` makes a lot of sense since it *is* a SQL standard 🧠 

But a few questions:
1. Do we _need_ a new default implementation here?
1. Do we _want_ a new default implementation here?
1. Why didn't the original author(s) do this in the first place?

### Context

This would replace the current implementation of `default__current_timestamp()` which is:

```sql
{% macro default__current_timestamp() -%}
  {{ exceptions.raise_not_implemented(
    'current_timestamp macro not implemented for adapter '+adapter.type()) }}
{%- endmacro %}
```

Tracing this implementation back in time goes to [here](https://github.com/dbt-labs/dbt-core/commit/5e8ab9ce4a8ac905d66bf85414a0be121fc58dbd#diff-29395c37ca37188832ed290a098706ca9b366055fa578d2524fde72ad6e3758dR233-R236).

Is this a [Chesterton's fence](https://sketchplanations.com/chestertons-fence)? I don't know! 🤷 We should figure it out one way or the other.

### Recommendation

If @jtcohen6 gives this new default implementation a 👍, I don't know of any technical reasons why _not_ to do it.

In a vacuum, my recommendation would be to leave the current implementation as-is _unless_ we can explain why the proposed replacement is more relevant than the previous implementation.

dbeatty10 (changes requested):
One easy way to test the backwards compatibility is just to compare the new SQL to the old SQL. If they are the same, then we are done.

An alternative testing method is to actually run the rendered SQL and compare the results. This alternative would be a lot more complicated since timestamp functions (and [sometimes even an explicit data type!](https://docs.snowflake.com/en/sql-reference/parameters.html#label-timestamp-type-mapping)) will vary their behavior based on session variables.

Could we take the easy approach here rather than the harder alternative? 🙏 

During manual testing, I didn't see the same rendered SQL between the `_backcompat` macros and the ones from dbt_utils:
- `dbt.current_timestamp_backcompat()`
- `dbt.current_timestamp_in_utc_backcompat()`

Could you take a look and see if my testing is giving the incorrect results?

<details>
<summary>Here's the output (with light manual re-formatting for readability)</summary>

&nbsp;

### dbt-postgres

```shell
dbt compile --target postgres
```

```sql
postgres dbt.current_timestamp():
    now()

postgres dbt_utils.current_timestamp(): 
    current_timestamp::TIMESTAMP

postgres dbt.current_timestamp_backcompat(): 
    current_timestamp::TIMESTAMP

postgres dbt_utils.current_timestamp_in_utc(): 
    (current_timestamp at time zone 'utc')::TIMESTAMP

postgres dbt.current_timestamp_in_utc_backcompat(): 
    current_timestamp::TIMESTAMP

```

&nbsp;&nbsp;

### dbt-snowflake

```shell
dbt compile --target snowflake
```

```sql
snowflake dbt.current_timestamp():
    convert_timezone('UTC', current_timestamp())

snowflake dbt_utils.current_timestamp(): 
    current_timestamp::TIMESTAMP

snowflake dbt.current_timestamp_backcompat(): 
    current_timestamp()::timestamp_ntz

snowflake dbt_utils.current_timestamp_in_utc(): 
    convert_timezone('UTC', current_timestamp::TIMESTAMP)::TIMESTAMP

snowflake dbt.current_timestamp_in_utc_backcompat(): 
    current_timestamp()::timestamp_ntz

```

&nbsp;&nbsp;

### dbt-redshift

```shell
dbt compile --target redshift
```

```sql
redshift dbt.current_timestamp():
    getdate()

redshift dbt_utils.current_timestamp(): 
    getdate()

redshift dbt.current_timestamp_backcompat(): 
    current_timestamp::TIMESTAMP

redshift dbt_utils.current_timestamp_in_utc():     
    getdate()

redshift dbt.current_timestamp_in_utc_backcompat(): 
    current_timestamp::TIMESTAMP

```

&nbsp;&nbsp;

### dbt-bigquery

```shell
dbt compile --target bigquery
```

```sql
bigquery dbt.current_timestamp():
    CURRENT_TIMESTAMP()

bigquery dbt_utils.current_timestamp(): 
    current_timestamp

bigquery dbt.current_timestamp_backcompat():
    CURRENT_TIMESTAMP()

bigquery dbt_utils.current_timestamp_in_utc():     
    current_timestamp

bigquery dbt.current_timestamp_in_utc_backcompat():
    CURRENT_TIMESTAMP()
```

&nbsp;&nbsp;

### dbt-spark

```shell
dbt compile --target spark
```

```sql
spark dbt.current_timestamp():
    current_timestamp()

spark dbt_utils.current_timestamp(): 
    current_timestamp::timestamp

spark dbt.current_timestamp_backcompat():
    current_timestamp()

spark dbt_utils.current_timestamp_in_utc(): 
    current_timestamp::timestamp

spark dbt.current_timestamp_in_utc_backcompat():
    current_timestamp()
```

&nbsp;&nbsp;

### dbt-databricks

```shell
dbt compile --target databricks
```

```sql
databricks dbt.current_timestamp():
    current_timestamp()

databricks dbt_utils.current_timestamp(): 
    current_timestamp::timestamp

databricks dbt.current_timestamp_backcompat():
    current_timestamp()

databricks dbt_utils.current_timestamp_in_utc(): 
    current_timestamp::timestamp

databricks dbt.current_timestamp_in_utc_backcompat():
    current_timestamp()
```

&nbsp;

</details>

<details>
<summary>Here's how to replicate the output</summary>

### `dev-requirements.txt`
```
# dbt-core
git+https://github.com/dbt-labs/dbt-core.git@consolidate_timestamp_logic#egg=dbt-core&subdirectory=core
git+https://github.com/dbt-labs/dbt-core.git@consolidate_timestamp_logic#egg=dbt-tests-adapter&subdirectory=tests/adapter

# dbt-postgres
git+https://github.com/dbt-labs/dbt-core.git@consolidate_timestamp_logic#egg=dbt-postgres&subdirectory=plugins/postgres

# dbt-snowflake
git+https://github.com/dbt-labs/dbt-snowflake.git@consolidateTimestampMacros#egg=dbt-snowflake

# dbt-bigquery
git+https://github.com/dbt-labs/dbt-bigquery.git@main#egg=dbt-bigquery

# dbt-redshift
git+https://github.com/dbt-labs/dbt-redshift.git@main#egg=dbt-redshift

# dbt-spark
git+https://github.com/dbt-labs/dbt-spark.git@main#egg=dbt-spark[PyHive]

# dbt-databricks
git+https://github.com/databricks/dbt-databricks.git@main#egg=dbt-databricks
```

### Install

```shell
python3 -m venv env
source env/bin/activate
pip install --upgrade pip
pip install -r dev-requirements.txt
source env/bin/activate
```

### Configure

Within `profiles.yml`, configure connections to each of the adapters to test.


### Create a dbt project
Create a dbt project however you prefer (`dbt init`, preexisting template, etc).

### `analysis/current_timestamp.sql`

:point_up: Create this file in your project:
```sql
{{ adapter.type() }} dbt.current_timestamp():
    {{ dbt.current_timestamp() }}

{{ adapter.type() }} dbt_utils.current_timestamp(): {{ dbt_utils.current_timestamp() }}
{{ adapter.type() }} dbt.current_timestamp_backcompat(): {{ dbt.current_timestamp_backcompat() }}
{{ adapter.type() }} dbt_utils.current_timestamp_in_utc(): {{ dbt_utils.current_timestamp_in_utc() }}
{{ adapter.type() }} dbt.current_timestamp_in_utc_backcompat(): {{ dbt.current_timestamp_in_utc_backcompat() }}
```

### Run

```shell
dbt compile --target postgres && (cat target/compiled/test_project/analyses/current_timestamp.sql > compiled_current_timestamp.md); \
dbt compile --target snowflake && (cat target/compiled/test_project/analyses/current_timestamp.sql >> compiled_current_timestamp.md); \
dbt compile --target redshift && (cat target/compiled/test_project/analyses/current_timestamp.sql >> compiled_current_timestamp.md); \
dbt compile --target bigquery && (cat target/compiled/test_project/analyses/current_timestamp.sql >> compiled_current_timestamp.md); \
dbt compile --target spark && (cat target/compiled/test_project/analyses/current_timestamp.sql >> compiled_current_timestamp.md); \
dbt compile --target databricks && (cat target/compiled/test_project/analyses/current_timestamp.sql >> compiled_current_timestamp.md);
```

### Examine

Visually compare the output within `compiled_current_timestamp.md`

</details>

Let me know if you run into issues setting up the replication case -- happy to help with the setup.

Alternatively, I could screenshare with you so you could see the output on my screen -- I can re-run in my local setup so you can see the effect of any changes.

Note: Databricks maintains the dbt-databricks adapter rather than dbt Labs. It was included because it was easy to do so and as a stand-in for vendor/community adapters.

colin-rogers-dbt:
> One easy way to test the backwards compatibility is just to compare the new SQL to the old SQL. If they are the same, then we are done.

Good idea, I've added an assertion to our functional tests to validate the compiled SQL

tests/adapter/dbt/tests/adapter/utils/test_timestamps.py
```diff
@@ -0,0 +1,51 @@
...
+"""
+
+_MODEL_EXPECTED_SQL = """
+select now() as current_timestamp,
+       (current_timestamp at time zone 'utc')::TIMESTAMP as current_timestamp_in_utc_backcompat,
+       current_timestamp::TIMESTAMP as current_timestamp_backcompat
+"""
+
+
+class TestCurrentTimestamps:
```
dbeatty10, line 18: @gshank could we get your take on something?

The patten I'm used to seeing is like [this](https://github.com/dbt-labs/dbt-core/blob/c2ab2971b073245366f1ccd8a9e1cbe198f18865/tests/adapter/dbt/tests/adapter/grants/test_invalid_grants.py):

```
import pytest
from dbt.tests.adapter.category.base_thing import BaseThing

class BaseSpecificThing(BaseThing):
    <a bunch of code>

class TestSpecificThing(BaseSpecificThing):
    pass
```

Is there any reason not to skip the last two lines and just do the following instead?
```
import pytest
from dbt.tests.adapter.category.base_thing import BaseThing

class TestSpecificThing(BaseThing):
    <a bunch of code>
```

plugins/postgres/dbt/include/postgres/macros/adapters.sql
```diff
@@ -117,21 +117,6 @@
...
-    {{ return(result) }}
-{%- endmacro %}
-
-
-{% macro postgres__snapshot_get_time() -%}
-  {{ current_timestamp() }}::timestamp without time zone
-{%- endmacro %}
-
 {#
   Postgres tables have a maximum length off 63 characters, anything longer is silently truncated.
```
dbeatty10, line 121: Not really your concern @colin-rogers-dbt 'cause your just innocently doing other work in this file, but maybe we can fix this small typo while you're already here?
```suggestion
  Postgres tables have a maximum length of 63 characters, anything longer is silently truncated.
```

tests/adapter/dbt/tests/adapter/utils/test_timestamps.py
colin-rogers-dbt: the only difference this makes is that if you import the base it pytest won't run it until the inheriting class is called because it doesn't have "test" in the class name.

core/dbt/include/global_project/macros/adapters/timestamps.sql
dbeatty10: [Quoting Chesterton](https://www.goodreads.com/quotes/833466-in-the-matter-of-reforming-things-as-distinct-from-deforming
):
> There exists in such a case a certain institution or law; let us say, for the sake of simplicity, a fence or gate erected across a road. The more modern type of reformer goes gaily up to it and says, “I don’t see the use of this; let us clear it away.” To which the more intelligent type of reformer will do well to answer: “If you don’t see the use of it, I certainly won’t let you clear it away. Go away and think. Then, when you can come back and tell me that you do see the use of it, I may allow you to destroy it.”

It isn't saying that something can't or shouldn't be reformed rather _how_ best to approach reforms.

dbeatty10 (commented):
There are two things I'd appreciate an additional set of eyes on before final approval (in order of importance):
1. New implementation for `default__current_timestamp` discussed [here](https://github.com/dbt-labs/dbt-core/pull/5838/files#r983680441)
1. Convention for test class names [here](https://github.com/dbt-labs/dbt-core/pull/5838/files#r983640640) and [here](https://github.com/dbt-labs/dbt-core/pull/5838/files#r984024424)

@jtcohen6 are you willing to give these a look (or suggest someone else)?

---

Other than that, this behavior looks as desired across postgres, snowflake, redshift, bigquery, spark, and databricks.

There are slight differences in the rendered SQL for the `*_backcompat` macros for the latter three vs. `dbt_utils`, but I _don't think_ they make a functional difference.

<details>
<summary>Output for the curious</summary>

&nbsp;

### dbt-postgres

```shell
dbt compile --target postgres
```

```sql
postgres dbt.current_timestamp():
    now()

postgres dbt_utils.current_timestamp(): 
    current_timestamp::TIMESTAMP

postgres dbt.current_timestamp_backcompat(): 
    current_timestamp::TIMESTAMP

postgres dbt_utils.current_timestamp_in_utc(): 
    (current_timestamp at time zone 'utc')::TIMESTAMP

postgres dbt.current_timestamp_in_utc_backcompat(): 
    (current_timestamp at time zone 'utc')::TIMESTAMP

```

&nbsp;&nbsp;

### dbt-snowflake

```shell
dbt compile --target snowflake
```

```sql
snowflake dbt.current_timestamp():
    convert_timezone('UTC', current_timestamp())

snowflake dbt_utils.current_timestamp(): 
    current_timestamp::TIMESTAMP

snowflake dbt.current_timestamp_backcompat(): 
  current_timestamp::TIMESTAMP

snowflake dbt_utils.current_timestamp_in_utc(): 
    convert_timezone('UTC', 
    current_timestamp::TIMESTAMP
)::TIMESTAMP

snowflake dbt.current_timestamp_in_utc_backcompat(): 
  convert_timezone('UTC', 
  current_timestamp::TIMESTAMP
)::TIMESTAMP

```

&nbsp;&nbsp;

### dbt-redshift

```shell
dbt compile --target redshift
```

```sql
redshift dbt.current_timestamp():
    getdate()

redshift dbt_utils.current_timestamp(): 
    getdate()

redshift dbt.current_timestamp_backcompat(): getdate()
redshift dbt_utils.current_timestamp_in_utc(): 
    
    getdate()


redshift dbt.current_timestamp_in_utc_backcompat(): getdate()
```

&nbsp;&nbsp;

### dbt-bigquery

```shell
dbt compile --target bigquery
```

```sql
bigquery dbt.current_timestamp():
    CURRENT_TIMESTAMP()

bigquery dbt_utils.current_timestamp(): 
    current_timestamp

bigquery dbt.current_timestamp_backcompat(): CURRENT_TIMESTAMP()
bigquery dbt_utils.current_timestamp_in_utc(): 
    
    current_timestamp


bigquery dbt.current_timestamp_in_utc_backcompat(): CURRENT_TIMESTAMP()
```

&nbsp;&nbsp;

### dbt-spark

```shell
dbt compile --target spark
```

```sql
spark dbt.current_timestamp():
    current_timestamp()

spark dbt_utils.current_timestamp(): 
    current_timestamp::timestamp

spark dbt.current_timestamp_backcompat(): current_timestamp()
spark dbt_utils.current_timestamp_in_utc(): 
    
    current_timestamp::timestamp


spark dbt.current_timestamp_in_utc_backcompat(): current_timestamp()
```

&nbsp;&nbsp;

### dbt-databricks

```shell
dbt compile --target databricks
```

```sql
databricks dbt.current_timestamp():
    current_timestamp()

databricks dbt_utils.current_timestamp(): 
    current_timestamp::timestamp

databricks dbt.current_timestamp_backcompat(): current_timestamp()
databricks dbt_utils.current_timestamp_in_utc(): 
    
    current_timestamp::timestamp


databricks dbt.current_timestamp_in_utc_backcompat(): current_timestamp()
```

&nbsp;

</details>

tests/adapter/dbt/tests/adapter/utils/test_timestamps.py
gshank: We can't import test classes with "Test" in their name from the adapter zone, because pytest will run both the imported test and the subclass, which really won't work.  In the specific test files there's no reason to do yet another layer of subclassing. Because you haven't said where the example files are located, I'm a little unclear as to which case you're illustrating. tl;dr:  nothing importable in a different adapter should start with "Test".

gshank: Took a look at the code, and you'll need to follow the Base<test_name>. Test(Base<test_name): pass pattern.

colin-rogers-dbt: if you inherit as <module name>.<base test class name> it won't run the test twice

gshank: At this point I think that it's more important to be consistent, so implementers of these tests don't have to do it in multiple ways. If you think we should switch everything over to inheriting from <module name>.<base test class name>, open a ticket for that and it can be discussed.

colin-rogers-dbt: yea that's a fair point, I've changed it to follow the standard pattern

core/dbt/include/global_project/macros/adapters/timestamps.sql
dbeatty10, line 7: ### Unblocking this PR
I don't think your change is needed to accomplish the purposes of this back compatibility PR, so let's leave the existing implementation as-is (but you _can_ move it from the old file to its new file) like so:

```suggestion
{% macro default__current_timestamp() -%}
  {{ exceptions.raise_not_implemented(
    'current_timestamp macro not implemented for adapter '+adapter.type()) }}
{%- endmacro %}
```

---
### Providing a path for innovation and change
You proposed change to `current_timestamp` seems like a reasonable one to propose, and you can raise it in a stand-alone PR.

dbeatty10 (changes requested):
@colin-rogers-dbt Here's what we still need to do prior to approving this:
1. Restore the implementation of `default__current_timestamp`. Rationale provided in the comments.
1. PRs in [dbt-snowflake](https://github.com/dbt-labs/dbt-snowflake/pull/273) and [dbt-redshift](https://github.com/dbt-labs/dbt-redshift/pull/191) need to be approved. There are code suggestions provided to unblock each.
1. @gshank question for you: this PR has a new class named `BaseCurrentTimestamps`. There are PRs to add this as a test case for dbt-snowflake and dbt-redshift, but there are not  PRs that add this test for dbt-bigquery and dbt-spark. Are there scenarios in which we don't add new test classes to all adapters?

Once those are done, I'll re-run [this](https://github.com/dbt-labs/dbt-core/pull/5838#pullrequestreview-1125563229) manual testing across the following adapters as additional confirmation: postgres, snowflake, redshift, bigquery, spark, and databricks.

gshank:
Yes, we won't be implementing all adapter zone tests for all adapters. We did expect that everything in the "basic" directory would be a requirement for all adapters, but tests in the adapter zone outside of that are optional. Not all adapters implement all features.

dbeatty10 (approved):
Looks good across all adapters 🎉 

Also confirmed manually:

<details>
<summary>Output</summary>

&nbsp;

### dbt-postgres

```shell
dbt compile --target postgres
```

```sql
postgres dbt.current_timestamp():
    now()

postgres dbt_utils.current_timestamp(): 
    current_timestamp::TIMESTAMP

postgres dbt.current_timestamp_backcompat(): 
    current_timestamp::TIMESTAMP

postgres dbt_utils.current_timestamp_in_utc(): 
    (current_timestamp at time zone 'utc')::TIMESTAMP

postgres dbt.current_timestamp_in_utc_backcompat(): 
    (current_timestamp at time zone 'utc')::TIMESTAMP

```

&nbsp;&nbsp;

### dbt-snowflake

```shell
dbt compile --target snowflake
```

```sql
snowflake dbt.current_timestamp():
    convert_timezone('UTC', current_timestamp())

snowflake dbt_utils.current_timestamp(): 
    current_timestamp::TIMESTAMP

snowflake dbt.current_timestamp_backcompat(): 
  current_timestamp::TIMESTAMP

snowflake dbt_utils.current_timestamp_in_utc(): 
    convert_timezone('UTC', 
    current_timestamp::TIMESTAMP
)::TIMESTAMP

snowflake dbt.current_timestamp_in_utc_backcompat(): 
  convert_timezone('UTC', 
  current_timestamp::TIMESTAMP
)::TIMESTAMP

```

&nbsp;&nbsp;

### dbt-redshift

```shell
dbt compile --target redshift
```

```sql
redshift dbt.current_timestamp():
    getdate()

redshift dbt_utils.current_timestamp(): 
    getdate()

redshift dbt.current_timestamp_backcompat(): getdate()
redshift dbt_utils.current_timestamp_in_utc(): 
    
    getdate()


redshift dbt.current_timestamp_in_utc_backcompat(): getdate()
```

&nbsp;&nbsp;

### dbt-bigquery

```shell
dbt compile --target bigquery
```

```sql
bigquery dbt.current_timestamp():
    CURRENT_TIMESTAMP()

bigquery dbt_utils.current_timestamp(): 
    current_timestamp

bigquery dbt.current_timestamp_backcompat(): current_timestamp
bigquery dbt_utils.current_timestamp_in_utc(): 
    
    current_timestamp


bigquery dbt.current_timestamp_in_utc_backcompat(): current_timestamp
```

&nbsp;&nbsp;

### dbt-spark

```shell
dbt compile --target spark
```

```sql
spark dbt.current_timestamp():
    CURRENT_TIMESTAMP()

spark dbt_utils.current_timestamp(): 
    current_timestamp::timestamp

spark dbt.current_timestamp_backcompat(): 
    current_timestamp::timestamp

spark dbt_utils.current_timestamp_in_utc(): 
    
    current_timestamp::timestamp


spark dbt.current_timestamp_in_utc_backcompat(): 
    current_timestamp::timestamp

```

&nbsp;&nbsp;

### dbt-databricks

```shell
dbt compile --target databricks
```

```sql
databricks dbt.current_timestamp():
    CURRENT_TIMESTAMP()

databricks dbt_utils.current_timestamp(): 
    current_timestamp::timestamp

databricks dbt.current_timestamp_backcompat(): 
    current_timestamp::timestamp

databricks dbt_utils.current_timestamp_in_utc(): 
    
    current_timestamp::timestamp


databricks dbt.current_timestamp_in_utc_backcompat(): 
    current_timestamp::timestamp

```

&nbsp;

</details>

PR #5839, opened 2022-09-14
remove source quoting setting in adapter tests

### LINKED ISSUE #5836: [Bug] Adapter doc tests override adapter quoting settings

### Current Behavior

The adapter tests for references in generated documentation define quoting policies for sources. By defining these policies, they override the default quoting policy set by the adapter.

### Expected Behavior

Use the default adapter quoting policies in the adapter tests.

### Steps To Reproduce

Implement a test based on `BaseDocsGenReferences`

### Relevant log output

_No response_

### Environment

```markdown
dbt-tests-adapter==1.2.1
```

### Additional Context

It makes sense for users of the adapters to specify their own quoting policies, but in this case every adapter maintainer would have to remove these quoting policies so that their tests would succeed. I think it doesn't really make sense to have these as part of the tests.

E.g. in my case one of the databases we run our tests on contains a `-` in its name. The adapter's quoting policy sets all database names to be quoted, but this is overridden in the source for this test so the test failed on that particular database.

sdebruyn:
resolves #5836 

### Description

As discussed in #5836, the adapter tests should not override the quoting policies defined in the adapters. I removed all quoting policies from this the docs generate test.

### Checklist

### DIFF
--- a/.changes/unreleased/Under the Hood-20220914-193933.yaml
+++ b/.changes/unreleased/Under the Hood-20220914-193933.yaml
@@ -1,0 +1,7 @@
+kind: Under the Hood
+body: remove source quoting setting in adapter tests
+time: 2022-09-14T19:39:33.688385+02:00
+custom:
+  Author: sdebruyn
+  Issue: "5836"
+  PR: "5839"
--- a/tests/adapter/dbt/tests/adapter/basic/test_docs_generate.py
+++ b/tests/adapter/dbt/tests/adapter/basic/test_docs_generate.py
@@ -432,9 +432,7 @@ class BaseDocsGenerate(BaseGenerateProject):
     # Test "--no-compile" flag works and produces no manifest.json
     def test_run_and_generate_no_compile(self, project, expected_catalog):
         start_time = run_and_generate(project, ["--no-compile"])
-        assert not os.path.exists(
-            os.path.join(project.project_root, "target", "manifest.json")
-        )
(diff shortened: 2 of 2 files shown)


sdebruyn:
I just had the same build failure in the dbt-sqlserver GH Actions as [this one(https://github.com/dbt-labs/dbt-core/actions/runs/3054923370/jobs/4927427142), something we can do about it or just ignore and retry?

ChenyuLInx:
It looks like there's some issue regarding pulling down `setup-tools`, maybe just retry in a bit.

As for the change, they look good to me! @gshank what was the reason those quoting setting were added in the beginning, do we expect this to break tests in other adapter repos?

gshank:
Setting the quoting was copied from the earlier integration tests. It's quite possible that it was originally there because we ran multiple adapter tests on the same test suite? I think we might want to check what this does to the tests in Redshift, BigQuery, and Snowflake before merging though. If it breaks those tests we'd have to quickly do PRs to fix.

ChenyuLInx:
Created a [test pr](https://github.com/dbt-labs/dbt-bigquery/pull/314) to test this change in dbt-bigquery repo

ChenyuLInx (approved):

ChenyuLInx:
The tests in test PR works as expected, so I think this is good to go. @sdebruyn  Thanks for contributing it!!

PR #5841, opened 2022-09-14
Add name validation for metrics

### LINKED ISSUE #5456: [Feature] dbt-core raising compile errors for invalid metric names (and potentially invalid model names)

### Describe the Feature

Continuing this conversation from [[CT-65] [Bug] Metrics' names shouldn’t be allowed to contain spaces](https://github.com/dbt-labs/dbt-core/issues/4572) after some more recent work on the metrics package. @joellabes had raised some really great points around additional constraints for metric names that weren't fixed when that issue was closed. Namely, leading numbers and periods in metric names that would cause column aliasing to fail. 

If I understand the issue correctly, there are two paths forward that I can see:

**1. Introduce additional naming constraints into dbt parsing logic based around db column naming**
   - Pros: 
     - Users will immediately know at compile time whether or not their metric names are valid
     - We can confidently use metric names as an alias in the metric package
   - Cons:
     - Not a small amount of work. To my knowledge we don't have these kind of db column/table naming constraints documented anywhere so we'd have to research those for the big 4, develop some kind of framework that can be used by adapter maintainers, and then implement the new conditions into the parsing logic.
 
**2. Accept run errors around column names** 
   - Pros:
     - We don't need to add additional functionality to dbt-core 
     - We can get away with clear documentation around the metric.
   - Cons: 
     - The error message users get will not help inform them around what the real error is. They'll get a column name not valid error but unless we have clear documentation around this error, it won't point them towards the fix.

Based on the amount of work laid out in the Con section of option 1, I am personally of the opinion that we move forward with option 2 until there is bandwidth or interest in the core team to tackle something like this. Granted, I could be totally off base around how much work it is and that could change the math here.

^See option 2. 

### Who will this benefit?

This will benefit anyone using the metrics package and potentially more if the scope is expanded to model names.

Potentially! Depends on the complexity.

### Anything else?

Nope! For those who read the entire issue, thank you - please enjoy this [very cute video of a bunny!](https://www.youtube.com/watch?v=dQw4w9WgXcQ)

emmyoop:
resolves #5456

### Description

Adds name validation rules:
* Metric name characters must be alphanumeric or underscore
* Metric name must begin with an letter (this allows us to get around the BQ naming prefixes which all begin with _ )
* Metric names must be 126 characters or less (Redshift constraint. BQ and Snowflake are longer)

Adding exposure name validation in #5844 

### Checklist

### DIFF
--- a/core/dbt/contracts/graph/unparsed.py
+++ b/core/dbt/contracts/graph/unparsed.py
@@ -489,8 +489,10 @@ class UnparsedMetric(dbtClassMixin, Replaceable):
             errors = []
             if " " in data["name"]:
                 errors.append("cannot contain spaces")
-            if len(data["name"]) > 126:
-                errors.append("cannot contain more than 126 characters")
+            # This handles failing queries due to too long metric names.
+            # It only occurs in BigQuery and Snowflake (Postgres/Redshift truncate)
+            if len(data["name"]) > 250:
+                errors.append("cannot contain more than 250 characters")
             if not (re.match(r"^[A-Za-z]", data["name"])):
                 errors.append("must begin with a letter")
             if not (re.match(r"[\w-]+$", data["name"])):
@@ -498,7 +500,7 @@ class UnparsedMetric(dbtClassMixin, Replaceable):
 
             if errors:
                 raise ParsingException(
-                    f"Metrics name '{data['name']}' is invalid.  It {', '.join(e for e in errors)}"
+                    f"The metric name '{data['name']}' is invalid.  It {', '.join(e for e in errors)}"
                 )
 
         if data.get("calculation_method") == "expression":
--- a/core/dbt/contracts/graph/unparsed.py
+++ b/core/dbt/contracts/graph/unparsed.py
@@ -1,3 +1,4 @@
+import re
 from dbt.node_types import NodeType
 from dbt.contracts.util import AdditionalPropertiesMixin, Mergeable, Replaceable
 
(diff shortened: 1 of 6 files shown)


core/dbt/contracts/graph/unparsed.py
```diff
@@ -484,8 +485,21 @@ class UnparsedMetric(dbtClassMixin, Replaceable):
...
+                errors.append("cannot contain more than 126 characters")
+            if not (re.match(r"^[A-Za-z]", data["name"])):
+                errors.append("must begin with a letter")
+            if not (re.match(r"[\w-]+$", data["name"])):
+                errors.append("must contain only letters, numbers and underscores")
+
+            if errors:
+                raise ParsingException(
+                    f"Metrics name '{data['name']}' is invalid.  It {', '.join(e for e in errors)}"
+                )
```
emmyoop, line 502: @callum-mcdata would love some help with error messaging here!  I went with concatenating all the errors together but open to other solutions.

emmyoop:
Sounds reasonable to me @callum-mcdata.  That's a simple change.  I'll defer to @jtcohen6 for the final decision on length.

core/dbt/contracts/graph/unparsed.py
callum-mcdata: Love the framework here! I'd only tweak it slightly to `The metric name` as opposed to `Metrics name`. Apart from that I think it is does a great job of describing the error and I LOVE that it provides all the errors in one failed compile instead of one at a time like I might have novicely implemented it

jtcohen6:
Fine by me! 250 chars feels like plenty. If you can tweet it, you can (probably) make it into a metric

callum-mcdata:
@jtcohen6 huh. Well .... I hadn't thought about that. Is it a fair answer to punt on that and say we'll solve that problem when we get to it? 😅  to me figuring out the pattern to account for that feels like we're optimizing for an edge case. 

If that problem did arise then I think your idea around fixing it in the package makes sense - it's a usage pattern more than it is a definition pattern so allowing that kind of naming convention **should** be supported in core.

gshank (approved):

PR #5844, opened 2022-09-14
Validate exposure names and add label

### LINKED ISSUE #5606: [CT-1184] [Feature] Exposures: stricter validation for `name`, support custom `label`

### Describe the feature

Currently, it's possible to define an exposure like:
```yml
exposures:
  - name: My Cool Exposure
     ...
```

A surprising number of things will keep working:
```
$ dbt list --resource-type exposure
exposure:my_project.My Cool Exposure
```

The problem is that it's not _guaranteed_ to work everywhere. In particular, the `dbt-docs` site DAG selection does not handle exposure names with spaces (https://github.com/dbt-labs/dbt-core/issues/2970).

Rather than fix it everywhere, I'd prefer to adopt the same approach we'll be taking for metrics (https://github.com/dbt-labs/dbt-core/issues/5456):
- Have stricter parse-time validation on the `name` property of exposures: snake-case, alphanumeric characters, _maybe_ periods (since this is supported for models)
- Support a `label` property that can be used in external metadata, and contain whatever characters the heart desires

```yml
exposures:
  - name: my_cool_exposures
    label: My Cool Exposure
     ...
```

- Continue to support whatever exposure names users want, and they only find out much later
- Add tons of strict validation _without_ supporting a custom human-friendly `label`

I don't like either very much!

### Who will this benefit?

- Users of exposures
- Programmatic integrations with exposures (e.g. https://hightouch.com/blog/introducing-dbt-exposures-for-hightouch/)

Perhaps :)

### Anything else?

_No response_

emmyoop:
resolves #5606 

### Description

Adds label and name validation for exposures.  Kept as a separate PR to keep the label addition with the name validation.

**Open Question**
Exactly what do we want to validate as a name?  I just copied over from metrics but that's likely incorrect.

### Checklist

### DIFF
--- a/.changes/unreleased/Under the Hood-20220914-150058.yaml
+++ b/.changes/unreleased/Under the Hood-20220914-150058.yaml
@@ -1,0 +1,7 @@
+kind: Under the Hood
+body: Validate exposure name and add label
+time: 2022-09-14T15:00:58.982982-05:00
+custom:
+  Author: emmyoop
+  Issue: "5606"
+  PR: "5844"
--- a/core/dbt/contracts/graph/unparsed.py
+++ b/core/dbt/contracts/graph/unparsed.py
@@ -1,5 +1,6 @@
 import re
 
+from dbt import deprecations
 from dbt.node_types import NodeType
 from dbt.contracts.util import (
     AdditionalPropertiesMixin,
@@ -447,22 +448,10 @@ class UnparsedExposure(dbtClassMixin, Replaceable):
     @classmethod
     def validate(cls, data):
         super(UnparsedExposure, cls).validate(data)
-        # TODO: what do we actually want to rules to be?
         if "name" in data:
-            errors = []
-            if " " in data["name"]:
-                errors.append("cannot contain spaces")
-            if len(data["name"]) > 126:
-                errors.append("cannot contain more than 126 characters")
-            if not (re.match(r"^[A-Za-z]", data["name"])):
-                errors.append("must begin with a letter")
+            # name can only contain alphanumeric chars and underscores
             if not (re.match(r"[\w-]+$", data["name"])):
-                errors.append("must contain only letters, numbers and underscores")
-
-            if errors:
-                raise ParsingException(
-                    f"Exposure name '{data['name']}' is invalid.  It {', '.join(e for e in errors)}"
-                )
+                deprecations.warn("exposure-name", exposure=data["name"])
 
 
 @dataclass
--- a/core/dbt/contracts/graph/unparsed.py
+++ b/core/dbt/contracts/graph/unparsed.py
@@ -433,10 +433,10 @@ class ExposureOwner(dbtClassMixin, Replaceable):
 @dataclass
 class UnparsedExposure(dbtClassMixin, Replaceable):
     name: str
-    label: str
     type: ExposureType
     owner: ExposureOwner
     description: str = ""
+    label: Optional[str] = None
     maturity: Optional[MaturityType] = None
     meta: Dict[str, Any] = field(default_factory=dict)
     tags: List[str] = field(default_factory=list)
--- a/core/dbt/contracts/graph/unparsed.py
+++ b/core/dbt/contracts/graph/unparsed.py
@@ -1,4 +1,5 @@
 import re
+
 from dbt.node_types import NodeType
 from dbt.contracts.util import (
     AdditionalPropertiesMixin,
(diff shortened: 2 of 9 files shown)


core/dbt/contracts/graph/unparsed.py
```diff
@@ -429,13 +431,34 @@ class UnparsedExposure(dbtClassMixin, Replaceable):
...
+            errors = []
+            if " " in data["name"]:
+                errors.append("cannot contain spaces")
+            if len(data["name"]) > 126:
+                errors.append("cannot contain more than 126 characters")
+            if not (re.match(r"^[A-Za-z]", data["name"])):
+                errors.append("must begin with a letter")
+            if not (re.match(r"[\w-]+$", data["name"])):
+                errors.append("must contain only letters, numbers and underscores")
+
```
emmyoop, line 456: @jtcohen6 this is just a copy of the metrics name validation.  It's overkill here I believe.  The issue wasn't super clean on what to validate on so I just copied it over to get something up.  Let me know what rules should apply!

.changes/unreleased/Under the Hood-20220914-150058.yaml
```diff
@@ -0,0 +1,7 @@
+kind: Under the Hood
```
emmyoop, line 1: Not sure where this falls in terms of changelog type?

core/dbt/contracts/graph/unparsed.py
```diff
@@ -429,13 +431,34 @@ class UnparsedExposure(dbtClassMixin, Replaceable):
...
+            if errors:
+                raise ParsingException(
+                    f"Exposure name '{data['name']}' is invalid.  It {', '.join(e for e in errors)}"
+                )
```
jtcohen6, line 460: Could we actually make this `deprecations.warn` instead of `ParsingException`? I think there are many users (including [this integration](https://hightouch.com/blog/introducing-dbt-exposures-for-hightouch/)) who will be in violation of the new rule.

I can provide copy for the deprecation message, but the big idea is: "Hey! We just added support for a human-friendly `label` propertyon exposures. We see that {exposure} has some disallowed characters in its name — why not turn that into label, and then come up with a simpler name?"

jtcohen6 (commented):
Thanks for this @emmyoop! Sorry I missed your questions the other

As a follow-up to this, let's also start using `label` instead of `name` for identifying exposures in `dbt-docs`, similar to the changes in https://github.com/dbt-labs/dbt-docs/pull/285

.changes/unreleased/Under the Hood-20220914-150058.yaml
emmyoop: Since we're deprecating instead of erroring now, this won't be breaking.  I'll leave as `Under The Hood` unless you disagree.

core/dbt/deprecations.py
```diff
@@ -35,6 +35,17 @@ def show(self, *args, **kwargs) -> None:
...
 
+class ExposureNameDeprecation(DBTDeprecation):
+    _name = "exposure-name"
+    # todo: word this better - it's bad.  Also do we want to set a timeline for when
+    # this will become an error?
+    _description = """\
+    Exposure names may not contain spaces and can only contain letters, numbers and underscores.
+    {exposure} does not follow this pattern.  Please use the label attribute for more complex
+    names.
+    """
```
emmyoop, line 46: @jtcohen6 please provide the promised wording 😄

jtcohen6: How about this:
```python
    _description = """\
    Starting in v1.3, the 'name' property of exposures should contain only contain letters, numbers, and underscores.
    Exposures support a new property, 'label', that can contain spaces, capital letters, and special characters.
    {exposure} does not follow this pattern.
    Please update its 'name', and use the 'label' property for its human-friendly title.
    This will raise an exception in a future version of 'dbt-core'.
    """
```
In the case of metrics, we'd explicitly documented the metric spec as "experimental," and reserved the right to make breaking changes. In that context, it's nice for us to provide backward compatibility for one minor version, and reasonable to expect a quick turnaround (breaking change by v1.4).

Exposures have been around for longer, and without an "experimental" denotation — so I don't think we should commit to a quick turnaround for actual deprecation. The soonest I could see doing this is if we significantly expand the capabilities of exposures, as part of a potential initiative next year. Otherwise, in dbt v2.0.

Given that, I also don't feel the need to open an issue to track this. We'll have an opportunity to revisit this either when we reinvest in exposures, or when we revisit all of our deprecations ahead of v2 (just as we did ahead of v1).

gshank (approved):

PR #5850, opened 2022-09-15
fix multiple dbt.config.get in python model

### LINKED ISSUE #5846: [CT-1189] [Bug] Missing configs when multiple dbt.config.get in Python model

### Current Behavior

Not all configs show up in generated code (`config_dict`) for Python model with multiple `dbt.config.get()` calls

### Expected Behavior

All configs are included in `config_dict`

### Steps To Reproduce

Define a model like:
```python
def model(dbt, session):

    PROJECT_ID = dbt.config.get("POSTHOG_PROJECT_ID")
    POSTHOG_TOKEN = dbt.config.get("POSTHOG_TOKEN")

    df = dbt.ref("orders")

    return df
```
With configs like:
```yml
version: 2
models:
  - name: something
    config:
      PROJECT_ID: something
      POSTHOG_TOKEN: something_else
```

Run `dbt compile`. See compiled output like:
```
def model(dbt, session):

    PROJECT_ID = dbt.config.get("POSTHOG_PROJECT_ID")
    POSTHOG_TOKEN = dbt.config.get("POSTHOG_TOKEN")

    df = dbt.ref("orders")

    return df


# This part is user provided model code
# you will need to copy the next section to run the code
# COMMAND ----------
# this part is dbt logic for get ref work, do not modify

def ref(*args,dbt_load_df_function):
    refs = {"orders": "ad_hoc.dbt_jcohen.orders"}
    key = ".".join(args)
    return dbt_load_df_function(refs[key])


def source(*args, dbt_load_df_function):
    sources = {}
    key = ".".join(args)
    return dbt_load_df_function(sources[key])


config_dict = {'POSTHOG_TOKEN': 'something_else'}


class config:
    def __init__(self, *args, **kwargs):
        pass

    @staticmethod
    def get(key):
        return config_dict.get(key)

class this:
    """dbt.this() or dbt.this.identifier"""
    database = 'ad_hoc'
    schema = 'dbt_jcohen'
    identifier = 'something'
    def __repr__(self):
        return 'ad_hoc.dbt_jcohen.something'


class dbtObj:
    def __init__(self, load_df_function) -> None:
        self.source = lambda *args: source(*args, dbt_load_df_function=load_df_function)
        self.ref = lambda *args: ref(*args, dbt_load_df_function=load_df_function)
        self.config = config
        self.this = this()
        self.is_incremental = False

# COMMAND ----------

# To run this in snowsight, you need to select entry point to be main
# And you may have to modify the return type to text to get the result back
# def main(session):
#     dbt = dbtObj(session.table)
#     df = model(dbt, session)
#     return df.collect()

# to run this in local notebook, you need to create a session following examples https://github.com/Snowflake-Labs/sfguide-getting-started-snowpark-python
# then you can do the following to run model
# dbt = dbtObj(session.table)
# df = model(dbt, session)
```

Namely:
```
config_dict = {'POSTHOG_TOKEN': 'something_else'}
```

### Relevant log output

_No response_

### Environment

```markdown
- OS:
- Python:
- dbt: 1.3.0-b2 (I think latest `main`)
```

snowflake

### Additional Context

_No response_

ChenyuLInx:
resolves #5846

### Description
Fix for config missing when multiple are specified

### Checklist

### DIFF
--- a/core/dbt/include/global_project/macros/python_model/python.sql
+++ b/core/dbt/include/global_project/macros/python_model/python.sql
@@ -30,8 +30,8 @@ def source(*args, dbt_load_df_function):
 
 {% macro build_config_dict(model) %}
     {%- set config_dict = {} -%}
-    {%- for key in model.config.utilized -%}
-        {# TODO: weird type testing with enum, would be much easier to write this logic in Python! #}
+    {%- for key in model.config.config_get_keys -%}
+        {# weird type testing with enum, would be much easier to write this logic in Python! #}
         {%- if key == 'language' -%}
           {%- set value = 'python' -%}
         {%- endif -%}
@@ -56,8 +56,8 @@ class config:
         pass
 
     @staticmethod
-    def get(key):
-        return config_dict.get(key)
+    def get(key, default=None):
+        return config_dict.get(key, default=None)
 
 class this:
     """dbt.this() or dbt.this.identifier"""
--- a/core/dbt/include/global_project/macros/python_model/python.sql
+++ b/core/dbt/include/global_project/macros/python_model/python.sql
@@ -57,7 +57,7 @@ class config:
 
     @staticmethod
     def get(key, default=None):
-        return config_dict.get(key, default=None)
+        return config_dict.get(key, default)
 
 class this:
     """dbt.this() or dbt.this.identifier"""
--- a/core/dbt/include/global_project/macros/python_model/python.sql
+++ b/core/dbt/include/global_project/macros/python_model/python.sql
@@ -30,7 +30,7 @@ def source(*args, dbt_load_df_function):
 
 {% macro build_config_dict(model) %}
     {%- set config_dict = {} -%}
-    {%- for key in model.config.config_get_keys -%}
+    {%- for key in model.config.config_keys_used -%}
         {# weird type testing with enum, would be much easier to write this logic in Python! #}
         {%- if key == 'language' -%}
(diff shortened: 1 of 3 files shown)


core/dbt/include/global_project/macros/python_model/python.sql
```diff
@@ -56,8 +56,8 @@ class config:
         pass
 
     @staticmethod
-    def get(key):
-        return config_dict.get(key)
+    def get(key, default=None):
+        return config_dict.get(key, default=None)
```
stu-k, line 60: This python looks slightly off, can you do `default=None` in a method invocation? I would think this should be
```suggestion
    def get(key, default=None):
        return config_dict.get(key, default)
```
since `default` already defaults to `None`.

core/dbt/parser/models.py
```diff
@@ -205,16 +205,16 @@ def parse_python_model(self, node, config, context):
 
         dbtParser = PythonParseVisitor(node)
         dbtParser.visit(tree)
-
+        config_get_keys = []
```
stu-k, line 208: Not a blocker, but `config_get_keys` is hard to reason about for me. `config_keys_used` is a little easier to understand, if that is what this is meant to hold.

ChenyuLInx: Thanks!! It's def easier to understand! Updated

stu-k (approved):
Comments addressed, nothing else stands out 👍

PR #5851, opened 2022-09-16
Adding perf testing GHA

leahwicz:
resolves https://github.com/dbt-labs/dbt-core/issues/4021

### Description
This PR creates an Action that automates running our perf regression testing framework after a release. It also reports those results and creates a PR for them so we can continue to keep track of the numbers over time

I pulled this work out of this draft PR https://github.com/dbt-labs/dbt-core/pull/4872 and just targeted the single action to gather results after a release is made (the other action is out of scope for this)

Example PRs created here
* https://github.com/dbt-labs/dbt-core/pull/5865
* https://github.com/dbt-labs/dbt-core/pull/5866

### Checklist

### DIFF
--- a/.github/workflows/model_performance.yml
+++ b/.github/workflows/model_performance.yml
@@ -227,6 +227,12 @@ jobs:
         uses: actions/checkout@v2
         with:
           ref: ${{ needs.latest-runner.outputs.release_branch }}
+          
+      - name: Create PR branch
+        run: |
+          git checkout -b performance-bot/release_${{needs.latest-runner.outputs.release_id}}_${{GITHUB.RUN_ID}}
+          git push origin performance-bot/release_${{needs.latest-runner.outputs.release_id}}_${{GITHUB.RUN_ID}}
+          git branch --set-upstream-to=origin/performance-bot/release_${{needs.latest-runner.outputs.release_id}}_${{GITHUB.RUN_ID}} performance-bot/release_${{needs.latest-runner.outputs.release_id}}_${{GITHUB.RUN_ID}}
 
       - name: '[DEBUG] ls baselines before artifact download'
         run: ls -R performance/baselines/
@@ -239,16 +245,15 @@ jobs:
       - name: '[DEBUG] ls baselines after artifact download'
         run: ls -R performance/baselines/
 
-      - name: Make Branch
-        uses: EndBug/add-and-commit@v8
+      - name: Commit baseline
+        uses: EndBug/add-and-commit@v9
         with:
           add: 'performance/baselines/*'
           author_name: 'Github Build Bot'
           author_email: 'buildbot@fishtownanalytics.com'
           message: 'adding performance baseline for ${{ needs.latest-runner.outputs.release_id }}'
-          new_branch: 'performance-bot/release_${{needs.latest-runner.outputs.release_id}}_${{GITHUB.RUN_ID}}'
-          push: true
-
+          push: 'origin origin/performance-bot/release_${{needs.latest-runner.outputs.release_id}}_${{GITHUB.RUN_ID}}'
+          
       - name: Create Pull Request
         uses: peter-evans/create-pull-request@v3
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/model_performance.yml
```diff
@@ -0,0 +1,271 @@
...
+        run: mv performance/runner/target/release/runner performance/app
+
+      - name: Change Runner Permissions
+        run: chmod +x ./performance/app
+
+      - name: '[DEBUG] ls baseline directory before run'
+        run: ls -R performance/baselines/
+
+      # `${{ github.workspace }}` is used to pass the absolute path
+      # TODO CHANGE NUMBER OF RUNS BEFORE MERGING
```
leahwicz, line 194: @nathaniel-may for number of runs- would I just change `-n 2` to something greater? 10? What were you thinking here?

emmyoop (commented):
Why do we only run against `main`?  When we release `1.2.7` it could have significantly diverged from `main`.

leahwicz:
> Why do we only run against main? 
@nathaniel-may  would you want to field this question?

.github/workflows/model_performance.yml
nathaniel-may: yeah. maybe 20? depends on how much time we want to spend. no less than 10.

nathaniel-may:
the model action should be pulling the action code from `main` always, but checking out dbt code from whatever branch we want to model.

.github/workflows/model_performance.yml
emmyoop: Would we ever want to change this number?  Ie. use different number of runs when running manually?  Should it be an input?

.github/workflows/model_performance.yml
```diff
@@ -0,0 +1,271 @@
...
+
+      - name: Create Pull Request
+        uses: peter-evans/create-pull-request@v3
+        with:
+          author: 'Github Build Bot <buildbot@fishtownanalytics.com>'
+          base: ${{ matrix.base-branch }}
+          title: 'Adding performance modeling for ${{needs.latest-runner.outputs.release_id}} to ${{ matrix.base-branch }}'
+          branch: '${{ matrix.target-branch }}'
+          labels: |
+            Skip Changelog
```
emmyoop, line 271: Should we also add the performance label so these PRs are easy to find in history?

emmyoop (changes requested):

.github/workflows/model_performance.yml
leahwicz: I don't think we would want that because we want to keep consistency between all results. It wouldn't be good to compare results from one version with 10 runs vs another version with 20 runs

emmyoop: That makes sense.  The only reason I would recommend it is that taking it as an input would allow for quicker testing of the action.

.github/workflows/model_performance.yml
```diff
@@ -0,0 +1,274 @@
...
+
+      - name: Change Runner Permissions
+        run: chmod +x ./performance/app
+
+      - name: '[DEBUG] ls baseline directory before run'
+        run: ls -R performance/baselines/
+
+      # `${{ github.workspace }}` is used to pass the absolute path
+      # TODO CHANGE NUMBER OF RUNS BEFORE MERGING
+      # TODO this isn't putting the baseline in the right directory. it's putting it one level up.
```
leahwicz, line 199: @nathaniel-may do you remember what this comment meant? It looks like the PR shows the file is in the right directory

leahwicz:
@emmyoop @nathaniel-may could I get a look at this again please? I do want to get these times documented for all the 1.0+ versions as we get further away from 1.0

.github/workflows/model_performance.yml
```diff
@@ -0,0 +1,278 @@
...
+        with:
+          ref: main
+
+      # collect all the variables that need to be used in subsequent jobs
+      - name: Set Variables
+        id: variables
+        run: |
+          # create a cache key that will be used in the next job. without this the
+          # next job would have to checkout from main an hash the files itself.
+          echo "::set-output name=cache_key::${{ runner.os }}-${{ hashFiles('performance/runner/Cargo.toml')}}-${{ hashFiles('performance/runner/src/*') }}"
```
emmyoop, line 70: `set-output` is being [deprecated](https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/).  All instances should use the new formats

```
echo "{name}={value}" >> $GITHUB_OUTPUT
```

emmyoop (changes requested):
Just updating to use the new env variable for outputs everywhere but nothing else jumps out.

emmyoop (approved):
Looks good to me.  There are several "remove/change before merging" comments to take care of before merging.

PR #5858, opened 2022-09-16
Ct 1191 event history cleanup

### LINKED ISSUE #5848: [CT-1191] Memory growth: stop saving Exceptions in logging events
Memory in a long-running server was continually growing because when an error occurred the node was saved in an Exception which was saved in a logging event which was stored in the logging event history. The references to the nodes meant that old manifests could not be garbage collected.

We should change Exceptions in logging events to strings, and add 'event_buffer_size' to UserConfig.

gshank:
resolves #5848

### Description

Some logging events were holding references to nodes in exceptions, which prevented garbage collection. Change Exceptions in events to strings. Enable setting event_buffer_size to 0.  Provide 'reset_event_history' and refactor code to add to event history. Add 'event_buffer_size' to user config. Stop providing an "end of buffer" event.

### Checklist


core/dbt/events/functions.py
```diff
@@ -22,24 +22,16 @@
...
 
-global LOG_VERSION
 LOG_VERSION = 2
-
-# create the global event history buffer with the default max size (10k)
-# python 3.7 doesn't support type hints on globals, but mypy requires them. hence the ignore.
-# TODO the flags module has not yet been resolved when this is created
-global EVENT_HISTORY
-EVENT_HISTORY = deque(maxlen=flags.EVENT_BUFFER_SIZE)  # type: ignore
+EVENT_HISTORY = None
```
gshank, line 26: Instead of creating the deque here, I check and create it when calling 'fire_event'.

core/dbt/events/functions.py
```diff
@@ -52,10 +44,6 @@
 
 
 def setup_event_logger(log_path, level_override=None):
-    # flags have been resolved, and log_path is known
```
gshank, line 55: I've removed setting the event_history here because of discussions we have about events not lasting for multiple calls. Does this make sense? Would it be better to reset it here too?

core/dbt/events/functions.py
```diff
@@ -344,3 +325,17 @@ def get_ts_rfc3339() -> str:
...
     return ts_rfc3339
+
+
+def add_to_event_history(event):
+    if flags.EVENT_BUFFER_SIZE == 0:
+        return
+    global EVENT_HISTORY
+    if EVENT_HISTORY is None:
+        reset_event_history()
+    EVENT_HISTORY.append(event)
```
gshank, line 336: I stopped emitting the EventBufferFull event here because for small deque sizes we'd get strange behavior. What do we do if the deque is set to 1? to 2? Event at 10, we'd get a really high numberof EventBufferFull messages. I'm also not sure what the utility of the EventBufferFull message is.

Is this the right way to go?

core/dbt/events/functions.py
```diff
@@ -344,3 +325,17 @@ def get_ts_rfc3339() -> str:
     ts = get_ts()
     ts_rfc3339 = ts.strftime("%Y-%m-%dT%H:%M:%S.%fZ")
     return ts_rfc3339
+
+
+def add_to_event_history(event):
+    if flags.EVENT_BUFFER_SIZE == 0:
```
gshank, line 331: We should be able to turn the event buffer off by setting to 0, which we couldn't do before.

core/dbt/events/functions.py
jtcohen6: Yes! Makes sense that `EVENT_HISTORY` wants to last for the duration of the application, rather than just the duration of the invocation

core/dbt/events/functions.py
jtcohen6: Agree that this message is oftentimes confusing / misleading for users. I do get why we added it, though. Let's keep it if EventBuffer is >=10000?

gshank: Done.

PR #5868, opened 2022-09-16
Disabled Models in schema files

### LINKED ISSUE #3992: [CT-232] Setting enabled to False in schema config does not move model to disabled
In the model node, NodeConfig.enabled equals False, but because the node isn't moved into the disabled dictionary, it doesn't have the correct error messages. Instead of the node is disabled messages, nodes referring to the disabled node cause an error "Model ... depends on a node named ... which was not found".

emmyoop:
resolves #3992

### Description

- throw correct exception for models disabled in schema files
- fix bug when chained disabled models incorrectly throw exception
- Add tests

### Checklist

### DIFF
--- a/core/dbt/parser/manifest.py
+++ b/core/dbt/parser/manifest.py
@@ -370,6 +370,7 @@ class ManifestLoader:
 
             self._perf_info.parse_project_elapsed = time.perf_counter() - start_parse_projects
 
+            # breakpoint()
             # patch_sources converts the UnparsedSourceDefinitions in the
             # Manifest.sources to ParsedSourceDefinition via 'patch_source'
             # in SourcePatcher
@@ -463,11 +464,13 @@ class ManifestLoader:
             for file_id in parser_files[parser_name]:
                 block = FileBlock(self.manifest.files[file_id])
                 if isinstance(parser, SchemaParser):
+                    # breakpoint()
                     assert isinstance(block.file, SchemaSourceFile)
                     if self.partially_parsing:
                         dct = block.file.pp_dict
                     else:
                         dct = block.file.dict_from_yaml
+                    # this is where the schema file gets parsed
                     parser.parse_file(block, dct=dct)
                     # Came out of here with UnpatchedSourceDefinition containing configs at the source level
                     # and not configs at the table level (as expected)
@@ -829,6 +832,7 @@ class ManifestLoader:
         for node in self.manifest.nodes.values():
             if node.created_at < self.started_at:
                 continue
+            # breakpoint()
             _process_refs_for_node(self.manifest, current_project, node)
         for exposure in self.manifest.exposures.values():
             if exposure.created_at < self.started_at:
--- a/core/dbt/parser/manifest.py
+++ b/core/dbt/parser/manifest.py
@@ -367,10 +367,8 @@ class ManifestLoader:
                 self.parse_project(
                     project, project_parser_files[project.project_name], parser_types
                 )
-            # breakpoint()
             self._perf_info.parse_project_elapsed = time.perf_counter() - start_parse_projects
 
-            # breakpoint()
             # patch_sources converts the UnparsedSourceDefinitions in the
             # Manifest.sources to ParsedSourceDefinition via 'patch_source'
             # in SourcePatcher
@@ -460,13 +458,10 @@ class ManifestLoader:
             parser_start_timer = time.perf_counter()
 
             # Parse the project files for this parser
-            # breakpoint()
             parser: Parser = parser_cls(project, self.manifest, self.root_project)
             for file_id in parser_files[parser_name]:
                 block = FileBlock(self.manifest.files[file_id])
-                # breakpoint()
                 if isinstance(parser, SchemaParser):
-                    # breakpoint()
                     assert isinstance(block.file, SchemaSourceFile)
                     if self.partially_parsing:
                         dct = block.file.pp_dict
@@ -477,7 +472,6 @@ class ManifestLoader:
                     # Came out of here with UnpatchedSourceDefinition containing configs at the source level
                     # and not configs at the table level (as expected)
                 else:
-                    # breakpoint()
                     parser.parse_file(block)
                 project_parsed_path_count += 1
 
@@ -1255,7 +1249,6 @@ def _process_refs_for_node(manifest: Manifest, current_project: str, node: Manif
             node.package_name,
         )
 
-        # breakpoint()
         if target_model is None or isinstance(target_model, Disabled):
             # This may raise. Even if it doesn't, we don't want to add
             # this node to the graph b/c there is no destination node
--- a/core/dbt/parser/manifest.py
+++ b/core/dbt/parser/manifest.py
@@ -389,6 +389,7 @@ class ManifestLoader:
             # These check the created_at time on the nodes to
             # determine whether they need processing.
             start_process = time.perf_counter()
+            self.process_nodes(self.root_project.project_name)
             self.process_sources(self.root_project.project_name)
             self.process_refs(self.root_project.project_name)
             self.process_docs(self.root_project)
@@ -926,6 +927,15 @@ class ManifestLoader:
             if exposure.created_at < self.started_at:
                 continue
             _process_sources_for_exposure(self.manifest, current_project, exposure)
+
+    def process_nodes(self, current_project: str):
+        # TODO: process that everything is in the right place
+        pass
+        # for node in self.manifest.nodes.values():
+        #     breakpoint()
+
+        # for node in self.manifest.disabled.values():
+        #     breakpoint()
 
 
 def invalid_ref_fail_unless_test(node, target_model_name, target_model_package, disabled):
--- a/core/dbt/parser/manifest.py
+++ b/core/dbt/parser/manifest.py
@@ -830,9 +830,7 @@ class ManifestLoader:
         for node in self.manifest.nodes.values():
             if node.created_at < self.started_at:
                 continue
-            # if the node is disabled, no need to resolve the refs
-            if node.config.enabled:
-                _process_refs_for_node(self.manifest, current_project, node)
+            _process_refs_for_node(self.manifest, current_project, node)
         for exposure in self.manifest.exposures.values():
             if exposure.created_at < self.started_at:
                 continue
--- a/core/dbt/parser/manifest.py
+++ b/core/dbt/parser/manifest.py
@@ -933,11 +933,13 @@ class ManifestLoader:
     def process_nodes(self):
         # make sure the nodes are in the manifest.nodes or the disabled dict,
         # correctly now that the schema files are also parsed
-        disable_node_copy = deepcopy(self.manifest.nodes)
-        for node in disable_node_copy.values():
+        disabled_nodes = []
+        for node in self.manifest.nodes.values():
             if not node.config.enabled:
-                self.manifest.nodes.pop(node.unique_id)
+                disabled_nodes.append(node.unique_id)
                 self.manifest.add_disabled_nofile(node)
+        for unique_id in disabled_nodes:
+            self.manifest.nodes.pop(unique_id)
 
         disabled_copy = deepcopy(self.manifest.disabled)
         for disabled in disabled_copy.values():
--- a/core/dbt/parser/manifest.py
+++ b/core/dbt/parser/manifest.py
@@ -368,6 +368,8 @@ class ManifestLoader:
                     project, project_parser_files[project.project_name], parser_types
                 )
 
+            self.process_nodes()
+
             self._perf_info.parse_project_elapsed = time.perf_counter() - start_parse_projects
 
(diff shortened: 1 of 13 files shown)


emmyoop:
This feels a little hacky but does resolve the exceptions thrown to be correct.  I opened #5869 to explore the possible underlying issue.

core/dbt/parser/schemas.py
```diff
@@ -887,9 +892,20 @@ def parse_patch(self, block: TargetBlock[NodeTarget], refs: ParserRef) -> None:
             if found_nodes:
                 # There might be multiple disabled nodes for this model
                 for node in found_nodes:
-                    # We're saving the patch_path because we need to schedule
-                    # re-application of the patch in partial parsing.
-                    node.patch_path = source_file.file_id
+                    # If this yaml file is enabled but the project config is not, we need to move
+                    # the node from disabled to manifest.nodes
+                    if patch.config.get("enabled"):
+                        test_from = {"key": block.target.yaml_key, "name": block.target.name}
```
gshank, line 898: The "test_from" variation on add_node is only for tests, not regular nodes. I don't think we should need to do "add_node" plus "remove_node'. I would just do something like popping it from "disabled" and adding it to "nodes".  Also we can't really do this if there are multiple disabled nodes, since we wouldn't know which one to enable. So I think we'll have to limit this to cases where there's only one disabled node with this unique_id.  You also shouldn't need to do a ref_lookup to get the unique_id, the node will already have a unique_id in it. So there would be an if/else after "if patch.config.get("enabled") with len(found_nodes) == 1, plus throwing an error if len is more than 1. 

I think it might be possible to just add the node to ref_lookup by manifest.ref_lookup().add_node(node) too.  Also need to remove the node from disabled_lookup. Could either rebuild it or add a function to remove node. Probably simpler to rebuild it.

core/dbt/parser/schemas.py
```diff
@@ -905,12 +921,39 @@ def parse_patch(self, block: TargetBlock[NodeTarget], refs: ParserRef) -> None:
...
                 raise_duplicate_patch_name(patch, existing_file_path)
+
             source_file.append_patch(patch.yaml_key, unique_id)
             # If this patch has config changes, re-calculate the node config
             # with the patch config
             if patch.config:
                 self.patch_node_config(node, patch)
-            node.patch(patch)
+
+            if node.config.enabled:
```
gshank, line 931: I think we don't need this if/else if we just remove it from disabled and put it in nodes as mentioned in the previous comment.  Just add back the node.patch(patch)

gshank (changes requested):
I think this needs a few changes, as described in the comments.

gshank (commented):
On thinking about this more, I think we're not handling some other cases. The precedence order is  1) dbt_project.yml config, 2) schema yaml config, 2) model config. So in theory the config in the model file could override the schema yaml config, and we have no code to move around the disabled/not disabled nodes for that case.

I really think we should switch to not having a separate dictionary for disabled, but that's probably for a later release.

For now I think that it might be best to always apply the patches to disabled nodes, except for the case where enabled is set to True and there is more than 1 matching disabled node (I kind of wish we never allowed that...) where we raise an error. Then when schema parsing is done, before refs are resolved, loop through nodes and disabled and make sure the nodes are in the right dictionary.

In addition, I think there's a hole in the "add_disabled" code that we didn't handle in the ticket for disabling metrics and exposures. It needs to be updated to check for test nodes (which used to be the only nodes from a SchemaSourceFile that could be updated), because the "test_from" piece only applies to tests.

Can you think of any additional holes?

core/dbt/parser/schemas.py
```diff
@@ -885,11 +888,20 @@ def parse_patch(self, block: TargetBlock[NodeTarget], refs: ParserRef) -> None:
...
+                    msg = (
+                        f"Found {len(found_nodes)} matching disabled nodes for '{patch.name}'. "
+                        "Multiple nodes for the same unique id cannot be disabled in the schema "
+                        "file. They must be disabled in `dbt_project.yml` or in the sql files."
+                    )
+                    raise ParsingException(msg)
+
+                # We're saving the patch_path because we need to schedule
+                # re-application of the patch in partial parsing.
+                node = found_nodes[0]
```
gshank, line 903: We should patch every node in found_nodes here, so loop through with for node in found_nodes. So the code doing the append_patch might need to be factored out.

gshank (changes requested):

core/dbt/parser/manifest.py
```diff
@@ -926,6 +929,33 @@ def process_sources(self, current_project: str):
...
+            self.manifest.nodes.pop(node.unique_id)
+            self.manifest.add_disabled_nofile(node)
+
+        enable_nodes = {}
+        for disabled in self.manifest.disabled.values():
+            for node in disabled:
+                if node.config.enabled:
+                    for dis_index, dis_node in enumerate(disabled):
+                        # Remove node from disabled and unique_id from disabled dict if necessary
+                        enable_nodes[dis_index] = dis_node
```
gshank, line 949: I commented on this last time, but it looks like github lost it somehow...  Making the index the key in the 'enable_nodes' dictionary means that if you have multiple enabled nodes at the same index, the first one will be overwritten.

I'm not sure why you are saving these in separate structures. It seems like for both looping through the nodes and looping through the disabled you could just move them when encountered.

gshank (changes requested):
A comment from last time seems to have gotten lost. The processing in parse_patch looks okay, but I think the 'process_nodes' code still has issues.

core/dbt/parser/manifest.py
emmyoop: @gshank I'm saving them as a separate structure because python doesn't like when you modify the dict you are looping through.

`disabled` on line 947 is the list of nodes for the unique id.  Probably worth renaming if it's confusing.  So the index is just the index of the list item.  There can't be multiple nodes at a single index since each index represents a single node.

core/dbt/parser/schemas.py
```diff
@@ -881,15 +884,31 @@ def parse_patch(self, block: TargetBlock[NodeTarget], refs: ParserRef) -> None:
...
-                # There might be multiple disabled nodes for this model
+                if len(found_nodes) > 1 and patch.config.get("enabled"):
+                    # There are multiple disabled nodes for this model and the schema file wants to enable one.
+                    # We have no way to know which one to enable.
+                    msg = (
+                        f"Found {len(found_nodes)} matching disabled nodes for '{patch.name}'. "
+                        "Multiple nodes for the same unique id cannot be disabled in the schema "
+                        "file. They must be disabled in `dbt_project.yml` or in the sql files."
+                    )
+                    raise ParsingException(msg)
```
emmyoop, line 900: @jtcohen6 can I get some input on the error message here, please?

core/dbt/parser/schemas.py
```diff
@@ -881,15 +884,31 @@ def parse_patch(self, block: TargetBlock[NodeTarget], refs: ParserRef) -> None:
...
             found_nodes = self.manifest.disabled_lookup.find(patch.name, patch.package_name)
             if found_nodes:
-                # There might be multiple disabled nodes for this model
+                if len(found_nodes) > 1 and patch.config.get("enabled"):
+                    # There are multiple disabled nodes for this model and the schema file wants to enable one.
+                    # We have no way to know which one to enable.
+                    msg = (
+                        f"Found {len(found_nodes)} matching disabled nodes for '{patch.name}'. "
+                        "Multiple nodes for the same unique id cannot be disabled in the schema "
+                        "file. They must be disabled in `dbt_project.yml` or in the sql files."
```
gshank, line 898: "must be enabled"

core/dbt/parser/schemas.py
```diff
@@ -881,15 +884,31 @@ def parse_patch(self, block: TargetBlock[NodeTarget], refs: ParserRef) -> None:
...
+                    )
+                    raise ParsingException(msg)
+
+                # all nodes in the disabled dict have the same unique_id so just grab the first one
```
gshank, line 902: Comment "just grab the first one" is no longer true.

core/dbt/parser/schemas.py
```diff
@@ -881,15 +884,31 @@ def parse_patch(self, block: TargetBlock[NodeTarget], refs: ParserRef) -> None:
...
+                # to append with the uniqe id
+                source_file.append_patch(patch.yaml_key, found_nodes[0].unique_id)
                 for node in found_nodes:
-                    # We're saving the patch_path because we need to schedule
-                    # re-application of the patch in partial parsing.
                     node.patch_path = source_file.file_id
+                    # re-calculate the node config with the patch config.  Always do this
+                    # for the case when no config is set to ensure the default of true gets captured
+                    self.patch_node_config(node, patch)
```
gshank, line 909: I think we can do "if patch.config:" before doing patch_node_config. If there's no config in this patch, then there's nothing to change.

core/dbt/parser/schemas.py
```diff
@@ -905,11 +924,12 @@ def parse_patch(self, block: TargetBlock[NodeTarget], refs: ParserRef) -> None:
...
-            source_file.append_patch(patch.yaml_key, unique_id)
-            # If this patch has config changes, re-calculate the node config
-            # with the patch config
-            if patch.config:
-                self.patch_node_config(node, patch)
+
+            source_file.append_patch(patch.yaml_key, node.unique_id)
+            # re-calculate the node config with the patch config.  Always do this
+            # for the case when no config is set to ensure the default of true gets captured
+            self.patch_node_config(node, patch)
```
gshank, line 931: I think we can do "if patch.config:" here too, unless I'm missing something?

emmyoop: @gshank  since the enabled config is not required, when it's not set it's the same as setting`enabled: true`.  So partial parsing was missing resetting it to true when the `enabled: false` config was removed.  This is why i ended up adding the tests to `068_partial_parsing`.

It didn't look like the expectation was for the patch to set values for things not required so we need to account for a config not being set here.

gshank: If the patch setting enabled to false was removed, the node should have been reparsed and it should then be true. I'll take a look at the tests.

core/dbt/parser/schemas.py
gshank: Read this too quickly. This is fine.

core/dbt/parser/schemas.py
gshank: We're not handling disabled nodes in 'delete_schema_mssa_links'.  Here's some code that makes that work:
```
        # remove elem node and remove unique_id from node_patches
        if elem_unique_id:
            # might have been already removed
            if elem_unique_id in self.saved_manifest.nodes or elem_unique_id in self.saved_manifest.disabled:
                if elem_unique_id in self.saved_manifest.nodes:
                    nodes = [self.saved_manifest.nodes.pop(elem_unique_id)]
                else:
                    # The value of disabled items is a list of nodes
                    nodes = self.saved_manifest.disabled.pop(elem_unique_id)
                # need to add the node source_file to pp_files
                for node in nodes:
                    file_id = node.file_id
                    # need to copy new file to saved files in order to get content
                    if file_id in self.new_files:
                        self.saved_files[file_id] = deepcopy(self.new_files[file_id])
                    if self.saved_files[file_id]:
                        source_file = self.saved_files[file_id]
                        self.add_to_pp_files(source_file)
            # remove from patches
            schema_file.node_patches.remove(elem_unique_id)
```

core/dbt/parser/manifest.py
```diff
@@ -926,6 +930,29 @@ def process_sources(self, current_project: str):
                 continue
             _process_sources_for_exposure(self.manifest, current_project, exposure)
 
+    def process_nodes(self):
+        # make sure the nodes are in the manifest.nodes or the disabled dict,
+        # correctly now that the schema files are also parsed
+        disable_node_copy = deepcopy(self.manifest.nodes)
+        for node in disable_node_copy.values():
```
gshank, line 937: I'm a little concerned by the overhead of deepcopying the entire nodes dictionary. Maybe we could have a list of disabled unique_ids, leave the 'add_disabled_nofile' in place, and just remove the disabled nodes afterward? I'm not so concerned by the disabled dictionary because it will be much smaller.

emmyoop: Oh, that's cleaner than creating a list of all nodes.  👍

gshank (approved):

core/dbt/parser/schemas.py
jtcohen6: Thanks for tagging me in — this looks mostly good to me!

Could we include the resource type? Is that available from the `patch`?

How about:
```
Found {len(found_nodes)} matching disabled nodes for {patch.resource_type} '{patch.name}'.
Multiple nodes with the same unique_id cannot be disabled in yaml resource properties.

If you need to have multiple disabled nodes with the same names, you should instead disable them
using in-file config, or resource-path config in `dbt_project.yml`.
```

PR #5870, opened 2022-09-16
Adding missing permissions on GHA

leahwicz:
### Description
This PR is to fill in any missing permissions in all our GHA in this repo. Right now our repo is defaulted to give read/write access to all GHA and I want to change it to default to read access only and then force us to explicitly opt into write access when needed. This is just a good security practice to keep the permission scope as limited as possible unless otherwise needed. When this setting is toggled to read only, the Actions that are missing listing the correct permissions will fail so this PR is to fill in those gaps now so this change doesn't break workflows.

Current setting:
image

⭐ I will not flip this switch on the setting until we are ready and can switch it back if it breaks any GHA in an unforeseen way

### Checklist

### DIFF
--- a/.github/workflows/schema-check.yml
+++ b/.github/workflows/schema-check.yml
@@ -21,6 +21,8 @@ on:
       - "*.latest"
       - "releases/*"
 
+# no permissions beyond the default needed here
+
 env:
   LATEST_SCHEMA_PATH: ${{ github.workspace }}/new_schemas
   SCHEMA_DIFF_ARTIFACT: ${{ github.workspace }}//schema_schanges.txt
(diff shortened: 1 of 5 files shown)


.github/workflows/schema-check.yml
```diff
@@ -21,6 +21,8 @@ on:
       - "*.latest"
       - "releases/*"
 
+# no permissions beyond the default needed here
```
emmyoop, line 24: Can we just put in the permission here?  I think it's a good standard to always specify the permissions, even when they match the defaults.

emmyoop (approved):
Looks good!  Just a tiny comment.  It keeps in line with what's stated in our `_README`

https://github.com/dbt-labs/dbt-core/blob/a6c37c948d7bbad4bf08436f7a70caa669cdd9c9/.github/_README.md?plain=1#L58

.github/workflows/schema-check.yml
leahwicz: That's a good point. Done!

PR #5873, opened 2022-09-16
Adding metric expression validation

### LINKED ISSUE #5871: [Bug] Metrics don't require `expression` field

### Current Behavior

Ran into this behavior when working with a beta customer. Right now, we provide the default value of empty string ("") to the expression/sql field. That then gets provided to ParsedMetric where it recognizes this field as a string and accepts it, even if the field isn't defined in the metric.

### Expected Behavior

If the expression field isn't present, the metric would fail to compile.

### Steps To Reproduce

1. Create a metric without a sql/expression field
2. Compile/parse

### Relevant log output

_No response_

### Environment

```markdown
- OS: 
- Python:
- dbt:
```

_No response_

### Additional Context

_No response_

callum-mcdata:
resolves #5871 

### Description

- changing default value for expression to None
- Adding test to confirm behavior

### Checklist

### DIFF
--- a/core/dbt/contracts/graph/unparsed.py
+++ b/core/dbt/contracts/graph/unparsed.py
@@ -476,7 +476,7 @@ class UnparsedMetric(dbtClassMixin, Replaceable):
     calculation_method: str
     timestamp: str
     description: str = ""
-    expression: Union[str, int] = ""
+    expression: Union[str, int] = None
     time_grains: List[str] = field(default_factory=list)
     dimensions: List[str] = field(default_factory=list)
     window: Optional[MetricTime] = None
--- a/core/dbt/contracts/graph/unparsed.py
+++ b/core/dbt/contracts/graph/unparsed.py
@@ -475,8 +475,8 @@ class UnparsedMetric(dbtClassMixin, Replaceable):
     label: str
     calculation_method: str
     timestamp: str
+    expression: str
     description: str = ""
-    expression: Union[str, int] = None
     time_grains: List[str] = field(default_factory=list)
     dimensions: List[str] = field(default_factory=list)
     window: Optional[MetricTime] = None
--- a/tests/functional/metrics/test_metrics.py
+++ b/tests/functional/metrics/test_metrics.py
@@ -195,6 +195,7 @@ class TestInvalidMetricMissingModel:
         with pytest.raises(ParsingException):
             run_dbt(["run"])
 
+
 invalid_metrics__missing_expression_yml = """
 version: 2
 metrics:
--- a/tests/functional/metrics/test_metrics.py
+++ b/tests/functional/metrics/test_metrics.py
@@ -201,6 +201,7 @@ version: 2
 metrics:
   - name: number_of_people
     label: "Number of people"
+    model: "ref(people)"
     description: Total count of people
     calculation_method: count
     timestamp: created_at
@@ -221,8 +222,7 @@ class TestInvalidMetricMissingExpression:
             "people.sql": models__people_sql,
         }
 
-    # tests that we get a ParsingException with an invalid model ref, where
-    # the model name does not have quotes
+    # tests that we get a ParsingException with a missing expression
     def test_simple_metric(
         self,
         project,
--- a/tests/functional/metrics/test_metrics.py
+++ b/tests/functional/metrics/test_metrics.py
@@ -195,6 +195,41 @@ class TestInvalidMetricMissingModel:
         with pytest.raises(ParsingException):
             run_dbt(["run"])
 
+invalid_metrics__missing_expression_yml = """
+version: 2
+metrics:
+  - name: number_of_people
+    label: "Number of people"
+    description: Total count of people
+    calculation_method: count
+    timestamp: created_at
+    time_grains: [day, week, month]
+    dimensions:
+      - favorite_color
+      - loves_dbt
+    meta:
+        my_meta: 'testing'
+"""
+
+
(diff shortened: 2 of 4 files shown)


callum-mcdata:
Interesting. I seem to be getting this error from the code quality check

> error: Incompatible types in assignment (expression has type "None", variable has type "Union[str, int]")  [assignment]

Is this method of implementation not correct? I suppose I could keep it the same and add another test under the `validate` function but that feels like we'd be building too many patterns into the function instead of having it live in the spec itself?

core/dbt/contracts/graph/unparsed.py
```diff
@@ -476,7 +476,7 @@ class UnparsedMetric(dbtClassMixin, Replaceable):
     calculation_method: str
     timestamp: str
     description: str = ""
-    expression: Union[str, int] = ""
+    expression: Union[str, int] = None
```
jtcohen6, line 479: @callum-mcdata I think you need to make the same change to `ParsedMetric` as well. Hoping that resolves the mypy error (though never guaranteed)

callum-mcdata: @jtcohen6 am I misunderstanding the `ParsedMetric`? The `expression` attribute is set to `expression: str`, which makes sense because I would think that enforces that it **must** be a string. I was hoping that the change to unparsed metric would then cause a None from `UnparsedMetric` to fail the requirement of string for `ParsedMetric`

jtcohen6: @callum-mcdata Ah, I understand now.

Two options here:
1. Leaving it as is, but adding a check in `UnparsedMetric.validate()` that raises a nice exception if `expression` is missing/`None` (after `rename_metric_attr` is handled of course)
2. Change the type to match exactly what's in `ParsedMetric`, that is a `str` field with no default (= required):
```
    expression: str
    description: str = ""
```
Those two have to be reordered, because required fields need to come before optional fields with defaults

I don't see any `mypy` errors because the types now match exactly. And when I define a metric that's missing `expression` + `sql` (old name), This is the exception I now see:
```
dbt.exceptions.ParsingException: Parsing Error
  Invalid metrics config given in FilePath(searched_path='models', relative_path='my_metric.yml', modification_time=1663957848.4168892, project_root='/Users/jerco/dev/scratch/testy') @ metrics: {'name': 'new_customers', 'label': 'New Customers', 'model': "ref('my_model')", 'description': 'The number of paid customers using the product', 'timestamp': 'signup_date', 'time_grains': ['day', 'week', 'month'], 'dimensions': ['plan', 'country'], 'filters': [{'field': 'is_paying', 'operator': 'is', 'value': 'true'}, {'field': 'lifetime_value', 'operator': '>=', 'value': '100'}, {'field': 'company_name', 'operator': '!=', 'value': "'Acme, Inc'"}, {'field': 'signup_date', 'operator': '>=', 'value': "'2020-01-01'"}], 'calculation_method': 'count'} - at path []: 'expression' is a required property
```
Could be nicer, but unfortunately that's our standard hologram validation error :/

Do you know why we added `int` as an option in https://github.com/dbt-labs/dbt-core/pull/5027? Is expression ever an `int`?

callum-mcdata: Got it. For sake of clean error messages, I'm leaning towards option 1 even though I don't like the idea of adding more things to the custom `validate`. Let me take a stab at this today and let you know how it goes!

As for the `int` function - I think it was to support things like "field + 1"? I'm not entirely sure though

callum-mcdata:
@jtcohen6 I went with the second option and additionally removed the `Union[str,int]` designation in `UnparsedMetric`. It passed all of the tests that we have in dbt core and just to make sure I ran it through the test suite in `dbt_metrics` as well. So assuming all the above tests complete we should be good to go!

gshank (approved):

emmyoop (approved):
Looks good.  If you can just update the test comment I can merge it in for you.

tests/functional/metrics/test_metrics.py
```diff
@@ -196,6 +196,42 @@ def test_simple_metric(
...
+class TestInvalidMetricMissingExpression:
+    @pytest.fixture(scope="class")
+    def models(self):
+        return {
+            "people_metrics.yml": invalid_metrics__missing_expression_yml,
+            "people.sql": models__people_sql,
+        }
+
+    # tests that we get a ParsingException with an invalid model ref, where
+    # the model name does not have quotes
```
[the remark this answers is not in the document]
callum-mcdata: I think this is still true, no? This PR doesn't change anything with regards to `model` validation, only `expression`. I still get a parsing exception if I remove the model field from a definition and then run `dbt parse`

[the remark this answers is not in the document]
emmyoop: @callum-mcdata  this specific test isn't testing an invalid ref missing quotes, it's testing a missing expression, right?  

You are right that we still expect the ParsingException, but the reason is different for this test.

colin-rogers-dbt (approved):
LGTM

tests/functional/metrics/test_metrics.py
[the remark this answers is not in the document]
callum-mcdata: AH - I see what you mean 🤦 . Yes I updated the comments in the test to better reflect what the actual issue is AND fixed the metric spec provided to that test because it also contained a missing model which would have error'd out and not actually tested this behavior.

Issue #5385, opened 2022-06-16
[CT-756] Stringify user-provided messages to `{{ log() }}`


jtcohen6:
### Reproduction case

```sql
-- models/any_model.sql

{% set some_list = ['a', 'b', 'c'] %}

{{ log(some_list, info = true) }}

select 1 as id
```
```
$ DBT_ENV_SECRET_WHATEVER=1234 dbt run
12:12:30  Running with dbt=1.2.0-a1
12:12:30  Encountered an error:
'list' object has no attribute 'replace'
```

I'm not even using the secret, but because I've defined one, dbt is searching log messages to scrub out any instances of the secret value. This error is _not_ related to the recent change re: secret rendering. It's just related to how we're scrubbing secrets from log messages in general, and have been since v1.0.0:

https://github.com/dbt-labs/dbt-core/blob/6e8388c653c6bbeee934aea7db93730bd88df091/core/dbt/events/functions.py#L122-L128

When a user passes a non-`string` object into `{{ log() }}`, we end up passing that object straight through to `scrub_secrets`. Which expects to be receiving a string!

The type signature of `log` actually expects to be receiving a string, too:

https://github.com/dbt-labs/dbt-core/blob/6e8388c653c6bbeee934aea7db93730bd88df091/core/dbt/context/base.py#L543-L562

### Resolution

We could put this logic anywhere, but it feels like the simplest and most appropriate fix may well be to stringify whatever users are passing into `log`, before firing it:
```python
    def log(msg: str, info: bool = False) -> str:
        ...
        # user-provided msg should be a string
        msg = str(msg)
        if info:
            fire_event(MacroEventInfo(msg=msg))
        else:
            fire_event(MacroEventDebug(msg=msg))
        return ""
```

This successfully handles some pretty weird edge cases:
```sql
{{ log(this, info = true) }}
{{ log(create_table_as, info = true) }}
{{ log(context, info = true) }}
```
```
12:11:26  "jerco"."dbt_jcohen"."model_a"
12:11:26  <dbt.clients.jinja.MacroGenerator object at 0x10dca87f0>
... giant context blob ...
```

Still, we should make sure it raises a useful error, in case users try passing in objects that aren't string serializable. (I can't think of an example right now, but maybe there is one?)

leahwicz:
@VersusFacit does this seem related to the ticket you just closed out? Should we solve this further down the stack? 
#5436 
#5874

VersusFacit:
[...]

PR #5874, opened 2022-09-18
Working solution serialization bug.

### LINKED ISSUE #5436: [CT-803] Serialization errors in AdapterLogger
Two places we've run into this:
- https://github.com/databricks/dbt-databricks/issues/115 (follow-up conversation in https://github.com/dbt-labs/dbt-spark/pull/371)
- https://github.com/dbt-labs/dbt-bigquery/issues/206

To quote @nathaniel-may from https://github.com/dbt-labs/dbt-spark/pull/371#issuecomment-1159184064, we should solve this in `dbt-core`, rather than pre-stringifying all values which might be passed into the `AdapterLogger`, because:

> This works because you're explicitly getting the value's string representation when you are making the log call. The reason the Python standard library doesn't do this, and instead uses the style of the original code is so that that string function is only called for the additional arguments if the log line is actually logged. For example, if you have a ton of debug lines with big values but you're only logging info level messages, those potentially expensive string conversions never actually happen.

> The place this should be handled is [here in core](https://github.com/dbt-labs/dbt-core/blob/d257d0b44c8118328a98caac5cb916eed4c3506c/core/dbt/events/types.py#L72) which was the original intent. If this is not working we should investigate modifying this line in core so that we can avoid calling string functions unnecessarily both to comply with a standard Python logging interface in adapters, and to avoid potential performance issues.

https://github.com/dbt-labs/dbt-core/blob/d257d0b44c8118328a98caac5cb916eed4c3506c/core/dbt/events/types.py#L72

VersusFacit:
resolves #5436

### Description
Ho, boy! What an odyssey.

#### reproducing the error

((Test case))
dbt --no-use-colors --log-format json run -s asdf | jq .msg
asdf.sql: select * from foo.bar

——————

dbt-bigquery==1.0.0
dbt-core==1.0.8

> 1 of 1 START table model my_dataset.asdf........................................ [RUN]
> Unhandled error while executing model.jaffle_shop.asdf
> Pickling client objects is explicitly not supported.
> Clients have non-trivial state that is local and unpickleable.
> 1 of 1 ERROR creating table model my_dataset.asdf............................... [ERROR in 0.61s]
> Finished running 1 table model in 1.47s.
> Completed with 1 error and 0 warnings:
> Pickling client objects is explicitly not supported.
> Clients have non-trivial state that is local and unpickleable.

dbt-bigquery==1.1.0
dbt-core==1.1.2

dbt-bigquery==1.2.0
dbt-bigquery==1.3.0b2

dbt-core==1.3.0b2
dbt-core==1.2.1

> 1 of 1 START table model my_dataset.asdf ....................................... [RUN]
> 1 of 1 ERROR creating table model my_dataset.asdf .............................. [ERROR in 0.53s]
> Finished running 1 table model in 1.17s.
> Completed with 1 error and 0 warnings:
> Compilation Error in macro statement (macros/etc/statement.sql)
>   Object of type NotFound is not JSON serializable
> 
>   > in macro materialization_table_bigquery (macros/materializations/table.sql)
>   > called by macro statement (macros/etc/statement.sql)

#### the error

`json.dumps` presumes that all elements within a dictionary can be simply rendered as a json string. But complex objects inside the argument will cause a serialization error. This happens precisely at line 209 of core/dbt/events/functions. 

The spec of `AdapterEventBase`--see line 56--assumes that the `base_msg` is a `str`. However, the `base_msg` can be anything, really, since `logger.debug` may take any exception type. So, if  we force the message given to logger functions to be a string, the spec matches reality and the error vanishes. Local proof of it working on development.

image

This error is what the user OwenKephart graciously documented over in [Bigquery issue 206](https://github.com/dbt-labs/dbt-bigquery/issues/206) as a desired output for my test case model, designed to fail.

QED.

### Checklist

### DIFF
--- a/core/dbt/events/adapter_endpoint.py
+++ b/core/dbt/events/adapter_endpoint.py
@@ -13,7 +13,7 @@ class AdapterLogger:
     name: str
 
     def debug(self, msg, *args, exc_info=None, extra=None, stack_info=False):
-        event = AdapterEventDebug(name=self.name, base_msg=msg, args=args)
+        event = AdapterEventDebug(name=self.name, base_msg=str(msg), args=args)
 
         event.exc_info = exc_info
         event.extra = extra
@@ -22,7 +22,7 @@ class AdapterLogger:
         fire_event(event)
 
     def info(self, msg, *args, exc_info=None, extra=None, stack_info=False):
-        event = AdapterEventInfo(name=self.name, base_msg=msg, args=args)
+        event = AdapterEventInfo(name=self.name, base_msg=str(msg), args=args)
 
         event.exc_info = exc_info
         event.extra = extra
@@ -31,7 +31,7 @@ class AdapterLogger:
         fire_event(event)
 
     def warning(self, msg, *args, exc_info=None, extra=None, stack_info=False):
-        event = AdapterEventWarning(name=self.name, base_msg=msg, args=args)
+        event = AdapterEventWarning(name=self.name, base_msg=str(msg), args=args)
 
         event.exc_info = exc_info
         event.extra = extra
@@ -40,7 +40,7 @@ class AdapterLogger:
         fire_event(event)
 
     def error(self, msg, *args, exc_info=None, extra=None, stack_info=False):
-        event = AdapterEventError(name=self.name, base_msg=msg, args=args)
+        event = AdapterEventError(name=self.name, base_msg=str(msg), args=args)
 
         event.exc_info = exc_info
         event.extra = extra
@@ -50,7 +50,7 @@ class AdapterLogger:
 
     # The default exc_info=True is what makes this method different
     def exception(self, msg, *args, exc_info=True, extra=None, stack_info=False):
-        event = AdapterEventError(name=self.name, base_msg=msg, args=args)
+        event = AdapterEventError(name=self.name, base_msg=str(msg), args=args)
 
         event.exc_info = exc_info
         event.extra = extra
@@ -59,7 +59,7 @@ class AdapterLogger:
         fire_event(event)
 
     def critical(self, msg, *args, exc_info=False, extra=None, stack_info=False):
-        event = AdapterEventError(name=self.name, base_msg=msg, args=args)
+        event = AdapterEventError(name=self.name, base_msg=str(msg), args=args)
 
         event.exc_info = exc_info
         event.extra = extra


core/dbt/events/adapter_endpoint.py
```diff
@@ -13,7 +13,7 @@ class AdapterLogger:
     name: str
 
     def debug(self, msg, *args, exc_info=None, extra=None, stack_info=False):
-        event = AdapterEventDebug(name=self.name, base_msg=msg, args=args)
+        event = AdapterEventDebug(name=self.name, base_msg=str(msg), args=args)
```
VersusFacit, line 16: Okay, but like Mila, why not change AdapterEventBase. I tried, but what this is force us to mutate data already encapsulated which feels wrong to my sensibilities.

However...this might not prevent the issues we have with efficiency that Nate called out? See, I can take a real crack at modifying the base class or even doing a stringification downstream in fire event. I tried all that. This seems most clean to me, but I realize now this may not  be efficient, only making sense within our current spec which maybe should change.

Thoughts, o hetairoi?

gshank (approved):
We're converting all of these Exception object to strings in the next structured logging update anyway, because everything in the event has to be encoded into a protobuf message.

core/dbt/events/adapter_endpoint.py
nathaniel-may: Huh.

To clarify the "performance" bit, it is about performance, but even more importantly it's about adhering to standard Python logging semantics- i.e. lazily converting arguments to strings. This doesn't actually break that because the arguments remain untouched.

Is this what's happening in the two user reported issues linked in the ticket? Logging a non-string value as the base message?

colin-rogers-dbt: Another idea would be to add a post init processing to convert any non-string input to string: https://docs.python.org/3/library/dataclasses.html#post-init-processing

colin-rogers-dbt (approved):
LGTM

nathaniel-may:
So it looks like #5385 has the exact same problem (passing a non-string to `log`). Since it's _jinja_ `log`, the first time exits Jinja-land into our control is in fire_event. My question to you is- do you think we should solve this lower in the call stack to knock both of these bugs out?

core/dbt/events/adapter_endpoint.py
VersusFacit: After the structured logging rework and Nate's additional ticket, post_init proved a very nice paradigm for insert a quick lazy-load-friendly type conversion.

VersusFacit:
Revised in light of the recent structured logging change. 

PR description revised, mostly expanded to describe the second bug. 

Employs the same subclass-as-trait paradigm to . And for good measure, I did check the new structured logging code on main still exhibits the bugs described in the linked issues before folding these changes into the code.

colin-rogers-dbt (commented):
Makes sense, just one question. Also, do we need unit tests to validate this behavior?

core/dbt/events/base_types.py
```diff
@@ -99,6 +99,23 @@ def level_tag(self) -> str:
         return "error"
 
 
+# Included to ensure classes with str-type message members are initialized correctly.
+@dataclass  # type: ignore[misc]
+class AdapterEventStringFunctor:
+    def __post_init__(self):
+        super().__post_init__()
+        if not isinstance(self.base_msg, str):
```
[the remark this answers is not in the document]
VersusFacit: Thought about this. We're talking data classes so these objects must be instantiated with params. A little toy example:
```python

@dataclass
class Item:
    a:str
    b:str
    c:str

asdf = Item()
```
Try running this and it complains:
> TypeError: __init__() missing 3 required positional arguments: 'a', 'b', and 'c'

Hence, if a dataclass exists, it's gotta have the params. And this functor should only be used in classes that have this param. I think that's clear? (if it's not, tell me)

VersusFacit:
Unsure about unit tests. Is it worth having tests to check that the intended spec holds for something like the structured logging? That feels extra to me, but I'm also open to writing some quick ones if it seems like a good idea from your POV!

colin-rogers-dbt (approved):
LGTM

Issue #5482, opened 2022-07-15
[Feature] Support grant-based node selection


dataders:
### Is this your first time opening an issue?

- [X] I have read the [expectations for open source contributors](https://docs.getdbt.com/docs/contributing/oss-expectations)

### Describe the Feature

a user should be able to, for the DAG subset of relations where the user `strongbad@homestarrunner.com` has SELECT access:
- run exactly said subset, and
- filter dbt docs to said subset

The fix is [this one-line change](https://github.com/dbt-labs/dbt-core/compare/jerco/config-selector-method-lists)

### Who will this benefit?

Everyone!

### Are you interested in contributing this feature?

sure?

jtcohen6:
Thanks for opening @dataders! I think this could be a `good_first_issue`, though we'll also want a test or two :)

> filter dbt docs to said subset

This will require another small change, in `dbt-docs`, possibly [hereabouts](https://github.com/dbt-labs/dbt-docs/blob/4d8cc4dcc734ec6105225ec08e430621b5a4fbc8/src/app/services/selector_methods.js#L68-L85)

Kip620:
I am taking a look at picking up this issue unless @dataders you're wanting to see this one through :)

dataders:
All you @Kip620! Appreciate your initiative.

jtcohen6:
resolved by https://github.com/dbt-labs/dbt-core/pull/5889

PR #5889, opened 2022-09-20
ConfigSelectorMethod should check for bools

danielcmessias:
# resolves

https://github.com/dbt-labs/dbt-core/issues/5890 
https://github.com/dbt-labs/dbt-core/issues/5482 

### Description

The `ConfigSelectorMethod` only checks for string equality. If you have a config property that is boolean, there is no way to match it.

Consider the example:

```yaml
# schema.yml
models:
  - name: some_model
    config:
      meta:
        pii: true
```

Then running `dbt  run --select config.meta.pii:true` would not match. The only way for it work would be to encapsulate the property as a string (`pii: "true"`), not ideal.

This is a small change such that if the property is a bool, check (case-insensitively) against the strings "true" / "false".

### Checklist

### DIFF
--- a/core/dbt/graph/selector_methods.py
+++ b/core/dbt/graph/selector_methods.py
@@ -356,7 +356,10 @@ class ConfigSelectorMethod(SelectorMethod):
             except AttributeError:
                 continue
             else:
-                if selector == value:
+                # if the config is a bool then check against the strings "true"/"false"
+                if ((selector == value) or
+                   (value is True and CaseInsensitive(selector) == "true") or
+                   (value is False and CaseInsensitive(selector) == "false")):
                     yield node
 
 
--- a/core/dbt/graph/selector_methods.py
+++ b/core/dbt/graph/selector_methods.py
@@ -356,11 +356,16 @@ class ConfigSelectorMethod(SelectorMethod):
             except AttributeError:
                 continue
             else:
-                # if the config is a bool then check against the strings "true"/"false"
-                if ((selector == value) or
-                   (value is True and CaseInsensitive(selector) == "true") or
-                   (value is False and CaseInsensitive(selector) == "false")):
-                    yield node
+                if isinstance(value, list):
+                    if ((selector in value) or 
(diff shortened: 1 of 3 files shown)


gshank:
Could you add a test?  test/unit/test_graph_selector_parsing.py looks like a good place...

core/dbt/graph/selector_methods.py
```diff
@@ -356,7 +356,10 @@ def search(
             except AttributeError:
                 continue
             else:
-                if selector == value:
+                # if the config is a bool then check against the strings "true"/"false"
```
jtcohen6, line 359: Should we support all forms of yaml truthy/falsy (0/1, `"t`/`"f"`, `yes/no`)?

I'm fine with just `"true"` + `"false"` personally

jtcohen6 (commented):
Thanks @danielcmessias! Do you feel up for adding some test cases? Relevant unit tests are [here](https://github.com/dbt-labs/dbt-core/blob/0bd6df0d1b1e4e382c413929403f7972108a5b8a/test/unit/test_graph_selector_methods.py#L771-L780).

I'm going to mark this as `ready_for_review` by Core-Execution in the meantime

core/dbt/graph/selector_methods.py
danielcmessias: I'm in favour of just true/false and avoid the yaml-specify truthy/falsy rules which don't apply in Python/JSON/Jinja templates

danielcmessias:
@gshank @jtcohen6 Added a resolution for https://github.com/dbt-labs/dbt-core/issues/5482 and some test cases. 

Went back and forth a couple of times on the structure; idk whether separate clauses for the list and string case that both yield the node is more or less readable than one giant IF with all the conditions 🤷‍♂️ . LMK if you'd prefer it another way :)

gshank:
Need to run pre-commit now, pip install pre-commit if it's not already installed, then run 'pre-commit run --all-files" and commit the changes.

gshank (approved):

PR #5897, opened 2022-09-21
.dbtignore

### LINKED ISSUE #5733: [CT-1102] [Feature] .dbtignore

### Describe the feature

A `.dbtignore` file that uses `.gitignore` syntax (including pattern matching) to define files that should NOT be parsed/loaded by dbt.

Code for this would go in [`read_files`](https://github.com/dbt-labs/dbt-core/blob/main/core/dbt/parser/read_files.py)

After thinking about this for 5 minutes, I don't think there would be unintended consequences for partial parsing. If a file is ignored, it's as if it were deleted; if it's un-ignored, it's as if it were created for the first time.

Regex support for `model-paths`, `test-paths`, etc. Gross

### Who will this benefit?

https://getdbt.slack.com/archives/C03QUA7DWCW/p1661873024869149

Concretely, people who have `.py` files in their dbt projects that are not actually Python models!

_No response_

### Anything else?

Another use case where this has come up: Removing unneeded files in dbt packages (https://github.com/dbt-labs/dbt-core/issues/4837). Unfortunately, because of how dbt installs Hub packages, this would still require downloading all files and then deleting the "ignored" ones. See conversation in https://github.com/dbt-labs/hubcap/issues/95.

ChenyuLInx:
resolves #5733

### Description
Add .dbtignore function

### Checklist

### DIFF
--- a/core/dbt/clients/system.py
+++ b/core/dbt/clients/system.py
@@ -12,6 +12,7 @@ import tarfile
 import requests
 import stat
 from typing import Type, NoReturn, List, Optional, Dict, Any, Tuple, Callable, Union
+from pathspec import PathSpec  # type: ignore
 
 from dbt.events.functions import fire_event
 from dbt.events.types import (
@@ -36,6 +37,7 @@ def find_matching(
     root_path: str,
     relative_paths_to_search: List[str],
     file_pattern: str,
+    ignore_spec: PathSpec,
 ) -> List[Dict[str, Any]]:
     """
     Given an absolute `root_path`, a list of relative paths to that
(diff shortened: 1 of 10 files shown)


jtcohen6 (commented):
Looks much simpler than I was afraid it would be. Big thanks to the maintainer of `pathspec`!

In some local testing, this seems to work just fine with partial parsing. Only real risk here is degraded parsing performance—adding more overhead to `read_files_elapsed`—which I haven't observed to a noticeable degree.

We'll want to document this, I just opened an issue so we don't lose track: https://github.com/dbt-labs/docs.getdbt.com/issues/2043

core/dbt/parser/read_files.py
```diff
@@ -1,3 +1,5 @@
+import os
+import pathspec  # type: ignore
```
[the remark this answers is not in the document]
ChenyuLInx: Thanks!!

core/dbt/parser/read_files.py
```diff
@@ -129,42 +131,84 @@ def get_source_files(project, paths, extension, parse_file_type, saved_files):
...
 
     project_files["MacroParser"] = read_files_for_parser(
-        project, files, project.macro_paths, [".sql"], ParseFileType.Macro, saved_files
+        project,
+        files,
+        project.macro_paths,
+        [".sql"],
+        ParseFileType.Macro,
+        saved_files,
+        dbt_ignore_spec,
```
ChenyuLInx, line 171: Don't really like the pass through here but we either do this or need to refactor the whole read files part of code I think

gshank (approved):
The signatures for the 'read_files' calls are getting kind of long, but I suspect that we'll be doing some major refactoring in that area sometime in the next year, so I think we can leave it until then to deal with it.

core/dbt/clients/system.py
```diff
@@ -57,19 +59,30 @@ def find_matching(
     reobj = re.compile(regex, re.IGNORECASE)
 
     for relative_path_to_search in relative_paths_to_search:
+        # potential speedup for ignore_spec
+        # if ignore_spec.matches(relative_path_to_search):
+        #     continue
         absolute_path_to_search = os.path.join(root_path, relative_path_to_search)
         walk_results = os.walk(absolute_path_to_search)
```
iknox-fa, line 66: We should really avoid using `os.walk` or `os.path` in favor of `Pathlib`.  All the os.path stuff is old AF and has weird operating system specific gotchas.

ChenyuLInx: Agree! I think we probably leave it to the larger refactor? Same as the signatures for `read_files` calls are getting long. I also don't like that but feels like right now might be okay to leave it.

PR #5915, opened 2022-09-22
expose macros in jinja context

### LINKED ISSUE #4919: [CT-403] [Feature] Allow programmatic access to macro properties

### Describe the Feature

I would like to be able to access the properties of macros, such as names and tags, at runtime in a similar way as you can access the properties of sources using `graph.sources.values()`. 

It is possible to write a python script to parse a `macros.yml` file that contains macro properties to locate specific keywords. However, since dbt cannot execute arbitrary python scripts during execution, any functionality gained by said scripts requires developers to take an extra step of executing manually, which is likely to be forgotten since it is not a standard part of dbt. 

### Who will this benefit?

This feature would benefit users who want to use jinja to dynamically generate sql based on the current properties of macros. For example, I am using the method described [here](https://discourse.getdbt.com/t/using-dbt-to-manage-user-defined-functions/18) to manage Snowflake user-defined functions in dbt. While it is not _difficult_ to add the "create_function" macro to a list in a `create_udfs()` macro, it would be really nice to save developers that step. Using jinja templating in the `create_udfs()` macro, one could loop through the macros configured in a yaml file (which, while not necessary, we want for the sake of documenting the arguments and output of each macro/function) and dynamically add any function with the tag `udf` to the `create_udfs()` macro. 

I am currently using a similar method to populate a macro using the tags applied to sources via `graph.sources.values()`. It would be useful for anything configured in a `yaml` file to have those properties available when running dbt.

Unfortunately I do not have the bandwidth right now to try this as a first contribution.

### Anything else?

- UDF management method (also linked above): https://discourse.getdbt.com/t/using-dbt-to-manage-user-defined-functions/18
- Slack thread: https://getdbt.slack.com/archives/C2JRRQDTL/p1647645997578909

Illustrative example of how this could work if macros were in the graph:

```
{% macro execute_macros() %}
{% if execute %}
        {%- for macro in graph.macros.values()
                | selectattr('resource_type', 'equalto', macro')
                | selectattr('tags') %}
            {%- if 'udf' in macro.tags -%}
                {{ macro.name }}();
                {%- do log("Ran operation " ~ macro.name , info=True) -%}
            {%- endif -%}
        {%- endfor -%}
        commit;
{% endif %}
{% endmacro %}
```

karunpoudel:
resolves #4919

### Checklist

### DIFF
--- a/core/dbt/context/providers.py
+++ b/core/dbt/context/providers.py
@@ -1264,7 +1264,7 @@ class ProviderContext(ManifestContext):
 
     @contextproperty
     def macros(self) -> Dict[str, Any]:
-        return self.manifest.macros
+        return self.manifest.flat_macros
 
 
 class MacroContext(ProviderContext):
--- a/core/dbt/context/providers.py
+++ b/core/dbt/context/providers.py
@@ -1262,6 +1262,10 @@ class ProviderContext(ManifestContext):
             )
         return self.adapter.submit_python_job(parsed_model, compiled_code)
 
+    @contextproperty
+    def macros(self) -> Dict[str, Any]:
+        return self.manifest.macros
+
 
 class MacroContext(ProviderContext):
     """Internally, macros can be executed like nodes, with some restrictions:
--- a/.changes/unreleased/Features-20220926-105313.yaml
+++ b/.changes/unreleased/Features-20220926-105313.yaml
@@ -1,0 +1,7 @@
+kind: Features
+body: expose macros in Jinja context
+time: 2022-09-26T10:53:13.4911753-05:00
+custom:
+  Author: karunpoudel
+  Issue: "4949"
+  PR: "5915"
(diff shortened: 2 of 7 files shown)


core/dbt/context/providers.py
```diff
@@ -1262,6 +1262,10 @@ def submit_python_job(self, parsed_model: Dict, compiled_code: str) -> AdapterRe
             )
         return self.adapter.submit_python_job(parsed_model, compiled_code)
 
+    @contextproperty
+    def macros(self) -> Dict[str, Any]:
+        return self.manifest.macros
```
karunpoudel, line 1267: Should the return type hint be changed to `MutableMapping[str, ParsedMacro]`
or does each macro needs to be converted to dict like
`return {k: v.to_dict() for k, v in self.manifest.macros.items()}`

jtcohen6: Yes, if we were to do this, we'd want to turn `manifest.macros` into a read-only dictionary, akin to what's done in [`build_flat_graph`](https://github.com/dbt-labs/dbt-core/blob/56783446dbc765d93c43afad2b763302f034ef6b/core/dbt/contracts/graph/manifest.py#L690-L701).

karunpoudel:
This PR might still be helpful for issue #4919

PR #5921, opened 2022-09-23
Fix adapter reset race condition in lib.py

### LINKED ISSUE #5919: [CT-1233] [Bug] Race condition in lib.py causes concurrent API calls to fail

### Current Behavior

The `lib.py` file in Core contains logic to "reset adapters" in the process of generating a RuntimeConfig object. While this logic is appropriate for single-threaded usage of dbt Core, we've seen that concurrent api calls to this method can result in race conditions:

https://github.com/dbt-labs/dbt-core/blob/56783446dbc765d93c43afad2b763302f034ef6b/core/dbt/lib.py#L27-L32

The `reset_adapters` call will clear out the set of adapters registered in the dbt Core adapter factory. If a concurrent api call accesses adapters (eg. to compile sql, or run dbt, or similar) before the `register_adapters` call is complete, there's a chance it will fail with a `KeyError` eg. [here](https://github.com/dbt-labs/dbt-core/blob/main/core/dbt/adapters/factory.py#L101) because the adapter dictionary was cleared in another thread.

### Expected Behavior

The `get_dbt_config` should not mutate global state to generate a RuntimeConfig.

### Steps To Reproduce

Here's a minimal repro case:

```python
import dbt.lib
import threading
import time
import random
import dbt.adapters.factory

PROJECT_PATH = "/Users/drew/dev/internal-analytics"


def repro_func(thread_no):
    time.sleep(random.random())

    try:
        # Simulate concurrent use of adapters
        config = dbt.lib.get_dbt_config(PROJECT_PATH)
        adapter = dbt.adapters.factory.get_adapter_by_type(config.credentials.type)
        adapter.acquire_connection()
        adapter.execute('select 1 as id')

        print(f"Thread {thread_no} OK")
    except KeyError:
        print("ERROR: KeyError")


threads = []
for i in range(25):
    thread = threading.Thread(target=repro_func, args=(i,))
    thread.start()
    threads.append(thread)

for thread in threads:
    thread.join()
```

Screen Shot 2022-09-23 at 9 51 27 AM

### Relevant log output

_No response_

### Environment

```markdown
- OS: macOS
- Python: 3.9.14
- dbt: 1.2.1
```

postgres, redshift, snowflake, bigquery, spark

### Additional Context

I am interested in submitting a patch for this issue

drewbanin:
resolves #5919

### Description

This PR updates the `lib.py` file to avoid mutating global adapter state when generating a RuntimeConfig. This logic was once necessary in order to work around issues with `dbt deps`, though it's fairly clear at this point that we'll need to employ a radically different approach for deps in the future (including higher-order, first-class APIs). In the meantime, this change fixes possible race conditions for concurrent callers of `get_dbt_config`

Additional changes:
- Change RuntimeArgs from a namedtuple to a dataclass fix an error in dbt v1.3.x (via [this PR](https://github.com/dbt-labs/dbt-core/pull/5782), i think)
- Tweaked / fixed up code comments :)
- Made it possible to control manifest sql operation nodes names. Historically, they've been hard-coded to `"name"`. The default value is `query` for backwards compatibility. Just added an optional arg for this

TODO:
 - write tests (would love guidance on the best place to test this! Not sure that we want/need to repro the race condition, but a couple of tests for `lib.py` would be dope here)
 - figure out how to run changie again
 - almost positive black and flake8 are going to yell at me

### Checklist

### DIFF
--- a/core/dbt/lib.py
+++ b/core/dbt/lib.py
@@ -1,10 +1,16 @@
-# TODO: this file is one big TODO
 import os
 from dbt.exceptions import RuntimeException
 from dbt import flags
-from collections import namedtuple
+from dataclasses import dataclass
 
-RuntimeArgs = namedtuple("RuntimeArgs", "project_dir profiles_dir single_threaded profile target")
+
+@dataclass
+class RuntimeArgs:
+    project_dir: str
+    profiles_dir: str
+    single_threaded: bool
+    profile: str
+    target: str
 
 
 def get_dbt_config(project_dir, args=None, single_threaded=False):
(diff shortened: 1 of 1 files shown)


gshank (approved):
I agree that at some point it would be good to have tests for lib.py. I'm not exactly sure what they'd look like though. I'm okay with doing that as a separate effort.

PR #5926, opened 2022-09-23
Runtime: Prevent introspective queries at compile (SL only)

racheldaniel:
resolves #[RUNTIME-443](https://dbtlabs.atlassian.net/browse/RUNTIME-443)

### Description

This PR conditionally no-ops warehouse connection at compile depending on an env var. This is a temporary solution to more complex permissions requirements for the semantic layer.

### Checklist

### DIFF
--- a/core/dbt/lib.py
+++ b/core/dbt/lib.py
@@ -1,5 +1,6 @@
 # TODO: this file is one big TODO
 # from dataclasses import dataclass
+from collections import namedtuple
 from dataclasses import dataclass
 import os
 from dbt.contracts.results import NodeStatus, RunningStatus, collect_timing_info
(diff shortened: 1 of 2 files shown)


leahwicz:
@racheldaniel for we need this to get into the 1.3 release?

racheldaniel:
> @racheldaniel for we need this to get into the 1.3 release?

@leahwicz yes, and back-ported as well if possible 😬 . What do I need to do to ensure that goes smoothly?

core/dbt/lib.py
```diff
@@ -13,6 +17,46 @@ class RuntimeArgs:
     target: str
 
 
+class SqlCompileRunnerNoIntrospection(SqlCompileRunner):
+    def compile_and_execute(self, manifest, ctx):
```
ChenyuLInx, line 21: Looks like this function is the same as `compile_and_execute` [here](https://github.com/dbt-labs/dbt-core/blob/7b1d61c956d47be759faebf2f479334e585c7585/core/dbt/task/base.py#L313) with line 313 `with self.adapter.connection_for(self.node):` removed. Looks good now but I would recommend add a TODO to refactor here when doing runtime(maybe?) so we are not duplicating majority of this function in two places.

ChenyuLInx (approved):

Issue #5696, opened 2022-08-22
[Spike] 


iknox-fa:
### Is this your first time submitting a feature request?

- [X] I have read the [expectations for open source contributors](https://docs.getdbt.com/docs/contributing/oss-expectations)
- [X] I have searched the existing issues, and I could not find an existing issue for this feature
- [X] I am requesting a straightforward extension of existing dbt functionality, rather than a Big Idea better suited to a discussion

### Describe the feature

In order to [switch to using `pyproject.toml`](#4674) we need to spike a few impact/feasibility things:
* Will this break things in `dbt-cloud`?  I'm specifically concerned about our docker images since I know they're a bit fragile install wise.
* Are all of the features we use in `setup.py` supported in `pyproject.toml`?  Our setuptools setup isn't terribly complex, but there's some uncertainty here (esp in adapter repos re the version checking regex stuff :grimace:)

jtcohen6:
> esp in adapter repos re the version checking regex stuff :grimace:

I'm incredibly open to simplifying this logic, or moving it into `bumpversion` logic!

bbroeksema:
There is this kind of pre-processing going:
```python
this_directory = os.path.abspath(os.path.dirname(__file__))
with open(os.path.join(this_directory, "README.md")) as f:
    long_description = f.read()
```

I don't think this is possible with `toml` so this would require some "templating" and moving this to the Makefile stage.

babaMar:
It seems such pre-processing is not needed as the `readme` field of the `[project]` table support relative file paths. See https://peps.python.org/pep-0621/#readme

DustinMoriarty:
I would be happy to help with migrating to poetry if there is interest.  Migrating to poetry is pretty easy because it is so structured.  Poetry seems to be the winner among all the build tools lately.  It is a back compatible change because the build output still meets the PEP standards.

DustinMoriarty:
> There is this kind of pre-processing going:
> 
> ```python
> 
> this_directory = os.path.abspath(os.path.dirname(__file__))
> 
> with open(os.path.join(this_directory, "README.md")) as f:
> 
>     long_description = f.read()
> 
> ```
> 
> 
> 
> I don't think this is possible with `toml` so this would require some "templating" and moving this to the Makefile stage.

That use case isn't needed if you use poetry.  You just specify the README.md.  All that manual stuff we used to do ad-hoc in setup.py is structured and standardized.  

[...]

PR #5947, opened 2022-09-28
Put black config in explicit config

### LINKED ISSUE #5946: [CT-1259] [Bug] Python formatter configs are not in config files

### Current Behavior

Not quite a bug, but what category should this be in? According to the PR template I need to open an issue.

Python formatter configs are not in config files, so IDEs can't read them, so saving a file in VSCode changes the formatting, which pre-commit then changes back.

The PR will put the configs in a single place where everything can read them!

### Expected Behavior

n/a

### Steps To Reproduce

n/a

### Relevant log output

_No response_

### Environment

```markdown
- OS:
- Python:
- dbt:
```

_No response_

### Additional Context

_No response_

max-sixty:
This lets editors etc use the correct config

resolves #5946

### Checklist

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -24,15 +24,10 @@ repos:
   rev: 22.3.0
   hooks:
   - id: black
-    args:
-    - "--line-length=99"
-    - "--target-version=py38"
   - id: black
     alias: black-check
     stages: [manual]
     args:
-    - "--line-length=99"
-    - "--target-version=py38"
     - "--check"
     - "--diff"
 - repo: https://gitlab.com/pycqa/flake8


iknox-fa:
Hi @max-sixty!
Thanks for opening this PR, however merging it would [break our pre-commit hooks](https://github.com/pre-commit/pre-commit/issues/1165) which we rely on for:
* [local development Makefile targets](https://github.com/dbt-labs/dbt-core/blob/7dbdfc88e001b59598af7f5902f6604eb30fb26f/Makefile#L57)
* [checking our code in the release process](https://github.com/dbt-labs/dbt-core/blob/main/.github/workflows/main.yml#L62)
* [cleaning up the PR from our version bump setup](https://github.com/dbt-labs/dbt-core/blob/7dbdfc88e001b59598af7f5902f6604eb30fb26f/.github/workflows/version-bump.yml#L95)

Additionally adding a `pyproject.toml` has a lot of implications for packaging that we haven't explored (although we have a [ticket in the queue to do so](https://github.com/dbt-labs/dbt-core/issues/5696) at some point) 

Can you say a bit more about what you're trying to achieve here?  If all you want is to use black in your IDE instead of as a pre-commit hook, may I suggest just configuring black directly in your IDE ([VSCode](https://dev.to/adamlombard/vscode-setting-line-lengths-in-the-black-python-code-formatter-1g62) / [Pycharm)](https://stackoverflow.com/a/57113896))?

iknox-fa (changes requested):
Needs further discussion
See: https://github.com/dbt-labs/dbt-core/pull/5947#issuecomment-1269025240

max-sixty:
> however merging it would [break our pre-commit hooks](https://github.com/pre-commit/pre-commit/issues/1165) which we rely on for:

I don't think that's the case, respectfully. In the existing code, pre-commit passes configs to black. In the proposed code, black already has the configs. So pre-commit is running with exactly the same configs in both cases. Note that the pre-commit tests pass; and feel free to have a go locally. 

This is the standard approach for using static tools with pre-commit, e.g. we've done this in https://github.com/pydata/xarray for years.

> Can you say a bit more about what you're trying to achieve here? If all you want is to use black in your IDE instead of as a pre-commit hook, may I suggest just configuring black directly in your IDE ([VSCode](https://dev.to/adamlombard/vscode-setting-line-lengths-in-the-black-python-code-formatter-1g62) / [Pycharm)](https://stackoverflow.com/a/57113896))?

Right but with the proposed approach, it avoids everyone having to add a repo-specific configuration themselves.

---

Though I don't have a religious belief here. I noticed it when doing development on the repo, and thought it would be helpful to make it conform to best practice. But if the whole team has an aesthetic view against it, I would not be offended!

iknox-fa:
My mistake @max-sixty -- I misremembered the process isolation pre-commit uses when running hooks as isolating them from non `pre-commit-config.yaml` config values.  Let's give it a whirl.

iknox-fa (approved):
LGTM

max-sixty:
Cheers @iknox-fa ! Appreciate your openness

PR #5954, opened 2022-09-28
Initial structured logging changes

### LINKED ISSUE #5662: [CT-1047] Define current logging events in proto files, generate python classes, pull into events in types.py
In pull request #5643 we implemented a few test events using proto definitions and betterproto generated python classes. We should convert all of the current events in a similar way, using betterproto.

This includes converting current Exception attributes to strings.

The existing test/unit/test_events.py should be modified to test the new proto based events.

gshank:
resolves #5662

### Description

Initial changes for using proto message in logging events.

### Checklist


core/setup.py
```diff
@@ -48,6 +48,7 @@
     install_requires=[
         "Jinja2==3.1.2",
         "agate>=1.6,<1.6.4",
+        "betterproto>=1.2.5",
```
emmyoop, line 51: Do we have a standard on how we're pinning packages?  Most (but not all) the time we seem to either pin to an exact version or a range with lower and upper bounds.

emmyoop (commented):
This is a big PR.  I had some questions in several places.

.github/workflows/structured-logging-schema-check.yml
```diff
@@ -69,10 +63,3 @@ jobs:
       # we actually care if these pass, because the normal test run doesn't usually include many json log outputs
       - name: Run integration tests
         run: tox -e integration -- -nauto
-
-      # apply our schema tests to every log event from the previous step
-      # skips any output that isn't valid json
-      - uses: actions-rs/cargo@v1
-        with:
-          command: run
-          args: --manifest-path test/interop/log_parsing/Cargo.toml
```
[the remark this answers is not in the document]
gshank: I removed the rust piece because a lot of it would have to be refactored and in my opinion it doesn't help that much. If someone disagrees or wants to do that re-implementation work, that's fine. I would personally vote for doing that in Python, so it can more easily be run locally.

The test as it is now does check that everything can be serialized and did catch a case where merging main had removed some updates to a fire_event call.

There's another ticket to implement log_format of protobuf and round trip the formatted logs.

core/dbt/events/types.py
```diff
@@ -1,22 +1,21 @@
...
-    NodeInfo,
     Cache,
 )
 from dbt.events.format import format_fancy_output_line, pluralize
-from dbt.events.serialization import EventSerialization
+
+# The generated classes quote the included message classes, requiring the following line
+from dbt.events.proto_types import EventInfo, RunResultMsg, ListOfStrings  # noqa
+from dbt.events.proto_types import NodeInfo, ReferenceKeyMsg
+from dbt.events import proto_types as pl
```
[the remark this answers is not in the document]
gshank: I had that thought when I read this too. I don't actually remember how it got to be "pl". I suspect that I had a different name before naming it proto_types. I'll change it.

core/dbt/events/types.py
```diff
@@ -38,908 +33,918 @@
...
+class ConnectionClosed2(DebugLevel, pl.ConnectionClosed2):
+    def code(self):
+        return "E010"
 
     def message(self) -> str:
-        return f'STDOUT: "{str(self.bmsg)}"'
+        return f"On {self.conn_name}: Close"
 
 
+# TODO: can we combine this with ConnectionLeftOpen?
```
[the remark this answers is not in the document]
gshank: That's a good point. I do think that in some cases we might have a logically equivalent location that's in different places in the code, but we want to use the same event.

core/dbt/adapters/base/connections.py
```diff
@@ -332,17 +334,21 @@ def _rollback_handle(cls, connection: Connection) -> None:
...
+                )
+            )
 
     @classmethod
     def _close_handle(cls, connection: Connection) -> None:
         """Perform the actual close operation."""
         # On windows, sometimes connection handles don't have a close() attr.
         if hasattr(connection.handle, "close"):
-            fire_event(ConnectionClosed2(conn_name=connection.name))
+            fire_event(ConnectionClosed2(conn_name=cast_to_str(connection.name)))
```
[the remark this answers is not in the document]
gshank: The generated betterproto Python classes only support the native proto types. There's no easy way to do that conversion in the generated classes.  Attributes in the proto classes can only be those native types plus embedded defined messages.

core/dbt/adapters/factory.py
```diff
@@ -63,7 +64,7 @@ def load_plugin(self, name: str) -> Type[Credentials]:
             # otherwise, the error had to have come from some underlying
             # library. Log the stack trace.
 
-            fire_event(PluginLoadError())
+            fire_event(PluginLoadError(exc_info=traceback.format_exc()))
```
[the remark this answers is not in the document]
gshank: The ShowException class was basically broken so we weren't including exc_info in places where it was supposed to be. Passing in the exc_info on these events was to fix the lack of that information.

core/setup.py
gshank: I don't know if there's a standard. Do you think we should put == instead?

jtcohen6: Our policies for these things are WIP ([here](https://www.notion.so/WIP-Core-Dependency-Policies-483df216b3c64c989c0720a766db6f03)), but basically: IMO we shouldn't tightly pin (`==`) if we don't need. (It makes sense to do that in "locked" requirements files that guarantee a working install, but not for setup requirements.) And the reason to add an upper bound is just to avoid breaking changes that could come in v2.0.0. Have we tested with older versions of betterproto v1, before 1.2.5? If it can be as simple as `betterproto>1,<2`, amazing. If we're using capabilities that are new since v1.0, then `betterproto>=1.2.5,<2` should do the trick.

emmyoop: v1.2.5 was released May 2020.  Looks like they've been actively working on a beta 2.0 since then.  Since the release is older, I would vote to pin it.  Seems unlikely they will release any new minor/patch releases for v1.  I can't find any timelines for the beta but have an open question in their slack about it.  @gshank were you able to find anything about timelines?

.github/workflows/structured-logging-schema-check.yml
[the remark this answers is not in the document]
emmyoop: @nathaniel-may do you remember why this was written in Rust?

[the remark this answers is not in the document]
nathaniel-may: We're consuming the logs from an "outside source" so it could be written in anything. I chose Rust because you can generate most of the code we'd be using here from a type definition. Plus Python's serialization-deserialization story is pretty abysmal so I didn't consider that a good option.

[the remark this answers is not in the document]
nathaniel-may: So the way I see it, we only have two ways to verify that our structured logs conform to anything even remotely reasonable for our users to consume: 

1. mypy in [base_types.py](https://github.com/dbt-labs/dbt-core/blob/main/core/dbt/events/base_types.py) but mypy is limited on what it can check because it has to be defined at the type level. (e.g. - it can't catch positive vs negative integers)
1. These interop tests written in Rust. There are lots of values we can't check at the type level so I think good we check values since it's a programmatic interface.

I will concede that I didn't add very many value checks so @gshank is correct to call out it's limited usefulness in its current state. The original idea is that we would add more to these tests over time which we haven't really done. If Rust isn't the right choice for the team I'm perfectly fine to see this rewritten.

core/setup.py
colin-rogers-dbt: As I recall the betterproto 2.0 beta has some great features but definitely some breaking/unexpected behavior. Not locking to the major version seems risky

colin-rogers-dbt: also there can be unexpected behavior if we compile with a different betterproto version than we use at runtime

dev-requirements.txt
```diff
@@ -1,3 +1,4 @@
+betterproto[compiler]
```
colin-rogers-dbt, line 1: we should lock this version as well

.github/workflows/structured-logging-schema-check.yml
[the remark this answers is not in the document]
gshank: Will open a ticket to look at restoring some of this testing.

core/dbt/events/functions.py
```diff
@@ -189,18 +177,17 @@ def create_debug_text_log_line(e: T_Event) -> str:
...
-    raw_log_line = json.dumps(values, sort_keys=True)
+    raw_log_line = event_to_json(e)
     return scrub_secrets(raw_log_line, env_secrets())
 
 
 # calls create_stdout_text_log_line() or create_json_log_line() according to logger config
-def create_log_line(e: T_Event, file_output=False) -> Optional[str]:
+def create_log_line(e: BaseEvent, file_output=False) -> Optional[str]:
     if this.format_json:
+        # TODO: Do we want to skip EmptyLine() like the TextOnly for logbook?
```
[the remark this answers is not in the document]
gshank: Removed comment.

dev-requirements.txt
gshank: Locked

core/setup.py
gshank: Pinned to 1.2.5

core/dbt/events/README.md
```diff
@@ -1,55 +1,38 @@
...
 
-    def message(self) -> str:
-        return f"before adding : {self.dump.force()}"
-```
+A matching class needs to be created in the core/dbt/events/types.py file, which will have two superclasses, the "Level" mixin and the generated class from proto_types.py. These classes will also generally have two methods, a "code" method that returns the event code, and a "message" method that is used to construct the "msg" from the event fields. In addition the "Level" mixin will provide a "level_tag" method to set the level (which can also be overridden using the "info" convenience function from functions.py)
 
+Note that no attributes can exist in these event classes except for fields defined in the protobuf definitions, because the betterproto metaclass will throw an error. Betterproto provides a to_dict() method to convert the generated classes to a dictionary and from that to json. However some attributes will successfully convert to dictionaries but not to serialized protobufs, so we need to test both output formats.jjkkkkkkkkkkj
```
emmyoop, line 15: I'm guessing the `jjkkkkkkkkkkj` at the end was accidental?

core/dbt/events/README.md
```diff
@@ -1,55 +1,38 @@
...
 
-# Adding a New Event
-In `events.types` add a new class that represents the new event. All events must be a dataclass with, at minimum, a code.  You may also include some other values to construct downstream messaging. Only include the data necessary to construct this message within this class. You must extend all destinations (e.g. - if your log message belongs on the cli, extend `Cli`) as well as the loglevel this event belongs to.  This system has been designed to take full advantage of mypy so running it will catch anything you may miss.
 
 ## Required for Every Event
 
-- a string attribute `code`, that's unique across events
-- assign a log level by extending `DebugLevel`, `InfoLevel`, `WarnLevel`, or `ErrorLevel`
+- a method `code`, that's unique across events
+- assign a log level by using the Level misin: `DebugLevel`, `InfoLevel`, `WarnLevel`, or `ErrorLevel`
```
emmyoop, line 21:
```suggestion
- assign a log level by using the Level mixin: `DebugLevel`, `InfoLevel`, `WarnLevel`, or `ErrorLevel`
```

core/dbt/events/types.py
```diff
@@ -38,908 +33,918 @@
...
-
+class MainReportVersion(InfoLevel, pt.MainReportVersion):  # noqa
+    def code(self):
+        return "A001"
 
-@dataclass
-class AdapterEventInfo(InfoLevel, AdapterEventBase, ShowException):
-    code: str = "E002"
+    def message(self):
+        return f"Running with dbt{self.version}"
```
emmyoop, line 62: Should this include the log version as well?  You're passing it in in `main.py`.

emmyoop (approved):
Just a few small things to tweak, overall looks good!

core/dbt/events/README.md
gshank: pfft

core/dbt/events/types.py
gshank: That would change the terminal output, which I don't think is necessary for the log format, just wanted to put it in a structured entry someplace.  We can discuss what to do about the log format.

PR #5971, opened 2022-09-28
[CT-1271] Test Query Comment conversion

### LINKED ISSUE #5962: [CT-1271] 051_query_comments_tests
051_query_comments_tests from https://github.com/dbt-labs/dbt-core/issues/4788

McKnight-42:
resolves #5962 

Description:
 converting integration test 051 to a new functional test
 todo: 
1. make sure new test works on `dbt-core` and all required adapters including:
- [x] Bigquery
 - [x] Snowflake
 - [x] Redshift
2. check to see if we need to add to `dbt-spark`

### Checklist


colin-rogers-dbt (approved):
LGTM

tests/adapter/dbt/tests/adapter/query_comment/fixtures.py
```diff
@@ -0,0 +1,64 @@
...
+{% endfor %}
+
+{% if 'type' not in target %}
+  {% do exceptions.raise_compiler_error('invalid target, missing "type"') %}
+{% endif %}
+
+{% set required = ['name', 'schema', 'type', 'threads'] %}
+
+{# Require what we docuement at https://docs.getdbt.com/docs/target #}
+{% if target.type == 'postgres' or target.type == 'redshift' %}
```
gshank, line 45: The problem with doing it this way is that it doesn't support external adapter maintainers very well, and we don't actually provide an example of how to get this working with an external adapter except by replacing this entire model. In addition, putting the test in jinja model code is kind of a bad pattern.

 It would be better to put the required fields in some kind of config that can be retrieved. Also, it would probably be better to move the checks into the actual test code, rather than embedding it in the model. I'd make the "required" fields a fixture that can be overridden in the adapter specific tests,  and then pull that into the test method and test for the existence of those fields in the target in the test.  I'm not exactly sure how you can get the target dictionary. Possibly from the adapter? I can help look for how to get that if you need.

tests/adapter/dbt/tests/adapter/query_comment/test_query_comment.py
```diff
@@ -0,0 +1,181 @@
...
+        for log in logs:
+            msg = self.query_comment("model.test.x", log)
+            if msg is not None and self.matches_comment(msg):
+                seen = True
+
+        for log in logs:
+            if seen:
+                "Never saw a matching log message! Logs:\n{}".format("\n".join(log["msg"]))
+
+    def test_comments(self, project):
```
gshank, line 91: I don't see a reason for just forwarding to another method here. Just

tests/adapter/dbt/tests/adapter/query_comment/fixtures.py
gshank: I'm not even sure why we're doing this in this test. What does this have to do with query comments?

tests/adapter/dbt/tests/adapter/query_comment/test_query_comment.py
McKnight-42: just? @gshank did something get cutoff here?

tests/adapter/dbt/tests/adapter/query_comment/fixtures.py
McKnight-42: as we talked about i'm a little hazy on why this is here other than validation logic possibly against if we did import it to a different database to make sure its not actually relying off say postgres. 

would love to find a better way to work on this will play around some of these other options if we do think this chucnk is needed at all.

tests/adapter/dbt/tests/adapter/query_comment/test_query_comment.py
```diff
@@ -0,0 +1,181 @@
...
+        }
+
+    @pytest.fixture(scope="class")
+    def macros(self):
+        return {
+            "macro.sql": MACROS__MACRO_SQL,
+        }
+
+    @pytest.fixture(scope="class", autouse=True)
+    def setUp(self, project):
```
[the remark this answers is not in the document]
McKnight-42: looks like I was able to remove it logging stuff is still a area I need to buff up on so this is a great learning part here.

tests/adapter/dbt/tests/adapter/query_comment/test_query_comment.py
```diff
@@ -0,0 +1,181 @@
...
+        self.initial_stdout = log_manager.stdout
+        self.initial_stderr = log_manager.stderr
+        self.stringbuf = io.StringIO()
+        log_manager.set_output_stream(self.stringbuf)
+
+    def run_get_json(self, expect_pass=True):
```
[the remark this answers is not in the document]
McKnight-42: modifying

tests/adapter/dbt/tests/adapter/query_comment/test_query_comment.py
```diff
@@ -0,0 +1,181 @@
...
+            if msg in {"COMMIT", "BEGIN", "ROLLBACK"}:
+                return None
+            return msg
+        return None
+
+    def run_assert_comments(self):
+        logs = self.run_get_json()
+
+        seen = False
+        for log in logs:
```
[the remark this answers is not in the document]
McKnight-42: pulling out to look at logs as a whole

tests/adapter/dbt/tests/adapter/query_comment/test_query_comment.py
gshank: maybe?  Probably was saying "just call the method directly"

tests/adapter/dbt/tests/adapter/query_comment/fixtures.py
McKnight-42: Hey @gerda had a few questions/ideas are we able to set up some global stuff for checks in a fixture that can't be overwritten by chance? like the 
`{% set required = ['name', 'schema', 'type', 'threads'] %}` that might not change between the adapters?

trying to get ahold of the related information that would be same as `target` in the fixture looks like from our `TestProjInfo` something like `profiles_dir` would contain the information we would want just not sure if theres a good way to tap into that possibly by some `os` read stuff?

gshank (approved):
Looks good!

PR #5972, opened 2022-09-28
tracking works with click

ChenyuLInx:
resolves #

### Checklist

### DIFF
--- a/core/dbt/cli/main.py
+++ b/core/dbt/cli/main.py
@@ -7,6 +7,7 @@ from dbt.adapters.factory import adapter_management
 from dbt.cli import params as p
 from dbt.cli.flags import Flags
 from dbt.profiler import profiler
+from dbt.tracking import initialize_from_flags
 
 
 def cli_runner():
@@ -48,18 +49,23 @@ def cli_runner():
 @p.version_check
 @p.warn_error
 @p.write_json
+@p.profile
+@p.profiles_dir
+@p.project_dir
 def cli(ctx, **kwargs):
     """An ELT tool for managing your SQL transformations and data models.
     For more documentation on these commands, visit: docs.getdbt.com
     """
     incomplete_flags = Flags()
-
     # Profiling
     if incomplete_flags.RECORD_TIMING_INFO:
         ctx.with_resource(profiler(enable=True, outfile=incomplete_flags.RECORD_TIMING_INFO))
 
     # Adapter management
     ctx.with_resource(adapter_management())
+
+    # tracking
+    initialize_from_flags()
 
     # Version info
     if incomplete_flags.VERSION:
@@ -79,9 +85,6 @@ def cli(ctx, **kwargs):
 @p.indirect_selection
 @p.log_path
 @p.models
-@p.profile
-@p.profiles_dir
-@p.project_dir
 @p.selector
 @p.show
 @p.state
@@ -100,9 +103,6 @@ def build(ctx, **kwargs):
 # dbt clean
 @cli.command("clean")
 @click.pass_context
-@p.profile
-@p.profiles_dir
-@p.project_dir
 @p.target
 @p.vars
 def clean(ctx, **kwargs):
@@ -126,9 +126,6 @@ def docs(ctx, **kwargs):
 @p.exclude
 @p.log_path
 @p.models
-@p.profile
-@p.profiles_dir
-@p.project_dir
 @p.selector
 @p.state
 @p.target
@@ -147,9 +144,6 @@ def docs_generate(ctx, **kwargs):
 @click.pass_context
 @p.browser
 @p.port
-@p.profile
-@p.profiles_dir
-@p.project_dir
 @p.target
 @p.vars
 def docs_serve(ctx, **kwargs):
@@ -167,9 +161,6 @@ def docs_serve(ctx, **kwargs):
 @p.log_path
 @p.models
 @p.parse_only
-@p.profile
-@p.profiles_dir
-@p.project_dir
 @p.selector
 @p.state
 @p.target
@@ -187,9 +178,6 @@ def compile(ctx, **kwargs):
 @cli.command("debug")
 @click.pass_context
 @p.config_dir
-@p.profile
-@p.profiles_dir
-@p.project_dir
 @p.target
 @p.vars
 @p.version_check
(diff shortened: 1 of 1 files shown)


core/dbt/cli/main.py
```diff
@@ -48,19 +49,24 @@ def cli_runner():
 @p.version_check
 @p.warn_error
 @p.write_json
+@p.profile
+@p.profiles_dir
+@p.project_dir
```
ChenyuLInx, line 54: Moving these 3 since they appeared in every task

ChenyuLInx:
@iknox-fa  current status of the code is that `active_user` for tracking will be initialized and ready to use. I am not sure how far I should go on for the refactor of tracking. There's a couple of directions of refactor
1. try to make `invocation_id` not a global object any more
   - this is kind of tricky as `invocation_id` is also being used by [logger](https://github.com/dbt-labs/dbt-core/blob/7b1d61c956d47be759faebf2f479334e585c7585/core/dbt/events/functions.py#L149), and also an [ArtifactMetadata](https://github.com/dbt-labs/dbt-core/blob/ae294b643be3b825865a4ce3ff9fe47b523256c6/core/dbt/contracts/util.py#L172) class(I tried a bit to take it out but mypy was not happy about it).
2. refactor other tracking function so that it doesn't [require a runtime config](https://github.com/dbt-labs/dbt-core/blob/391e8bc1165eda31f0a9493d451a6fa3b9716db3/core/dbt/tracking.py#L363)
   - this is kind of straightforward, but since we haven't moved project, profiles object to the new framework yet, I am not sure how to test things.

Another thread we can try to pull is do some tracking validation, I don't think we have anything to validate certain events being emitted after certain invocation, we could try to build some tests validating the events for future code change.

Let me know which direction you want me to go!

core/dbt/cli/flags.py
```diff
@@ -32,13 +33,20 @@ def assign_params(ctx):
 
         assign_params(ctx)
 
+        # Get the invoked command flags
```
iknox-fa, line 36: FYI-- I reworked this to be [a lot cleaner and remove some potential edge cases.](https://github.com/dbt-labs/dbt-core/blob/c3ea0d1e01adc70d000375fb73b9897ffe45ec4d/core/dbt/cli/flags.py#L38)  No need to fix it now unless you want to since it's part of the logging branch I'm working on, but worth noting.

core/dbt/cli/main.py
```diff
@@ -6,7 +6,10 @@
 from dbt.adapters.factory import adapter_management
 from dbt.cli import params as p
 from dbt.cli.flags import Flags
+from dbt.events.functions import setup_event_logger
```
iknox-fa, line 9: Why are we pulling in logging stuff?

core/dbt/cli/main.py
```diff
@@ -52,17 +55,30 @@ def cli(ctx, **kwargs):
...
+
+    dbt.tracking.initialize_from_flags(flags)
+    # TODO we need to have config to get the projectID
+    project_id = ""
+    # TODO we need to get the credentials or we need to know that this will make adapter info
+    # not available in invocation start/end events
+    credentials = None
+    ctx.with_resource(dbt.tracking.track_run(project_id, credentials, ctx.invoked_subcommand))
+
+    # Logging
```
iknox-fa, line 68: I think you merged all the changes from my logging branch, not just the changes to flags.
:)

As above, feel free to leave it if you want, but it's all changed since you pulled it in from my branch.

core/dbt/cli/main.py
```diff
@@ -6,7 +6,10 @@
 from dbt.adapters.factory import adapter_management
 from dbt.cli import params as p
 from dbt.cli.flags import Flags
+from dbt.events.functions import setup_event_logger
 from dbt.profiler import profiler
+import dbt.tracking
```
iknox-fa, line 11: Let's use the [google styleguide for importing](https://google.github.io/styleguide/pyguide.html#22-imports) and import specific things from the module instead of the whole module.
`from dbt.tracking import initalize_from_flags, track_run`

core/dbt/cli/main.py
```diff
@@ -52,17 +55,30 @@ def cli(ctx, **kwargs):
     """An ELT tool for managing your SQL transformations and data models.
     For more documentation on these commands, visit: docs.getdbt.com
     """
-    incomplete_flags = Flags()
+    flags = Flags(invoked_subcommand=globals()[ctx.invoked_subcommand])
+
+    dbt.tracking.initialize_from_flags(flags)
```
iknox-fa, line 60: We're only using two flag values-- for better visibility can we pull those values out specifically?

`initalize_from_flags(
    flags.SEND_ANONYMOUS_USAGE_STATS,
    flags.PROFILES_DIR
)`

core/dbt/cli/main.py
ChenyuLInx: I think this is because I cherry picked your commit to be able to access those attributes in flags.

core/dbt/tracking.py
```diff
@@ -470,9 +435,51 @@ def process(self, record):
...
+        except Exception:
+            fire_event(TrackingInitializeFailure())
+            active_user = User(None)
     else:
-        do_not_track()
+        active_user = User(None)
+
+
+@contextmanager
+def track_run(project_id, credentials, run_command):
```
iknox-fa, line 453: So, I think in the long run we'll want to take a little of the logic out of this function.  I don't think we want to require adapter data or profile name in order to start `track_run`, instead I think we need to start with nothing but the `flags` and `active_user` and then add another tracking event or two later on to ensure we still track that data.. I'm picturing something like this:

_for this PR_
```
@contextmanager
def track_run(run_command):
    invocation_context = get_base_invocation_context()

    track_invocation_start(invocation_context)
    try:
        yield
        track_invocation_end(invocation_context, result_type="ok")
    except (NotImplementedException, FailedToConnectException) as e:
        fire_event(MainEncounteredError(exc=str(e)))
        track_invocation_end(invocation_context, result_type="error")
    except Exception:
        track_invocation_end(invocation_context, result_type="error")
        raise
    finally:
        flush()
``` 

_for a future PR_
_gets called whenever we actually need the adapters_
```
def track_adapter_info(options):
    context = [SelfDescribingJson(ADAPTER_INFO, options)]
    assert active_user is not None, "Cannot track adapter info when active user is None"

    track(
        active_user,
        category="dbt",
        action="adapter_info",
        label=get_invocation_id(),
        context=context,
    )
```


Obvs, this will need to be done in tandem with changes to the internal analytics project/iglu schemas and is work outside the scope of _this_ ticket, but I wanted

iknox-fa (changes requested):
A few small changes

iknox-fa (approved):
LGTM!

PR #5975, opened 2022-09-29
CT-1150: Include flat_graph in manifest.deepcopy(), tests

### LINKED ISSUE #5809: [CT-1150] Include `flat_graph` in `manifest.deepcopy()`
Why isn't `flat_graph` included in `manifest.deepcopy()`? Probably not for any reason in particular.

The original implementation of `manifest.deepcopy()` was the early days of the dbt-RPC server (https://github.com/dbt-labs/dbt-core/commit/1a0df174c9f252d0f93e03dbf4ef5005b2429072), and that's still the only place we really call it. The idea is to support dynamically adding new temporary nodes (arbitrary SQL compilation/execution) that might need to [`add_refs`](https://github.com/dbt-labs/dbt-rpc/blob/b45d09e487fedb1b12fd3214081e593d52332bb4/dbt_rpc/task/sql_commands.py#L30-L50) in the manifest — or rather, a full copy of the manifest.

This does mean, however, that any time you want to deepcopy of a manifest, and you want that copied manifest to support introspection of the `graph` variable in the Jinja context, you actually need an extra step:
```python
from dbt.contracts.graph.manifest import Manifest

manifest = Manifest(...)

# should just be this, right?
copy_of_manifest = manifest.deepcopy()

# still missing something
copy_of_manifest.build_flat_graph()
# now you're good to go
```

The fix here might be as simple as taking this:
https://github.com/dbt-labs/dbt-core/blob/b26280d1cf1af1c715753f73eb3b46f6eaa717f5/core/dbt/contracts/graph/manifest.py#L781-L794
And adding in:
```python
import copy
...

    def deepcopy(self):
        return Manifest(
            ...
            flat_graph={k: copy.deepcopy(v) for k, v in self.flat_graph.items()},
            ...
        )
```

peterallenwebb:
resolves #5809 

### Description

Adds the flat_graph attribute to set of Manifest class elements that are copied by deepcopy(), correcting what was presumably an oversight.

### Checklist

- [x] This PR includes tests

### DIFF
--- a/core/dbt/contracts/graph/manifest.py
+++ b/core/dbt/contracts/graph/manifest.py
@@ -780,7 +780,7 @@ class Manifest(MacroMethods, DataClassMessagePackMixin, dbtClassMixin):
         return frozenset(x.database for x in chain(self.nodes.values(), self.sources.values()))
 
     def deepcopy(self):
-        return Manifest(
+        copy = Manifest(
             nodes={k: _deepcopy(v) for k, v in self.nodes.items()},
             sources={k: _deepcopy(v) for k, v in self.sources.items()},
             macros={k: _deepcopy(v) for k, v in self.macros.items()},
@@ -793,6 +793,8 @@ class Manifest(MacroMethods, DataClassMessagePackMixin, dbtClassMixin):
             files={k: _deepcopy(v) for k, v in self.files.items()},
             state_check=_deepcopy(self.state_check),
         )
+        copy.build_flat_graph()
+        return copy
 
     def build_parent_and_child_maps(self):
         edge_members = list(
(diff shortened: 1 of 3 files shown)


core/dbt/contracts/graph/manifest.py
```diff
@@ -793,6 +793,8 @@ def deepcopy(self):
             files={k: _deepcopy(v) for k, v in self.files.items()},
             state_check=_deepcopy(self.state_check),
         )
+        copy.build_flat_graph()
```
gshank, line 796: I think instead of building this again, the idea was that we'd do a copy of the flat_graph dictionary, like the nodes/sources/macros/files/state_check above. So add a line like:  flat_graph={k: _deepcopy(v) for k, v in self.flat_graph.items()},

gshank (changes requested):

core/dbt/contracts/graph/manifest.py
gshank: On second thought, this is probably better than doing a deepcopy of the flat graph, which will make yet another copy of the nodes and is probably slower.

gshank (approved):

PR #5983, opened 2022-10-02
Make `mypy` pass from the command line 

### LINKED ISSUE #6029: [CT-1315] [Internal] Make mypy pass from the command line

### Describe the feature

Stub issue for https://github.com/dbt-labs/dbt-core/pull/5983. Contributing.md states it's required, and only my PRs with issues have been reviewed

_No response_

### Who will this benefit?

_No response_

_No response_

### Anything else?

_No response_

max-sixty:
Similar issue to https://github.com/dbt-labs/dbt-core/pull/5947 ~(which this includes, and so requires merging #5947 first)~ — currently running `mypy` from the command line fails, because the configs are in `.pre-commit-config.yaml`, which isn't available to editors etc.

### Description

closes #6029 

### Checklist


lostmygithubaccount:
closes #6029

pyproject.toml
```diff
@@ -1,3 +1,9 @@
+[tool.mypy]
+exclude = ['core/build/.*']
+files = 'core'
```
emmyoop, line 3: This is actually less restrictive than now.  We currently restrict to `core/dbt` in `.pre-commit-config.yaml`.  If we alter this to only look there, you no longer need the exclude.

Would it be possible to only define the `files` key here and remove it from `.pre-commit-config.yaml` so we don't have it defined in multiple places for mypy?

emmyoop (changes requested):

pyproject.toml
max-sixty: > Would it be possible to only define the `files` key here and remove it from `.pre-commit-config.yaml` so we don't have it defined in multiple places for mypy?

Yeah, annoyingly mypy doesn't have a notion of "force exclude", where it excludes something even if explicitly passed it on the command line. And pre-commit passes file names. So it's still necessary to have the duplication unfortunately

max-sixty: > This is actually less restrictive than now. We currently restrict to `core/dbt` in `.pre-commit-config.yaml`. If we alter this to only look there, you no longer need the exclude.

Nice, done

emmyoop (approved):
Thanks @max-sixty!  Really appreciate the contribution here!

PR #5999, opened 2022-10-03
Allow custom metadata to be attached to every log

### LINKED ISSUE #5228: [CT-624] [Feature] Allow for custom metadata to be attached to logs

### Describe the Feature

I run my dbt jobs using [Argo Workflows](https://argoproj.github.io/argo-workflows/), then I use [Vector](https://vector.dev/) to grab those logs from stdout and dump them into my logging provider. Argo and Vector don't make it super easy to set and attach something like the Argo Workflow step name and / or k8s pod name to the log line on a per-step basis.

It'd be great if `dbt` had functionality where you could set custom metadata properties (and their values) in each log line. Presumably, these could be both raw values as well as Jinja templateized values, from the project config. Something like this would be ideal:

```shell
dbt --log-format json --log-properties '{ "job_name": "my_job_name", "domain_name": "{{ project_name }}" } run'
```

It's possible this kind of thing already exists, but I couldn't easily find it. Thanks in advance for the consideration!

_No response_

### Who will this benefit?

Anyone needing to run dbt jobs in a DAG-like scheduler that needs to log to some APM / Logging provider.

_No response_

### Anything else?

_No response_

emmyoop:
resolves #5228 

### Description

Adds all env vars prefixed with the `METADATA_ENV_PREFIX` to all logs.

### Checklist


core/dbt/constants.py
```diff
@@ -1,2 +1,3 @@
 SECRET_ENV_PREFIX = "DBT_ENV_SECRET_"
 DEFAULT_ENV_PLACEHOLDER = "DBT_DEFAULT_PLACEHOLDER"
+METADATA_ENV_PREFIX = "DBT_ENV_CUSTOM_ENV_"
```
peterallenwebb, line 3: This name might be natural in terms of some context I don't have yet, but I wonder if it can be improved. Maybe DBT_LOGGING_VAR_ or DBT_METADATA_VAR_?

emmyoop: This is actually an existing name ([docs](https://docs.getdbt.com/reference/artifacts/dbt-artifacts#:~:text=env%3A%20Any%20environment%20variables%20prefixed%20with%20DBT_ENV_CUSTOM_ENV_%20will%20be%20included%20in%20a%20dictionary%2C%20with%20the%20prefix%2Dstripped%20variable%20name%20as%20its%20key.)).  Metadata is in the manifest and these vars end up there currently.  I centralized the constant into this (newish) file as part of my changes.

gshank: It would be my preference to have a new key for this behavior, such as DBT_LOG_CUSTOM_ENV.  Users already have existing DBT_ENV_CUSTOM_ENV environment variables, which they might actually be surprised to find in every log line. @jtcohen6 could you weigh in on this?

To me, this seems to be a different use case than we already have for DBT_ENV_CUSTOM_ENV.

emmyoop: I'm not sure specifically how users are using `DBT_ENV_CUSTOM_ENV` currently.

I do see what you're saying with users possibly being surprised these env vars show up in logs now.  But is there a use case for these being the exact variables they want to show up?

gshank: Right now those DBT_ENV_CUSTOM_ENV variables show up in the artifacts that are written up in the end, and there was a recent change to put them in the context for access in Jinja.

I'm not familiar with how users might use those, but from some of our test cases, it does look like they were intended for putting things like a run_id or a job_id. So maybe it would be fine to use this prefix string.

gshank:
In general, this looks fine though. We are in the process of an initiative to move global things to be non-global, but since the good places to put this particular global haven't been defined yet, making it global for now is probably fine.

Does need tests :)

core/dbt/constants.py
jtcohen6: Thanks for tagging me in, and for the good conversation here!

Users set these env vars with the explicit intent of having them show up in metadata artifacts. It makes it easier to debug and quickly cross-reference. I see logs as a metadata artifact, in very much the same vein.

This was also our intent by crafting such an unwieldy prefix. It's hard to type `DBT_ENV_CUSTOM_ENV_` by accident. I'm definitely open to a better name (`DBT_ENV_CUSTOM_METADATA_`?) that's equally clear about what it's doing. By convention, "functional" env vars should just be prefixed with `DBT_`, including env vars setting global configs.

I think it's an okay assumption to automatically coordinate between these. If later on, we hear a valid use case for having them show up in one, and not the other, I'm willing to admit that I was wrong! But I'd rather do this than require users to set two sets of identical env vars from the get-go.

emmyoop: Thanks for the input @jtcohen6!  I do think `DBT_ENV_CUSTOM_METADATA_` does a better job of saying what the var is for sure, so lean towards this new name.

My understanding is that if we change the name we should add a deprecation warning for the old name and remove it in 2.0.  My question is would `DBT_ENV_CUSTOM_ENV_` only show up in the artifacts (since that's maintaining current behavior) and `DBT_ENV_CUSTOM_METADATA_` shows up in both logs and artifacts?

jtcohen6: I'd still vote in favor of coordinating the information so that it shows up in both places.

How would we feel about implementing this as `DBT_ENV_CUSTOM_ENV_` for now, and then opening a new ticket for the rename (+ deprecation warning)?

core/dbt/events/base_types.py
```diff
@@ -3,18 +3,62 @@
...
+import functools
+import collections
 
 # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # #
 # These base types define the _required structure_ for the concrete event #
 # types defined in types.py                                               #
 # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # #
 
 
+# TODO: this should not live here - duplicated and modified from utils.py because of circular imports for now
```
emmyoop, line 15: This is here temporarily to test out if I could get it working.  Importing the one in `util.py` causes some circular dependency issues.

PR #6025, opened 2022-10-07
CT-625: Fail with clear message for invalid materialized vals

peterallenwebb:
resolves #5229 

### Description

Parsing now fails and provides a clear error message if an inappropriate value of 'materialized' is set for seeds, snapshots, or tests.

### Checklist

- [x] I have read [the contributing guide](https://github.com/dbt-labs/dbt-core/blob/main/CONTRIBUTING.md) and understand what's expected of me
- [x] I have signed the [CLA](https://docs.getdbt.com/docs/contributor-license-agreements)
- [x] I have run this code in development and it appears to resolve the stated issue
- [x] This PR includes tests, or tests are not required/relevant for this PR
- [x] I have [opened an issue to add/update docs](https://github.com/dbt-labs/docs.getdbt.com/issues/new/choose), or docs changes are not required/relevant for this PR
- [x] I have run `changie new` to [create a changelog entry](https://github.com/dbt-labs/dbt-core/blob/main/CONTRIBUTING.md#adding-a-changelog-entry)

PR #6038, opened 2022-10-10
Validate enabled config is a bool for sources, exposures and metrics

### LINKED ISSUE #6030: [CT-1317] Validate source config enabled flag
There is no validation of the "enabled" config flag for sources, whether in dbt_project.yml or in a schema file. This means that people can set the flag to strange things such as "False and False" and it will be saved with no error message.

emmyoop:
resolves #6030 

### Description

Minor test cleanup to meet new standards

Validate source, exposure and metric enabled config is a bool.

### Checklist


gshank (changes requested):
We have a pattern of doing validation in a 'validate' method, and this should be consistent. (See #6025).  We should already have the standard hologram validation available for these configs. I think the problem is that the other schema configs do validation in "update_parsed_node_config', and that's not called for schema-file only configs. Instead of adding a post_init validation, I think we should be calling "validate" on the configs when they're generated in parsing, such as after _generate_exposure_config, _generate_metric_config are called, and some similar place for sources.

gshank (changes requested):
The validation is in the wrong places. There is separate validation for nodes and code configs. We need to validate the node configs. That needs to be done after the dbt_project configs are merged in, otherwise the configs set in dbt_project won't be validated. 

The hologram validate methods for MetricConfig, ExposureConfig, and SourceConfig will already be validating that enabled is a bool, so all that's necessary is to call the Config.validate(...) method in the right places. In addition if we actually call the config's validate method, if we add other config attributes the validation will already be in place. I think you might only need to add a config = config.finalize_and_validate() call after the _generate_metric_config and _generate_exposure_config calls.

gshank (approved):
Looks good!

Issue #5958, opened 2022-09-28
[CT-1268] [Spike] Integrate dbt exceptions into structured logging system


gshank:
Right now we have structured logging events for the majority of our logging, except for exceptions. In some but not all cases we have a separate logging event. Ideally all exceptions should be logged automatically. This ticket is labeled a "spike" because in order to scope this work we need to identify a mechanism/policies for doing this.

jtcohen6:
Copying from Nate's comment in https://github.com/dbt-labs/dbt-core/issues/4836#issuecomment-1104038063, because the underlying thrust of it continues to feel important:

> whenever possible, we want to be able to programmatically distinguish when the root cause of an error is a warehouse issue, an adapter issue, or a core issue. This will be a big improvement in the process for debugging failed runs.

Basically, when a dbt run fails, how can we tell whether it was user error, database error, or internal dbt error?

In a Pythonic world, we'd expect to do this via class inheritance and type introspection. Given that our log events are now proto-composed, we probably need a field on each exception-log. Something like `exception_cause` or `exception_category`. Categories I can imagine:
- parsing / validation error (problem in user code)
- disallowed behavior, e.g. using deprecated functionality or a version of `dbt-core` that's disallowed by `require-dbt-version`
- database error due to connection timeout
- database error due to malformed query
- internal / unhandled error
- ... more? ...

emmyoop:
This can probably be accomplished with an enum that defines all the categories.  Without verifying the syntax I think something like this would work:

```proto
enum ExceptionCause {
{
        PARSING = 0;
        DISALLOWED = 1;
        DATABASE_CONNECTION = 2;
        DATABASE_QUERY = 3;
        UNHANDLED = 4;
        ...
 }
```

And then within the message add
 ```proto
message SomeException {
    EventInfo info = 1;
    repeated ExceptionCause exception_cause = 2;
    ...
}
```

This also allows the flexibility to add more types to the end of the enum which I'm sure we'll end up needing.

jtcohen6:
Blocking question we need to answer: Should we have separate `LogCategory` versus `ExceptionCategory`, or combine the two into a single field?

(Idea with `LogCategory` is to formally serve the role that the event codes' starting letters have been informally playing: this is parsing-related, deps-related, adapter-related, ...)

emmyoop:
[...]

PR #6064, opened 2022-10-14
Convert messages to struct logs

### LINKED ISSUE #5957: [CT-1267] Convert non-structured messages generated in exceptions.py to structured events
We have a number of logging events in which the string message is generated by external utilities, mostly in various functions in exceptions.py.  Because of this, the actual information used to structure the message is not included in the logging event. These events can be identified by the fact that they contain a "msg" field instead of various structured fields.

Refactor these messages so that more of the formatting code is moved into the logging event code, and so that the events capture the strings used to construct the message.

emmyoop:
resolves #5957 

### Description

Convert msg to be built in struct logs

### Checklist


core/dbt/events/helpers.py
```diff
@@ -0,0 +1,16 @@
...
+    return [v for k, v in os.environ.items() if k.startswith(SECRET_ENV_PREFIX) and v.strip()]
+
+
+def scrub_secrets(msg: str, secrets: List[str]) -> str:
+    scrubbed = msg
+
+    for secret in secrets:
+        scrubbed = scrubbed.replace(secret, "*****")
+
+    return scrubbed
```
emmyoop, line 16: I'm not sure I love what I did here.  These were moved out because of some circular dependency issues.  I'm not sure they even belong in the `/events` dir since they're used in various places.

emmyoop:
Happy to wait on this until after the stuct logging changes are merged in main - can change the target for the PR.

gshank (changes requested):
This is a really good start. But I'd like us to handle the "warn_or_error" calls somewhat differently, and convert as many of the "GeneralWarningMsg" instances to specific events instead. So warn_or_error might look something like:
```
def warn_or_error(event, node):
    if flags.WARN_ERROR:
        from dbt.exceptions import raise_compiler_error

        raise_compiler_error(scrub_secrets(event.info.msg, env_secrets()), node)
    else:
        fire_event(event)
```

We're also going to be doing some work on handling and logging exceptions better too, so that "raise_compiler_error" is also likely to see some changes.

One thing people who consume our logs is asking for is being able to categorize the log events and exceptions, and that GeneralWarningMsg does not help.

core/dbt/events/functions.py
```diff
@@ -220,6 +215,15 @@ def send_to_logger(l: Union[Logger, logbook.Logger], level_tag: str, log_line: s
         )
 
 
+def warn_or_error(event, node=None):
+    if flags.WARN_ERROR:
+        from dbt.exceptions import raise_compiler_error
+
+        raise_compiler_error(scrub_secrets(event.info.msg, env_secrets()), node)
```
peterallenwebb, line 222: I know this code has been moved and isn't new, but should we perhaps fire the event on both branches? Otherwise, I would guess the exception won't be logged.

core/dbt/deprecations.py
```diff
@@ -21,66 +21,50 @@ def track_deprecation_warn(self) -> None:
...
             active_deprecations.add(self.name)
 
 
 class PackageRedirectDeprecation(DBTDeprecation):
     _name = "package-redirect"
-    _description = """\
-    The `{old_name}` package is deprecated in favor of `{new_name}`. Please update
-    your `packages.yml` configuration to use `{new_name}` instead.
-    """
+    _event = "PackageRedirectDeprecation"
```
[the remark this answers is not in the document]
emmyoop: The description got moved into the event message.  It lives in `types.py` now.

core/dbt/events/helpers.py
gshank: Yeah, but I can't think of a great place either. If a superior location emerges, we can move it.

gshank (approved):
This looks really good!  It's so big it's possible I missed something, but I didn't see any issues.

core/dbt/adapters/base/impl.py
```diff
@@ -1327,7 +1327,7 @@ def catch_as_completed(
         elif isinstance(exc, KeyboardInterrupt) or not isinstance(exc, Exception):
             raise exc
         else:
-            warn_or_error(f"Encountered an error while generating catalog: {str(exc)}")
+            warn_or_error(CatalogGenerationError(exc=str(exc)))
```
VersusFacit, line 1330: Nit: this paradigm of stringifying arguments to event types now has the option of adding [this class](https://github.com/dbt-labs/dbt-core/blob/8145eed603266951ce35858f7eef3836012090bd/core/dbt/events/base_types.py#L112) as a subclass to the event class definition.

core/dbt/events/base_types.py
```diff
@@ -3,6 +3,7 @@
 import threading
 from datetime import datetime
 
+
```
VersusFacit, line 6: nit: extra whitespace or demanded by flake8?

core/dbt/events/functions.py
VersusFacit: We comfy with importing the exception down yonder inline rather than at the top? Do we have a style convention here?

core/dbt/config/runtime.py
```diff
@@ -16,17 +16,16 @@
...
 from dbt.exceptions import (
     RuntimeException,
     DbtProjectError,
     validator_error_message,
-    warn_or_error,
     raise_compiler_error,
 )
-
+from dbt.events.functions import warn_or_error
+from dbt.events.types import UnsedResourceConfigPath
```
jtcohen6, line 28: very cool PR! caught one typo while reading through
```suggestion
UnusedResourceConfigPath
```

core/dbt/events/functions.py
emmyoop: Doing it here actually prevents a circular dependency.  My next task #5958 should help resolve this to a better solution.

core/dbt/adapters/base/impl.py
emmyoop: That class is super cool.  However, it is causing some issues because protobuf requires us to define the type of each field being passed in in a very narrow scope (types cannot be classes ie Exception).  So `exc` comes in as an `Exception` and then does become a string in the message but that's not enough since proto expects the field itself to be a string already.

core/dbt/config/runtime.py
emmyoop: Good eye!

PR #6070, opened 2022-10-14
added optional incremental_key config

### LINKED ISSUE #6069: [CT-1351] [Feature] Add optional `incremental_key` config to incremental models

### Describe the feature

I would like the ability to pass in a column (or SQL expression string) to incremental models that specifies an incremental column to use for incremental processing.

This follows the pattern of the optional `unique_key` config.

This change would remove the need to repeatedly add the same `is_incremental()` check by hand. 

So instead of:
```sql
{{
  config(
    materialized='incremental',
    unique_key='id'
 )
}}

select 
  id, 
  ingestion_time 
from {{ref('my_parent')}}

{% if is_incremental() %}
  -- this filter will only be applied on an incremental run
  WHERE {{ incremental_key }} >= (SELECT MAX(THIS.{{ incremental_key }}) FROM {{ this }} THIS)
{% endif %}

```

A user can write:
```sql
{{
  config(
    materialized='incremental',
    unique_key='id',
    incremental_key='ingestion_time'
 )
}}

select 
  id, 
  ingestion_time 
from {{ref('my_parent')}}
```

The obvious alternative is creating this as a custom materialization, maybe throw it in a library for folks to use.

I think this is such a common pattern though that it deserves to be built in to incrementals.

### Who will this benefit?

Users that process incremental data in relatively simple ways.

Yes! Will PR next : )

### Anything else?

I know I brought this up once, but I can't find if that ever made its way to a github issue / PR.

I have ensured that users can still use is_incremental checks simply by NOT providing an incremental_key config.

randypitcherii:
resolves #6069 

### Description
This change only allows users to supply an optional `incremental_key` config to incremental models

### Checklist

### DIFF
--- a/core/dbt/include/global_project/macros/materializations/models/incremental/incremental.sql
+++ b/core/dbt/include/global_project/macros/materializations/models/incremental/incremental.sql
@@ -11,6 +11,7 @@
 
   -- configs
   {%- set unique_key = config.get('unique_key') -%}
+  {%- set incremental_key  = config.get('incremental_key') -%}
   {%- set full_refresh_mode = (should_full_refresh()  or existing_relation.is_view) -%}
   {%- set on_schema_change = incremental_validate_on_schema_change(config.get('on_schema_change'), default='ignore') -%}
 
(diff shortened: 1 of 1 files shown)


core/dbt/include/global_project/macros/materializations/models/incremental/incremental.sql
```diff
@@ -34,11 +35,23 @@
...
+
   {% else %}
-    {% do run_query(get_create_table_as_sql(True, temp_relation, sql)) %}
+    {% set build_sql %}
+      {{sql}} 
+
+      {#-- If an incremental key was provided, append the incremental filtering logic to the model sql --#}
+      {% if incremental_key is not none %}
+        -- this filter will only be applied on an incremental run
+        WHERE {{ incremental_key }} >= (SELECT MAX(THIS.{{ incremental_key }}) FROM {{ this }} THIS)
```
owenprough-sift, line 50: You include the `WHERE` so `incremental_key` would only work for models which have no other filters (otherwise you'd end up with multiple `WHERE` clauses, which is a syntax error). Is there a way to more gracefully handle that situation, or is `incremental_key` intended only for the simplest of use cases?

owenprough-sift, line 50: Is it safe/appropriate to assume that the `incremental_key` field has the same name in the source and destination?

jtcohen6:
Thanks for taking initiative on this one @randypitcherii! Code shows like prose could only hope.

I agree with @owenprough-sift's comments. This approach makes a number of simplifying assumptions that, while keeping it easy for end users, could go wrong. The question is whether those assumptions hold true in a large enough % of cases (50? 80?) to justify built-in support for the happiest path, at the risk of some obfuscation/confusion when things get more complicated.

Let's keep the convo going in https://github.com/dbt-labs/dbt-core/issues/6069. In the interest of housekeeping, I'm going to close this PR in the meantime.

Issue #6067, opened 2022-10-14
[CT-1349] [Feature] Dynamic sources


yang6n:
### Is this your first time submitting a feature request?

- [X] I have read the [expectations for open source contributors](https://docs.getdbt.com/docs/contributing/oss-expectations)
- [X] I have searched the existing issues, and I could not find an existing issue for this feature
- [X] I am requesting a straightforward extension of existing dbt functionality, rather than a Big Idea better suited to a discussion

### Describe the feature

I love the lineage that DBT provided to us for free when using `source()`, but we have to define the source schema explicitly. 
As a Developer, I want a way to define the sources by:

- Specify the schema if not default,
- Include or exclude a list with wild card support to define the source tables.

```
version: 2

sources:
  - name: dbt_alice
    schema: dbt_alice
    include_tables: src_*
    or 
    exclude_tables: stg_*_tmp
```

### Describe alternatives you've considered

Use the codegen package to generating the source YML file before `dbt run`
https://github.com/dbt-labs/dbt-codegen/tree/0.8.1/#generate_source-source
generate

### Who will this benefit?

For developers they want lineage but want to include ALL tables as the source, they don't need to define or generate the source.yml

### Are you interested in contributing this feature?

I'd love to contribute if someone point me to the right piece of code and provide some guidance on the source code.

yang6n:
Well, the alternatives are not gonna work because if we don't have the source.yml but the models are using `source()`, then it will fail to generate the source.

yang6n:
Draft PR - https://github.com/dbt-labs/dbt-core/pull/6072

lostmygithubaccount:
thanks for the issue and potential contribution @yang6n! we've been a bit busy with our Coalesce conference this past week, and many of us on the Core team (myself and @jtcohen6 included) are still traveling this week. we'll try to get to this shortly -- you might also consider posting this PR and asking for feedback here in the meantime: https://getdbt.slack.com/archives/C50NEBJGG

jtcohen6:
@yang6n Thanks for opening up the issue, and sorry for the delay in getting back to you!

There's a hard line we draw in dbt today, and don't wish to soon erase: **dbt must know the full DAG before it connects to the database.** This is a useful abstraction and simplification for us. We can keep "control plane" code (DAG definitions) clearly separate from "data plane" code (tabular queries / dataframe manipulations). Every dbt resource maps to exactly one database object (or zero, in the case of ephemeral models, analyses, etc).

[...]

PR #6072, opened 2022-10-15
CT-1349, create dynamic sources

yang6n:
resolves CT-1349(https://github.com/dbt-labs/dbt-core/issues/6067)

### Description

Creates a sub-type of "sources" named "dynamic_sources", user can define the source by specifying the following:
- The schema
- The regex for tables includes pattern
- The regex for tables excludes pattern
i.e.
```dynamic_sources:
  - name: dbt_alice
    schema: dbt_alice
    tables_exclude_pattern: raw_\S+_tmp
    tables_include_pattern: raw_\S+
```

### Checklist

### DIFF
--- a/core/dbt/parser/schemas.py
+++ b/core/dbt/parser/schemas.py
@@ -11,6 +11,7 @@ from dbt.contracts.graph.unparsed import UnparsedSourceTableDefinition
 from dbt.dataclass_schema import ValidationError, dbtClassMixin
 
 from dbt.adapters.factory import get_adapter, get_adapter_package_names
+from dbt.adapters.base import BaseRelation
 from dbt.clients.jinja import get_rendered, add_rendered_test_kwargs
 from dbt.clients.yaml_helper import load_yaml_text
 from dbt.parser.schema_renderer import SchemaYamlRenderer
@@ -80,6 +81,7 @@ from dbt.parser.generic_test_builders import (
 from dbt.ui import warning_tag
 from dbt.utils import get_pseudo_test_path, coerce_dict_str
 
+import re
 
 TestDef = Union[str, Dict[str, Any]]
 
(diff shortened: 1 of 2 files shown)


core/dbt/parser/schemas.py
```diff
@@ -737,6 +746,50 @@ def add_source_definitions(self, source: UnparsedSourceDefinition) -> None:
...
+        include_pattern = re.compile(source.tables_include_pattern)
+        
+        tables = []
+        adapter = get_adapter(self.root_project)
+        with adapter.connection_named("dynamic_sources"):
+            schema_relation = BaseRelation.create(
+                # database = self.root_project.default_database,
+                schema = source.schema
+            )
+            relations = adapter.list_relations_without_caching(schema_relation)
```
jtcohen6, line 766: Hard rule in dbt: We shouldn't be making database connections at parse time. I said more about this over in https://github.com/dbt-labs/dbt-core/issues/6067#issuecomment-1296516720:

> There's a hard line we draw in dbt today, and don't wish to soon erase: dbt must know the full DAG before it connects to the database. This is a useful abstraction and simplification for us. We can keep "control plane" code (DAG definitions) clearly separate from "data plane" code (tabular queries / dataframe manipulations). Every dbt resource maps to exactly one database object (or zero, in the case of ephemeral models, analyses, etc).

> For that reason, it's essential that we don't require dbt to make a database connection — even if it's just a caching query — to enumerate the set of available sources.

jtcohen6 (commented):
@yang6n Really appreciative of the hard work you put into this! It's one thing to write up an issue with a Big Idea, and another entirely to actually dive into the code to get something working. That's all the more true when it's code related to project parsing in `dbt-core`, which is gnarly at points.

I left a lengthier explanation in the linked issue of why I don't think we'll be able to move forward with the current proposal. Let's keep discussing the bigger ideas around "dynamic sources" over in https://github.com/dbt-labs/dbt-core/discussions/5686!

PR #6101, opened 2022-10-19
Create --macro-state-in-respective-node option

### LINKED ISSUE #5922: [CT-1234] [Feature] State-based selection able to handle models that refresh in different schedules or jobs

### Describe the feature

**Background:**
When we have models that run in different schedules (like hourly, daily, weekly) then manifest.json from any of those jobs on its own won't represent current state of database. Two different teams using same dbt project may want to setup separate jobs for better monitoring and alerting even thought the schedule may be the same. In these scenario, we can't use SlimCI with `state:modified` selector because not all models are run in same job.

**Approach:**
Since the structure of manifest.json is well document and easily parsed we can consolidate those manifests from different jobs and create single manifest file that represents current state of the database. But the problem appears if those models are using macros. If an hourly model and a weekly model is referencing same macro, and the macro sql changes, then while using `state:modified` selector, single definition of the macro won't give the correct picture. For this scenario, we need to store the macro definition within each of the models. Hourly run model will soon execute with latest macro definition while weekly model might take a week to come to latest state (unless someone manually refreshes those weekly models sooner).

On a side note: How those different manifest are consolidated might be a separate topic and might not be in scope for dbt-core because different companies may be running dbt differently (unless dbt-core defines a single location where dbt object's current state is tracked as they are run/build in database). In dbt Cloud, it could keep the latest manifest.json from each jobs, and combine the nodes based on unique_id and which one was executed latest by looking at respective run_result.json. For someone using external scheduler like Airflow and running each model as a separate tasks in a dag, they can use macro and post_hook to store the model and macro definition in a database table as they are executed (I found this straight forward rather than parsing run_result and manifest for each model). When running CI, they can run a script to get latest definition for each unique_id and generate a manifest file to be used with `--state` argument.

**Proposal:**
Add an optional field in ParsedNode called `dependent_macro_state` that will store the model's dependent macro's definitions. Since we don't want to change the default behavior, create a new cli parameter called `--macro-state-in-respective-node`. When `--state` parameter is used with this new parameter, the `--select state:modified` selector will read the macro_sql from `dependent_macro_state` field instead of the existing `macros` field in manifest.

manifest.json for state file would look like this with `dependent_macro_state` field. It will have complete definition of each macro used in the model, extracted recursively. **dependent_macro_state field will be empty in the manifest.json generated automatically by dbt commands; it's upon the user to populate this field correctly for state selector** *Showing only fields required for current discussion:*

```json
{
    "metadata": {},
    "nodes": {
        "model.my_analytics.model_a": {
            "config": {},
            "raw_sql": "some sql",
            "depends_on": {
                "macros": [
                    "macro.my_analytics.macro_1"
                ],
                "nodes": []
            },
            "dependent_macro_state": {
                "macro.my_analytics.macro_1": {
                    "macro_sql": "{% macro macro_1() %} some code {% endmacro %}",
                    "depends_on": {
                        "macros": [
                            "macro.my_analytics.macro_2"
                        ]
                    },
                    "name": "macro_1",
                    "unique_id": "macro.my_analytics.macro_1"
                },
                "macro.my_analytics.macro_2": {
                    "macro_sql": "{% macro macro_2() %} some code {% endmacro %}",
                    "depends_on": {
                        "macros": []
                    },
                    "name": "macro_2",
                    "unique_id": "macro.my_analytics.macro_2"
                },
            },
            "unique_id": "model.my_analytics.model_a",
        }
    }
}
```

Currently we are doing an workaround while setting up jobs so that there is one Master job that includes all models. This job is being used for state:modified selector for slim CI. Master job runs once a day and includes all model that need to run daily, hourly and weekly. We have a gap in hourly job schedule when Master job starts so that same models doesn't get executed in parallel and cause inconsistency. For weekly job, we are thinking of adding some checks/filters so that the query doesn't do anything when run daily outside its weekly schedule.

### Who will this benefit?

Anyone who have models running in different schedule or trying to execute models in groups in separate jobs.

Yes, I have an working example of what I have proposed

### Anything else?

_No response_

karunpoudel:
resolves #5922

### Description
Add an optional field in ParsedNode called dependent_macro_state that will be referenced for the model's dependent macro's definitions when --state parameter is used with new parameter --macro-state-in-respective-node. The --select state:modified selector will read the macro_sql from dependent_macro_state field instead of the existing macros field in manifest when parameter --macro-state-in-respective-node is provided

### Checklist

### DIFF
--- a/.changes/unreleased/Features-20221018-221612.yaml
+++ b/.changes/unreleased/Features-20221018-221612.yaml
@@ -1,0 +1,7 @@
+kind: Features
+body: Option for --macro-state-in-respective-node
+time: 2022-10-18T22:16:12.4911753-05:00
+custom:
+  Author: karunpoudel
+  Issue: "5922"
+  PR: "6101"
--- a/core/dbt/contracts/graph/parsed.py
+++ b/core/dbt/contracts/graph/parsed.py
@@ -198,12 +198,68 @@ class NodeInfoMixin:
 
 
 @dataclass
+class ParsedPatch(HasYamlMetadata, Replaceable):
+    name: str
+    description: str
+    meta: Dict[str, Any]
+    docs: Docs
+    config: Dict[str, Any]
+
+
+# The parsed node update is only the 'patch', not the test. The test became a
+# regular parsed node. Note that description and columns must be present, but
+# may be empty.
+@dataclass
+class ParsedNodePatch(ParsedPatch):
+    columns: Dict[str, ColumnInfo]
+
+
+@dataclass
+class ParsedMacroPatch(ParsedPatch):
+    arguments: List[MacroArgument] = field(default_factory=list)
+
+
+@dataclass
+class ParsedMacro(UnparsedBaseNode, HasUniqueID):
+    name: str
+    macro_sql: str
+    resource_type: NodeType = field(metadata={"restrict": [NodeType.Macro]})
+    # TODO: can macros even have tags?
+    tags: List[str] = field(default_factory=list)
+    # TODO: is this ever populated?
+    depends_on: MacroDependsOn = field(default_factory=MacroDependsOn)
+    description: str = ""
+    meta: Dict[str, Any] = field(default_factory=dict)
+    docs: Docs = field(default_factory=Docs)
+    patch_path: Optional[str] = None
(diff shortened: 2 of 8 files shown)


gshank:
Closing because ticket has been closed as stale.

PR #6142, opened 2022-10-25
Project working with Click

ChenyuLInx:
resolves #

### Checklist

### DIFF
--- a/core/dbt/config/runtime.py
+++ b/core/dbt/config/runtime.py
@@ -13,6 +13,7 @@ from dbt import flags
 from dbt.adapters.factory import get_relation_class_by_name, get_include_paths
 from dbt.helper_types import FQNPath, PathSet, DictDefaultEmptyStr
 from dbt.config.profile import read_user_config
+from dbt.config.project import load_raw_project
 from dbt.contracts.connection import AdapterRequiredConfig, Credentials
 from dbt.contracts.graph.manifest import ManifestMetadata
 from dbt.contracts.relation import ComponentName
@@ -189,24 +190,32 @@ class RuntimeConfig(Project, Profile, AdapterRequiredConfig):
         return Profile.render_from_args(args, profile_renderer, profile_name)
 
     @classmethod
-    def collect_parts(cls: Type["RuntimeConfig"], args: Any) -> Tuple[Project, Profile]:
-        # profile_name from the project
-        project_root = args.project_dir if args.project_dir else os.getcwd()
-        version_check = bool(flags.VERSION_CHECK)
-        partial = Project.partial_load(project_root, verify_version=version_check)
-
+    def get_profile(
+        cls: Type["RuntimeConfig"], args: Any, cli_vars: Dict[str, Any], raw_profile_name: str
+    ):
         # build the profile using the base renderer and the one fact we know
         # Note: only the named profile section is rendered. The rest of the
         # profile is ignored.
-        cli_vars: Dict[str, Any] = parse_cli_vars(getattr(args, "vars", "{}"))
         profile_renderer = ProfileRenderer(cli_vars)
-        profile_name = partial.render_profile_name(profile_renderer)
+        profile_name = profile_renderer.render_value(raw_profile_name)
         profile = cls._get_rendered_profile(args, profile_renderer, profile_name)
         # Save env_vars encountered in rendering for partial parsing
         profile.profile_env_vars = profile_renderer.ctx_obj.env_vars
+        return profile
 
-        # get a new renderer using our target information and render the
-        # project
+    @classmethod
+    def collect_parts(cls: Type["RuntimeConfig"], args: Any) -> Tuple[Project, Profile]:
+        # profile_name from the project
+        project_root = args.project_dir if args.project_dir else os.getcwd()
+        raw_project = load_raw_project(project_root)
+        raw_profile_name: str = raw_project.get("profile")  # type: ignore
+        cli_vars: Dict[str, Any] = parse_cli_vars(getattr(args, "vars", "{}"))
+
+        profile = cls.get_profile(args, cli_vars, raw_profile_name)
+
+        # get the project with all of the provided information
+        version_check = bool(flags.VERSION_CHECK)
+        partial = Project.partial_load(project_root, verify_version=version_check)
         project_renderer = DbtProjectYamlRenderer(profile, cli_vars)
         project = partial.render(project_renderer)
         # Save env_vars encountered in rendering for partial parsing
--- a/core/dbt/config/profile.py
+++ b/core/dbt/config/profile.py
@@ -421,8 +421,9 @@ class Profile(HasCredentials):
         """
         threads_override = getattr(args, "threads", None)
         target_override = getattr(args, "target", None)
+        profile_name_override = getattr(args, "profile", None)
         raw_profiles = read_profile(flags.PROFILES_DIR)
-        profile_name = cls.pick_profile_name(getattr(args, "profile", None), project_profile_name)
+        profile_name = cls.pick_profile_name(profile_name_override, project_profile_name)
         return cls.from_raw_profiles(
             raw_profiles=raw_profiles,
             profile_name=profile_name,
--- a/core/dbt/config/project.py
+++ b/core/dbt/config/project.py
@@ -156,7 +156,7 @@ def value_or(value: Optional[T], default: T) -> T:
         return value
 
 
-def _raw_project_from(project_root: str) -> Dict[str, Any]:
+def load_raw_project(project_root: str) -> Dict[str, Any]:
 
     project_root = os.path.normpath(project_root)
     project_yaml_filepath = os.path.join(project_root, "dbt_project.yml")
@@ -485,7 +485,7 @@ class PartialProject(RenderComponents):
         cls, project_root: str, *, verify_version: bool = False
     ) -> "PartialProject":
         project_root = os.path.normpath(project_root)
-        project_dict = _raw_project_from(project_root)
+        project_dict = load_raw_project(project_root)
         config_version = project_dict.get("config-version", 1)
         if config_version != 2:
             raise DbtProjectError(


core/dbt/config/renderer.py
```diff
@@ -107,7 +107,7 @@ def __init__(
         if cli_vars is None:
             cli_vars = {}
         if profile:
```
ChenyuLInx, line 109: Still need to pass in profile here since this renderer is being used in integration tests all around the place(dbt-databricks even), hard to do the update.

core/dbt/cli/main.py
```diff
@@ -73,6 +76,14 @@ def cli(ctx, **kwargs):
     if flags.RECORD_TIMING_INFO:
         ctx.with_resource(profiler(enable=True, outfile=flags.RECORD_TIMING_INFO))
 
+    project_root = flags.PROJECT_DIR if flags.PROJECT_DIR else os.getcwd()
+    cli_vars = parse_cli_vars(getattr(flags, "vars", "{}"))
+
+    # TODO need profile to exisit
+    profile = None
```
ChenyuLInx, line 83: @iknox-fa this dependency not removable, because of the need for target in profile. And due to the other comment we still need to pass in the whole profile object to load project

core/dbt/cli/main.py
```diff
@@ -73,6 +76,14 @@ def cli(ctx, **kwargs):
     if flags.RECORD_TIMING_INFO:
         ctx.with_resource(profiler(enable=True, outfile=flags.RECORD_TIMING_INFO))
 
+    project_root = flags.PROJECT_DIR if flags.PROJECT_DIR else os.getcwd()
```
MichelleArk, line 79: Is the if/else statement necessary here given `flags.PROJECT_DIR` [defaults](https://github.com/dbt-labs/dbt-core/blob/chenyu/project_with_click/core/dbt/cli/params.py#L233) to `Path.cwd()` via the [default_project_dir](https://github.com/dbt-labs/dbt-core/blob/374fce4e5b86882250db79945daf0b5cd3419a68/core/dbt/cli/resolvers.py#L4) method? 

Also, is there a semantic difference between the `project_root` and `project_dir` variable names or can the naming be standardized here?

iknox-fa: +1 to both of these comments.

core/dbt/cli/main.py
iknox-fa: I think we can integrate the logic in `parse_cli_vars` into the custom option type for YAML.  It's already doing 90% of this anyway, we just need to check that the yaml string evaluates to a dict and set a default value to `{}`.  If we do that we can just use `flags.vars` when calling `load_project`.

iknox-fa: Side note:  We'll want to support the "short form" of vars @jtcohen6 pointed out here: https://dbt-labs.slack.com/archives/C03P5DB4GA3/p1667422670997319

iknox-fa: Regarding the entire profile being a dependency of project:

I _really_ don't like that these two things are coupled right now.  Especially since the project requires the _entire_ profile.  I see what you're saying about how widely spread the renderer is across the codebase, but I'm going to take a little more time and see if I can find a way past it.

I won't hold up the PR on that account though-- We can always ticket decoupling the two as another piece of work.

core/dbt/cli/main.py
```diff
@@ -73,6 +76,14 @@ def cli(ctx, **kwargs):
...
+    cli_vars = parse_cli_vars(getattr(flags, "vars", "{}"))
+
+    # TODO need profile to exisit
+    profile = None
+    # project need profile to render because it requires knowing Target
+    ctx.obj["project"] = load_project(project_root, profile, cli_vars)
```
iknox-fa, line 85: Good call-- might as well start stuffing that context obj with all the things we want to pass on.

core/dbt/config/runtime.py
```diff
@@ -189,28 +205,31 @@ def _get_rendered_profile(
...
+        profile_name = profile_renderer.render_value(raw_profile_name)
         profile = cls._get_rendered_profile(args, profile_renderer, profile_name)
         # Save env_vars encountered in rendering for partial parsing
         profile.profile_env_vars = profile_renderer.ctx_obj.env_vars
+        return profile
+
+    @classmethod
+    def collect_parts(cls: Type["RuntimeConfig"], args: Any) -> Tuple[Project, Profile]:
+        # profile_name from the project
+        project_root = args.project_dir if args.project_dir else os.getcwd()
```
iknox-fa, line 224: Can we not use any `os` methods for paths?  It's the "old" filesystem interface, PathLib will save us a lot of headaches, esp when it comes to cross-platform stuff.

core/dbt/config/runtime.py
```diff
@@ -189,28 +205,31 @@ def _get_rendered_profile(
...
     @classmethod
-    def collect_parts(cls: Type["RuntimeConfig"], args: Any) -> Tuple[Project, Profile]:
-        # profile_name from the project
-        project_root = args.project_dir if args.project_dir else os.getcwd()
-        version_check = bool(flags.VERSION_CHECK)
-        partial = Project.partial_load(project_root, verify_version=version_check)
-
+    def get_profile(
+        cls: Type["RuntimeConfig"], args: Any, cli_vars: Dict[str, Any], raw_profile_name: str
+    ):
```
iknox-fa, line 210: Minor detail: needs return type annotation

core/dbt/config/runtime.py
```diff
@@ -30,6 +31,21 @@
 from dbt.dataclass_schema import ValidationError
 
 
+def load_project(
+    project_root: str,
+    profile: HasCredentials,
+    cli_vars: Optional[Dict[str, Any]] = None,
+):
```
iknox-fa, line 38: Minor detail: needs return type annotation

iknox-fa (changes requested):
A few small changes.  Fee free to ignore my carping about the Profile/Project coupling though.  I'll work on a solution for that.

core/dbt/cli/main.py
jtcohen6: Explaining this to myself: `dbt_project.yml` can include Jinja expressions that use [`{{ target }}`](https://docs.getdbt.com/reference/dbt-jinja-functions/target), so we need to have loaded the profile before we can render those parts of `dbt_project.yml`.

_Another comment that's very much out of scope for this PR:_ We might also think about a future where one project/DAG could to be deployed using multiple adapters / targets. (They _could_ all be defined within the same profile, though.) At that point, we might especially want a clear decoupling between projects and connection profiles. The info arch around `target`, `profile`, `envs` (!?) is something we should look to refactor next year.

ChenyuLInx: Read my mind @jtcohen6 , I think I was writing somewhere that we should separate out target and the actual connection cred stuff.

@iknox-fa  I will also spend some more time try to figure out if we can get that rendered in some other form, so that maybe we just need to pass in target other than the whole object

core/dbt/cli/main.py
ChenyuLInx: @MichelleArk  good catch, I didn't dig in there.
I will also update the name

core/dbt/config/runtime.py
ChenyuLInx: I think this part of the code will go away in the new CLI? just refactored it to make the new parts easier, we are removing this line in the new code and just use [this](https://github.com/dbt-labs/dbt-core/pull/6142#discussion_r1012244317)

ChenyuLInx: And kept this function so all tests still passin

core/dbt/cli/main.py
ChenyuLInx: @iknox-fa I didn't do the

core/dbt/cli/main.py
```diff
@@ -73,6 +75,13 @@ def cli(ctx, **kwargs):
     if flags.RECORD_TIMING_INFO:
         ctx.with_resource(profiler(enable=True, outfile=flags.RECORD_TIMING_INFO))
 
+    cli_vars = parse_cli_vars(getattr(flags, "vars", "{}"))
+
+    # TODO need profile to exisit
+    profile = None
+    # project need profile to render because it requires knowing Target
+    ctx.obj["project"] = load_project(flags.PROJECT_DIR, profile, cli_vars)
```
MichelleArk, line 83: Do we need to initialize ctx.obj to an empty dictionary prior to the assignment here? Testing this out locally, I get the following error: 

```    
   ctx.obj["project"] = load_project(project_root, profile, cli_vars)
TypeError: 'NoneType' object does not support item assignment
```

core/dbt/cli/main.py
```diff
@@ -73,6 +75,13 @@ def cli(ctx, **kwargs):
     if flags.RECORD_TIMING_INFO:
         ctx.with_resource(profiler(enable=True, outfile=flags.RECORD_TIMING_INFO))
 
+    cli_vars = parse_cli_vars(getattr(flags, "vars", "{}"))
```
MichelleArk, line 78: Will `flags.vars` ever be a valid attribute reference? My understanding is that [Flags sets attribute names as uppercase](https://github.com/dbt-labs/dbt-core/blob/main/core/dbt/cli/flags.py#L29).

Any reason not to default the [`vars` param to `{}`](https://github.com/dbt-labs/dbt-core/blob/main/core/dbt/cli/params.py#L347-L352)? That way this line could be simplified to: 
```python
cli_vars = parse_cli_vars(flags.VARS)
```

core/dbt/cli/main.py
ChenyuLInx: And we talked about testing this file during standup today, @iknox-fa will add some tests to test calls to the new CLI interface will run without error. Currently not possible due to profile doesn't exist

core/dbt/cli/main.py
iknox-fa: Just to close the loop-- neither of us came up with a good way to break this apart without adding a lot to this ticket so we're leaving it as is.  https://github.com/dbt-labs/dbt-core/issues/6257 is the follow on ticket to decouple things.

.pre-commit-config.yaml
```diff
@@ -30,7 +30,7 @@ repos:
     args:
     - "--check"
     - "--diff"
-- repo: https://gitlab.com/pycqa/flake8
+- repo: https://github.com/pycqa/flake8
```
iknox-fa, line 33: Erm.  What's this about @ChenyuLInx ?

iknox-fa (approved):
LGTM with the outstanding Q about the flake8 repo change.  I'm assuming they just moved where that is stored, but I'd like to verify that before merging.

.pre-commit-config.yaml
ChenyuLInx: This is for porting the change @MichelleArk  did for fixing the code quality check

core/dbt/config/runtime.py
```diff
@@ -30,6 +31,21 @@
 from dbt.dataclass_schema import ValidationError
 
 
+def load_project(
+    project_root: str,
+    profile: HasCredentials,
+    cli_vars: Optional[Dict[str, Any]] = None,
+) -> Project:
+    # get the project with all of the provided information
+    version_check = bool(flags.VERSION_CHECK)
```
MichelleArk, line 40: should we be getting `VERSION_CHECK` from the [click params](https://github.com/dbt-labs/dbt-core/blob/main/core/dbt/cli/params.py#L361) at this point?

MichelleArk (approved):
LGTM! One outstanding comment re: flags usage. The unsafe access issue flags.VARS/flags.PROJECT_DIR would make sense to tackle generally in a separate PR.

core/dbt/config/runtime.py
ChenyuLInx: Updated to pass `VERSION_CHECK` as a param

PR #6157, opened 2022-10-27
More ergonomic profile name handling

### LINKED ISSUE #6156: [CT-1426] Easier programmatic profile selection
Per a discussion with the switchboard team this PR allows for overriding the profile name.  It represents a minor improvement  of the existing code and will be oblivated by the Click profile/config work, but there's an urgent need for this now ergo this ticket/PR

iknox-fa:
resolves #6156

### Description
Allows for overriding the profile name when creating new runtime configs.  
N.B. Per SB team we need to backport this to 1.2.

### Checklist

### DIFF
--- a/core/dbt/config/runtime.py
+++ b/core/dbt/config/runtime.py
@@ -3,31 +3,42 @@ import os
 from copy import deepcopy
 from dataclasses import dataclass, field
 from pathlib import Path
-from typing import Dict, Any, Optional, Mapping, Iterator, Iterable, Tuple, List, MutableSet, Type
+from typing import (
+    Any,
+    Dict,
+    Iterable,
+    Iterator,
+    List,
+    Mapping,
+    MutableSet,
+    Optional,
+    Tuple,
+    Type,
+    Union,
+)
 
-from .profile import Profile
-from .project import Project
-from .renderer import DbtProjectYamlRenderer, ProfileRenderer
-from .utils import parse_cli_vars
 from dbt import flags
-from dbt.adapters.factory import get_relation_class_by_name, get_include_paths
-from dbt.helper_types import FQNPath, PathSet, DictDefaultEmptyStr
+from dbt.adapters.factory import get_include_paths, get_relation_class_by_name
 from dbt.config.profile import read_user_config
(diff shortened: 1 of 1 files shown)


core/dbt/config/runtime.py
```diff
@@ -3,31 +3,42 @@
 from copy import deepcopy
 from dataclasses import dataclass, field
 from pathlib import Path
-from typing import Dict, Any, Optional, Mapping, Iterator, Iterable, Tuple, List, MutableSet, Type
```
iknox-fa, line 6: FYI-- Most of these import changes are just sorting... Nothing major changed here but having properly sorted imports matters so I'm doing a little cleanup whilst here.

core/dbt/config/runtime.py
```diff
@@ -190,28 +201,52 @@ def _get_rendered_profile(
 
     @classmethod
     def collect_parts(cls: Type["RuntimeConfig"], args: Any) -> Tuple[Project, Profile]:
```
iknox-fa, line 203: The code in this function will be basically what the SB team will use to generate the profile and project, which can then be used with the `from_parts` method to generate Runtime configs.

The difference will be potentially overriding the profile name when it's collected like so:
```
profile = cls.collect_profile(args=args, profile_name="a_valid_profile_name")
```

leahwicz:
@iknox-fa will this need to go into 1.3 as well then? Do we see any risk or breaking change with this going into a patch? (still catching up on this in Slack...)

ChenyuLInx (approved):
LGTM

ChenyuLInx:
> @iknox-fa will this need to go into 1.3 as well then? Do we see any risk or breaking change with this going into a patch? (still catching up on this in Slack...)

@leahwicz  this won't be a breaking change.

iknox-fa:
> @iknox-fa will this need to go into 1.3 as well then? Do we see any risk or breaking change with this going into a patch? (still catching up on this in Slack...)

Yep-- apparently they support 1.2+ so we'll ned this change in both.  Do I need anotherPR for that?  (I forget how our backport system works.. I haven't had to use it in ages).

PR #6172, opened 2022-10-28
CT 1443 remove root path

### LINKED ISSUE #6171: [CT-1443] Remove 'root_path' from most node definitions
We are moving towards enabling manifests to be portable. Currently we include the 'root_path' of the project in every node, which is duplicative and non-portable. In addition, this field is hardly used.

Remove root_path from the node class definitions.

Note: because we don't load seed csv file contents, we do use the root_path to load the seed file when the seeds command is run. For now, we will include the root_path in seed nodes, but longer term we need to come up with a solution for this, whether making it possible to get the root_path for every project in the 'load_agate_table' context method or some other solution.

This is done as part of the structured logging effort, as cleanup of the node attributes that need to be included in the contracts for node information.

gshank:
resolves #6171

### Description

As part of cleanup for the structured logging effort, remove the 'root_path' field from most nodes.

### Checklist


schemas/dbt/run-results/v4.json
```diff
@@ -37,12 +37,12 @@
         },
         "dbt_version": {
           "type": "string",
-          "default": "1.2.0a1"
+          "default": "1.4.0a1"
```
emmyoop, line 40: It looks like only the default and description fields are changing.  Do we really need to commit this?

schemas/dbt/manifest/v7.json
```diff
@@ -366,9 +365,6 @@
         "package_name": {
           "type": "string"
         },
-        "root_path": {
-          "type": "string"
-        },
```
emmyoop, line 371: Do we need to bump the schema to v8 instead of modifying?

emmyoop (changes requested):
A few schema comments

gshank:
Yes, we did need to bump the schema to v8. Have done that work.

emmyoop (changes requested):
The run-results schema is just updating defaults and descriptions so `run-results/v4.json` should be removed from the PR.

gshank:
Yeah, I've removed that run-results file a couple of times and then I run the script again... Gone now, I hope.

emmyoop (approved):

PR #6175, opened 2022-10-28
fix: check if a relation exists before creating one with ref

### LINKED ISSUE #6155: [CT-1420] [Bug] Ref creates a new relation every time, meaning that model data (such as `type` cannot be inherited)

### Current Behavior

At the moment, `ref` creates a new relation object each time rather than checking if something already exists.

Context for the problem this is causing is in slack:

https://getdbt.slack.com/archives/C50NEBJGG/p1666805781564349

This is something that I've run into when developing the duckdb adapter.

https://github.com/jwills/dbt-duckdb/issues/19#issuecomment-1292375802

The relevant code is: https://github.com/dbt-labs/dbt-core/blob/19167842873ea882c2dbb70aeb525953812d8ef3/core/dbt/context/providers.py#L463

which uses:

https://github.com/dbt-labs/dbt-core/blob/19167842873ea882c2dbb70aeb525953812d8ef3/core/dbt/adapters/base/relation.py#L243

### Expected Behavior

If model `x` has a materialization of `foo` (and the returned relation from the materialization sets this up correctly) i'd expect `ref('x')` to be a relation of type `foo`, up to the enum being supported under the hood etc.

### Steps To Reproduce

Make a relation (via a model) that has a type that is not `None`, and then ref that model. A relation of type none is returned, with the same name etc

### Relevant log output

_No response_

### Environment

_No response_

_No response_

### Additional Context

_No response_

jpmmcneill:
resolves #6155

### Description

Use the adapter to get a relation during ref execution (importantly just during runtime) to preserve relation metadata such as type when it has been defined.
When no relation is found, create a new one as was previously done

### Checklist

### DIFF
--- a/core/dbt/context/providers.py
+++ b/core/dbt/context/providers.py
@@ -483,7 +483,16 @@ class RuntimeRefResolver(BaseRefResolver):
                 disabled=isinstance(target_model, Disabled),
             )
         self.validate(target_model, target_name, target_package)
-        return self.create_relation(target_model, target_name)
+        relation = self.db_wrapper._adapter.get_relation(
+            database=target_model.database,
+            schema=target_model.schema,
+            identifier=target_model.identifier
+        )
+        if relation is not None:
+            return relation
+        else:
+            # we need to create a relation if no existing relation was found
+            return self.create_relation(target_model, target_name)
 
     def create_relation(self, target_model: ManifestNode, name: str) -> RelationProxy:
         if target_model.is_ephemeral_model:


jpmmcneill:
I have now validated that this solves the issue I was having previously (where overwriting the `render` function depending on the relation type wasnt working. With this, I have `ref` resolving to `read_parquet(...)` in a duckdb adapter 😃

core/dbt/context/providers.py
```diff
@@ -483,7 +483,16 @@ def resolve(self, target_name: str, target_package: Optional[str] = None) -> Rel
...
+        relation = self.db_wrapper._adapter.get_relation(
+            database=target_model.database,
+            schema=target_model.schema,
+            identifier=target_model.identifier
+        )
+        if relation is not None:
+            return relation
+        else:
+            # we need to create a relation if no existing relation was found
+            return self.create_relation(target_model, target_name)
```
jtcohen6, line 495: This is more or less identical to the logic in the [`get_or_create_relation` macro](https://github.com/dbt-labs/dbt-core/blob/8145eed603266951ce35858f7eef3836012090bd/core/dbt/include/global_project/macros/adapters/relation.sql#L68-L86)

The risk here: If the relation `type` isn't needed, _and_ the `ref`'d relation hasn't been cached, this will cause dbt to run an additional caching query, which could slow down compilation. Caching queries can add a lot of overhead to the start of a run, so we've been thinking about / experimenting with ways to trim them down: https://docs.getdbt.com/reference/global-configs#cache-database-objects-for-selected-resource

So the question is: Is there merit to an additional flag/argument that turns this on? Or _only_ returning the relation type if the relation is also cached (rather than triggering a metadata query)? In particular, I'm thinking about:
- `run-operation`, which does not populate the cache
- semantic layer queries, which do not perform introspective queries against the database (to minimize overhead)

jtcohen6 (commented):
Nice work on this one! The solution is straightforward and self-contained. Just a few edge cases we need to think through, to prevent a performance regression when the relation type isn't really needed.

core/dbt/context/providers.py
jpmmcneill: Thanks @jtcohen6!

Just so we're on the same page, I assume that you'd see this extra argument being inside the ref() function?

jtcohen6: @jpmmcneill After thinking about this some more, I appreciate why this seemed to be a bug. `ref` returns a `Relation` object, which maps to an object in the database. In that sense, it _seems_ like it should have metadata about that database object, including its type — `view`, `table`, `external`, `materialized_view`, `parquet` (!), etc.

But getting that type does require running a database query, at one point or another. During `run`-style invocations, we run caching queries so that subsequent lookups hit the cache and return very quickly. But in all other cases, this would be a real performance penalty.

The workaround I proposed in https://github.com/dbt-labs/dbt-core/issues/6155#issuecomment-1295631083 is actually the thing we already do in materializations: First, you grab the relation (whether `{{ this }}` or `{{ ref('thing') }}`), and then you use [`load_cached_relation()`](https://github.com/dbt-labs/dbt-core/blob/c3ccbe3357d15e45153c70b62bea626d37fc0d05/core/dbt/include/global_project/macros/materializations/models/view/view.sql#L3) (= `adapter.get_relation`) to get the extra info about its relation `type`.

So, the two choices ahead of us are:
```sql
-- current state
{% set ref_model = ref('some_model') %}
{% set with_type = load_cached_relation(ref_model) or ref_model %}
```
```sql
-- proposed future state
{% set ref_model = ref('some_model', lookup=True) %}
```
At that point, it's just ergonomics. I don't mind the greater explicitness of the former. What do you think?

jtcohen6:
Going to close this for the time being - happy to continue discussing in the linked issue

jpmmcneill:
Thanks @jtcohen6, I promise that I will get back to discussing this soon! Have had a few things sapping my out of hours energy 🙂

PR #6200, opened 2022-11-02
validation on non-public models

dave-connors-3:
### Description
adding a really simple framework to make sure there are values present for metric values that are optional for semantic metrics and not-optional

### DIFF
--- a/core/dbt/parser/manifest.py
+++ b/core/dbt/parser/manifest.py
@@ -1251,6 +1251,18 @@ def _process_semantic_information_for_metric(
             metric.time_grains = metric.config._extra["default_time_grains"]
 
 
+def _validate_non_public_metric(metric: ParsedMetric):
+    """
+    Ensure that there are the necessary fields for metrics from non-public models
+    """
+    keys_to_check = ["time_grains"]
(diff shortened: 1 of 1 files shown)


core/dbt/parser/manifest.py
```diff
@@ -1291,9 +1303,12 @@ def _process_refs_for_metric(manifest: Manifest, current_project: str, metric: P
 
         metric.depends_on.nodes.append(target_model_id)
 
-        _process_semantic_information_for_metric(
-            manifest, target_model, metric, target_model_package, current_project
-        )
+        if target_model.is_public:
```
gshank, line 1306: Normally we want to limit the post-processing things we do in parsing as much as possible, to only those things that CAN'T be done earlier. Is there some reason that the process_semantic_information_for_metric and  _validate_non_public_metric things aren't being done when nodes are normally validated?

callum-mcdata: When you say `normally validated` are you talking about the classes in `unparsed.py` and `parsed.py`?

gshank: Yes

callum-mcdata (approved):
I think we're good to go with the understanding that we'll probably be re-writing/fixing most of this in the future!

PR #6233, opened 2022-11-09
dbt Constraints

### LINKED ISSUE #6079: [Feature] Add column type constraints as dbt native configs

### Describe the feature

**Problem**:
- Data types are created after the fact(assumed by the database) and/or a dbt user has to explicitly cast data types in their SQL. Because of this, there isn't much incentive to assure dbt models have expected data types for consumers to depend on(it's tedious work :/). This may not be a problem for 50 dbt models, but becomes a big problem when you have 2000+ dbt models. There should be an easier way to configure data type constraints for any dbt models and provide options to enable/disable data type enforcement(think: what [mypy](http://mypy-lang.org/) does for the python static typing experience).

**Solution**:
- Bring data types as native to the dbt experience
- Example BigQuery Constraints Macro:https://gist.github.com/sungchun12/f7ea081773ae824a83294649530d6e41
- Related discussion: https://github.com/dbt-labs/dbt-core/discussions/5244#discussioncomment-3880254

Custom Materialization Macro in BigQuery: https://www.loom.com/share/1f1f190e66254d12984962c613e8082d

This is a general design pattern, but uses meta configs to enable this functionality. 

### Who will this benefit?

dbt users with large projects that need extra robustness in their developer experience and downstream users get the data they expect: data types and all. 

yes, and I'll be working with someone in the community @jonathanneo

### Anything else?

**Research:** 
- Postgres Constraints: https://www.postgresql.org/docs/current/ddl-constraints.html#id-1.5.4.6.6
- Snowflake ONLY enforces `not null` constraints, the rest are nice to have metadata: https://docs.snowflake.com/en/sql-reference/constraints-overview.html#supported-constraint-types
- BigQuery Macro Demo: https://www.loom.com/share/1f1f190e66254d12984962c613e8082d

**Considerations:**
- have the configs embedded in dbt-core and then each adapter can take stable configs to adjust table materialization macros
- Enforce column positions based on schema config
- dbt-core contains the hub of native configs while adapters own specific implementations as each database has nuances to which enforceable constraints are valid.
-  think about not null and default values for a table
- Work with Jon Neo from Canva
- Include a `check` constraint as pgsql enables it? 
- What if constraints could be defined once within a subfolder path in dbt_project.yml and then in automatically creates not null constraints in the DDL?
- Leverage how seed configs enable data types

sungchun12:
resolves #6079

### Checklist

### DIFF
--- a/core/dbt/exceptions.py
+++ b/core/dbt/exceptions.py
@@ -296,6 +296,17 @@ class UndefinedMacroException(CompilationException):
         )
 
 
+# TODO: create a new class that displays in the terminal that a data type config is missing/wrong
+class UndefinedDataTypeException(CompilationException):
+    def __str__(self, prefix="! ") -> str:
+        msg = super().__str__(prefix)
+        return (
+            f"{msg}. This can happen when a data_type configuration does "
+            "not exist or is defined as null when `constraints_enabled: true` for a dbt model. \
+            Check for typos and/or verify data_type is defined for all columns for the dbt models with constraints. "
+        )
+
+
 class UnknownAsyncIDException(Exception):
     CODE = 10012
     MESSAGE = "RPC server got an unknown async ID"
--- a/core/dbt/parser/generic_test_builders.py
+++ b/core/dbt/parser/generic_test_builders.py
@@ -180,6 +180,7 @@ class GenericTestBlock(TestBlock[Testable], Generic[Testable]):
         )
 
 
+# TODO: create a ConstraintsBuilder equivalent
 class TestBuilder(Generic[Testable]):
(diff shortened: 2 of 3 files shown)


core/dbt/exceptions.py
```diff
@@ -296,6 +296,17 @@ def __str__(self, prefix="! ") -> str:
         )
 
 
+# TODO: create a new class that displays in the terminal that a data type config is missing/wrong
```
sungchun12, line 299:
```suggestion
# TODO: create a new class that displays in the terminal that a data type config is missing/wrong. Test Comment
```

sungchun12:
Closing in favor of #6271

Issue #5547, opened 2022-07-25
[CT-923] `dbt deps` works in click


iknox-fa:

jtcohen6:
Resolved by #6260

PR #6260, opened 2022-11-16
DepsTask with Click

MichelleArk:
resolves #https://github.com/dbt-labs/dbt-core/issues/5547

### Checklist

### DIFF
--- a/core/dbt/deps/local.py
+++ b/core/dbt/deps/local.py
@@ -8,6 +8,7 @@ from dbt.contracts.project import (
 )
 from dbt.events.functions import fire_event
 from dbt.events.types import DepsCreatingLocalSymlink, DepsSymlinkNotAvailable
+from dbt.config.project import PartialProject, package_config_from_data
 
 
 class LocalPackageMixin:
@@ -39,9 +40,11 @@ class LocalPinnedPackage(LocalPackageMixin, PinnedPackage):
             project.project_root,
         )
 
-    def _fetch_metadata(self, project, renderer):
-        loaded = project.from_project_root(self.resolve_path(project), renderer)
-        return ProjectPackageMetadata.from_project(loaded)
+    def _fetch_metadata(self, project, renderer) -> ProjectPackageMetadata:
+        partial = PartialProject.from_project_root(self.resolve_path(project))
+        packages_data = renderer.render_data(partial.packages_dict)
+        packages_config = package_config_from_data(packages_data)
+        return ProjectPackageMetadata(partial.project_name, packages_config.packages)
 
     def install(self, project, renderer):
         src_path = self.resolve_path(project)
--- a/core/dbt/deps/local.py
+++ b/core/dbt/deps/local.py
@@ -8,7 +8,8 @@ from dbt.contracts.project import (
 )
 from dbt.events.functions import fire_event
 from dbt.events.types import DepsCreatingLocalSymlink, DepsSymlinkNotAvailable
-from dbt.config.project import PartialProject, package_config_from_data
+from dbt.config.project import PartialProject, Project
+from dbt.config.renderer import PackageRenderer
 
 
 class LocalPackageMixin:
@@ -40,11 +41,11 @@ class LocalPinnedPackage(LocalPackageMixin, PinnedPackage):
             project.project_root,
         )
 
-    def _fetch_metadata(self, project, renderer) -> ProjectPackageMetadata:
+    def _fetch_metadata(
+        self, project: Project, renderer: PackageRenderer
+    ) -> ProjectPackageMetadata:
         partial = PartialProject.from_project_root(self.resolve_path(project))
-        packages_data = renderer.render_data(partial.packages_dict)
-        packages_config = package_config_from_data(packages_data)
-        return ProjectPackageMetadata(partial.project_name, packages_config.packages)
+        return partial.render_package_metadata(renderer)
 
     def install(self, project, renderer):
         src_path = self.resolve_path(project)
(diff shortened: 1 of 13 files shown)


core/dbt/deps/resolver.py
```diff
@@ -116,21 +116,21 @@ def _check_for_duplicate_project_names(
 
 
 def resolve_packages(
-    packages: List[PackageContract], config: RuntimeConfig
+    packages: List[PackageContract], project: Project, profile, cli_vars
 ) -> List[PinnedPackage]:
     pending = PackageListing.from_contracts(packages)
     final = PackageListing()
-    renderer = DbtProjectYamlRenderer(config, config.cli_vars)
+    renderer = DbtProjectYamlRenderer(profile, cli_vars)
```
ChenyuLInx, line 123: If you don't need to use Target for resolving the package, you don't need to initialize this render with `profile`.

core/dbt/task/base.py
```diff
@@ -169,7 +170,7 @@ def get_nearest_project_dir(args):
 
 
 def move_to_nearest_project_dir(args):
-    nearest_project_dir = get_nearest_project_dir(args)
+    nearest_project_dir = get_nearest_project_dir(args.project_dir)
```
ChenyuLInx, line 173: since `get_nearest_project_dir` doesn't need `args` anymore, maybe we can also have `move_to_nearest_project_dir` just taking  `project_dir` as argument?

core/dbt/task/deps.py
```diff
@@ -23,11 +23,17 @@
 from dbt.task.base import BaseTask, move_to_nearest_project_dir
 
 
+from dbt.config import RuntimeConfig, Project
+from dbt.task.base import NoneConfig
+
 class DepsTask(BaseTask):
```
ChenyuLInx, line 29: Since we are doing the refactor here, does this still have to be a Task object? Seems like we are not getting much stuff from the `BaseTask'?

ChenyuLInx (commented):
Looks great!

core/dbt/task/base.py
MichelleArk: yes yes! I actually thought I had that change in, not sure where it went. Will update

core/dbt/deps/resolver.py
MichelleArk: Given that any dbt_project.yml (including those from packages) _can_ include Jinja expressions that use {{ target }}, it feels necessary to continue to pass `UnsetProfile` through the deps command to stay consistent with the current behaviour and not raise `CompilationException: 'target' is undefined` errors

ChenyuLInx: But shouldn't we already rendered everything at the time we start running the task?

core/dbt/task/deps.py
```diff
@@ -82,9 +98,30 @@ def run(self):
                 fire_event(EmptyLine())
                 fire_event(DepsNotifyUpdatesAvailable(packages=packages_to_upgrade))
 
+    @classmethod
+    def _get_unset_profile(cls) -> UnsetProfile:
```
ChenyuLInx, line 102: Is function [`_get_rendered_profile`](https://github.com/dbt-labs/dbt-core/blob/a0ec0b6f9db74701ee701f2415ad47c8b91300ba/core/dbt/config/runtime.py#L585) still needed after adding this new function?

MichelleArk: I believe the commands that still depend on UnsetProfileConfig (just clean) still require it at this point. But once CleanTask is implemented with click, we should be able to rip out that method and all of UnsetProfileConfig!

core/dbt/deps/local.py
```diff
@@ -39,9 +41,11 @@ def resolve_path(self, project):
             project.project_root,
         )
 
-    def _fetch_metadata(self, project, renderer):
-        loaded = project.from_project_root(self.resolve_path(project), renderer)
-        return ProjectPackageMetadata.from_project(loaded)
+    def _fetch_metadata(
```
ChenyuLInx, line 44: Seems like this function and the function in `git.py` is mostly similar other than the way path is extracted, maybe consider refactor both into one function?

core/dbt/task/deps.py
ChenyuLInx: Maybe leave a note in the clean task so when @stu-k  works on it he has the infomation?

ChenyuLInx (approved):
Looks great with one small question regarding `_fetch_metadata`

core/dbt/task/deps.py
MichelleArk: There may be a more appropriate base class for more util-like commands such as clean, deps, init and debug. At the risk of abstracting too early I'm thinking to leave this line as-is for this PR and revisit it once we've got clean/init/debug combed through - there's probably a common set of functionality that makes sense for util-type commands that can be separate from BaseTask.

core/dbt/deps/local.py
MichelleArk: I was itching to do the same but I keep going back on forth on whether it makes sense refactored elsewhere. At this point the only deviation between the two _fetch_metadata implementations is in how they construct a `PartialProject` to call `render_package_metadata` on. Figuring out how the `PartialProject` gets constructed (if one is necessary at all) feels like the right level of responsibility for the `PinnedPackage` subclasses.

Additionally, I'm not sure where a better fit for this common logic would be. A couple alternatives considered: 
1. In `ProjectPackageMetadata`, perhaps something like `ProjectPackageMetadata.from_project_dir(project_dir, renderer)`. This is what I went with initially but ran into circular imports between `dbt.config` and `dbt.contracts`. The logic also felt better contained in `PartialProject` where the other relevant helpers were already defined. 
2. In the common base class, something like `PinnedPackage._fetch_metadata_from_project_dir(project_dir, renderer)`. It doesn't feel great for the base class to be implemented with so much awareness of what just a subset of its subclasses are doing.

ChenyuLInx: I was thinking of the second option you have, didn't realize that there's also a `registry` method also overwrite this function. Although it looks like that one doesn't even use the two arguments passed in.
I agree that this doesn't feel like something we can come up with a easy-to-implement better solution very soon. So I am okay leaving it as is for now.

PR #6266, opened 2022-11-16
Set Flags from UserConfig

MichelleArk:
resolves #https://github.com/dbt-labs/dbt-core/issues/6327

### Description
* Incorporates `UserConfig` (if available) to `Flags`, preferring `UserConfig` values to environment variables and defaults.
* Adds some initial tests for `cli.Flags`

Another option considered was to implement resolvers for each user-config param in cli/resolvers.py and use them to set defaults in params.py. However, this approach would still need access to profiles_dir prior to Flags initialization, and be much less succinct. 

### Checklist

### DIFF
--- a/core/dbt/cli/flags.py
+++ b/core/dbt/cli/flags.py
@@ -43,8 +43,11 @@ class Flags:
         assign_params(ctx, params_assigned_from_default)
 
         # Get the invoked command flags
-        if hasattr(ctx, "invoked_subcommand") and ctx.invoked_subcommand is not None:
-            invoked_subcommand = getattr(import_module("dbt.cli.main"), ctx.invoked_subcommand)
+        invoked_subcommand_name = (
+            ctx.invoked_subcommand if hasattr(ctx, "invoked_subcommand") else None
+        )
+        if invoked_subcommand_name is not None:
+            invoked_subcommand = getattr(import_module("dbt.cli.main"), invoked_subcommand_name)
             invoked_subcommand.allow_extra_args = True
             invoked_subcommand.ignore_unknown_options = True
             invoked_subcommand_ctx = invoked_subcommand.make_context(None, sys.argv)
@@ -64,9 +67,7 @@ class Flags:
                     )
 
         # Hard coded flags
-        object.__setattr__(
-            self, "WHICH", ctx.protected_args[0] if ctx.protected_args else ctx.info_name
-        )
+        object.__setattr__(self, "WHICH", invoked_subcommand_name or ctx.info_name)
         object.__setattr__(self, "MP_CONTEXT", get_context("spawn"))
 
         # Support console DO NOT TRACK initiave
--- a/core/dbt/cli/flags.py
+++ b/core/dbt/cli/flags.py
@@ -11,6 +11,7 @@ from click import Context, get_current_context
 from click.core import ParameterSource
 
 from dbt.config.profile import read_user_config
+from dbt.contracts.project import UserConfig
 
 if os.name != "nt":
     # https://bugs.python.org/issue41567
@@ -19,7 +20,7 @@ if os.name != "nt":
 
 @dataclass(frozen=True)
 class Flags:
-    def __init__(self, ctx: Context = None) -> None:
+    def __init__(self, ctx: Context = None, user_config: UserConfig = None) -> None:
 
         if ctx is None:
             ctx = get_current_context()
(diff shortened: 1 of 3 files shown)


core/dbt/cli/flags.py
```diff
@@ -29,29 +34,47 @@ def assign_params(ctx):
...
+                user_config_param_value = getattr(user_config, param_assigned_from_default, None)
+                if user_config_param_value is not None:
+                    object.__setattr__(
+                        self, param_assigned_from_default.upper(), user_config_param_value
+                    )
 
         # Hard coded flags
-        object.__setattr__(self, "WHICH", ctx.info_name)
+        object.__setattr__(
+            self, "WHICH", ctx.protected_args[0] if ctx.protected_args else ctx.info_name
```
iknox-fa, line 68: I'm not sure I follow what this is doing... According to [the docs](https://click.palletsprojects.com/en/8.1.x/api/#click.Context.protected_args) `protected_args` is used for certain extra parsing scenarios and that doesn't sound like that we're looking for when using `flags.WHICH`?

core/dbt/cli/flags.py
```diff
@@ -29,29 +34,47 @@ def assign_params(ctx):
...
+        )
         object.__setattr__(self, "MP_CONTEXT", get_context("spawn"))
 
         # Support console DO NOT TRACK initiave
         object.__setattr__(
             self,
             "ANONYMOUS_USAGE_STATS",
             False
-            if os.getenv("DO_NOT_TRACK", "").lower() in (1, "t", "true", "y", "yes")
+            if os.getenv("DO_NOT_TRACK", "").lower() in ("1", "t", "true", "y", "yes")
```
iknox-fa, line 77: Good catch!  env vars != ints. Ever.

iknox-fa (commented):
LGTM other than my question about `protected_args`

core/dbt/cli/flags.py
MichelleArk: Here I was looking to get an accurate value for `WHICH` regardless of whether the context was based on the CLI group or a subcommand. I assume we'll want callers going through the CLI group context because it does a lot of setup (and even implements some CLI functionality, such as `--version`).

In either case the caller would be responsible for figuring out what to set `name` to when constructing a `Context` object programmatically, which is then used to set `flags.VERSION`:
```python
>>> from dbt.cli.main import cli
>>> from dbt.cli.flags import Flags
# Caller would need to infer that "run" is the right name to set based on args.
>>> ctx = cli.make_context(cli.name, ["--partial-parse", "run"])
>>> flags = Flags(ctx)
>>> flags.WHICH
'cli'
```

Perhaps `invoked_subcommand` is what we should use here, if available. However `invoked_subcommand` is only available if the command is actually being _invoked_, as opposed to based on the provided `args`. So it would be `None` in the example above. 

I went with `protected_args` because of the docs hinting that it is used to implement nested parsing, but it does feel fragile as well.

MichelleArk: Related note - Currently on feature/click-cli, flags.WHICH is `dbt` [at this point](https://github.com/dbt-labs/dbt-core/compare/feature/click-cli#diff-43832707bfec8cdaa49c13f004d9b3296585827f7e68ef07fba7ad1d5e6eb7c6R62), not the subcommand name when called directly. So it's technically not accurate unless we [recreate flags further along in the subcommand method](https://github.com/dbt-labs/dbt-core/compare/feature/click-cli#diff-43832707bfec8cdaa49c13f004d9b3296585827f7e68ef07fba7ad1d5e6eb7c6R231), which I understand we're looking to move away from doing. Using `invoked_subcommand` when available should at least fix the issue from the CLI interface.

jtcohen6: As far as I'm aware, the real function of `flags.WHICH` is to support conditional checks in user code for `{{ if flags.WHICH = ... }}`. That is, conditional behavior that only wants to run during certain commands, and not others.

links:
- the `flags` module is exposed to the context [here](https://github.com/dbt-labs/dbt-core/blob/c3be975783e065c4dc42c10fca29aca8e5584945/core/dbt/context/base.py#L626-L642)
- https://docs.getdbt.com/reference/dbt-jinja-functions/flags
- [functional test showing this behavior](https://github.com/dbt-labs/dbt-core/blob/main/tests/functional/schema_tests/test_schema_v2_tests.py)

iknox-fa: Oh I'm glad we had this conversation!  Yes `invoked_subcommand` is exactly what we should be using-- In my head that's what `info_name` did-- I probably C/P'd the wrong line in the original code (`info_name` is right above `invoked_subcommand` in the output of `pp vars(flags)`).

MichelleArk: Updated to `invoked_subcommand` if available, falling back on ctx.info_name. I think it'd still be worth looking into `protected_args` more but I'll punt that to a follow-up if that's alright; feels a bit out of scope for this work.

iknox-fa (approved):
LGTM!

Issue #2986, opened 2020-12-31
Rationalize quoting configs + properties


jtcohen6:
### Describe the feature

_Picks up from issues like #2468 and #2975, which are narrower in scope and offer more straightforward near-term fixes_

- Why do we call it [`quoting`](https://docs.getdbt.com/reference/project-configs/quoting/) when configuring database/schema/identifier names, but then [`quote`](https://docs.getdbt.com/reference/resource-properties/quote/) when describing properties of column names?
- Why does each adapter have to implement its quoting character sort-of twice? (#2243)
- There's also [quote_columns](https://docs.getdbt.com/reference/resource-configs/quote_columns/), which is a seed-only config item that lives on its own level but surely belongs inside the quoting config item (or will it be quote?!) as columns 😬
- Not to mention [`quoted`](https://docs.getdbt.com/reference/dbt-classes/#properties), which while not itself a config, returns the quoted version of a column name from a Relation based on the configs above

Instead, we should have a single config/property, and I think it should be `quote`. This would take over from the current project-level `quoting` config:
```
quote:
  database: true|false # or project on bq
  schema: true|false # or dataset on bq
  identifier: true|false
  columns: true|false
```

The `quote: {columns: true}` would also replace `quote_columns` as a bespoke config for seeds. If that config is specified in `dbt_project.yml`, it can be superseded by:
- setting `quote: {}` inside the `config()` block for a specific moel
- `quote: true|false` set for a specific column in `models/*.yml` (it's implied that this really means `quote: {column: true}`
- in a post-#2401 world, a `model` can set its `quote: {}` config within `models/*.yml`, too

If `quote` is not set, it falls back to the default behavior of the adapter plugin, which also sets the character used for quoting (almost always `"` or ``` ` ```).

### Questions

Here's what we have in the [docs FAQs](https://docs.getdbt.com/docs/building-a-dbt-project/using-sources/) today for sources:

> By default, dbt will not quote the database, schema, or identifier for the source tables that you've specified.

Should sources start respecting project-level `quote` settings? Or they continue to act independently, but we should enable turning this config-property on or off for _all_ sources in `dbt_project.yml`:
```yml
sources:
  quote:
    schema: true
```

### Describe alternatives you've considered
Retaining all of these configs/properties/adapter methods and documenting them exceptionally well so as to avoid confusion

### Additional context
[...]

PR #6271, opened 2022-11-16
dbt Constraints / model contracts

### LINKED ISSUE #6079: [Feature] Add column type constraints as dbt native configs

sungchun12:
resolves #6079

### Description

This is a building block to empower multi-project experiences in the future.

Brings defining data types for columns within a SQL select statement as native to dbt. This PR focuses on injecting new config defaults within the manifest, and introduces compilation defense code to ensure configs follow a clear standard.  Other adapter PRs will pair with this one to ensure configs cascade appropriately. 

Note: It is the responsibility of the adapter to provide extra guardrails for configs like `check` that do not work universally across databases. 

**Related Adapter Pull Requests**
> Must be reviewed with passing tests

- https://github.com/dbt-labs/dbt-core/pull/6271
- https://github.com/dbt-labs/dbt-snowflake/pull/341
- https://github.com/dbt-labs/dbt-redshift/pull/229
- https://github.com/dbt-labs/dbt-bigquery/pull/426
- https://github.com/dbt-labs/dbt-spark/pull/574

**dbt Core Test Cases**
- [x] Fix failing test regressions (Thanks @MichelleArk for helping me across the finish line 💯 )
- [x] Test: Verify compilation errors show expected dynamic messages
- [x] Test: Verify config hierarchy overrides work as expected(ex: constraints enabled at the model level override constraints enabled at the project_yml level with `True or False` values)
- [x] Test: Verify only SQL table materializations work
- [x] Test: Verify all or nothing `data_type` configs are enforced

**dbt-postgres Adapter Test Cases**
- [x] Test: Verify global macro is dispatched and can be overridden elegantly in the postgres adapter
- [x] Test: Verify DDL matches an expectation
- [x] Test: rollback the full DDL transaction if it fails with an ad hoc query verifying the table is in its previous state

### Checklist


sungchun12:
**Progress Update 11/17/2022**

- defaults for `data_type, constraint, check` appear in the `manifest.json`
- Add a constraints validator within `parsed.py` before a node can be written to compiled SQL code
image
- only SQL table materializations are allowed
- performance is fast
image

**Next Steps**
- Verify if my approach is an accepted pattern
- Write down test cases

core/dbt/contracts/graph/parsed.py
```diff
@@ -17,6 +17,7 @@
 
 from dbt.dataclass_schema import dbtClassMixin, ExtensibleDbtClassMixin
 
+from dbt.exceptions import CompilationException
```
sungchun12, line 20: I noticed no exceptions exist in this file except for the one I introduce. Is that intentional?

core/dbt/exceptions.py
```diff
@@ -1000,6 +1000,12 @@ def warn(msg, node=None):
     return ""
 
 
+# TODO: remove this later as Sung created this to avoid errors with running dbt-snowflake==1.3.0
+def warn_or_error(msg, node=None):
```
sungchun12, line 1004: Is there a reason this function was removed from the `main` branch while `1.3.latest` still has it? https://github.com/dbt-labs/dbt-core/blob/1.3.latest/core/dbt/exceptions.py#L1090

core/dbt/parser/generic_test_builders.py
```diff
@@ -180,6 +180,7 @@ def from_test_block(
         )
 
 
+# TODO: create a ConstraintsBuilder equivalent
 class TestBuilder(Generic[Testable]):
```
sungchun12, line 184: I should NOT need to build a `ConstraintsBuilder` because the configs extend `column_info` within the `schemas.py` file.

core/dbt/parser/models.py
```diff
@@ -177,6 +177,21 @@ def verify_python_model_code(node):
         raise ParsingException("No jinja in python model code is allowed", node=node)
 
 
+# TODO: Do we need to do this?
+# def verify_model_constraints(node):
```
sungchun12, line 181: We don't need to verify within this model because that happens `parsed.py`

schemas/dbt/manifest/v8.json
```diff
@@ -585,6 +585,10 @@
           "type": "boolean",
           "default": true
         },
+        "constraints_enabled": {
```
sungchun12, line 588: How do you test the manifest?

sungchun12:
A lot of tests will fail because I've adjusted the `manifest.json` and `column_info` to hold 2 new values: `constraint`, `check`. Before I adjust those tests, I need to verify my implementation is an accepted pattern.

sungchun12:
I got a dry run of this working to produce this DDL successfully against snowflake

```sql

  
    

        create or replace transient table ANALYTICS.dbt_sung.constraints_example  
        
        
          (
          
            
            id integer not null ,
          
            
            color string  ,
          
            
            date_day string  
          
        )
        
        as
        (

select 
  1 as id, 
  'blue' as color, 
  cast('2019-01-01' as date) as date_day
        );
```

With this custom macro to override the `table` materialization in snowflake

```sql

{% macro snowflake__create_table_as(temporary, relation, compiled_code, language='sql') -%}
  {%- if language == 'sql' -%}
    {%- set transient = config.get('transient', default=true) -%}
    {%- set cluster_by_keys = config.get('cluster_by', default=none) -%}
    {%- set enable_automatic_clustering = config.get('automatic_clustering', default=false) -%}
    {%- set copy_grants = config.get('copy_grants', default=false) -%}

    {%- if cluster_by_keys is not none and cluster_by_keys is string -%}
      {%- set cluster_by_keys = [cluster_by_keys] -%}
    {%- endif -%}
    {%- if cluster_by_keys is not none -%}
      {%- set cluster_by_string = cluster_by_keys|join(", ")-%}
    {% else %}
      {%- set cluster_by_string = none -%}
    {%- endif -%}

    {# # get the list of columns in my schema file#}
    {%- set user_provided_columns = model['columns'] -%}

    {{ sql_header if sql_header is not none }}

        create or replace {% if temporary -%}
          temporary
        {%- elif transient -%}
          transient
        {%- endif %} table {{ relation }} {% if copy_grants and not temporary -%} copy grants {%- endif %} 
        {# loop through user_provided_columns to create DDL with data types and constraints #}
        {% if config.get('constraints_enabled', False) %}
          (
          {% for i in user_provided_columns %}
            {% set col = user_provided_columns[i] %}
            {{ col['name'] }} {{ col['data_type'] }} {{ col['constraint'] or "" }} {{ "," if not loop.last }}
          {% endfor %}
        )
        {% endif %}
        as
        (
          {%- if cluster_by_string is not none -%}
            select * from(
              {{ compiled_code }}
              ) order by ({{ cluster_by_string }})
          {%- else -%}
            {{ compiled_code }}
          {%- endif %}
        );
      {% if cluster_by_string is not none and not temporary -%}
        alter table {{relation}} cluster by ({{cluster_by_string}});
      {%- endif -%}
      {% if enable_automatic_clustering and cluster_by_string is not none and not temporary  -%}
        alter table {{relation}} resume recluster;
      {%- endif -%}

  {%- elif language == 'python' -%}
    {{ py_write_table(compiled_code=compiled_code, target_relation=relation, temporary=temporary) }}
  {%- else -%}
      {% do exceptions.raise_compiler_error("snowflake__create_table_as macro didn't get supported language, it got %s" % language) %}
  {%- endif -%}

{% endmacro %}
```

core/dbt/exceptions.py
jtcohen6: Ah, this was related to work that Emily has been doing on more structured logging/exceptions. Just opened a ticket for it: https://github.com/dbt-labs/dbt-core/issues/6297

schemas/dbt/manifest/v8.json
jtcohen6: quis verificat ipsos verificantia? quis manifestat ipsos manifesta?

core/dbt/contracts/graph/unparsed.py
```diff
@@ -93,6 +93,8 @@ class HasDocs(AdditionalPropertiesMixin, ExtensibleDbtClassMixin, Replaceable):
     description: str = ""
     meta: Dict[str, Any] = field(default_factory=dict)
     data_type: Optional[str] = None
+    constraint: Optional[str] = None
```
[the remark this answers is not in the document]
jtcohen6: Could there ever be multiple constraints? Should this be a list (or [either a string or a list](https://github.com/dbt-labs/dbt-core/blob/ff2f1f42c394d27ff8d49ca7e285fcf2b76faa05/core/dbt/contracts/graph/model_config.py#L122-L126))?

jtcohen6:
@sungchun12 This code largely makes sense to me!

### A few questions to confirm my understanding

_Let me know if these miss the mark / don't align with your understanding_

This "simpler" implementation will only work on databases that support CTAs with column specs, which (to my most recent knowledge) is Snowflake + BigQuery. This is _not_ supported by Redshift, Databricks, ... where we'd instead need multiple queries to validate the column spec before actually creating/populating it.

We're requiring users to provide **the full set of column names + data types** if they want to define _any_ constraints. (Again, we could work around this with an additional `where false limit 0` query, similar to how we handle `persist_docs` for Snowflake views today — it's just not a very fun time, and gets gnarly when quoting/casing enter the picture.)

### Next steps

_Questions that might be out of scope for this issue/PR, but worth discussing_

Should we be clever about "translating" existing `not_null` tests into `not null` column constraints? Or leave that as a refactoring exercise to the end user?

Should we go beyond just `constraints` here? Should we also support things like (e.g.) [adding a masking policy to a column in Snowflake](https://docs.snowflake.com/en/sql-reference/sql/create-table.html#access-control-requirements:~:text=for%20a%20column.-,MASKING%20POLICY,-%3D%20policy_name)? We might not want to have to hard-code all possible options ourselves in `dbt-core`, but instead define a standalone macro — `{{ get_one_column_spec_ddl(col) }}` — that users can easily reimplement/extend if they want to define + use custom properties in the column's `meta` dict?
```sql
-- default built into dbt-core
{% macro get_one_column_spec_ddl(col) %}
    {{ col['name'] }} {{ col['data_type'] }} {{ col['constraint'] or "" }}
{% endmacro %}
```
```sql
-- my custom version
{% macro get_one_column_spec_ddl(col) %}
    {% set masking_policy_name = col['meta'].get('masking_policy_name') %}
    {{ col['name'] }} {{ col['data_type'] }} {{ col['constraint'] or "" }}
    {{ if masking_policy_name %} with masking policy {{ masking_policy_name }}
{% endmacro %}
```
```yml
# models/my_custom_model_config.yml
models:
  - name: my_model
    columns:
      - name: email
        data_type: string
        constraints: not null
        meta:
          masking_policy_name: masque_de_mail # this will be picked up by my custom macro
```

core/dbt/contracts/graph/unparsed.py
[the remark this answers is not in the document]
sungchun12: Ooooo great question. For developer UX, I'd bias towards a list even though it may be more "efficient" to allow arbitrary strings, developer mental load better reasons about multiple constraints as a list.

**I'll make the change for the `constraints` config to be an optional list**

**String Format:** 
Harder to reason about separate constraints
```
# models/my_custom_model_config.yml
models:
  - name: my_model
    columns:
      - name: email
        data_type: string
        constraints: not null unique 
```

**List Format:** 
Easier to reason about separate constraints
```
# models/my_custom_model_config.yml
models:
  - name: my_model
    columns:
      - name: email
        data_type: string
        constraints: ['not null','unique']
```

sungchun12:
@jtcohen6 Thanks for the thorough review. Glad to know we're in the right direction.

I imagine implementation will be easy across, Postgres, Snowflake, BigQuery, and [Databricks](https://docs.databricks.com/tables/constraints.html)

**Yes, full set of column names + data types** to define ANY constraints because as a building block for multi-project workflows, we need clear data promises, not half-hearted ones. 

**Next Steps**
I recommend we NOT be clever about refactoring existing `not_null` tests into column constraints because there are materializations that will never work with `constraints`(think: sources because they're already created before dbt touches them), and in large dbt projects, this will lead to more confusion than not. If this pain becomes more emphasized, we could probably add a macro to `codegen` to solve this problem.

For going beyond `constraints`, I imagine things like `masking policies` and other database-specific configs would be macros built into adapter-specific code vs. in `dbt-core`. However, I do like like your default macro `get_one_column_spec_ddl` a LOT. I'll work to incorporate this standard macro. Actually, as I'm writing this out, we're super aligned on this point. Hah!

core/dbt/include/global_project/macros/materializations/models/table/columns_spec_ddl.sql
```diff
@@ -0,0 +1,13 @@
+{% macro get_columns_spec_ddl() %}
```
sungchun12, line 1: Reference PR on how to test global macros: https://github.com/dbt-labs/dbt-core/pull/5838/files#diff-649f9b4416cee71ccd51b3b9cfce09b374d393e2a125d7423c9a428928d7bbaf

core/dbt/contracts/graph/parsed.py
```diff
@@ -221,8 +224,34 @@ class ParsedNodeDefaults(NodeInfoMixin, ParsedNodeMandatory):
...
+                language_error = {"language": language}
+
+            for column, column_info in self.columns.items():
+                if column_info.data_type is None:
+                    data_type_error = {column: {"data_type": column_info.data_type}}
+                    data_type_errors.update(data_type_error)
+
+        if materialization_error or language_error or data_type_errors:
+            raise CompilationException(
+                f"Only the SQL table materialization is supported for constraints. \n`data_type` values must be defined for all columns and NOT be null or blank.\n  Materialization Error: {materialization_error}\n  Language Error: {language_error}\n  Data Type Errors: {data_type_errors}"
```
b-per, line 249: Would we want to restrict the message to the specific type of error? i.e. not showing `materialization_error` if the error is about data type?

sungchun12: Yep I was just thinking about this last night! I'll figure out how to make conditional errors display!

sungchun12:
> I really like the simple way this is implemented. One thing that I can't resonate completely with is the _constraints_ and _test_ in schema.yml. In some cases, I just want to write _not_null_ in the test and that to be added as a _contraints_ automatically. Or adding _constraints_ and that automatically add a _test_? Would there be a scenario where I don't want to test on a _contraints_?
> 
> Maybe a test for the constraints flag (yes/no) could be a way to go. Want I want to away write the same thing in two places. Hope it makes sense.

@boje great questions! We wrestled with that internally about the schema config experience. We decided to NOT be clever about automatically converting `not_null` tests into `not null` constraints because constraints will only work with SQL tables where as the `not_null` tests can work with all materializations. For SQL tables in the first version of this release, you can replace all your `not_null` tests with `not null` constraints. You'll leave `not_null` tests for testing things like sources, incrementals, and views.

boje:
> > I really like the simple way this is implemented. One thing that I can't resonate completely with is the _constraints_ and _test_ in schema.yml. In some cases, I just want to write _not_null_ in the test and that to be added as a _contraints_ automatically. Or adding _constraints_ and that automatically add a _test_? Would there be a scenario where I don't want to test on a _contraints_?
> > Maybe a test for the constraints flag (yes/no) could be a way to go. Want I want to away write the same thing in two places. Hope it makes sense.
> 
> @boje great questions! We wrestled with that internally about the schema config experience. We decided to NOT be clever about automatically converting `not_null` tests into `not null` constraints because constraints will only work with SQL tables where as the `not_null` tests can work with all materializations. For SQL tables in the first version of this release, you can replace all your `not_null` tests with `not null` constraints. You'll leave `not_null` tests for testing things like sources, incrementals, and views.

@sungchun12 I see. Then it may be a job for the dbt_projectevaluate to check this (if possible). I think best practise would be to add a test for `not_null` when you add constraints about the same thing. In general, when the constraints NOT are enforced, the understanding can be a bit misleading in Snowflake. Sidenote: The Snowflake docs says that it is a few of the tables tyoes: Snowflake supports defining constraints on permanent, transient, and temporary tables.

core/dbt/exceptions.py
sungchun12: Looks like the selector error message was fixed, but is there a reason `warn_or_error` doesn't exist in the `main` branch but in `1.3.latest`? @jtcohen6

jtcohen6: @sungchun12 It still exists, it just moved:
```
- from dbt.exceptions import warn_or_error
+ from dbt.events.functions import warn_or_error
```

sungchun12: Ahhhh I know what happened, I needed to update my snowflake adapter to the main branch vs. relying on `1.3.latest` Thanks Jerco! I'm unblocked!

core/dbt/contracts/graph/parsed.py
```diff
@@ -221,8 +224,43 @@ class ParsedNodeDefaults(NodeInfoMixin, ParsedNodeMandatory):
...
     config_call_dict: Dict[str, Any] = field(default_factory=dict)
+    constraints_enabled: bool = False
+
+    def constraints_validator(self):
+        materialization_error = {}
+        language_error = {}
+        data_type_errors = set()
+
+        if self.resource_type == "model" and self.config.constraints_enabled is True:
+            if self.config.materialized != "table":
```
sungchun12, line 235: Should this be so restrictive to `table`? 

Some folks may be bold enough to create their own `incremental` materializations based on the the new configs available.

tests/functional/postgres/test_postgres_constraints.py
```diff
@@ -0,0 +1 @@
+# Verify global macro is dispatched and can be overridden elegantly in any adapter(I can do this by creating a custom macro in the dbt-postgres equivalent adapter)
```
sungchun12, line 1: Reference for adapter tests: https://github.com/dbt-labs/dbt-snowflake/pull/341

core/dbt/contracts/graph/parsed.py
sungchun12: Let's leave it alone as `table` only for now and see what can be done for `incremental` after we get more user feedback

cladden:
This is very exciting - looking forward to this feature!

core/dbt/contracts/graph/nodes.py
```diff
@@ -144,6 +146,8 @@ class ColumnInfo(AdditionalPropertiesMixin, ExtensibleDbtClassMixin, Replaceable
     description: str = ""
     meta: Dict[str, Any] = field(default_factory=dict)
     data_type: Optional[str] = None
+    constraints: Optional[List[str]] = None
+    check: Optional[str] = None
```
mikealfare, line 150: `check` is pretty vague. What are we checking? The constraints? The data_type? A more specific name might be useful here.

sungchun12: `check` is a normal term for checking arbitrary things like `check (id > 0)`, so for SQL table config context, it's not vague.

https://www.postgresql.org/docs/current/ddl-constraints.html#DDL-CONSTRAINTS-CHECK-CONSTRAINTS

Let me know if you referring to a different angle of vague.

core/dbt/contracts/graph/nodes.py
```diff
@@ -241,10 +245,53 @@ class ParsedNode(NodeInfoMixin, ParsedNodeMandatory, SerializableType):
     unrendered_config: Dict[str, Any] = field(default_factory=dict)
     created_at: float = field(default_factory=lambda: time.time())
     config_call_dict: Dict[str, Any] = field(default_factory=dict)
+    constraints_enabled: bool = False
     relation_name: Optional[str] = None
     raw_code: str = ""
 
+    def constraints_validator(self):
```
mikealfare, line 252: At first glance, this appears to be four types of errors (schema, materialization, language, and data type) in the same method. Can we break them out into separate methods? Perhaps this is a controller that provides a single interface, but there's a lot going on here.

core/dbt/include/global_project/macros/materializations/models/table/columns_spec_ddl.sql
```diff
@@ -0,0 +1,22 @@
+{%- macro get_columns_spec_ddl() -%}
+    {{ adapter.dispatch('get_columns_spec_ddl', 'dbt')() }}
```
mikealfare, line 2: This should be a consistent indent (2 spaces).

core/dbt/include/global_project/macros/materializations/models/table/columns_spec_ddl.sql
```diff
@@ -0,0 +1,22 @@
...
+{% macro default__get_columns_spec_ddl() -%}
+  {{ return(columns_spec_ddl()) }}
+{%- endmacro %}
+
+{% macro columns_spec_ddl() %}
+  {# loop through user_provided_columns to create DDL with data types and constraints #}
+  {% if config.get('constraints_enabled', False) %}
+    {%- set user_provided_columns = model['columns'] -%}
+    (
+    {% for i in user_provided_columns %}
```
mikealfare, line 14: `{% for i in model['columns'] %}` and remove line 12.

plugins/postgres/dbt/include/postgres/macros/adapters.sql
```diff
@@ -4,14 +4,27 @@
...
 
-  create {% if temporary -%}
-    temporary
-  {%- elif unlogged -%}
-    unlogged
-  {%- endif %} table {{ relation }}
-  as (
-    {{ sql }}
-  );
+  {% if config.get('constraints_enabled', False) %}
```
mikealfare, line 7: The initial piece (lines 8-12) is the same for both conditions. This if block should be moved down to just the piece that is the difference.

core/dbt/contracts/graph/nodes.py
sungchun12: Will do!

core/dbt/contracts/graph/nodes.py
sungchun12: Same language used by Databricks: https://docs.databricks.com/tables/constraints.html

plugins/postgres/dbt/include/postgres/macros/adapters.sql
sungchun12: done

core/dbt/include/global_project/macros/materializations/models/table/columns_spec_ddl.sql
sungchun12: done

core/dbt/include/global_project/macros/materializations/models/table/columns_spec_ddl.sql
sungchun12: done

b-per:
Not sure if we can do anything about it but the error message threw me off quite a bit when I provided a list to my `check` config rather than an individual item.

```
12:26:52  Running with dbt=1.4.0-b1
12:26:53  Encountered an error:
Parsing Error
  Invalid models config given in models/schema.yml @ models: {'name': 'customers', 'description': "This table has basic information about a customer, as well as some derived facts based on a customer's orders", 'config': {'constraints_enabled': True}, 'columns': [{'name': 'customer_id', 'description': 'This is a unique identifier for a customer', 'data_type': 'number(38,0)', 'constraints': ['unique', 'not null'], 'check': ['my test check'], 'tests': ['not_null', 'unique']}, {'name': 'first_name', 'data_type': 'varchar(100)', 'description': "customer's first name. pii.", 'check': 'abcd'}, {'name': 'last_name', 'data_type': 'varchar(5)', 'description': "customer's last name. pii."}, {'name': 'first_order', 'description': "date (utc) of a customer's first order", 'data_type': 'date'}, {'name': 'most_recent_order', 'description': "date (utc) of a customer's most recent order", 'data_type': 'date'}, {'name': 'number_of_orders', 'description': 'count of the number of orders a customer has placed', 'data_type': 'number(18,0)'}, {'name': 'total_order_amount', 'description': "total value (aud) of a customer's orders", 'data_type': 'number(18,0)'}], 'original_file_path': 'models/schema.yml', 'yaml_key': 'models', 'package_name': 'jaffle_shop'} - at path ['columns'][0]['check']: ['my test check'] is not valid under any of the given schemas
```

The message is `['columns'][0]['check']: ['my test check'] is not valid under any of the given schemas` and it took me some time to realize that it was because it was a list rather than a string.

sungchun12:
> Not sure if we can do anything about it but the error message threw me off quite a bit when I provided a list to my `check` config rather than an individual item.
> 
> ```
> 12:26:52  Running with dbt=1.4.0-b1
> 12:26:53  Encountered an error:
> Parsing Error
>   Invalid models config given in models/schema.yml @ models: {'name': 'customers', 'description': "This table has basic information about a customer, as well as some derived facts based on a customer's orders", 'config': {'constraints_enabled': True}, 'columns': [{'name': 'customer_id', 'description': 'This is a unique identifier for a customer', 'data_type': 'number(38,0)', 'constraints': ['unique', 'not null'], 'check': ['my test check'], 'tests': ['not_null', 'unique']}, {'name': 'first_name', 'data_type': 'varchar(100)', 'description': "customer's first name. pii.", 'check': 'abcd'}, {'name': 'last_name', 'data_type': 'varchar(5)', 'description': "customer's last name. pii."}, {'name': 'first_order', 'description': "date (utc) of a customer's first order", 'data_type': 'date'}, {'name': 'most_recent_order', 'description': "date (utc) of a customer's most recent order", 'data_type': 'date'}, {'name': 'number_of_orders', 'description': 'count of the number of orders a customer has placed', 'data_type': 'number(18,0)'}, {'name': 'total_order_amount', 'description': "total value (aud) of a customer's orders", 'data_type': 'number(18,0)'}], 'original_file_path': 'models/schema.yml', 'yaml_key': 'models', 'package_name': 'jaffle_shop'} - at path ['columns'][0]['check']: ['my test check'] is not valid under any of the given schemas
> ```
> 
> The message is `['columns'][0]['check']: ['my test check'] is not valid under any of the given schemas` and it took me some time to realize that it was because it was a list rather than a string.

@b-per I intentionally kept it as a string because checks typically follow this pattern in a way that doesn't help even when you put it in a list format. I aimed to keep it simple! For `dbt-snowflake`, assume `check` is a string in your macro. 
```sql
CREATE TABLE products (
    product_no integer,
    name text,
    price numeric,
    CHECK (price > 0),
    discounted_price numeric,
    CHECK (discounted_price > 0),
    CONSTRAINT valid_discount CHECK (price > discounted_price)
);
```

sungchun12:
Can a reviewer help explain how a test can fail here?: https://github.com/dbt-labs/dbt-core/actions/runs/3734850681/jobs/6337394974#step:7:227

but pass here: https://github.com/dbt-labs/dbt-core/actions/runs/3734850876/jobs/6337423780#step:8:1033

sungchun12:
Here's the workflow (start to finish) that we typically use for changes in core that affect adapters:

1. Open PR for dbt-core
2. Open PRs for each of the adapters

- dbt-snowflake
- dbt-redshift
- dbt-bigquery
- dbt-spark

3. Temporarily update the dev-requirements.txt file within each adapter to point to your PR in Step 1, like [this](https://github.com/dbt-labs/dbt-bigquery/pull/308/commits/7f795796abe4e15b8627d6cf086accce772b5685).
4. Once all PRs are passing CI, reviewed, and approved, then merge the dbt-core PR
5. Undo the temporary update to `dev-requirements.txt` in each of the adapter PRs (so that they point at the `main` branch of dbt-core again) like [this](https://github.com/dbt-labs/dbt-bigquery/pull/313/commits/27d28176b0148263cb05b41be2230ffc43f42700)
6. Once CI passes, then all the adapter PRs can be merged as well.

core/dbt/contracts/graph/nodes.py
colin-rogers-dbt: +1 that this is pretty vague, maybe `constraints_check`? In the case you shared the user already has context that the check relates to constraints. In our case there isn't an obvious link between check and constraints

core/dbt/contracts/graph/nodes.py
```diff
@@ -241,10 +245,77 @@ class ParsedNode(NodeInfoMixin, ParsedNodeMandatory, SerializableType):
     unrendered_config: Dict[str, Any] = field(default_factory=dict)
     created_at: float = field(default_factory=lambda: time.time())
     config_call_dict: Dict[str, Any] = field(default_factory=dict)
+    constraints_enabled: bool = False
     relation_name: Optional[str] = None
     raw_code: str = ""
 
+    def constraints_validator_handler(self):
```
colin-rogers-dbt, line 252: this might be cleaner if we moved it to it's own file, maybe `contracts/graph/contracts.py`? I'll defer to others with more experience in this part of the code

gshank (dismissed):
The "constraints" attribute should be in the CompiledNode class, not the ParsedNode class. I also don't think that calling that validation code from "write_node" is the right place. We generally do validation in a "validate" method on the class, which calls the jsonschema validation. When we add additional validation it's best to do it by overriding "validate" and adding the validation there.

Is this supposed to work with snapshots? Because where you're calling it in "write_node' the validation will not run for snapshots.

core/dbt/contracts/graph/nodes.py
sungchun12: I can update this to `constraints_check`

sungchun12: All `check` configs have been updated to `constraints_check`

joshuataylor:
Is having `constraints_enabled` going to be a temporary thing, or will it exist forever? It's going to be kind of annoying to have to add that to every model that has these. Could we just check if the columns have constraints?

It would be like having `tests_enabled` whenever we wanted to define tests.

sungchun12:
@joshuataylor You can add it to any hierarchy of config, so you don't have to do it model by model. You can do it by the subfolder level. See the example below.

```yml
models:
  tpch:
    staging:
      +materialized: view
      +docs:
        node_color: "#cd7f32"

    marts:
      core:
        +constraints_enabled: true # it will enforce data type constraints across all thee models in the "/core" subfolder
        materialized: table

```

joshuataylor:
> @joshuataylor You can add it to any hierarchy of config, so you don't have to do it model by model. You can do it by the subfolder level. See the example below.
> 
> We decided to make it opt-in, so people have lots of control over when to use this feature.
> 
> It'll likely exist for a long time, but I can't predict infinity ;)
> 
> ```yml
> models:
>   tpch:
>     staging:
>       +materialized: view
>       +docs:
>         node_color: "#cd7f32"
> 
>     marts:
>       core:
>         +constraints_enabled: true # it will enforce data type constraints across all thee models in the "/core" subfolder
>         materialized: table
> 
> ```

Love it! Incrementally rolling this out is our plan as well.

sungchun12:
https://github.com/dbt-labs/dbt-core/issues/6511 FYI For future reference

sungchun12:
> The "constraints" attribute should be in the CompiledNode class, not the ParsedNode class. I also don't think that calling that validation code from "write_node" is the right place. We generally do validation in a "validate" method on the class, which calls the jsonschema validation. When we add additional validation it's best to do it by overriding "validate" and adding the validation there.
> 
> Is this supposed to work with snapshots? Because where you're calling it in "write_node' the validation will not run for snapshots.

This is NOT supposed to work with snapshots and shouldn't given the latest code changes

sungchun12:
Alrighty! We finally have this in working order to validate constraints at `parse time`, which required more effort than expected but glad we're finally here. Huge shout out to @gshank for guiding me through the twists and turns!

For this next review cycle, can someone review BOTH this PR and https://github.com/dbt-labs/dbt-snowflake/pull/341 and specifically compare tests for the `postgres` adapter in this PR vs. the `snowflake` adapter and recommend how tests should be consistent vs. differ. This will serve to streamline the review process for the rest of the adapters afterwards!

Quick note on a failing test: How do I resolve this?
- It fails for structured logging schema here: https://github.com/dbt-labs/dbt-core/actions/runs/3857477232/jobs/6574912680#step:7:527
- but then it passes here: https://github.com/dbt-labs/dbt-core/actions/runs/3857477234/jobs/6574912906#step:8:1061

schemas/dbt/manifest/v8.json
colin-rogers-dbt: google translate:
> Who verifies the verifiers? Who makes them manifest?

core/dbt/parser/base.py
```diff
@@ -306,6 +311,19 @@ def update_parsed_node_config(
...
+        # If we have constraints_enabled in the config, copy to node level, for backwards
+        # compatibility with earlier node-only config.
+        if config_dict.get("constraints_enabled", False):
+            parsed_node.constraints_enabled = True
+
+            parser_name = type(self).__name__
+            if parser_name == "ModelParser":
+                original_file_path = parsed_node.original_file_path
+                error_message = "\n    `constraints_enabled=true` can only be configured within `schema.yml` files\n      NOT within a model file(ex: .sql, .py) or `dbt_project.yml`."
+                raise CompilationException(
```
gshank, line 323: This should actually be a ParsingException. Ideally exceptions generated at parse time should all be ParsingExceptions and not CompilationExceptions, but it was tricky to set up the distinction in the Jinja code so it hasn't been completed yet.

gshank:
Could you merge main into your branch so we can get a clean build?

gshank:
You're getting a test failure in the "structured logging schema check" because that test runs the entire test suite with log format set to json. Probably you just need to ensure that the string it's looking for is the same in the json output and the non-json output. You can set DBT_LOG_FORMAT environment variable and run the test.

sungchun12:
@gshank everything passes now!

core/dbt/parser/schemas.py
```diff
@@ -914,6 +920,75 @@ def parse_patch(self, block: TargetBlock[NodeTarget], refs: ParserRef) -> None:
...
+                self.constraints_materialization_validator(patched_node),
+                self.constraints_language_validator(patched_node),
+                self.constraints_data_type_validator(patched_node),
+            ]
+            error_messages = [validator for validator in validators if validator != "None"]
+
+        if error_messages:
+            original_file_path = patched_node.original_file_path
+            raise ParsingError(
+                f"Original File Path: ({original_file_path})\nConstraints must be defined in a `yml` schema configuration file like `schema.yml`.\nOnly the SQL table materialization is supported for constraints. \n`data_type` values must be defined for all columns and NOT be null or blank.{self.convert_errors_to_string(error_messages)}"
```
jtcohen6, line 942: > Only the SQL table materialization is supported for constraints.

Good on you for having the explicit callout here!

This does not work for:
- Views, because databases don't allow data type specifications or constraints when creating views. (They only allow column names, and those aren't actually cross-checked, just used for implicit renaming.)
- Incremental models - unless, of course, they have configured `'on_schema_change: 'fail'`! Should we disallow the more capable mechanisms for handling schema changes, though? This feels like a pretty big limitation.

*Unless!*:
1. We add a check for column types _after_ a view or temp table has been created. (For views: This could be within-transaction and before the alter-rename-swap on "older-school" databases, or after a view has already been created on "newer-school" databases that support `create or replace`.) That feels like a pretty big limitation, though - a big part of the appeal of "constraints," as opposed to `dbt test`, is that they should be checked _before_ the object has been created/replaced, or before expensive incremental processing has happened.
2. We add a check for column names + types _before_ the view / temp table has been created, by means of [`get_columns_in_query`](https://github.com/dbt-labs/dbt-core/blob/43e3fc22c4eae4d3d901faba05e33c40f1f1dc5a/core/dbt/include/global_project/macros/adapters/columns.sql#L24-L31), a.k.a. `where false limit 0`. That's a "free" query on most databases (including BQ!), there's just a bit of risk in extracting metadata + doing our own data type comparisons, rather than pushing it all down to the database.

I lean toward option 2, as a way to expand this support beyond just tables. In a future where we support project-level governance rules ("all public models must have constraints enabled"), without that support, we'd also be requiring that every public model be `materialized: table` — and that feels like a pretty big limitation. I've already talked to a handful of folks who have grown quite fond of a view-driven "consumption layer," and feel it offers them an additional lever of control between the underlying data (table) and its downstream users.

core/dbt/parser/base.py
```diff
@@ -306,6 +306,19 @@ def update_parsed_node_config(
...
+        if config_dict.get("constraints_enabled", False):
+            parsed_node.constraints_enabled = True
+
+            parser_name = type(self).__name__
+            if parser_name == "ModelParser":
+                original_file_path = parsed_node.original_file_path
+                error_message = "\n    `constraints_enabled=true` can only be configured within `schema.yml` files\n      NOT within a model file(ex: .sql, .py) or `dbt_project.yml`."
+                raise ParsingError(
+                    f"Original File Path: ({original_file_path})\nConstraints must be defined in a `yml` schema configuration file like `schema.yml`.\nOnly the SQL table materialization is supported for constraints. \n`data_type` values must be defined for all columns and NOT be null or blank.{error_message}"
+                )
```
jtcohen6, line 320: Confirming that this is necessary only:
- in order to enable parse-time validation that every column for this model has a `data_type` defined
- because of this tech debt ticket: https://github.com/dbt-labs/dbt-core/issues/6511

As @gshank commented earlier, parse time is the right time for this sort of validation, versus compile-time, per our slightly nonstandard meanings of "parsing" + "compilation." I do see the downside of, we return parse-time errors as soon as we hit them, rather than batching up all the errors we can and returning them all. If it were a _warning_ instead, then we could return them all together!

FWIW - I'm aligned with `constraints_enabled` being a _config_, with the idea that it should be possible to enable a broad swath of models (e.g. "everything in the `marts/public` folder") to have type-checked column specs. That's true even though it depends on the `columns` attribute, which is _not_ a config, but a property specific to one model.

If we can't resolve that tech debt before the v1.5 release, I'd probably opt for removing the parse-time check, and letting the database return its own validation error at runtime. It's a less-nice UX to see the error later in the dev cycle, but I think it's better to offer the full flexibility around configuring `constraints_enabled`.

jtcohen6 (commented):
Amazing work pulling this together @sungchun12 & team!! This gave me an opportunity to really dig into some of the gnarlier details at work here.

I wouldn't consider any of my comments here hard blockers for getting this merged. If anything, I think right next steps here are:
- Discuss the observations + limitations I've noted below with you + @MichelleArk
- Merge this + adapter PRs, to get this from WIP &rarr; beta state (as it will be included in v1.5.0-b1, targeting ~February)
- Go open a bunch more follow-up issues, as part of the larger multi-project initiative
- Gather community feedback!

core/dbt/contracts/graph/model_config.py
```diff
@@ -446,6 +446,7 @@ class NodeConfig(NodeAndTestConfig):
         default_factory=Docs,
         metadata=MergeBehavior.Update.meta(),
     )
+    constraints_enabled: Optional[bool] = False
```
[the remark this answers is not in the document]
sungchun12: I'm open to changing it after community feedback! There are exceptions that I flow through into `DatabaseErrors` vs. `ParsingErrors` because we get error handling for free(think: data type validation, number of columns mismatch SQL). 

This does NOT validate order or name consistency across the SQL statement to your point. This is a much larger effort for `static typing` for `dbt`.

core/dbt/contracts/graph/unparsed.py
```diff
@@ -93,6 +93,8 @@ class HasDocs(AdditionalPropertiesMixin, ExtensibleDbtClassMixin, Replaceable):
     description: str = ""
     meta: Dict[str, Any] = field(default_factory=dict)
     data_type: Optional[str] = None
+    constraints: Optional[List[str]] = None
+    constraints_check: Optional[str] = None
```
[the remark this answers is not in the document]
sungchun12: Yes, just postgres + databricks. We'll include adapter-specific disclaimers in this PR: https://github.com/dbt-labs/docs.getdbt.com/pull/2601

test/unit/test_macro_calls.py
```diff
@@ -27,7 +27,7 @@ def setUp(self):
             "{% macro some_test(model) -%} {{ return(adapter.dispatch('test_some_kind4', 'foo_utils4')) }} {%- endmacro %}",
             "{% macro some_test(model) -%} {{ return(adapter.dispatch('test_some_kind5', macro_namespace = 'foo_utils5')) }} {%- endmacro %}",
         ]
-
+        # TODO: do I need to adjust this for `get_columns_spec_ddl`
```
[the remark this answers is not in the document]
sungchun12: I'll remove all my TODOs

plugins/postgres/dbt/include/postgres/macros/utils/columns_spec_ddl.sql
```diff
@@ -0,0 +1,23 @@
+{% macro postgres__get_columns_spec_ddl() %}
```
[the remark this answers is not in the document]
sungchun12: Yeah, this was a toughie to discern. I biased towards the global macro default bundling the data type DDL and wrapping the select statement because other adapters may benefit from it(think: bigquery, snowflake, etc.). At the same time, the `postgres` implementation will likely be the most universally applicable minus the `check` constraint which isn't common across data warehouses. Overall, there is no good default implementation, just good template code to override in a specific adapter. 

**I recommend keeping it as is given the above, but given your team maintains this and you have a bigger picture view of how adapters will use this, I'm happy to make the `postgres` implementation the default.**

plugins/postgres/dbt/include/postgres/macros/adapters.sql
```diff
@@ -9,7 +9,13 @@
   {%- elif unlogged -%}
     unlogged
   {%- endif %} table {{ relation }}
-  as (
+  {% if config.get('constraints_enabled', False) %}
+    {{ get_columns_spec_ddl() }} ;
+    insert into {{ relation }} {{ get_column_names() }}
```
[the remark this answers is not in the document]
sungchun12: Good to know! I guess this is where dbt's opinions aim for explicit over implicit.

plugins/postgres/dbt/include/postgres/macros/utils/columns_spec_ddl.sql
```diff
@@ -0,0 +1,23 @@
...
+  {# loop through user_provided_columns to create DDL with data types and constraints #}
+    {%- set user_provided_columns = model['columns'] -%}
+    (
+    {% for i in user_provided_columns %}
+      {% set col = user_provided_columns[i] %}
+      {% set constraints = col['constraints'] %}
+      {% set constraints_check = col['constraints_check'] %}
+      {{ col['name'] }} {{ col['data_type'] }} {% for x in constraints %} {{ x or "" }} {% endfor %} {% if constraints_check -%} check {{ constraints_check or "" }} {%- endif %} {{ "," if not loop.last }}
```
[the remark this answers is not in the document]
sungchun12: I vote option 1 because it has less database performance hits because option 2 using subqueries depends a lot more on the database to optimize it(which varies). 

I'll make this change!

core/dbt/parser/base.py
sungchun12: You are correct on why this is a Parsing Error today. I'm fine with the less-nice UX in favor of `constraints_enabled=true` being available in all configs come v1.5.

core/dbt/parser/schemas.py
sungchun12: For Option 2, I'm aligned on approach to get `incremental` models working. Did you want this implemented in this PR or another?

jtcohen6: Happy to have this in a follow-up issue/PR! I don't think we need it for the initial cut / enough to unlock a beta experience

plugins/postgres/dbt/include/postgres/macros/utils/columns_spec_ddl.sql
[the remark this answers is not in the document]
dave-connors-3: I'm not sure we create any temp tables in the current implementation, so where would we do the mid-transaction check in option 1?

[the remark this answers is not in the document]
jtcohen6: @dave-connors-3 The way we replace preexisting views/tables on Postgres + Redshift is by creating a "temp" relation (with a suffix `__dbt_tmp`, not always _actually_ a temporary table), and then `alter rename` swapping the existing/old + new relations, before dropping the old one. That ensures zero downtime, and prevents the view/table from being missing if the model SQL runs into an error.

So we _could_ try to stick another check in the middle of that process. But I actually think `get_columns_in_query` might be preferable across the board, because it will also catch mismatches _before_ we run a potentially expensive query to create the table with new data.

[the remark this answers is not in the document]
sungchun12: I made the change by raising a compile error if the names and order of the columns don't match exactly in a global macro. I added 2 tests to prove this. 

image

core/dbt/include/global_project/macros/materializations/models/table/columns_spec_ddl.sql
```diff
@@ -0,0 +1,45 @@
...
+
+{%- macro get_assert_columns_equivalent(sql) -%}
+  {{ adapter.dispatch('get_assert_columns_equivalent', 'dbt')(sql) }}
+{%- endmacro -%}
+
+{% macro default__get_assert_columns_equivalent(sql) -%}
+  {{ return(assert_columns_equivalent(sql)) }}
+{%- endmacro %}
+
+{% macro assert_columns_equivalent(sql) %}
```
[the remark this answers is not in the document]
jtcohen6: @b-per Great point. And that `quote_columns` config is specific to seeds - not to be confused with the column-level `quote` property: https://docs.getdbt.com/reference/resource-properties/quote

(Sorry everyone: #2986)

plugins/postgres/dbt/include/postgres/macros/utils/columns_spec_ddl.sql
dave-connors-3, line 1:
```suggestion
{% macro assert_columns_equivalent(sql) %}
  {# loop through user_provided_columns to get column names #}
    {%- set user_provided_columns = model['columns'] -%}
    {%- set column_names_config_only = [] -%}
    {%- for i in user_provided_columns -%}
      {%- set col = user_provided_columns[i] -%}
      {%- set col_name = col['name'] -%}
      {%- set column_names_config_only = column_names_config_only.append(col_name) -%}
    {%- endfor -%}
    {%- set sql_file_provided_columns = get_columns_in_query(sql) -%}

    {%- if column_names_config_only != sql_file_provided_columns -%}
      {%- do exceptions.raise_compiler_error('Please ensure the name and order of columns in your `yml` file match the columns in your SQL file.\nSchema File Columns: ' ~ column_names_config_only ~ '\nSQL File Columns: ' ~ sql_file_provided_columns ~ ' ' ) %}
    {%- endif -%}
{% endmacro %}
```

to make sure there's not unnecessary whitespace in the log file!

core/dbt/include/global_project/macros/materializations/models/table/columns_spec_ddl.sql
dave-connors-3, line 30:
```suggestion
{% macro assert_columns_equivalent(sql) %}
  {# loop through user_provided_columns to get column names #}
    {%- set user_provided_columns = model['columns'] -%}
    {%- set column_names_config_only = [] -%}
    {%- for i in user_provided_columns -%}
      {%- set col = user_provided_columns[i] -%}
      {%- set col_name = col['name'] -%}
      {%- set column_names_config_only = column_names_config_only.append(col_name) -%}
    {%- endfor -%}
    {%- set sql_file_provided_columns = get_columns_in_query(sql) -%}

    {%- if column_names_config_only != sql_file_provided_columns -%}
      {%- do exceptions.raise_compiler_error('Please ensure the name and order of columns in your `yml` file match the columns in your SQL file.\nSchema File Columns: ' ~ column_names_config_only ~ '\nSQL File Columns: ' ~ sql_file_provided_columns ~ ' ' ) %}
    {%- endif -%}
{% endmacro %}
```

For whitespace control in the `dbt.log` file!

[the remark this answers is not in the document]
sungchun12: @b-per Great feedback. I'll update the macro to uppercase all the columns across both the schema and sql files. I'll also figure out quote columns!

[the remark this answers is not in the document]
sungchun12: Updated this to uppercase all columns and added extra uppercase checks and quotes to the columns in the SQL file and added a `quote:true` config in `test_constraints.py`. @b-per can you pull in these changes and verify that it's working for snowflake?
image

sungchun12:
@dave-connors-3 can you re-suggest your code change? I made a lot of code changes which then made your code suggestion funky when committed.

core/dbt/contracts/graph/unparsed.py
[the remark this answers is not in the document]
jtcohen6: _[For follow-up issues, out of scope for this PR]_

This didn't occur to me on the first read-through, but is there a reason why we want this to be a separate configuration, rather than just a subset of the available `constraints` (on platforms that support it)?

Answering my own question: It looks like it's because Databricks requires that `check` constraints have to be applied _after_ table creation via `alter` statements. (It looks like Postgres supports them within the `create table` statement.)

I still think I might prefer a configuration like:
```python
constraints: Optional[List[Union[str, Dict[str, str]]]] = None
```
```yml
columns:
  - name: price
    data_type: numeric
    constraints:
      - not null
      - check: price > 0
        name: positive_price  # Postgres supports naming constraints
```

Postgres + Databricks also support table-level constraints that are composed of multiple columns. We should also support `constraints` at the model level, in addition to the column level. Should they be a config, or a non-config attribute?

core/dbt/parser/schemas.py
jtcohen6: _[For follow-up issues, out of scope for this PR]_

After a bit more investigation, we'll need to use a modified version of `get_columns_in_query`, that pulls data types off the database cursor/API, if we want to also enforce that data types match before a table/view is created/replaced.

core/dbt/parser/schemas.py
```diff
@@ -914,6 +920,75 @@ def parse_patch(self, block: TargetBlock[NodeTarget], refs: ParserRef) -> None:
...
+        materialization_error_msg = f"\n    Materialization Error: {materialization_error}"
+        materialization_error_msg_payload = (
+            f"{materialization_error_msg if materialization_error else None}"
+        )
+        return materialization_error_msg_payload
+
+    def constraints_language_validator(self, patched_node):
+        language_error = {}
+        language = str(patched_node.language)
+        if language != "sql":
```
jtcohen6, line 975: _[For follow-up issues, out of scope for this PR]_

It should be _possible_ to support dbt-py models, by saving the Python model in a temp table, and then inserting it into a preexisting table (created per contract). The downside of that approach is, the full (potentially expensive) transformation has already taken place before the contract could be validated.

jtcohen6 (commented):
I've left a handful of comments, all of which are intended as footholds for future issues, not blockers for this PR. I'd like to get this PR (+ associated adapter PRs) merged, so that we can include in v1.5 beta (targeting Feb 15) and enable a clean handoff. Core devs will take it from here!

The last thing I want to do, before approving for merge, is take a closer look at the functional tests. We want our test cases to be as concise as possible, and also clearly separate between:
1. Functional tests of core functionality that are not adapter-specific (e.g. raising parsing exceptions)
2. Tests of adapter-specific behavior that want to live in the "adapter zone" (`dbt-tests-adapter`), which ought to be inherited + run in adapter repos in consistent ways.

jtcohen6:
I've reviewed + refactored tests on the dbt-core PR, with the specific goal of having a clearer separation between:
1. What can be tested in `dbt-core` only: validation, parsing exceptions, basic end-to-end functionality
2. What needs testing on each adapter: dbt's "preflight" checks within materialization; platform-enforced runtime checks (via DDL); successful rollback if contract isn't met

There are now only two "adapter zone" test classes that need inheriting/modifying. I'll take a pass over the adapter PRs tomorrow or Wednesday.

Ideally, I'd like to get all of this merged by Wednesday (Feb 1), but I'm going to wait until `feature/click-cli` is merged (https://github.com/dbt-labs/dbt-core/issues/6631), so as to avoid creating additional merge conflicts for that crew.

tests/adapter/dbt/tests/adapter/constraints/test_constraints.py
```diff
@@ -0,0 +1,272 @@
...
+  )
+}}
+
+select
+  1 as error,
+  'blue' as color,
+  cast('2019-01-01' as date) as date_day
+"""
+
+my_model_error_sql = """
```
dave-connors-3, line 46: this looks the same as `my_model_sql` -- what's the expected error? @jtcohen6

tests/adapter/dbt/tests/adapter/constraints/test_constraints.py
```diff
@@ -0,0 +1,272 @@
...
+        with open("./target/run/test/models/my_model.sql", "r") as fp:
+            generated_sql = fp.read()
+
+        generated_sql_check = re.sub(r"\s+", " ", generated_sql).lower().strip()
+        expected_sql_check = re.sub(r"\s+", " ", expected_sql).lower().strip()
+        assert (
+            expected_sql_check == generated_sql_check
+        ), f"generated sql did not match expected: \n{generated_sql} \n{expected_sql}"
+
+    def test__constraints_enforcement_rollback(self, project, unique_schema):
```
dave-connors-3, line 231: is `unique_schema

tests/adapter/dbt/tests/adapter/constraints/test_constraints.py
```diff
@@ -0,0 +1,159 @@
...
+        # Make a contract-breaking change to the model
+        my_model_with_nulls = my_model_sql.replace("1 as id", "null as id")
+        write_file(my_model_with_nulls, "models", "my_model.sql")
+
+        failing_results = run_dbt(["run", "-s", "my_model"], expect_pass=False)
+        assert len(failing_results) == 1
+
+        # Verify the previous table still exists
+        old_model_exists_sql = """
+            select id from dbt.{0}.my_model where id = 1
```
dave-connors-3, line 131: @jtcohen6 I think this will also need the templated database!

jtcohen6: done

core/dbt/parser/schemas.py
sungchun12: I recommend the syntax be: `contract: true | false` because `contracted` invokes imagery of pregnancy(a beautiful human experience but dissonant related to data contracts) at first glance and is in past tense, which may confuse the reader. Whenever the term `data contracts` come up, it's a noun in plural or singular, not a verb. 

I'm on board with this example config, feels a lot more elegant compared to the first draft: 

```yml
columns:
  - name: price
    data_type: numeric
    constraints:
      - not null
      - check: price > 0
        name: positive_price  # Postgres supports naming constraints
```

I'm aligned with you on `get_columns_in_query` relying on the database cursor/API for more robustness and cheaper operations.

I'm aligned on python models as long as we document/warn the user of performance hits as a result. 

I'm aligned with you on "validators". My main goal is to provide maximally useful terminal logs that invoke clear and quick actions to fix. This makes iteration a delight rather than a chore. 

Thanks for such a thorough review and eye towards the future. I'm juiced up to finish these PRs!

jtcohen6 (commented):
I'll work on fixing up the merge conflicts. Let's make sure all the TODOs are captured in our tracking issues, then let's get this baby merged 😎

core/dbt/include/global_project/macros/materializations/models/table/columns_spec_ddl.sql
```diff
@@ -0,0 +1,51 @@
...
+  {#- loop through user_provided_columns to get column names -#}
+    {%- set user_provided_columns = model['columns'] -%}
+    {%- set column_names_config_only = [] -%}
+    {%- for i in user_provided_columns -%}
+      {%- set col = user_provided_columns[i] -%}
+      {%- set col_name = col['name'] -%}
+      {%- set column_names_config_only = column_names_config_only.append(col_name) -%}
+    {%- endfor -%}
+    {%- set sql_file_provided_columns = get_columns_in_query(sql) -%}
```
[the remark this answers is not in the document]
jtcohen6: Thanks @MichelleArk for opening https://github.com/dbt-labs/dbt-core/issues/6975!

core/dbt/include/global_project/macros/materializations/models/table/create_table_as.sql
```diff
@@ -25,6 +25,10 @@
 
   create {% if temporary: -%}temporary{%- endif %} table
     {{ relation.include(database=(not temporary), schema=(not temporary)) }}
+    {% if config.get('constraints_enabled', False) %}
+      {{ get_assert_columns_equivalent(sql) }}
+      {{ get_columns_spec_ddl() }}
+    {% endif %}
```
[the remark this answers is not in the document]
jtcohen6: Assigning myself to #6750 for further refinement, before we pick this up for implementation

MichelleArk (approved):

jtcohen6 (approved):

sungchun12:
Thanks team for getting this done. I just shouted out loud, "FINALLY!" With the utmost glee.

core/dbt/contracts/graph/model_config.py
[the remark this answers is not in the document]
davehowell: _nit_: If it has a default value `False` then it doesn't need to be an `Option type`, it could just be `bool`.

It's a shame the python typing lib uses the word `Optional` as a keyword; it's the default value that makes providing the attribute _optional_ when instantiating the class.  

On another note, a bool with a default value of `None` forces the use of `Optional[bool]` which only makes sense if you must have a lenient (public) API or if there's inheritance compatibility issues.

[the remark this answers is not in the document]
jtcohen6: @davehowell Good catch! It looks like @gshank already had the same thought, and implemented this change over in #7002, along with the config rename :)

PR #6283, opened 2022-11-17
CT 1509 reorg protobuf events to have top-level Event message

### LINKED ISSUE #6256: [CT-1509] Reorganize protobuf structured logging events
As per discussion with Dave Berton, attempt to reorganize the existing protobuf messages so that instead of an EventInfo submessage, those fields will be moved to a top-level Event, and the rest of the logging events will be submessages specified in a "oneof" in the Event message. This should make it easier to deserialize. 

The current creation of events for the "fire_event" function should be able to stay the same, except for tweaks (improvements) to the way that we set "level". Instead of setting the event keys in the a post_init of the BaseEvent class, we will construct an Event in "fire_event" and set the passed in message as the "detail" part of the event.

This is part of the structured logging initiative, such as we see in https://github.com/dbt-labs/dbt-core/pull/6136.

gshank:
resolves #6256

### Description

Reorganize the protobuf logging events so that the top level is the common Event message, which will contain a "oneof" of the entire list of detailed events.

### Checklist


gshank:
Note: there is a lot of noise in this pull request because of the need to touch every one of our logging events, fix the tests, etc. The significant parts regarding protobuf message organization are in the core/dbt/events/types.proto and core/dbt/events/event.proto files.

gshank:
Looks like there might be one of those logging-error-causing-hang problems now. Not happening with a local run, of course.

core/dbt/events/functions.py
```diff
@@ -116,23 +115,35 @@ def stop_capture_stdout_logs() -> None:
...
+        # need defaults in the json log output, but that includes a huge unnecessary
+        # pile of oneof field dictionaries, so we'll have to munge it.
         event_dict = event.to_dict(casing=betterproto.Casing.SNAKE, include_default_values=True)  # type: ignore
     except AttributeError as exc:
         event_type = type(event).__name__
-        raise Exception(f"type {event_type} is not serializable. {str(exc)}")
-    return event_dict
+        raise BaseException(f"type {event_type} is not serializable. {str(exc)}")
+
+    # Kludge warning: create a new dictionary with only the relevant keys. In the
```
emmyoop, line 141: Can you create an issue to investigate this and link to the issue in the comment?  That way it's less likely to get lost and when we read this we can know where to find more context.

gshank: Created a ticket and added a comment.

gshank:
The performance of this approach has proved to be unacceptable, so I'm closing this PR.

Issue #6268, opened 2022-11-16
[CT-1514] [Feature] Finalize and Merge Binary Logging


peterallenwebb:
### Is this your first time submitting a feature request?

- [X] I have read the [expectations for open source contributors](https://docs.getdbt.com/docs/contributing/oss-expectations)
- [X] I have searched the existing issues, and I could not find an existing issue for this feature
- [X] I am requesting a straightforward extension of existing dbt functionality, rather than a Big Idea better suited to a discussion

### Describe the feature

We created a binary structured-logging prototype in branch for issue #6138 and are now ready to incorporate feedback and merge the branch. In order to complete the work, we will:

- Encode event blocks using [Consistent Overhead Byte Stuffing](https://en.wikipedia.org/wiki/Consistent_Overhead_Byte_Stuffing). This will provide the ability to recover from errors in the event stream, make it easier to document the format, and allow downstream consumers to use standard protobuf and COBS libraries for processing.
- Work with Gerda to finalize message modeling in protobuf, preferring to use a single message type to frame more specific messages, so that we can use the simplest format possible and guarantee binary format stability in the future.

PR #6291, opened 2022-11-18
CT-1405: Refactor event logging code

### LINKED ISSUE #6139: [CT-1405] [Feature] Decouple event logging functionality

### Describe the feature

We should improve the encapsulation and flexibility of our event logging code, most of which is currently located in core/dbt/events/functions.py.

### Areas needing improvement:
- Event logging behavior is tied directly to global variables which are open to unpredictable access
- Log redirection for testing (as implemented by capture_stdout_logs) and other aspects of the logging setup, are likely thread-unsafe. We run our integration tests multi-threaded, so that might already be a problem
- We are using the python logging library, but few features of the library
- The event logging code is hard to test because it is coded directly against stdout and file output rather than i/o streams

### Design and implement a class which improves the situation by:
- Relying on instance variables rather than global variables for configuration
- Supporting any number of output configurations, each of which has its own type (text/json/protobuf), output stream (passed in rather than created internally), and level (info/warn/debug/etc)
- (?) Bypassing python's logging module entirely
- (?) Removing the EVENT_HISTORY global, either encapsulating it in this class, or replacing it with a subscription/callback mechanism

Current logging configuration, as peformed in setup_event_logger(), should be refactored to use the new class.

_No response_

### Who will this benefit?

The improved testability, clarity, and encapsulation should increase the velocity and overall experience of any dbt engineer or community member who needs to work with this code, or perform tests which involve monitoring dbt's output.

Yes!

### Anything else?

_No response_

peterallenwebb:
resolves #6139 

### Description

This PR is intended to make the logging code simpler to reason about, and easier to use in future multi-threaded scenarios. It does the following:

* Centralizes management of events into a single class, EventManager located in dbt.events.eventmgr
* Keeps the eventmgr module free of global variables or references to global variables
* Encapsulates code that deals with the python logging subsystem in private portions of of eventmgr
* Eliminates global logging variables (FILE_LOG, STDOUT_LOG, format_color, format_json) replacing them with a single EVENT_MANAGER
* Removes the global EVENT_HISTORY dequeue, replacing it with callback functionality in EventManager
* Uses flag values in functions.py only at configuration-time
* Retains support for legacy logger, for now

### Suggested Future Work

* Design and implement a dbt configuration manager, so we can remove references to global flags in functions.py, and make an EventManager part of the configuration context for a dbt command, removing the global EVENT_MANAGER
* Remove logbook related code
* Add working color support to new log system (I'm pretty sure its missing?)
* Remove uses fire_event() from the integration tests

### Testing

Since the code was substantially rearranged and rewritten, I used the following test strategy:

1. Record all BaseEvent objects generated during an integration test run on the main branch, pickle them, and write them to a file.
2. Replay the events against the old logging framework, recording the file and console output.
3. Replay the events against the new logging framework, recording the file and console output.
4. Repeat 2 & 3 for every combinations of flags.DEBUG, flags.QUIET, flags.LOG_CACHE_EVENTS, flags.USE_COLORS, flags.LOG_FORMAT
4. Compare console and file recordings for binary equality

### Checklist


core/dbt/flags.py
```diff
@@ -39,7 +39,6 @@
 WHICH = None
 INDIRECT_SELECTION = None
 LOG_CACHE_EVENTS = None
-EVENT_BUFFER_SIZE = 100000
```
emmyoop, line 42: @jtcohen6 is this flag considered experimental?  Does it need to be flagged as deprecated without being removed entirely.  It obviously would no longer do anything but if we remove it as a flag entirely, it causes an error for anyone using it as a parameter.

gshank: I don't think anybody would have actually used the event history buffer because prior to the protobuf work it totally lacked all useful logging information (such as the timestamp, level, invocation_id, thread, etc). So in my opinion we shouldn't have to go through a deprecation cycle.

gshank:
@peterallenwebb Could you merge main and resolve the existing errors?  I'll start looking at this meanwhile. Also -- I'm not sure why the tests aren't running, and it's a shorter list than usual. Is something changed or broken in CI?

core/dbt/flags.py
emmyoop: I brought it up because I remember an issue when someone in the community was asking about it.  An issue I cannot find so I'm possibly mistaken.

gshank (commented):
The EventManager code looks nice and neat and much better than the existing code. I'm a little unclear about how somebody using dbt (such as dbt-server or shipments) would specify a custom logger. How would that work?

There's no protobuf logger config or code. Do you have plans for doing that?

Also I think we'll have to get the use_colors config from the existing flags.

core/dbt/flags.py
gshank: You do dream about these things, right? ;-)

peterallenwebb:
@gshank I am not sure why CI is having issues, but I'll merge in main and see if that helps. The `make unit` and `make integration` commands ran find for me locally.

emmyoop:
When merge conflicts exists it won't run because it is running off the workflows in `main`.

core/dbt/events/functions.py
```diff
@@ -1,111 +1,129 @@
...
+# generates events, without configuring the event manager.
+EVENT_MANAGER: EventManager = EventManager()
 
-def setup_event_logger(log_path, level_override=None):
-    global format_json, format_color, STDOUT_LOG, FILE_LOG
+
+def setup_event_logger(log_path: str, level_override: Optional[EventLevel] = None):
+    cleanup_event_logger()
     make_log_dir_if_missing(log_path)
+    if flags.ENABLE_LEGACY_LOGGER:
```
ChenyuLInx, line 27: @peterallenwebb This looks great!! Is this the place cloud integrate their desired logger in?

peterallenwebb:
And yes, I'm planning to address the protobuf logging in my next issue, #6268.

You're right that this change did not include a mechanism for client code to include its own logger. I'm definitely interested in discussing how we should do that.

peterallenwebb:
@gshank I added a commit to this PR last night because some adapter test checks were failing. The issue was that pytest redefines sys.stdout stream so it can capture test output, but it closes the stream between tests. So if dbt functions are called during a test run, but the logging setup from the previous run is still in place, events may fire and write to a closed stream.

I added code to close the streams at the end of a run, and put in an additional fixture to reset logging at the beginning of each test. If you think there is a better way to deal with the issue, let me know.

core/dbt/events/functions.py
peterallenwebb: No here, but by adding a function to EventManager.callbacks. As you know, we now have a conversation started with @racheldaniel to get the details sorted out around how that callback registration will work.

core/dbt/flags.py
jtcohen6: We did document this: https://docs.getdbt.com/reference/events-logging#python-interface

But I don't think anyone was really using this, so let's remove it entirely for now. We can add it back with a deprecation cycle if we encounter significant pushback/outcry.

core/dbt/events/functions.py
jtcohen6: Very happy to see this thread — let me know if/how I can be helpful in that conversation w/ Rachel et al!

gshank:
I've always wanted to be able to get text logging to stdout and json logging to a file. That would require being able to configure the loggers differently. @jtcohen6 Is that something we'd have appetite for or not? Particularly once we have protobuf logging set up, I do not think *anybody* will want to have protobuf logs sent to stdout.

I took a quick look at what you're doing for capturing test output and it does have a bit of a hacky feeling to it. It's possible that there might not be a better way to do it, but I do wonder if we couldn't configure the test logger to do things differently once we have a clear idea of how the custom loggers will work.

peterallenwebb:
I started discussing this with @ChenyuLInx and @MichelleArk yesterday. The idea taking shape, as I understand it, is that each dbt command would be executed with a new function replacing handle()/handle_with_args(). That function would take a more general context object as a parameter, and that context object would likely be a Click context, which also has the capacity to include things like a callback function, or a configured Event Manager.

The hack (it definitely is a hack) to deal with stdout capture could be avoided if the calling code had the ability to control the execution context in this way, but the Click code still coming into focus, and I don't want to delay merging this.

That said, everyone seems to be thinking in this direction, so I'll provide add an additional commit to this PR with a proposed method of configuring the EventManager more elegantly.

jtcohen6:
> I've always wanted to be able to get text logging to stdout and json logging to a file. That would require being able to configure the loggers differently. @jtcohen6 Is that something we'd have appetite for or not? Particularly once we have protobuf logging set up, I do not think anybody will want to have protobuf logs sent to stdout.

I buy this. Would the challenge here be around finding the right UX for managing that more complex / granular configuration? I'm open to jamming on some flag/yaml specs.

gshank:
Yeah, we'd have to have flags to set them separately, I guess, and it might get a big muddled with trying to preserve compatibility with the old versions... New flags, maybe?

jtcohen6:
@gshank Maybe something like:
```sh
$ dbt --log-format json run                                # sets default for all, backwards-compatible
$ dbt --file-log-format json --stdout-log-format text run  # json file, text stdout
$ dbt --log-format json --stdout-log-format text run       # same: json file (default), text stdout
```

gshank:
That seems reasonable. Though I did start to wonder if we'd want to enable custom log handlers via config

jtcohen6:
> enable custom log handlers via config

Hm. At that point, it's probably more appropriate for someone to start invoking dbt-core via its Python API instead (i.e. the callback mechanism added in this PR). Good news is, we're hard at work (re)building a sane, defensible, documented Python API for dbt-core :)

peterallenwebb:
After today's discussions, I think we have concluded that further work on logging integration points will be deferred to future issues, and some of that may be done in the Click feature branch.

@gshank or @emmyoop Do you have any remaining concerns about merging this work?

I will open an issue for adding official API support for callback registration.

gshank (changes requested):
I'm a little unclear about why you chose to copy attributes from the LoggerConfig objects into the _Logger subclasses individually. Wouldn't it be simpler to just set LoggerConfig as an attribute or a superclass or something? It seems like some of the code in the _create_logger function could be handled by the individual logger classes.  I'm guessing that it's in the function in an attempt to make creating additional loggers simpler, but since we're not going to be doing that for a while, this feels harder to read.  Perhaps a "from_config" constructor in the _JsonLogger and _TextLogger classes?

In addition, you're not actually setting the "use_colors" in the json logger, and for compatibility sake with the current cloud use of the logger, I think we need to have that set until they've changed that.

In addition, I've made some changes in my current branch to the "event_to_dict" function, so I think we'd be better off leaving that as a function call for now, otherwise merging is going to get weird. It's also used in the tests...

peterallenwebb:
> I'm a little unclear about why you chose to copy attributes from the LoggerConfig objects into the _Logger subclasses individually. Wouldn't it be simpler to just set LoggerConfig as an attribute or a superclass or something? It seems like some of the code in the _create_logger function could be handled by the individual logger classes. I'm guessing that it's in the function in an attempt to make creating additional loggers simpler, but since we're not going to be doing that for a while, this feels harder to read. Perhaps a "from_config" constructor in the _JsonLogger and _TextLogger classes?

@gshank You are correct. I meant for this to be a factory function which simplified log creation, but I'll try the more direct approach you suggest.

> In addition, you're not actually setting the "use_colors" in the json logger, and for compatibility sake with the current cloud use of the logger, I think we need to have that set until they've changed that.

This part is intentional. The current code never uses this setting when JSON logging is being used, unless there is something I'm missing, and I think I would have caught a change of behavior in my testing.

> In addition, I've made some changes in my current branch to the "event_to_dict" function, so I think we'd be better off leaving that as a function call for now, otherwise merging is going to get weird. It's also used in the tests...

Yeah, I saw your change this morning. My intention was to avoid circular imports, but I'll find another way around that.

gshank:
Yeah, you're right about not using colors in the new json logging. It was in the legacy logger with json logging...  I do think that it's possible that cloud might want colors in the new json logging, but I'm okay with not doing it now. If it becomes an issue we can put it in then.

core/dbt/events/proto_types.py
```diff
@@ -2159,7 +2159,6 @@ class TrackingInitializeFailure(betterproto.Message):
     exc_info: str = betterproto.string_field(2)
 
 
-@dataclass
 class EventBufferFull(betterproto.Message):
     """Z045"""
```
gshank, line 2163: This is a generated file. You need to remove the EventBufferFulll message from types.proto and run the protoc compiler (as described in the events README.md).

gshank (approved):
You need to fix the EventBufferFull thing, but otherwise it looks good!

Approving so we don't have to go around again.

core/dbt/events/proto_types.py
peterallenwebb: Thanks! Will take care of this!

PR #6293, opened 2022-11-18
CT 1537 fix event test and rename a couple of fields

### LINKED ISSUE #6292: [CT-1537] Fix logging event test and rename a few fields
While working on the pull request for #6256, issues were fixed in the logging event tests and a few issues with event names were fixed. This is a ticket to process that work.

gshank:
resolves #6292

### Description

The event unit test was broken and some of the tests weren't actually doing anything. This fixes that, removes the lists of events from the bottom of the types.py and test_types.py files and moves them into test_events.py test file. It also renames a few tests.

### Checklist


core/dbt/context/base.py
```diff
@@ -16,7 +16,7 @@
     disallow_secret_env_var,
 )
 from dbt.events.functions import fire_event, get_invocation_id
-from dbt.events.types import MacroEventInfo, MacroEventDebug
+from dbt.events.types import JinjaLogInfo, JinjaLogDebug
```
emmyoop, line 19: Along these same lines there's also a `GeneralMacroWarning` that is the catchall for `warn` defined in `exceptions.py`.  It's worth renaming that along with these.

core/dbt/events/types.proto
```diff
@@ -15,6 +15,7 @@ message EventInfo {
     string thread = 7;
     google.protobuf.Timestamp ts = 8;
     map<string, string> extra = 9;
+    string category = 10;
```
emmyoop, line 18: Is this eventually going to be the log category?  Am I missing how this gets populated?

emmyoop (changes requested):
One more rename and one question.

core/dbt/events/types.proto
gshank: Yeah, right now it's just a placeholder. Eventually we'll populate it somehow...

core/dbt/context/base.py
gshank: It doesn't look like it's actually used anywhere. Can we just remove it?

core/dbt/events/types.proto
emmyoop: Is there an issue so we don't lose the fact we still need to actually populate it?

core/dbt/context/base.py
emmyoop: It's used within macros ([docs](https://docs.getdbt.com/reference/dbt-jinja-functions/exceptions#warn))

gshank: So it's exported into the context. So should we call in JinjaLogWarning?

gshank: Or JinjaLogWarn?

gshank: I went with JinjaLogWarning.

core/dbt/events/types.proto
gshank: Mostly it's your ticket #5958.

PR #6303, opened 2022-11-22
profile work in click

ChenyuLInx:
resolves #

### Checklist

### DIFF
--- a/core/dbt/config/profile.py
+++ b/core/dbt/config/profile.py
@@ -399,7 +399,7 @@ class Profile(HasCredentials):
         )
 
     @classmethod
-    def render_from_args(
+    def render(
         cls,
         renderer: ProfileRenderer,
         project_profile_name: Optional[str],
--- a/core/dbt/config/profile.py
+++ b/core/dbt/config/profile.py
@@ -401,9 +401,11 @@ class Profile(HasCredentials):
     @classmethod
     def render_from_args(
         cls,
-        args: Any,
         renderer: ProfileRenderer,
         project_profile_name: Optional[str],
+        threads_override: Optional[int] = None,
+        target_override: Optional[str] = None,
+        profile_name_override: Optional[str] = None,
     ) -> "Profile":
         """Given the raw profiles as read from disk and the name of the desired
         profile if specified, return the profile component of the runtime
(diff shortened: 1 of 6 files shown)


core/dbt/config/runtime.py
```diff
@@ -48,6 +48,25 @@ def load_project(
     return project
 
 
+def load_profile(
```
ChenyuLInx, line 51: @MichelleArk  I imagine we would use something like this to generate the profile object, we maybe should refactor this code to pass in a loaded project dict directly instead of reading from disk, that way we can reuse that loaded project to also construct project object

core/dbt/config/runtime.py
```diff
@@ -196,39 +215,18 @@ def validate(self):
...
-    @classmethod
-    def get_profile(
-        cls: Type["RuntimeConfig"], args: Any, cli_vars: Dict[str, Any], raw_profile_name: str
-    ) -> Profile:
-        # build the profile using the base renderer and the one fact we know
-        # Note: only the named profile section is rendered. The rest of the
-        # profile is ignored.
-        profile_renderer = ProfileRenderer(cli_vars)
-        profile_name = profile_renderer.render_value(raw_profile_name)
-        profile = cls._get_rendered_profile(args, profile_renderer, profile_name)
```
MichelleArk, line 218: This method is overwritten in UnsetProfileConfig to return an `UnsetProfile` for tasks that don't require a profile. Is there still a place for something like `UnsetProfile` in the new high-level API to avoid loading the profile for tasks that don't require it?

ChenyuLInx: Rewired things to make sure existing test case passes

core/dbt/config/profile.py
```diff
@@ -401,9 +401,11 @@ def from_raw_profiles(
     @classmethod
     def render_from_args(
```
MichelleArk, line 402: should probably consider renaming this method now that it doesn't even take `args`. Maybe just `render`?

MichelleArk:
Looks like the failing tests are all of the flavor:
```
render_from_args() got an unexpected keyword argument 'args'
```
and just need some refactoring to conform to the new signature for `render_from_args`. 

Otherwise looks good!

core/dbt/config/profile.py
ChenyuLInx: Updated

MichelleArk (approved):

ChenyuLInx:
@iknox-fa @MichelleArk  gonna close this in favor of https://github.com/dbt-labs/dbt-core/pull/6336

PR #6305, opened 2022-11-22
Convert test 067. 

VersusFacit:
### Description

Test conversion ready to go. Added some comments and used some new tricks to make sure everything places nicely.

In place of modifying unique_schema, I used the `+schema` tag to prevent the test storage schema name from becoming its expect 65 characters (two over limit 63).

### Checklist


VersusFacit:
I proposed an exception for the behavior that stumped me here since the failures it produces mask what's really going on. Now, we'll have an error like this:
image

tests/functional/store_test_failures_tests/test_store_test_failures.py
```diff
@@ -0,0 +1,152 @@
...
+    tests__passing_test,
+)
+
+# used to rename test audit schema to help test schema meet max char limit
+# the default is _dbt_test__audit but this runs over the postgres 63 schema name char limit
+# without which idempotency conditions will not hold (i.e. dbt can't drop the schema properly)
+TEST_AUDIT_SCHEMA_SUFFIX = "dbt_test__aud"
+
+
+class TestStoreTestFailures:
```
McKnight-42, line 28: nit: should this be a Base class instead?

McKnight-42 (approved):
other than my one thought about class naming all looks good here

tests/functional/store_test_failures_tests/test_store_test_failures.py
colin-rogers-dbt: I assume this is fine because we aren't expecting adapters to implement this test. @VersusFacit can you confirm?

VersusFacit: Great call, Matt. Doing this before hitting merge.

VersusFacit: As for your comment, Colin, that's right, but it was an oversight when renaming the original base class. All it would do is be silently incorrect since pytest would look for a test method and find nothing.

Issue #6099, opened 2022-10-18
[CT-1369] Human-written docs for top level API via Sphinx


iknox-fa:
In #5528 we added automatic documentation for the top level API.  This ticket encompasses the work to expand that with a human-written section with examples/how to sections as needed to flesh out the documentation for said API.

AC: The documentation is fully ready for internal consumption but this doesn't cover styling beyond the basic markup needed for readability.

iknox-fa:
Updated from BLG:  This work is intended to be stand alone from the extension created in #5528.  Implement as block at top of automated docs in order to KISS.

PR #6307, opened 2022-11-23
Example click API usage

MichelleArk:
resolves https://github.com/dbt-labs/dbt-core/issues/6310
resolves https://github.com/dbt-labs/dbt-core/issues/6099

### Description
This PR introduces a script that will serve as a base to be iterated on as we work through API-ification so that it serves as a live spec of the API-ified effort. Currently implemented with `deps` but ideally swapped out with `run` as soon as that is available in the click API. 

### Open Question(s)
* Currently dbt is callable programmatically by constructing a click `Context` and invoking the relevant command. Should we provide a thin abstraction over this so callers don't need to be concerned with the details + extra dependency management of click?

### Checklist

### DIFF
--- a/core/dbt/cli/example.py
+++ b/core/dbt/cli/example.py
@@ -1,0 +1,25 @@
+import click
+import sys
+
+from dbt.cli.main import cli as dbt, deps
+
+
+def make_context(args):
+    try: 
+        ctx = dbt.make_context('dbt', args)
+    except click.exceptions.Exit:
+        exit()
+
+    ctx.invoked_subcommand = ctx.protected_args[0] if ctx.protected_args else None
+    return ctx
+
+# python core/dbt/cli/example.py
+# python core/dbt/cli/example.py --version
+# python core/dbt/cli/example.py deps --project-dir <project-dir-path>
+if __name__ == "__main__":
+    ctx = make_context(sys.argv[1:])
+    dbt.invoke(ctx)
+    
+    # Skips group-level context setting
+    # dbt.commands['deps'].invoke(ctx)
+    # deps.invoke(ctx)
--- a/core/dbt/cli/example.py
+++ b/core/dbt/cli/example.py
@@ -1,25 +1,48 @@
 import click
+import os
 import sys
+from typing import Optional
 
 from dbt.cli.main import cli as dbt, deps
+from dbt.tracking import track_run
+from dbt.adapters.factory import adapter_management
+from dbt.profiler import profiler
+from dbt.config.runtime import load_project
 
 
-def make_context(args):
-    try: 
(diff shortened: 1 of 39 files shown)


core/dbt/cli/example.py
```diff
@@ -0,0 +1,25 @@
...
+        exit()
+
+    ctx.invoked_subcommand = ctx.protected_args[0] if ctx.protected_args else None
+    return ctx
+
+# python core/dbt/cli/example.py
+# python core/dbt/cli/example.py --version
+# python core/dbt/cli/example.py deps --project-dir <project-dir-path>
+if __name__ == "__main__":
+    ctx = make_context(sys.argv[1:])
```
MichelleArk, line 20: @ChenyuLInx - Idea here being that `make_context` could be passed in objects to add to `ctx.obj`, bypassing/modifying what's configured by the CLI.

ChenyuLInx: Should the example be constructing args explicitly rather than trying to take sys.args? given that we are trying to show other teams how to use the interface programmatically?

MichelleArk: I think constructing context objects like a Project or Profile (or an EventHandler with registered callbacks once we have that!) would be useful, yes! `sys.args` is just for the CLI arguments.

core/dbt/docs/source/index.rst
```diff
@@ -1,4 +1,70 @@
...
+    ctx.with_resource(profiler(enable=True, outfile="output.profile"))
+    project_dir = os.path.expanduser(ctx.params.get('project_dir'))
+    ctx.obj["project"] = load_project(project_dir, True, None, None)
+
+    # invoke the command
+    dbt_cli.commands[command].invoke(ctx)
+
+
+For the full code example, you can refer to `example.py <https://github.com/dbt-labs/dbt-core/blob/feature/click-cli/core/dbt/cli/example.py>`_
+
```
stu-k, line 68: Could you add another section header here as well, to separate the example from the rest of the documentation?

ChenyuLInx:
@iknox-fa  this is updated with latest interface, and ready to review

iknox-fa (approved):
LGTM

core/dbt/cli/example.py
```diff
@@ -0,0 +1,21 @@
+from dbt.cli.main import dbtRunner
+from dbt.config.runtime import load_profile, load_project
+
+if __name__ == "__main__":
+    project_dir = "/Users/chenyuli/git/jaffle_shop"
+    # Bypass cli group context configuration entirely and invoke deps directly
```
MichelleArk, line 6: don't think this comment makes sense here anymore - will remove it.

PR #6314, opened 2022-11-23
reformatting of test after some spike investigation

McKnight-42:
resolves #

### Description
Reformatting Base class test locations to test classes as part of fix for bigquery artifact spike

### Checklist


tests/adapter/dbt/tests/adapter/aliases/test_aliases.py
```diff
@@ -138,24 +123,33 @@ def models(self):
...
-    def test_alias_model_name_diff_database(self, project):
-        results = run_dbt(["run"])
-        assert len(results) == 2
-        res = run_dbt(["test"])
-        assert len(res) > 0
-
 
 class TestAliases(BaseAliases):
-    pass
+    def test_alias_model_name(self, project):
```
colin-rogers-dbt, line 128: feel like I'm missing something here, shouldn't the base class define the test and the implementation class handle adapter specific configuration?

McKnight-42: I made this change here to have the tests be designed in test classes because adapters like https://github.com/dbt-labs/dbt-bigquery/pull/382/files#diff-ca3edbc2e0b6572e21f4fcf04747fb68718878bdf9eee5d93efb079c2a08cf19R93-R104 that need to have the cleanup fixture for these odd cases in which we change pointer either in database or schema has to be passed in as a argument to the test

McKnight-42: this way we don't have a duplicate test method being called on import that doesn't cleanup properly

colin-rogers-dbt: Interesting, my understanding is pytest should only invoke a test method if the class has "Test" in the name but maybe it's because "test_aliases" is the module name?

McKnight-42: Exactly! it looks for Test in the class name and test in the method name so the default test_aliases in base class we inherit would still run as well and not have the clean_up fixture passing through it is correct from what i've seen locally anyway.

McKnight-42: okay reformated this in core a bit and playing around with it in bigquery currently as that is spike hitter.
Biguqery:
```
import pytest
import os
from dbt.tests.util import run_dbt
import dbt.tests.adapter.aliases.test_aliases as aliases

MACROS__BIGQUERY_CAST_SQL = """
{% macro bigquery__string_literal(s) %}
    cast('{{ s }}' as string)
{% endmacro %}
"""

MACROS__EXPECT_VALUE_SQL = """
-- cross-db compatible test, similar to accepted_values

{% test expect_value(model, field, value) %}

select *
from {{ model }}
where {{ field }} != '{{ value }}'

{% endtest %}
"""

MODELS_DUPE_CUSTOM_DATABASE_A = """
select {{ string_literal(this.name) }} as tablename
"""

MODELS_DUPE_CUSTOM_DATABASE_B = """
select {{ string_literal(this.name) }} as tablename
"""

MODELS_SCHEMA_YML = """
version: 2
models:
- name: model_a
  tests:
  - expect_value:
      field: tablename
      value: duped_alias
- name: model_b
  tests:
  - expect_value:
      field: tablename
      value: duped_alias
"""

class TestAliasesBigQuery(aliases.BaseAliases):
    def macros(self):
        return {
            "bigquery_cast.sql": MACROS__BIGQUERY_CAST_SQL,
            "expect_value.sql": MACROS__EXPECT_VALUE_SQL
        }


class TestSameTestSameAliasDifferentDatabasesBigQuery(aliases.BaseSameAliasDifferentDatabases):
    def project_config_update(self):
        return {
            "config-version": 2,
            "macro-paths": ["macros"],
            "models": {
                "test": {
                    "alias": "duped_alias",
                    "model_b": {"database": os.getenv("BIGQUERY_TEST_ALT_DATABASE")},
                },
            },
        }

    def macros(self):
        return {
            "bigquery_cast.sql": MACROS__BIGQUERY_CAST_SQL,
            "expect_value.sql": MACROS__EXPECT_VALUE_SQL
        }

    def models(self):
        return {
            "schema.yml": MODELS_SCHEMA_YML,
            "model_a.sql": MODELS_DUPE_CUSTOM_DATABASE_A,
            "model_b.sql": MODELS_DUPE_CUSTOM_DATABASE_B
        }

    @pytest.fixture(scope="function")
    def clean_up(self, project):
        yield
        with project.adapter.connection_named('__test'):
            relation = project.adapter.Relation.create(database=os.getenv("BIGQUERY_TEST_ALT_DATABASE"), schema=project.test_schema)
            project.adapter.drop_schema(relation)

    def test_alias_model_name_diff_database_bigquery(self, project, clean_up):
        results = run_dbt(["run"])
        assert len(results) == 2
        res = run_dbt(["test"])
        assert len(res) > 0

```

following this and installing local into adapter running tests gets me 

```
tests/functional/adapter/test_aliases.py::TestAliasesBigQuery::test_alias_model_name PASSED                                                                                                                                                                              [ 33%]
tests/functional/adapter/test_aliases.py::TestSameTestSameAliasDifferentDatabasesBigQuery::test_alias_model_name_diff_database PASSED                                                                                                                                    [ 66%]
tests/functional/adapter/test_aliases.py::TestSameTestSameAliasDifferentDatabasesBigQuery::test_alias_model_name_diff_database_bigquery PASSED
```
which shows us that the original test method is still being called on, also tried this just importing file and using actual `test_aliases.BaseSameAliasDifferentDatabases`

McKnight-42: continued playing around with it, reset some envs and got it working in the original way I had planned.

colin-rogers-dbt (approved):
Awesome work digging into the vagaries of pytest inheritance!

PR #6317, opened 2022-11-24
Fix dbt.config.get default values  (python-model)

### LINKED ISSUE #6309: [CT-1545] [Bug] dbt.config.get returns Undefined instead of the default value in Python model

### Current Behavior

`dbt.config.get("missing_key", "default_value")` default value dosen't work.  This behavior creates an Undefined variable inside the context after the compilation like this.

```python
config_dict = {'KEY_AVAILABLE': 'everthing_is_ok',  'MISSING_KEY': Undefined}
```

### Expected Behavior

```python
config_dict = {'KEY_AVAILABLE': 'everthing_is_ok',  'MISSING_KEY': "default_value"}
```

### Steps To Reproduce

Define a model like:

```python
def model(dbt, session):

    CONFIG_KEY = dbt.config.get("KEY_EXIST", "DEFAULT_VALUE")
    MISSING_KEY = dbt.config.get("MISSING_KEY", "DEFAULT_VALUE")

    df = dbt.ref("orders")

    return df
```
With configs like:

```yaml
version: 2
models:
  - name: something
    config:
      CONFIG_KEY:  everthing_is_ok
```
Run dbt compile. 

The compiled output will create a `config_dict` like this

```python
config_dict = {'CONFIG_KEY': 'everthing_is_ok',  'MISSING_KEY: Undefined}
```

In that case, Undefined is a undeclared variable in the context. Because that, this will create an exception. 

### Relevant log output

_No response_

### Environment

```markdown
- OS: Linux/Ubuntu
- Python:3.9.12
- dbt-core==1.3.0 
- dbt-databricks==1.3.0
```

spark

### Additional Context

_No response_

devmessias:
resolves #6309

### Description

After this, will be possible to use default values for dbt.config.get

### Checklist

### DIFF
--- a/core/dbt/parser/models.py
+++ b/core/dbt/parser/models.py
@@ -218,7 +218,7 @@ class ModelParser(SimpleSQLParser[ParsedModelNode]):
                     )
                 if num_args > 2:
                     raise ParsingException(
-                        f"dbt.config.get() takes at most  arguments ({num_args} given)",
+                        f"dbt.config.get() takes at most 2 arguments ({num_args} given)",
                         node=node,
                     )
                 key = args[0]
--- a/core/dbt/parser/models.py
+++ b/core/dbt/parser/models.py
@@ -210,8 +210,19 @@ class ModelParser(SimpleSQLParser[ParsedModelNode]):
         config_keys_defaults = []
         for (func, args, kwargs) in dbtParser.dbt_function_calls:
             if func == "get":
+                num_args = len(args)
+                if num_args == 0:
+                    raise ParsingException(
+                        "dbt.config.get() requires at least one argument",
+                        node=node,
+                    )
+                if num_args > 2:
+                    raise ParsingException(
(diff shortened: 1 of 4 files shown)


core/dbt/parser/models.py
```diff
@@ -207,15 +207,33 @@ def parse_python_model(self, node, config, context):
...
-                config_keys_used.append(args[0])
+                num_args = len(args)
+                if num_args == 0:
+                    raise ParsingException(
+                        "dbt.config.get() requires at least one argument",
+                        node=node,
+                    )
+                if num_args > 2:
+                    raise ParsingException(
+                        f"dbt.config.get() takes at most  arguments ({num_args} given)",
```
ChenyuLInx, line 221:
```suggestion
                        f"dbt.config.get() takes at most 2 arguments ({num_args} given)",
```

devmessias: thanks for this catch @ChenyuLInx

PR #6321, opened 2022-11-27
Add support for Python 3.11

joshuataylor:
resolves #6147

### Description
To support Python 3.11, various changes need to be made. Hopefully it won't be too much. I know that some BC Mashumaro changes will need to be made ([see here](https://github.com/Fatal1ty/mashumaro/releases/tag/v3.1)).

The other adapters will need to rely on this. I believe the Snowflake adapter (the only one I have experience with) might be a bit painful, due to Snowflake not planning on releasing wheels soon. Pyarrow 10.0.1 does support 3.11, so we'll see. :shrug: 

There are a few changes to be made with the dataclasses - see here: https://docs.python.org/3.11/whatsnew/3.11.html#dataclasses

### Checklist

- [x] I have read [the contributing guide](https://github.com/dbt-labs/dbt-core/blob/main/CONTRIBUTING.md) and understand what's expected of me
- [x] I have signed the [CLA](https://docs.getdbt.com/docs/contributor-license-agreements)
- [x] I have run this code in development and it appears to resolve the stated issue
- [ ] This PR includes tests, or tests are not required/relevant for this PR
- [ ] I have [opened an issue to add/update docs](https://github.com/dbt-labs/docs.getdbt.com/issues/new/choose), or docs changes are not required/relevant for this PR
- [ ] I have run `changie new` to [create a changelog entry](https://github.com/dbt-labs/dbt-core/blob/main/CONTRIBUTING.md#adding-a-changelog-entry)

PR #6326, opened 2022-11-28
Add support for Python 3.11

### LINKED ISSUE #6147: [CT-1411] [Feature] Support Python 3.11

### Describe the feature

Support Python 3.11.x in dbt Core and adapters:
- https://github.com/dbt-labs/dbt-snowflake/issues/331
- https://github.com/dbt-labs/dbt-bigquery/issues/398
- https://github.com/dbt-labs/dbt-redshift/issues/225
- https://github.com/dbt-labs/dbt-spark/issues/524

_No response_

### Who will this benefit?

Folks using the most recent release of Python

Will mark as a "good first issue" and let someone else take a shot at this

### Last time
It looks like this work was split across multiple PRs last time, but it would be nice to have a single PR this time:
1. https://github.com/dbt-labs/dbt-core/pull/4866/files
1. https://github.com/dbt-labs/dbt-core/pull/5025/files

### How to implement

Add 3.11 (and 3.10 when it is missing) to these locations:
1. https://github.com/dbt-labs/dbt-core/blob/19167842873ea882c2dbb70aeb525953812d8ef3/core/setup.py#L82-L84
1. https://github.com/dbt-labs/dbt-core/blob/19167842873ea882c2dbb70aeb525953812d8ef3/tests/adapter/setup.py#L53-L54
1. https://github.com/dbt-labs/dbt-core/blob/19167842873ea882c2dbb70aeb525953812d8ef3/plugins/postgres/setup.py#L84-L86
1. https://github.com/dbt-labs/dbt-core/blob/19167842873ea882c2dbb70aeb525953812d8ef3/tox.ini#L5
1. https://github.com/dbt-labs/dbt-core/blob/19167842873ea882c2dbb70aeb525953812d8ef3/tox.ini#L17
1. https://github.com/dbt-labs/dbt-core/blob/19167842873ea882c2dbb70aeb525953812d8ef3/.github/workflows/main.yml#L74
2. https://github.com/dbt-labs/dbt-core/blob/19167842873ea882c2dbb70aeb525953812d8ef3/.github/workflows/main.yml#L119
3. https://github.com/dbt-labs/dbt-core/blob/19167842873ea882c2dbb70aeb525953812d8ef3/Dockerfile.test#L46-L51
4. https://github.com/dbt-labs/dbt-core/blob/19167842873ea882c2dbb70aeb525953812d8ef3/CONTRIBUTING.md?plain=1#L59
5. https://github.com/dbt-labs/dbt-core/blob/19167842873ea882c2dbb70aeb525953812d8ef3/CONTRIBUTING.md?plain=1#L163
1. Any other relevant file I may have missed in the two pull requests from last time (when we added Python 3.10)

iknox-fa:
resolves #6147

### Description

Adds support for Python 3.11 :sonic:

### Checklist

### DIFF
--- a/core/dbt/events/eventmgr.py
+++ b/core/dbt/events/eventmgr.py
@@ -128,7 +128,7 @@ class _TextLogger(_Logger):
             log_line = f"\n\n{separator} {datetime.utcnow()} | {self.event_manager.invocation_id} {separator}\n"
         ts: str = datetime.utcnow().strftime("%H:%M:%S.%f")
         scrubbed_msg: str = self.scrubber(e.message())  # type: ignore
-        log_line += f"{self._get_color_tag()}{ts} [{e.log_level():<5}]{self._get_thread_name()} {scrubbed_msg}"
+        log_line += f"{self._get_color_tag()}{ts} [{e.log_level().value:<5}]{self._get_thread_name()} {scrubbed_msg}"
         return log_line
 
     def _get_color_tag(self) -> str:
--- a/core/dbt/events/eventmgr.py
+++ b/core/dbt/events/eventmgr.py
@@ -128,7 +128,11 @@ class _TextLogger(_Logger):
             log_line = f"\n\n{separator} {datetime.utcnow()} | {self.event_manager.invocation_id} {separator}\n"
         ts: str = datetime.utcnow().strftime("%H:%M:%S.%f")
         scrubbed_msg: str = self.scrubber(e.message())  # type: ignore
-        log_line += f"{self._get_color_tag()}{ts} [{e.log_level().value:<5}]{self._get_thread_name()} {scrubbed_msg}"
+        # log_level() for DynamicLevel events returns str instead of EventLevel
+        level = e.log_level().value if isinstance(e.log_level(), EventLevel) else e.log_level()
+        log_line += (
+            f"{self._get_color_tag()}{ts} [{level:<5}]{self._get_thread_name()} {scrubbed_msg}"
+        )
         return log_line
 
     def _get_color_tag(self) -> str:
--- a/core/dbt/helper_types.py
+++ b/core/dbt/helper_types.py
@@ -3,7 +3,7 @@
 # necessary for annotating constructors
 from __future__ import annotations
 
-from dataclasses import dataclass
+from dataclasses import dataclass, field
 from datetime import timedelta
 from pathlib import Path
 from typing import Tuple, AbstractSet, Union
@@ -80,12 +80,15 @@ class NVEnum(StrEnum):
     def __eq__(self, other):
         return isinstance(other, NVEnum)
 
+    def __call__(self):
+        return self.novalue
+
 
 @dataclass
 class NoValue(dbtClassMixin):
     """Sometimes, you want a way to say none that isn't None"""
 
-    novalue: NVEnum = NVEnum.novalue
+    novalue: NVEnum = field(default_factory=NVEnum.novalue)
 
 
 dbtClassMixin.register_field_encoders(
--- a/core/dbt/helper_types.py
+++ b/core/dbt/helper_types.py
@@ -80,15 +80,12 @@ class NVEnum(StrEnum):
     def __eq__(self, other):
         return isinstance(other, NVEnum)
 
-    def __call__(self):
-        return self.novalue
-
 
 @dataclass
 class NoValue(dbtClassMixin):
     """Sometimes, you want a way to say none that isn't None"""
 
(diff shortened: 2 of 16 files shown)


core/dbt/helper_types.py
```diff
@@ -80,12 +80,15 @@ class NVEnum(StrEnum):
...
+    def __call__(self):
+        return self.novalue
+
 
 @dataclass
 class NoValue(dbtClassMixin):
     """Sometimes, you want a way to say none that isn't None"""
 
-    novalue: NVEnum = NVEnum.novalue
+    novalue: NVEnum = field(default_factory=NVEnum.novalue)
```
jtcohen6, line 91: The approach in #6321 is to instead do:
```suggestion
    novalue: NVEnum = field(default_factory=lambda: NVEnum.novalue)
```
I think that avoids the need to define the `__call__` method above

jtcohen6 (commented):
More tests passing locally. `mypy` is unhappy with the hack I've taken here, where I "call" our `Policy` objects as a way of initializing all their dataclass properties (now that they're `default_factory` instead of `default`). I'm sure there's a better way to do this!

jtcohen6:
Per failing unit tests, looks like this needs fixing for py311:
https://github.com/dbt-labs/dbt-core/blob/b5d303f12a71ba17ac0e4707419a6439a032c6c4/tests/unit/test_version.py#L638-L648

MichelleArk:
> Per failing unit tests, looks like this needs fixing for py311:

Took a look into this - I believe the root cause here is that `mock.patch` now uses `importlib.import_module` indirectly python 3.11 and thus uses the mocked object (in this case, `mock_import`) instead of the real implementation in subsequent `mock.patch` calls, raising an `ImportError`.

The change in `unittest` is [here](https://github.com/python/cpython/blame/3.11/Lib/unittest/mock.py#L1608). A [related-looking issue was opened in cpython](https://github.com/python/cpython/issues/98771), but closed as not planned. One (not great, fragile) way to work around this is to [swap the order of execution in test_versions.py](https://github.com/dbt-labs/dbt-core/blob/main/tests/unit/test_version.py#L676-L677), such that the `importlib.import_module` patch is the last to be mocked.

jtcohen6:
> (not great, fragile)

@MichelleArk Given that this is just a method in our unit tests (quite old, probably warranting a rewrite), I don't hate the idea of a quick & expedient fix, whose fragility we can document for our future selves

core/dbt/helper_types.py
MichelleArk: I think the approach in #6321 is preferable for a couple reasons:
1. It's more consistent with how other `field.default_factory` values are set throughout the codebase, and 
2. It's more locally contained and easier to reason about - no need to create/rely on a `__call__` method on another class to get `NoValue` working

MichelleArk (approved):
Just one [comment](https://github.com/dbt-labs/dbt-core/pull/6326#discussion_r1041655874) - otherwise LGTM! 🐍 🏃💨

core/dbt/helper_types.py
jtcohen6: Thought so!

jtcohen6 (commented):
I think I managed to fix the failure in `test_logging.py` for py311.

Less sure why `tests/adapter/grants` suddenly started failing, claiming that the env vars for test users weren't set:
```
Env var required but not provided: 'DBT_TEST_USER_X'
```

**Update:** This is due to a breaking change in `tox==4.0.0`, see https://github.com/dbt-labs/dbt-core/pull/6405

core/dbt/events/eventmgr.py
```diff
@@ -128,7 +128,7 @@ def create_debug_line(self, e: BaseEvent) -> str:
             log_line = f"\n\n{separator} {datetime.utcnow()} | {self.event_manager.invocation_id} {separator}\n"
         ts: str = datetime.utcnow().strftime("%H:%M:%S.%f")
         scrubbed_msg: str = self.scrubber(e.message())  # type: ignore
-        log_line += f"{self._get_color_tag()}{ts} [{e.log_level():<5}]{self._get_thread_name()} {scrubbed_msg}"
+        log_line += f"{self._get_color_tag()}{ts} [{e.log_level().value:<5}]{self._get_thread_name()} {scrubbed_msg}"
```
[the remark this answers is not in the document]
jtcohen6: I noticed that `e.log_level()` generally returns an `EventLevel` enum, but sometimes it just returns a `str` for `DynamicLevel` events (added in https://github.com/dbt-labs/dbt-core/pull/6174)

PR #6363, opened 2022-12-02
Reverting back to older ubuntu image

leahwicz:
### Description
GitHub changed the `ubuntu-latest` image from Ubuntu-20.04 to Ubuntu-22.04 which caused our CI to break. As a temporary fix to unblock CI, let's revert back to the older image and use that until we know why things broke

### Checklist

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -119,7 +119,7 @@ jobs:
       fail-fast: false
       matrix:
         python-version: ["3.7", "3.8", "3.9", "3.10"]
-        os: [ubuntu-latest]
+        os: [ubuntu-20.04]
         include:
           - python-version: 3.8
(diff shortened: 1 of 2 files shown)


.github/workflows/main.yml
```diff
@@ -119,7 +119,7 @@ jobs:
       fail-fast: false
       matrix:
         python-version: ["3.7", "3.8", "3.9", "3.10"]
-        os: [ubuntu-latest]
```
emmyoop, line 122: Is there a reason you only pinned this one?  We point to `ubuntu-latest` in multiple places in this file.

emmyoop (commented):
Is it because this is the only place we sudo?

emmyoop (approved):

PR #6366, opened 2022-12-02
Adds 'buildable' selection mode

### LINKED ISSUE #6365: [CT-1590] [Feature] Improving dbt selection options

### Describe the feature

Our organization's code typically has tests that compare models with their dependencies. (Ex: if Model A contains $1 million in transactions, and Model B transforms that data, we should still have $1 million in transactions).

image

When building only part of the graph the existing selection modes do not provide good options for running the tests. For example, let's say we're making a change to Model B and want to make sure that it is working correctly. We run `dbt build -s model_b`

The "eager" mode will run Model B, Test AB and Test CB. But Test CB is likely to fail because we haven't built Model C yet!
The "cautious" mode will run Model B and _no_ tests - so we're not sure if the change worked or not. 

Instead we'd like a selection mode that will run only Model B and Test AB. Test AB should pass because Model A was required to have been already been built to run Model B.

In fact - the existing cautious mode will not run Test AB _even if Model A is a source_! Sources don't need to be built at all, so that test should always run.

Instead of adding a new selection mode we might want to consider changing the behavior of the cautious mode. I'm hesitant to suggest changing existing behavior - but I believe the suggested new mode would be preferable to cautious mode, and having only two modes is easier for users to understand. At the least cautious mode should be updated to allow tests that reference sources.

### Who will this benefit?

This will benefit developers who are testing changes on a subset of the graph, rather than building the entire graph.

Yes - I have code changes working that add this new mode and am willing to make updates to it based on feedback.

### Anything else?

_No response_

AGPapa:
resolves #6365

### Description

This PR adds a new selection mode called 'buildable'. This selection mode runs tests if the test's parents are selected - or if the tests's parents are ancestors of what is selected. This change makes it easier to develop part of the graph and test changes on it. See the issue for a more detailed description.

I am able to make changes to this PR in response to feedback! As mentioned in the issue, I think that some or all of these changes could be made directly to the existing cautious mode, without making a new selection mode at all.

### Checklist

    - https://github.com/dbt-labs/docs.getdbt.com/issues/2568

### DIFF
--- a/core/dbt/graph/selector_spec.py
+++ b/core/dbt/graph/selector_spec.py
@@ -24,6 +24,7 @@ SELECTOR_METHOD_SEPARATOR = "."
 class IndirectSelection(StrEnum):
     Eager = "eager"
     Cautious = "cautious"
+    Buildable = "buildable"
 
 
 def _probably_path(value: str):
@@ -173,12 +174,14 @@ class BaseSelectionGroup(dbtClassMixin, Iterable[SelectionSpec], metaclass=ABCMe
     def __init__(
         self,
         components: Iterable[SelectionSpec],
+        indirect_selection: IndirectSelection = IndirectSelection.Eager,
         expect_exists: bool = False,
         raw: Any = None,
     ):
         self.components: List[SelectionSpec] = list(components)
         self.expect_exists = expect_exists
         self.raw = raw
+        self.indirect_selection = indirect_selection
 
     def __iter__(self) -> Iterator[SelectionSpec]:
         for component in self.components:
--- a/.changes/unreleased/Features-20230102-091335.yaml
+++ b/.changes/unreleased/Features-20230102-091335.yaml
@@ -1,0 +1,6 @@
+kind: Features
+body: Adds buildable selection mode
+time: 2023-01-02T09:13:35.663627-05:00
+custom:
+  Author: agpapa
+  Issue: "6365"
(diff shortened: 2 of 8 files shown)


core/dbt/graph/selector_spec.py
```diff
@@ -173,12 +174,14 @@ class BaseSelectionGroup(dbtClassMixin, Iterable[SelectionSpec], metaclass=ABCMe
     def __init__(
         self,
         components: Iterable[SelectionSpec],
+        indirect_selection: IndirectSelection = IndirectSelection.Eager,
```
AGPapa, line 177: In case this is confusing - in `selector.py` on line 138 we need to get the indirect selection method from a composite spec (a union, difference, or intersection). This change here, and the changes in cli.py, add the indirect selection method to these composite specs through the BaseSelectionGroup  so we can get them at that line in selector.py. Previously the indirect selection method was only on the SelectionCriteria object.

It's possible I misunderstood how something works here, but I couldn't find an easier way to get the indirect selection method at that point.

dbeatty10:
@AGPapa 
One thing you can do to get ready for review is to run `changie new` to enter the details for the changelog entry. Running `changie` will require installing the development dependencies in your local environment. The [changelog](https://github.com/dbt-labs/dbt-core/blob/9ecb6e50e4b1c7f1ac99143a5a2cce4e545a9ac3/CONTRIBUTING.md) describes the specific details for this step.

Excited about the work you've done on this! 🤩

core/dbt/graph/selector_spec.py
ChenyuLInx: Thanks for this note, really helpful!

ChenyuLInx (approved):
LGTM! Thanks @AGPapa  for adding this new selection method!

PR #6384, opened 2022-12-05
CT 1604 remove compiled classes

### LINKED ISSUE #6383: [CT-1604] Remove Compiled Node classes
As part of the structured logging updates, we are creating proto definitions for node output. Having separate "Parsed" and "Compiled" classes complicates this and also complicates the code.

Remove the core/dbt/contracts/graph/compiled.py file and references to it.

Remove "Parsed" from the Parsed node classes, so that they are ModelNode, SnapshotNode, etc.

gshank:
resolves #6383

### Description

Remove compiled classes in order to simplify the artifact generation and code.

### Checklist


gshank:
@cla-bot check

.changes/unreleased/Under the Hood-20221205-164948.yaml
```diff
@@ -0,0 +1,7 @@
+kind: Under the Hood
+body: Remove compiled node classes
```
jtcohen6, line 2: Feels more like "unify" or "consolidate" ParsedNode + CompiledNode classes

core/dbt/contracts/sql.py
```diff
@@ -32,7 +32,7 @@ class RemoteResult(VersionedSchema):
 class RemoteCompileResultMixin(RemoteResult):
     raw_code: str
     compiled_code: str
-    node: CompileResultNode
+    node: ResultNode
```
jtcohen6, line 35: This may have implications for `dbt-rpc` ([here](https://github.com/dbt-labs/dbt-rpc/blob/514691d7b2d6387e879df461918a0e78fe956c96/dbt_rpc/contracts/rpc.py#L261)). I don't _think_ this should have any direct implications for `dbt-server` / `lib.py`, but worth testing to be sure.

core/dbt/contracts/graph/nodes.py
```diff
@@ -99,6 +99,49 @@ def add_macro(self, value: str):
...
+    sql: str
+
+
+@dataclass
+class CompiledNode:
+    compiled: bool = False
+    compiled_code: Optional[str] = None
+    extra_ctes_injected: bool = False
+    extra_ctes: List[InjectedCTE] = field(default_factory=list)
+    relation_name: Optional[str] = None
```
jtcohen6, line 114: It feels weird and incorrect that `relation_name` is only available [during compilation](https://github.com/dbt-labs/dbt-core/blob/16f529e1d4e067bdbb6a659a622bead442f24b4e/core/dbt/compilation.py#L399), rather than after parsing. I think the reason had been, we need to know [the adapter's `quoting` config](https://github.com/dbt-labs/dbt-core/blob/16f529e1d4e067bdbb6a659a622bead442f24b4e/core/dbt/compilation.py#L202-L204) — but I'm pretty sure we'd be able to know that at parse time too, right?

jtcohen6 (commented):
Very satisfying to see this cleanup underway. Appreciate the thought given to back-compat for plugins (adapters, and `dbt-rpc` as well). Let's see what we can do — and if we can't, we'll very call out the required upgrade paths.

core/dbt/contracts/graph/nodes.py
gshank: I think so. I agree that it would be nice to have this available at parse time. I think "relation_name" was added by an external contributor, so I'm guessing it was at compile time because it was easier to find a place to put it.

gshank: We do set "relation_name" for SourceDefinition at parse time.

gshank:
I checked the adapter repos and the only imports from a changed file are for a tests/unit/utils.py file in methods which are not actually used anywhere, so it doesn't look like the CI tests should fail. Those repos should be updated to remove the dead code though, so I'll open tickets to do that after this is merged.

I don't see any imports in dbt-server from dbt.contracts.graph, so I think we're good there.

dbt-rpc will have to be updated. I'll open a ticket to do that after this is merged.

core/dbt/contracts/graph/nodes.py
gshank: I'll look at setting relation_name in the right place in the pull request to implement the nodes in protobuf messages.

jtcohen6: I think you're right! Sounds good.

peterallenwebb (approved):

PR #6385, opened 2022-12-05
Adding Semantic Structure: The First Draft

callum-mcdata:
### Description

This is a draft PR. It is not even close to being completed and this code isn't tested at all 😅 . That being said, we've got entities!

image


callum-mcdata:
### What This PR Adds:
- Add the entity node to dbt-core
- Add the name, model, description properties to the entity node
- Create EntityDimension
- Create the entity output in the manifest based upon the below mockup
- Allow entities to be disabled

Issue #6096, opened 2022-10-18
Test framework works with Click


iknox-fa:
### Is this your first time submitting a feature request?

- [X] I have read the [expectations for open source contributors](https://docs.getdbt.com/docs/contributing/oss-expectations)
- [X] I have searched the existing issues, and I could not find an existing issue for this feature
- [X] I am requesting a straightforward extension of existing dbt functionality, rather than a Big Idea better suited to a discussion

### Describe the feature

Our current test framework utilizes the legacy CLI code found in `core/dbt/main.py`.  This ticket is to refactor it such that is uses the new cli framework found at `core/dbt/cli/main.py`

Issue #6381, opened 2022-12-05
[CT-1602] [Feature] a thin layer for calling dbt top level commands programatically


ChenyuLInx:
### Describe the feature

In the current iteration of integrating dbt-core into dbt-server, we are doing something similar to [the example](https://github.com/dbt-labs/dbt-core/pull/6307/files) @MichelleArk added.

But this require use to setup the click context manual, for example([link](https://github.com/dbt-labs/dbt-server/blob/ac1f183ea5304b6fb40790441cee90c8038f0e69/dbt_server/views.py#L471-L472))

```
    project_dir_override = os.path.expanduser(ctx.params.get('project_dir'))
    ctx.obj["project"] = load_project(project_dir_override, True, None, None)  # type: ignore
    dbt.commands[command].invoke(ctx)
```

In this example, we are constructing a project object and put it in the context, so that the following command can use it.

It is not ideal that we are doing this kind of operation on the raw click context, dbt-core side should provide an abstraction layer on top to hide the click implementation underneath. Option 3 out of the discussion with @peterallenwebb  and @MichelleArk  is what we are currently thinking about.([link](https://github.com/dbt-labs/dbt-server/blob/ac1f183ea5304b6fb40790441cee90c8038f0e69/dbt_server/views.py#L490-L494))

```
    # option3
    # we need to customize making context
    # ctx.set_logger_for_dbt(new_logger)
    # ctx.set_profie()
    #     # maybe have a customize invoke function from click
```

Here we imagine the `ctx` being some dbt-core object that has certain functions that other program can use to customize setup.

This issue is for implementing such thin layer on top of the click framework core uses to model it's cli interface.

This would be the replacement(maybe part of the replacement) for current [lib.py](https://github.com/dbt-labs/dbt-core/blob/main/core/dbt/lib.py), we should also think about what are the things that we don't like about `lib.py` and avoid doing similar thing in the new thin layer. We would want this layer tested, and also avoid any hard code happening in it.

### Describe alternatives you've considered

Updating the click context directly

### Who will this benefit?

Users who wants to use dbt as a python package

jtcohen6:
First cut of this was included in https://github.com/dbt-labs/dbt-core/pull/6387

PR #6387, opened 2022-12-06
Functional test framework working with Click, dbtRunner

MichelleArk:
resolves https://github.com/dbt-labs/dbt-core/issues/6096, https://github.com/dbt-labs/dbt-core/issues/6381

### Description
- adds `dbtRunner` to run dbt CLI programmatically. Accepts profile, project, and manifest (for later), and uses them in combination with CLI args during `invoke` calls to invoke the CLI.  
- runs `tests/functional` tests through `dbtRunner`
  - many integration tests are failing because most commands are not yet implemented, but expected to start passing as we implement commands. 
  - did some manual inspection and found `tests/functional/dependencies/test_simple_dependency.py::TestSimpleDependencyUnpinned::test_simple_dependency`, which only depends on running `deps` - and succeeds! 
- runs `test/integration` tests last so that we can start surfacing failing functional tests in CI logs. 
- sets a couple param defaults to callables so they are set at runtime as opposed to param.py import. This is needed for correct fixture setup which set the project/profile dirs dynamically. 

### Checklist


core/dbt/cli/context.py
```diff
@@ -0,0 +1,44 @@
...
+
+class DBTUsageException(Exception):
+    pass
+
+class DBTContext(Context):
+    def __init__(self, args: List[str]) -> None:
+        try: 
+            ctx = cli.make_context(cli.name, args)
+            if args:
+                cmd_name, cmd, cmd_args = cli.resolve_command(ctx, args)
```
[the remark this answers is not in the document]
MichelleArk: yes!

core/dbt/tests/util.py
```diff
@@ -73,7 +73,11 @@ def run_dbt(args: List[str] = None, expect_pass=True):
         args = ["run"]
 
     print("\n\nInvoking dbt with {}".format(args))
-    res, success = handle_and_check(args)
+    # ctx = DBTContext.from_args(args)
+    ctx = DBTContext(args)
```
ChenyuLInx, line 77: This `ctx` object here is actually a context object for the actual command( for example `run`), and in the bottom we are calling the invoke of the top level command(`cli` group) with it. 

What actually happened during the process of calling a click framework from the cli is(lets use `run` as an example):
- invoke the `cli group cmd` with context built for that group(doesn't actually have anything related to run except that the `invoked_subcommand ` is set) [link](https://github.com/pallets/click/blob/main/src/click/core.py#L1653-L1654)
- invoke the run command with context that was built for run, with context that went through `cli group cmd` as parent. [link](https://github.com/pallets/click/blob/main/src/click/core.py#L1655-L1657)
- not sure what exactly happens if we have multi level group like `docs`

So this actually doesn't mimic what happens in cli, we probably need to figure out what is the best way to do this

MichelleArk: > This ctx object here is actually a context object for the actual command( for example run)

I believe the ctx object is for the top-level command, as constructed [here](https://github.com/dbt-labs/dbt-core/pull/6387/files#diff-6b14f723aa6f9a670e0994eea8dbddf5a973ef12c3c4a1e48461bad094c2de58R13). The subcommand context isn't used, just created for validation [here](https://github.com/dbt-labs/dbt-core/pull/6387/files#diff-6b14f723aa6f9a670e0994eea8dbddf5a973ef12c3c4a1e48461bad094c2de58R16).

ChenyuLInx: But maybe what you are doing here will mimic the behavior of calling in from https://github.com/pallets/click/blob/main/src/click/core.py#L1055 ?

core/dbt/cli/context.py
```diff
@@ -0,0 +1,39 @@
...
+import sys
+
+
+class DBTUsageException(Exception):
+    pass
+
+
+class DBTContext(Context):
+    def __init__(self, command: Command, **kwargs) -> None:
+        invocation_args = kwargs.pop("args", sys.argv[1:])
```
ChenyuLInx, line 14: I think we probably can't use sys.argv here? For anything invoking it programmatically

MichelleArk: agree, still need to fix this up 👍

core/dbt/cli/context.py
```diff
@@ -0,0 +1,39 @@
...
+        super().__init__(command, **kwargs)
+
+        # Bubble up validation errors for top-level commands
+        if not self.parent:
+            self._validate_args(command, invocation_args)
+
+        if not self.obj:
+            flags = Flags(self, args=invocation_args)
```
ChenyuLInx, line 22: How do you imagine this work with the current setup stuff we do in `cli` group?

MichelleArk: Removed it from the `cli` group method, instead the runner sets `cli.context_class = DBTContext` here: https://github.com/dbt-labs/dbt-core/pull/6387/files#diff-43832707bfec8cdaa49c13f004d9b3296585827f7e68ef07fba7ad1d5e6eb7c6R24

core/dbt/cli/context.py
```diff
@@ -0,0 +1,39 @@
...
+from dbt.cli.flags import Flags
+import sys
+
+
+class DBTUsageException(Exception):
+    pass
+
+
+class DBTContext(Context):
+    def __init__(self, command: Command, **kwargs) -> None:
```
ChenyuLInx, line 13: Trying to understand what different behavior will I get by passing in command `cli(group)` with all args vs `run` with all args

MichelleArk: So far this only really works for `cli`. The more I think it, the more it makes sense that only `the cli.context_class` should be `DBTContext` (as opposed to the subcommands) given all the top-level configuration and initialization it's doing.

core/dbt/cli/context.py
```diff
@@ -0,0 +1,39 @@
...
+            self.obj["profile"] = profile
+            self.obj["project"] = project
+
+    def _validate_args(self, command, args) -> None:
+        try:
+            command.parse_args(self, args)
+            if isinstance(command, Group):
+                _, cmd, cmd_args = command.resolve_command(self, args)
+                self._validate_args(cmd, cmd_args)
+        except (NoSuchOption, UsageError) as e:
```
MichelleArk, line 38: Note: catching the click exceptions and raising a custom exception (or any other type of handling that suppresses the original exceptions) in the context object breaks user-friendly stdout for the CLI, so it's probably not the right place to do this type of handling that assumes a programmatic invocation.

core/dbt/cli/context.py
```diff
@@ -0,0 +1,39 @@
...
+            self.obj = {}
+            self.obj["flags"] = flags
+            self.obj["profile"] = profile
+            self.obj["project"] = project
+
+    def _validate_args(self, command, args) -> None:
+        try:
+            command.parse_args(self, args)
+            if isinstance(command, Group):
+                _, cmd, cmd_args = command.resolve_command(self, args)
```
MichelleArk, line 36: We probably shouldn't rely on this method - it's not publicly documented and assumes the first command of `args` is the invoked subcommand, which we don't really have a way to guarantee without taking on more parsing responsibility here.

core/dbt/cli/main.py
```diff
@@ -25,6 +28,52 @@ def cli_runner():
...
+    # TODO: generalize safe access to threads
+    threads = getattr(flags, "THREADS", None)
+    profile = load_profile(flags.PROJECT_DIR, flags.VARS, flags.PROFILE, flags.TARGET, threads)
+    # Project
+    if ctx.obj.get("project") is None:
+        project = load_project(flags.PROJECT_DIR, flags.VERSION_CHECK, profile, flags.VARS)
+        ctx.obj["project"] = project
+
+    # Context for downstream commands
+    ctx.obj["flags"] = flags
```
ChenyuLInx, line 73: Not saying we need to change it but since we actually construct flags, profile, project within the command, is `ctx.obj` still the best place to put it?

core/dbt/cli/main.py
```diff
@@ -25,6 +28,52 @@ def cli_runner():
...
+def _initialize_context(ctx):
+    flags = Flags()
+
+    # Tracking
+    initialize_from_flags(flags.ANONYMOUS_USAGE_STATS, flags.PROFILES_DIR)
+    ctx.with_resource(track_run(run_command=flags.WHICH))
+
+    ctx.obj = ctx.obj or {}
+
+    # Profile
```
ChenyuLInx, line 63: I think we talked about having some sort of mapping defining which command would need what object(Deps doesn't need profile for example). Should we define those in the actual command and use it here so that we don't always load all objects?

Also thinking how can we expose that to server so that they don't need to fetch things that is not needed during initialize a task.

I think all I am talking about here can be a follow up thing after the current change.

core/dbt/cli/main.py
```diff
@@ -25,6 +28,52 @@ def cli_runner():
     cli()
 
 
+class dbtUsageException(Exception):
+    pass
+
+
+# Programmatic invocation
+class dbtRunner:
```
jtcohen6, line 36: Is this first rough version of thin wrapper, for end users who want to programmatically invoke `dbt-core` via Python API? Thinking:
```python
    # initialize longer-lived dbtRunner
    # by implicitly loading project/profile/etc based on config + cwd
    dbt = dbtRunner()
    
    # submit commands to dbtRunner
    # each command returns
    #   results: List[Result]
    #   success: bool
    results, success = dbt.run(...)
    results, success = dbt.test(...)
```

core/dbt/cli/main.py
MichelleArk: This actually resolved itself quite nicely via [`requires.profile`, `requires.project`](https://github.com/dbt-labs/dbt-core/pull/6387/files#diff-43832707bfec8cdaa49c13f004d9b3296585827f7e68ef07fba7ad1d5e6eb7c6R234-R236) decorators once we pivoted to initializing flags / context objects at the subcommand level.

core/dbt/cli/main.py
ChenyuLInx:
```python
    results, success = dbt.run(...)
    results, success = dbt.test(...)
```
this part is tbd I think? current way would be pass in command through args, the main advantage is some CLI args need to be specified before the actual command. So exposing things this way provide best flexibility at the moment.

```python
args = ['run', '--project-dir', 'something']
res, success = dbt.invoke(args)
```

core/dbt/cli/requires.py
```diff
@@ -0,0 +1,87 @@
...
+def project(func):
+    def wrapper(*args, **kwargs):
+        ctx = args[0]
+        assert isinstance(ctx, Context)
+
+        if ctx.obj.get("project") is None:
+            # TODO: Decouple target from profile, and remove the need for profile here:
+            # https://github.com/dbt-labs/dbt-core/issues/6257
+            if not ctx.obj.get("profile"):
+                raise DbtProjectError("profile required for project")
```
MichelleArk, line 77: Is there a more appropriate exception to raise here? I'm not too concerned either way because we should be able to get away from requiring a profile at which point the exception won't be necessary here.

core/dbt/cli/main.py
jtcohen6: > current way would be pass in command through args

We'll definitely want to support this generic pattern ("given an arbitrary list of CLI arg strings, invoke"). I could also see us wanting to support a more structured approach, where users can build up a dictionary of attributes and pass it into a named command. All of which is to say, we should design this programmatic interface for real, before we commit to it as a public-facing API — and that means sitting down, writing some user stories, talking with community members. All good stuff to do next year :)

> the main advantage is some CLI args need to be specified before the actual command

Lukewarm take: This is pretty confusing for end users! We should really seek to support flags in any order (and resolve any resulting ambiguities by actually renaming flags). I know we implemented the new CLI modeling in accordance with the old CLI model — parity is a good goal to shoot for with any refactor — but it would be _great_ to just support both of these, seamlessly and equivalently:
```bash
$ dbt --no-use-color run
$ dbt run --no-use-color
```

core/dbt/cli/main.py
MichelleArk: Hmmm, definitely worth questioning! An alternative could be to inject profile, project, manifest, etc into the command method's `arg`s, similar to how `click.pass_context` works. This way we'd be a little less coupled to click, although we'd probably still need to use the click context to store any project/profile overrides passed to `dbtRunner`. There would be some naming conflicts to resolve (ex: params from options named `--profile` and a `profile` override).

I think it's worth keeping in mind alternatives to using `ctx.obj` for passing user-configured objects to commands in mind as we go as the current approach doesn't preclude us from changing this up in the future.

core/dbt/cli/requires.py
ChenyuLInx: Can we also check for `requires.profile` is added to the current command or not?

core/dbt/cli/main.py
ChenyuLInx: cc @iknox-fa

core/dbt/cli/requires.py
MichelleArk: I think that may be overly-rigid. Technically this method only requires profile to exist in `ctx.obj`, and shouldn't need to be concerned with _how_ it gets there. Adding that check would make it unnecessarily difficult to change the 'how'. Also not sure it's worth the effort given we're looking to decouple profile from project creation.

stu-k (approved):
Looks good to me. There is one `test.py` file that seems to be empty that you can probably remove.

ChenyuLInx (approved):
Looks good to me, also tried to extend this in another PR, works great!

core/dbt/cli/main.py
iknox-fa: Well, that wasn't part of the original plan, but I guess we can see about supporting arbitrary flag ordering--  It's not something we should take on in this PR though.  @jtcohen6, can you write up a ticket that we can put into the estimation process.

core/dbt/cli/requires.py
iknox-fa: > Is there a more appropriate exception to raise here
This seems appropriate to me.  I do know that exceptions are getting some closer attention from @emmyoop and others soon so I wouldn't stress it.

tox.ini
```diff
@@ -25,9 +25,10 @@ passenv =
   POSTGRES_TEST_*
   PYTEST_ADDOPTS
 commands =
-  {envpython} -m pytest --cov=core -m profile_postgres {posargs} test/integration
   {envpython} -m pytest --cov=core {posargs} tests/functional
   {envpython} -m pytest --cov=core {posargs} tests/adapter
+  {envpython} -m pytest --cov=core -m profile_postgres {posargs} test/integration
```
iknox-fa, line 30: Nothing important, but I'm curious why the test run-order changed?

core/dbt/tests/util.py
```diff
@@ -73,7 +73,8 @@ def run_dbt(args: List[str] = None, expect_pass=True):
         args = ["run"]
 
     print("\n\nInvoking dbt with {}".format(args))
-    res, success = handle_and_check(args)
+    dbt = dbtRunner()
```
iknox-fa, line 76: I love how simple this change ended up being.  It's a sign that we made the right abstractions when we implemented the new functional testing framework.

iknox-fa (approved):
This looks great @MichelleArk!
Ship it!

core/dbt/cli/requires.py
emmyoop: There's also a `DbtProfileError` Exception that is also a subclass of `DbtConfigError.

tox.ini
MichelleArk: Given the test/integration tests will continue to fail on this branch, I moved them to run last so that errors from `tests/functional` would be visible in CI logs

core/dbt/cli/main.py
jtcohen6: @iknox-fa Totally - opened here: https://github.com/dbt-labs/dbt-core/issues/6497

PR #6402, opened 2022-12-07
Making Metric Timestamps Optional

### LINKED ISSUE #6398: [CT-1618] [Feature] Metric timestamp and time_grains should be optional

### Describe the feature

In today's world, we state that dbt metrics are inherently time bound - they must be associated with a timestamp and any query of them is aggregated at the grain provided. This was initially bound with #6379 but this is actually something we can pursue before beginning that work - they're not bound to one another.

**scratch** oh wait that's not entirely true 😅 

Based on feedback, we implemented a feature in the metrics package to support the 'all_time' grain where we aggregate to the base grain of the metric. So we already support a world where timestamp isn't really needed.

Additionally we have `derived` metrics which don't need timestamps because they combine the timestamps of other metrics.

So lets support this world we've created!

### The Proposal
- `timestamp` becomes an optional input
- `time_grains` becomes an optional input
- Rewrite dbt_metrics to make `grain` an optional input and the base behavior is all time. This will follow in another issue.

Not doing this but we already "support" it!

### Who will this benefit?

Anyone who uses metrics but doesn't want to aggregate them by time!

You betcha!

### Anything else?

_No response_

callum-mcdata:
resolves #6398 

### Description
This makes the timestamp property optional. This is in order to support non time-grain behavior in dbt_metrics. PR here: https://github.com/dbt-labs/dbt_metrics/pull/204

Concrete changes:
- Makes timestamp optional
- Raises a ValidationError if time_grains are defined but are missing timestamp
- Raises a ValidationError if window is defined but is missing timestamp
- Adds tests to ensure that all three above pieces of functionality are working as expected

### Checklist


jtcohen6:
@callum-mcdata I think this requires an update to docs (for v1.4+), since we currently say that `timestamp` is required in [this table](https://docs.getdbt.com/docs/build/metrics#available-properties). Does that sound right to you?

callum-mcdata:
@jtcohen6 yep! I've got it on my to-do - we're going to mirror the readme in dbt_metrics over to the docs page so people don't need to reference two locations.

PR #6409, opened 2022-12-08
Ct 1517/convert 060 persist docs

### LINKED ISSUE #6273: [CT-1517] 060_persist_docs_tests

VersusFacit:
resolves #6273 

### Description

Convert the test. There was a seed in here unused for whatever reason. I added it into the code env and bumped the assertion rhs in a line to complete the re-add.

### Checklist


mikealfare:
If I understand the intent correctly, we're moving all of the files into fixtures. So instead of moving the *.sql/*.yml files from integration to functional, shouldn't we just delete them? From spot checking, it looks like we already put those in `fixtures.py`.

tests/functional/persist_docs_tests/test_persist_docs.py
```diff
@@ -1,17 +1,47 @@
...
+    _MODELS__VIEW,
+    _PROPERITES__SCHEMA_MISSING_COL,
+    _PROPERTIES__SCHEMA_YML,
+    _SEEDS__SEED,
+)
+
+
+class BasePersistDocsTest:
+    @pytest.fixture(scope="class", autouse=True)
+    def setUp(self, project):
```
mikealfare, line 23: I don't think this is a fixture, is it? If you want this to run once per class, changing it to this should work:
```
@classmethod
def setup_class(cls):
    ...
```
If you need it for each method, then this should work:
```
def setup_method(self, method):
    ...
```

tests/functional/persist_docs_tests/test_persist_docs.py
```diff
@@ -40,8 +70,12 @@ def _assert_has_table_comments(self, table_node):
             table_comment, table_id_comment, table_name_comment
         )
 
-    def _assert_has_view_comments(self, view_node, has_node_comments=True,
-                                  has_column_comments=True):
+    def _assert_has_view_comments(
```
mikealfare, line 73: This and `_assert_has_table_comments()` look like a good use case for parameterized tests (`@pytest.mark.paraterize(...)`. I think that's what's effectively happening here, but it could be maintained more easily with the pytest built-ins.

tests/functional/persist_docs_tests/test_persist_docs.py
VersusFacit: It's not a fixture in the project fixture definition, and I agree with  the setup_class,setup_method methods. However, this is the scheme the team has been using. I'm doing this for current sake of consistency more than anything.

tests/functional/persist_docs_tests/test_persist_docs.py
VersusFacit: I looked into this, and agree it's a neat built in. But I'm not sure it matches this use case. This flag appears to accomodate fixed values whereas we're passing in dynamically generated nodes. Am I missing something from your thinking here?

VersusFacit:
Yep, you're right on those old files. They were added with a sloppy git add. I like how the one time I use -a, the reason you shouldn't use -a happens.

colin-rogers-dbt (approved):
LGTM

PR #6412, opened 2022-12-09
feat: add full_refresh option to on_schema_change of incremental models

### LINKED ISSUE #4473: Support `on_schema_change: full_refresh` mode
_Sujay [in Slack](https://getdbt.slack.com/archives/C2JRTGLLS/p1638993832110500):_

_> Can we have a config for dbt full refresh on schema change for incremental modes?_

_I started writing a response in Slack, but figured it's better for visibility here!_

This was definitely a goal of our initial implementation (#3288), and it's worth revisiting in the future.

It's a bit trickier than the other modes, because (on most databases):
- We need to create the temp table of "new data" first, in order to infer its column schema
- Then, if `on_schema_change` is not `ignore` (default), we compare column schemas and (depending on mode) adjust the existing column schema

If we detected a change and wanted to trigger a full-refresh, we'd need to do either/both of:
- Throw away that initial temp table, and jump into a totally different logical branch of the materialization (full-refresh logic)
- Try to [infer the new column schema from the model SQL](https://github.com/dbt-labs/dbt-core/blob/3fc715f06673f24718c761e5d0cf173766954816/core/dbt/include/global_project/macros/adapters/columns.sql#L20-L34) (`where false limit 0`), rather than creating an actual temp table of new data. If the column schemas differ, proceed with full-refresh; otherwise, proceed as normal.

Both are possible, it'll just take some work to get there. We should use this opportunity to clean up and modularize some of the incremental materialization logic in the process (no breaking changes).

pol-defont-reaulx:
resolves #4473

### Description

I added a full_refresh option on on_schema_change option for incremental models. Now, with this option and when finding a schema change on a model it will do a full refresh on this model

### Checklist

### DIFF
--- a/core/dbt/include/global_project/macros/materializations/models/incremental/incremental.sql
+++ b/core/dbt/include/global_project/macros/materializations/models/incremental/incremental.sql
@@ -44,16 +44,22 @@
              to_relation=target_relation) %}
     {#-- Process schema changes. Returns dict of changes if successful. Use source columns for upserting/merging --#}
     {% set dest_columns = process_schema_changes(on_schema_change, temp_relation, existing_relation) %}
-    {% if not dest_columns %}
-      {% set dest_columns = adapter.get_columns_in_relation(existing_relation) %}
-    {% endif %}
 
-    {#-- Get the incremental_strategy, the macro to use for the strategy, and build the sql --#}
-    {% set incremental_strategy = config.get('incremental_strategy') or 'default' %}
-    {% set incremental_predicates = config.get('incremental_predicates', none) %}
-    {% set strategy_sql_macro_func = adapter.get_incremental_strategy_macro(context, incremental_strategy) %}
-    {% set strategy_arg_dict = ({'target_relation': target_relation, 'temp_relation': temp_relation, 'unique_key': unique_key, 'dest_columns': dest_columns, 'predicates': incremental_predicates }) %}
-    {% set build_sql = strategy_sql_macro_func(strategy_arg_dict) %}
+    {% if dest_columns and dest_columns.get("full_refresh") %}
+      {% set build_sql = get_create_table_as_sql(False, intermediate_relation, sql) %}
+      {% set need_swap = true %}
+    {% else %}
+        {% if not dest_columns %}
+          {% set dest_columns = adapter.get_columns_in_relation(existing_relation) %}
+        {% endif %}
+
+        {#-- Get the incremental_strategy, the macro to use for the strategy, and build the sql --#}
+        {% set incremental_strategy = config.get('incremental_strategy') or 'default' %}
+        {% set incremental_predicates = config.get('incremental_predicates', none) %}
+        {% set strategy_sql_macro_func = adapter.get_incremental_strategy_macro(context, incremental_strategy) %}
+        {% set strategy_arg_dict = ({'target_relation': target_relation, 'temp_relation': temp_relation, 'unique_key': unique_key, 'dest_columns': dest_columns, 'predicates': incremental_predicates }) %}
+        {% set build_sql = strategy_sql_macro_func(strategy_arg_dict) %}
+    {% endif %}
 
   {% endif %}
 
--- a/core/dbt/include/global_project/macros/materializations/models/incremental/on_schema_change.sql
+++ b/core/dbt/include/global_project/macros/materializations/models/incremental/on_schema_change.sql
@@ -1,6 +1,6 @@
 {% macro incremental_validate_on_schema_change(on_schema_change, default='ignore') %}
 
-   {% if on_schema_change not in ['sync_all_columns', 'append_new_columns', 'fail', 'ignore'] %}
+   {% if on_schema_change not in ['sync_all_columns', 'append_new_columns', 'full_refresh', 'fail', 'ignore'] %}
 
      {% set log_message = 'Invalid value for on_schema_change (%s) specified. Setting default value of %s.' % (on_schema_change, default) %}
      {% do log(log_message) %}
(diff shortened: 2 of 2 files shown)


core/dbt/include/global_project/macros/materializations/models/incremental/on_schema_change.sql
```diff
@@ -128,6 +128,9 @@
 
           {% do exceptions.raise_compiler_error(fail_msg) %}
 
+        {% elif on_schema_change == 'full_refresh' %}
+          {{ return({"full_refresh": True}) }}
```
jtcohen6, line 132: What do you think about some additional debug-level logging here? Something like:
```
      {% do log("Full refreshing " ~ target_relation ~ " on account of schema change") %}
```

core/dbt/include/global_project/macros/materializations/models/incremental/incremental.sql
```diff
@@ -44,16 +44,22 @@
...
-    {#-- Get the incremental_strategy, the macro to use for the strategy, and build the sql --#}
-    {% set incremental_strategy = config.get('incremental_strategy') or 'default' %}
-    {% set incremental_predicates = config.get('incremental_predicates', none) %}
-    {% set strategy_sql_macro_func = adapter.get_incremental_strategy_macro(context, incremental_strategy) %}
-    {% set strategy_arg_dict = ({'target_relation': target_relation, 'temp_relation': temp_relation, 'unique_key': unique_key, 'dest_columns': dest_columns, 'predicates': incremental_predicates }) %}
-    {% set build_sql = strategy_sql_macro_func(strategy_arg_dict) %}
+    {% if dest_columns is mapping and dest_columns.get("full_refresh") %}
+      {% set build_sql = get_create_table_as_sql(False, intermediate_relation, sql) %}
+      {% set need_swap = true %}
+    {% else %}
```
jtcohen6, line 51: We'll need logic like this in every `incremental` materialization, including the modified versions in other adapter plugins we maintain:
- [dbt-snowflake](https://github.com/dbt-labs/dbt-snowflake/blob/e18c501175c84d5b08d66244533e636a5a04dba1/dbt/include/snowflake/macros/materializations/incremental.sql#L74-L75)
- [dbt-bigquery](https://github.com/dbt-labs/dbt-bigquery/blob/da0d0d7e21c0bec09a9fd35baa264f915eddc55a/dbt/include/bigquery/macros/materializations/incremental.sql#L130-L131)
- [dbt-spark](https://github.com/dbt-labs/dbt-spark/blob/5ca20be56ec2d557b4fff5e42c320949040650d3/dbt/include/spark/macros/materializations/incremental/incremental.sql#L56)

jtcohen6 (changes requested):
@pol-defont-reaulx Thanks for the contribution! I know many folks have wanted this for a while, so props to you for rolling up your sleeves and starting to make it happen.

After some local testing, I've realized there's a more foundational issue at play here. Given a model like this one:
```sql
{{ config(
  materialized = 'incremental',
  on_schema_change = 'full_refresh'
) }}

with source_data as (
  
    select * from {{ ref('some_model') }}
    
    {% if is_incremental() %}
    
      -- always last 3 days (PostgreSQL)
      where date_day > (current_date - 3 days)
    
    {% endif %}
  
),

select ...
```
If the `on_schema_change` behavior is detected, and we need to move into "full refresh mode" — **we actually need to re-compile the model's SQL,** with `is_incremental` now returning `false`. Why? Otherwise, we'll be replacing the existing table with _only the past 3 days of data_, when what we really want to do is start the whole process over again, as if we had configured the model with `full_refresh: true` or passed in the `--full-refresh` flag.

I'm not sure of the right way to trigger that re-compilation, let alone how to do it from within the Jinja (materialization) context. These are [two clearly separated steps today](https://github.com/dbt-labs/dbt-core/blob/60f80056b1b09b35a22ac91c93319827bdac0e0e/core/dbt/task/base.py#L315-L332), and dbt expects a model's compiled SQL to be available before it starts doing anything materialization-related.

There's also a performance consideration: By the time we know there's been a schema change, we've _already created the temp table_ with new data, which now needs to be thrown away in favor of the "full" version. We could try hacking around that, by using [`get_columns_in_query`](https://github.com/dbt-labs/dbt-core/blob/43e3fc22c4eae4d3d901faba05e33c40f1f1dc5a/core/dbt/include/global_project/macros/adapters/columns.sql#L24-L34) (`where false limit 0`) to detect schema changes, instead of actually running + saving the query, but there's always risk involved with that sort of proxy approach.

(cc @colin-rogers-dbt - this could be a fun problem to jam on, but definitely not trivial)

---

Some more tactical considerations, typed out before I started thinking about the point above:
- Ultimately, we'll want automated testing for this. The existing functional tests for `on_schema_change` behavior in incremental models are defined here: https://github.com/dbt-labs/dbt-core/tree/main/tests/functional/incremental_schema_tests
- In order to support this behavior on all our adapters, we'll need similar changes to the incremental materializations in `dbt-snowflake`, `dbt-bigquery`, and `dbt-spark`. Several of these don't use an "alter-rename-swap" mechanism, but instead just an atomic `create or replace`.

Fleid:
It wasn't picked up on our side, so no.

@dbeatty10 do you think you could pick that up as part of the incremental effort?

jtcohen6:
This will be quite tricky to implement! As described in my comment above - we to re-compile the model's SQL mid-materialization, now with `is_incremental()` returning `False`. I think that's _possible_, but it's not something for which we have an existing pattern, and it likely requires changes outside the scope of just the incremental materialization (Jinja) code.

Fleid:
Oh, by that I meant the refinement, not the implementation!
Thanks for the reminder that it's not an easy task Jeremy :)

So no @nbsmo4, no progress on it sadly :/

dbeatty10:
Thanks for taking the time to open this PR @pol-defont-reaulx! Since opening, we've [decoupled dbt Adapters from dbt Core](https://github.com/dbt-labs/dbt-core/discussions/9171), and this code now lives in a separate repo: [dbt-adapters](https://github.com/dbt-labs/dbt-adapters).

Issue #6434, opened 2022-12-13
[CT-1652] [Regression] [main/1.4] Failure to raise initialization exception


jtcohen6:
### Is this a regression in a recent version of dbt-core?

- [X] I believe this is a regression in dbt-core functionality
- [X] I have searched the existing issues, and I could not find an existing issue for this regression

### Current Behavior

If I run `dbt` with a profile that expects an adapter to be installed, and the adapter is not in fact installed, I do not see any error from `dbt run`:

```bash
$ dbt run
```
(no output)
```bash
$ dbt --debug run
$ echo $?
2
```
(no other output)

I do see the expected error from `dbt debug`:
```bash
$ dbt debug
15:46:40  Running with dbt=1.4.0-a1
dbt version: 1.4.0-a1
python version: 3.10.9
python path: /Users/jerco/dev/product/dbt-core/env/bin/python3
os info: macOS-12.4-arm64-arm-64bit
Using profiles.yml file at /Users/jerco/.dbt/profiles.yml
Using dbt_project.yml file at /Users/jerco/dev/scratch/testy/dbt_project.yml

15:46:40  Error importing adapter: No module named 'dbt.adapters.bigquery'
Configuration:
  profiles.yml file [ERROR invalid]
  dbt_project.yml file [OK found and valid]

Required dependencies:
 - git [OK found]

1 check failed:
Profile loading failed for the following reason:
Runtime Error
  Credentials in profile "sandbox-bigquery", target "dev" invalid: Runtime Error
    Could not find adapter type bigquery!
```

### Expected/Previous Behavior

`dbt run` should raise the same error as `dbt debug`, if there is an error during initialization (e.g. trying to use an uninstalled adapter)

### Steps To Reproduce

1. Install dbt from the `main` branch
2. Specify a profile that requires an adapter that isn't installed
3. `dbt run`

### Environment

```markdown
- OS: macOS 12.4
- Python: Python 3.10.9
- dbt (working version): 1.3
- dbt (regression version): 1.4.0-a1
```

jtcohen6:
If I drop a breakpoint here (line 197):
https://github.com/dbt-labs/dbt-core/blob/83b1fee06228055f6fe573d778775160c7127454/core/dbt/main.py#L196-L198

On `main` branch:
```python
task, res = run_from_args(parsed)
*** dbt.exceptions.RuntimeException: Runtime Error
  Could not run dbt
```
On `1.3.latest` branch:
```python
[...]

Issue #6207, opened 2022-11-03
Whither UserConfig?


jtcohen6:
https://docs.getdbt.com/reference/profiles.yml#user-config

The idea in the past has been: This is tied to my specific "machine," and should hold true / set defaults for all of my dbt activity. The original options were things like: don't use colors in the terminal, don't send anonymous usage tracking, don't use partial parsing.

However, these configs don't _really_ have anything to do with connection profiles / credentials. We're all about splitting up our monolithic files. And `profiles.yml` really deserves a bigger refresh next year (https://github.com/dbt-labs/dbt-core/issues/1958 and then some).

**Questions:**
- Where should these "user configs" go instead? Should they be settable in `dbt_project.yml`? In a new root-level file that lives alongside the project? Should that file be `.gitignored` (a la `dbt.env`), or version-controlled (`dbt_config.yml`)?
- Is the distinction between ["global configs"](https://docs.getdbt.com/reference/global-configs) and ["project configs"](https://docs.getdbt.com/reference/dbt_project.yml) really meaningful? The addition of `--log-path` and `--target-path` as CLI flag / env var overrides already blurs that boundary. That said, I don't think `--model-paths` should really be a thing :)

Wherever we move the `UserConfig`, we'll need to support it in its current location for a period of time, with a clear deprecation warning.

jtcohen6:
How would we feel about a precedence order like:
1. CLI flag
2. Env var
3. `UserConfig`, in a dedicated file, **always in the root directory**, optionally version controlled

Idea being:
- All "global" configs can be set in `dbt_project.yml`, too, to set a default for everyone running this particular project. Will be overridden by the same config set at the user/machine level, env level, or invocation level
- No ifs-ands-or-buts-ing about where `UserConfig` is located, so we can safely load it first & foremost. I don't even want to have to think about `--profiles-dir` or `--project-dir`.

jtcohen6:
This comment is my attempt to summarize my latest thinking on the next steps here. Given the last few months of Fun With Flags, I think a solid answer here could provide us with much cleaner codepaths & initialization flows going forward. As always, we should aim to provide one minor version of back-compat for existing behavior, with deprecation warnings, if we can manage it.

Big idea: "Runtime configuration" can be provided by users in one of three ways:
1. CLI flags
2. Environment variables (read from system OR defined in `dbt.env` file)
3. Long-lived configuration file

Commentary:
[...]

PR #6443, opened 2022-12-14
CT-1685: Restore certain aspects of legacy logging behavior important…

### LINKED ISSUE #6437: [CT-1658] Logging changes break dbt-rpc

### Current Behavior

Due to recent changes, logs are now written to the logbook GLOBAL_LOGGER using the same numeric priority as used for native Python logs, but they are not compatible. Furthermore, the GLOBAL_LOGGER is not written to by default, and dbt-rpc does not do its own log set up. We need to correct both issues so that dbt-core logging behavior with respect to the legacy logbook code remains the same in 1.4 as it was in 1.3.

### Expected Behavior

When DBT_ENABLE_LEGACY_LOGGING is true, dbt-core should exhibit the same log output behaviors as it did in 1.3.

### Steps To Reproduce

The issue can be seen in the GitHub actions for the dbt-rpc build, which are currently failing.

peterallenwebb:
resolves #6437 

### Description

These changes restore the behavior described in the associated issue. The legacy (logbook) logger is now active by default, and the appropriate log level is applied to ensure the events are seen by dbt-rpc.

### Checklist

### DIFF
--- a/core/dbt/events/functions.py
+++ b/core/dbt/events/functions.py
@@ -18,10 +18,6 @@ import uuid
 LOG_VERSION = 3
 metadata_vars: Optional[Dict[str, str]] = None
 
-# The default event manager will not log anything, but some tests run code that
-# generates events, without configuring the event manager.
-EVENT_MANAGER: EventManager = EventManager()
-
 
 def setup_event_logger(log_path: str, level_override: Optional[EventLevel] = None):
     cleanup_event_logger()
@@ -113,6 +109,16 @@ def cleanup_event_logger():
     EVENT_MANAGER.loggers.clear()
     EVENT_MANAGER.callbacks.clear()
 
+
+# The default event manager will not log anything, but some tests run code that
+# generates events, without configuring the event manager, so we create an empty
+# manager here until there is a better testing strategy in place.
+EVENT_MANAGER: EventManager = EventManager()
+
+# Since dbt-rpc does not do its own log setup, we set up logbook if legacy
+# logging is enabled.
+if flags.ENABLE_LEGACY_LOGGER:
+    EVENT_MANAGER.add_logger(_get_logbook_log_config(None))
 
 # This global, and the following two functions for capturing stdout logs are
 # an unpleasant hack we intend to remove as part of API-ification. The GitHub
--- a/.changes/unreleased/Under the Hood-20221213-214106.yaml
+++ b/.changes/unreleased/Under the Hood-20221213-214106.yaml
@@ -1,0 +1,7 @@
+kind: Under the Hood
+body: Restore important legacy logging behaviors, following refactor which removed
+  them
+time: 2022-12-13T21:41:06.815133-05:00
+custom:
+  Author: peterallenwebb
+  Issue: "6437"
--- a/core/dbt/events/eventmgr.py
+++ b/core/dbt/events/eventmgr.py
@@ -47,6 +47,33 @@ _log_level_map = {
 }
 
 
+# We should consider fixing the problem, but log_level() can return a string for
+# DynamicLevel events, even thought it is supposed to return an EventLevel. This
+# function gets a string for the level, no matter what.
+def _get_level_str(e: BaseEvent) -> str:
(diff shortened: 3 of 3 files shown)


core/dbt/events/functions.py
```diff
@@ -114,6 +110,16 @@ def cleanup_event_logger():
     EVENT_MANAGER.callbacks.clear()
 
 
+# The default event manager will not log anything, but some tests run code that
+# generates events, without configuring the event manager, so we create an empty
+# manager here until there is a better testing strategy in place.
+EVENT_MANAGER: EventManager = EventManager()
```
jtcohen6, line 116: I noticed a functional regression related to the fact that `EVENT_MANAGER` does not have any logs configured, prior to `setup_event_logger` being called. If dbt runs into configuration errors during initialization, they are not surfaced to the user: https://github.com/dbt-labs/dbt-core/issues/6434

gshank: How would that work in dbt-rpc?

gshank (approved):

core/dbt/events/functions.py
jtcohen6: IMO only stdout logging is acceptable for now. My sense is, we really only expect these logs to fire if something goes wrong during initialization. It's okay to have some hardcoded, generic, unconfigurable behavior that just ensures we're showing the error to the user.

We'll need to figure out the story for `dbt-server`, since I think the initial plan is to tail the `logs/dbt.log` file. If we think the longer-term answer is registering a custom callback, that might solve our problem.

As far as dbt-rpc, all we need is ongoing compatibility with the legacy logger for v1.4. The way I understand the code in this PR, if the `ENABLE_LEGACY_LOGGER` env var is enabled, we'll immediately flip on the legacy logger, without waiting for the call to `setup_event_logger`.

peterallenwebb: @jtcohen6 Your understanding is correct w.r.t. dbt-rpc

From speaking to @racheldaniel, I have a fairly good idea how dbt-server is using logging now. They do their own configuration of the EVENT_MANAGER before running tasks, so I feel confident that we have a good story for them.

I'm going to prepare a PR to address #6434, now. I believe that will restore all of the important logging behaviors we had in 1.3, and I have to apologize for not considering how important it was to get something set up before the call to setup_event_logger() is made.

With all that said, there are some interesting thing to note about the file logging behavior of the legacy logger. I found earlier this week that it starts recording messages after it is set up, which is immediately dbt starts. It then writes the recorded messages to the log file (if any) when the correct file path has been determined, and stops recording. My hope is that in 1.5 we can find a way to start all logging, both console and file, immediately when dbt starts. Does that seem feasible?

jtcohen6: > I have to apologize for not considering how important it was to get something set up before the call to setup_event_logger() is made.

I'm glad we caught it early! I think the failure here was lack of sufficient inline code comment around the event system initialization, to scare you away from what was a perfectly sensible refactor. "We need to configure these defaults here, even though it looks like we shouldn't, because some log events _might_ arrive before we have the real inputs for proper configuration — namely, if we encounter an early exception."

> I found earlier this week that it starts recording messages after it is set up, which is immediately dbt starts. It then writes the recorded messages to the log file (if any) when the correct file path has been determined, and stops recording.

IIRC, it does this by storing logs in a "buffer," and then "replaying" the buffer once the proper log path has been determined. A little too magical for my taste buds, but this might be an acceptable pattern to emulate.

https://github.com/dbt-labs/dbt-core/blob/5e4e917de5f4da7b24862a2d9b14a2709686e3ff/core/dbt/logger.py#L417-L421

> My hope is that in 1.5 we can find a way to start all logging, both console and file, immediately when dbt starts. Does that seem feasible?

Maybe! We could roll our own similar version of a stateful "buffer"? Or split `log-path` out of `dbt_project.yml`, so that we can know where we're logging without having to parse any config/files? (Feels related: #6207)

FWIW in the happy path, when config initialization succeeds (99% of the time), I don't think we're firing any log events before `setup_event_logger` is called.

PR #6463, opened 2022-12-20
remove packaging<22 cap

### LINKED ISSUE #6461: [CT-1688] [Feature] Remove packaging <22 version cap

### Describe the feature

This isn't really a "feature", but no other category fit. Please consider lifting `dbt-core`'s current upper version bound of `packaging<22`, it prevents installation in the same environment as tox 4, and I doubt the cap actually needs to be there-- just delete any references to `packaging.LegacyVersion` (which used to be a possible return from `packaging.version.parse`) and you can probably remove it.

_No response_

### Who will this benefit?

Anyone working with `dbt-core` in an environment with a lot of other packages, and in particular in a dev environment with tox 4.

I'd be happy to put up a PR, I grepped dbt-core codebase and packaging only used in a few places

### Anything else?

_No response_

smackesey:
resolves #6461 

### Description

This PR removes the version cap on `packaging`. This cap was preventing dbt-core from being installed in the same environment as tox 4. It also did not appear to be necessary-- the main breaking change from packaging <22 to 22+ is the drop of `LegacyVersion` (which used to be a possible return from `packaging.parse`). `dbt-core` did not contain any direct references to `LegacyVersion`, and AFAICT does not depend on any functionality specifically in `packaging<22`, but I might be wrong.

### Checklist


gshank (changes requested):
dbt Core does actually use LegacyVersion under the covers in that the version code falls back to legacy version if the version method doesn't work. So switching to later versions of packaging might break version strings that are currently accepted.  Which is why a number of the test cases are currently failing.

If you could identify some versions which succeed currently and would fail under a later version of packaging, we could make a decision about whether this is a change that we want to do at this point.

gshank:
Another alternative might be to use our own version function. I haven't really looked into how feasible that is though. Pinning this forever is certainly not optimal.

core/dbt/semver.py
```diff
@@ -142,13 +142,10 @@ def compare(self, other):
...
-            with warnings.catch_warnings():
-                warnings.simplefilter("ignore", category=DeprecationWarning)
-                if packaging_version.parse(a) > packaging_version.parse(b):
-                    return 1
-                elif packaging_version.parse(a) < packaging_version.parse(b):
-                    return -1
+            if packaging_version.parse(a) > packaging_version.parse(b):
+                return 1
+            elif packaging_version.parse(a) < packaging_version.parse(b):
+                return -1
```
jtcohen6, line 148: _adapted from Slack: @jtcohen6 + @gshank on 6 Jan_

There are two things at play here, leading to failures in [`test/unit/test_semver.py`](https://github.com/dbt-labs/dbt-core/blob/main/test/unit/test_semver.py):
1. This logic is not quite right! We're passing just _part_ of a version string into `packaging.version.parse()`, rather than the whole string. This is the cause of some of the failing tests.
2. Unfortunately, there are also some real discrepancies between [PEP 440](https://peps.python.org/pep-0440) and [SemVer](https://semver.org/), and `packaging==22.0` lands firmly on the side of PEP 440. This is the cause of the rest of the failing tests.

These versions are still supported by `packaging==22.0`. (Although PEP 440 says not to use a hyphen, and SemVer says to use a hyphen.) The failure is because we're trying to parse just the prerelease identifier (a1, b2) as if it's an entire version string:
- `'1.2.0a1'`
- `'1.2.0-a1'`
- `'1.2.0b2'`
- `'1.2.0-b2'`

These are valid per SemVer, but they are not valid per PEP 440, and so they fail on `packaging.version.parse()` starting with `packaging==22.0`:
- `'2.2.0-fishtown-beta'`
- `'2.2.0asdf'`

https://peps.python.org/pep-0440/#pre-releases
> The pre-release segment consists of an alphabetical identifier for the pre-release phase, along with a non-negative integer value. Pre-releases for a given release are ordered first by phase (alpha, beta, release candidate) and then by the numerical component within that phase.
but they are valid

https://semver.org/#spec-item-9
> A pre-release version MAY be denoted by appending a hyphen and a series of dot separated identifiers immediately following the patch version. Identifiers MUST comprise only ASCII alphanumerics and hyphens [0-9A-Za-z-].

@gshank
> It would be nice to not pin "packaging" because of it... Maybe at this point we should defer this to a re-write

dbeatty10:
@smackesey we couldn't easily lift the upper bound from `packaging` without a re-write of how we compare prerelease versions. We ended up doing that re-write in https://github.com/dbt-labs/dbt-core/pull/6838 which supersedes this PR.

Thank you for opening the original issue in https://github.com/dbt-labs/dbt-core/issues/6461, drafting this PR, and spawning the discussion in https://github.com/dbt-labs/dbt-core/discussions/6495 -- we appreciate all of them!

PR #6488, opened 2022-12-26
add defer_to_manifest in before_run to fix faulty deferred docs generate

### LINKED ISSUE #6124: [CT-1386] [Bug] `dbt docs generate` with `--defer`

### Current Behavior

It's great that dbt 1.3 makes it possible to use `dbt docs generate` with `--defer`, but it appears to be not quite working, unless I'm not understanding something.

When running `dbt docs generate (--select state:modified+) --defer --state /path/to/manifest` I'm getting an error `Object X does not exist or not authorized.` The cause for this error is that X indeed does not exist, so the `ref` referring to it should have been resolved using the manifest.

In my situation, the faulty `ref` actually happens in a macro, perhaps that's why it wasn't resolved properly? What is strange is that this doesn't seem to be a problem for `dbt run`.

### Expected Behavior

When using `dbt docs generate` with `--defer`, all `ref` calls should be resolved as described in the [dbt doc](https://docs.getdbt.com/reference/node-selection/defer).

By the way, it's unclear to me whether I'm supposed to use a `--select` when doing `dbt docs generate` with `--defer`. I would assume so, since it's supposed to play a role in resolving  the `ref`s according to the doc, but it doesn't seem to make a difference.

### Steps To Reproduce

1. Create a macro that `ref`s some model X.
2. Run `dbt docs generate` with `--defer` (and possibly other flags).
3. Observe that the `ref` in the macro is not resolved properly and produces an error.

### Relevant log output

_No response_

### Environment

```markdown
- OS:
- Python:
- dbt:
```

snowflake

### Additional Context

_No response_

mivanicova:
resolves #6124

### Description

The `dbt docs generate --defer` command does not work as expected as it ignores the `state`. This PR fixes the issue by:
- adding call to defer_to_manifest to before_run  of compile/docs generate
- extending integration tests to validate that the  generated catalog is  indeed coming from the deferred state

### Checklist

### DIFF
--- a/.changes/unreleased/Fixes-20221226-010211.yaml
+++ b/.changes/unreleased/Fixes-20221226-010211.yaml
@@ -1,0 +1,6 @@
+kind: Fixes
+body: fix docs generate --defer by adding defer_to_manifest to before_run
+time: 2022-12-26T01:02:11.630614+01:00
+custom:
+  Author: mivanicova
+  Issue: "6488"
--- a/core/dbt/task/freshness.py
+++ b/core/dbt/task/freshness.py
@@ -147,6 +147,10 @@ class FreshnessSelector(ResourceTypeSelector):
 
 
 class FreshnessTask(GraphRunnableTask):
+    def defer_to_manifest(self, adapter, selected_uids):
+        # freshness don't defer
+        return
+
     def result_path(self):
         if self.args.output:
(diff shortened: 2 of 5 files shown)


ChenyuLInx (approved):
@mivanicova  Thanks for adding this fix and adding the base function! Validated that removing the line would make the test fail.

PR #6493, opened 2022-12-30
Adds the meta field to node_info in logs

### LINKED ISSUE #6216: [CT-1474] [Feature] Let users add fields to strucutred logs

No, I have made other feature requests.

### Describe the feature

At the moment, structured logs have a predefined set of JSON properties ([docs](https://docs.getdbt.com/reference/events-logging#structured-logging)). 

I would like to to be able to add fields to the structured logs to consume in downstream systems.

This ask was inspired by the [query-comment marco](https://docs.getdbt.com/reference/project-configs/query-comment#advanced-use-a-macro-to-generate-a-comment). By customizing the macro, I can "log" exactly the details I need, removing info I don't need and adding info I do need.

Here's my work-around to augment logs with additional data. 

Instead of consuming logs as they are generated to stdout, I write the logs to file. Then I execute a Python script that adds data from `manifest.json` into the logs file (by `data.node_info.unique_id` in logs) and then re-emits the logs.

### Who will this benefit?

Use-case: add node tags to the `data.node_info` property in logs. By tagging a node with team ownership, the log consumer can alert different teams based on the failures.

Example:
```
{"level": "error", "data": { "node_info": {"tags": ["finance_team"], ...}, ...}, ...}
```
The consumer would see the `"level": "error"` and prepare to alert. Then the consumer would see that the node is tagged with `"finance_team"`, so instead of sending a general alert, a more narrow, finance alert is sent.

Yes, I could help if we are interested in the feature.

### Anything else?

_No response_

tmastny:
resolves #6216

### Description

Added `meta` to the suggested protocols and classes to ensure it is logged in `node_info`. Also added a unit and functional test to confirm `meta` is added to the logs.

### Checklist

(documentation issue: https://github.com/dbt-labs/docs.getdbt.com/issues/2632#issue-1514640465 )


gshank:
Protobuf is kind of picky about what exactly goes into the message fields. I don't think that putting a string into the "meta" key of node_info will actually work. In addition, "meta" is defined as only having string keys and string values, so I think you would probably have to convert a meta dictionary into all strings or use a different kind of structure. (protobuf struct?) You should probably add a test for a meta dictionary with non-string keys.

tests/unit/test_proto_events.py
```diff
@@ -80,6 +80,7 @@ def test_exception_events():
 
 
 def test_node_info_events():
+    meta_dict = {"string-key1": ["value1", 2], "string-key2": {"nested-dict-key": "value2"}, 1: "value-from-non-string-key"}
```
tmastny, line 83: Hi @gshank, thanks for the feedback! On your suggestion, I added a unit test to see if protobuf could serialize and de-serialize a more complicated meta object, including non-string keys and nested dictionaries.

The test passes, so if I understand this test correctly, I think it is correctly handling these cases. Does this satisfy your concerns, or should I look into protobuf structs?

Thanks!

gshank: I misspoke before. The problem is with non-string values in the meta dictionary, not non-string keys. Also tests are failing.

tmastny: Hi @gshank, I fixed the broken test and also added tests for various kinds of non-string values and non-string keys in the meta dictionary: https://github.com/dbt-labs/dbt-core/pull/6493/files#diff-f9fe1fc7c69b0361dab53616751b87850948b5efd845478da472d2a68eec08afR83

gshank:
I stringified the meta dictionary before adding it to ndoe_info, because protobuf serialization is much more rigid than python. Unfortunately we don't have a test case to cover it. We can work on making the format of meta more flexible after this.

PR #6494, opened 2023-01-02
Partial parsing bug with empty schema file - ensure None is not passed to load_yaml_text

### LINKED ISSUE #4850: [CT-343] partial parse conflicts with generate_schema_name changes

### Current Behavior

We have a project with a custom generate_schema_name macro.
If I change the macro generate_schema_name, project compilation fails with error `a string or stream input is required`.
If I run a `dbt clean` before compiling, compilation completes without errors.

### Expected Behavior

Compilation should not depend on `dbt clean` 

### Steps To Reproduce

dbt clean
dbt deps
dbt compile --> no errors

modify generate_schema_name macro (even adding a jinja comment):

dbt compile --> error

10:01:21  Running with dbt=1.0.3
10:01:22  Change detected to override macro used during parsing. Starting full parse.
10:01:34  Encountered an error:
a string or stream input is required

### Relevant log output

_No response_

### Environment

```markdown
- OS: Windows 10
- Python: 3.7.9
- dbt-core: 1.0.3
- dbt-bigquery: 1.0.0
```

bigquery

### Additional Context

_No response_

gshank:
resolves #4850

### Description

The "contents" of a file object is defined as Optional[str] but load_yaml_text does not support passing in None from contents. Ensure that if field is set to None, an empty string is passed.

### Checklist

### DIFF
--- a/core/dbt/flags.py
+++ b/core/dbt/flags.py
@@ -1,7 +1,9 @@
-import os
+# Do not import the os package because we expose this package in jinja
+from os import name as os_name, path as os_path, getcwd as os_getcwd, getenv as os_getenv
 import multiprocessing
+from argparse import Namespace
 
-if os.name != "nt":
+if os_name != "nt":
     # https://bugs.python.org/issue41567
     import multiprocessing.popen_spawn_posix  # type: ignore
 from pathlib import Path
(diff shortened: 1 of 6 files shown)


core/dbt/parser/schemas.py
```diff
@@ -100,8 +100,11 @@
 def yaml_from_file(source_file: SchemaSourceFile) -> Dict[str, Any]:
     """If loading the yaml fails, raise an exception."""
     path = source_file.path.relative_path
+    # File contents filed is Optional, and is sometimes None, so ensure
+    # we are not passing in None to load_yaml_text.
+    contents = source_file.contents if source_file.contents else ""
```
mikealfare, line 105: I can go either way on this, but this could be:
```contents = source_file.contents or ""```

Taken further, this function could be:
```
def yaml_from_file(source_file: SchemaSourceFile) -> Dict[str, Any]:
    try:
        return load_yaml_text(source_file.contents or "", source_file.path)
    except ValidationException as e:
        raise YamlLoadFailure(source_file.project_name, source_file.path.relative_path, e)
```
The initial variables of `path` and `contents` are only used once, so the only benefit is readability. However, in both cases I need to refer to the variable definition to see what's being passed in; i.e. `contents` is forced to have a value, hence is not quite the same as source_file.contents, and `path` is actually the relative path, and not the absolute path. If these are moved into the function call, it's more readable to me (though that may differ by user). Also, the docstring is basically the code at that point, so it's no longer needed.

As an alternative for the None vs "" issue with `load_yaml_text()`, should we consider overriding None with "" in `load_yaml_text()` itself? In other words, will this happen elsewhere? Will we always want to pass in an empty string when we have a None? Or are there valid scenarios where we want to pass in a None for this argument and generate the resulting error?

File: test/integration/068_partial_parsing_tests/test_partial_parsing.py [hunk 2]
```diff
@@ -477,6 +477,8 @@ def test_postgres_skip_macros(self):
         # initial run so we have a msgpack file
         self.setup_directories()
         self.copy_file('test-files/model_one.sql', 'models/model_one.sql')
+        # use empty_schema file for bug #4850
+        self.copy_file('test-files/empty_schema.yml', 'models/eschema.yml')
```
mikealfare, line 481: This test used to run always without this file being present, but now runs always with this file being. So we effectively lost the integration test scenario where this file doesn't exist at all. Do we care about that? We may not; I don't know enough about this functionality.

gshank: No, we don't care about that.

core/dbt/parser/schemas.py
gshank: I changed yaml_from_file as per your suggestion. I'd rather not change load_yaml_text at this point because we're going to backport this and we want to make backport changes as minimal as possible.

core/dbt/flags.py
```diff
@@ -1,7 +1,9 @@
-import os
+# Do not import the os package because we expose this package in jinja
```
mikealfare, line 1: I don't understand this comment. I believe the `os` package is still running, even if only a few objects are pulled in from that namespace. In other words, I don't think this is doing anything different than the original version.

leahwicz (approved):

Issue #6165, opened 2022-10-27
[CT-1436] Treat warnings as errors for specific events, based on user configuration


jtcohen6:
Today, `--warn-error` is a simple on/off switch that turns _all_ warnings into errors. We've gotten multiple requests (https://github.com/dbt-labs/dbt-core/issues/4383, https://github.com/dbt-labs/dbt-core/issues/6080) from users who want finer-grained control for raising errors on specific types of warnings.

We should allow users to configure exactly which event classes they want to raise as exceptions instead of warnings. This will be enabled by https://github.com/dbt-labs/dbt-core/pull/6064, which will turn all `warn_or_error` call sites into structured events.

**Where should this configuration live?** It feels like "runtime" config, rather than something that goes in project code. But it might be awkward to pass as a CLI flag or env var. The best places feels like the "user config" which currently lives in `profiles.yml`.
```yml
config:
  warn_error: true     # all of them

config:
  warn_error_events:  # scoped down to specific events
    - NoNodesForSelectionCriteria
    - UnusedResourceConfigPath
```

```python
def warn_or_error(event, node):
    # pseudo code!
    if flags.WARN_ERROR or type(event) in flags.WARN_ERROR_EVENTS:
        from dbt.exceptions import raise_compiler_error
        raise_compiler_error(scrub_secrets(event.info.msg, env_secrets()), node)
    else:
        fire_event(event)
```

emmyoop:
We need to define how we want this to behave a bit more concretely by allowing include or exclude of specific warn events.

Error for all events _except_ the ones listed
```yaml
config:
  warn_error: true     # all of them

config:
  warn_error_override:
    - NoNodesForSelectionCriteria
    - UnusedResourceConfigPath
```

Warn for all events _except_ the one listed
```yaml
config:
  warn_error: false     # none of them (default)

config:
  warn_error_override:
    - NoNodesForSelectionCriteria
    - UnusedResourceConfigPath
```

Question: Should there be separate configs for what to include/exclude from the list based on if `warn_error` is true/false or a single config that just swaps including or excluding specific events.

jtcohen6:
I like the framing of this as `_override`!

emmyoop:
As part of this work, we should remove `warn_or_error` entirely to make the `warn_error` config work for _all_ warnings, not just a subset.

github-christophe-oudar:
[...]

PR #6520, opened 2023-01-04
Add --warn-error-options

MichelleArk:
resolves https://github.com/dbt-labs/dbt-core/issues/6165

### Checklist

- [x] I have run this code in development and it appears to resolve the stated issue 🎩 

--warn-error
```sh
❯ dbt run --select nada
20:29:07  Running with dbt=1.4.0-b1
20:29:07  Found 5 models, 20 tests, 0 snapshots, 0 analyses, 303 macros, 0 operations, 3 seed files, 0 sources, 1 exposure, 0 metrics
20:29:07  The selection criterion 'nada' does not match any nodes
20:29:07  
20:29:07  Nothing to do. Try checking your model configs and model specification args

❯ dbt --warn-error run -m nada
20:29:12  Running with dbt=1.4.0-b1
20:29:12  Found 5 models, 20 tests, 0 snapshots, 0 analyses, 303 macros, 0 operations, 3 seed files, 0 sources, 1 exposure, 0 metrics
20:29:13  Encountered an error:
Compilation Error
  The selection criterion 'nada' does not match any nodes
```
Mutually exclusive
```sh
❯ dbt --warn-error --warn-error-options '{"include": "all"}' run -m nada
usage: dbt [-h] [--version] [-r RECORD_TIMING_INFO] [-d] [--log-format {text,json,default}] [--no-write-json]
           [--use-colors | --no-use-colors] [--printer-width PRINTER_WIDTH] [--warn-error | --warn-error-options WARN_ERROR_OPTIONS]
           [--no-version-check] [--partial-parse | --no-partial-parse] [--use-experimental-parser] [--no-static-parser]
           [--profiles-dir PROFILES_DIR] [--no-anonymous-usage-stats] [-x] [-q] [--no-print]
           [--cache-selected-only | --no-cache-selected-only]
           {docs,source,init,clean,debug,deps,list,ls,build,snapshot,run,compile,parse,test,seed,run-operation} ...
dbt: error: argument --warn-error-options: not allowed with argument --warn-error

# warn_error: True in user config (profiles.yml)
❯ dbt --warn-error-options '{"include": "all"}' run -m nada
20:32:05  Encountered an error:
warn_error_options: not allowed with argument warn_error
20:32:05  Traceback (most recent call last):
  File "/Users/michelleark/src/dbt-core/core/dbt/main.py", line 135, in main
    results, succeeded = handle_and_check(args)
  File "/Users/michelleark/src/dbt-core/core/dbt/main.py", line 180, in handle_and_check
    flags.set_from_args(parsed, user_config)
  File "/Users/michelleark/src/dbt-core/core/dbt/flags.py", line 153, in set_from_args
    _check_mutually_exclusive(["WARN_ERROR", "WARN_ERROR_OPTIONS"], args, user_config)
  File "/Users/michelleark/src/dbt-core/core/dbt/flags.py", line 201, in _check_mutually_exclusive
    raise ValueError(f"{flag.lower()}: not allowed with argument {set_flag.lower()}")
ValueError: warn_error_options: not allowed with argument warn_error
```
--warn-error-options
```sh
❯ dbt --warn-error-options '{"include": "all"}' run -m nada
20:32:56  Running with dbt=1.4.0-b1
20:32:57  Found 5 models, 20 tests, 0 snapshots, 0 analyses, 303 macros, 0 operations, 3 seed files, 0 sources, 1 exposure, 0 metrics
20:32:57  Encountered an error:
Compilation Error
  The selection criterion 'nada' does not match any nodes

❯ dbt --warn-error-options '{"include": "all", "exclude": [NoNodesForSelectionCriteria,NothingToDo]}' run -m nada
20:36:53  Running with dbt=1.4.0-b1
20:36:53  Found 5 models, 20 tests, 0 snapshots, 0 analyses, 303 macros, 0 operations, 3 seed files, 0 sources, 1 exposure, 0 metrics
20:36:53  The selection criterion 'nada' does not match any nodes
20:36:53  
20:36:53  Nothing to do. Try checking your model configs and model specification args

❯ dbt --warn-error-options '{"include": [NoNodesForSelectionCriteria]}' run -m nada
20:37:21  Running with dbt=1.4.0-b1
20:37:22  Found 5 models, 20 tests, 0 snapshots, 0 analyses, 303 macros, 0 operations, 3 seed files, 0 sources, 1 exposure, 0 metrics
20:37:22  Encountered an error:
Compilation Error
  The selection criterion 'nada' does not match any nodes
```
- [x] I have [opened an issue to add/update docs](https://github.com/dbt-labs/docs.getdbt.com/issues/new/choose), or docs changes are not required/relevant for this PR: https://github.com/dbt-labs/docs.getdbt.com/pull/2691


core/dbt/helper_types.py
```diff
@@ -88,6 +90,59 @@ class NoValue(dbtClassMixin):
     novalue: NVEnum = field(default_factory=lambda: NVEnum.novalue)
 
 
+@dataclass
+class IncludeExclude(dbtClassMixin):
```
jtcohen6, line 94: nice!

core/dbt/cli/params.py
```diff
@@ -358,9 +358,20 @@
...
     help="If dbt would normally warn, instead raise an exception. Examples include --models that selects nothing, deprecations, configurations with no associated models, invalid test configurations, and missing sources/refs in tests.",
+    default=None,
+    flag_value=True,
+)
+
+warn_error_options = click.option(
+    "--warn-error-options",
+    envvar="DBT_WARN_ERROR_OPTIONS",
+    default=None,
+    help="""If dbt would normally warn, instead raise an exception based on include/exclude configuration. Examples include --models that selects nothing, deprecations, configurations with no associated models, invalid test configurations,
```
jtcohen6, line 372: careful, this is how you prove you've been using dbt for a while :)

`--select` has been canonical since v0.21
```suggestion
    help="""If dbt would normally warn, instead raise an exception based on include/exclude configuration. Examples include --select that selects nothing, deprecations, configurations with no associated models, invalid test configurations,
```

core/dbt/main.py
```diff
@@ -1018,6 +1019,19 @@ def parse_args(args, cls=DBTArgumentParser):
         """,
     )
 
+    warn_error_flag.add_argument(
+        "--warn-error-options",
+        default=None,
+        help="""
+        If dbt would normally warn, instead raise an exception based on
+        include/exclude configuration. Examples include --models that selects
```
jtcohen6, line 1027: `--models` → `--select` here as well

core/dbt/main.py
```diff
@@ -1006,7 +1006,8 @@ def parse_args(args, cls=DBTArgumentParser):
         """,
     )
 
-    p.add_argument(
+    warn_error_flag = p.add_mutually_exclusive_group()
```
jtcohen6, line 1009: Is this the argparse way to say `--warn-error` + `--warn-error-options` are mutually exclusive? I'm guessing there's also a `click`-y way to do it?

I'm thinking we'll need to add a check within `flags.py` as well, because it's possible that one of these configs was supplied as a CLI flag, and the other was set as an env var or UserConfig attr

jtcohen6 (commented):
quick first pass at the quick first pass

MichelleArk:
@stu-k - A couple big files were deleted by the API doc generation bot (`core/dbt/docs/build/html/_static/_sphinx_javascript_frameworks_compat.js` and `core/dbt/docs/build/html/_static/jquery-3.6.0.js`) - is that expected?

core/dbt/events/types.py
```diff
@@ -108,12 +108,12 @@ def message(self) -> str:
 
 
 @dataclass
-class InvalidVarsYAML(ErrorLevel, pt.InvalidVarsYAML):
+class InvalidOptionYAML(ErrorLevel, pt.InvalidOptionYAML):
```
MichelleArk, line 111: @gshank - Does this seem like a reasonable change - or would a separate event make more sense? `warn_error_options` is similar to `vars` in its need to be parsed from a yaml string to a dictionary and can reuse [same `parse_cli_vars` method in its implementation](https://github.com/dbt-labs/dbt-core/pull/6520/files#diff-a99d2dd08715c05aebaa52b4cb4471289a0821b57fe46a9677402415feb6cf89R15-R19).

That said, if the goal is to have finer-grain events without much parametrization - we could alternatively create a new event specifically for `warn_error_options` and still implement the underlying parsing in a reusable way. I'm open to either!

stu-k:
@MichelleArk All the workflow does if it detects changes is run `make clean` and `make html`, so if those files were removed as part of either of those operations, I would assume it's sphinx running as expected.

core/dbt/helper_types.py
```diff
@@ -88,6 +90,63 @@ class NoValue(dbtClassMixin):
...
+        pass
+
+
+class WarnErrorOptions(IncludeExclude):
+    # TODO: this method can be removed once the click CLI is in use
+    @classmethod
+    def from_yaml_string(cls, warn_error_options_str: Optional[str]):
+
+        # TODO: resolve circular import
+        from dbt.config.utils import parse_cli_yaml_string
```
jtcohen6, line 132: Getting a weird error from this method when I defined my config as:
```yml
# profiles.yml
config:
  warn_error_options:
    include:
      - NoNodesForSelectionCriteria
      - NothingToDo
```
```python
ipdb> warn_error_options_str
{'include': ['NoNodesForSelectionCriteria', 'NothingToDo']}
ipdb> from dbt.clients import yaml_helper
ipdb> yaml_helper.load_yaml_text(warn_error_options_str)
*** TypeError: a string or stream input is required
```
I could be doing something really dumb? I do think it's valid yaml, though, which is why I'm not seeing the nicer exception

core/dbt/flags.py
```diff
@@ -188,20 +193,36 @@ def get_flag_value(flag, args, user_config):
     return flag_value
 
 
+def _check_mutually_exclusive(group, args, user_config):
+    set_flag = None
+    for flag in group:
+        flag_set_by_user = not _flag_value_from_default(flag, args, user_config)
+        if flag_set_by_user and set_flag:
+            raise ValueError(f"{flag.lower()}: not allowed with argument {set_flag.lower()}")
```
jtcohen6, line 201: This could be nicer, in cases where I've set `warn_error` in `UserConfig` and I'm also passing the conflicting `--warn-error-options` flag on the CLI:
```
ValueError: warn_error_options: not allowed with argument warn_error
```
Sounds like it will be easier to handle in the new `click` CLI. It serves its purpose for now, and I'm happy with making it nicer in v1.5+

jtcohen6 (commented):
Tried a bunch of different combinations, all of them looking good!

I did run into a weird issue with yaml loading when `warn_error_options` is defined in `UserConfig` (`profiles.yml`)

core/dbt/events/types.py
gshank: Opinions might differ, but in generally right now I'm in favor of having fewer logging events rather than more.

core/dbt/helper_types.py
```diff
@@ -88,6 +90,63 @@ class NoValue(dbtClassMixin):
...
+            exclude=warn_error_options.get("exclude", []),
+        )
+
+    def _validate_items(self, items: List[str]):
+        valid_exception_names = set(
+            [name for name, cls in dbt_event_types.__dict__.items() if isinstance(cls, type)]
+        )
+        for item in items:
+            if item not in valid_exception_names:
+                raise ValidationException(f"{item} is not a valid dbt error name.")
```
jtcohen6, line 147: It seems like env var values might not preserve casing:
```bash
$ DBT_WARN_ERROR_OPTIONS='{"include": ["NoNodesForSelectionCriteria"]}' dbt run --select alsiuhfasd
21:07:25  Running with dbt=1.4.0-b1
21:07:25  Found 3 models, 0 tests, 0 snapshots, 0 analyses, 291 macros, 0 operations, 0 seed files, 0 sources, 0 exposures, 0 metrics
21:07:25  Encountered an error:
Runtime Error
  nonodesforselectioncriteria is not a valid dbt error name.
```
Could we just make this a case-insensitive comparison?

MichelleArk: > It seems like env var values might not preserve casing:

Yeah, looks like we lower any env var values here: https://github.com/dbt-labs/dbt-core/blob/main/core/dbt/flags.py#L214-L218. 

Does this make sense to do for the [`_NON_BOOLEAN_FLAGS`](https://github.com/dbt-labs/dbt-core/blob/main/core/dbt/flags.py#L50-L57) though? For example, lowercasing `LOG_PATH` or `PROFILES_DIR` could lead to surprising (log path different from what  was configured) or non-functional (profiles dir should be case-sensitive) behaviour. Would also be different behaviour depending on whether the flag was configured from the environment vs arg/user config.

jtcohen6: I take it you're asking, should we be doing that lowercasing at all? I agree it's an odd choice. Specifically the bit about functionally different behavior if passed as env var vs. CLI flag does not feel right! Shall we rip it out? No time like a new minor version to do it

core/dbt/helper_types.py
MichelleArk: Good catch! Fixed 👍

core/dbt/helper_types.py
MichelleArk: done!

jtcohen6 (approved):

MichelleArk:
Addressed comments, resolved some merge conflicts, and re-tested (with `dbt clean` this time, to avoid having to fix adapter error messages): 

The following all worked as expected 🎩 
* `dbt --warn-error clean` -> errors because of `PackageInstallPathDeprecation`
* `dbt --warn-error-options '{"include": [PackageInstallPathDeprecation]}' clean` -> errors
* `dbt --warn-error-options '{"include": "all"}' clean` -> errors
*  `dbt --warn-error-options '{"include": "all", "exclude": [PackageInstallPathDeprecation]}' clean` -> no error
* `dbt --warn-error-options '{"include": "all"}' --warn-error clean` -> mutually exclusive error
* `dbt --warn-error-options '{"include": "all"}' clean` (warn_error: True in user_config` -> mutually exclusive error
* `DBT_WARN_ERROR_OPTIONS='{"include": ["PackageInstallPathDeprecation"]}' dbt clean` -> errors


```
# profiles.yml
config:
  warn_error_options:
    include:
      - PackageInstallPathDeprecation
 ```
-> errors

```
# profiles.yml
config:
  warn_error_options:
    include: 'all'
    exclude: 
      - PackageInstallPathDeprecation
 ```
-> does not error!

core/dbt/task/runnable.py
```diff
@@ -154,7 +154,7 @@ def get_node_selector(self) -> NodeSelector:
 
     @abstractmethod
     def defer_to_manifest(self, adapter, selected_uids: AbstractSet[str]):
-        raise NotImplementedException(f"defer_to_manifest not implemented for task {type(self)}")
+        raise NotImplementedError(f"defer_to_manifest not implemented for task {type(self)}")
```
MichelleArk, line 157: @emmyoop - Noticed this import error via [mypy from the latest changes](https://github.com/dbt-labs/dbt-core/actions/runs/3888238595/jobs/6635350189), addressed it here as part of my merge.

emmyoop: thanks.  no idea how i missed that.

PR #6539, opened 2023-01-06
Rename exceptions

### LINKED ISSUE #6460: [CT-1687] Rename every exception to end in `Error`
Exceptions have been cleaned up, but let's also standardize the naming pattern.

> Looks great, but should we rename every exception to end in `Error` similar to Python's [built-in exceptions](https://docs.python.org/3/library/exceptions.html#concrete-exceptions)?
> 
> At the very least, we should standardize on using either `Error` or `Exception` uniformly everywhere?

_Originally posted by @aranke in https://github.com/dbt-labs/dbt-core/pull/6347#pullrequestreview-1223688719_

emmyoop:
resolves #6460 

### Description

Standardized exception classes to end in `Error` to meet PEP recommendations [here](https://peps.python.org/pep-0008/#exception-names).

### Checklist


colin-rogers-dbt:
do people rely on these exceptions?  Is it worth making this change just to comply with pep style guide?

gshank (approved):
Wow. So many lines changed... But consistency is a good thing.

colin-rogers-dbt (changes requested):
Do we have a sense on the broader impact this will have? Could this break services invoking us programmatically?

emmyoop:
@colin-rogers-dbt and i had a conversation offline and decided to remove the deprecations on the classes.  Leaving them deprecated has the potential to introduce bugs into adapters if they are catching specific exceptions.  They will never catch the old exception, but because the class still exists they will also not break.  While removing the old classes entirely will break anyone referring to the old exception classes, it will also prevent any silent bugs.

colin-rogers-dbt (approved):

elyobo:
Some are certainly in use, e.g. see https://github.com/sqlfluff/sqlfluff/issues/4325.

Removing deprecated stuff seems reasonable, but it does seem like something you'd want to do in a major release.

For others running into this, it seems like 1.3.x is the last before [this breaking change](https://github.com/dbt-labs/dbt-core/blob/1.4.latest/CHANGELOG.md#breaking-changes).

Issue #6350, opened 2022-12-01
[Spike] [CT-1575] Improve Console Log Capture Mechanism


peterallenwebb:
We substantially revised the logging system for 1.4, but it still includes an unpleasant hack to allow test code to capture and examine what is being logged to stdout. We should remove the hack, and provide an official API to capture dbt logging during a run. The result of this spike should be a one-pager describing the technical implementation of that API, confirmed with the Execution team.

jtcohen6:
@peterallenwebb Just to confirm, this is the "unpleasant hack" you mean?

https://github.com/dbt-labs/dbt-core/blob/b1d597109f6cf195077ca29b848c69725c40f412/core/dbt/tests/util.py#L106-L121

peterallenwebb:
@jtcohen6 Sorry for the late reply. Yes, this and also ad-hoc use of a global in functions.py:

https://github.com/dbt-labs/dbt-core/blob/a2213abbc0ba932ab164160ce98f8b2c63e74d5c/core/dbt/events/functions.py#L45-L49

PR #6541, opened 2023-01-07
Ct 738/dbt debug log fix

### LINKED ISSUE #5353: [CT-738] [Enhancement] dbt debug command not outputting logs correctly

### Current Behavior

When running the command `dbt debug` the logs are not output correctly. When the command is run with json log formatting turned on the log entries are still printed to the console in the same textual way. It also appears they are not output to the logs/dbt.log file either. When `--debug` is turned on alongside the json formatter then only the debug level logs are output using the json format. 

### Expected Behavior

It would be good to have the loglines of the debug command output using the json formatter for all levels of logs. We use `dbt debug` as an pre run test to check for any connections issues. We would like to be able to parse the logs for any lines with `'level': 'error'` across our entire process, but this connection test doesn't get captured because of the lack of json logging.

### Steps To Reproduce

Simply run `dbt --log-format=json debug` in any situation. With `--debug` included the difference becomes even more apparent.

### Relevant log output

```shell
I've removed some pieces/lines which contain sensitive information.


{"code": "A001", "data": {"v": "=1.1.0"}, "invocation_id": "27f7d084-0279-4455-8a5e-b937004b5930", "level": "info", "log_version": 2, "msg": "Running with dbt=1.1.0", "pid": 13, "thread_name": "MainThread", "ts": "2022-06-09T13:58:50.708619Z", "type": "log_line"}
{"code": "A003", "data": {"user_state": "tracking"}, "invocation_id": "27f7d084-0279-4455-8a5e-b937004b5930", "level": "debug", "log_version": 2, "msg": "Tracking: tracking", "pid": 13, "thread_name": "MainThread", "ts": "2022-06-09T13:58:50.711659Z", "type": "log_line"}
{"code": "Z040", "data": {"kwargs": "{'category': 'dbt', 'action': 'invocation', 'label': 'start', 'context': [<snowplow_tracker.self_describing_json.SelfDescribingJson object at 0x7fb85d827e50>, <snowplow_tracker.self_describing_json.SelfDescribingJson object at 0x7fb85d827e80>, <snowplow_tracker.self_describing_json.SelfDescribingJson object at 0x7fb85d8f1fa0>]}"}, "invocation_id": "27f7d084-0279-4455-8a5e-b937004b5930", "level": "debug", "log_version": 2, "msg": "Sending event: {'category': 'dbt', 'action': 'invocation', 'label': 'start', 'context': [<snowplow_tracker.self_describing_json.SelfDescribingJson object at 0x7fb85d827e50>, <snowplow_tracker.self_describing_json.SelfDescribingJson object at 0x7fb85d827e80>, <snowplow_tracker.self_describing_json.SelfDescribingJson object at 0x7fb85d8f1fa0>]}", "pid": 13, "thread_name": "MainThread", "ts": "2022-06-09T13:58:50.713623Z", "type": "log_line"}
dbt version: 1.1.0
python version: 3.8.13
python path: /usr/local/bin/python
os info: Linux-5.10.104-linuxkit-x86_64-with-glibc2.2.5
Using profiles.yml file at /root/.dbt/profiles.yml
Using dbt_project.yml file at /usr/app/dbt_dir/dbt_project.yml

Configuration:
  dbt_project.yml file [OK found and valid]

Required dependencies:
{"code": "Z006", "data": {"cmd": ["git", "--help"]}, "invocation_id": "27f7d084-0279-4455-8a5e-b937004b5930", "level": "debug", "log_version": 2, "msg": "Executing \"git --help\"", "pid": 13, "thread_name": "MainThread", "ts": "2022-06-09T13:58:50.827315Z", "type": "log_line"}
{"code": "Z008", "data": {"bmsg": ""}, "invocation_id": "27f7d084-0279-4455-8a5e-b937004b5930", "level": "debug", "log_version": 2, "msg": "STDERR: \"b''\"", "pid": 13, "thread_name": "MainThread", "ts": "2022-06-09T13:58:50.836976Z", "type": "log_line"}
 - git [OK found]

Connection:
  host: 
  port: 443
  cluster: 
  endpoint: None
  schema: 
  organization: 0
{"code": "E005", "data": {"conn_name": "debug", "conn_type": "spark"}, "invocation_id": "27f7d084-0279-4455-8a5e-b937004b5930", "level": "debug", "log_version": 2, "msg": "Acquiring new spark connection \"debug\"", "pid": 13, "thread_name": "MainThread", "ts": "2022-06-09T13:58:50.845000Z", "type": "log_line"}
{"code": "E015", "data": {"conn_name": "debug", "conn_type": "spark"}, "invocation_id": "27f7d084-0279-4455-8a5e-b937004b5930", "level": "debug", "log_version": 2, "msg": "Using spark connection \"debug\"", "pid": 13, "thread_name": "MainThread", "ts": "2022-06-09T13:58:50.846550Z", "type": "log_line"}
{"code": "E016", "data": {"conn_name": "debug", "sql": "select 1 as id"}, "invocation_id": "27f7d084-0279-4455-8a5e-b937004b5930", "level": "debug", "log_version": 2, "msg": "On debug: select 1 as id", "pid": 13, "thread_name": "MainThread", "ts": "2022-06-09T13:58:50.847964Z", "type": "log_line"}
{"code": "E037", "data": {"connection_state": "init"}, "invocation_id": "27f7d084-0279-4455-8a5e-b937004b5930", "level": "debug", "log_version": 2, "msg": "Opening a new connection, currently in state init", "pid": 13, "thread_name": "MainThread", "ts": "2022-06-09T13:58:50.849569Z", "type": "log_line"}
{"code": "E001", "data": {"args": [], "base_msg": "Error while running:\nselect 1 as id", "name": "Spark"}, "invocation_id": "27f7d084-0279-4455-8a5e-b937004b5930", "level": "debug", "log_version": 2, "msg": "Spark adapter: Error while running:\nselect 1 as id", "pid": 13, "thread_name": "MainThread", "ts": "2022-06-09T14:00:54.908987Z", "type": "log_line"}
  Connection test: [ERROR]

1 check failed:
dbt was unable to connect to the specified database.
The database returned the following error:

  >Object of type FailedToConnectException is not JSON serializable

Check your database credentials and try again. For more information, visit:
https://docs.getdbt.com/docs/configure-your-profile


{"code": "Z040", "data": {"kwargs": "{'category': 'dbt', 'action': 'invocation', 'label': 'end', 'context': [<snowplow_tracker.self_describing_json.SelfDescribingJson object at 0x7fb85d6462b0>, <snowplow_tracker.self_describing_json.SelfDescribingJson object at 0x7fb85d62da60>, <snowplow_tracker.self_describing_json.SelfDescribingJson object at 0x7fb85d62dd90>]}"}, "invocation_id": "27f7d084-0279-4455-8a5e-b937004b5930", "level": "debug", "log_version": 2, "msg": "Sending event: {'category': 'dbt', 'action': 'invocation', 'label': 'end', 'context': [<snowplow_tracker.self_describing_json.SelfDescribingJson object at 0x7fb85d6462b0>, <snowplow_tracker.self_describing_json.SelfDescribingJson object at 0x7fb85d62da60>, <snowplow_tracker.self_describing_json.SelfDescribingJson object at 0x7fb85d62dd90>]}", "pid": 13, "thread_name": "MainThread", "ts": "2022-06-09T14:00:54.913884Z", "type": "log_line"}
{"code": "Z042", "data": {}, "invocation_id": "27f7d084-0279-4455-8a5e-b937004b5930", "level": "debug", "log_version": 2, "msg": "Flushing usage events", "pid": 13, "thread_name": "MainThread", "ts": "2022-06-09T14:00:54.914749Z", "type": "log_line"}
{"code": "E008", "data": {"conn_name": "debug"}, "invocation_id": "27f7d084-0279-4455-8a5e-b937004b5930", "level": "debug", "log_version": 2, "msg": "Connection 'debug' was properly closed.", "pid": 13, "thread_name": "MainThread", "ts": "2022-06-09T14:00:55.472184Z", "type": "log_line"}
```

### Environment

```markdown
- OS:Linux-5.10.104-linuxkit-x86_64-with-glibc2.2.5
- Python: 3.8.13
- dbt:1.1.0
```

### Additional Context

databricks

VersusFacit:
resolves #5353 

> The world is a ~~book~~ repo and those who do not travel read only a ~~page~~ file. ~Augustine...or something

And what a fun, instructive excursion this was. I got to edit some CLI jobs and walk around some execution parts of the repo -- how exciting 😊 

#### `dbt debug` structured logging
1. `dbt --log-format json debug` and `DBT_LOG_FORMAT=json dbt debug` both now render as structured logs
    - they were only doing so piecemeal before this PR
2. there's not a single `print` left in the file -- sweet!
3. tested for a working case (my own trusty snowflake connection) -- screenshots omitted
4. tested for a failing class (profile missing) -- screen shots below 👇 
 
#### `dbt {ls,list}` structured logging
5. `dbt --log json {ls,list}` and `DBT_LOG_FORMAT=json dbt {ls,list}` prints structured logs now, under `info` level
6. `dbt {ls,list}` continues to print just the names (🔪 -ed out some overrides and a `TODO` or two)
 
### Before and After Screenshots

My debugging strategy:

```sh
clear; dbt debug; echo "======"; dbt --log-format json debug
```
Check evolving structured logs against plain ol' `dbt debug`.

Before, several printed lines:
Screen Shot 2023-01-07 at 12 18 22 AM

After, they're all structured logs now:
Screen Shot 2023-01-07 at 12 29 41 AM

Minor caveat: These were taken in two different iterations so the connections listed are slightly different, but I just have lots of profiles :p

### Outstanding Questions
1. I generated my protobuf types with `protoc -I. --python_betterproto_out=. types.proto`. Hope that's right 🙇‍♀️ 
2. Jeremy over on the issue thread might suggest a different spec than what I served here.

### Checklist


core/dbt/events/types.proto
```diff
@@ -1289,6 +1289,85 @@ message NoNodesSelected {
     EventInfo info = 1;
 }
 
+// Q039
+message DebugEnvironmentDetails {
```
VersusFacit, line 1293: So I added a lot of these.

1. Jerco considered differentiating between [Successes and Failures](https://github.com/dbt-labs/dbt-core/issues/5353#issuecomment-1152299423). I erred on that side.

2. I tacked these onto Q types, but Node Execution is sort of a bit loose (may even be erroneous). My considered alternative was to split this off into, like, a "Task Execution" subcategory 🤷‍♀️ Figured I'd check with reviewers.

core/dbt/events/types.proto
```diff
@@ -1289,6 +1289,85 @@ message NoNodesSelected {
...
+    EventInfo info = 1;
+    string msg = 2;
```
VersusFacit, line 1295: All of these are `msg` because that's really all I'm passing to the exceptions: various messages. The messages are so flexible that trying to define messages in `def message` functions felt "wrong" on gut reaction. Happy to adjust!

core/dbt/events/types.py
```diff
@@ -2039,6 +2039,123 @@ def message(self) -> str:
         return "No nodes selected!"
 
 
+@dataclass
+class DebugEnvironmentDetails(InfoLevel, pt.DebugEnvironmentDetails):
+    def code(self):
+        return "Q039"
+
+    def message(self) -> str:
```
VersusFacit, line 2047: see above comment

core/dbt/tests/util.py
```diff
@@ -86,22 +89,27 @@ def run_dbt(args: List[str] = None, expect_pass=True):
...
 # will turn the logs into json, so you have to be prepared for that.
-def run_dbt_and_capture(args: List[str] = None, expect_pass=True):
+def run_dbt_and_capture(args: List[str] = None, expect_pass=True, capture_stderr=False):
     try:
         stringbuf = StringIO()
         capture_stdout_logs(stringbuf)
-        res = run_dbt(args, expect_pass=expect_pass)
-        stdout = stringbuf.getvalue()
+        if capture_stderr:
+            with redirect_stdout(stringbuf):
```
VersusFacit, line 97: Lots of `dbt debug` wasn't being piped out to `stdout` for whatever reason. I think we need this trick in our toolkit for future use cases.

core/dbt/events/types.proto
```diff
@@ -1585,6 +1586,85 @@ message RunResultWarningMessage {
     string msg = 2;
 }
 
+// Z047
+message DebugEnvironmentDetails {
```
VersusFacit, line 1590: So I added a lot of these.

1. Jerco considered differentiating between [Successes and Failures](https://github.com/dbt-labs/dbt-core/issues/5353#issuecomment-1152299423). I erred on that side.

2. I tacked these onto Q types, but Node Execution is sort of a bit loose (may even be erroneous). My considered alternative was to split this off into, like, a "Task Execution" subcategory 🤷‍♀️ Figured I'd check with reviewers.

core/dbt/events/types.proto
```diff
@@ -1585,6 +1586,85 @@ message RunResultWarningMessage {
...
+    EventInfo info = 1;
+    string msg = 2;
```
VersusFacit, line 1592: All of these are msg because that's really all I'm passing to the exceptions: various messages. The messages are so flexible that trying to define messages in def message functions felt "wrong" on gut reaction. Happy to adjust!

core/dbt/task/list.py
```diff
@@ -148,9 +139,14 @@ def run(self):
...
     def output_results(self, results):
+        """Log, or output a plain, newline-delimited, and ready-to-pipe list of nodes found."""
         for result in results:
             self.node_results.append(result)
-            print(result)
+            if get_flag_dict()["log_format"] == "json":
+                fire_event(ListRunDetails(msg=result))
+            else:
+                # Cleaner to leave as print than to mutate the logger.
+                print(result)
```
jtcohen6, line 149: This feels like the right idea. I'm not sure if it's the exact implementation we'll want, versus swapping this out for a more generic `clients.system.print` or `events.functions.print` method that handles this conditional logic for us.

I opened an issue to encapsulate the larger need here: https://github.com/dbt-labs/dbt-core/issues/6543

core/dbt/tests/util.py
jtcohen6: Potentially relevant: https://github.com/dbt-labs/dbt-core/issues/6350

dbeatty10 (commented):
@VersusFacit The branch for the PR has conflicts to be resolved. Could you resolve those?

core/dbt/tests/util.py
dbeatty10: I think I'm the one that added [`redirect_stdout`](https://docs.python.org/3/library/contextlib.html#contextlib.redirect_stdout) here originally.

If I remember correctly, it was _only_ added in order to capture the print statements from `dbt debug` for the TODO test (see my other PR comment below on that test).

If all the print statements are removed, we might not need the `capture_stderr` parameter and the associated logic. We'd have to check one way or the other.

If we can remove it altogether to simplify, then great!

If we keep it, I'd suggest renaming `capture_stderr` to `redirect_stdout` (or some other alternative) -- the former feels like a misnomer.

tests/functional/profiles/test_profile_dir.py
```diff
@@ -106,8 +100,8 @@ def dbt_debug(self, project_dir_cli_arg=None, profiles_dir_cli_arg=None):
         if profiles_dir_cli_arg:
             command.extend(["--profiles-dir", str(profiles_dir_cli_arg)])
 
-        # get the output of `dbt debug` regarless of the exit code
-        return run_dbt_and_capture_stdout(command, expect_pass=None)
+        # get the output of `dbt debug` regardless of the exit code
+        return run_dbt_and_capture(command, expect_pass=False, capture_stderr=True)
```
dbeatty10, line 104: I'm the one who originally wrote this section of code.

Good catch on my misspelling of "regardless"! 🙏 

I didn't confirm this one way or the other, but my naive expectation is that we would preserve `expect_pass=None`. This is because this test intends to inspect the output of the `dbt debug` command _regardless_ if the exit code is zero or non-zero (corresponding to PASS vs FAIL, respectively).

Put another way, `dbt debug` will pass in some test scenarios and fail in others, but in both cases, we don't care about the pass/fail status; we only care about the text in stdout. If we uniformly expect_pass=False or expect_pass=True, it wouldn't allow us to reach all the test cases in the desired manner.

```suggestion
        # get the output of `dbt debug` regardless of the exit code
        return run_dbt_and_capture(command, expect_pass=None, capture_stderr=True)
```

It might also be that the refactor of `print` statements in `dbt debug` allows and/or requires us to refactor this test as well. If so, we'll just want to carefully consider things like `expect_pass=False`.

dbeatty10 (changes requested):

core/dbt/compilation.py
```diff
@@ -482,7 +483,13 @@ def compile(self, manifest: Manifest, write=True, add_test_edges=False) -> Graph
...
         if write:
             self.write_graph_file(linker, manifest)
-        print_compile_stats(stats)
+
+        # Only in the case the object is specifically a ListTask do we avoid printing these stats.
+        try:
+            if not isinstance(self.config.args.get("cls", None), list_task.ListTask):
+                print_compile_stats(stats)
+        except AttributeError:
+            print_compile_stats(stats)
```
gshank, line 492: It's not clear to me what you're trying to do here. If the "try" of doing print_compile_stats fails you do it again?

core/dbt/events/proto_types.py
```diff
@@ -2016,6 +2016,118 @@ class RunResultWarningMessage(betterproto.Message):
     msg: str = betterproto.string_field(2)
 
 
+@dataclass
+class DebugEnvironmentDetails(betterproto.Message):
```
gshank, line 2020: We have been going through an exercise of combining various logging events. These new logging events are pretty much identical except for the names and codes. I think it would be better to combine some of these messages. I would personally prefer a sort of "DebugCmdResult" message. Or maybe some of those could be put in a DebugCmdOut event.  The touchstone is to think of what would be most convenient for someone who's looking for particular information in the logs.

core/dbt/events/types.proto
gshank: I think Z is fine for the debug command messages.

core/dbt/task/list.py
gshank: Normally I would expect to see somebody use flags.LOG_FORMAT to get the flag setting. Though I'm not sure what's happening with the flags object in the about-to-me merged click branch.

Also I'm not sure what the reference to having to mutate the logger is about. What issue were you avoiding here?

core/dbt/tests/util.py
gshank: I would rather that we didn't add an additional argument to run_dbt_and_capture if we can avoid it.

gshank (changes requested):
This PR needs to have main merged because there have been a lot of changes in the events code and the messages as implemented in types.proto here will no longer work.

Rather than a bunch of new logging events I would prefer to have them combined as much as makes sense. We are going through an exercise of reducing the number of events and this is going in the opposite direction.

core/dbt/task/list.py
VersusFacit: re: the flags question.

For sure! I tried that aaaaaand it doesn't work here. 😕 For whatever reason, under certain combinations of passing in the `json` type, the LOG_FORMAT object is None. Maybe that's a deeper bug with `flags.py`?

core/dbt/compilation.py
VersusFacit: Read another way: only print compile stats if the task is ListTask. If the task object is None (sometimes this happens), the `get` will fail. In that case, still print the stats.

core/dbt/tests/util.py
VersusFacit: As I recall, I had to override this to capture the output because before other fixes in this PR, the `debug` output was being kicked out to err as warnings. I'll double check if that's still relevant now because this change may no longer be needed (which would be nice)

core/dbt/compilation.py
gshank: I don't think the try/except serves any purpose here.  Wouldn't the "if not... / print_compile_stats" be enough?

core/dbt/task/list.py
gshank: Using "get_flag_dict" should be no different, since it just copies the value of LOG_FORMAT into the dictionary. Use flags.LOG_FORMAT and if you get a None we should figure out why.

core/dbt/compilation.py
VersusFacit: We need the try catch guard between the `if not` can throw an `AttributeError` if a `self.config.args` equal to `None` is passed in, which happens numerous times through execution. I'll double check though :)

core/dbt/task/list.py
VersusFacit: I'll for sure double check this, but at the time I did this work, the values I was getting were different for whatever reason. It could have been due to logger mutations that no longer exist or something else. Rest assured, I'll be sure to asses.

tests/functional/profiles/test_profile_dir.py
VersusFacit: This is interesting. I'll have to double check this. My thinking was "None" and "False" are evaluated the same way in Python `if`-`else` blocks. I'll dive a bit deeper into the code here to ensure there isn't something else going on behind the scenes.

core/dbt/compilation.py
VersusFacit: Just double checked and if I remove the try-catch, this is what happens:
image

If I have the try-catch in place, that behavior doesn't happen
image

There could be a more elegant way to observe this is a ListTask but with the `self.config.args` object, we have to worry about `None` for `args`.

core/dbt/tests/util.py
VersusFacit: Checked it! This is _no longer needed_ 🎉 It was needed when I started (before I changed the text of `dbt list` to no longer be a warning, which gets piped to `stderr`). Great catch

tests/functional/profiles/test_profile_dir.py
VersusFacit: Oh I didn't know we had this feature. This is great and there were several times I wish I had known about this sooner! 
```python
if expect_pass is not None:
    assert success == expect_pass, "dbt exit state did not match expected"
```
Thanks for correcting me. I'll fix this back.

core/dbt/compilation.py
VersusFacit: I went through all this only to find out the underlying object has changed to a `Namespace`. Pardon me, under this setup, what you're saying is right. My changes were developed before merging main. I'll fix this up 👍

core/dbt/task/list.py
VersusFacit: Oh good news, some upstream change on merge fixed this for me. I'll simplify as you suggest 👏

core/dbt/events/proto_types.py
VersusFacit: Did my best with this!

gshank (approved):
Looks good!

tests/functional/profiles/test_profile_dir.py
```diff
@@ -84,18 +88,8 @@ def environ(env):
...
-    with redirect_stdout(stringbuf):
-        res = run_dbt(args, expect_pass=expect_pass)
-    stdout = stringbuf.getvalue()
-
-    return res, stdout
-
-
 class TestProfiles:
-    def dbt_debug(self, project_dir_cli_arg=None, profiles_dir_cli_arg=None):
+    def dbt_debug(self, project, project_dir_cli_arg=None, profiles_dir_cli_arg=None):
```
dbeatty10, line 92: Excited about all the simplification you were able to do for the `tests/functional/profiles/test_profile_dir.py` test!

Is adding `project` essential? If not, what do you think about leaving it as the following?
```suggestion
    def dbt_debug(self, project_dir_cli_arg=None, profiles_dir_cli_arg=None):
```

dbeatty10 (dismissed):

tests/functional/profiles/test_profile_dir.py
VersusFacit: @dbeatty10 It is because you need the profile for the run command to work with run_dbt or run_dbt_and_capture

VersusFacit: Yikes, I was wrong. This _was_ needed at some point, but now I can go in and remove it. Thanks for the callout 🙇‍♀️ 😅 I need to dive in the next time I find the error I described and I identify it's a context-specific bug or something. I've run into it too many times for it to be fake, but also, my understanding of it has now been shifting as a result of this convo.

tl;dr merged your change

dbeatty10: No worries!

Will be a very happy day once this is merged and my custom `run_dbt_and_capture_stdout` is no longer a thing!

Will re-review shortly.

dbeatty10 (commented):
I _only_ reviewed the changes to `tests/functional/profiles/test_profile_dir.py`, and they look great!

Relying on other reviewers to cover everything else.

PR #6432, opened 2022-12-13
[CT-1284] Change Python model default materialization to table

aranke:
Resolves #5989 

### Description

* Change Python model default materialization to table and add tests
* Refactor tests to read from Python source files instead of storing as an inline string
* Clean up imports

### Checklist

- [x] I have read [the contributing guide](https://github.com/dbt-labs/dbt-core/blob/main/CONTRIBUTING.md) and understand what's expected of me
- [x] I have signed the [CLA](https://docs.getdbt.com/docs/contributor-license-agreements)
- [x] I have run this code in development, and it appears to resolve the stated issue
- [x] This PR includes tests, or tests are not required/relevant for this PR
- [x] I have [opened an issue to add/update docs](https://github.com/dbt-labs/docs.getdbt.com/issues/new/choose), or docs changes are not required/relevant for this PR
- [x] I have run `changie new` to [create a changelog entry](https://github.com/dbt-labs/dbt-core/blob/main/CONTRIBUTING.md#adding-a-changelog-entry)

Issue #5989, opened 2022-10-03
[Bug] default dbt-py models to view materialization


lostmygithubaccount:
### Is this a new bug in dbt-core?

- [X] I believe this is a new bug in dbt-core
- [X] I have searched the existing issues, and I could not find an existing issue for this bug

### Current Behavior

it's a bit unclear whether the default is truly `view` (reproduction below + comment in Slack thread indicates it is), or people have configured that elsewhere (e.g. `dbt_project.yml`) and are running into this. regardless, we should ensure `table` is the default

### Expected Behavior

a dbt-py model with no materialization configuration defaults to `table`

### Steps To Reproduce

```bash
pip install --pre dbt-core dbt-snowflake
mkdir test
cd test
dbt init
# use project name "testing"
cd testing
vi models/test.py
# python code
def model(dbt, session):
    return session.sql("select 1 as id")
# end python code
dbt run -s test
```

this results in the views not supported for Python language error. the `dbt_project.yml` only configures the default for the `models/example` directory and should not affect `test.py` above

### Relevant log output

```shell
(test) @lostmygithubaccount ➜ /workspaces/codyspace/test/testing (main ✗) $ dbt run -s test 
15:53:39  Running with dbt=1.3.0-rc2
15:53:39  Partial parse save file not found. Starting full parse.
15:53:40  Found 3 models, 4 tests, 0 snapshots, 0 analyses, 301 macros, 0 operations, 0 seed files, 0 sources, 0 exposures, 0 metrics
15:53:40  
Initiating login request with your identity provider. A browser window should have opened for you to complete the login. If you can't see it, check existing browser windows, or your OS settings. Press CTRL+C to abort and try again...
15:53:44  Concurrency: 8 threads (target='dev')
15:53:44  
15:53:44  1 of 1 START python view model dbt_cody.test ................................... [RUN]
15:53:45  1 of 1 ERROR creating python view model dbt_cody.test .......................... [ERROR in 0.02s]
15:53:45  
15:53:45  Finished running 1 view model in 0 hours 0 minutes and 4.32 seconds (4.32s).
15:53:45  
15:53:45  Completed with 1 error and 0 warnings:
15:53:45  
15:53:45  Runtime Error in model test (models/test.py)
15:53:45    Materialization "materialization_view_snowflake" only supports languages ['sql']; got "python"
15:53:45  
15:53:45  Done. PASS=0 WARN=0 ERROR=1 SKIP=0 TOTAL=1
```

[...]

PR #6560, opened 2023-01-10
Fix changelog entry for 5989/6432

jtcohen6:
This also happened in https://github.com/dbt-labs/dbt-core/pull/6496. I haven't been doing any exhaustive checking, just some spot-checking here or there while working on docs for v1.4.

Now that we're only linking to the issue, rather than both the issue + the PR—if we get one link wrong, it's the _only_ link we're showing in the changelog / release notes. It's not a huge deal, but it could be misleading for some community members.

Is there any automation that could help us here? Or just an additional manual check during review? E.g. https://github.com/dbt-labs/dbt-core/pull/6432 was linked to #5989 in the PR description (+ therefore also GitHub "Development" flow).


dbeatty10 (approved):
Glad you caught this @jtcohen6 !

Approving now and will merge immediately thereafter.

Issue #6546, opened 2023-01-08
[CT-1766] [DX] Review command + param help text in CLI / top-level API


jtcohen6:
Let's give the "help" text defined below another review from a DX (Developer Experience) lens:

- https://github.com/dbt-labs/dbt-core/blob/main/core/dbt/cli/main.py
- https://github.com/dbt-labs/dbt-core/blob/main/core/dbt/cli/params.py

Anything marked `TODO` is especially in need, but I think all help text deserves another look-over.

This is also our opportunity to remove params / help text that are completely unused or irrelevant. For example, [`dbt compile --parse-only`](https://github.com/dbt-labs/dbt-core/blob/94d6d19fb43f1ed466c9049dea40e203b23f89da/core/dbt/cli/params.py#L165-L170) doesn't do anything, and I don't believe it ever has.

### Context

This is what users will see when running `dbt --help` (invoked as module for now):
```bash
$ python3 -m core.dbt.cli.main --help
Usage: python -m core.dbt.cli.main [OPTIONS] COMMAND [ARGS]...

  An ELT tool for managing your SQL transformations and data models. For more
  documentation on these commands, visit: docs.getdbt.com

Options:
  --anonymous-usage-stats / --no-anonymous-usage-stats
                                  Send anonymous usage stats to dbt Labs.
  --cache-selected-only / --no-cache-selected-only
                                  Pre cache database objects relevant to
                                  selected resource only.
  -d, --debug / --no-debug        Display debug logging during dbt execution.
                                  Useful for debugging and making bug reports.
```

And it's also what they'll see in our auto-generated "API docs."

<img width="666" alt="Screenshot 2023-01-08 at 18 19 28" src="https://user-images.githubusercontent.com/13897643/211209857-9def46e5-e934-477f-8480-4f57a9c474ea.png">

jtcohen6:
@dbeatty10 Let's figure out a sprint where you have capacity to pick this up & drive this forward, between now and mid-April (RC release)

jtcohen6:
Might be nice: using the `click-option-group` extension to group together related commands in the `--help` output (and in the codebase).
- https://click-option-group.readthedocs.io/en/stable/index.html

Not necessary for the scope for this issue, only if inspiration strikes while looking at existing params.

PR #2712, opened 2020-08-18
Feature: adapter cte generation

beckjake:
resolves #2660 (well, makes it feasible)
resolves #2609 

### Description
This PR makes it possible for the exasol plugin (or any others) to override their `Relation` class with one that has a custom `add_ephemeral_prefix` method. For the basic case, this should solve cte-naming issues.

The adapter class is further modified: It now provides a `get_compiler` method that returns a `Compiler` instance, so plugins can provide plugin-specific compilation changes.

In particular, it should now be possible to override how dbt creates ephemeral models. No promises, of course, but this should make it feasible for adapters that support subqueries but not CTEs to make their own "ephemeral" models. This is definitely left as an exercise to the plugin author, but overriding the handy `_insert_ctes` method should help. Note that `_insert_ctes` also does special work for making data tests work, so authors will need to re-implement that. Which is because...

This PR also makes data tests behave _much_ more like CTEs. Instead of performing a subquery, dbt inserts a custom CTE at the end of the CTE chain that includes the test body, and then the final select clause is a `select count(*) from` the CTE. Adapters that need to override CTE generation will need to do this as well, but I think that's ok. My understanding is that this will make #2609  work as it should out of the box.

Most adapters should not need to change anything. Any adapters that don't support CTEs but do support subqueries might break here. We should be able tof come up with a backwards-compatibility behavior here if we need to, and make only our adapters use this behavior.

Finally, while I was working on this I found that we were not properly excluding some things from the sandbox. This isn't a huge deal, but as a matter of hygiene we should be prefixing more attributes of objects in the context with `_` so jinja won't allow access.

### Checklist
 - [X] I have signed the [CLA](https://docs.getdbt.com/docs/contributor-license-agreements)
 - [X] I have run this code in development and it appears to resolve the stated issue
 - [X] This PR includes tests, or tests are not required/relevant for this PR
 - [X] I have updated the `CHANGELOG.md` and added information about my change to the "dbt next" section.

PR #6565, opened 2023-01-10
Abstract manifest generation from tasks

### LINKED ISSUE #6357: [CT-1582] [Feature] Refactor tasks to breakout parsing the manifest into a separate piece

### Describe the feature

Right now we are doing the parsing and generating a manifest object as part of the `runtime_initialization` under task. 
This makes this important step of `dbt` hidden under multiple layer of inheritance.
This also makes use need to do hacky solutions to fit the usecases of dbt-server where we want to have some endpoint do parsing upon files being modified, and saves some kind of manifest object for faster dbt invocation.

In order to resolve this, we should refactor out the [`ManifestTask`](https://github.com/dbt-labs/dbt-core/blob/main/core/dbt/task/runnable.py#L66) out of the inheritance chain of tasks into it's own module. Then we will initialize tasks that need parsing of the project with a constructed manifest that generated by the refactored out Manifest loader.

One thing tbd is where we want the [compile step](https://github.com/dbt-labs/dbt-core/blob/main/core/dbt/task/runnable.py#L91) to happen after this refactor. (Stu: created #6708 to address abstract graph generation as well)

stu-k:
resolves #6357

### Description

Abstract manifest generation from directly inside the task classes to their instantiation.

### Checklist


core/dbt/cli/main.py
```diff
@@ -453,6 +469,30 @@ def test(ctx, **kwargs):
...
 
+def _get_config_from_ctx(ctx):
+    return RuntimeConfig.from_parts(
+        ctx.obj["project"],
+        ctx.obj["profile"],
+        ctx.obj["flags"],
+    )
+
+
+def _set_manifest_on_task(task, ctx, write_perf_info=False):
```
ChenyuLInx, line 480: Why do we choose this over passing in manifest as an object during initialization of task

core/dbt/task/base.py
```diff
@@ -45,10 +45,11 @@
...
 
-from dbt.adapters.factory import register_adapter
+from dbt.adapters.factory import get_adapter, register_adapter
 from dbt.config import RuntimeConfig, Project
 from dbt.config.profile import read_profile
 import dbt.exceptions
+from dbt.graph import Graph
 
 
 class NoneConfig:
```
ChenyuLInx, line 55: Do we plan to keep this in longer term? Or just something to help during the transition period?

core/dbt/task/compile.py
```diff
@@ -85,4 +86,4 @@ def defer_to_manifest(self, adapter, selected_uids: AbstractSet[str]):
             selected=selected_uids,
         )
         # TODO: is it wrong to write the manifest here? I think it's right...
-        self.write_manifest()
+        write_manifest(self.manifest, self.config.target_path)
```
ChenyuLInx, line 89: I remember in changes above we write manifest in the click command, any reason we do this here not following the same convention?

core/dbt/task/runnable.py
```diff
@@ -490,7 +449,12 @@ def run(self):
             )
 
         if flags.WRITE_JSON:
-            self.write_manifest()
+            write_manifest(
```
ChenyuLInx, line 452: same question here

core/dbt/cli/main.py
```diff
@@ -453,6 +469,30 @@ def test(ctx, **kwargs):
...
+    config = _get_config_from_ctx(ctx)
+    manifest = ManifestLoader.get_full_manifest(config, write_perf_info=write_perf_info)
```
ChenyuLInx, line 482: Thoughts on creating a decorator `@requires.manifest` and do the load of manifest there?

ChenyuLInx (changes requested):
Overall looks good!

core/dbt/cli/main.py
ChenyuLInx, line 482: `register_adapter(runtimeConfig)` need to be called before/within get_full_manifest

core/dbt/task/base.py
stu-k: Compilation needed to live somewhere, and since the ask of this ticket is only abstracting manifest generation, not compilation, I moved the behavior to `ConfiguredTask`. If we want to move this behavior elsewhere, we can do so in the future.

core/dbt/task/compile.py
stu-k: This is to maintain parity with existing functionality. I am not well versed in why writing the manifest happens when it does, however I didn't want to change anything currently happening, only abstract manifest generation.

core/dbt/cli/requires.py
```diff
@@ -85,3 +87,64 @@ def wrapper(*args, **kwargs):
...
+
+            return func(*args, **kwargs)
+
+        return update_wrapper(wrapper, func)
+
+    # if there are no args, the decorator was used without params @decorator
+    # otherwise, the decorator was called with params @decorator(arg)
+    if len(args0) == 0:
+        return outer_wrapper
+    return outer_wrapper(args0[0])
```
stu-k, line 150: Is there a better more pythonic way to have a decorator accept args?

core/dbt/cli/main.py
stu-k: I initial thought there wasn't much of a difference in either approach, but after speaking with @MichelleArk I believe setting the manifest directly on task init is favorable given users may attempt to set a different manifest after a run, which may have unknown consequences given how stateful our task classes are.

ChenyuLInx: Makes sense! Yes I agree seems like re-create the whole task would be a cleaner interface than setting things after it being constructed!

core/dbt/cli/requires.py
```diff
@@ -85,3 +87,67 @@ def wrapper(*args, **kwargs):
...
+            ctx.obj["profile"],
+            ctx.obj["flags"],
+        )
+
+        return func(*args, **kwargs)
+
+    return update_wrapper(wrapper, func)
+
+
+def manifest(*args0, write_perf_info=False):
```
ChenyuLInx, line 118: I think we would want to do something similar to what we did at line 55 of this file to allow people pass in a pre constructed manifest and skip the loading part, but that could be a separate issue we tackle later on.

CC @MichelleArk  thoughts?

core/dbt/task/base.py
```diff
@@ -167,12 +168,31 @@ def move_to_nearest_project_dir(project_dir: Optional[str]) -> str:
...
+        self.manifest = manifest
+
+    def compile_manifest(self):
+        if self.manifest is None:
+            raise InternalException("compile_manifest called before manifest was loaded")
+
+        start_compile_manifest = time.perf_counter()
+
+        # we cannot get adapter in init since it will break rpc #5579
+        adapter = get_adapter(self.config)
```
ChenyuLInx, line 189: Was this because adapter haven't been initialized at the time? I kinda forgot about the change

ChenyuLInx (approved):
Overall looks good, I was able to confirm a run works as expected locally.

core/dbt/cli/requires.py
stu-k: Yes, that makes sense. I thought this was being handled by the `dbtRunner`, but if that is not implemented in this decorator as well it will be overwritten every time.

ChenyuLInx: The dbtRunner only set it. The logic in these decorators are the ones that handles skip doing loading when things are set.

core/dbt/task/base.py
jtcohen6: Compilation requires access to the adapter. Way back in https://github.com/dbt-labs/dbt-core/pull/2712, we made it possible for adapters to:
- Change the prefix used for ephemeral model CTEs [here](https://github.com/dbt-labs/dbt-core/blob/93619a9a3761d3606d257c1274828ac4eef4f627/core/dbt/compilation.py#L187-L190), since certain data platforms (e.g. Exasol) have very particular rules about which naming conventions are supported
- [Totally reimplement the `Compiler`](https://github.com/dbt-labs/dbt-core/blob/93619a9a3761d3606d257c1274828ac4eef4f627/core/dbt/adapters/base/impl.py#L1117-L1120), from scratch—intended for cases where an adapter needs to do something fundamentally different for ephemeral models, because the data platform doesn't support nested CTEs (e.g. SQLServer / Azure Synapse). But the [`Compiler`](https://github.com/dbt-labs/dbt-core/blob/93619a9a3761d3606d257c1274828ac4eef4f627/core/dbt/compilation.py#L161) also defines methods related to DAG construction (node/edge linking), which really should _not_ vary by data platform.

Compilation is part of execution/runtime, so requiring adapter-specific inputs for certain bits of its functionality can be appropriate. But I'd say that both of these are pretty wonky behaviors today, and are worth additional consideration as we move toward a future where the adapter is no longer a global singleton.

core/dbt/cli/requires.py
```diff
@@ -85,3 +87,70 @@ def wrapper(*args, **kwargs):
...
+
+            # a manifest has already been set on the context, so don't overwrite it
+            if ctx.obj.get("manifeset") is None:
+                manifest = ManifestLoader.get_full_manifest(
+                    runtime_config, write_perf_info=write_perf_info
+                )
+
+                ctx.obj["manifest"] = manifest
+                if ctx.obj["flags"].write_json:
+                    write_manifest(manifest, ctx.obj["runtime_config"].target_path)
```
jtcohen6, line 146: This mean we always write the manifest (`manifest.json`) at the end of project parsing, unless an already-generated manifest was passed in explicitly. That makes sense to me!

jtcohen6 (changes requested):
Great work on this @stu-k! I left some comments to contextualize a few of the interrelated pieces that this PR is touching.

I called out two required changes that I'd consider blocking to merge:
- Now that we've converted more manifest-using commands, we'll need `@requires.manifest` on those commands as well
- The `build` command/task needs different behavior for `compile_manifest`

Then, I think we should open up a separate ticket to take another look at abstracting manifest compilation / graph generation from tasks.

core/dbt/cli/requires.py
stu-k: I refactored this to not change any existing writing functionality.

core/dbt/cli/main.py
```diff
@@ -120,10 +119,10 @@ def cli(ctx, **kwargs):
 @requires.preflight
 @requires.profile
 @requires.project
+@requires.runtime_config
 def build(ctx, **kwargs):
```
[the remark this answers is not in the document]
stu-k: Thank you! I had added `runtime_config` to the newer tasks but forgot to add the manifest decorator and pass it in.

core/dbt/task/build.py
```diff
@@ -63,10 +62,3 @@ def get_node_selector(self) -> ResourceTypeSelector:
 
     def get_runner_type(self, node):
         return self.RUNNER_MAP.get(node.resource_type)
-
-    def compile_manifest(self):
-        if self.manifest is None:
-            raise InternalException("compile_manifest called before manifest was loaded")
-        adapter = get_adapter(self.config)
-        compiler = adapter.get_compiler()
-        self.graph = compiler.compile(self.manifest, add_test_edges=True)
```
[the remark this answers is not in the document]
stu-k: Abstracting graph generation into a decorator was something I wanted to do for sure, thank you for catching this!

core/dbt/cli/main.py
```diff
@@ -341,9 +341,13 @@ def list(ctx, **kwargs):
...
 @requires.preflight
+@requires.profile
+@requires.project
+@requires.runtime_config
+@requires.manifest(write_perf_info=True)
 def parse(ctx, **kwargs):
     """Parses the project and provides information on performance"""
-    click.echo(f"`{inspect.stack()[0][3]}` called\n flags: {ctx.obj['flags']}")
+    # manifest generation and writing happens in @requires.manifeset
     return None, True
```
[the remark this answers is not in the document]
stu-k: Thank you for the context. I had spoken with Gerda about the function of the existing `ParseTask` class, and she had said I could take out the tracking. Adding it back in shouldn't be difficult, and could be done conditionally.

For your proposal of returning the manifest from the task directly, I think that's something we need to discuss on the exec team. I believe we should have a standard output for each of these functions, perhaps a dict with a `result` or `message` or whatever key should contain result, in case we want to add meta information to that result object later.

core/dbt/task/build.py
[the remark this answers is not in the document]
jtcohen6: > Abstracting graph generation into a decorator was something I wanted to do for sure, thank you for catching this!

Want to open a new issue for this? We can plan to tackle it in the next few months, whether as part of "API-ification" (phase 2) or "library-ification" (support for programmatic invocations / `dbt-server`).

core/dbt/cli/main.py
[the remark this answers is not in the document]
jtcohen6: > I had spoken with Gerda about the function of the existing `ParseTask` class, and she had said I could take out the tracking. Adding it back in shouldn't be difficult, and could be done conditionally.

This is fine with me too. Gerda consolidated all the events in this task, anyway, to just `ParseCmdOut` (in the `main` branch). We'll just want to remove that event now that it's no longer being called anywhere. For simplicity, let's track that in a separate issue, rather than try to do branch merging shenanigans here.

---

The point I was making above was around, `dbt parse --compile` would also conditionally include the step around manifest compilation / graph generation, which can be very slow (as we've seen in recent reports/issues from folks with large projects). Let's track that in the new issue, "Abstract manifest compilation / graph generation from tasks."

---

> For your proposal of returning the manifest from the task directly, I think that's something we need to discuss on the exec team. I believe we should have a standard output for each of these functions, perhaps a dict with a `result` or `message` or whatever key should contain result, in case we want to add meta information to that result object later.

Agreed! Let's keep discussing in the separate linked issue. The biggest considerations are:
- How to keep the `results` _relatively_ consistent for all commands (although different tasks already return differently typed result objects)
- Should we expose the full manifest, as part of our public API? Just a part of it? Should we call this "experimental" functionality (liable to change in future versions)? We have the power to decide these things!

jtcohen6 (approved):
Nice work @stu-k!

Could I ask you to open two follow-up issues that came up in conversation?
1. Remove `ParseCmdOut`, now that the `parse` command isn’t doing any of its own custom event firing / logging. (Or, conditionally add the logging back.)
2. Abstracting graph generation into a decorator, rather than a step buried within task initialization. We can plan to tackle it in the next few months, whether as part of "API-ification" (phase 2) or "library-ification" (support for programmatic invocations / `dbt-server`).

stu-k:
@jtcohen6 Created
- #6707
- #6708

core/dbt/cli/main.py
```diff
@@ -178,10 +181,11 @@ def docs(ctx, **kwargs):
 @requires.preflight
 @requires.profile
 @requires.project
+@requires.runtime_config
+@requires.manifest
 def docs_generate(ctx, **kwargs):
     """Generate the documentation website for your project"""
-    config = RuntimeConfig.from_parts(ctx.obj["project"], ctx.obj["profile"], ctx.obj["flags"])
-    task = GenerateTask(ctx.obj["flags"], config)
+    task = GenerateTask(ctx.obj["flags"], ctx.obj["runtime_config"])
```
MichelleArk, line 188: sorry that I'm catching this post-merge but - should `GenerateTask` be passed the manifest?

ChenyuLInx: I think so, it is inheriting `CompileTask`, the reason it need manifest is that generate task would go to the warehouse and fetch all info about the models in current dbt project.

jtcohen6: Good catch! `docs generate` also performs a full `compile` of all manifest nodes — unless `--select` is passed, in which case only those nodes; or if `--no-compile` is passed, in which case it doesn't

MichelleArk: Looks like the issue is being addressed by @aranke in the [cli test work here](https://github.com/dbt-labs/dbt-core/pull/6723/files#diff-43832707bfec8cdaa49c13f004d9b3296585827f7e68ef07fba7ad1d5e6eb7c6R193).

core/dbt/cli/main.py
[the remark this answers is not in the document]
dbeatty10: We might want to call these changes out here:
https://docs.getdbt.com/guides/migration/versions/upgrading-to-v1.5

Did the proposed **"Abstract manifest compilation / graph generation from tasks."** issue ever get created?

In the meantime, I'm going to delete both of these from `params.py` as part of https://github.com/dbt-labs/dbt-core/issues/6546 since they don't appear to do anything:

https://github.com/dbt-labs/dbt-core/blob/3f76f82c88ad35436d9c7e35d27402d3fe3b892c/core/dbt/cli/params.py#L44-L49

https://github.com/dbt-labs/dbt-core/blob/3f76f82c88ad35436d9c7e35d27402d3fe3b892c/core/dbt/cli/params.py#L491-L496

PR #6567, opened 2023-01-10
Check length of escaped strings in the adapter test

### LINKED ISSUE #6566: [CT-1782] [Bug] Check length of escaped strings in the adapter test

### Current Behavior

Adapter maintainers can choose to extend one of the following pytests (which are expected to be mutually exclusive):
- BaseEscapeSingleQuotesQuote
- BaseEscapeSingleQuotesBackslash

But we've observed at least one case where the adapter made the wrong choice, but the test passed anyways.

Example
The Spark and Databricks adapters were inheriting the `BaseEscapeSingleQuotesQuote` pytest instead of `BaseEscapeSingleQuotesBackslash`. 

This resulted in the following bug report:
https://github.com/dbt-labs/dbt-spark/issues/572

### Expected Behavior

If the adapter maintainer accidentally makes the wrong choice, the test should not pass.

One way to do this would be to also check the length of the resulting string [here](https://github.com/dbt-labs/dbt-core/blob/b13b0e949232ed7fefb626ab0fe6499979621482/tests/adapter/dbt/tests/adapter/utils/fixture_escape_single_quotes.py).

Run the following in postgres:
```sql
select 'let''s' as contraction, length('let''s') as contraction_length
```

We'd expect it to yield:
| contraction | contraction_length |
|-------------|--------------------|
| let's       | 5                  |

Any length other than `5` would indicate a problem. Such a check would have caught the spark/databricks bug.

### Steps To Reproduce

See https://github.com/dbt-labs/dbt-spark/issues/572

### Relevant log output

_No response_

### Environment

```markdown
- OS:
- Python:
- dbt:
```

_No response_

### Additional Context

_No response_

dbeatty10:
resolves #6566

See the following PRs for context and validation that it doesn't break any dbt Labs adapters:
- https://github.com/dbt-labs/dbt-core/pull/6567
- https://github.com/dbt-labs/dbt-spark/pull/586
- https://github.com/dbt-labs/dbt-redshift/pull/260
- https://github.com/dbt-labs/dbt-bigquery/pull/462
- https://github.com/dbt-labs/dbt-snowflake/pull/386

### Description

Adapter maintainers can choose to extend one of the following pytests (which are expected to be mutually exclusive):
- BaseEscapeSingleQuotesQuote
- BaseEscapeSingleQuotesBackslash

We've observed [at least one case](https://github.com/dbt-labs/dbt-spark/issues/572) where the adapter made the wrong choice but the test passed. This PR aims to trigger a CI failure if the wrong choice is made accidentally, and I validated that is does so [here](https://github.com/dbt-labs/dbt-spark/pull/586).

### Checklist

### DIFF
--- a/.changes/unreleased/Under the Hood-20230110-145648.yaml
+++ b/.changes/unreleased/Under the Hood-20230110-145648.yaml
@@ -1,0 +1,6 @@
+kind: Under the Hood
+body: Check length of escaped strings in the adapter test
+time: 2023-01-10T14:56:48.044198-07:00
+custom:
+  Author: dbeatty10
+  Issue: "6566"
--- a/tests/adapter/dbt/tests/adapter/utils/fixture_escape_single_quotes.py
+++ b/tests/adapter/dbt/tests/adapter/utils/fixture_escape_single_quotes.py
@@ -1,15 +1,15 @@
 # escape_single_quotes
 
 models__test_escape_single_quotes_quote_sql = """
-select '{{ escape_single_quotes("they're") }}' as actual, 'they''re' as expected union all
-select '{{ escape_single_quotes("they are") }}' as actual, 'they are' as expected
+select '{{ escape_single_quotes("they're") }}' as actual, {{ length(string_literal(escape_single_quotes("they're"))) }} as actual_length, 'they''re' as expected, 7 as expected_length union all
+select '{{ escape_single_quotes("they are") }}' as actual, {{ length(string_literal(escape_single_quotes("they are"))) }} as actual_length, 'they are' as expected, 8 as expected_length
 """
 
 
 # The expected literal is 'they\'re'. The second backslash is to escape it from Python.
 models__test_escape_single_quotes_backslash_sql = """
-select '{{ escape_single_quotes("they're") }}' as actual, 'they\\'re' as expected union all
-select '{{ escape_single_quotes("they are") }}' as actual, 'they are' as expected
+select '{{ escape_single_quotes("they're") }}' as actual, {{ length(string_literal(escape_single_quotes("they're"))) }} as actual_length, 'they\\'re' as expected, 7 as expected_length union all
+select '{{ escape_single_quotes("they are") }}' as actual, {{ length(string_literal(escape_single_quotes("they are"))) }} as actual_length, 'they are' as expected, 8 as expected_length
 """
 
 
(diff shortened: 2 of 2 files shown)


VersusFacit (commented):
Really intriguing! So, the adapter maintainer picked a macro quoting policy that did not coordinate to the realities of their database system? Curious!

I like how the tests are defined and the numeric confirmation is key. 

My only concern is this absolutely insane case that does happen and might be important to clarify:
> '{{ escape_single_quotes('they\'re') }}'

So like, this is a thing...that works
image

I figure we should add a case to account for this. What's your thought?

VersusFacit:
Oops didn't get my message in before you went on the merge

dbeatty10:
> the adapter maintainer picked a macro quoting policy that did not coordinate to the realities of their database system?

Yep! In this case, it was us as the maintainers of dbt-spark 😅. Since it was so easy to accidentally overlook, it provided motivation to open https://github.com/dbt-labs/dbt-core/issues/6566 to make it hard to get wrong.

Based on our DM's, it seems like part of your concern is that we're covering test cases like:
`'{{ escape_single_quotes("they're") }}'`
but not:
`'{{ escape_single_quotes('they're') }}'`

I tried the latter in my local environment, and it gave the following error:
```
>           raise CompilationException(str(e), node) from e
E           dbt.exceptions.CompilationException: Compilation Error in model test_escape_single_quotes (models/test_escape_single_quotes.sql)
E             expected token ',', got 're'
E               line 2
E                 '{{ escape_single_quotes('they're') }}' as actual,

projects/dbt-core/core/dbt/clients/jinja.py:516: CompilationException
```

I think the issue is that `"they're"` can be parsed as valid string literal within Jinja whereas `'they're'` can not.

Based on my current understanding, I'm comfortable with the changes made by this PR. But I'm open to continued discussion if we're missing something important.

VersusFacit:
Got it. So, on some db's (like Snowflake) it'll be a syntax error and on others, it'll be treated similar identical -- what a nightmare hah. 

>  '{{ escape_single_quotes("they're") }}'

1. Outermost: SQL level
2. Innermost: will be escaped 
3. In-between: what's inside these is passed to the escape function, but not these themselves

My concern was around that last of the three. I hope we have literature or guidance somewhere that makes this clear, because otherwise, that's some deep arcane type magic that would have bothered (among others) early career data engineer me 😄 

Agreed this Issue is resolved. Reclosing the issue. Thanks Doug!

PR #6605, opened 2023-01-13
CT 1440 Fix code to emit ConnectionReused event

### LINKED ISSUE #6168: [CT-1440] [Bug] Connection reused event is not shown

### Current Behavior

The connection reused event does not show in the logs: `Re-using an available connection from the pool (formerly {self.conn_name})`

I think this is because [the connection reused event](https://github.com/dbt-labs/dbt-core/blob/9b84b6e2e87f4e6ada4ccc343906749ba013b138/core/dbt/adapters/base/connections.py#L175) is never fired.

I am not completely sure if I am reading the code below correctly, but I think the `ConnectionReused` is misplaced and maybe never reached:

``` python
        conn = self.get_if_exists()
        if conn is None:
            conn = Connection(
                type=Identifier(self.TYPE),
                name=None,
                state=ConnectionState.INIT,
                transaction_open=False,
                handle=None,
                credentials=self.profile.credentials,
            )
            self.set_thread_connection(conn)

        if conn.name == conn_name and conn.state == "open":
            return conn

        fire_event(NewConnection(conn_name=conn_name, conn_type=self.TYPE))

        if conn.state == "open":
            fire_event(ConnectionReused(conn_name=conn_name))
        else:
            conn.handle = LazyHandle(self.open)

        conn.name = conn_name
        return conn
```

I read the following
1. The code fetches an existing connection
2. If no connection exists, it creates a new one and adds the connection to be used later
3. If the connection has the same name and is open, the connection is returned immediately
4. If not, a new connection event is sent
5. Then if the connection is open, a connection reused event is sent
6. Else lazy handler is set
7. Connection is returned

I do not understand why the connection reused event might be sent right after the new connection event is sent. 

Also, I do not understand when this `conn.name == conn_name` evaluates to `False`. When is a different connection name used within the same thread?

### Expected Behavior

``` python
        conn = self.get_if_exists()
        if conn is None:
            conn = Connection(
                type=Identifier(self.TYPE),
                name=None,
                state=ConnectionState.INIT,
                transaction_open=False,
                handle=None,
                credentials=self.profile.credentials,
            )
            self.set_thread_connection(conn)

        if conn.state == ConnectionState.OPEN.
            fire_event(ConnectionReused(conn_name=conn_name))
        else:
            fire_event(NewConnection(conn_name=conn_name, conn_type=self.TYPE))
            conn.handle = LazyHandle(self.open)
            conn.name = conn_name
        return conn
```

### Steps To Reproduce

NA

### Relevant log output

```shell
NA
```

### Environment

```markdown
- OS: NA
- Python: NA
- dbt: v.1.3.0
```

_No response_

### Additional Context

_No response_

gshank:
resolves #6168

### Description

The code in the adapter "set_connection_name" method was not correctly emitting a ConnectionReused event. Refactor code to make it more understandable and emit the right logging event.

### Checklist


core/dbt/adapters/base/connections.py
```diff
@@ -142,44 +142,44 @@ def exception_handler(self, sql: str) -> ContextManager:
...
-                    f"For connection name, got {name} - not a string!"
-                )
-            assert isinstance(name, str)
-            conn_name = name
+        """Called by 'acquire_connection' in BaseAdapter, which is called by
+        'connection_named', called by 'connection_for(node)'.
+        Creates a connection for this thread if one doesn't already
+        exist, and will rename an existing connection."""
 
+        conn_name: str = "master" if name is None else name
```
JCZuurmond, line 150: Type hint should be `: str | None` or `: Optional[str]`

core/dbt/adapters/base/connections.py
```diff
@@ -142,44 +142,44 @@ def exception_handler(self, sql: str) -> ContextManager:
...
+
+        # Get a connection for this thread
         conn = self.get_if_exists()
+
+        if conn and conn.name == conn_name and conn.state == "open":
```
JCZuurmond, line 155: Do we only consider a connection to be reused if the name changed? Reading the code I would expect this to also be a connection reused.

core/dbt/adapters/base/connections.py
```diff
@@ -142,44 +142,44 @@ def exception_handler(self, sql: str) -> ContextManager:
...
+            # Add the connection to thread_connections for this thread
+            self.set_thread_connection(conn)
+            fire_event(
+                NewConnection(conn_name=conn_name, conn_type=self.TYPE, node_info=get_node_info())
+            )
+        else:  # existing connection either wasn't open or didn't have the right name
+            if conn.state != "open":
+                conn.handle = LazyHandle(self.open)
+            if conn.name != conn_name:
+                orig_conn_name: str = conn.name or ""
```
JCZuurmond, line 179: Why switch to an empty string here? I would expect the same type being used everywhere either `str | None` or `str`

JCZuurmond (commented):
Hi @gshank, thanks for cleaning up this code and fixing the reused connection message. I added some comments.

core/dbt/adapters/base/connections.py
gshank: It can't be Optional[str] because the logging events are defined in protobufs which don't have Optional[str].

core/dbt/adapters/base/connections.py
gshank: It's currently not considered reused. Of course the code is so strangely factored that it's a bit hard to tell how it ought to be used.

core/dbt/adapters/base/connections.py
JCZuurmond: Ah I misread it!

core/dbt/adapters/base/connections.py
JCZuurmond: I misread the type above

peterallenwebb (approved):

PR #5859, opened 2022-09-16
feature/favor-state-node

josephberni:
resolves #5016 

### Description

Add --favor-state option which enables --defer to favor using --state node even if node exists in current target.

This optionally removes the below second --defer criterion when using the --favor-state flag.

This branch has been created and rebased against an [older branch](https://github.com/dbt-labs/dbt-core/pull/5017) so that the owner of the branch has signed the CLA.

### Checklist

- [x] I have read [the contributing guide](https://github.com/dbt-labs/dbt-core/blob/main/CONTRIBUTING.md) and understand what's expected of me
- [x] I have signed the [CLA](https://docs.getdbt.com/docs/contributor-license-agreements)
- [x] I have run this code in development and it appears to resolve the stated issue
- [x] This PR includes tests, or tests are not required/relevant for this PR
- [x] I have [opened an issue to add/update docs](https://github.com/dbt-labs/docs.getdbt.com/issues/new/choose), or docs changes are not required/relevant for this PR
- [x] I have run `changie new` to [create a changelog entry](https://github.com/dbt-labs/dbt-core/blob/main/CONTRIBUTING.md#adding-a-changelog-entry)

PR #6616, opened 2023-01-14
convert 062_defer_state_tests

### LINKED ISSUE #6613: [CT-1820] 062_defer_state_tests

### LINKED ISSUE #6617: [CT-1823] [Bug] [1.4.0rc1] --favor-state not working as intended
### Not working

In situations of "deferral," the `--favor-state` flag should tell dbt to _always_ use the upstream resources defined in the `--state` manifest, even in cases where that resource has already been materialized in the current (dev/CI) namespace.

While https://github.com/dbt-labs/dbt-core/pull/5859 added the flag appropriately, and the `favor_state` argument to [`Manifest.merge_from_artifact`](https://github.com/dbt-labs/dbt-core/blob/20c95a49933cd48dfbba9bd2ad5bc372ad440c6b/core/dbt/contracts/graph/manifest.py#L997), the flag is not actually being piped through [the task](https://github.com/dbt-labs/dbt-core/blob/20c95a49933cd48dfbba9bd2ad5bc372ad440c6b/core/dbt/task/compile.py#L82) and into that method. So it never takes effect!

I think this wasn't caught because the old-style integration tests for this functionality were really tricky to read & understand. I only caught this while converting those tests to the new framework (https://github.com/dbt-labs/dbt-core/pull/6616).

### Repro case

```sql
-- models/model_a.sql
select 1 as id

-- models/model_b.sql
```

```bash
$ dbt run --target prod
$ mkdir state
$ dbt run --target dev
$ dbt run --target dev --select model_b --defer --favor-state --state state/
```
Because the `--favor-state` flag was passed, `model_b` _should_ select from the "prod" version of `model_a`, even though it exists in the dev schema. Instead, it still selects from the dev version:
```sql
  create view "jerco"."dbt_jcohen"."model_b__dbt_tmp" as (
    select * from "jerco"."dbt_jcohen"."model_a"
  );
```

### Missing from some relevant tasks

Also in v1.4, we're adding support for `--defer` to `dbt compile` + `dbt docs generate`. (Makes sense!) Let's make sure `--favor-state` is available on all the same commands as `--defer`.

### [Tech debt] Inconsistent implementation

Also, the way `--favor-state` was implemented is parallel to `--defer`, but different from most of our other flags since v1.0:

https://github.com/dbt-labs/dbt-core/blob/20c95a49933cd48dfbba9bd2ad5bc372ad440c6b/core/dbt/flags.py#L117-L118

(Namely, the argparse default is set based on the flag, rather than the other way around!)

https://github.com/dbt-labs/dbt-core/blob/20c95a49933cd48dfbba9bd2ad5bc372ad440c6b/core/dbt/main.py#L504-L515

Let's just solve for this when we move these flags/etc into the new CLI (#6585). We can always add back-compat for the older env vars + flag names — since, in the meantime, some users might add code to their projects like `{{ flags.FAVOR_STATE_MODE }}`.

jtcohen6:
resolves #6613
resolves #6617

Backport to `1.4.latest` after merging (see https://github.com/dbt-labs/dbt-core/issues/6617#issuecomment-1382732958)

### Description

- Converted to `pytest` framework
- Replace direct file reads/writes with common test utilities (`write_file`, `rm_file`). I was a bit inconsistent in terms of what I defined in + imported from `fixtures.py`, and what I just kept as strings defined inline.
- These are complex & intricate test sequences. I did some light refactoring in places where tests had been over-aggressively duplicated over time, but I kept most of them.
- As part of that refactoring work, I found a bug in the implementation of `--favor-state` that's currently included in v1.4.0-rc1 (#5859). ~I'll open a separate issue + PR to include that commit.~ We should backport this PR to `1.4.latest`, and include the fix for v1.4.0-rc2.
- I do not believe this test needs to be run on any other adapters; it's completely core functionality. As such, I added it to `tests/functional` rather than `tests/adapter`, and I'll open follow-up PRs in adapter repos to remove this test entirely.

### Adapter PRs
- https://github.com/dbt-labs/dbt-redshift/pull/268
- https://github.com/dbt-labs/dbt-snowflake/pull/397
- https://github.com/dbt-labs/dbt-bigquery/pull/468

### Checklist

- ~I have run `changie new` to [create a changelog entry](https://github.com/dbt-labs/dbt-core/blob/main/CONTRIBUTING.md#adding-a-changelog-entry)~


McKnight-42 (approved):
Favor state isn't a config i've heard of so will look into that a bit more, but the tests themselves all make sense and I don't see any logical issues

jtcohen6:
I forgot I included a commit (fa9d2a0) that reverted the fix (083d310), just to prove that the converted test really does catch the bug + prove the fix!

Reverting the revert now...

PR #6620, opened 2023-01-16
Respect quoting config in dbt-py models

### LINKED ISSUE #6103: [Bug] on snowflake "quoting" is not respected in python models

### Current Behavior

Using a quoting: true parameter results in the quoting not being respected.

```
def model(dbt, session):
    print(type(dbt))
    dbt.config(materialized="incremental", unique_key="id")

    leads_created = dbt.source("products_eventstore", "events").filter(
        "\"type\" = 'LeadCreated'"
    )

    leads_df = leads_created["aggregateId"]

    return leads_df

```
yields this query being executed: 
```
SELECT  *  FROM ( SELECT  *  FROM (PRODUCTS.EVENTSTORE.events)) WHERE "type" = 'LeadCreated'
```
when the source specifies quoting the identifier.

### Expected Behavior

The identifier is correctly quoted (as per https://github.com/dbt-labs/dbt-core/blob/main/tests/adapter/dbt/tests/adapter/python_model/test_python_model.py)

### Steps To Reproduce

As above.

### Relevant log output

_No response_

### Environment

```markdown
- OS: OSX 12.5.1
- Python: 3.10.6
- dbt: 1.3.0
```

snowflake

### Additional Context

_No response_

### LINKED ISSUE #6619: [CT-1824] [Bug] Python models do not respect quoting for `dbt.this()`

### Current Behavior

`dbt.this()` should provide access to the full stringified relation name, including quotes (or not) depending on the project- and adapter-level `quoting` config.

This is important especially for incremental modeling logic, which needs the ability to template the preexisting table name into the model code:
```python
    if dbt.is_incremental:

        # only new rows compared to max in current table
        max_from_this = f"select max(updated_at) from {dbt.this}"
        df = df.filter(df.updated_at >= session.sql(max_from_this).collect()[0][0])

        # or only rows from the past 3 days
        df = df.filter(df.updated_at >= F.date_add(F.current_timestamp(), F.lit(-3)))
```

In the current compiled code, `dbt.this()` doesn't include quotes:
```
class this:
    """dbt.this() or dbt.this.identifier"""
    database = 'project-name'
    schema = 'dbt_jcohen'
    identifier = 'python_model'
    def __repr__(self):
        return 'project-name.dbt_jcohen.python_model'
```

So users have to resort to hacks like:
```
          this = "`".join(str(dbt.this).split("."))
          this = f"`{dbt.this.database}`.`{dbt.this.schema}`.`{dbt.this.identifier}`"
```

### Expected Behavior

While `dbt.this.<partname>` should return the unquoted value, `dbt.this()` should return the full rendered relation name, with (or without) quoting as appropriate:
```python
class this:
    """dbt.this() or dbt.this.identifier"""
    database = 'dbt-test-env'
    schema = 'dbt_jcohen'
    identifier = 'python_model'
    def __repr__(self):
        return '`dbt-test-env`.`dbt_jcohen`.`python_model`'
```

### Steps To Reproduce

1. Create a Python model on `dbt-bigquery` in a hyphenated project/dataset name
2. Try to materialize it incrementally, per [our docs](https://docs.getdbt.com/docs/build/python-models#materializations)
3. It fails because the relation name needs to be quoted

### Relevant log output

_No response_

### Environment

```markdown
- OS: 12.6.2
- Python: 3.10.9
- dbt: 1.3.2
```

bigquery

### Additional Context

[Slack thread](https://getdbt.slack.com/archives/C03QUA7DWCW/p1671045654467469)

Related issues:
- https://github.com/dbt-labs/dbt-core/issues/6103
- https://github.com/dbt-labs/dbt-snowflake/issues/330

jtcohen6:
resolves #6103
resolves #6619

### Description

From our code comment:
> Quoting seems like something very specific to sql so far. For all python implementations we are seeing, there's no quoting.

Initial user feedback has proven otherwise!
- We do need to respect quoting in methods that return stringified relation names (`dbt.ref()`, `dbt.source()`, `dbt.this()`)
- Because relation names are interpolated into Python dataframe code as strings, we need to escape relation names that contain double quotes (`"`). This is principally relevant to Snowflake today, but several other data platforms also use `"` as their quoting character.

### Checklist

- [x] This PR includes tests, or tests are not required/relevant for this PR — relevant tests in adapter repos

### DIFF
--- a/core/dbt/compilation.py
+++ b/core/dbt/compilation.py
@@ -351,13 +351,6 @@ class Compiler:
         )
 
         if node.language == ModelLanguage.python:
-            # TODO could we also 'minify' this code at all? just aesthetic, not functional
-
-            # quoating seems like something very specific to sql so far
-            # for all python implementations we are seeing there's no quating.
-            # TODO try to find better way to do this, given that
-            original_quoting = self.config.quoting
-            self.config.quoting = {key: False for key in original_quoting.keys()}
             context = self._create_node_context(node, manifest, extra_context)
 
             postfix = jinja.get_rendered(
@@ -367,8 +360,6 @@ class Compiler:
             )
             # we should NOT jinja render the python model's 'raw code'
             node.compiled_code = f"{node.raw_code}\n\n{postfix}"
-            # restore quoting settings in the end since context is lazy evaluated
-            self.config.quoting = original_quoting
 
         else:
             context = self._create_node_context(node, manifest, extra_context)
--- a/core/dbt/include/global_project/macros/python_model/python.sql
+++ b/core/dbt/include/global_project/macros/python_model/python.sql
@@ -3,7 +3,7 @@
     {%- set ref_dict = {} -%}
     {%- for _ref in model.refs -%}
         {%- set resolved = ref(*_ref) -%}
-        {%- do ref_dict.update({_ref | join("."): resolved.quote(database=False, schema=False, identifier=False) | string}) -%}
+        {%- do ref_dict.update({_ref | join("."): resolved | string | replace('"', '\"')}) -%}
     {%- endfor -%}
 
 def ref(*args,dbt_load_df_function):
@@ -18,7 +18,7 @@ def ref(*args,dbt_load_df_function):
     {%- set source_dict = {} -%}
     {%- for _source in model.sources -%}
         {%- set resolved = source(*_source) -%}
-        {%- do source_dict.update({_source | join("."): resolved.quote(database=False, schema=False, identifier=False) | string}) -%}
+        {%- do source_dict.update({_source | join("."): resolved | string | replace('"', '\"')}) -%}
     {%- endfor -%}
 
 def source(*args, dbt_load_df_function):
--- a/core/dbt/include/global_project/macros/python_model/python.sql
+++ b/core/dbt/include/global_project/macros/python_model/python.sql
@@ -33,8 +33,8 @@ def source(*args, dbt_load_df_function):
     {% set config_dbt_used = zip(model.config.config_keys_used, model.config.config_keys_defaults) | list %}
     {%- for key, default in config_dbt_used -%}
         {# weird type testing with enum, would be much easier to write this logic in Python! #}
-        {%- if key == 'language' -%}
-          {%- set value = 'python' -%}
+        {%- if key == "language" -%}
+          {%- set value = "python" -%}
         {%- endif -%}
         {%- set value = model.config.get(key, default) -%}
         {%- do config_dict.update({key: value}) -%}
@@ -62,11 +62,12 @@ class config:
 
 class this:
     """dbt.this() or dbt.this.identifier"""
-    database = '{{ this.database }}'
-    schema = '{{ this.schema }}'
-    identifier = '{{ this.identifier }}'
+    database = "{{ this.database }}"
+    schema = "{{ this.schema }}"
+    identifier = "{{ this.identifier }}"
+    {% set this_relation_name = this | string | replace('"', '\\"') %}
     def __repr__(self):
-        return '{{ this }}'
+        return "{{ this_relation_name  }}"
 
 
 class dbtObj:
--- a/.changes/unreleased/Fixes-20230116-123645.yaml
+++ b/.changes/unreleased/Fixes-20230116-123645.yaml
@@ -1,0 +1,6 @@
+kind: Fixes
+body: Respect quoting config for dbt.ref() + dbt.source() in dbt-py models
+time: 2023-01-16T12:36:45.63092+01:00
+custom:
+  Author: jtcohen6
+  Issue: "6103"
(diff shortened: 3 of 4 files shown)


core/dbt/compilation.py
```diff
@@ -351,13 +351,6 @@ def _compile_node(
         )
 
         if node.language == ModelLanguage.python:
-            # TODO could we also 'minify' this code at all? just aesthetic, not functional
-
-            # quoating seems like something very specific to sql so far
-            # for all python implementations we are seeing there's no quating.
-            # TODO try to find better way to do this, given that
-            original_quoting = self.config.quoting
```
ChenyuLInx, line 359: I am so glad these lines are removed!

core/dbt/include/global_project/macros/python_model/python.sql
```diff
@@ -3,7 +3,7 @@
     {%- set ref_dict = {} -%}
     {%- for _ref in model.refs -%}
         {%- set resolved = ref(*_ref) -%}
-        {%- do ref_dict.update({_ref | join("."): resolved.quote(database=False, schema=False, identifier=False) | string}) -%}
+        {%- do ref_dict.update({_ref | join("."): resolved | string | replace('"', '\"')}) -%}
```
ChenyuLInx, line 6: is the `"` here tied to specific quoting for snowflake?

ChenyuLInx (approved):
Looks good over all, as long as the quoting `"` is not tied only to snowflake.

core/dbt/compilation.py
VersusFacit: Any day will kill a todo is a good day.

VersusFacit (commented):
I'm reading the principle change here as quoting all necessary `"`s very `\` escapes _within_ the Python environment. 

Will any downstream adapters need to override this or can we trust this behavior holds for all? I'm concern about the quote rules being different in, idk, Bigquery making this a bug introducing PR.

Or is this quoting restricted to the Python environment. [And as a corollary, whatever the underlying db receives from Python will evaluate if the relation name string makes sense for the engine you're on]

core/dbt/include/global_project/macros/python_model/python.sql
jtcohen6: I believe it would be relevant to any data platform that uses `"` as its quoting character ([it's our default](https://github.com/dbt-labs/dbt-core/blob/main/core/dbt/adapters/base/relation.py#L32)), or any identifier that contains a quote character (`"`) for any reason

jtcohen6:
> I'm reading the principle change here as quoting all necessary "s very \ escapes within the Python environment.

Correct! This isn't specific to Snowflake, it would also be relevant on any other data platform that uses `"` as its quoting character. (As it happens, both BigQuery + Spark/Databricks use backticks, and don't allow other special characters in identifier names.)

Big idea: in Python models, `dbt.ref()` is really producing this code:
```python
session.table("database.schema.table")
```
Where the table identifier is a Python string, i.e. quoted. If there are _any_ instances of quotation marks (`"`) in `database`, `schema`, or `table`, we need to escape, just as you would for any Python string containing quotes:
```python
>>> string = "hello, \"world\"!"
>>> print(string)
hello, "world"!
```

The automated test confirming that this change works is really in https://github.com/dbt-labs/dbt-snowflake/pull/399, which also resolves a `dbt-snowflake`-specific bug in its own right.

core/dbt/include/global_project/macros/python_model/python.sql
```diff
@@ -62,11 +62,12 @@ class config:
 
 class this:
     """dbt.this() or dbt.this.identifier"""
-    database = '{{ this.database }}'
-    schema = '{{ this.schema }}'
-    identifier = '{{ this.identifier }}'
+    database = "{{ this.database }}"
+    schema = "{{ this.schema }}"
+    identifier = "{{ this.identifier }}"
+    {% set this_relation_name = this | string | replace('"', '\\"') %}
```
ChenyuLInx, line 68: One potential option is to make this a macro and have adapters overwrite it

ChenyuLInx: This comment is related to the change of respecting quoting schema would break bigquery adapter

jtcohen6: Just chatted with @colin-rogers-dbt about this. Simplest solution is to turn this processing into a dispatched macro that can be overridden on `dbt-bigquery`.

```
{% macro render_relation_for_python_model(relation) %}
  {{ adapter.dispatch('render_relation_for_python_model', 'dbt')(relation)
{% endmacro %}

{% macro default__render_relation_for_python_model(relation) %}
  {{ return(relation | string | replace('"', \\"') }}
{% endmacro %}

{% macro bigquery__render_relation_for_python_model(relation) %}
  {{ return(relation.quote(database=False, schema=False, identifier=False) | string) }}
{% endmacro %}
```

And use that macro in all three spots that I changed here (`ref`, `source`, `this`)

PR #6650, opened 2023-01-19
fix MP_CONTEXT is not JSON serializable

### LINKED ISSUE #6509: [CT-1746] Flags.MP_CONTEXT is not JSON serializable
One of the many exceptions raised during integration test runs on the feature/click-cli branch is: 

`TypeError: Object of type SpawnContext is not JSON serializable`

There are a [couple options](https://github.com/dbt-labs/dbt-core/pull/6504#issuecomment-1371002518) to mitigate the issue:
1. Remove MP_CONTEXT from flags (where would be the best place to put it instead?)
2. Avoid serializing MP_CONTEXT

ChenyuLInx:
Resolves #6509

This fixes the issue by adding logic to skip duplicated capped attributes introduced in Flags, also remove `mp_context` that is added.
Test that failed before now passes
* [x] `tests/adapter/dbt/tests/adapter/grants/test_snapshot_grants.py`

### DIFF
--- a/core/dbt/utils.py
+++ b/core/dbt/utils.py
@@ -642,7 +642,10 @@ def args_to_dict(args):
     dict_args = {}
     # remove args keys that clutter up the dictionary
     for key in var_args:
-        if key == "cls":
+        if key.lower() in var_args and key == key.upper():
+            # skip all caped keys being introduced by Flags in dbt.cli.flags
+            continue
+        if key in ["cls", "mp_context"]:
             continue
         if var_args[key] is None:
             continue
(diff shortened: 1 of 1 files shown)


core/dbt/utils.py
```diff
@@ -642,7 +642,10 @@ def args_to_dict(args):
     dict_args = {}
     # remove args keys that clutter up the dictionary
     for key in var_args:
-        if key == "cls":
+        if key.lower() in var_args and key == key.upper():
+            # skip all caped keys being introduced by Flags in dbt.cli.flags
```
MichelleArk, line 646: typo/nit: 'caped' - perhaps you meant 'capped'?

PR #6682, opened 2023-01-20
Dynamically list all `.latest` branches for scheduled testing

### LINKED ISSUE #5773: [CT-1129] Auto-generate list of release branches to test against
Refer to PR #5767 where we added `1.2.latest` to our testing matrix. This is to auto-generate the list of release branches instead of manually updating.

To future proof, we should make this auto add any `.latest` branches we push.  This could be done if we add a job at the beginning of the workflow that essentially sets a branches variable using:

```bash
git branch -r --list | grep "**.latest"
```

For clarity the first job could have a second step that appends `main` onto the list.

You can then set the matrix in the next job from the variable output of that first job.  This way we wouldn't have to modify this at all.

(mildly related [reference](https://thekevinwang.com/2021/09/19/github-actions-dynamic-matrix/))

_Originally posted by @emmyoop in https://github.com/dbt-labs/dbt-core/pull/5767#pullrequestreview-1097578998_

emmyoop:
resolves #5773 

### Description

This auto generates the list of all `*.latest` branches (plus `main`) for our scheduled integration tests.

Eventually we would want to start removing versions once we no longer support them but this iteration does not deal with that.

### Checklist

### DIFF
--- a/.github/workflows/release-branch-tests.yml
+++ b/.github/workflows/release-branch-tests.yml
@@ -34,6 +34,9 @@ jobs:
       # Will look like '["1.0.latest", "1.2.latest"]'
       branches: ${{ toJSON(steps.generate.outputs.branch_matrix) }}
     steps:
+      - name: Check out the repository
+        uses: actions/checkout@v3
+
       - name: "Get all branches"
         run: |
           branches=($(git branch --format "%(refname:short)" | grep "^1.*.latest"))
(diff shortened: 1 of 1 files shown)


emmyoop:
Tested here: https://github.com/dbt-labs/dbt-core/actions/runs/4088492211

leahwicz (approved):
LGTM

PR #6691, opened 2023-01-20
CT-1718: Add Note and Formatting event types

### LINKED ISSUE #6481: [CT-1718] Add an "FYI" Message Type
### Describe the feature

Sometimes we have messages we would would like to write to the console without creating a custom event type. For example, we often want to add formatting by inserting blank lines or break lines like "=====...". We also have a variety of messages we print for informational purposes, but which are unlikely to be consumed programmatically by structured logging users.

To solve this problem, let's introduce a new event type which contains the standard event fields, but only one additional field (called "Text" or "Message"?) which can include miscellaneous text and formatting.

After this change, we should also be able to remove the EmptyLine event.

### Who will this benefit?

Consumers of structured logging will benefit from this change, since they will no longer see events like EmptyLine which are useless to them. Since this change will also make it possible to remove other events, like StarterProjectPath and project initialization related events, it will reduce the complexity and size of our event schema.

peterallenwebb:
resolves #6481 

### Description

This PR replaces the EmptyLine event with a more general Formatting event. This new event isn't any semantically useful than EmptyLine, but let's us replace a broader set of events which are used only for formatting console output with a single event for that purpose which can be ignored for the purposes of structured logging.

This PR also adds a Note event, which is closer in spirit to the "FYI" event type mentioned in the original issue, and which dbt-server has asked us to include so that they can insert messages into the dbt log.

### Checklist


gshank (approved):
Looks good!

Issue #6073, opened 2022-10-16
[Bug] Performance issues with DBT build when nodes more than 2K in a project


china-cse:
### Is this a new bug in dbt-core?

- [X] I believe this is a new bug in dbt-core
- [X] I have searched the existing issues, and I could not find an existing issue for this bug

### Current Behavior

In case project has more than 2K-3K nodes(includes models, tests etc), **dbt build** taking time to start its first model. Just to build a single model itself takes 15 mins. Have disabled analytics tracking event thought that could have caused but still no luck. However, dbt run is faster but I can't go with just run as tests required immediately before proceed to next model in DAG.

### Expected Behavior

Models could be more depends project size but building single model not supposed to take 15 mins and it does not look realistic when we have multiple dbt commands to run.

### Steps To Reproduce

version: dbt 1.0 till latest release 1.3

total nodes count: 3K+ including models, tests, snapshots, seeds

build single model:
$dbt build -s test_model

### Environment

```markdown
- OS: Windows/DBT Cloud
- Python: 3.7.4
- dbt: 1.2
```

### Which database adapter are you using with dbt?

snowflake

jtcohen6:
Thanks for opening @china-cse!

This is an issue we saw after introducing the `dbt build` command last year. The fact is that `build` requires constructing a more complex DAG, with tests as additional upstream dependencies, versus the simpler DAG used for other commands. After the v0.21 release, we decided the right move was to create genuinely different DAGs for `build` and non-`build` commands.

Some ideas we could explore to improve performance here:
- A more-performant `networkx` algorithm (in the vein of https://github.com/dbt-labs/dbt-core/pull/4155)
- Caching the `networkx` DAG between invocations — with cache invalidation for project code / config changes (like partial parsing), or when switching between `build` and non-`build` commands
- Where possible, perform certain types of node selection _before_ adding test edges to the graph. This either wouldn't work with graph operators, or it would represent a functional change, if we resolve graph operators _before_ adding test edges. (Imagine a case with `model_a --> some_test` and `model_a --> model_b`. What should `dbt build --select some_test+` resolve to? Today, it includes `model_b`; if we reordered the logic so graph selection happens before the `some_test --> model_b` edge is added, it no longer would.) But it could make a big difference in the case of `dbt build --select single_model`.

[...]

Issue #5527, opened 2022-07-25
[CT-901] [Epic] API-ification + CLI - Phase 1


iknox-fa:
This Epic encompasses the work to create a new cli using the click CLI framework—and, in so doing, to start the process of turning `dbt-core` into a mature Python library with stable APIs & programmatic entry points.

Generally, the goal of this epic is **functional parity.** From the perspective of most end users, everything should work just the same as before. We are _not_ aiming to introduce new functionality into `dbt-core`—except what we get "for free" from the click framework, and by (re)designing & (finally) documenting our internal APIs—but we _are_ laying foundations that will enable us to add new capabilities, in https://github.com/dbt-labs/dbt-core/issues/6356 and subsequent initiatives.

### Get `dbt run` working
* [x] #5526 
* [x] #5528
* [x] #5529
* [x] #6327
* [x] #5530
* [x] #5531 
* [x] #5533
* [x] #5534
* [x] #6096
* [x] #5551

### Parity for all commands (ordered list based off usage)

* [x] #5550 - tracked by https://github.com/dbt-labs/dbt-core/issues/6357
* [x] #5553
* [x] #5545
* [x] #5547
* [x] #5556
* [x] #5554
* [x] #5542
* [x] #5555
* [x] #5549
* [x] #5541 
* [x] #5543
* [x] #5552
* [x] #5544
* [x] #5546 
* [x] #5548
* [x] #6757

### Parity for all params

* [x] #6577
* [x] #6585
* [x] #6579

jtcohen6:
Closing this initiative as completed :)

dbeatty10:
The dbt-core CLI uses `click` for v1.5+ and used `argparse` before that.

PR #6694, opened 2023-01-22
Cheeky performance improvement on big DAGs

### LINKED ISSUE #6697: [CT-1881] [Feature] When possible, bypass an expensive method during graph construction

### Describe the feature

https://github.com/dbt-labs/dbt-core/issues/6073 describes how projects with 1000's of nodes can take ~15 minutes to `dbt build` even a single node.

The `_already_known` method call is a relatively expensive method call during DAG parsing. There are some cases where it is unnecessarily called, namely when `in_degree != 0` [here](https://github.com/dbt-labs/dbt-core/blob/93619a9a3761d3606d257c1274828ac4eef4f627/core/dbt/graph/queue.py#L159-L166).

Swapping the conditions of the if statement would short-circuit calling `_already_known`  while keeping the logic the same.  This is shown in https://github.com/dbt-labs/dbt-core/commit/9f9608731263b7b3b3e8dfd22a9d579cb650cd54. (We know we can safely swap the order by [commutativity of conjunction](https://en.wikipedia.org/wiki/Commutativity_of_conjunction)).

There's at least two alternatives:
1. Default approach of changing nothing
1. More complicated refactor

This proposal seems obviously superior to the first alternative while not preventing us from taking up the second alternative at a later date. So it's an easy win.

### Who will this benefit?

This will (slightly) benefit teams with very large DAGs (thousands of nodes).

There's already a PR ready for this!

### Anything else?

_No response_

boxysean:
resolves #6697

### Description

Small optimization on manifest parsing benefitting large DAGs.

Swapping the conditions of the `if` statement as shown in 9f9608731263b7b3b3e8dfd22a9d579cb650cd54 short-circuits the relatively expensive `_already_known` method call. In a test DAG provided by my client, with 8468 models and 17103 tests, this significantly reduced calls to `_already_known` and reduced runtime in my test environment. See profiling screenshot of a `dbt compile` command:

Screenshot 2023-01-22 at 9 12 15 PM

I don't know precisely what this will translate to in real-world results, but it's safe to say (1) there are fewer calls to a relatively expensive method, and (2) the logic is the same. My client reports any `dbt build` command takes 15 minutes startup time -- I hope afterwards the startup time will be between 11-12 minutes.

### Checklist

### DIFF
--- a/.changes/unreleased/Under the Hood-20230122-215235.yaml
+++ b/.changes/unreleased/Under the Hood-20230122-215235.yaml
@@ -3,4 +3,4 @@ body: Small optimization on manifest parsing benefitting large DAGs
 time: 2023-01-22T21:52:35.549814+01:00
 custom:
   Author: boxysean
-  Issue: "6322"
+  Issue: "6073"
--- a/.changes/unreleased/Under the Hood-20230122-215235.yaml
+++ b/.changes/unreleased/Under the Hood-20230122-215235.yaml
@@ -3,4 +3,4 @@ body: Small optimization on manifest parsing benefitting large DAGs
 time: 2023-01-22T21:52:35.549814+01:00
 custom:
   Author: boxysean
-  Issue: "6073"
+  Issue: "6697"
--- a/.changes/unreleased/Under the Hood-20230122-215235.yaml
+++ b/.changes/unreleased/Under the Hood-20230122-215235.yaml
@@ -1,0 +1,6 @@
+kind: Under the Hood
+body: Small optimization on manifest parsing benefitting large DAGs
+time: 2023-01-22T21:52:35.549814+01:00
+custom:
+  Author: boxysean
+  Issue: "6322"
--- a/core/dbt/graph/queue.py
+++ b/core/dbt/graph/queue.py
@@ -161,7 +161,7 @@ class GraphQueue:
         queue and add them.
         """
         for node, in_degree in self.graph.in_degree():
-            if not self._already_known(node) and in_degree == 0:
+            if in_degree == 0 and not self._already_known(node):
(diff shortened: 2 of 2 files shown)


.changes/unreleased/Under the Hood-20230122-215235.yaml
```diff
@@ -0,0 +1,6 @@
+kind: Under the Hood
+body: Small optimization on manifest parsing benefitting large DAGs
+time: 2023-01-22T21:52:35.549814+01:00
+custom:
+  Author: boxysean
+  Issue: "6073"
```
dbeatty10, line 6: I created a stand-alone issue for this surgical change. That way the changelog doesn't accidentally imply that this PR completely solves #6073.

But I'm open to the idea of multiple PRs leading back to the same issue if that's a precedent we've already set (or one we want to start using).
```suggestion
  Issue: "6697"
```

dbeatty10 (changes requested):
Very smart @boxysean ! :brain: Commutativity for the win! 🏆 

One small suggestion:
- As an alternative to [#6073](https://github.com/dbt-labs/dbt-core/issues/6073), I created a stand-alone issue for this precision change that we could link back to instead
- We can treat [#6073](https://github.com/dbt-labs/dbt-core/issues/6073) more like [#5527](https://github.com/dbt-labs/dbt-core/issues/5527)

Open to discussion if this seems unnecessary or undesirable.

boxysean:
Perfect, plan looks good @dbeatty10! Thanks, let me know if you need anything else from me.

dbeatty10 (approved):
Let's goooo!

I will merge once it finishes its additiona round of CI checks.

PR #6711, opened 2023-01-24
Make project version Optional

seub:
resolves #6603 

### Description

I changed `Project.version` to `Optional` (default `None`) in core/dbt/config and core/dbt/contracts. 

(This change was suggested by @jtcohen6 in the issue #6603)

This config was not being used by dbt, so I believe no further changes are needed. `make test` and `make integration` both passed.

### Checklist

- [x] I have read [the contributing guide](https://github.com/dbt-labs/dbt-core/blob/main/CONTRIBUTING.md) and understand what's expected of me
- [x] I have signed the [CLA](https://docs.getdbt.com/docs/contributor-license-agreements)
- [x] I have run this code in development and it appears to resolve the stated issue
- [ ] This PR includes tests, or tests are not required/relevant for this PR
- [x] I have [opened an issue to add/update docs](https://github.com/dbt-labs/docs.getdbt.com/issues/new/choose), or docs changes are not required/relevant for this PR
- [x] I have run `changie new` to [create a changelog entry](https://github.com/dbt-labs/dbt-core/blob/main/CONTRIBUTING.md#adding-a-changelog-entry)

PR #6718, opened 2023-01-24
Made project version optional (default None)

### LINKED ISSUE #6603: [Feature] Remove required 'version' from dbt_project.yml

### Describe the feature

At the moment, it is required that dbt projects include a `version` in their dbt_project.yml, which should be a semantic version, as mentioned in the dbt doc [here](https://docs.getdbt.com/reference/project-configs/version).

My dbt project is a git package with its own process for creating and publishing its semantic versioning. The semantic version is not hard coded in the code base, but it is generated semi-automatically by our CI/CD. I imagine this is a common practice.

Therefore our dbt project has 2 semantic versions, the one in `dbt_project.yml`, which does not serve any meaningful purpose according to the dbt doc, and the git semantic version (tag).

In summary, I suggest removing the required 'version' from dbt_project.yml, because:
- it does not serve any purpose
- it is confusing as it is potentially in conflict with the git repo's own versioning.

_No response_

### Who will this benefit?

All?

Sure

### Anything else?

https://docs.getdbt.com/reference/project-configs/version

seub:
resolves #6603 

### Description

This MR makes the project version optional.

I changed `Project.version` to `Optional` (default `None`) in core/dbt/config and core/dbt/contracts. 

(This change was suggested by @jtcohen6 in the issue #6603)

This config was not being used by dbt, so I believe no further changes are needed. `make test` and `make integration` both passed.

### Checklist

    - https://github.com/dbt-labs/docs.getdbt.com/issues/2747

### DIFF
--- a/.changes/unreleased/Features-20230124-135550.yaml
+++ b/.changes/unreleased/Features-20230124-135550.yaml
@@ -1,0 +1,6 @@
+kind: Features
+body: Make project version optional
+time: 2023-01-24T13:55:50.86071024-08:00
+custom:
+  Author: seub
+  Issue: "6603"
--- a/core/dbt/config/project.py
+++ b/core/dbt/config/project.py
@@ -525,7 +525,7 @@ class VarProvider:
 @dataclass
 class Project:
     project_name: str
-    version: Union[SemverString, float]
+    version: Optional[Union[SemverString, float]]
     project_root: str
     profile_name: Optional[str]
     model_paths: List[str]
--- a/core/dbt/contracts/project.py
+++ b/core/dbt/contracts/project.py
@@ -184,8 +184,8 @@ BANNED_PROJECT_NAMES = {
 @dataclass
 class Project(HyphenatedDbtClassMixin, Replaceable):
     name: Identifier
-    version: Union[SemverString, float]
     config_version: int
+    version: Optional[Union[SemverString, float]] = None
     project_root: Optional[str] = None
     source_paths: Optional[List[str]] = None
     model_paths: Optional[List[str]] = None
--- a/core/dbt/tests/fixtures/project.py
+++ b/core/dbt/tests/fixtures/project.py
@@ -181,7 +181,6 @@ def dbt_project_yml(project_root, project_config_update, logs_dir):
     project_config = {
         "config-version": 2,
         "name": "test",
-        "version": "0.1.0",
         "profile": "test",
         "log-path": logs_dir,
     }
--- a/tests/functional/basic/test_project.py
+++ b/tests/functional/basic/test_project.py
@@ -1,0 +1,31 @@
+import pytest
+from dbt.tests.util import run_dbt, update_config_file
+from dbt.exceptions import ProjectContractError
+
+
+class TestProjectYamlVersionMissing:
+    # default dbt_project.yml does not fill version
+
+    def test_empty_version(self, project):
+        run_dbt(["run"], expect_pass=True)
+
+
+class TestProjectYamlVersionValid:
+    @pytest.fixture(scope="class")
+    def project_config_update(self):
+        return {"version": "1.0.0"}
+
+    def test_valid_version(self, project):
+        run_dbt(["run"], expect_pass=True)
+
+
+class TestProjectYamlVersionInvalid:
+    def test_invalid_version(self, project):
+        # we need to run it so the project gets set up first, otherwise we hit the semver error in setting up the test project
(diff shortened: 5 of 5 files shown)


seub:
This PR replaces https://github.com/dbt-labs/dbt-core/pull/6711 which was broken.

dbeatty10:
Thanks for submitting this @seub !

At least one check is complaining:
image

This particular check is related to the way we automatically generate our changelogs (like [this one for 1.3](https://github.com/dbt-labs/dbt-core/blob/3f1704438363ceb1b17ca3315a89cdb61c1e78df/CHANGELOG.md)).

You can resolve this by running `changie new` as described [here](https://github.com/dbt-labs/dbt-core/blob/cbb9117ab997289062cd5bacaf4cfd4466ab9b1e/CONTRIBUTING.md#adding-or-modifying-a-changelog-entry) and then committing & pushing the file it creates.

seub:
@dbeatty10 Thanks! I had forgotten to push the changie commit, my bad

emmyoop (changes requested):
@seub thanks for the contribution!  These changes look good but we do want to make sure we have tests.  Right now the version is filled for all tests by default, which is why they continue to pass.

Since this is an optional field now, we should remove it from the fixture we use in all tests here:

https://github.com/dbt-labs/dbt-core/blob/3cee9d16fa39012e30a02c66770d6b6c7a445f6c/core/dbt/tests/fixtures/project.py#L184

Then, make sure all the tests still pass.  As long as they do you'll want to write a test where we add the version in using the `project_config_update` method.  There are other tests that do define a version here but let's be sure to test this specifically while we're making the change.

Add the test as a new file named `test_project.py` in `dbt/tests/functional/basic`.  You can leverage some fixtures that already exist.

Test no version, a valid semver compliant version, and a non-semver compliant version.

This involves a bit of knowledge of the utilities we use in out tests so I went ahead and wrote up the test file you should use.

```python
import pytest
from dbt.tests.util import run_dbt, update_config_file
from dbt.exceptions import ProjectContractError


class TestProjectYamlVersionMissing:
    # default dbt_project.yml does not fill version

    def test_empty_version(self, project):
        run_dbt(["run"], expect_pass=True)

class TestProjectYamlVersionValid:
    def project_config_update(self):
        return {"version": "1.0.0"}

    def test_valid_version(self, project):
        run_dbt(["run"], expect_pass=True)

class TestProjectYamlVersionInvalid:

    def test_invalid_version(self, project):
        # we need to run it so the project gets set up first, otherwise we hit the semver error in setting up the test project
        run_dbt()
        update_config_file({"version": "invalid"}, "dbt_project.yml")
        with pytest.raises(ProjectContractError) as excinfo:
            run_dbt()
        assert "at path ['version']: 'invalid' is not valid under any of the given schemas" in str(excinfo.value)

```

You'll still want to make sure it works as expected as well as making sure other tests don't break.

seub:
@emmyoop Thank you very much for your help!

There just seems to be a small issue in your test `test_invalid_version` that I'm hoping you can help me fix. 

```
======================================================================== short test summary info =========================================================================
FAILED tests/functional/basic/test_project.py::TestProjectYamlVersionInvalid::test_invalid_version - dbt.exceptions.DbtRuntimeError: Runtime Error
====================================================================== 1 failed, 2 passed in 3.05s ======================================================================
```

It looks to me like the problem is that the `run_dbt()` inside the `with` raises a `DbtRuntimeError` and not a `ProjectContractError`, so the assert statement is never reached.

In the captured stdout call, I do see:
```bash
Invoking dbt with ['run']
14:45:54  Encountered an error while reading the project:  ERROR: Runtime Error
  at path ['version']: 'invalid' is not valid under any of the given schemas
  ```
  But I don't know how to extract that error message.

I tried a couple things like inserting `expect_pass=True` in `run_dbt()` but it didn't help. Any idea?

One thing that does fix the test "of course" is to replace `ProjectContractError` by `DbtRuntimeError` and `"at path ['version']: 'invalid' is not valid under any of the given schemas"` by `"Could not run dbt"`. But I assume that's not what we want to do.

emmyoop:
@seub I tested this locally and it's working for me.  The tests also pass in CI.  I'm not able to reproduce the behavior you're experiencing.  What command are you using to run your tests?

emmyoop:
@seub I have been unable to reproduce your issue.  Since it's working locally for me as well as in CI I'm going to go ahead and merge this change in.  Thank you again for the contribution!

PR #6720, opened 2023-01-25
Graph Analysis Optimization for Large Dags

### LINKED ISSUE #6759: [CT-1927] [Feature] Consider only successors of the deleted nodes when populating the GraphQueue

### Describe the feature

The current implementation [here](https://github.com/dbt-labs/dbt-core/blob/c65ba11ae6dca5203228320a7f6388ce1796f104/core/dbt/graph/queue.py#L159-L166) looks through all four of these nodes, no matter what.

image

We can do better! In fact, we only need to add nodes with an in-degree of 0. This means we only need to visit nodes that could possibly have an in-degree of zero. This only happens to a node whose predecessor is removed, so we can restrict visits to successor nodes of removed nodes.

#### Example
We can make it so that if Node 2 is removed, then only its successors (Nodes 3 and 4) will be visited, and only Node 3 will change to an in-degree of zero. Only nodes with an in-degree of zero need to be re-added to the queue -- in this case, only Node 3.

### Average case
We can see that before it would visit all X nodes in the average case. But now the average case is just the typical number of successors (i.e., the mean number of out-degrees per node).

I would guess for most dbt projects, the mean number of out-degrees will be WAAAAY closer to 0 than it will be to X, hence "taking almost no time at all" 😉

The status quo is _an_ alternative, but not a very competitive one 😄

There might be some other optimizations we can pursue, but this is a good place to start.

### Who will this benefit?

This should help somewhat with large dbt projects with 1000+ nodes like described in https://github.com/dbt-labs/dbt-core/issues/6073

See https://github.com/dbt-labs/dbt-core/pull/6720

### Anything else?

_No response_

peterallenwebb:
resolves #6759

### Description

This optimization reduces by 25% the time needed to run dbt compile in a large scenario provided to me by @boxysean. In that scenario, dbt originally spent 25% of its time in the GraphQueue._find_new_additions() function, which has been modified in the PR so that it takes almost no time at all.

The optimization works because the the in-degree of a node in the GraphQueue's graph only changes when we remove a node, and the only nodes whose in-degree will change are those that were successors of the deleted node. Instead of looking at every node in the graph to see whether its in-degree has changed to zero, we now look only at successors of the deleted node.

### Checklist

### DIFF
--- a/.changes/unreleased/Under the Hood-20230126-135939.yaml
+++ b/.changes/unreleased/Under the Hood-20230126-135939.yaml
@@ -1,0 +1,6 @@
+kind: Under the Hood
+body: ' Optimized GraphQueue to remove graph analysis bottleneck in large dags.'
+time: 2023-01-26T13:59:39.518345-05:00
+custom:
+  Author: peterallenwebb
+  Issue: "6759"
--- a/core/dbt/graph/queue.py
+++ b/core/dbt/graph/queue.py
@@ -40,7 +40,7 @@ class GraphQueue:
         # store the 'score' of each node as a number. Lower is higher priority.
         self._scores = self._get_scores(self.graph)
         # populate the initial queue
-        self._find_new_additions()
+        self._find_new_additions(list(self.graph.nodes()))
         # awaits after task end
         self.some_task_done = threading.Condition(self.lock)
 
@@ -156,12 +156,12 @@ class GraphQueue:
         """
         return node in self.in_progress or node in self.queued
 
-    def _find_new_additions(self) -> None:
+    def _find_new_additions(self, candidates) -> None:
         """Find any nodes in the graph that need to be added to the internal
         queue and add them.
         """
-        for node, in_degree in self.graph.in_degree():
-            if in_degree == 0 and not self._already_known(node):
+        for node in candidates:
+            if self.graph.in_degree(node) == 0 and not self._already_known(node):
                 self.inner.put((self._scores[node], node))
                 self.queued.add(node)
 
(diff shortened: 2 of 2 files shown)


dbeatty10 (approved):
> it takes almost no time at all

🎉 

This looks awesome @peterallenwebb 🧠 

If I'm understanding correctly, the original implementation would have looked through all four of these nodes, no matter what.

image

In your new implementation, if Node 2 is removed, then only its successors (Nodes 3 and 4) will be visited, and only Node 3 will change to an in-degree of zero.

We can see that before it would visit all X nodes in the average case. But now the average case is just the typical number of successors (i.e., the mean number of out-degrees per node).

I would guess for most dbt projects, the mean number of out-degrees will be WAAAAY closer to 0 than it will be to X, hence "taking almost no time at all" 😉

peterallenwebb:
@dbeatty10 Yes, your summarization is completely correct. @iknox-fa, since you've likely spent a lot more time with this code than me, can you let me know what you think about this change? If you think it looks good, I'll plan to merge soon.

dbeatty10 (dismissed):
@peterallenwebb I didn't see a changelog entry or an associated GitHub issue.

Is there a reason we wouldn't want to do both of those here?

peterallenwebb:
> @peterallenwebb I didn't see a changelog entry or an associated GitHub issue.
> 
> Is there a reason we wouldn't want to do both of those here?

No reason. Since I'm new to this part of the code, I just wanted a sanity check before I went to the trouble, but I'll create those tomorrow.

dbeatty10 (approved):
🚀

PR #6723, opened 2023-01-25
[CT-920][CT-1900] Create Click CLI runner and use it to fix dbt docs …

### LINKED ISSUE #5544: `dbt docs serve` works in click

aranke:
resolves #5544 #6722 

### Checklist

### DIFF
--- a/tests/unit/test_events.py
+++ b/tests/unit/test_events.py
@@ -1,29 +1,18 @@
 # flake8: noqa
-from dbt.events.test_types import UnitTestInfo
-from dbt.events import AdapterLogger
-from dbt.events.functions import event_to_json, LOG_VERSION, event_to_dict
-from dbt.events.types import *
-from dbt.events.test_types import *
+import re
+from typing import TypeVar
 
+from dbt.contracts.files import FileHash
+from dbt.contracts.graph.nodes import ModelNode, NodeConfig, DependsOn
+from dbt.events import AdapterLogger
 from dbt.events.base_types import (
     BaseEvent,
-    DebugLevel,
-    WarnLevel,
-    InfoLevel,
-    ErrorLevel,
     TestLevel,
 )
-from dbt.events.proto_types import ListOfStrings, NodeInfo, RunResultMsg, ReferenceKeyMsg
-from importlib import reload
-import dbt.events.functions as event_funcs
-import dbt.flags as flags
-import inspect
-import json
-from dbt.contracts.graph.nodes import ModelNode, NodeConfig, DependsOn
-from dbt.contracts.files import FileHash
-from mashumaro.types import SerializableType
-from typing import Generic, TypeVar, Dict
-import re
+from dbt.events.functions import event_to_json, event_to_dict
+from dbt.events.test_types import *
+from dbt.events.types import *
+
 
 # takes in a class and finds any subclasses for it
 def get_all_subclasses(cls):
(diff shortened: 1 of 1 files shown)


core/dbt/task/serve.py
```diff
@@ -1,46 +1,28 @@
...
-            httpd.server_close()
+        if self.args.browser:
+            webbrowser.open_new_tab(f"http://localhost:{port}")
 
-        return None
+        with socketserver.TCPServer(("", port), SimpleHTTPRequestHandler) as httpd:
+            click.echo(f"Serving docs at {port}")
+            click.echo(f"To access from your browser, navigate to: http://localhost:{port}")
+            click.echo("\n\n")
+            click.echo("Press Ctrl+C to exit.")
```
jtcohen6, line 27: This change makes sense to me. `docs serve` is CLI-only functionality; it might make more sense to just `click.echo`, rather than firing a real event.

If we did want these to keep being "real" events / log messages, we could use the `Formatting` ("FYI") message type that @peterallenwebb added in https://github.com/dbt-labs/dbt-core/pull/6691. I leave it up to your discretion.

In any case, I think we could also just remove the one-off event types that are no longer being used anywhere else in the codebase: `ServingDocsPort`, `ServingDocsAccessInfo`, `ServingDocsExitInfo `

tests/cli/dbt_project.yml
```diff
@@ -0,0 +1,5 @@
+name: jaffle_shop
+profile: jaffle_shop
+version: '0.1.0'
+config-version: 2
+clean-targets: [target, dbt_packages, logs]
```
MichelleArk, line 5: What is the motivation for defining the project files in this manner as opposed to what's in place in terms of fixtures for the existing functional tests? 

For example, this file could be overwritten in an individual test method or class via the [dbt_project_yaml fixture](https://github.com/dbt-labs/dbt-core/blob/main/core/dbt/tests/fixtures/project.py#L180). The existing fixtures have some nice test isolation + cleanup functionality that should get rid of the need to [.gitignore `dbt_packages` in this PR](https://github.com/dbt-labs/dbt-core/pull/6723/files#diff-25a58f2510df1a432759522298c0ec0fc19b09b9e432947fc4369a7e823f87aeR1) and make it more straightforward to write additional cli tests that have their own isolated project environment.

tests/cli/profiles.yml
```diff
@@ -0,0 +1,11 @@
+jaffle_shop:
```
ChenyuLInx, line 1: The way these test files being stored is similar to the old testing framework. Can we test it similar to what the functional tests does?
cc @gshank

tests/cli/test_cli_run.py
```diff
@@ -0,0 +1,48 @@
+import pytest
+from click.testing import CliRunner
+
+from dbt.cli.main import cli
+from tests.cli.fixtures import BaseConfigProject
+
+
+class TestCli(BaseConfigProject):
```
ChenyuLInx, line 8: Intention is to have this best the basic happy path of click entry point while more complex cases are being tested by functional tests? Love that!

Nit: If that's the case, can we add some comment in this file to make it a bit clear?

ChenyuLInx (approved):
LGTM

MichelleArk (commented):
We should add another command to [tox.ini](https://github.com/dbt-labs/dbt-core/blob/main/tox.ini#L28-L30) to run tests for `tests/cli` locally and in CI via tox, or consider moving `cli/test_cli_run.py` under `tests/functional` (my preference, since it's using the same underlying framework)

MichelleArk (approved):

PR #6742, opened 2023-01-26
Use IEC standard abbreviations (GiB, TiB, etc)

### LINKED ISSUE #6741: [CT-1911] [Bug] Non-standard abbreviations for kibibyte, etc

### Context

This is a copy of https://github.com/dbt-labs/dbt-bigquery/issues/477 since there is an (unused!) copy of the relevant code in dbt-core as well.

### Current Behavior

Described in https://github.com/dbt-labs/dbt-bigquery/issues/477

### Expected Behavior

Described in https://github.com/dbt-labs/dbt-bigquery/issues/477

### Steps To Reproduce

Described in https://github.com/dbt-labs/dbt-bigquery/issues/477

### Relevant log output

_No response_

### Environment

```markdown
- OS:
- Python:
- dbt:
```

_No response_

### Additional Context

_No response_

dbeatty10:
resolves #6741
equivalent to https://github.com/dbt-labs/dbt-bigquery/pull/482

### Description

When using multiples of 1024, the IEC standard abbreviations are "GiB", "TiB", "PiB" instead of "GB", "TB", "PB" (which are for multiples of 1000).

### More info
Today I Learned (TIL), `format_bytes` (and `format_rows_number`) are defined in dbt-core but redefined in dbt-bigquery and only used in that adapter. So the dbt-core version is unused to best I can tell.

#### Context for the duplication
Removing them from dbt-core was mentioned [here](https://github.com/dbt-labs/dbt-bigquery/pull/48#discussion_r742152754) and [here](https://github.com/dbt-labs/dbt-bigquery/pull/48#discussion_r743485981) as part of the discussion for #48.

It feels to me like one of the following should have happened along with https://github.com/dbt-labs/dbt-bigquery/pull/48:
1. a simultaneous PR should have removed `format_bytes` and `format_rows_number` from dbt-core; or
2. [#48](https://github.com/dbt-labs/dbt-bigquery/pull/48) should have been added to dbt-core instead of dbt-bigquery.

As it stands, the implementations in dbt-core are unused (to the best that we can tell), but would probably need to be maintained to preserve backwards compatibility in case some unknown adapter is using them.

#### Proposal

Going forward, I'd propose that we choose one (and only one) of the following:
1. Keep separate implementations in dbt-core and dbt-bigquery (and fix the KB vs. KiB issue in both)
1. Remove the implementations in dbt-core (so they exist only in dbt-bigquery)
1. Completely move the implementation in dbt-bigquery back into dbt-core

### Checklist

### DIFF
--- a/core/dbt/utils.py
+++ b/core/dbt/utils.py
@@ -453,26 +453,6 @@ class classproperty(object):
         return self.func(objtype)
 
 
-def format_bytes(num_bytes):
-    for unit in ["Bytes", "KiB", "MiB", "GiB", "TiB", "PiB"]:
-        if abs(num_bytes) < 1024.0:
-            return f"{num_bytes:3.1f} {unit}"
-        num_bytes /= 1024.0
-
-    num_bytes *= 1024.0
-    return f"{num_bytes:3.1f} {unit}"
-
-
-def format_rows_number(rows_number):
-    for unit in ["", "k", "m", "b", "t"]:
-        if abs(rows_number) < 1000.0:
-            return f"{rows_number:3.1f}{unit}".strip()
-        rows_number /= 1000.0
-
-    rows_number *= 1000.0
-    return f"{rows_number:3.1f}{unit}".strip()
-
-
 class ConnectingExecutor(concurrent.futures.Executor):
     def submit_connected(self, adapter, conn_name, func, *args, **kwargs):
         def connected(conn_name, func, *args, **kwargs):
--- a/core/dbt/utils.py
+++ b/core/dbt/utils.py
@@ -454,7 +454,7 @@ class classproperty(object):
 
 
 def format_bytes(num_bytes):
-    for unit in ["Bytes", "KB", "MB", "GB", "TB", "PB"]:
+    for unit in ["Bytes", "KiB", "MiB", "GiB", "TiB", "PiB"]:
         if abs(num_bytes) < 1024.0:
             return f"{num_bytes:3.1f} {unit}"
         num_bytes /= 1024.0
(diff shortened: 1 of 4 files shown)


ChenyuLInx (approved):
The updated change looks good to me! Learned a bit new things about IEC standard also

core/dbt/utils.py
```diff
@@ -454,7 +454,7 @@ def __get__(self, obj, objtype):
 
 
 def format_bytes(num_bytes):
-    for unit in ["Bytes", "KB", "MB", "GB", "TB", "PB"]:
+    for unit in ["Bytes", "KiB", "MiB", "GiB", "TiB", "PiB"]:
```
jtcohen6, line 457: I don't think we actually use this method anywhere in `dbt-core`. It's a holdover from when it was added, for the sole benefit of `dbt-bigquery`, and we forgot to remove it during the Great Adapter Split-Apart of 2021.

What do you think about just removing it?

dbeatty10: > What do you think about just removing it?

Removed!

I did a quick cursory check for both `format_bytes` and `format_rows_number` in the dbt Labs adapters using a GitHub search like [this](https://github.com/dbt-labs/dbt-redshift/search?q=format_rows_number&type=code). DIdn't see any references other than in dbt-bigquery, which we know about.

I didn't check any non-dbt Labs adapters, although our expectation is that it is not being used. If it _did_ happen to be used by some adapter out there, they'd be on their own to resolve it.

jtcohen6 (approved):
lgtm aussi

dbeatty10:
Les remerciements @ChenyuLInx and @jtcohen6

Merging now.

PR #6764, opened 2023-01-26
Add clearer directions for custom test suite vars in Makefile.

### LINKED ISSUE #6689: [CT-1874] [Bug] Erroneous Makefile variable causing `make integration` to fail

### Current Behavior

My friend and coworker @seub has followed the Contributing guide for setting up the development environment for DBT-core, but repeatedly failed the `make integration` pytest suite. Many of the error messages referenced the fact that some environment variables weren't defined (the environment variables provided in the $CI_FLAGS variable). Looking into lines 66-69 of the makefile, there's this conditional statement contributed by @VersusFacit [one or two commits ago](https://github.com/dbt-labs/dbt-core/commit/1feeb804f44230cfadda8d910130b4d53646a233):

```
.PHONY: integration
integration: .env ## Runs postgres integration tests with py-integration
	@\
	$(if $(USE_CI_FLAGS), $(CI_FLAGS)) $(DOCKER_CMD) tox -e py-integration -- -nauto
```

Because the variable $USE_CI_FLAGS is not defined anywhere, the conditional fails and the CI_FLAGS are not provided to the DOCKER_CMD. After removing the if statement and simply providing the $CI_FLAGS unconditionally, `make integration` succeeded.

### Expected Behavior

I expected the `make integration` test to succeed after following the Contributing guide.

### Steps To Reproduce

Create a fresh dev environment for dbt-core and try running `make integration` off the main branch.

### Relevant log output

_No response_

### Environment

```markdown
- OS: macOS 12.2 (M1)
- Python: 3.9.10
```

_No response_

### Additional Context

Suggested solutions:

1. Remove the conditional from the Makefile (unconditionally providing the CI_FLAGS), line 69 becomes:
```
$(CI_FLAGS) $(DOCKER_CMD) tox -e py-integration -- -nauto
```

2. Update the contributing guide to use a different make tag than `integration`, at least until this issue is resolved.

VersusFacit:
resolves #6689 

### Description

The makefile is being used by community members after all! Hooray! Except, it's unclear how to use it. Booo.

Fixing that up:
* more explicit comments
* a dedicated spot in the Makefile for users to add "overrides"

#### History

I remember now! We did it this way because at the time it was super unclear how to inline an if-else in 1976's GNU Make. Well, I found a jacked up way to do it with `filter`!

### Checklist


VersusFacit:
For testing this, I made sure with `echo` that the command was being rendered properly.

image
1. `make integration`
2. `USE_DEFAULT_FLAGS=true make integration`
2. `USE_DEFAULT_FLAGS=false make integration`

VersusFacit:
image
Additional test case since I realize I should show how the new system works when you've commented the custom ci flags var in the Makefile.

dbeatty10 (commented):
### Today I Learned
Reading [here](https://stackoverflow.com/questions/24263291/define-a-makefile-variable-using-a-env-variable-or-a-default-value), it looks like the `make` command allows for overriding variables!

Consider a `Makefile` defined like:
```
DBT_TEST_USER_1 := "dbt_test_user_1"
DBT_TEST_USER_2 := "dbt_test_user_2"
DBT_TEST_USER_3 := "dbt_test_user_3"
RUSTFLAGS := "-D warnings"
LOG_DIR := "./logs"
DBT_LOG_FORMAT := "json"

CI_FLAGS =\
	DBT_TEST_USER_1=$(DBT_TEST_USER_1)\
	DBT_TEST_USER_2=$(DBT_TEST_USER_2)\
	DBT_TEST_USER_3=$(DBT_TEST_USER_3)\
	RUSTFLAGS=$(RUSTFLAGS)\
	LOG_DIR=$(LOG_DIR)\
	DBT_LOG_FORMAT=$(DBT_LOG_FORMAT)

test:
	@echo $(CI_FLAGS)
```

Then run:
```shell
make test
```

The output is:
```
DBT_TEST_USER_1=dbt_test_user_1 DBT_TEST_USER_2=dbt_test_user_2 DBT_TEST_USER_3=dbt_test_user_3 RUSTFLAGS=-D warnings LOG_DIR=./logs DBT_LOG_FORMAT=json
```

Then run this instead:
```shell
make test LOG_DIR=/tmp DBT_LOG_FORMAT=text
```

The output changes accordingly:
```
DBT_TEST_USER_1=dbt_test_user_1 DBT_TEST_USER_2=dbt_test_user_2 DBT_TEST_USER_3=dbt_test_user_3 RUSTFLAGS=-D warnings LOG_DIR=/tmp DBT_LOG_FORMAT=text
```

### Suggestion

If we use the variable definitions from the `Makefile` above, then we can just use the following in both places where it is relevant (without if statements or a filter or a new `USE_DEFAULT_CI_FLAGS` variable):
```
$(CI_FLAGS) $(DOCKER_CMD) tox -e py-integration -- -nauto
```

This would allow the user to override any variables when desired without needing to make any edits within the `Makefile`.

Interested to hear your feedback on this suggestion!

VersusFacit:
Yeah, that's one way to do it. With so many though, and thinking how unix environment variable overriding might not be the most intuitive thing, that dbt users might prefer a way to just load up their Makefile and forget about it rather than reconstructing the massive env_var string on the command line every time. 🤷‍♀️ Like what you said is true, and I wish I would have said in the PR originally that I had considered this.

But suffice to say, my guess was that wasn't the right solution for here since we like to template out everything for users.

If we trust that anyone using the Makefile is a power user who can be bothered to futz around with Bash and write their own scripts/aliases (not sure I'm confident about that), then sure, I'd prefer what you wrote.

VersusFacit:
Also what you wrote uses `:=` which is fine, but we'd have to explain to people how that works since the standard is to set env vars ahead of the executable; which requires use of the `?=` operator. 

tl;dr let's decide upon a use of syntax that is consistent across all tooling.

Makefile
```diff
@@ -6,18 +6,29 @@ ifeq ($(USE_DOCKER),true)
 	DOCKER_CMD := docker-compose run --rm test
 endif
 
-LOGS_DIR := ./logs
```
VersusFacit, line 9: Collapsed this variable into the `LOG_DIR` variable used in our CI

VersusFacit:
Alright @dbeatty10 what about this scheme? Trying for best of all worlds, with more complicated code to support a user-friendly, no expertise required override entrypoint (`makefile.text.env`).

When it's empty or not made, here's the result of `CI_FLAGS`:
image

With an override:
image

Also, don't mind missing quotes. They are present in actual jobs. No I don't understand why it displays it that way 😭

CONTRIBUTING.md
```diff
@@ -155,6 +155,21 @@ make integration
...
+DBT_TEST_USER_1=
+DBT_TEST_USER_2=
+DBT_TEST_USER_3=
+RUSTFLAGS=
+LOG_DIR=
+DBT_LOG_FORMAT=
+EOF
+```
+
+Note: This `makefile.test.env` file is git-ignored, but do not check this into the repo.
```
stu-k, line 171: If this file is git ignored it shouldn't matter if the file is checked into the repo, I would remove this line as it may cause confusion.

stu-k (approved):
One nonblocking comment on the contributing doc, otherwise nothing stands out. I'd recommend getting a 👍 from someone other than myself before merging as I am not familiar with our Makefile or CI flags.

CONTRIBUTING.md
VersusFacit: Right on!

Makefile
```diff
@@ -6,18 +6,26 @@ ifeq ($(USE_DOCKER),true)
...
+#
+# To override CI_flags, create a file at this repo's root dir named `makefile.test.env`. Fill it
```
VersusFacit, line 10: Doug and I decided for now it's better to just have this note here, since anyone interested in overriding the Makefile would be the type of person who needs this information. Better not to clutter the `CONTRIBUTING.md` with very specific information.

dbeatty10 (approved):
🚀 Thanks for your work on this @VersusFacit !

PR #6787, opened 2023-01-31
[CT-1947] Alias `--models` to `--select` for all commands except `dbt ls`

### LINKED ISSUE #6782: --models
Per #6774
@ChenyuLInx I figured it out! The tests fail for me locally, and then pass when I switch `-m` to `-s` in the arguments to `run_dbt` (https://github.com/dbt-labs/dbt-core/commit/83850f51214c0b61aaf4ecd0b0b4123acced392f)! It seems like this might be a genuine bug where we're not passing through `--models`/`-m` appropriately. That's important for backward compatibility.

**Desirable behavior:**
- For almost all commands, `--models` should be a pure alias for `--select` (and the two should be mutually exclusive)
- EXCEPTION: `dbt list` historically accepted _both_ `--select` and `--models`, with slightly different behavior, [which we've preserved for back compat](https://github.com/dbt-labs/dbt-core/blob/a8abc496323f741d3218d298d5d2bb118fa01017/core/dbt/task/list.py#L153-L177)

aranke:
resolves #6782 

### Checklist

### DIFF
--- a/tests/unit/test_cli.py
+++ b/tests/unit/test_cli.py
@@ -1,8 +1,5 @@
-import ast
-from inspect import getsource
-
 import click
-from dbt.cli import params
+
 from dbt.cli.main import cli
 
 
@@ -42,10 +39,3 @@ class TestCLI:
                     run_test(command)
 
         run_test(cli)
-
-    def test_params_are_alpha_sorted(self):
-        root_node = ast.parse(getsource(params))
-        param_var_names = [
-            node.targets[0].id for node in ast.walk(root_node) if isinstance(node, ast.Assign)
-        ]
-        assert param_var_names == sorted(param_var_names)


core/dbt/cli/params.py
```diff
@@ -274,14 +264,22 @@
...
+    "type": tuple,
+}
+
+# `--select` and `--models` are analogous for most commands except `dbt list` for legacy reasons.
+# Most CLI arguments should use the combined `select` option that aliases `--models` to `--select`.
+# However, if you need to split out these separators (like `dbt ls`), use the `models` and `raw_select` options instead.
+# See https://github.com/dbt-labs/dbt-core/pull/6774#issuecomment-1408476095 for more info.
+models = click.option(*model_decls, **select_attrs)
+raw_select = click.option(*select_decls, **select_attrs)
+select = click.option(*select_decls, *model_decls, **select_attrs)
```
ChenyuLInx, line 282: Nit: this is simpler code but slightly harder to read compared to just write the options here.

ChenyuLInx (approved):
LGTM.

MichelleArk:
Did some manual testing by locally overwriting a command that accepts the new `select` (--select, --model) alias command with `print(ctx.obj["flags"].SELECT)` and found some issues when passing `--select` and `--model` to a command together. 

The options are not treated as mutually exclusive via MultiOption and the resulting `flags.SELECT` value is clobbered by the value from the last alias option:
```
❯ python3 core/dbt/cli/main.py docs generate --select model_a model_b  --model model_c
('model_c',)

❯ python3 core/dbt/cli/main.py docs generate --model model_c --select model_a model_b
('model_a', 'model_b')
```

This was unexpected behaviour to me, but for better or for worse this is also how main's parsing currently works (printing args.select and exiting on main to test): 
```
❯ dbt docs generate --select model_a model_b --model model_c
['model_c']
❯ dbt docs generate --select model_a model_b --model model_c
['model_c']
❯ dbt docs generate --model model_c --select model_a model_b 
['model_a', 'model_b']
```

We should probably open an issue to document this behaviour and investigate a fix that gets us --select/--model option mutual exclusivity if possible. But given that it's the current behaviour on main I don't see it as pressing to fix as part of this PR.

MichelleArk (approved):

PR #6788, opened 2023-01-31
consolidate flags

### LINKED ISSUE #6701: [CT-1885] [Feature] Remove flags.py
### Describe the feature

Moving to the new Flags module instead of old flags.py

## Things to watch out for:
* dbt cloud would import this module [link](https://github.com/dbt-labs/dbt-cloud/blob/master/sinter/services/connection.py#L547)
* a few places where we would read env vars in while starting the dbt process
  https://github.com/dbt-labs/dbt-core/blob/7a6160273875853525c6709cc196b4835617530a/core/dbt/flags.py#L116-L121
  https://github.com/dbt-labs/dbt-core/blob/cbb9117ab997289062cd5bacaf4cfd4466ab9b1e/core/dbt/logger.py#L597-L601
* For there are functions that is buried pretty deep
  https://github.com/dbt-labs/dbt-core/blob/4480d05cfbe211bb78cdac721ff2664d0a8fc6a7/core/dbt/ui.py#L21-L35

ChenyuLInx:
resolves #6701 
This moves the remaining default to cli/flags.py
We still have a few being accessed by legacy logger that remains in flags module.

# TODO
- [x] with the new dbtRunner, we would initialize the global flag object again when running the actually dbt command. This would make previously set `project_dir` and `profiles_dir` at [link](https://github.com/dbt-labs/dbt-core/blob/d6ac340df09762040ef737c0463c418b78be5818/core/dbt/tests/fixtures/project.py#L263) not carried over and cause some additional tests fail.


core/dbt/cli/flags.py
```diff
@@ -17,11 +17,28 @@
...
+}
+
 
 @dataclass(frozen=True)
 class Flags:
     def __init__(self, ctx: Context = None, user_config: UserConfig = None) -> None:
 
+        # set the default flags
+        for key, value in flag_defaults.items():
+            object.__setattr__(self, key, value)
```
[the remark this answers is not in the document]
ChenyuLInx: Yes, I was chatting with @MichelleArk  about similar thing also. @iknox-fa  might have something around it, but we are gonna defer that to work later on instead of in this PR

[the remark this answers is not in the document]
iknox-fa: I didn't do that the first time around because I wanted to keep things as immutable as python allows so adding even a "private" method to do that seemed like a recipe for mutated flags in the codebase in a few months.

Maybe as a compromise for readability we just make nested `get` and `set` functions in __init__?  That way no-one uses them unless they're directly editing the flags? We can discuss it further but given how much of a tangled mess the flags are currently I think we're going to get a lot of mileage out of having reasonable assurance of immutability and I don't want to jeopardize that.

core/dbt/cli/flags.py
```diff
@@ -12,27 +12,61 @@
...
 from dbt.contracts.project import UserConfig
+from dbt.cli.option_types import WarnErrorOptionsType
 
 if os.name != "nt":
     # https://bugs.python.org/issue41567
     import multiprocessing.popen_spawn_posix  # type: ignore  # noqa: F401
 
+# TODO anything that has a default in params should be removed here?
+# Or maybe only the ones that's in the root click group
+flag_defaults = {
```
MichelleArk, line 23: nit: constants should be uppercase by convention

core/dbt/cli/flags.py
```diff
@@ -63,6 +97,12 @@ def assign_params(ctx, params_assigned_from_default):
             for param_assigned_from_default in params_assigned_from_default:
                 user_config_param_value = getattr(user_config, param_assigned_from_default, None)
                 if user_config_param_value is not None:
+                    # TODO Conversion should happen here to make sure types are consistent with the flag
+                    # extract into a function
+                    if param_assigned_from_default in ["warn_error_options"]:
+                        user_config_param_value = WarnErrorOptionsType().convert(
+                            user_config_param_value, None, None
```
MichelleArk, line 104: `warn_error_options` is a [dictionary](https://github.com/dbt-labs/dbt-core/blob/main/core/dbt/contracts/project.py#L252) when loaded via `UserConfig`, but `WarnErrorOptionsType().convert` expects a string, which leads to runtime errors when `warn_error_options` is set in the user config. 

For example: 
```yaml
#profiles.yml
config:
  warn_error_options:
    exclude:
    - PackageInstallPathDeprecation
    include: all
```

```sh
❯ python3 core/dbt/cli/main.py run --project-dir ~/src/jaffle_shop
Usage: main.py run [OPTIONS]
Try 'main.py run -h' for help.

Error: Invalid value: Cannot load YAML from type <class 'dict'>
```

I think our options would be to either: 
1. Pass `'include'` and `'exclude'` keys from from `user_config_param_value` dict to tthe `WarnErrorOptions` constructor directly instead of going through `WarnErrorOptionsType`, or 
2. convert `user_config_param_value` to a string representation before passing it to `WarnErrorOptionsType().convert`

My preference is the first option so we don't go through an unnecessary series of str -> dict -> str conversions assuming there are no circular dependency issues in using `WarnErrorOptions` directly.

ChenyuLInx: Adjusted things via the first way and moved conversion for original args and user config to a single place.

core/dbt/cli/flags.py
```diff
@@ -12,27 +12,61 @@
...
+                        "version_check",
+                        "fail_fast",
+                        "indirect_selection",
+                        "store_failures",
+                    ]:
+                        raise Exception(
+                            f"Duplicate flag names found in click command: {param_name}"
+                        )
+                    else:
+                        if param_name not in params_assigned_from_default:
```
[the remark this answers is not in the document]
ChenyuLInx: fixed with new logic, slightly cleaner, I would like to defer the duplicate param issue to a new issue, maybe after merge. @iknox-fa  thoughts?

core/dbt/config/profile.py
```diff
@@ -181,6 +181,11 @@ def pick_profile_name(
         args_profile_name: Optional[str],
         project_profile_name: Optional[str] = None,
     ) -> str:
+        def default_profiles_dir():
```
MichelleArk, line 184: add comment

tests/unit/test_dbt_runner.py
```diff
@@ -5,7 +5,7 @@
 
 class TestDbtRunner:
     @pytest.fixture
-    def dbt(self) -> dbtRunner:
```
ChenyuLInx, line 8: This feels like gonna introduce some overhead of setting up the project to unittests?

MichelleArk (approved):

PR #6374, opened 2022-12-03
Remove PR from most changelog kinds

emmyoop:
resolves #6332 

### Description

Removes changelog from all but `Dependency` and `Security` kinds.  Removes issue from those.

Allows for multiple issues (or PRs depending on kind) to be entered for a single change.

Note: I also found that the `changelog-bot` was using the `Dependency` kind while the `.changie.yaml` was using `Dependencies`.  I synced them to be `Dependencies`.

### Checklist

- [ ] I have read [the contributing guide](https://github.com/dbt-labs/dbt-core/blob/main/CONTRIBUTING.md) and understand what's expected of me
- [ ] I have signed the [CLA](https://docs.getdbt.com/docs/contributor-license-agreements)
- [ ] I have run this code in development and it appears to resolve the stated issue
- [ ] This PR includes tests, or tests are not required/relevant for this PR
- [ ] I have [opened an issue to add/update docs](https://github.com/dbt-labs/docs.getdbt.com/issues/new/choose), or docs changes are not required/relevant for this PR
- [ ] I have run `changie new` to [create a changelog entry](https://github.com/dbt-labs/dbt-core/blob/main/CONTRIBUTING.md#adding-a-changelog-entry)

PR #6808, opened 2023-01-31
[BACKPORT] fix contributor list generation (#6799)

emmyoop:
Backport #6374 #6799 

Necessary to allow other back ports to work without conflicts. Since PR is no longer required on main, when a change is back ported it will not have the PR in the changelog yaml and will break changelog generation.


VersusFacit (approved):
I'd say LGTM but that code is obtuse, wow! Bet you had "fun" trudging through that.

PR #6831, opened 2023-02-01
[CT-1976] [Feature] Allow setting postgres port for integration tests

### LINKED ISSUE #6794: [Feature] Provide additional Jinja tests on top of the built-in ones

### Describe the feature

Jinja comes with [built-in tests](https://jinja.palletsprojects.com/en/3.1.x/templates/#list-of-builtin-tests), but it is also possible to create custom ones.

When inspecting objects with Jinja in dbt a few extra tests could be useful to make the code shorter and more readable.

The following code inspects the graph to search for tests that depend on at lease one node.
```jinja
{{ graph.nodes.values() 
    | selectattr("config.materialized","==","test") 
    | selectattr("depends_on.nodes") 
    | list | tojson(indent=1)
}}
```

If I want to search for tests applying to a given node I then need to loop through the results. If we had a test called `contains`, we could do something like 

```jinja
{{ graph.nodes.values() 
    | selectattr("config.materialized","==","test") 
    | selectattr("depends_on.nodes","contains","model.project.my_model") 
    | list | tojson(indent=1)
}}
```
which would return the list of tests that are applied to `my_model`.

The list of tests I can think of for now would be:
- contains: checking if a list contains a given string
- contains_substring: checking if a string contains a substring
- matches: checking if a string matches a given regex
- startswith: checking if a string starts by a specific substring
- endswith: checking if a string ends by a specific substring

We can already achieve all the outcome by writing longer and more complex Jinja code. This feature would make it 
- easier to understand what some Jinja code does 
- and easier to write it in the first place

### Who will this benefit?

People writing Jinja macros as part of packages or custom logic

Yes

### Anything else?

_No response_

### LINKED ISSUE #6795: [Feature] Allow setting postgres port for integration tests

### Describe the feature

Currently, it is not possible to configure the port used by postgres in either `make setup-db` or `make integration`. It would be very useful if one could override the default port e.g. by passing a make parameter or using an environment variable, so one can run dbt integration tests without interfering with a postgres instance one may already have installed locally.

After a quick check, this looks like a relatively small feature to implement. There is already support for a `POSTGRES_TEST_PORT` environment variable in almost all of the test suite, so mainly small changes to the Makefile and docker-compose.yml would be required.

*Leaving as is.* This works, just causes a bit of extra friction: stop local postgres before starting work on dbt-core, then restart the local postgres after done.

### Who will this benefit?

People wanting to contribute to `dbt-core` who already have postgres installed locally

Yes

### Anything else?

_No response_

flexponsive:
resolves #6795
### Description

Allows running the test-suite against a postgres database running on a port other than the default 5432 by adding a make parameter:

```bash
make setup-db POSTGRES_TEST_PORT=6100
make integration POSTGRES_TEST_PORT=6100
```

If no parameter is specified, the default port is specified in the Makefile.

Comments:

- On my shell (bash 5.1 on ubuntu 22.04) I received an error relating to CI_FLAGS, which I fixed in 72ec289d87958f721a5f76fb3fdfb78aca4b6f1c

- Consider changing the default port for the integration test database to something other than 5432, to eliminate clashes with a locally installed postgres altogether?

### Checklist


tests/functional/hooks/test_model_hooks.py
```diff
@@ -165,7 +164,7 @@ def models(self):
 
     def test_pre_and_post_run_hooks(self, project, dbt_profile_target):
         run_dbt()
-
+        print(dbt_profile_target)
```
flexponsive, line 167: TODO remove overlooked print

PR #6838, opened 2023-02-01
CT 2000 fix semver prerelease comparisons

### LINKED ISSUE #6834: [CT-2000] Fix broken semver comparison logic
_Originally posted by @jtcohen6 in https://github.com/dbt-labs/dbt-core/pull/6463#discussion_r1064541063_

_adapted from Slack: @jtcohen6 + @gshank on 6 Jan_

https://github.com/dbt-labs/dbt-core/blob/1a6e4a00c7be2ea88492192de51d190a46c19561/core/dbt/semver.py#L145-L151

There are two things at play here, leading to failures in [`test/unit/test_semver.py`](https://github.com/dbt-labs/dbt-core/blob/main/test/unit/test_semver.py):
1. This logic is not quite right! We're passing just _part_ of a version string into `packaging.version.parse()`, rather than the whole string. This is the cause of some of the failing tests.
2. Unfortunately, there are also some real discrepancies between [PEP 440](https://peps.python.org/pep-0440) and [SemVer](https://semver.org/), and `packaging==22.0` lands firmly on the side of PEP 440. This is the cause of the rest of the failing tests.

These versions are still supported by `packaging==22.0`. (Although PEP 440 says not to use a hyphen, and SemVer says to use a hyphen.) The failure is because we're trying to parse just the prerelease identifier (a1, b2) as if it's an entire version string:
- `'1.2.0a1'`
- `'1.2.0-a1'`
- `'1.2.0b2'`
- `'1.2.0-b2'`

These are valid per SemVer, but they are not valid per PEP 440, and so they fail on `packaging.version.parse()` starting with `packaging==22.0`:
- `'2.2.0-fishtown-beta'`
- `'2.2.0asdf'`

https://peps.python.org/pep-0440/#pre-releases
> The pre-release segment consists of an alphabetical identifier for the pre-release phase, along with a non-negative integer value. Pre-releases for a given release are ordered first by phase (alpha, beta, release candidate) and then by the numerical component within that phase.

https://semver.org/#spec-item-9
> A pre-release version MAY be denoted by appending a hyphen and a series of dot separated identifiers immediately following the patch version. Identifiers MUST comprise only ASCII alphanumerics and hyphens [0-9A-Za-z-].

@gshank
> It would be nice to not pin "packaging" because of it... Maybe at this point we should defer this to a re-write

gshank:
resolves #6834

### Description

In semver.py we were using packaging.version.parse to parse the prerelease portion of versions, but since that version parsing is incompatible with SemVer, it started failing once the LegacyVersion fallback was removed.

This removes the dependency on packaging and subsitutes some comparison code specifically for prepreleases, so that we can remove the pin on packaging.

### Checklist

### DIFF
--- a/.changes/unreleased/Fixes-20230201-154418.yaml
+++ b/.changes/unreleased/Fixes-20230201-154418.yaml
@@ -1,0 +1,6 @@
+kind: Fixes
+body: Remove pin on packaging and stop using it for prerelease comparisons
+time: 2023-02-01T15:44:18.279158-05:00
+custom:
+  Author: gshank
+  Issue: "6834"
--- a/core/dbt/semver.py
+++ b/core/dbt/semver.py
@@ -1,9 +1,6 @@
 from dataclasses import dataclass
 import re
-import warnings
 from typing import List
-
-from packaging import version as packaging_version
 
 from dbt.exceptions import VersionsNotCompatibleError
 import dbt.utils
@@ -68,6 +65,11 @@ $
 )
 
 _VERSION_REGEX = re.compile(_VERSION_REGEX_PAT_STR, re.VERBOSE)
+
+
+def _cmp(a, b):
+    """Return negative if a<b, zero if a==b, positive if a>b."""
+    return (a > b) - (a < b)
 
 
 @dataclass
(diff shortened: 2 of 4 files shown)


dbeatty10 (approved):
#### Highlights
- 🎉  This looks like it does as described and removes the dependency on `packaging.version.parse()`
- 💡 It seems like we should still swing back later and try to reduce our need for custom [PEP 440](https://peps.python.org/pep-0440) and [SemVer](https://semver.org/) parsing and comparison logic 

#### More detail
I double-checked that CI indeed used `packaging>=23.0`:
https://github.com/dbt-labs/dbt-core/actions/runs/4068580210/jobs/7007304928#step:4:55

I gave a reading of the logic in `compare()` and `_nat_cmp()` and `cmp_prerelease_tag()`, and I didn't notice anything amiss or a key case not covered.

My only nits are the variable names `prcmp` (for "prerelease comparison"?) and `_nat_cmp` (for "natural comparison"?) felt a little terse and not 100% obvious. But given the time-sensitive nature of your fixes here, these didn't feel quite important enough to block or even raise as formal suggestions.

✅ Ultimately, it looks like the key test cases are covered by `test__filter_installable`.

Side note: I'm guessing `black` doesn't run against code in the `test` directory, and I won't raise any nits about white space formatting within `test/unit/test_semver.py`.

gshank:
We should definitely look at a more consistent solution to these version problems, and look into using some existing package.

jtcohen6:
Let's include this in the next v1.4.x patch release as well, since users will continue to have issues installing `dbt-core` with the tight pin. The change was small & precise, and fixed a narrow bug (the previous logic was genuinely incorrect).

PR #6839, opened 2023-02-01
update regex to match all iterations

emmyoop:
### Description
The existing regex (new since nightly releases were added) had a bug and did not recognize versions without any kind of prerelease (ie. `1.5.0`).  

This change passes for the following
- `1.5.0`
- `1.5.0a1`
- `1.5.0a1.dev123457+nightly`

and fails for
- `1.5`
- `1`
- `1.5.2-a1`

Tested with release [here](https://github.com/dbt-labs/dbt-core/actions/runs/4069752682)

### Checklist

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,8 +1,11 @@
 [bumpversion]
 current_version = 1.5.0a1
 
+# `parse` allows parsing the version into the parts we need to check.  There are some
+# unnamed groups and that's okay because they do not need to be audited.  If any part
+# of the version passed in does not match the regex, it will fail.
 # expected matches: `1.5.0`, `1.5.0a1`, `1.5.0a1.dev123457+nightly`
-# excepted failures: `1`, `1.5`, `1.5.2-a1`
+# excepted failures: `1`, `1.5`, `1.5.2-a1`, `text1.5.0`
 parse = (?P<major>[\d]+) # major version number
 	\.(?P<minor>[\d]+) # minor version number
 	\.(?P<patch>[\d]+) # patch version number
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,12 +1,14 @@
 [bumpversion]
 current_version = 1.5.0a1
-parse = (?P<major>\d+)
-	\.(?P<minor>\d+)
-	\.(?P<patch>\d+)
-	((?P<prekind>a|b|rc)
-	(?P<pre>\d+)  # pre-release version num
-	)(\.(?P<nightly>[a-z..0-9]+)
+parse = (?P<major>[\d]+)
+	\.(?P<minor>[\d]+)
+	\.(?P<patch>[\d]+)
+	(((?P<prekind>a|b|rc)
+	?(?P<pre>[\d]+?))
+	\.?(?P<nightly>[a-z..0-9]+\+[a-z]+)?
 	)?
+
+
 serialize =
 	{major}.{minor}.{patch}{prekind}{pre}.{nightly}
 	{major}.{minor}.{patch}{prekind}{pre}
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,11 +1,14 @@
 [bumpversion]
 current_version = 1.5.0a1
-parse = (?P<major>[\d]+)
-	\.(?P<minor>[\d]+)
-	\.(?P<patch>[\d]+)
-	(((?P<prekind>a|b|rc)
-	?(?P<num>[\d]+?))
-	\.?(?P<nightly>[a-z..0-9]+\+[a-z]+)?
+
+# expected matches: `1.5.0`, `1.5.0a1`, `1.5.0a1.dev123457+nightly`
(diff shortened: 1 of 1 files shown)


emmyoop:
Once this is merged it needs to be added to all the core release backports (#6792, #6791, #6790 ) and also backported to `1.4.latest` since that one was already merged.

.bumpversion.cfg
```diff
@@ -1,12 +1,14 @@
...
-	\.(?P<patch>\d+)
-	((?P<prekind>a|b|rc)
-	(?P<pre>\d+)  # pre-release version num
-	)(\.(?P<nightly>[a-z..0-9]+)
+parse = (?P<major>[\d]+)
+	\.(?P<minor>[\d]+)
+	\.(?P<patch>[\d]+)
+	(((?P<prekind>a|b|rc)
+	?(?P<pre>[\d]+?))
+	\.?(?P<nightly>[a-z..0-9]+\+[a-z]+)?
```
nathaniel-may, line 8:
```suggestion
	\.?(?P<nightly>[a-z0-9]+\+[a-z]+)?
```
The two dots in this character class definition are literals, so this would match `1.5.0a1.dev123.457+nightly` as well. If you'd like to match on dots, remove the duplicate.

nathaniel-may (commented):
Only found one true issue.

It could be worth commenting each line for the next person who might have to edit this. Also adding the examples in the PR description as a comment could help as well.

Two additional notes:
- depending on how this matching is applied you may want to have start and stop anchors (`^`, `$`) to prevent matching in the middle of bad strings. e.g. - `1.2.3.4.5.6` matches twice: once on `1.2.3` and again on `4.5.6`
- There are also two unnamed groups. Namely the two groups quantified by `?` that for the string `1.5.0a1.dev123457+nightly` would match `a1.dev123457+nightly` and `a1`. I don't suspect this is an issue, but I wanted to make sure you were aware it still matches as an unnamed group since every other group is named.

emmyoop:
@nathaniel-may 

> depending on how this matching is applied you may want to have start and stop anchors (^, $) to prevent matching in the middle of bad strings. e.g. - 1.2.3.4.5.6 matches twice: once on 1.2.3 and again on 4.5.6

bumpversion doesn't need this.  If there is anything that falls outside the match it will fail.  Tested [here](https://github.com/dbt-labs/dbt-core/actions/runs/4076997736/jobs/7025462367)

> There are also two unnamed groups. Namely the two groups quantified by ? that for the string 1.5.0a1.dev123457+nightly would match a1.dev123457+nightly and a1. I don't suspect this is an issue, but I wanted to make sure you were aware it still matches as an unnamed group since every other group is named.

It is not an issue.  I've added a comment to clarify in the code.

.bumpversion.cfg
```diff
@@ -1,15 +1,18 @@
...
-	)(\.(?P<nightly>[a-z..0-9]+)
+
+# expected matches: `1.5.0`, `1.5.0a1`, `1.5.0a1.dev123457+nightly`
+# excepted failures: `1`, `1.5`, `1.5.2-a1`
+parse = (?P<major>[\d]+) # major version number
+	\.(?P<minor>[\d]+) # minor version number
+	\.(?P<patch>[\d]+) # patch version number
+	(((?P<prekind>a|b|rc) # optional pre-release type
+	?(?P<num>[\d]+?)) # optional pre-release version number
+	\.?(?P<nightly>[a-z0-9]+\+[a-z]+)? # optional nightly release indicator
```
colin-rogers-dbt, line 14: would need to give this some thought but I think we want `1.5.0a1` to be valid but `1.5.0a` to be invalid right?

nathaniel-may: the `+` in `?(?P<num>[\d]+?))` should handle that.

.bumpversion.cfg
```diff
@@ -1,15 +1,21 @@
...
-parse = (?P<major>\d+)
-	\.(?P<minor>\d+)
-	\.(?P<patch>\d+)
-	((?P<prekind>a|b|rc)
-	(?P<pre>\d+)  # pre-release version num
-	)(\.(?P<nightly>[a-z..0-9]+)
+
+# `parse` allows parsing the version into the parts we need to check.  There are some
+# unnamed groups and that's okay because they do not need to be audited.  If any part
+# of the version passed in does not match the regex, it will fail.
```
nathaniel-may, line 6:
```suggestion
# of the version passed and does not match the regex, it will fail.
```

nathaniel-may (approved):
Looks good!

.bumpversion.cfg
colin-rogers-dbt: I missed the parenthesis

colin-rogers-dbt (approved):

.bumpversion.cfg
emmyoop: @colin-rogers-dbt `1.5.0a` is not valid because of the `?` in the middle.  it means what come before is optional, but if it's filled, what's after is required.
`((?P<prekind>a|b|rc)?(?P<num>[\d]+?))`

PR #4076, opened 2021-10-15
Adjust logic when finding approx matches for model or test matching

emilieschario:
resolves # https://github.com/dbt-labs/dbt-core/issues/3835

### Checklist

- [x] I have signed the [CLA](https://docs.getdbt.com/docs/contributor-license-agreements)
- [x] I have run this code in development and it appears to resolve the stated issue
- [ ] This PR includes tests, or tests are not required/relevant for this PR
- [x] I have updated the `CHANGELOG.md` and added information about my change

Issue #3835, opened 2021-08-30
Cache miss when alias includes explicit quotes, instead of configuring quoting: {identifier: true}


magnusfors:
### Describe the bug
If updating a quoted model the dbt tests are not executed against the newly built model, instead they are deferring to the models in the previous state. For regular (non quoted) models the tests run against the correct newly built models.

In the provided example dbt 0.19.1 has been used but the same issue occurs in dbt 0.20.1.

Attached is a dbt project used to reproduce the issue (dbt_bug_project.zip)

### Steps To Reproduce

0. Created some source data in Snowflake like below.
```
create schema afs_source;
create table sourcedata(col1 number);
insert into sourcedata(col1)  values(1),(2),(3);
```
1.  dbt run --target=dev
```
Running with dbt=0.19.1
Found 2 models, 2 tests, 0 snapshots, 0 analyses, 144 macros, 0 operations, 0 seed files, 1 source, 0 exposures

13:13:00 | Concurrency: 1 threads (target='dev')
13:13:00 |
13:13:00 | 1 of 2 START view model [dbt_bug_project.zip](https://github.com/dbt-labs/dbt/files/7077915/dbt_bug_project.zip).not_quoted........................... [RUN]
13:13:02 | 1 of 2 OK created view model AFS_DEV.not_quoted...................... [SUCCESS 1 in 2.20s]
13:13:02 | 2 of 2 START view model AFS_DEV."quoted"............................. [RUN]
13:13:04 | 2 of 2 OK created view model AFS_DEV."quoted"........................ [SUCCESS 1 in 1.91s]
13:13:06 |
13:13:06 | Finished running 2 view models in 12.25s.

Completed successfully

Done. PASS=2 WARN=0 ERROR=0 SKIP=0 TOTAL=2
```
2.  dbt test --target=dev
```
Running with dbt=0.19.1
Found 2 models, 2 tests, 0 snapshots, 0 analyses, 144 macros, 0 operations, 0 seed files, 1 source, 0 exposures

13:16:01 | Concurrency: 1 threads (target='dev')
13:16:01 |
13:16:01 | 1 of 2 START test sanity_check_model_not_quoted_..................... [RUN]
13:16:04 | 1 of 2 PASS sanity_check_model_not_quoted_........................... [PASS in 2.61s]
13:16:04 | 2 of 2 START test sanity_check_model_quoted_......................... [RUN]
[...]

PR #727, opened 2018-04-06
Implement relations api

cmcarthur:
WIP

PR #6844, opened 2023-02-02
Optimize BaseRelation.matches()

### LINKED ISSUE #6842: [CT-2018] Performance Bottleneck in BaseRelation.get_relation()
A profile of @boxysean's "really big project" showed that 60% of the time of a `dbt build` run is spent in BaseRelation.get_relation(). Let's look at how we can improve the performance of this function in a way with substantial end-user impact.

peterallenwebb:
resolves #6842 

### Description

This PR optimizes the BaseRelation.matches() function in order to avoid costly string processing and comparison operations which are being done a very large number of times during certain large project runs of `dbt build`. On the large scenario described in #6842, it took my local run time for `dbt build` from 23m to 14m.

As written, we would lose the ApproximateMatchError exception, since determining whether a relation is a an approximate match was a large part of the time spent. We'll need to determine whether that is justified by the savings, or if there is a better way to avoid performing the check a large number of times.

At any rate, there is a lot of room for improvement in this bottleneck.

### Checklist

### DIFF
--- a/core/dbt/adapters/base/relation.py
+++ b/core/dbt/adapters/base/relation.py
@@ -12,7 +12,6 @@ from dbt.contracts.relation import (
     Path,
 )
 from dbt.exceptions import (
-    ApproximateMatchError,
     DbtInternalError,
     MultipleDatabasesNotAllowedError,
 )
@@ -77,36 +76,24 @@ class BaseRelation(FakeAPIObject, Hashable):
         schema: Optional[str] = None,
         identifier: Optional[str] = None,
     ) -> bool:
-        search = filter_null_values(
-            {
-                ComponentName.Database: database,
-                ComponentName.Schema: schema,
-                ComponentName.Identifier: identifier,
-            }
-        )
 
-        if not search:
-            # nothing was passed in
+        if identifier is not None and not self._is_exactish_match(
+            ComponentName.Identifier, identifier
+        ):
+            return False
+
+        if schema is not None and not self._is_exactish_match(ComponentName.Schema, schema):
+            return False
+
+        if database is not None and not self._is_exactish_match(ComponentName.Database, database):
+            return False
+
+        if database is None and schema is None and identifier is None:
             raise dbt.exceptions.DbtRuntimeError(
                 "Tried to match relation, but no search path was passed!"
             )
 
-        exact_match = True
-        approximate_match = True
-
-        for k, v in search.items():
-            if not self._is_exactish_match(k, v):
-                exact_match = False
-            if str(self.path.get_lowered_part(k)).strip(self.quote_character) != v.lower().strip(
-                self.quote_character
-            ):
-                approximate_match = False  # type: ignore[union-attr]
-
-        if approximate_match and not exact_match:
-            target = self.create(database=database, schema=schema, identifier=identifier)
-            raise ApproximateMatchError(target, self)
-
-        return exact_match
+        return True
 
     def replace_path(self, **kwargs):
         return self.replace(path=self.path.replace(**kwargs))


peterallenwebb:
@boxysean If you have a chance to kick the tires on this change, please do and let me know how it works for you. If it looks good to you, I'll discuss it with other stakeholders.

jtcohen6:
@peterallenwebb This is awesome!!

## Perf bottlenecks in dbt

First, I want to offer some higher-level context for this change:
- There are some "fixed costs" at the start of every run: parsing (files &rarr; manifest), graph generation (manifest &rarr; graph), cache population (run queries to build up relational cache on `adapter`).
- All of those have to happen on the main thread, before we can do anything else, and they scale with the size of the overall project, rather than the number of nodes actually selected to run. That's particularly frustrating for developers in large projects who may just want to be running `dbt build --select one_model`. (There is [an experimental config](https://docs.getdbt.com/reference/global-configs#cache-database-objects-for-selected-resource) to modify cache population based on node selection.)
- Once all those fixed costs are paid, we move into a [queue](https://github.com/dbt-labs/dbt-core/blob/b2ea2b8b256e5db1da0b712dfedd7973e1e50a37/core/dbt/task/runnable.py#L276-L316). Now, each node can run, in parallel, on an independent thread—up to the number of `--threads`, depending on the shape of the DAG (interdependencies), and _only the subset of nodes actually selected to run._
- So: I would expect `BaseRelation.get_relation()` to run within those threads, and only for nodes that are actually selected to run. This does add up to a lot of time if running thousands of models, but I'd expect it to scale in proportion with that number of models, and to be parallelized across multiple threads. (Exception: In the case of [deferral](https://docs.getdbt.com/reference/node-selection/defer), we run [adapter.get_relation](https://github.com/dbt-labs/dbt-core/blob/b2ea2b8b256e5db1da0b712dfedd7973e1e50a37/core/dbt/contracts/graph/manifest.py#L1018) as a fixed-cost step earlier on, to determine which upstream models do/don't exist in the target namespace.)
- Finally, once we actually get into running models against the warehouse, dbt doesn't tend to be the performance bottleneck anymore. We're "I/O bound" by the time it takes to actually execute each query on the warehouse, which can be anywhere from <1s (`create view`) to many minutes (`create table`, `merge`). At this point, when users are thinking about performance, they're thinking about _database_ performance, and dbt's role is mostly around making sure it's templating the right DDL/DML, with just the right keywords or magic incantations that make the data platform purr.
- Of course, dbt can also try to avoid running duplicative queries. That's the motivation behind having the relational cache to begin with, populated at the beginning & accessed/update as we go along. That cache lookups should be much faster (overall) than running the same metadata queries, over and over, against the warehouse, while materializing every single model ("does this relation already exist?" "is it a table?").

## This PR: proposed trade-off

> As written, we would lose the ApproximateMatchError exception, since determining whether a relation is a an approximate match was a large part of the time spent. We'll need to determine whether that is justified by the savings, or if there is a better way to avoid performing the check a large number of times.

I could be open to making this change. The `ApproximateMatchError` is not particularly delightful for end users to see today. It is our attempt to surface an explicit error when we've missed matching up a user's defined relation to one in the catch, because of a _very_ subtle discrepancy in casing/quoting. The alternative, unfortunately, is even more-confusing behavior, whatever is the fallout of missing the match. As it is, though, it's already pretty hard to debug, and the UX loss of removing this extra detection does feel outweighed by the UX improvement of a significant speedup.

A little history:
- The original logic here goes all the way back to 2018: https://github.com/dbt-labs/dbt-core/pull/727
- We modified this logic ~15 months ago (https://github.com/dbt-labs/dbt-core/pull/4076) as the resolution to an issue that was quite tricky to debug (https://github.com/dbt-labs/dbt-core/issues/3835)

### A concrete example

Here's a simple case for reproducing when this exception would be helpful. On Snowflake, the default case is uppercase (both ANSI-compliant _and_ annoying), and quoted identifiers are case-sensitive. This was the bane of my existence in 2018; since then, we've disable quoting for relation identifiers by default, and it's much more pleasant.

So if I create a model like:
```sql
-- models/my_model.sql
select 1 as id
```
I `dbt run`, it templates out a SQL statement like
```
create or replace view analytics.dbt_jcohen.my_model as (select 1 as id);
```
In Snowflake, this relation has a much more boisterous name: `ANALYTICS.DBT_JCOHEN.MY_MODEL`. It's unquoted, ergo case-insensitive, ergo uppercase. What happens, though, if I [turn `quoting` on](https://docs.getdbt.com/reference/project-configs/quoting), for all dbt-created relations in my project?
```yml
# dbt_project.yml
quoting:
  identifier: true
```
Now, dbt is going to try to template a SQL statement like:
```sql
create or replace view analytics.dbt_jcohen."my_model" as (select 1 as id);
```
But we don't actually get all the way there, because first dbt populates the adapter cache, then it tries to match up my model with an entry in the cache, and it sees there's there's an _almost but not quite matching_ entry. And we stop the whole thing in its tracks, because we want to avoid an ugly scenario.
```
$ dbt run
...
11:12:25  Compilation Error in model my_model (models/my_model.sql)
11:12:25    When searching for a relation, dbt found an approximate match. Instead of guessing
11:12:25    which relation to use, dbt will move on. Please delete "ANALYTICS"."DBT_JCOHEN"."MY_MODEL", or rename it to be less ambiguous.
11:12:25    Searched for: ANALYTICS.DBT_JCOHEN.my_model
11:12:25    Found: "ANALYTICS"."DBT_JCOHEN"."MY_MODEL"
11:12:25
11:12:25    > in macro create_or_replace_view (macros/materializations/models/view/create_or_replace_view.sql)
11:12:25    > called by macro materialization_view_snowflake (macros/materializations/view.sql)
11:12:25    > called by model my_model (models/my_model.sql)
...
```
If I try the same, having checked out your branch, I don't get that exception—the model succeeds!—because we didn't get a match, and we didn't check for an _approximate_ match either. dbt successfully created a view named `analytics.dbt_jcohen."my_model"`. Of course, depending on which of these queries I run in Snowflake, I will actually be querying _a different view_:
```sql
select * from analytics.dbt_jcohen.my_model;
select * from analytics.dbt_jcohen."my_model";
```
It's a gross situation, no question. But, in keeping with what I said above, I'm not convinced that the `ApproximateMatchError` exception does a whole lot to make the situation less gross—it just shoves the grossness in the user's face, earlier and a bit more explicitly.

boxysean:
Thanks for the explanation @jtcohen6!

I'd be curious to see some real-world results, but I won't have time in the next 1-2 weeks to review due to travel. A similar analysis to [what I did here](https://github.com/dbt-labs/dbt-core/issues/6073#issuecomment-1398198015) would help us determine the impact of @peterallenwebb's proposed change. I will ask my client to see if they could support.

I'd also be curious to see some unit tests on `get_relation()` 🙈

peterallenwebb:
@jtcohen6 Yes, thanks very much for this clear explanation of where the practical performance concerns really are! I'll keep it in mind as I inevitably continue to tinker with performance. I definitely don't feel strongly about getting this change in if it is unlikely to make an impact under real world conditions.

jtcohen6:
@peterallenwebb If this proves to be a significant perf boost in an "in-the-wild" scenario, I'd be supportive of moving forward! Sounds like the next step here is testing with a real large project. We do have one of these for our own internal analytics :)

peterallenwebb:
@jtcohen6 @boxysean I'm probably going to close this particular PR for now, since there are risks to any effective improvement, and we have not demonstrated that it is needed in the field.

That said, I have some interesting parting observations...

Our current strategy for looking up relations in the cache is compute-intensive, since it has to account for all the ways a relation name might be quoted or cased. Our strategy also takes time linear in the number of tables in the cache. The entire list of tables in the cache is scanned for a match every time a relation is looked up. A lookup in a cache with 1000 entries will be ten times slower than one with 100 on average. With some development effort we could make this a constant-time lookup with much lower overhead.

It's not clear how important this bottleneck is in real-world production scenarios, but the anonymized client project which @boxysean provided me has the following runtimes for `dbt compile` on my local machine:

- Stock dbt: 23 minutes
- With the optimization from this PR: 14 minutes
- With a quick/dirty implementation of constant-time lookup: 7 minutes

As @jtcohen6 has pointed out to me, our multithreading model might blunt the impact of the compute savings in real-world scenarios. It's still interesting how much overhead is being spent on this operation, though.

jtcohen6:
> It's still interesting how much overhead is being spent on this operation, though.

Agree that it's very interesting. If you think the right next step is to close this specific PR for now, given some unknowns in the risks & benefits, I won't argue. I do think we should keep #6842 open as a promising lead to revisit for perf improvements in the future.

PR #6859, opened 2023-02-03
[CT-1959]: moving simple_seed tests to adapter zone

nssalian:
resolves #

### Description

Moving the simple seed test code over to the adapter zone to help convert tests in adapters.

### Checklist


tests/adapter/dbt/tests/adapter/simple_seed/test_seed.py
```diff
@@ -229,32 +234,24 @@ class TestSimpleSeedWithBOM(SeedConfigBase):
...
     def test_simple_seed(self, project):
-        results = run_dbt(["seed"])
-        assert len(results) == 1
-
-        # encoding param must be specified in open, so long as Python reads files with a
-        # default file encoding for character sets beyond extended ASCII.
-        with open(
-            project.project_root / Path("seeds") / Path("seed_bom.csv"), encoding="utf-8"
-        ) as fp:
-            assert fp.read(1) == BOM_UTF8.decode("utf-8")
```
nssalian, line 249: Couldn't exactly map this to the new test. CC: @VersusFacit 
```
codec = "utf-8"
encoded = seed__bom_csv.encode(codec)
assert seed__bom_csv == encoded.decode(codec)
```
is what can possibly replace but it's not great.

tests/adapter/dbt/tests/adapter/simple_seed/test_seed.py
```diff
@@ -229,32 +234,24 @@ class TestSimpleSeedWithBOM(SeedConfigBase):
     @pytest.fixture(scope="class", autouse=True)
     def setUp(self, project):
         """Create table for ensuring seeds and models used in tests build correctly"""
-        project.run_sql_file(project.test_data_dir / Path("seed_expected.sql"))
+        project.run_sql(seeds__expected_sql)
 
-        # manual copy because seed has a special and tricky-to-include unicode character at 0
```
colin-rogers-dbt, line 234: do we need to preserve these as actual `.csv` files to validate this? I'm wondering if it would be easier to have a core-only test that looks at some of these edge cases (BOM, unicode etc.) and an actually "simple" seed test in the adapter zone that just validates loading csv files

nssalian: I think that would be a better approach. I'll modify the code.

colin-rogers-dbt (approved):

VersusFacit (approved):
A good rework of this test. Now, I remember how hard this was when I did it because of the BOM unprintable character -- it stumped everyone on the team until I chanced upon the hidden character one day.

 Can we comment referencing that on which test case that is, just so people understand we have a hidden character there? That feels very important for future readers who may run into unprintable character issues.
 
 Otherwise, LGTM

PR #6873, opened 2023-02-06
CT 1808 diff based partial parsing

### LINKED ISSUE #6592: [CT-1808] [Spike] "diff"-based entry point into partial parsing
See https://github.com/dbt-labs/dbt-server/issues/128

The gist of current partial parsing, as I understand it:
1. Perform a number of checks to determine if a full re-parse is needed, or which files need to be re-parsed on account of changes to `--vars`, env vars, target/profile, etc ([`build_manifest_state_check`](https://github.com/dbt-labs/dbt-core/blob/d43c070007b823588e7a170c6c37495ce319b011/core/dbt/parser/manifest.py#L694-L751))
2. If partial parsing is possible: inspect the full relevant file system, using system timestamps + file hashes to detect changes to source files
3. Based on files changed: schedule nodes defined in those files (+ other nodes related to those nodes + other files affecting those nodes) for re-parsing

Rather than requiring `dbt-core` to inspect the full relevant file system, could we provide some sort of interface for saying, "These files have been added / modified / deleted," and skip straight to step 3? (This assumes that the check in step 1 has already taken place, even if the logic for it needs to live elsewhere.)

The potential data structure for this "diff" could (but doesn't have to) match up closely with the [`file_diff`](https://github.com/dbt-labs/dbt-core/blob/main/core/dbt/parser/partial.py#L146-L153) object that partial parsing constructs + uses today. Three thoughts:
- The current `file_diff` distinguishes between "schema files" (`.yml`) and other files
- The current `file_diff` is just a set of `file_id` pointers, not the actual content of those files — that _could_ be sufficient here too, but the most compelling version of this capability also enables us to pass in file contents programmatically
- While it would be great to use `dbt-core`'s internal `file_id` as the unique identifier here (`project_name://subdir/path/file.ext`), it would require an external process to know the project's `name` (defined in `dbt_project.yml`, which might be a faulty assumption.

Imagining something like:
```json
{
  "added": [
    {
        "path": "models/new_model.sql",
        "content": "... lots of jinja + sql ...",
    },
    {
        "path": "models/docs.md",
        "content": "... some markdown ...",
    },
  ],
  "changed": [
    {
        "path": "models/my_py_model.py",
        "content": "... some python ...",
    },
  ],
  "deleted": [
    "models/other.sql",
    "models/other/schema.yml",
  ],
}
```
Programmatically, we'd call a `parse` with pointers to both:
1. existing manifest — although I think this would need to be `partial_parse.msgpack`, not just `manifest.msgpack`, given the need for all the between-file links that power partial parsing
2. "diff," either in a file or passed in as a data structure from memory

gshank:
resolves #6592

### Description

This is at the proof of concept stage, for discussion purposes. There are a number of edge cases and questions to resolve before being ready to officially review.

### Checklist


jtcohen6 (commented):
Great start to discussion! Next step: pull together a group of colleagues to talk much more about this :)

gshank:
A minor nit is that currently we do a checksum on the content of the file, then do a "strip" and store it. To keep the checksums from changing we should either do the checksum after the strip or not do the strip. My inclination is to do the checksum after the strip. If we start storing empty files, that would be slightly easier.

core/dbt/parser/read_files.py
```diff
@@ -153,104 +172,243 @@ def generate_dbt_ignore_spec(project_root):
...
+                source_file.contents = input_file.content
+                source_file.checksum = FileHash.from_contents(input_file.content)
+                source_file.path.modification_time = input_file.modification_time
+                # Handle creation of dictionary version of schema file content
+                if isinstance(source_file, SchemaSourceFile) and source_file.contents:
+                    dfy = yaml_from_file(source_file)
+                    if dfy:
+                        validate_yaml(source_file.path.original_file_path, dfy)
+                        source_file.dfy = dfy
+                    # TODO: ensure we have a file object even for empty files, such as schema files
```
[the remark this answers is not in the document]
ChenyuLInx: What if we leave the empty part to where we figure out what are the file diffs? Would that leads to simpler code on the parsing side?
I think it is reasonable to thing the input to parsing could be a manifest + file diff in the following format
```
{
  "updated":{
    "file1": "full content of file1",
    "file2": "full content of file2",
    ...
  },
  "added":{
    "file3": "full content of file3",
    "file4": "full content of file4",
    ...
  },
 "deleted": ["file4", "file5"],
}
```
`updated` and `added` could also be merged into one dictionary.

From dat-core's perspective an empty file would be the same as that file doesn't exist

tests/functional/partial_parsing/test_file_diff.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+second_file_diff = {
+    "deleted": [],
+    "changed": [],
+    "added": [{"path": "models/model_two.sql", "content": "select 123 as notfun"}],
+}
+
+
+class TestFileDiffs:
```
ChenyuLInx, line 19: Since we plan to use manifest as some sort of interface between parsing/execution, should we test the parsing based on checking the manifest instead of running dbt?

core/dbt/parser/read_files.py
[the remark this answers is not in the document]
gshank: This is the format that we're using now -- are you suggesting something different?

The empty file thing only applies to schema files, and we can't depend on it being simply an empty string -- it might have comments or whitespace, etc. I've changed not storing a schema file if it doesn't reduce to a real dictionary.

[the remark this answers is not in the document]
ChenyuLInx: I think I missed the fact that this is just for schema files. And the comments/whitespace are good call out!

gshank:
I put a little thought into how we would handle non-base project updates, and I think we have some open questions about how this will work with regard to "deps". The "deps" command takes both remote and local dependencies and copies them into the "dbt_packages" directory. Would these files have to be reflected back to the UI server which is providing the file_diff? Or would these files be "installed" only on the remote server which is accepting the file diffs? So would dependencies be local to the UI server or local to the remote dbt server?  If the "deps" command is processing a local dependency, would the ui server need to know that and provide all of those files, in a different kind of interface?

core/dbt/parser/partial.py
```diff
@@ -73,7 +73,6 @@ def __init__(self, saved_manifest: Manifest, new_files: MutableMapping[str, AnyS
         self.project_parser_files: Dict = {}
         self.saved_files = self.saved_manifest.files
         self.project_parser_files = {}
-        self.deleted_manifest = Manifest()
```
gshank, line 76: I removed the deleted_manifest because it's not necessary. Originally I thought it might be helpful for diagnosing, but it wasn't.

core/dbt/parser/read_files.py
```diff
@@ -49,17 +70,17 @@ def load_source_file(
             skip_loading_schema_file = True
 
     if not skip_loading_schema_file:
-        file_contents = load_file_contents(path.absolute_path, strip=False)
-        source_file.checksum = FileHash.from_contents(file_contents)
-        source_file.contents = file_contents.strip()
+        # We strip the file_contents before generating the checksum because we want
```
gshank, line 73: This is a change in behavior. It felt really wrong that we couldn't actually generate the checksum from the stored file contents, since it was done on the file contents prior to the contents being stripped and stored.

The alternative would have been to store the unstripped version of the contents.

core/dbt/parser/read_files.py
```diff
@@ -153,104 +172,243 @@ def generate_dbt_ignore_spec(project_root):
...
-        project.docs_paths,
-        [".md"],
-        ParseFileType.Documentation,
-        saved_files,
-        dbt_ignore_spec,
-    )
+        # Now that we have a copy of the files, remove deleted files
+        # For now, we assume that all files are in the root_project, until
+        # we've determined whether project name will be provided or deduced
+        # from the directory.
```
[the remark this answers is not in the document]
gshank: As it turns out, the only other local files besides the root project are local dependencies, and there currently is no way to connect up a local package with its directory, and in addition local dependency "files" are generated off of where they're copied by deps. This issue has not been resolved, because it's tied up in the question about what to do about deps.

gshank:
Although we aren't moving forward with more complete plans for file diff parsing, this pull request contains some refactoring which we don't want to lose, and it can serve as a base for using file_diff parsing in our tests, as a possible performance improvement and additional proof-of-concept.

core/dbt/parser/read_files.py
peterallenwebb: Makes sense to me.

core/dbt/parser/read_files.py
```diff
@@ -153,104 +174,244 @@ def generate_dbt_ignore_spec(project_root):
...
-        project,
-        files,
-        project.macro_paths,
-        [".sql"],
-        ParseFileType.Macro,
-        saved_files,
-        dbt_ignore_spec,
-    )
+@dataclass
+class ReadFilesFromFileSystem:
```
peterallenwebb, line 178: Just generally, more typing would have helped me understand more of this file more quickly. That said, I think the best way to capture shared behavior between classes like ReadFilesFromFileSystem and ReadFilesFromDiff is by defining a Protocol. But Protocols were introduced in Python 3.8, so we have a few more months without them.

peterallenwebb (approved):
Generally looks good and makes sense as far as I can tell, but I think we should clean up the unused event.

core/dbt/parser/read_files.py
gshank: We do use protocols in a few places already, so I think there are compatibility things for before 3.8.

core/dbt/events/types.py
```diff
@@ -1891,13 +1900,7 @@ def message(self) -> str:
...
 
 
-@dataclass
-class SystemErrorRetrievingModTime(ErrorLevel, pt.SystemErrorRetrievingModTime):
-    def code(self):
-        return "Z004"
-
-    def message(self) -> str:
-        return f"Error retrieving modification time for file {self.path}"
+# Skipped Z004
```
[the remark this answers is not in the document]
gshank: pfft. Good catch. Have removed.

PR #6897, opened 2023-02-08
Add double to float types for columns

### LINKED ISSUE #6876: [CT-2043] [Bug] Missing type (double) in `is_float` for databricks

### Current Behavior

I'm not sure if this belongs in here or the spark or the databricks adapator, but a `double` type column in databricks returns false for all of `is_float`, `is_numeric`, and `is_number`.

### Expected Behavior

It would return true for all those class method.

### Steps To Reproduce

1. Create a databricks table with a double column
2. Use `adapter.get_columns_in_relation(your_table)` and loop over the columns, logging or adding comments for the `.data_type` and the `.is_number()`
3. Observe the column with type `double` has `false` for `is_number()`

### Relevant log output

_No response_

### Environment

```markdown
- OS:
- Python:
- dbt: 1.4.1
```

### Additional Context

Using databricks as unable to test on spark, it is a delta table if that is relevant.

I believe it just needs to be added to the list here:
https://github.com/dbt-labs/dbt-core/blob/b2ea2b8b256e5db1da0b712dfedd7973e1e50a37/core/dbt/adapters/base/column.py#L55-L63

rlh1994:
resolves #6876 

### Description

Adds a `double` type for float type checks, used within databricks and also as an alias in snowflake warehouses.

### Checklist

### DIFF
--- a/core/dbt/adapters/base/column.py
+++ b/core/dbt/adapters/base/column.py
@@ -60,6 +60,7 @@ class Column:
             "float",
             "double precision",
             "float8",
+            "double",
         ]
 
     def is_integer(self) -> bool:
--- a/.changes/unreleased/Fixes-20230208-104230.yaml
+++ b/.changes/unreleased/Fixes-20230208-104230.yaml
@@ -1,0 +1,6 @@
+kind: Fixes
+body: Add double type to list of float column types for the column class
+time: 2023-02-08T10:42:30.918479Z
+custom:
+  Author: rlh1994
+  Issue: "6876"


core/dbt/adapters/base/column.py
```diff
@@ -60,6 +60,7 @@ def is_float(self):
             "float",
             "double precision",
             "float8",
+            "double",
```
nssalian, line 63: Is there a way to test this change?

rlh1994: I might be mistaken but I think those existing tests (based on the types) are valid for a postgres database, postgres doesn't have a `double` type, it's an alias in snowflake and a pure type in databaricks, so the tests would need to run on one of those to be valid?

nssalian: That tracks. @jtcohen6, do you know of any way to help test this? The change itself is simple enough though.

nssalian (approved):

PR #6856, opened 2023-02-03
Test fix: `TestProfileEnvVars::test_profile_env_vars`

iknox-fa:
resolves # 
https://media.tenor.com/5WTzTOGLLyEAAAAC/no-ticket-indiana-jones.gif

### Description
Fixes issue where `FailedToConnectError` errors stop test from being completed due to not using `pytest.raises`

### Checklist

- [x] I have read [the contributing guide](https://github.com/dbt-labs/dbt-core/blob/main/CONTRIBUTING.md) and understand what's expected of me
- [x] I have signed the [CLA](https://docs.getdbt.com/docs/contributor-license-agreements)
- [x] I have run this code in development and it appears to resolve the stated issue
- [x] This PR includes tests, or tests are not required/relevant for this PR
- [x] I have [opened an issue to add/update docs](https://github.com/dbt-labs/docs.getdbt.com/issues/new/choose), or docs changes are not required/relevant for this PR
- [x] I have run `changie new` to [create a changelog entry](https://github.com/dbt-labs/dbt-core/blob/main/CONTRIBUTING.md#adding-a-changelog-entry)

PR #6908, opened 2023-02-08
dbt_project.yml log-path respects project directory

MichelleArk:
Resolves: https://github.com/dbt-labs/dbt-core/issues/6957

Fixes log_path setting in flags to respect project_dir, if available

On [before_click_merge branch](https://github.com/dbt-labs/dbt-core/tree/before_click_merge)
```
❯ dbt run --project-dir ~/src/jaffle_shop
...
❯ ls logs
ls: logs: No such file or directory
❯ ls ~/src/jaffle_shop/logs
dbt.log
```

currently, on main
```
❯ dbt run --project-dir ~/src/jaffle_shop
...
❯ ls logs
dbt.log #bad!
❯ ls ~/src/jaffle_shop/logs
ls: /Users/michelleark/src/jaffle_shop/logs: No such file or directory #also bad!
```

🎩 after fix: 
```
❯ dbt run --project-dir ~/src/jaffle_shop
...
❯ ls logs
ls: logs: No such file or directory
❯ ls ~/src/jaffle_shop/logs
dbt.log
```

### Checklist


core/dbt/tests/fixtures/project.py
```diff
@@ -356,8 +356,8 @@ def project_files(project_root, models, macros, snapshots, properties, seeds, te
 
 # We have a separate logs dir for every test
 @pytest.fixture(scope="class")
-def logs_dir(request, prefix):
-    return os.path.join(request.config.rootdir, "logs", prefix)
+def logs_dir(project_root, prefix):
+    return os.path.join(project_root, "logs", prefix)
```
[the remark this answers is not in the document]
MichelleArk: Ah! I didn't realize this was the desired behaviour but I can see the utility. I changed it because this `logs_dir` is taking precedence over the default, and we're [expecting the default log location for in a functional test](https://github.com/dbt-labs/dbt-core/pull/6908/files#diff-5c2acd7b48060d11487f9bbb724169956b01dc7daf12d4e62a29dd03ae673ab4R337). I'll do a bit more digging here to understand what main does better and find an alternative.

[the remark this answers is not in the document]
MichelleArk: Okay - I see what was going on. The test was relying on a default location of log path that was incorrect on the feature branch. When we fixed the default location of log path and properly read it from the dbt_project.yml, this test started to break again. The fix is just to read from `logs_dir` directly in the test, since that's where the tests are written. 

Thanks for pointing this out!

tests/functional/partial_parsing/test_pp_vars.py
```diff
@@ -334,7 +334,7 @@ def test_profile_env_vars(self, project):
         with pytest.raises(FailedToConnectError):
             run_dbt(["run"], expect_pass=False)
 
-        log_output = Path(project.project_root, "logs", "dbt.log").read_text()
+        log_output = Path(logs_dir, "dbt.log").read_text()
```
gshank, line 337: This test in main has the code:
```
        # Change env_vars, the user doesn't exist, this should fail
        os.environ["ENV_VAR_USER"] = "fake_user"                     
        (results, log_output) = run_dbt_and_capture(["run"], expect_pass=False)
        assert "env vars used in profiles.yml have changed" in log_output
        manifest = get_manifest(project.project_root)                                       
        assert env_vars_checksum != manifest.state_check.profile_env_vars_hash.checksum            
```

Is there some reason that this test is now throwing an error in click-cli when it doesn't in main? And is there some reason that it's not doing "run_dbt_and_capture"?  In a general way it's better to read the output for a single command using run_dbt_and_capture rather than reading the entire log file, because there might be other commands running which will confuse the checks for output in the log for this particular command. For this particular test, it doesn't look like it would matter, but in a general way those two ways of getting the logs are not exactly equivalent.

MichelleArk: The test was originally changed in [this PR](https://github.com/dbt-labs/dbt-core/pull/6856/files) to read the logs this way instead of via `run_dbt_and_capture`. 

> Is there some reason that this test is now throwing an error in click-cli when it doesn't in main?

It does appear that this exception is swallowed on main but not on the feature branch. It's not  clear why that's the case or whether it's something to do with the test setup/framework or the code, but will keep looking into it.

gshank: I think I remember being surprised that it didn't raise an error. Would be kind of nice to know why, but It's probably not worth spending time on, since it does seem like the branch is right and main is wrong.

MichelleArk:
Removing the fix to `tests/functional/partial_parsing/test_pp_vars.py::TestProfileEnvVars` in favor of https://github.com/dbt-labs/dbt-core/pull/6916. Scope of this PR is now just the difference in behaviour with log-path between click and legacy cli.

PR #6921, opened 2023-02-09
parse group resource

### LINKED ISSUE #6822: [CT-1989] Parse group resource
Parse a group resource as defined in schema yml files. Groups have a required (at least for now!) owner that should be unified with the existing [`ExposureOwner`](https://github.com/dbt-labs/dbt-core/blob/main/core/dbt/contracts/graph/unparsed.py#L427) dataclass internally.

Example spec: 
```yaml
# models/marts/github/github.yml

groups:
  - name: github
    owner: 
      name: Jeremy
      email: data@jer.co
      slack: talk-jerco-memes
      github: jtcohen6
      whatever: you want
```

Eventually ([issue](https://github.com/dbt-labs/dbt-core/issues/6823)), group names will be used as values for the optional `group` model config:
```yaml
# models/marts/github/github.yml

models:
  - name: int__github_issue_label_history
    config: 
      group: github  # explicit opt-in
  - name: fct_github_issues
    config: 
      group: github  # explicit opt-in
```

### Considerations
* `ExposureOwner` currently has a required email, and an optional name. What should the unified unparsed `Owner` schema be? Proposal for a unified schema/dataclass in https://github.com/dbt-labs/dbt-core/issues/6833

MichelleArk:
resolves #6822 

### Checklist


gshank:
I just merged some changes to tests/functional/artifacts/data/state/v8/manifest.json, you probably need to merge main.

MichelleArk:
@gshank - Thanks for letting me know! I've merged main, should be ready for review.

core/dbt/parser/partial.py
```diff
@@ -903,6 +926,19 @@ def delete_schema_exposure(self, schema_file, exposure_dict):
             elif unique_id in self.saved_manifest.disabled:
                 self.delete_disabled(unique_id, schema_file.file_id)
 
+    # groups are created only from schema files, so just delete the group
```
gshank, line 929: When there are references to groups in other nodes, then we'll have to go through and schedule every node with this groups for reparsing.

gshank (approved):
Looks good.

core/dbt/parser/partial.py
MichelleArk: Makes sense. That would also go for modifications to existing groups in addition to deletions, right? 

I'm thinking it makes more sense to add that to the follow-up PR that parses group configs on group-able nodes: https://github.com/dbt-labs/dbt-core/pull/6965.

gshank: We'd only need to reparse for modifications of existing groups if information from the group is pulled into the node that needs to be rebuilt. If we only have the name in the node then probably not.

gshank: On thinking about this... Would people be pulling group information into model jinja code? Like model.group.owner? Then we'd have to reparse.

jtcohen6: We're not storing any additional group information directly on the node, right? Users would need to look up `.groups()` in the manifest / `flat_graph` (a.k.a. `{{ graph }}` in Jinja code) to get information about the `owner`

MichelleArk: > We're not storing any additional group information directly on the node, right? Users would need to look up .groups() in the manifest / flat_graph (a.k.a. {{ graph }} in Jinja code) to get information about the owner

Right - we're just storing the name. Changes to which will be captured as a deleted + new group. We're not currently adding groups to the flat graph - should we be?

jtcohen6: Ooh, I'd just seen [this change](https://github.com/dbt-labs/dbt-core/pull/6921/files#diff-33c94e40283afdfeba411a98c0628fc610c287e41c3477d88559f1e0244259c3R421-R426) and thought we were.

I think we probably should. This feels like valid metadata that someone might want access to in the Jinja context (including the folks who maintain the [`dbt_artifacts`](https://github.com/brooklyn-data/dbt_artifacts) package), and I can't see potential for anti-pattern shenanigans. (There are indeed some things we exclude from the `flat_graph`, such as macros. I would say that exclusion is somewhat intentional. I just remembered this issue that I tagged myself to refine forever ago: https://github.com/dbt-labs/dbt-core/issues/4919)

It would just require an update to [this method](https://github.com/dbt-labs/dbt-core/blob/daea7d59a72da3a164d6e28839db8e240efd7688/core/dbt/contracts/graph/manifest.py#L679-L690), and associated tests

MichelleArk: okay, makes sense! I had actually added it on first pass of this work but wasn't convinced it should be part of the flat graph so removed it. Will add now

PR #6950, opened 2023-02-11
postgres relations add foreign_tables

lialzm:
resolves dbt-labs/dbt-postgres#53

When using incremental mode, if it is a remote table, data cannot be inserted


lialzm:
I submitted the cla agreement, but there seems to be no feedback

graciegoheen:
Thanks for opening this PR @lialzm ! Sorry for the delay here, I see this PR is missing an associated issue - that's typically where we give product/dx feedback. 

Could you do the following?

1. Open an associated issue (so there is a more detailed description of the problem, including a simple code example, that we can link to it in the changelog).
2. Add a changelog entry

dbeatty10:
This looks like it would resolve https://github.com/dbt-labs/dbt-core/issues/7663, and I've added `resolves #7663` to the PR description to link it to that issue.

@lialzm 

1\. is taken care of -- no need to create a new issue.

You'd still need to do:

2\. [create a changelog entry](https://github.com/dbt-labs/dbt-core/blob/6461f5aacfcb72bd51ca2fd352369685938bcb2d/CONTRIBUTING.md#adding-or-modifying-a-changelog-entry) (via `changie new`).

plugins/postgres/dbt/include/postgres/macros/adapters.sql
```diff
@@ -71,6 +71,14 @@
 
 {% macro postgres__list_relations_without_caching(schema_relation) %}
   {% call statement('list_relations_without_caching', fetch_result=True) -%}
+    select
+      '{{ schema_relation.database }}' as database,
+      foreign_table_name as name,
+      foreign_table_schema as schema,
+      'table' as type
+    from information_schema.foreign_tables
+    where foreign_table_schema ilike '{{ schema_relation.schema }}'
```
dbeatty10, line 80: @lialzm could you take a look at the prototype version of this macro [here](https://github.com/dbt-labs/dbt-core/issues/7663#issuecomment-1584854886)?

### Key thing
The key thing is that it uses `external` as the type rather than `table`. I don't know for certain, but I'm guessing that we'll want to to be `external`.

i.e., we want it to be the same as `dbt_utils.get_table_types_sql` [here](https://github.com/dbt-labs/dbt-utils/blob/6ba7b660b1d1b4e3b41cb0cf6c3c0e4a70ae54e4/macros/sql/get_table_types_sql.sql#L18).

### Something else to consider

I don't have strong opinions on using the `information_schema.foreign_tables` view versus the `information_schema.tables` view. Using `information_schema.foreign_tables` is a little more simple since it doesn't need an extra filter in the where clause.

dbeatty10:
Thanks for taking the time to open this PR @lialzm! Since opening, we've [decoupled dbt Adapters from dbt Core](https://github.com/dbt-labs/dbt-core/discussions/9171), and this code now lives in a separate repo: [dbt-postgres](https://github.com/dbt-labs/dbt-postgres).

PR #6964, opened 2023-02-14
add helper function for load project/profile

ChenyuLInx:
resolves #

This PR added a (hacky) helper function to load project file and profile file.
It is currently used both in lib.py, and dbt-server directly
The [dbt-server direct usecase ](https://github.com/dbt-labs/dbt-server/pull/167/files)is because there's no way to programmatically specify a Project Dir right now.
We should look into remove all function here in the second phase of API-ification once we improve our programatic interface core provides.

### Checklist

### DIFF
--- a/core/dbt/lib.py
+++ b/core/dbt/lib.py
@@ -1,6 +1,5 @@
 import os
 from dbt.config.project import Project
-from dbt.config.renderer import DbtProjectYamlRenderer
 from dbt.contracts.results import RunningStatus, collect_timing_info
 from dbt.events.functions import fire_event
 from dbt.events.types import NodeCompiling, NodeExecuting
@@ -9,6 +8,8 @@ from dbt import flags
 from dbt.task.sql import SqlCompileRunner
 from dataclasses import dataclass
 from dbt.cli.resolvers import default_profiles_dir
+from dbt.config.runtime import load_profile, load_project
+from dbt.flags import set_from_args
 
 
 @dataclass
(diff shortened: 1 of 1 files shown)


racheldaniel (approved):
Thank you!

core/dbt/lib.py
```diff
@@ -74,10 +81,6 @@ def get_dbt_config(project_dir, args=None, single_threaded=False):
 
     profile_name = getattr(args, "profile", None)
 
-    profile_name = getattr(args, "profile", None)
```
iknox-fa, line 77: :)
:)
:)

core/dbt/lib.py
```diff
@@ -74,10 +81,6 @@ def get_dbt_config(project_dir, args=None, single_threaded=False):
...
-
-    profile_name = getattr(args, "profile", None)
-
     runtime_args = RuntimeArgs(
```
iknox-fa, line 84: :)
:)
:)

iknox-fa (approved):
LGTM

core/dbt/lib.py
```diff
@@ -86,9 +89,8 @@ def get_dbt_config(project_dir, args=None, single_threaded=False):
         target=getattr(args, "target", None),
```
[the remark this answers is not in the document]
ChenyuLInx: Does passing in an `args` that has target specified in it not working?

But yes I think the current way of only allowing modification of flags in the text form of args has a lot of drawbacks. See the comment [here](https://github.com/dbt-labs/dbt-server/pull/167/files#diff-a7d2baed1389102b086c2df869095347d3735c640855254dbc4473b99194035cR274).

core/dbt/lib.py
ChenyuLInx: Once we have a better interface this whole file should be gone

Issue #6751, opened 2023-01-26
[CT-1919] Create get_column_schema_from_query macro


MichelleArk:
[`get_columns_in_query`](https://github.com/dbt-labs/dbt-core/blob/main/core/dbt/include/global_project/macros/adapters/columns.sql#L20-L24) retrieves a list of column names from a model's sql by running an inexpensive `select * from {{ sql }} where false limit 0`  query. Currently, this macro is leveraged to efficiently [enforce that a model's contracted names match](https://github.com/dbt-labs/dbt-core/pull/6271/files#diff-28fef0f8080904ca9d7c04e61050beded83aaecc39d332b8f0486eea686d44f1R39) with what its sql would actually produce. 

However, we'll need to be able to retrieve column data types _in addition_ to column names from a model's sql in order to enforce a model contract complete with data types. 

Let's create a new macro, `get_column_schema_from_query` with a default implementation that returns a list of `Column` objects with column name and type. This macro will need to rely on adapter-specific implementations to data type off the connection cursor/API. Because aapter-specific connections may return data type representations that do not match up with user-provided values, we’ll need to use [dbt adapters’ type translation (aliases → standard)](https://github.com/dbt-labs/dbt-core/blob/eb200b46877c31469f410b7d47dfd245473ed7cb/core/dbt/adapters/base/column.py#L23-L25), to match up data types with user-provided values. 

@jtcohen6 has taken a very thorough first stab at validating this approach in [jerco/get-column-schema-from-query](https://github.com/dbt-labs/dbt-core/compare/jerco/get-column-schema-from-query)

PR #6986, opened 2023-02-15
get_column_schema_from_query macro

MichelleArk:
resolves https://github.com/dbt-labs/dbt-core/issues/6751

### Description
* `adapters/base/impl.py`
  *  Adds [`get_column_schema_from_query(self, sql: str) -> List[Tuple[str, Any]]`]((https://github.com/dbt-labs/dbt-core/pull/6986/files#diff-aa1e9e602a0f1e2930b465b30a35a28778fbe131da4fbf2c4e88dd52ca44f140R273)) to `BaseAdapter` 
   * calls the connection manager's `get_column_schema_from_query` method
* `adapters/sql/connection.py`
  * Implements [default](https://github.com/dbt-labs/dbt-core/pull/6986/files#diff-b7e353a2e80718c8fb00ecd3b2d5ffcedd51cefa91f1cfabe862954de8feb567R168) `get_column_schema_from_query` on `SQLConnectionManager`, which depends on two new class methods:
    1. `get_column_schema_from_cursor`
    2.  `data_type_code_to_name`
   * These new class methods (including `get_column_schema_from_query`) can be overwritten in adapter connection manager implementations of `get_column_schema_from_query`.
* `plugins/postgres/dbt/adapters/postgres/connections.py`
  * `PostgresConnectionManager` overwrites [`data_type_code_to_name`](https://github.com/dbt-labs/dbt-core/pull/6986/files#diff-7d1dade4a84dfc685fd8907fe32d21f269a1a40178322926d9f0b3db6fa4ca37R196) using `psycopg2.extensions.string_types`. This is the only change necessary to get model contracts extended to `data_type`s for Postgres.

**get_column_schema_from_query Usage**
* Used in [assert_columns_equivalent](https://github.com/dbt-labs/dbt-core/pull/6986/files#diff-28fef0f8080904ca9d7c04e61050beded83aaecc39d332b8f0486eea686d44f1L30)
  * `get_column_schema_from_query` is [called twice](https://github.com/dbt-labs/dbt-core/pull/6986/files#diff-28fef0f8080904ca9d7c04e61050beded83aaecc39d332b8f0486eea686d44f1R31-R32): 
    1. Once on the model SQL. This obtains the column schema for the _actual_ model based on its SQL. 
    2. Once on the [`empty_schema_sql`](https://github.com/dbt-labs/dbt-core/pull/6986/files#diff-fc4b08268274f470faed2ab4527aa52f1691b86859917d0a0f623129aa0cc67dR37-R43), which is simply a `SELECT cast(null as {{ col['data_type'] }}` for each column specified in the _expected_ model schema yaml. 
  * Once the expected and actual column schemas are obtained, they are formatted and compared, [raising a compiler error](https://github.com/dbt-labs/dbt-core/pull/6986/files#diff-28fef0f8080904ca9d7c04e61050beded83aaecc39d332b8f0486eea686d44f1R38) if any differences are detected. 
 

### Checklist

- [x] [Jerco has] [opened an issue to add/update docs](https://github.com/dbt-labs/docs.getdbt.com/issues/new/choose), or docs changes are not required/relevant for this PR: https://github.com/dbt-labs/docs.getdbt.com/pull/2890/files#diff-9056e8c425c88b44d27796d1131ceb68c5b5b15ccc57a140d609be3d4b5358e7R47

🎩
```
❯ dbt run --select test_schema
02:53:06  Running with dbt=1.5.0-a1
02:53:06  Found 6 models, 20 tests, 0 snapshots, 1 analysis, 503 macros, 0 operations, 3 seed files, 0 sources, 1 exposure, 1 metric
02:53:06  
02:53:07  Concurrency: 1 threads (target='dev')
02:53:07  
02:53:07  1 of 1 START sql table model test_arky.test_schematest_v2 ...................... [RUN]
02:53:07  1 of 1 ERROR creating sql table model test_arky.test_schematest_v2 ............. [ERROR in 0.04s]
02:53:07  
02:53:07  Finished running 1 table model in 0 hours 0 minutes and 0.46 seconds (0.46s).
02:53:07  
02:53:07  Completed with 1 error and 0 warnings:
02:53:07  
02:53:07  Compilation Error in model test_schema (models/test_schema.sql)
02:53:07    Please ensure the name, data_type, order, and number of columns in your `yml` file match the columns in your SQL file.
02:53:07    Schema File Columns: 
02:53:07      a INT, b TEXT
02:53:07    
02:53:07    SQL File Columns: 
02:53:07      a INT, c TEXT 
02:53:07    
02:53:07    > in macro assert_columns_equivalent (macros/materializations/models/table/columns_spec_ddl.sql)
02:53:07    > called by model test_schema (models/test_schema.sql)
02:53:07  
02:53:07  Done. PASS=0 WARN=0 ERROR=1 SKIP=0 TOTAL=1
```


core/dbt/adapters/sql/connections.py
```diff
@@ -128,6 +128,31 @@ def get_result_from_cursor(cls, cursor: Any) -> agate.Table:
 
         return dbt.clients.agate_helper.table_from_data_flat(data, column_names)
 
+    @classmethod
+    def data_type_code_to_name(cls, int) -> str:
```
MichelleArk, line 132: TODO: improve `int` variable name, add type annotations

core/dbt/include/global_project/macros/adapters/columns.sql
```diff
@@ -17,23 +17,60 @@
 {% endmacro %}
 
 
+{% macro get_empty_subquery_sql(select_sql) -%}
+  {{ return(adapter.dispatch('get_empty_subquery_sql', 'dbt')(select_sql)) }}
+{% endmacro %}
+
+{% macro default__get_empty_subquery_sql(select_sql) %}
```
MichelleArk, line 24: TODO: add comment about why we have `where false limit 0`

core/dbt/adapters/sql/connections.py
```diff
@@ -128,6 +128,31 @@ def get_result_from_cursor(cls, cursor: Any) -> agate.Table:
...
+    @classmethod
+    def data_type_code_to_name(cls, type_code: Union[int, str]) -> str:
+        """Get the string representation of the data type from the type_code."""
+        # https://peps.python.org/pep-0249/#type-objects
+        raise dbt.exceptions.NotImplementedError(
+            "`data_type_code_to_name` is not implemented for this adapter!"
+        )
+
+    @classmethod
+    def get_column_schema_from_cursor(cls, cursor: Any) -> List[Tuple[str, str]]:
```
VersusFacit, line 140: Are models assessed with this guaranteed to be preexisting or does the empty list indication that it doesn't exist yet?

core/dbt/include/global_project/macros/adapters/columns.sql
```diff
@@ -17,23 +17,60 @@
...
+
+{% macro get_empty_schema_sql(columns) -%}
+  {{ return(adapter.dispatch('get_empty_schema_sql', 'dbt')(columns)) }}
+{% endmacro %}
+
+{% macro default__get_empty_schema_sql(columns) %}
+    select
+    {% for i in columns %}
+      {%- set col = columns[i] -%}
+      cast(null as {{ col['data_type'] }}) as {{ col['name'] }}{{ ", " if not loop.last }}
```
VersusFacit, line 41: We talk about this. This can possibly lead to some weird type resolution outcomes...we think. This is so far the "best" option and so far looks promising. I just know SQLs typing mechanisms can get a mind of their own for the worse.

core/dbt/include/global_project/macros/adapters/columns.sql
```diff
@@ -17,23 +17,60 @@
...
+    {%- endfor -%}
+{% endmacro %}
+
+{% macro get_column_schema_from_query(select_sql) -%}
+  {{ return(adapter.dispatch('get_column_schema_from_query', 'dbt')(select_sql)) }}
+{% endmacro %}
+
+{% macro default__get_column_schema_from_query(select_sql) %}
```
VersusFacit, line 49: I don't like this name. I'm not sure why. I feel like its saying "get column schema from some query" when it's far more deliberate than that in practice, as I understand it. 

This macro is said to "Get a list of the column names and data types from the given sql." What is that given SQL? Why is it? These are questions my brain latches onto as though part of the abstraction that'd help explain this is missing.

VersusFacit (commented):
My thoughts are all design minded as opposed to implementation minded. This might be an error on my part since Jinja is harder to track than Python. That said, I feel like deliberate design will help indicate/signpost what the implementation is expected to do.

core/dbt/include/global_project/macros/adapters/columns.sql
MichelleArk: > What is that given SQL? Why is it? These are questions my brain latches onto as though part of the abstraction that'd help explain this is missing.

Agree - but I do want to avoid an overly-specific name to this more general function. The deliberate use case here is to both 'get a list of column names and data types based on the schema yml' and 'get a list of column names and data types based on the user-provided SQL'. Both use cases call this underlying method `get_column_schema_from_query`, and presumably could be used for other applications in the future! 

I went with `get_column_schema..` here as opposed to `get_columns..` because the method is not returning the `Column` class but a list of tuples representing the schema. Don't have a strong preference between `from_query` or `from_sql`. I think I just went with `from_query` for consistency with the existing `get_columns_from_query` adapter method. 

I'll add some more comments here in terms of the deliberate use case in [assert_columns_equivalent](https://github.com/dbt-labs/dbt-core/pull/6986/files#diff-28fef0f8080904ca9d7c04e61050beded83aaecc39d332b8f0486eea686d44f1R31-R32), where we're composing these more generic methods to get a very deliberate and narrow functionality.

core/dbt/adapters/sql/connections.py
MichelleArk: At this point, callers using the method now are guaranteed to have run a query successfully, and yes, an empty `cursor.description` would lead to an empty list result here. I'll add a docstring around the expected behaviour of this method given a cursor.

core/dbt/include/global_project/macros/adapters/columns.sql
peterallenwebb: @VersusFacit I had the same concern, and discussed it with Michelle synchronously. On the plus side, this approach will automatically account for any new types which appear, so if it works in practice it will be a lot easier than trying to maintain our own list of type aliases. I like that. I was also reassured that this code path will only affect people using contracts, so there isn't much regression risk and we should hear pretty quickly if there are databses/drivers that this approach doesn't work for.

peterallenwebb (approved):

core/dbt/include/global_project/macros/adapters/columns.sql
MichelleArk: One more thing this approach has going for it is that there's a non-opaque definition of what `data_type` values should be - it's 'the thing you'd write in SQL when casting to the desired type', as opposed to 'the value returned by mapping the connection cursor's type code to a string'.

mikealfare:
@MichelleArk I could be naive to the context here, so take this with a grain of salt. What is the benefit gained from putting these methods largely on `SQLConnectionManager`? They rely very little on the connection attribution, mainly just `cursor`. And then the result is pulled through in a pass-through method in `BaseAdapter` with no logic applied.

I'm concerned that we might be polluting the interface on `SQLConnectionManager`. If I'm thinking about it correctly, `SQLConnectionManager` deals with the general task of how we interrogate the database and `BaseAdapter` deals with what we're asking of the database. `BaseAdapter` effectively uses `SQLConnectionManager` to answer questions; `SQLConnectionManager` doesn't ask questions itself.

I would think some of this logic belongs on `BaseAdapter`. At first glance, I would expose a way to add a standard query on `SQLConnectionManager` without executing it, basically the body of `get_column_schema_from_query`. And then I'd probably also put the data_type map there because it has no knowledge of the question being asked:
```python
class SQLConnectionManager(BaseConnectionManager):
    ...

    def add_select_query(self, sql: str) -> Tuple[Connection, Any]:
        sql = self._add_query_comment(sql, auto_begin=False)
        return self.add_query(sql)

    def add_begin_query(self):
        return self.add_query("BEGIN", auto_begin=False)

    ...

    @property
    def data_type_map(self) -> Dict[Union[str, int], str]:
        # build the dictionary somehow, needs to be a method to raise NotImplementedError
        # e.g., abstract
        raise dbt.exceptions.NotImplementedError(...)
        # e.g., Snowflake:
        return snowflake.connector.constants.FIELD_ID_TO_NAME

    ...
```

Then I'd put the logic from `get_column_schema_from_cursor` into `get_column_schema_from_query` on `BaseAdaptor`, e.g.:
```python

ColumnSchema = List[Tuple[str, str]]


class BaseAdapter(metaclass=AdapterMeta):
    ...

    @available.parse(lambda *a, **k: [])
    def get_column_schema_from_query(self, sql: str) -> ColumnSchema:
        _, cursor = self.connections.add_select_query(sql)
        data_type_map = self.connections.data_type_map
        columns: ColumnSchema = [
            column_name, data_type_map.get(column_code, "UNDEFINED")
            for column_name, column_code, *_ in cursor.description
        ]
        return columns

    ...
```

Am I thinking about this correctly?

core/dbt/include/global_project/macros/adapters/columns.sql
VersusFacit: this discussion should be seen as a context dump for posterity and is not blocking.

VersusFacit (approved):
I'm going to approve this as to not hold things up. However, please do consider Mike's comment before hitting merge. Our interfaces are in flux and this is the right time I think to ask these questions. If it doesn't end up in this PR, we should consider what sorts of "contracts" we want these objects to observe.

I admit I haven't considered until now the questions Mike raises, but they feel cogent, insofar as we should weigh these tradeoffs now (early).

MichelleArk:
@mikealfare - Thank you for this really thoughtful comment!

> I'm concerned that we might be polluting the interface on SQLConnectionManager.

I had the same concern here and went back and forth for a bit between where this functionality should live. I agree with the overall responsibility of each `SQLConnectionManager` and `BaseAdapter` you've described, and that the `get_column_schema_from_query` functionality in `BaseAdapter` goes beyond the general task of _how_ we interrogate the database. 

Your proposal makes a lot of sense and should work across adapter implementations, but my main concern was that the `SQLConnectionManager` would be leaking the `cursor` object, which has some standardization from the [DB-API standard](https://peps.python.org/pep-0249/#cursor-objects), but will vary across adapters and has to do with _how_ the database runs queries. Should the BigQueryAdapter need to know about `bigquery.job.query.QueryJob` (and know that a query has a temp destination table?) or `bigquery.table.RowIterator`? But.. it looks like the [bigquery adapter is already doing that, for a very similar purpose](https://github.com/dbt-labs/dbt-bigquery/blob/main/dbt/adapters/bigquery/impl.py#L456). I didn't see that was the case already 🤦

All that said - I do think the `BaseAdapter` is still the better spot for the implementation for the reasons you described. It additionally is configured with the `Column` class and is better positioned to create a `ColumnSchema = List[Column]` result for end callers to use, eliminating the need for [this messiness](https://github.com/dbt-labs/dbt-core/pull/6986/files#diff-fc4b08268274f470faed2ab4527aa52f1691b86859917d0a0f623129aa0cc67dR58-R63) in jinja.

I'm going to work in your proposed structure - it shouldn't require a huge refactoring lift in adapter-specific implementations. Longer term, I think we'd want to explore introducing an abstraction around a `Cursor` or `QueryResult` (similar to what we've done with `Connection`).

tests/adapter/dbt/tests/adapter/constraints/fixtures.py
```diff
@@ -116,4 +130,21 @@
...
+        constraints: ['not null','primary key']
+        constraints_check: (id > 0)
+        tests:
+          - unique
+      - name: color
+        data_type: text
+      - name: date_day
+        data_type: date
+      - name: num_array
+        data_type: int[]
```
dbeatty10, line 149: Would there be value in expanding this to cover more possible values for `data_type`? e.g., `boolean`, `timestamp without time zone`, `timestamp with time zone`, etc?

Or would that type of coverage be better in some other test?

core/dbt/adapters/sql/connections.py
```diff
@@ -128,6 +132,14 @@ def get_result_from_cursor(cls, cursor: Any) -> agate.Table:
...
         return dbt.clients.agate_helper.table_from_data_flat(data, column_names)
 
+    @classmethod
+    def data_type_code_to_name(cls, type_code: Union[int, str]) -> str:
+        """Get the string representation of the data type from the type_code."""
+        # https://peps.python.org/pep-0249/#type-objects
+        raise dbt.exceptions.NotImplementedError(
+            "`data_type_code_to_name` is not implemented for this adapter!"
+        )
+
```
dbeatty10, line 142: I was interested in seeing a tangible example of how data type codes map to a string representation for a database connector.

This table from the Snowflake docs was useful to me:
https://docs.snowflake.com/en/user-guide/python-connector-api#label-python-connector-type-codes

| type_code | String Representation | Data Type      |
|-----------|-----------------------|----------------|
| 0         | FIXED                 | NUMBER/INT     |
| 1         | REAL                  | REAL           |
| 2         | TEXT                  | VARCHAR/STRING |
| 3         | DATE                  | DATE           |
| 4         | TIMESTAMP             | TIMESTAMP      |
| 5         | VARIANT               | VARIANT        |
| 6         | TIMESTAMP_LTZ         | TIMESTAMP_LTZ  |
| 7         | TIMESTAMP_TZ          | TIMESTAMP_TZ   |
| 8         | TIMESTAMP_NTZ         | TIMESTAMP_TZ   |
| 9         | OBJECT                | OBJECT         |
| 10        | ARRAY                 | ARRAY          |
| 11        | BINARY                | BINARY         |
| 12        | TIME                  | TIME           |
| 13        | BOOLEAN               | BOOLEAN        |

(Side note: I _suspect_ there is a typo for code 8 and `TIMESTAMP_TZ` there should be `TIMESTAMP_NTZ` instead.)

tests/adapter/dbt/tests/adapter/constraints/fixtures.py
MichelleArk: Definitely - I'm working on rejigging the tests to have more coverage here + be easier to test adapter-specific types.

dbeatty10: One idea for testing using adapter-specific values is to use the data type functions [here](https://docs.getdbt.com/reference/dbt-jinja-functions/cross-database-macros#:~:text=type_timestamp-,Data%20type%20functions,-type_bigint).

image

e.g., something like:
```
      - name: color
        data_type: {{ dbt.type_string() }}
```

One downside of this idea is that it doesn't cover all the data types you might expect. For starters, it doesn't include `date` or `time`. Also, most databases support at least two different timestamp data types, but `type_timestamp()` only covers one of those.

tests/adapter/dbt/tests/adapter/constraints/test_constraints.py
```diff
@@ -32,7 +34,35 @@ def models(self):
...
+    @pytest.fixture
+    def int_type(self):
+        return "INT"
+
+    @pytest.fixture
+    def schema_int_type(self, int_type):
+        return int_type
+
+    @pytest.fixture
+    def data_types(self, schema_int_type, int_type, string_type):
```
MichelleArk, line 50: I would have loved to use pytest's `parametrized` functionality here (and did implement it that way to start), but unfortunately couldn't find a way to make it work with the testing inheritance pattern we have for adapter tests given the parametrized array would need to be provided dynamically - which pytest.parametrize does not support.

tests/adapter/dbt/tests/adapter/constraints/test_constraints.py
```diff
@@ -62,14 +97,91 @@ def test__constraints_wrong_column_names(self, project):
...
         assert expected_compile_error in log_output
         assert expected_schema_file_columns in log_output
         assert expected_sql_file_columns in log_output
 
+    def test__constraints_wrong_column_data_types(
+        self, project, string_type, int_type, schema_int_type, data_types
+    ):
+        for (sql_column_value, schema_data_type, error_data_type) in data_types:
+            # Write parametrized data_type to sql file
+            write_file(
```
mikealfare, line 117: It would be nice for this function to be a fixture instead, so that we can yield the file and then clean it up after. Maybe the `project` fixture does that if the file is setup in the project directory?

mikealfare (approved):

tests/adapter/dbt/tests/adapter/constraints/test_constraints.py
MichelleArk: The `project` fixture [drops the test schema on teardown](https://github.com/dbt-labs/dbt-core/blob/main/core/dbt/tests/fixtures/project.py#L519), but from what I can tell, it does not clean up the temp project directory. I understand this might actually be by design, for ease of debugging failing tests - [similar to intentionally preserving log files](https://github.com/dbt-labs/dbt-core/pull/6908#discussion_r1100677783). Worth revisiting, but probably out of scope for this PR.

Issue #6847, opened 2023-02-02
[CT-2022] Non-error log is not silenced by -q flag, Click Version


peterallenwebb:
### Current Behavior

This issue is a follow-up to #6749, which pertains to fixing a regression in 1.4. This issue pertains to fixing the same regression in the new click-cli branch, which will require a different approach.

### Expected Behavior

The -q/--quiet cli flag suppresses all non-error log messages, even before the entire logging config can be determined.

### Steps To Reproduce

See #6749

peterallenwebb:
Per [Jeremy's comment in the PR](https://github.com/dbt-labs/dbt-core/pull/7048#issuecomment-1449031767), I'm closing this (no fix needed).

PR #6990, opened 2023-02-16
fix small issues with flags and logger to support dbt-rpc

### LINKED ISSUE #6979: [CT-2123] [Bug] Flags are referenced incorrectly (and sometimes at the wrong time) in functions.py
### Current Behavior

Since the merge of the click-cli feature branch, the flags USE_COLOR and LOG_CACHE_EVENTS are accessed inconsistently and sometimes before their values are ready. Consider the following from functions.py:
```python
def _get_logbook_log_config(debug: bool) -> LoggerConfig:
    # use the default one since this code should be removed when we remove logbook
    flags = get_flags()
    config = _get_stdout_config("", debug, bool(flags.USE_COLORS))
    config.name = "logbook_log"
    config.filter = NoFilter if flags.LOG_CACHE_EVENTS else lambda e: not isinstance(e.data, Cache)
    config.logger = GLOBAL_LOGGER
    return config
```
Here LOG_CACHE_EVENTS is looked up using the wrong access pattern.

Also consider:
```python
EVENT_MANAGER: EventManager = EventManager()
EVENT_MANAGER.add_logger(
    _get_logbook_log_config(flags_module.DEBUG)  # type: ignore
    if flags_module.ENABLE_LEGACY_LOGGER
    else _get_stdout_config(flags_module.LOG_FORMAT, flags_module.DEBUG, flags_module.USE_COLORS)  # type: ignore
)
```
Here the flag values are being checked at module load, before they can reasonably be expected to have values set. To remedy this, we will replace the references to flags with specific default values. We also need to remove direct access to flags from functions like _get_stdout_config() and accept their values as parameters instead, to avoid the same access-before-set issue.

### Expected Behavior

The USE_COLOR and LOG_CACHE_EVENTS are not referenced before their values can reasonably be expected to be set, and they are accessed via their global variables in the dbt.flags module, not on the new flags object.

Other flags in the logging code should be inspected for these same issues.

### Steps To Reproduce

Run dbt with ENABLE_LEGACY_LOGGER=True

### Relevant log output

_No response_

### Environment

```markdown
- OS: macOS
- Python: 3.10
- dbt: main/head
```

_No response_

### Additional Context

_No response_

ChenyuLInx:
resolves #6979 https://github.com/dbt-labs/dbt-rpc/issues/122

The helper functions are not in ideal situation but I think we can revisit them in the revisit of Flags module that's happening soon


core/dbt/events/functions.py
```diff
@@ -133,7 +140,7 @@ def cleanup_event_logger():
 EVENT_MANAGER.add_logger(
     _get_logbook_log_config(debug=False)  # type: ignore
     if ENABLE_LEGACY_LOGGER
-    else _get_stdout_config(log_format="text", debug=False, use_colors=True)  # type: ignore
+    else _get_stdout_config(log_format="text", debug=False, use_colors=True, log_cache_events=True, quiet=False)  # type: ignore
```
jtcohen6, line 143: I think we'll want `log_cache_events=False` by default

FYI @peterallenwebb since you're:
- assigned to #6979
- thinking about how to handle `--quiet` in the default logger, before we have full config (#6847) — it's just being set to `False` here

To be honest, I don't know if it's possible, and it might be motivation for us to really finally uncouple `UserConfig` from `profiles.yml` (#6207), so that we can have a configured logger before we need to parse the profile (and encounter any warnings)

ChenyuLInx: Updated, what exactly does it do?

racheldaniel (approved):

core/dbt/cli/flags.py
```diff
@@ -49,13 +49,33 @@ def convert_config(config_name, config_value):
     # This function should take care of converting the values from config and original
     # set_from_args to the correct type
     ret = config_value
-    if config_name.lower() == "warn_error_options":
+    if config_name.lower() == "warn_error_options" and type(config_value) == dict:
```
MichelleArk, line 52: Why is this change necessary? Shouldn't warn_error_options always be a dictionary?

Docs: https://docs.getdbt.com/reference/global-configs#warnings-as-errors

core/dbt/cli/flags.py
```diff
@@ -49,13 +49,33 @@ def convert_config(config_name, config_value):
...
         ret = WarnErrorOptions(
             include=config_value.get("include", []), exclude=config_value.get("exclude", [])
         )
     return ret
 
 
+def args_to_context(args):
```
MichelleArk, line 59: nit: missing typing

core/dbt/lib.py
```diff
@@ -89,14 +88,18 @@ def get_dbt_config(project_dir, args=None, single_threaded=False):
         target=getattr(args, "target", None),
     )
 
+    # set global flags from arguments
```
ChenyuLInx, line 91: The functions in this file are still by large untested, gonna revisit to replace this file and test the interfaces we have to replace it.

core/dbt/cli/flags.py
ChenyuLInx: There are other places that passes in something that's not a dict. I can do another round of verify it is a dict and update input from other places.

core/dbt/cli/flags.py
```diff
@@ -49,13 +49,33 @@ def convert_config(config_name, config_value):
...
+    from dbt.cli.main import cli
+
+    cli_ctx = cli.make_context(cli.name, args)
+    # args would get converted during make context
+    if len(args) == 1 and "," in args[0]:
+        args = args[0].split(",")
+    sub_command_name, sub_command, args = cli.resolve_command(cli_ctx, args)
+
+    # handle source and docs group
+    if type(sub_command) == Group:
```
MichelleArk, line 70: I think we only have one level of `Group` nesting at the moment (sources and docs, as you mentioned in the comment), but we could make this a `while` instead so that it works for multiple levels of `Group`s too.

core/dbt/cli/flags.py
MichelleArk, line 59: What is the intended usage of this method? Is it just to create a click context that can be used to construct a `Flags` instance, or would we also want callers invoking the click command on the built context? If it's just the first use case, perhaps we could add this as a class method on Flags instead, and get eventually use it to get rid of `set_from_args`?

core/dbt/cli/flags.py
ChenyuLInx: Planning on revisit all of this and add more things to it as we rethink the flags object.

core/dbt/cli/flags.py
ChenyuLInx: Fixed

core/dbt/cli/flags.py
ChenyuLInx: I remember why, it is because the default value for `warn_error_options` in [UserConfig](https://github.com/dbt-labs/dbt-core/blob/main/core/dbt/contracts/project.py#L245) is None, so if you pass in a UserConfig object the original logic would fail

MichelleArk: 👍 gotcha. `config_value is not None` would be more precise here then

core/dbt/cli/flags.py
ChenyuLInx: Yes it is, and that's what I mean by Flag object is leaky.
The only place that uses this right now is in `dbt-rpc`, and we are gonna deprecate it in 1.6. So I am thinking of merging this change for now to make the 1.5 beta release, and do a revisit of the whole flags object before final cut of 1.5

MichelleArk (approved):

core/dbt/events/functions.py
jtcohen6: Super verbose logging every time a relation is added, removed, updated from the cache. Adds a ton of noise in the general case, but very useful to turn on when debugging issues that might be due to cache misses/inconsistencies

Issue #6993, opened 2023-02-16
[CT-2130] Respect `log.ini` config file if provided


jtcohen6:
> However, for CLI use a file config is more common and appropriate in my mind. The dictionary config is for calling directly within python. A --logging-config-file argument would probably meet all future logging needs. [file config docs](https://docs.python.org/3/library/logging.config.html#logging.config.fileConfig)

If an end user provides logging config in a `log.ini` configuration file, we should do our best to respect that file's configuration. I'd be curious to see how many other users / community members would be interested in the same.

_Original discussion in https://github.com/dbt-labs/dbt-core/issues/6639_

DustinMoriarty:
Here is some background. The request is to allow a user to provide a standard [logging.config](https://docs.python.org/3/library/logging.config.html) file which bypasses all DBT logging configuration. For CLI use, the user would provide the logging configuration file path. The user defines the log levels, log formats, log destinations (e.g. stderr, stdout, file, etc) in that file using the existing python log configuration format. There would be no need to add any new parsing of the configuration file schema. DBT would simply expose the existing built in python logging configuration options provided by the built in logging library provided by the language.

As far as all the extra event metadata available with json logging is concerned, as long as that metadata is provided in the "extra" argument in any log statement (e.g. `log.info("Something happened", extra={"data": {}, "info": {}})`), then it can be picked up by any log formatter which wants to use it, including json structured log formatters like [python-json-logger](https://pypi.org/project/python-json-logger/).

DustinMoriarty:
It looks like DBT uses [logbook](https://logbook.readthedocs.io/en/stable/) for logging rather than the standard library. I would take a hard look at if logbook is really a good fit. It looks like logbook adds a lot of complexity to solve problems which probably should not be handled within the application logger in the first place (sending emails, notifications, webhooks, raising tickets, etc.). These are all things which should be handled by log layers outside of the actual application (e.g. cloudwatch, ELK, splunk, etc.) and that DBT is not using anyway. I don't at first glace see a place in logback to provide a logging config, which in itself makes me question it's design since log config should be a runtime decision and never hard coded.  I would be curious what logbook feature is needed by DBT.


There must be a way that logbook exposes the logging config though.

jtcohen6:
@DustinMoriarty We don't use `logbook` anymore, except in our "legacy" logging system (pre-1.0), which is still around to support a few quite old integrations that are still being deprecated.

For the new event/logging system, see:
[...]

PR #6994, opened 2023-02-16
Paw/ct 1844 log params take 2

### LINKED ISSUE #6639: [CT-1844] Improve Granularity of Log Configuration
### Describe the feature

This is a proposal which would subsume the following four issues:
- https://github.com/dbt-labs/dbt-core/issues/6580
- https://github.com/dbt-labs/dbt-core/issues/6312
- https://github.com/dbt-labs/dbt-core/issues/5208
- https://github.com/dbt-labs/dbt-core/issues/6615

The proposal is to modify log configuration so that the following command line help would be accurate, allowing for independent file and console log configuration of color, format, and level:

```
  --log-format {text,json,default}
  --log-format-file {text,json,default}
                        Specify the log format, overriding the command's
                        default. The standard option will apply the format
                        to both console and file output, unless the -file
                        variant (which can also appear alone) is used to 
                        override the log file format.
   --log-level {debug, info, warn, error, none}
   --log-level-file {debug, info, warn, error, none}
                        Set the logging level. Dbt will log events at or above
                        the severity level specified. The standard option will
                        apply to both console and file output, unless the
                        -file variant (which can also appear alone) is used to 
                        override the log file format. The 'none' level disables
                        all log output, including log file creation.
  --use-colors          Colorize (or do not colorize when using the 'no' 
  --use-colors-file     variants) dbt's log output. Output is colorized by
  --no-use-colors       by default and may also be set in a profile or at 
  --no-use-colors-file  the command line. The standard option will apply to
                        to both console and file output, unless the -file
                        variant (which can also appear alone) is used to 
                        override the log file colorization.
```

peterallenwebb:
resolves #6639

### Description

Adds the following cli parameters for finer grained control of console and file log behavior, as described in #6639:

--log-format
--log-format-file
--log-level
 --log-level-file
--use-colors 
--use-colors-file
--no-use-colors
--no-use-colors-file  the 

### Checklist


gshank:
This doesn't touch UserConfig (in core/dbt/contracts/project.py) where users can set log_format and use_colors. So I'm wondering how those settings integrate with these cli params, and whether we shouldn't allow setting the new params (somehow) in user config also. If I have a logging configuration that I want to use all the time, I would prefer not to have to pass in cli params every time, but have the settings persistent in user config.

peterallenwebb:
@gshank Yes, that is an interesting point which I did not consider. 

@jtcohen6 Do you agree that we should let users set these same flags in their user config, following the pattern of log_format and use_colors? I worry that the interaction patterns between all the different places these get set could start to become complicated, but Gerda's observations do make sense to me.

On the other hand, maybe Gerda's use case could be supported by the work we are considering for #6993.

jtcohen6:
@peterallenwebb @gshank Two-part answer:
1. Every "global config" that is settable via CLI flag + env var should also be settable in `UserConfig`. I know that is not the case today, but it is the documented behavior, and we ought to make it so. Let's do this for these logging parameters as well.
2. We should move `UserConfig` out of `profiles.yml`, and into its own standalone config file, in order to simplify the initialization flow for dbt-core. I agree it's a tangled web we weave today. My latest proposal for this: https://github.com/dbt-labs/dbt-core/issues/6207#issuecomment-1433007301

peterallenwebb:
@gshank Still waiting for the checks to pass here, but let me know whether the latest changes address your concerns.

core/dbt/cli/flags.py
```diff
@@ -172,6 +177,12 @@ def assign_params(ctx, params_assigned_from_default):
         for param in params:
             object.__setattr__(self, param.lower(), getattr(self, param))
 
+    # If the value of the lead parameter was set explicitly, apply the value to follow,
+    # unless follow was also set explicitly.
```
MichelleArk, line 181: tiny nit: documenting the method in a docstring below the signature would be more pythonic + consistent with the rest of this file.

core/dbt/events/functions.py
```diff
@@ -18,103 +18,146 @@
...
 
-def setup_event_logger(
-    log_path: str,
-    log_format: str,
-    use_colors: bool,
-    debug: bool,
-    log_cache_events: bool,
-    quiet: bool,
-) -> None:
+def setup_event_logger(flags) -> None:
```
MichelleArk, line 21: are we able to annotate the type of `flags` here?

core/dbt/events/functions.py
```diff
@@ -18,103 +18,146 @@
...
-    else:
         EVENT_MANAGER.add_logger(
-            _get_stdout_config(log_format, debug, use_colors, log_cache_events, quiet)
+            _get_logbook_log_config(
+                flags.DEBUG, flags.USE_COLORS, flags.LOG_CACHE_EVENTS, flags.QUIET
+            )
         )
+    else:
+        if flags.LOG_LEVEL != "none":
+            log_format = _line_format_from_str(flags.LOG_FORMAT, LineFormat.PlainText)
```
MichelleArk, line 33: is there a difference between `log_format` and `line_format`? If not, `line_format` would be more intuitive given the naming of `_line_format_from_str` + parameter name of `_get_stdout_config`.

MichelleArk (approved):
Just a couple small nits, otherwise LGTM!

core/dbt/events/functions.py
peterallenwebb: I agree, the variable being assigned here should be line_format. There is no difference between the two things, and the difference in naming is unfortunate.

PR #7002, opened 2023-02-17
Rename "constraints_enabled" to "contract"

### LINKED ISSUE #6748: [CT-1916] Rename 'constraints_enabled' to 'contract'
In the foundational [dbt Constraints work](https://github.com/dbt-labs/dbt-core/pull/6271), we added a boolean config named `constraints_enabled` that “turns on” contracting for this model. Given that the scope of model contracts is broader than “constraints,” and that dbt’s enforcement may go beyond the built-in behavior of analytical data platforms, we should rename this config to `contracted` instead.

gshank:
resolves #6748

### Description

After more discussion and considering the scope of this setting, we decided to rename  "constraints_enabled" to "contract".

### Checklist


MichelleArk (approved):
Changes here look good! There is some boilerplate in each adapter implementation that references `config.get('constraints_enabled')` ([dbt-snowflake example](https://github.com/dbt-labs/dbt-snowflake/pull/341/files#diff-79f8a1393fce3d64ce05238ec32cdf841b7ed89cf053ece9de0c50d763d3db79R25)), which will need to be updated as well. I can make corresponding issues in the adapter repos 👍

MichelleArk:
We should close the corresponding adapter issues prior to merging this to prevent breaking tests in adapters, since they inherit from [functional tests modified in this PR](https://github.com/dbt-labs/dbt-core/pull/7002/files#diff-8cb7e0269a9b9a6117115f688e3a6eafc9d4924df1a7e44a320f5cd06a3dd63cR22).

gshank:
We can't close the adapter issues prior to merging this because the "contract" config wouldn't actually exist in core until after this is merged. The has to merge first, then all of the adapter issues.

gshank:
Actually, I have a question about this now that I think about it. We're "promoting"  the "contract" attribute to the model level. Shouldn't we be accessing it via the model, like "model.contract", and not looking at the config? This is one of those fields where the config is just used to shuttle the value in node level...

gshank:
So actually the config.get calls should still work in the adapter repos since it doesn't require that the config attribute actually work. But if we did model.contract they might fail.

sungchun12:
@gshank thanks for evolving contracts by actually having the config called `contract`. Even that simple visual cue as I tested this out today was so much more intuitive than `constraints_enabled`.

Issue #6747, opened 2023-01-26
[CT-1915] [Epic] Multi-project collaboration - Milestone 1


MichelleArk:
Develop new constructs that enable dbt developers to create, contract, and communicate data models like software APIs. This work should enable more scalable mono-repos, while also laying the foundation for Milestone 2.

Discussions:
- https://github.com/dbt-labs/dbt-core/discussions/6725
- https://github.com/dbt-labs/dbt-core/discussions/6726
- https://github.com/dbt-labs/dbt-core/discussions/6730
- https://github.com/dbt-labs/dbt-core/discussions/6736

```[tasklist]
## [Model contracts](https://github.com/dbt-labs/dbt-core/discussions/6726)

* [ ] https://github.com/dbt-labs/dbt-core/pull/6271
* [ ] https://github.com/dbt-labs/dbt-core/issues/6748
* [ ] https://github.com/dbt-labs/dbt-core/issues/6750
* [ ] https://github.com/dbt-labs/dbt-core/issues/7066
* [ ] https://github.com/dbt-labs/dbt-core/issues/7067
* [ ] https://github.com/dbt-labs/dbt-core/issues/6751
* [ ] https://github.com/dbt-labs/dbt-core/issues/6754
* [ ] https://github.com/dbt-labs/dbt-core/issues/6756
* [ ] https://github.com/dbt-labs/dbt-core/issues/7034
* [ ] https://github.com/dbt-labs/dbt-core/issues/6755
* [ ] https://github.com/dbt-labs/dbt-core/issues/7154
* [ ] https://github.com/dbt-labs/dbt-core/issues/6975
* [ ] https://github.com/dbt-labs/dbt-core/issues/7064
* [ ] https://github.com/dbt-labs/dbt-core/issues/7184
* [ ] https://github.com/dbt-labs/dbt-core/issues/7209
* [ ] https://github.com/dbt-labs/dbt-core/issues/7121
```

```[tasklist]
## Model contracts - Adapters
* [x] Introducing constraints and contract:
* [ ] https://github.com/dbt-labs/dbt-snowflake/pull/341
* [ ] https://github.com/dbt-labs/dbt-redshift/pull/229
* [ ] https://github.com/dbt-labs/dbt-bigquery/pull/426
* [ ] https://github.com/dbt-labs/dbt-spark/pull/574
* [x] Renaming constraints_enabled to contract:
* [ ] https://github.com/dbt-labs/dbt-snowflake/issues/479
[...]

PR #7007, opened 2023-02-18
CT 1990 add access property to parsed nodes

### LINKED ISSUE #6824: [CT-1990] Parse access property on model node
### Requirements
- `access` is a new property on model nodes. For phase 1, cannot be provided for any other ['groupable' nodes](https://github.com/dbt-labs/dbt-core/issues/6823)
- `access` property should be defined as an enum, for potential extension in the future. Currently must be one of `private` or `public`, or `protected`:
* private - ref works within the same group
* protected - ref works within the same project/package
* public - ref works across projects/packages
- `access` defaults to `protected`, irrespective of the group config value

### Reference spec
```yaml
# models/marts/github/github.yml

groups:
  - name: github
    owner: 
      name: Jeremy
      email: data@jer.co

models:
  - name: int__github_issue_label_history
    access: private
    config: 
      group: github  # explicit opt-in
  - name: fct_github_issues
    access: protected
    config: 
      group: github  # explicit opt-in
  - name: fct_github_issues
    access: public
```

gshank:
resolves #6824

### Description

Add access property to parsed nodes. (Did not add to metrics.) Update tests and bump schema version.

### Checklist


MichelleArk:
Would adding the `access` property to parsed nodes would enable configuring `access` on all `ManifestNode`s?

```python
ManifestSQLNode = Union[
    AnalysisNode,
    SingularTestNode,
    HookNode,
    ModelNode,
    RPCNode,
    SqlNode,
    GenericTestNode,
    SnapshotNode,
]

# All SQL nodes plus SeedNode (csv files)
ManifestNode = Union[
    ManifestSQLNode,
    SeedNode,
]
```
(From: https://github.com/dbt-labs/dbt-core/blob/main/core/dbt/contracts/graph/nodes.py#L1153-L1168)

gshank:
The code and the tests assume that all of those nodes have the same set of attributes. It's certainly possible to limit it to just ModelNode and SnapshotNode, but it's actually rather more work, which is why I did it this way for commenting. Are we going to limit "group" to the same set of nodes?

gshank:
Also the way that "patches" are processed, all of the manifest nodes are handled with the same code, so unless we split that up we still have to add "access" to ParsedNodePatch. Would we want to throw an error if somebody tried to set "access" (or group?) for non-supported node types?

MichelleArk:
> Are we going to limit "group" to the same set of nodes?

Nodes that should be 'groupable' are nodes that have the ability to `ref` other nodes. On the other hand, nodes that can have the `access` property are nodes that are 'refable'. So no - they should be different sets of nodes. 

The configured group value will be used to determine whether the `ref` is permitted or not for `private` models. If there is _no_ configured group on a model accessing a private model, the `ref` should be restricted.
 
> Would we want to throw an error if somebody tried to set "access" (or group?) for non-supported node types?

I'd say an error or at least a warning, given that the user may be expecting dbt to actually something given the set `access` property and we have no plans to act on this property value for other nodes in [Phase 1 of Multi-project deployments](https://github.com/dbt-labs/dbt-core/issues/6747)

> It's certainly possible to limit it to just ModelNode and SnapshotNode...
I'd actually advocate to limit this to just `ModelNode`, since that is the most concrete product story we have around multi-project deployments. Ultimately, a snapshot or seed could be wrapped in a model to use the `access` property. Would definitely be open to relaxing this given community / user feedback though.

gshank:
Note: we didn't move to version 9 of the manifest schema when we should have, so I updated all the v8 manifests that shouldn't have the new fields in them, and switched to v9.

core/dbt/contracts/util.py
```diff
@@ -298,6 +298,8 @@ def upgrade_manifest_json(manifest: dict) -> dict:
         if "root_path" in doc_content:
             del doc_content["root_path"]
         doc_content["resource_type"] = "doc"
+    if "groups" not in manifest:
+        manifest["groups"] = {}
```
MichelleArk, line 302: Looks like this is a duplicate of lines 279-290: https://github.com/dbt-labs/dbt-core/pull/7007/files#diff-f009cc111693f698a64598745128407e594cb9679e14b4c40faf6c072699c897R279-R280

gshank: Ooops. It was failing with "no groups field", but that was because I hadn't yet updated the version below which we need to update the manifest. Removed.

core/dbt/contracts/graph/nodes.py
```diff
@@ -365,6 +367,28 @@ def patch(self, patch: "ParsedNodePatch"):
...
+            if self.resource_type == NodeType.Model:
+                if AccessType.is_valid(patch.access):
+                    self.access = AccessType(patch.access)
+                else:
+                    warn_or_error(
+                        InvalidValueForField(
+                            field_value=patch.access,
+                            field_name="access",
+                        )
+                    )
```
MichelleArk, line 383: Instead of a warning (which a user can promote to an error), Is it possible to raise a `ParsingError` here instead? I think that's more consistent with existing product expectations for other entities. For example, configuring `type: invalid` on an exposure gets: 

> dbt.exceptions.YamlParseDictError: Parsing Error
  Invalid exposures config given in FilePath(searched_path='models', relative_path='schema.yml', modification_time=1677270162.964944, project_root='/Users/michelleark/src/jaffle_shop') @ exposures: {'name': 'jaffle_shop_report', 'type': 'invalid', 'owner': {'name': 'testsss'}} - at path ['type']: 'invalid' is not one of ['dashboard', 'notebook', 'analysis', 'ml', 'application']

MichelleArk: Also, can we define `access` on the dataclass declaration for the `ParsedNode`, simialr to the other class attributes set by this method (created_at, description, patch_path). Ideally `access` is a `AccessType` if possible.

gshank: Sure. It's an exception, so it aborts everything, which is not my favorite pattern because if you have multiple errors you have to address them one at a time, but we can certainly do that.

gshank: I'm not sure what you're suggesting with regard to ParsedNode. It's set in the dataclass declaration for ModelNode, which is where you wanted it.

MichelleArk: >I'm not sure what you're suggesting with regard to ParsedNode. It's set in the dataclass declaration for ModelNode, which is where you wanted it.

I see, makes sense. Still getting my head around patches. Sorry for the confusion!

MichelleArk: > which is not my favorite pattern because if you have multiple errors you have to address them one at a time

Agree - but I'm nervous about gracefully handling this invalid config because, given that the default `protected` value is not the 'most' restricted access option. For example, a typo to `access:priivate` would lead to `access:protected` on the model, which is probably a more open level of access than the user meant to specify. Hopefully they'd catch it in review / via logs but I think it's worth raising an exception for this case.

MichelleArk (approved):

PR #7008, opened 2023-02-19
[CT-1584] New top level commands: interactive compile

### LINKED ISSUE #6358: [Feature] New top level commands: interactive compile

### Describe the feature

In `dbt-rpc` and `dbt-server`, we support using some kind of manifest object(in mem object for `dbt-rpc` and `manifest.msgpack` for `dbt-server`) that regenerated whenever a dbt project file got modified to to support interactive compile.
Here's the link of how it is done in [`dbt-rpc`](https://github.com/dbt-labs/dbt-rpc/blob/main/dbt_rpc/task/sql_commands.py#L207) and [`dbt-server`](https://github.com/dbt-labs/dbt-server/blob/main/dbt_server/views.py#L430)

We would want to support this as a top level dbt-command where user would be able to compile 1 file/string leveraging existing state of project(likely manifest.msgpack generated by previous command)

Open questions for @jtcohen6 :
1. We need a mock command that shows what invoking this looks like. Is this what it should look like?
    ```
    dbt compile_sql --state {somepath to manifest.msgpack, optional} file_name
    ```
    then we print the compiled code out?
2. how do we imagine update that `manifest.msgpack` when something changes but before user do another command that could generate that file? Should we just load the outdated ones? or should we try to do something like partial parsing to check project and determine whether we need a full parsing?

_No response_

### Who will this benefit?

_No response_

_No response_

### Anything else?

_No response_

aranke:
resolves #6358 

### Checklist

    - https://github.com/dbt-labs/docs.getdbt.com/issues/2994

### DIFF
--- a/core/dbt/task/compile.py
+++ b/core/dbt/task/compile.py
@@ -61,7 +61,7 @@ class CompileTask(GraphRunnableTask):
         return CompileRunner
 
     def task_end_messages(self, results):
-        if self.args.inline and len(results) == 1:
+        if getattr(self.args, "inline", None):
             result = results[0]
             fire_event(
                 CompiledNode(node_name=result.node.name, compiled=result.node.compiled_code)


core/dbt/cli/flags.py
```diff
@@ -42,6 +43,7 @@
     "fail_fast",
     "indirect_selection",
     "store_failures",
+    "introspect",
```
ChenyuLInx, line 46: why we want to add introspect here?

core/dbt/events/types.proto
```diff
@@ -1600,7 +1600,16 @@ message ConcurrencyLineMsg {
     ConcurrencyLine data = 2;
 }
 
-// Skipped Q028
+// Q028
```
ChenyuLInx, line 1603: Just curious, what is Q stand for?

core/dbt/task/base.py
```diff
@@ -309,7 +310,7 @@ def skip_result(self, node, message):
 
     def compile_and_execute(self, manifest, ctx):
         result = None
-        with self.adapter.connection_for(self.node):
+        with self.adapter.connection_for(self.node) if get_flags().INTROSPECT else nullcontext():
```
ChenyuLInx, line 313: If a user set `DBT_INTROSPECT` to False, but we are not running compile/preview command, `get_flags().INTROSPECT` will be `True` since it is using default value right?

core/dbt/task/compile.py
```diff
@@ -1,17 +1,17 @@
...
 from dbt.exceptions import DbtInternalError, DbtRuntimeError
 from dbt.graph import ResourceTypeSelector
-from dbt.events.functions import fire_event
-from dbt.events.types import CompileComplete
-from dbt.parser.manifest import write_manifest
 from dbt.node_types import NodeType
+from dbt.parser.manifest import write_manifest, process_node
+from .base import BaseRunner
+from .runnable import GraphRunnableTask
+from ..parser.sql import SqlBlockParser
```
ChenyuLInx, line 14: Nit: Looks like we are using `from dbt.parser.xxx import xxx` above, do we want to standarize on one(Could be either one)

ChenyuLInx (commented):
Looks great! Thanks for traversing through the codebase and add this feature!

I did run into one issue that running `dbt compile -s customers`(I have a model named that), compiled result was not logged.

The rest of the comments are mostly informational/confirm things.

stu-k (approved):
With the exception of those relative imports, nothing stands out 👍

core/dbt/tests/util.py
```diff
@@ -160,6 +161,11 @@ def write_file(contents, *paths):
         fp.write(contents)
 
 
+def file_exists(*paths):
+    """Check if file exists at path"""
+    return os.path.exists(os.path.join(*paths))
+
+
```
dbeatty10, line 168: Not a problem to solve in this PR, but I wonder why `os.path` was used  in this module rather than `pathlib`?

dbeatty10: Yep, going with the flow make sense in this PR.

Something for us to consider in a refactor down the road though.

core/dbt/events/types.py
```diff
@@ -1711,7 +1711,13 @@ def message(self) -> str:
...
 
 
-# Skipped Q028
+@dataclass
+class CompiledNode(InfoLevel, pt.CompiledNode):
+    def code(self):
+        return "Q028"
+
+    def message(self) -> str:
+        return f"Compiled node '{self.node_name}' is:\n{self.compiled}"
```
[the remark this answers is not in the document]
dbeatty10: Do you have an example what this message would look like with realistic values?

core/dbt/task/compile.py
```diff
@@ -43,19 +43,32 @@ def raise_on_first_error(self):
         return True
 
     def get_node_selector(self) -> ResourceTypeSelector:
+        if getattr(self.args, "inline", None):
+            resource_types = [NodeType.SqlOperation]
+        elif getattr(self.args, "select", None):
```
ChenyuLInx, line 48: This will be a behavior change, we can do it in task_end_message if that change is not acceptable.(I don't know whether it is or not)

core/dbt/task/compile.py
```diff
@@ -43,19 +43,32 @@ def raise_on_first_error(self):
...
+            resource_types = [NodeType.Model]
```
jtcohen6, line 49: Discussed offline yesterday with @aranke @ChenyuLInx:

This is to handle the fact that `--select my_model` will also select the tests on `my_model`, and we really only want to be compiling one node. The right way to handle this behavior is to turn off "indirect selection," and the cleanest approach there is probably to introduce a new option to turn off indirect selection entirely (`--indirect-selection empty`).

Depending on how much work is involved there, we can scope & estimate that as a separate issue.

tests/functional/compile/test_compile.py
```diff
@@ -0,0 +1,131 @@
...
+        (results, log_output) = run_dbt_and_capture(
+            ["compile", "--inline", "select * from {{ ref('first_model') }}"]
+        )
+        assert len(results) == 1
+        assert "Compiled node 'inline_query' is:" in log_output
+
+    def test_select_pass(self, project):
+        (results, log_output) = run_dbt_and_capture(["compile", "--select", "second_model"])
+        assert len(results) == 3
+        assert "Compiled node 'second_model' is:" in log_output
```
ChenyuLInx, line 118: I thought in this case we would't print the compiled result for second model?

core/dbt/cli/flags.py
ChenyuLInx: Do you mean removing it would cause things to fail? This doesn't feel right given that `introspect` is only a parameter for `compile` task command, and all parameter in this list is defiled both as `cli` click group level and a task command level.

ChenyuLInx (commented):
Looks good over all! Last two questions

tests/functional/compile/test_compile.py
ChenyuLInx: Then `--indirect-selection empty ` is only to make sure we don't compile test etc? makes sense to me!

core/dbt/task/compile.py
dbeatty10: Did y'all consider any alternative names to `empty`? For example, `none` feels more intuitive to me, but maybe y'all considered it already.

Adding `empty` to the [current options](https://docs.getdbt.com/reference/node-selection/test-selection-examples#indirect-selection) would give:
1. `eager` (default) - include ANY test that references the selected nodes
1. `cautious` - restrict to tests that ONLY refer to selected nodes
1. `buildable` - restrict to tests that ONLY refer to selected nodes (or their ancestors)
1. `empty` - don't include any tests

If it is named something like `none`, then the list would be this instead:
1. `eager`
1. `cautious`
1. `buildable`
1. `none`

core/dbt/task/compile.py
```diff
@@ -43,19 +43,40 @@ def raise_on_first_error(self):
...
             previous_state=self.previous_state,
-            resource_types=NodeType.executable(),
+            resource_types=resource_types,
         )
 
     def get_runner_type(self, _):
         return CompileRunner
 
     def task_end_messages(self, results):
+        if self.args.inline and len(results) == 1:
```
ChenyuLInx, line 64: Is there a case where `inline` is `True` but `len(results)` is not 1?

ChenyuLInx (approved):
LGTM with one question

dbeatty10:
@aranke Do we have an issue opened in the [docs.getdbt.com repo](https://github.com/dbt-labs/docs.getdbt.com/issues/new?assignees=&labels=content%2Cupcoming+release&template=new-dbt-feature.yml) for this new feature?

If not, then we'd want to fill out the template [here](https://github.com/dbt-labs/docs.getdbt.com/issues/new?assignees=&labels=content%2Cupcoming+release&template=new-dbt-feature.yml).

We'd want to add the `dbt-core v1.5` label so that it gets tracked with the other [1.5 features](https://github.com/dbt-labs/docs.getdbt.com/issues?q=is%3Aissue+label%3A%22dbt-core+v1.5%22+) that need documentation.

dbeatty10:
> @dbeatty10 done, thank you! [dbt-labs/docs.getdbt.com#2994](https://github.com/dbt-labs/docs.getdbt.com/issues/2994)

Sure thing @aranke !

Could you also add the `dbt-core v1.5` label to https://github.com/dbt-labs/docs.getdbt.com/issues/2994 ?

I just updated the PR checklist so it's easy to see the linked issue and its status:

image

Issue #6749, opened 2023-01-26
[CT-1917] [Regression] Non-error log is not silenced by `-q` flag.


elongl:
### Is this a regression in a recent version of dbt-core?

- [X] I believe this is a regression in dbt-core functionality
- [X] I have searched the existing issues, and I could not find an existing issue for this regression

### Current Behavior

According to `-q`:
> Suppress all non-error logging to stdout. Does not affect {{ print() }} macro calls.

A new log is printed even though it's not an error.

```bash
(elementary) ☁  dbt_project [main] dbt run-operation hello
14:22:47  target not specified in profile 'elementary', using 'default'
14:22:47  Running with dbt=1.4.0
hello
(elementary) ☁  dbt_project [main] dbt -q run-operation hello
14:22:56  target not specified in profile 'elementary', using 'default'
hello
```

### Expected/Previous Behavior

Previously, the log was omitted when passing the `-q` flag.

```bash
(elementary) ☁  dbt_project [main] dbt run-operation hello
14:22:15  target not specified in profile 'elementary', using 'default'
14:22:16  Running with dbt=1.3.0
hello
(elementary) ☁  dbt_project [main] dbt -q run-operation hello
hello
```

### Steps To Reproduce

1. Name a `default` target without having a `target` defined.
2. Run a dbt command with the `-q` flag.

### Environment

```markdown
- OS: macOS
- Python: 3.9.6
- dbt (working version): 1.3.0
- dbt (regression version): 1.4.0
```

### Additional Context

I'd love to contribute.

jtcohen6:
My suspicion here is:
- We set up a basic event logger before we have the full set of information we need to initialize it for real (all flags & configs resolved). This ensures that any real initialization errors don't get swallowed. Relevant: #6434, #6447.
- The event you're seeing, `MissingProfileTarget`, is raised during `render_profile`, which is happening during initialization and before we call `setup_event_logger` here:
https://github.com/dbt-labs/dbt-core/blob/c65ba11ae6dca5203228320a7f6388ce1796f104/core/dbt/main.py#L225-L232

Should we simply

elongl:
Hi @jtcohen6 , thanks a lot for replying.
I personally tend to favor (2) because I think that if a user is using `--quiet` they'd generally like to keep a program quiet at its entirety except for logs made by them (`print()`) and it doesn't really matter at what phase the execution is.
What do you think?

If that resonates, care to share any guidelines on how can I make this contribution?

[...]

PR #7048, opened 2023-02-24
CT-2022: Fix --quiet and DBT_ENABLE_LEGACY_LOGGER during initialization

### LINKED ISSUE #6847: [CT-2022] Non-error log is not silenced by -q flag, Click Version

peterallenwebb:
resolves #6847

### Description

Checks the values of --quiet and --enable-legacy-logger after the cli and env parameters are parsed, but before the user profile is parsed. This way, the value of those parameters can be respected even for events which are fired during profile parsing.

### Checklist


peterallenwebb:
@aranke I'm glad you prodded me to write a test, because once I started on that I realized that I could not reproduce the original issue on main, and that in fact we do not (as far as I could see after searching for 10-15 minutes) have any code paths which would potentially fire non-error events before the call to setup_event_logger() in requires.py.

I'm curious whether you think it's worth merging the code under review at all. It would protect us from regressions to the desired behavior if anyone were to ever fire events before log initialization in the future, but that's the only advantage I can imagine. In the meantime it's hard to create a test which would fail without the fix under review.

jtcohen6:
> we do not (as far as I could see after searching for 10-15 minutes) have any code paths which would potentially fire non-error events before the call to setup_event_logger() in requires.py

@peterallenwebb The original regression report in #6749 was for the `MissingProfileTarget` event.

Repro case:
1. Define a `profile` in `profiles.yml` that has a target named `default`, and doesn't have `target: default` in it
2. Use that profile (`profile` field in `dbt_project.yml`)
3. `dbt run`

FWIW, this seems to "just work" on the `main` branch, even without this change:
```
$ dbt run
17:55:10  Running with dbt=1.5.0-b1
17:55:10  target not specified in profile 'garage-postgres', using 'default'
17:55:10  Unable to do partial parsing because profile has changed
17:55:11  Found 0 models, 0 tests, 0 snapshots, 0 analyses, 298 macros, 0 operations, 1 seed file, 0 sources, 0 exposures, 0 metrics, 0 groups
17:55:11
17:55:11  Nothing to do. Try checking your model configs and model specification args
$ dbt --quiet run
```

I'm not sure exactly how, but: I was under the impression that `Flags` read `user_config`, and therefore the `profiles.yml`, and therefore would raise this `info`-level message, before `setup_event_logger` is called. But it looks like it's going down a different codepath — `read_profile` instead of `render_profile`.

peterallenwebb:
@jtcohen6 I was under the same impression you were.

What's happening is that `preflight()` (triggered by the `@requires.profile` decorator on the click command) completes before the profile is completely analyzed. While `profiles.yml` is loaded during `Flags.__init__`, it is not analyzed or validated in any depth. Later, `load_profile()` (triggered by the `@requires.profile` decorator) loads the profile _again_, this time performing validation which may fire the MissingProfileTarget warning.

So, we don't need to do anything to resolve the original issue on main. 

The only question is whether it is worth it to keep this fix in place, just in case someone adds non-error events to the Flag generation code, or any other code which they might execute between command line parsing and log configuration. I am leaning toward no at the moment, since it adds complexity, but happy to let you make the call.

jtcohen6:
Thanks for digging in to confirm!

> The only question is whether it is worth it to keep this fix in place, just in case someone adds non-error events to the Flag generation code, or any other code which they might execute between command line parsing and log configuration. I am leaning toward no at the moment, since it adds complexity, but happy to let you make the call.

I agree with "no."

In the medium term, we should stop reading `profiles.yml` _at all_ during app initialization (including `Flags`), by moving `UserConfig` to its own separate file.

I called out in https://github.com/dbt-labs/dbt-core/issues/6207#issuecomment-1433007301 that we shouldn't do any more-complex logging while initializing configs:

> If we can locate `dbt_config.yml`, and there's an error while reading it, it's always an error. We immediately raise the error and stop.

Basically: Logging is not allowed until we have an initialized app, including a fully initialized & configured logger. That codepath should be incredibly clear, and if anything goes wrong during that initialization, it's a show-stopping error.

peterallenwebb:
Closing based on the consensus that this is already fixed on main, and no further work is warranted.

PR #7069, opened 2023-02-27
CT 1993 handle invalid access to private models

### LINKED ISSUE #6826: [CT-1993] Raise an exception from ref if referencing a private model across groups
During resolution, raise an exception if a node attempts to ref a private model from a different group than the referenced private model’s group. 

Nodes not in a group (group is `None`) should not be able to reference _any_ private models because they are not part of any group.

Note: exposures can have `ref` calls, but are not 'groupable'. Following on the logic above, they should _not_ be able to reference _any_ private models because they are not part of any group.

gshank:
resolves #6826

### Description

Throw a DbtReferenceError if a private model is accessed from a non-group model, or from an exposure or metric.

### Checklist


core/dbt/parser/schemas.py
```diff
@@ -788,6 +788,7 @@ def parse(self) -> List[TestBlock]:
                 refs = ParserRef()
 
             # There's no unique_id on the node yet so cannot add to disabled dict
+            print(f"--- node_plock in parse: {node_block}")
```
MichelleArk, line 791: s/plock/block

gshank: Oops. Just some testing I did that got left in.

.changes/unreleased/Under the Hood-20230227-115821.yaml
```diff
@@ -0,0 +1,6 @@
+kind: Under the Hood
+body: Throw errors for invalid model access
+time: 2023-02-27T11:58:21.133458-05:00
+custom:
+  Author: gshank
+  Issue: "6826"
```
MichelleArk, line 6: Is a second changelog entry necessary?

core/dbt/parser/manifest.py
```diff
@@ -1239,6 +1248,14 @@ def _process_refs_for_metric(manifest: Manifest, current_project: str, metric: M
                 should_warn_if_disabled=False,
             )
             continue
+        elif (
+            target_model.resource_type == NodeType.Model
+            and target_model.access == AccessType.Private
+        ):
+            # Metrics do not have a group and so can never reference private models
```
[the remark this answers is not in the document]
gshank: There is no "group" field on a Metric node, but I see that there is on MetricConfig, so if it were set in config on a metric it would blow up in "update_parsed_node_config". Should I add a "group" field on Metric?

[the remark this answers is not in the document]
gshank: Actually metrics doesn't even use update_parsed_node_config, so it wouldn't have blown up, but there wasn't a "group" field on metric. I added the group field and copied it from config to node level. I used the node level field in '_process_refs_..." calls.  If we're going to copy that field to node level we should probably use that consistently. Or not bother with the copying.

core/dbt/exceptions.py
```diff
@@ -1114,6 +1114,19 @@ def __init__(self, exc: ValidationError, node):
...
+class DbtReferenceError(ParsingError):
+    def __init__(self, unique_id: str, ref_unique_id: str):
+        self.unique_id = unique_id
+        self.ref_unique_id = ref_unique_id
+        super().__init__(msg=self.get_message())
+
+    def get_message(self) -> str:
+        return (
+            f"Node {self.unique_id} attempted to reference node {self.ref_unique_id}, "
+            "which is not allowed because the referenced node is private."
```
MichelleArk, line 1126: Could we also add the target node's group as part of this error message? Perhaps something like:
"...because the referenced node is private to the {group} group."

gshank: Updated.

core/dbt/exceptions.py
```diff
@@ -1114,6 +1114,19 @@ def __init__(self, exc: ValidationError, node):
...
+class DbtReferenceError(ParsingError):
+    def __init__(self, unique_id: str, ref_unique_id: str, group: str):
+        self.unique_id = unique_id
+        self.ref_unique_id = ref_unique_id
+        super().__init__(msg=self.get_message())
+
+    def get_message(self) -> str:
+        return (
+            f"Node {self.unique_id} attempted to reference node {self.ref_unique_id}, "
+            "which is not allowed because the referenced node is private to the {self.group} group."
```
MichelleArk, line 1126: Error message looks good, but this is missing the f-string + initializion self.group

MichelleArk (approved):

PR #7100, opened 2023-03-01
CT 2251 model yaml frontmatter

### LINKED ISSUE #7099: [CT-2251] Inline model documentation + tests
As per discussion in #6853, add code to process yaml frontmatter in model nodes. Copying information from that discussion by jtcohen:

What about YFM?
Here's an idea, which you have every right to hate: What if we did this as yaml front matter (YFM)? After scanning through the older linked issues, I didn't see this come up; I may have missed it, or I may be proposing something that everyone else has tacitly agreed would be terrible. For whatever reason, I don't hate the idea, at least not at first glance.

Idea being, these remain two separate & independent sets of information, which just happen to live in the same file. The yaml in the front matter can include any of the key-value pairs that could also be passed to the models: list in a yaml file. It would be parsed at the same time as the rest of the model file, but using our yaml parser, instead of the ModelParser.

The front matter would either take priority over, or be mutually exclusive with, defining yaml properties for this model in a separate yaml file. If the model also defines {{ config() }} (Jinja-SQL) or dbt.config() (Python), it would take precedence over, or be mutually exclusive with, the config defined in the front matter.

```
---
description: This is my model's description. It can reference a {{ doc('block') }} if it wants to.
config:
  materialized: incremental
  unique_key: id
columns:
  - name: first_col
    tests:
      - unique
      - not_null
---

select ...
---
columns:
  - name: first_col
    tests:
      - unique
      - not_null
---

def model(dbt, session):
    ...

```

The model file might get too big (dozens of columns in a model might be confusing, especially if all the docs are inlined)

This approach definitely doesn't help with the sheer model size, but yaml does offer a few line-saving options if you're willing to get a bit JSON-y:

```
---
columns:
  - {name: first_col, tests: [not_null, unique]}
  - {name: second_col, tests: [not_null], description: ...}
  - ...
---

select ...
```

To be clear, what I'm not proposing here is the ability to use Jinja or yaml to dynamically create/populate the front matter contents. I do think that remains, rightly, the domain of code-generation plugins/packages. (Our work on refactoring dbt-core's foundational APIs could eventually unlock even more-capable code generation, in the form of more-direct node generation.)

gshank:
resolves #7099

### Description

This is a work in progress, not at all ready for code review. It's here for commenting and further development.

The changes in core/dbt/clients/yaml_helper.py came from a pull request by jakebiesinger (https://github.com/jakebiesinger/dbt-core/pull/1)

### Checklist

### DIFF
--- a/core/dbt/clients/yaml_helper.py
+++ b/core/dbt/clients/yaml_helper.py
@@ -1,5 +1,6 @@
+import re
 import dbt.exceptions
-from typing import Any, Dict, Optional
+from typing import Any, Dict, Optional, Tuple
 import yaml
 
 # the C version is faster, but it doesn't always exist
@@ -8,6 +9,9 @@ try:
 except ImportError:
     from yaml import Loader, SafeLoader, Dumper  # type: ignore  # noqa: F401
 
+FRONTMATTER_CHECK = ["---\n", "---\r\n"]
+FRONTMATTER_DELIMITER = re.compile(r"^---$", re.MULTILINE)
+NON_WHITESPACE = re.compile(r"\S")
 
 YAML_ERROR_MESSAGE = """
 Syntax error near line {line_number}
@@ -61,3 +65,47 @@ def load_yaml_text(contents, path=None):
             error = str(e)
 
         raise dbt.exceptions.DbtValidationError(error)
+
+
+def split_yaml_frontmatter(content: str, original_file_path: str) -> Tuple[Optional[str], str]:
+    """Splits `content` into raw YAML frontmatter (as a raw string) and everything else proceeding.
+
+    Frontmatter is defined as a block of YAML appearing between two `---` tokens in an otherwise non-YAML document.
+    The frontmatter must be placed at the beginning of the file: anything other than whitespace preceding the first `---`
+    will cause the frontmatter block to be ignored, with a warning.
+    """
+    parts = FRONTMATTER_DELIMITER.split(content, 2)
+    if len(parts) != 3:
+        # Zero or one `---` token, so return the original string
+        return None, content
+    elif NON_WHITESPACE.search(parts[0]) is not None:
+        # No frontmatter section or non-whitespace preceding the first `---`, so skip frontmatter
+        return None, content
+
+    frontmatter_content, after_footer = parts[1:]
+    return frontmatter_content, after_footer
+
+
+def parse_yaml_frontmatter(
+    frontmatter_content: str, original_content: str
+) -> Optional[dict[str, Any]]:
+    try:
+        parsed_yaml = safe_load(frontmatter_content)
+    except (yaml.scanner.ScannerError, yaml.YAMLError) as e:
+        if hasattr(e, "problem_mark"):
+            error = contextualized_yaml_error(original_content, e)
+        else:
+            error = str(e)
+        error = f"Error parsing YAML frontmatter:  {error}"
+        raise dbt.exceptions.DbtValidationError(error)
+
+    return parsed_yaml
+
+
+def maybe_has_yaml_frontmatter(content: str) -> bool:
+    """Return if `content` *might* have YAML frontmatter
+
+    This weak filter allows us to skip the more-expensive regex + YAML parsing.
+    """
+    # The manual [0] and [1] here are perf optimizations.
+    return FRONTMATTER_CHECK[0] in content or FRONTMATTER_CHECK[1] in content
--- a/core/dbt/clients/yaml_helper.py
+++ b/core/dbt/clients/yaml_helper.py
@@ -67,7 +67,7 @@ def load_yaml_text(contents, path=None):
         raise dbt.exceptions.DbtValidationError(error)
 
 
-def split_yaml_frontmatter(content: str, original_file_path: str) -> Tuple[Optional[str], str]:
+def split_yaml_frontmatter(content: str) -> Tuple[Optional[str], str]:
     """Splits `content` into raw YAML frontmatter (as a raw string) and everything else proceeding.
 
     Frontmatter is defined as a block of YAML appearing between two `---` tokens in an otherwise non-YAML document.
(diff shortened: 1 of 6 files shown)


joellabes:
@gshank A drive-by comment, based on [this message](https://github.com/dbt-labs/dbt-core/discussions/6853#discussioncomment-5172074):
> There's still a fair amount of work to do [...] dealing with config in both the SQL files and a schema file, etc

I don't think we should support this! [I think](https://github.com/dbt-labs/dbt-core/discussions/6853#discussioncomment-5074299) you should only be allowed to have YAML in the SQL file itself _or_ a schema file:
>It feels good to me that there is not a new layer of inheritance and override order of operations for people to get their heads around.
>
>☝️ This is provided that if they try, they get a warning that some properties are being ignored, and the winner is chosen in a deterministic way. I don't really care which one wins if one is technically easier than the other, but all else being equal I'd vote for YFM being prioritised over schema.yml

(To reiterate, this is just what I reckon, not an official spec)

gshank:
I wasn't clear enough in my comment... I didn't mean that we'd support doing config in both places, I mean figuring out how to detect that it was happening and issue an error....

jtcohen6:
Quoting @jakebiesinger in https://github.com/dbt-labs/dbt-core/discussions/6853#discussioncomment-5001630:

> There is https://pypi.org/project/python-frontmatter/ which I'm using in my PR above, but I looked around at the internals and it's literally just [regex parsing](https://github.com/eyeseast/python-frontmatter/blob/main/frontmatter/default_handlers.py#L233) combined with a `documentString.split(regex)`. This unfortunately accepts frontmatter from anywhere in the doc (not something we can configure, even by implementing a custom handler in that world. While there aren't many deps for that package, the latest release is March 2021...
> 
> Honestly, I think I'd rather provide our own stricter and simpler parser, given the above's flexibility + complexity. A few dozen lines of code is not much of a price to pay for parsing exactly how you want IMHO.

gshank:
That was the same conclusion that I came to after looking at it. It's doing more than we want and the actually relevant code is pretty small.

gshank:
This only implements yaml frontmatter for model nodes. Would we want to also support it for other node types?

core/dbt/contracts/graph/nodes.py
```diff
@@ -684,6 +686,19 @@ def same_contents(self, other) -> bool:
     def test_node_type(self):
         return "generic"
 
+    def write_node(self, target_path: str, subdirectory: str, payload: str):
+        # If this test came from yaml frontmatter, it will try to write to a directory
+        # with the same name as the model file, which won't work. So change the
+        # directory to model file name + .yaml.
+        file_path = self.original_file_path
+        if file_path.endswith(".sql"):
```
jtcohen6, line 694: - Need to check for `".py"` here as well, for Python models
- Maybe it should just write to the same directory that the model is defined in? Need to think a bit more about this. I don't think it's a high-stakes decision one way or the other

jtcohen6 (commented):
> This only implements yaml frontmatter for model nodes. Would we want to also support it for other node types?

If we're thinking about this as a proof of concept, `model` is definitely the node type most worth proving out :)

We could extend this to other node types that are 1:1 with a file (singular tests, analyses), mostly for the sake of consistency. I think resources that are defined as Jinja blocks (snapshots, macros), where you can have multiple in a file, would be tricky. Even though they're one-to-a-file, I think seeds would also be tricky, and probably not the right UX. 

Not sure if you also tested with Python models... but it (mostly) "just works"! Which is pretty darn cool.
```py
---
description: "This is my dbt-py model"
config:
  alias: my_cool_alias
---

def model(dbt, session):
    df = dbt.ref("my_model")
    return df
```

core/dbt/contracts/graph/nodes.py
gshank: I *think* that if we put it in the same directory that the model is in we could have collisions with other tests from other models in the same directory.

jtcohen6:
Assigning myself to take this for another spin :)

@gshank If we decide this is good-to-go from a user/product perspective, how much reworking would you want to do before putting this up for formal review of the code/implementation?

gshank:
If we want to start out by just supporting models, I think it's good to review. There might eventually be some additional work related to some of the upcoming version project relating to checking that this is the only yaml for a group of versions.

gshank:
Oh, I was wondering about the location of the new "yaml_config_dict" key. I put it in the node, but it could go somewhere else like the file object.

core/dbt/clients/yaml_helper.py
```diff
@@ -61,3 +65,42 @@ def load_yaml_text(contents, path=None):
...
+        error = f"Error parsing YAML frontmatter:  {error}"
+        raise dbt.exceptions.DbtValidationError(error)
+
+    return parsed_yaml
+
+
+def has_yaml_frontmatter(content: str) -> bool:
+    """Check first line for yaml frontmatter"""
+
+    return content.startswith(FRONTMATTER_CHECK[0]) or content.startswith(FRONTMATTER_CHECK[1])
```
jakebiesinger, line 106: I think `content` could start with whitespace (including newlines) and this method would return `False`, while `split_yaml_frontmatter` would skip over leading whitespace.

Perhaps this would be better?

```suggestion
    stripped = content.lstrip()
    return stripped.startswith(FRONTMATTER_CHECK[0]) or stripped.startswith(FRONTMATTER_CHECK[1])
```

gshank: Content should be stripped already, I think, when it's loaded in read_files. Did you actually see an error? Wouldn't hurt to double up on that though.

jakebiesinger: Nope! I honestly just drove by and looked over your approach. Cool that it's already stripped. It might be worth a test case?

callum-mcdata:
@gshank is there any chance this might slip in with 1.7? Been writing a lot more dbt-sql these days and keep coming back to how much I would **love** yaml frontmatter!

gshank:
Unfortunately we are right on top of code freeze for 1.7. I'll update the branch though, so we can see what's involved.

guizsantos:
Just for the record, I was just thinking about this, came up with exactly the idea to implement YFM and eventually found the disucssion, issue and finally, very gladly, this PR! I would really love to see this getting released and for that, I'd be happy to contribute!

zqureshi:

PR #7101, opened 2023-03-01
Selector Upgrade

### LINKED ISSUE #5009: [CT-469] Combine --select/--exclude with --selector when used together?

### Current Behavior

I try to test my models like this:
```bash
% dbt test --selector staging --exclude stg_client
```
However, tests that apply to `stg_client` (a staging model) are still executed. 

### Expected Behavior

I expected the tests that relate to the model `stg_client` to be skipped, since I specified `--exclude stg_client`.

### Steps To Reproduce

1. Create a YAML selector that includes several models with tests.
2. Try to test with the selector, but use `--exclude` to exclude a single model from testing.

### Relevant log output

_No response_

### Environment

```markdown
- OS: Mac OS 11.6.5
- Python: 3.9.10
- dbt: 1.0.3
```

bigquery

### Additional Context

_No response_

acurtis-evi:
resolves https://github.com/dbt-labs/dbt-core/issues/5009

### Description

There are some really cool features in DBT such as the ability to choose models and their descendents, graph ordered builds, and much more.  However, I found that some of the syntax seemed to be lacking (particularly around the selectors yml syntax).

The definition part of the selector yaml supports some model selection operators, but requires that the end user use a different approach to describe what can be done in short-hand on the command line.  For example, on the command line, I can select the children of a model using 

```bash
dbt ls --select some_model+
```

In the yaml, 
```yaml
- name: some_selector
  description:
    - some_model+
```
does not work and instead

```yaml
- name: some_selector
  description:
    - method: fqn
       value: some_model
       children: true    
```
must be used.

The yaml is very useful in defining routine selections.  This inconsistency with the command line forces the end user to learn two approaches.

Furthermore, there is support from the command line to union and intersect using the --select operator.  However, if one wishes to exclude, the separate --exclude operator is needed.  The syntax is very straight forward for the command line approach, but again the yaml approach is different.  I see value in the yaml approach.  I just believe that the two should be compatible.  In this PR, I propose that the --select can include NOT (all caps - but easy to change) as a way of embedding excluded models in a single string.

```bash
dbt ls --select 'selector:some_selector+1 NOT selector:some_selector'
```

In addition, dbt doesn't allow selectors to be expanded using the children/parent/ancestor operators.  This PR addresses that and makes selectors just like any other method.

Syntax in the yaml is also supported
```yaml
- definition: selector:some_other_selector+1 NOT some_model+2 some_other_model
    name: some_selector
```

Most of the code removes the selectors dictionary from being passed around and instead opts to pass the project to the graph/cli get_selected and get_graph_queue methods.

### Checklist


core/dbt/graph/selector.py
```diff
@@ -62,8 +62,11 @@ def select_included(
         """Select the explicitly included nodes, using the given spec. Return
         the selected set of unique IDs.
         """
-        method = self.get_method(spec.method, spec.method_arguments)
-        return set(method.search(included_nodes, spec.value))
+        if spec.method == 'selector':
+            return self.get_selected(self.project.get_selector(spec.value))
+        else:
+            method = self.get_method(spec.method, spec.method_arguments)
+            return set(method.search(included_nodes, spec.value))
```
acurtis-evi, line 70: The self.project.get_selector is the only place that checks if selectors actually exist

core/dbt/config/selectors.py
```diff
@@ -141,33 +141,28 @@ def validate_selector_default(selector_file: SelectorFile) -> None:
...
         if key == "exclude":
             definition = {key: value}
         elif len(definition) == 1:
             definition = {"method": key, "value": value}
-        elif key == "method" and value == "selector":
-            sel_def = definition.get("value")
-            if sel_def not in selector_dict:
-                raise DbtSelectorsError(f"Existing selector definition for {sel_def} not found.")
-            return selector_dict[definition["value"]]["definition"]
         return definition
```
acurtis-evi, line 157: Only checked in graph/selector.py (or config/project.py)

core/dbt/config/selectors.py
```diff
@@ -198,8 +193,7 @@ def parse_from_selectors_list(cls, selectors):
         for selector in selectors:
             sel_name = selector["name"]
             selector_dict[sel_name] = selector
-            definition = cls.parse_from_definition(
-                selector["definition"], selector_dict=deepcopy(selector_dict)
+            selector_dict[sel_name]["definition"] = cls.parse_from_definition(
+                selector["definition"]
             )
```
acurtis-evi, line 198: The selector_dict doesn't initially resolve the selectors.  This is left until the end to allow for it to be more dynamic

core/dbt/graph/cli.py
```diff
@@ -33,28 +33,64 @@ def parse_union(
     # turn ['a b', 'c'] -> ['a', 'b', 'c']
     raw_specs = itertools.chain.from_iterable(r.split(" ") for r in components)
     union_components: List[SelectionSpec] = []
+    exclude_components: List[SelectionSpec] = []
+    in_exclude = False
+
     flags = get_flags()
```
acurtis-evi, line 39: Allow for --select 'some_expression --exclude some_other_expression', having it be one string allows for the yaml syntax to support union/intersection/exclude in a single string

core/dbt/graph/cli.py
```diff
@@ -194,15 +230,10 @@ def parse_dict_definition(definition: Dict[str, Any], result={}) -> SelectionSpe
             "method": key,
             "value": value,
         }
-    elif definition.get("method") == "selector":
-        sel_def = definition.get("value")
-        if sel_def not in result:
-            raise DbtValidationError(f"Existing selector definition for {sel_def} not found.")
-        return result[definition["value"]]["definition"]
     elif "method" in definition and "value" in definition:
```
acurtis-evi, line 233: moved to graph/selector.py / config/project.py

core/dbt/graph/selector_methods.py
```diff
@@ -45,6 +45,7 @@ class MethodName(StrEnum):
     Metric = "metric"
     Result = "result"
     SourceStatus = "source_status"
+    Selector = "selector"
```
acurtis-evi, line 49: The "selector" is added without a corresponding class

core/dbt/task/list.py
```diff
@@ -58,7 +58,7 @@ def __init__(self, args, config, manifest):
     def _iterate_selected_nodes(self):
         selector = self.get_node_selector()
         spec = self.get_selection_spec()
-        nodes = sorted(selector.get_selected(spec))
+        nodes = sorted(selector.get_selected(spec, self.config))
```
acurtis-evi, line 61: get_selected needs the project to call get_selector

core/dbt/task/runnable.py
```diff
@@ -125,7 +125,7 @@ def defer_to_manifest(self, adapter, selected_uids: AbstractSet[str]):
     def get_graph_queue(self) -> GraphQueue:
         selector = self.get_node_selector()
         spec = self.get_selection_spec()
-        return selector.get_graph_queue(spec)
+        return selector.get_graph_queue(spec, self.config)
```
acurtis-evi, line 129: get_graph_queue needs the project to call get_selector

acurtis-evi:
The --exclude operator was meant to work like a NOT operator like ...
```
dbt ls --select 'selector:some_selector+1 NOT selector:some_selector,... something_else'
```
where everything to the right of NOT would be used to exclude from the left.

The reason to choose --exclude was that it was very unlikely that it would conflict with another model and break anything.  Also, the cli could just append --exclude to the --select.

With the selector being added, it effectively allows for parenthetical support and very advanced expressions.

acurtis-evi:
@gshank and @stu-k - I believe this is ready from a functional perspective.  I've never actually contributed anything to DBT, so not sure what else should be done

jtcohen6:
@acurtis-evi Very cool! :)

I definitely want to take a look at this, and play around with some of the proposed additions from a UX perspective, before moving into more-detailed code review with the dev team. (@dbeatty10 too, if you're able & interested!)

dbeatty10:
> The --exclude operator was meant to work like a NOT operator like ...
> 
> dbt ls --select 'selector:some_selector+1 NOT selector:some_selector,... something_else'
> where everything to the right of NOT would be used to exclude from the left.

Just to make sure I'm following, this is a [set difference](https://en.wikipedia.org/wiki/Complement_(set_theory)#Relative_complement), right?

A quick brainstorm of ideas relating to the syntax:
- NOT
- not
- exclude
- minus
- except
- `-` (minus sign)
- `∖` (Set Minus Unicode [U+2216](https://www.compart.com/en/unicode/U+2216))
- `\` (backslash)

acurtis-evi:
I think EXCEPT might be a better name than NOT

acurtis-evi:
Changed NOT => EXCEPT

Support using --selector + --select / --exclude at the same time.  A new selector named arg_selector is introduced which is the combination of --select/--exclude and if --selector is specified, that remains the primary action.


```
dbt ls --selector arg_selector --select some_model
```

acurtis-evi:
There is also support for the arg_selector in the selectors.yml.  This PR moves the selectors.yml validation to runtime, allowing for selectors to be out of order and to refer to the arg_selector.  It is possible to have cycles with this approach, but I tend to think that the benefit of the selection flexibility significantly outweighs the downside of it.

mroy-seedbox:
It would be nice to be able to run selectors minus one or two models.

Right now, the only way to do so is to define another selector, which is a cumbersome process in some environment (but would be very simple on the command line).

mroy-seedbox:
Alternative (and perhaps simpler) implementation proposed here: #10992.

PR #7102, opened 2023-03-02
[feat] Explainable state modified node comparisons

### LINKED ISSUE #7109: [Feature] Explainable state modified

### Describe the feature

**Problem:**
I want to be able to explain a state:modified selection with fine grained diffs. I can code up a lot of ways to solve this but let me riff on what I think would be ideal.

**Solution:**
There are methods returning primitive types that should ideally be wrapper types that duck-type the effect of the primitive. Specifically, the methods I am calling out are planted all over the sort-of core dataclasses, and they return bools when, in an ideal world, they should return a specific class which implements `__bool__` but is, by effect of not being a binary primitive, significantly more extensible. We can do this whilst maintaining the contract with dependents of the method. 

All of the `same_*` methods from `same_body` to `same_database_representation`
These should pass around a `Comparator` which has a baked in method called output diff which leverages `difflib` to generate a log-able diff of the node. This means `Comparator`'s will carry references to old and new, so that it is available as needed. And lastly, the diff is opt-in based on where we decide to plant the method calls. I can see the `-vv` verbosity level being the impetus to output the diff.

So long as the `same_*` methods are restricted to `bool`s, there is no alternative.

### Who will this benefit?

Anyone troubleshooting why a model is running when they don't think or know if it was modified by providing a highly detailed `unified_diff`.

Yes

### Anything else?

Slack Thread: https://getdbt.slack.com/archives/C50NEBJGG/p1675216311733109

z3z1ma:
resolves #7109

### Description

Spike to try updating node comparisons to leverage an object with a `__bool__` method that additionally carries requisite data to log detailed diffs during state based node selection.

### Checklist

### DIFF
--- a/core/dbt/contracts/comparators.py
+++ b/core/dbt/contracts/comparators.py
@@ -1,0 +1,106 @@
+import typing as t
+from difflib import unified_diff
+
+T = t.TypeVar("T")
+
+
+class SimpleComparator(t.Generic[T]):
+    """A base class for comparing two objects of type T."""
+
+    def __init__(
+        self,
+        left: T,
+        right: T,
+        operator: t.Callable[[T, T], bool] = lambda x, y: x == y,
+        left_name: str = "current",
+        right_name: str = "previous",
+    ) -> None:
+        """Create a new SimpleComparator."""
+        self.operator = operator
+        self.left = left
+        self.right = right
+        self.left_name = left_name
+        self.right_name = right_name
+        self.chained_comparisons = [self]
+
+    def compare_operands(self) -> bool:
+        """Return a boolean comparison of the two objects."""
+        return self.operator(self.left, self.right)
+
+    def diff_operands(self) -> str:
+        """Return a str diff of the two objects.
+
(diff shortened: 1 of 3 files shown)


gshank:
Because of a number of changes to nodes.py and because we're in the middle of removing the old metric code, this PR is in a difficult to resolve state. Also the same_contents method has switched to doing "|" between the different checks, which is different behavior and might be breaking, so I'm thinking we don't want to do that at this time. The comparison improvements do look valuable, however. My suggestion would be to wait a bit until the metric switchover is a bit farther along, and then update this (removing the switch to "|" and try to get this included in 1.6.

z3z1ma:
Makes sense to me. We can close this out and I can follow up after the release with a new PR with fresh eyes. 
cc: @gshank

jtcohen6:
Thanks @z3z1ma @gshank! For what it's worth - this is still a really cool idea, and I'd love to keep making progress toward it (for v1.7 if not for v1.6)

PR #7125, opened 2023-03-05
Make MAXIMUM_SEED_SIZE_MIB configurable

### LINKED ISSUE #7117: [Feature] Make MAXIMUM_SEED_SIZE configurable

### Describe the feature

There is a need to allow the 1MB seed limit to be relaxed

Slack thread for context
https://getdbt.slack.com/archives/C50NEBJGG/p1677856760823939

_No response_

### Who will this benefit?

_No response_

_No response_

### Anything else?

_No response_

### LINKED ISSUE #7124: [Feature] Compute seed file hashes incrementally

### Describe the feature

This relates to a conversion on [Slack](https://getdbt.slack.com/archives/C50NEBJGG/p1677856760823939) which resulted in issue https://github.com/dbt-labs/dbt-core/issues/7117 and corresponding PR.

If we allow dbt to hash larger seed files I think these should be computed incrementally to avoid allocation of too much memory. This will be most important in CI where memory is more limited.

My idea is to add another classmethod to the `FileHash` dataclass which has the same signature as `from_contents`. This method can then be used specifically for seed file hashes and from_contents can be continued to be used alongside it.

```python
@classmethod
  def from_path(cls, path: str, name="sha256") -> "FileHash":
      """Create a file hash from the file at given path.
      """
      pass
```

The alternative is to go with what we have which could spike memory usage for those who override the default limit and have a lot of large seed files. Yes this could be an anti pattern but there's no reason to limit the user like that here.

### Who will this benefit?

I would argue that the default limit should be increased from 1 MB but as we have the environment variable override possibility in above PR I would be fine with keeping the default at 1 MB for now.

Therefore, the most benefit are more advanced projects which will override the 1 MB limit with the environment variable.

Yes, I will create the PR

### Anything else?

_No response_

noppaz:
resolves https://github.com/dbt-labs/dbt-core/issues/7117
resolves https://github.com/dbt-labs/dbt-core/issues/7124

### Description

Increasing the maximum size of seed files where the content is hashed for state comparison will enable a greater use of deferred runs with updated seed file contents. By making this a configuration with an environment variable users are able to override the default 1 MiB limit.

Furthermore, to support reading larger files in memory constrained environments, a new method is added to read and compute file contents incrementally. As this is performed on bytes for better performance we continue using the previous UTF-8 content method for small files to not mess up current states where seed files are not stored with UTF-8.

### Checklist


acurtis-evi:
I might recommend removing this from dbt/core/contracts/file.py

```
def seed_too_large(self) -> bool:
        """Return whether the file this represents is over the seed size limit"""
        return os.stat(self.full_path).st_size > MAXIMUM_SEED_SIZE
```

and replacing it with 

```
    def file_size(self): -> int:
        return os.stat(self.full_path).st_size
```

And then you could take [my change](https://github.com/dbt-labs/dbt-core/pull/7125) directly into your PR and add a DEFAULT_MAXIMUM_SEED_SIZE (or just add DEFAULT_MAXIMUM_SEED_SIZE as 1 * 1024 * 1024)

Finally, you could change the logic to do the old thing if the file size <= DEFAULT_MAXIMUM_SEED_SIZE and the new thing if it is larger.

acurtis-evi:
This looks good to me!

noppaz:
Hey @dbt-labs/oss-maintainers, really looking forward to your feedback here.

core/dbt/constants.py
```diff
@@ -1,9 +1,18 @@
...
-MAXIMUM_SEED_SIZE = 1 * 1024 * 1024
-MAXIMUM_SEED_SIZE_NAME = "1MB"
+
+def get_max_seed_size():
+    mx = os.getenv("DBT_MAXIMUM_SEED_SIZE", "1")
+    return int(mx)
+
+
+DEFAULT_MAXIMUM_SEED_SIZE = 1 * 1024 * 1024
+MAXIMUM_SEED_SIZE = get_max_seed_size() * DEFAULT_MAXIMUM_SEED_SIZE
```
jtcohen6, line 14: I don't think we need both of these constants defined. `MAXIMUM_SEED_SIZE` should just be `get_max_seed_size() * 1024 * 1024`. The `get_max_seed_size()` function will return the default value of `1` if the user does not set/override.

jtcohen6 (commented):
Thanks for the PR @noppaz! A few comments after a first read through the code.

core/dbt/constants.py
noppaz: Thank you for checking @jtcohen6!

One reason for keeping `DEFAULT_MAXIMUM_SEED_SIZE` would be to be able to keep the logic of `load_seed_source_file()` in `read_files.py`. Since we do not convert the file from bytes in the new incremental logic we opted to continue computing the checksum with the "old" method for files smaller than 1 MiB so that we'd ensure backward compatibility with current manifests. This is since the checksum of a file encoded as anything other than UTF-8 will be different depending on if it gets UTF-8 encoded or is read as raw bytes.

The reason for not converting from bytes to UTF-8 string is performance and somewhat cleaner code but if we want to remove `DEFAULT_MAXIMUM_SEED_SIZE` maybe we should remove the elif and only compute the seed file hash incrementally instead. I see two options then:
1. Read the file as string and ensure UTF-8 encoding before hashing
2. Continue reading the raw bytes and accept that some projects with seed files that are not utf-8 might have a diff in their manifest deferral when upgrading to this version of dbt-core.

What's your thoughts on this?

core/dbt/constants.py
```diff
@@ -1,9 +1,18 @@
...
+
 SECRET_ENV_PREFIX = "DBT_ENV_SECRET_"
 DEFAULT_ENV_PLACEHOLDER = "DBT_DEFAULT_PLACEHOLDER"
 METADATA_ENV_PREFIX = "DBT_ENV_CUSTOM_ENV_"
 
-MAXIMUM_SEED_SIZE = 1 * 1024 * 1024
-MAXIMUM_SEED_SIZE_NAME = "1MB"
+
+def get_max_seed_size():
+    mx = os.getenv("DBT_MAXIMUM_SEED_SIZE", "1")
```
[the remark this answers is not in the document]
noppaz: I have addressed this in https://github.com/dbt-labs/dbt-core/pull/7125/commits/fa836bb2dfa41dc39360449aba7db92245a2420c. It required some additional moving of the usage of the global flag as it is not a constant anymore and the constants.py file is referenced much earlier in the import cycle.

core/dbt/constants.py
noppaz: I've implemented this suggestion with removal of `DEFAULT_MAXIMUM_SEED_SIZE ` and went with alternative 1 above. The changes can be seen in https://github.com/dbt-labs/dbt-core/pull/7125/commits/53021d942863376f13a90f4dadacb4acfb82868f.

noppaz:
Hey @jtcohen6, I was hoping we'd manage to find some more of your time after 1.5 release (which is looking extremely good btw). Would appreciate some more looks at this. I can't manage to tag @dbt-labs/oss-maintainers mentioned [here](https://docs.getdbt.com/community/resources/oss-expectations#pull-requests), therefore pinging you directly.

dbeatty10:
@noppaz sorry for our silence here!

@jtcohen6 has limited availability currently, so I'm going to give a read through the discussion so far and try out this branch.

dbeatty10:
Will close and re-open to try to trigger the CI checks.

noppaz:
@dbeatty10 thanks for the tests, I've made adjustments and I'm able to get them passing locally now so I think another round of tests and eyes on the code would be the next step.

JavierLopezT:
@noppaz @dbeatty10 Currently my CI/CD doesn't work because we have one seed bigger than 1MiB, so we would need this.

Is there any intention of finishing it up? Thank you very much!

noppaz:
> @noppaz @dbeatty10 Currently our CI/CD doesn't work because we have one seed bigger than 1MiB, so we would need this.
> 
> Is there any intention of finishing it up? Thank you very much!

Hey @JavierLopezT, with no intention of sounding bitter, I gave up on this work due to low interest from dbt Labs. Would be happy to finish it up if they'd be up for it.

jtcohen6:
@noppaz We've heard some renewed interest in making seed size configurable (for the purposes of `state:modified` comparison), so I ported the changes from this PR into a new one:
- https://github.com/dbt-labs/dbt-core/pull/11177

Issue #7067, opened 2023-02-27
[CT-2199] Generalize constraint compatibility warnings


MichelleArk:
Refinement discussion: https://github.com/dbt-labs/dbt-core/issues/6750

Currently, constraints that are set but unenforceable or unsupported raise warnings in adapter-specific jinja macros. For example: [`snowflake__get_columns_spec_ddl` in dbt-snowflake](https://github.com/dbt-labs/dbt-snowflake/pull/341/files#diff-c96dce207e585a460348ff3f7fbd4c39d5822de2b13e2443648c0d8948066178R13-R15).

To generalize this warning mechanism, we'll need a new adapter method that can tell dbt whether a constraint is:
1. supported & enforced
2. supported but not enforced
3. not supported

Here's what that might look like for `dbt-snowflake`, in pseudo code:
```python
@staticmethod
def constraint_support:
    return {
        "not_null": ConstraintSupport.ENFORCED,
        "primary_key": ConstraintSupport.NOT_ENFORCED,
        "foreign_key": ConstraintSupport.NOT_ENFORCED,
        "unique": ConstraintSupport.NOT_ENFORCED,
        "check": ConstraintSupport.NOT_SUPPORTED,
    }
```
This mapping could be a static method, or a property, or maybe just a constant class attribute. As long as subclasses of `BaseAdapters` are able to overwrite it. This mapping should also not be available in the jinja context as it's intended for use within the python adapter implementation and not by end users in user space (jinja).

This method would be called in order to determine whether  to raise warnings for unenforced/unsupported constraints (unless the user has disabled via `warn_unsupported` or `warn_unenforced` (https://github.com/dbt-labs/dbt-core/issues/7066)). Ideally, warnings would be raised at runtime - during materialization, for legible user-facing logs. The actual method that raises warnings could also be implemented in python, on the adapter, calling `warn_or_error` and firing an event under the hood. It should be used to replace the existing adapter-specific [warning mechanisms](https://github.com/dbt-labs/dbt-snowflake/pull/341/files#diff-c96dce207e585a460348ff3f7fbd4c39d5822de2b13e2443648c0d8948066178R14). 

The `constraint_support` method should be implemented for each adapter, but the warning implementation should be generalized and only implemented in dbt-core.

---

Is blocked by:
- #7066

Blocks:
- #7065

jtcohen6:
### Adapter-specific support

[...]

Issue #6754, opened 2023-01-26
Support 'constraints' as a model-level config


MichelleArk:
In addition to `constraints` as a column-level property, support constraints as a model-level config for row-level boolean expressions that require multiple columns, or combined primary keys (unenforced & informational-only).

Will depend on what we land on in terms of a spec for the column-level `constraints` property in https://github.com/dbt-labs/dbt-core/issues/6750#issue-1558234542

MichelleArk:
From: https://github.com/dbt-labs/dbt-core/issues/6750#issuecomment-1436064688

```yaml
models:
  constraints: Optional[List[ModelLevelConstraint]]
  columns:
    - name: str
      data_type: Optional[str]
      not_null: Optional[bool]
      constraints: Optional[List[ColumnLevelConstraint]]
```

MichelleArk:
From https://github.com/dbt-labs/dbt-core/issues/7066: 

```python
class ConstraintOption(StrEnum):
    check = "check"
    not_null = "not_null"
    unique = "unique"
    primary_key = "primary_key"
    foreign_key = "foreign_key"
    custom = "custom"  # I have no idea -- but let's leave it open for the wacky & wild data platforms out there

ColumnLevelConstraint:
    type: ConstraintOption
    name: Optional[str]
    warn_unenforced: bool = True  # raise a warning if this constraint is not enforced by this data platform (but will still be included in templated DDL)
    warn_unsupported: bool = True  # raise a warning if this constraint is not supported by this data platform (and will be excluded from templated DDL)
    expression: Optional[str] # free for all. required if constraint is 'check' or 'foreign_key' -- can we validate via post-init ?
    # additional properties allowed: yes! there are going to be lots of constraint-specific and platform-specific shenanigans

# the same, with an added 'columns' attribute (which isn't needed when the constraint is defined for one column)
ModelLevelConstraint(ColumnLevelConstraint):
    columns: List[str]
```

```yml
models:
  constraints: Optional[List[ModelLevelConstraint]]
```
If `constraints` is set a the model level, it should be a model-level attribute, _not_ a "config." This means it isn't settable in `{{ config() }}` or in `dbt_project.yml`, and that's okay.

jtcohen6:
### Example

(Borrowing heavily from Postgres docs: https://www.postgresql.org/docs/current/ddl-constraints.html)

```yml
models:
  - name: my_model
    columns:
      - name: first_id
        data_type: int
      # ... rest of columns ...
[...]

PR #7130, opened 2023-03-06
ct-2198: Unify constraints and check_constraints fields

### LINKED ISSUE #7066: [CT-2198] Unify constraints and constraints_check configs
Refinement discussion: https://github.com/dbt-labs/dbt-core/issues/6750

In https://github.com/dbt-labs/dbt-core/pull/6271, constraints: List[str] and constraints_check: str are optional column configs that configure column-level constraints and row-level checks respectively.

## Current data structure
```yaml
columns:
  - name: price
    data_type: numeric
    constraints:
      - not null
    constraints_check: price > 0
```

## Proposed data structure

`Constraint` is a new object type. Pseudo code:
```python
class ConstraintOption(StrEnum):
    check = "check"
    not_null = "not_null"
    unique = "unique"
    primary_key = "primary_key"
    foreign_key = "foreign_key"
    custom = "custom"  # I have no idea -- but let's leave it open for the wacky & wild data platforms out there

ColumnLevelConstraint:
    type: ConstraintOption
    name: Optional[str]
    warn_unenforced: bool = True  # raise a warning if this constraint is not enforced by this data platform (but will still be included in templated DDL)
    warn_unsupported: bool = True  # raise a warning if this constraint is not supported by this data platform (and will be excluded from templated DDL)
    expression: Optional[str] # free for all. required if constraint is 'check' or 'foreign_key' -- can we validate via post-init ?
    # additional properties allowed: yes! there are going to be lots of constraint-specific and platform-specific shenanigans

# the same, with an added 'columns' attribute (which isn't needed when the constraint is defined for one column)
ModelLevelConstraint(ColumnLevelConstraint):
    columns: List[str]
```

```yml
models:
  constraints: # type: Optional[List[ModelLevelConstraint]]
  columns:
    - name: # type: str
      data_type: # type: Optional[str]
      constraints: # type: Optional[List[ColumnLevelConstraint]]
```

**Notes:**
- `constraints` is an optional attribute at both the model level and the column level
- If `constraints` is set a the model level, it should be a model-level attribute, _not_ a "config." This means it isn't settable in `{{ config() }}` or in `dbt_project.yml`, and that's okay.
- `warn_unenforced` & `warn_unsupported` could alternatively be `Optional[bool]`, and flipped/renamed to `ignore_warning_unenforced` & `ignore_warning_unsupported`, so that the data is only included when it's a non-default value
- This will require changes in adapter macros to obtain constraints information from the new data structure to build constraint ddl.

peterallenwebb:
resolves #7066

### Checklist


core/dbt/contracts/graph/unparsed.py
```diff
@@ -88,7 +88,7 @@ class Docs(dbtClassMixin, Replaceable):
 
 
 @dataclass
-class HasDocs(AdditionalPropertiesMixin, ExtensibleDbtClassMixin, Replaceable):
+class HasColumnProps(AdditionalPropertiesMixin, ExtensibleDbtClassMixin, Replaceable):
```
gshank, line 91: The "access" property should not be in a class that is attached to columns. Probably should be moved into UnparsedNodeUpdate.

gshank (changes requested):
This looks pretty good, except that the field I put in the wrong place should be moved :)

core/dbt/contracts/graph/unparsed.py
gshank: If you look at ColumnInfo in core/dbt/contracts/graph/nodes.py, the ColumnInfo should be a match for what was HasDocs. Looks like the "docs" field should also be moved out of that class, and into UnparsedNodeUpdate.

MichelleArk (approved):
The issue https://github.com/dbt-labs/dbt-core/issues/7066 also describes adding a `ModelLevelConstraint` dataclass that is available as a model property. However, we won't implement any usage of `ModelLevelConstraint` until https://github.com/dbt-labs/dbt-core/issues/6754, so I'd be fine shipping this work as-is and introducing `ModelLevelConstraint` as part of https://github.com/dbt-labs/dbt-core/issues/6754 instead.

core/dbt/contracts/graph/unparsed.py
peterallenwebb: @gshank When I moved `access` to UnparsedNodeUpdate, it caused a the TestAnalyses::test_postgres_analyses test to break as follows:
```
def parse_patch(self, block: TargetBlock[NodeTarget], refs: ParserRef) -> None:
        # We're not passing the ParsedNodePatch around anymore, so we
        # could possibly skip creating one. Leaving here for now for
        # code consistency.
        patch = ParsedNodePatch(
            name=block.target.name,
            original_file_path=block.target.original_file_path,
            yaml_key=block.target.yaml_key,
            package_name=block.target.package_name,
            description=block.target.description,
            columns=refs.column_info,
            meta=block.target.meta,
            docs=block.target.docs,
            config=block.target.config,
>           access=block.target.access,
        )
E       AttributeError: 'UnparsedAnalysisUpdate' object has no attribute 'access'
```

I was able to fix that by adding a separate `access` property to `UnparsedAnalysisUpdate` but it makes me nervous. What do you think?

Moving docs also breaks a few tests.

gshank: That should be okay. Yet another sign that we should go in and rearrange those classes...

MichelleArk: Not sure it makes sense functionally to have `access` on an analysis node given they can't be `ref`'d. Is there an alternative workaround? If not, let's at least leave a comment that `access` isn't actually used to restrict `ref`s for analyses.

gshank: It's not on the analysis node. It's on the class that's used for pulling in fields from schema files in unparsed,py, which is classes that are used for ferrying things from yaml to nodes (plus a few secondary classes that are used both there and in nodes.py). There is no "access" on the AnalysisNode, and setting it is prevented by code  in the the "patch" method on ParsedNode.

We really need to do some cleanup in unparsed.py.

core/dbt/contracts/graph/unparsed.py
```diff
@@ -145,12 +145,12 @@ class HasConfig:
 
 
 @dataclass
-class UnparsedAnalysisUpdate(HasConfig, HasColumnDocs, HasDocs, HasYamlMetadata):
+class UnparsedAnalysisUpdate(HasConfig, HasColumnDocs, HasColumnProps, HasYamlMetadata):
```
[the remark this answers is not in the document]
gshank: Yes, agree. But probably not in this ticket :)

core/dbt/contracts/graph/unparsed.py
```diff
@@ -113,18 +113,18 @@ def __post_init__(self):
...
 
 @dataclass
 class HasColumnDocs(dbtClassMixin, Replaceable):
-    columns: Sequence[HasDocs] = field(default_factory=list)
+    columns: Sequence[HasColumnProps] = field(default_factory=list)
 
 
 @dataclass
-class HasColumnTests(HasColumnDocs):
+class HasColumnTests(dbtClassMixin, Replaceable):
```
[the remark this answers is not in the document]
gshank: Sounds good.

core/dbt/parser/schemas.py
```diff
@@ -132,11 +125,10 @@ def add(
...
         self.column_info[column.name] = ColumnInfo(
             name=column.name,
-            description=description,
-            data_type=data_type,
-            constraints=constraints,
-            constraints_check=constraints_check,
-            meta=meta,
+            description=column.description,
+            data_type=column.data_type,
+            constraints=[ColumnLevelConstraint.from_dict(c) for c in column.constraints],
```
gshank, line 130: This will blow up if the "type" in the constraint isn't valid. You should make a test case with a column level constraint with an invalid type string.

gshank (changes requested):
I don't think this handles invalid column level constraint types.

core/dbt/contracts/graph/unparsed.py
MichelleArk: Ahh I see, thank you for the explanation!

peterallenwebb: @gshank I verified that the line you highlighted throws a `mashumaro.exceptions.InvalidFieldValue` exception. Should we catch it here and raise a `ParsingError` exception instead, or do you think we should be validating this somewhere else?

core/dbt/parser/schemas.py
peterallenwebb: @gshank I verified that the line you highlighted throws a mashumaro.exceptions.InvalidFieldValue exception. Should we catch it here and raise a ParsingError exception instead, or do you think we should be validating this somewhere else?

gshank: I validated separately in a similar circumstance. In the "patch" method on ParsedNode in nodes.py, I did: 
```
               if AccessType.is_valid(patch.access):
                    self.access = AccessType(patch.access)
                else:
                    raise InvalidAccessTypeError(
                        unique_id=self.unique_id,
                        field_value=patch.access,
                    )

```
Added an "is_valid" method to the AccessType enum.

gshank: I would validate right there before you want to construct those ColumnLevelConstraints.

gshank: Not sure whether we want to proliferate these kind of errors or not. I suppose we could do something more generic...

core/dbt/include/global_project/macros/materializations/models/table/columns_spec_ddl.sql
```diff
@@ -13,8 +13,8 @@
     {% for i in user_provided_columns %}
       {% set col = user_provided_columns[i] %}
       {% set constraints = col['constraints'] %}
-      {% set constraints_check = col['constraints_check'] %}
-      {{ col['name'] }} {{ col['data_type'] }} {% for x in constraints %} {{ x or "" }} {% endfor %} {% if constraints_check -%} check {{ constraints_check or "" }} {%- endif %} {{ "," if not loop.last }}
+      {{ col['name'] }} {{ col['data_type'] }} {% for x in constraints %} {{ "check" if x.type == "check" else "not null" if x.type == "not_null" else "unique" if x.type == "unique" else "primary key" if x.type == "primary_key" else "foreign key" if x.type == "foreign key" else ""
```
mikealfare, line 16: Can we add a property to the constraint object solely for jinja templates so that this if block is calculated in python? I feel like it would be easier to maintain this across adapters in python than in jinja.

core/dbt/contracts/graph/unparsed.py
```diff
@@ -88,23 +88,21 @@ class Docs(dbtClassMixin, Replaceable):
 
 
 @dataclass
-class HasDocs(AdditionalPropertiesMixin, ExtensibleDbtClassMixin, Replaceable):
+class HasColumnProps(AdditionalPropertiesMixin, ExtensibleDbtClassMixin, Replaceable):
```
mikealfare, line 91: For my own understanding, is this supposed to reflect a Column, but with additional properties? In other words, the new column object would inherit from both `ColumnBase` (assumed name, not sure of the real name) and this `HasColumnProperties` mixin; and the original `ColumnBase` object is very basic with very simple methods/attributes, but this mixin adds additional properties, and methods that use those properties, as a bundle so to speak.

core/dbt/parser/schemas.py
```diff
@@ -114,29 +116,28 @@ class ParserRef:
...
         quote: Optional[bool]
         if isinstance(column, UnparsedColumn):
             quote = column.quote
         else:
             quote = None
+
+        if any(
+            c
+            for c in column.constraints
+            if not c["type"] or not ConstraintType.is_valid(c["type"])
```
mikealfare, line 131: Along the same lines as @gshank's argument below, we should be validating the `constraint.type` when we set it, not when we use it. Once we get to this point, we should be able to assume that `type` exists for each `constraint`, and is always a valid type.

core/dbt/include/global_project/macros/materializations/models/table/columns_spec_ddl.sql
peterallenwebb: I can see the argument for it, because this code (or similar) is repeated in several adapters. At the same time, I think the current approach allows for the greatest flexibility and support for syntax variations. What do you think @MichelleArk?

core/dbt/parser/schemas.py
peterallenwebb: This is the spot @gshank suggested validating, if I understood her correctly. I think validating earlier would require doing what I attempted in an earlier revision: Adding more structure to the `constraints` column on HasColumnProps. I encountered an issue with that approach, which is that it is serialized by hologram for partial parsing, which can't deal with enumerations.

core/dbt/contracts/graph/unparsed.py
gshank: No, it doesn't reflect a column with additional properties. It's really just a step on the path to making these mixins consistent. This particular set of fields is used in both nodes and columns.

core/dbt/parser/schemas.py
gshank: In general it's not very clean to handle enums in classes that represent yaml dictionaries (which is what most of the classes in unparsed.py are) since we need classes that will take the dictionary of strings that we get from yaml. So the validation has to be done in between the yaml-representing classes and instantiating nodes/configs. Somewhere.

In theory we could validate the dictionaries, and we do that in some places, but that's really early in parsing and unfortunately we don't distinguish between a bunch of overlapping nodes there, so in order to do that we'd have to separate out some parsing methods which are now common to a bunch of different nodes.

core/dbt/contracts/graph/unparsed.py
peterallenwebb: The mixins used for "unparsed" objects have become confusing, and this rename is just meant to be one step in a journey of making them make more sense. Gerda and I have discussed eventually eliminating the mixing approach in favor of concrete objects with minimal inheritance.

This change doesn't make HasDocs a perfectly coherent class in one stroke. It just gives it a name closer to what it is today, in order to make the code more readable. I'm still not sure what this thing is. It's true that it "has docs", but that wasn't a very useful way of thinking about it, since it has accumulated many more properties, and is really more like a column. Confusingly, things that are clearly not columns inherit from it, and they inherit properties that don't make much sense for them. We intend to continue addressing these issues over time.

core/dbt/parser/schemas.py
gshank: At some point we should look at all of the places where validation is done and regularize it all and do it in a methodical way.

core/dbt/contracts/graph/unparsed.py
peterallenwebb: Sorry, I took too long to write my comment. What Gerda said. :D

core/dbt/include/global_project/macros/materializations/models/table/columns_spec_ddl.sql
MichelleArk: The idea is that the python property on the `Constraint` object would represent the rendered out constraint, correct? Unfortunately the way constraints are rendered out is pretty adapter-specific (e.g. `primary key not enforced` for BigQuery, `primary key` for postgres). So we'd need this to be a method on the adapter - provided a `Constraint` object, render out its DDL. Most DDL generation is currently done in jinja already, but perhaps we could make this easier to maintain by splitting it out into a separate adapter.dispatch macro - perhaps `default__render_constraint(constraint)`?

plugins/postgres/dbt/include/postgres/macros/utils/columns_spec_ddl.sql
```diff
@@ -5,8 +5,8 @@
     {% for i in user_provided_columns %}
```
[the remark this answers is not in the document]
peterallenwebb: Confirmed they are exactly the same. I'll remove this macro.

core/dbt/include/global_project/macros/materializations/models/table/columns_spec_ddl.sql
MichelleArk: Thinking about this more, I do see another argument for this DDL rendering logic living in python as opposed to user-space (jinja). Although it's adapter-specific logic, there's really only one correct way to render constraint DDL given a `Constraint` object and it's very unlikely that individual dbt projects will want the ability to override that behaviour (maybe just to get a new constraint working faster than we can implement / accept contributions to the adapter python implementation). 

It feels similar to how the BigQuery adapter renders `PartitionConfig` via a `render` method: https://github.com/dbt-labs/dbt-bigquery/blob/main/dbt/adapters/bigquery/impl.py#L89-L97. The difference here being that it doesn't make sense as much sense to define the `Constraint` class in the `adapter` module given core requires it for parsing.

core/dbt/contracts/graph/unparsed.py
mikealfare: > Gerda and I have discussed eventually eliminating the mixin approach in favor of concrete objects with minimal inheritance.

Sign me up

core/dbt/parser/schemas.py
mikealfare: Thanks @gshank for all of that context. That makes sense now.

mikealfare (approved):

core/dbt/include/global_project/macros/materializations/models/table/columns_spec_ddl.sql
MichelleArk: So concretely what I'm suggesting here is:
* A new `@available.parse` method on [BaseAdapter](https://github.com/dbt-labs/dbt-core/blob/main/core/dbt/adapters/base/impl.py#L151), perhaps with the interface: `render_column_constraint(raw_column_constraint: Dict[str, str]) -> str`
* This new method would rely on an import of the `ColumnLevelConstraint` class from core to parse the `raw_constraint` to a `ColumnLevelConstraint` object, validate it, and render out the DDL. Unfortunately I don't see how we could implement a `render` method on the core-defined `ColumnLevelConstraint` class itself given the adapter-specific implementations.
* The adapter implementation would go from: 
```
{{ col['name'] }} {{ col['data_type'] }} {% for x in constraints %} {{ "check" if x.type == "check" else "not null" if x.type == "not_null" else "unique" if x.type == "unique" else "primary key" if x.type == "primary_key" else "foreign key" if x.type == "foreign key" else "" }} {{ x.expression or "" }} {% endfor %} {{ "," if not loop.last }}
```
to: 
```
{{ col['name'] }} {{ col['data_type'] }} {% for constraint in constraints %} {{ adapter.render_column_constraint(constraint) }}
```

Notes for dependent issues: 
* The warning logic that will end up using [`constraint_support`](https://github.com/dbt-labs/dbt-core/issues/7067) for raising warnings when users configure unsupported/unenforced constraints could live in this new method!
* We'd end up needing to implement a similar pattern for rendering [model-level constraints](https://github.com/dbt-labs/dbt-core/issues/6754) via a `render_model_constraint` method unless we can figure out a nice way to unify the two.

jtcohen6: ^ This all tracks for me!

Rationale:
- This differs significantly by adapter, but _for each data platform_, there's really only one correct way to render a constraint string. There's no "business logic" here that can really differ by user of that platform.
- This ends up being a massive `if-else` engine, which is much more a strength of Python than of Jinja (which is nice for old-fashioned string templating, but that's about it). That's only truer with consideration for model-level constraints. We could also more easily write a unit test for this function, if it's written in Python :)
- The [`custom` constraint type](https://github.com/dbt-labs/dbt-core/pull/7130/files#diff-42cfc99b2fcd89ab75957b8999bee2b1af822f7d39953ca99f1f285c170557c1R151) should just render out its entire `expression` as a string. This is the escape hatch, in case a user can't get some wacky constraint string working with one of the more-strongly-typed options. dbt should assume that a `custom` constraint is supported and not enforced.

peterallenwebb: @jtcohen6 @MichelleArk @mikealfare The latest changeset should implement the suggested changes. Let me know if you have any concerns!

core/dbt/adapters/base/impl.py
```diff
@@ -1262,6 +1265,39 @@ def get_incremental_strategy_macro(self, model_context, strategy: str):
         # This returns a callable macro
         return model_context[macro_name]
 
+    @classmethod
+    def _parse_constraint(cls, raw_constraint: Dict[str, Any]) -> ColumnLevelConstraint:
```
MichelleArk, line 1269: naming nit: `_parse_column_constraint`

core/dbt/adapters/base/impl.py
```diff
@@ -1262,6 +1265,39 @@ def get_incremental_strategy_macro(self, model_context, strategy: str):
...
+        try:
+            ColumnLevelConstraint.validate(raw_constraint)
+            return ColumnLevelConstraint.from_dict(raw_constraint)
+        except Exception:
+            raise DbtValidationError(f"Could not parse constraint: {raw_constraint}")
+
+    @available
+    @classmethod
+    def render_raw_column_constraint(cls, raw_constraint: Dict[str, Any]) -> str:
```
MichelleArk, line 1278: What do you think about adding a unit test for this method, perhaps in https://github.com/dbt-labs/dbt-core/blob/main/test/unit/test_postgres_adapter.py?

MichelleArk (approved):
LGTM! A couple comments, nothing blocking.

core/dbt/adapters/base/impl.py
peterallenwebb: It sounds good, but I am desperate to be done with the current round of work. I'll look to add more testing (even if it is a bit out of scope like this) in the model-level portion of the work.

gshank (approved):

PR #7134, opened 2023-03-07
Make mypy version consistently 0.981

gshank:
### Description

We specified mypy version 0.981 in dev-requirements.txt but specified mypy 0.942 in github main workflow and pre-commit config.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -55,10 +55,9 @@ jobs:
           python -m pip --version
           python -m pip install pre-commit
           pre-commit --version
-          python -m pip install mypy==0.981
-          mypy --version
           python -m pip install -r requirements.txt
           python -m pip install -r dev-requirements.txt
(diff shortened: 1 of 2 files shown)


.github/workflows/main.yml
```diff
@@ -55,7 +55,7 @@ jobs:
           python -m pip --version
           python -m pip install pre-commit
           pre-commit --version
-          python -m pip install mypy==0.942
+          python -m pip install mypy==0.981
```
emmyoop, line 58: Why do we need to do this here?

emmyoop (changes requested):

.github/workflows/main.yml
gshank: Have no idea why it was there. Removed the pip install and moved the line checking mypy version after dev-requirements.

emmyoop (approved):

Issue #7064, opened 2023-02-27
[CT-2196] Make constraints ddl agnostic to column order


MichelleArk:
The [foundational work on constraints](https://github.com/dbt-labs/dbt-core/pull/6271)  introduced pre-flight contract validation on column names, expecting that the user-provided column names are identical both in name and order the column names obtained from the model SQL. This is currently by design, so that the database associates any configured constraints to the correct columns from the query result.

However, it will be cumbersome for users to ensure that the order of configured columns in a model yaml file always matches with the order the database provides. 

We can mitigate this while still ensuring any constraints are applied to the correct columns by wrapping user-provided SQL in a subquery which SELECTs columns in the order they are provided in the schema yaml (which is the same order used to generate the constraints ddl). 

 Implementation Notes: 
* If `contract: True`, wrap the [user-provided SQL](https://github.com/dbt-labs/dbt-core/blob/main/core/dbt/include/global_project/macros/materializations/models/table/create_table_as.sql#L33) in a `SELECT ..` statement generated from `model['columns']` (the user-provided columns). 

Could look something like: 
```
{% macro <adapter>__create_table_as(temporary, relation, sql) -%}
    ... 
    {% if config.get('contract', False) %}
      {{ get_assert_columns_equivalent(sql) }}
      {{ get_columns_spec_ddl() }}
      {%- set sql = get_select_subquery(sql) %}
    {% endif %}
  as (
    {{ sql }}
  );
{%- endmacro %}

{% macro get_select_subquery(sql) -%}
  {{ return(adapter.dispatch('get_select_subquery', 'dbt')(sql)) }}
{% endmacro %}

{% macro default__get_select_subquery(sql) -%}
    select 
    {% for column in model['columns'] %}
      {{ column }}{{ ", " if not loop.last }}
    {%- endfor %}
    from (
        {{ sql }}
    ) as model_subq
{%- endmacro %}
```
* This will require changes to adapter-specific macros for create table statements.
* We won't need to do this for `view` materializations as they don't support constraints.
 
Relevant discussions: https://github.com/dbt-labs/dbt-core/pull/6271#discussion_r1106122988, https://github.com/dbt-labs/dbt-core/pull/6271/files#r1069332715

jtcohen6:
What this should look like, roughly:

```sql
create table dbt_jcohen.my_model (
    -- returned by get_columns_spec_ddl
    column_a integer not null,
    column_b text,
    ...
)
as (
-- this part is new
select
[...]

Issue #7147, opened 2023-03-09
[CT-2287] Clarify contract error message


emmyoop:
Right now the error message for errors related to having contracts enabled does not mention that the error is being hit _because_ contracts are enabled for the model.  Add some verbiage to the effect that the model is contracted so all the following have more context.

https://github.com/dbt-labs/dbt-core/blob/c02ddf8c0ea84f25298bad35c99ed04a030d89e4/core/dbt/include/global_project/macros/materializations/models/table/columns_spec_ddl.sql#L44

```
16:00:13  Compilation Error in model my_model (models/contracted_models/my_model.sql)
16:00:13    Please ensure the name, data_type, order, and number of columns in your `yml` file match the columns in your SQL file.
16:00:13    Schema File Columns:
16:00:13
16:00:13    SQL File Columns: id INT, color TEXT, date_day DATE
16:00:13
16:00:13    > in macro assert_columns_equivalent (macros/materializations/models/table/columns_spec_ddl.sql)
16:00:13    > called by macro default__get_assert_columns_equivalent (macros/materializations/models/table/columns_spec_ddl.sql)
16:00:13    > called by macro get_assert_columns_equivalent (macros/materializations/models/table/columns_spec_ddl.sql)
16:00:13    > called by macro default__create_view_as (macros/materializations/models/view/create_view_as.sql)
16:00:13    > called by macro create_view_as (macros/materializations/models/view/create_view_as.sql)
16:00:13    > called by macro default__get_create_view_as_sql (macros/materializations/models/view/create_view_as.sql)
16:00:13    > called by macro get_create_view_as_sql (macros/materializations/models/view/create_view_as.sql)
16:00:13    > called by macro statement (macros/etc/statement.sql)
16:00:13    > called by macro materialization_view_default (macros/materializations/models/view/view.sql)
16:00:13    > called by model my_model (models/contracted_models/my_model.sql)
16:00:13
16:00:13  Done. PASS=0 WARN=0 ERROR=2 SKIP=1 TOTAL=3
```

jtcohen6:
Resolved by https://github.com/dbt-labs/dbt-core/pull/7161

Issue #7065, opened 2023-02-27
[CT-2197] [Feature] Detect breaking changes to constraints in state:modified check


MichelleArk:
If a model with `contract: True` has its constraints modified in a way that represents a breaking change, dbt should raise an error as part of the `state:modified` check. 

What constitutes a breaking change?
- Removing a (column or model-level) constraint that is _enforced_ by the data warehouse
- Changing the materialization from `table` to `view` or `ephemeral` on a model that has at least one enforced constraint.

---

_Blocked by:_
- #7066
- #7067
- #6754

MichelleArk:
> array

jtcohen6:
Good point! We really shouldn't allow `ephemeral` models to be `contract: true` or `access: public` in the first place

emmyoop:
We are unclear on exactly what we would be checking has changed and how often.  There's concern this would cause a slowdown in detecting changes if we have to check every single column for every model with a contract.

sadahry:
We think this feature will be useful for us, so I'd like to ask you about this spec.

We've tried to using `dbt-core==1.5.0-b5` ,
and found disallowing adding new columns to yml in this version.

e.g., 

before
```yml
version: 2

models:
  - name: sometable
    config:
      contract:
        enforced: true
    columns:
      - name: order_id
        data_type: number

      - name: order_name
        data_type: varchar
```

after -> then failed with `dbt.exceptions.ModelContractError`
```yml
version: 2

models:
  - name: sometable
    config:
      contract:
        enforced: true
    columns:
      - name: order_id
        data_type: number

      - name: order_name
        data_type: varchar
```

That is what you excepted?

I read it and found this description.
> Additive changes are not considered breaking:
>* adding a new column to a contracted model
>https://docs.getdbt.com/reference/resource-configs/contract#detecting-breaking-changes

My understand is that it means allow adding columns in yaml, but current behavior does not.
Allowing adding columns is useful for us because our model's are updated in a fast cycle.

I'd like to know if this specification is as expected
or if there is a mistake in me or dbt-core.

jtcohen6:
@sadahry Thanks for trying out the beta, and for the bug report! I agree that's _not_ the intended behavior. It should be possible to add a new column without raising a `ModelContractError` (breaking change). I'll open a bug report now.

sadahry:
Thank you for your quick fix!

PR #7161, opened 2023-03-13
CT 2196, CT2121 constraints column order

### LINKED ISSUE #6975: [CT-2121] Make model contracts agnostic to ordering
The current model contract comparison (`assert_columns_equivalent` [macro](https://github.com/dbt-labs/dbt-core/pull/6986/files#diff-28fef0f8080904ca9d7c04e61050beded83aaecc39d332b8f0486eea686d44f1L30)) validates that the user-provided column names and data types are identical: in name, type, _and order_ the column names obtained from the model SQL.

However, it will be cumbersome for users to ensure that the order of configured columns in a model yaml file always matches with the order the database provides. 

For example, the SQL model configured with the schema below fails its contract check:

```sql
-- model.sql
SELECT 1 as a, 2 as b
```
```yaml
-- schema.yml
models:
- model: 
  contract: True
  columns:
    - name: b
       data_type: INT
    - name: a
       data_type: INT
```

The [equality comparison is done directly on formatted columns](https://github.com/dbt-labs/dbt-core/pull/6986/files#diff-28fef0f8080904ca9d7c04e61050beded83aaecc39d332b8f0486eea686d44f1R37) from the SQL + schema yaml, and could instead be done on the intermediate collection of unordered columns. 

Relevant discussion: https://github.com/dbt-labs/dbt-core/pull/6271#discussion_r1106122988

Note: should not be merged before https://github.com/dbt-labs/dbt-core/issues/7064, as this ordering check is currently done in `assert_columns_equivalent` and prevents constraint ddl generation from being generated in an invalid order.

gshank:
resolves #6975, #7064

### Description

Modify create_table_as ddl to use the order specified in the model config "columns". Modify "assert_columns_equivalent" to not check column order.

### Checklist


core/dbt/include/global_project/macros/materializations/models/table/columns_spec_ddl.sql
```diff
@@ -37,13 +37,34 @@
...
+  {#-- Build compiler error msg #}
   {%- set sql_file_provided_columns_formatted = format_columns(sql_file_provided_columns)  -%}
   {%- set schema_file_provided_columns_formatted = format_columns(schema_file_provided_columns)  -%}
+  {%- set compiler_error_msg = 'Please ensure the name, data_type, and number of columns in your `yml` file match the columns in your SQL file.\nSchema File Columns: ' ~ (schema_file_provided_columns_formatted|trim) ~ '\n\nSQL File Columns: ' ~ (sql_file_provided_columns_formatted|trim) ~ ' ' -%}
 
-  {%- if sql_file_provided_columns_formatted != schema_file_provided_columns_formatted -%}
-    {%- do exceptions.raise_compiler_error('Please ensure the name, data_type, order, and number of columns in your `yml` file match the columns in your SQL file.\nSchema File Columns: ' ~ (schema_file_provided_columns_formatted|trim) ~ '\n\nSQL File Columns: ' ~ (sql_file_provided_columns_formatted|trim) ~ ' ' ) %}
+
+  {%- if sql_file_provided_columns|length != schema_file_provided_columns|length -%}
+    {%- do exceptions.raise_compiler_error(compiler_error_msg) -%}
```
emmyoop, line 47: Should we be raising a unique (new) error here?  I know it was already raising a CompilerError but it was added for multi project.  In the view materialization changes, I created a new error for it.

core/dbt/include/global_project/macros/materializations/models/table/create_table_as.sql
```diff
@@ -25,11 +25,26 @@
...
+
+{% macro get_select_subquery(sql) -%}
+  {{ return(adapter.dispatch('get_select_subquery', 'dbt')(sql)) }}
+{% endmacro %}
+
+{% macro default__get_select_subquery(sql) -%}
+    select
+    {% for column in model['columns'] %}
+      {{ column }}{{ ", " if not loop.last }}
+    {%- endfor %}
```
emmyoop, line 46: to match other patterns 
```suggestion
    {%- for column in model['columns'] -%}
      {{ column }}{{ ", " if not loop.last }}
    {%- endfor -%}
```

core/dbt/include/global_project/macros/materializations/models/table/columns_spec_ddl.sql
```diff
@@ -37,13 +37,34 @@
...
   {%- endif -%}
 
+  {%- for sql_col in sql_file_provided_columns -%}
+    {%- set yaml_col = [] -%}
```
emmyoop, line 51: This won't catch when there's a column defined in the schema file but missing from sql.

emmyoop (changes requested):

core/dbt/include/global_project/macros/materializations/models/table/columns_spec_ddl.sql
gshank: I think that either the length will be wrong or there would be the same number of columns with two different names, and both cases will cause an error message.

core/dbt/include/global_project/macros/materializations/models/table/columns_spec_ddl.sql
gshank: What was the new error?

emmyoop: These are the two new pieces I had to add.

https://github.com/dbt-labs/dbt-core/blob/f24452a3ab789d8e678f9833184286dfaf0b16ed/core/dbt/context/exceptions_jinja.py#L96-L97

https://github.com/dbt-labs/dbt-core/blob/f24452a3ab789d8e678f9833184286dfaf0b16ed/core/dbt/exceptions.py#L2095-L2110

You would need to add a new one.  I started a slack thread about larger picture how we wanted to handle these types of errors generated in jinja.

core/dbt/include/global_project/macros/materializations/models/table/create_table_as.sql
gshank: Will add some hyphens.

gshank: Hyphens on these lines breaks the generated sql.

core/dbt/include/global_project/macros/materializations/models/table/columns_spec_ddl.sql
gshank: I added "rasie_contract_error", ContractError.

core/dbt/exceptions.py
```diff
@@ -2136,6 +2136,22 @@ def get_message(self) -> str:
...
+        self.sql_columns = sql_columns
+        super().__init__(msg=self.get_message())
+
+    def get_message(self) -> str:
+        msg = (
+            "Please ensure the name, data_type, and number of columns in your `yml` file "
+            "match the columns in your SQL file.\n"
+            f"Schema File Columns: {self.yaml_columns}\n"
+            f"SQL File Columns: {self.sql_columns}"
+        )
```
emmyoop, line 2151: Were you going to update this as related to https://github.com/dbt-labs/dbt-core/issues/7147?

emmyoop (approved):
Looks good to me!  Just a matter of if you wanted to update the wording of the error to include something about the model having contracts enabled or not.

core/dbt/include/global_project/macros/materializations/models/table/columns_spec_ddl.sql
```diff
@@ -37,13 +37,32 @@
...
+      {%- if schema_col.name == sql_col.name -%}
+        {%- do yaml_col.append(schema_col) -%}
+        {%- break -%}
+      {%- endif -%}
+    {%- endfor -%}
+    {%- if not yaml_col -%}
+      {#-- Column with name not found in yaml --#}
+      {%- do exceptions.raise_contract_error(yaml_columns, sql_columns) -%}
+    {%- endif -%}
+    {%- if sql_col.dtype != yaml_col[0].dtype -%}
```
MichelleArk, line 60: Previously column types were compared based on their formatted representation, which could be data platform specific as implemented by the `<adapter>__format_column` macro. 

For example, [BigQuery's `format_column` implementation](https://github.com/dbt-labs/dbt-bigquery/pull/551/files#diff-ba5d875a4144a1b739519c0133ae7847a58715ca5374e841c06df089894778faR19) compares data_type values by `column.data_type` as opposed to `column.dtype` in order to make comparisons on nested data types. 

Strictly comparing SQL and yml data_type values by `dtype` would allow the following contract to be accepted in BigQuery:

```sql
SELECT
STRUCT("test" AS name, [1.0,2.0] AS laps) as some_struct
```

```yaml
models:
  - name: test_schema
    config:
      contract: true
    columns:
      - name: some_struct
        data_type: STRUCT<name FLOAT64, laps STRING> #wrong! but accepted because dtype == STRUCT for both SQL and schema.yml
```

One workaround would be to do this comparison using `format_column` instead, i.e: `adapter.dispatch('format_column', 'dbt')(sql_col) != adapter.dispatch('format_column', 'dbt')(yaml_col[0])`. This would also ensure the comparison and error messaging are using consistent logic. An alternative would be to implement a `default__compare_data_types` macro to enable adapter-specific implementations.

gshank: I looked at using format_column but the default implementation of that just uses the dtype. Is that different in the other adapters?

gshank: I don't actually see any implementations of format_column in the adapters.

core/dbt/include/global_project/macros/materializations/models/table/columns_spec_ddl.sql
```diff
@@ -37,13 +37,32 @@
   {#--Obtain the column schema provided by the schema file by generating an 'empty schema' query from the model's columns. #}
```
MichelleArk, line 37: small nit (can't comment directly on the line) - we can remove 'order' from 'If any differences in name, data_type or order of columns' from the comment on line 32 above!

core/dbt/include/global_project/macros/materializations/models/table/columns_spec_ddl.sql
gshank: Oh, I was on the wrong bigquery branch. Bigquery is the only adapter that implements it.

gshank: Could we include the formatted column in the Column structures returned by "get_column_schema_from_query"?

gshank: Or concatenate the other parts of the column structure for comparison?

gshank:
I don't see how a "ModelContract" would be different than model.columns.

MichelleArk:
> I don't see how a "ModelContract" would be different than model.columns. 

Currently, these two are really the same thing but I could see a definition of `ModelContract` that includes model columns (`name`s + `data_type`s) as well as `constraints`, and potentially materializations. Basically any model attribute that could be modified in a way that would constitute a breaking change (https://github.com/dbt-labs/dbt-core/issues/7065, https://github.com/dbt-labs/dbt-core/issues/6869).

In Mike's proposal, `assert_schema_meets_contract` would just be comparing the model columns component of `ModelContract` to implement the pre-flight model contract checks necessary for the scope of this work. However, a `ModelContract` structure to represent the more general model contract could be useful as we implement `state:modified` checks to detect breaking changes to model contracts (https://github.com/dbt-labs/dbt-core/issues/7065, https://github.com/dbt-labs/dbt-core/issues/6869). I pushed up some prototyping I did on the state:modified check for column names + data_types and could imagine a `ModelContract` being instantiated [here](https://github.com/dbt-labs/dbt-core/pull/7168/files#diff-42cfc99b2fcd89ab75957b8999bee2b1af822f7d39953ca99f1f285c170557c1R481-R482). 

> Also this code has the same issue that the earlier jinja code did in that it's only checking dtype, not the formatted column. Would we also have to implement a "format_column" method in the python adapters?

I agree with this - we'd need to add an overridable `formatted_dtype` property/method on the `Column` interface to provide a complete and user-friendly representation of the `data_type`. Note: we should _not_ just use the existing `data_type` definitions for this comparison because it does some dbt-specific formatting, which would make the mismatched contract error message less reflective of what the user actually has to specify as a valid value for `data_type` in their schema.yml (more context here: https://github.com/dbt-labs/dbt-core/pull/6986#discussion_r1120849287)

gshank:
This all might be a good idea, but is totally out of scope for this ticket as it was defined. If you want us to close this one and wait until this other implementation is defined, let me know.

MichelleArk:
> out of scope for this ticket as it was defined

100% agree! I think if we decide to go down the route of introducing a more generalized `ModelContract` structure as part of the`state:modified` issues (https://github.com/dbt-labs/dbt-core/issues/7065, https://github.com/dbt-labs/dbt-core/issues/6869), we can revisit making use of it to implement more of the contract validation in python. I can make an issue to keep track of the idea.

tests/adapter/dbt/tests/adapter/constraints/test_constraints.py
```diff
@@ -190,17 +178,25 @@ def test__constraints_correct_column_data_types(self, project, data_types):
...
-    date_day date
+    date_day text
 ) ;
 insert into {0} (
     id ,
     color ,
     date_day
-) (
+)
+(
```
mikealfare, line 188: @gshank Is this the nested sql that's inexplicably failing? I recall you mentioning one, but can't remember which. I wanted to help look into it.

gshank: I don't think anything is inexplicably failing right now. We were having a problem on dbt-spark because we hadn't specified file_format: delta, and the constraints were failing because they were referencing the original column name (not the spoofed one). But it's working now.

core/dbt/include/global_project/macros/materializations/models/table/create_table_as.sql
```diff
@@ -25,11 +25,26 @@
```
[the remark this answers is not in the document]
gshank: We need the get_select_subquery in order to put the columns in the right order.  We removed validating the order because we added the subquery. As far as splitting out the macro into two, probably @jtcohen6 and @MichelleArk should weigh in on that.

PR #7169, opened 2023-03-14
Allow `--select` and `--exclude` multiple times

### LINKED ISSUE #6800: [CT-1981] Issue when passing `--select` and `--model` to a command together
Did some manual testing by locally overwriting a command that accepts the new `select` (--select, --model) alias command with `print(ctx.obj["flags"].SELECT)` and found some issues when passing `--select` and `--model` to a command together. 

The options are not treated as mutually exclusive via MultiOption and the resulting `flags.SELECT` value is clobbered by the value from the last alias option:
```
❯ python3 core/dbt/cli/main.py docs generate --select model_a model_b  --model model_c
('model_c',)

❯ python3 core/dbt/cli/main.py docs generate --model model_c --select model_a model_b
('model_a', 'model_b')
```

This was unexpected behaviour to me, but for better or for worse this is also how main's parsing currently works (printing args.select and exiting on main to test): 
```
❯ dbt docs generate --select model_a model_b --model model_c
['model_c']
❯ dbt docs generate --select model_a model_b --model model_c
['model_c']
❯ dbt docs generate --model model_c --select model_a model_b 
['model_a', 'model_b']
```

We should probably open an issue to document this behaviour and investigate a fix that gets us --select/--model option mutual exclusivity if possible. But given that it's the current behaviour on main I don't see it as pressing to fix as part of this PR.

_Originally posted by @MichelleArk in https://github.com/dbt-labs/dbt-core/issues/6787#issuecomment-1410299008_

### LINKED ISSUE #7158: [Feature] warn/error when specifying `--select` or `--exclude` multiple times in single invocation

### Describe the feature

dbt invocations containing multiple selections/exclusion specifications do not behave intuitively and offer no guidance as to the resultant behavior. For example `dbt run --select a --select b` does not necessarily run models `a` and `b`. Neither does `dbt run --exclude a --exclude b` exclude `a` and `b` (empirically, it seems like the first specification of each option is used and the rest are ignored).

Assuming that the intended command-line interface is "at most one `--select` option per invocation" and "at most one `--exclusion` option per invocation", it would be terrific if the CLI raised a warning or error when an invocation violates that expectation.

Or, if multiple `--select`/`--exclude` options are supported, then the current behavior is non-intuitive and poorly-documented.

Update the [node selection documentation](https://docs.getdbt.com/reference/node-selection/syntax), but it is much more user-friendly to have dbt prevent you from doing unsupported things rather than expecting everyone to have thoroughly read all the documentation (sadly, haha).

### Who will this benefit?

Folks learning dbt

Not at this time, sorry

### Anything else?

Sourced from this Slack thread: https://getdbt.slack.com/archives/CBSQTAPLG/p1678480247560059, with @mattdelac

dbeatty10:
resolves #7158
resolves #6800

### Problem
Currently, if a user does the following, only the last selector (`users_rollup`) will be take effect and earlier ones (`users`) will not:
```shell
dbt run --select users --select users_rollup
```

Selection syntax is confusing enough as-is, and the fact that it accepts multiples but only uses the last one makes it even more confusing.

### Options

#7158 describes two options for how to move forward:
1. Combine multiple instances of `--select` as if it were a [union](https://docs.getdbt.com/reference/node-selection/set-operators#unions). Similar for `--exclude`.
1. Raise an error (with a helpful message) if more than 1 `--select` is supplied. Same thing for `--exclude`.

This PR fleshes out **option 1**.

Note: most of it would also be applicable if we choose to implement **option 2** instead.

Implicitly, there's also an **option 0** that keeps the status quo. If we choose option 0, I'd still advocate that we cherry-pick the two test cases in https://github.com/dbt-labs/dbt-core/pull/7169/commits/2824f4ac82c08c07ef46f32e05defcc44f924de4 so we can actively affirm that this is the expected and desired behavior.

### Breaking change

This is described as a breaking change in the changelog because the behavior changes for certain edge cases (see above).

### 🎩 
Our custom `MultiOption` class itself doesn't currently have any unit tests. Its main form of testing is the functional tests that depend upon it.

We expect that any parameter with MultiOption should also have (`type=tuple` or `type=ChoiceTuple`) and `multiple=True`.

Here's the exceptions that are raised when either of those expectations are violated:

Not `type=tuple` and not `type=ChoiceTuple`:
```
  File "/Users/dbeatty/projects/dbt-core/core/dbt/cli/options.py", line 28, in __init__
    assert issubclass(option_type, tuple), msg
AssertionError: MultiOption named `output_keys` must be tuple or ChoiceTuple (rather than <class 'list'>)
```

Not `multiple=True`:
```
  File "/Users/dbeatty/projects/dbt-core/core/dbt/cli/options.py", line 22, in __init__
    assert multiple, msg
AssertionError: MultiOption named `output_keys` must have multiple=True (rather than None)
```

### Checklist

    - https://github.com/dbt-labs/docs.getdbt.com/issues/3003

### DIFF
--- a/core/dbt/cli/options.py
+++ b/core/dbt/cli/options.py
@@ -1,4 +1,6 @@
 import click
+import typing as t
+from click import Context
 
 
 # Implementation from: https://stackoverflow.com/a/48394004
@@ -42,3 +44,17 @@ class MultiOption(click.Option):
                 our_parser.process = parser_process
                 break
         return retval
+
+    def type_cast_value(self, ctx: Context, value: t.Any) -> t.Any:
+        def flatten(data):
+            if isinstance(data, tuple):
+                for x in data:
+                    yield from flatten(x)
+            else:
+                yield data
+
+        # there will be nested tuples to flatten when multiple=True
+        value = super(MultiOption, self).type_cast_value(ctx, value)
+        if value:
+            value = tuple(flatten(value))
+        return value
--- a/core/dbt/cli/options.py
+++ b/core/dbt/cli/options.py
@@ -1,6 +1,8 @@
 import click
+import inspect
 import typing as t
 from click import Context
+from dbt.cli.option_types import ChoiceTuple
 
 
 # Implementation from: https://stackoverflow.com/a/48394004
@@ -13,6 +15,19 @@ class MultiOption(click.Option):
         super(MultiOption, self).__init__(*args, **kwargs)
         self._previous_parser_process = None
         self._eat_all_parser = None
+
+        # validate that multiple=True
+        multiple = kwargs.pop("multiple", None)
+        msg = f"MultiOption named `{self.name}` must have multiple=True (rather than {multiple})"
+        assert multiple, msg
+
+        # validate that type=tuple or type=ChoiceTuple
+        option_type = kwargs.pop("type", None)
+        msg = f"MultiOption named `{self.name}` must be tuple or ChoiceTuple (rather than {option_type})"
+        if inspect.isclass(option_type):
+            assert issubclass(option_type, tuple), msg
+        else:
+            assert isinstance(option_type, ChoiceTuple), msg
 
     def add_to_parser(self, parser, ctx):
         def parser_process(value, state):
--- a/core/dbt/cli/params.py
+++ b/core/dbt/cli/params.py
@@ -83,6 +83,7 @@ exclude = click.option(
     envvar=None,
     type=tuple,
     cls=MultiOption,
+    multiple=True,
     help="Specify the nodes to exclude.",
 )
 
@@ -318,6 +319,7 @@ select_attrs = {
     "envvar": None,
     "help": "Specify the nodes to include.",
     "cls": MultiOption,
+    "multiple": True,
     "type": tuple,
 }
 
--- a/core/dbt/cli/params.py
+++ b/core/dbt/cli/params.py
@@ -183,7 +183,7 @@ output_keys = click.option(
         "Space-delimited listing of node properties to include as custom keys for JSON output "
         "(e.g. `--output json --output-keys name resource_type description`)"
     ),
-    type=list,
+    type=tuple,
     cls=MultiOption,
     default=[],
 )
--- a/core/dbt/cli/params.py
+++ b/core/dbt/cli/params.py
@@ -185,6 +185,7 @@ output_keys = click.option(
     ),
     type=tuple,
     cls=MultiOption,
+    multiple=True,
     default=[],
 )
 
@@ -310,6 +311,7 @@ resource_type = click.option(
         case_sensitive=False,
     ),
     cls=MultiOption,
+    multiple=True,
     default=(),
 )
 
--- a/core/dbt/cli/params.py
+++ b/core/dbt/cli/params.py
@@ -79,7 +79,11 @@ enable_legacy_logger = click.option(
 )
 
 exclude = click.option(
-    "--exclude", envvar=None, type=tuple, cls=MultiOption, help="Specify the nodes to exclude."
+    "--exclude",
+    envvar=None,
+    type=tuple,
+    cls=MultiOption,
+    help="Specify the nodes to exclude.",
 )
 
 fail_fast = click.option(
--- a/.changes/unreleased/Breaking Changes-20230314-161505.yaml
+++ b/.changes/unreleased/Breaking Changes-20230314-161505.yaml
@@ -1,0 +1,6 @@
+kind: Breaking Changes
+body: Allow `--select` and `--exclude` multiple times
+time: 2023-03-14T16:15:05.81741-06:00
+custom:
+  Author: dbeatty10
+  Issue: "7158"
--- a/tests/functional/graph_selection/test_graph_selection.py
+++ b/tests/functional/graph_selection/test_graph_selection.py
@@ -163,6 +163,12 @@ class TestGraphSelection(SelectionFixtures):
         results = run_dbt(["run", "--select", "@emails_alt", "users_rollup"], expect_pass=False)
         check_result_nodes_by_name(results, ["users_rollup", "users", "emails_alt"])
 
+    def test_concat_multiple(self, project):
+        results = run_dbt(
+            ["run", "--select", "@emails_alt", "--select", "users_rollup"], expect_pass=False
+        )
+        check_result_nodes_by_name(results, ["users_rollup"])
+
     def test_concat_exclude(self, project):
         results = run_dbt(
             [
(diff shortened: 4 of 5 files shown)


dbeatty10:
@jtcohen6 we need to make a functional decision between one of these three options:
- [ ] option 0 - status quo: accept multiple `--select` but only use the last one
- [ ] option 1 - combine multiple instances of `--select` as if they were provided together, space delimited
- [ ] option 2 - raise an error if multiple `--select` are supplied

We have same options for `--exclude` also, and I'm assuming we'll align the choice with the one made for `--select`.

I'm most attracted to option 1 as I think it would the least surprising to the end user.

dbeatty10:
Besides `--select` and `--exclude`, there are two other parameters that use `MultiOption`:
- `--output-keys`
- `--resource-type`

So we'd need to make the make a similar decision amongst the three options for each of them as well. They could either be covered in this PR or deferred to a separate issue(s).

jtcohen6:
@dbeatty10 Thanks for laying out the options so clearly!

I agree that option 0 (status quo) is the most confusing & least desirable.

My initial instinct was for option 2 (raise an error), but I think you've convinced me that option 1 is preferable. It's actually the default expected behavior for many CLIs ([including `click`](https://click.palletsprojects.com/en/8.1.x/options/#multiple-options)) when they accept multiple options.

core/dbt/cli/params.py
```diff
@@ -178,7 +183,7 @@
         "Space-delimited listing of node properties to include as custom keys for JSON output "
         "(e.g. `--output json --output-keys name resource_type description`)"
     ),
-    type=list,
+    type=tuple,
```
dbeatty10, line 186: ### Context
When I left this as `type=list`, a couple tests failed:
- https://github.com/dbt-labs/dbt-core/actions/runs/4420754548/

But when I updated it to `type=tuple`, they passed again:
- https://github.com/dbt-labs/dbt-core/actions/runs/4422283702/

Although it is not enforced, there is a comment [here](https://github.com/dbt-labs/dbt-core/blob/311a57a21e61955904202bdb469c2077202e9073/core/dbt/cli/options.py#L5) that says:
> MultiOption options must be specified with type=tuple or type=ChoiceTuple

### Suggestion
To avoid unforeseen issues in the future, maybe we should enforce that the configured `type` is one of the following whenever `cls=MultiOption`?
- subclass of `tuple`
- instance of `ChoiceTuple`

One option for enforcing this within the [`__init__` method of `MultiOption`](https://github.com/dbt-labs/dbt-core/blob/311a57a21e61955904202bdb469c2077202e9073/core/dbt/cli/options.py#L4-L13):
```
option_type = kwargs.pop("type", None)
if inspect.isclass(option_type):
    assert issubclass(option_type, tuple), "type must be tuple or ChoiceTuple not {}".format(option_type)
else:
    assert isinstance(option_type, ChoiceTuple), "type must be tuple or ChoiceTuple not {}".format(option_type)
```

dbeatty10:
> you've convinced me that option 1 is preferable. It's actually the default expected behavior for many CLIs ([including `click`](https://click.palletsprojects.com/en/8.1.x/options/#multiple-options)) when they accept multiple options.

@jtcohen6 

Shall we apply option 1 to each of the following as well? Or just `--select` and `--exclude` for now?
- `--output-keys`
- `--resource-type`

I tested both `--output-keys` and `--resource-type` out by hand, and they worked well.

I think those four are all the ones using our custom `MultiOption` class right now.

Even when specifying `cls=MultiOption`, `multiple=True` also needs to be specified in order to opt-in to allowing multiple instances of `--{argument}`, so we need to explicitly decide which parameters should have `multiple=True` added.

jtcohen6:
@dbeatty10 I agree with applying this change consistently across the board, to all existing `MultiOption` params.

Going forward, it sounds like any new param with `MultiOption` should also have `type=tuple` and `multiple=True`.

core/dbt/cli/options.py
```diff
@@ -12,6 +16,19 @@ def __init__(self, *args, **kwargs):
...
+        msg = f"MultiOption named `{self.name}` must have multiple=True (rather than {multiple})"
+        assert multiple, msg
+
+        # validate that type=tuple or type=ChoiceTuple
+        option_type = kwargs.pop("type", None)
+        msg = f"MultiOption named `{self.name}` must be tuple or ChoiceTuple (rather than {option_type})"
+        if inspect.isclass(option_type):
+            assert issubclass(option_type, tuple), msg
+        else:
+            assert isinstance(option_type, ChoiceTuple), msg
```
dbeatty10, line 30: Very open to feedback here -- I just stuffed quick-n-dirty validation in the easiest place possible.

core/dbt/cli/options.py
```diff
@@ -42,3 +59,17 @@ def parser_process(value, state):
...
+            if isinstance(data, tuple):
+                for x in data:
+                    yield from flatten(x)
+            else:
+                yield data
+
+        # there will be nested tuples to flatten when multiple=True
+        value = super(MultiOption, self).type_cast_value(ctx, value)
+        if value:
+            value = tuple(flatten(value))
```
dbeatty10, line 74: I don't know why, but the value can be `None`. We only want to flatten it when it has a non-None value.

dbeatty10:
> Maybe it's just me, but should the behavior be set intersection instead of set union?

Good eyes @aranke ! 🤩

Suppose a user expresses `--select` multiple times  like:
```
dbt run --select users --select users_rollup
```

### Key question
- would they expect this to be equivalent to a ` ` or a `,` in [dbt selection syntax](https://docs.getdbt.com/reference/node-selection/set-operators)?

### Options

In this case, we have two options of how to combine them each of the arguments:
1. set union
1. set intersection

### Explanations

1. With a set union
    - chaining uses a logical `OR`
    - [union](https://docs.getdbt.com/reference/node-selection/set-operators#unions) operator in dbt selection syntax
    - `dbt run --select users --select users_rollup` would be equivalent to `dbt run --select users users_rollup`
1. With a set intersection
    - chaining uses a logical `AND`
    - [intersection](https://docs.getdbt.com/reference/node-selection/set-operators#intersections) operator in dbt selection syntax
    - `dbt run --select users --select users_rollup` would be equivalent to `dbt run --select users,users_rollup`

### Pros of set union
- Aligns with the [union](https://docs.getdbt.com/reference/node-selection/set-operators#unions) within the dbt selection syntax
    - `dbt run --select users --select users_rollup` is equivalent to `dbt run --select users users_rollup`
- Users can still achieve intersections with [this](https://docs.getdbt.com/reference/node-selection/set-operators#intersections) syntax
- Can express slightly more complex selection on the command line before needing [YAML selectors](https://docs.getdbt.com/reference/node-selection/yaml-selectors)

### Cons of set union
- Surprising if you are expecting a logical `AND` like [method chaining](https://towardsdatascience.com/using-pandas-method-chaining-to-improve-code-readability-d8517c5626ac)/cascading with Pandas DataFrames or piping in Unix systems.

### Pros of set intersection
- Behaves similar to [method chaining](https://towardsdatascience.com/using-pandas-method-chaining-to-improve-code-readability-d8517c5626ac)/cascading using dataframes like:
  ```
  df
    .filter(f1)
    .filter(f2)
  ```

### Cons of set intersection
- Surprising if you are expecting a logical `AND`

### Proposal
My belief is that a majority of dbt users will expect `dbt list -s this that` and `dbt list -s this -s that` to be equivalent, so I'm in favor of **set union**.

PR #7185, opened 2023-03-17
Add deprecation warnings for `log-path`, `target-path` in dbt_project.yml

### LINKED ISSUE #6882: Move LOG_PATH, TARGET_PATH from dbt_project.yml to user config dict
# Issue
Currently, user would be able to set `log-path` and `target-path` in `dbt_project.yml`.

But since we initialize all of the flags before loading the `project` object, we need to 
manually load the project object from a profile dir, read out the log-path and set it there.

One other issue with this setup is that dbt-server would load the project object before 
invoking the rest of the core functionality, during that process we are actually calling 
get_flags function with will create a global flag. This causes circular dependency between 
project and flags.

# Solution
Move `log_path` and `target-path` out of dbt_project and move them into user_config that's 
currently in profiles.yml. This would be a breaking change for certain user but clean up the 
circular dependency of project and Flags.

jtcohen6:
resolves #6882

### Description

- We have back-compat (for now) for `log-path` and `target-path` being set in `dbt_project.yml`
- Since these are now properly `Flags` (settable via CLI flag or env var), it is inconsistent (and added complexity in our codebase / initialization flow) that they can also be set in `dbt_project.yml`
- Add a deprecation warning that we plan to remove support for these configs in a future version of `dbt-core`. Because so many users have the configs set, with just the default values, only raise the deprecation warning if a custom (non-default) value is set in `dbt_project.yml`

### Checklist

- [x] I have [opened an issue to add/update docs](https://github.com/dbt-labs/docs.getdbt.com/issues/new/choose), or docs changes are not required/relevant for this PR: https://github.com/dbt-labs/docs.getdbt.com/issues/3014

### DIFF
--- a/core/dbt/config/project.py
+++ b/core/dbt/config/project.py
@@ -298,7 +298,7 @@ class PartialProject(RenderComponents):
             raise DbtProjectError("Package dbt_project.yml must have a name!")
         return ProjectPackageMetadata(self.project_name, packages_config.packages)
 
-    def check_config_path(self, project_dict, deprecated_path, exp_path):
+    def check_config_path(self, project_dict, deprecated_path, exp_path=None, default=None):
         if deprecated_path in project_dict:
             if exp_path in project_dict:
                 msg = (
@@ -310,11 +310,20 @@ class PartialProject(RenderComponents):
                 raise DbtProjectError(
                     msg.format(deprecated_path=deprecated_path, exp_path=exp_path)
                 )
-            deprecations.warn(
-                f"project-config-{deprecated_path}",
-                deprecated_path=deprecated_path,
-                exp_path=exp_path,
-            )
+            # this field has been renamed
+            if exp_path:
+                deprecations.warn(
+                    f"project-config-{deprecated_path}",
+                    deprecated_path=deprecated_path,
+                    exp_path=exp_path,
+                )
+            # this field is no longer supported, but many projects may specify it with the default value
+            # if so, let's only raise this deprecation warning if they set a custom value
+            if not default or project_dict[deprecated_path] != default:
+                deprecations.warn(
+                    f"project-config-{deprecated_path}",
+                    deprecated_path=deprecated_path,
+                )
 
     def create_project(self, rendered: RenderComponents) -> "Project":
         unrendered = RenderComponents(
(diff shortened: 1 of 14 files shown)


core/dbt/config/project.py
```diff
@@ -310,11 +310,20 @@ def check_config_path(self, project_dict, deprecated_path, exp_path):
...
-            )
+            # this field has been renamed
+            if exp_path:
+                deprecations.warn(
+                    f"project-config-{deprecated_path}",
+                    deprecated_path=deprecated_path,
+                    exp_path=exp_path,
+                )
+            # this field is no longer supported, but many projects may specify it with the default value
+            # if so, let's only raise this deprecation warning if they set a custom value
```
jtcohen6, line 321: ❗I welcome disagreement ❗

core/dbt/config/project.py
```diff
@@ -298,7 +298,7 @@ def render_package_metadata(self, renderer: PackageRenderer) -> ProjectPackageMe
             raise DbtProjectError("Package dbt_project.yml must have a name!")
         return ProjectPackageMetadata(self.project_name, packages_config.packages)
 
-    def check_config_path(self, project_dict, deprecated_path, exp_path):
+    def check_config_path(self, project_dict, deprecated_path, exp_path=None, default=None):
```
VersusFacit, line 301: I don't know what `exp_path` is.....expected path? Explicit path? Something else? An explicit variable would be nice for readable/reasoning.

core/dbt/config/project.py
VersusFacit: I'm fine with this. Why rouse them to action when there's really no need. It's just a syntactic artifact. At most, you could tell people "hey this doesn't even do anything anymore tbh" but also idk if it really impacts anyone.

core/dbt/config/project.py
VersusFacit: Also what's `default` exactly?

VersusFacit (commented):
Comments mostly around readability. I'm generally on board for this and how you've gone about it.

Also, just an aside, man, I used to use these flags way back when. Crazy that they're going away. tempus fugit.

core/dbt/config/project.py
jtcohen6: I'll rename these!
- `exp_path` &rarr; `renamed_path`
- `default` &rarr; `default_value`

VersusFacit (approved):
Thanks for making the variable names explicit. Last night without knowing exactly what they were, I was just having trouble parsing the code. It's much clearer now!

jtcohen6:
@VersusFacit Thank you for the review! :)

Issue #7113, opened 2023-03-03
[CT-2264] Don't raise exception during event serialization failure, just log warning


jtcohen6:
Instead of raising as an `Exception` that stops the whole run, let's instead fire a different warn-level event to call out the event serialization failure. Ideally, this would include enough information about the event's contents (serialized as a string?) that would be helpful in debugging, but let's make extra sure that doesn't silently/loudly fail either.

In `main` branch:

https://github.com/dbt-labs/dbt-core/blob/72076b3fe5a9a97884bbc4b6951a984eb50556a9/core/dbt/events/functions.py#L205-L209

In `1.3.latest` (should be similar in `1.2.latest` and `1.1.latest`):

https://github.com/dbt-labs/dbt-core/blob/7f2e9f94c0dcb27feb4ac1f6b545140e2fd301b8/core/dbt/events/functions.py#L129-L135

gshank:
Duh. An obvious improvement. Should have thought of this before...

jtcohen6:
Opened this up while chatting with @QMalcolm earlier! We were taking a look at:

- https://github.com/dbt-labs/dbt-core/issues/7108
- https://github.com/dbt-labs/dbt-core/issues/6568

Both of those have actually been fixed in `main` / `1.4.latest`, but it's worth:
- adding tests for the odd behavior to `main` (to ensure we don't regress on either one in the future)
- making this change, to prevent serialization errors from ever being a deal-breaker
- backporting those patches to older versions (v1.1, v1.2, v1.3)

jtcohen6:
Proposed catch-all fixes for:
- `1.0.latest`: https://github.com/dbt-labs/dbt-core/commit/fd83cc7f9ba37bf7d85eba6651356ae38fcfb84d
- `1.1.latest`, `1.2.latest`, `1.3.latest`: https://github.com/dbt-labs/dbt-core/commit/dd9bcddaf10b436d53322ed9c05a097008dea71b (need to catch during dict serialization and also JSON serialization)

I tested those changes with the reproduction case from https://github.com/dbt-labs/dbt-core/issues/7108, and it gracefully handled the bad log message:
```json
[...]

PR #7190, opened 2023-03-18
CT 1998 use google protobuf to enable more flexible dictionaries

### LINKED ISSUE #6832: [CT-1998] Improve handling of dictionaries in protobuf messages
In pull request #6820 we implemented protobuf message versions of nodes and node configs. However the match between Python dictionaries that allow values of any type and protobuf maps isn't straightforward. As a result we have temporarily forced a number of dictionaries into string/string dictionaries.

We should look at other options for this and improve the representation of dictionaries in protobuf messages. In particularly this applies to the "meta" dictionary, which is already included in 1.4 in the node_info structures in the logs.

In addition to "meta", the following dictionaries are affected:
* in NodeConfig - persist_docs, quoting, column_types, grants
* in ColumnInfo - _extra
* in various nodes - unrendered_config
* in TestMetadata - kwargs
* in SourceDefinitino - source_meta

gshank:
resolves #6832

### Description

In order to have more flexible dictionaries in our logging events, we need to have the protobuf Struct data type. This isn't supported by betterproto. In addition, Optional is only supported with a beta version. In order to have Structs, we are switching to using google protobuf instead.

There are a number of differences in implementation between betterproto and google protobuf.  In google protobuf the generated "Python" classes are not very Python-like, and the interfaces are odd. Betterproto had support for a datetime-like timestamp field, which is different than protobuf.  Protobuf can instantiate messages from nested dictionaries. It also catches type mismatches that betterproto didn't. When using betterproto it was necessary to pre-construct nested messages; this can't be done in protobuf. The interfaces for serializing are different. 

### Checklist


core/dbt/contracts/results.py
```diff
@@ -67,7 +68,7 @@ def __exit__(self, exc_type, exc_value, traceback):
         with TimingProcessor(self.timing_info):
             fire_event(
                 TimingInfoCollected(
-                    timing_info=self.timing_info.to_msg(), node_info=get_node_info()
+                    timing_info=self.timing_info.to_msg_dict(), node_info=get_node_info()
```
jtcohen6, line 71: In practice, do we expect this change (nested message &rarr; nested dict) to be breaking in any way? My sense is that it looks the same, if the event is being serialized to JSON, which is our only official current interface for consuming structured logs

core/dbt/events/base_types.py
```diff
@@ -61,24 +57,48 @@ class EventLevel(str, Enum):
...
+            len(args) == 0
+        ), f"[{class_name}] Don't use positional arguments when constructing logging events"
+        if "base_msg" in kwargs:
+            kwargs["base_msg"] = str(kwargs["base_msg"])
+        if "msg" in kwargs:
+            kwargs["msg"] = str(kwargs["msg"])
+        try:
+            self.pb_msg = ParseDict(kwargs, msg_cls())
+        except Exception as exc:
+            raise Exception(f"[{class_name}]: Unable to parse dict {kwargs},\n exc: {exc}")
```
jtcohen6, line 79: As discussed during the live walkthrough, we want to stop throwing an exception (and stopping the run), and instead just log a warning that we failed to serialize the event:
- https://github.com/dbt-labs/dbt-core/issues/7113

@gshank made the point that it's a bit weird to fire an event _within the code_ for firing events. I don't have strong opinions here on the implementation — just that, as an end user, I would want dbt to more gracefully handle serialization errors.

docs/arch/adr-005-betterproto.md
```diff
@@ -28,8 +28,24 @@ You can use the google protobuf package to generate Python "classes", using the
 * Additional benefits listed: [betterproto](https://github.com/danielgtaylor/python-betterproto)
 
 
+## Revisited
```
jtcohen6, line 31: Nice! Thanks for reviewing/rewriting. Let's put a date here (and up top)? I know it's available from `git blame`, but just for quick reference

docs/arch/adr-005-betterproto.md
```diff
@@ -28,8 +28,24 @@ You can use the google protobuf package to generate Python "classes", using the
...
+We are switching away from using betterproto because of the following reasons:
+* betterproto only suppports Optional fields in a beta release
+* betterproto has had only beta releases for a few years
+* betterproto doesn't support Struct, which we really need
 
-# Consequences
+Steps taken to mitigate the drawbacks of Google protobuf from above:
+* We are using a wrapping class around the logging events to enable a constructor that looks more like a Python constructor, as long as only keyword arguments are used.
+* The generated file is skipped in the pre-commit config
+* We can live with the awkward interfaces. It's just code.
```
jtcohen6, line 41: :)

jtcohen6 (commented):
Very cool! Really appreciated the walkthrough & recording :)

I tried this out locally, and I didn't see any obvious changes in JSON-formatted logs. No blocking comments from me.

core/dbt/events/functions.py
```diff
@@ -12,11 +11,14 @@
 import sys
 from typing import Callable, Dict, Optional, TextIO
 import uuid
+from google.protobuf.json_format import MessageToDict
 
 
 LOG_VERSION = 3
 metadata_vars: Optional[Dict[str, str]] = None
 
+nofile_codes = ["Z012", "Z013", "Z014", "Z015"]
```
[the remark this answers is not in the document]
gshank: I added a comment. In theory we could still use something like classes or method flags, but we'd have to get the class from msg.info.name, do getattr(types, class_name) and then check for the existence of another class or execute a class method. So that's still on option if some other flag thing comes along that makes it worth it.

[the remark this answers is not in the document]
jtcohen6: cool! agree that, in this case, it doesn't feel worth it — hard-coding the codes is a right-sized solution

core/dbt/contracts/results.py
```diff
@@ -45,11 +44,13 @@ def begin(self):
...
-            name=self.name, started_at=self.started_at, completed_at=self.completed_at
-        )
-        return timsg
+    def to_msg_dict(self):
+        msg_dict = {"name": self.name}
+        if self.started_at:
+            msg_dict["started_at"] = self.started_at.strftime("%Y-%m-%dT%H:%M:%SZ")
+        if self.completed_at:
+            msg_dict["completed_at"] = self.completed_at.strftime("%Y-%m-%dT%H:%M:%SZ")
+        return msg_dict
```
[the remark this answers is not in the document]
gshank: I created a datetime_to_json_format_string utility method, and used it here and in base_types.py. The other one (timestampe_to_datetime_string) goes in a different direction, from the timestamp to a string and string is shorter

core/dbt/contracts/results.py
gshank: I think it should have the same output. In betterproto we had to provide the pre-constructed sub-messages (which is the way python works) and with google protobuf, you can't.

docs/arch/adr-005-betterproto.md
gshank: Dated.

core/dbt/events/base_types.py
gshank: The nested fire_event seems to be working okay. Added a test for it.

peterallenwebb (approved):

PR #7194, opened 2023-03-20
Remove trivial nodes before building subdag

ttusing:
Resolves [#7195](https://github.com/dbt-labs/dbt-core/issues/7195)

### Description

This implements the "smart trimming" suggested in (https://github.com/dbt-labs/dbt-core/issues/7195).

Local testing on my project:
DBT 1.4.4:
```
dbt build --select modelname
00:31:26  Running with dbt=1.4.4
00:31:27  Found 307 models, 3708 tests...
00:34:02  
00:34:08 
00:34:09  Concurrency: 16 threads (target='default')
00:34:09  
00:34:09  1 of 179 START sql table model  .................................... [RUN]
```
Build time: 2 minutes 45 seconds

Version 1.5.0-b4:
```
dbt build --select modelname
00:39:02  Running with dbt=1.5.0-b4
00:39:05  Found 307 models, 3708 tests...
00:41:14  
00:41:19  
00:41:20  Concurrency: 16 threads (target='default')
00:41:20  
00:41:20  1 of 179 START sql table model  .................................... [RUN]
```
Build time: 2 minutes 18 seconds

This branch:
```
dbt build --select modelname
05:28:42  Running with dbt=1.5.0-b4
05:28:45  Found 307 models, 3708 tests...
05:28:45  
05:28:52  
05:28:52  Concurrency: 16 threads (target='default')
05:28:52  
05:28:52  1 of 179 START sql table model  .................................... [RUN]
```
Build time: 10 seconds

### Checklist

### DIFF
--- a/core/dbt/graph/graph.py
+++ b/core/dbt/graph/graph.py
@@ -76,7 +76,7 @@ class Graph:
 
         while (
             True
-        ):  # remove nodes that have 0 or 1 degree and are not in include_nodes then repeat
+        ):
 
             nodes_to_remove = list(
                 node
--- a/core/dbt/graph/graph.py
+++ b/core/dbt/graph/graph.py
@@ -74,9 +74,7 @@ class Graph:
         new_graph = self.graph.copy()
         include_nodes = set(selected)
 
-        while (
-            True
-        ):
+        while True:
 
             nodes_to_remove = list(
                 node
@@ -87,7 +85,9 @@ class Graph:
                 break
             new_graph.remove_nodes_from(nodes_to_remove)
 
-        remaining_nodes = set(new_graph.nodes())
+        # sort remaining nodes by degree
+        remaining_nodes = list(new_graph.nodes())
+        remaining_nodes.sort(key=lambda node: new_graph.degree(node))
 
         for node in remaining_nodes:
             if node not in include_nodes:
--- a/core/dbt/graph/graph.py
+++ b/core/dbt/graph/graph.py
@@ -74,7 +74,22 @@ class Graph:
         new_graph = self.graph.copy()
         include_nodes = set(selected)
 
-        for node in self:
+        while (
+            True
+        ):  # remove nodes that have 0 or 1 degree and are not in include_nodes then repeat
+
+            nodes_to_remove = list(
+                node
+                for node in new_graph
+                if node not in include_nodes and new_graph.degree(node) < 2
+            )
+            if not nodes_to_remove:
+                break
+            new_graph.remove_nodes_from(nodes_to_remove)
+
+        remaining_nodes = set(new_graph.nodes())
+
+        for node in remaining_nodes:
             if node not in include_nodes:
                 source_nodes = [x for x, _ in new_graph.in_edges(node)]
                 target_nodes = [x for _, x in new_graph.out_edges(node)]
--- a/core/dbt/graph/graph.py
+++ b/core/dbt/graph/graph.py
@@ -74,20 +74,24 @@ class Graph:
         new_graph = self.graph.copy()
         include_nodes = set(selected)
 
-        while True:
-
+        still_removing = True
+        while still_removing:
             nodes_to_remove = list(
                 node
                 for node in new_graph
-                if node not in include_nodes and new_graph.degree(node) < 2
+                if node not in include_nodes
+                and (new_graph.in_degree(node) * new_graph.out_degree(node)) == 0
             )
-            if not nodes_to_remove:
-                break
-            new_graph.remove_nodes_from(nodes_to_remove)
+            if len(nodes_to_remove) == 0:
+                still_removing = False
+            else:
+                new_graph.remove_nodes_from(nodes_to_remove)
 
         # sort remaining nodes by degree
         remaining_nodes = list(new_graph.nodes())
-        remaining_nodes.sort(key=lambda node: new_graph.degree(node))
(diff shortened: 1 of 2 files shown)


ttusing:
> Thanks for your pull request, and welcome to our community! We require contributors to sign our [Contributor License Agreement](https://docs.google.com/forms/d/e/1FAIpQLScfOV7K4enYRHozrDRP6BBIXjOij-JDGca6WBTHyP_ANXSqlg/viewform) and we don't seem to have your signature on file. Check out [this article](https://docs.getdbt.com/docs/contributor-license-agreements) for more information on why we have a CLA.
> 
> In order for us to review and merge your code, please submit the Individual Contributor License Agreement form attached above above. If you have questions about the CLA, or if you believe you've received this message in error, please reach out through a comment on this PR.
> 
> CLA has not been signed by users: @ttusing

Believe this is in error! Filled it out a few minutes before opening PR, might be a timing issue.

core/dbt/graph/graph.py
```diff
@@ -74,7 +74,22 @@ def get_subset_graph(self, selected: Iterable[UniqueId]) -> "Graph":
         new_graph = self.graph.copy()
         include_nodes = set(selected)
 
-        for node in self:
+        while True:
```
peterallenwebb, line 77: The logic here looks correct, but we suggest a slightly different formulation which avoids using while True:

```python
        still_removing = True
        while still_removing:
            nodes_to_remove = list(
                node
                for node in new_graph
                if node not in include_nodes and new_graph.degree(node) < 2
            )
            if len(still_removing) == 0:
                still_removing = False
            else:
                new_graph.remove_nodes_from(nodes_to_remove)
```

core/dbt/graph/graph.py
```diff
@@ -74,7 +74,22 @@ def get_subset_graph(self, selected: Iterable[UniqueId]) -> "Graph":
...
+                for node in new_graph
+                if node not in include_nodes and new_graph.degree(node) < 2
+            )
+            if not nodes_to_remove:
+                break
+            new_graph.remove_nodes_from(nodes_to_remove)
+
+        # sort remaining nodes by degree
+        remaining_nodes = list(new_graph.nodes())
+        remaining_nodes.sort(key=lambda node: new_graph.degree(node))
```
peterallenwebb, line 90: This is an interesting heuristic. We agree it makes sense to consider nodes with smaller degree first for the reasons you outlined in the original issue. Would you consider evaluating a slightly different heuristic which we think might work as well or better?

The idea is that the number of edges (potentially) created by removing a node is the product of its in-degree and out-degree, and that perhaps degree is working because it is a rough proxy for that product. If that is the case, maybe sorting by the product instead would be even better?

```python
remaining_nodes.sort(key=lambda node: new_graph.in_degree(node) * new_graph.out_degree(node))
```

peterallenwebb (changes requested):

peterallenwebb:
Overall this looks like an exciting improvement. We're thinking through whether it is likely to work well for all of the graph topologies that come up in practice, but I'm pretty optimistic about that.

core/dbt/graph/graph.py
ttusing: Smart! Agreed.

ttusing:
> Overall this looks like an exciting improvement. We're thinking through whether it is likely to work well for all of the graph topologies that come up in practice, but I'm pretty optimistic about that.

I think this is strictly better than random selection in all cases, especially with switching to the product of `in_degree` and `out_degree`.

peterallenwebb:
@iknox-fa Looks like the issues we discussed have been addressed here. What do you think about this PR?

ttusing:
> @iknox-fa Looks like the issues we discussed have been addressed here. What do you think about this PR?

@iknox-fa  If part of the hesitation is the pruning part and the sorting is good to go, I can open a separate PR.

peterallenwebb:
@iknox-fa @jtcohen6 (cc: @leahwicz)

I'd like to approve this PR and get it into the RC for 1.5, and what follows is my case for doing that.

What we know:

- The code is correct. There is no disagreement on that point.
- It is 5-10x faster on long runs involving narrow node selections on large graphs, as benchmarked by @ttusing and at least one other user.
- The code involves heuristic changes to the order in which nodes are removed from the graph, and there is a strong basis for concluding that the heuristic will improve performance across a variety of graph types.

Based on my general algorithmic/performance experience, I'd guess the following:

- There is a 85% chance that this is faster for all users all of the time.
- There is a 10% chance that this is faster for users with graphs similar to the ones the change has been tried on, but no change or slightly slower for other graphs.
- There is 5% chance that the algorithm is significantly slower for some users, even though it dramatically improves performance in other cases.

Since we will have time to test the performance in the wild during the RC interval, I think it is worth a 5% risk of some performance regressions if it lets us dramatically improve the experience of many users in the 95% case. Even if there are performance regressions, I believe we could probably address them by rolling forward instead of back, but rolling back is easy.

jtcohen6:
@peterallenwebb Thanks for laying out the rationale - I'm down to give this a shot, and merge for inclusion in v1.5.0-rc1.

leahwicz:
@peterallenwebb thanks as well for the analysis. I feel good with getting this in the RC based off of that and monitoring to see if we see any regressions in perf from different DAG shapes.

peterallenwebb (approved):

PR #7202, opened 2023-03-21
Skip catalog generation dbt docs generate

### LINKED ISSUE #6980: [Feature] `--no-connection` flag for `dbt docs generate`

### Describe the feature

Provide an optional flag to be used with doc generation that skips the metadata queries and generates `catalog.json` based only off what it can get from compilation (and/or the existing `manifest.json if `--no-compile` is also passed).

The problem I am solving for is around enabling the docs to be used locally for exploratory purposes during local development in large projects where building the catalog would otherwise take a long time (in the case of our project with 11k nodes building the catalog without compile takes ~minutes).

This problem and solution is actually described exactly in https://github.com/dbt-labs/dbt-core/issues/3947#issuecomment-1349013372:

> I work with a few teams in the company and, as a developer, I would love to see their DAGs and understand better how they are developing/deploying their environment. How do other teams generate certain metrics, how do they use those metrics, and what kinds of tests do they typically perform across their metrics? I know that some teams use pre-hooks and post-hooks to determine things like row counts before and after runs to ensure things look right. I'd love to be able to explore this in the `dbt docs serve` format.
> 
> I can manually parse through this info on my own but digging through dozens of sql files across multiple folders is... a bit intense. That said, I'm not as interested in the actual data and don't have access anyways.
> 
> I understand the huge benefit to having connectivity to the database when generating docs but it would be great if you could put a flag on the command that allows you to forego any benefits of connectivity. Something like `dbt docs generate --no_connection` so that I could create enough of the docs to view their processes but knowing that I wouldn't have access to some of the metadata that I might get if I didn't use the flag.

The intention would be that it's not recommended to use this flag for building production-quality docs, but it would unlock quick exploration in large projects.

1.  Using dbt Cloud's in-editor DAG
2.  Parsing the manifest manually to create an explorable DAG (maybe this could be bundled as part of a vscode extension 😉)

### Who will this benefit?

Developers of large projects.

Provided it is within my capabilities, then yes!

### Anything else?

I imagine this feature would be (ab)used by many folks who want to build their docs in CI without a DB connection.  I'm not sure that this is behaviour we should actually be preventing though, as long as they understand the tradeoffs.

AndyBys:
resolves #6980 

### Checklist


jaypeedevlin:
Since I created the original issue, putting in my two cents that I don't love the `skip_catalogue_generation` name since it does generate a `catalog.json` (albeit empty)

AndyBys:
Hey, thanks for feedback. Sounds good to me. I still haven't made up my mind between renaming it to `generate_empty_catalog` (or any meaningful alternative) or rework codebase into the case where we skip generation indeed. Would be happy if you have thoughts on that.

core/dbt/task/generate.py
```diff
@@ -227,10 +228,14 @@ def run(self) -> CatalogArtifact:
...
-            fire_event(BuildingCatalog())
-            catalog_table, exceptions = adapter.get_catalog(self.manifest)
+        catalog_table: agate.Table = agate.Table([])
+        exceptions: List[Exception] = []
+
+        if not self.args.empty_catalog:
+            adapter = get_adapter(self.config)
+            with adapter.connection_named("generate_catalog"):
+                fire_event(BuildingCatalog())
+                catalog_table, exceptions = adapter.get_catalog(self.manifest)
```
iknox-fa, line 238: Just for readability sake, can we re-write this such that it's a little more clear what code is being executed on either side of the conditional and do so without re-assigning variables?  Maybe something like:

```
if self.args.empty_catalog:
    catalog_table: agate.Table = agate.Table([])
    exceptions: List[Exception] = []
else:
    adapter = get_adapter(self.config)
    with adapter.connection_named("generate_catalog"):
        fire_event(BuildingCatalog())
        catalog_table, exceptions = adapter.get_catalog(self.manifest)
```

iknox-fa (approved):
LGTM.  I have one small suggestion for readability but it looks good logically.  Thanks for your contribution!

core/dbt/task/generate.py
AndyBys: Thanks for flagging, fixed!

Issue #6866, opened 2023-02-06
[CT-2035] [Spike] Parsing: model ‘version’ attribute


MichelleArk:
Introduce an optional ‘versions’ attribute on model yaml entry. `versions` is a list of 'version' structures. Pseudocode: 

```python
ModelVersion:
    version: str
    implemented_by: Optional[str] = None # defaults to <model_name>.v<version>
    # all model properties (e.g. config, description, etc) 
    columns: Optional[List[Union[ColumnInfo, IncludeExclude]]]
```

`versions` will be specified within a single model yaml entry, but each version should maps to an individual node in the manifest with a `unique_id`: `model.<project_name>.<model_name>.v<version>`.

Example: 
```yml
models:
  - name: dim_customers
    latest_version: 2
    config:
      materialized: table
    columns:
      - name: customer_id
        description: This is the primary key
        data_type: int

    versions: # new!
      - version: 2
        # implemented_by: dim_customers.v2 - optional, this is the default naming convention
      - version: 1
        # implemented_by: dim_customers.v1 - optional, this is the default naming convention
        config:
          alias: dim_customers
        columns:
          - include: * 
            exclude: customer_id
          - name: customer_id
            data_type: float  # this was previously a float in version 1 for some reason
```

Notes:
- The model `name` is defined in the yaml file, and is globally unique but shared among `versions` of that model
- Each version of this model is "implemented by" a model SQL/python file. The `implemented_by` property points to the name of a `.sql` (or `.py`) file that implements that model version. By convention, this file is `<model_name>.v<version>.sql`, e.g. `dim_customers.v2.sql`. However, users could set `implemented_by` to point to a different SQL file that doesn't meet the default naming convention. (These file names must be globally unique, even though they are not the actual model `name`.)

jtcohen6:
> version should be a valid integer (for now)

https://github.com/dbt-labs/dbt-core/discussions/6736#discussioncomment-4881591

We should leave space for adding other versioning schemes in the future (namely SemVer). For now, that can be as simple as creating lightweight abstractions over the logic to:
- validate a version input (this issue)
- sort versions & determine which is "latest" (https://github.com/dbt-labs/dbt-core/issues/6867)

jtcohen6:
@MichelleArk tagging us for refinement :)

emmyoop:
[...]

PR #7204, opened 2023-03-21
[Spike] Model versions parsing

MichelleArk:
resolves https://github.com/dbt-labs/dbt-core/issues/6866

### Description

Implements most of the spec for versions described from the discussion: https://github.com/dbt-labs/dbt-core/discussions/6736#discussioncomment-5387269

Example spec: 
```yaml
models:
  - name: dim_customers
    latest_version: 3
    config:
      contract: true
      materialized: table
      meta:
        key: value
    columns:
      - name: customer_id
        description: This is the primary key
        data_type: float
      - name: country_name
        description: Where this customer lives
        data_type: string
    versions:
      - name: 2
      - name: 1
        config:
          alias: dim_customers
          deprecation_date: '2022-01-01'
        columns:
          - include: * 
            exclude: customer_id
          - name: customer_id
            data_type: int
      - name: 3
        defined_in: my_new_customers_definition # matches SQL or python file name
        columns:
          - include: *
            exclude: 
              - country_name 

# Manifest representation
# model.project_name.dim_customers.v1
# * name: dim_customers
# * version: 1
# * path: dim_customers_v1.sql
# * alias: dim_customers

# model.project_name.dim_customers.v2
# * name: dim_customers
# * version: 2
# * path: dim_customers_v2.sql
# * alias: dim_customers_v2

# model.project_name.dim_customers.v3
# * name: dim_customers
# * version: 3
# * path: my_new_customers_definition.sql
# * alias: dim_customers_v3
```


core/dbt/context/providers.py
```diff
@@ -212,16 +212,19 @@ def __call__(self, *args: str) -> Union[str, RelationProxy, MetricReference]:
...
 
-    def _repack_args(self, name: str, package: Optional[str]) -> List[str]:
-        if package is None:
-            return [name]
-        else:
-            return [package, name]
+    def _repack_args(self, name: str, package: Optional[str], version: Optional[str]) -> List[str]:
+        repacked_args = [name] if package is None else [package, name]
+        if version:
+            repacked_args.append(f"version:{version}")
```
MichelleArk, line 223: Opted to repack the `version` arg as a formatted string to keep the structure of `model.refs` as `List[List[str]]` (result of this method is [appended to model.refs here](https://github.com/dbt-labs/dbt-core/pull/7204/files#diff-d14b7d00e52da5da96bfb3bdef414e9289d923efe0a9cdadd716d792c7199535R465). 

Leads to some [not-so-fun string parsing](https://github.com/dbt-labs/dbt-core/pull/7204/files#diff-a25a77362db78f4ddc0675bcbd4775e158631076c9918ce5b64cfcd9f6db6bceR1190) when refs are processed. 

Could also add it as a dictionary, which would hold all kwargs (if we ever add more) going forward. This would be a breaking change in the manifest (model.refs becomes `List[List[Union[str, Dict[str,str]]]]`, but would be more straightforward to parse when processing refs.

core/dbt/contracts/graph/manifest.py
```diff
@@ -145,19 +145,22 @@ def perform_lookup(self, unique_id: UniqueID, manifest: "Manifest") -> SourceDef
...
     # model, seed, snapshot
     _lookup_types: ClassVar[set] = set(NodeType.refable())
+    _versioned_types: ClassVar[set] = set(NodeType.versioned())
 
     # refables are actually unique, so the Dict[PackageName, UniqueID] will
     # only ever have exactly one value, but doing 3 dict lookups instead of 1
     # is not a big deal at all and retains consistency
     def __init__(self, manifest: "Manifest"):
-        self.storage: Dict[str, Dict[PackageName, UniqueID]] = {}
+        self.storage: Dict[str, Dict[PackageName, List[UniqueID]]] = {}
```
MichelleArk, line 154: `ref` lookups now need to handle 3 access patterns (`package` handling remains unchanged):
1. Lookup versioned model by name and version => lookup specified version
2. Lookup versioned model by name _without_ version => get the latest version
3. Lookup unversioned model by name only => get only 'version'

For example, a manifest with the following nodes: 
```json
"model.jaffle_shop.dim_unversioned": {
  "name": "dim_unversioned",
  "version": null, 
  "is_latest_version": null
},
"model.jaffle_shop.dim_customers.v1": {
  "name": "dim_customers",
  "version": "1", 
  "is_latest_version": false
},
"model.jaffle_shop.dim_customers.v2": {
  "name": "dim_customers",
  "version": "2", 
  "is_latest_version": true
}
```
should find unique_ids for the following refs via `RefableLookup.get_unique_id`:
```
ref("dim_unversioned") => "model.jaffle_shop.dim_unversioned"
ref("dim_customers", version="1") => "model.jaffle_shop.dim_customers.v1"
ref("dim_customers") => "model.jaffle_shop.dim_customers.v2"
```

MichelleArk: I considered a few options for the `storage` structure here: 
* Dict[str, Dict[PackageName, List[UniqueID]]]
   * "latest" unique_id stored at the front of list
   * first element of `UniqueID` list always returned, unless a `version` is specified - in which case iterating through the list is necessary.  
   * implemented in this draft
```python
{
"dim_unversioned": {
    "jaffle_shop": ["model.jaffle_shop.dim_unversioned"]
  }, "dim_customers": {
    "jaffle_shop": ["model.jaffle_shop.dim_customers.v2", "model.jaffle_shop.dim_customers.v1"]
  }
}
```
* Dict[str, Dict[PackageName, Dict[VersionName, UniqueID]]]
  * doesn't handle unversioned or "latest" versioned lookups models nicely
```python
{ 
"dim_unversioned": {
  "jaffle_shop": {
    "??": "model.jaffle_shop.dim_unversioned"
  },
},
"dim_customers": {
  "jaffle_shop": {
      "2": "model.jaffle_shop.dim_customers.v2", 
      "1": "model.jaffle_shop.dim_customers.v1"
    }
  }
}
```
* Dict[str, Dict[PackageName, UniqueID]]
  * doesn't handle "latest" versioned lookups models nicely

```python
{ 
"dim_unversioned": {
  "jaffle_shop": model.jaffle_shop.dim_unversioned"
},
"dim_customers.v1": {
  "jaffle_shop": "model.jaffle_shop.dim_customers.v1"
},
"dim_customers.v2": {
  "jaffle_shop": "model.jaffle_shop.dim_customers.v2"
}
}
```

core/dbt/parser/schemas.py
```diff
@@ -878,12 +898,68 @@ def parse_patch(self, block: TargetBlock[NodeTarget], refs: ParserRef) -> None:
...
+                    package_name=versioned_model_node.package_name,
+                    description=unparsed_version.description or versioned_model_node.description,
+                    columns=version_refs.column_info,
+                    meta=unparsed_version.meta or versioned_model_node.meta,
+                    docs=unparsed_version.docs or versioned_model_node.docs,
+                    config=unparsed_version.config,
+                    access=versioned_model_node.access,
+                    version=unparsed_version.name,
+                    is_latest_version=latest_version == unparsed_version.name,
+                )
```
MichelleArk, line 957: some of these fields need a closer inspection - but the jist of it is that the versioned model node should inherit properties and configs from the base node patch (where the `versions` block is configured), extending or overriding them as appropriate.

core/dbt/parser/schemas.py
```diff
@@ -878,12 +898,68 @@ def parse_patch(self, block: TargetBlock[NodeTarget], refs: ParserRef) -> None:
...
+                versioned_model_node.patch(versioned_model_patch)
+                self.patch_node_config(versioned_model_node, versioned_model_patch)
```
MichelleArk, line 959: alias gets updated here

core/dbt/parser/schemas.py
```diff
@@ -878,12 +898,68 @@ def parse_patch(self, block: TargetBlock[NodeTarget], refs: ParserRef) -> None:
...
+            unique_id = self.manifest.analysis_lookup.get_unique_id(patch.name, None, None)
         else:
             raise DbtInternalError(
                 f"Unexpected yaml_key {patch.yaml_key} for patch in "
                 f"file {source_file.path.original_file_path}"
             )
+
+        # patch versioned nodes
+        versions = block.target.versions
+        if versions:
```
MichelleArk, line 910: this is the meat and potatoes of the change. Currently it's in `NonSourceParser`, which is subclassed by `TestablePatchParser` for parsing models, seeds, and snapshots. Given the increasing number of model-specific patching that's emerged as part of the 'Models as APIs' work (access, constraints, now versions) - I'm thinking this could be split out into a new subclass for parsing models patches independently of seeds and snapshots. Perhaps simply `ModelPatchParser`. What do you think @gshank, cc @peterallenwebb?

core/dbt/context/providers.py
jtcohen6: Let's go with the dictionary. IMO:
- consumers of the manifest should use `depends_on.nodes`, rather than `refs`, in almost all cases
- the list-of-lists always felt a bit janky

This may require some changes in some of the weirder packages out there, but in a way that should make their lives easier, ultimately

MichelleArk: I'm good to go with a dictionary as well!

What I'd originally thought up in terms of representation in the manifest.json for `model.refs` was a `List[List[Union[str, Dict[str,str]]]]`. For example, 
```json
"refs": ["package_name", "model_name", {"version":"3", ...}]
```

But what might be best is actually just a `List[Dict[str, str]]`. For example, 
```json
"refs": {
"package": "package_name",  # what should we actually call this key..?
"model": "model_name",
"version": "version_name"
}
```

That would enable us to simplify how we parse `ref` [here](https://github.com/dbt-labs/dbt-core/pull/7204/files#diff-a25a77362db78f4ddc0675bcbd4775e158631076c9918ce5b64cfcd9f6db6bceR1190) as well.

core/dbt/parser/schemas.py
```diff
@@ -878,12 +898,68 @@ def parse_patch(self, block: TargetBlock[NodeTarget], refs: ParserRef) -> None:
...
+                    meta=unparsed_version.meta or versioned_model_node.meta,
+                    docs=unparsed_version.docs or versioned_model_node.docs,
+                    config=unparsed_version.config,
+                    access=versioned_model_node.access,
+                    version=unparsed_version.name,
+                    is_latest_version=latest_version == unparsed_version.name,
+                )
+                versioned_model_node.patch(versioned_model_patch)
+                self.patch_node_config(versioned_model_node, versioned_model_patch)
+            self.manifest.rebuild_ref_lookup()  # TODO: is this necessary?
```
gshank, line 960: I wouldn't think you would need to rebuild the whole thing. There is a self.manifest.ref_lookup.add_node(...) method if you need to add the versioned nodes. It's not clear to me whether that would be necessary or not...

core/dbt/context/providers.py
gshank: I vote for List[Dict[str, str]] too. It's always a bit messy to have two different types.  We're not allowing versioned sources, right?  Sources currently List[List[str]] but always have two strings in the list, whereas refs only have one. Eventually it might be nice to switch all of those things to dictionaries.

MichelleArk: > We're not allowing versioned sources, right?
right - agree it'd be nice to switch them all to dictionaries. I can make an issue for sources, perhaps we can tackle it sa follow-on tech debt work.

core/dbt/contracts/graph/manifest.py
MichelleArk: For the third option, we _could_ store an additional key here that maps the model name to its latest unique id. For example, 
```

```python
{ 
"dim_unversioned": {
  "jaffle_shop": model.jaffle_shop.dim_unversioned"
},
"dim_customers.v1": {
  "jaffle_shop": "model.jaffle_shop.dim_customers.v1"
},
"dim_customers.v2": {
  "jaffle_shop": "model.jaffle_shop.dim_customers.v2"
},
"dim_customers.
}
```
```
This has the benefits of:
* keeping the storage data structure consistent across different lookup implementations
* mapping closely with the access patterns => easier to reason about + constant-time lookup for all access patterns

gshank: I assume that the ref_lookup code is going to have to handle external references eventually too, right? Not yet sure how that would impact these design choices.

I am a bit wary of having "latest" only implied by the order of versions in the first option.  That feels a bit fragile to me.  It doesn't feel like checking the versions would be that much overhead. And it's always going to be numerical order, so the highest integer is always latest, right?

MichelleArk: > I am a bit wary of having "latest" only implied by the order of versions in the first option. That feels a bit fragile to me. 

Same here. 

> And it's always going to be numerical order, so the highest integer is always latest, right?

This is something I haven't put in much polish on for the spike yet but - but we'll be accepting strings for the version identifier, as well as a model-level `latest_version`. If the latest_version is provided - then the latest is that value, regardless of what the highest version is (to support the concept of prereleases). 

If latest_version is not provided, dbt will compute a default latest version by attempting to order the version identifiers numerically if possible (casting to int/float), and fallback to alphanumeric ordering. This is to support a variety of versioning conventions out of the box, even if our recommendation is to use major versioning only. If

gshank: Yes, I think the latest proposed structure is appealing.

MichelleArk:
Got answers to how we'll implement `RefLookup` storage to handle versioned look ups, as well as more clarity in the spec from this spike. Closing now in favor of the new implementation ticket: https://github.com/dbt-labs/dbt-core/issues/7263

PR #7211, opened 2023-03-22
ADAP-387: Stub materialized view as a materialization

### LINKED ISSUE #6911: [CT-2068] [Feature] Let's add Materialized View as a materialization, finally

### A new materialization

___
#### Default implementation

- Configurations : none
- Lifecycle (`dbt run`) : 
      - `fail` : Fail the run
   - **If** necessary drop
   - Create

    1. ALTER in place 

___
#### Postgres

- Configurations:
- Lifecycle : _default_

___
#### Redshift

- Configurations:

___
#### BigQuery

- Configurations:

___

### Backward compatibility

___
___
___

### Risks

- **Default**
  - Configurations
  - Lifecycle : 
    - DDL : 
- **Postgres**
  - Configurations:
    - auto_refresh : mandatory no
- **Redshift**
  - Configurations: 
- **BigQuery**
  - Configurations:
    - refresh_interval_minutes
    - max_staleness

___

### Who will this benefit?

A lot of people:

### Anything else?

mikealfare:
resolves #6911

### Description

- Add “materialized view” is a materialization option
- A model that is configured as a “materialized view” implements a view materialization (for now)
- A model that is configured as a “materialized view” has a refresh option (despite that not applying to traditional views)

### Checklist


core/dbt/include/global_project/macros/materializations/models/materialized_view/create_materialized_view_as.sql
```diff
@@ -0,0 +1,11 @@
+-- MATERIALIZED_VIEW IMPLMENTATION
+{% macro create_materialized_view_as(relation, sql, config) %}
+    {{ adapter.dispatch('create_materialized_view_as',(relation, sql, config)) }}
```
McKnight-42, line 3: will we need to keep up the `macro_namespace convention from the experimental repo?`  {{ return(adapter.dispatch('create_materialized_view_as', macro_namespace = 'dbt_labs_materialized_views')(relation, sql, config)) }}

core/dbt/include/global_project/macros/materializations/models/materialized_view/create_materialized_view_as.sql
```diff
@@ -0,0 +1,11 @@
+-- MATERIALIZED_VIEW IMPLMENTATION
+{% macro create_materialized_view_as(relation, sql, config) %}
```
mikealfare, line 2: What does `config` get used for here? Will that be for the `on_configuration_change` piece?

core/dbt/include/global_project/macros/materializations/models/materialized_view/materialized_view.sql
```diff
@@ -0,0 +1,74 @@
...
+  {{ run_hooks(pre_hooks, inside_transaction=True) }}
+
+
+  -- cleanup
+  -- move the existing view out of the way
+  {% if existing_relation is none %}
+      {% set build_sql = create_materialized_view_as(target_relation, sql, config) %}
+
+  {% elif full_refresh_mode or existing_relation.type != 'materializedview' %}
+      {#-- Make sure the backup doesn't exist so we don't encounter issues with the rename below #}
```
mikealfare, line 46: I think three lines are already handled with lines 24, 26, and 34 above. I would look at `intermediate.sql` and `table.sql` as well to see how those were implemented.

Also, we will need an intermediate object for the recreation because we'll have data in the existing object and we'll need to load the new object up; so there's a non-trivial amount of downtime.

core/dbt/include/global_project/macros/materializations/models/materialized_view/materialized_view.sql
```diff
@@ -0,0 +1,74 @@
...
+      {% set backup_relation = existing_relation.incorporate(path={"identifier": backup_identifier}) %}
+      {% do adapter.drop_relation(backup_relation) %}
+
+      {% do adapter.rename_relation(target_relation, backup_relation) %}
+      {% set build_sql = create_materialized_view_as(target_relation, sql, config) %}
+      {% do to_drop.append(backup_relation) %}
+
+  {% else %}
+      {% set build_sql = refresh_materialized_view(target_relation, config) %}
+  {% endif %}
```
mikealfare, line 57: We don't execute `build_sql` after setting it.

core/dbt/include/global_project/macros/materializations/models/materialized_view/materialized_view.sql
```diff
@@ -0,0 +1,74 @@
...
+
+  {% elif full_refresh_mode or existing_relation.type != 'materializedview' %}
+      {#-- Make sure the backup doesn't exist so we don't encounter issues with the rename below #}
+      {% set backup_identifier = existing_relation.identifier ~ "__dbt_backup" %}
+      {% set backup_relation = existing_relation.incorporate(path={"identifier": backup_identifier}) %}
+      {% do adapter.drop_relation(backup_relation) %}
+
+      {% do adapter.rename_relation(target_relation, backup_relation) %}
+      {% set build_sql = create_materialized_view_as(target_relation, sql, config) %}
+      {% do to_drop.append(backup_relation) %}
```
mikealfare, line 53: This is handled by line 68.

mikealfare (commented):
I added a few comments, some of it might be things you're already working on given this is in draft status.

core/dbt/include/global_project/macros/materializations/models/materialized_view/create_materialized_view_as.sql
McKnight-42: sorry that is leftover from experimental implementation I believe similar to `config.get` which seems to be more of the norm for us so will remove

core/dbt/include/global_project/macros/materializations/models/materialized_view/materialized_view.sql
McKnight-42: yeah this was a point of interest. in `view` materialization they actually toss out that statement call for sql to a macro, would we want to do the same or keep with the `build_sql`?

core/dbt/include/global_project/macros/materializations/models/materialized_view/materialized_view.sql
McKnight-42: the entirety of the if statement or just specific part? both `table.sql` and `view.sql` have both a similar setup around the `backup_relation` setup, then the `build model` where we build out the main call statement and then this cleanup conditional. but for the MV would we not need to know if full refresh is triggered/dropped already before we build the sql_model?

mikealfare: Sorry, could have been more clear. I was referring to lines 43-5. But it now looks like we're not dropping the backup relation above if it already exists, just the intermediate one. I would put all of that setup before this if block since it doesn't really cost much to try to drop something that doesn't exist; and we would want to make sure we're clean before doing work, regardless of the logic in the if block.

McKnight-42: ah, that makes sense will re-add and remove from the conditional

core/dbt/include/global_project/macros/materializations/models/materialized_view/create_or_replace_materialized_view.sql
```diff
@@ -0,0 +1,11 @@
+{% macro create_or_replace_materialized_view() %}
```
mikealfare, line 1: I don't think we need both `create_as` and `create_or_replace_as`. I think we call them both `create_as`, and then use "create or replace" if it's supported. This would impact the materialization itself (whether we need to drop first or not).

core/dbt/include/global_project/macros/materializations/models/materialized_view/refresh_materialized_view.sql
```diff
@@ -0,0 +1,10 @@
+{% macro refresh_materialized_view(relation, config) %}
```
mikealfare, line 1: I doubt we'll need `config` here. We'll need it in determining *if* we need to refresh, but this should just be the logic to refresh, with the assumption that we've verified that we have to refresh prior to calling this macro.

core/dbt/include/global_project/macros/materializations/models/materialized_view/materialized_view.sql
```diff
@@ -0,0 +1,71 @@
...
+  {% set should_revoke = should_revoke(existing_relation, full_refresh_mode=True) %}
+  {% do apply_grants(target_relation, grant_config, should_revoke=should_revoke) %}
+
+  {% do persist_docs(target_relation, model) %}
+
+  {{ run_hooks(post_hooks, inside_transaction=True) }}
+
+  {{ adapter.commit() }}
+
+  {{ drop_relation_if_exists(backup_relation) }}
```
mikealfare, line 65: Drop the intermediate as well, just like above.

core/dbt/include/global_project/macros/materializations/models/materialized_view/materialized_view.sql
```diff
@@ -0,0 +1,71 @@
...
+  -- drop the temp relations if they exist already in the database
+  {{ drop_relation_if_exists(preexisting_intermediate_relation) }}
+  {{ drop_relation_if_exists(preexisting_backup_relation) }}
+
+  -- `BEGIN` happens here:
+  {{ run_hooks(pre_hooks, inside_transaction=True) }}
+
+  -- cleanup
+  -- move the existing view out of the way
+  {% if existing_relation is none %}
```
mikealfare, line 41: Referencing @Fleid's direction here for convenience:

> For the full refresh, I think we should try to, in order:
> - CREATE OR REPLACE - if CREATE OR REPLACE is supported
> - CREATE, DROP, ALTER NAME - if rename is supported
> - DROP, CREATE - if none of the above
> 
> For on_configuration_changes, we should try to, in order:
> - ALTER in place
> - Apply the full refresh order

This is going to vary by adapter; but I'd like to avoid creating versions of this materialization in adapters if possible. Hence I think we should adopt something like the strategies approach that we did with the incremental materialization. In other words, we have "replace" strategy, an "alter" strategy, a "refresh" strategy, etc. In some cases, those strategies may point to the same macros (e.g. create_materialized_view_as()); but that would allow maintainers to update just the piece that is specific to that platform without needing to maintain the rest of the materialization.

core/dbt/include/global_project/macros/db_api/materialized_view/create.sql
```diff
@@ -0,0 +1,10 @@
+{% macro db_api__create_materialized_view_as(relation, sql) %}
+    {{ adapter.dispatch('db_api__create_materialized_view_as', 'dbt')(relation, sql) }}
+{% endmacro %}
+
+{% macro default_db_api__create_materialized_view_as(relation, sql) -%}
```
mikealfare, line 5: A few things:

1. Typo: you need a __ after default.
2. We should go with a naming convention of __-separated path. So that would make this macro `db_api__materialized_view__create()` and `default__db_api__materialized_view__create()`.

core/dbt/include/global_project/macros/db_api/materialized_view/refresh.sql
```diff
@@ -0,0 +1,10 @@
+{% macro db_api__materialized_view__refresh(relation) %}
+    {{ adapter.dispatch('db_api__materialized_view__refresh', 'dbt')(relation) }}
+{% endmacro %}
+
+{% macro default__db_api_refresh_materialized_view(relation) -%}
```
mikealfare: Same comment here regarding the naming convention, though the interface macro is correctly named.

core/dbt/include/global_project/macros/materializations/models/materialized_view/materialized_view.sql
```diff
@@ -0,0 +1,72 @@
...
+      {% set build_sql = strategy__materialized_view__create(target_relation, sql) %}
+  {% elif full_refresh_mode or existing_relation.type != 'view' %}
```
mikealfare, line 43: We may not always be able to rename the materialized view, so that step needs to be in the strategy; it can't be here. Also, this should be the `full_refresh` strategy, not the `create` strategy.

core/dbt/include/global_project/macros/strategies/materialized_view/create.sql
```diff
@@ -0,0 +1,3 @@
+{% macro strategy__materialized_view__create(relation, sql) %}
```
mikealfare, line 1: I think we need defaults on these, just like the `db_api` macros. So these would dispatch (`adapter.dispatch(...)`) and then the default would have the body you have here (`db_api__...`). In contrast to the `db_api` macros, I think it's fine to a default strategy that's functional (strategies are probably more reusable and don't directly rely on platform specific sql); so don't raise an exception like we do in the `db_api` macros.

core/dbt/include/global_project/macros/strategies/materialized_view/full_refresh.sql
```diff
@@ -0,0 +1,4 @@
+{% macro strategy__materialized_view__full_refresh(relation, sql) %}
```
mikealfare, line 1: Same comment as on `/create.sql`

core/dbt/include/global_project/macros/materializations/models/materialized_view/materialized_view.sql
McKnight-42: added `backup_relatiojn` as an optional arg for full_refresh strategy and passed down the rename trigger to it

core/dbt/include/global_project/macros/strategies/materialized_view/create.sql
McKnight-42: added the two defaults

core/dbt/include/global_project/macros/strategies/materialized_view/full_refresh.sql
```diff
@@ -0,0 +1,12 @@
+{% macro strategy__materialized_view__full__refresh(relation, sql, backup_relation=None) %}
```
mikealfare, line 1: There's an extra _ in `full__refresh` (also appears below and when called in the materialization).

Do we need to pass in `backup_relation`? I'm on the fence about it. But if we do, then we would need the intermediate relation as well, so that we can do the swap. I think the order would look like:
1. Create the new view as `intermediate_relation`
2. Swap names: `target_relation` > `backup_relation`
3. Swap names: `intermediate_relation` > `target_relation`
4. Drop `backup_relation`

core/dbt/include/global_project/macros/materializations/models/materialized_view/materialized_view.sql
mikealfare: The arg order is swapped here (`backup_relation` <> `sql`).

Also, see my comment below regarding `backup_relation`. I think I'm leaning towards including it, but then also include `intermediate_relation`.

core/dbt/include/global_project/macros/materializations/models/materialized_view/materialized_view.sql
```diff
@@ -0,0 +1,71 @@
...
+      {% set build_sql = strategy__materialized_view__full__refresh(target_relation, backup_relation, sql) %}
+  {% else %}
+      {% set build_sql = strategy__materialized_view__full__refresh(target_relation) %}
+  {% endif %}
+
+    -- possible conditional wrapper for build_sql statement based on configs as a option?
+    -- build model
+  {% call statement("main") %}
```
mikealfare, line 51: We need to make sure this can actually exist outside of the if block above. For example, does `adapter.rename()` occur when you call it? Or does it just return the sql to do so? We're fine if it's the latter. But if it's the former, we'll do all the thing that happen immediately during the if block, and then do all the things that we return sql for during this phase. That may not be the right order.

nssalian (approved):
Took a look, and overall this looks to be in good shape after all the comments and changes. I would be ok putting it in the feature branch to move forward the work.

PR #7214, opened 2023-03-22
CT-1928: dbtRunner to EventManager callback support

### LINKED ISSUE #6763: [CT-1928] Support passing event callbacks to EventManager via dbtRunner
Accept event [callbacks](https://github.com/dbt-labs/dbt-core/blob/ed50877c4fe5a74aa8cdd203bc9aceac979d9243/core/dbt/events/eventmgr.py#L186) as part of [`dbtRunner`'s constructor](https://github.com/dbt-labs/dbt-core/blob/ed50877c4fe5a74aa8cdd203bc9aceac979d9243/core/dbt/cli/main.py#L34-L41) and use them to setup the `EventManager` for programmatic invocations of dbt.

QMalcolm:
resolves #6763 

# Description
We need to be able to set callbacks for the `EventManager` when using the `dbtRunner` for programmatic implementations, as described in [CT-1928](https://github.com/dbt-labs/dbt-core/issues/6763). The way it works is on instantiation of `dbtRunner` one can now provide `callbacks`. These callbacks are for the `EventLogger`. When `invoke` is called on a `dbtRunner`, the `callbacks` are added to the cli context object. In the preflight decorator these callbacks are extracted from the cli context and then passed to the `setup_event_logger`, finally `setup_event_logger` ensures the callbacks are added to the global `EVENT_MANAGER`.

<a href="https://www.loom.com/share/2e9fb9999f9849c7ba9f1cc1d51febb3">
    
</a>

## Problems
This implementation has two problems: DevEx and Global State Shenanigans

### DevEx Gripe
To make this work we had to set `callbacks` on the cli context `obj`. This `obj` is basically a grab bag of stuff which may or may not exist, and may or may not be necessary. It would be nice to have a more explicit passing of parameters so that it's clearer where and why they are necessary

### `EVENT_MANAGER` Global State Shenanigans
Every invocation of the CLI shares the same `EventManager`, which is the global `EVENT_MANAGER`. This isn't problematic so long as only one invocation is running at a time. However, as soon as there are two or more invocations at a time (I don't think we currently support this?) things will start breaking. Things will start breaking because `setup_event_logger` is destructive and called during every invocation. Specifically it clears the existing loggers and callbacks associated with the `EVENT_MANAGER`. Though even if this didn't happen, the global `EVENT_MANAGER` would still be problematic because `setup_event_logger` would be adding extra callbacks to other unrelated invocations 🫠

Loom video coming soon™️ 

### Checklist

[CT-1928]: https://dbtlabs.atlassian.net/browse/CT-1928?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/core/dbt/events/eventmgr.py
+++ b/core/dbt/events/eventmgr.py
@@ -215,12 +215,8 @@ class EventManager:
         self.loggers.append(logger)
 
     def add_callbacks(self, callbacks: List[Callable[[EventMsg], None]]):
-        """Helper for bulk adding a sequence of Callbacks to the EventManager"""
+        """Helper for adding a sequence of Callbacks to the EventManager"""
         self.callbacks.extend(callbacks)
-
-    def add_callback(self, callback: Callable[[EventMsg], None]):
-        """Helper which adds a singular Callback to the EventManager"""
-        self.callbacks.append(callback)
 
     def flush(self):
         for logger in self.loggers:
--- a/core/dbt/events/eventmgr.py
+++ b/core/dbt/events/eventmgr.py
@@ -214,6 +214,14 @@ class EventManager:
         logger.event_manager = self
         self.loggers.append(logger)
 
+    def add_callbacks(self, callbacks: List[Callable[[EventMsg], None]]):
+        """Helper for bulk adding a sequence of Callbacks to the EventManager"""
+        self.callbacks.extend(callbacks)
+
+    def add_callback(self, callback: Callable[[EventMsg], None]):
+        """Helper which adds a singular Callback to the EventManager"""
+        self.callbacks.append(callback)
+
     def flush(self):
         for logger in self.loggers:
             logger.flush()
--- a/.changes/unreleased/Features-20230322-124615.yaml
+++ b/.changes/unreleased/Features-20230322-124615.yaml
@@ -1,0 +1,6 @@
+kind: Features
+body: Support setting of callbacks for programmatic uses of `dbtRunner`
+time: 2023-03-22T12:46:15.877884-07:00
+custom:
+  Author: QMalcolm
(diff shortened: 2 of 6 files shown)


core/dbt/events/eventmgr.py
```diff
@@ -214,6 +214,14 @@ def add_logger(self, config: LoggerConfig):
         logger.event_manager = self
         self.loggers.append(logger)
 
+    def add_callbacks(self, callbacks: List[Callable[[EventMsg], None]]):
+        """Helper for bulk adding a sequence of Callbacks to the EventManager"""
+        self.callbacks.extend(callbacks)
+
+    def add_callback(self, callback: Callable[[EventMsg], None]):
```
ChenyuLInx, line 221: Do we need this function?

QMalcolm: We do not need it, no. I included it as syntactic sugar for anyone who has a use case wherein they know they are only adding one callback or have to add callbacks one at a time due to having to track some extra context whenever adding a callback. Essentially when adding only one callback
```
EVENT_MANAGER.add_callback(singular_callback)
```
is more idiomatic than having people do
```
EVENT_MANAGER.add_callbacks([singular_callback])
```
Functionally though there is no difference :shrug:

QMalcolm: Dropped in https://github.com/dbt-labs/dbt-core/pull/7214/commits/9bceac887084a248e3f5d52612aa928af9e00569

ChenyuLInx (approved):
Looks great!!

PR #7216, opened 2023-03-23
Detect breaking changes to column names and data types in state:modified check

### LINKED ISSUE #6869: [CT-2038] Detect breaking changes to column names and data types in state:modified check
If a model with `contract:true` has its column name/data types modified in a way that represents a breaking change, dbt should raise an error as part of the `state:modified` check. 

What constitutes a breaking change?
- Removal of column
- name change of column
- data_type change of column 
- changing `contract:true` to `contract:false`

gshank:
resolves #6869

### Description

Add "same_contract", "state:modified.contract" to defer_state checks, create new exception and tests.

### Checklist


gshank:
Please look at the exception message to see what you want.

core/dbt/exceptions.py
```diff
@@ -209,6 +209,24 @@ def _fix_dupe_msg(self, path_1: str, path_2: str, name: str, type_name: str) ->
...
+        super().__init__(self.message(), node)
+
+    @property
+    def type(self):
+        return "Contract"
+
+    def message(self):
+        return (
+            "There is a breaking change in the model contract; "
+            "you may need to create a new version. See: https://docs.getdbt.com/docs/collaborate/publish/model-versions"
```
dbeatty10, line 226: This error message looks good to me 👍

MichelleArk: If the user is aware they're making a breaking change, and just needs some guidance regarding making a new version of the model - this error message looks good to me too. 

That said, I am worried that this error message could be a little opaque for unintented breaking changes, especially as [our set of conditions that constitute breaking changes becomes more nuanced](https://github.com/dbt-labs/dbt-core/issues/7065). For example, a user updates a contracted model with constraints from `table` to `view` in an attempt to optimize some spend, which raises a breaking change error because the existing constraints can't be validated. Just seeing `There is a breaking change in the model contract` could be confusing, and perhaps the user would actually choose to just revert that change and make it at a later point (bundle it with some other breaking changes). In that scenario, I'd (as an imaginary user) love to see an error message that looks more like: 
```
There is a breaking change in the model contract: 
 * Updated materialization from `table` to `view` for model with constraints
 * Updated config from contract: true to contract: false
 * Column a removed
 * Column c updated data_type from string to int
 * ...
You may need to create a new version. See: https://docs.getdbt.com/docs/collaborate/publish/model-versions
```

This may not be concern with the current set of conditions for breaking changes (any change to column name/data_type => breaking change), but given that we're planning to extend this to be more nuanced soon in https://github.com/dbt-labs/dbt-core/issues/7065, I'd leave some design space for it if this scenario seems important to prioritize the UX of (cc @jtcohen6).

gshank: Currently the change of a materialization is not a contract change, it's a config change. The ticket doesn't mention materialization -- do you want that to be part of the contract change too?

Creating that more user-friendly error message will probably be about twice as much work than everything else so far. If you want me to spend my time doing that, I can.

MichelleArk: Materializations are not part of this issue, but they will be part the work to extend detecting breaking changes in https://github.com/dbt-labs/dbt-core/issues/7065 (which i still blocked by https://github.com/dbt-labs/dbt-core/issues/7067). If we know we'll want more granular error messaging as part of that work, we could set up the foundations here or leave it to https://github.com/dbt-labs/dbt-core/issues/7065.

gshank: I added a brief "reason" to the exception message. We can do more elaborate messaging once the other parts are in place.

jtcohen6: (good conversation, no notes!)

core/dbt/exceptions.py
```diff
@@ -209,6 +209,25 @@ def _fix_dupe_msg(self, path_1: str, path_2: str, name: str, type_name: str) ->
...
+        self.reason = reason
+        super().__init__(self.message(), node)
+
+    @property
+    def type(self):
+        return "Contract"
+
+    def message(self):
+        return (
+            "There is a breaking change in the model contract because {self.reason}; "
```
MichelleArk, line 226: missing the f-string for formatting `self.reason` here.

tests/functional/defer_state/test_modified_state.py
```diff
@@ -261,3 +263,38 @@ def test_changed_exposure(self, project):
...
+        expected_unrendered_config = {"contract": True, "materialized": "table"}
+        assert model.unrendered_config == expected_unrendered_config
+
+        # Run it again with "state:modified:contract", still finds modified due to contract: true
+        results = run_dbt(["run", "--models", "state:modified.contract", "--state", "./state"])
+        assert len(results) == 1
+        # save a new state
+        self.copy_state()
+
+        # This should raise because a column has added a constraint
```
MichelleArk, line 292: Adding a constraint shouldn't be a breaking change - just a regular change to the model contract (returns `False` from `same_contract`, but does not raise `ModelContractError`.

core/dbt/contracts/graph/nodes.py
```diff
@@ -497,6 +507,39 @@ def depends_on_nodes(self):
...
+        if old.contract is False and self.contract is False:
+            # Not a change
+            return True
+        if old.contract is False and self.contract is True:
+            # A change, but not a breaking change
+            return False
+        if old.contract is True and self.contract is False:
+            # Breaking change: throw an error
+            raise (ModelContractError(reason="contract has been disabled", node=self))
+        if self.contract_checksum == old.contract_checksum:
```
MichelleArk, line 536: Should this be a `!=` instead? I'd expect the `contract_checksum` values to be different between `self` and `old` if the model contract has changed.

core/dbt/contracts/graph/nodes.py
```diff
@@ -497,6 +507,39 @@ def depends_on_nodes(self):
...
+            # Breaking change: throw an error
+            raise (ModelContractError(reason="column definitions have changed", node=self))
```
MichelleArk, line 538: The reasons for a breaking change can occur at the same time in a state modification check. For example, column definitions changing as well as a contract going from enabled to disabled. Given that these checks are typically used in CI - could we batch up the reasons and raise a single `ModelContractError` with an array of `reasons`? That way the user would have the full picture in their CI logs.

core/dbt/parser/base.py
```diff
@@ -314,6 +314,7 @@ def update_parsed_node_config(
         # compatibility with earlier node-only config.
         if config_dict.get("contract", False):
             parsed_node.contract = True
+            parsed_node.build_contract_checksum()
```
MichelleArk, line 317: This should be called once the `parsed_node` has `columns`, since the contract checksum is derived from the columns. Otherwise `parsed_node.columns` will always be `{}` and produce the same checksum. I think that's what's currently happening. Perhaps we could call `build_contract_checksum` [here](https://github.com/dbt-labs/dbt-core/blob/main/core/dbt/parser/schemas.py#L940) instead, after the config and columns are patched, and constraints are validated.

core/dbt/contracts/graph/nodes.py
```diff
@@ -492,6 +502,40 @@ def depends_on_nodes(self):
...
+        breaking_change_reasons = []
+        if old.contract is True and self.contract is False:
+            # Breaking change: throw an error
+            breaking_change_reasons.append("contract has been disabled")
+        if self.contract_checksum != old.contract_checksum:
+            # Breaking change: throw an error
+            breaking_change_reasons.append("column definitions have changed")
+        else:
+            # No change
+            return False
```
MichelleArk, line 536: should this be `True` instead, since there is no change?

Also, it's possible to hit this line and return early even if there are elements in `breaking_change_reasons` from the first case (`contract_checksum` has not changed but `self.contract` changed from `True` to `False`). Perhaps instead of this else statement we need a separate block at the end of this method: 

```python
if breaking_change_reasons: 
   raise (ModelContractError(reasons=" and ".join(breaking_change_reasons), node=self))
else: 
  # No change
  return True
```

gshank: Yeah, that last bit should be True. But there's another issue here, which is that right now I'm not generating the contract_checksum if contract is false, so we're not actually ever going to get both "contract has been disabled" and "column definitions have changed".  It feels kind of wrong to build the contract_checksum if contract enabled is false, but do you want it anyway?

gshank: Well, technically it would be a generated_checksum != None. But there might not even be any column changes for that None, so it would be wrong to flag it.

gshank: I've rearranged it so I think it should work now.

core/dbt/contracts/graph/nodes.py
```diff
@@ -492,6 +502,46 @@ def depends_on_nodes(self):
...
+            return True
+        if old.contract is False and self.contract is True:
+            # A change, but not a breaking change
+            return False
+
+        breaking_change_reasons = []
+        if old.contract is True and self.contract is False:
+            # Breaking change: throw an error
+            # Note: we don't have contract_checksum for current node, so build
+            self.build_contract_checksum()
```
MichelleArk, line 532: Just checking my understanding: instead of building the contract checksum for every model, we only build it for models with enabled contracts or previously enabled contracts (which we detect at this point, in the state:modified check).

gshank: That's the current state, yes. Does that feel right to you?

MichelleArk: Yes - this feels right! Nice way to ensure we can detect the full set of breaking changes without having to unnecessarily compute a checksum for every model, even those that with unenforced contracts. The comment is definitely helpful here though! 👍

core/dbt/contracts/graph/nodes.py
```diff
@@ -492,6 +502,46 @@ def depends_on_nodes(self):
...
+        # have contract enabled, because it won't be used.
+        # This needs to be executed after contract config is set
+        if self.contract is True:
+            contract_state = ""
+            # We need to sort the columns so that order doesn't matter
+            # columns is a str: ColumnInfo dictionary
+            sorted_columns = sorted(self.columns.values(), key=lambda col: col.name)
+            for column in sorted_columns:
+                contract_state += f"|{column.name}"
+                contract_state += str(column.data_type)
```
MichelleArk, line 516: Makes sense to leave out constraints for the scope of this issue 👍 

We should revisit whether constraints should go in the checksum when we go to implement https://github.com/dbt-labs/dbt-core/issues/7065, where:
* breaking changes to constraints should raise a `ModelContractError` from `same_contract`
* non-breaking changes to constraints (e.g. adding a constraint, removing an unenforced constraint) should return `False` from `same_contract`. 

Determining the nature of a change to a constraint will need additional logic beyond comparing checksums, and adding constraints to the checksum will make understanding changes to column data_type and names a bit trickier. I could see the checksum being a useful to determine, crudely, whether there was a _change at all_, and needing to actually inspect the columns/constraints more closely to determine the type of change - breaking vs non-breaking.

mikealfare: What is the benefit of putting these methods on `CompiledNode` versus carrying the columns on `Contract` and overriding `__eq__()` on `Contract` (in place of `same_contract()`)?

gshank: We are not prepared to move those fields at this time, since it would be a breaking change for packages which use those fields. In addition, not all of the attributes of a column count for breaking changes, so a simple equal doesn't work.

MichelleArk (approved):

PR #7217, opened 2023-03-23
Use function for datetime in CommandCompleted and throw error when testing

gshank:
### Description

Use utility function "get_json_string_utcnow" to set the time in the CommandCompleted message. Throw errors when testing and unable to parse the kwarg dictionary in the logging event base class.

### Checklist

### DIFF
--- a/core/dbt/events/base_types.py
+++ b/core/dbt/events/base_types.py
@@ -74,8 +74,13 @@ class BaseEvent:
             from dbt.events.types import Note
             from dbt.events.functions import fire_event
 
-            fire_event(Note(msg=f"[{class_name}]: Unable to parse dict {kwargs}"))
-            self.pb_msg = msg_cls()
+            error_msg = f"[{class_name}]: Unable to parse dict {kwargs}"
+            # If we're testing throw an error so that we notice failures
+            if "pytest" in sys.modules:
+                raise Exception(error_msg)
+            else:
+                fire_event(Note(msg=error_msg))
+                self.pb_msg = msg_cls()
 
     def __setattr__(self, key, value):
         if key == "pb_msg":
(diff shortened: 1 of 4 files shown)


core/dbt/events/base_types.py
```diff
@@ -74,8 +74,13 @@ def __init__(self, *args, **kwargs):
...
 
-            fire_event(Note(msg=f"[{class_name}]: Unable to parse dict {kwargs}"))
-            self.pb_msg = msg_cls()
+            error_msg = f"[{class_name}]: Unable to parse dict {kwargs}"
+            # If we're testing throw an error so that we notice failures
+            if "pytest" in sys.modules:
+                raise Exception(error_msg)
+            else:
+                fire_event(Note(msg=error_msg))
+                self.pb_msg = msg_cls()
```
jtcohen6, line 83: ah! this is good!

jtcohen6: does this also mean that "fire a warning instead of raising an exception on event serialization errors" was effectively resolved by the proto changes?
- https://github.com/dbt-labs/dbt-core/issues/7113

gshank: Well, probably, but mainly because the google protobuf messages are more strongly typed, so that errors are caught here instead of on serialization. As a belt-and-suspenders thing it's probably not a bad idea to catch on serialization too (though I haven't actually seen any errors there).

core/dbt/events/base_types.py
```diff
@@ -74,8 +74,13 @@ def __init__(self, *args, **kwargs):
             from dbt.events.types import Note
             from dbt.events.functions import fire_event
 
-            fire_event(Note(msg=f"[{class_name}]: Unable to parse dict {kwargs}"))
-            self.pb_msg = msg_cls()
+            error_msg = f"[{class_name}]: Unable to parse dict {kwargs}"
+            # If we're testing throw an error so that we notice failures
+            if "pytest" in sys.modules:
```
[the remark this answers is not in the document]
gshank: I don't think that we want to issue warnings when users are running dbt though.

peterallenwebb (approved):

PR #7222, opened 2023-03-24
convert contract to dict

### LINKED ISSUE #7184: [CT-2314] [Feature] Do we _have_ to call the boolean property `contract`?

### Describe the feature

From reading https://github.com/dbt-labs/dbt-core/issues/6748, it looks like I'm about 3 weeks late to this party. But better now than in another 3 weeks! 

I haven't been following this closely, and when I saw `contract:` I assumed it was going to be the key of a dictionary containing all the rules of the contract, not a simple bool. 

Could I make a late breaking pitch for `has_contract`? 

- I would also accept `is_contracted`, but I don't think I'll get as far on that one. 
- Or go back to the original naming scheme, but just change the noun (i.e. `contract_enabled` instead of `constraints_enabled`)

### Who will this benefit?

People who assume we follow [our style guide](https://github.com/dbt-labs/corp/blob/main/dbt_style_guide.md#:~:text=Booleans%20should%20be%20prefixed%20with%20is_%20or%20has_.) (yes I know this is the SQL style guide not the Python style guide, but it's going to be side-by-side with all the YAML that does follow this scheme)

I could if it's the only way it'll get done, but I won't be the fastest way to get it done

### Anything else?

_No response_

emmyoop:
resolves #7184

### Description

Converts contract from a bool to it's own object

`contract: bool = False`
to 

```python
@dataclass
class Contract(dbtClassMixin, Replaceable):
    enforced: bool = False
```

### Checklist


gshank (commented):
I have a question about the use of a dictionary here. In my PR (also up for review) I'm adding a "contract_checksum" as a separate field, but which I assume will go in this new contract "thing". Do we actually want it to be a dictionary or an object with fields, such as "enabled" and "checksum"? There are pros and cons for each...

core/dbt/parser/base.py
```diff
@@ -310,10 +310,11 @@ def update_parsed_node_config(
             else:
                 parsed_node.docs = Docs(show=docs_show)
 
-        # If we have "contract" in the config, copy to node level, for backwards
+        # If we have contract in the config, copy to node level, for backwards
         # compatibility with earlier node-only config.
-        if config_dict.get("contract", False):
-            parsed_node.contract = True
+        if "contract" in config_dict and config_dict["contract"]:
+            # breakpoint()
```
MichelleArk, line 316: nit: this was probably left in by accident!

core/dbt/parser/schemas.py
```diff
@@ -940,7 +944,9 @@ def parse_patch(self, block: TargetBlock[NodeTarget], refs: ParserRef) -> None:
 
     def validate_constraints(self, patched_node):
         error_messages = []
-        if patched_node.resource_type == "model" and patched_node.config.contract is True:
+        contract_config = patched_node.config.get("contract")
+        # breakpoint()
```
MichelleArk, line 948: ditto

core/dbt/parser/base.py
emmyoop: oops.  fixed.

core/dbt/contracts/graph/unparsed.py
```diff
@@ -87,6 +87,11 @@ class Docs(dbtClassMixin, Replaceable):
     node_color: Optional[str] = None
 
 
+@dataclass
+class Contract(dbtClassMixin, Replaceable):
```
gshank, line 91: It doesn't look like the Contract class is used by anything in unparsed.py, so it ought to go in nodes.py.  We should limit what nodes.py uses from unparsed.py as much as possible.

gshank (changes requested):
Looks good, except for the location of the Contract class.

core/dbt/contracts/graph/unparsed.py
emmyoop: That creates circular imports with model_config.  Instead of `nodes.py` I moved it into `model_config.py`.  The only other thing defined in there is `Hook` but I think it still makes sense since it's not used in `unparsed.py`.

gshank (approved):

PR #7223, opened 2023-03-26
CT-2322/improve contracts error message

### LINKED ISSUE #7209: [CT-2322] [Feature] Side-by-side diff when an enforced model contract fails

### Describe the feature

See this Loom starting at 12:50 for context of this feature proposal:
https://www.loom.com/share/f436e5e336e4433695dc63d53ee72fcf?t=770

Two things:
- Update "Contracts are enabled for this model." to "This model has an enforced contract."
- Do a side-by-side diff rather than a stacked listing of column names

Where a preflight check fails for a model contract, could the output look more like this?

image

Right now, the output looks like this:

image snip...
image

Sometimes the status quo is the way to go ...
... and sometimes it isn't 😉 

### Who will this benefit?

This will benefit folks that:
- are using enforced model contracts
- there is a mismatch in the contract leading to a preflight failure
- have a model with tons of columns (or even just a few!)

_No response_

### Anything else?

_No response_

kentkr:
resolves #7209 

### Description

If a model has a contract error dbt currently prints the yaml columns and sql columns, respectively, as a single string with delimited column names and data type. This is hard to read. The changes made here split the strings and print the yaml and sql columns side by side.

Note that the columns are printed in the order that they are listed in the model or in the yaml file. So columns may not line up correctly. Because of this I decided not to add an extra column that checks if the two columns are the same, a feature requested in the original issue. 

I'm not all that familiar with OOP and don't know how `self.yaml_columns` (and sql columns) is inherited or where it comes from. But surely it can be inherited as something other than a delimited string. If it was a dict (`{col : data_type}`) it'd be easier to compare and print in an ordered format. That may be more effort than its worth. @jtcohen6 any suggestions?

### Checklist

### DIFF
--- a/core/dbt/exceptions.py
+++ b/core/dbt/exceptions.py
@@ -2136,7 +2136,7 @@ class ContractError(CompilationError):
             "Please ensure the name, data_type, and number of columns in your `yml` file "
             "match the columns in your SQL file.\n"
             # separate column headers by 20 spaces
-            f"{'yaml col:' : <20} {'sql col:' : <20} {'issue:' : <20}\n"
+            f"{'sql col:' : <20} {'yaml col:' : <20} {'issue:' : <20}\n"
         )
 
         # list of ordered matches (or not matches) for printing
--- a/core/dbt/exceptions.py
+++ b/core/dbt/exceptions.py
@@ -2132,7 +2132,7 @@ class ContractError(CompilationError):
 
     def get_message(self) -> str:
         msg = (
-            "Contracts are enabled for this model. "
+            "This model has an enforced contract that failed.\n"
             "Please ensure the name, data_type, and number of columns in your `yml` file "
             "match the columns in your SQL file.\n"
             f"Schema File Columns: {self.yaml_columns}\n"
--- a/core/dbt/exceptions.py
+++ b/core/dbt/exceptions.py
@@ -2136,23 +2136,44 @@ class ContractError(CompilationError):
             "Please ensure the name, data_type, and number of columns in your `yml` file "
             "match the columns in your SQL file.\n"
             # separate column headers by 20 spaces
-            f"{'yaml col:' : <20} sql col:\n"
+            f"{'yaml col:' : <20} {'sql col:' : <20} {'issue:' : <20}\n"
         )
 
-        # print columns pretty
-        # split yaml and sql strings into list
-        yaml_list = self.yaml_columns.split(', ')
-        sql_list = self.sql_columns.split(', ')
-        # make sure lists are the same length, add empty string if not
-        yaml_list += ['']*(len(sql_list) - len(yaml_list))
-        sql_list += ['']*(len(yaml_list) - len(sql_list))
-        # for each col
-        for i in range(len(yaml_list)):
-            # separate yaml and sql columns by 20 spaces, append to message
-            msg += f"{yaml_list[i] : <20} {sql_list[i]}\n"
+        # list of ordered matches (or not matches) for printing
+        ordered_match = []
+        # track sql cols so we don't need another for loop later
+        sql_col_set = set()
+        # for each sql col list
+        for sql_col in self.sql_columns:
+            # add sql col to set
+            sql_col_set.add(sql_col['name']) 
+            # for each yaml col list
+            for i, yaml_col in enumerate(self.yaml_columns):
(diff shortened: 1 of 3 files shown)


kentkr:
Just signed the CLA but don't know how to rerun the github check.

core/dbt/exceptions.py
```diff
@@ -2132,12 +2132,25 @@ def __init__(self, yaml_columns, sql_columns):
...
-            f"Schema File Columns: {self.yaml_columns}\n"
-            f"SQL File Columns: {self.sql_columns}"
+            # separate column headers by 20 spaces
+            f"{'yaml col:' : <20} sql col:\n"
         )
+
+        # print columns pretty
+        # split yaml and sql strings into list
+        yaml_list = self.yaml_columns.split(', ')
+        sql_list = self.sql_columns.split(', ')
```
jtcohen6, line 2145: Ah - I didn't realize that these are the stringified versions of `yaml_columns` and `sql_columns`, but that is indeed exactly how they're being passed in here (`string_yaml_columns` and `string_sql_columns`):

https://github.com/dbt-labs/dbt-core/blob/73ff49720021334bd793e55d9d1a198a85ccef5d/core/dbt/include/global_project/macros/materializations/models/table/columns_spec_ddl.sql#L45-L65

It is also true that the logic we're using to compare columns one-by-one, and the logic we'd want to use to display pretty output, should be quite similar.

So, a couple of potential approaches here:
1. Pass the dict (not str) representation of these columns into the Python exception, so that we can do the same sort of comparison over here again
2. Since the Jinja macro is already detecting differences - pass only the _ordered_ set of _disagreeing_ columns, from the `assert_columns_equivalent` macro, into this exception. (But what about the case where the two column lengths differ? We'd still want to detect which columns are present in one, and missing from the other)
3. Move this comparison logic into a Python method, and have `assert_columns_equivalent` macro call that method instead of writing this logic in Jinja. That method would call this exception if it detects differences.

(1) is probably simplest with the existing implementation - just switch out `string_yaml_columns` + `string_sql_columns` for `yaml_columns` and `sql_columns` in the macro definition linked above.

@MichelleArk Curious if you have an opinion on what the right way to go here would be!

jtcohen6 (commented):
@kentkr Thanks for picking this up so quickly!! :)

> But surely it can be inherited as something other than a delimited string.

Surely it can! I don't think this should be too tricky, and it would let us do much nicer things when formatting the error message, to highlight / show only the diffs.

Re: the CLA bot: unfortunately, you might need to do a quick `git rebase` where you edit your previous commits to include the same email address that's associated with your GitHub username. If that's too tricky to sort out, no worries - I can see you've signed the CLA - once the changes here are good to go, one of us can take your commits, rebase, and open as a new PR to make the bot happy. (We'd of course still give you all credit for the contribution!)

core/dbt/exceptions.py
kentkr: @jtcohen6 Alright, what I just pushed should work. It's not my favorite solution because the columns are actually passed as a list of dictionaries `[{col1 : col1 INT}, ...]`. I could've removed the list to avoid using so many for loops, but decided mimic the exception call in the .sql file as best I could. Cuz if it aint broke don't fix it. Attached is an image of how the error prints!
Screenshot 2023-04-02 at 6 42 59 PM

As for the rebase issue, if you're happy with these changes I'll open a new PR with the modified files.

kentkr: @jtcohen6 Alright, what I just pushed should work. It's not my favorite solution because the columns are actually passed as a list of dictionaries `[{col1 : col1 INT}, ...]`. I could've removed the list to avoid using so many for loops, but decided mimic the exception call in the .sql file as best I could. Cuz if it aint broke don't fix it. Attached is an image of how the error prints!
Screenshot 2023-04-02 at 6 42 59 PM

As for the rebase issue, if you're happy with these changes I'll open a new PR with the modified files.

jtcohen6: @kentkr This is a big improvement!

Sorry for the delay getting back to you. To have a better chance of getting this over the finish line, I'm going to open a new PR with
- rebased commits to fix the CLA issue
- a few more aesthetic touch-ups (e.g. highlighting columns with mismatches, and using our existing library for printing markdown tables to stdout)

Issue #6750, opened 2023-01-26
[CT-1918] [Discussion] Unify constraints and constraints_check configs


MichelleArk:
In [the foundational dbt constraints work](https://github.com/dbt-labs/dbt-core/pull/6271),`constraints: List[str]` and `constraints_check: str` are optional column configs that configure column-level constraints and row-level checks respectively.

Let's unify these two configs into a single optional `constraints` (`[List[Union[str, Dict[str, str]]]]`) that encompasses both sets of functionality. Relevant discussion: https://github.com/dbt-labs/dbt-core/pull/6271#discussion_r1084033137

Before: 
```yaml
columns:
  - name: price
    data_type: numeric
    constraints:
      - not null
    constraints_check: price > 0
```

After: 
```yaml
columns:
  - name: price
    data_type: numeric
    constraints:
      - not null
      - check: price > 0
        name: positive_price  # support database-specific check options (Postgres supports naming constraints)
```

jtcohen6:
I definitely think this will be a step in the right direction, by unifying into a single `constraints` attribute.

I'm not 100% convinced that `[List[Union[str, Dict[str, str]]]]` is the perfect data structure either, though it might be close.

- [On Postgres](https://www.postgresql.org/docs/current/ddl-constraints.html), each column can define exactly one `check` constraint, but then the table can define as many additional `check` constraints as it wants (named or unnamed)
- [On Databricks](https://docs.databricks.com/tables/constraints.html#set-a-check-constraint-in-databricks), `check` constraints are named and table-level only. They also support [several additional parameters](https://docs.databricks.com/sql/language-manual/sql-ref-syntax-ddl-alter-table-add-constraint.html), which makes me think that defining each constraint entry as a dictionary is the right move.
- Instead of just accepting strings like "not null," would we want this to be `not_null: true|false`? Would that make it easier or harder for us to compare constraints across project states, in order to detect breaking changes to contracts in CI and require a version bump?

MichelleArk:
> Instead of just accepting strings like "not null," would we want this to be not_null: true|false? Would that make it easier or harder for us to compare constraints across project states, in order to detect breaking changes to contracts in CI and require a version bump?

I'm all for standardizing on a separate `not_null: true|false` (or `nullable: true|false`) config - that would definitely make it more straightforward to detect breaking changes.

jtcohen6:
[...]

PR #7230, opened 2023-03-27
CT-1922: Parsing and postgres implementations for model level constraints

### LINKED ISSUE #6754: Support 'constraints' as a model-level config

peterallenwebb:
resolves #6754

### Description

Adds code to parse and validate constraints on models. Implements default (and postgres) adapter support for model level constraint rendering.

### Checklist


core/dbt/contracts/graph/nodes.py
```diff
@@ -168,6 +168,12 @@ class ColumnLevelConstraint(dbtClassMixin):
     warn_unsupported: bool = (
         True  # Warn if constraint is not supported by the platform and won't be in DDL
     )
+    columns: List[str] = field(default_factory=list)
+
+
+@dataclass
+class ModelLevelConstraint(ColumnLevelConstraint):
+    pass  # REVIEW: The columns property above should be on this class instead
```
peterallenwebb, line 176: For reasons I am still working to understand, moving the columns property above to this class, where it would actually makes sense, results in the columns property being missing when the raw dictionary is passed to the adapter implementation of render_raw_model_constraint().

@gerda It seems like there is a mechanism at work which adds defaults values to and removes unexpected values from the raw dictionaries, but is confused about which type this one has. Can you help me understand what's going on here?

core/dbt/include/global_project/macros/materializations/models/table/create_table_as.sql
```diff
@@ -27,7 +27,7 @@
     {{ relation.include(database=(not temporary), schema=(not temporary)) }}
   {% if config.get('contract', False) %}
     {{ get_assert_columns_equivalent(sql) }}
-    {{ get_columns_spec_ddl() }}
+    {{ get_table_columns_and_constraints() }}
```
mikealfare, line 30: I like this name a lot more. But we'll also need to update that macro in each of the adapters, correct?

core/dbt/contracts/graph/nodes.py
gshank: "constraints" is not processed as a config, it's a node-level field. So you need to actually copy it at the top of the "patch" method on parsed node, where "description" and "columns" and "access" are copied. It's only on model nodes so you'd also want to check the node.resource_type. (At some point we may want to split out more finely grained "patch" methods...)

mikealfare: I couldn't find the answer, but I want to make sure I actually understand the question. It sounds like if `columns` is a key in `raw_constraint`, then `BaseAdapter._parse_model_constraint(raw_constraint)` does not have a `columns` attribute, but `BaseAdapter._parse_column_constraint(raw_constraint)` does have a `columns` attribute. Is that right?

Is `raw_constraint["columns"]` just a `List[str]`? Or is it more complicated?

MichelleArk: >  (At some point we may want to split out more finely grained "patch" methods...)

Thinking along the same lines for patching other new node-level fields that are only on model nodes [in the model version parsing spike (version, is_latest_version)](https://github.com/dbt-labs/dbt-core/pull/7204/files#diff-42cfc99b2fcd89ab75957b8999bee2b1af822f7d39953ca99f1f285c170557c1R396-R399).

core/dbt/include/global_project/macros/materializations/models/table/create_table_as.sql
peterallenwebb: That's right. It's a new macro in 1.5, though, so it should be contained only to our officially supported adapters.

core/dbt/parser/schemas.py
```diff
@@ -936,59 +938,41 @@ def parse_patch(self, block: TargetBlock[NodeTarget], refs: ParserRef) -> None:
...
-                self.constraints_language_validator(patched_node),
-            ]
-            error_messages = [validator for validator in validators if validator != "None"]
-
-        if error_messages:
-            original_file_path = patched_node.original_file_path
-            raise ParsingError(
-                f"Original File Path: ({original_file_path})\nConstraints must be defined in a `yml` schema configuration file like `schema.yml`.\nOnly the SQL table and view materializations are supported for constraints. \n`data_type` values must be defined for all columns and NOT be null or blank.{self.convert_errors_to_string(error_messages)}"
+
+            if isinstance(node, ModelNode) and node.config.contract is True:
```
gshank, line 942: This code works, but right now we're doing node-level field setting in the patch method so ideally this should happen in that method if possible.  I presume that access to "validate_constraints" is the issue? Can that be changed into a function that can be called in both places?

core/dbt/parser/schemas.py
```diff
@@ -936,59 +938,41 @@ def parse_patch(self, block: TargetBlock[NodeTarget], refs: ParserRef) -> None:
...
+                self.validate_constraint_prerequisites(node)
+
+                if any(
+                    c
+                    for c in block.target.constraints
+                    if "type" not in c or not ConstraintType.is_valid(c["type"])
+                ):
+                    raise ParsingError(f"Invalid constraint type on model {block.target.name}")
```
MichelleArk, line 950: Suggestion to make this error message more actionable: 
```python
raise ParsingError(f"Invalid constraint type on model {block.target.name}: type must be one of {[ct.value for ct in ConstraintType]}")
```

core/dbt/parser/schemas.py
gshank: Alternatively, we could pull that "patch" method into where this code is called from. It doesn't need to stay on the node object...

peterallenwebb: Per huddle convo, I've added some comments for now and we'll plan to make this more consistent soon.

core/dbt/parser/schemas.py
peterallenwebb: Done!

gshank (approved):

MichelleArk (approved):

mikealfare (approved):

core/dbt/adapters/base/impl.py
```diff
@@ -1298,6 +1298,44 @@ def render_column_constraint(cls, constraint: ColumnLevelConstraint) -> str:
...
+        try:
+            ModelLevelConstraint.validate(raw_constraint)
+            c = ModelLevelConstraint.from_dict(raw_constraint)
+            return c
+        except Exception:
+            raise DbtValidationError(f"Could not parse constraint: {raw_constraint}")
+
+    @available
+    @classmethod
+    def render_raw_model_constraints(cls, raw_constraints: Dict[str, Any]) -> str:
```
emmyoop, line 1312: Shouldn't the return type be a List here?
```suggestion
    def render_raw_model_constraints(cls, raw_constraints: Dict[str, Any]) -> List:
```

core/dbt/adapters/base/impl.py
```diff
@@ -1298,6 +1298,44 @@ def render_column_constraint(cls, constraint: ColumnLevelConstraint) -> str:
...
+        elif constraint.type == ConstraintType.unique:
+            return f"{constraint_prefix}unique ({column_list})"
+        elif constraint.type == ConstraintType.primary_key:
+            return f"{constraint_prefix}primary key ({column_list})"
+        elif constraint.type == ConstraintType.foreign_key:
+            return f"{constraint_prefix}foreign key ({column_list})"
+        elif constraint.type == ConstraintType.custom and constraint.expression:
+            return f"{constraint_prefix}{constraint.expression}"
+        else:
+            return None
```
VersusFacit, line 1337: This might be completely over pedantic but I'm curious about the fact that some of these functions return an empty string and others a true `None`. Is there some kind of significance here are we just enjoying pythons implicit type conversion?

core/dbt/adapters/base/impl.py
```diff
@@ -1298,6 +1298,44 @@ def render_column_constraint(cls, constraint: ColumnLevelConstraint) -> str:
...
+    def render_raw_model_constraint(cls, raw_constraint: Dict[str, Any]) -> Optional[str]:
+        constraint = cls._parse_model_constraint(raw_constraint)
+        return cls.render_model_constraint(constraint)
+
+    @classmethod
+    def render_model_constraint(cls, constraint: ModelLevelConstraint) -> Optional[str]:
+        """Render the given constraint as DDL text. Should be overriden by adapters which need custom constraint
+        rendering."""
+        constraint_prefix = f"constraint {constraint.name} " if constraint.name else ""
+        column_list = ", ".join(constraint.columns)
```
VersusFacit, line 1325: Haven't looked over all of the adapter PR's, I'm assuming we can trust that these predicates are fairly consistent across warehouses. Are we expecting to add more in the near future or are these a canonical subset that we've chosen to support?

MichelleArk: We're not expecting to add more `ConstraintType`s in the near future - these are the set we've chosen to support based on current warehouse support. More discussion on `ConstraintType` and the `Constraint` here: https://github.com/dbt-labs/dbt-core/issues/6750

This method (`render_model_constraint`), as well as `render_column_constraint` would be overridable in specific adapter implementations, but this base is pretty consistent across implementations thus far.

PR #7239, opened 2023-03-29
ADAP-2: Materialized Views

### LINKED ISSUE #6911: [CT-2068] [Feature] Let's add Materialized View as a materialization, finally

mikealfare:
resolves #6911

### Description

Add support for materialized views.

TODO: add more context as the feature gets built

### Checklist


core/dbt/include/global_project/macros/materializations/models/materialized_view/materialized_view.sql
```diff
@@ -0,0 +1,121 @@
...
+        {% if configuration_changes == {} %}
+            {% set build_sql = refresh_materialized_view(target_relation) %}
+
+        {% elif on_configuration_change == 'apply' %}
+            {% set build_sql = get_alter_materialized_view_as_sql(target_relation, configuration_changes, sql, existing_relation, backup_relation, intermediate_relation) %}
+        {% elif on_configuration_change == 'skip' %}
+            {% set build_sql = '' %}
+            {{ exceptions.warn("Configuration changes were identified and `on_configuration_change` was set to `skip` for `" ~ target_relation ~ "`") }}
+        {% elif on_configuration_change == 'fail' %}
+            {{ exceptions.raise_compiler_error("Configuration changes were identified and `on_configuration_change` was set to `fail` for `" ~ target_relation ~ "`") }}
```
mikealfare, line 76: Raise a different error here. It's not really a compiler error since it results from user input.

plugins/postgres/dbt/adapters/postgres/relation.py
```diff
@@ -20,3 +54,52 @@ def __post_init__(self):
...
+            "context": {
+                "name": "index_abc",  # we only need this to drop, but we need the rest to compare
+                "columns": ["column_1"],
+                "type": "btree",
+                "unique": True
+            }
+        },
+        """
+        # the columns show up as a comma-separated list in the query from postgres
+        for index in indexes:
```
mikealfare, line 94: Add a post_init to do this split

plugins/postgres/dbt/include/postgres/macros/materializations/materialized_view.sql
```diff
@@ -0,0 +1,83 @@
...
+
+{% macro postgres__refresh_materialized_view(relation) %}
+    refresh materialized view {{ relation }};
+{% endmacro %}
+
+
+{% macro postgres__update_indexes_on_materialized_view(relation, index_updates) %}
+    {{ log("Applying UPDATE INDEXES to: " ~ relation) }}
+
+    {% for _index_update in index_updates %}
```
mikealfare, line 66: Make sure these are ordered in such a way that the drops are executed first, so that recreated indexes are created in the right order.

tests/functional/materializations/materialized_view_tests/test_materialized_view.py
```diff
@@ -0,0 +1,99 @@
...
+from tests.adapter.dbt.tests.adapter.materialized_view.base import Base, run_model
+
+from tests.functional.materializations.materialized_view_tests.fixtures import (
+    PostgresOnConfigurationChangeBase,
+)
+
+
+class TestBasic(Base):
+    def test_relation_is_materialized_view_on_initial_creation(self, project):
+        self.assert_relation_is_materialized_view(project)
```
mikealfare, line 11: Pass the relation into the assertion to make it more clear.

mikealfare: Add a test for the table to show that it is not a materialized view.

core/dbt/include/global_project/macros/materializations/models/materialized_view/create_materialized_view.sql
```diff
@@ -0,0 +1,9 @@
+{% macro get_create_materialized_view_as_sql(relation, sql) -%}
+    {{- log('Applying CREATE to: ' ~ relation) -}}
```
mikealfare, line 2: Make this debug logging

tests/functional/materializations/materialized_view_tests/test_materialized_view.py
```diff
@@ -0,0 +1,99 @@
...
+class TestOnConfigurationChangeSkip(OnConfigurationChangeCommon):
+
+    on_configuration_change = "skip"
+
+    def test_model_is_skipped_with_configuration_changes(
+        self, configuration_changes, configuration_change_skip_message
+    ):
+        results, logs = run_model(self.base_materialized_view.name)
+        self.assert_proper_scenario(
+            results, logs, RunStatus.Success, messages_in_logs=[configuration_change_skip_message]
```
mikealfare, line 85: Use RunStatus.Skipped

core/dbt/include/global_project/macros/materializations/models/materialized_view/materialized_view.sql
```diff
@@ -0,0 +1,121 @@
...
+        -- get config options
+        {% set on_configuration_change = config.get('on_configuration_change') %}
+        {% set configuration_changes = get_materialized_view_configuration_changes(existing_relation, config) %}
+
+        {% if configuration_changes == {} %}
+            {% set build_sql = refresh_materialized_view(target_relation) %}
+
+        {% elif on_configuration_change == 'apply' %}
+            {% set build_sql = get_alter_materialized_view_as_sql(target_relation, configuration_changes, sql, existing_relation, backup_relation, intermediate_relation) %}
+        {% elif on_configuration_change == 'skip' %}
```
mikealfare, line 72: Make this a RunStatus.Skipped

core/dbt/include/global_project/macros/materializations/models/materialized_view/create_materialized_view.sql
mikealfare: It looks like this is already the lowest level logging:
https://docs.getdbt.com/reference/dbt-jinja-functions/log

Adding `info=True` would also throw it into stdout.

core/dbt/include/global_project/macros/materializations/models/materialized_view/materialized_view.sql
mikealfare: For now I've replaced this with a `FailFastError`. I think this aligns since the entire run fails upon the first sight of a new config setting.

core/dbt/include/global_project/macros/materializations/models/materialized_view/materialized_view.sql
mikealfare: I cannot figure out how to update the NodeStatus from within the materialization. It appears that the jinja context only has Relation and Config level attributes at run (i.e. we're already "inside" the node). And the runner either captures that an exception was thrown, or not. Seed is able to work because it's a separate command on the CLI, and get's dispatched to its own task and runner. We don't want to do that here since we want this to behave like any other model materialization (e.g. table, view).

plugins/postgres/dbt/include/postgres/macros/materializations/materialized_view.sql
mikealfare: This doesn't happen. The indexes are defined in such a way that if we get an old and a new with the same attribution, we skip it.

core/dbt/contracts/graph/model_config.py
```diff
@@ -287,11 +299,18 @@ def same_contents(cls, unrendered: Dict[str, Any], other: Dict[str, Any]) -> boo
...
     mergebehavior = {
         "append": ["pre-hook", "pre_hook", "post-hook", "post_hook", "tags"],
-        "update": ["quoting", "column_types", "meta", "docs", "contract"],
+        "update": [
+            "quoting",
+            "column_types",
+            "meta",
+            "docs",
+            "contract",
+            "on_configuration_change",
```
mikealfare, line 312: Remove this from this list

core/dbt/include/global_project/macros/adapters/indexes.sql
```diff
@@ -21,3 +21,21 @@
...
   {% endfor %}
 {% endmacro %}
+
+
+{% macro get_drop_index_sql(relation, index_name) -%}
+    {{ adapter.dispatch('get_drop_index_sql', 'dbt')(relation, index_name) }}
+{%- endmacro %}
+
+{% macro default__get_drop_index_sql(relation, index_name) -%}
+    {% do return(None) %}
```
mikealfare, line 31: Update this to throw an exception like the other macros.

core/dbt/include/global_project/macros/materializations/models/materialized_view/get_materialized_view_configuration_changes.sql
```diff
@@ -0,0 +1,23 @@
+{% macro get_materialized_view_configuration_changes(existing_relation, new_config) %}
```
mikealfare, line 1: What do we do with the rest of NodeConfig?

core/dbt/contracts/graph/model_config.py
```diff
@@ -445,6 +464,9 @@ class NodeConfig(NodeAndTestConfig):
     # sometimes getting the Union order wrong, causing serialization failures.
     unique_key: Union[str, List[str], None] = None
     on_schema_change: Optional[str] = "ignore"
+    on_configuration_change: OnConfigurationChangeOption = field(
```
mikealfare, line 467: Make the validation for `on_configuration_change` match that of `on_schema_change`.

plugins/postgres/dbt/include/postgres/macros/materializations/materialized_view.sql
dataders, line 1: nit, but i think convention is that `materialization/materialization.sql` is reserved for `materialization` blocks. and when there's only dispatched macros then the file is called `helpers.sql`? though now I can't find an example

core/dbt/contracts/graph/model_config.py
```diff
@@ -189,6 +191,16 @@ class Severity(str):
 register_pattern(Severity, insensitive_patterns("warn", "error"))
 
 
+class OnConfigurationChangeOption(StrEnum):
+    apply = "apply"
+    skip = "skip"
```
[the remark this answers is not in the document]
Fleid: Continue I like!
so `skip ="contine"` would be my first pick, but I understand if you'd rather have `continue="continue"`

core/dbt/contracts/graph/model_config.py
mikealfare: `on_schema_change` does not appear to be controlled anywhere besides in jinja. the valid values are enforced in jinja. So this is effectively an Enum. I've removed the meta entry on `on_configuration_change` to match.

core/dbt/include/global_project/macros/materializations/models/materialized_view/get_materialized_view_configuration_changes.sql
mikealfare: We don't actually track anything else at the relation level for materialized views in Postgres. It looks pretty close to table. We have not implemented contracts or constraints yet for materialized views, otherwise it's the same as table. However, I did refactor how I'm determining the differences so that it can be more extensible for the other adapters, or when we add functionality to materialized views.

core/dbt/include/global_project/macros/materializations/models/materialized_view/materialized_view.sql
mikealfare: We've decided to leave this is a RunStatus.Success, since the node did finish running and should not impede downstream models from running. However, we updated the `on_configuration_change` setting to be "continue" to avoid confusion.

plugins/postgres/dbt/include/postgres/macros/materializations/materialized_view.sql
mikealfare: I just parked this next to `incremental_strategies.sql` and `snapshot_merge.sql` since that looked the most similar. I'm open to moving it, and I'm sure it's in a slightly different spot in the other adapters. But I don't know that something as generic as `helpers.sql` is appropriate. These are all macros that are called directly from the materialization block, similar to `create_table_as_sql`. I just have more macros since we're attempting to alter and refresh the object. So considering that, I'd also like to keep them in their own file if possible, to communicate how they're related.

plugins/postgres/dbt/adapters/postgres/database_configs/index.py
```diff
@@ -0,0 +1,180 @@
...
+        name: Optional[str] = None,
+    ) -> "IndexConfig":
+        kwargs = {
+            "unique": unique,
+            "method": method,
+            "name": name,
+        }
+
+        if isinstance(column_names, set) and len(column_names) > 0:
+            # TODO: do we need to look at the quoting policy?
```
mikealfare, line 106: We need to look at the quote policy.

plugins/postgres/dbt/adapters/postgres/database_configs/materialized_view.py
```diff
@@ -0,0 +1,133 @@
...
+    query: Optional[str] = None
+    indexes: Set[IndexConfig] = field(default_factory=set)
+
+    @classmethod
+    def from_model_node(cls, model_node: ModelNode) -> "MaterializedViewConfig":
+        node_config = model_node.config
+
+        kwargs = {}
+
+        table_name = model_node.identifier
```
mikealfare, line 39: Follow the include policy for `table_name`.

core/dbt/contracts/relation.py
```diff
@@ -117,3 +122,22 @@ def replace_dict(self, dct: Dict[ComponentName, str]):
...
+
+
+"""
+These classes are used to define an interface for updates on objects in place.
+This originated with materialized views, which attempt to update things like indexes without dropping the underlying
+materialized view.
+"""
+
+
+class ChangeAction(StrEnum):
```
mikealfare, line 134: Move these to `adapters/base/relation.py`.

core/dbt/include/global_project/macros/materializations/models/materialized_view/get_materialized_view_configuration_changes.sql
```diff
@@ -0,0 +1,23 @@
...
+    /* {#
+    It's recommended that configuration changes be formatted as follows:
```
mikealfare, line 3: Update these comments to reflect `ChangeAction` and `Change`

tests/functional/materializations/materialized_view_tests/test_materialized_view.py
```diff
@@ -0,0 +1,143 @@
...
+            logs,
+            RunStatus.Success,
+            messages_in_logs=[configuration_change_continue_message],
+            messages_not_in_logs=[refresh_message],
+        )
+
+
+class TestOnConfigurationChangeFail(OnConfigurationChangeCommon):
+
+    on_configuration_change = OnConfigurationChangeOption.Fail
```
MichelleArk, line 131: nit: inlining `on_configuration_change` in the `assert_proper_scenario` call below here + in other tests that inherit from `OnConfigurationChangeCommon` (similar to RunStatus values) would be more readable

plugins/postgres/dbt/include/postgres/macros/materializations/materialized_view.sql
```diff
@@ -0,0 +1,84 @@
...
+
+    -- apply a full refresh immediately if needed
+    {% if configuration_changes.requires_full_refresh %}
+
+        {{ get_replace_materialized_view_as_sql(relation, sql, existing_relation, backup_relation, intermediate_relation) }}
+
+    -- otherwise apply individual changes as needed
+    {% else %}
+
+        {{ postgres__update_indexes_on_materialized_view(relation, configuration_changes.indexes) }}
```
MichelleArk, line 18: idea here being - additional updates to other configuration can occur sequentially in this `else` block?

e.g: 
```
{{ postgres__update_indexes_on_materialized_view(relation, configuration_changes.indexes) }}
{{ postgres__update_<something-else>_on_materialized_view(relation, configuration_changes.something_else) }}
```

core/dbt/contracts/relation.py
```diff
@@ -17,9 +17,13 @@ class RelationType(StrEnum):
     Table = "table"
     View = "view"
     CTE = "cte"
-    MaterializedView = "materializedview"
+    MaterializedView = "materialized_view"
     External = "external"
 
+    @classmethod
+    def default(cls) -> "RelationType":
```
MichelleArk, line 24: is this used anywhere in this PR? I can see merits to this approach of defining a default on the enum type itself - but I could also see it becoming confusing if the sensible default is based on the calling context

plugins/postgres/dbt/adapters/postgres/relation_configs/materialized_view.py
```diff
@@ -0,0 +1,129 @@
...
+    table_name: Optional[str] = None
+    query: Optional[str] = None
+    indexes: Set[PostgresIndexConfig] = field(default_factory=set)
+
+    @property
+    def validation_rules(self) -> Set[RelationConfigValidationRule]:
+        # index rules get run by default with the mixin
+        return {
+            RelationConfigValidationRule(
+                validation_check=self.table_name is None or len(self.table_name) <= 63,
```
MichelleArk, line 45: nit - 63 looks like magic number and should probably be refactored as a constant defined either as a class attribute or somewhere more appropraite

plugins/postgres/dbt/adapters/postgres/relation_configs/index.py
```diff
@@ -0,0 +1,174 @@
...
+            "name": config_dict.get("name"),
+            "method": config_dict.get("method"),
+            "unique": config_dict.get("unique"),
+            "column_names": frozenset(column for column in config_dict.get("column_names", {})),
+        }
+        index: "PostgresIndexConfig" = super().from_dict(kwargs_dict)  # type: ignore
+        return index
+
+    @classmethod
+    def parse_model_node(cls, model_node_entry: ModelNodeEntry) -> dict:
```
MichelleArk, line 80: looks like this method and `parse_relation_results` are both returning intermediate dictionaries that eventually get passed to `from_dict`. What do you think about refactoring them both to be `from_model_node` and `from_relation_results` respectively - each calling `from_dict` under the hood. That way the caller has less to worry about when initializing a `PostgresIndexConfig`.

plugins/postgres/dbt/adapters/postgres/relation_configs/index.py
```diff
@@ -0,0 +1,174 @@
...
+        config_dict = {
+            "unique": model_node_entry.get("unique"),
+            "method": model_node_entry.get("type"),
+        }
+
+        if column_names := model_node_entry.get("columns", []):
+            # TODO: include the QuotePolicy instead of defaulting to lower()
```
MichelleArk, line 87: is this an outstanding TODO for this PR or is there a follow-up issue?

plugins/postgres/dbt/adapters/postgres/relation_configs/index.py
MichelleArk: hmm, looks like something similar has been done for `PostgresMaterializedViewConfig`.

plugins/postgres/dbt/include/postgres/macros/materializations/materialized_view.sql
mikealfare: That's the idea. It would be one or more changes.

core/dbt/contracts/relation.py
mikealfare: No, it's not used explicitly, but it's the default that I generally see. Yes, it could depend on the calling context, but then the calling context would set it explicitly instead of defaulting to the Enum. I'm fine pulling this back out. It was just something I noticed and something I was doing on my Enums.

plugins/postgres/dbt/adapters/postgres/relation_configs/index.py
mikealfare: I did that so that `MaterializedView.parse_model_node()` returns a standard dictionary, and not a dictionary with `RelationConfig` objects. More specifically, I did that so that `MaterializedView.from_dict()` can accept a standard dictionary, in case there are other things that want to use this that don't want to build the `IndexConfig` objects manually before submitting them to `MaterializedView.from_dict()`. Ultimately this might all be replaced with something like `mashumaro` or `pydantic` under the hood, so I want `from_dict()` to look as close as possible to methods in those frameworks. I could build the `from_model_node()` and `from_relation_results()` methods for `IndexConig`; I just didn't need them, so I didn't build them.

plugins/postgres/dbt/adapters/postgres/relation_configs/index.py
mikealfare: Tomato tomahto. I will do this prior to release, but I think it can go into the beta as is. Ultimately that's your call.

plugins/postgres/dbt/adapters/postgres/database_configs/index.py
mikealfare: Added to cleanup items story.

plugins/postgres/dbt/adapters/postgres/database_configs/materialized_view.py
mikealfare: Added to cleanup items story.

plugins/postgres/dbt/adapters/postgres/relation_configs/index.py
MichelleArk: Cool -- as long as its tracked for prior to release 👍

plugins/postgres/dbt/adapters/postgres/relation_configs/materialized_view.py
mikealfare: Done

tests/functional/materializations/materialized_view_tests/test_materialized_view.py
mikealfare: Yeah, I can see that. When I started the framework I reference this attribute often, making ideal for a class attribute. But I only reference in one spot now, so it's really not saving me a lot of work. I'll update it.

plugins/postgres/dbt/adapters/postgres/relation_configs/materialized_view.py
```diff
@@ -0,0 +1,129 @@
...
+        return config_dict
+
+
+@dataclass
+class PostgresMaterializedViewConfigChangeCollection:
+    indexes: Optional[Set[PostgresIndexConfigChange]] = None
+
+    @property
+    def requires_full_refresh(self) -> bool:
+        return any(index.requires_full_refresh for index in self.indexes)
```
MichelleArk, line 123: this probably needs the same wrapping of `if isinstance(self.indexes, Iterable)` that `has_changes` has because `indexes` is optional

core/dbt/adapters/relation_configs/config_change.py
```diff
@@ -0,0 +1,27 @@
...
+class RelationConfigChangeAction(StrEnum):
+    alter = "alter"
+    create = "create"
+    drop = "drop"
+
+
+@dataclass(frozen=True, eq=True, unsafe_hash=True)
+class RelationConfigChange(RelationConfigBase, ABC):
+    action: RelationConfigChangeAction
+    context: Hashable  # this is usually a RelationConfig, e.g. IndexConfig, but shouldn't be limited
```
MichelleArk, line 18: should those both be marked as optional? `is_change` checks for `None` values for both

plugins/postgres/dbt/adapters/postgres/relation.py
```diff
@@ -20,3 +34,57 @@ def __post_init__(self):
...
+        config_change_collection = PostgresMaterializedViewConfigChangeCollection()
+
+        existing_materialized_view = PostgresMaterializedViewConfig.from_relation_results(
+            relation_results
+        )
+        new_materialized_view = PostgresMaterializedViewConfig.from_model_node(
+            runtime_config.model
+        )
+
+        if index_config_changes := self._get_index_config_changes(
```
MichelleArk, line 55: ooooh we can use the walrus operator now because of python 3.8+ support! :=)

tests/functional/materializations/materialized_view_tests/test_materialized_view.py
mikealfare: Done

plugins/postgres/dbt/adapters/postgres/relation_configs/materialized_view.py
mikealfare: Done

core/dbt/adapters/relation_configs/config_change.py
mikealfare: Neither is optional actually. A change needs both. I'll look at what I did with `is_change` and update that then.

plugins/postgres/dbt/adapters/postgres/relation.py
```diff
@@ -20,3 +34,57 @@ def __post_init__(self):
...
+            existing_materialized_view.indexes, new_materialized_view.indexes
+        ):
+            config_change_collection.indexes = index_config_changes
+
+        if config_change_collection.has_changes:
+            return config_change_collection
```
MichelleArk, line 61: if there is _any_ change, we return the entire collection, right? Is this what downstream callers expect? Or do they assume that all config changes in a non-null PostgresMaterializedViewConfigChangeCollection from this method are changes

core/dbt/adapters/relation_configs/config_change.py
mikealfare: I changed it to `and` instead of `or`. Ideally every instance of `RelationConfigChange` is a change, but I'm allowing for a small degree of sloppiness I guess. In other words, I can pass in a `context` of `None` and let the object determine that's not a change, instead of checking for `None` when creating the instance.

plugins/postgres/dbt/adapters/postgres/relation.py
mikealfare: Yup, and I sprinkled it everywhere. We do that operation really often, so I'm excited now.

plugins/postgres/dbt/adapters/postgres/relation.py
mikealfare: Yes, we return the entire collection either way, because we might need to inspect the collection to know what to do. For example, if there are changes that don't require a full refresh, and changes that do, I don't want to apply changes and then refresh anyway. And all changes should be changes, if I understand that correctly.

core/dbt/contracts/relation.py
mikealfare: I removed this from `RelationType` but kept it in the relation configs since those model the database and its constraints.

tests/adapter/dbt/tests/adapter/materialized_view/on_configuration_change.py
```diff
@@ -0,0 +1,91 @@
...
+):
+    assert len(results.results) == 1
+    result = results.results[0]
+
+    assert result.node.config.on_configuration_change == on_configuration_change
+    assert result.status == status
+    for message in messages_in_logs or []:
+        assert_message_in_logs(logs, message)
+    for message in messages_not_in_logs or []:
+        assert_message_in_logs(logs, message, expected_fail=True)
```
MichelleArk, line 49: discussed async - we're primarily testing that the expected scenarios occurred by inspecting logs (dbt logs) which can be brittle if we want to change/remove logs or error-prone if we change functionality but keep logs as-is. Let's see if we can instead assert the expected sql statements were sent to the db for execution.

mikealfare: Agreed, it's on the list of follow up items to take care of ahead of the full release.

MichelleArk (approved):
Great work, this is a really huge lift and a solid start 💪 I'm excited to see this new pattern for applying configuration changes more precisely and cost-effectively to materialized views.  The materialization macro itself was also structured in a way that was super readable which made this much easier to review, so thank you! ✨ 

We've identified a few pieces as follow-on work but nothing blocking from my end. I'm looking forward to get this in beta for more hands-on testing and hardening! 🎉

mikealfare:
> Great work, this is a really huge lift and a solid start 💪 I'm excited to see this new pattern for applying configuration changes more precisely and cost-effectively to materialized views. The materialization macro itself was also structured in a way that was super readable which made this much easier to review, so thank you! ✨
> 
> We've identified a few pieces as follow-on work but nothing blocking from my end. I'm looking forward to get this in beta for more hands-on testing and hardening! 🎉

Thanks for the compliments! And I want to make sure @VersusFacit gets credit for the materialization structure. It was her idea to break it up and I agree it's much easier to digest. Some of the pieces might even be reusable across materializations as they are somewhat generalized.

PR #7250, opened 2023-03-31
Generalize constraint warnings

### LINKED ISSUE #7067: [CT-2199] Generalize constraint compatibility warnings

emmyoop:
resolves #7067 

### Description

Centralizes warnings so adapters just need to set the `constraint_support` property.

### Checklist


core/dbt/include/global_project/macros/materializations/models/table/columns_spec_ddl.sql
```diff
@@ -9,13 +9,24 @@
...
-    {% for i in user_provided_columns %}
+    {% for i in user_provided_columns -%}
       {%- set col = user_provided_columns[i] -%}
       {%- set constraints = col['constraints'] -%}
-      {{ col['name'] }} {{ col['data_type'] }}{% for c in constraints %} {{ adapter.render_raw_column_constraint(c) }}{% endfor %}{{ "," if not loop.last }}
-    {% endfor -%}
+      {{ col['name'] }} {{ col['data_type'] }}
+      {% for c in constraints -%}
+        {% do all_constraints.append(c) %}
+        {%- if c["type"] not in unsupported_contraints -%}
```
MichelleArk, line 21: I'm thinking we can move a lot of this filtering logic to the python base implementation, potentially within the existing `adapter.render_raw_column_constraint` method. Motivation being - this filtering / validation logic should be consistent across adapters (with flexibility in configuring _which_ constraints are supported/enforced). There's only one correct way to render this DDL for a given adapter - so any overrides should ideally be implemented by the adapter-maintainer in python, as opposed to by end users in "user-space" (jinja).

plugins/postgres/dbt/adapters/postgres/impl.py
```diff
@@ -131,3 +131,13 @@ def valid_incremental_strategies(self):
         Not used to validate custom strategies defined by end users.
         """
         return ["append", "delete+insert"]
+
+    @property
+    def constraint_support(self):
+        return {
+            "not_null": ConstraintSupport.ENFORCED,
```
MichelleArk, line 138: we could use the [`ConstraintType` enum](https://github.com/dbt-labs/dbt-core/blob/main/core/dbt/contracts/graph/nodes.py#L143) for keys in this dictionary instead of strings.

core/dbt/adapters/base/impl.py
```diff
@@ -1298,6 +1301,57 @@ def render_column_constraint(cls, constraint: ColumnLevelConstraint) -> str:
         else:
             return ""
 
+    @available
+    def render_ddl(self, columns: Dict[str, Dict]) -> List:
```
MichelleArk, line 1305: this name feels too generic - a lot of what dbt does is render ddl. how about `render_column_constraint_ddl`?

core/dbt/adapters/base/impl.py
```diff
@@ -1298,6 +1301,57 @@ def render_column_constraint(cls, constraint: ColumnLevelConstraint) -> str:
...
+        ddl = []
```
MichelleArk, line 1306: similar naming suggestion: `rendered_column_constraints` would be more precise here.

core/dbt/adapters/base/impl.py
```diff
@@ -1298,6 +1301,57 @@ def render_column_constraint(cls, constraint: ColumnLevelConstraint) -> str:
...
+        ]
+
+        if not_supported:
+            warn_or_error(ConstraintNotSupported(constraints=not_supported))
+        if not_enforced:
+            warn_or_error(ConstraintNotEnforced(constraints=not_enforced))
+
+    @staticmethod
+    def constraint_support() -> Dict:
+        # TODO: should this instead default all to NOT_ENFORCED
```
MichelleArk, line 1352: In the issue the [recommendation](https://github.com/dbt-labs/dbt-core/issues/7067#issuecomment-1447281379) was to default to: 
```
{
"not_null": ConstraintSupport.ENFORCED,
 "primary_key": ConstraintSupport.NOT_ENFORCED,
 "foreign_key": ConstraintSupport.NOT_ENFORCED,
 "unique": ConstraintSupport.NOT_ENFORCED,
 "check": ConstraintSupport.NOT_SUPPORTED,
}
```

core/dbt/adapters/base/impl.py
```diff
@@ -1298,6 +1301,62 @@ def render_column_constraint(cls, constraint: ColumnLevelConstraint) -> str:
...
+        for k, v in columns.items():
+            column_ddl = [f"{v['name']} {v['data_type']}"]
+            for con in v["constraints"]:
+                all_constraints.append(con)
+                if con["type"] not in unsupported_contraints:
+                    constraint = self._parse_column_constraint(con)
+                    column_ddl.append(self.render_column_constraint(constraint))
+            rendered_column_constraints.append(" ".join(column_ddl))
+
+        self.process_constraint_warnings(constraints=all_constraints)
```
MichelleArk, line 1320: This may just come down to personal preference, but I find this method a bit difficult to read / duplicative because its helpers are also iterating over constraints to extract their type / support before performing some sort of action (filtering, warning, etc). I think much of it could be simplified if we inlined more here. For example, 

```python
rendered_column_constraints = []

# maybe this should be a constant class attribute instead of a static method, ie self.CONSTRAINT_SUPPORT
constraint_type_to_support = self.constraint_support()
constraints_by_type = {cs: [] for cs in ConstraintSupport}

for k, v in columns.items():
    rendered_column_constraint = [f"{v['name']} {v['data_type']}"]
    for con in v["constraints"]:
        constraint = self._parse_column_constraint(con) # parsing the raw column constraint earlier so we have a ColumnConstraint object to work with
        constraints_by_type[constraint.type].append(constraint)
        if constraint_type_to_support[con.type] != ConstraintSupport.NOT_SUPPORTED:
           rendered_column_constraint.append(self.render_column_constraint(constraint))
    rendered_column_constraints.append(" ".join(rendered_column_constraint))

self.raise_constraint_warnings(constraints_by_type) # this probably ends up being pretty simple, could consider inlining it as well. 
return rendered_column_constraints
```

emmyoop: I took some of this into account.  But I discovered we can't bundle the error messages per table since the warn bools can be set per constraint.

core/dbt/events/types.py
```diff
@@ -746,6 +746,30 @@ def message(self) -> str:
...
+        return msg
+
+
+class ConstraintNotSupported(WarnLevel):
+    def code(self):
+        return "E049"
+
+    def message(self) -> str:
+        msg = (
+            f"We noticed you have `{ self.constraint }` configs, these are NOT supported "
```
MichelleArk, line 767: I think these messages were carried over from the initial implementation of model contracts. @dbeatty10 - curious to hear if you have any feedback on them!

core/dbt/adapters/base/impl.py
```diff
@@ -1298,6 +1310,79 @@ def render_column_constraint(cls, constraint: ColumnLevelConstraint) -> str:
...
+                c = cls.process_parsed_constraint(constraint, cls.render_raw_column_constraint)
+                if c is not None:
+                    rendered_column_constraint.append(c)
+            rendered_column_constraints.append(" ".join(rendered_column_constraint))
+
+        return rendered_column_constraints
+
+    @classmethod
+    def process_parsed_constraint(
+        cls, parsed_constraint: Union[ColumnLevelConstraint, ModelLevelConstraint], render_func
```
peterallenwebb, line 1331: This implementation makes sense to me if we want to tie column-level and model-level support together, but I question whether we do. For example, Spark supports check constraints, but only at the model/table level, and of course not-null only makes sense at the column level anyway. What do you think @MichelleArk ?

core/dbt/adapters/base/impl.py
```diff
@@ -1298,6 +1310,79 @@ def render_column_constraint(cls, constraint: ColumnLevelConstraint) -> str:
         else:
             return ""
 
+    @available
+    @classmethod
+    def render_raw_columns_constraints(cls, raw_columns: Dict[str, Any]) -> List:
```
peterallenwebb, line 1315: I suggest we be more specific about the types here and annotate as:

```python
def render_raw_columns_constraints(cls, raw_columns: Dict[str, Dict[str, Any]]) -> List:
```

peterallenwebb (commented):
Overall this is looking really good, just had one final question and one final suggestion.

core/dbt/events/types.py
jtcohen6: The first-person plural here, and the capitalized "NOT," feel out of keeping with our style in other warnings/error messages.

How about something like (total untested pseudo-code):
```python
# ConstraintNotEnforced
description = (
    f"The constraint type {constraint.type.value} is not enforced by {self.adapter_type}. "
    "The constraint will be included in this model's DDL statement, but it will not guarantee anything about the underlying data. "
    "Set 'warn_unenforced: false' on this constraint to ignore this warning."
)
msg = (line_wrap_message(warning_tag(description))
return msg
```

```python
# ConstraintNotSupported
description = (
    f"The constraint type {constraint.type.value} is not supported by {self.adapter_type}, and will be ignored. "
    "Set 'warn_unsupported: false' on this constraint to ignore this warning."
)
msg = (line_wrap_message(warning_tag(description))
return msg
```
That would require adding `{adapter.type()}` to these event definitions & call sites

core/dbt/adapters/base/impl.py
jtcohen6: I wouldn't want to be too strict about this. As a general rule, on most data platforms, every constraint should be supported at both the model-level (multiple columns) and the single column-level. Of all the constraint types, `not_null` is the only one that is supported _only_ for a single column, and so it only really makes sense to define it at the column-level. IMO the funkiness of `check` on Spark/Databricks is more of a detail of their current implementation, which they're working to evolve/improve.

jtcohen6: To summarize - I think it is totally legit to use this same method for both column-level and model-level constraints

core/dbt/include/global_project/macros/materializations/models/table/columns_spec_ddl.sql
```diff
@@ -1,19 +1,21 @@
...
-    {%- set user_provided_columns = model['columns'] -%}
+    {%- set raw_column_constraints = adapter.render_raw_columns_constraints(raw_columns=model['columns']) -%}
+    {%- set raw_model_constraints = adapter.render_raw_model_constraints(raw_constraints=model['constraints']) -%}
     (
-    {% for i in user_provided_columns %}
-      {%- set col = user_provided_columns[i] -%}
-      {%- set constraints = col['constraints'] -%}
-      {{ col['name'] }} {{ col['data_type'] }}{% for c in constraints %} {{ adapter.render_raw_column_constraint(c) }}{% endfor %}{{ "," if not loop.last }}
+    {% for c in raw_column_constraints -%}
+      {{ c }}{{ "," if not loop.last or raw_model_constraints }}
```
VersusFacit, line 15: Can we trust Jinja to reliably treat `Null` as `false`? Can there be weird edge cases like whitespace messing up logic things?

core/dbt/events/types.py
```diff
@@ -748,6 +748,32 @@ def message(self) -> str:
...
+
+
+class ConstraintNotSupported(WarnLevel):
+    def code(self):
+        return "E049"
+
+    def message(self) -> str:
+        msg = (
+            f"The constraint type {self.constraint} is not supported by {self.adapter}, and will "
+            "be ignored. Set 'warn_unsupported: false' on this constraint to ignore this warning."
```
VersusFacit, line 772: Why a warning and not a compiler error?

core/dbt/adapters/base/impl.py
```diff
@@ -1298,6 +1310,83 @@ def render_column_constraint(cls, constraint: ColumnLevelConstraint) -> str:
...
+        elif constraint.type == ConstraintType.unique:
+            return f"{constraint_prefix}unique ({column_list})"
+        elif constraint.type == ConstraintType.primary_key:
+            return f"{constraint_prefix}primary key ({column_list})"
+        elif constraint.type == ConstraintType.foreign_key:
+            return f"{constraint_prefix}foreign key ({column_list})"
+        elif constraint.type == ConstraintType.custom and constraint.expression:
+            return f"{constraint_prefix}{constraint.expression}"
+        else:
+            return None
```
VersusFacit, line 1388: I noticed that sometimes these functions return an option while other times these functions return an empty string – is there a meaningful intent there? It would be great to have semantic consistency around that syntax if nothing else to avoid the appearance of arbitrary code

core/dbt/adapters/base/impl.py
```diff
@@ -1298,6 +1310,83 @@ def render_column_constraint(cls, constraint: ColumnLevelConstraint) -> str:
...
+
+    @classmethod
+    def render_model_constraint(cls, constraint: ModelLevelConstraint) -> Optional[str]:
+        """Render the given constraint as DDL text. Should be overriden by adapters which need custom constraint
+        rendering."""
+        constraint_prefix = f"constraint {constraint.name} " if constraint.name else ""
+        column_list = ", ".join(constraint.columns)
+        if constraint.type == ConstraintType.check and constraint.expression:
+            return f"{constraint_prefix}check {constraint.expression}"
+        elif constraint.type == ConstraintType.unique:
```
VersusFacit, line 1379: I'm guessing no but should there be a `column_list` is not `None` check here?

VersusFacit (commented):
Just a few comments but this one looks largely good. Also, I don't know if we have a set style in mind, but in few places we do

core/dbt/events/types.py
emmyoop: By design.  @MichelleArk may be able to speak more directly to why that decision was made.

core/dbt/adapters/base/impl.py
emmyoop: Good callout, I'll make the 2 methods return None

core/dbt/adapters/base/impl.py
emmyoop: It's not possible.  We only make it here for a constraint that is defined in `CONSTRAINT_SUPPORT` and is either supported` or `not_enforced` (but not `not_supported`)

core/dbt/include/global_project/macros/materializations/models/table/columns_spec_ddl.sql
VersusFacit: We don't have a standard, and give him some ginger I actually brought this morning, I do think that `None` (whatever name they've changed it to for this package) is falsey. 

For context, I was especially concerned about an empty str and being unclear on how it's truthiness would be evaluated, but we've eliminated that as a possibility with your recent changes so I think this is less of a concern than it was for me before

tests/functional/dbt_runner/test_dbt_runner.py
```diff
@@ -12,20 +13,20 @@ def dbt(self) -> dbtRunner:
         return dbtRunner()
 
     def test_group_invalid_option(self, dbt: dbtRunner) -> None:
```
mikealfare, line 15: Minor style/readability suggestion. You could use `pytest.mark.parameterize` here to group these test cases, basically saying "all of these scenarios result in `DbtUsageException`:
```python
@pytest.mark.parameterize("invoke_args",[
    (["--invalid-option"]),
    (["deps", "--invalid-option"]),
    (["--warn-error", "--warn-error-options", '{"include": "all"}', "deps"]),
    (["--invalid-command"]),
])
def test_commands_raise_usage_exception(self, dbt: dbtRunner, invoke_args: list) -> None:
    res = dbt.invoke(invoke_args)
    assert type(res.exception) == DbtUsageException
```

tests/unit/test_base_context.py
```diff
@@ -0,0 +1,22 @@
+import os
+
+from dbt.context.base import BaseContext
+from jinja2.runtime import Undefined
+
+
+class TestBaseContext:
+    def test_log_jinja_undefined(self):
```
mikealfare, line 8: These can also be combined with `@pytest.mark.parameterize()`, though the exception message might be a bit long to do so.

tests/unit/test_functions.py
```diff
@@ -43,3 +43,19 @@ def test_warn_or_error_warn_error(warn_error, expect_compilation_exception):
...
+    class BadEvent(InfoLevel):
+        """A spoof Note event which breaks dictification"""
+
+        def __init__(self):
+            self.__class__.__name__ = "Note"
+
+    event = BadEvent()
+    try:
+        msg_to_dict(event)
+    except Exception as exc:
```
mikealfare, line 58: Same as above, should this be a more specific version of `Exception`?

core/dbt/adapters/base/impl.py
```diff
@@ -1273,14 +1291,8 @@ def _parse_column_constraint(cls, raw_constraint: Dict[str, Any]) -> ColumnLevel
         except Exception:
             raise DbtValidationError(f"Could not parse constraint: {raw_constraint}")
```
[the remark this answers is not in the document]
emmyoop: We were just looking at a single constraint int eh macro but we decided to move all the logic into Python so that we pass in the entire column.  The same logic is handled, just differently.

core/dbt/contracts/graph/nodes.py
```diff
@@ -175,6 +175,16 @@ class ModelLevelConstraint(ColumnLevelConstraint):
     columns: List[str] = field(default_factory=list)
 
 
+@dataclass
+class ModelLevelConstraint(ColumnLevelConstraint):
+    columns: List[str] = field(default_factory=list)
+
+
+@dataclass
```
mikealfare, line 183: Unless I'm missing something, this looks like a duplicate.

mikealfare (approved):

PR #7253, opened 2023-03-31
CT 2346 fix read and check versions

### LINKED ISSUE #7252: [CT-2346] Fix incorrect handling of run_results.json and sources.json in read_and_check_versions
In the "read_and_check_versions" method in the VersionedSchema class in. ore/dbt/contracts/util.py there is some code that only applies to manifests, but the method is also called for sources.json and run_results.json

gshank:
resolves #7252

### Description

The code to fix older versions of manifest.json was also being called for sources.json and run_results.json, which is incorrect. Refactor to only call it for manifests.

### Checklist

### DIFF
--- a/core/dbt/contracts/graph/manifest.py
+++ b/core/dbt/contracts/graph/manifest.py
@@ -1231,10 +1231,11 @@ class WritableManifest(ArtifactMixin):
 
     @classmethod
     def upgrade_schema_version(cls, data):
-        # This is specific to manifests, so call this from ArtifactMixin code
+        """This overrides the "upgrade_schema_version" call in VersionedSchema (via
+        ArtifactMixin) to modify the dictionary passed in from earlier versions of the manifest."""
         if get_manifest_schema_version(data) <= 8:
             data = upgrade_manifest_json(data)
-        return cls.from_dict(data)  # type: ignore
+        return cls.from_dict(data)
 
     def __post_serialize__(self, dct):
         for unique_id, node in dct["nodes"].items():
--- a/core/dbt/contracts/graph/manifest.py
+++ b/core/dbt/contracts/graph/manifest.py
@@ -37,6 +37,7 @@ from dbt.contracts.graph.nodes import (
     BaseNode,
 )
 from dbt.contracts.graph.unparsed import SourcePatch
+from dbt.contracts.graph.manifest_upgrade import upgrade_manifest_json
 from dbt.contracts.files import SourceFile, SchemaSourceFile, FileHash, AnySourceFile
 from dbt.contracts.util import BaseArtifactMetadata, SourceKey, ArtifactMixin, schema_version
 from dbt.dataclass_schema import dbtClassMixin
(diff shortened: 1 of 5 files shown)


core/dbt/contracts/graph/manifest.py
```diff
@@ -1228,13 +1229,27 @@ def compatible_previous_versions(self):
             ("manifest", 8),
         ]
 
+    @classmethod
+    def upgrade_schema_version(cls, data):
```
QMalcolm, line 1233: Hmm I got turned around because the function's name led me to believe that the schema's version is upgraded, but I don't think of `from_dict` alone as an upgrading action. Is it generally expected that the verb "upgrade" for dbt-core functions means "change if changes are needed"? Additionally does `from_dict` actually do an upgrade of some sort?

core/dbt/contracts/util.py
```diff
@@ -342,9 +226,12 @@ def read_and_check_versions(cls, path: str):
...
                     )
-        if get_manifest_schema_version(data) <= 8:
-            data = upgrade_manifest_json(data)
-        return cls.from_dict(data)  # type: ignore
+
+        return cls.upgrade_schema_version(data)
+
+    @classmethod
+    def upgrade_schema_version(cls, data):
+        return cls.from_dict(data)
```
QMalcolm, line 234: Somewhat of a repeat of my other comment. Does `from_dict` actually do an upgrade?

QMalcolm (commented):
I love that we're separating things out to apply only to their relevant parts! I'm slightly confused about if `from_dict` is a upgrading action though? It seems like we're treating it as an upgrading action in `upgrade_schema_version` of both `VersionedSchema` and `WritableManifest`. I couldn't find a ton of detail `from_dict` but it looks like it simply instantiates an instance of the referenced class using the relevant keys from the provided dict as the necessary attributes.

core/dbt/contracts/graph/manifest.py
gshank: The part that does the "upgrade" is "upgrade_manifest_json". That code implements any changes that are necessary to have an older manifest schema version loaded by the current manifest code.  "from_dict" is used all over the codebase and is implemented by mashumaro, which is pulled in by dbtClassMixin. It just instantiates an object from a dictionary.

core/dbt/contracts/graph/manifest.py
```diff
@@ -1228,13 +1229,27 @@ def compatible_previous_versions(self):
...
+        # This is specific to manifests, so call this from ArtifactMixin code
+        if get_manifest_schema_version(data) <= 8:
+            data = upgrade_manifest_json(data)
+        return cls.from_dict(data)  # type: ignore
```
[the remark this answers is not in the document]
gshank: That was just left over from the code that was moved out of VersionedSchema which could return a number of different classes. I can check to see if it can be removed.

core/dbt/contracts/util.py
gshank: No. This is just the "base" implementation of "upgrade_schema_version" which is to do nothing to the schema version, but does return an object from the dictionary that was loaded from the json.  It's only the manifest version of this method which actually does an "upgrade".

core/dbt/contracts/graph/manifest.py
QMalcolm: Gotcha! Thank you for walking me through that 🙂

core/dbt/contracts/util.py
QMalcolm: Ah gotcha! So this method exists because a default implementation of "upgrade_schema_version" must exist. That makes a ton of sense, and was the bit of context I was missing. Thank you 🙏

QMalcolm (approved):

Issue #4646, opened 2022-01-31
[CT-112] Better UX for macro dispatch


jtcohen6:
## Big idea

We're moving, slowly but surely, in the direction of sensible defaults for [macro dispatch](https://docs.getdbt.com/reference/dbt-jinja-functions/dispatch). Now that we've established conventions, let's find ways to make it easier/quicker to write.

Take [`dbt_utils.hash`](https://github.com/dbt-labs/dbt-utils/blob/main/macros/cross_db_utils/hash.sql) for example. This has to include a lot of supporting code to do something that's ultimately pretty simple:
```sql
{% macro hash(field) -%}
  {{ return(adapter.dispatch('hash', 'dbt_utils') (field)) }}
{%- endmacro %}

{% macro default__hash(field) -%}
    md5(cast({{field}} as {{dbt_utils.type_string()}}))
{%- endmacro %}

{% macro bigquery__hash(field) -%}
    to_hex({{dbt_utils.default__hash(field)}})
{%- endmacro %}
```

What are the problems here?
- The first macro, which dispatches a macro by its own name, is redundant / boilerplate. Could we support dispatch for _all_ macros, and make this the default behavior?
- The need to specify the `default__` adapter decorator is unintuitive for folks who are looking to take advantage of dispatch's package override capabilities, and who might be unfamiliar with its history as (primarily) a way to support cross-database macros

Imagine the same macro, like:
```sql
{% macro hash(field) -%}
    md5(cast({{field}} as {{dbt_utils.type_string()}}))
{%- endmacro %}

{% macro bigquery__hash(field) -%}
    to_hex({{dbt_utils.default__hash(field)}})
{%- endmacro %}
```

The implied (default) behavior is that a macro named `hash`, in the package named `dbt_utils`:
- Is going to search for macros named `hash` in the `dbt_utils` namespace
- Is itself the `default__` implementation of `hash`

The prime beneficiaries of this change would be package maintainers (+ us, as `global_project` maintainers).

There's two additional considerations we need to make, for "special kinds of macros": **materializations** and **generic tests**.

### Materializations

Adapter-specific materializations work differently from adapter-specific macros. Namely:
- `{% materialization table, adapter = 'snowflake' %}` creates a macro named `materialization_table_snowflake`, rather than (as one would expect) `snowflake__materialization_table`
[...]

PR #7258, opened 2023-04-01
New command: `dbt clone`

### LINKED ISSUE #7256: [Feature] `dbt clone` command

### Describe the feature

Big idea:
- Sorta similar to `--defer`. Using the `--state` manifest, but actually create zero-copy clones/pointers to the "prod" versions
- If the data platform supports `create or replace table ... clone`, use it!
- If it doesn't, or if the prod version isn't a table, just create a simple view instead: `select * from {prod_version}`
- By default, we shouldn't _recreate_ already-existing relations in the current target, but users could override this via `--full-refresh`
- Support a **ton** of threads! No need to go in DAG order for this.

Previously, I had thought (& written) that this should happen as part of `dbt run --defer`, as another deferral mode (instead of rewriting upstream references).

But I think that "cloning" should support node selection, too. I don't think there's a nice way to implicitly clone unselected resources and build selected resources, even if it's the ultimate UX we want to achieve.

Instead, you could do something like:
```bash
$ dbt clone --exclude state:modified+ --state path/to/manifest
$ dbt build --select state:modified+ --state path/to/manifest
```

### Who will this benefit?

Anyone who wants to avoid rerunning lots of models in dev/CI — but who still want to have a "full" dev/CI schema, with a pointer/clone of every relational object, in order to hook up to downstream querying (e.g. BI)

I'll push up a draft, and we can start some conversation there :)

### Anything else?

Lengthy previous discussion in https://github.com/dbt-labs/dbt-core/discussions/5095!

jtcohen6:
resolves #7256

```
$ dbt clone --state state/ --full-refresh
```

### Description

Introduce `clone` command, which also makes use of a `clone` materialization.
- On data platforms that support `create table <dev> clone <prod>`, let's use it
- Otherwise, create views that are simple pointers (`create view <dev> as select * from <prod>`)
- Don't recreate objects that already exist in the target schemas, unless `--full-refresh` is passed

Some interesting behaviors around:
- accessing the `--state` manifest in a way that's _similar to_, but _not exactly the same as_, `--defer`
- caching: we want to cache the other/prod schemas, in addition to the dev/target ones

### TODOs
- [x] Drop some comments below where I've left TODOs, or made some more inspired choices
- [x] Timing comparison of `dbt clone --threads 50 --full-refresh` (~3.5 minutes) versus `create schema <dev> clone <prod>` (~10 min) using our real Snowflake project with ~1k models in it. That's ~65% faster!
- [x] Add functional tests :) — in progress!

### Example

I have a seed, a view model, and a table model.

From `logs/dbt.log`:
```
19:07:08.285835 [debug] [Thread-1  ]: Using snowflake connection "model.test.another_model"
19:07:08.286490 [debug] [Thread-1  ]: On model.test.another_model: /* {"app": "dbt", "dbt_version": "1.5.0b5", "profile_name": "sandbox-snowflake", "target_name": "dev", "node_id": "model.test.another_model"} */
create or replace
      transient
      table analytics.dbt_jcohen.another_model
      clone analytics.dbt_jcohen_prod.another_model
19:07:08.287087 [debug] [Thread-2  ]: Using snowflake connection "model.test.my_model"
19:07:08.287513 [debug] [Thread-1  ]: Opening a new connection, currently in state closed
19:07:08.287933 [debug] [Thread-2  ]: On model.test.my_model: /* {"app": "dbt", "dbt_version": "1.5.0b5", "profile_name": "sandbox-snowflake", "target_name": "dev", "node_id": "model.test.my_model"} */
create or replace   view analytics.dbt_jcohen.my_model
  
   as (
    
        select * from analytics.dbt_jcohen_prod.my_model
    
  );
19:07:08.288501 [debug] [Thread-3  ]: Using snowflake connection "seed.test.my_seed"
19:07:08.289204 [debug] [Thread-2  ]: Opening a new connection, currently in state closed
19:07:08.289592 [debug] [Thread-3  ]: On seed.test.my_seed: /* {"app": "dbt", "dbt_version": "1.5.0b5", "profile_name": "sandbox-snowflake", "target_name": "dev", "node_id": "seed.test.my_seed"} */
create or replace
      transient
      table analytics.dbt_jcohen.my_seed
      clone analytics.dbt_jcohen_prod.my_seed
19:07:08.290207 [debug] [Thread-3  ]: Opening a new connection, currently in state init
19:07:09.648042 [debug] [Thread-2  ]: SQL status: SUCCESS 1 in 1.0 seconds
19:07:09.669806 [debug] [Thread-2  ]: Timing info for model.test.my_model (execute): 19:07:08.250276 => 19:07:09.669642
19:07:09.670302 [debug] [Thread-2  ]: On model.test.my_model: Close
19:07:10.163990 [debug] [Thread-2  ]: Finished running node model.test.my_model
19:07:10.267498 [debug] [Thread-1  ]: SQL status: SUCCESS 1 in 2.0 seconds
19:07:10.273490 [debug] [Thread-1  ]: Timing info for model.test.another_model (execute): 19:07:08.231088 => 19:07:10.273041
19:07:10.274599 [debug] [Thread-1  ]: On model.test.another_model: Close
19:07:10.411055 [debug] [Thread-3  ]: SQL status: SUCCESS 1 in 2.0 seconds
19:07:10.415530 [debug] [Thread-3  ]: Timing info for seed.test.my_seed (execute): 19:07:08.275505 => 19:07:10.415249
```

### Checklist

- [x] This PR includes tests, or tests are not required/relevant for this PR — in progress!


core/dbt/context/providers.py
```diff
@@ -1362,6 +1362,20 @@ def this(self) -> Optional[RelationProxy]:
             return None
         return self.db_wrapper.Relation.create_from(self.config, self.model)
 
+    @contextproperty
+    def state_relation(self) -> Optional[RelationProxy]:
```
jtcohen6, line 1366: Open to naming suggestions! This will _only_ be available in the context for the `clone` command currently

core/dbt/contracts/graph/manifest.py
```diff
@@ -1032,6 +1033,30 @@ def merge_from_artifact(
...
+        """
+        refables = set(NodeType.refable())
+        for unique_id, node in other.nodes.items():
+            current = self.nodes.get(unique_id)
+            if current and (node.resource_type in refables and not node.is_ephemeral):
+                other_node = other.nodes[unique_id]
+                state_relation = RelationalNode(
+                    other_node.database, other_node.schema, other_node.alias
+                )
+                self.nodes[unique_id] = current.replace(state_relation=state_relation)
```
jtcohen6, line 1054: We're storing information about each node's production-state counterpart, right on the node entry in the manifest. I'm open to discussing whether this is the right approach. It feels better than passing the entire other manifest around into other methods/runners.

core/dbt/contracts/graph/nodes.py
```diff
@@ -567,6 +571,7 @@ class HookNode(CompiledNode):
 class ModelNode(CompiledNode):
     resource_type: NodeType = field(metadata={"restrict": [NodeType.Model]})
     access: AccessType = AccessType.Protected
+    state_relation: Optional[RelationalNode] = None
```
jtcohen6, line 574: Given the approach of storing stateful information on each node about its prod counterpart, we do need a place on the node object to do that. For now, I'm adding a new attribute (default `None`) to models, seeds, and snapshots — the three `refable` nodes that are eligible for deferral & cloning.

core/dbt/include/global_project/macros/materializations/models/clone.sql
```diff
@@ -0,0 +1,107 @@
...
+
+
+{% macro default__can_clone_tables() %}
+    {{ return(False) }}
+{% endmacro %}
+
+
+{% macro snowflake__can_clone_tables() %}
+    {{ return(True) }}
+{% endmacro %}
```
jtcohen6, line 13: This kind of True/False conditional behavior might be better as an adapter property/method (Python), since it's really a property of the adapter / data platform, rather than something a specific user wants to reimplement. Comparable to the ["boolean macros"](https://github.com/dbt-labs/dbt-core/blob/c25260e5dd2afa237a30db115605ece9629443d1/core/dbt/include/global_project/macros/adapters/apply_grants.sql#L1-L39) we defined for logic around `grants`. Open to thoughts!

core/dbt/include/global_project/macros/materializations/models/clone.sql
```diff
@@ -0,0 +1,107 @@
...
+
+  {%- if other_existing_relation and other_existing_relation.type == 'table' and can_clone_tables -%}
+
+      {%- set target_relation = this.incorporate(type='table') -%}
+      {% if existing_relation is not none and not existing_relation.is_table %}
+        {{ log("Dropping relation " ~ existing_relation ~ " because it is of type " ~ existing_relation.type) }}
+        {{ drop_relation_if_exists(existing_relation) }}
+      {% endif %}
+
+      -- as a general rule, data platforms that can clone tables can also do atomic 'create or replace'
```
jtcohen6, line 73: This is me being a _little_ bit lazy, but it does hold true for Snowflake/BigQuery/Databricks. My goal is for all of our data platforms to be able to use this materialization "as is," and only reimplement a targeted set of macros where the behavior genuinely differs.

core/dbt/include/global_project/macros/materializations/models/clone.sql
```diff
@@ -0,0 +1,107 @@
...
+      }) -%}
+
+      -- reuse the view materialization
+      -- TODO: support actual dispatch for materialization macros
+      {% set search_name = "materialization_view_" ~ adapter.type() %}
+      {% if not search_name in context %}
+          {% set search_name = "materialization_view_default" %}
+      {% endif %}
+      {% set materialization_macro = context[search_name] %}
+      {% set relations = materialization_macro() %}
```
jtcohen6, line 102: This is some of the most inspired (read: horrifying) code I've written in years. Big idea:
- We want to create a view.
- We want the SQL in that view definition to be `select * from {other_relation}`.
- We already know how to create views for this data platform — it's the `view` materialization! — with all its atomicity and accoutrements.
- What if we could just _call and reuse_ the `view` materialization, within this one? Well... we actually can :)

Specific things to call out:
- Writing back into the `context` (via `context.update`) feels a little bit illegal... but I don't think it's _actually_ dangerous, the context is specific to this node
- It would be all-around better if materialization (+ test) macros followed the same naming convention as other dispatched macros, with `adapter__` prefixes, so that this could just be `adapter.dispatch("materialization_view")` (https://github.com/dbt-labs/dbt-core/issues/4646)

core/dbt/task/clone.py
```diff
@@ -0,0 +1,185 @@
...
+            thread_id=threading.current_thread().name,
+            execution_time=0,
+            message=message,
+            adapter_response=adapter_response,
+            failures=None,
+        )
+
+    def compile(self, manifest):
+        # no-op
+        return self.node
```
jtcohen6, line 51: This doesn't feel quite right, but the alternative is making this method `pass` and reimplementing `BaseRunner.compile_and_execute`

core/dbt/task/clone.py
```diff
@@ -0,0 +1,185 @@
...
+
+    def _materialization_relations(self, result: Any, model) -> List[BaseRelation]:
```
jtcohen6, line 53: A decent amount of this code is copy-pasted from `RunTask`. It also doesn't seem right for `CloneTask` to inherit from `RunTask`, since there's a lot of other stuff that we _don't_ want, among them hooks & the requirement to run in DAG order

core/dbt/task/clone.py
```diff
@@ -0,0 +1,185 @@
...
+            if node.is_relational and not node.is_ephemeral:
+                relation = adapter.Relation.create_from(self.config, node)
+                result.add(relation.without_identifier())
+
+                # cache the 'other' schemas too!
+                if node.state_relation:  # type: ignore
+                    other_relation = adapter.Relation.create_from_node(
+                        self.config, node.state_relation  # type: ignore
+                    )
+                    result.add(other_relation.without_identifier())
```
jtcohen6, line 118: We want to cache the other/state schemas as well, to know their relation types (view vs. table)

core/dbt/task/clone.py
```diff
@@ -0,0 +1,185 @@
...
+
+    def before_run(self, adapter, selected_uids: AbstractSet[str]):
+        with adapter.connection_named("master"):
+            # unlike in other tasks, we want to add information from the --state manifest *before* caching!
+            self.defer_to_manifest(adapter, selected_uids)
+            # only create *our* schemas, but cache *other* schemas in addition
+            schemas_to_create = super().get_model_schemas(adapter, selected_uids)
+            self.create_schemas(adapter, schemas_to_create)
+            schemas_to_cache = self.get_model_schemas(adapter, selected_uids)
+            self.populate_adapter_cache(adapter, schemas_to_cache)
```
jtcohen6, line 130: This is a bit tricksy of me:
- We want to use the standard `get_model_schemas` logic for creation (our schemas only)
- We want to use this task's custom `get_model_schemas` reimplementation for caching (our schemas _plus_ the state/other schemas)

core/dbt/task/clone.py
```diff
@@ -0,0 +1,185 @@
...
+                "--state is required for cloning relations from another environment"
+            )
+
+        if state.manifest is None:
+            raise DbtRuntimeError(f'Could not find manifest in --state path: "{self.args.state}"')
+        return state.manifest
+
+    # Note that this is different behavior from --defer with other commands, which *merge*
+    # selected nodes from this manifest + unselected nodes from the other manifest
+    def defer_to_manifest(self, adapter, selected_uids: AbstractSet[str]):
```
jtcohen6, line 175: This method has the same name, but different behavior, in that it calls `Manifest.add_from_artifact` instead of `Manifest.merge_from_artifact`. Could give it a different name for clarity

core/dbt/task/clone.py
```diff
@@ -0,0 +1,185 @@
...
+        deferred_manifest = self._get_deferred_manifest()
+        if deferred_manifest is None:
+            return
+        if self.manifest is None:
+            raise DbtInternalError(
+                "Expected to defer to manifest, but there is no runtime manifest to defer from!"
+            )
+        self.manifest.add_from_artifact(other=deferred_manifest)
+        # TODO: is it wrong to write the manifest here? I think it's right...
```
jtcohen6, line 184: This is an inherited TODO, from the code in the `compile` task :)

core/dbt/task/run.py
```diff
@@ -430,7 +431,10 @@ def before_run(self, adapter, selected_uids: AbstractSet[str]):
         with adapter.connection_named("master"):
             required_schemas = self.get_model_schemas(adapter, selected_uids)
             self.create_schemas(adapter, required_schemas)
-            self.populate_adapter_cache(adapter, required_schemas)
+            if get_flags().CACHE_SELECTED_ONLY is True:
+                self.populate_adapter_cache(adapter, required_schemas)
+            else:
+                self.populate_adapter_cache(adapter)
```
jtcohen6, line 437: IMO it makes more sense to have the `CACHE_SELECTED_ONLY` conditional check here, rather than in `populate_adapter_cache`. The behavior of `populate_adapter_cache` should just be: If I'm handed a specific set of schemas, cache those schemas; otherwise, cache all schemas defined in the manifest.

(Relevant to the current PR because the `clone` task wants to cache _more_ schemas than just the ones defined in _this_ manifest — we also want to cache the _other_ manifest's schemas)

core/dbt/task/runnable.py
```diff
@@ -384,6 +381,11 @@ def populate_adapter_cache(self, adapter, required_schemas: Set[BaseRelation] =
 
     def before_run(self, adapter, selected_uids: AbstractSet[str]):
         with adapter.connection_named("master"):
+            if get_flags().CACHE_SELECTED_ONLY is True:
+                required_schemas = self.get_model_schemas(adapter, selected_uids)
+                self.populate_adapter_cache(adapter, required_schemas)
+            else:
+                self.populate_adapter_cache(adapter)
```
jtcohen6, line 388: As in `RunTask.before_run` (comment above)

tests/functional/defer_state/test_defer_state.py
```diff
@@ -272,3 +272,68 @@ def test_run_defer_deleted_upstream(self, project, unique_schema, other_schema):
...
+        project.create_test_schema(other_schema)
+        self.build_and_save_state()
+
+        clone_args = ["clone", "--state", "state", "--target", "otherschema"]
+
+        results = run_dbt(clone_args)
+        # TODO: need an "adapter zone" version of this test that checks to see
+        # how many of the cloned objects are "pointers" (views) versus "true clones" (tables)
+        # e.g. on Postgres we expect to see 4 views
+        # whereas on Snowflake we'd expect to see 3 cloned tables + 1 view
```
jtcohen6, line 322: Either this test, or some extension of it, should be in the "adapter zone" so that we can verify this functionality:
> - On data platforms that support `create table <dev> clone <prod>`, let's use it
> - Otherwise, create views that are simple pointers (`create view <dev> as select * from <prod>`)

core/dbt/include/global_project/macros/materializations/models/clone.sql
```diff
@@ -0,0 +1,107 @@
...
+  {%- set existing_relation = load_cached_relation(this) -%}
+  {%- set other_existing_relation = load_cached_relation(state_relation) -%}
+
+  {%- if existing_relation and not flags.FULL_REFRESH -%}
+      -- noop!
+      {{ return(relations) }}
+  {%- endif -%}
+
+  -- If this is a database that can do zero-copy cloning of tables, and the other relation is a table, then this will be a table
+  -- Otherwise, this will be a view
```
jtcohen6, line 61: Data platforms that support table cloning:
- Snowflake ([docs](https://docs.snowflake.com/en/sql-reference/sql/create-clone))
- BigQuery ([docs](https://cloud.google.com/bigquery/docs/table-clones-create))
- Databricks ([docs](https://docs.databricks.com/sql/language-manual/delta-clone.html), with two modes: "shallow" (zero-copy) and "deep" (full copy)

Data platforms that don't:
- Postgres
- Redshift
- Trino

core/dbt/include/global_project/macros/materializations/models/clone.sql
```diff
@@ -0,0 +1,107 @@
...
+      {% endif %}
+
+      -- as a general rule, data platforms that can clone tables can also do atomic 'create or replace'
+      {% call statement('main') %}
+          {{ get_clone_table_sql(target_relation, state_relation) }}
+      {% endcall %}
+
+      {% set should_revoke = should_revoke(existing_relation, full_refresh_mode=True) %}
+      {% do apply_grants(target_relation, grant_config, should_revoke=should_revoke) %}
+      {% do persist_docs(target_relation, model) %}
```
jtcohen6, line 80: I'm thinking we should still apply grants & table/column-level comments. I have a suspicion that whether these things are copied over, during cloning, varies by data platform; I should really look to confirm/reject that suspicion. It's also possible that the user has defined conditional logic for these that differs between dev & prod, especially `grants`.

core/dbt/contracts/graph/manifest.py
```diff
@@ -1232,6 +1257,8 @@ def __post_serialize__(self, dct):
         for unique_id, node in dct["nodes"].items():
             if "config_call_dict" in node:
                 del node["config_call_dict"]
+            if "state_relation" in node:
+                del node["state_relation"]
```
jtcohen6, line 1261: It doesn't feel like `state_relation` is a thing we should need/want to include in serialized `manifest.json`. It's really just for our internal use.

core/dbt/include/global_project/macros/materializations/models/clone.sql
jtcohen6: Just to state the obvious: any `snowflake__` macros would want to move to `dbt-snowflake` as part of implementing & testing this on our adapters! I'm just defining them here for now for the sake of comparison & convenience (= laziness)

sungchun12:
Holy crap. Jerco and team, how are you shipping soooo much value right now?!?!

jtcohen6:
A decent amount of time is spent checking/updating the cache, which locks across all the concurrent threads. This could be even faster (2 min for 1k models) with the change proposed in https://github.com/dbt-labs/dbt-core/pull/6844.

I think we'd want to modify the behavior of `run_queue` for this command, such that it:
- Does not skip on the first failure
- Does not require running nodes in graph order, or skipping nodes downstream of any failed nodes

core/dbt/task/run.py
```diff
@@ -444,7 +445,10 @@ def before_run(self, adapter, selected_uids: AbstractSet[str]):
         with adapter.connection_named("master"):
```
VersusFacit, line 445: probably should not be master if we're trying to avoid that language elsewhere

core/dbt/context/providers.py
```diff
@@ -1427,6 +1427,20 @@ def this(self) -> Optional[RelationProxy]:
...
+        """
+        For commands which add information about this node's corresponding
+        production version (via a --state artifact), access the Relation
+        object for that stateful other
```
VersusFacit, line 1435: other feels vague (well, really, just Latinate/French in phrasing :) ). "source node", etc.?

core/dbt/include/global_project/macros/materializations/models/clone.sql
VersusFacit: I concur with both comments

core/dbt/include/global_project/macros/materializations/models/clone.sql
```diff
@@ -0,0 +1,115 @@
...
+{% endmacro %}
+
+
+{% macro get_pointer_sql(to_relation) %}
+    {{ return(adapter.dispatch('get_pointer_sql', 'dbt')(to_relation)) }}
+{% endmacro %}
+
+
+{% macro default__get_pointer_sql(to_relation) %}
+    {% set pointer_sql %}
```
VersusFacit, line 22: I feel uncomfortable with this use of pointer. I'm not exactly sure why were drawing that comparison, when perhaps we should use something closer to reference or allude to the fact that there's a shadow copy. In my mind pointer is just a very specific thing and someone anachronistic.

core/dbt/context/providers.py
VersusFacit: Is "relation" being used as a byword for table/view. Relation convene, so many things, and I feel like we want an additional term or a better term that can be more specific about what you're trying to achieve. Perhaps "stateful_db_relation". I really don't know the subtleties here though, so you might have picked the best one.

jtcohen6:
closing in favor of https://github.com/dbt-labs/dbt-core/pull/7881 :)

PR #7268, opened 2023-04-04
if need_swap then rename from existing - not target

### LINKED ISSUE #7267: [CT-2371] [Bug] incremental macro - need_swap rename needs to be from existing not target relation

### Current Behavior

Using the default implementation of the incremental macro, when executing the functional test https://github.com/tglunde/dbt-core/blob/main/tests/adapter/dbt/tests/adapter/basic/test_base.py - the incremental execution from the view model to be changed back to an incremental model (line 94) fails because it tries to rename the target_relation instead of the existing_relation - where the latter is the view previously created.

### Expected Behavior

in case of need_swap = true - the existing_model needs to be renamed - not the target model, because according to the test_base the target_model is a table - where the existing model could be a view. Using the existing_model the adapter macro could change from alter table to alter view - as we do in case of exasol.

### Steps To Reproduce

1. Execute test suite with default postgres implementation
2. alter table swappable rename ... - gets generated instead of - alter view swappable rename ...

### Relevant log output

_No response_

### Environment

```markdown
- OS: MacOS
- Python: 3.10.6
- dbt: 1.3.3
```

postgres

### Additional Context

_No response_

tglunde:
resolves #7267

### Description

in incremental.sql defaullt macro - use existing_model instead of target model - in order to reflect the correct type(table/view) of the existing model - in case materialization changes from view to incrmental.
 
### Checklist

### DIFF
--- a/core/dbt/include/global_project/macros/materializations/models/incremental/incremental.sql
+++ b/core/dbt/include/global_project/macros/materializations/models/incremental/incremental.sql
@@ -62,7 +62,7 @@
   {% endcall %}
 
   {% if need_swap %}
-      {% do adapter.rename_relation(target_relation, backup_relation) %}
+      {% do adapter.rename_relation(existing_relation, backup_relation) %}
       {% do adapter.rename_relation(intermediate_relation, target_relation) %}
       {% do to_drop.append(backup_relation) %}
   {% endif %}


jtcohen6:
Blocked by need for reliable reproduction case, in order to test the change accurately: https://github.com/dbt-labs/dbt-core/issues/7267#issuecomment-1529938989

tglunde:
removing this change - my fault, got confused with variable naming.

Issue #7191, opened 2023-03-19
[Feature] Ignore duplicate edges while building the subgraph for the job queue 


MatthieuBlais:
### Is this your first time submitting a feature request?

- [X] I have read the [expectations for open source contributors](https://docs.getdbt.com/docs/contributing/oss-expectations)
- [X] I have searched the existing issues, and I could not find an existing issue for this feature
- [X] I am requesting a straightforward extension of existing dbt functionality, rather than a Big Idea better suited to a discussion

### Describe the feature

Context: I'm investigating an issue with the command `dbt build` that became very slow in our project (1,300+ models and 20,000+ tests). It impacts user productivity (waiting time) and time-sensitive scheduled jobs. 

I suspect the function [Graph.get_subset_graph](https://github.com/dbt-labs/dbt-core/blob/main/core/dbt/graph/graph.py#L68) can become slow when running `dbt build` in a project with wide tables and many tests.

Let's consider a simple example:
```
Source A -----> Model B --------> Model C
     |                           |                               |
    \/                          \/                              \/ 
  Test A                Test B                     Test C
```

The [function](https://github.com/dbt-labs/dbt-core/blob/main/core/dbt/graph/graph.py#L68) loops through all the graph nodes and extract the upstream & downstream nodes. It creates the product of these nodes and adds the edges to the subgraph.
```
for node in self:
         if node not in include_nodes:
                source_nodes = [x for x, _ in new_graph.in_edges(node)]
                target_nodes = [x for _, x in new_graph.out_edges(node)]

                new_edges = product(source_nodes, target_nodes)
                non_cyclic_new_edges = [
                    (source, target) for source, target in new_edges if source != target
                ]  # removes cyclic refs

                new_graph.add_edges_from(non_cyclic_new_edges)
                new_graph.remove_node(node)
```

During `dbt build`, when evaluating the node `Test B` (and looking at the test nodes only), `source_nodes = [Test A]` and `target_nodes = [Test C]`. So, the edges to add to the subgraph are `[(Test A, Test C)]`. 
(_ -> Are these dependencies related to [add_test_edges](https://github.com/dbt-labs/dbt-core/blob/4186f99b742b47e0e95aca4f604cc09e5c67a449/core/dbt/compilation.py#L427)?_ ) 

Now, let's consider we have 30 tests in `Source A` and 500 tests in `Model C` --> We get 15000 new edges to add to the graph for node `Test B`.
[...]

PR #7271, opened 2023-04-04
Sort nodes in graph object by memory impact when building subdag

### LINKED ISSUE #7195: [CT-2316] [Feature] Improve subset_graph selection by removing trivial nodes and improving removal order

### Describe the feature

This is a performance suggestion separate from but similar to ideas explored in https://github.com/dbt-labs/dbt-core/issues/7191, https://github.com/dbt-labs/dbt-core/issues/6073 .

I have a DBT project with ~300 models and ~4200 tests. When selecting a single model to build, `dbt build` takes ~3 minutes to start. This is a major hurdle for development iterations.

I have diagnosed that this is because of the performance of subset_graph using `record_timing_info`. Using `dbt run` for the same model builds almost instantly, as does `dbt build` for the entire DAG.

There appear to be a number of major efficiency gains possible for this function. I propose a "smart trimming" of the DAG before starting the more expensive calculation in the function.

The function currently (randomly?) chooses a node, and if it is not in the selection, connects all immediate dependents and successors, then removes the node and continues iterating. This is presumably to preserve build order when nodes aren't directly connected and perhaps for other implications I haven't considered.

For a variety of reasons described in related issues, this can be expensive and cause the creation of *many* new edges, which most commonly will be deleted later. I see spikes of memory usage above 1 gig on my local machine when building for a single node and not above 200MB for the full DAG.

https://github.com/dbt-labs/dbt-core/issues/7191 describes some of these situations. For an additional example, one can imagine two wide tables with lots of tests, where removing one of the tables causes all tests in model A to be linked to model B on the scale of tens of thousands or hundreds of thousands of new connections, only for all tests and models to be removed later because they were not in the selector.

We can avoid these expensive operations to add new edges by first addressing the removal of nodes that do not require adding new edges. These trivial nodes are nodes with degree 0 or 1 (and are not part of the selector). If a node is not linked to at least 2 other nodes, removing it definitally does not require constructing new edges. By iterating over removing trivial nodes, the graph will collapse on a much simplier subgraph that the existing algorithm can much more efficiency prune.

Similarly, when moving on to remove non-selected nodes from the DAG, starting with the lowest degree nodes is likely to improve performance by reducing the number of edges added.

In the case of many common simple selectors, like single models with tests or simple paths, this algorithm gives the full intended subgraph without needing to invoke the more expensive algorithm.

There are other approaches to making this function more efficient, including an alternate trimming approach in https://github.com/dbt-labs/dbt-core/issues/6073#issuecomment-1312831392. It appears that suggestion ultimately did not end up being accepted into `dbt-core` for not making efficiency gains. My suggestion should be compatible with most other efficiency suggestions.

### Who will this benefit?

* Users of DBT with many nodes, especially tests
* Developers running a single model with `dbt build` during development

Yes!

### Anything else?

_No response_

ttusing:
Partially resolves #7195 

### Description

This implements the node sorting suggested in #7195. This is contained partially in https://github.com/dbt-labs/dbt-core/pull/7194 and a separate PR is opened here in case the other changes in #7194 are not ready to be reviewed. This also includes a performance check for this change alone with the other new performance enhancements merged into the `1.5` beta (https://github.com/dbt-labs/dbt-core/issues/7191).

Performance tests:

Current `1.5` beta:
```
dbt=1.5.0-b5
(env)  % dbt build --no-version-check -s mymodel
19:05:38  Running with dbt=1.5.0-b5
19:05:41  Found 306 models, 3346 tests... 
19:06:10  
19:06:16  
19:06:16  Running 1 on-run-start hook
19:06:16  1 of 1 START hook: on-run-start.0 ...................................... [RUN]
```
48 seconds

This branch:
```
(env)  % dbt build --no-version-check -s mymodel
19:03:39  Running with dbt=1.5.0-b5
19:03:42  Found 306 models, 3346 tests...
19:03:42  
19:03:48  
19:03:48  Running 1 on-run-start hook
19:03:48  1 of 1 START hook: on-run-start.0 ...................................... [RUN]
```
9 seconds

Because of the huge performance gains, I believe many teams would benefit from getting this change into DBT as soon as possible.

### Checklist

- [XX] I have read [the contributing guide](https://github.com/dbt-labs/dbt-core/blob/main/CONTRIBUTING.md) and understand what's expected of me


ttusing:
@iknox-fa @peterallenwebb I spun out this PR in case some of the changes in https://github.com/dbt-labs/dbt-core/pull/7194 require more in-depth review. I also re-ran performance checks, this sorting change alone improves at least my DBT project for developing a single model by 5x compared to `1.5.0-b5` and it seems highly unlikely to have any unintended impacts.

ttusing:
Closing since https://github.com/dbt-labs/dbt-core/pull/7194 was merged

PR #7272, opened 2023-04-04
ADAP-391: Add configuration change option

### LINKED ISSUE #6911: [CT-2068] [Feature] Let's add Materialized View as a materialization, finally

McKnight-42:
resolves #6911 

### Description

add new `on_configuration_change` config option into materialized_view materialization

### Checklist


core/dbt/contracts/graph/model_config.py
```diff
@@ -42,6 +42,20 @@ def _set_meta_value(obj: M, key: str, existing: Optional[Dict[str, Any]] = None)
     return result
 
 
+class OnConfigurationChange(str, Enum):
```
nssalian, line 45: I wonder if some naming like OnConfigurationChangeOptions might be better here? The name doesn't feel intuitive.

McKnight-42: kinda had the same thought for the NodeConfig option as well planned on raising it as it seems to wide range and people might not know its only for MV's

core/dbt/contracts/graph/model_config.py
```diff
@@ -445,6 +459,7 @@ class NodeConfig(NodeAndTestConfig):
     # sometimes getting the Union order wrong, causing serialization failures.
     unique_key: Union[str, List[str], None] = None
     on_schema_change: Optional[str] = "ignore"
+    on_configuration_change: List[OnConfigurationChangeOptions] = field(default_factory=list)
```
mikealfare, line 462: Can we have more than one setting for `on_configuration_change`? I think this might just be `Optional[OnConfigurationChangeOptions]`.

core/dbt/contracts/graph/model_config.py
```diff
@@ -42,6 +42,20 @@ def _set_meta_value(obj: M, key: str, existing: Optional[Dict[str, Any]] = None)
     return result
 
 
+class OnConfigurationChangeOptions(str, Enum):
```
mikealfare, line 45: I prefer the singular for a class name, so `OnConfigurationChangeOption`. Any objections?

core/dbt/contracts/graph/model_config.py
McKnight-42: changed name, trying to use the `Optional[OnConfigurationChangeOptions]` by itself just lead to a missing type issue, trying to provide a string "apply" value lead to a type mismatch error so tried to be more inline with the other Enum related classes and used the field call to set it

core/dbt/include/global_project/macros/materializations/models/materialized_view/get_materialized_view_configuration_changes.sql
```diff
@@ -0,0 +1,9 @@
+{% macro get_materialized_view_configuration_changes(existing_relation, new_config) %}
+    {{ log('Determining configuration changes on: ' ~ existing_relation) }}
+    {% do return(adapter.dispatch('get_materialized_view_configuration_changes', 'dbt')(existing_relation, new_config)) %}
+{% endmacro %}
+
+
+{% macro default__get_materialized_view_configuration_changes(existing_relation, new_config) %}
+    {% do return([{"indexes": ["id"]}]) %}
```
mikealfare, line 8: Need to back this out to `[]`

core/dbt/include/global_project/macros/materializations/models/materialized_view/materialized_view.sql
```diff
@@ -38,14 +44,20 @@
...
+    -- determine the scenario we're in: create, full_refresh, alter, refresh data
     {% if existing_relation is none %}
         {% set build_sql = get_create_materialized_view_as_sql(target_relation, sql) %}
-    {% elif full_refresh_mode or existing_relation.type != 'view' %}
+    {% elif full_refresh_mode or not existing_relation.is_view %}
         {% set build_sql = get_replace_materialized_view_as_sql(target_relation, sql, existing_relation, backup_relation, intermediate_relation) %}
+    {% elif config_updates and on_configuration_change == 'apply' %}
+        {% set build_sql = get_alter_materialized_view_as_sql(target_relation, config_updates, sql, existing_relation, backup_relation, intermediate_relation) %}
+    {% elif config_updates and on_configuration_change == 'skip' %}
+        {% set build_sql = 'select 1 as skip_configuration_changes where 0 = 1;' %}
```
mikealfare, line 55: Is there a better way to pass a "no-op" sql string into the expected call below?

tests/adapter/dbt/tests/adapter/materialized_views/test_materialized_views.py
```diff
@@ -4,6 +4,7 @@
 
 from dbt.tests.util import run_dbt, get_manifest
 from dbt.contracts.relation import RelationType
+from dbt.contracts.results import RunStatus
 
 
 class MaterializedViewTestsBase:
```
McKnight-42, line 10: Would we expect all of these to be passable for the `skip` and `fail` option or do we need to rename these to stop test call for automatic run and just call them as we would expect in the inherited class?

tests/adapter/dbt/tests/adapter/materialized_views/test_materialized_views.py
```diff
@@ -74,3 +81,50 @@ def test_updated_base_table_data_only_shows_in_materialized_view_after_rerun(sel
...
+        results = run_dbt(
+            ["run", "--models", self.materialized_view, "--vars", "quoting: {identifier: True}"],
+            expect_pass=False,
+        )
+        # breakpoint()
+        results.results[0].node.config.on_configuration_change == "fail"
+        assert results.results[0].status == RunStatus.Error
+
+
+class TestMaterializedViewFailTestsBase(MaterializedViewTestsFailConfigChangeBase):
```
McKnight-42, line 124: Screenshot 2023-04-17 at 3 23 23 PM

not sure if the exception we are wanting to throw will kick off as we might be hitting this one first

tests/adapter/dbt/tests/adapter/materialized_views/test_materialized_views.py
mikealfare: I need to think through the scenarios, but I suspect that some may not run appropriately until we are done with the feature. Either way, I would not rename the test. I would instead use the `@pytest.mark.skip()` decorator.

McKnight-42: will look into how flexible we can be with the skip, see if we can possibly use the configuration option to control if it skips or not

mikealfare: We need to test the matrix of `['apply', 'skip', 'fail']` X `[--full-refresh, not --full-refresh]` X `[configuration_change_1, ..., configuration_change_n]`. The third one is the tricky one because it changes by database platform. I think we wind up with a base class, or set of classes, that take care of the first two dimensions, and then the adapter maintainer needs to supply the last dimension as part of implementing materialized views.

We should only expose the scenarios where the configuration change matters. For example, the `on_configuration_change` setting shouldn't matter if --full-refresh is specified, even if there are configuration changes and 'fail' is selected. In that scenario, we know the outcome, regardless of what the adapter maintainer supplies.

We should also test the scenario where there are multiple configuration changes, one of which can be applied via an alter statement, but one of which requires a full refresh. In this case, we should show that we only do a full refresh. This would be a specific test within the 'apply' scenario.

I'm not sure this entirely answers your question. We can pair up on it if you'd like.

core/dbt/include/global_project/macros/materializations/models/materialized_view/materialized_view.sql
```diff
@@ -38,20 +44,28 @@
...
+        {% set build_sql = '' %}
+        {{ exceptions.warn("Updates were identified and `on_configuration_change` was set to `skip` for `" ~ target_relation ~ "`") }}
+    {% elif config_updates and on_configuration_change == 'fail' %}
+        {{ exceptions.raise_compiler_error("Updates were identified and `on_configuration_change` was set to `fail`") }}
     {% else %}
-        {% set build_sql = get_refresh_data_in_materialized_view_sql(target_relation) %}
+        {% set build_sql = refresh_materialized_view(target_relation) %}
     {% endif %}
 
     -- build model
```
mikealfare, line 63: @gshank Here's the scenario we discussed.

Note, we use an empty string on `line 55` to denote that we don't want to update the model if there are updates and the user has selected `skip` for `on_configuration_change`.

We condition here on whether `build_sql` is populated so that we don't try to execute an empty string. However, if we don't run any sql statement (which I assume includes prehooks and posthooks), then `line 79` causes an error because we don't have an open transaction.

We skip `line 79` if we have no `build_sql` (again this needs to consider prehooks and posthooks). But when we do that, we get the error `main is not being called during running model`.

Is there a way to avoid the check on main?

gshank: That check is in _build_run_model_result, which is called at the end of ModelRunner.execute. That references a "result" which is loaded by code in the provider context... see store_result, load_result, _sql_results, etc.  Possibly you could execute a "store_results" call, and provide a name and response?

"store_results" is called at the end of the "statement" macro, and also in core/dbt/include/global_project/macros/materializations/seeds/seed.sql.  The agate table is optional, so perhaps just a name ("main") and a response? ("OK")

mikealfare: Thanks for the suggestion. This appears to work. Does this align with what you were thinking? Are there any other updates we should make that would accompany this?

gshank: Looks good to me. If you're not seeing any errors, I think you're okay.

gshank: Note that the "store_results" provider method seems to be changing in https://github.com/dbt-labs/dbt-core/pull/7371, so you'll want to be sure to keep on eye on that.

tests/adapter/dbt/tests/adapter/materialized_views/test_materialized_views.py
```diff
@@ -74,3 +81,54 @@ def test_updated_base_table_data_only_shows_in_materialized_view_after_rerun(sel
...
+    @pytest.mark.skip("This currently fails since we're mocking with a traditional view")
+    def test_updated_base_table_data_only_shows_in_materialized_view_after_rerun(self, project):
+        pass
+
+    def test_on_configuration_change_fails_with_update(self, project):
```
McKnight-42, line 133: We can make this last failing test pass by removing the `expect_pass=fail` and the line on 125 currently

tests/adapter/dbt/tests/adapter/materialized_views/test_on_configuration_change.py
```diff
@@ -0,0 +1,173 @@
...
+        return {"models": {"on_configuration_change": str(self.on_configuration_change)}}
+
+    def assert_proper_scenario(
+        self,
+        results,
+        logs,
+        status: RunStatus,
+        messages_in_logs: List[str],
+        messages_not_in_logs: List[str],
+        rows_affected: int,
```
McKnight-42, line 27: on `skip or fail` scenarios would we expect the `rows_affected` to matchup? since nothing should be happening on the full refresh

PR #7275, opened 2023-04-05
Moving simple_snapshot test from individual adapters to core adapter zone

nssalian:
resolves #

### Description

Moving the simple_snapshot tests to the adapter zone so it can be used across adapters

### Checklist


tests/adapter/dbt/tests/adapter/simple_snapshot/test_snapshot.py
```diff
@@ -0,0 +1,203 @@
...
+            "id between 11 and 20",
+        )
+        run_dbt(["snapshot"])
+        self._assert_results(
+            ids_with_current_snapshot_records=range(1, 21),
+            ids_with_closed_out_snapshot_records=range(11, 21),
+        )
+
+
+class TestSnapshotCheck(SimpleSnapshotTestBase):
```
VersusFacit, line 184: To make it accessible to downstream repos, we're gonna need to strip off the `Test` from these names and redefined them with the pass keyword
[Example](https://github.com/dbt-labs/dbt-core/blob/a5131ecc7d6450ffdc3abe44a75df33c7b1b41e7/tests/adapter/dbt/tests/adapter/dbt_debug/test_dbt_debug.py#L67) -- if you don't do it this way, you'll end up re-running tests from the `import`

nssalian: Upon looking, I noticed that in that example, the other tests are similarly have their own checks that aren't in the Base class. The one you linked is only relying on the check_project and has no tests of its own, which isn't the case here.
To change the class name here and add `pass` means to move the `test_` methods all into the base.
Let me know if I understood

VersusFacit (approved):
I'm not gonna hold of this individual PR because apparently we have a bigger code style issue to work out & that's happening off-line.

PR #7276, opened 2023-04-05
Initial implementation of cross-project ref

### LINKED ISSUE #2327: Sprint 45: Core (4/15/2020 - 4/29/2020)

### LINKED ISSUE #7227: [CT-2327] [Spike] Produce publication artifact and resolve cross-project refs
Write a new 'publication artifact' as a result of dbt invocations. This artifact should contain the set of _logical_ project state necessary for downstream projects to take dependencies (resolve refs) to public models in the producer project. 

Minimally, the publication artifact would contain: 
* common artifact metadata (version, custom env vars, etc) 
* project name
* quoting policy
* the relation components (database, schema, alias) for _each public_ model in the project
* project-level dependencies (to prevent project-level cycles) or model-level dependencies (to prevent model-level cycles across projects)

Let's prove out that this publication artifact can be used to resolve two-variant refs, where the first variant is a project (not package) name. If the two-variant ref _can_ be resolved with a package - prefer the package resolution to using the publication artifact.

gshank:
resolves #7227

### Description

Initial implementation of cross project refs, including writing out publication artifacts, ingesting publication artifacts from other projects, constructing public nodes and resolving references to public nodes at runtime.

### Checklist

There's a bunch of unavoidable noise in this pr which can be skipped over for code review purposes.
* types_pb2.py update (binary-ish file which changes when logging event is added)
* schema changes for switching to a new schema version of the manifest (from 9 to 10)
* Refactored the Linker in compilation.py slightly to allow constructing a Graph object without pulling in Compiler code
* test updates due to the addition of new manifest fields (public_nodes mainly)

The new publication objects are in core/dbt/contracts/publication.py. There's a new Protocol to make a ModelNode and a ManifestNode look the same to the lookup code, plus the addition of "public_nodes" to DependsOn which is in core/dbt/contracts/graph/nodes.py. There are changes to fields in the Manifest and WritableManifest in core/dbt/contracts/graph/manifest.py. There are updates to the RefResolver code in core/dbt/context/providers.py which resolves the public refs at execution time.

Most of the functional changes are in core/dbt/parser/manifest, including "build_public_nodes" and "write_publication_artifact".


core/dbt/parser/manifest.py
```diff
@@ -598,6 +600,60 @@ def write_manifest_for_partial_parse(self):
...
+        for node in self.manifest.nodes.values():
+            if node.resource_type == NodeType.Model and node.access == AccessType.Public:
+                public_model_ids.append(node.unique_id)
+
+        set_of_public_unique_ids = set(public_model_ids)
+
+        # Get the Graph object from the Compiler
+        from dbt.compilation import Compiler
+
+        compiler = Compiler(self.root_project)
```
gshank, line 625: This is the first time that we're constructing the networkx graph at parse time. I will probably factor out the graph code from the Compiler object a bit better, because it feels wrong to be constructing a Compiler at parse time.

There was some test breakage relating to doing this. Probably we want to skip this for some errors, will have to investigate.

gshank: Have refactored to put the linker methods in the linker. The test breakage was because of a docs generate test that checked for the existence of manifest.json (it exists now), and because the experimental_parser tests had a model referring to itself, which prior to this caused no error. Now that I'm building the networkx graph we get cycle errors at parse time. Which seems better anyway.

gshank:
I handled the following from the walkthrough:
* move rebuild_ref_lookup
* clean up depends_on_public_nodes
* Remove writing manifest.json because it's already happening
* Change if/if to if/elif in "perform_lookup"

We have a number of open questions about names (particularly those pesky dependencies).

core/dbt/parser/manifest.py
```diff
@@ -599,6 +652,173 @@ def write_manifest_for_partial_parse(self):
         except Exception:
             raise
 
+    def write_artifacts(self):
+        # The manifest.json is written out in a task, so we're not writing it here
```
MichelleArk, line 656: What do you think about: 
1. renaming this method to `write_publication` (accepting a manifest + target_path, similar to write_path) and calling it alongside `write_manifest` in requires.py
2. Bundling the two calls (write_manifest, write_publication) together in a `write_artifact` method which is  called in requires.py. 

Given we're going to be introducing a semantic graph artifact soon, it'd be good to set a pattern for writing manifest-like objects consistently. Especially if there isn't a use case where we'd want to write one of the manifest, publication, or semantic graph separately from one another.

core/dbt/contracts/publication.py
```diff
@@ -0,0 +1,110 @@
...
+from dataclasses import dataclass, field
+
+from dbt.contracts.util import BaseArtifactMetadata, ArtifactMixin, schema_version
+from dbt.contracts.graph.unparsed import NodeVersion
+from dbt.contracts.graph.nodes import ManifestOrPublicNode
+from dbt.node_types import NodeType, AccessType
+
+
+@dataclass
+class DependentProjects(dbtClassMixin):
```
MichelleArk, line 14: naming suggestion: `ProjectDependency`

core/dbt/contracts/publication.py
```diff
@@ -0,0 +1,110 @@
...
+    name: str
+    environment: str
+
+
+@dataclass
+class Dependencies(dbtClassMixin):
```
MichelleArk, line 20: naming suggestion: `ProjectDependencies`

core/dbt/contracts/publication.py
```diff
@@ -0,0 +1,110 @@
...
+    """Used to represent cross-project models"""
+
+    name: str
+    package_name: str
+    unique_id: str
+    relation_name: str
+    version: Optional[NodeVersion] = None
+    latest_version: Optional[NodeVersion] = None
+    # list of model unique_ids
+    public_dependencies: List[str] = field(default_factory=list)
```
MichelleArk, line 44: What about - `public_node_dependencies`?

core/dbt/contracts/publication.py
gshank: Done.

core/dbt/contracts/publication.py
gshank: Done.

core/dbt/contracts/publication.py
gshank: Done

core/dbt/adapters/base/relation.py
```diff
@@ -244,6 +244,28 @@ def create_from_node(
...
+            schema = parts.pop(0)
+        if include_policy["identifier"]:
+            identifier = parts.pop(0)
+
+        return cls.create(
+            database=database,
+            schema=schema,
+            identifier=identifier,
+            quote_policy=quoting,
+        )
```
MichelleArk, line 267: I could see an argument for passing in the relation_name string directly from the publication artifact if it meant less processing hops in core (e.g. relation_name from upstream manifest -> relation_name in publication artifact -> relation_name used to create a Relation). But given we still need to unpack the relation name, I don't see many benefits beyond more explicit readability in the publication artifact files (which arguably aren't supposed to be human-friendly)

A few pros in favor of using the database, schema, and alias components 
- using an existing (tested and battle-tested) method (create_from_node)
- individual components would be available for user-space code when accessing a public_node
- individual components would be more flexible for potential core implementations (validation, error messaging for example)

Downsides could be - a larger publication artifact file.. any others?

MichelleArk: side implementation note - we'd need to add PublicNode to `ResultNode` in nodes.py to pass a public node to `Relation.create_from`. Should we do that either way?

gshank: I'd prefer the database/schema/identifier terminology instead of alias, because when constructing relations the field is called "identifier".

core/dbt/contracts/publication.py
```diff
@@ -0,0 +1,110 @@
...
+class PublicationMetadata(BaseArtifactMetadata):
+    dbt_schema_version: str = field(
+        default_factory=lambda: str(PublicationArtifact.dbt_schema_version)
+    )
+    adapter_type: Optional[str] = None
+    quoting: Dict[str, Any] = field(default_factory=dict)
+
+
+@dataclass
+class PublicModel(dbtClassMixin, ManifestOrPublicNode):
```
MichelleArk, line 34: hmm - what about `ProjectDependencyModel`, or simply `CrossProjectModel`? To avoid ambiguity with internal models that are public.

gshank: Would we change the name of the "public_nodes" dictionary then?

gshank: Also the same class is used to construct the public_models in the publication artifact.

MichelleArk: > Also the same class is used to construct the public_models in the publication artifact.

Ah, okay! I hadn't realized this when starring at this name in isolation. In that case I think PublicModel is fine 👍

core/dbt/adapters/base/relation.py
gshank: Adding PublicNode to ResultNode would give mypy fits. I think we probably don't want to do that. We'd have to end up creating dummy properties for almost every single field in a ModelNode.

MichelleArk:
Right now running a model with `{{ ref('invalid_project', 'model_name') }}` or `{{ ref('valid_project', 'invalid_model_name') }}`  throws the error: 
```
Compilation Error
  Model 'model.consumer_project.cross_project_ref' (models/cross_project_ref.sql) depends on a node named 'public_model' in package 'unknown' which was not found
```
Let's update the message in TargetNotFound to indicate this could be a node in a package or project: https://github.com/dbt-labs/dbt-core/blob/main/core/dbt/exceptions.py#L1363

core/dbt/parser/manifest.py
```diff
@@ -599,6 +652,173 @@ def write_manifest_for_partial_parse(self):
...
+    def build_public_nodes(self) -> bool:
+        """This method loads the dependencies from dependencies.yml, reads in the
+        the publication artifacts and adds the PublicModels to the manifest
+        "public_nodes" dictionary. It returns a boolean that indicates that
+        public nodes have been rebuilt."""
+        public_nodes_changed = False
+
+        # Load the dependencies from the dependencies.yml file
+        dependencies_filepath = resolve_path_from_base(
+            "dependencies.yml", self.root_project.project_root
```
MichelleArk, line 722: nit: should use DEPENDENCIES_FILE_NAME

core/dbt/contracts/publication.py
```diff
@@ -0,0 +1,110 @@
...
+
+
+@dataclass
+class PublicationMandatory:
+    project_name: str
+
+
+@dataclass
+@schema_version("publication", 1)
+class PublicationArtifact(ArtifactMixin, PublicationMandatory):
```
MichelleArk, line 92: naming consideration - is the `Artifact` here redundant? Should we just call this a `Publication`?

core/dbt/contracts/publication.py
```diff
@@ -0,0 +1,110 @@
...
+from dbt.contracts.util import BaseArtifactMetadata, ArtifactMixin, schema_version
+from dbt.contracts.graph.unparsed import NodeVersion
+from dbt.contracts.graph.nodes import ManifestOrPublicNode
+from dbt.node_types import NodeType, AccessType
+
+
+@dataclass
+class ProjectDependency(dbtClassMixin):
+    name: str
+    environment: str
```
MichelleArk, line 16: We can get rid of this field for now as we don't have plans to use it in core for 1.6. That said, allowing for additional attributes in `ProjectDependency` would enable projects to set up additional configuration that could be useful to them in fetching publication artifacts.

core/dbt/parser/manifest.py
```diff
@@ -599,6 +652,173 @@ def write_manifest_for_partial_parse(self):
...
+                version=model.version,
+                latest_version=model.latest_version,
+                public_node_dependencies=list(public_node_dependencies),
+                generated_at=metadata.generated_at,
+            )
+            public_models[unique_id] = public_model
+
+        dependencies = []
+        if self.manifest.project_dependencies:
+            for project in self.manifest.project_dependencies.projects:
```
MichelleArk, line 700: We'll need to include transitive dependencies (available through `self.manifest.publications`) in the publication artifact to [detect project-level cycles](https://github.com/dbt-labs/dbt-core/issues/7468). 

The high-level logic here would be to loop through the available publications in self.manifest.publications, and include their `dependencies` in the dependencies being constructed in addition to including the dependencies in self.manifest.project_dependencies.

core/dbt/contracts/publication.py
gshank: Do we want to allow random keys here? We could do that by including an additional base class that puts them in an _extra dictionary.

core/dbt/contracts/publication.py
gshank: I thought PublicationArtifact and PublicationConfig made the difference clearer, but I don't have strong feelings.

MichelleArk: I agree it makes the difference between the two clearer. And verbally I keep finding myself saying the full 'publication artifact' at this point. So I'm good to leave as-is!

core/dbt/parser/manifest.py
gshank: Moved some file_name constants to core/dbt/constants.py and used DEPENDENCIES_FILE_NAME

core/dbt/contracts/publication.py
MichelleArk: Yep, I think that'd be appropriate here.

core/dbt/parser/manifest.py
gshank: I separated this out into a function in core/dbt/parser/manifest.py, but didn't change to calling it in requires.py because it would get re-created on every single command invocation, which feels like too much to me. Right now it's recreated if parsing happens or the publication artifacts changed. Since it's now a function it can be easily moved somewhere else if necessary.

core/dbt/contracts/graph/manifest.py
```diff
@@ -640,6 +653,9 @@ class Manifest(MacroMethods, DataClassMessagePackMixin, dbtClassMixin):
     source_patches: MutableMapping[SourceKey, SourcePatch] = field(default_factory=dict)
     disabled: MutableMapping[str, List[GraphMemberNode]] = field(default_factory=dict)
     env_vars: MutableMapping[str, str] = field(default_factory=dict)
+    public_nodes: MutableMapping[str, PublicModel] = field(default_factory=dict)
+    project_dependencies: Optional[ProjectDependencies] = None
```
MichelleArk, line 657: for follow-up consideration: potentially should be moved to RuntimeConfig. Think it could be addressed as part of https://github.com/dbt-labs/dbt-core/issues/7469

core/dbt/parser/manifest.py
```diff
@@ -1385,7 +1546,20 @@ def _process_refs_for_node(manifest: Manifest, current_project: str, node: Manif
...
+            node.depends_on.add_public_node(target_model_id)
+        else:
+            node.depends_on.add_node(target_model_id)
+
+
+def remove_dependent_project_references(manifest: Manifest, publication: PublicationConfig):
+    for unique_id in publication.public_model_ids:
+        for child_id in manifest.child_map[unique_id]:
+            node = manifest.expect(child_id)
+            if hasattr(node.depends_on, "public_nodes"):
```
MichelleArk, line 1559: not blocking, for my own understanding - which nodes dont have depends_on.public_nodes?

core/dbt/parser/manifest.py
```diff
@@ -1482,6 +1656,74 @@ def process_node(config: RuntimeConfig, manifest: Manifest, node: ManifestNode):
...
+        quoting=root_project.quoting,
+    )
+
+    # get a list of public model ids first so it can be used in constructing dependencies
+    public_model_ids = []
+    for node in manifest.nodes.values():
+        if node.resource_type == NodeType.Model and node.access == AccessType.Public:
+            public_model_ids.append(node.unique_id)
+
+    set_of_public_model_ids = set(public_model_ids)
```
MichelleArk, line 1674: refactoring suggestion: 
``` python
public_node_ids = {node.unique_id for node in manifest.nodes.values() if node.resource_type == NodeType.Model and node.access == AccessType.Public}
```

tests/functional/multi_project/test_publication.py
```diff
@@ -0,0 +1,184 @@
...
+        results = run_dbt(["run"])
+        assert len(results) == 3
+
+        publication_dict = get_artifact(project.project_root, "target", "test_publication.json")
+        publication = PublicationArtifact.from_dict(publication_dict)
+        assert publication
+        assert len(publication.public_models) == 2
+        assert publication.public_models["model.test.model_three"].public_node_dependencies == [
+            "model.test.model_one"
+        ]
```
MichelleArk, line 109: Could we add a test with a publication that contains dependencies as well?

core/dbt/parser/manifest.py
```diff
@@ -599,6 +651,115 @@ def write_manifest_for_partial_parse(self):
...
+        """This method loads the dependencies from dependencies.yml, reads in the
+        the publication artifacts and adds the PublicModels to the manifest
+        "public_nodes" dictionary. It returns a boolean that indicates that
+        public nodes have been rebuilt."""
+        public_nodes_changed = False
+
+        # Load the dependencies from the dependencies.yml file
+        dependencies_filepath = resolve_path_from_base(
+            DEPENDENCIES_FILE_NAME, self.root_project.project_root
+        )
```
MichelleArk, line 664: Not blocking: we may need to rework this to implement https://github.com/dbt-labs/dbt-core/issues/7484

core/dbt/parser/manifest.py
```diff
@@ -442,6 +477,23 @@ def load(self):
...
+            # publications have changed and we need to run this to capture that.
+            self.manifest.build_parent_and_child_maps()
+            public_nodes_changed = self.build_public_nodes()
+            if public_nodes_changed:
+                self.manifest.rebuild_ref_lookup()
+                self.process_refs(self.root_project.project_name)
+                # parent and child maps will be rebuilt by write_manifest
+
+        if not skip_parsing or public_nodes_changed:
+            # Following adds publications to manifest too...
```
MichelleArk, line 495: is this comment still applicable? my understanding is that publications are now added to the manifest in build_public_nodes

core/dbt/parser/manifest.py
```diff
@@ -599,6 +651,115 @@ def write_manifest_for_partial_parse(self):
...
+            self.manifest.project_dependencies = dependencies
+        else:
+            self.manifest.project_dependencies = None
+
+        # Return False if there weren't any dependencies before and aren't any now.
+        if saved_manifest_dependencies is None and self.manifest.project_dependencies is None:
+            return False
+
+        # collect the names of the projects for later use
+        dependent_project_names = []
```
MichelleArk, line 679: naming nit: `project_dependency_names`

core/dbt/parser/manifest.py
```diff
@@ -599,6 +651,115 @@ def write_manifest_for_partial_parse(self):
         except Exception:
             raise
 
+    def build_public_nodes(self) -> bool:
+        """This method loads the dependencies from dependencies.yml, reads in the
+        the publication artifacts and adds the PublicModels to the manifest
+        "public_nodes" dictionary. It returns a boolean that indicates that
+        public nodes have been rebuilt."""
```
MichelleArk, line 658: This method is pretty long and involved - If there is a way to split it up into two - one for building public nodes (perhaps returning relevant saved artifacts), and one (private method) for detecting changes to public nodes - it would be helpful for readability. Could be a follow-on refactor

MichelleArk:
> Would it be possible to break up this PR?

@aranke - Technically this could be split up into (1) writing the publication artifact and (2) reading it and using it to resolve references. However most of the code changes are coming from autogenerated code (event types (~1k), new manifest  version (~5k)). So splitting it up into those functional components wouldn't help reduce the size of the PR. Additionally,  seeing both parts has been personally helpful as a reviewer as they inform each others implementation.

core/dbt/parser/manifest.py
gshank: SeedNodes, Macro, SourceDefinition

gshank:
@aranke There's a video walkthrough of the pr that's linked in the language channel.

core/dbt/parser/manifest.py
gshank: Yep, bad comment. Have changed.

core/dbt/parser/manifest.py
gshank: Yep. I do wonder if moving away from "get_full_manifest" to more directly constructing the ManifestLoader might be better if we're going to want to pass a lot of things in to get_full_manifest.

core/dbt/parser/manifest.py
gshank: Suggestion taken. Also changed the name of "public_model_ids" in the PublicationConfig to "public_node_ids", which appealed because it makes the difference a bit more clear.

core/dbt/parser/manifest.py
gshank: I moved the straightforward loading of new public nodes into a separate method. May not help a lot... The rest is a bit harder to pull apart because of detecting changes.

tests/functional/multi_project/test_publication.py
gshank: We need a bunch more tests. Haven't done a lot of them because in many cases it will take longer to write the test than the code, and I've been thinking about whether there's a way to include two projects in a test case, which our pytest fixtures makes pretty difficult.

gshank:
I added some more information on what changes are mostly noise and which are pertinent.

core/dbt/contracts/publication.py
gshank: Let's make that a follow-on ticket.

tests/functional/multi_project/test_publication.py
gshank: Have added check for dependencies in the PublicationArtifact and a test that actually uses two projects in the same test and copies the publication artifact.

MichelleArk:
Added one more follow-on item to consider graph selection / ls: https://github.com/dbt-labs/dbt-core/issues/7496

MichelleArk (approved):
Looks great. This is a big lift and covers the majority of our considerations for cross-project refs in dbt-core + sets a nice foundation for [the identified follow-on work](https://github.com/dbt-labs/dbt-core/issues/7372). Thank you for the all the in-depth spiking / exploration work leading up to this! 🚀

PR #7287, opened 2023-04-06
first pass - model versions

### LINKED ISSUE #7263: [CT-2352] Versioned Models - parsing and referencing
Introduce an optional ‘versions’ attribute on model yaml entry. versions is a list of 'version' structures. Spec:

```
models:
  - name: dim_customers
    latest_version: 2
    config:
      materialized: table
    columns:
      - name: customer_id
        description: This is the primary key
        data_type: int

    versions: # new!
      - v: 2
        # implemented_by: dim_customers_v2 - optional, this is the default naming convention
      - v: 1
        defined_in: dim_customers_any_filename
        config:
          alias: dim_customers
        columns:
          - include: * 
            exclude: customer_id
          - name: customer_id
            data_type: float 
```

### Acceptance Criteria
* each version of a model is its own node in the manifest - with its unique id `model.{project_name}.{model_name}.v{version}`
* all versions of a model have the same value for their `name` attribute
* each version of this model is "defined in" a model SQL/python file. The `defined_in` property points to the name of a .sql (or .py) file that implements that model version. By convention, this file is <model_name>_v<version>.sql, e.g. dim_customers_v2.sql. However, users could set defined_in to point to a different SQL file that doesn't meet the default naming convention. (These file names must be globally unique, even though they are not the actual model name.)
* the alias of a versioned model is not based on its file name. Instead, it is either the `alias` from the versioned `config` (if available) or defaults to `{model_name}_v{version}` if not. 
* the `v` key for versions accepts a string - and orders versions within a model by attempting to cast to an int or float, and falling back on string / alphanumeric ordering
* `ref` supports a new keyword argument - `version` (or `v` alias) - that retrieves the specified version of a model. if `version` is not provided to `ref` for a versioned model, `ref` retrieves the latest version of the model. 
* versioned models inherit any properties / configurations from the 'base' model. Additionally, they may have at most one `IncludeExclude` object in their `columns` list to avoid re-specifying most columns from the 'root' model.

MichelleArk:
resolves #7263

### Checklist

### DIFF
--- a/core/dbt/contracts/graph/nodes.py
+++ b/core/dbt/contracts/graph/nodes.py
@@ -405,6 +405,11 @@ class ParsedNode(NodeInfoMixin, ParsedNodeMandatory, SerializableType):
         self.created_at = time.time()
         self.description = patch.description
         self.columns = patch.columns
+
+        # TODO - these are model specific, so is access
+        self.version = patch.version
+        self.is_latest_version = patch.is_latest_version
+        self.name = patch.name
         # This might not be the ideal place to validate the "access" field,
         # but at this point we have the information we need to properly
         # validate and we don't before this.
@@ -567,6 +572,8 @@ class HookNode(CompiledNode):
 class ModelNode(CompiledNode):
     resource_type: NodeType = field(metadata={"restrict": [NodeType.Model]})
     access: AccessType = AccessType.Protected
+    version: Optional[str] = None
+    is_latest_version: Optional[bool] = None
 
 
 # TODO: rm?
@@ -1244,6 +1251,8 @@ class ParsedPatch(HasYamlMetadata, Replaceable):
 class ParsedNodePatch(ParsedPatch):
     columns: Dict[str, ColumnInfo]
     access: Optional[str]
+    version: Optional[Union[str, float]]
+    is_latest_version: Optional[bool]
 
 
 @dataclass
--- a/core/dbt/contracts/graph/unparsed.py
+++ b/core/dbt/contracts/graph/unparsed.py
@@ -103,11 +103,7 @@ TestDef = Union[Dict[str, Any], str]
 
 @dataclass
 class HasColumnAndTestProps(HasColumnProps):
-    tests: Optional[List[TestDef]] = None
-
-    def __post_init__(self):
-        if self.tests is None:
-            self.tests = []
+    tests: List[TestDef] = field(default_factory=list)
 
 
 @dataclass
@@ -143,6 +139,54 @@ class HasConfig:
 
 
 @dataclass
+class UnparsedVersion(HasConfig, dbtClassMixin):
+    # TODO: Consider separate version type var
+    v: Union[str, float] = ""
+    description: str = ""
+    meta: Dict[str, Any] = field(default_factory=dict)
+    constraints: List[Dict[str, Any]] = field(default_factory=list)
+    docs: Docs = field(default_factory=Docs)
+    defined_in: Optional[str] = None
+    tests: List[TestDef] = field(default_factory=list)
+    columns: Sequence[Union[dbt.helper_types.IncludeExclude, UnparsedColumn]] = field(
+        default_factory=list
+    )
+
+    def __lt__(self, other):
+        try:
+            v = type(other.v)(self.v)
+            return v < other.v
+        except ValueError:
+            try:
+                other_v = type(self.v)(other.v)
+                return self.v < other_v
+            except ValueError:
+                return str(self.v) < str(other.v)
+
+    @property
+    def include_exclude(self) -> dbt.helper_types.IncludeExclude:
+        return self._include_exclude
+
+    @property
+    def unparsed_columns(self) -> List:
+        return self._unparsed_columns
+
+    def __post_init__(self):
+        has_include_exclude = False
+        self._include_exclude = dbt.helper_types.IncludeExclude(include=[])
+        self._unparsed_columns = []
+        for column in self.columns:
+            if isinstance(column, dbt.helper_types.IncludeExclude):
+                if not has_include_exclude:
+                    self._include_exclude = column
+                    has_include_exclude = True
+                else:
+                    raise ParsingError("version can have at most one include/exclude element")
+            else:
+                self._unparsed_columns.append(column)
+
+
+@dataclass
 class UnparsedAnalysisUpdate(HasConfig, HasColumnDocs, HasColumnProps, HasYamlMetadata):
     access: Optional[str] = None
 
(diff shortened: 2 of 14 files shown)


core/dbt/parser/schemas.py
```diff
@@ -1005,6 +1056,76 @@ def _target_type(self) -> Type[UnparsedNodeUpdate]:
...
+                    access=unparsed_version.access or target.access,
+                    version=unparsed_version.v,
+                    is_latest_version=latest_version == unparsed_version.v,
+                )
+                versioned_model_node.patch(versioned_model_patch)
+                self.patch_node_config(versioned_model_node, versioned_model_patch)
+                source_file.append_patch(
+                    versioned_model_patch.yaml_key, versioned_model_node.unique_id
+                )
+            self.manifest.rebuild_ref_lookup()
```
MichelleArk, line 1123: TODO: look into whether this can be more granular and moved into the loop for a performance improvement

core/dbt/contracts/graph/nodes.py
```diff
@@ -405,6 +427,11 @@ def patch(self, patch: "ParsedNodePatch"):
         self.created_at = time.time()
         self.description = patch.description
         self.columns = patch.columns
+
```
MichelleArk, line 430: consider moving this to parse_patch code on PatchParser

core/dbt/parser/generic_test_builders.py
```diff
@@ -473,7 +515,9 @@ def build_raw_code(self) -> str:
 
     def build_model_str(self):
         targ = self.target
-        if isinstance(self.target, UnparsedNodeUpdate):
+        if isinstance(self.target, UnparsedModelUpdate) and self.version:
```
gshank, line 518: What happens here if the target is an UnparsedModelUpdate without self.version?

gshank (approved):
Looks really good!  I just had one question about "build_model_str"...

core/dbt/parser/generic_test_builders.py
MichelleArk: it's super subtle actually - but it would get handled by the `elif isinstance(self.target, UnparsedNodeUpdate)` case below because `UnparsedModelUpdate` subclasses `UnparsedNodeUpdate`. Handling that case more explicitly would probably be less error prone if there is a change to the inheritance structure between these two going forward. What do you think?

gshank: Yeah, I think I'd prefer that case to be handled explicitly, for the reason you give. Also more readable. Though a comment would probably be okay at this point.

core/dbt/context/providers.py
```diff
@@ -232,18 +235,26 @@ def validate_args(self, name: str, package: Optional[str]):
...
         if len(args) == 1:
             name = args[0]
         elif len(args) == 2:
             package, name = args
         else:
             raise RefArgsError(node=self.model, args=args)
-        self.validate_args(name, package)
-        return self.resolve(name, package)
+
+        version = kwargs.get("version") or kwargs.get("v")
```
jtcohen6, line 255: :)

core/dbt/contracts/graph/unparsed.py
```diff
@@ -142,6 +138,62 @@ class HasConfig:
...
+    def __lt__(self, other):
+        try:
+            v = type(other.v)(self.v)
+            return v < other.v
+        except ValueError:
+            try:
+                other_v = type(self.v)(other.v)
+                return self.v < other_v
+            except ValueError:
+                return str(self.v) < str(other.v)
```
jtcohen6, line 168: Cool! This behaves how I would expect if `latest_version` isn't set. Defaults to greatest (numeric) or alphabetically last (string).

core/dbt/contracts/graph/unparsed.py
```diff
@@ -142,6 +138,62 @@ class HasConfig:
...
+    description: str = ""
+    access: Optional[str] = None
+    config: Dict[str, Any] = field(default_factory=dict)
+    meta: Dict[str, Any] = field(default_factory=dict)
+    constraints: List[Dict[str, Any]] = field(default_factory=list)
+    docs: Docs = field(default_factory=Docs)
+    tests: Optional[List[TestDef]] = None
+    columns: Sequence[Union[dbt.helper_types.IncludeExclude, UnparsedColumn]] = field(
+        default_factory=list
+    )
```
jtcohen6, line 157: As discussed offline: I'd expect the default `columns` value for a version, if not otherwise set, to be `[include: "*"]`.

For example:
```yml
models:
  - name: my_model
    columns:
      - name: id
        data_type: int
    versions:
      - v: 2
      - v: 1
        columns: 
          - include: 'all'
          - name: id
            data_type: text
```
My expectation is that the `columns` in the top-level model definition == the `columns` in `v2`. In the current implementation, `v2` has `columns: []`.

(Not sure if that makes more sense to implement as the default here or in `get_columns_for_version` below)

core/dbt/include/global_project/macros/get_custom_name/get_custom_alias.sql
```diff
@@ -19,13 +19,17 @@
 
 {% macro default__generate_alias_name(custom_alias_name=none, node=none) -%}
 
-    {%- if custom_alias_name is none -%}
+    {%- if custom_alias_name -%}
```
jtcohen6, line 22: This if/else order makes much more sense to me, and is how it's always been in my head

core/dbt/node_types.py
```diff
@@ -56,6 +56,12 @@ def refable(cls) -> List["NodeType"]:
             cls.Snapshot,
         ]
 
+    @classmethod
+    def versioned(cls) -> List["NodeType"]:
+        return [
+            cls.Model,
+        ]
```
jtcohen6, line 63: 💥

jtcohen6 (commented):
I found a small partial parsing bug, and you just fixed it.

That's ... really it! I've done my best to break this, and it's been holding up well. A handful of comments, none of them a blocker to merging.

Nice work :)

core/dbt/parser/schemas.py
```diff
@@ -1005,6 +1056,132 @@ def _target_type(self) -> Type[UnparsedNodeUpdate]:
...
+                )
+
+                versioned_model_patch = ParsedNodePatch(
+                    name=target.name,
+                    original_file_path=target.original_file_path,
+                    yaml_key=target.yaml_key,
+                    package_name=target.package_name,
+                    description=unparsed_version.description or target.description,
+                    columns=version_refs.column_info,
+                    meta={**target.meta, **unparsed_version.meta},
```
gshank, line 1160: This is where the problem with the meta dictionaries is happening. There's a method in core/dbt/context/context_config.py, "_add_config_call" which does the equivalent of "_merge_field_value" in core/dbt/contracts/graph/model_config.py.  Maybe you could pull out just the parts of _add_config_call which merges the dictionaries (BaseConfig.mergebejavior["update"]})

MichelleArk: taking a look, thank you for the pointer!

core/dbt/contracts/graph/unparsed.py
```diff
@@ -142,6 +138,62 @@ class HasConfig:
...
+
+
+@dataclass
+class UnparsedVersion(dbtClassMixin):
+    v: NodeVersion
+    defined_in: Optional[str] = None
+    description: str = ""
+    access: Optional[str] = None
+    config: Dict[str, Any] = field(default_factory=dict)
+    meta: Dict[str, Any] = field(default_factory=dict)
```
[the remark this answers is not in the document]
MichelleArk: I removed `meta` as a top-level key from versions to stay consistent with the current behaviour for models. 

And made an update to handle merging "meta" within "config" 👍

core/dbt/parser/schemas.py
MichelleArk: Initially went down this route to merge "meta" nested within "config", but then found `deepmerge` in `dbt.utils` so went with that!

gshank: Sounds good. That duplicative merge code is some technical debt that it would be nice to clean up sometime.

dbeatty10 (commented):
I found some un-submitted review comments relating to changie 😅

Issue #6547, opened 2023-01-08
[CT-1767] [Feature] `dbt parse` should return a manifest


jtcohen6:
_Prerequisite: `dbt parse` supported as top-level command in new API (https://github.com/dbt-labs/dbt-core/issues/5550)_

`dbt parse` should return the [`Manifest`](https://github.com/dbt-labs/dbt-core/blob/94d6d19fb43f1ed466c9049dea40e203b23f89da/core/dbt/contracts/graph/manifest.py#L590) as its result. This makes it clearer that _the output of parsing is the manifest*_, and it would also enable programmatic users to do handy things like:
```python
# just for type hint below
from dbt.contracts.graph.manifest import Manifest

# initialize
from dbt.cli.main import dbtRunner
dbt = dbtRunner()

# 'parse' command
results, success = dbt.invoke(['parse'])

# hand waving this part for now -- we'd want 'results' to have a consistent type for all 'invoke' calls
manifest: Manifest = results...['manifest']

# now I can programmatically inspect the project contents!
# pseudo code: assert every non-ephemeral model has an "owner"
for model in [n for n in manifest.nodes if n.resource_type == 'model']:
    if model.materialized != 'ephemeral':
        assert (model.meta['owner'] is not None)
```

_*There are caveats to this statement, of course—the manifest is modified during `compile`, and technically there is other config resolution (profile, project, packages) that happens during parsing and is not included in the manifest—but it still holds 85% true today, and I'd like to see us getting to a place where it's 95% true._

### Current behavior

Currently, `dbt parse` returns `None` as its results.

It's possible to call the (undocumented, internal) `get_full_manifest` method directly and return a manifest. This is what `dbt.lib.parse_to_manifest` does:

https://github.com/dbt-labs/dbt-core/blob/94d6d19fb43f1ed466c9049dea40e203b23f89da/core/dbt/lib.py#L186-L189

The proposal in this issue is to expose that capability within `dbt-core`'s documented, contracted, top-level API.

### Motivation

**Programmatic, read-only access to project manifests**, as part of `dbt-core`'s initial public Python API + "library" capabilities.

This would make it quite easy to write "project quality" rules, that run in `pre-commit` or CI. Because this would be the **actual Manifest Python object**, it would also makes available "helper" methods specific to each node class, which are lost during serialization to flat dictionaries (Jinja `model` + `graph` variables) and JSON (`manifest.json`).

[...]

Issue #6526, opened 2023-01-05
[CT-1751] Config to optionally skip population of relation cache


jtcohen6:
Users should have the ability to turn off relation cache population, if they really need to. It should still be "on" by default.

```bash
$ dbt --no-populate-cache [run|test|...]
$ DBT_POPULATE_CACHE=0|1 dbt [run|test|...]
```
```yml
# profiles.yml (UserConfig)
config:
  populate_cache: true|false
```

This is different from entirely disabling or skipping over the cache — we're just skipping the population of the cache on startup. When dbt needs to run caching queries, I think it should still report "cache miss," and then cache the result of the metadata query, if it needs to be used again in the same invocation.

### Where

https://github.com/dbt-labs/dbt-core/blob/54538409509e0d677876f1102aa3bc0c67007278/core/dbt/task/runnable.py#L407-L421

### Who is it for?
- Users in large projects, or of data warehouses that are quite slow at running metadata queries.
- Users running into inexplicable issues with specific relations not showing up in the relation cache (https://github.com/dbt-labs/dbt-core/issues/6050)
- Interactive compile & preview (#6358, #6359), which need to be blazing-fast

### YMMV
End users will need to experiment with the approach that's most efficient for them, between:
- full cache enabled
- "cache selected only" ([docs](https://docs.getdbt.com/reference/global-configs#cache-database-objects-for-selected-resource))
- `--no-populate-cache`

I expect mileage may vary between dev, CI, and prod environments.

### Questions

Would this break behavior around `--defer`, which expects to use the relation cache to determine if model X already exists in the dev schema, or should have its reference rewritten to use the schema defined in the other manifest?

https://github.com/dbt-labs/dbt-core/blob/54538409509e0d677876f1102aa3bc0c67007278/core/dbt/contracts/graph/manifest.py#L1018

Imagining a future where interactive compile/preview want to be _both_ very fast, and able to correctly leverage `--defer`: We should also think more about making the adapter cache pluggable, as something that can live & persist outside of a single `dbt-core` invocation. It would be the responsibility of that other application wrapping `dbt-core` to handle cache invalidation (fun!).

AnotherGuitar:
We're using the dbt-databricks adaptor and would be very interested in seeing this feature get developed because compile times are taking a while for us. Thanks for writing up this issue!

PR #7292, opened 2023-04-06
dbt compile work with server, remove lib.py associated with it

ChenyuLInx:
Update compile query to clean up the node added during runtime initialization, and partially remove `lib.py`


core/dbt/task/compile.py
```diff
@@ -40,6 +40,18 @@ def compile(self, manifest):
 
 
 class CompileTask(GraphRunnableTask):
+    # We add a new inline node to the manifest during initialization
+    # it should be removed before the task is complete
+    _inline_node_id = None
+
+    # TODO remove when stu's PR about skip adapter cache is merged
```
ChenyuLInx, line 47: As stated, this should be removed once Stu's work is done.

jtcohen6: Tracked in https://github.com/dbt-labs/dbt-core/issues/6526

core/dbt/lib.py
```diff
@@ -111,88 +59,6 @@ def get_dbt_config(project_dir, args=None, single_threaded=False):
...
-    from dbt.task.sql import SqlExecuteRunner
-
-    config, node, adapter = _get_operation_node(manifest, project_path, sql, node_name)
-
-    runner = SqlExecuteRunner(config, adapter, node, 1, 1)
-
-    return runner.safe_run(manifest)
-
-
 def parse_to_manifest(config):
```
jtcohen6, line 62: If we implement #6547, we would be able to remove this method as well, right?

jtcohen6 (commented):
Screenshot 2023-04-10 at 14 01 49

you simply love to see it

ChenyuLInx:
Closing due to git history looks odd, will reopen

PR #7296, opened 2023-04-07
Fix for simple snapshot for seed and test assert

nssalian:
resolves #

### Description

ran into some issues while porting over big query and snowflake, this PR will address those changes.

### Checklist

### DIFF
--- a/tests/adapter/dbt/tests/adapter/simple_snapshot/test_snapshot.py
+++ b/tests/adapter/dbt/tests/adapter/simple_snapshot/test_snapshot.py
@@ -104,7 +104,8 @@ class BaseSimpleSnapshotBase:
             {(i, True) for i in ids_with_current_snapshot_records},
             {(i, False) for i in ids_with_closed_out_snapshot_records},
         )
-        assert records == expected_records
+        for record in records:
+            assert record in expected_records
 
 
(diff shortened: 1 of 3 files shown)


McKnight-42 (approved):

tests/adapter/dbt/tests/adapter/simple_snapshot/test_snapshot.py
```diff
@@ -104,7 +104,8 @@ def _assert_results(
             {(i, True) for i in ids_with_current_snapshot_records},
             {(i, False) for i in ids_with_closed_out_snapshot_records},
         )
-        assert records == expected_records
+        for record in records:
```
McKnight-42, line 107: do we know what makes the two objects unalike for snowflake?

nssalian: it was matching sets where are unordered. the equality breaks because it expects a set to be in a particular order. The record and expected_records have the same items but the ordering is not deterministic (as expected in a set)

McKnight-42: oh neat, thank you for the detailed explanation!.

PR #7303, opened 2023-04-10
Handle duplicated flags better

### LINKED ISSUE #6913: [CT-2069] Unexpected behavior with global/subcommand flags
For backwards compatibility with older behavior, some flags are implemented at both the "global" ("parent") level, and also at the "subcommand" ("child") level. We have some tricky logic to handle this potential duplication:
https://github.com/dbt-labs/dbt-core/blob/77748571b46593999dfc45b717a1bc1189ec2533/core/dbt/cli/flags.py#L70-L95

On `feature/click-cli` branch, I'm observing behavior where:
- parent flag alone isn't respected
- child flag alone is respected
- flag provided to both parent + child, parent flag takes precedence and is respected

**Expected behavior:**
- parent flag alone **is** (!) respected
- child flag alone is respected
- if both parent + child flags are passed, raise an error (ideal) — but if needed, we can document that the parent flag always wins

### Example

Create a new dbt project, and add `require-dbt-version: "1.4.1"` to `dbt_project.yml`.

```bash
$ python3 -m dbt.cli.main --no-version-check parse
...
dbt.exceptions.DbtProjectError: Runtime Error
  This version of dbt is not supported with the 'test' package.
    Installed version of dbt: =1.5.0-a1
    Required version of dbt for 'test': ['=1.4.1']
  Check for a different version of the 'test' package, or run dbt again with --no-version-check
```
```
$ python3 -m dbt.cli.main parse --no-version-check
23:52:24  Performance info: target/perf_info.json
```
```
$ python3 -m dbt.cli.main --no-version-check parse --version-check
23:52:27  Performance info: target/perf_info.json
```

iknox-fa:
resolves #6913

### Description
Re-works a considerable portion of the flags code and changes the handling of duplicated flags from parent-flag-wins to throws-an-error-when-duplicated-flags-exist.

### Checklist

### DIFF
--- a/tests/unit/test_cli_flags.py
+++ b/tests/unit/test_cli_flags.py
@@ -345,5 +345,5 @@ class TestFlags:
         parent_context = self.make_dbt_context("parent", ["--version-check"])
         context = self.make_dbt_context("child", ["--version-check"], parent_context)
 
-        with pytest.raises(click.BadOptionUsage):
+        with pytest.raises(DbtUsageException):
             Flags(context)


core/dbt/cli/flags.py
```diff
@@ -89,102 +69,121 @@ def args_to_context(args: List[str]) -> Context:
...
+                        )
 
-                    if (
-                        param_name not in EXPECTED_DUPLICATE_PARAMS
-                        and ctx.get_parameter_source(param_name) != "kwargs"
-                    ):
+                    # Rename for clarity.
+                    dep_name = param_name
+                    new_name = DEPRECATED_PARAMS.get(dep_name)
+                    assert isinstance(new_name, str)
```
stu-k, line 133: I believe the `except IndexError` catch below should handle `dep_name` not existing in `DEPRECATED_PARAMS` and `new_name` being set to `None`.

core/dbt/cli/flags.py
```diff
@@ -89,102 +69,121 @@ def args_to_context(args: List[str]) -> Context:
...
+                        new_name=new_param.envvar,
+                    )
+
+                # Set the flag value.
+                is_duplicate = hasattr(self, param_name.upper())
+                is_default = (
+                    True
+                    if ctx.get_parameter_source(param_name) == ParameterSource.DEFAULT
+                    else False
+                )
```
stu-k, line 163: This can be simplified
```suggestion
                is_default = ctx.get_parameter_source(param_name) == ParameterSource.DEFAULT
```

core/dbt/cli/flags.py
iknox-fa: While that's probably true, mypy isn't happy unless we actually assert that it's a string. /shrug

stu-k: Would `new_name = DEPRECATED_PARAMS.get(dep_name, "")` satisfy this instead?

core/dbt/cli/flags.py
iknox-fa: Good catch! Fixed.

core/dbt/cli/flags.py
```diff
@@ -1,56 +1,43 @@
...
 from pprint import pformat as pf
-from typing import Callable, Dict, List, Set
-
-from click import Context, get_current_context, BadOptionUsage
-from click.core import ParameterSource, Command, Group
+from typing import Callable, Dict, List, Set, Union
 
+from click import BadOptionUsage, Context, get_current_context
+from click.core import Command, Group, ParameterSource
+from dbt.cli.resolvers import default_log_path, default_project_dir
```
iknox-fa, line 11: No material changes here, just sorting imports.

core/dbt/cli/flags.py
iknox-fa: No, but we can wrap it and raise a better looking exception. ✅

stu-k (approved):
LGTM 👍

PR #7304, opened 2023-04-10
Add relation info to `node_info` in logging

### LINKED ISSUE #6724: [Feature] Add `relation_name` to the `node_info` dictionary

### Describe the feature

As described by @jtcohen6 [here](https://github.com/dbt-labs/dbt-core/issues/6712#issuecomment-1402625556) this feature would:

- Add `relation_name` (`database.schema.identifier`) to the `node_info` dictionary, which is present on _all_ node-related logging events, and offers a convenience for looking up the most salient information. (The rest can be looked up in the manifest, using the node's `unique_id`.)

This would provide a mapping from `relation_name` → node `unique_id` (or vice versa). Other properties can be looked up via the node's `unique_id` within [`manifest.json`](https://docs.getdbt.com/reference/artifacts/manifest-json).

I'm not aware of any reasonable alternatives to determine the final `database.schema.identifier` for a model that has been built.

For those who love a good unstructured text parsing challenge, I suppose they could try to extract this from `logs/dbt.log` somehow.

### Who will this benefit?

There is at least one issue where a user wants to monitor logs for successful model execution and trigger another process downstream:
- https://github.com/dbt-labs/dbt-core/issues/6712

_No response_

### Anything else?

_No response_

jtcohen6:
resolves #6724

### Description

Add to `node_info` dictionary another nested message/dictionary for `node_relation`, including:
- `database`
- `schema`
- `alias`
- `relation_name` (with include + quote policies applied)

```json
{
  "data": {
    "description": "sql view model dbt_jcohen.my_model",
    "index": 1,
    "node_info": {
      "materialized": "view",
      "meta": {},
      "node_finished_at": "",
      "node_name": "my_model",
      "node_path": "my_model.sql",
      "node_relation": {
        "alias": "my_model",
        "database": "jerco",
        "relation_name": "\"jerco\".\"dbt_jcohen\".\"my_model\"",
        "schema": "dbt_jcohen"
      },
      "node_started_at": "2023-04-10T19:21:43.288413",
      "node_status": "started",
      "resource_type": "model",
      "unique_id": "model.test.my_model"
    },
    "total": 1
  },
  "info": {
    "category": "",
    "code": "Q011",
    "extra": {},
    "invocation_id": "a0234d14-04ef-4eec-94bc-f409100f92e4",
    "level": "info",
    "msg": "1 of 1 START sql view model dbt_jcohen.my_model ................................ [RUN]",
    "name": "LogStartLine",
    "pid": 8576,
    "thread": "Thread-1",
    "ts": "2023-04-10T19:21:43.289185Z"
  }
}
```

### Checklist

- [ ] This PR includes tests, or tests are not required/relevant for this PR — do we need more tests?
- [ ] I have [opened an issue to add/update docs](https://github.com/dbt-labs/docs.getdbt.com/issues/new/choose), or docs changes are not required/relevant for this PR — TODO!


core/dbt/events/types.proto
```diff
@@ -37,6 +37,10 @@ message NodeInfo {
     string node_started_at = 7;
     string node_finished_at = 8;
     google.protobuf.Struct meta = 9;
+    string node_database = 10;
+    string node_schema = 11;
+    string node_alias = 12;
+    string node_relation_name = 13;
```
jtcohen6, line 43: @gshank @peterallenwebb Does either of you have a strong feeling around whether:
- We should prefix these with `node_` (like the other ones)
- We should nest these as fields in a `relation_info` dictionary, instead of having them all as top-level fields (strings)

gshank: tbh I kind of wish none of the other fields started with "node_". It feels duplicative. But that's moot, I guess :)

Nested or not I don't have strong feelings.

I wonder if we need to make these "optional" though. Did you test this with a node with database = None?  The version of betterproto we were using before didn't have "optional", and google protobuf behaves a bit differently anyway. I'm not quite sure what it will do with a None passed in. Changing the event in test_events.py should tell.

gshank: How about node_relation: { database='...', schema='', alias='', name: '...' ?

jtcohen6: If I pass in `None` for `database`, it comes back as `""`. I'll add a test for this. I think this is how proto works, right? All fields are technically optional, and there's no way to really distinguish between empty string (`""`) and missing (`None`), short of adding another boolean field `has_database`

gshank: There is an "optional" thing, but it's not exactly the same as in Python, so I suspect we'd still get an empty string. betterproto was trying to be more python-like, so I think they implemented it in a somewhat more pythonic way, but I haven't done any optional fields with google protobuf yet...

jtcohen6: I went with
```
node_relation: {database='...', schema='...', alias='...', relation_name: '...'}
```
To match the nomenclature in the existing top-level node attributes, and because `name` by itself feels closer to alias/identifier

gshank (approved):

Issue #6240, opened 2022-11-13
[CT-1498] Update cache after cache misses in `list_relations`


jtcohen6:
_Original issue context: https://github.com/dbt-labs/dbt-core/issues/6237_

dbt encounters a "cache miss" whenever a user calls `adapter.get_relation` on a schema that hasn't been cached. When does this come up in practice? Perhaps if:
- user code is trying to introspect high-level information about source tables (`sources` aren't cached, since they aren't materialized)
- user code is introspecting information about another materialized relation that isn't selected to run, _and_ is using the [experimental config](https://docs.getdbt.com/reference/global-configs#cache-database-objects-for-selected-resource) that limits caching to only schemas containing selected relations

When a "cache miss" happens, dbt runs a caching-style query for that schema, and returns the results — but it doesn't actually update the cache with those results. There's a comment hinting at why:
https://github.com/dbt-labs/dbt-core/blob/ff2f1f42c394d27ff8d49ca7e285fcf2b76faa05/core/dbt/adapters/base/impl.py#L713-L735

That code annotation is pretty direct ("we can't build the relations cache here"), but it's also super old (https://github.com/dbt-labs/dbt-core/commit/f4afd495ad8164cdf034e36528ce3b635a2fba92), from the very PR that added the relations cache in the first place.

**Possible implementation:**

```python
    def list_relations(self, database: Optional[str], schema: str) -> List[BaseRelation]:
        if self._schema_is_cached(database, schema):
            return self.cache.get_relations(database, schema)

        schema_relation = self.Relation.create(
            database=database,
            schema=schema,
            identifier="",
            quote_policy=self.config.quoting,
        ).without_identifier()

        relations = self.list_relations_without_caching(schema_relation)
        
        # if the cache is already populated, add this schema in
        # otherwise, skip updating the cache and just ignore
        if self.cache:
            for relation in relations:
                self.cache.add(relation)
            if not relations:
                # it's possible that there were no relations in some schemas. We want
                # to insert the schemas we query into the cache's `.schemas` attribute
                # so we can check it later
                cache_update: Set[Tuple[Optional[str], Optional[str]]] = set()
                for relation in cache_schemas:
                    cache_update.add((database, schema))
                self.cache.update_schemas(cache_update)

        return relations
```

This way, if a relation (same or different) from that originally uncached schema is asked about again, it can be a _cache hit_ rather than another miss, requiring another go with the same query.

PR #7307, opened 2023-04-10
Add option to skip relation cache population

### LINKED ISSUE #6526: [CT-1751] Config to optionally skip population of relation cache

stu-k:
resolves #6526

### Description

Add a `--populate-cache` flag to optionally skip relation cache population, defaults to `True`.

### Checklist

### DIFF
--- a/core/dbt/adapters/base/impl.py
+++ b/core/dbt/adapters/base/impl.py
@@ -728,12 +728,6 @@ class BaseAdapter(metaclass=AdapterMeta):
                 # it's possible that there were no relations in some schemas. We want
                 # to insert the schemas we query into the cache's `.schemas` attribute
                 # so we can check it later
-
-                # Jeremy: what was the intent behind this inner loop?
-                # cache_update: Set[Tuple[Optional[str], Optional[str]]] = set()
-                # for relation in cache_schemas:
-                #     cache_update.add((database, schema))
-
                 self.cache.update_schemas([(database, schema)])
 
         fire_event(
--- a/core/dbt/adapters/base/impl.py
+++ b/core/dbt/adapters/base/impl.py
@@ -734,7 +734,7 @@ class BaseAdapter(metaclass=AdapterMeta):
                 # for relation in cache_schemas:
                 #     cache_update.add((database, schema))
 
-                self.cache.update_schemas(set((database, schema)))
+                self.cache.update_schemas([(database, schema)])
 
         fire_event(
             ListRelations(
--- a/core/dbt/adapters/base/impl.py
+++ b/core/dbt/adapters/base/impl.py
@@ -718,6 +718,24 @@ class BaseAdapter(metaclass=AdapterMeta):
         # we can't build the relations cache because we don't have a
         # manifest so we can't run any operations.
         relations = self.list_relations_without_caching(schema_relation)
+
+        # if the cache is already populated, add this schema in
+        # otherwise, skip updating the cache and just ignore
+        if self.cache:
+            for relation in relations:
+                self.cache.add(relation)
+            if not relations:
+                # it's possible that there were no relations in some schemas. We want
+                # to insert the schemas we query into the cache's `.schemas` attribute
+                # so we can check it later
+
+                # Jeremy: what was the intent behind this inner loop?
+                # cache_update: Set[Tuple[Optional[str], Optional[str]]] = set()
+                # for relation in cache_schemas:
+                #     cache_update.add((database, schema))
+
+                self.cache.update_schemas(set((database, schema)))
+
         fire_event(
             ListRelations(
                 database=cast_to_str(database),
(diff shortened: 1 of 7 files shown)


core/dbt/task/runnable.py
```diff
@@ -371,6 +371,9 @@ def _mark_dependent_errors(self, node_id, result, cause):
             self._skipped_children[dep_node_id] = cause
 
     def populate_adapter_cache(self, adapter, required_schemas: Set[BaseRelation] = None):
+        if not self.args.populate_cache:
+            return
```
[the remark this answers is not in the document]
stu-k: There is no real difference, no. `self.args` is `Flags` which we should be using much more where we can. I think the places we use`get_flags` instead of `self.args` was just to get the click feature branch over the line to be merged.

core/dbt/adapters/base/impl.py
```diff
@@ -718,6 +718,24 @@ def list_relations(self, database: Optional[str], schema: str) -> List[BaseRelat
...
+                # it's possible that there were no relations in some schemas. We want
+                # to insert the schemas we query into the cache's `.schemas` attribute
+                # so we can check it later
+
+                # Jeremy: what was the intent behind this inner loop?
+                # cache_update: Set[Tuple[Optional[str], Optional[str]]] = set()
+                # for relation in cache_schemas:
+                #     cache_update.add((database, schema))
+
+                self.cache.update_schemas(set((database, schema)))
```
stu-k, line 737: @jtcohen6 While pairing with @ChenyuLInx we weren't sure what this inner loop was doing, especially since `cache_schemas` isn't defined anywhere.

jtcohen6: @stu-k The cache is keyed on `database` + `schema`. When dbt does a cache lookup, it asks: Have I already cached this `database` + `schema` combo? If the combo isn't present as a key in the cache, it's a cache miss, and dbt needs to go run a query. If those keys are present, and no relations are found, then the assumption is that the schema is _empty_ (= missing from the database), rather than just missing from the cache.

If there are no `relations` returned by the query, we still want to record that, by inserting an _empty_ set for each `database` + `schema` pair. That way, the next time dbt wants to know if any relations are in that `database.schema`, it doesn't need to run the same query over again — the cache says, I already know there aren't any relations there.

stu-k: I understand that, but I do not understand the intent of your inner loop I've commented on here.
```python
cache_update: Set[Tuple[Optional[str], Optional[str]]] = set()
for relation in cache_schemas:
    cache_update.add((database, schema))
self.cache.update_schemas(cache_update)
```
What is this code trying to accomplish? It is difficult to determine because `cache_schemas` isn't defined in [your possible implementation on the original bug ticket](https://github.com/dbt-labs/dbt-core/issues/6240).

ChenyuLInx (approved):
LGTM, also tested with `dbt-server` compile endpoint

PR #7242, opened 2023-03-29
Handle internal exceptions in postflight

stu-k:
resolves #7118 

### Description

- Handle all `dbt` thrown exceptions in the `postflight` decorator
- Add `DbtRunnerResult` dataclass with the intent of `dbtRunner.invoke` not throwing exceptions
- Handle all exceptions thrown from `cli.invoke` within `dbtRunner.invoke`
- Add exception handling to `dbt/cli/README.md`
- Change `run_dbt` method from `tests/util.py` to raise any exceptions returned from `dbtRunner.invoke`
- Move cli exceptions into `dbt/cli/exceptions.py`

### Checklist

- [X] I have read [the contributing guide](https://github.com/dbt-labs/dbt-core/blob/main/CONTRIBUTING.md) and understand what's expected of me
- [X] I have signed the [CLA](https://docs.getdbt.com/docs/contributor-license-agreements)
- [X] I have run this code in development and it appears to resolve the stated issue
- [X] This PR includes tests, or tests are not required/relevant for this PR
- [X] I have [opened an issue to add/update docs](https://github.com/dbt-labs/docs.getdbt.com/issues/new/choose), or docs changes are not required/relevant for this PR
- [X] I have run `changie new` to [create a changelog entry](https://github.com/dbt-labs/dbt-core/blob/main/CONTRIBUTING.md#adding-a-changelog-entry)

PR #7314, opened 2023-04-10
`Parse` command should return manifest when invoked via `dbtRunner`

### LINKED ISSUE #6547: [CT-1767] [Feature] `dbt parse` should return a manifest

iknox-fa:
resolves #6547

### Description
Allows easy programmatic access to the manifest as generated by the `cli.requires.manifest` decorator.  Currently a very, VERY simple implementation.

### Checklist


core/dbt/cli/main.py
```diff
@@ -48,7 +48,7 @@ def __init__(
         self.manifest = manifest
         self.callbacks = callbacks
 
-    def invoke(self, args: List[str], **kwargs) -> Tuple[Optional[List], bool]:
+    def invoke(self, args: List[str], **kwargs) -> Tuple[Optional[Any], bool]:
```
jtcohen6, line 51: @stu-k updated this in https://github.com/dbt-labs/dbt-core/pull/7242 (rightly so!) so it now returns a `dbtRunnerResult` class. I think we want to update [this union](https://github.com/dbt-labs/dbt-core/blob/7a4d3bd2dc770885b74322b1c6811e80207ad342/core/dbt/cli/main.py#L52-L59) to now also include `Manifest`.

iknox-fa: Ahhh gotcha.  I thought you wanted it to be `Any` because it was going to be something different, but you didn't know what it was going to be and you didn't want things to break when Stu merged the other PR.  Makes more sense now.

stu-k (approved):
I'm sure this could benefit from a test somewhere, but this implementation looks correct 👍

iknox-fa:
> I'm sure this could benefit from a test somewhere, but this implementation looks correct 👍
I was thinking about that but I'm not sure what I could test that wasn't already handled by the functional tests for every command that uses a manifest--  other than that it's just basic passthrough logic.

If you think of a useful thing to test let me know-- I'm certainly not opposed to more testing.

PR #7317, opened 2023-04-11
Reset invocation_id Between Commands

### LINKED ISSUE #7197: [CT-2317] `invocation_id` is cached between invocations in same Python process
Originally reported by @jeremyyeo:

> With dbt >= 1.4, we seem to be maintaining invocation_id across steps in a single run.

Starting in v1.4, `invocation_id` is now stored on `EventManager`:
https://github.com/dbt-labs/dbt-core/blob/65fcbe7b79c66b15a833fd9033117cc3f8ad3b3b/core/dbt/events/eventmgr.py#L185

I think we should clear or reset `invocation_id` as part of the steps in `cleanup_event_logger`:

https://github.com/dbt-labs/dbt-core/blob/0ef9931d197c996ccd4e3caf641729176eceda99/core/dbt/events/functions.py#L269-L272

https://github.com/dbt-labs/dbt-core/blob/0ef9931d197c996ccd4e3caf641729176eceda99/core/dbt/events/functions.py#L160-L165

### Reproduction case

- Create an `on-run-start` hook with `{{ log(invocation_id, info = true) }}`
- Create a dbt Cloud job with multiple run steps
- See if the invocation IDs are the same or different

peterallenwebb:
resolves #7197 

### Description

Reset invocation_id between command invocations.

### Checklist

### DIFF
--- a/core/dbt/cli/requires.py
+++ b/core/dbt/cli/requires.py
@@ -9,7 +9,7 @@ from dbt.cli.exceptions import (
 from dbt.cli.flags import Flags
 from dbt.config import RuntimeConfig
 from dbt.config.runtime import load_project, load_profile, UnsetProfile
-from dbt.events.functions import setup_event_logger, fire_event, LOG_VERSION
+from dbt.events.functions import fire_event, LOG_VERSION, set_invocation_id, setup_event_logger
 from dbt.events.types import (
     CommandCompleted,
     MainReportVersion,
@@ -48,6 +48,8 @@ def preflight(func):
         # Logging
         # N.B. Legacy logger is not supported
         callabcks = ctx.obj.get("callbacks", [])
+
+        set_invocation_id()
         setup_event_logger(flags=flags, callbacks=callabcks)
 
         # Now that we have our logger, fire away!
--- a/tests/unit/test_invocation_id.py
+++ b/tests/unit/test_invocation_id.py
@@ -1,0 +1,24 @@
+from dbt.cli.main import dbtRunner
+
+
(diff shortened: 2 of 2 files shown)


tests/functional/sources/test_source_fresher_state.py
```diff
@@ -90,7 +90,6 @@ def _assert_freshness_results(self, path, state):
             == "https://schemas.getdbt.com/dbt/sources/v3.json"
         )
         assert data["metadata"]["dbt_version"] == dbt.version.__version__
-        assert data["metadata"]["invocation_id"] == dbt.tracking.active_user.invocation_id
```
peterallenwebb, line 93: Unless I'm misunderstanding the nature of this test, the invocation_id shouldn't really matter, and the change in dbt's behavior, which now resets the invocation_id on every invocation, was causing this check to fail. Therefore, I removed it.

core/dbt/cli/requires.py
```diff
@@ -41,14 +41,15 @@ def wrapper(*args, **kwargs):
         ctx.obj["flags"] = flags
         set_flags(flags)
 
-        # Tracking
         initialize_from_flags(flags.SEND_ANONYMOUS_USAGE_STATS, flags.PROFILES_DIR)
```
MichelleArk, line 44: does this (sets up active_user) need to happen logger initialization, or just before tracking? If it's only necessary for tracking, let's move it down where tracking is initialized for readability

peterallenwebb: Sounds good.

PR #7325, opened 2023-04-11
compile inline query doesn't add node

ChenyuLInx:
Related to #6358
When compile a inline node, we add one additional node to the manifest. That node should be removed when compile is done.


ChenyuLInx:
@stu-k Once your PR is merged I will update this to point to main

core/dbt/lib.py
```diff
@@ -1,10 +1,5 @@
 import os
 from dbt.config.project import Project
-from dbt.contracts.results import RunningStatus, collect_timing_info
```
ChenyuLInx, line 3: This whole file is removed on main. The diff will disappear once removed

core/dbt/task/compile.py
```diff
@@ -40,6 +40,10 @@ def compile(self, manifest):
 
 
 class CompileTask(GraphRunnableTask):
+    # We add a new inline node to the manifest during initialization
```
ChenyuLInx, line 43: This is the only change.

stu-k (approved):

PR #7333, opened 2023-04-12
Adding a new column is not a breaking contract change

### LINKED ISSUE #7332: [CT-2401] [Bug] Adding a new column to a model with an enforced contract should not be a breaking change
Reported in https://github.com/dbt-labs/dbt-core/issues/7065#issuecomment-1505047614

### Reproduction case

```sql
-- models/sometable.sql
select 1 as id
```

```yml
models:
  - name: sometable
    config:
      contract:
        enforced: true
    columns:
      - name: order_id
        data_type: number
      - name: order_name
        data_type: varchar
```
Run:
```
$ mkdir state
$ dbt ls
```

Change the yaml to:
```yml
models:
  - name: sometable
    config:
      contract:
        enforced: true
    columns:
      - name: order_id
        data_type: number
      - name: order_name
        data_type: varchar
      - name: categories
        data_type: varchar
```

```
$ dbt ls -s state:modified --state state/
10:38:49  Running with dbt=1.5.0-b5
10:38:50  Found 2 models, 0 tests, 0 snapshots, 1 analysis, 536 macros, 0 operations, 1 seed file, 0 sources, 0 exposures, 0 metrics, 0 groups
10:38:50  Encountered an error:
Contract Error in model sometable (models/sometable.sql)
  There is a breaking change in the model contract because column definitions have changed; you may need to create a new version. See: https://docs.getdbt.com/docs/collaborate/publish/model-versions
```

### Also

Let's rename [`ModelContractError`](https://github.com/dbt-labs/dbt-core/blob/main/core/dbt/exceptions.py#L210) to `ModelContractBreakingChangeError`, to better disambiguate it from [`ContractError`](https://github.com/dbt-labs/dbt-core/blob/main/core/dbt/exceptions.py#L2144) (when the model's contract fails at build time).

jtcohen6:
resolves #7332

### Description

If there's a mismatch in the `checksum`, then we need to dig into _why_ to understand if the change is actually breaking. I'm not sure if this is the most performant way to do it, but it yields the intended functionality.

---

I also renamed the exception from `ModelContractError` to `ContractBreakingChangeError`, and reworked the message that we display to users:
```
$ dbt ls -s state:modified --state state/
11:31:16  Running with dbt=1.5.0-b5
11:31:17  Found 2 models, 0 tests, 0 snapshots, 1 analysis, 536 macros, 0 operations, 1 seed file, 0 sources, 0 exposures, 0 metrics, 0 groups
11:31:17  Encountered an error:
Breaking Change to Contract Error in model sometable (models/sometable.sql)
  While comparing to previous project state, dbt detected a breaking change to an enforced contract.

  The contract's enforcement has been disabled.

  Columns were removed:
   - order_name

  Columns with data_type changes:
   - order_id (number -> int)

  Consider making an additive (non-breaking) change instead, if possible.
  Otherwise, create a new model version: https://docs.getdbt.com/docs/collaborate/publish/model-versions
```

I first did this with fancier string formatting within `CompiledNode.same_contract`, but I added another commit that took the approach of storing more structured information on `ContractBreakingChangeError`, and pretty-formatting the message there.

### Checklist

### DIFF
--- a/core/dbt/contracts/graph/nodes.py
+++ b/core/dbt/contracts/graph/nodes.py
@@ -539,6 +539,7 @@ class CompiledNode(ParsedNode):
             self.contract.checksum = hashlib.new("sha256", data).hexdigest()
 
     def same_contract(self, old) -> bool:
+        # If the contract wasn't previously enforced
         if old.contract.enforced is False and self.contract.enforced is False:
             # Not a change
             return True
@@ -546,29 +547,40 @@ class CompiledNode(ParsedNode):
             # A change, but not a breaking change
             return False
 
+        # Otherwise: the contract was previously enforced
+        self.build_contract_checksum()
+
+        # If the checksums match up, the contract has not changed, so same_contract: True
+        if self.contract.checksum == old.contract.checksum:
+            return True
+
+        # The checksums don't match up, so there has been a change.
+        # We need to determine if it's a **breaking** change.
+        # These are the categories of breaking changes:
         contract_enforced_disabled: bool = False
         columns_removed: List[str] = []
         column_type_changes: List[Tuple[str, str, str]] = []
+
         if old.contract.enforced is True and self.contract.enforced is False:
-            # Breaking change: throw an error
-            # Note: we don't have contract.checksum for current node, so build
+            # Breaking change: the contract was previously enforced, and it no longer is
+            # Note: we don't have contract.checksum for current node, so build it now
             self.build_contract_checksum()
             contract_enforced_disabled = True
 
-        if self.contract.checksum != old.contract.checksum:
-            for key, value in sorted(old.columns.items()):
-                # Has this column been removed?
-                if key not in self.columns.keys():
-                    columns_removed.append(value.name)
-                # Has this column's data type changed?
-                elif value.data_type != self.columns[key].data_type:
-                    column_type_changes.append(
-                        (str(value.name), str(value.data_type), str(self.columns[key].data_type))
-                    )
-                # Otherwise, this was an additive change -- not breaking
-                else:
-                    continue
+        for key, value in sorted(old.columns.items()):
+            # Has this column been removed?
+            if key not in self.columns.keys():
+                columns_removed.append(value.name)
+            # Has this column's data type changed?
+            elif value.data_type != self.columns[key].data_type:
+                column_type_changes.append(
+                    (str(value.name), str(value.data_type), str(self.columns[key].data_type))
+                )
+            # Otherwise, this was an additive change -- not breaking
+            else:
+                continue
 
+        # Did we find any changes that we consider breaking? If so, throw an error
         if contract_enforced_disabled or columns_removed or column_type_changes:
             raise (
                 ContractBreakingChangeError(
@@ -578,9 +590,10 @@ class CompiledNode(ParsedNode):
                     node=self,
                 )
             )
+
+        # Otherwise, the contract has still changed, so same_contract: False
         else:
-            # no breaking changes
-            return True
+            return False
 
 
 # ====================================
--- a/core/dbt/contracts/graph/nodes.py
+++ b/core/dbt/contracts/graph/nodes.py
@@ -546,42 +546,36 @@ class CompiledNode(ParsedNode):
             # A change, but not a breaking change
             return False
 
-        breaking_change_reasons = []
+        contract_enforced_disabled: bool = False
+        columns_removed: List[str] = []
+        column_type_changes: List[Tuple[str, str, str]] = []
         if old.contract.enforced is True and self.contract.enforced is False:
             # Breaking change: throw an error
             # Note: we don't have contract.checksum for current node, so build
             self.build_contract_checksum()
-            breaking_change_reasons.append("The contract has been disabled.")
+            contract_enforced_disabled = True
 
         if self.contract.checksum != old.contract.checksum:
-            columns_removed = []
-            column_type_changes = []
             for key, value in sorted(old.columns.items()):
                 # Has this column been removed?
                 if key not in self.columns.keys():
-                    columns_removed.append(key)
+                    columns_removed.append(value.name)
                 # Has this column's data type changed?
                 elif value.data_type != self.columns[key].data_type:
                     column_type_changes.append(
-                        f"{key} ({value.data_type} -> {self.columns[key].data_type})"
+                        (str(value.name), str(value.data_type), str(self.columns[key].data_type))
                     )
                 # Otherwise, this was an additive change -- not breaking
                 else:
                     continue
 
-            if columns_removed:
-                columns_removed_str = "\n  - ".join(columns_removed)
-                breaking_change_reasons.append(f"Columns were removed: \n - {columns_removed_str}")
-            if column_type_changes:
-                column_type_changes_str = "\n  - ".join(column_type_changes)
-                breaking_change_reasons.append(
-                    f"Columns with data_type changes: \n - {column_type_changes_str}"
-                )
-
-        if breaking_change_reasons:
+        if contract_enforced_disabled or columns_removed or column_type_changes:
             raise (
                 ContractBreakingChangeError(
-                    reasons="\n\n".join(breaking_change_reasons), node=self
+                    contract_enforced_disabled=contract_enforced_disabled,
+                    columns_removed=columns_removed,
+                    column_type_changes=column_type_changes,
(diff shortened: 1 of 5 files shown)


gshank (approved):

core/dbt/exceptions.py
```diff
@@ -207,22 +207,44 @@ def _fix_dupe_msg(self, path_1: str, path_2: str, name: str, type_name: str) ->
             )
 
 
-class ModelContractError(DbtRuntimeError):
+class ContractBreakingChangeError(DbtRuntimeError):
     CODE = 10016
-    MESSAGE = "Contract Error"
+    MESSAGE = "Breaking Change to Contract"
```
jtcohen6, line 212: I changed this, but I don't actually know where this `MESSAGE` appears. Should, this and the `type` below, be `Contract Breaking Change Error` for consistency with the exception class name?

core/dbt/contracts/graph/nodes.py
```diff
@@ -546,19 +546,38 @@ def same_contract(self, old) -> bool:
...
-            breaking_change_reasons.append("contract has been disabled")
+            contract_enforced_disabled = True
 
         if self.contract.checksum != old.contract.checksum:
-            # Breaking change, throw error
-            breaking_change_reasons.append("column definitions have changed")
-
-        if breaking_change_reasons:
-            raise (ModelContractError(reasons=" and ".join(breaking_change_reasons), node=self))
+            for key, value in sorted(old.columns.items()):
```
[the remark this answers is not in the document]
gshank: Yes. Something like this was always going to have to happen for constraints.

core/dbt/exceptions.py
gshank: A few exceptions use it to construct the actual "message", but mostly it's not. Another thing we might want to clean up at some point. So I don't think it matters here.

core/dbt/contracts/graph/nodes.py
```diff
@@ -546,19 +546,38 @@ def same_contract(self, old) -> bool:
...
+                ContractBreakingChangeError(
+                    contract_enforced_disabled=contract_enforced_disabled,
+                    columns_removed=columns_removed,
+                    column_type_changes=column_type_changes,
+                    node=self,
+                )
+            )
         else:
             # no breaking changes
             return True
```
MichelleArk, line 583: It might just be a confusing comment - but this method should return True when there are _no_ changes (breaking or otherwise) to the contract, and False if there is a change to the contract but it is non-breaking. Are we actually handling the latter case?

gshank: It does look like the logic isn't right here.  I think we need to keep track of adds and return False, so we need "elif columns_added, return False, else True.

jtcohen6: I'm taking a look at this logic now. It's genuinely tricky. I'll add more inline comments.

jtcohen6: I inverted the logic a bit, so now it's:
- Was the contract being enforced previously?
    - no &rarr; return `False` or `True`, but never `ContractBreakingChangeError`
- If so, do the checksums match up?
    - yes &rarr; `return True`
- If the checksums don't match up, are there any changes we consider breaking?
    - breaking &rarr; throw a `ContractBreakingChangeError` with details
    - non-breaking, still a change &rarr; `return False`
    
Let me know if you think that makes more or less sense than before

core/dbt/contracts/graph/nodes.py
```diff
@@ -539,29 +539,61 @@ def build_contract_checksum(self):
...
+        contract_enforced_disabled: bool = False
+        columns_removed: List[str] = []
+        column_type_changes: List[Tuple[str, str, str]] = []
+
         if old.contract.enforced is True and self.contract.enforced is False:
-            # Breaking change: throw an error
-            # Note: we don't have contract.checksum for current node, so build
+            # Breaking change: the contract was previously enforced, and it no longer is
+            # Note: we don't have contract.checksum for current node, so build it now
             self.build_contract_checksum()
```
MichelleArk, line 567: have we not already done this up [here](https://github.com/dbt-labs/dbt-core/pull/7333/files#diff-42cfc99b2fcd89ab75957b8999bee2b1af822f7d39953ca99f1f285c170557c1R551)?

MichelleArk (approved):
Just one question, not necessarily blocking. Thanks so much for fixing this ✨

core/dbt/contracts/graph/nodes.py
```diff
@@ -539,29 +539,61 @@ def build_contract_checksum(self):
...
             # A change, but not a breaking change
             return False
 
-        breaking_change_reasons = []
+        # Otherwise: the contract was previously enforced
+        self.build_contract_checksum()
+
+        # If the checksums match up, the contract has not changed, so same_contract: True
+        if self.contract.checksum == old.contract.checksum:
+            return True
```
gshank, line 555: I think this is wrong. If old.contract.enforced is True and self.contract.enforced is False it's a breaking change even when checksums are still equal, and this will always return True for that case. Also the "build_contract_checksum" only needs to run for self.contract.enforced is False, and the line above runs it for both True and False.

core/dbt/contracts/graph/nodes.py
jtcohen6: I'm pretty sure we can remove both calls to `build_contract_checksum()`. It's being called within [`NodePatchParser.parse_patch`](https://github.com/dbt-labs/dbt-core/blob/main/core/dbt/parser/schemas.py#L999)

gshank: Oh, that's a change. Initially I only had it building when contract.enforced is True, and it's been changed so that it always builds.

gshank: Actually no, it's "build_contract_checksum" that checks for contract enforced. So it is a no-op, but it probably shouldn't be, because Michelle wanted to also capture that there were additional changes to the columns. So we should remove the check for contract.enforced in build_contract_checksum and only call it in parse_patch if contract.enforced is True, and THEN it won't be a no-op.

core/dbt/contracts/graph/nodes.py
jtcohen6: > If old.contract.enforced is True and self.contract.enforced is False it's a breaking change even when checksums are still equal, and this will always return True for that case.

Yes! Good point! Let me add a test for this

> Also the "build_contract_checksum" only needs to run for self.contract.enforced is False, and the line above runs it for both True and False.

Per thread above, I don't think we need to call `build_contract_checksum` at any point within `same_contract`

core/dbt/contracts/graph/nodes.py
gshank: Boy this logic makes your head hurt.

core/dbt/contracts/graph/nodes.py
gshank: Per other thread, we want to capture that the contract has been disabled AND the columns are different.

jtcohen6: Yeah, the previous way of expressing this logic does make more sense. I'm going to switch it back.

jtcohen6: > If old.contract.enforced is True and self.contract.enforced is False it's a breaking change even when checksums are still equal, and this will always return True for that case. 

After looking more closely, the current formulation of the logic **does** work, because `self.contract.checksum` is `None`!

```python
ipdb> self.contract.checksum == old.contract.checksum
False
ipdb> self.contract
Contract(enforced=False, checksum=None)
ipdb> old.contract
Contract(enforced=True, checksum='1698cf5a415f00ae0dee2c6e97bb51c020d46955847b2e9cec53a8e40d1afb13')
```

core/dbt/contracts/graph/nodes.py
jtcohen6: @gshank Right 😅

We want the full column spec to be captured in the checksum, so that additions would cause a checksum mismatch — but if the checksums don't match, we then need to dig in and understand whether a column was actually removed/changed, or just added. I'm happy with the current implementation for now, but we can open up another ticket to revisit this logic.

jtcohen6:
I'm going to merge this for inclusion in v1.5.0-rc1. Even if the logic is a bit verbose, and perhaps still not 100% perfect, we have better test coverage now than before, and it feels like a step closer to the behavior that we want.

PR #7334, opened 2023-04-12
ADAP-402: Add configuration change option - postgres

mikealfare:

### Description

Configure the plumbing for the update scenario for materialized views.

- Update index > drop/create index
- Else > full refresh

### Checklist

- [x] I have read [the contributing guide](https://github.com/dbt-labs/dbt-redshift/blob/main/CONTRIBUTING.md) and understand what's expected of me
- [ ] I have run `changie new` to [create a changelog entry](https://github.com/dbt-labs/dbt-redshift/blob/main/CONTRIBUTING.md#Adding-CHANGELOG-Entry)


plugins/postgres/dbt/include/postgres/macros/materializations/materialized_view.sql
```diff
@@ -0,0 +1,30 @@
...
+
+    alter view {{ intermediate_relation }} rename to {{ relation.include(database=False, schema=False) }};
+
+{% endmacro %}
+
+
+{% macro postgres__get_alter_materialized_view_sql(relation, updates, sql, existing_relation, backup_relation, intermediate_relation) %}
+    {% if 'index' in updates.keys() %}
+    --  should this be placed lower into the chain somehow to be below call to main? othe
+        {% do create_indexes(target_relation) %}
```
McKnight-42, line 26: I am a little concerned about this placement most other materializations that call do it a little below the call to `main` for statement builds but not sure if we want this just in the base materialization or to recreate the materialization for this small addition in postgres

McKnight-42:
this test `test_on_configuration_change_fails_with_update` is the only test currently failing due to `expect_pass='fail` being false and having a `RunStatus` of `Success` instead of `Error` or anything else, this may change when we swap over to `materializedview` over `view` on the target_relation. so would we want to skip for now?

plugins/postgres/dbt/include/postgres/macros/materializations/materialized_view.sql
```diff
@@ -27,11 +42,28 @@
 {% endmacro %}
 
 
-{% macro default__get_materialized_view_configuration_changes(existing_relation, new_config) %}
-    {%- do return([]) -%}
+{% macro postgres__get_materialized_view_configuration_changes(existing_relation, new_config) %}
```
McKnight-42, line 45: @mikealfare curious to what ideas you have for this would you want to move it to a impl method?

plugins/postgres/dbt/include/postgres/macros/materializations/materialized_view.sql
mikealfare: Oh, good point, in other materializations we call the macro to create the indexes after the object is created. However, that's the change that we're looking for in this case, so we also call it during the build. I think we would need to put that in the create and replace steps then, in the postgres versions. Let's leave that for the native implementation story since creating indexes on the proxy view doesn't make sense.

PR #7336, opened 2023-04-12
ADAP-395: Implement native materialized view DDL

mikealfare:

### Description

Replace all stubs with actual Redshift DDL for materialized views.

### Checklist

- [x] I have read [the contributing guide](https://github.com/dbt-labs/dbt-redshift/blob/main/CONTRIBUTING.md) and understand what's expected of me
- [ ] I have run `changie new` to [create a changelog entry](https://github.com/dbt-labs/dbt-redshift/blob/main/CONTRIBUTING.md#Adding-CHANGELOG-Entry)


McKnight-42:
@mikealfare macro `drop_relation_if_exists` which we call on the materialziation calls `drop materialized_viiew...` instead of `drop materialized view ...` due to macro `drop_relation` calling for `relation.type`

mikealfare:
> @mikealfare macro `drop_relation_if_exists` which we call on the materialziation calls `drop materialized_viiew...` instead of `drop materialized view ...` due to macro `drop_relation` calling for `relation.type`
> 
> should we change the string from RelationType to `materialized view` from `matieralized_view`?

Oof. I don't know if I want to put a space in that. I'm not sure where else it gets used. But I'm also not too keen on updating `drop_relation()` to account for mapping `_` to `' '`. Let's ask around first.

McKnight-42:
> > @mikealfare macro `drop_relation_if_exists` which we call on the materialziation calls `drop materialized_viiew...` instead of `drop materialized view ...` due to macro `drop_relation` calling for `relation.type`
> > should we change the string from RelationType to `materialized view` from `matieralized_view`?
> 
> Oof. I don't know if I want to put a space in that. I'm not sure where else it gets used. But I'm also not too keen on updating `drop_relation()` to account for mapping `_` to `' '`. Let's ask around first.

From quick glance at core code looks like `relation.type` almost acts as a string replacement for whatever the set type is in general or used to assert what the type should be.

McKnight-42 (approved):

PR #7337, opened 2023-04-12
remove providing profile/project to dbtRunner

ChenyuLInx:
Given that dbt-core doesn't provide a way to load profile/project, and the fact that we are uncertain about the interface that's gonna be provided here. We are removing the functionality to pass those objects to `dbtRunner`.

### DIFF
--- a/core/dbt/cli/main.py
+++ b/core/dbt/cli/main.py
@@ -15,8 +15,6 @@ from dbt.cli.exceptions import (
     DbtInternalException,
     DbtUsageException,
 )
-from dbt.config.profile import Profile
-from dbt.config.project import Project
 from dbt.contracts.graph.manifest import Manifest
 from dbt.contracts.results import (
     CatalogArtifact,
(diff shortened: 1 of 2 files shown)


core/dbt/cli/main.py
```diff
@@ -64,13 +62,9 @@ class dbtRunnerResult:
 class dbtRunner:
     def __init__(
         self,
-        project: Project = None,
-        profile: Profile = None,
```
stu-k, line 68: Does removing these lines break anything outside of `dbt-core`?

stu-k (approved):
Nothing stands out, one question about apps with a dependency on `dbtRunner` that are already providing `project` and `profile`.

core/dbt/cli/main.py
ChenyuLInx: No, did a full look up on places I know uses this runner and they don't use it at all.

Issue #7348, opened 2023-04-13
[CT-2409] [Bug] FQN selection for versioned models in subdirectory


jtcohen6:
If a versioned model is **nested in one or more subdirectories**, it can't be selected by `name`, `v{v}`, etc.

**Expected behavior:**
```
dbt list -s dim_customers    --> all
dbt list -s dim_customers.*  --> all
dbt list -s dim_customers.v1 --> v1 only
dbt list -s *.v1             --> any model with v1
```

**Current behavior:**
None of the above work, but it does work when I add in the `*` wildcards:
```
dbt ls -s *.dim_customers.*  --> all
dbt ls -s *.dim_customers.v1 --> v1 only
```

~Weirdly, I could've sworn this was working when I was trying out #7287. Worth adding some tests.~ Realized that this only happens for a model defined in a subdirectory, rather than top-level in the `models/` folder.

PR #7351, opened 2023-04-13
[Fix] fix versioned model selection in subdirectories

MichelleArk:
resolves https://github.com/dbt-labs/dbt-core/issues/7348 

### Checklist

### DIFF
--- a/core/dbt/graph/selector_methods.py
+++ b/core/dbt/graph/selector_methods.py
@@ -53,13 +53,18 @@ class MethodName(StrEnum):
     Version = "version"
 
 
-def is_selected_node(
-    fqn: List[str],
-    node_selector: str,
-) -> bool:
+def is_selected_node(fqn: List[str], node_selector: str, is_versioned: bool) -> bool:
     # If qualified_name exactly matches model name (fqn's leaf), return True
-    if fqn[-1] == node_selector:
-        return True
+    if is_versioned:
+        flat_node_selector = node_selector.split(".")
+        if fqn[-2] == node_selector:
+            return True
+        # If this is a versioned model, then the last two segments should be allowed to exactly match
+        elif fqn[-2:] == flat_node_selector[-2:]:
+            return True
+    else:
+        if fqn[-1] == node_selector:
+            return True
     # Flatten node parts. Dots in model names act as namespace separators
     flat_fqn = [item for segment in fqn for item in segment.split(".")]
     # Selector components cannot be more than fqn's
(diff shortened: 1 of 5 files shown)


tests/functional/graph_selection/test_schema_test_graph_selection.py
```diff
@@ -12,7 +12,7 @@ def run_schema_and_assert(project, include, exclude, expected_tests):
     run_dbt(["deps"])
     run_dbt(["seed"])
     results = run_dbt(["run", "--exclude", "never_selected"])
-    assert len(results) == 10
+    assert len(results) == 12
```
MichelleArk, line 15: added 2 (non-ephemeral) models to the fixtures [here](https://github.com/dbt-labs/dbt-core/pull/7351/files#diff-a12c5c36034799ba76ea9118119d6cc4966a66714899dea076f76d1ffaf5c31fR217-R221)

core/dbt/graph/selector_methods.py
```diff
@@ -320,7 +326,7 @@ def search(self, included_nodes: Set[UniqueId], selector: str) -> Iterator[Uniqu
                 ymlfp = Path(pfp)
                 if ymlfp in paths:
                     yield node
-            elif any(parent in paths for parent in ofp.parents):
+            if any(parent in paths for parent in ofp.parents):
```
MichelleArk, line 329: this `elif` was not being hit if a node had a `patch_path`. Given these 3 conditions are all independent predicates for path selection matching, separate if statements seemed appropriate to me.

dave-connors-3:
LGTM!

dave-connors-3 (approved):

core/dbt/graph/selector_methods.py
```diff
@@ -53,13 +53,18 @@ class MethodName(StrEnum):
...
     # If qualified_name exactly matches model name (fqn's leaf), return True
-    if fqn[-1] == node_selector:
-        return True
+    if is_versioned:
+        flat_node_selector = node_selector.split(".")
+        if fqn[-2] == node_selector:
+            return True
+        # If this is a versioned model, then the last two segments should be allowed to exactly match
+        elif fqn[-2:] == flat_node_selector[-2:]:
+            return True
```
jtcohen6, line 64: Nice! Ok, this roughly approximates the logic in the docs site now (https://github.com/dbt-labs/dbt-docs/pull/411).

The only differences:
- In the docs site we support selecting as `model_name_v1`, in addition to `model_name.v1`, simply because that's what we visually display in the docs site. This doesn't feel like a massive inconsistency to me.
- In the docs site we support selecting with _just_ `v1`. We could support that here too, by allowing `fqn[-1] == node_selector` for versioned models as well (rather than in an `else` clause below). But this works if I select with the wildcard `*.v1`, and tbh I almost like that better (more explicit).

In both cases, I think I prefer the behavior here, and it should define the standard. The small differences in `dbt-docs` are just there for expediency.

jtcohen6 (approved):

PR #7358, opened 2023-04-13
CT-2414: Add graph summaries to target directory output

### LINKED ISSUE #7357: [CT-2414] Output Graph Summary Information on When dbt Compiles
The `manifest.json` and `graph.gpickle` files would often be useful for investigating performance issues in Core's graph algorithms, but there are a few of issues:
- They may be large
- They often contain sensitive proprietary information that cannot be shared internally
- They only record graph structure _after_ test edges are added

I propose that we add an additional output file called `graph_summary.json`, which would be more compact and anonymized. This file would contain only the edge structure, and the name and type of each node. It will include that information at two separate points in time: Immediately after the graph is linked together, and then after test edges have been added.

My hope is that this file could be shared with the Core team on a routine basis, and would allow us to build and maintain a library of realistic graph structures from production environments, helping us improve and test our graph algorithms.

Immediate applications include:

- Building a library of user DAGs for analysis and testing
- Analyzing whether or not we are executing DAGs as quickly as we reasonably could
- Checking whether add_test_edges() is doing what we want, and optimizing it

peterallenwebb:
resolves #7357 

### Description

Adds two new graph summary output files to the target directory output.

### Checklist


dbeatty10:
To keep the number of artifacts to a minimum, could we just write out `graph_with_test_edges.json`?

From there, consumers could filter out any resource types they don't want.

I know we started to discuss that idea [here](https://github.com/dbt-labs/dbt-core/issues/7357#issuecomment-1508634328), but now that you have this pull request, it's probably useful to move that conversation to here.

dbeatty10:
@peterallenwebb Would you be open to just a single JSON artifact that is a superset of both artifacts?

That way, there's just one artifact to produce/consumer, and the end user can filter out any nodes they want to ignore.

Some benefits is that it would make the docs and implementation more simple.

peterallenwebb:
@dbeatty10 I am just getting back to working on this, and not disregarding your earlier request, I promise. Using a single file will complicate the implementation, especially since the tests are added on one code path but not another, so I'm trying to figure out how to keep both the implementation and the output simple.

I think I am going to spin the snowplow stats off as a separate case, though.

dbeatty10:
No worries @peterallenwebb !

Spinning off the snowplow stats seems like nice effort that is complementary.

I was mainly wondering what would be lost if you just dropped this line:
```
linker.write_graph_summary("linked", out_stream, manifest)
```

Is it because `graph_with_test_edges.json` would _not_ be a simple superset of `graph.json`?

peterallenwebb:
@dbeatty10 There is some debate about whether the algorithm for adding the test edges is completely correct, and we know it has performance issues that we'd like to address, so my idea here is to make sure we know exactly what the graph looks like before and after that algorithm runs. That way we'll have a known starting point for refining and/or optimizing that algorithm.

I'm going to take one more pass at making this simpler.

iknox-fa (approved):
LGTM-- I bet we'll want to make some tweaks at some point (esp wrt test edges), but this seems like a very sane starting place.

peterallenwebb:
@boxysean You may find this interesting.

PR #7371, opened 2023-04-17
Clear cached `statement` results when retrieved from `ProviderContext` and subclasses

### LINKED ISSUE #7281: [CT-2377] [Spike] Performance improvement to memory footprint
When a macro runs a call statement with `fetch_result` set to `True`, that result may be held in memory longer than it needs to be.

iknox-fa:
resolves #7281 

### Description
Reduces the memory footprint of `statement` calls by removing the cached results from the `ProviderContext` subclass when `load_result` is called except when the macro is named "main" (special case).

Note: If we decide to merge this we'll want to update the docs site to mention that a `{% set foo = load_result("bar") %}` can only be done _once_ per value of bar.

### Checklist


core/dbt/context/providers.py
```diff
@@ -732,12 +733,28 @@ def invocation_args_dict(self):
...
+            elif self.sql_results[name] is None:
+                raise MacroResultAlreadyLoadedError(name)
+
+            # Handle the regular use case
+            else:
+                ret_val = self.sql_results[name]
+                self.sql_results[name] = None
+                return ret_val
+        else:
+            # Handle trying to load a result that was never stored
```
iknox-fa, line 756: @jtcohen6 Just pointing out that this, while it maintains parity to the previous code, doesn't seem like the best way to handle calls to `load_result` in instances where there is no matching key in `sql_results`.  Should we throw an error or something else here or have people possibly baked this behavior into their workflows and we should just let it slide?

jtcohen6: You raise a good point! After giving it a bit of thought, I do think we want this to keep returning `None`, rather than raising, because we do a first pass through this code **at parse time** — at which time, the `statement` has not been called, and so there's no matching value in `sql_results`.

Basically, this would start raising during parsing:
```sql
{% call statement('my_statement') %}
  select 1 as id
{% endcall %}
{% set res = load_result('my_statement') %}
```
We could insist that users stick the whole thing in an `if execute` block:
```
{% if execute %}
    {% call statement('my_statement') %}
      select 1 as id
    {% endcall %}
    {% set res = load_result('my_statement') %}
{% else %}
    {% set res = ... %}
{% endif %}
```
But that feels like putting more of the brunt on them — even if it would make it easier to catch instances where you just had a typo, e.g. `load_result('my_statment')`.

(Through a bit of cleverness, even if users try to operate on, or access values of the result while it's `None` - e.g. `res['table']` - at parse time, we catch and ignore those Jinja Undefined exceptions. I think.)

core/dbt/exceptions.py
```diff
@@ -1019,6 +1019,17 @@ def get_message(self) -> str:
         return msg
 
 
+class MacroResultAlreadyLoadedError(CompilationError):
+    def __init__(self, result_name):
+        self.result_name = result_name
+        super().__init__(msg=self.get_message())
+
+    def get_message(self) -> str:
+        msg = f'The result "{self.result_name}" has already been loaded into a variable'
```
jtcohen6, line 1028: Let's mention the word `statement` explicitly, since that should give users an extra hint. We're pretty inconsistent about how we quote variables in our exception/logging messages today. I tend to opt for single quotes or backticks.
```suggestion
        msg = f"The 'statement' result named '{self.result_name}' has already been loaded into a variable"
```

jtcohen6 (commented):
nice work! thanks for digging into this :)

core/dbt/context/providers.py
iknox-fa: Ok sounds good-- We'll leave it as is.

gshank (approved):
Looks okay to me.

PR #7383, opened 2023-04-18
dbt.include import issue for ci

mikealfare:
### Description

The code quality check is not properly installing `dbt-core`. This fixes that.

### Checklist

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -54,7 +54,7 @@ jobs:
           python -m pip install --user --upgrade pip
           python -m pip --version
           make dev_req
-          make dev
+          pre-commit install
           mypy --version
           dbt --version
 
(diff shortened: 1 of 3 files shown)


.github/workflows/main.yml
```diff
@@ -53,6 +53,7 @@ jobs:
         run: |
           python -m pip install --user --upgrade pip
           python -m pip --version
+          make dev_req
           make dev
```
mikealfare, line 57: This (`make dev`) should include `make dev_req`. And when both are run, the logs show an attempt to install packages a second time. However, adding this creates a successful installation. The main difference I can see between the two is that running both commands results in running `setup.py develop` for the three editable packages. That isn't explicitly called out when only running `make dev`. This isn't an ideal solution, but it resolves the problem in our CI jobs and makes it slightly more clear what may be happening here.

mikealfare: This was working because we were forcing the editable installs to happen after the dev requirements were done. This is now done explicitly in the recipe for `make dev` and this is no longer relevant.

nathaniel-may (approved):

PR #7412, opened 2023-04-19
fix target-path not working for sources.json

### LINKED ISSUE #7411: [CT-2439] [Bug] dbt source freshness not using target-path specified in dbt_project.yml

### Current Behavior

`dbt source freshness` will write to default running dir instead of being able to specify a target dir.

### Expected Behavior

When a target-path is specified in `dbt_project.yml`, dbt-core would write sources.json to that path.

### Steps To Reproduce

Specify a `target-path` in `dbt_project.yml` that's different from default path. Run source freshness. You will see `sources.json` being written to the target dir in the current path not the path specified in `dbt_project.yml`

ChenyuLInx:
resolves #7411 
Edited test would fail before the fix

### DIFF
--- a/.changes/unreleased/Fixes-20230419-142150.yaml
+++ b/.changes/unreleased/Fixes-20230419-142150.yaml
@@ -1,0 +1,6 @@
+kind: Fixes
+body: Fix dbt command missing target-path param
+time: 2023-04-19T14:21:50.959786-07:00
+custom:
+  Author: ChenyuLInx
+  Issue: "\t7411"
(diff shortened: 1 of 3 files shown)


gshank (approved):

iknox-fa (approved):
LGTM!

MichelleArk:
@ChenyuLInx - does this need a backport to 1.5?

MichelleArk:
Confirmed this does need a backport offline 👍

PR #7426, opened 2023-04-21
Address mypy issues affecting scheduled CI runs on 1.1.latest

mikealfare:
### Description

- upgrade `mypy` to .961 (the version in `1.2.latest`)
- explicitly add `.core/` to mypy_path to avoid double path error
- add missing types libraries to `dev-requirements.txt`
- add `mypy --install-types` to code quality workflow to install missing types libraries
- address one deprecation warning by using recommended replacement decorator
- address remaining mypy violations by ignoring with `type: ignore

### Checklist

### DIFF
--- a/mypy.ini
+++ b/mypy.ini
@@ -1,3 +1,4 @@
 [mypy]
-mypy_path = ./third-party-stubs
+mypy_path = ./third-party-stubs,./core
 namespace_packages = True
+explicit_package_bases = True
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -54,7 +54,7 @@ jobs:
           pip --version
           pip install pre-commit
           pre-commit --version
-          pip install mypy==0.782
+          pip install mypy==0.961
           mypy --version
           pip install -r editable-requirements.txt
           dbt --version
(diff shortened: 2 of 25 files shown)


mypy.ini
```diff
@@ -1,3 +1,4 @@
 [mypy]
-mypy_path = ./third-party-stubs
+mypy_path = ./third-party-stubs,./core
```
MichelleArk, line 2: If you still have it handy - what was the message from the error that this resolved? I'm curious why we don't need to do the same for later versions of core

MichelleArk (approved):

mypy.ini
mikealfare: I think we added or removed an init file in the later versions. I can't recall verbatim, but the issue was that files could be found via two paths, one starting at `/dbt` and another starting at `/core`; it seems like these were both in $PATH. So it was something like this:

```
Module found with multiple paths: `/core/dbt/adapters/base/meta.py` and `/dbt/adapters/base/meta.py`.
```

But a side effect seemed to be that we were not type checking `/core/dbt/` at all. Because once I turned it on, I got 46 errors in 30 files (I guess I remember numbers better than words).

Issue #5193, opened 2022-04-29
[CT-584] [Bug] `dbt build` doesn't respect n-plus operator


foundinblank:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

I feel I might’ve run into a node selection bug when using `dbt run` vs `dbt build`. My understanding is that dbt build is supposed to respect [standard selection syntax](https://docs.getdbt.com/reference/commands/build) but it does not seem to be handling n-plus selection correctly. 

The behavior I’m seeing:
`dbt run -s 3+dim_users`  - only runs dim_users and 1st, 2nd, and 3rd degree upstream models
`dbt build -s 3+dim_users` - runs all upstream models, as if it’s just doing `+dim_users`

### Expected Behavior

`dbt build -s 3+dim_users` should run only `dim_users` and 1st, 2nd, and 3rd degree upstream models. 

### Steps To Reproduce

1. Be in a regular dbt environment with a basic DAG
2. Run `dbt run -s 1+model_name` - the output should be that dbt runs only `model_name` and all 1st degree upstream models
3. Run `dbt build -s 1+model_name` - the output is that dbt runs _all_ upstream models

### Environment

```markdown
- OS: macOS Big Sur 11.6
- Python: 3.7.9
- dbt: 1.0.1
```

### What database are you using dbt with?

snowflake

iknox-fa:
Hi @foundinblank , thanks for the bugreport.  Unfortunately, I'm having a hard time replicating the issue locally.  Could you provide a sample project so I can try to sort out what's going wrong?

patkearns10:
tested this on 1.4 and can confirm: https://www.loom.com/share/2dd237b0da834a6589211a8e23c8d77f

Confirmed this is happening in the IDE and locally cli.
`dbt run -s 2+some_model`  and `dbt build -s 2+some_model` return a different number of models
Confirmed this is also happening in the other direction, ie. 
`dbt run -s some_model+2`  and `dbt build -s some_model+2` return a different number of models

ttusing:
I can also confirm this behavior.

jtcohen6:
From a little poking around - I think the fix in https://github.com/dbt-labs/dbt-core/pull/7431 could address this one too! By filtering out the extra edges that we add during `build`, to block/skip downstream models on upstream models' tests.

PR #7431, opened 2023-04-21
Ignore parent tests added edges for build selection

### LINKED ISSUE #7289: [CT-2381] [Bug] `state:modified+` selection doesn't behave as expected for `dbt build` when tests are modified

### Current Behavior

`dbt build --select state:modified+` should behave like the other commands.

Specifically the following scenario doesn't yield the expected selection:
- modify a single/bunch of tests (but only tests)
- `PREVIOUS_TARGET=~/previous_target`
- `cp -r $DBT_PROJECT_DIR/target/* $PREVIOUS_TARGET`
- run `dbt list --select state:modified+ --state $PREVIOUS_TARGET`  => correctly lists only the changed tests (nothing downstream)
- run `dbt build --select state:modified+ --state $PREVIOUS_TARGET`  => runs the changed tests but also ALL downstream nodes!!
- while `dbt run --select state:modified+ --state $PREVIOUS_TARGET` does as expected: "Nothing to run"
- and `dbt test --select state:modified+ --state $PREVIOUS_TARGET` as well: only runs the changed tests.

### Expected Behavior

`dbt build --select state:modified+` should NOT run models downstream from modified tests (if only the tests have been modified, not the tests' model(s) )

### Steps To Reproduce

- modify a single/bunch of tests (but only tests)
- `PREVIOUS_TARGET=~/previous_target`
- `cp -r $DBT_PROJECT_DIR/target/* $PREVIOUS_TARGET`
- run `dbt list --select state:modified+ --state $PREVIOUS_TARGET`  => correctly lists only the changed tests (nothing downstream)
- run `dbt build --select state:modified+ --state $PREVIOUS_TARGET`  => runs the changed tests but also ALL downstream nodes!!
- while `dbt run --select state:modified+ --state $PREVIOUS_TARGET` does as expected: "Nothing to run"
- and `dbt test --select state:modified+ --state $PREVIOUS_TARGET` as well: only runs the changed tests.

### Relevant log output

_No response_

### Environment

```markdown
- OS: PopOS 22.04
- Python: 3.7
- dbt: 1.2.5
```

snowflake

### Additional Context

_No response_

b-luu:
resolves #7289 #5193

### Description
Removes the added tests edges (specific to build task) for children and/or parent selection.
This allows to avoid selecting any other nodes that aren't selected with the other dbt commands (list, run, ...).
The added tests edges are still respected for the GraphQueue to follow the execution.

### Checklist

- [x] This PR includes tests, ~~or tests are not required/relevant for this PR~~
- [x] ~~I have [opened an issue to add/update docs](https://github.com/dbt-labs/docs.getdbt.com/issues/new/choose), or~~ docs changes are not required/relevant for this PR

### DIFF
--- a/core/dbt/graph/graph.py
+++ b/core/dbt/graph/graph.py
@@ -28,9 +28,10 @@ class Graph:
         """Returns all nodes having a path to `node` in `graph`"""
         if not self.graph.has_node(node):
             raise DbtInternalError(f"Node {node} not found in the graph!")
+        filtered_graph = self.exclude_edge_type("parent_test")
         return {
             child
-            for _, child in nx.bfs_edges(self.graph, node, reverse=True, depth_limit=max_depth)
+            for _, child in nx.bfs_edges(filtered_graph, node, reverse=True, depth_limit=max_depth)
         }
 
     def descendants(
@@ -39,17 +40,19 @@ class Graph:
         """Returns all nodes reachable from `node` in `graph`"""
         if not self.graph.has_node(node):
             raise DbtInternalError(f"Node {node} not found in the graph!")
-        filtered_graph = nx.restricted_view(
+        filtered_graph = self.exclude_edge_type("parent_test")
+        return {child for _, child in nx.bfs_edges(filtered_graph, node, depth_limit=max_depth)}
+
+    def exclude_edge_type(self, edge_type_to_exclude):
+        return nx.restricted_view(
             self.graph,
             nodes=[],
             edges=(
                 (a, b)
                 for a, b in self.graph.edges
-                if not include_tests_children
-                and self.graph[a][b].get("edge_type") == "parent_test"
+                if self.graph[a][b].get("edge_type") == edge_type_to_exclude
             ),
         )
-        return {child for _, child in nx.bfs_edges(filtered_graph, node, depth_limit=max_depth)}
 
     def select_childrens_parents(self, selected: Set[UniqueId]) -> Set[UniqueId]:
         ancestors_for = self.select_children(selected) | selected
--- a/core/dbt/graph/graph.py
+++ b/core/dbt/graph/graph.py
@@ -34,9 +34,7 @@ class Graph:
             for _, child in nx.bfs_edges(filtered_graph, node, reverse=True, depth_limit=max_depth)
         }
 
-    def descendants(
-        self, node: UniqueId, max_depth: Optional[int] = None, include_tests_children: bool = False
-    ) -> Set[UniqueId]:
+    def descendants(self, node: UniqueId, max_depth: Optional[int] = None) -> Set[UniqueId]:
         """Returns all nodes reachable from `node` in `graph`"""
         if not self.graph.has_node(node):
             raise DbtInternalError(f"Node {node} not found in the graph!")
--- a/core/dbt/graph/graph.py
+++ b/core/dbt/graph/graph.py
@@ -33,11 +33,18 @@ class Graph:
             for _, child in nx.bfs_edges(self.graph, node, reverse=True, depth_limit=max_depth)
         }
 
-    def descendants(self, node: UniqueId, max_depth: Optional[int] = None) -> Set[UniqueId]:
+    def descendants(
+        self, node: UniqueId, max_depth: Optional[int] = None, include_tests_children: bool = False
+    ) -> Set[UniqueId]:
         """Returns all nodes reachable from `node` in `graph`"""
         if not self.graph.has_node(node):
             raise DbtInternalError(f"Node {node} not found in the graph!")
-        return {child for _, child in nx.bfs_edges(self.graph, node, depth_limit=max_depth)}
+        return {
+            child
+            for parent, child in nx.bfs_edges(self.graph, node, depth_limit=max_depth)
+            if include_tests_children
+            or self.graph[parent][child].get("edge_type") != "parent_test"
+        }
 
     def select_childrens_parents(self, selected: Set[UniqueId]) -> Set[UniqueId]:
         ancestors_for = self.select_children(selected) | selected
--- a/core/dbt/graph/graph.py
+++ b/core/dbt/graph/graph.py
@@ -39,12 +39,17 @@ class Graph:
         """Returns all nodes reachable from `node` in `graph`"""
         if not self.graph.has_node(node):
             raise DbtInternalError(f"Node {node} not found in the graph!")
-        return {
-            child
-            for parent, child in nx.bfs_edges(self.graph, node, depth_limit=max_depth)
-            if include_tests_children
-            or self.graph[parent][child].get("edge_type") != "parent_test"
-        }
+        filtered_graph = nx.restricted_view(
+            self.graph,
+            nodes=[],
+            edges=(
+                (a, b)
+                for a, b in self.graph.edges
+                if not include_tests_children
+                and self.graph[a][b].get("edge_type") == "parent_test"
+            ),
+        )
+        return {child for _, child in nx.bfs_edges(filtered_graph, node, depth_limit=max_depth)}
 
     def select_childrens_parents(self, selected: Set[UniqueId]) -> Set[UniqueId]:
         ancestors_for = self.select_children(selected) | selected
--- a/.changes/unreleased/Fixes-20230421-172428.yaml
+++ b/.changes/unreleased/Fixes-20230421-172428.yaml
@@ -1,0 +1,6 @@
+kind: Fixes
+body: dbt build selection of tests' descendants
+time: 2023-04-21T17:24:28.335866975+02:00
+custom:
+  Author: b-luu
+  Issue: "7289"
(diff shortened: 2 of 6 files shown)


jtcohen6:
@b-luu Thanks for the PR - nice work! :)

I think [`tests/functional/build/test_build`](https://github.com/dbt-labs/dbt-core/blob/main/tests/functional/build/test_build.py) would be a good place for this. Specifically, a test case that captures the scenario you described as being fixed in this PR:

> `dbt build --select <a_test_of_a_parent_model>+` only runs that single test (without this fix, it also runs all children of the model of the test!!)

Something like:
```python
class TestDownstreamSelection:
    def models(self):
        return {
            "model_a.sql": models_simple_blocking__model_a_sql,
            "model_b.sql": models_simple_blocking__model_b_sql,
            "test.yml": models_simple_blocking__test_yml,
        }

    def test_downstream_selection(self, project):
        """Ensure that selecting test+ does not select model_a's other children"""
        results = run_dbt(["build", "--select", "model_a not_null_model_a_id+"], expect_pass=False)
        assert len(results) == 2
````

(While here, I noticed that we need to update our `.gitignore` to un-ignore the `tests/functional/build` directory: https://github.com/dbt-labs/dbt-core/pull/7436)

**Comment by b-luu** [comment]
Thanks @jtcohen6 :pray: 

I've added that, just changing `expect_pass` to `True` in the call to `dbt.tests.utils.run_dbt` as the build shouldn't fail in this current case.

So everything should be ready, please let me know if I can help in any other way.

File: core/dbt/graph/graph.py [hunk 1]
```diff
@@ -33,11 +33,23 @@ def ancestors(self, node: UniqueId, max_depth: Optional[int] = None) -> Set[Uniq
...
 
-    def descendants(self, node: UniqueId, max_depth: Optional[int] = None) -> Set[UniqueId]:
+    def descendants(
+        self, node: UniqueId, max_depth: Optional[int] = None, include_tests_children: bool = False
+    ) -> Set[UniqueId]:
         """Returns all nodes reachable from `node` in `graph`"""
         if not self.graph.has_node(node):
             raise DbtInternalError(f"Node {node} not found in the graph!")
-        return {child for _, child in nx.bfs_edges(self.graph, node, depth_limit=max_depth)}
+        filtered_graph = nx.restricted_view(
```
jtcohen6, line 42: Good news! If we apply this exact same filtering logic to `ancestors` above, we could also resolve a similar bug reported for upstream selection (https://github.com/dbt-labs/dbt-core/issues/5193).

@b-luu Is that something you'd be willing to take on, in the scope of this same PR? I think it would just require factoring this out into a reusable method, and one more (slightly more involved) test case, asserting that `dbt ls 1+model_c` and `dbt build 1+model_c` return the same number of results.

b-luu: Oh yeah sure: 2 birds in one stone 👍

I'll factor that in during the week (probably Wednesday).

b-luu: Thanks for spotting that!
I had thought about adding it more generally just in case, but didn't see that particular corner case that really justifies it.

Come to think of it, I think these added test edges should simply be ignored from the whole selection process in itself. I'll try to frame it in more generally if I can.

b-luu: Turns out I couldn't sleep this morning, so here it is ;)

b-luu: I limited myself to fitting the new method in ancestors and descendants because I didn't have time to thoroughly dig through and test all the other graph methods to see if we could/should expand this filtering further... :shrug: 

But just covering ancestors and children is enough to solve both issues listed :+1:

b-luu:
Ok, the first test is clearly a false positive: now that we've fixed dbt build to match the other commands (by excluding the parents' te4sts added edges from selections), some tests that were initially set up to replicate that are now failing.

I've fixed the first one in `TestBuildRunResultsState` lines 218 & 220.

But I can't seem to wrap my head around the second one at `TestConcurrentSelectionBuildRunResultsState` line 485. I'm tempted to apply the exact same fix: len(results) -= 1 and remove `table_model` from the expect results set, but I'm not sure that's the actual culprit since we're not using dbt list for comparison in this particular test case...

Also, I'm having a hard time reproducing because my `docker compose up` is failing to build successfully... :-/

b-luu:
> Ok, the first test is clearly a false positive: now that we've fixed dbt build to match the other commands (by excluding the parents' tests added edges from selections), some tests that were initially set up to replicate that are now failing.
> 
> I've fixed the first one in `TestBuildRunResultsState` lines 218 & 220.
> 
> But I can't seem to wrap my head around the second one at `TestConcurrentSelectionBuildRunResultsState` line 485. I'm tempted to apply the exact same fix: len(results) -= 1 and remove `table_model` from the expected results set, but I'm not sure that's the actual culprit since we're not using dbt list for comparison in this particular test case...
> 
> Also, I'm having a hard time reproducing because my `docker compose up` is failing to build successfully... :-/

Actually it's the same for the second one: `table_model` should no longer be present in the selection => it was only present as an artificailly added dependency (parent test) to the `unique_view_model_id` test (whereas it's initial dependency is only on the `view_model`).

So I've also fixed this second test, it should be better now :+1:

core/dbt/graph/graph.py
```diff
@@ -28,16 +28,31 @@ def ancestors(self, node: UniqueId, max_depth: Optional[int] = None) -> Set[Uniq
...
+        filtered_graph = self.exclude_edge_type("parent_test")
         return {
             child
-            for _, child in nx.bfs_edges(self.graph, node, reverse=True, depth_limit=max_depth)
+            for _, child in nx.bfs_edges(filtered_graph, node, reverse=True, depth_limit=max_depth)
         }
 
-    def descendants(self, node: UniqueId, max_depth: Optional[int] = None) -> Set[UniqueId]:
+    def descendants(
+        self, node: UniqueId, max_depth: Optional[int] = None, include_tests_children: bool = False
```
iknox-fa, line 38: `Include_test_children` doesn't appear to be used.

b-luu: Sorry forgot to remove it (had initially added it during testing...)

b-luu: Removed.

b-luu:
Thanks for your reviews @jtcohen6 and @iknox-fa :pray: 

I've applied your suggestions and also fixed the failing tests. :heavy_check_mark:

iknox-fa:
Hey @b-luu.  Sorry for the delay in reviewing this one, since there was a non-trivial chance it would make an impact on performance it took me a little longer to try it out across a series of test projects and benchmark results.

From what I'm seeing the performance is within the bounds of standard deviation for the use cases I'd expect it to effect.  I'm attaching the testing results for the largest and most complex project below for future reference.  

It's probably worth noting that there were several times where the compile step took considerably longer than expected.  Since I'm testing on a local laptop and I wasn't able to reliably duplicate it I've tossed those datapoints since they probably represent a background task on my local machine or some other non-dbt related factor.  Still worth noting though :)

<details>
  <summary>Test Results</summary>

# Compile
`dbt compile --full-refresh --project-dir ../dbt-projects/really_big_project --target-path ../dbt-projects/really_big_project/target`

PR times: 
    1:34.64
    1:30.29
    1:32.03
    1:32.76
    1:35.46
    ----
    avg: 1:33.03

Main times:
    1:38.61
    1:36.74
    1:36.61
    1:35.52
    1:36.01
    ----
    avg: 1:36.70

Result: PR = 3.80% shorter run time
 
# List
`(modify all tests)`
`time dbt list --project-dir ../dbt-projects/really_big_project --select state:modified+ --state ../dbt-projects/really_big_project/previous_target`

PR times: 
    1:10.89
    1:11.37
    1:10.59
    1:09.30
    1:09.31
    ----
    avg: 1:10.29

Main times: 
    1:15.18
    1:13.14
    1:13.54
    1:15.23
    1:15.20
    ----
    avg: 1:14.45

Result: PR = 5.79% shorter run time

# Run
`(modify all tests)`
`time dbt run --project-dir ../dbt-projects/really_big_project --select state:modified+ --state ../dbt-projects/really_big_project/previous_target --target-path ../dbt-projects/really_big_project/target`

PR times:
    1:11.10
    1:13.86
    1:14.19
    1:14.59
    1:15.73
    ----
    avg: 1:13.89

Main times: 
    1:16.28
    1:15.55
    1:10.44
    1:11.97
    1:11.37
    ----
    avg: 1:13.12

Result: PR = 1.05% longer run time

# Test
`(modify all tests)`
`time dbt test --project-dir ../dbt-projects/really_big_project --select state:modified+ --state ../dbt-projects/really_big_project/previous_target --target-path ../dbt-projects/really_big_project/target`

PR times:
    1:12.19
    1:14.01
    1:15.97
    1:16.12
    1:15.57
    ----
    avg: 1:14.77

Main times: 
    1:11.68
    1:12.26
    1:11:66
    1:13.17
    1:12.34
    ----
    avg: 1:12.22

Result: PR = 3.53% longer run time

</details>

iknox-fa (approved):
Looks good to me.

b-luu:
Awesome :tada: 

Thanks for the thorough performance testing and taking the relay on this @iknox-fa :pray:

PR #7435, opened 2023-04-22
UX improvements to model versions

### LINKED ISSUE #7443: [CT-2469] Quick UX improvements to model versions
Discussed in: https://github.com/dbt-labs/dbt-core/discussions/6736#discussioncomment-5695170

Three quick hits that allow us to tell a more compelling story around model versions in the first cut (v1.5):
- By default, allow defining the `latest_version` of dim_customers in a file named `dim_customers.sql` **(no suffix)**. The convention of `dim_customers_vX.sql` would also be supported — so this isn't actually a behavior change.
- Provide users with a mechanism (probably `post-hook` macro) to create an un-suffixed alias pointing to the latest model version, until we support this natively: https://github.com/dbt-labs/dbt-core/issues/7442
- Info-level log if you have an unpinned `ref` to a model that has a newer-than-latest (i.e. `prerelease`) version defined: "Heads up, your unpinned `ref` will roll over to the new version as soon as it's the latest."

jtcohen6:
resolves #7443

Naive implementations for some of the proposed improvements I mention here:
- https://github.com/dbt-labs/dbt-core/discussions/6736#discussioncomment-5695170

### Overview of changes
1. ~By default, land the `latest_version` of `dim_customers` in `<db>.<schema>.dim_customers` **(no suffix)**. This is a behavior change to the built-in `generate_alias_name`, so if we're going to do it, we should do it before the final release.~ We are going to recommend accomplishing this with a `post-hook` in the meantime. We are still discussing if it makes sense to vendor that macro code within `dbt-core`.
2. By default, allow defining the `latest_version` of dim_customers in a file named `dim_customers.sql` **(no suffix)**. The convention of `dim_customers_vX.sql` would also be supported — so this isn't actually a behavior change.
3. Info-level log if you have an unpinned `ref` to a model that has a newer-than-latest (i.e. `prerelease`) version defined: "Heads up, your unpinned `ref` will roll over to the new version as soon as it's the latest." (Should this be a warning? If you're not actually referencing a model slated for deprecation...)
```
11:47:08  3 of 3 START sql view model dbt_jcohen.another_model ........................... [RUN]
11:47:09  While compiling 'another_model':
Found an unpinned reference to versioned model 'my_model' in 'my_dbt_project'.
Resolving to latest version: my_model.v2
A prerelease version 3 is available. It has not yet been marked 'latest' by its maintainer.
When that happens, this reference will resolve to my_model.v3 instead.

  Try out v3: {{ ref('my_dbt_project', 'my_model', v='3') }}
  Pin to v2:  {{ ref('my_dbt_project', 'my_model', v='2') }}
```

(Bonus) Raise a nicer error if I've written a `ref` pinned to a specific version, and the model I'm referencing isn't actually versioned. This came up during training, when someone didn't correctly define the `versions` block on their model — but it wasn't clear if the error was in the model configuration, or in the downstream `ref`. This isn't a behavior change, and the codepath is slightly trickier, so I'll probably just kick it out of this PR. **Update:** Kicked out of scope.

### Checklist

- [x] This PR includes tests, or tests are not required/relevant for this PR — no test for the logging event
- [x] I have [opened an issue to add/update docs](https://github.com/dbt-labs/docs.getdbt.com/issues/new/choose), or docs changes are not required/relevant for this PR — captured in https://github.com/dbt-labs/docs.getdbt.com/pull/3232


gshank (approved):

core/dbt/contracts/graph/manifest.py
```diff
@@ -169,7 +170,33 @@ def find(
     ):
         unique_id = self.get_unique_id(key, package, version)
         if unique_id is not None:
-            return self.perform_lookup(unique_id, manifest)
+            node = self.perform_lookup(unique_id, manifest)
+            if version is None and node.is_versioned and get_node_info():
```
MichelleArk, line 174: what is `get_node_info()` doing here?

core/dbt/contracts/graph/manifest.py
```diff
@@ -169,7 +170,33 @@ def find(
...
+                # This was an 'unpinned' reference, resolved to latest_version.
+                # Check to see if newer versions are available, and log about it.
+                max_version: Union[str, float, None] = max(  # type: ignore
+                    [
+                        v.version
+                        for k, v in manifest.nodes.items()
+                        if k.startswith(f"{node.unique_id[:-len(str(node.version))]}")
+                    ]
+                )
```
MichelleArk, line 183: think this can be simplified to: 
```python
                max_version: Union[str, float, None] = max(  # type: ignore
                    [
                        v.version
                        for v in manifest.nodes.values()
                        if v.name == node.name
                    ]
                )
```

MichelleArk:
Looking at the logged message: 
```
Try out v3: {{ ref('my_dbt_project', 'my_model', v='3') }}
Pin to v2:  {{ ref('my_dbt_project', 'my_model', v='3') }}
```
Should the second line be: 
```
Pin to v2:  {{ ref('my_dbt_project', 'my_model', v='2') }}
```

core/dbt/contracts/graph/manifest.py
jtcohen6: This was my janky way of logging the message only at runtime, rather than at parsing time. We could log it twice.

MichelleArk: Ahhh I see! Worth a comment

core/dbt/contracts/graph/manifest.py
jtcohen6: this is ... clearly better

MichelleArk: Broader refactoring suggestion: using `UnparsedVersion` for the comparisons, similar to [what we do in the graph selector method here](https://github.com/dbt-labs/dbt-core/blob/main/core/dbt/graph/selector_methods.py#L661)

```
                max_version = max(
                    [
                        UnparsedVersion(v.version)
                        for v in manifest.nodes.values()
                        if v.name == node.name and v.version is not None
                    ]
                )
                assert node.latest_version
                if max_version > UnparsedVersion(node.latest_version):
                    fire_event(
                        UnpinnedRefNewVersionAvailable(
                            node_info=get_node_info(),
                            ref_node_name=node.name,
                            ref_node_package=node.package_name,
                            ref_node_version=str(node.version),
                            ref_max_version=str(max_version.v),
                        )
                    )
```

The TypeError issues in unit testing should be fixed by updating the MockNode constructor: 
```
+++ b/test/unit/utils.py
@@ -350,6 +350,7 @@ def MockNode(package, name, resource_type=None, **kwargs):
         **kwargs,
     )
     node.name = name
+    node.is_versioned = resource_type is NodeType.Model and version is not None
     return node
```

jtcohen6: yes!!!

jtcohen6: at the risk of expanding the intended scope of a partial parsing test - I'm going to borrow your commit 84cc3cabb69e704b0e09737121f3f08a90aabc24 and check that this message is being logged in the case of an unpinned `ref` to a latest-not-newest model

jtcohen6:
Will rebase after #7439 is merged

PR #7447, opened 2023-04-24
Copy dir if symlink fails

### LINKED ISSUE #7428: [CT-2457] [Bug] dbt deps commands error our in databricks DBFS

### Current Behavior

`dbt deps `command fails to run in databricks DBFS. 

-  `deps` command internally creating a symlink while running dbt deps command. After more debugging, I found dbt internally creating a symlink when installing` dbt deps.`

- In current code the`  supports_symlinks` is based on if python `lib(os)` has the `symlink` method available or not. https://github.com/dbt-labs/dbt-core/blob/main/core/dbt/clients/system.py#L161 

- As DBT is running on 3.7+ Python we can get rid of `supports_symlinks` method. This will be always true in databricks DBFS because `getattr(os, "symlink", None) `will never return False  with the python versions dbt supporting

- DBFS can not work with symlink so we should use copytree.

https://stackoverflow.com/questions/64314341/creating-symbolic-link-in-a-databricks-filestores-folder
Screenshot 2023-04-21 at 3 38 03 pm

### Expected Behavior

* If symlink fails copy directory

### Steps To Reproduce

1. Install dbt under the DBFS path
2. Run dbt deps command

### Relevant log output

_No response_

### Environment

```markdown
- OS: NA (DBFS)
- Python: 3.8
- dbt: dbt-databricks==1.2.0
```

### Additional Context

_No response_

### LINKED ISSUE #8223: [CT-2876] [Feature] add possibility to copy dbt local packages instead of make it symlink

### Describe the feature

Hello team,

This is a feature request that comes from the following issues/discussions:
- https://github.com/dbt-labs/dbt-core/issues/5271
- https://github.com/dbt-labs/dbt-core/issues/4538

In this piece of code https://github.com/dbt-labs/dbt-core/blob/main/core/dbt/deps/local.py#L62 

```python
 if can_create_symlink: 
     fire_event(DepsCreatingLocalSymlink())
     system.make_symlink(src_path, dest_path)
 else: 
     fire_event(DepsSymlinkNotAvailable())
     shutil.copytree(src_path, dest_path)
```

we are evaluating whether symlink can be setuped. In case the OS does not support symlink (i.e. Windows) we do a deep copy.

We would like to be able to do a dbt local packages deep copy on Linux machines, for instance. 

The reasons for this aim are:
- `gsutils rsync` [does not support](https://cloud.google.com/storage/docs/gsutil/commands/rsync#be-careful-when-synchronizing-over-os-specific-file-types-symlinks,-devices,-etc.) symlinks as well as `rsync` does not support symlinks
- `Google Cloud Storage` [does not support](https://stackoverflow.com/questions/57777777/symlink-or-an-alternative-solution-on-gcs) symlinks

As a workaround, I had to write custom Python code, that iterates thru all the packages we have, find the symlink, and make a deep copy of all the content.

### Who will this benefit?

In my humble opinion, at least, GCP users would benefit a lot

There is a chance, with all the context and pitfalls to be provided

### Anything else?

Probably, no. However, if I miss something please feel free to ask all the questions and raise all the concerns you might be having

anjutiwari:
resolves #7428
resolves #8223

### Description

dbt deps command fails to run in databricks DBFS. Copy dir in case if any symlink fails with expection 

### Checklist

### DIFF
--- a/core/dbt/deps/local.py
+++ b/core/dbt/deps/local.py
@@ -51,19 +51,15 @@ class LocalPinnedPackage(LocalPackageMixin, PinnedPackage):
         src_path = self.resolve_path(project)
         dest_path = self.get_installation_path(project, renderer)
 
-        can_create_symlink = system.supports_symlinks()
-
         if system.path_exists(dest_path):
             if not system.path_is_symlink(dest_path):
                 system.rmdir(dest_path)
             else:
                 system.remove_file(dest_path)
-
-        if can_create_symlink:
+        try:
             fire_event(DepsCreatingLocalSymlink())
             system.make_symlink(src_path, dest_path)
-
-        else:
+        except OSError:
             fire_event(DepsSymlinkNotAvailable())
             shutil.copytree(src_path, dest_path)
 
--- a/.changes/unreleased/Fixes-20230424-210734.yaml
+++ b/.changes/unreleased/Fixes-20230424-210734.yaml
@@ -3,4 +3,4 @@ body: Copy dir during `dbt deps` if symlink fails
 time: 2023-04-24T21:07:34.336797+05:30
 custom:
   Author: anjutiwari
-  Issue: "7428" "8223"
+  Issue: "7428 8223"
--- a/.changes/unreleased/Fixes-20230424-210734.yaml
+++ b/.changes/unreleased/Fixes-20230424-210734.yaml
@@ -1,0 +1,6 @@
+kind: Fixes
+body: Copy dir during `dbt deps` if symlink fails
+time: 2023-04-24T21:07:34.336797+05:30
+custom:
+  Author: anjutiwari
+  Issue: "7428" "8223"
--- a/tests/unit/test_deps.py
+++ b/tests/unit/test_deps.py
@@ -6,7 +6,7 @@ from unittest import mock
 import dbt.deps
 import dbt.exceptions
 from dbt.deps.git import GitUnpinnedPackage
-from dbt.deps.local import LocalUnpinnedPackage
+from dbt.deps.local import LocalUnpinnedPackage, LocalPinnedPackage
 from dbt.deps.tarball import TarballUnpinnedPackage
 from dbt.deps.registry import RegistryUnpinnedPackage
 from dbt.clients.registry import is_compatible_version
@@ -91,6 +91,21 @@ class TestGitPackage(unittest.TestCase):
         self.assertEqual(a_pinned.get_version(), "0.0.1")
         self.assertEqual(a_pinned.source_type(), "git")
         self.assertIs(a_pinned.warn_unpinned, True)
+
+    @mock.patch("shutil.copytree")
+    @mock.patch("dbt.deps.local.system.make_symlink")
+    @mock.patch("dbt.deps.local.LocalPinnedPackage.get_installation_path")
+    @mock.patch("dbt.deps.local.LocalPinnedPackage.resolve_path")
+    def test_deps_install(
+        self, mock_resolve_path, mock_get_installation_path, mock_symlink, mock_shutil
+    ):
+        mock_resolve_path.return_value = "/tmp/source"
+        mock_get_installation_path.return_value = "/tmp/dest"
+        mock_symlink.side_effect = OSError("Install deps symlink error")
+
+        LocalPinnedPackage("local").install("dummy", "dummy")
+        self.assertEqual(mock_shutil.call_count, 1)
+        mock_shutil.assert_called_once_with("/tmp/source", "/tmp/dest")
 
     def test_invalid(self):
         with self.assertRaises(ValidationError):


core/dbt/deps/local.py
```diff
@@ -51,19 +51,15 @@ def install(self, project, renderer):
...
 
-        can_create_symlink = system.supports_symlinks()
-
         if system.path_exists(dest_path):
             if not system.path_is_symlink(dest_path):
                 system.rmdir(dest_path)
             else:
                 system.remove_file(dest_path)
-
-        if can_create_symlink:
```
nitinbhojwani, line 62: `if can_create_symlink` check shouldn't be removed. It helps for the cases where `symlink` function doesn't exist on `os`, in which case `AttributeError` would be thrown.

`except OSError` should be added as an edge case like ours, where `os.symlink` function exists but it still fails.

Something like:
```
symlink_supported = False

if can_create_symlink:
    try:
        fire_event(DepsCreatingLocalSymlink())
        system.make_symlink(src_path, dest_path)
    except OSError:
        pass
    else:
        symlink_supported = True
if not symlink_supported:
    fire_event(DepsSymlinkNotAvailable())
    shutil.copytree(src_path, dest_path)
```

test/unit/test_deps.py
```diff
@@ -92,6 +92,21 @@ def test_init(self):
...
+        self, mock_resolve_path, mock_get_installation_path, mock_symlink, mock_shutil
+    ):
+        mock_resolve_path.return_value = "/tmp/source"
+        mock_get_installation_path.return_value = "/tmp/dest"
+        mock_symlink.side_effect = OSError("Install deps symlink error")
+
+        LocalPinnedPackage("local").install("dummy", "dummy")
+        self.assertEqual(mock_shutil.call_count, 1)
+        mock_shutil.assert_called_once_with("/tmp/source", "/tmp/dest")
+
```
nitinbhojwani, line 109: add test where `system.supports_symlinks()` returns False

nitinbhojwani (changes requested):
`system.supports_symlinks()` check should continue to exist before trying to invoke `symlink` on `os`

Along with that, the edge case handling of `OSError` exception on `system.make_symlink` should be done.

core/dbt/deps/local.py
anjutiwari: I do not think we need it, `make_symlink`  calling it. The functionality remains the same. Check make_symlink method.

test/unit/test_deps.py
anjutiwari: My idea was to handle the exception for only the newly added code.

core/dbt/deps/local.py
```diff
@@ -51,19 +51,15 @@ def install(self, project, renderer):
...
+        try:
             fire_event(DepsCreatingLocalSymlink())
             system.make_symlink(src_path, dest_path)
-
-        else:
+        except OSError:
```
anjutiwari, line 62: https://docs.getdbt.com/faqs/core/install-pip-os-prereqs.md
Handling exception case of DBFS

core/dbt/deps/local.py
nitinbhojwani: `make_symlink` method will throw `dbt.exceptions.SymbolicLinkError`, in which case `except OSError` block here won't handle it.  
The behaviour is changed for the case where `symlink` function attribute isn't present on `os`

anjutiwari: yes, I think we should use `dbt.exceptions.SymbolicLinkError,` and we should also catch the SymbolicLinkError. 

Also, I am curious about which OS does not have symlinks function in python `lib(os)` package, as mentioned in my Issue description.
As per the dbt, we have a list of supported operating systems and condition on python3.7+. https://docs.getdbt.com/faqs/core/install-pip-os-prereqs.md

If my understanding is true, we may not have any scenarios where  `lib(os)` package will not have a `symlink` function and we may not need to handle the SymbolicLinkError exception. 

Waiting for the core reviewers to help us here.

chronitis:
Just a note that I think this PR fixes wider problems than just those related to databricks DBFS.

When using a `local: PATH` package on Windows, DBT will attempt to create a symlink (since `supports_symlink()` is true), but unless the user has administrator permissions or specific group policy, creating the symlink fails. If I read this PR right, it would also fix this issue.

anjutiwari:
Thanks @chronitis for pointing out another edge case. Let me try to reproduce this scenario and verify if this solution is already fixing the problem.

anjutiwari:
Hey @jtcohen6, 
Just wanted to check for any update on this. Please let me know if I could make this PR more smooth to get it prioritized for the review.

Thank you

anjutiwari:
Thanks @gshank. Fixing it.

dbeatty10:
> Just a note that I think this PR fixes wider problems than just those related to databricks DBFS.
> 
> When using a `local: PATH` package on Windows, DBT will attempt to create a symlink (since `supports_symlink()` is true), but unless the user has administrator permissions or specific group policy, creating the symlink fails. If I read this PR right, it would also fix this issue.

Very good point @chronitis ! This PR does look like it addresses use cases beyond Databricks DBFS, namely any time a symlink doesn't work. For example, it would also resolve https://github.com/dbt-labs/dbt-core/issues/8223.

gshank (approved):

PR #7450, opened 2023-04-24
Use profile specified in --profile with dbt init

### LINKED ISSUE #6154: [CT-1418] [Bug] CLI argument `--profile` ignored in `dbt init`

### Current Behavior

According to `dbt init --help`, the `--profile` option lets the user load an existing profile to use with a new project. However, that option currently does nothing when you call it with `dbt init`. 

Example:
```
dbt init --profile default proj1
```

Result:
proj1 created with `proj1/dbt_project.yml` showing `profile: 'proj1'`. `~/.dbt/profiles.yml` shows a new profile `proj1` created with placeholders. **Note**: if a profile named `proj1` already exists, dbt *still* creates another profile named `proj1` again.

In contrast, calling `dbt run --profile default` actually does make the project use the profile `default`, so the `--profile` argument does work elsewhere at least. 

### Expected Behavior

proj1 created with `proj1/dbt_project.yml` showing `profile: 'default'`. `~/.dbt/profiles.yml` should not show any new profile created.

### Steps To Reproduce

1. Create new project directory to store project in
2. Open `~/.dbt/profiles.yml` and create a new profile `default`.
3. Run `dbt init --profile default proj1`
4. `cd` into `/proj1` and open `dbt_project.yml`. The `profile` will be set to `proj1` instead of `default`.
5. Check `~/.dbt/profiles.yml` again and see that dbt has created a new blank `proj1` profile to go with the new project, instead of using the existing `default` profile.

### Relevant log output

_No response_

### Environment

```markdown
- OS: Ubuntu 22.04.1 LTS
- Python: 3.10.6
- dbt: 1.3.0
```

postgres

### Additional Context

_No response_

ezraerb:
resolves #6154

### Description

If --profile is specified with dbt init, and the profile exists, initialize the project with that as the profile instead of creating a new profile. If the profile does not exist, or the command is run inside an existing project, it fails with an error.

### Checklist


dbeatty10:
Thanks for submitting this pull request @ezraerb !

The core team is hard at work making lots of improvements so it may be a while before an engineer is assigned to work on this PR.

In the meantime, https://github.com/dbt-labs/dbt-core/pull/7609 is updating the same area of code (to resolve a different issue), so one or the other will need to be updated depending on the merge order. We'll probably want to merge https://github.com/dbt-labs/dbt-core/pull/7609 first, so you'd need to resolve the conflicts afterwards.

.changes/unreleased/Fixes-20230424-161642.yaml
```diff
@@ -0,0 +1,7 @@
+kind: Fixes
+body: If --profile specified with dbt-init, create the project with the specified
+  profile
+time: 2023-04-24T16:16:42.994547-04:00
+custom:
+  Author: ezraerb
+  Issue: CT-1418
```
dbeatty10, line 7: For `changie`, we use the GitHub issue number rather than the Jira issue number.
```suggestion
  Issue: "6154"
```

ezraerb:
Sure thing. Let me know when I need to make edits after the other merge.

dbeatty10:
@ezraerb https://github.com/dbt-labs/dbt-core/pull/7609 is merged now ✅

ezraerb:
Merged code has been pushed to my branch

core/dbt/task/init.py
```diff
@@ -274,9 +276,18 @@ def run(self):
             in_project = False
 
         if in_project:
+            # If --profile was specified, it means use an existing profile, which is not
+            # applicable to this case
+            if getattr(get_flags(), "PROFILE", None):
+                print(
```
VersusFacit, line 282: Let's import and use an appropriate exception from `dbt.exceptions` which is defined in `core/dbt/exceptions`. Maybe `DbtRuntimeError`?

core/dbt/task/init.py
```diff
@@ -285,11 +296,25 @@ def run(self):
...
 
-        # Ask for adapter only if skip_profile_setup flag is not provided.
-        if not self.args.skip_profile_setup:
+            # If the user specified an existing profile to use, use it instead of generating a new one
+            user_profile_name = getattr(get_flags(), "PROFILE", None)
+            if user_profile_name:
+                # Verify it exists. Can't use the regular profile validation routine because it assumes
+                # the project file exists
+                raw_profiles = read_profile(profiles_dir)
+                if user_profile_name not in raw_profiles:
```
VersusFacit, line 306: ditto for exceptions

tests/functional/init/test_init.py
```diff
@@ -84,6 +85,12 @@ def test_init_task_in_project_with_existing_profiles_yml(
 """
             )
 
+    def test_init_task_in_project_specifying_profile_errors(self):
+        # This triggers a call to sys.exit(), requring the following to test it
+        with pytest.raises(SystemExit) as error:
```
VersusFacit, line 90: we have a builtin flag for this

`run_dbt(..., expect_pass=False)`

tests/functional/init/test_init.py
```diff
@@ -159,6 +166,21 @@ def exists_side_effect(path):
...
+        def exists_side_effect(path):
+            # Override responses on specific files, default to 'real world' if not overriden
+            return {"profiles.yml": False}.get(path.name, os.path.exists(path))
+
+        exists.side_effect = exists_side_effect
+
+        # Even through no profiles.yml file exists, the init will not modify project.yml,
+        # so this errors
+        # This triggers a call to sys.exit(), requring the following to test it
+        with pytest.raises(SystemExit) as error:
```
VersusFacit, line 180: ditto

tests/functional/init/test_init.py
```diff
@@ -708,3 +730,130 @@ def test_init_inside_project_and_skip_profile_setup(
...
+
+        # profiles.yml is NOT overwritten, so assert that the text matches that of the
+        # original fixture
+        with open(os.path.join(project.profiles_dir, "profiles.yml"), "r") as f:
+            assert f.read() == yaml.safe_dump(dbt_profile_data)
+
+        with open(os.path.join(project.project_root, project_name, "dbt_project.yml"), "r") as f:
+            assert (
+                f.read()
+                == f"""
```
VersusFacit, line 763: Can we make this a fixture in a constant at file top or in a `fixture.py`?

tests/functional/init/test_init.py
```diff
@@ -708,3 +730,130 @@ def test_init_inside_project_and_skip_profile_setup(
...
+    ):
+        manager = Mock()
+        manager.attach_mock(mock_prompt, "prompt")
+        manager.prompt.side_effect = [
+            project_name,
+        ]
+        mock_get_adapter.return_value = [project.adapter.type()]
+
+        # This triggers a call to sys.exit(), requring the following to test it
+        with pytest.raises(SystemExit) as error:
```
VersusFacit, line 818: another refactor opportunity

tests/functional/init/test_init.py
```diff
@@ -708,3 +730,130 @@ def test_init_inside_project_and_skip_profile_setup(
...
+        # Override responses on specific files, default to 'real world' if not overriden
+        original_isfile = os.path.isfile
+        with mock.patch(
+            "os.path.isfile",
+            new=lambda path: {"profiles.yml": False}.get(
+                os.path.basename(path), original_isfile(path)
+            ),
+        ):
+            # This triggers a call to sys.exit(), requring the following to test it
+            with pytest.raises(SystemExit) as error:
```
VersusFacit, line 851: ditto

core/dbt/task/init.py
```diff
@@ -274,9 +276,18 @@ def run(self):
...
+            if getattr(get_flags(), "PROFILE", None):
+                print(
+                    "Can not init existing project with specified profile, edit dbt_project.yml instead"
+                )
+                sys.exit(1)
+
             # When dbt init is run inside an existing project,
             # just setup the user's profile.
             profile_name = self.get_profile_name_from_current_project()
+            profile_specified = False
```
VersusFacit, line 290: Sentinel values like this in if-else branches can make it hard to read code and track execution semantics. Would much prefer a different strategy like calling functions that make sense in each if-else branch.

tests/functional/init/test_init.py
```diff
@@ -708,3 +730,130 @@ def test_init_inside_project_and_skip_profile_setup(
...
+
+class TestInitOutsideOfProjectWithSpecifiedProfile(TestInitOutsideOfProjectBase):
+    @mock.patch("dbt.task.init._get_adapter_plugin_names")
+    @mock.patch("click.prompt")
+    def test_init_task_outside_of_project_with_specified_profile(
+        self, mock_prompt, mock_get_adapter, project, project_name, unique_schema, dbt_profile_data
+    ):
+        manager = Mock()
+        manager.attach_mock(mock_prompt, "prompt")
+        manager.prompt.side_effect = [
```
VersusFacit, line 743: Curious why Black didn't reformat this or the line below it to be the same.

core/dbt/task/init.py
```diff
@@ -274,9 +276,18 @@ def run(self):
             in_project = False
 
         if in_project:
+            # If --profile was specified, it means use an existing profile, which is not
+            # applicable to this case
+            if getattr(get_flags(), "PROFILE", None):
```
VersusFacit, line 281: `self.args` will give you access to the profile flag's value, so use that instead of `getattr(get_flags(), "PROFILE", None)`, as this is how we do it elsewhere in core tasks

VersusFacit (commented):
Hey there! Thanks a bundle for all this great work to pull these arguments through such that they obey the stated flag interface!! 🙇‍♀️ 

I had some time for reviewing PRs and do a bunch of `task` directory work, so I offered some thought for how to rework your code to be inline with our conventions and practices.

I may not be the one to pull this PR over the finish line, but I hope my comments help you on your way. Welcome to the community as a first-time contributor 🎉

core/dbt/task/init.py
```diff
@@ -285,11 +296,25 @@ def run(self):
...
+                    sys.exit(1)
+                profile_name = user_profile_name
+                profile_specified = True
+            else:
+                profile_name = project_name
+                profile_specified = False
+            self.create_new_project(project_name, profile_name)
+
+        # Ask for adapter only if skip_profile_setup flag is not provided and no profile to use was specified.
+        if not self.args.skip_profile_setup and not profile_specified:
```
dbeatty10, line 317: ### User experience
From a user experience perspective, I don't think we need to raise an error if the specified profile is not found. I'd rather just create it anytime it doesn't exist.

### Implementation
I defer to whoever ends up being the code reviewer for this, but see below for some suggestions for refactoring.

The `run` method is long and has a lot of conditionals, which makes it harder to read. Refactoring this would make it easier to maintain in the future.

So I'd suggest refactoring this logic into its own method (similar to how `check_if_can_write_profile` is its own method). Maybe something similar this (completely untested!):

```python
def check_if_profile_exists(self, profile_name: Optional[str] = None) -> bool:
    profile_exists = False  # assume it doesn't exist unless proven otherwise
    user_profile_name = getattr(get_flags(), "PROFILE", None)
    profiles_dir = get_flags().PROFILES_DIR
    if user_profile_name:
        raw_profiles = read_profile(profiles_dir)
        profile_exists = user_profile_name in raw_profiles
    return profile_exists
```

Then this method can be applied in one or more places to use the specified profile if it exists (and create it otherwise).

core/dbt/task/init.py
ezraerb: ditto for the changes.

tests/functional/init/test_init.py
ezraerb: Checking the code for run_dbt() shows that expect_pass only applies to the reported status of the task, not any generated exceptions. Those get reraised.

tests/functional/init/test_init.py
ezraerb: ditto

tests/functional/init/test_init.py
ezraerb: Changed along the lines of the the other tests referenced above.

tests/functional/init/test_init.py
ezraerb: ditto

tests/functional/init/test_init.py
ezraerb: Checking the other init tests shows that all of them use this pattern. Its probably worth getting more opinions before changing it.

core/dbt/task/init.py
ezraerb: For user experience, I think more input is needed. The risk with creating the profile if it does not exist is the classic typo problem, where a misspelling creates a new profile instead of using the one the user actually wanted. The requirements said "existing profile" so I put in the check.

For implementation, the test is only done once currently, but shrinking a big method is usually a good idea. Refactored.

dbeatty10: Did you push the refactored code @ezraerb?

This is the most recent that I'm seeing:
image

ezraerb: Still working on a few other comments, so have not pushed yet.

core/dbt/task/init.py
ezraerb: I personally can handle sentinel values as long as the names are descriptive of the state they capture, and use is not excessive. However, there was a request elsewhere to refactor run() into smaller methods. Since creating the profile is triggered in two places, its an excellent candidate for a refactor. Changed as requested.

core/dbt/task/init.py
ezraerb: Changes have been pushed.

core/dbt/task/init.py
```diff
@@ -274,9 +304,18 @@ def run(self):
...
+                raise DbtRuntimeError(
+                    msg="Can not init existing project with specified profile, edit dbt_project.yml instead"
+                )
+
             # When dbt init is run inside an existing project,
             # just setup the user's profile.
-            profile_name = self.get_profile_name_from_current_project()
+            if not self.args.skip_profile_setup:
+                profile_name = self.get_profile_name_from_current_project()
+                self.setup_profile(profile_name)
```
dbeatty10, line 318: @ezraerb Can you help me understand why we'd want or need `self.get_profile_name_from_current_project()` to be guarded by `if not self.args.skip_profile_setup:`?

Could we safely do the following outside of `if` statement?
```python
            profile_name = self.get_profile_name_from_current_project()
```

If so, then we could avoid repeating `self.setup_profile(profile_name)` twice (once for each case) and instead we can just do it once at the very end (IFF profile setup is being skipped).

ezraerb: There are four cases that need to be considered:
1. Initializing inside an existing project, no profile specified. The profile is created (the code being commented on)
2. Initializing inside an existing project, profile specified. This raises an error.
3. Initializing outside an existing project, no profile specified. Create the project and then create the profile.
4. Initializing outside an existing project, profile specified. Create the project only.

The difference between cases 3 and 4 prohibits having a catch-all profile creation at the end of the method. Either signal variables are needed or the calls need to be duplicated in the various paths. Previous viewers expressed a desire for the latter as they believe it is easier to follow.

core/dbt/task/init.py
jtcohen6: My first instinct was to agree with @dbeatty10 on this point:

> From a user experience perspective, I don't think we need to raise an error if the specified profile is not found. I'd rather just create it anytime it doesn't exist.

I definitely appreciate the typo annoyance: `dbt init --profile defaut` would lead to the writing of a whole new `defaut` profile, when you _intended_ to use the existing `default` profile. In that case, it would be better to get an explicit error.

But it means that, when initializing a new project from scratch, you have exactly two options:
1. Do not pass `--profile` flag: Initialize a new project and a new profile, with the same name as your project.
2. Pass the `--profile` flag. It must match an existing profile.

If we took Doug's recommendation, there would be three options:
1. Do not pass `--profile` flag to initialize a new project and a new profile. The profile name will match your supplied project name (reasonable default behavior).
2. Pass the `--profile` flag, and its value matches an existing profile: Initialize a new project, do not write a new profile.
3. Pass the `--profile` flag, and its value doesn't match an existing profile: Initialize a new project and a new profile. The profile name will match what you passed into the `--profile` flag.

We could even take this one step further, and provide [the same flexibility that we offer when running `dbt init` within an existing project](https://github.com/dbt-labs/dbt-core/blob/b39eeb328c9a245f21b54fb48f2761ad03d4d6ad/core/dbt/task/init.py#L202-L205):
```
The profile <profile-name> already exists in /Users/jerco/.dbt/profiles.yml. Do you wish to overwrite it? [y/N]: N
```

---

I don't have strong feelings either way. As a heuristic to make this determination, I'm thinking about how there are two "modes" of `dbt init`:
1. **Interactive.** Likely first time using dbt. Need to set up everything.
2. **Programmatic.** Someone who has used dbt before, likely on this machine. Wants to skip the click interactivity and jump straight to `dbt init <project_name> --skip-profile-setup`, or `dbt init <project_name> --profile <existing_profile_name>`.

This `--profile` flag feels designed for persona / use case (2). The first user is less likely to want fine-grained control over exactly how the profile is being named — we should provide the easiest path from start to finish, with some sensible defaults along the way. And second user (slightly more experienced) is less likely to want the interactive walkthrough for setting up their profile.

Which is to say: I'm convinced enough that the behavior implemented in this PR is an acceptable user experience. We'll need to document the behavior in a new "Existing profile" section here: https://docs.getdbt.com/reference/commands/init

emmyoop (changes requested):
@ezraerb this is looking great!  We have some time set aside to help you get this over the finish line an merged in.

There's just one more small change I'd like to request.  Updating the help text will make it more obvious that the `--profile` flag must refer to an existing profile.

The help text is [here](https://github.com/dbt-labs/dbt-core/blob/b39eeb328c9a245f21b54fb48f2761ad03d4d6ad/core/dbt/cli/params.py#L318).  Just adding in the word "existing" would be beneficial:

```
help="Which existing profile to load. Overrides setting in dbt_project.yml.",
```

Other than that you'll need to rebase off `main` to pull in the most recent version of core so we can run some tests.  I'm not seeing a reason the current test is failing but we can get a better picture once you branch is updated.

ezraerb:
I pushed an update with the help text update. When I did a merge from the main branch from upstream, git claimed there was nothing to merge. Should I do a full rebase instead?

emmyoop (approved):
@ezraerb this looks great!  Thanks for making that last small change.  The tests are passing and we're ready to merge this in!  Thanks for your work on getting this fixed!

Issue #7335, opened 2023-04-12
[CT-2402] [Feature] Improve warnings/errors for constraints <> materialization types


jtcohen6:
### Views

If your model is materialized as a `view` and has `constraints` defined, those constraints are not supported & will be ignored, on ALL data platforms. We should raise a warning saying this, similar to the `warn_unsupported` functionality added in #7250.

**Update:** This might require calling a different macro within `create_view_as`, since all constraint parsing/rendering is currently happening within `create_table_as` &rarr; `get_table_columns_and_constraints`.

---

### Ephemeral models
```
Parsing Error
  Constraint validation failed for: (models/sometable.sql)
  Only table, view, and incremental materializations are supported for constraints, but found 'ephemeral'
```
Let's just update the language here: "constraint validation" &rarr; "contract enforcement"

https://github.com/dbt-labs/dbt-core/blob/87e25e8692197c093e9be2c0df5fe38dea48eec5/core/dbt/parser/schemas.py#L969-L988

jtcohen6:
Let's review the `validate_constraint_prerequisites` method as well, see which validation we're actually still using/needing, and either remove or update the warnings accordingly

PR #7476, opened 2023-04-28
Detect breaking changes to constraints in state:modifed

### LINKED ISSUE #7065: [CT-2197] [Feature] Detect breaking changes to constraints in state:modified check

emmyoop:
resolves #7065 

### Description

Detects breaking changes on state modified to enforced constraints.  Accounts for constraints only considered enforced on table & incremental materializations.

### Checklist

### DIFF
--- a/core/dbt/adapters/factory.py
+++ b/core/dbt/adapters/factory.py
@@ -158,7 +158,7 @@ class AdapterContainer:
     def get_adapter_type_names(self, name: Optional[str]) -> List[str]:
         return [p.adapter.type() for p in self.get_adapter_plugins(name)]
 
-    def get_adapter_constraint_support(self, name: str) -> List[str]:
+    def get_adapter_constraint_support(self, name: Optional[str]) -> List[str]:
         return self.lookup_adapter(name).CONSTRAINT_SUPPORT  # type: ignore
 
 
@@ -217,7 +217,7 @@ def get_adapter_type_names(name: Optional[str]) -> List[str]:
     return FACTORY.get_adapter_type_names(name)
 
 
-def get_adapter_constraint_support(name: str) -> List[str]:
+def get_adapter_constraint_support(name: Optional[str]) -> List[str]:
     return FACTORY.get_adapter_constraint_support(name)
 
 
--- a/core/dbt/adapters/factory.py
+++ b/core/dbt/adapters/factory.py
@@ -158,6 +158,9 @@ class AdapterContainer:
     def get_adapter_type_names(self, name: Optional[str]) -> List[str]:
         return [p.adapter.type() for p in self.get_adapter_plugins(name)]
 
+    def get_adapter_constraint_support(self, name: str) -> List[str]:
+        return self.lookup_adapter(name).CONSTRAINT_SUPPORT  # type: ignore
+
 
 FACTORY: AdapterContainer = AdapterContainer()
 
@@ -212,6 +215,10 @@ def get_adapter_package_names(name: Optional[str]) -> List[str]:
 
 def get_adapter_type_names(name: Optional[str]) -> List[str]:
     return FACTORY.get_adapter_type_names(name)
+
+
+def get_adapter_constraint_support(name: str) -> List[str]:
+    return FACTORY.get_adapter_constraint_support(name)
 
 
 @contextmanager
--- a/core/dbt/contracts/graph/nodes.py
+++ b/core/dbt/contracts/graph/nodes.py
@@ -18,9 +18,6 @@ from typing import (
 
 from dbt.dataclass_schema import dbtClassMixin, ExtensibleDbtClassMixin
 
-# causes circular imports... also BaseAdapter isn't right
-# from dbt.adapters.base import BaseAdapter, ConstraintSupport
-
 from dbt.clients.system import write_file
 from dbt.contracts.files import FileHash
 from dbt.contracts.graph.unparsed import (
@@ -183,14 +180,6 @@ class ConstraintType(str, Enum):
             cls(item)
         except ValueError:
             return False
-        return True
-
-    @classmethod
-    def is_enforced(cls, item):
-        # TODO: where can I get this from?
-        # if BaseAdapter.CONSTRAINT_SUPPORT[item] == ConstraintSupport.ENFORCED:
-        #     return True
-        # return False
         return True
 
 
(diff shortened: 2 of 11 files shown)


core/dbt/adapters/factory.py
```diff
@@ -214,6 +217,10 @@ def get_adapter_type_names(name: Optional[str]) -> List[str]:
     return FACTORY.get_adapter_type_names(name)
 
 
+def get_adapter_constraint_support(name: str) -> List[str]:
```
MichelleArk, line 220: nit: looks like `Optional[str]` would be more accurate given usage in https://github.com/dbt-labs/dbt-core/pull/7476/files#diff-42cfc99b2fcd89ab75957b8999bee2b1af822f7d39953ca99f1f285c170557c1R630

core/dbt/contracts/graph/nodes.py
```diff
@@ -584,10 +620,14 @@ def build_contract_checksum(self):
             for column in sorted_columns:
                 contract_state += f"|{column.name}"
                 contract_state += str(column.data_type)
+                contract_state += str(column.constraints)
+            if self.config.materialized in ["table", "incremental"]:
```
MichelleArk, line 624: Can be in a follow-up issue (perhaps when we tackle https://github.com/dbt-labs/dbt-core/issues/7335), but finding a way to avoid [duplicating](https://github.com/dbt-labs/dbt-core/blob/87e25e8692197c093e9be2c0df5fe38dea48eec5/core/dbt/parser/schemas.py#L976) this hard-coded list of supported materializations would be neat!

MichelleArk: Hmm, looks like we're referencing `["table", "incremental"]` explicitly in this PR as well. Could we add a property/helper on the compiled node to indicate whether a materialization can enforce constraints that does this check under the hood? Something like `self.materialization_enforces_constraints`

core/dbt/contracts/graph/nodes.py
```diff
@@ -608,32 +648,96 @@ def same_contract(self, old) -> bool:
...
+                    and constraint_support[old_constraint.type] == ConstraintSupport.ENFORCED
+                ):
+                    enforced_model_constraint_removed.append(
+                        (str(old_constraint.type), old_constraint.columns)
+                    )
+
+        # Check for materialization changes.
+        if (
+            old.config.materialized in ["table", "incremental"]
+            and old.config.materialized != self.config.materialized
```
MichelleArk, line 717: should changing between table and incremental materializations actually raise a breaking contract error? I can see how going from a table -> view is effectively like deleting any existing constraints but going from table -> incremental should maintain any enforced constraints.

MichelleArk: along the same lines - we should probably only raise a breaking contract error if the materialization changes to one where constraints are unenforceable if were any enforceable (by the adapter) constraints defined to begin with. 

dbt can still validate contracts (column names + types) in materializations where constraints are unsupported. so if you move from a table -> view and there were no enforceable constraints to begin with or breaking changes to the contract, no error should be raised.

emmyoop: That makes sense.  Updated.

core/dbt/contracts/graph/nodes.py
```diff
@@ -572,6 +572,48 @@ def depends_on_nodes(self):
...
+
+    @property
+    def search_name(self):
+        if self.version is None:
+            return self.name
+        else:
+            return f"{self.name}.v{self.version}"
+
+    @property
+    def materialization_enforces_constraints(self):
```
MichelleArk, line 612: couple nits: 
- missing return type annotation `-> bool`
- simplifying to `return self.config.materialized in ["table", "incremental"]` feels more pythonic to me!

core/dbt/graph/selector_methods.py
```diff
@@ -567,17 +591,24 @@ def search(self, included_nodes: Set[UniqueId], selector: str) -> Iterator[Uniqu
         manifest: WritableManifest = self.previous_state.manifest
 
         for node, real_node in self.all_nodes(included_nodes):
```
MichelleArk, line 593: what do you think about refactoring this bit to look more like: 
```python
        for node, real_node in self.all_nodes(included_nodes):
            previous_node: Optional[SelectorTarget] = None

            if node in manifest.nodes:
                previous_node = manifest.nodes[node]
            elif node in manifest.sources:
                previous_node = manifest.sources[node]
            elif node in manifest.exposures:
                previous_node = manifest.exposures[node]
            elif node in manifest.metrics:
                previous_node = manifest.metrics[node]

            keyword_args = {}
            if checker.__name__ in ["same_contract", "check_modified_content"]:
                keyword_args["adapter_type"] = adapter_type  # type: ignore

            if checker(previous_node, real_node, **keyword_args):  # type: ignore
                yield node
```

My thinking is that it's more clear / readable to isolate the `adapter_type` optionality in keyword_args from setting `previous_node` and `real_node`. I tested it out locally but could be missing something on why it wouldn't work!

core/dbt/contracts/graph/nodes.py
```diff
@@ -608,32 +654,99 @@ def same_contract(self, old) -> bool:
         contract_enforced_disabled: bool = False
         columns_removed: List[str] = []
         column_type_changes: List[Tuple[str, str, str]] = []
+        enforced_column_constraint_removed: List[List[str]] = []  # column, constraint_type
```
MichelleArk, line 657: any reason not to use a list of `Tuple[str, str]` here and for `enforced_model_constraint_removed`? Feels more consistent with the implementation so far (of `columns_removed`) given the nested lists are of fixed length.

MichelleArk (approved):
Couple nits/suggestions, nothing blocking. Great testing! ✨

core/dbt/graph/selector_methods.py
emmyoop: This is better!

PR #7494, opened 2023-05-03
Adding Docker workflow to release pipeline

leahwicz:
### Description
This adds 2 new steps to the release workflow which will call the Docker workflow to build and publish the Docker images for `dbt-core` and `dbt-postgress`

I also updated the Docker workflow to leverage the sem-ver action and other action version bumps

This does not include testing of the Docker workflow in the testing of the overall release workflow. That can be done as a next step

### Checklist

### DIFF
--- a/.github/workflows/release-docker.yml
+++ b/.github/workflows/release-docker.yml
@@ -5,7 +5,7 @@
 # Docker images for dbt are used in a number of important places throughout the dbt ecosystem.  This is how we keep those images up-to-date.
 
 # **when?**
-# This is triggered manually
+# This is triggered manually or by the release workflow
 
 # **next steps**
 # - build this into the release workflow (or conversly, break out the different release methods into their own workflow files)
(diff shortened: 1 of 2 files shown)


.github/workflows/release-docker.yml
emmyoop, line 1: This workflow should live in the `dbt-release` repo to be consistent with the rest of the release.

.github/workflows/release-docker.yml
```diff
@@ -16,41 +16,48 @@ permissions:
...
+      package:
+       description: The package to release. _One_ of [dbt-core, dbt-redshift, dbt-bigquery, dbt-snowflake, dbt-spark, dbt-postgres]
+       required: true
+       type: string
   workflow_dispatch:
     inputs:
       package:
        description: The package to release. _One_ of [dbt-core, dbt-redshift, dbt-bigquery, dbt-snowflake, dbt-spark, dbt-postgres]
        required: true
+       type: string
```
emmyoop, line 30: You can make this [`type: choice`](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#:~:text=%27warning%27%0A%20%20%20%20%20%20%20%20type%3A-,choice,-options%3A%0A%20%20%20%20%20%20%20%20%2D) to make manual submission easier.

.github/workflows/release.yml
emmyoop, line 1: Need to add a needs for `docker-core-release` and `docker-postgres-release` down on line 239 for the slack notifications.

.github/workflows/release.yml
```diff
@@ -209,6 +210,23 @@ jobs:
...
+    if: ${{ !inputs.test_run }}
+    needs: [pypi-release]
+
+    uses: dbt-labs/dbt-core/.github/workflows/release-docker.yml@main
+    with:
+      package: dbt-core
+
+  docker-postgres-release:
+    name: Docker dbt-postgres Release
+    needs: [docker-core-release]
```
emmyoop, line 224: I would explicitly add this for clarity.
```suggestion
  docker-postgres-release:
    name: Docker dbt-postgres Release
    if: ${{ !inputs.test_run }}
    needs: [docker-core-release]
```

emmyoop (changes requested):
Overall this looks good.  Just a few comments.

I think the right move is to move the workflow over in `dbt-release` long term.  If we don't want to tackle that now we should definitely open an issue over there to get it done.

Issue #7444, opened 2023-04-24
[CT-2470] Package-namespaced macro resolution


MichelleArk:
Resources in a package should be rendered with a rendering context specific to that package (its vars and macros). This issue focuses on macros resolution.

**Acceptance Criteria**
* Macros defined in imported packages should be respected when compiling code defined in same imported package (this is already the case - just not for generate_x_macros)
* MacroManifest.find_generate_macro_by_name should consider imported packages if the calling node is defined in an imported package: https://github.com/dbt-labs/dbt-core/blob/fd7306643fd3a98e3232cbc15bf50eb2f3733603/core/dbt/contracts/graph/manifest.py#L575-L576
  * e.g. package_a uses its version of `get_custom_alias`, package_b uses its version of `get_custom_alias` during parse-time

* If the root project wants to override a package_a macro being called within a package_a model, it would be possible to achieve this desirable behavior with [dispatch](https://docs.getdbt.com/reference/project-configs/dispatch-config) — as long as the package_a macro in question is dispatched to the package_a namespace.

PR #7509, opened 2023-05-03
Package-namespaced generate_x_name macro resolution

MichelleArk:
resolves https://github.com/dbt-labs/dbt-core/issues/7444

### Description
* Updates `find_generate_macro_by_name` to accept an optional `imported_package` argument.
  * If `imported_package` is provided, `find_generate_macro_by_name` returns the associated `generate_x_name` macro if it exists in an associated imported package.
  *  Otherwise, return the default `generate_x_name` package (existing behaviour) 

* `RelationUpdate` constructs a `default_updater` by calling `find_generate_macro_by_name` without any packages, as well as a `package_updaters` mapping that contains the `generate_x_name` for each imported package. 
  * When called, `RelationUpdate` uses the imported package `generate_x_name` macro if the node is imported. If not, it falls back to the default `generate_x_name` (existing behaviour)

### Checklist


core/dbt/parser/base.py
```diff
@@ -71,20 +71,44 @@ def __init__(
...
             raise DbtInternalError(f"No macro with name generate_{component}_name found")
 
-        root_context = generate_generate_name_macro_context(macro, config, manifest)
-        self.updater = MacroGenerator(macro, root_context)
+        default_macro_context = generate_generate_name_macro_context(
+            default_macro, config, manifest
+        )
+        self.default_updater = MacroGenerator(default_macro, default_macro_context)
+
+        package_names = {macro.package_name for macro in manifest.macros.values()}
```
MichelleArk, line 86: @gshank - this doesn't seem too bad to do for each RelationUpdate instantiation (3 - one per component) per manifest (1-2). Is there somewhere where this information is already loaded during parsing? We could add it to the manifest but that would increase the scope of this PRs changes pretty significantly so I'd opt to do it as a follow-on if that feels like the right way forward.

MichelleArk: I updated this bit to grab the dependency names from RuntimeConfig instead of computing them based on nodes in the manifest 👍

jtcohen6 (approved):
I sent myself into a bit of a tizzy trying to figure out how this behavior should (or should not) theoretically intersect with `dispatch`. After taking a step back, I think there are three actual use cases that we want to be possible:

1. **Package always wins.** As the developer of dbt models in a project that will be deployed elsewhere as a package, I want full control over the macros used to define & materialize my models, including their `generate_x_name` rules. My macros should always take precedence for my models, and there should not be any way to override them.
2. **Conditional application (root project wins).** As the maintainer of one dbt project in a mesh of multiple, my team wants conditional application of these rules. When running my project standalone (e.g. in dev), I want my to apply my custom behavior; but when installed as a package and deployed alongside several other projects, I want the root-level project's `generate_x_name` rules to apply.
3. **Same rules everywhere all the time.** As a member of the data platform team responsible for consistency across teams at my organization, I want to create a "macro package" that every team can install & use. 

I wanted to validate that every use case described above is achievable, one way or another, even if the mechanism is a bit counterintutive. (Prior to this PR, 2+3 were possible; 1 was not.) I believe all three are now possible:
1. Each project/package fully overrides the macro by its name - `generate_schema_name`. Thanks to the changes in this PR, that implementation will apply for all resources in the project, whether it's being run from its root or installed as a package by another project.
2. Each package implements its "local" override by registering a candidate for dispatch with an adapter prefix, e.g. `default__generate_schema_name`. Then, the root-level project can register its own candidate for dispatch (`default__generate_schema_name`), winning the default search order, or by explicitly overriding the macro by name (`generate_schema_name`).
3. Create a standalone package of macros only, including `default__generate_schema_name`, and add project-level `dispatch` rules in every project's `dbt_project.yml`, as well as in the root "pass-through" project installing them all as packages.

---

A cool thing that I think you realized a long while ago, but which I only really came to appreciate in this testing, is that this same behavior always held for _other_ macros. So if I have a custom reimplementation of `create_view_as` defined in a package, it takes effect for the `view`-materialized models in that package, but not any other models. They will respect all the same rules described above. So really, this PR brings the `generate_x_name` macros into compliance with all other macro resolution — neat!

---

I'm not sure exactly if/how this behavior would make sense for _public_ modeling packages, maintained outside the organization. Which of the patterns would I want the Fivetran team to follow, for instance? I think the answer is "none of the above" — it's not really the use case we're targeting here — but (2) retains the most flexibility. I could always write some tricky logic to accomplish "conditional dispatch":
```yml
dispatch:
  - macro_namespace: dbt
    search_order: ['my_project', 'installed_package', 'dbt']
```
```sql
{% macro generate_schema_name(custom_schema_name, node) -%}
    {% if node.package_name == 'installed_package' %}
        {{ return(installed_package.generate_schema_name(custom_schema_name, node)) }}
    {% else %}
        -- my logic
        {{ generate_schema_name_for_env(custom_schema_name, node) }}
    {% endif %}
{%- endmacro %}
```

PR #7538, opened 2023-05-06
remove depends on when creating new manifest

ChenyuLInx:
fix compilation error when mocked model have a ref to another model that's not referenced by any unittest


ChenyuLInx:
Realized that if we remove the lineage here then the selection will no longer work for situations like
```
raw_order -> stg_order -> orders
```
adding this will result in `raw_order+` only selecting `stg_order` even if `orders` also have tests on it.

PR #7540, opened 2023-05-06
copy nodes being referenced over

ChenyuLInx:
This would resolve compilation error when there are nodes being referenced not copied to unittest manifest


MichelleArk:
Added manually in https://github.com/dbt-labs/dbt-core/commit/4dc6dd395edce960dfb0f563ea14376d118f2af9

PR #7545, opened 2023-05-08
fix #7390: push down limit filtering to adapter

### LINKED ISSUE #7390: `dbt show` should include `--limit` in compiled query
`dbt show --limit` will limit the number of rows of data previewed on the CLI.

Currently, `dbt show` applies a `limit` _after_ it's already run the query and loaded the results into memory:

https://github.com/dbt-labs/dbt-core/blob/121fa5793fda43aa62ece302114e9cf9ee7bb8a7/core/dbt/task/show.py#L20-L22

I would expect the query templated by `dbt show` to include `limit {limit}`.

Otherwise, dbt could run a query that returns 10k rows, load all of them into memory (in an `agate.Table`), and only then filter down to the default (`5`). My sense is that risks significant slowdown, and OOM issues.

**Acceptance criteria:**
- Instead of just directly executing `compiled_node.compiled_code`, template it into a subquery that includes `limit`
- In order to write the subquery in a cross-database-compatible way, we should either
     - Use a (Jinja) macro or materialization to template the subquery with `limit`
     - Use an adapter (Python) method — much easier to implement, and probably sufficient

---

**Out of scope:** Supporting BigQuery's no-cost "preview" API (`bq head`), which would require us to do a diff check / cache validation on the logical/applied state of this model. (logical = current SQL, applied = most recently materialized) Let's keep thinking about if there are ways to support that in the future.

aranke:
resolves #7390 

### Checklist

### DIFF
--- a/tests/functional/show/test_show.py
+++ b/tests/functional/show/test_show.py
@@ -100,7 +100,7 @@ class TestShow:
         dbt_args = ["show", "--inline", models__second_ephemeral_model, *args]
         results, log_output = run_dbt_and_capture(dbt_args)
         assert len(results.results[0].agate_table) == expected
-    
+
     def test_seed(self, project):
         (results, log_output) = run_dbt_and_capture(["show", "--select", "sample_seed"])
         assert "Previewing node 'sample_seed'" in log_output


tests/functional/show/test_show.py
```diff
@@ -85,3 +85,24 @@ def test_second_ephemeral_model(self, project):
...
+        assert len(results.results[0].agate_table) == 5
+
+    def test_limit_positive_number(self, project):
+        run_dbt(["build"])
+        (results, log_output) = run_dbt_and_capture(
+            ["show", "--inline", models__second_ephemeral_model, "--limit", 3]
+        )
+        assert len(results.results[0].agate_table) == 3
+
+    def test_limit_negative_number(self, project):
```
mikealfare, line 103: I'm probably just missing context. Why would a negative number return 7 records? Is that the entire dataset?

tests/functional/show/test_show.py
```diff
@@ -85,3 +85,24 @@ def test_second_ephemeral_model(self, project):
             ["show", "--inline", models__second_ephemeral_model]
         )
         assert "col_hundo" in log_output
+
+    def test_limit_default(self, project):
```
mikealfare, line 89: This is a nice use case for `@pytest.mark.parameterize`:
```python
@pytest.mark.parameterize("args,expected", [
    ([], 5),
    (["--limit", 3], 3),
    (["--limit", -1], 7),
])
def test_limit(self, project, args, expected):
    run_dbt(["build"])
    dbt_args = ["show", "--inline", models__second_ephemeral_model]
    dbt_args.extend(args)
    results, _ = run_dbt_and_capture(dbt_args)
    assert len(results.results[0].agate_table) == expected
```

.changes/unreleased/Fixes-20230508-060926.yaml
```diff
@@ -0,0 +1,6 @@
+kind: Fixes
```
mikealfare, line 1: Is this a fix or a feature? In other words, did we expect it to be at the cursor level and we just didn't do that? Or did users expect it to be at the cursor level and we're meeting their needs now?

core/dbt/adapters/sql/connections.py
```diff
@@ -138,13 +141,13 @@ def data_type_code_to_name(cls, type_code: Union[int, str]) -> str:
...
     def execute(
-        self, sql: str, auto_begin: bool = False, fetch: bool = False
+        self, sql: str, auto_begin: bool = False, fetch: bool = False, limit: Optional[int] = None
     ) -> Tuple[AdapterResponse, agate.Table]:
         sql = self._add_query_comment(sql)
         _, cursor = self.add_query(sql, auto_begin)
         response = self.get_response(cursor)
         if fetch:
-            table = self.get_result_from_cursor(cursor)
+            table = self.get_result_from_cursor(cursor, limit)
```
dataders, line 150: because `limit` is only relevant if `fetch=true`, might it make more sense to have `fetch` become an `int` where:
- `0`: don't fetch i.e. `false`
- `-1`: fetch all
- `n`: fetch `n` rows

dataders (commented):
love a great simpl impl (simple implementation). my first though is one @jtcohen6 already brought up in the source issue
> In order to write the subquery in a cross-database-compatible way, we should either
>     - Use a (Jinja) macro or materialization to template the subquery with limit
>     - Use an adapter (Python) method — much easier to implement, and probably sufficient

with any mention of `LIMIT` my "incompatible with TSQL" sense tingles and prompts me to think
> how will this work with SELECT TOP N?
> Is this supported by `pyodbc` and MSFT's `msodbc18`?

However, the use of `limit` here is semantic. In reality, the change here is to make use of a DB API 2.0 Cursor method, `fetchmany()` ([PEP 249: fetchmany()](https://peps.python.org/pep-0249/#fetchmany)).

Now my brain is thinking about how our ConnectionManager class methods map to PEP249. As a less-in-the-weeds person, I have a hard time understanding when dbt makes use of all the execute* and fetch* variants of the DB API 2.0.

Also, is there a clear implementation recommendation for non-dbapi2 adapters (e.g. BQ, DB, and (increasingly RS)

.changes/unreleased/Fixes-20230508-060926.yaml
mikealfare: Ah, makes sense.

core/dbt/adapters/sql/connections.py
mikealfare: I would be hesitant to change the type and function of `fetch`. This would certainly break our tests, which we can fix as part of the PR. But it would also break tests for a lot of community maintainers. Perhaps this is the right way to handle it, but my recommendation is to leave that as a 2.0 thing. Also, we would lose the option of a default `limit`; maybe not a big deal, but worth calling out.

mikealfare (approved):

jtcohen6:
@aranke This is causing failures in `dbt-bigquery`, when I run locally and [in automated tests](https://github.com/dbt-labs/dbt-bigquery/actions/runs/4936330986/jobs/8823728537):

```
13:49:25  Compilation Error in model my_model (models/my_model.sql)
13:49:25    execute() got an unexpected keyword argument 'limit'
13:49:25
13:49:25    > in macro create_or_replace_view (macros/materializations/models/view/create_or_replace_view.sql)
13:49:25    > called by macro materialization_view_bigquery (macros/materializations/view.sql)
13:49:25    > called by model my_model (models/my_model.sql)
```

`dbt-bigquery` overrides the `execute` method [here](https://github.com/dbt-labs/dbt-bigquery/blob/f717e83927e5ac71c4a085985ecc95e7c0c70099/dbt/adapters/bigquery/connections.py#L459-L461). Granted, `dbt-bigquery` is something we maintain, so we can make this change ourselves — but this could impact other externally-maintained adapters as well. Is there any way to make this change without breaking/modifying the signature of the `execute` method?

Issue #5267, opened 2022-05-18
[CT-662] [Feature] Allow the configuration of the timezone for timestamps in event logging


Stefan-Rann:
### Is this your first time opening an issue?

- [X] I have read the [expectations for open source contributors](https://docs.getdbt.com/docs/contributing/oss-expectations)

### Describe the Feature

Currently the timestamp in event logging is always output in utc.
It would be helpful to be able to set the time zone via a configuration or parameter (either local or any other time zone).

### Describe alternatives you've considered

I have not found any alternatives. One has to live with utc.

### Who will this benefit?

This would make it easier to evaluate the logs and help to avoid confusion. 
Additionally the evaluation of dbt logs in combination with other logs using lokal or any other timezone would be easier.

gshank:
This sounds like a reasonable request. We're going to be looking at logging updates later in this quarter. I'm not sure if it would get onto the schedule or not, but we can take a look.

jtcohen6:
Reopening because we just got a duplicate opened in https://github.com/dbt-labs/dbt-core/issues/6319

> Things I'm curious about: How would you want to configure this? Should there only be two options (local system time versus UTC)?

ahuiiiiiii:
> Reopening because we just got a duplicate opened in #6319
> 
> > Things I'm curious about: How would you want to configure this? Should there only be two options (local system time versus UTC)?

can i add an option that set zonetime in dbt-core project configuration file?

ahuiiiiiii:
sorry spelling error above should be timezone

Stefan-Rann:
> can i add an option that set zonetime in dbt-core project configuration file?

Yes, a project level setting would be cool to set the timezone by name.
Additionally the option "local" would be great to use the system timezone without the need to explicitly define the timezone.

Gunnnn:
Please! I see timestamps in log in UTC and its annoying =(

Frunkh:
Atleast change 
`record.extra[self.name] = datetime.utcnow().isoformat()`
to 
`record.extra[self.name] = datetime.now().isoformat()`

in logger.py
So it respects the timezone of the system.. this is really annoying.

sgaraycoa-de:
Adding my vote in favor of this issue

jtcohen6:
[...]

PR #7562, opened 2023-05-09
Model Deprecation

### LINKED ISSUE #7433: [CT-2461] Support deprecating models
**Acceptance Criteria**

**Parsing:** 
  * Any model (not just versioned ones) can provide a non-null `deprecation_date` as a top-level property (not under config). Other node types (e.g. seeds, snapshots, etc) can not. 
    * Unparsed `deprecation_date` is a string - formatted as a date, optionally with a timezone offset (e.g. `'2020-10-23+00:00'` or equivalently `'2020-10-23'`. Parsed representation of `deprecation_date` on `ModelNode` should be a datetime (parsed from one of the two supported formats)
    * Models with `deprecation_date` values that are set to a past date (as compared to `datetime.now(pytz.utc)`) should raise a `DeprecatedModel` warning.
      * This is not feasible to do during initial parsing of `ModelNode` objects because we'd need to re-evaluate whether to raise `DeprecatedModel` based on the current time, __even__ when partial parsing. So we'd need to do it once the manifest is loaded (perhaps here: https://github.com/dbt-labs/dbt-core/blob/main/core/dbt/parser/manifest.py#L434)

**`ref` behaviour:** 
* parse-time references to a model that has a deprecation_date set to a: 
   * future date (as compared to `datetime.now(pytz.utc)`) -> raise an `UpcomingDeprecationReference` warning, containing both the deprecation date and its relative time
   * past date (as compared to `datetime.now(pytz.utc)`) -> raise a `DeprecatedReference` warning
   * (We'd want these two to be separate so they can be configured for error-promotion via WARN_ERROR_OPTIONS individually)

**Warning messages:**
  * DeprecatedModel (warning when parsing a project that defines deprecated model(s)): `[WARNING] Model {model_name}[.v{version} if versioned] has passed its deprecation date of {deprecation_date}. This model should be disabled or removed.`
  * UpcomingDeprecationReference (warning when referencing a model with a future deprecation date): `[WARNING] While compiling '{this_model_name}': Found a reference to {ref_model_name}[.v{version} if versioned], which is slated for deprecation on '{ref_model_deprecation_date}'. [if versioned: A new version of '{ref_model_name}' is available. Try it out: {{ ref('ref_model_package', 'ref_model_name', v='{ref_model_latest_version}') }}.`
  * DeprecatedReference (warning when referencing a model with a past deprecation date): `[WARNING] While compiling '{this_model_name}': Found a reference to {ref_model_name}[.v{version} if versioned], which was deprecated on '{ref_model_deprecation_date}'. [if versioned: A new version of '{ref_model_name}' is available. Migrate now: {{ ref('ref_model_package', 'ref_model_name', v='{ref_model_latest_version}') }}.`
  
For inspiration: [`UnpinnedRefNewVersionAvailable`](https://github.com/dbt-labs/dbt-core/blob/d2f963e20ea945355db57ff3d34a318429a8bf1e/core/dbt/events/types.py#L1121-L1135). I would also be happy if we cleaned up that "info"-level log, perhaps by moving it to an end-of-run summary. We don't have a great way to surface & summarize warnings on the CLI currently—and facilitating this type of cross-team communication **is** the feature!

peterallenwebb:
resolves #7433

### Description

Support model deprecation, with appropriate warnings and errors.

### Checklist

    - [x] https://github.com/dbt-labs/docs.getdbt.com/issues/3357


core/dbt/parser/manifest.py
```diff
@@ -125,6 +126,36 @@
...
+def extended_mashumuro_decoder(data):
+    return msgpack.unpackb(data, ext_hook=extended_msgpack_decoder, raw=False)
+
+
+def extended_msgpack_decoder(code, data):
+    if code == 1:
+        d = datetime.date.fromisoformat(data.decode())
+        return d
+    elif code == 2:
+        dt = datetime.datetiem.fromisoformat(data.decode())
```
dbeatty10, line 153: This looks like a small misspelling:
```suggestion
        dt = datetime.datetime.fromisoformat(data.decode())
```

core/dbt/parser/manifest.py
```diff
@@ -867,14 +898,14 @@ def read_manifest_for_partial_parse(self) -> Optional[Manifest]:
...
-                manifest: Manifest = Manifest.from_msgpack(manifest_mp)  # type: ignore
+                manifest: Manifest = Manifest.from_msgpack(manifest_mp, decoder=extended_mashumuro_decoder)  # type: ignore
                 # keep this check inside the try/except in case something about
                 # the file has changed in weird ways, perhaps due to being a
                 # different version of dbt
                 is_partial_parsable, reparse_reason = self.is_partial_parsable(manifest)
                 if is_partial_parsable:
                     # We don't want to have stale generated_at dates
-                    manifest.metadata.generated_at = datetime.utcnow()
+                    manifest.metadata.generated_at = datetime.datetime.utcnow()
```
dbeatty10, line 908: `dt.utcnow()` yields a naive timestamp rather than an aware one.

Maybe we want an aware datetime like `dt.now(timezone.utc).isoformat()` instead?

☝️ assuming an import like this:
```python
from datetime import timezone, datetime as dt
```

Slightly more detail [here](https://github.com/dbt-labs/dbt-core/issues/5267#issuecomment-1513306428).

| Python                                    | Output                             |   | UTC Offset       |
|-------------------------------------------|------------------------------------|------------------|------------------|
| `dt.utcnow().isoformat()`           | `2023-04-18T13:42:18.050150`       | Naive            | N/A              |
| `dt.now().isoformat()`              | `2023-04-18T07:42:18.050150`       | Naive            | N/A              |
| `dt.now().astimezone().isoformat()` | `2023-04-18T07:42:18.050150-06:00` | Aware            | System time zone |
| `dt.now(timezone.utc).isoformat()`  | `2023-04-18T13:42:18.050150+00:00` | Aware            | UTC              |

gshank:
Wow. This was a lot more complicated than anticipated, but does seem like functionality that we want to have.

At one point the server code was also creating a msgpack'd manifest, so we might want to check their codebase to see if they need to use this too.

core/dbt/parser/manifest.py
```diff
@@ -125,6 +126,36 @@
...
 
 
+def extended_mashumaro_encoder(data):
+    return msgpack.packb(data, default=extended_msgpack_encoder, use_bin_type=True)
+
+
+def extended_msgpack_encoder(obj):
+    if type(obj) is datetime.date:
+        date_bytes = msgpack.ExtType(1, obj.isoformat().encode())
+        return date_bytes
```
MichelleArk, line 136: The [msgpack docs](https://github.com/msgpack/msgpack-python#packingunpacking-of-custom-data-type)  outline an alternative method to encoding custom types by returning something more like: 
```python
{'__datetime__': True, 'as_str': obj.strftime("%Y%m%dT%H:%M:%S.%f")}
```

It looks to be interchangeable with `ExtType` usage as you've implemented. The `ExtType` approach feels more readable + extensible to custom data types beyond datetime

core/dbt/parser/manifest.py
```diff
@@ -125,6 +126,36 @@
 PERF_INFO_FILE_NAME = "perf_info.json"
 
 
+def extended_mashumaro_encoder(data):
+    return msgpack.packb(data, default=extended_msgpack_encoder, use_bin_type=True)
+
+
+def extended_msgpack_encoder(obj):
+    if type(obj) is datetime.date:
```
MichelleArk, line 134: if deprecation_date always be a datetime - do we strictly need this case for the deprecation_date addition? or is it just provided here in the event that we do want to serialize a date object in the future

dbeatty10: Piggy-backing off your question @MichelleArk, here are some of my beliefs:

The happiest path is if `deprecation_date` is converted to an "aware" datetime in Python that can compare with another "aware" datetime that represents "now" for dbt purposes.

That way, the user can specify an unambiguous timestamp in [RFC 3339](https://dbeatty10.github.io/rfc3339-iso8601/) format like `2023-05-09 11:15:00-0600`, and it can compare nicely to the current system time like `2023-05-09 17:13:58Z`.

Anything other than two aware datetimes is likely to be confusing to dbt users and/or dbt-core developers.

peterallenwebb: @MichelleArk If the user puts a date into the deprecation_date field, such as '2023-05-09', the YAML parser produces a raw parsed object which contains a date, rather than a date-time. That raw parser output is one of the things that gets put in the manifest, which is why I have included date. Writing and reading the manifest fails, otherwise.

peterallenwebb: @dbeatty10 I agree, and will work on getting dates converted to timezone-aware as early as possible in the parsing process.

dbeatty10:
@MichelleArk For the implementation, would all models have their `deprecation_date`s compared against a shared `datetime.now(pytz.utc)`? Or would they each get their own `datetime.now(pytz.utc)` depending on when their `deprecation_date` is evaluated?

When the dbt Developer Experience (DX) team was first reviewing the design for `deprecated_date` as part of dbt model versions, @joellabes raised the idea of using [`run_started_at`](https://docs.getdbt.com/reference/dbt-jinja-functions/run_started_at) as a consistent datetime to use across all models in a single dbt execution.

There could be other ways of generating a shared `datetime.now(pytz.utc)` too besides `run_started_at`, but it seemed like an easy one to reach for.

peterallenwebb:
@dbeatty10 I like the idea of comparing against `run_started_at`. I'll try that.

dbeatty10:
> @dbeatty10 Do you think we need a documentation case for this work?

@peterallenwebb Yes, indeed -- thanks for noticing and asking!

Just created the relevant placeholder issue for the related documentation [here](https://github.com/dbt-labs/docs.getdbt.com/issues/3357) and checked it off in the checklist:

image

core/dbt/parser/manifest.py
peterallenwebb: I am adding this to my list of Known Python Facts, but since I did not add this call, and am just changing the way the namespace is imported, I'm inclined to leave it alone for now.

core/dbt/parser/manifest.py
```diff
@@ -497,8 +532,46 @@ def load(self):
...
+        self.check_for_model_deprecations()
+
         return self.manifest
 
+    def check_for_model_deprecations(self):
+        for node in self.manifest.nodes.values():
+            if isinstance(node, ModelNode):
+                if (
+                    node.deprecation_date
+                    and node.deprecation_date < datetime.datetime.now().astimezone()
```
dbeatty10, line 544: @peterallenwebb want to double-check two things with you:
1. naive `deprecation_date`s
2. single stable time

### naive `deprecation_date`s
I believe this would be dependent on the system time wherever the python is being executed. So we'll probably want to double-check that any `deprecation_date` expressed without an offset like `2023-05-17` is understood to have its offset from the system time zone.

### single stable time
Were you able to try out something akin to [`run_started_at`](https://docs.getdbt.com/reference/dbt-jinja-functions/run_started_at)?

It might be useful for using a single stable value over the course of the entire execution. It might also be useful from a testing perspective as well.

gshank (approved):
Looks good, just need to merge main.

core/dbt/contracts/publication.py
```diff
@@ -45,6 +45,7 @@ class PublicModel(dbtClassMixin, ManifestOrPublicNode):
     # list of model unique_ids
     public_node_dependencies: List[str] = field(default_factory=list)
     generated_at: datetime = field(default_factory=datetime.utcnow)
+    deprecation_date: Optional[datetime] = None
```
MichelleArk, line 48: ✨

core/dbt/parser/manifest.py
```diff
@@ -511,8 +545,46 @@ def load(self):
...
+                        else:
+                            event_cls = UpcomingReferenceDeprecation
+
+                        fire_event(
+                            event_cls(
+                                model_name=node.name,
+                                ref_model_package=resolved_ref.package_name,
+                                ref_model_name=resolved_ref.name,
+                                ref_model_version=str(resolved_ref.version),
+                                ref_model_latest_version=str(resolved_ref.latest_version),
```
MichelleArk, line 583: ditto here and on the line above re: string conversion and error messages: https://github.com/dbt-labs/dbt-core/pull/7562/files#r1199287517

core/dbt/events/types.py
```diff
@@ -1146,6 +1146,62 @@ def message(self) -> str:
...
 
+class DeprecatedModel(WarnLevel):
+    def code(self):
+        return "I065"
+
+    def message(self) -> str:
+        version = ".v" + self.model_version if self.model_version else ""
+        return (
+            f"Model {self.model_name}{version} has passed its deprecation date of {self.deprecation_date}."
+            "This model should be disabled or removed."
```
MichelleArk, line 1157: nit: missing space between this line and the previous one

core/dbt/parser/manifest.py
```diff
@@ -511,8 +545,46 @@ def load(self):
...
+        for node in self.manifest.nodes.values():
+            if isinstance(node, ModelNode):
+                if (
+                    node.deprecation_date
+                    and node.deprecation_date < datetime.datetime.now().astimezone()
+                ):
+                    fire_event(
+                        DeprecatedModel(
+                            model_name=node.name,
+                            model_version=str(node.version),
```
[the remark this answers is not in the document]
peterallenwebb: Good point. I'm not sure why I did this.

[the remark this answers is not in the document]
peterallenwebb: The unit tests reminded me that it is because of protobuf serialization, but I'll find a better compromise.

core/dbt/contracts/graph/unparsed.py
```diff
@@ -652,3 +659,18 @@ def validate(cls, data):
...
+
+def normalize_date(d: Optional[datetime.date]) -> Optional[datetime.datetime]:
+    """Convert date to datetime (at midnight), and add local time zone if naive"""
+    if d is None:
+        return None
+
+    # convert date to datetime
+    dt = d if type(d) == datetime.datetime else datetime.datetime(d.year, d.month, d.day)
+
+    if not dt.tzinfo:
```
dbeatty10, line 672: The Python docs [here](https://docs.python.org/3/library/datetime.html#determining-if-an-object-is-aware-or-naive) say the following:

image

Translating those into code gives something like [this](https://github.com/dbt-labs/dbt-core/blob/265e09dc93914a14caf0cb0ee778264e20cf75c8/tests/adapter/dbt/tests/adapter/utils/test_current_timestamp.py#L16-L17).

The slight change suggested below would cover the case where `tzinfo` is not `None` but `dt.tzinfo.utcoffset(dt)` returns `None`.

To fully encode the Python datetime definition, I think we'd need to do something like this:
```suggestion
    if not dt.tzinfo or not dt.tzinfo.utcoffset(dt):
```

Granted, it might "never" happen for a `tzinfo` object from `pytz` to return `None` for the offset, but I was able to hand construct such a `tzinfo` object locally.

tests/functional/deprecations/model_deprecations.py
```diff
@@ -0,0 +1,78 @@
...
+"""
+
+deprecating_model__yml = """
+version: 2
+
+models:
+  - name: my_model
+    description: deprecating in the future
+    deprecation_date: 2999-01-01
+"""
```
dbeatty10, line 21: Are `deprecation_date` formats like the following accepted?
- `2022-04-08T22:00:00.0-04:00`
- `2022-04-08 22:00:00.0-04:00`
- `2022-04-08T22:00:00.0`
- `2022-04-08 22:00:00.0`

If so, could we add test cases for the relevant ones?

@MichelleArk Could you or someone else add which cases do or don't work in the docs issue [here](https://github.com/dbt-labs/docs.getdbt.com/issues/3357)?

dbeatty10 (commented):
Left a few comments on some edge cases.

tests/functional/deprecations/model_deprecations.py
peterallenwebb: @dbeatty10 We are using PyYAML to parse JSON, and relying on its date detection and parsing for now, which claims to follow the YAML spec. I'll figure out what we can say definitively about the dates we accept.

core/dbt/contracts/graph/unparsed.py
peterallenwebb: I'd be happy to tighten up this check if you're willing to open the issue. Sorry we didn't get it in v1.

PR #7577, opened 2023-05-09
CT 2510 Throw error for duplicate versioned and non versioned model names

### LINKED ISSUE #7487: [CT-2510] [Bug] Raise an error if non-versioned model is defined with same name as existing versioned model
Set up:
```sql
-- models/my_model_v1.sql
select 1 as id
```
```yml
# models/whatever.yml
models:
  - name: my_model
    versions:
      - v: 1
```

```
$ dbt parse
12:48:12  Running with dbt=1.5.0
12:48:14  Performance info: target/perf_info.json
```
Now:
- create `models/my_model.sql` (no version suffix)
- do not add any new versions to the `versions` block in the yaml file

Re-parse with partial parsing, and everything looks fine:
```
$ dbt parse
12:48:29  Running with dbt=1.5.0
12:48:30  Performance info: target/perf_info.json
```
And now without partial parsing, I see an error (good! I should!):
```
$ dbt --no-partial-parse parse
12:48:43  Running with dbt=1.5.0
12:48:46  Encountered an error:
Compilation Error
  dbt found two models with the name "my_model".

  Since these resources have the same name, dbt will be unable to find the correct resource
  when looking for ref("my_model").

  To fix this, change the name of one of these resources:
  - model.my_dbt_project.my_model (models/my_model.sql)
  - model.my_dbt_project.my_model.v1 (models/my_model_v1.sql)
```

gshank:
resolves #7487

### Description

The _check_resource_uniqueness method in core/dbt/parser/manifest.py has been modified to check for non-versioned models with the same "name" as versioned models.

### Checklist

### DIFF
--- a/core/dbt/exceptions.py
+++ b/core/dbt/exceptions.py
@@ -2201,7 +2201,8 @@ Since these resources have the same name, dbt will be unable to find the correct
 when looking for ref({self.versioned_node.name}).
 
 To fix this, change the name of the unversioned resource:
-- {self.unversioned_node.unique_id} ({self.unversioned_node.original_file_path})
+{self.unversioned_node.unique_id} ({self.unversioned_node.original_file_path})
+or add the unversioned model to the versions in {self.versioned_node.patch_path}
     """.strip()
         return msg
 
--- a/core/dbt/exceptions.py
+++ b/core/dbt/exceptions.py
@@ -2187,6 +2187,25 @@ To fix this, change the name of one of these resources:
         return msg
 
 
+class DuplicateVersionedUnversionedError(ParsingError):
+    def __init__(self, versioned_node, unversioned_node):
+        self.versioned_node = versioned_node
+        self.unversioned_node = unversioned_node
+        super().__init__(msg=self.get_message())
+
+    def get_message(self) -> str:
+        msg = f"""
+dbt found versioned and unversioned models with the name "{self.versioned_node.name}".
+
+Since these resources have the same name, dbt will be unable to find the correct resource
+when looking for ref({self.versioned_node.name}).
+
+To fix this, change the name of the unversioned resource:
+- {self.unversioned_node.unique_id} ({self.unversioned_node.original_file_path})
+    """.strip()
+        return msg
+
+
 class PropertyYMLError(CompilationError):
     def __init__(self, path: str, issue: str):
         self.path = path
--- a/core/dbt/exceptions.py
+++ b/core/dbt/exceptions.py
@@ -2198,9 +2198,9 @@ class DuplicateVersionedUnversionedError(ParsingError):
 dbt found versioned and unversioned models with the name "{self.versioned_node.name}".
 
 Since these resources have the same name, dbt will be unable to find the correct resource
-when looking for ref({self.versioned_node.name}).
+when looking for ref('{self.versioned_node.name}').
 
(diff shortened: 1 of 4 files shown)


core/dbt/exceptions.py
```diff
@@ -2187,6 +2187,25 @@ def get_message(self) -> str:
...
+        self.versioned_node = versioned_node
+        self.unversioned_node = unversioned_node
+        super().__init__(msg=self.get_message())
+
+    def get_message(self) -> str:
+        msg = f"""
+dbt found versioned and unversioned models with the name "{self.versioned_node.name}".
+
+Since these resources have the same name, dbt will be unable to find the correct resource
+when looking for ref({self.versioned_node.name}).
```
MichelleArk, line 2201: tiny nit - adding quotes around the node: `ref('{self.versioned_node.name}')`

core/dbt/parser/manifest.py
```diff
@@ -1280,6 +1287,17 @@ def _check_resource_uniqueness(
 
         alias_resources[full_node_name] = node
 
+    intersection_versioned = set(versioned_resources.keys()).intersection(
+        set(unversioned_resources.keys())
+    )
+    if intersection_versioned:
```
MichelleArk, line 1293: I think raising this sets us back in terms of being able to handle ref-able resource with duplicate names across packages. For example - if I define project_a.model (unversioned) and project_b.model.v1 (versioned) - those work when run in their respective projects as root, but when installed as packages would raise `DuplicateVersionedUnversionedError`

gshank: I'm thinking that we need to store the versioned_resources and unversioned_resources by package. Maybe we'd only need to check the root package?

gshank: I've updated the code to check versioned/unversioned by packages and tweaked the error message to also suggest adding the unversioned model to the versions.

core/dbt/exceptions.py
gshank: Added quotes.

core/dbt/parser/manifest.py
```diff
@@ -1280,6 +1288,18 @@ def _check_resource_uniqueness(
 
         alias_resources[full_node_name] = node
 
+    for ver_unver_dict in name_resources.values():
+        versioned_names = ver_unver_dict["ver"].keys()
+        unversioned_names = ver_unver_dict["unver"].keys()
```
MichelleArk, line 1293: Non blocking at all, just a suggestion: 

I think it'd be more straightforward to read this method if `name_resources` just have had level of nesting (packages) and omit "ver" and "unver". and then construct the verioned_names and un_versioned names at this point by: 

```python
for name_resources in name_resources_by_package.values():
  versioned_names = {resource.name for resource in name_resources if resource.is_versioned}
  versioned_names = {resource.name for resource in name_resources if not resource.is_versioned}
```

just a little bit less state/nesting to keep in mind while reading.

MichelleArk (approved):
Looks good, although I've found an unrelated bug in version parsing across packages: https://github.com/dbt-labs/dbt-core/issues/7593

core/dbt/parser/manifest.py
gshank: It kind of bothers me that this way of doing it means looping over an n-sized array 3 times instead of 1 though.

PR #7626, opened 2023-05-15
Feature/drop relation/ct 2581

### LINKED ISSUE #7625: [CT-2581] [Feature] Allow for more flexible drop statements in drop_relation()

### Describe the feature

The current implementation of `drop_relation()` looks like this:
```
{% macro drop_relation(relation) -%}
    {{ return(adapter.dispatch('drop_relation', 'dbt')(relation)) }}
{% endmacro %}

{% macro default__drop_relation(relation) -%}
    {% call statement('drop_relation', auto_begin=False) -%}
        drop {{ relation.type }} if exists {{ relation }} cascade
    {%- endcall %}
{% endmacro %}
```
That forces a specific string for `relation.type`. In particular, it forces spaces in `relation.type` for some scenarios (e.g. materialized views) and it forces the dbt implementation of a model to match that of the underlying database. The latter is an issue if, as an example, dbt decides to implement "dbt materialized views" in `dbt-snowflake` by using dynamic tables. The issue in the end is that `relation.type` is playing two roles that can't always be guaranteed to overlap.

The proposal is to turn `drop_relation()` into a dispatch function and create functions such as `drop_table()`, `drop_materialized_view()`, etc. There should still be a default condition in `drop_relation()` in order to preserve backwards compatibility. The proposal would look something like this:
```
{% macro drop_relation(relation) -%}
    {{ return(adapter.dispatch('drop_relation', 'dbt')(relation)) }}
{% endmacro %}

{% macro default__drop_relation(relation) -%}
    {% call statement('drop_relation', auto_begin=False) -%}
        {%- if relation.type == 'table' -%}
            {{- drop_table(relation) -}}
        {%- elif relation.type == 'materialized_view' -%}
            {{- drop_materialized_view(relation) -}}
        {%- else -%}
            drop {{ relation.type }} if exists {{ relation }} cascade
        {%- endif -%}
    {%- endcall %}
{% endmacro %}


{% macro drop_table(relation) -%}
  {{ return(adapter.dispatch('drop_table', 'dbt')(relation)) }}
{% endmacro %}

{% macro default__drop_table(relation) -%}
    drop table if exists {{ relation }} cascade
{% endmacro %}


{% macro drop_materialized_view(relation) -%}
  {{ return(adapter.dispatch('drop_materialized_view', 'dbt')(relation)) }}
{% endmacro %}

{% macro default__drop_materialized_view(relation) -%}
    drop materialized view if exists {{ relation }} cascade
{% endmacro %}
```

An alternative would be to keep the current approach. This would mean creating dbt materializations that match the underlying database's implementation. In this scenario, there would be no materialized view implementation for `dbt-snowflake` because we implement that concept via dynamic tables. Likewise, there would be no dynamic table implementation for any other adapter besides `dbt-snowflake`.

Another alternative is to overwrite `drop_relation()` for each adapter where necessary. That would address materialized view / dynamic table scenario since we could check for that in the `dbt-snowflake` adapter. But that would encourage overwriting `drop_relation()` in general, which is a central macro (it gets used in almost every run). The goal of the proposal above is to allow adapter maintainers to configure how an object is dropped without the need to worry about dbt-level functionality (e.g. logging, how to implement `call statement`, etc.).

### Who will this benefit?

This will benefit adapter maintainers and `dbt-core` maintainers. Adapter maintainers can be more flexible with how they manage the drop logic while also being less concerned with breaking `dbt-core` logic (or losing logic added in the future). `dbt-core` maintainers can be more confident that any logic added in the future will be implemented while still allowing adapter maintainers a path for configuration.

Yes

### Anything else?

_No response_

mikealfare:
resolves #7625

### Description

Allow for a more configurable `drop_relation()` macro; preserve backwards compatibility.

- create `adapters/drop_relation.sql`
- update `drop_relation()` to condition on `relation.type` and dispatch to the appropriate macro
- create `drop_table()`, `drop_view()`, and `drop_materialized_view()`

### Checklist

### DIFF
--- a/core/dbt/include/global_project/macros/adapters/drop_relation.sql
+++ b/core/dbt/include/global_project/macros/adapters/drop_relation.sql
@@ -19,26 +19,26 @@
 
 {% macro drop_table(relation) -%}
   {{ return(adapter.dispatch('drop_table', 'dbt')(relation)) }}
-{% endmacro %}
+{%- endmacro %}
 
 {% macro default__drop_table(relation) -%}
     drop table if exists {{ relation }} cascade
-{% endmacro %}
+{%- endmacro %}
 
 
 {% macro drop_view(relation) -%}
   {{ return(adapter.dispatch('drop_view', 'dbt')(relation)) }}
-{% endmacro %}
+{%- endmacro %}
 
 {% macro default__drop_view(relation) -%}
     drop view if exists {{ relation }} cascade
-{% endmacro %}
+{%- endmacro %}
 
 
 {% macro drop_materialized_view(relation) -%}
   {{ return(adapter.dispatch('drop_materialized_view', 'dbt')(relation)) }}
-{% endmacro %}
+{%- endmacro %}
 
 {% macro default__drop_materialized_view(relation) -%}
     drop materialized view if exists {{ relation }} cascade
-{% endmacro %}
+{%- endmacro %}
--- a/core/dbt/include/global_project/macros/adapters/drop_relation.sql
+++ b/core/dbt/include/global_project/macros/adapters/drop_relation.sql
@@ -7,7 +7,7 @@
         {%- if relation.is_table -%}
             {{- drop_table(relation) -}}
         {%- elif relation.is_view -%}
-            {{- drop_table(relation) -}}
+            {{- drop_view(relation) -}}
         {%- elif relation.is_materialized_view -%}
             {{- drop_materialized_view(relation) -}}
         {%- else -%}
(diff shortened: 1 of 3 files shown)


core/dbt/include/global_project/macros/adapters/drop_relation.sql
mikealfare, line 1: I moved all of the `drop_relation()` macros into their own file because now there's several macros that are all related to each other. However, I could also see an argument for keeping `drop_relation()` and `default__drop_relation()` in `adapters/relation.sql` to act as an interface, and then putting `drop_view()` in a file `adapter/view.sql` along with similar implementations of other macros (e.g. `create_view()`, `rename_view()`, etc.).

PR #7640, opened 2023-05-16
Move node patch method to schema parser patch_node_properties and refactor schema parsing

### LINKED ISSUE #7430: Move ParsedNode.patch and Macro.patch to respective PatchParsers
Currently a couple node types (Macro, ParsedNode) implement their own `patch` method which accepts a patch (either ParsedNodePatch or ParsedMacroPatch) and are responsible for applying the contents of the patch to the node attributes. The PatchParsers then call these node `patch` methods once after constructing a patch. 

This splits the overall patching responsibilities into creating a patch and applying it across the PatchParsers and Nodes (which is not really that clean of a split in reality because NodePatchParser implement its own `patch_node_config` method that applies patches to node configs specifically). Most of the complexity is in creating a patch rather than 'applying' it, and this split requires changes to patching functionality in multiple places each time a new patch attribute is introduced. Additionally, this makes the PatchParsers [difficult to test because applying actual result of their processing is mocked out](https://github.com/dbt-labs/dbt-core/pull/7415#discussion_r1172800797).

Let's move the Node.patch methods into the respective PatchParsers instead to consolidate overall patching logic. We should also consider creating a separate ParsedModelPatch for model-specific configs that are patched by ModelPatchParser if possible (could be tricky with the inheritance currently in place).

gshank:
resolves #7430

### Description

Moves the node patch method out of the node classes and into the schema parser.

* The SchemaParser was serving as two separate parsers, parsing schema files and parsing generic tests from yaml files. I've split out the parsing of generic tests into SchemaGenericTestParser in schema_generic_tests.py. The base generic tests which turn into macros are still parsed by the GenericTestParser. The SchemaGenericTestParsers are created in the SchemaParser __init__ and the SourcePatcher.
* I moved the "patch" method from the ParsedNode class to the "patch_node_properties" method in the PatchParser and ModelPatchParser.
* Moving the patch method was easier if the "constraints" dictionary was added to the UnparsedNodeUpdate class instead of being retrieved from the block target.
* Because mypy started complaining I had to change a number of places that checked the resource_type of a node to using "isinstance".
* I moved some common classes and types into core/dbt/parser/common.py and out of "generic_test_builders.py" because the includes were getting weird and some of them didn't logically belong there anyway.
* I moved the exposure, metric and group schema parsers out of schemas.py and into "schema_yaml_readers.py" to simplify and shorten schemas.py
* I renamed NonSourceParser to PatchParser because it's a better name
* I removed the YamlDocsReader class because it was pointless
* I moved the patch_constraints code to the ModelPatchParser where it belonged
* I removed some tests which no longer work because we don't have a "patch" method to call.

### Checklist

### DIFF
--- a/core/dbt/parser/common.py
+++ b/core/dbt/parser/common.py
@@ -1,0 +1,222 @@
+from dbt.contracts.graph.unparsed import (
+    HasColumnProps,
+    UnparsedColumn,
+    UnparsedNodeUpdate,
+    UnparsedMacroUpdate,
+    UnparsedAnalysisUpdate,
+    UnparsedExposure,
+    UnparsedModelUpdate,
+)
+from dbt.contracts.graph.unparsed import NodeVersion, HasColumnTests, HasColumnDocs
+from dbt.contracts.graph.nodes import (
+    UnpatchedSourceDefinition,
+    ColumnInfo,
+    ColumnLevelConstraint,
+    ConstraintType,
+)
+from dbt.parser.search import FileBlock
+from typing import List, Dict, Any, TypeVar, Generic, Union, Optional
+from dataclasses import dataclass
+from dbt.exceptions import DbtInternalError, ParsingError
+
+
+def _trimmed(inp: str) -> str:
+    if len(inp) < 50:
+        return inp
+    return inp[:44] + "..." + inp[-3:]
+
+
+TestDef = Union[str, Dict[str, Any]]
+
+
+Target = TypeVar(
+    "Target",
+    UnparsedNodeUpdate,
+    UnparsedMacroUpdate,
+    UnparsedAnalysisUpdate,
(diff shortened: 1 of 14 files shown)


core/dbt/parser/common.py
```diff
@@ -0,0 +1,222 @@
...
+    ColumnLevelConstraint,
+    ConstraintType,
+)
+from dbt.parser.search import FileBlock
+from typing import List, Dict, Any, TypeVar, Generic, Union, Optional
+from dataclasses import dataclass
+from dbt.exceptions import DbtInternalError, ParsingError
+
+
+def _trimmed(inp: str) -> str:
```
MichelleArk, line 23: nit: maybe we drop the `_` prefix given this method is imported outside the file

MichelleArk (approved):

PR #7641, opened 2023-05-16
Missed PR feedback

iknox-fa:
resolves #
No Ticket.

This fixes some PR feedback that got missed on #7628


iknox-fa:
Not what was intended :/

PR #7645, opened 2023-05-16
Add "other" relation to reffable node classes

### LINKED ISSUE #7550: [CT-2544] Add "other" relation attribute to reffable node classes
## Description
As part of the work for clone #7256, reffable nodes need to hold a reference to their "other" counterparts in other schemas.

Jeremy already has [a draft PR](https://github.com/dbt-labs/dbt-core/pull/7258) up that accomplishes this and uses the name `"state_relation"` as seen in the links below, however we are not beholden to this name.

As always, the language team has the final say in all things artifacts so make sure to get their sign off on whatever approach is taken.

## Dependencies
Adapter work from #7549 will be dependent on the work in this ticket.

## Acceptance Criteria
- The reffable node classes (i.e. `ModelNode`, `SeedNode`, and `SnapshotNode`) have a field that stores a relation to an "other" counterpart in another schema
- This new attribute is not written to `manifest.json`
- This new attribute is somehow tested

## Links from the draft PR
Separating `RelationalNode` from `ParsedNodeMandatory`:
https://github.com/dbt-labs/dbt-core/blob/5f08d70336af77908890784fa6ac1399d50ca2d5/core/dbt/contracts/graph/nodes.py#L231-L243

Adding a relation to a reffable node:
https://github.com/dbt-labs/dbt-core/blob/5f08d70336af77908890784fa6ac1399d50ca2d5/core/dbt/contracts/graph/nodes.py#L570-L574

Serializing the new relation attribute out of manifest writing:
https://github.com/dbt-labs/dbt-core/blob/5f08d70336af77908890784fa6ac1399d50ca2d5/core/dbt/contracts/graph/manifest.py#L1256-L1262

Adding the new relation attribute to required node keys for manifest tests:
https://github.com/dbt-labs/dbt-core/blob/5f08d70336af77908890784fa6ac1399d50ca2d5/test/unit/test_manifest.py#L91-L92

stu-k:
resolves #7550

### Description

As part of `dbt clone` #7256, we need to add an additional field to reffable nodes (i.e. `ModelNode`, `SeedNode`, `SnapshotNode`). This field holds a relation to another node that will be used later in the `dbt clone` project to be exposed in a jinja context to enable the `clone` materialization. The name `state_relation` is very general and we do not need to use this name.

This code is taken from [an existing draft PR](https://github.com/dbt-labs/dbt-core/pull/7258) that Jeremy has up.

### Checklist

### DIFF
--- a/core/dbt/contracts/graph/nodes.py
+++ b/core/dbt/contracts/graph/nodes.py
@@ -287,6 +287,10 @@ class ParsedNodeMandatory(GraphNode, HasRelationMetadata, Replaceable):
     checksum: FileHash
     config: NodeConfig = field(default_factory=NodeConfig)
 
+    @property
+    def identifier(self):
+        return self.alias
+
 
 # This needs to be in all ManifestNodes and also in SourceDefinition,
 # because of "source freshness"
--- a/core/dbt/contracts/graph/nodes.py
+++ b/core/dbt/contracts/graph/nodes.py
@@ -271,14 +271,18 @@ class DependsOn(MacroDependsOn):
 
 
 @dataclass
-class ParsedNodeMandatory(GraphNode, HasRelationMetadata, Replaceable):
+class RelationalNode(HasRelationMetadata):
     alias: str
-    checksum: FileHash
-    config: NodeConfig = field(default_factory=NodeConfig)
 
     @property
     def identifier(self):
         return self.alias
+
+
+@dataclass
+class ParsedNodeMandatory(GraphNode, RelationalNode, Replaceable):
+    checksum: FileHash
+    config: NodeConfig = field(default_factory=NodeConfig)
 
 
 # This needs to be in all ManifestNodes and also in SourceDefinition,
@@ -615,6 +619,7 @@ class ModelNode(CompiledNode):
     constraints: List[ModelLevelConstraint] = field(default_factory=list)
     version: Optional[NodeVersion] = None
     latest_version: Optional[NodeVersion] = None
+    state_relation: Optional[RelationalNode] = None
 
     @property
     def is_latest_version(self) -> bool:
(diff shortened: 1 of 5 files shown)


core/dbt/contracts/graph/nodes.py
```diff
@@ -615,6 +619,7 @@ class ModelNode(CompiledNode):
     constraints: List[ModelLevelConstraint] = field(default_factory=list)
     version: Optional[NodeVersion] = None
     latest_version: Optional[NodeVersion] = None
+    state_relation: Optional[RelationalNode] = None
```
stu-k, line 622: As mentioned in the PR description, `state_relation` is a very general name that doesn't explain much of what this field holds. Please feel free to suggest better names for this attribute.

stu-k: Maybe `other_relation`, `target_relation`, `cross_schema_relation`

core/dbt/contracts/graph/nodes.py
```diff
@@ -271,16 +271,20 @@ def add_public_node(self, value: str):
 
 
 @dataclass
-class ParsedNodeMandatory(GraphNode, HasRelationMetadata, Replaceable):
+class RelationalNode(HasRelationMetadata):
```
gshank, line 274: I think it's confusing to use Node in the name here. It's not a Node. These classes are already complicated enough, I don't think it's necessarily a good idea to reuse the existing classes in nodes.py.  You're actually reimplementing a Relation class.  Do you actually need the separate database/schema/identifier? ("identifier" is what it's actually called in the Relation classes, and makes more sent to me, at least.).  I'm not exactly sure how this is going to be used. Would a relation_name (clone_relation_name?) field work?

gshank (approved):

core/dbt/contracts/graph/nodes.py
stu-k: Spoke with Gerda about this name and she thought it made sense for the context it was in.

PR #7419, opened 2023-04-20
Add --target-path to snapshot command

dwreeves:
resolves #7418

### Description

`dbt snapshot` uses the target path (since it runs a subclass of `RunTask`), but does not contain the option to specify it. This PR adds `--target-path` as a valid option for `dbt snaphot`.

### Checklist

- [x] I have read [the contributing guide](https://github.com/dbt-labs/dbt-core/blob/main/CONTRIBUTING.md) and understand what's expected of me
- [x] I have signed the [CLA](https://docs.getdbt.com/docs/contributor-license-agreements)
- [x] I have run this code in development and it appears to resolve the stated issue
- [x] This PR includes tests, or tests are not required/relevant for this PR
- [x] I have [opened an issue to add/update docs](https://github.com/dbt-labs/docs.getdbt.com/issues/new/choose), or docs changes are not required/relevant for this PR
- [x] I have run `changie new` to [create a changelog entry](https://github.com/dbt-labs/dbt-core/blob/main/CONTRIBUTING.md#adding-a-changelog-entry)

Issue #7418, opened 2023-04-20
[CT-2443] [Regression] `dbt snapshot` command uses target path, but does not have a CLI option to specify it.


dwreeves:
### Is this a new bug in dbt-core?

- [X] I believe this is a new bug in dbt-core
- [X] I have searched the existing issues, and I could not find an existing issue for this bug

### Current Behavior

the `dbt snapshot` command does not have a `--target-path` option.

### Expected Behavior

Because the `dbt snapshot` command uses the target path as part of its runtime, I expected that you would also be able to specify it. This is also a regression from our end (we are upgrading from dbt 1.1 to 1.5), as we were using a custom target path to run everything out of.

### Steps To Reproduce

Attempt the command `dbt snapshot --target-path my/custom/path`.

### Environment

```markdown
- OS: OSX
- Python: 3.10
- dbt: 1.5 beta
```

### Which database adapter are you using with dbt?

snowflake

dbeatty10:
Thanks for trying out 1.5 and reporting this @dwreeves 🏅 

I suspect that the behavior you are seeing is because `@p.target_path` is missing from here:
https://github.com/dbt-labs/dbt-core/blob/1722079a43d1e3a94ffb73e66c8712ee0837e9fa/core/dbt/cli/main.py#L638-L655

I'm going to re-label this as a regression.

dbeatty10:
Lol, I just noticed you already realized the fix and submitted https://github.com/dbt-labs/dbt-core/pull/7419 accordingly. Thanks for doing that @dwreeves!

PR #7647, opened 2023-05-16
add target-path to more cli commands that use it.

### LINKED ISSUE #7646: [Regression] --target-path is needed in more CLI commands.

### Current Behavior

`dbt [subcommand] --target-path` is needed in more subcommands.

If I run `dbt [subcommand]` and a `target/` is created in the PWD (when one previously did not exist), then I expect that I should be able to parameterize this in the CLI.

### Steps To Reproduce

Test all the subcommands with `dbt [subcommand]` and see which ones create a `target/` dir if one does not exist in the PWD.

### Relevant log output

_No response_

### Environment

```markdown
- OS: OSX
- Python: 3.10
- dbt: 1.5.0
```

snowflake

### Additional Context

_No response_

dwreeves:
resolves #7646

### Description

Related to my previous PR #7419 and issue #7418.

There are additional subcommands that do not have `--target-path` as an option, even though they create a `target/` folder when one does not exist. Because we (at our company with our dbt setup) do not have our target set up in `${PWD}/target`, and we specify the `--target-path` via the CLI, this ends up being an annoyance.

In particular, we were using dbt 1.1.x and ran into this as a regression for the command `dbt run-operation`. However, I went through the rest of the CLI and added `--target-path` for everything that met the following criteria:

- `${PWD}/target/` does not exist.
- I run `dbt [subcommand]`.
- After the command executes, `${PWD}/target/` does exist.

I also validated this against source code, e.g. `GraphRunnableTask` uses `self.config.target_path`, so all subclasses of this should have a `--target-path`.

### Checklist

### DIFF
--- a/core/dbt/cli/main.py
+++ b/core/dbt/cli/main.py
@@ -213,6 +213,7 @@ def build(ctx, **kwargs):
 @p.profiles_dir
 @p.project_dir
 @p.target
+@p.target_path
 @p.vars
 @requires.postflight
 @requires.preflight
@@ -284,6 +285,7 @@ def docs_generate(ctx, **kwargs):
 @p.profiles_dir
 @p.project_dir
 @p.target
+@p.target_path
 @p.vars
 @requires.postflight
 @requires.preflight
@@ -477,6 +479,7 @@ def init(ctx, **kwargs):
 @p.state
 @p.deprecated_state
 @p.target
+@p.target_path
 @p.vars
 @requires.postflight
 @requires.preflight
@@ -577,6 +580,7 @@ def run(ctx, **kwargs):
 @p.profiles_dir
 @p.project_dir
 @p.target
+@p.target_path
 @p.vars
 @requires.postflight
 @requires.preflight
@@ -691,6 +695,7 @@ def source(ctx, **kwargs):
 @p.state
 @p.deprecated_state
 @p.target
+@p.target_path
 @p.threads
 @p.vars
 @requires.postflight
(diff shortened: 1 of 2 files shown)


core/dbt/cli/main.py
```diff
@@ -213,6 +213,7 @@ def build(ctx, **kwargs):
 @p.profiles_dir
 @p.project_dir
 @p.target
+@p.target_path
```
stu-k, line 216: I don't believe the clean task uses `target_path`. It writes neither a results artifact nor a manifest.

dwreeves: I was basing this off of the following in `project.py`:

```python
        flag_target_path = str(flags.TARGET_PATH) if flags.TARGET_PATH else None
        target_path: str = flag_or(flag_target_path, cfg.target_path, "target")
        log_path: str = str(flags.LOG_PATH)

        clean_targets: List[str] = value_or(cfg.clean_targets, [target_path])
```

Plus the fact that `CleanTask` refers to `self.project.clean_targets` and `self.config.clean_targets`.

Additionally, the command outputs `Checking target/*` and `Cleaned target/*` when I run it.

Let me know if I'm misunderstanding something though, but I would have guessed based on all this that it is using the target path.

stu-k (approved):

dbeatty10:
@jtcohen6 would you approve backporting this one?

jtcohen6:
@dbeatty10 Good by me! Sorry I missed this comment last month

PR #5935, opened 2022-09-26
Tests for `current_timestamp`

dbeatty10:
resolves #5933

### Description

For `{{ dbt.current_timestamp() }}`, verify two things we want to test for this macro across adapters:

1. Does it return a value "close enough" to the Python current time in UTC?
1. Does the semantic semantic Python data type ([timezone-naive vs. timezone-aware](https://docs.python.org/3/library/datetime.html#aware-and-naive-objects)) of the result match our expectations?

Some of all of the logic can also be re-used for other macros that return timestamps (like `current_timestamp_in_utc`), `convert_timezone`, etc.

### Checklist

- [x] I have read [the contributing guide](https://github.com/dbt-labs/dbt-core/blob/main/CONTRIBUTING.md) and understand what's expected of me
- [x] I have signed the [CLA](https://docs.getdbt.com/docs/contributor-license-agreements)
- [x] I have run this code in development and it appears to resolve the stated issue
- [ ] This PR includes tests, or tests are not required/relevant for this PR
- [x] ~I have [opened an issue to add/update docs](https://github.com/dbt-labs/docs.getdbt.com/issues/new/choose), or~ docs changes are not required/relevant for this PR
- [x] ~I have run `changie new` to [create a changelog entry](https://github.com/dbt-labs/dbt-core/blob/main/CONTRIBUTING.md#adding-a-changelog-entry)~

PR #7665, opened 2023-05-19
Unified to UTC

### LINKED ISSUE #7664: [Bug] Integration test gives an error for the JST timezone

### Current Behavior

Error at `dbt-core/tests/functional/simple_snapshot/test_hard_delete_snapshot.py:97` when running `make integration`.

### Expected Behavior

Integration tests complete successfully.

### Steps To Reproduce

I forked the repository and built the necessary environment according to “[CONTRIBUTING.md](https://github.com/dbt-labs/dbt-core/blob/main/CONTRIBUTING.md)”.
Below is the command I executed.
1. Build venv environment
- python3 -m venv env
- source env/bin/activate
- pip install --upgrade pip
2. Build dbt environment
- make dev
3. Build DB environment
- make setup-db
4. Run unit tests
- make test
5. Run integration tests
- make integration

In addition, in “~/.dbt/profiles.yml”,
```
default:
   target: dev
   outputs:
     devs:
       type: postgres
       host: localhost
       user: root
       password: password
       port: 5432
       dbname: postgres
       schema: public
       threads: 4
```
### Relevant log output

```shell
>           assert result[-1].replace(tzinfo=currenttz()) >= invalidated_snapshot_datetime
E           AssertionError: assert datetime.datetime(2023, 5, 15, 7, 36, 0, 679815, tzinfo=datetime.timezone(datetime.timedelta(seconds=32400), 'JST')) >= datetime.datetime(2023, 5, 15, 7, 36, 0, 58485, tzinfo=<UTC>)
E            +  where datetime.datetime(2023, 5, 15, 7, 36, 0, 679815, tzinfo=datetime.timezone(datetime.timedelta(seconds=32400), 'JST')) = <built-in method replace of datetime.datetime object at 0x11a20f2a0>(tzinfo=datetime.timezone(datetime.timedelta(seconds=32400), 'JST'))
E            +    where <built-in method replace of datetime.datetime object at 0x11a20f2a0> = datetime.datetime(2023, 5, 15, 7, 36, 0, 679815).replace
E            +    and   datetime.timezone(datetime.timedelta(seconds=32400), 'JST') = currenttz()
```

### Environment

```markdown
- OS:macOS Catalina 10.15.7
- Python:3.10.5
- dbt:1.6.0-b1
```

postgres

### Additional Context

_No response_

d-kaneshiro:
resolves #7664 

### Description

Unified to UTC to avoid integration test errors.

### Checklist


dbeatty10:
### Code history
Digging into the history, it used to be [standardized to UTC](https://github.com/dbt-labs/dbt-core/pull/5760/files#r962579430), but apparently there was some kind of intermittent error. I don't know for certain what those error messages were or their ultimate root cause. 🤷 

💡 I suspect one (or both) of these things may have caused those intermittent issues:
1. Slight differences between the system clock of the Python machine and the system clock of the database.
2. Differences between the ["awareness"](https://docs.python.org/3/library/datetime.html#aware-and-naive-objects) (naive vs. aware) of the Python `datetime`s returned by different dbt adapters.

https://github.com/dbt-labs/dbt-core/pull/5935 has examples of how to resolve both of those:
1. Give some [wiggle room](https://github.com/dbt-labs/dbt-core/blob/87ea28fe84edd65d589fd3cf658d8dd1f65c000a/tests/adapter/dbt/tests/adapter/utils/test_current_timestamp.py#L42-L46)
2. Create a Python datetime that [matches the awareness](https://github.com/dbt-labs/dbt-core/blob/87ea28fe84edd65d589fd3cf658d8dd1f65c000a/tests/adapter/dbt/tests/adapter/utils/test_current_timestamp.py#L48-L52) of the database timestamp

### Logic of the integration test

If I'm reading it correctly, this [integration test](https://github.com/dbt-labs/dbt-core/blob/87ea28fe84edd65d589fd3cf658d8dd1f65c000a/tests/functional/simple_snapshot/test_hard_delete_snapshot.py) does the following:

1. Run the first snapshot
1. Check the expected result
1. Remove 10 source records
1. Take another snapshot (and save the current time)
1. Check the number of expected rows
1. Check the `dbt_valid_to` values are after the time from 4)
1. Revive 2 source records
1. Take another snapshot (and save the current time)
1. Check the number of expected rows
1. Check the `dbt_valid_to` values are after the time from 4)
1. Look at two specific records
1. Check the number of expected rows
1. Check the `dbt_valid_from` values are earlier than the time from 8)

### Moving forward

Steps 6, 10, and 13 are checking for **external** consistency between the database timestamps and the current timestamps of the Python process. It's possible that the system clocks of the Python process and the database aren't perfectly synchronized, so testing for external consistency is complicated.

To simplify, I'd suggest changing these to check for **internal** consistency instead. i.e., just verify that the database timestamps after each snapshot are in the expected order.

If checks for external consistency are desired, then I'd suggest checking that the database timestamps are within some delta of the Python timestamps instead of a strict ordering. i.e., just verify that they are "close enough" to each other rather than asserting that one is less than the other. The key is to give them enough wiggle room: a delta of 1 minute should be plenty enough whereas 1 millisecond would be too small.

See [here](https://github.com/dbt-labs/dbt-core/blob/87ea28fe84edd65d589fd3cf658d8dd1f65c000a/tests/adapter/dbt/tests/adapter/utils/test_current_timestamp.py#L43-L46) for a test with some wiggle room.

d-kaneshiro:
Thanks for commenting on the issue in detail.
I have decided to accept @dbeatty10 's suggestion regarding external consistency for the following reasons:

1. In the case of internal consistency, (4) 'dbt_valid_to' timestamps can only be obtained after 'invalidated_snapshot_datetime', so the order of timestamps before that cannot be compared (it will be NULL before that)
2. External consistency with a certain margin is sufficiently effective at the moment (error was within 1 second in my environment)

tests/functional/simple_snapshot/test_hard_delete_snapshot.py
```diff
@@ -94,7 +82,13 @@ def test_snapshot_hard_delete(project):
...
         # result is a tuple, the dbt_valid_to column is the latest
         assert isinstance(result[-1], datetime)
-        assert result[-1].replace(tzinfo=currenttz()) >= invalidated_snapshot_datetime
+        # Plenty of wiggle room if clocks aren't perfectly sync'd, etc
+        tolerance = timedelta(minutes=1)
+        assert (
+            result[-1].replace(tzinfo=pytz.UTC) > (invalidated_snapshot_datetime - tolerance)
+        ) and (
+            result[-1].replace(tzinfo=pytz.UTC) < (invalidated_snapshot_datetime + tolerance)
+        ), f"SQL timestamp {result[-1].replace(tzinfo=pytz.UTC).isoformat()} is not close enough to Python UTC {result[-1].replace(tzinfo=pytz.UTC).isoformat()}"
```
dbeatty10, line 91:
```suggestion
        ), f"SQL timestamp {result[-1].replace(tzinfo=pytz.UTC).isoformat()} is not close enough to Python UTC {invalidated_snapshot_datetime.isoformat()}"
```

tests/functional/simple_snapshot/test_hard_delete_snapshot.py
```diff
@@ -133,7 +127,13 @@ def test_snapshot_hard_delete(project):
...
         # result is a tuple, the dbt_valid_to column is the latest
         assert isinstance(result[1], datetime)
-        assert result[1].replace(tzinfo=currenttz()) >= invalidated_snapshot_datetime
+        # Plenty of wiggle room if clocks aren't perfectly sync'd, etc
+        tolerance = timedelta(minutes=1)
+        assert (
+            result[1].replace(tzinfo=pytz.UTC) > (invalidated_snapshot_datetime - tolerance)
+        ) and (
+            result[1].replace(tzinfo=pytz.UTC) < (invalidated_snapshot_datetime + tolerance)
+        ), f"SQL timestamp {result[1].replace(tzinfo=pytz.UTC).isoformat()} is not close enough to Python UTC {result[1].replace(tzinfo=pytz.UTC).isoformat()}"
```
dbeatty10, line 136:
```suggestion
        ), f"SQL timestamp {result[1].replace(tzinfo=pytz.UTC).isoformat()} is not close enough to Python UTC {invalidated_snapshot_datetime.isoformat()}"
```

dbeatty10 (changes requested):
Heading in the right direction @d-kaneshiro 

### Two high-level things
1. Could you run [`changie new`](https://github.com/dbt-labs/dbt-core/blob/main/CONTRIBUTING.md#adding-a-changelog-entry) and commit the result?
1. I gave two suggestions to update the message to print if the assertion fails

### One low-level thing
- I'm not certain if this is always guaranteed to work:
    ```
    result[-1].replace(tzinfo=pytz.UTC)
    ```

i.e., if `result[-1]` is an aware Python `datetime` with `tzinfo` other than UTC, then it will just overwrite it, which would cause this test to fail erroneously. Examples of dbt adapters that _might_ be effected by that would be dbt-spark and dbt-databricks.

Instead, I'd suggest [checking the awareness](https://github.com/dbt-labs/dbt-core/blob/87ea28fe84edd65d589fd3cf658d8dd1f65c000a/tests/adapter/dbt/tests/adapter/utils/test_current_timestamp.py#L16-L21) of the `datetime` given by `result[-1]` and then only do `replace(tzinfo=pytz.UTC)` if it is naive.

Of course that all assumes that any naive timestamp returned by the database is meant to represent UTC. In the case of dbt snapshots, that is the expectation and a safe assumption.

d-kaneshiro:
Thank you @dbeatty10. your comment is very helpful.
It was pointed out and fixed. Please leave a review.

tests/functional/simple_snapshot/test_hard_delete_snapshot.py
d-kaneshiro: Resolved in [c736197](https://github.com/dbt-labs/dbt-core/pull/7665/commits/c7361976f9b0dcab63059000759a89728d64d9ab)

tests/functional/simple_snapshot/test_hard_delete_snapshot.py
d-kaneshiro: Resolved in [c736197](https://github.com/dbt-labs/dbt-core/pull/7665/commits/c7361976f9b0dcab63059000759a89728d64d9ab)

PR #7672, opened 2023-05-21
Include null checks in utils test base

### LINKED ISSUE #7670: [Bug] dbt adapter tests for utils unreliable

### Current Behavior

The current adapter tests for the [utils](https://github.com/dbt-labs/dbt-core/tree/main/tests/adapter/dbt/tests/adapter/utils) use an incorrect way to verify the end results of the tests.

```sql
{% test assert_equal(model, actual, expected) %}
select * from {{ model }} where {{ actual }} != {{ expected }}

{% endtest %}
```

The above macro is currently used. However, `NULL` works [in a strange way in SQL](https://www.sqlite.org/nulls.html).

This is documented [for SQL Server](https://stackoverflow.com/questions/50917152/sql-server-compare-to-null) or [Postgres](https://www.postgresql.org/docs/9.1/functions-comparison.html) in the same way as it is the ANSI standard that dictates this behaviour.

The result is that the test above will completely ignore results where either `actual` or `expected` is `null` and the other one is not.

I just noticed this in the dbt-sqlserver tests. Apparently the hashing test should have been failing from as soon it was implemented, but it never did.

### Expected Behavior

There are multiple routes we can take here. I'd be glad to submit a PR once we agree on the way forward.

1. `distinct from`
[This SQL syntax](https://wiki.postgresql.org/wiki/Is_distinct_from) is not supported in all common databases (e.g. SQL Server just added support in the 2022 version), but it effectively compares the values as was probably intended in the mentioned macro.

2. `coalesce`
If we would coalesce the result values to an invalid value (e.g. `coalesce(actual, 'invalid')` for strings), then the tests could continue using the `!=` operator to compare `actual` and `expected`. I think this would mean that we have to adapt every utils test as some work on strings, others on integers etc.

3. add null checks to where clause
IMHO the best way forward. Just add conditions to see if `actual` is `null` while `expected` is not and vice versa.

```sql
{% test assert_equal(model, actual, expected) %}
select * from {{ model }}
where {{ actual }} != {{ expected }}
or ({{ actual }} is null and {{ expected }} is not null)
or ({{ expected }} is null and {{ actual }} is not null)
{% endtest %}
```

### Steps To Reproduce

This issue is visible in the latest version of the dbt-sqlserver tests, but it will probably have caused more false positives for other adapters as well.

sdebruyn:
resolves #7670 

### Description

As discussed in #7670, add null into the conditions when running the utils tests. Thanks @dbeatty10 for the excellent code suggestion!

### Checklist

### DIFF
--- a/tests/adapter/dbt/tests/adapter/utils/base_utils.py
+++ b/tests/adapter/dbt/tests/adapter/utils/base_utils.py
@@ -1,10 +1,17 @@
 import pytest
 from dbt.tests.util import run_dbt
 
+macros__equals_sql = """
+{% macro equals(actual, expected) %}
+{# -- actual is not distinct from expected #}
+(({{ actual }} = {{ expected }}) or ({{ actual }} is null and {{ expected }} is null))
+{% endmacro %}
+"""
+
 macros__test_assert_equal_sql = """
 {% test assert_equal(model, actual, expected) %}
-select * from {{ model }} where {{ actual }} != {{ expected }}
-
+select * from {{ model }}
+where not {{ equals(actual, expected) }}
 {% endtest %}
 """
 
@@ -13,7 +20,10 @@ class BaseUtils:
     # setup
     @pytest.fixture(scope="class")
     def macros(self):
-        return {"test_assert_equal.sql": macros__test_assert_equal_sql}
+        return {
+            "equals.sql": macros__equals_sql,
+            "test_assert_equal.sql": macros__test_assert_equal_sql,
+        }
 
     # make it possible to dynamically update the macro call with a namespace
     # (e.g.) 'dateadd', 'dbt.dateadd', 'dbt_utils.dateadd'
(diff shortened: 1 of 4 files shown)


VersusFacit:
Thanks so much for the contribution for a number of reasons @sdebruyn . Among other things, this started a thread internally wherein we realized we hadn't given this very important macro a test of its own! A small but significant oversight we're correcting along with you macro chang.

Logistics:
1. I pushed these tests onto your branch here and am going to merge the whole kit and caboodle. 
2. This will be 🚢 -ed out in this week's release.

P.S. thanks @dbeatty10 for helping this one along and being a part of the testing conversation/effort. I felt so "wrong" leaving this macro untested.

tests/adapter/dbt/tests/adapter/utils/base_utils.py
```diff
@@ -1,10 +1,17 @@
 import pytest
 from dbt.tests.util import run_dbt
 
+macros__equals_sql = """
+{% macro equals(actual, expected) %}
+{# -- actual is not distinct from expected #}
+(({{ actual }} = {{ expected }}) or ({{ actual }} is null and {{ expected }} is null))
```
VersusFacit, line 7: I really despise how the "absence of value is not equal to itself" definition is formally correct but does not reflect how anyone uses a Null in practice. Maybe we need a ANSI sequel patch while we're at it 😁

VersusFacit (approved):
I added tests for all Nulls and some Null, some not.

PR #7706, opened 2023-05-25
Target path should be relative to project dir, rather than current working directory

### LINKED ISSUE #7465: [CT-2491] [Bug] Default `--target-path` should be relative to `--project-dir`, rather than current working directory

### Current Behavior

Executing a command that targets a project in a specified `--project-dir` results in the json artifacts being written to `current-working-directory/target/` rather than `project-dir/target/`.

### Expected Behavior

In previous versions, specifying a `target-path` in your dbt_project.yml was relative to the project directory, rather than relative to the current working directory from which the command was executed.

### Steps To Reproduce

In a new virtual environment (just using duckdb so I don't have to get docker involved):

```
pip install dbt-core==1.5.0rc2 dbt-duckdb==1.5.0rc1

gh repo clone dbt-labs/jaffle_shop

cd jaffle_shop

echo "jaffle_shop:
  outputs:
   dev:
     type: duckdb
     path: /tmp/dbt.duckdb
  target: dev" > profiles.yml
  
 cd some/other/path
 
 dbt build --profiles-dir ~/jaffle_shop --project-dir ~/jaffle_shop
 
 ls some/other/path 
```

At this point, you will observe that a new `target` directory has been created in `some/other/path`, and no target directory exists in `~/jaffle_shop`.

### Relevant log output

_No response_

### Environment

```markdown
- OS: Ventura 13.12
- Python: 3.9.10
- dbt: 1.5.0rc2
```

### Additional Context

For this example, I'm using the duckdb adapter just for ease of reproducibility, but I don't think that has an impact on this specific behavior.

gshank:
resolves #7465

### Description

With version 1.5 we no longer did a chdir to the project directory in the same places as in previous releases.

There are two ways this could be fixed: 1) Always do a chdir to the project directory for every command, 2) construct paths based on the project root directory.

In this pull request I've chosen to construct paths based on the project root directory, because going back to doing a chdir feels like going in the wrong direction. Both options probably have a similar number of code changes, but adding a contextmanager (or something similar) to do a chdir feels more fragile and more likely to have unexpected side effects. The only real reason to chdir to the project directory is to avoid having to construct the right kind of paths, and it feels more robust to just construct the paths.

A new property has been added to the Project/RuntimeConfig classes, "project_target_path", and I've switched to using that in most of the places where we used config.target_path. The "compiled_path" and "build_path" required a bit more refactoring, since they used a method which combined building the path for writing and returning the path to be stored in the node. That method has been split into two in order to not save an absolute path in the node. Changes were also necessary in the PreviousState class, to use the project_root.

For the defer_state "state" directory, I tweaked the PreviousState object and had to set "resolve_path" to False on the click.Path object for that directory, because it was setting it to a subdirectory of cwd before even constructing the args.

Tests have been modified in a couple of places to do a chdir and verify that things still end up in the right places. The code should handle the difference between absolute and relative paths correctly.

### Checklist

### DIFF
--- a/core/dbt/contracts/state.py
+++ b/core/dbt/contracts/state.py
@@ -43,7 +43,7 @@ class PreviousState:
                 exc.add_filename(str(sources_path))
                 raise
 
-        sources_current_path = self.target_path / "sources.json"
+        sources_current_path = self.project_root / self.target_path / "sources.json"
         if sources_current_path.exists() and sources_current_path.is_file():
             try:
                 self.sources_current = FreshnessExecutionResultArtifact.read_and_check_versions(
--- a/core/dbt/contracts/state.py
+++ b/core/dbt/contracts/state.py
@@ -7,15 +7,17 @@ from dbt.exceptions import IncompatibleSchemaError
 
 
 class PreviousState:
-    def __init__(self, path: Path, current_path: Path):
-        self.path: Path = path
-        self.current_path: Path = current_path
+    def __init__(self, state_path: Path, target_path: Path, project_root: Path):
+        self.state_path: Path = state_path
+        self.target_path: Path = target_path
+        self.project_root: Path = project_root
         self.manifest: Optional[WritableManifest] = None
         self.results: Optional[RunResultsArtifact] = None
         self.sources: Optional[FreshnessExecutionResultArtifact] = None
         self.sources_current: Optional[FreshnessExecutionResultArtifact] = None
 
-        manifest_path = self.path / "manifest.json"
+        # Note: if state_path is absolute, project_root will be ignored.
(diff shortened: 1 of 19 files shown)


core/dbt/contracts/state.py
```diff
@@ -41,7 +43,7 @@ def __init__(self, path: Path, current_path: Path):
                 exc.add_filename(str(sources_path))
                 raise
 
-        sources_current_path = self.current_path / "sources.json"
+        sources_current_path = self.target_path / "sources.json"
```
MichelleArk, line 46: for my own understanding - why isn't self.project_root considered for sources_current_path?

gshank: Good catch. It was just an oversight on my part.

PR #7710, opened 2023-05-26
fix StopIteration error when publication not found

MichelleArk:
resolves https://github.com/dbt-labs/dbt-core/issues/7711

🎩 -ed with reproduction case from issue + added new test case.

### Description
* represents `ManifestLoader.publications` as a dictionary mapping project name strings -> publications, instead of a flat list for more straightforward lookups downstream. 
* This feels like an internal concern to ManifestLoader. User-facing interface `dbtRunner.invoke` still makes sense as a flat list of `PublicationArtifacts` for simplicity. 

### Checklist

### DIFF
--- a/core/dbt/parser/manifest.py
+++ b/core/dbt/parser/manifest.py
@@ -243,7 +243,11 @@ class ManifestLoader:
         self.root_project: RuntimeConfig = root_project
         self.all_projects: Mapping[str, Project] = all_projects
         self.file_diff = file_diff
-        self.publications = publications
+        self.publications: Mapping[str, PublicationArtifact] = (
+            {publication.project_name: publication for publication in publications}
+            if publications
+            else {}
+        )
         self.manifest: Manifest = Manifest()
         self.new_manifest = self.manifest
         self.manifest.metadata = root_project.get_metadata()
@@ -841,11 +845,7 @@ class ManifestLoader:
 
     def load_new_public_nodes(self):
         for project in self.manifest.project_dependencies.projects:
-            publication = (
-                next(p for p in self.publications if p.project_name == project.name)
-                if self.publications
-                else None
-            )
+            publication = self.publications.get(project.name)
             if publication:
                 publication_config = PublicationConfig.from_publication(publication)
                 self.manifest.publications[project.name] = publication_config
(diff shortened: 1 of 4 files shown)


gshank (approved):

core/dbt/parser/manifest.py
```diff
@@ -841,11 +845,7 @@ def build_public_nodes(self) -> bool:
 
     def load_new_public_nodes(self):
         for project in self.manifest.project_dependencies.projects:
-            publication = (
-                next(p for p in self.publications if p.project_name == project.name)
-                if self.publications
-                else None
-            )
+            publication = self.publications.get(project.name)
```
mikealfare, line 848: What about putting the `raise` closer to the cause of the exception? For example:

```python
try:
    publication=self.publications[project.name]
except KeyError:
    raise PublicationConfigNotFound(project=project.name)

publication_config = PublicationConfig.from_publication(publication)
self.manifest.publications[project.name] = publication_config

for public_node in publication.public_models.values():
    self.manifest.public_nodes[public_node.unique_id] = public_node
```

MichelleArk: I like it! will update

core/dbt/parser/manifest.py
```diff
@@ -243,7 +243,11 @@ def __init__(
         self.root_project: RuntimeConfig = root_project
         self.all_projects: Mapping[str, Project] = all_projects
         self.file_diff = file_diff
-        self.publications = publications
+        self.publications: Mapping[str, PublicationArtifact] = (
+            {publication.project_name: publication for publication in publications}
+            if publications
```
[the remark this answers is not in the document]
MichelleArk: Setting a default parameter value as a mutable object can lead to surprising behaviour in python, and its [recommended in the docs](https://docs.python.org/3/reference/compound_stmts.html#function-definitions) to use `None` as a default instead: 

> Default parameter values are evaluated from left to right when the function definition is executed. This means that the expression is evaluated once, when the function is defined, and that the same “pre-computed” value is used for each call. This is especially important to understand when a default parameter value is a mutable object, such as a list or a dictionary: if the function modifies the object (e.g. by appending an item to a list), the default parameter value is in effect modified. This is generally not what was intended. A way around this is to use None as the default, and explicitly test for it in the body of the function, e.g.:

```python
def whats_on_the_telly(penguin=None):
    if penguin is None:
        penguin = []
    penguin.append("property of the zoo")
    return penguin
```

We could have callers of the ManifestLoader constructor pass in an empty list instead of `None` but I prefer to have the input sanitization in a single place to make things less error prone / sprawled out.

PR #7741, opened 2023-05-31
Adap 496/add test connection mode to debug

### LINKED ISSUE #7104: [CT-2254] `dbt debug` for connection testing
_Much older issue in the same vein: https://github.com/dbt-labs/dbt-core/issues/3735_

It should be possible to use `dbt debug` for connection testing. Currently, `debug` also includes checks for:
1. Valid `profiles.yml` file
2. Valid `dbt_project.yml` file
3. `git` as an installed dependency (for `deps`)
4. dbt can connect and submit a "debug" query to the selected profile

Of those checks, we really only need the last one for connection testing. Ideally, we'd be able to skip the other checks (perhaps by passing a flag?) and return success/error (programmatically & as CLI exit code) on the basis of the "debug" query only.

If we had to hack this today, it would be _sorta_ possible by examining the structured events (`DebugCmdOut`) returned from the `debug` command, but this hardly feels like a stable interface.

### Why `dbt debug`?

Why not just an interactive query (new `preview` command) with the query hard-coded to `select 1 as id`? Reasons to prefer:
- leveraging an adapter method, [`debug_query`](https://github.com/dbt-labs/dbt-core/blob/cbb9117ab997289062cd5bacaf4cfd4466ab9b1e/core/dbt/adapters/base/impl.py#L229-L230)
- it's part of dbt-core's top-level API, and clearly the intent of `debug`

### Questions
- How should the profile be provided? In `profiles.yml`, or could it be provided as a Python data structure?
- Could we avoid the need to have a `dbt_project.yml`? Just use a "dummy"/empty `dbt_project.yml`?

### Acceptance criteria

- Given the contents of a profile _only_, validate that profile via `dbt debug`, and return success/error

VersusFacit:
resolves #7104 (among other things like ADAP-6)

### Description

`dbt debug` has been reworked from the inside out. It's ready for further extension for future flags. It supports new methods. I'm also proposing removal of old code.

usage: `dbt debug` and `dbt debug --connection`

## Summary of changes 
see output screenshots [in this comment](https://github.com/dbt-labs/dbt-core/pull/7741#issuecomment-1575502178)
1. the flag that does connection testing only
2. adapter version is added
3. standardized `_connection_keys` and `debug_query()` in all 4 our of our adapters (see [bigquery](https://github.com/dbt-labs/dbt-bigquery/pull/754) [redshift](https://github.com/dbt-labs/dbt-redshift/pull/478) [snowflake](https://github.com/dbt-labs/dbt-snowflake/pull/649) [spark](https://github.com/dbt-labs/dbt-spark/pull/794))

## Solicitation for feedback at this stage

* Comments around the output, workflow, and code design!
* ~~Code review around basically everything -- we just have an open conversation around the future of `--config-dir`~~

## Next steps for me / what's left
- [X] double check debug_query in all repos overrides fine (so far, so good on Snow and PG)
- [X] ensure `_connection_keys` makes sense for all our 4 supported repos
- [X] update the tests in this repo

### Checklist


core/dbt/task/debug.py
```diff
@@ -126,6 +144,35 @@ def run(self):
...
+        """Using profile, return the adapter instance."""
+        register_adapter(profile)
+        return get_adapter(profile)
+
+    def _read_adapter_info(self, cls):
+        """Read the adapter name and version from the setup.py file of the adapter plugin."""
+        current_adapter_plugin = next(
+            filter(lambda x: x.adapter == cls, get_adapter_plugins(None))
+        )
+        adapter_setup_script = current_adapter_plugin.include_path / Path("../../../setup.py")
```
VersusFacit, line 157: Well, I don't love this for a number of reasons but this _is_ the only idea out of 5 I had to get this working. If you have a suggestion, let me know. I dug around the objects/classes available to the `DebugTask` and found nothing.

Fleid:
I'm not sure I'm a fan of `--test-connection-only`.

If `dbt debug` does 4 things. When there is no flag it does all of them, when you use flag it does only the one you explicitely ask for:

- `dbt debug`  does all
- `dbt debug --connection` tests the connection
- `dbt debug --config-dir` this one actually exists already
- `dbt debug --configuration` checks the yml files
- `dbt debug --dependencies` check the dependencies
- `dbt debug --info` prompt versions and file paths and such
- `dbt debug --info --dependencies` does 2 things

Right now we only implement `--connection`, but I like the potential picture that paints.

@dbeatty10, @dataders what's your take?

dbeatty10:
> @dbeatty10, @dataders what's your take?

Agreed with your take @Fleid ✅ 

I appreciate a couple things in your proposal:
1. That particular flag is easier to type since it is shorter (`--connection` instead of `--test-connection-only`)
2. Gives flexibility to choose which flags to flip on by having them each be independent of each other

The downside is that it could lead to gymnastics from an implementation perspective. Maybe all the flags would need to default to `None` instead of `True`. Then if all are `None`, then treat as-if all were `True`. If any are not `None`, then can leave the flags as-is.

Another related downside is that we're refining this on-the-fly here and I want to be thoughtful to @VersusFacit.

@VersusFacit would you prefer that we send this issue back to refinement or continue having this discussion within this PR?

VersusFacit:
Thanks for prompting us to take this back to refinement. 🙏 I'm going to talk to Florian about that ASAP since we need a sensible flag design. I like shades of this, but the code under the code makes this nontrivial, believe it or not. Moreover, this might actually be too much overhead for users. 

Things we can consider:
1. Making the underlying code execute in an order agnostic way (surprisingly tough lift of hard complexity)
* Can these flags be order agnostic on the command line and then be processed in order (this is so unclear right now, and I suspect 'no')
  * would we then have to sort the tasks internally?
* How can we make these flags order agnostic when certain commands are dependencies of each other? (spoiler: 
2. Adding something that explains the order of options (easier to implement, but leaves us with some spaghetti code)
3. Choosing which buckets to group similar items into so that we end up with fewer flags
dependency hell not a fun place to be)

Note this flag and such is based on recommendations from #7104, which is to say we'll have to be intentional here in changing the experience even further (i.e. beyond original scope).

Like I said, I'm going to talk this over internally, but this interface change involves a big(ger) departure from the status quo above and below the hood 🖖

VersusFacit:
Talked with Florian. For this PR, we are going to stick to `--connection` which amounts to a mere cosmetic change in design. However, there is a hunger when resources are available to make flags more flexible. TBD at a later date!

VersusFacit:
So `--config-dir` does something than we probably all expected it to -- also it's broken #7774 . I've got a fix for this branch here. 

However, it's job is to make you a command to show you how to open a window to your profiles dir file. This is 6 year old code and clearly from the early days of dbt (context given in the bug issue linked above).

What I think we want this command to do is to show us the command dir the `profiles.yml` is being searched for. I can make that design change with all your sign off @Fleid @dbeatty10 . If we want something else that's quick, let me know here, or if bigger, let's make an issue to figure out this (and other wishlist) `dbt debug` flags.

core/dbt/task/debug.py
VersusFacit: I looked hard into this but the method used here is not truly installation agnostic. It's really difficult to test locally because it won't work with `-e` pip installs. It won't find adapters for beta versions of dbt either. I think what I have here is better for that reason...even if I don't love it. I'm going to split the difference and steal it's way of obtaining the `version`.

VersusFacit: I _did_ find something far better 
```
module_name = f"{'.'.join(cls.__module__.split('.')[:-1])}.__version__"
version = importlib.import_module(module_name).version
```

VersusFacit:
Well, after staring at this code, considering future next steps, and seeing a plate of 🍝 code (i.e. the way state variables in this class were handled is understated but so confusing/conflicting) it's practically impossible to debug code.

I therefore gave it a deep refactor, and I strove to make it far easier to track what's happening. I could tweak knobs more, but the 80% has been reached. 

Test cases:
- existing unit tests
- `dbt debug`
- `dbt debug --connection`
- `dbt debug --profiles-dir NOPE # to see how it fails`
- `dbt debug` after commenting out my `dbt_project.yml`
- `dbt debug` after commenting out my `profiles.yml`

The behavior in all cases is cogent or more often exactly matches what the "stock" `debug` task is on `main` today. Because of how confusing the existing code is, how I've recoded this may see some minor adjustment to logging (e.g. changing order of lines, a newline here, a slightly different wording there).

Fleid:
Hey @VersusFacit,

So you're proposing to yank `--config-dir` out, and replace it by 2 new options?
I'm a bit skittish about doing that here to be honest. It's a big enough change in the interface that it should be in its own issue.

So I would go through this PR just adding `--connection`, then fix `--config-dir` via that other issue (not you, whoever picks it up from Support Rotation), then opening another issue/pr with your most recent ideas (I like them btw).

VersusFacit:
Pardon me, I was merely electing removal of `--config-dir` because it doesn't seem to serve any purpose. 

`--profile-dir` and `--project-dir` already exist

dbeatty10:
@VersusFacit I don't know how often `dbt debug --config-dir` is used, but its documentation is [here](https://docs.getdbt.com/reference/commands/debug):

> The --config-dir option to dbt debug will show the configured location for the profiles.yml file and exit

Since removing it would be a breaking change, I agree with you and @Fleid to leave it in this PR and then split that into a separate issue for further discussion.

.changes/unreleased/Features-20230604-025956.yaml
```diff
@@ -0,0 +1,7 @@
+kind: Features
+body: Revamp debug, add --connection flag, and drop --config-dir flag. Prepare for
+  future refactors/interface changes.
```
dbeatty10, line 3: After https://github.com/dbt-labs/dbt-core/pull/7741/commits/92fcffbd54fa16a1b8c61514f2895de53a3a97ec, the changelog entry can be shortened:
```suggestion
body: Revamp debug, add --connection flag. Prepare for future refactors/interface changes.
```

VersusFacit:
@dbeatty10 , good find on the docs! Weird that and no other flags are documented. It's a shame really. I can make sure there's a docs ticket out of this merge that we change that 😄 

And the changelog here has been duly changed!

---- 

🚚  `--config-dir` background and code has been moved to [here](https://github.com/dbt-labs/dbt-core/pull/7793)

tests/adapter/dbt/tests/adapter/dbt_debug/test_dbt_debug.py
```diff
@@ -51,6 +51,13 @@ def test_ok(self, project):
         run_dbt(["debug"])
         assert "ERROR" not in self.capsys.readouterr().out
 
+    def test_connection_flag(self, project):
+        run_dbt(["debug", "--connection"])
+
+        run_dbt(["debug", "--connection", "--target", "NONE"], expect_pass=False)
+
+        run_dbt(["debug", "--connection", "--profiles-dir", "NONE"], expect_pass=False)
```
colin-rogers-dbt, line 59: Do we need the assert from `test_ok` here? i.e.
`assert "ERROR" not in self.capsys.readouterr().out`

VersusFacit: Oh good point. Adding output checks also indirectly tests a _lot_ of logic changes about how messages are rendered so that's a worthy addition even if not clear ta first glance what's happening.

core/dbt/task/debug.py
```diff
@@ -95,81 +115,140 @@ def path_info(self):
...
+        # Load profile if possible, then load adapter info (which requires the profile)
+        load_profile_status: SubtaskStatus = self._load_profile()
+        fire_event(DebugCmdOut(msg="Using profiles dir at {}".format(self.profiles_dir)))
         fire_event(DebugCmdOut(msg="Using profiles.yml file at {}".format(self.profile_path)))
         fire_event(DebugCmdOut(msg="Using dbt_project.yml file at {}".format(self.project_path)))
-        self.test_configuration()
-        self.test_dependencies()
-        self.test_connection()
+        if load_profile_status.run_status == RunStatus.PASS:
+            if self.profile is not None:
```
colin-rogers-dbt, line 135: This might be easier to read if we flip the if condition:

```
if self.profile is None:
    raise dbt.exceptions.DbtInternalError(
                    "Profile should not be None if loading profile completed"
                )
```

core/dbt/task/debug.py
```diff
@@ -95,81 +115,140 @@ def path_info(self):
...
+    # ===============
+    # Loading profile
+    # ===============
 
-        try:
-            self.project = Project.from_project_root(
-                self.project_dir,
-                renderer,
-                verify_version=self.args.VERSION_CHECK,
+    def _load_profile(self) -> SubtaskStatus:
```
colin-rogers-dbt, line 193: Is there some commonality with https://github.com/dbt-labs/dbt-core/blob/587bbcbf0def587c876cfd7ffe984889ac776e9f/core/dbt/config/runtime.py#L58 that we could extract into a common/shared function?

core/dbt/task/debug.py
```diff
@@ -310,72 +437,57 @@ def test_configuration(self):
...
-        """Return a string containing the error message, or None if there was
-        no error.
-        """
+    def attempt_connection(profile) -> Optional[str]:
+        """Return a string containing the error message, or None if there was no error."""
         register_adapter(profile)
         adapter = get_adapter(profile)
         try:
             with adapter.connection_named("debug"):
+                # set in adapter repos
```
colin-rogers-dbt, line 451: super nit: would `# implemented in adapter class` be more accurate? (technically postgres is in the same repo after all

colin-rogers-dbt (commented):
minor feedback and one question

core/dbt/task/debug.py
VersusFacit: There is some commonalty but I'm concerned about messing with that. I redid this function to get rid of its 3+ side effects. The function you're pointing at affects a lot more than just `debug` and seems to approach loading the profile a bit different, so I'm hesitant to bring them in alignment in this PR.

Also, I have a feeling others will ask us to avoid affecting more than we have to with the size/user changes received from this PR.

core/dbt/task/debug.py
VersusFacit: You're so right lol. On it

core/dbt/task/debug.py
```diff
@@ -58,6 +62,23 @@
...
+)
+
+
+class RunStatus(Enum):
+    PASS = 1
+    FAIL = 2
+    SKIP = 3
+
+
+DEBUG_RUN_STATUS: Dict[str, bool] = {
```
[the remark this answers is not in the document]
VersusFacit: I'd really rather not have fandom boolean literals if possible because of loss of context, even if it seems obvious. This is just syntax sugar but it is more descriptive than having people calculate the boolean on the fly or think about how True/False corresponds to the return value.

core/dbt/task/debug.py
```diff
@@ -95,81 +115,140 @@ def path_info(self):
         open_cmd = dbt.clients.system.open_dir_cmd()
         fire_event(OpenCommand(open_cmd=open_cmd, profiles_dir=self.profiles_dir))
 
-    def run(self):
+    def run(self) -> bool:
         if self.args.config_dir:
             self.path_info()
-            return not self.any_failure
+            return DEBUG_RUN_STATUS["pass"]
```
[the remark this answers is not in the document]
VersusFacit: It should actually. If you trace the original code, it was always rendered as `not False`.

core/dbt/task/debug.py
```diff
@@ -95,81 +115,140 @@ def path_info(self):
...
+        # Load profile if possible, then load adapter info (which requires the profile)
+        load_profile_status: SubtaskStatus = self._load_profile()
+        fire_event(DebugCmdOut(msg="Using profiles dir at {}".format(self.profiles_dir)))
         fire_event(DebugCmdOut(msg="Using profiles.yml file at {}".format(self.profile_path)))
         fire_event(DebugCmdOut(msg="Using dbt_project.yml file at {}".format(self.project_path)))
-        self.test_configuration()
-        self.test_dependencies()
-        self.test_connection()
+        if load_profile_status.run_status == RunStatus.PASS:
+            if self.profile is None:
```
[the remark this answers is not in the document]
VersusFacit: I just flipped per Colin because of the request to avoid `not None` and put the positive case first. I think his _is_ more inline with our inhouse style....I think at least 😅

core/dbt/task/debug.py
```diff
@@ -95,81 +115,140 @@ def path_info(self):
...
+                    path=self.profile_path, url=ProfileConfigDocs
+                ),
             )
-        except dbt.exceptions.DbtConfigError as exc:
-            self.project_fail_details = str(exc)
-            return red("ERROR invalid")
 
-        return green("OK found and valid")
+        raw_profile_data = load_yaml_text(dbt.clients.system.load_file_contents(self.profile_path))
+        if isinstance(raw_profile_data, dict):
```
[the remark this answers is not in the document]
VersusFacit: What/how exactly do you want me to assess? (note this was merely shuffled as opposed to changed code)

core/dbt/task/debug.py
```diff
@@ -95,81 +115,140 @@ def path_info(self):
...
+        else:
+            return SubtaskStatus(
+                log_msg=green("OK found and valid"),
+                run_status=RunStatus.PASS,
+                details="",
+                summary_message="Profile is valid",
+            )
 
-    def _choose_profile_names(self) -> Optional[List[str]]:
+    def _choose_profile_names(self) -> Tuple[List[str], str]:
```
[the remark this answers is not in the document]
VersusFacit: Nah, I'm giving back two distinct pieces of info. In fact, I had to change this at one point but don't have to now. I'll just revert it. Good catch.

core/dbt/task/debug.py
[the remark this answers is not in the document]
VersusFacit: We talked offline and found that this isn't necessary per se although it's a future nice to have

core/dbt/task/debug.py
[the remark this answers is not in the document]
VersusFacit: Using a bool value enum now

core/dbt/task/debug.py
```diff
@@ -227,75 +320,109 @@ def _choose_target_name(self, profile_name: str):
...
+            return green("OK found")
 
-    def test_git(self):
+    # ============
+    # Config tests
+    # ============
+
+    def test_git(self) -> SubtaskStatus:
         try:
             dbt.clients.system.run_cmd(os.getcwd(), ["git", "--help"])
```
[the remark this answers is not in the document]
VersusFacit: Jerco's out till Fri, so I'd propose bringing this back up in the followup issue

Fleid (approved):

dbeatty10 (approved):

PR #7763, opened 2023-06-02
Fixes #7299: dbt retry

### LINKED ISSUE #7299: [Feature] `dbt retry` command

### Describe the feature

If provided with an artifact (`run_results.json`) from a previous command, dbt should be able to `retry` that command.

The `run_results.json` artifact already includes:
- The command (`build`, `seed`, etc)
- The full set of parameters passed into that command
- The full set of nodes that were selected to run
- Which nodes succeeded, errored, skipped

The existing `result` selection method:
- https://docs.getdbt.com/docs/deploy/project-state#the-result-status
- https://docs.getdbt.com/reference/node-selection/methods#the-result-method

This still requires users/applications to store a lot of core-specific logic around:
- which `result` statuses to use, based on the command
- what to do for commands that _don't_ accept selection criteria (i.e. `run-operation`)
- what to do in the case of `--fail-fast` and other "unhandled" errors (etc)

We could do a better job of offering a standard: "This is how you retry a failed dbt command."

### Who will this benefit?

Orchestrators that want an easy way to rerun dbt jobs from point of failure, without needing to split up dbt graphs into standalone tasks for every node

_No response_

### Anything else?

Spiked by @ChenyuLInx back in February:
- https://github.com/dbt-labs/dbt-core/pull/7020

aranke:
resolves #7299 

### Checklist

### DIFF
--- a/tests/functional/retry/test_retry.py
+++ b/tests/functional/retry/test_retry.py
@@ -169,3 +169,11 @@ class TestRetry:
             run_dbt(["retry"], expect_pass=False)
 
         write_file(models__sample_model, "models", "sample_model.sql")
+
+    def test_removed_file_leaf_node(self, project):
+        write_file(models__sample_model, "models", "third_model.sql")
+        run_dbt(["build"], expect_pass=False)
+
+        rm_file("models", "third_model.sql")
+        with pytest.raises(ValueError, match="Couldn't find model 'model.test.third_model'"):
+            run_dbt(["retry"], expect_pass=False)


core/dbt/cli/main.py
```diff
@@ -576,6 +577,48 @@ def run(ctx, **kwargs):
     return results, success
 
 
+# dbt run
+@cli.command("retry")
+@click.pass_context
+@p.defer
```
ChenyuLInx, line 583: I think most of the params listed here should not be supported as retry will reconstruct them from previous run result?

core/dbt/cli/main.py
```diff
@@ -576,6 +577,48 @@ def run(ctx, **kwargs):
...
+@p.vars
+@p.version_check
+@requires.postflight
+@requires.preflight
+@requires.profile
+@requires.project
+@requires.runtime_config
+@requires.manifest
+def retry(ctx, **kwargs):
+    """Compile SQL and execute against the current target database."""
```
ChenyuLInx, line 610: Update comment?

core/dbt/task/retry.py
```diff
@@ -0,0 +1,116 @@
...
+            [
+                result.unique_id
+                for result in self.previous_state.results.results
+                if result.status in RETRYABLE_STATUSES
+            ]
+        )
+
+        cli_command = CMD_DICT.get(self.previous_command_name)
+
+        if "show" in self.previous_args:
```
ChenyuLInx, line 83: Should we just tell the user we can't retry show here?

core/dbt/task/retry.py
```diff
@@ -0,0 +1,116 @@
...
+            del self.previous_args["show"]
+
+        if "resource_types" in self.previous_args and self.previous_args["resource_types"] == []:
```
ChenyuLInx, line 86: Can we add some comment about why this and L89 happens

core/dbt/cli/utils.py
```diff
@@ -0,0 +1,32 @@
+from dbt.cli.flags import Flags
+from dbt.config import RuntimeConfig
+from dbt.config.runtime import Profile, Project, load_project, load_profile
+from dbt.flags import get_flags
+
+
+def get_profile(flags: Flags) -> Profile:
+    # TODO: Generalize safe access to flags.THREADS:
```
ChenyuLInx, line 8: It appears that the functions here are copied from `requires.py`. Please just consolidate it in one place

tests/functional/retry/test_retry.py
ChenyuLInx, line 1: Thanks for the test here! This is great!

.changes/unreleased/Features-20230602-083302.yaml
```diff
@@ -0,0 +1,6 @@
+kind: Features
+body: dbt retry
+time: 2023-06-02T08:33:02.410456-07:00
+custom:
+  Author: aranke
```
ChenyuLInx, line 5: Please make sure @stu-k  is also added to the author.

ChenyuLInx (commented):
Thanks for adding this patch and the tests. I have some question/thoughts regarding some implementation detail.

core/dbt/task/retry.py
```diff
@@ -0,0 +1,140 @@
...
+            "warn_error_options": lambda x: x == {"exclude": [], "include": []},
+        }
+
+        for k, v in bad_args.items():
+            if k in self.previous_args and v(self.previous_args[k]):
+                del self.previous_args[k]
+
+        retry_flags = Flags.from_dict(cli_command, self.previous_args)
+
+        retry_profile = load_profile(
```
ChenyuLInx, line 100: Why we need to reload the profile and project here? Given that the current runtime config already contains these information.
Is it just because we need to reconstruct the runtime_config?
If that's the case we should just overwrite attributes on the current runtime config with the values in retry flags.

core/dbt/task/run_operation.py
```diff
@@ -108,5 +110,6 @@ def run(self) -> RunResultsArtifact:
 
         return results
 
-    def interpret_results(self, results):
+    @classmethod
```
ChenyuLInx, line 113: Why this becomes a class method where all other places are still just a normal method?

core/dbt/task/retry.py
```diff
@@ -0,0 +1,140 @@
...
+    "generate": GenerateTask,
+    "run": RunTask,
+    "seed": SeedTask,
+    "show": ShowTask,
+    "snapshot": SnapshotTask,
+    "test": TestTask,
+    "run-operation": RunOperationTask,
+}
+
+CMD_DICT = {
```
ChenyuLInx, line 39: Can we add two unit test to check
1. the TASK_DICT's key always matches the CMD_DICT's key?
2. for all commands that are in CLI, we either have them defined in this CMD_DICT, or it is defined in the test as something we don't support retry on.

So that next time we add new command a test would fail to tell us to see whether we should add it to here.

ChenyuLInx (changes requested):
Let's rethink a bit about loading profile and project. I would like to avoid it unless there's no way to make things works without load project/profile in the task again.

core/dbt/task/retry.py
```diff
@@ -0,0 +1,140 @@
...
+            [
+                result.unique_id
+                for result in self.previous_state.results.results
+                if result.status in RETRYABLE_STATUSES
+            ]
+        )
+
+        cli_command = CMD_DICT.get(self.previous_command_name)
+
+        bad_args = {
```
ChenyuLInx, line 88: What does bad args mean? is it bad because it doesn't make sense to configure these for retry? Can we add a bit more detail in a comment?

core/dbt/cli/main.py
```diff
@@ -576,6 +577,36 @@ def run(ctx, **kwargs):
     return results, success
 
 
+# dbt run
+@cli.command("retry")
+@click.pass_context
+@p.project_dir
+@p.profiles_dir
+@p.vars
```
ChenyuLInx, line 585: I think we don't support `vars` in retry? As it might change the graph

core/dbt/cli/main.py
```diff
@@ -576,6 +577,36 @@ def run(ctx, **kwargs):
...
+@p.profile
+@p.target
+@p.state
+@p.threads
+@p.fail_fast
```
ChenyuLInx, line 590: We should just get fail fast option from the previous args in run results.json?

core/dbt/cli/main.py
ChenyuLInx: Or is the idea that if the user specified `vars` in the previous command, they have to specify it in the retry command also to make everything work?

core/dbt/task/retry.py
```diff
@@ -0,0 +1,140 @@
...
+            raise DbtRuntimeError(
+                f"Could not find previous run in '{state_path}' target directory"
+            )
+
+        self.previous_args = self.previous_state.results.args
+        self.previous_command_name = self.previous_args.get("which")
+        self.task_class = TASK_DICT.get(self.previous_command_name)
+
+    def run(self):
+        unique_ids = set(
```
ChenyuLInx, line 78: What will happen if I delete the model that failed in my last command?

ChenyuLInx:
Thanks for thinking through the usecase here! There's def more complexity than I originally expected regarding modify files etc. Happy to chat about this live. If there are more questions we can loop Jerco in as needed.

core/dbt/task/retry.py
```diff
@@ -0,0 +1,113 @@
...
+
+
+class RetryTask(ConfiguredTask):
+    def __init__(self, args, config, manifest):
+        super().__init__(args, config, manifest)
+
+        state_path = self.args.state or self.config.target_path
+
+        if self.args.warn_error:
+            RETRYABLE_STATUSES.add(NodeStatus.Warn)
```
jtcohen6, line 52: My understanding is:
- If the previous command passed `--warn-error`, then we will rerun tests or freshness checks that warned
- Otherwise, we will not rerun those warn-status tests/checks

I buy that - let's be sure to document it.

jtcohen6 (commented):
Took for a quick spin, looking cool! Might have more comments later on as I (or @dbeatty10) get a chance to review further

tests/functional/retry/test_retry.py
```diff
@@ -0,0 +1,171 @@
...
+        result = run_dbt(["--warn-error", "build", "--fail-fast"], expect_pass=False)
+
+        assert result.status == RunStatus.Error
+        assert result.node.name == "sample_model"
+
+        results = run_dbt(["retry"], expect_pass=False)
+
+        assert len(results.results) == 1
+        assert results.results[0].status == RunStatus.Error
+        assert results.results[0].node.name == "sample_model"
```
jtcohen6, line 155: From discussing https://github.com/dbt-labs/dbt-core/pull/7763/files#r1218248760 live: @aranke added a test case for this. Current behavior of `--fail-fast` is to _only_ return _one_ result, and only for the node that encountered the error (which triggered the fast failure).

Let's consider this out of scope for this PR (adding the `retry` command), but open a new issue that we may want to tackle (if we have time) in this epic

jtcohen6, line 155: From discussing https://github.com/dbt-labs/dbt-core/pull/7763/files#r1218248760 live: @aranke added a test case for this. Current behavior of `--fail-fast` is to only return **exactly one** result, and only for the node that encountered the error (which triggered the fast failure). That's true no matter how many nodes are being run, and even if previous nodes had already succeeded.

Let's consider this out of scope for this PR (adding the `retry` command), but open a new issue that we may want to tackle (if we have time) in this epic

core/dbt/task/run_operation.py
ChenyuLInx: Cool! I think making things more consistent is not in scope of this PR so current implementation works!

core/dbt/task/retry.py
```diff
@@ -0,0 +1,113 @@
...
+
+        self.previous_state = PreviousState(
```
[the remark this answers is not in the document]
ChenyuLInx: Maybe we should call it a different name so that semantically we distinguish it from previous state in the other place?

core/dbt/cli/main.py
```diff
@@ -576,6 +577,36 @@ def run(ctx, **kwargs):
     return results, success
 
 
+# dbt run
```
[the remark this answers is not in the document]
dbeatty10: Fixed in https://github.com/dbt-labs/dbt-core/pull/7932

Issue #7503, opened 2023-05-03
Begin "basic" parsing of SemanticModel to the Manifest as nodes


QMalcolm:
In #7499 we defined dbt-core's representation of a Semantic node. In #7456 we defined the the expected YAML representation of a SemanticModel. This issue is to begin parsing SemanticModels defined in YAML to SemanticModel nodes on the Manifest. We do have a prototype of one way this could be done, #7187, if inspiration is needed.

SemanticModel node parsing is slight more complicated than Metric Parsing (#7501) in that additional nodes can be created. Specifically if a `Measure` on the SemanticModel is defined with `create_metric: true` then a Metric Node should be created for the measure. This issue **_IS NOT_** expected to handle that edge case. We have a separate issue, #7504, for tackling that specific issue.

jtcohen6:
During `ref` resolution (end of parsing), we will need to include an object representing the resolved `relation` for the model that the SemanticModel is referencing, via `model: ref('something')`.

We should include an object that contains:
- relation path components (`database`, `schema`, `identifier`)
- stringified `relation_name`, which will already apply quoting + include policies that vary by adapter (e.g. `"schema"."identifier"`)

If a SemanticModel references a disabled model, raise an error :)

PR #7769, opened 2023-06-02
Paw/semantic models

peterallenwebb:
resolves #

### Checklist

### DIFF
--- a/core/dbt/contracts/graph/manifest.py
+++ b/core/dbt/contracts/graph/manifest.py
@@ -1212,6 +1212,9 @@ class Manifest(MacroMethods, DataClassMessagePackMixin, dbtClassMixin):
         self.docs[doc.unique_id] = doc
         source_file.docs.append(doc.unique_id)
 
+    def add_semantic_model(self, source_file: SchemaSourceFile, semantic_model: Metric):
+        pass
+
     # end of methods formerly in ParseResult
 
     # Provide support for copy.deepcopy() - we just need to avoid the lock!
--- a/core/dbt/contracts/graph/nodes.py
+++ b/core/dbt/contracts/graph/nodes.py
@@ -552,6 +552,30 @@ class CompiledNode(ParsedNode):
         return self.depends_on.macros
 
 
+@dataclass
+class FileSlice(dbtClassMixin, Replaceable):
+    """Provides file slice level context about what something was created from.
+
+    Implementation of the dbt-semantic-interfaces `FileSlice` protocol
+    """
+
+    filename: str
+    content: str
+    start_line_number: int
+    end_line_number: int
+
+
+@dataclass
+class Metadata(dbtClassMixin, Replaceable):
+    """Provides file context about what something was created from.
+
+    Implementation of the dbt-semantic-interfaces `Metadata` protocol
+    """
+
+    repo_file_path: str
+    file_slice: FileSlice
+
+
 # ====================================
 # CompiledNode subclasses
 # ====================================
--- a/core/dbt/contracts/graph/nodes.py
+++ b/core/dbt/contracts/graph/nodes.py
@@ -1424,6 +1424,20 @@ class Group(BaseNode):
 
 
 # ====================================
+# SemanticModel
+# ====================================
+
+
+@dataclass
+class SemanticModel(GraphNode):
+    description: Optional[str]
+    # node_relation: NodeRelation
+    # entities: Sequence[Entity]
+    # measures: Sequence[Measure]
+    # dimensions: Sequence[Dimension]
+
+
+# ====================================
 # Patches
 # ====================================
 
--- a/core/dbt/contracts/graph/unparsed.py
+++ b/core/dbt/contracts/graph/unparsed.py
@@ -661,6 +661,63 @@ class UnparsedGroup(dbtClassMixin, Replaceable):
             raise ValidationError("Group owner must have at least one of 'name' or 'email'.")
 
 
+#
+# semantic interfaces unparsed objects
+#
+
+
+@dataclass
+class UnparsedEntity(dbtClassMixin, Replaceable):
+    name: str
+    type: str  # actually an enum
+    description: Optional[str] = None
+    role: Optional[str] = None
+    expr: Optional[str] = None
+
+
+@dataclass
+class UnparsedMeasureAggregationParameters(dbtClassMixin, Replaceable):
+    percentile: Optional[float] = None
+    use_discrete_percentile: bool = False
+    use_approximate_percentile: bool = False
+
+
+@dataclass
+class UnparsedMeasure(dbtClassMixin, Replaceable):
+    name: str
+    agg: str  # enum
+    description: Optional[str] = None
+    create_metric: Optional[bool] = None
+    expr: Optional[str] = None
+    agg_params: Optional[UnparsedMeasureAggregationParameters] = None
+    non_additive_dimension: Optional[Dict[str, Any]] = None  # TODO: Refine type as class?
+    agg_time_dimension: Optional[str] = None
+
+
+@dataclass
+class UnparsedDimension(dbtClassMixin, Replaceable):
+    name: str
+    description: Optional[str]
+    type: str  # actually an enum
+    is_partition: bool = False
+    type_params: Optional[Dict[str, Any]] = field(
+        default_factory=dict
+    )  # TODO: Refine type as class?
+    expr: Optional[str] = None
+    # TODO metadata: Optional[Metadata] (this would actually be the YML for the dimension)
+
+
+@dataclass
+class UnparsedSemanticModel(dbtClassMixin, Replaceable):
+    name: str
+    description: Optional[str]
+    model: str  # looks like ref(blah, blah) TODO: Further parsing for ref?
+    # node_relation:   # does not exist in YML, will be populated from "model"
+    entities: List[UnparsedEntity] = field(default_factory=list)
+    measures: List[UnparsedMeasure] = field(default_factory=list)
+    dimensions: List[UnparsedDimension] = field(default_factory=list)
+
+
 def normalize_date(d: Optional[datetime.date]) -> Optional[datetime.datetime]:
     """Convert date to datetime (at midnight), and add local time zone if naive"""
     if d is None:
--- a/core/dbt/parser/schema_yaml_readers.py
+++ b/core/dbt/parser/schema_yaml_readers.py
@@ -1,8 +1,13 @@
 from dbt.parser.schemas import YamlReader, SchemaParser
 from dbt.parser.common import YamlBlock
 from dbt.node_types import NodeType
-from dbt.contracts.graph.unparsed import UnparsedExposure, UnparsedMetric, UnparsedGroup
-from dbt.contracts.graph.nodes import Exposure, Metric, Group
+from dbt.contracts.graph.unparsed import (
+    UnparsedExposure,
+    UnparsedGroup,
+    UnparsedMetric,
+    UnparsedSemanticModel,
+)
+from dbt.contracts.graph.nodes import Exposure, Group, Metric, SemanticModel
 from dbt.exceptions import DbtInternalError, YamlParseDictError, JSONValidationError
 from dbt.context.providers import generate_parse_exposure, generate_parse_metrics
 from dbt.contracts.graph.model_config import MetricConfig, ExposureConfig
@@ -269,3 +274,37 @@ class GroupParser(YamlReader):
                 raise YamlParseDictError(self.yaml.path, self.key, data, exc)
 
             self.parse_group(unparsed)
+
+
+class SemanticModelParser(YamlReader):
+    def __init__(self, schema_parser: SchemaParser, yaml: YamlBlock):
+        super().__init__(schema_parser, yaml, "semantic_models")
+        self.schema_parser = schema_parser
+        self.yaml = yaml
+
+    def parse_semantic_model(self, unparsed: UnparsedSemanticModel):
+        package_name = self.project.project_name
+        unique_id = f"{NodeType.SemanticModel}.{package_name}.{unparsed.name}"
+        path = self.yaml.path.relative_path
+
+        fqn = self.schema_parser.get_fqn_prefix(path)
+        fqn.append(unparsed.name)
+
+        parsed = SemanticModel(
+            fqn=fqn,
+            name=unparsed.name,
+            package_name=package_name,
+            resource_type=NodeType.SemanticModel,
+        )
+
+        self.manifest.add_semantic_model(self.yaml.file, parsed)
+
+    def parse(self):
+        for data in self.get_key_dicts():
+            try:
+                UnparsedSemanticModel.validate(data)
+                unparsed = UnparsedSemanticModel.from_dict(data)
+            except (ValidationError, JSONValidationError) as exc:
+                raise YamlParseDictError(self.yaml.path, self.key, data, exc)
+
+            self.parse_semantic_model(unparsed)
--- a/tests/functional/semantic_models/test_semantic_model_parsing.py
+++ b/tests/functional/semantic_models/test_semantic_model_parsing.py
@@ -1,0 +1,51 @@
+import pytest
+
+from dbt.cli.main import dbtRunner
+
+schema_yml = """models:
+  - name: fct_revenue
+    description: This is the model fct_revenue. It should be able to use doc blocks
+
+semantic_models:
+  - name: revenue
+    description: This is the revenue semantic model. It should be able to use doc blocks
+    model: ref('fct_revenue')
+
+    measures:
+      - name: txn_revenue
+        expr: revenue
+        agg: sum
+
+    dimensions:
+      - name: ds
+        type: time
+        expr: created_at
+        type_params:
+          is_primary: True
+          time_granularity: day
+
+    entities:
+      - name: user
+        type: foreign
+        expr: user_id
+"""
+
+fct_revenue_sql = """select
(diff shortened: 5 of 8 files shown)


core/dbt/contracts/graph/nodes.py
```diff
@@ -552,6 +540,30 @@ def depends_on_macros(self):
...
+    """
+
+    filename: str
+    content: str
+    start_line_number: int
+    end_line_number: int
+
+
+@dataclass
+class Metadata(dbtClassMixin, Replaceable):
```
gshank, line 557: We already have a Metadata class in model_config.py.  Could this be given a more descriptive name?

core/dbt/contracts/graph/nodes.py
```diff
@@ -1399,6 +1411,28 @@ class Group(BaseNode):
     resource_type: NodeType = field(metadata={"restrict": [NodeType.Group]})
 
 
+# ====================================
+# SemanticModel and related classes
+# ====================================
+
+
+@dataclass
+class NodeRelation(dbtClassMixin, Replaceable):
```
gshank, line 1420: This a lot like the StateRelation class and a number of other relation classes that we have. I personally don't like using "alias" in these classes instead of "identifier" because without an actual identifier there's nothing for it to be an alias of. Can we use "identifier" and create an "alias" property as a synonym?

core/dbt/contracts/graph/nodes.py
```diff
@@ -552,6 +540,30 @@ def depends_on_macros(self):
         return self.depends_on.macros
 
 
+@dataclass
+class FileSlice(dbtClassMixin, Replaceable):
```
gshank, line 544: I'm not a fan of including the "Replaceable" class in every single class we create and in general think that it's a bad pattern anyway. Can we leave it off and if somebody really wants that functionality on a particular object it can be specifically added?

gshank (commented):
In general it looks like a perfectly reasonable starting point. I did have a few nits... 

The discussion about our various Relation classes might be a larger thing and I'm not actually asking for a change there, those classes just seem to be proliferating lately.

core/dbt/contracts/graph/nodes.py
peterallenwebb: I will leave it out, with pleasure!

core/dbt/contracts/graph/nodes.py
peterallenwebb: @QMalcolm What do you think about calling this class SemanticMetadata? I don't think it should affect serialization.

core/dbt/contracts/graph/nodes.py
peterallenwebb: In this case, a name change would affect the serialization format we're targeting, unless we control it another way. Is there still flexibility on that format, Quigley? What do you think about Gerda's suggestion?

jtcohen6: fwiw - I also prefer `identifier` to `alias` here

QMalcolm: I agree that we should make the move to `identifier`. Is it okay if we do that as a follow up task though? It'd require the change in DSI, re-releasing DSI, and then making the update here. Not the biggest deal. There shouldn't be any gotchas. However, as this is an under the hood change and we're on a tight deadline, I think we could push this to extra work to be done after Wednesday (June 7th)

schemas/dbt/manifest/v10.json
```diff
@@ -4020,138 +4086,6 @@
       "additionalProperties": false,
       "description": "FreshnessThreshold(warn_after: Optional[dbt.contracts.graph.unparsed.Time] = <factory>, error_after: Optional[dbt.contracts.graph.unparsed.Time] = <factory>, filter: Optional[str] = None)"
     },
-    "FreshnessMetadata": {
```
QMalcolm, line 4023: These removals seem unrelated to the PR. Any idea what happened here?

QMalcolm (commented):
This looks good to me. I like the direction we're heading 🙂

core/dbt/contracts/graph/nodes.py
QMalcolm: This was me by the way 🙈 I believe I included it because we included it on everything 🙃 I'll make a note for myself that it's not a pattern we want to continue.

core/dbt/contracts/graph/nodes.py
QMalcolm: I think that'd be a great distinction to make. And yea the class name shouldn't affect the serialization 👍

core/dbt/contracts/graph/nodes.py
gshank: Yeah, I wouldn't mind a tech debt ticket to remove it from every place that it's not needed, and look at the places where it's actually called to see if it should be. Some of that was due to having Compiled and non-compiled versions of nodes, I think.

gshank: https://github.com/dbt-labs/dbt-core/issues/7802

core/dbt/parser/schema_yaml_readers.py
```diff
@@ -269,3 +274,46 @@ def parse(self):
...
+
+        fqn = self.schema_parser.get_fqn_prefix(path)
+        fqn.append(unparsed.name)
+
+        parsed = SemanticModel(
+            description=unparsed.description,
+            fqn=fqn,
+            model=unparsed.model,
+            name=unparsed.name,
+            node_relation=None,  # Resolved from the value of "model" after parsing
```
MichelleArk, line 298: for my own understanding -- we need to resolve the relation later on because not all model nodes are available in the manifest for lookup at this point?

core/dbt/parser/manifest.py
```diff
@@ -1176,6 +1179,28 @@ def process_metrics(self, config: RuntimeConfig):
...
+
+                    refd_node = self.manifest.ref_lookup.find(
+                        input_model_name, input_package_name, None, self.manifest
+                    )
+                    if isinstance(refd_node, ModelNode):
+                        semantic_model.node_relation = NodeRelation(
+                            alias=refd_node.alias,
+                            schema_name=refd_node.schema,
+                            database=refd_node.database,
+                        )
```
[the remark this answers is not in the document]
MichelleArk: similarly refd_node could be `None` if the ref_lookup doesn't find anything -- let's handle that case as well.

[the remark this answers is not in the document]
MichelleArk: We can handle it by doing a secondary lookup in `disabled_lookup` if the model is disabled we should probably [raise a separate error](https://github.com/dbt-labs/dbt-core/issues/7503#issuecomment-1538583101) than if it is not found at all.

core/dbt/contracts/graph/nodes.py
```diff
@@ -1411,6 +1423,28 @@ class Group(BaseNode):
...
+# ====================================
+# SemanticModel and related classes
+# ====================================
+
+
+@dataclass
+class NodeRelation(dbtClassMixin):
+    alias: str
+    schema_name: str  # TODO: Could this be called simply "schema" so we could reuse StateRelation?
+    database: Optional[str] = None
```
MichelleArk, line 1435: From [this comment](https://github.com/dbt-labs/dbt-core/issues/7503#issuecomment-1538583101) -- does this still need to include a unified relation_name that respects quoting + include policies?

MichelleArk:
Looks like a good starting point! Most of my comments are around error handling -- those aren't necessarily blocking this chunk of work but we should create follow-up issues to track the ones we want to address after this is merged.

core/dbt/contracts/graph/nodes.py
MichelleArk: if we do -- we'll probably need the adapter.Relation to do something like this during compilation:
```python
adapter = get_adapter(self.config)
relation_cls = adapter.Relation
relation_name = str(relation_cls.create_from(self.config, node))
```
(from https://github.com/dbt-labs/dbt-core/blob/main/core/dbt/compilation.py#L486-L488)

core/dbt/contracts/graph/nodes.py
peterallenwebb: Done. I chose the name SourceFileMetadata.

core/dbt/contracts/graph/manifest.py
```diff
@@ -706,6 +707,7 @@ class Manifest(MacroMethods, DataClassMessagePackMixin, dbtClassMixin):
     public_nodes: MutableMapping[str, PublicModel] = field(default_factory=dict)
     project_dependencies: Optional[ProjectDependencies] = None
     publications: MutableMapping[str, PublicationConfig] = field(default_factory=dict)
+    semantic_models: MutableMapping[str, SemanticModel] = field(default_factory=dict)
```
gshank, line 710: The existing pattern is the node objects use "Model" and the dictionaries use "nodes", that is Model nodes are found in "nodes" dictionary, "PublicModel" nodes are found in public_nodes. Could this be the "semantic_nodes" to preserve that pattern?

core/dbt/contracts/files.py
```diff
@@ -228,6 +228,7 @@ class SchemaSourceFile(BaseSourceFile):
     groups: List[str] = field(default_factory=list)
     # node patches contain models, seeds, snapshots, analyses
     ndp: List[str] = field(default_factory=list)
+    semantic_models: List[str] = field(default_factory=list)
```
gshank, line 231: semantic_nodes? (see other comment)

tests/functional/semantic_models/test_semantic_model_parsing.py
```diff
@@ -0,0 +1,57 @@
...
+"""
+
+fct_revenue_sql = """select
+  1 as id,
+  10 as user_id,
+  1000 as revenue,
+  current_timestamp as created_at"""
+
+
+class TestSemanticModelParsing:
```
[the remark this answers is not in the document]
peterallenwebb: I agree that testing is too light in this changeset. The directive Jeremy has given Quigley and I is to get the MetricFlow team unblocked ASAP and clean things up in future issues. So I've opened this one to address testing: https://github.com/dbt-labs/dbt-core/issues/7820, and we will work on it in the current sprint.

core/dbt/contracts/graph/manifest.py
```diff
@@ -1246,6 +1249,11 @@ def add_doc(self, source_file: SourceFile, doc: Documentation):
         self.docs[doc.unique_id] = doc
         source_file.docs.append(doc.unique_id)
 
+    def add_semantic_model(self, source_file: SchemaSourceFile, semantic_model: SemanticModel):
```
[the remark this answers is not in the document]
gshank: We already have an "add_node" method, and currently the pattern is that we have a different "add_<n>" method for each dictionary type. So I don't think that's the kind of change we would like to make at this point.

core/dbt/contracts/graph/nodes.py
```diff
@@ -1411,6 +1423,28 @@ class Group(BaseNode):
     resource_type: NodeType = field(metadata={"restrict": [NodeType.Group]})
 
 
+# ====================================
+# SemanticModel and related classes
+# ====================================
+
+
+@dataclass
+class NodeRelation(dbtClassMixin):
```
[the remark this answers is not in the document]
gshank: That one doesn't actually have the relation_name in it, and I think they want that. Settling on a common relation class is the "larger issue" that I mentioned we didn't need to solve at this particular moment.

core/dbt/contracts/graph/nodes.py
gshank: I don't think the SemanticModels actually get compiled, since they're yaml-only. There is a question of whether they need the individual pieces (identifier/schema/database) or just the relation_name...

core/dbt/contracts/graph/unparsed.py
```diff
@@ -661,6 +661,60 @@ def validate(cls, data):
             raise ValidationError("Group owner must have at least one of 'name' or 'email'.")
 
 
+#
+# semantic interfaces unparsed objects
+#
+
+
+@dataclass
+class Entity(dbtClassMixin, Replaceable):
```
gshank, line 670: There's a bunch more Replaceable classes that we can skip including here.

core/dbt/parser/schema_yaml_readers.py
gshank: SemanticModels come from yaml only, and the referenced models should always be sql based and so already resolvable when this is parsed.

core/dbt/parser/manifest.py
```diff
@@ -1176,6 +1179,28 @@ def process_metrics(self, config: RuntimeConfig):
                 continue
             _process_metrics_for_node(self.manifest, current_project, exposure)
 
+    def process_semantic_models(self) -> None:
+        for semantic_model in self.manifest.semantic_models.values():
+            if semantic_model.model:
+                statically_parsed = py_extract_from_source(f"{{{{ {semantic_model.model} }}}}")
```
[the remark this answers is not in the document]
peterallenwebb: Agree, and we'll follow up in the current sprint with #7822.

core/dbt/parser/manifest.py
```diff
@@ -1176,6 +1179,28 @@ def process_metrics(self, config: RuntimeConfig):
...
+                if statically_parsed["refs"]:
+
+                    ref = statically_parsed["refs"][0]
+                    if len(ref) == 2:
+                        input_package_name, input_model_name = ref
+                    else:
+                        input_package_name, input_model_name = None, ref[0]
+
+                    refd_node = self.manifest.ref_lookup.find(
+                        input_model_name, input_package_name, None, self.manifest
```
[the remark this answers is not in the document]
peterallenwebb: Agree, and we'll follow up in the current sprint with #7822.

core/dbt/parser/manifest.py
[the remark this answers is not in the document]
peterallenwebb: Agree, and we'll follow up in the current sprint with #7822.

core/dbt/parser/schema_yaml_readers.py
peterallenwebb: @MichelleArk Yes, I discussed this with Gerda and she thought we could possibly handle it immediately, since models would likely have been created, but doing it at the end would be most likely to succeed without complications.

@aranke I'm can follow up the idea of lazy evaluation in the current sprint. If you have a specific method of doing it in mind can you add it as a comment to #7822?

core/dbt/contracts/graph/nodes.py
peterallenwebb: @QMalcolm I'm not sure of the answer to this, but I suspect the answer may be yes. What makes sense from the perspective of MetricFlow integration?

core/dbt/contracts/graph/manifest.py
peterallenwebb: In that case, do you think we just rely on the existing nodes dictionary, and not have a special case for SemanticModel objects? Would it matter that the SemanticModels are not yet linked into the compiled graph?

gshank: The original "metrics" went in a "metrics" dictionary...  This has been a bit ad-hoc and not closely thought through, but it feels like the existing pattern is that SQL things go in nodes (plus seeds...) and yaml-generated things go in their own dictionaries. There are some assumptions in the rest of the code that match, such as the links in the file objects, etc. So I think it still makes sense to put semantic models in their own dictionary. If anything I'd be tempted to separate out some of the existing things that are in the nodes dictionary into their own dictionaries and maybe make some combined "indexes" (for cases where we don't want to loop over "nodes" all the time...)

core/dbt/contracts/graph/unparsed.py
peterallenwebb: Done in latest commit.

core/dbt/contracts/graph/nodes.py
QMalcolm: @MichelleArk Yep! We need to include the unified relation_name, and it's desirable that it respects quoting + include policies.

As for whether we need both the individual pieces as well as the `relation_name`, for this PR yes, long term we don't have to. MetricFlow is fine with having the parts or the relation_name. It was suggested that we use this node_relation object as it was a pattern that was already used elsewhere, thus the shape of the object currently. Though MetricFlow currently uses the `relation_name` attribute as seen [here](https://github.com/dbt-labs/metricflow/blob/4e64f1656382a9b2001e597681823f0b0cf751a3/metricflow/dataset/convert_semantic_model.py#L407), and ignores the individual properties. However, the individual properties as well as the `relation_name` are currently required by the protocol. So for the scope of this PR, we should include the individual properties as well as the `relation_name`. Otherwise we'd first need to update and release a new DSI (which would include a number of other schema changes that have been made), and then propagate all those changes in core. Although the `node_relation` changes might be trivial in core, propagating the other changes with the new DSI could eat a fair bit of time.

core/dbt/contracts/graph/manifest.py
peterallenwebb: This is all really good to know. It seems like the right direction to me.

core/dbt/contracts/files.py
peterallenwebb: @gshank To clarify, do you think we should change the name in the yml file, or just for our internal property names?

gshank: Just for for the internal property names. I think "semantic_models" in the yaml files is fine, it matches the "models" section we already have. Naming is so hard.

gshank (approved):

Issue #6997, opened 2023-02-17
[Feature] Cross-database implementation of is_distinct_from


joellabes:
### Is this your first time submitting a feature request?

- [X] I have read the [expectations for open source contributors](https://docs.getdbt.com/docs/contributing/oss-expectations)
- [X] I have searched the existing issues, and I could not find an existing issue for this feature
- [X] I am requesting a straightforward extension of existing dbt functionality, rather than a Big Idea better suited to a discussion

### Describe the feature

If I had a dollar for every time I had done 

```sql
select a, b, 
  a = b or (a is null and b is null) as are_the_same
from table
```
or 
```sql
select a, b, 
  coalesce(a, -1) = coalesce(b, -1) as are_the_same --very dangerous! do not do!
from table
```
,
I would have a lot of money. 

Turns out that there is a solution for this in the sql standard! 

![image](https://user-images.githubusercontent.com/7335046/219536480-9b19f856-3e94-40d7-9eae-21a1a382482e.png)

We should add this as a cross-db macro - a speedy Google looks like most of our key adapters support it natively [except for Redshift](https://popsql.com/learn-sql/redshift/how-to-compare-two-values-when-one-is-null-in-redshift)

### Describe alternatives you've considered

Doing nothing. 

### Who will this benefit?

This seems useful for two reasons: 
- By creating and documenting it, more people might stumble over it
- I would have given my left leg for this when building the first version of the metrics package, when I was trying to do joins that were potentially null 

### Are you interested in contributing this feature?

Yep! But it'll be my first cross-adapter contribution so I'll need some handholding

dbeatty10:
Thanks for opening @joellabes !

What kind of syntax are you imagining, something like this?
```
{{ dbt.is_distinct_from(a, b) }}
```

Would you imagine it being being paired with `is_not_distinct_from` too? Or just rely on the user to do the negation?

For the Redshift implementation (or any other non-conforming database), we'd want to be diligent with the implementation.
As you noted as `dangerous!`, we wouldn't want to do [this](https://popsql.com/learn-sql/redshift/how-to-compare-two-values-when-one-is-null-in-redshift) implementation, but one that works correctly for all possible input values. (I suspect the SQL in that link would be incorrect when one of `width` & `height` is `0` and the other is `NULL`.)

The truth table [here](https://wiki.postgresql.org/wiki/Is_distinct_from) would be handy for integration tests.

[...]

Issue #4725, opened 2022-02-14
[CT-230] DBT seed converts empty strings to nulls


snagapuri-mdsol:
### Describe the bug
DBT seed process converts 
- ,"",
- ,,
- ,NULL,

all above cases to NULLs

I would say atleast quoted empty strings should be allowed to be as blank as its being explicit enough to be used as blank: ,"",

### Steps to reproduce
Add any csv seed file with data like below:
column1,column2,column3
NULL,"",

All the above three columns produce NULLs in database table

### Expected results
NULL,,NULL
Atleast second column in the above example can be left alone to be blank in order to have an option in seed process to add blanks in database tables

### Actual results
NULL,NULL,NULL

### System information

**Which database are you using dbt with?**
- snowflake

**The output of `dbt --version`:**
```
installed version: 1.0.1
   latest version: 1.0.1

Up to date!

Plugins:
  - snowflake: 1.0.0
```

ChenyuLInx:
@snagapuri-mdsol  thanks for opening this! We are still going to stick to the behavior for empty strings in CSVs. Since it is implemented in python standard CSV reader, and it seems to be standard outside python.
To achieve the specific result you want, you can opt for a "placeholder" empty string value (e.g. `EMPTY`), coupled with a staging model that wraps around the seed.

PR #5428, opened 2022-06-30
Move data type macros into dbt-core

jtcohen6:
- Part of resolution for https://github.com/dbt-labs/dbt-utils/issues/598
- Related to https://github.com/dbt-labs/dbt-core/issues/5317, but does not actually resolve that tech debt

### Description

- Move `type_{X}` macros from `dbt-utils` into `dbt-core`. In almost all cases, these macros are just a sugarier interface for [`api.Column.translate_type`](https://github.com/dbt-labs/dbt-core/blob/d63ad4cd82619dafed7d68194df08d256c15f85b/core/dbt/adapters/base/column.py#L8-L24), which leverages built-in type capabilities
- Add functional tests for these data types

After some back-and-forth on what the right interface ought to be, we've landed on keeping the existing one. Users can confidently access a "string" type for data type casting, by calling macro named `type_string`, which returns the correct word for string (`text`, `varchar`, etc) on that database

### Checklist

- [x] I have read [the contributing guide](https://github.com/dbt-labs/dbt-core/blob/main/CONTRIBUTING.md) and understand what's expected of me
- [x] I have signed the [CLA](https://docs.getdbt.com/docs/contributor-license-agreements)
- [x] I have run this code in development and it appears to resolve the stated issue
- [x] This PR includes tests, or tests are not required/relevant for this PR
- [ ] I have [opened an issue to add/update docs](https://github.com/dbt-labs/docs.getdbt.com/issues/new/choose), or docs changes are not required/relevant for this PR
- [x] I have run `changie new` to [create a changelog entry](https://github.com/dbt-labs/dbt-core/blob/main/CONTRIBUTING.md#Adding-CHANGELOG-Entry)

PR #7776, opened 2023-06-04
Fix not equals comparison to be null-safe for adapters/utils tests

### LINKED ISSUE #7778: [Bug] `not {{ equals(actual, expected) }}` isn't working properly

### Current Behavior

Surfaced by @lllong33 in https://github.com/dbt-labs/dbt-core/issues/7670#issuecomment-1563840006:

> @sdebruyn  @dbeatty10 It looks like `not null` is `not true`, can't write it like that. FYI
> ```sql
> select version();  -- PostgreSQL 12.3 on x86_64-pc-linux-gnu
> 
> with t1 as (
> 	select 1 as actual, 2 as expected
> 	union all 
> 	select 1 as actual, 1 as expected
> 	union all 
> 	select 1 as actual, cast(null as int) as expected
> 	union all 
> 	select cast(null as int) as actual, cast(null as int) as expected 
> )
> 
> select * from t1 
> --where actual = expected or (actual is null and expected is null) -- normal: 1,1; null,null
> --where actual != expected 
> --	or (actual is null and expected is not null )
> --	or (actual is not null and expected is null ) -- normal: 1,2; 1,null
> where not (actual = expected or (actual is null and expected is null)) -- issue: actual=1, expected=null is not work 
> ```
> image
> 
> 

### Expected Behavior

Negation of `equals` should work.

### Steps To Reproduce

See https://github.com/dbt-labs/dbt-core/issues/7670#issuecomment-1564713670 for a reproduction case.

### Relevant log output

_No response_

### Environment

```markdown
- OS:
- Python:
- dbt:
```

_No response_

### Additional Context

The [three-valued logic of SQL](https://modern-sql.com/concept/three-valued-logic) is tricky, and we should use a "[fully equivalent substitute](https://modern-sql.com/feature/is-distinct-from#conforming-alterantives) for A is not distinct from B that works in all SQL databases."

And then we should make the sure the "adapters zone" includes relevant tests.

dbeatty10:
resolves #7778

follow-up on #7672

### Description

The bug report describes how functional tests in adapters/utils are not currently null-safe (specifically for "not equals" comparisons).

#### Main things
This pull request does two main things:
1. Implements a ["SQL standard conforming alternative"](https://modern-sql.com/feature/is-distinct-from#conforming-alterantives) to `A is not distinct from B`.
2. Implements comprehensive 🤞 tests inspired by the [Snowflake docs](https://docs.snowflake.com/en/sql-reference/functions/is-distinct-from#examples) and demo'd [here](https://github.com/dbt-labs/dbt-core/issues/7670#issuecomment-1564713670).

#### Ancillary things
This PR needed some ancillary things as well that arose with the improved `equals` comparisons:
- Fix tests for `concat` and `hash` by using empty strings ('') instead of `null`

See a combination of [three-valued logic of SQL](https://modern-sql.com/concept/three-valued-logic) and https://github.com/dbt-labs/dbt-core/issues/4725 for context why these changes were necessary and done this particular way.

### Future-facing opportunities
The code in this PR would make the implementation of #6997 as easy as possible:
1. Move the definition of the `equals` macro from `tests/adapter/dbt/tests/adapter/utils/base_utils.py` to:

    `core/dbt/include/global_project/macros/utils/equals.sql`
    ```
    {% macro equals(expr1, expr2) -%}
      {{ return(adapter.dispatch('equals', 'dbt')(expr1, expr2)) }}
    {%- endmacro %}
    
    {% macro default__equals(expr1, expr2) -%}
    case when (({{ expr1 }} = {{ expr2 }}) or ({{ expr1 }} is null and {{ expr2 }} is null))
        then 0
        else 1
    end = 0
    {%- endmacro %}
    ```
1. Remove any other code that is no longer necessary
1. Adopt this test within each adapter
1. Resolve this issue: https://github.com/dbt-labs/docs.getdbt.com/issues/3469
1. Add the new macro to the migration guide for adapter maintainers

### Checklist

- [x] This PR includes tests 👍 
- [x] Updates to docs are not relevant

### DIFF
--- a/tests/adapter/dbt/tests/adapter/utils/base_utils.py
+++ b/tests/adapter/dbt/tests/adapter/utils/base_utils.py
@@ -2,9 +2,11 @@ import pytest
 from dbt.tests.util import run_dbt
 
 macros__equals_sql = """
-{% macro equals(actual, expected) %}
-{# -- actual is not distinct from expected #}
-(({{ actual }} = {{ expected }}) or ({{ actual }} is null and {{ expected }} is null))
+{% macro equals(expr1, expr2) -%}
+case when (({{ expr1 }} = {{ expr2 }}) or ({{ expr1 }} is null and {{ expr2 }} is null))
+    then 0
+    else 1
+end = 0
 {% endmacro %}
 """
 
--- a/tests/adapter/dbt/tests/adapter/utils/base_utils.py
+++ b/tests/adapter/dbt/tests/adapter/utils/base_utils.py
@@ -17,6 +17,15 @@ where not {{ equals(actual, expected) }}
 {% endtest %}
 """
 
+macros__replace_empty_sql = """
+{% macro replace_empty(expr) -%}
+case
+    when {{ expr }} = 'EMPTY' then ''
+    else {{ expr }}
+end
+{% endmacro %}
+"""
+
 
 class BaseUtils:
     # setup
@@ -25,6 +34,7 @@ class BaseUtils:
         return {
             "equals.sql": macros__equals_sql,
             "test_assert_equal.sql": macros__test_assert_equal_sql,
+            "replace_empty.sql": macros__replace_empty_sql,
         }
 
     # make it possible to dynamically update the macro call with a namespace
--- a/tests/adapter/dbt/tests/adapter/utils/test_equals.py
+++ b/tests/adapter/dbt/tests/adapter/utils/test_equals.py
@@ -15,19 +15,6 @@ class BaseEquals:
             "equals.sql": macros__equals_sql,
         }
 
-    # make it possible to dynamically update the macro call with a namespace
-    # (e.g.) 'dateadd', 'dbt.dateadd', 'dbt_utils.dateadd'
-    def macro_namespace(self):
-        return ""
-
-    def interpolate_macro_namespace(self, model_sql, macro_name):
-        macro_namespace = self.macro_namespace()
-        return (
-            model_sql.replace(f"{macro_name}(", f"{macro_namespace}.{macro_name}(")
-            if macro_namespace
-            else model_sql
-        )
-
     @pytest.fixture(scope="class")
     def seeds(self):
         return {
@@ -37,12 +24,8 @@ class BaseEquals:
     @pytest.fixture(scope="class")
     def models(self):
         return {
-            "equal_values.sql": self.interpolate_macro_namespace(
-                MODELS__EQUAL_VALUES_SQL, "equals"
-            ),
-            "not_equal_values.sql": self.interpolate_macro_namespace(
-                MODELS__NOT_EQUAL_VALUES_SQL, "equals"
-            ),
+            "equal_values.sql": MODELS__EQUAL_VALUES_SQL,
+            "not_equal_values.sql": MODELS__NOT_EQUAL_VALUES_SQL,
         }
 
     def test_equal_values(self, project):
--- a/tests/adapter/dbt/tests/adapter/utils/test_equals.py
+++ b/tests/adapter/dbt/tests/adapter/utils/test_equals.py
@@ -1,0 +1,71 @@
+import pytest
+from dbt.tests.adapter.utils.base_utils import macros__equals_sql
+from dbt.tests.adapter.utils.fixture_equals import (
+    SEEDS__DATA_EQUALS_CSV,
+    MODELS__EQUAL_VALUES_SQL,
+    MODELS__NOT_EQUAL_VALUES_SQL,
+)
+from dbt.tests.util import run_dbt, relation_from_name
+
+
+class BaseEquals:
+    @pytest.fixture(scope="class")
+    def macros(self):
+        return {
+            "equals.sql": macros__equals_sql,
+        }
+
+    # make it possible to dynamically update the macro call with a namespace
+    # (e.g.) 'dateadd', 'dbt.dateadd', 'dbt_utils.dateadd'
+    def macro_namespace(self):
+        return ""
+
+    def interpolate_macro_namespace(self, model_sql, macro_name):
+        macro_namespace = self.macro_namespace()
+        return (
+            model_sql.replace(f"{macro_name}(", f"{macro_namespace}.{macro_name}(")
+            if macro_namespace
+            else model_sql
+        )
+
+    @pytest.fixture(scope="class")
+    def seeds(self):
+        return {
+            "data_equals.csv": SEEDS__DATA_EQUALS_CSV,
+        }
+
+    @pytest.fixture(scope="class")
+    def models(self):
+        return {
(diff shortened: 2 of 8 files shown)


dbeatty10:
Good news: the functional test implemented in https://github.com/dbt-labs/dbt-core/pull/7776/commits/617cf1e3b7218953f584434140b465dacada17d1 is now catching the exact problem reported by @lllong33 in https://github.com/dbt-labs/dbt-core/issues/7670#issuecomment-1563840006.

https://github.com/dbt-labs/dbt-core/actions/runs/5170455117/jobs/9313364211?pr=7776

image

tests/adapter/dbt/tests/adapter/utils/base_utils.py
```diff
@@ -15,6 +17,15 @@
 {% endtest %}
 """
 
+macros__replace_empty_sql = """
+{% macro replace_empty(expr) -%}
+case
+    when {{ expr }} = 'EMPTY' then ''
```
VersusFacit, line 23: Is EMPTY something `dbt` does? I have a vague recollection of that.

tests/adapter/dbt/tests/adapter/utils/base_utils.py
```diff
@@ -2,9 +2,11 @@
...
 
 macros__equals_sql = """
-{% macro equals(actual, expected) %}
-{# -- actual is not distinct from expected #}
-(({{ actual }} = {{ expected }}) or ({{ actual }} is null and {{ expected }} is null))
+{% macro equals(expr1, expr2) -%}
+case when (({{ expr1 }} = {{ expr2 }}) or ({{ expr1 }} is null and {{ expr2 }} is null))
+    then 0
+    else 1
+end = 0
```
VersusFacit, line 9: `1`s and `0`s to be platform agnostic?

tests/adapter/dbt/tests/adapter/utils/test_null_compare.py
```diff
@@ -14,8 +14,8 @@ class BaseMixedNullCompare(BaseUtils):
     @pytest.fixture(scope="class")
     def models(self):
         return {
-            "test_mixed_null_compare.yml": MODELS__TEST_MIXED_NULL_COMPARE_SQL,
-            "test_mixed_null_compare.sql": MODELS__TEST_MIXED_NULL_COMPARE_YML,
+            "test_mixed_null_compare.yml": MODELS__TEST_MIXED_NULL_COMPARE_YML,
+            "test_mixed_null_compare.sql": MODELS__TEST_MIXED_NULL_COMPARE_SQL,
```
VersusFacit, line 18: good catch!

VersusFacit (approved):
Glad to know there's even more ways to  test equality than I thought. When we do adapter zone tests, we'll have to watch our backs on the fact that things like `distinct from` may not exist on all dbms.

Regardless, approved (assuming your responses are "yep" to comments as I'm expecting them to be)

tests/adapter/dbt/tests/adapter/utils/base_utils.py
dbeatty10: Yep, you got it!

Markus Winand explains it nicely [here](https://modern-sql.com/feature/is-distinct-from#conforming-alterantives).

The SQL standard way to express this is:
```
a is not distinct from b
```

But not all databases support that syntax.

tests/adapter/dbt/tests/adapter/utils/test_equals.py
```diff
@@ -0,0 +1,71 @@
...
+        return {
+            "equals.sql": macros__equals_sql,
+        }
+
+    # make it possible to dynamically update the macro call with a namespace
+    # (e.g.) 'dateadd', 'dbt.dateadd', 'dbt_utils.dateadd'
+    def macro_namespace(self):
+        return ""
+
+    def interpolate_macro_namespace(self, model_sql, macro_name):
```
VersusFacit, line 23: Help me understand the background of this step if ya please.

tests/adapter/dbt/tests/adapter/utils/base_utils.py
VersusFacit: Oh no
> This is often not a problem because [SQL generally treats unknown like false when making binary decisions](https://modern-sql.com/concept/three-valued-logic#general-rule) such as accepting or rejecting a row for a where clause.

tests/adapter/dbt/tests/adapter/utils/base_utils.py
dbeatty10: The `EMPTY` here is from @ChenyuLInx's advice in https://github.com/dbt-labs/dbt-core/issues/4725#issuecomment-1047266185

Basically, dbt seed .csv files can't produce an empty string -- they can only produce `NULL`. So we do a .csv files like this instead:
```csv
input_1,input_2,output
a,b,ab
a,EMPTY,a
EMPTY,b,b
```

After creating an intermediate model that uses the `replace_empty` macro to convert `EMPTY` to empty strings, then the intermediate model will look like:

| input_1 | input_2 | output |
|---------|---------|--------|
| a       | b       | ab     |
| a       |         | a      |
|         | b       | b      |

tests/adapter/dbt/tests/adapter/utils/test_equals.py
dbeatty10: Good question!

This is copy-pasta from [here](https://github.com/dbt-labs/dbt-core/pull/5428/files#diff-68f6cf649856a2177e399f35632114882e33d31d6378260d2ee73063ec6d52c8R5-R16).

I don't know all the details, but it's in a [couple places](https://github.com/search?q=repo%3Adbt-labs%2Fdbt-core+%22def+interpolate_macro_namespace%22&type=code) currently.

I don't remember what happens if I pull it out.

VersusFacit: I just tested removing it locally. It seems like the answer is "nothing" for this test at least 🤷‍♀️

dbeatty10: Cool -- thanks for checking. I pulled it out of this PR in 9aca97a00b60f4ea2af5afd546c6d54ff9a69a13.

We might want/need it to go back in if/when we do #6997 in order for it to be consistent with all the other cross-database macros [here](https://github.com/dbt-labs/dbt-core/tree/main/core/dbt/include/global_project/macros/utils) that inherit [this](https://github.com/dbt-labs/dbt-core/blob/dc35f56baa6876e6381b46f9b99b45f29d58f857/tests/adapter/dbt/tests/adapter/utils/base_utils.py#L28-L39).

But it might also be that we don't need it and the macro namespace interpolation logic was only needed for the migration of macros from dbt-utils to dbt-core 🤷

PR #7783, opened 2023-06-05
log PublicationArtifactAvailable even when partially parsing & public models unchanged

### LINKED ISSUE #7782: partial parsing + PublicationArtifactAvailable

MichelleArk:
resolves #7782

### Checklist

### DIFF
--- a/tests/functional/partial_parsing/test_partial_parsing.py
+++ b/tests/functional/partial_parsing/test_partial_parsing.py
@@ -69,6 +69,7 @@ from tests.functional.partial_parsing.fixtures import (
     groups_schema_yml_one_group_model_in_group2,
     groups_schema_yml_two_groups_private_orders_valid_access,
     groups_schema_yml_two_groups_private_orders_invalid_access,
+    public_models_schema_yml,
 )
 
 from dbt.exceptions import CompilationError, ParsingError, DuplicateVersionedUnversionedError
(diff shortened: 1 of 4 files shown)


tests/functional/partial_parsing/test_partial_parsing.py
```diff
@@ -804,3 +805,31 @@ def test_pp_groups(self, project):
         )
         with pytest.raises(ParsingError):
             results = run_dbt(["--partial-parse", "run"])
+
+
+class TestPublicationArtifactAvailable:
```
ChenyuLInx, line 810: This seems like a lot of things to run in a test for a small functionality, is it possible to write unit for this case instead of functional tests? Totally understand if that's really tricky to do.

tests/functional/partial_parsing/test_partial_parsing.py
```diff
@@ -804,3 +805,31 @@ def test_pp_groups(self, project):
...
+    @pytest.fixture(scope="class")
+    def models(self):
+        return {
+            "orders.sql": orders_sql,
+            "schema.yml": public_models_schema_yml,
+        }
+
+    def test_pp_publication_artifact_available(self, project):
+        # initial run with public model logs PublicationArtifactAvailable
+        results, log_output = run_dbt_and_capture(["--debug", "--log-format", "json", "run"])
```
ChenyuLInx, line 820: Does `parse` work here? Or it has to be `run`(which will be slightly slower).

ChenyuLInx (approved):
The changes looks good to me, feel free to merge as is. I hope we can write tests that run faster for changes like this. But also understand that it might be hard with the current codebase.

tests/functional/partial_parsing/test_partial_parsing.py
gshank: Yeah, the "test_partial_parsing" file is getting pretty long. Might be a good idea to split it up into smaller pieces. But probably not for this pull request. Maybe we should open another ticket...

gshank (approved):

tests/functional/partial_parsing/test_partial_parsing.py
MichelleArk: good call! updated

tests/functional/partial_parsing/test_partial_parsing.py
MichelleArk: Unit testing partial parsing in particular may be especially tricky given how dependent it is on reading & writing files.. but I agree that splitting out some of the longer tests into their own pieces would help for debugging / understanding coverage / readability.

PR #496, opened 2017-07-26
Fix/dist bug

drewbanin:
lower/strip dist key after null check

Issue #77, opened 2016-07-26
first time project configuration


drewbanin:
- [x] create ~/.dbt/profiles.yml on `dbt init` (if it doesn't exist)
- [x] clone a starter repo (cc @jthandy)
- [x] show/open profiles dir/file at end of setup
  - [x] `dbt debug --config-dir` ?

jthandy:
david krevitt recently had the excellent suggestion that `dbt init` go through a series of interactive prompts to set up the config file. I don't think that's a must for this round, but it's a solid suggestion.

jthandy:
@drewbanin I think we should remove this from the Minor improvements milestone. This is one that I think would benefit from @cmcarthur's input, so would rather leave it for just a bit longer if you're ok with that.

drewbanin:
@jthandy sure, sounds good

drewbanin:
let's think about the optimal first-time flow for dbt

PR #7793, opened 2023-06-05
Remove --config-dir instead of fixing #7774

### LINKED ISSUE #7774: [CT-2632] [Regression] --config-dir with dbt debug is thowing a cannot parse error

### Current Behavior

image

Looks like the `--config-dir` flag is broken! Bug arises likely in [this structured logging related PR](https://github.com/dbt-labs/dbt-core/pull/4183) 

So this has been broken since `November 2021` or thereabouts.

The fix is to add a `str()` call on line ~108.

### Expected Behavior

Going back to [this positively ancient 6 year old issue](https://github.com/dbt-labs/dbt-core/pull/439/files), it seems like the original purpose is "if specified, dbt will show path information for this project."

Thankfully the current flag description is sort of consistent with this: `Show the configured location for the profiles.yml file and exit`. Really though, it just renders `open <what dbt thinks the profiles dir is>` which you manually copy-paste to the CLI. 

**Instead of fixing this, because it's causing confusion on our own team, I'm proposing, we just remove this.**

### Steps To Reproduce

1. kick a `dbt debug --config-dir`

### Relevant log output

```shell
see screenshot above
```

### Environment

```markdown
- OS: MacOS 13.3.1
- Python: 3.9.9
- dbt: dev
```

_No response_

### Additional Context

_No response_

VersusFacit:
resolves #7774
### Description

So `--config-dir` is broken.

Also, it's less a flag and more a subcommand which prevents the running of any other `dbt debug` runtime. 

This PR aims to streamline the experience and remove the ambiguously-named "config" dir.

### Preferred alternative to fixing (breaking change to interface)

This may affect day to day workflows somewhat, but this flag doesn't fit in orchestrated jobs (unless you jam a square peg into a round hole). To be clear, `dbt` users **_already_** get the useful info out of `--config-dir` in the output of `dbt debug` on every run, by default.

Because it is redundant, I'd rather nix this flag and the code it relies on.

There are no tests for it, and we document it rarely (see below). [Some users reference it](https://stackoverflow.com/questions/71482406/how-to-set-location-of-profiles-yml-and-dbt-project-yml-files-in-dbt), but it causes confusion (which of these do I use and why?), and the output in `debug` already has the tools to do what we want `--config-dir` to do (i.e. show where `dbt` thinks the profile dir is). 

Even that SO post has some confusion over how to use the flag.

### Aligned with existing API

Rather than keep this as a distinct flag, we can rely on these existing ones:
image
and slip the profile dir info into the existing output.
image 
The first line in this trio so what I'm proposing in lieu of keeping `--config_dir` around.

Revised `dbt debug --help` output.

image

#### Appendix -- Docs references to adjust
1. https://docs.getdbt.com/docs/core/connect-data-platform/connection-profiles
2. https://docs.getdbt.com/reference/commands/debug
3. https://docs.getdbt.com/guides/best-practices/debugging-errors#failed-to-connect

We should update with guidance on how to use the existing `profiles-dir` flag and how to read the new and improved `dbt debug` output from #496 instead (i.e. a plain old `dbt debug` command will always -- even before this PR -- give you the _profiles dir + profiles file_, no separate subcommand needed). 

### Checklist

### DIFF
--- a/core/dbt/cli/params.py
+++ b/core/dbt/cli/params.py
@@ -43,7 +43,7 @@ compile_docs = click.option(
 config_dir = click.option(
     "--config-dir",
     envvar=None,
-    help="Show the configured location for the profiles.yml file and exit",
+    help="Print a system-specific command to access the directory that the current dbt project is searching for a profiles.yml. Then, exit. This flag renders other debug step flags no-ops.",
     is_flag=True,
 )
 
--- a/core/dbt/task/debug.py
+++ b/core/dbt/task/debug.py
@@ -102,13 +102,14 @@ class DebugTask(BaseTask):
             return None
         return self.project.profile_name
 
-    def path_info(self):
-        open_cmd = dbt.clients.system.open_dir_cmd()
-        fire_event(OpenCommand(open_cmd=open_cmd, profiles_dir=self.profiles_dir))
-
     def run(self) -> bool:
+        # WARN: this is a legacy workflow that is not compatible with other runtime flags
         if self.args.config_dir:
-            self.path_info()
+            fire_event(
+                OpenCommand(
+                    open_cmd=dbt.clients.system.open_dir_cmd(), profiles_dir=str(self.profiles_dir)
+                )
+            )
             return DebugRunStatus.SUCCESS.value
 
         version: str = get_installed_version().to_version_string(skip_matcher=True)
--- a/.changes/unreleased/Fixes-20230621-030733.yaml
+++ b/.changes/unreleased/Fixes-20230621-030733.yaml
@@ -1,0 +1,6 @@
+kind: Fixes
+body: Fix a bad implicit string conversion regression in debug --config-dir code.
+time: 2023-06-21T03:07:33.815966-07:00
+custom:
+  Author: versusfacit
+  Issue: "7774"


dbeatty10:
Reading the bug report, it sounds like the fix for https://github.com/dbt-labs/dbt-core/issues/7774 might be relatively simple:
> The fix is to add a str() call on line ~108.

If we choose to remove `--config-dir`, I'd rather mark the bug report in https://github.com/dbt-labs/dbt-core/issues/7774 as `wont_fix` and create a stand-alone feature request related to removing this feature.

@VersusFacit I'm not _trying_ to create extra work, but would you be willing to open a separate feature request for your proposal to remove the `dbt debug --config-dir` flag? That would allow us to separate the bug report and the feature request. That way we can discuss the pros/cons of that approach there instead of in this PR. Also, 

If that's too much to ask, I'd be willing to create that feature request -- just let me know. I'd just be relying on you to advocate for the removal (since I won't be able to do that effectively because my default position is frequently "keep things the same to avoid breaking changes").

dbeatty10:
Offline, we decided to continue discussing here within this PR whether or not to drop `dbt debug --config-dir`. If we go with removing the flag we will create a separate ticket than https://github.com/dbt-labs/dbt-core/issues/7774 since this not a fix for a regression.

Reading https://github.com/dbt-labs/dbt-core/issues/77, it sounds like the original intent of this flag was to:
> show/open profiles dir/file [after `dbt init`]

I'm not sure how often it is actually used like that in practice. I don't recall seeing any other regression reports related to this other than #7774 by @VersusFacit, which suggests it's used relatively infrequently, and possibly never 😎 

That being said, my default approach is most frequently to avoid breaking changes -- less chance of accidentally breaking someone and less work to add it to a [migration guide](https://docs.getdbt.com/guides/migration/versions).

What say you @jtcohen6? Drop it? Keep it?

Fleid:
In an ideal world I would keep it and fix it :P

core/dbt/task/debug.py
```diff
@@ -102,13 +102,14 @@ def project_profile(self):
             return None
         return self.project.profile_name
 
-    def path_info(self):
-        open_cmd = dbt.clients.system.open_dir_cmd()
-        fire_event(OpenCommand(open_cmd=open_cmd, profiles_dir=self.profiles_dir))
-
     def run(self) -> bool:
+        # WARN: this is a legacy workflow that is not compatible with other runtime flags
```
VersusFacit, line 106: Flattened the code and commented it for the next person who wonders why this here in the `run` command.

core/dbt/cli/params.py
```diff
@@ -43,7 +43,7 @@
 config_dir = click.option(
     "--config-dir",
     envvar=None,
-    help="Show the configured location for the profiles.yml file and exit",
+    help="Print a system-specific command to access the directory that the current dbt project is searching for a profiles.yml. Then, exit. This flag renders other debug step flags no-ops.",
```
VersusFacit, line 46: Disambiguated flag.

colin-rogers-dbt (approved):

PR #7811, opened 2023-06-06
Fixes #7753: Fix regression in `run-operation` to not require the name of the package to run

### LINKED ISSUE #7753: [Pre-Regression] `run-operation` requires the name of the package to run macros from 1.6b2

### Current Behavior

Since 1.6b2, `run-operation` requires the name of the package when the macro is from another package, otherwise an error is raised.

As part of the integration tests check for dbt-external-table, we do a `dbt run-operation stage_external_sources`. This command used to work until 1.6b1 but form 1.6b2 it started raising

```
dbt could not find a macro with the name "stage_external_sources" in the "dbt_external_tables_integration_tests" package
```

The current workaround I found was to qualify the name of the macro with the package it comes from: `dbt run-operation dbt_external_tables.stage_external_sources`

You can see all the details in the different Circle CI checks for [this PR](https://github.com/dbt-labs/dbt-external-tables/pull/211)

It shoud either be listed as a breaking change or we should still be able to call the macro without the package name

### Steps To Reproduce

Described under "Current Behavior"

### Relevant log output

```shell
see Circle CI logs
```

### Environment

```markdown
- OS: NA
- Python: NA
- dbt (working version): 1.6b1
- dbt (regression version): 1.6b2
```

_No response_

### Additional Context

_No response_

aranke:
resolves #7753 

### Checklist

### DIFF
--- a/.changes/unreleased/Under the Hood-20230615-163217.yaml
+++ b/.changes/unreleased/Under the Hood-20230615-163217.yaml
@@ -1,0 +1,6 @@
+kind: Under the Hood
+body: Upgrade to dbt-semantic-interfaces v0.1.0dev5
+time: 2023-06-15T16:32:17.602189-07:00
+custom:
+  Author: QMalcolm
+  Issue: "7853"
--- a/.changes/unreleased/Under the Hood-20230619-135201.yaml
+++ b/.changes/unreleased/Under the Hood-20230619-135201.yaml
@@ -1,0 +1,6 @@
+kind: Under the Hood
+body: Move from dbt-semantic-intefaces 0.1.0dev5 to 0.1.0dev7
+time: 2023-06-19T13:52:01.947206-07:00
+custom:
+  Author: QMalcolm
+  Issue: "7898"
--- a/core/dbt/contracts/graph/nodes.py
+++ b/core/dbt/contracts/graph/nodes.py
@@ -53,8 +53,7 @@ from dbt_semantic_interfaces.references import (
     SemanticModelReference,
 )
 from dbt_semantic_interfaces.references import MetricReference as DSIMetricReference
-from dbt_semantic_interfaces.type_enums.metric_type import MetricType
-from dbt_semantic_interfaces.type_enums.time_granularity import TimeGranularity
+from dbt_semantic_interfaces.type_enums import MetricType, TimeGranularity
 
 from .model_config import (
     NodeConfig,
@@ -1308,7 +1307,7 @@ class MetricInputMeasure(dbtClassMixin):
     def measure_reference(self) -> MeasureReference:
         return MeasureReference(element_name=self.name)
 
-    def post_aggregation_measure_referenc(self) -> MeasureReference:
+    def post_aggregation_measure_reference(self) -> MeasureReference:
         return MeasureReference(element_name=self.alias or self.name)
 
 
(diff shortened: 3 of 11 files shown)


core/dbt/task/run_operation.py
```diff
@@ -66,8 +67,17 @@ def run(self) -> RunResultsArtifact:
...
+            macro_name, self.config.project_name, package_name
+        )
+
+        try:
+            unique_id = macro.unique_id  # type: ignore[union-attr]
+            fqn = unique_id.split(".")
+        except AttributeError:
+            raise DbtRuntimeError(
+                f"dbt could not find a macro with the name '{macro_name}' in any package"
+            )
```
MichelleArk, line 80: This section could be refactored as following to avoid the #type: ignore workarounds and make it more clear what's happening with the `macro.unique_id` access. Catching the error on the `None` value of the returned macro feels like an indirect way of checking whether the macro was found or not. 

Additionally, we could avoid making a breaking change to the `RunResult` unique id (from `operation.` to `macro.`) if we use the attributes on the resolved macro: 

```python
        macro = (
            self.manifest.find_macro_by_name(macro_name, self.config.project_name, package_name)
            if self.manifest
            else None
        )

        if macro:
            fqn = [NodeType.Operation, macro.package_name, macro.name]
            unique_id = macro.unique_id
        else:
            raise DbtRuntimeError(
                f"dbt could not find a macro with the name '{macro_name}' in any package"
            )
```

tests/functional/retry/test_retry.py
```diff
@@ -134,7 +134,7 @@ def test_run_operation(self, project):
         )
 
         expected_statuses = {
-            "operation.test.alter_timezone": RunStatus.Error,
+            "macro.test.alter_timezone": RunStatus.Error,
```
MichelleArk, line 137: this is a breaking change from the user perspective - unless it's an intentional change we want to make or its strictly necessary for implementation, i think we should avoid making it.

core/dbt/task/run_operation.py
```diff
@@ -66,8 +67,17 @@ def run(self) -> RunResultsArtifact:
...
-        unique_id = ".".join(fqn)
+        macro = self.manifest.find_macro_by_name(  # type: ignore[union-attr]
+            macro_name, self.config.project_name, package_name
+        )
+
+        try:
+            unique_id = macro.unique_id  # type: ignore[union-attr]
+            fqn = unique_id.split(".")
+        except AttributeError:
+            raise DbtRuntimeError(
```
MichelleArk, line 78: for my own understanding: if the macro is not found - we'll actually hit this error earlier on from the `self._run_unsafe` call via `adapter.execute_macro`, right? If so, maybe this should actually be a DbtInternalError - since we don't expect it to be reached if everything is running as expected.

core/dbt/task/run_operation.py
```diff
@@ -66,8 +67,17 @@ def run(self) -> RunResultsArtifact:
         end = datetime.utcnow()
 
         package_name, macro_name = self._get_macro_parts()
```
MichelleArk, line 69: non-blocking suggestion: we don't use the `agate.Table` return value from `_run_unsafe` calls anywhere in this file, so perhaps we could return `package_name, macro_name` from `_run_unsafe` instead to avoid having to call it again here.

tests/functional/retry/test_retry.py
MichelleArk: I don't have a strong opinion on whether the unique_ids should look like `operation.*` or `macro.*`. I'm not sure what the context is but if there was some rationale for opting for `operation.*` when we introduced this initially, that probably still holds.

core/dbt/task/run_operation.py
jtcohen6: > Additionally, we could avoid making a breaking change to the `RunResult` unique id (from `operation.` to `macro.`)

FWIW - I believe this is new in v1.6 (https://github.com/dbt-labs/dbt-core/pull/7655), so it's not a breaking change for existing functionality in prior versions. Whether this node is better considered a `macro` or an `operation` is a fair question. IIRC, we use the `operation` node type for `on-run-*` hooks ("project-level" hooks). Whereas the resource run by `run-operation` is really just a `macro`... but of course, the command is named `run-operation`.

MichelleArk (approved):

Issue #7404, opened 2023-04-18
[CT-2434] [Feature] Create SemanticManifest artifact with every run


callum-mcdata:
### Is this your first time submitting a feature request?

- [X] I have read the [expectations for open source contributors](https://docs.getdbt.com/docs/contributing/oss-expectations)
- [X] I have searched the existing issues, and I could not find an existing issue for this feature
- [X] I am requesting a straightforward extension of existing dbt functionality, rather than a Big Idea better suited to a discussion

### Describe the feature

With the introduction of semantic objects and metrics planned for 1.6 (eagle-eyed among you keep an eye out for a discussion), we should plan on creating another artifact that is a subset of the larger manifest.

This artifact should contain:
- semantic_models
- metrics

These nodes should also be contained in the larger manifest but the separate artifact makes it easier for MetricFlow to consume the information that it needs without needing to parse the entire manifest.

### Describe alternatives you've considered

Not doing this

### Who will this benefit?

The teams working on the integration with MetricFlow and anyone who wants to use the Semantic Layer!

### Are you interested in contributing this feature?

Yes!

QMalcolm:
### Blocked by:

- #7501
- #7503
- https://github.com/dbt-labs/dbt-semantic-interfaces/issues/17

jtcohen6:
We will want to use the serialization strategy defined in `dbt-semantic-interfaces` to produce this artifact

Issue #7800, opened 2023-06-06
[SPIKE] [CT-2651] Add SemanticModels and Metrics to the DAG


QMalcolm:
We've added SemanticModel and Metric nodes to the manifest, but we haven't added them to the dag. We need to get them added to the dag so that visualizations based on the DAG include SemanticModels and Metrics.

Note that inclusion of the SemanticModel node in the dag should include tested support for partial parsing. After initial implementation, changes to the semantic_model section of schema files is not triggering partial (or any) re-parse, as it should.

jtcohen6:
@QMalcolm Are you thinking about the visualization in `dbt-docs`? There & elsewhere, the main requirement from within `dbt-core` is that these resource types are included in `manifest.json` and the output of `dbt list`, with requisite information in their `depends_on.nodes`

jtcohen6:
Should `metrics` and `semantic_models` land in the same top-level key (`semantic_nodes`) of the manifest + `graph` context var? I think maybe!

PR #7812, opened 2023-06-07
Update spec MetricNode for dbt x MetricFlow integration

### LINKED ISSUE #7500: [CT-2522] Redefine Metric node for the Manifest and update parsing
There are two new node types being added to the manifest as part of the MetricFlow integration work: SemanticModel nodes and Metric nodes. This ticket addresses the Metric nodes. In [dbt-semantic-interfaces](https://github.com/dbt-labs/dbt-semantic-interfaces) we're adding a protocol of a [Metric](https://github.com/dbt-labs/dbt-semantic-interfaces/issues/18), which will have the shared standard definition of a Metric. In dbt-core we need to compose this protocol with base node classes to define dbt-core's representation of a Metric. Additionally there are some additional fields we should expose as defined in #7456

### Blocked by:

- https://github.com/dbt-labs/dbt-core/issues/7470
- https://github.com/dbt-labs/dbt-semantic-interfaces/issues/18

QMalcolm:
resolves #7500 #7404

### Checklist


core/dbt/contracts/graph/nodes.py
```diff
@@ -1285,27 +1311,76 @@ def same_contents(self, old: Optional["Exposure"]) -> bool:
 # ====================================
 
 
+@dataclass
+class WhereFilter(dbtClassMixin, Replaceable):
```
peterallenwebb, line 1315: Gerda recommended in my review that we not use Replaceable unless we need it, and I don't think we need it on these classes.

core/dbt/parser/schema_yaml_readers.py
```diff
@@ -125,6 +144,102 @@ def __init__(self, schema_parser: SchemaParser, yaml: YamlBlock):
         self.schema_parser = schema_parser
         self.yaml = yaml
 
+    def _get_input_measure(
```
peterallenwebb, line 147: These helper functions look good enough for a beta, but I think we should look at ways to make the code less repetitive and mechanical.

More specifically, if we are going to continue along these lines, it would be better to have a way to declare how the unparsed objects map to the parsed objects, rather than performing the conversion manually for these and every new object.

core/dbt/contracts/graph/manifest.py
```diff
@@ -1319,14 +1319,7 @@ class WritableManifest(ArtifactMixin):
...
     def compatible_previous_versions(self):
-        return [
-            ("manifest", 4),
-            ("manifest", 5),
-            ("manifest", 6),
-            ("manifest", 7),
-            ("manifest", 8),
-            ("manifest", 9),
-        ]
+        return []
```
peterallenwebb, line 1322: As discussed in slack, we'll probably want to add these back.

QMalcolm:
I'd prefer to rebase off of main, but for the sanity of those in the midst of reviewing, I'm gonna merge main into this branch

core/dbt/contracts/graph/manifest.py
```diff
@@ -985,6 +988,20 @@ def analysis_lookup(self) -> AnalysisLookup:
             self._analysis_lookup = AnalysisLookup(self)
         return self._analysis_lookup
 
+    @property
+    def pydantic_semantic_manifest(self) -> Optional[PydanticSemanticManifest]:
```
peterallenwebb, line 992: Borderline nitpicky, the return type is actually a non-Optional PydanticSemanticManifest, which is the more attractive return type anyway.

QMalcolm: Oh right, good catch! I had planned on having it return `None` if there were not any SemanticModel or Metric nodes 🤦 I think returning an empty semantic model is a pretty good default. I'll get the typing updated, and then also update `write_semantic_manifest` to not have to anticipate it ever being `None`

peterallenwebb (approved):
Overall this looks good to me as a beta/starting-point. We will be following up the major remaining issues I see in this sprint with:

#7800 
#7821 
#7827

core/dbt/contracts/graph/nodes.py
```diff
@@ -566,7 +568,7 @@ class FileSlice(dbtClassMixin, Replaceable):
 
 
 @dataclass
-class SourceFileMetadata(dbtClassMixin, Replaceable):
+class Metadata(dbtClassMixin, Replaceable):
```
MichelleArk, line 571: is this rename intentional? re: https://github.com/dbt-labs/dbt-core/pull/7769#discussion_r1222086236

QMalcolm: Nope! looks like the merge of `main` did screw some things up 😞

PR #7816, opened 2023-06-07
Display public models in output of list command

### LINKED ISSUE #7496: [CT-2517] Graph selection for public models from dependency projects
Currently `dbt ls` does not include any models from publication artifacts (dependency projects). Should it? My inclination would be to output public model dependencies the same way that models from installed packages output. 

E.g, a project with `project_a.public_model` available s a dependency: 
```
❯ dbt ls
14:43:02  Running with dbt=1.6.0-a1
14:43:02  target not specified in profile 'postgres', using 'default'
14:43:02  Found 2 models, 0 tests, 0 snapshots, 0 analyses, 307 macros, 0 operations, 0 seed files, 2 sources, 0 exposures, 0 metrics, 0 groups
consumer_project.cross_project_ref
consumer_project.model
source:consumer_project.jaffle_shop.customers
source:consumer_project.jaffle_shop.orders
```

gshank:
resolves #7496 

### Description

Add public models to the graph and the selection code to enable listing public models and using selection syntax for upstream nodes of public nodes.

Note: There's some noise in the selector_methods.py because I couldn't stand the non-standard node/real_node thing and switched it all to unique_id/node, which doesn't totally confuse people.

### Checklist


peterallenwebb (approved):
Should we consider having the PublicModel class inherit from GraphNode? I'm concerned that introducing a "node that isn't really a node" adds more complexity than we will want to manage. On the other hand, maybe the solution is to use the ManifestOrPublicNode protocol more widely, in which case I'd suggest renaming it to something more evocative.

Another lingering question: Aren't all nodes "graph nodes"? If they aren't, what do we mean by node?

I'm going to approve, since I think these changes do what they say on the tin and don't constrain our options going forward, but I'd be interested in your thoughts.

gshank:
We're not doing it this way now.

PR #7838, opened 2023-06-09
Fix: `dbt show --inline` with `private` models

### LINKED ISSUE #7837: [Bug] Allow referencing private models within `dbt show` "inline" queries
Currently, dbt will raise a `DbtReferenceError` when trying to preview an `--inline` query that `ref`s a private model, because the inline query is not part of the same model `group` as the private model.

In my opinion:
- It should be possible to preview a private model, or a query that references that model
- If you then save that query in a model in a different `group`, you'll start seeing the reference error — which makes sense!

### Reproduction case

```yml
groups:
  - name: my_cool_group
    owner:
      name: Me

models:
  - name: my_model
    access: private
    config:
      group: my_cool_group
```

```bash
$ dbt show --inline "select * from {{ ref('my_model') }}"
22:51:45  Running with dbt=1.6.0-b2
22:51:45  Encountered an error:
Parsing Error
  Node sql operation.my_dbt_project.inline_query attempted to reference node model.my_dbt_project.my_model, which is not allowed because the referenced node is private to the my_cool_group group.
```

(While here: let's change the string rendering of `sql operation` to something without spaces - `sqloperation`)

jtcohen6:
resolves #7837

### Description

Allow `dbt show --inline` queries to select from grouped models with `access: private`. These queries will still raise a `DbtReferenceError` if saved as a "permanent" resource (model, snapshot, analysis, etc) outside the `group`.

### Checklist

- ~I have [opened an issue to add/update docs](https://github.com/dbt-labs/docs.getdbt.com/issues/new/choose), or docs changes are not required/relevant for this PR~


core/dbt/node_types.py
```diff
@@ -26,14 +26,14 @@ class NodeType(StrEnum):
...
-    SqlOperation = "sql operation"
+    SqlOperation = "sqloperation"
     Documentation = "doc"
     Source = "source"
     Macro = "macro"
     Exposure = "exposure"
     Metric = "metric"
     Group = "group"
-    SemanticModel = "semantic model"
+    SemanticModel = "semanticmodel"
```
jtcohen6, line 36: This is not in scope of the current fix; it just annoys me that there's a space in the `unique_id`. I can kick out and open as a separate PR

jtcohen6: Opened as separate issue: https://github.com/dbt-labs/dbt-core/issues/7841

Will move commit to separate PR

MichelleArk (approved):

PR #7844, opened 2023-06-12
Remove Replaceable class

### LINKED ISSUE #7802: [CT-2652] Remove "Replaceable" class usage as much as possible
We pull in a Replaceable class which contains a "replace" method in a lot of our classes. This method is designed to replace the current object with a newly constructed one based on kwargs, and is in general not a very good idea. If we want to replace a current object it would be better to be a lot more explicit and intentional about what's happening.

Remove the usage Replaceable class where it's found in dbt-core, look for breakage, and determine if we really need to do this or could just update the current object.

gshank:
resolves #7802

### Description

Remove the Replaceable class.

### Checklist


peterallenwebb:
@gshank I think the failing test may be due to new uses of replaceable (booooo) added since you branched.

peterallenwebb (approved):

jtcohen6:
@gshank @peterallenwebb Is this still something we're looking to merge?

Issue #7827, opened 2023-06-08
[CT-2671] Check Semantic Interface Protocol Implementations in Core


peterallenwebb:
We recently overhauled the Metric and SemanticModel objects in Core, and they are now intended to be concrete implementations of interfaces defined as protocols in dbt-semantic-interfaces. We need to put in place a mechanism to automatically verify that these objects (and their child objects) do in fact implement their respective protocols.

PR #7848, opened 2023-06-13
Update SemanticModel node to match DSI 0.1.0dev3 protocols

### LINKED ISSUE #7833: [CT-2677] [BUG] Update SemanticModel definitions to match DSI 0.1.0dev3 protocols
In #7769 we added SemanticModel nodes in core. We started this work while pointed at [DSI 0.1.0dev2](https://pypi.org/project/dbt-semantic-interfaces/0.1.0.dev2/) but midway updated to depend on [DSI 0.1.0dev3](https://pypi.org/project/dbt-semantic-interfaces/0.1.0.dev3/). The main problem is that there are some protocol changes from DSI dev2 to DSI dev3, specifically `SemanticModel.dimensions[x].is_primary` has moved to `SemanticModel.defaults.agg_time_dimension`. Due to our serialization process, the semantic manifest artifact contains neither `is_primary` or `agg_time_dimension` 🙃 This issue was able to slip through because we don't check currently in core that our node implementations satisfy the protocol. Thus in solving this issue we should not only correct the nodes,  but also add tests to verify that our nodes satisfy the DSI protocols.

QMalcolm:
resolves #7833, #7827 

### Description

In the midst of #7769 we moved from DSI `0.1.0dev2` to `0.1.0dev3`. This led to a situation where we were parsing and defining `SemanticModel` nodes in a `0.1.0dev2` shape instead of a `0.1.0dev3` shape (more detail can be found in #7833). This was possible because we weren't doing anything to ensure that semantic layer nodes we were adding satisfied the protocols we were targeting (the focus of #7827). This PR first adds testing to ensure our semantic layer nodes satisfy the protocols, and then goes through updating the nodes (and sub objects) to pass the tests (and therefore ensure satisfaction of the protocols).

### Checklist

### DIFF
--- a/core/dbt/contracts/graph/semantic_models.py
+++ b/core/dbt/contracts/graph/semantic_models.py
@@ -129,7 +129,7 @@ class Measure(dbtClassMixin):
     name: str
     agg: AggregationType
     description: Optional[str] = None
-    create_metric: Optional[bool] = False
+    create_metric: bool = False
     expr: Optional[str] = None
     agg_params: Optional[MeasureAggregationParameters] = None
     non_additive_dimension: Optional[NonAdditiveDimension] = None
--- a/core/dbt/contracts/graph/semantic_models.py
+++ b/core/dbt/contracts/graph/semantic_models.py
@@ -1,0 +1,71 @@
+from dataclasses import dataclass
+from dbt.dataclass_schema import dbtClassMixin
+from dbt_semantic_interfaces.references import DimensionReference, TimeDimensionReference
+from dbt_semantic_interfaces.type_enums.dimension_type import DimensionType
+from dbt_semantic_interfaces.type_enums.time_granularity import TimeGranularity
+from typing import Optional
+
+
+@dataclass
+class FileSlice(dbtClassMixin):
+    """Provides file slice level context about what something was created from.
+
+    Implementation of the dbt-semantic-interfaces `FileSlice` protocol
+    """
+
+    filename: str
+    content: str
+    start_line_number: int
+    end_line_number: int
+
+
+@dataclass
+class SourceFileMetadata(dbtClassMixin):
+    """Provides file context about what something was created from.
+
+    Implementation of the dbt-semantic-interfaces `Metadata` protocol
+    """
+
+    repo_file_path: str
+    file_slice: FileSlice
+
+
+@dataclass
+class DimensionValidityParams(dbtClassMixin):
+    is_start: bool = False
+    is_end: bool = False
+
+
+@dataclass
+class DimensionTypeParams(dbtClassMixin):
+    time_granularity: TimeGranularity
+    validity_params: Optional[DimensionValidityParams] = None
+
+
+@dataclass
+class Dimension(dbtClassMixin):
+    name: str
+    type: DimensionType
+    description: Optional[str] = None
+    is_partition: bool = False
+    type_params: Optional[DimensionTypeParams] = None
+    expr: Optional[str] = None
+    metadata: Optional[SourceFileMetadata] = None
+
+    @property
+    def reference(self) -> DimensionReference:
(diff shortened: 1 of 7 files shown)


QMalcolm:
This PR history "should" be clean enough that I recommend reviewing commit by commit. It's easier to see what's going on each step of the way 🙂

core/dbt/contracts/graph/semantic_models.py
```diff
@@ -0,0 +1,152 @@
+from dataclasses import dataclass
```
peterallenwebb, line 1: I am very happy about moving these out of nodes.py, but I wonder if we should move them out of the graph subdirectory?  I could see arguments either way. We don't need to decide now.

core/dbt/parser/schema_yaml_readers.py
```diff
@@ -408,6 +423,79 @@ def __init__(self, schema_parser: SchemaParser, yaml: YamlBlock):
         self.schema_parser = schema_parser
         self.yaml = yaml
 
+    def _get_dimension_type_params(
```
peterallenwebb, line 426: The code in these methods continues to be verbose and mechanical. We need a better way, though not today.

peterallenwebb (approved):
I think it makes sense to move forward with the changes as is. That said, we're already getting close to a point where it will make sense to focus on some of the mounting tech debt I've identified, even under schedule pressure.

tests/unit/test_semantic_layer_nodes_satisfy_protocols.py
```diff
@@ -0,0 +1,123 @@
...
+    pass
+
+
+@runtime_checkable
+class RuntimeCheckableMetric(DSIMetric, Protocol):
+    pass
+
+
+def test_semantic_model_node_satisfies_protocol():
+    test_semantic_model = SemanticModel(
```
peterallenwebb, line 49: It's awkward that we have to instantiate dummy classes to do this check. It makes the tests verbose and could be a minor maintenance issue. I think we could probably come up with a more compact and flexible way of doing these checks.

core/dbt/parser/schema_yaml_readers.py
QMalcolm: I'm not sure I wholly agree 🤔 There are fields which will require remapping / conversion. To you're point though, a lot of the fields are just direct mappings. We could do something like `params = unparsed.to_dict()`, edit the dict to spell out the things that do indeed need to be changed, and then `SemanticModel(**params)`. However, this can hide some problems. For instance if a "direct mapping" has a typo'd property in either the node or the unparsed version, the key value will get dropped. Or if someone makes a change to the node (renaming or adding properties), it isn't immediately obvious that anything needs to be done in the `schema_yaml_parser.py` or `unparsed.py` (whereas this current implementation will start complaining pretty quick). 

Another way is we can push the mapping logic into the nodes/subobjects themselves. We could add `parse` functions to the nodes/subobjects which take in their unparsed counter part and return the "parsed" version. This doesn't make it less mechnanical, but it does change where the logic is to perhaps somewhere more tasteful.

Yet another way would be what we do for the Pydantic implementations in DSI wherein we put things through the ManifestTransformer. However I'd advocated against this, as you lose some type guarantees as a lot more things become `Optional`.

I could be wrong, but I don't imagine this code being to high touch after 1.6. The current implementation makes it easy to see whats going on and raises discrepancies better than if we add magic, which I think are both important for a future where this code is infrequently touched.

tests/unit/test_semantic_layer_nodes_satisfy_protocols.py
QMalcolm: How would it be a maintenance issue? The main thing I can think of is if the DSI version is incompatible, then the tests would start failing. However, that's kind of what we want, right?

Issue #7799, opened 2023-06-06
[SPIKE] [CT-2650] Materialization macros should support dispatch


jtcohen6:
_Copied from https://github.com/dbt-labs/dbt-core/issues/4646_

Adapter-specific materializations work differently from adapter-specific macros. Namely:
- `{% materialization table, adapter = 'snowflake' %}` creates a macro named `materialization_table_snowflake`, rather than (as one would expect) `snowflake__materialization_table`
- "Parent" adapter inheritance is not supported for materializations, as it is for all other macros. This causes confusion for external adapter maintainers: https://github.com/dbt-msft/dbt-synapse/pull/74#issuecomment-997102690, https://github.com/databricks/dbt-databricks/pull/40#discussion_r795513529 

Let's make materialization macro generation / discovery work more like dispatch. We'd need to offer backwards compatibility / avoiding breaking changes for folks who currently rely on the `materialization_<name>_<adapter>` construction.

---

This is important in cases where we want to "call" a materialization macro from within another macro (perhaps from within another materialization). I found a concrete use case for doing this in my spiking on `dbt clone` (https://github.com/dbt-labs/dbt-core/pull/7258), whereby the `clone` materialization would in certain instances want to "shell out" to the `view` materialization.

Specific comment & spot in the spiking code where this is relevant: https://github.com/dbt-labs/dbt-core/pull/7258#discussion_r1155320591

Depending on the implementation we pursue for #7442, this could also be a relevant capability there.

gshank:
Materializations are a variant of Jinja macro (jinja extension) which is created in core/dbt/clients/jinja.py, in MaterializationExtension.  They are saved in the manifest.macros dictionary with names such as: "macro.dbt.materialization_incremental_default"

In order to use adapter.dispatch and use its macro resolution code we would need to normalize the names. (Could we have synonyms?)  (Note: we currently have two macro resolvers, MacroResolver and MacroNamespace.)

In the "def execute" method in dbt/task/run.py and dbt/task/clone.py we do manifest.get_materialization_macro_name. Instead we would have to execute adapter.dispatch to get the materialization macro, which would return a MacroGenerator object so it wouldn't need to be created. We don't currently call adapter.dispatch in Python, but the materialization macros are the "launcher" macros that get everything started so this might be an exception.

[...]

PR #7796, opened 2023-06-06
Fixes #7551: Create `add_from_artifact` to populate `state_relation` field of nodes

aranke:
resolves #7551 

### Checklist

- [x] I have read [the contributing guide](https://github.com/dbt-labs/dbt-core/blob/main/CONTRIBUTING.md) and understand what's expected of me
- [x] I have signed the [CLA](https://docs.getdbt.com/docs/contributor-license-agreements)
- [x] I have run this code in development and it appears to resolve the stated issue
- [x] This PR includes tests, or tests are not required/relevant for this PR
- [x] I have [opened an issue to add/update docs](https://github.com/dbt-labs/docs.getdbt.com/issues/new/choose), or docs changes are not required/relevant for this PR
- [x] I have run `changie new` to [create a changelog entry](https://github.com/dbt-labs/dbt-core/blob/main/CONTRIBUTING.md#adding-a-changelog-entry)

Issue #7432, opened 2023-04-21
[Feature] Infer contracts from prod, to enforce & detect breaking changes in dev


jtcohen6:
In v1.5, we added support for:

- Declaring model [contracts](https://docs.getdbt.com/reference/resource-configs/contract) in yaml (which can require a lot of yaml!)
- Enforcing that contract at build time, and raising a nice error message if there's a mismatch

What if we could manage to do that, *without* requiring all that yaml to be defined upfront? We infer the contract on the basis of the **current version of this model in production**. In that way, we also marry together the contract’s enforcement and its breaking change detection.

This would require a mechanism similar to what's proposed in https://github.com/dbt-labs/dbt-core/issues/7256 / https://github.com/dbt-labs/dbt-core/pull/7258. If provided with a manifest for deferral / state comparison, we should store a `state_relation` associated with the prod counterpart of each node, in the current manifest.

During the model’s materialization:

1. Dynamically access the columns of the “stateful” (prod) relation
2. Call the [`same_contract`](https://github.com/dbt-labs/dbt-core/blob/main/core/dbt/contracts/graph/nodes.py#L590) method to detect breaking changes — *while building the model*, instead of / in addition to while selecting from it — where the columns being passed in are actually the columns inferred from the database, rather than the ones defined in yaml

Pseudo logic for (1):

```sql
{% if state_relation and contract.inferred %}
    {% set dev_columns = get_column_schema_from_query(sql) %}
    {% set prod_columns = get_column_schema_from_query(get_empty_subquery_sql("select * from " ~ state_relation)) %}

    {% if not dev_columns | sorted == prod_columns | sorted %}
        {%- do exceptions.raise_contract_error(yaml_columns, sql_columns) -%}
    {% endif %}
{% endif %}
```

```yaml
models:
  - name: my_model_with_dynamic_contract
    contract:
      enforced: true
      inferred: true
```

### Considerations

We'd need some special handling when a model is being renamed, or created for the first time. If there is no `state_relation`, there is no contract to infer/enforce — it's whatever the model SQL says.

We would need to do more work to support *partial* contract definitions—a mix of explicit & inferred—whereby a user could define certain columns with constraints, dynamic data masking rules, etc, but wouldn’t have to define them all. (Since these require us to include the full column spec during `create table as`.)
- A lot of that work might be around type aliasing, which gets tricky quickly. Many database cursors return column types with a genuinely different naming convention from how that database supports receiving them.

[...]

PR #7881, opened 2023-06-15
`dbt clone`

### LINKED ISSUE #7256: [Feature] `dbt clone` command

aranke:
resolves #7256

### Checklist

### DIFF
--- a/tests/functional/semantic_models/test_semantic_model_parsing.py
+++ b/tests/functional/semantic_models/test_semantic_model_parsing.py
@@ -111,6 +111,7 @@ class TestSemanticModelParsing:
         semantic_model = manifest.semantic_models["semantic_model.test.revenue"]
         assert semantic_model.dimensions[0].type_params.time_granularity == TimeGranularity.WEEK
 
+    @pytest.mark.skip(reason="Tracking in https://github.com/dbt-labs/dbt-core/issues/7977")
     def test_semantic_model_deleted_partial_parsing(self, project):
         # First, use the default schema.yml to define our semantic model, and
         # run the dbt parse command


core/dbt/include/global_project/macros/adapters/clone.sql
```diff
@@ -0,0 +1,115 @@
...
+{% macro snowflake__get_clone_table_sql(this_relation, state_relation) %}
+    create or replace
+      {{ "transient" if config.get("transient", true) }}
+      table {{ this_relation }}
+      clone {{ state_relation }}
+      {{ "copy grants" if config.get("copy_grants", false) }}
+{% endmacro %}
+
+
+{%- materialization clone, default -%}
```
McKnight-42, line 48: does this need to be moved to `core/dbt/include/global_project/macros/materializations/models` we are creating a new matiaerlization

McKnight-42:
if we are technically creating a materialization do we need to add a new `is_type` check in`BaseRelation`?

tests/functional/defer_state/test_defer_state.py
```diff
@@ -345,3 +350,63 @@ def test_defer_state_flag(self, project, unique_schema, other_schema):
...
+    -- put seeds into a separate schema in "prod", to verify that cloning in "dev" still works
+    {%- elif target.name == 'default' and node.resource_type == 'seed' -%}
+        {{ return(default_schema ~ '_' ~ 'seeds') }}
+    {%- else -%}
+        {{ return(default_schema) }}
+    {%- endif -%}
+{%- endmacro %}
+"""
+
+
```
McKnight-42, line 369: To circle back to talk of test placement does creating two base classes ex. `BaseClonePossible` and `BaseCloneNotPossible` and define like the tests below to look for the correct types `view or table` so users can pull in one or both options to test? or potentially move both those to their own file to be the adapter zone tests? as I don't believe `defer_state` is pulled into all adapters

McKnight-42: will make some mock stuff for snowflake pr we can look at

core/dbt/include/global_project/macros/materializations/models/clone/clone.sql
```diff
@@ -0,0 +1,68 @@
...
+
+      -- reuse the view materialization
+      -- TODO: support actual dispatch for materialization macros
+      {% set search_name = "materialization_view_" ~ adapter.type() %}
+      {% if not search_name in context %}
+          {% set search_name = "materialization_view_default" %}
+      {% endif %}
+      {% set materialization_macro = context[search_name] %}
+      {% set relations = materialization_macro() %}
+      {{ return(relations) }}
```
jtcohen6, line 64: This is definitely the bit I felt worst about in https://github.com/dbt-labs/dbt-core/pull/7258.

I opened an issue to track real dispatch support for materialization macros: https://github.com/dbt-labs/dbt-core/issues/7799. Until we do that, this will work _much_ of the time, but it won't support adapters that inherit from other adapters. E.g. `dbt-postgres` &rarr; `dbt-redshift`, ergo Redshift is allowed to use `postgres__` implementations.

As far as hard-overwriting the context values for `sql` + `compiled_code`: Gross, right? The other potential approach I can think of is, making materialization macros accept `kwargs`, so that `sql`/`compiled_code` can be passed in explicitly rather than set + pulled out of the ~aether~ context.

core/dbt/task/clone.py
```diff
@@ -0,0 +1,183 @@
...
+                "--state is required for cloning relations from another environment"
+            )
+
+        if state.manifest is None:
+            raise DbtRuntimeError(f'Could not find manifest in --state path: "{self.args.state}"')
+        return state.manifest
+
+    # Note that this is different behavior from --defer with other commands, which *merge*
+    # selected nodes from this manifest + unselected nodes from the other manifest
+    def defer_to_manifest(self, adapter, selected_uids: AbstractSet[str]):
```
jtcohen6, line 173: It's compelling to want access to `state_relation` much more frequently. If we have it, we can do cool things with it; `clone` is just one of those cool things. This has come up in a handful of conversations over the past few weeks/months:
- Inferring model contracts/diffs based on the prod db object (#7432)
- Interactive preview + `audit_helper` to calculate diffs between prod/dev

I know that was split off as a separate bit of work, and it was already implemented (#7796). It might be worth reconsidering this as either:
- not mutually exclusive with `merge_from_artifact`
- the default behavior when `--defer` is set, rather than something totally unique and special to the `clone` command

My intent in this comment is to share context - no action needed right now.

core/dbt/context/providers.py
```diff
@@ -1432,6 +1432,20 @@ def this(self) -> Optional[RelationProxy]:
             return None
         return self.db_wrapper.Relation.create_from(self.config, self.model)
 
+    @contextproperty
+    def state_relation(self) -> Optional[RelationProxy]:
```
jtcohen6, line 1436: Is now the time for us to actually decide what this thing should be called? (Or was that back in https://github.com/dbt-labs/dbt-core/pull/7796?) `state_relation` is the least bad thing I could come up with at the time, but it's definitely a bit wonky.

Another question: Should this even have the same name as the node attribute, given that we're actually converting it to a real (adapter-informed) `Relation` object here?

jtcohen6 (commented):
It's happening !!

core/dbt/context/providers.py
ChenyuLInx: I would prefer a narrower scoped name, we can always expand the scope later on but it's hard to narrow it. Also making names slightly more verbose sometimes makes the code way easy to read, it took me a while to make the link of `state_relation` means a relation for current model in the provided state.

With that being said, I do not have a great suggestion on what it should be named. Throwing some potentially bad options here: `source_relation`, `provided_state_relation`.

tests/unit/test_context.py
```diff
@@ -230,7 +230,7 @@ def assert_has_keys(required_keys: Set[str], maybe_keys: Set[str], ctx: Dict[str
     "submit_python_job",
     "dbt_metadata_envs",
 }
-REQUIRED_MODEL_KEYS = REQUIRED_MACRO_KEYS | {"this", "compiled_code"}
+REQUIRED_MODEL_KEYS = REQUIRED_MACRO_KEYS | {"this", "compiled_code", "state_relation"}
```
ChenyuLInx, line 233: Should state_relation in required or maybe?

ChenyuLInx (commented):
Thanks for adding this patch!! Really glad it's happening! As this work is the foundation piece for adapters, I am hoping that we can think through what's strictly needed at this stage to make this function easy to maintain in the long run and make sure all of those are included in this patch.

core/dbt/context/providers.py
ChenyuLInx: I have strong preference against `state_relation` as it is too vague. What does `state` mean? You can even specify two previous state to dbt now.
`defer_state_relation` or something that gives reader of the code more hint would be my preference. Feel free to update it to any name that's not `state_relation`.

jtcohen6: I like the idea of including "defer" in the name, since `clone` should always be using the deferral state, rather than comparison state (if two are provided separately).

How about: `defer_relation` or `defer_to_relation`?

core/dbt/task/clone.py
jtcohen6: I totally agree with removing `state_relation` before writing the manifest. That's exactly how I (naively) implemented this in my spike PR: https://github.com/dbt-labs/dbt-core/pull/7258/files#r1155323650

core/dbt/include/global_project/macros/materializations/models/clone/clone.sql
jtcohen6: Good breakdown of the questions @ChenyuLInx!

Thinking through the logic we'd want from the `view` materialization:
- Handling the cases where the object already exists as a table, and needs to be dropped first
- Knowing whether this adapter wants to `create or replace view` versus create temp + alter/rename/swap
- Applying grants, persisting descriptions as comments - I think we still might actually want these things!

So - I would argue that's the right way to go. I'm definitely interested in a world where materializations are more composable than they are currently, so that more-complex build strategies ("clone an object from one env into another") can call into simpler ones ("make a view").

Macros are already pretty composable. Materializations are just macros, but they pull all their inputs from this node's g-d context, rather than receiving them as explicit arguments.

core/dbt/task/clone.py
```diff
@@ -0,0 +1,183 @@
...
+
+    def get_runner_type(self, _):
+        return CloneRunner
+
+    def _get_deferred_manifest(self) -> Optional[WritableManifest]:
+        state = self.previous_state
+        if state is None:
+            raise DbtRuntimeError(
+                "--state is required for cloning relations from another environment"
+            )
```
jtcohen6, line 165: This should support `--defer-state` in addition to / instead of `--state`, right?

core/dbt/include/global_project/macros/materializations/models/clone/clone.sql
```diff
@@ -0,0 +1,68 @@
...
+      {% do persist_docs(target_relation, model) %}
+
+      {{ return({'relations': [target_relation]}) }}
+
+  {%- else -%}
+
+      {%- set target_relation = this.incorporate(type='view') -%}
+
+      -- TODO: this should probably be illegal
+      -- I'm just doing it out of convenience to reuse the 'view' materialization logic
```
jtcohen6, line 50: If this is the approach we're going to go with, we should remove the TODO :)

core/dbt/task/clone.py
```diff
@@ -0,0 +1,183 @@
...
+
+        for relation in self._materialization_relations(result, model):
+            self.adapter.cache_added(relation.incorporate(dbt_created=True))
+
+        return self._build_run_model_result(model, context)
+
+
+class CloneTask(GraphRunnableTask):
+    def raise_on_first_error(self):
+        return False
```
jtcohen6, line 97: One thing I noticed in my previous spike, and I'm seeing in this implementation as well: Even though each node's `clone` operation is completely independent of any other node, we are still running `clone` in DAG order. That significantly limits how fast this can be with `--threads 1000`.

I'm not sure how to fix that, and it will require digging into our base task / graph queue logic — but I suspect that will be key to unlocking a lot of the value here.

Repro
```
-- models/model_a.sql
select 1 as id

-- models/model_b.sql
```
```
$ dbt run --target prod
$ mkdir state
$ dbt clone --target dev --state state --full-refresh
```
From the logs: Notice that `model_b` runs in serial after `model_a`, even though they're on different threads. dbt understands that `model_b` depends on `model_a`, even though for the purposes of the `clone` task, it really doesn't.
```
============================== 11:36:51.356566 | 2fab3bf3-3220-47f5-a842-a8e7aee780d1 ==============================
[0m11:36:51.356566 [info ] [MainThread]: Running with dbt=1.6.0-b4
[0m11:36:51.360523 [debug] [MainThread]: running dbt with arguments {'printer_width': '80', 'indirect_selection': 'eager', 'write_json': 'True', 'log_cache_events': 'False', 'partial_parse': 'True', 'cache_selected_only': 'False', 'warn_error': 'None', 'version_check': 'False', 'debug': 'False', 'log_path': '/Users/jerco/dev/scratch/testy/logs', 'fail_fast': 'False', 'profiles_dir': '/Users/jerco/.dbt', 'use_colors': 'True', 'use_experimental_parser': 'False', 'no_print': 'None', 'quiet': 'False', 'warn_error_options': 'WarnErrorOptions(include=[], exclude=[])', 'static_parser': 'True', 'log_format': 'default', 'introspect': 'True', 'target_path': 'None', 'send_anonymous_usage_stats': 'False'}
[0m11:36:51.616469 [info ] [MainThread]: Registered adapter: snowflake=1.6.0-b3
[0m11:36:51.632107 [debug] [MainThread]: checksum: 85def1bdb9aa2c18a6a11d52b89108c001273905a0f5fd72c238caeba233ef8d, vars: {}, profile: , target: dev, version: 1.6.0b4
[0m11:36:51.669565 [debug] [MainThread]: Partial parsing enabled: 0 files deleted, 0 files added, 0 files changed.
[0m11:36:51.669910 [debug] [MainThread]: Partial parsing enabled, no changes found, skipping parsing
[0m11:36:51.670469 [debug] [MainThread]: Publication artifact available
[0m11:36:51.692800 [info ] [MainThread]: Found 2 models, 0 tests, 0 snapshots, 0 analyses, 371 macros, 0 operations, 0 seed files, 0 sources, 0 exposures, 0 metrics, 0 groups
[0m11:36:51.694034 [info ] [MainThread]: 
[0m11:36:51.694695 [debug] [MainThread]: Acquiring new snowflake connection 'master'
[0m11:36:51.700555 [debug] [ThreadPool]: Acquiring new snowflake connection 'list_analytics'
[0m11:36:51.712206 [debug] [ThreadPool]: Using snowflake connection "list_analytics"
[0m11:36:51.712583 [debug] [ThreadPool]: On list_analytics: /* {"app": "dbt", "dbt_version": "1.6.0b4", "profile_name": "sandbox-snowflake", "target_name": "dev", "connection_name": "list_analytics"} */
show terse schemas in database analytics
    limit 10000
[0m11:36:51.712825 [debug] [ThreadPool]: Opening a new connection, currently in state init
[0m11:36:52.957601 [debug] [ThreadPool]: SQL status: SUCCESS 119 in 1.0 seconds
[0m11:36:53.011881 [debug] [ThreadPool]: On list_analytics: Close
[0m11:36:53.372661 [debug] [ThreadPool]: Re-using an available connection from the pool (formerly list_analytics, now list_analytics_dbt_jcohen_dev)
[0m11:36:53.381793 [debug] [ThreadPool]: Using snowflake connection "list_analytics_dbt_jcohen_dev"
[0m11:36:53.382231 [debug] [ThreadPool]: On list_analytics_dbt_jcohen_dev: /* {"app": "dbt", "dbt_version": "1.6.0b4", "profile_name": "sandbox-snowflake", "target_name": "dev", "connection_name": "list_analytics_dbt_jcohen_dev"} */
show terse objects in analytics.dbt_jcohen_dev limit 10000
[0m11:36:53.382470 [debug] [ThreadPool]: Opening a new connection, currently in state closed
[0m11:36:54.382890 [debug] [ThreadPool]: SQL status: SUCCESS 2 in 1.0 seconds
[0m11:36:54.387887 [debug] [ThreadPool]: On list_analytics_dbt_jcohen_dev: Close
[0m11:36:54.748809 [info ] [MainThread]: Concurrency: 8 threads (target='dev')
[0m11:36:54.749807 [info ] [MainThread]: 
[0m11:36:54.755801 [debug] [Thread-1 (]: Began running node model.my_dbt_project.model_a
[0m11:36:54.756728 [debug] [Thread-1 (]: Re-using an available connection from the pool (formerly list_analytics_dbt_jcohen_dev, now model.my_dbt_project.model_a)
[0m11:36:54.757151 [debug] [Thread-1 (]: Began compiling node model.my_dbt_project.model_a
[0m11:36:54.757645 [debug] [Thread-1 (]: Timing info for model.my_dbt_project.model_a (compile): 11:36:54.757426 => 11:36:54.757430
[0m11:36:54.758014 [debug] [Thread-1 (]: Began executing node model.my_dbt_project.model_a
[0m11:36:54.775804 [debug] [Thread-1 (]: On "model.my_dbt_project.model_a": cache miss for schema "analytics.dbt_jcohen_prod", this is inefficient
[0m11:36:54.778532 [debug] [Thread-1 (]: Using snowflake connection "model.my_dbt_project.model_a"
[0m11:36:54.778925 [debug] [Thread-1 (]: On model.my_dbt_project.model_a: /* {"app": "dbt", "dbt_version": "1.6.0b4", "profile_name": "sandbox-snowflake", "target_name": "dev", "node_id": "model.my_dbt_project.model_a"} */
show terse objects in analytics.dbt_jcohen_prod limit 10000
[0m11:36:54.779255 [debug] [Thread-1 (]: Opening a new connection, currently in state closed
[0m11:36:55.787694 [debug] [Thread-1 (]: SQL status: SUCCESS 5 in 1.0 seconds
[0m11:36:55.793394 [debug] [Thread-1 (]: While listing relations in database=analytics, schema=dbt_jcohen_prod, found: ANOTHER_MODEL, MODEL_A, MODEL_B, MY_MODEL, MY_SEED
[0m11:36:55.842689 [debug] [Thread-1 (]: Writing runtime sql for node "model.my_dbt_project.model_a"
[0m11:36:55.845516 [debug] [Thread-1 (]: Using snowflake connection "model.my_dbt_project.model_a"
[0m11:36:55.845921 [debug] [Thread-1 (]: On model.my_dbt_project.model_a: /* {"app": "dbt", "dbt_version": "1.6.0b4", "profile_name": "sandbox-snowflake", "target_name": "dev", "node_id": "model.my_dbt_project.model_a"} */
create or replace   view analytics.dbt_jcohen_dev.model_a
  
   as (
    
        select * from analytics.dbt_jcohen_prod.model_a
    
  );
[0m11:36:56.184105 [debug] [Thread-1 (]: SQL status: SUCCESS 1 in 0.0 seconds
[0m11:36:56.215770 [debug] [Thread-1 (]: Timing info for model.my_dbt_project.model_a (execute): 11:36:54.758266 => 11:36:56.215606
[0m11:36:56.216137 [debug] [Thread-1 (]: On model.my_dbt_project.model_a: Close
[0m11:36:56.587539 [debug] [Thread-1 (]: Finished running node model.my_dbt_project.model_a
[0m11:36:56.589991 [debug] [Thread-3 (]: Began running node model.my_dbt_project.model_b
[0m11:36:56.591451 [debug] [Thread-3 (]: Acquiring new snowflake connection 'model.my_dbt_project.model_b'
[0m11:36:56.592146 [debug] [Thread-3 (]: Began compiling node model.my_dbt_project.model_b
[0m11:36:56.592824 [debug] [Thread-3 (]: Timing info for model.my_dbt_project.model_b (compile): 11:36:56.592575 => 11:36:56.592579
[0m11:36:56.593252 [debug] [Thread-3 (]: Began executing node model.my_dbt_project.model_b
[0m11:36:56.600984 [debug] [Thread-3 (]: Writing runtime sql for node "model.my_dbt_project.model_b"
[0m11:36:56.602917 [debug] [Thread-3 (]: Using snowflake connection "model.my_dbt_project.model_b"
[0m11:36:56.603500 [debug] [Thread-3 (]: On model.my_dbt_project.model_b: /* {"app": "dbt", "dbt_version": "1.6.0b4", "profile_name": "sandbox-snowflake", "target_name": "dev", "node_id": "model.my_dbt_project.model_b"} */
create or replace   view analytics.dbt_jcohen_dev.model_b
  
   as (
    
        select * from analytics.dbt_jcohen_prod.model_b
    
  );
[0m11:36:56.603966 [debug] [Thread-3 (]: Opening a new connection, currently in state init
[0m11:36:57.722641 [debug] [Thread-3 (]: SQL status: SUCCESS 1 in 1.0 seconds
[0m11:36:57.728537 [debug] [Thread-3 (]: Timing info for model.my_dbt_project.model_b (execute): 11:36:56.593543 => 11:36:57.728157
[0m11:36:57.729557 [debug] [Thread-3 (]: On model.my_dbt_project.model_b: Close
[0m11:36:58.086873 [debug] [Thread-3 (]: Finished running node model.my_dbt_project.model_b
[0m11:36:58.089987 [debug] [MainThread]: Connection 'master' was properly closed.
[0m11:36:58.090496 [debug] [MainThread]: Connection 'model.my_dbt_project.model_a' was properly closed.
[0m11:36:58.090840 [debug] [MainThread]: Connection 'model.my_dbt_project.model_b' was properly closed.
[0m11:36:58.091862 [debug] [MainThread]: Command end result
[0m11:36:58.108549 [info ] [MainThread]: 
[0m11:36:58.109036 [info ] [MainThread]: [32mCompleted successfully[0m
[0m11:36:58.109432 [info ] [MainThread]: 
[0m11:36:58.109785 [info ] [MainThread]: Done. PASS=2 WARN=0 ERROR=0 SKIP=0 TOTAL=2
[0m11:36:58.110456 [debug] [MainThread]: Command `dbt clone` succeeded at 11:36:58.110325 after 6.42 seconds
[0m11:36:58.110747 [debug] [MainThread]: Flushing usage events
```

jtcohen6 (commented):
IMO A prerequisite to merging this is having a working implementation in one of the adapters that actually supports `create table ... clone` (Snowflake, BigQuery, or Databricks), to make sure we have the right constructs in place.

core/dbt/task/clone.py
jtcohen6: Meta comment: We can open this one as a separate follow-on issue, rather than a blocker to merging the current PR. But it does sound like you & I are thinking about this differently!

---

> If model A doesn't exist in the warehouse source, should we even bother trying to clone model B?

I totally agree with this when it comes to _building_ model B, based on its SQL definition. The act of _cloning_ model B doesn't actually depend on model A, though.

> This assumption also breaks the user's mental model that each command in dbt runs in DAG order, except for clone.

IIRC - `compile` and `docs generate` don't run in DAG order. (They also interrupt the entire command on the first failure encountered during compilation.) I think the inconsistency is justifiable, if the parallelism delivers significantly better performance.

> Supporting `retry` for `clone` also becomes much more tricky here.

We wouldn't `SKIP` any models while cloning them, but failure to clone should still be recorded as an `ERROR`. The `retry` command should pick up those `ERROR` statuses in the subsequent invocation. Does that sound right?

> Select wrote a great blog post about this, and most of the gains were from using clone and not from parallelism: https://select.dev/posts/snowflake-clones.

My read of that article: It's **all** about the parallelism! Snowflake doesn't use parallelism when cloning databases/schemas, so submitting a ton of table-level `clone` commands asynchronously / in parallel gets the whole thing done much faster. Crucially: "The bottleneck is still the rate at which queries can be dispatched."

> If people want a 100% solution, they can always write a macro that will far outperform what we can build.

I don't think they can! Parallelism / async queries aren't available to the macro context. For this reason, users have been completely blocked from taking advantage of the performance optimizations recommended in that Select.Dev article within dbt. Lots of background on this in https://github.com/dbt-labs/dbt-snowflake/issues/315.

---

We can collect some real-world anecdata on what the performance difference looks like; my suspicion is that it will be meaningful. The fastest `clone` of an entire DAG will be `[time to clone one model] * [longest left-to-right dependency chain]`, rather than `[time to clone one model] * [number of models] / [number of concurrent queries Snowflake can handle]`.

core/dbt/include/global_project/macros/materializations/models/clone/clone.sql
```diff
@@ -0,0 +1,66 @@
...
+      {% do apply_grants(target_relation, grant_config, should_revoke=should_revoke) %}
+      {% do persist_docs(target_relation, model) %}
+
+      {{ return({'relations': [target_relation]}) }}
+
+  {%- else -%}
+
+      {%- set target_relation = this.incorporate(type='view') -%}
+
+      {%- do context.update({
```
ChenyuLInx, line 49: The way that this update works is to overwrite the jinja context so that when we execute `materialization_macro` later on will be running with the updated context to actually create the view.
This logic is pretty hidden, can we spend a bit more time thinking about how to avoid changing context in macro space? @gshank  Any suggestion on whether it is possible to do this in a better way?

I did a quick search in our codebase and it seems that we are creating a new pattern here. If this is the only way we can do it but **we should add comment and call out this is not a pattern we would want to do** in the codebase.

core/dbt/include/global_project/macros/materializations/models/clone/clone.sql
```diff
@@ -0,0 +1,66 @@
...
+          'sql': get_clone_target(defer_relation),
+          'compiled_code': get_clone_target(defer_relation)
+      }) -%}
+
+      -- reuse the view materialization
+      -- TODO: support actual dispatch for materialization macros
+      {% set search_name = "materialization_view_" ~ adapter.type() %}
```
ChenyuLInx, line 56: Moving an existing comment over:

> What happens if we try to run this code with clone on Redshift today?

> In theory, it should do the same thing as Postgres (create views) since neither support zero-copy cloning.I'll have to verify what happens in practice.

Is this verified?

I understand this will also be revisited in the spike about materialization dispatch, can we make sure we link that issue in the comment here and link this line in that issue?

ChenyuLInx (approved):
Thanks for the iteration on this and getting the adapter test setup!! Great stuff.

We should see @gshank 's thoughts on updating context before this is merged, but I also understand there might not be a better way to do it.

gshank:
In a general way, updating the context is a very bad idea. The jinja documentation says all over the place that the context is immutable (though it does also store locally defined variables, temporarily), and I suspect that the only reason updating it works at all is because we implemented our own context object. It's also sets a bad example for users -- we don't want them changing things in the context either.  I'm not all that familiar with the code execution path here -- I'll look at it some more and discuss with Kshitij later this afternoon.

gshank:
I think I'd be happier with a specific context method to set something (the cloned/alternate model? just the compiled_code?) and then modify the sql and compiled_code methods in the provider context to return the compiled_code from the cloned model if it exists, otherwise return the usual compiled_code.

core/dbt/context/providers.py
```diff
@@ -1367,20 +1367,26 @@ def post_hooks(self) -> List[Dict[str, Any]]:
...
         # only doing this in sql model for backward compatible
-        if (
-            getattr(self.model, "extra_ctes_injected", None)
-            and self.model.language == ModelLanguage.sql  # type: ignore[union-attr]
-        ):
-            # TODO CT-211
-            return self.model.compiled_code  # type: ignore[union-attr]
-        return None
+        if self.model.language == ModelLanguage.sql:  # type: ignore[union-attr]
+            if getattr(self.model, "defer_relation", None):
```
ChenyuLInx, line 1371: Can we leave some comment here about this logic for future us? You can maybe find the clue in git history but they can also be tricky to find. So I prefer direct comment

ChenyuLInx:
Great!!!!

> we need to modify the clone target directory to be different from the previous directory.

My understanding is that having a separate state directory is all you need? Do we have to have `--target snowclone --target-path target_clone` specified? Am I missing something?

core/dbt/task/clone.py
jtcohen6: Thinking more about this: Do we actually need separate methods for `merge_from_artifact` and `add_from_artifact`? Or could it simply be that:
- for nodes which are "being deferred," we completely clobber/replace this manifest's version with the other manifest's version
- for nodes which _aren't_ "being deferred" (because they're being selected, or already exist in this schema), we keep the existing node entry and add the `defer_relation` attribute

Then, all tasks could call this same method. We don't need the divergent behavior between `clone` and `run`, and we set ourselves up more nicely for future work around contract inference, dev/prod diff, ...

Pseudo code:
```python
    def merge_from_artifact(
        self,
        adapter,
        other: "WritableManifest",
        selected: AbstractSet[UniqueID],
        favor_state: bool = False,
    ) -> None:
        """Given the selected unique IDs and a writable manifest, update this
        manifest by replacing any unselected nonexisting nodes with their counterpart.
        For any selected/existing nodes, add information about each node's location
        in the other manifest.

        Only non-ephemeral refable nodes are examined.
        """
        refables = set(NodeType.refable())
        merged = set()
        for unique_id, node in other.nodes.items():
            current = self.nodes.get(unique_id)
            
            # do the replace thing
            if current and (
                node.resource_type in refables
                and not node.is_ephemeral
                and unique_id not in selected
                and (
                    not adapter.get_relation(current.database, current.schema, current.identifier)
                    or favor_state
                )
            ):
                merged.add(unique_id)
                self.nodes[unique_id] = node.replace(deferred=True)
           
           # do the additive thing
           elif current and (node.resource_type in refables and not node.is_ephemeral):
                defer_state_relation = RelationalNode(node.database, node.schema, node.alias)
                self.nodes[unique_id] = current.replace(defer_relation=defer_relation)

        # Rebuild the flat_graph, which powers the 'graph' context variable,
        # now that we've deferred some nodes
        self.build_flat_graph()

        # log up to 5 items
        sample = list(islice(merged, 5))
        fire_event(MergedFromState(num_merged=len(merged), sample=sample))
```

The complication I could foresee here: Traditional defer behavior requires the population of the adapter cache (for the "nonexisting" part). I had suggested [doing things differently in `clone`'s `before_run` setup](https://github.com/dbt-labs/dbt-core/blob/4bce05fb0c991d239a9d9d8a5a392dee5d3502d2/core/dbt/task/clone.py#L118-L126) so that we defer _first_, and then cache, so that we can also cache the ["other" schemas (indicated by `defer_relation`)](https://github.com/dbt-labs/dbt-core/blob/4bce05fb0c991d239a9d9d8a5a392dee5d3502d2/core/dbt/task/clone.py#L109-L114). I'm not sure if that's actually necessary, or if it's extra complication that isn't really worth it.

jtcohen6 (commented):
> I've also removed the ugly `context.update` hack (thanks @gshank @ChenyuLInx).

Good riddance! Thanks for making it happen, sorry I ever introduced it into this world :)

McKnight-42:
@aranke only questions I currently have will be around possible ways of universally adding the clean_up methods and some possible stuff around spark implementation. this pr looks good on my end

core/dbt/context/providers.py
```diff
@@ -1367,20 +1367,28 @@ def post_hooks(self) -> List[Dict[str, Any]]:
...
+            else:
+                return None
+        else:
+            return None
 
     @contextproperty
     def compiled_code(self) -> Optional[str]:
-        if getattr(self.model, "extra_ctes_injected", None):
+        if getattr(self.model, "defer_relation", None):
+            return f"select * from {self.model.defer_relation.database}.{self.model.defer_relation.schema}.{self.model.defer_relation.alias}"  # type: ignore[union-attr]
```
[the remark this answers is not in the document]
jtcohen6: Yup, that's better!

gshank (approved):

sungchun12:
Love this

PR #7895, opened 2023-06-16
Resolve SemanticModel refs in the same way as other refs

### LINKED ISSUE #7822: [CT-2668] Improve ref() Parsing for SemanticModel Objects
As observed during review of our SemanticModels work, there are a couple important improvements which could be made to the way we are parsing the ref() expressions in SemanticModel.

- We should catch internal parsing exceptions and [raise a more informative one](https://github.com/dbt-labs/dbt-core/pull/7769#discussion_r1221993547).
- We should raise an exception if the ref [does not resolve to a model](https://github.com/dbt-labs/dbt-core/pull/7769#discussion_r1221997254).
- We should consider resolving refs at parse time, rather than deferring until the entire manifest is (almost) ready.
- We should check whether the ref'd model is disabled, and raise a specific exception if it is not.

gshank:
resolves #7822

### Description

Refactor SemanticModel ref resolution to work in the same way as other nodes. This includes adding several fields on the SamnticModel: created_at (for partial parsing), refs (for holding intermediate representation of the ref), and depends_on for holding the resolved unique_id of the model that's ref'd (also necessary for partial parsing.  We do a "get_rendered" on the model field to generate the "ref", calling common code to process the ref (_process_refs), and setting the fields in the NodeRelation class.

I also added relation_name to the NodeRelation class, because without something like that you have to call adapter code to get a properly quoted relation name. Let me know if there's some reason that's not necessary.

### Checklist

### DIFF
--- a/.changes/unreleased/Under the Hood-20230616-155749.yaml
+++ b/.changes/unreleased/Under the Hood-20230616-155749.yaml
@@ -1,0 +1,6 @@
+kind: Under the Hood
+body: Resolve SemanticModel ref is the same way as other refs
+time: 2023-06-16T15:57:49.245262-04:00
+custom:
+  Author: gshank
+  Issue: "7822"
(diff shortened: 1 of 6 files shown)


peterallenwebb (approved):
Looks good, and like a nice reduction in repetition!

PR #7910, opened 2023-06-20
Do not jinja render packages from dependencies yml

### LINKED ISSUE #7905: [CT-2717] Disallow conditional configuration (Jinja-in-yaml rendering) within `dependencies.yml`
_Follow-up to https://github.com/dbt-labs/dbt-core/pull/7857_

As described in https://github.com/dbt-labs/dbt-core/issues/7631#issuecomment-1564965453, we do not want to allow conditional configuration of `projects` and `packages` within `dependencies.yml` (i.e. Jinja-in-yaml rendering). I foresee this making it much trickier to parse this configuration outside of `dbt-core`, and it significantly expands the complexity of use cases we'll need to support in the future.

To avoid breaking changes for existing users & use cases, we should continue to support conditional configuration of `packages` if (and onlyif) defined in the old file name, `packages.yml`.

gshank:
resolves #7905

### Description

We don't want to render packages from the dependencies.yml file.

### Checklist

### DIFF
--- a/core/dbt/config/project.py
+++ b/core/dbt/config/project.py
@@ -283,6 +283,7 @@ class RenderComponents:
 
 @dataclass
 class PartialProject(RenderComponents):
+    # This class includes the project_dict, packages_dict, selectors_dict, etc from RenderComponents
     profile_name: Optional[str] = field(
         metadata=dict(description="The unrendered profile name in the project, if set")
     )
@@ -324,7 +325,7 @@ class PartialProject(RenderComponents):
             selectors_dict=rendered_selectors,
         )
 
-    # Called by 'collect_parts' in RuntimeConfig
+    # Called by Project.from_project_root (not PartialProject.from_project_root!)
     def render(self, renderer: DbtProjectYamlRenderer) -> "Project":
         try:
             rendered = self.get_rendered(renderer)
@@ -538,6 +539,7 @@ class PartialProject(RenderComponents):
         project_name = project_dict.get("name")
         profile_name = project_dict.get("profile")
 
+        # Create a PartialProject
         return cls(
             profile_name=profile_name,
             project_name=project_name,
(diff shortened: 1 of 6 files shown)


core/dbt/config/project.py
```diff
@@ -116,6 +116,7 @@ def package_and_project_data_from_root(project_root):
     dependent_projects_dict = {}
     if "packages" in dependencies_yml_dict:
         packages_dict["packages"] = dependencies_yml_dict["packages"]
+        packages_dict["packages_from_dependencies"] = True
```
jtcohen6, line 119: Not related to the change in this PR, as much as the changes previously made in https://github.com/dbt-labs/dbt-core/pull/7857:

This will now start raising an error if any `packages.yml` is an empty file, whether defined in the project root or in an installed package. (For example, the [`dbt-date` package](https://github.com/calogica/dbt-date/blob/main/packages.yml).) That's not exactly supported behavior, but it didn't used to raise an error, and now it does.

```yml
packages:
  - package: calogica/dbt_date
    version: 0.7.2
```
```
$ dbt deps && dbt parse
...
  File "/Users/jerco/dev/product/dbt-core/core/dbt/config/project.py", line 108, in package_and_project_data_from_root
    if "packages" in packages_yml_dict and "packages" in dependencies_yml_dict:
TypeError: argument of type 'NoneType' is not iterable
```

Quick fix on lines 103-106 above:
```python
    if path_exists(package_filepath):
        packages_yml_dict = _load_yaml(package_filepath) or {}
    if path_exists(dependencies_filepath):
        dependencies_yml_dict = _load_yaml(dependencies_filepath) or {}
```

gshank: Good catch. I've added a test and made this change.

peterallenwebb (approved):

jaklan:
@gshank if Jinja is not rendered in `dependencies.yml`, what's the expected workflow for specifying different `revision`s for different environments? E.g. on DEV we would like to refer to packages from `develop` branch, on PRD - from `master` branch.

With `packages.yml` - afaik we can use `env_var()`. I know it's still supported, but I can imagine it's gonna be deprecated one day when you resolve issues with i.a. private packages, but the above question would be still valid then. In theory we can maintain multiple `dependencies.yml` files, e.g. `dev-dependencies.yml`, `prd-dependencies.yml`, and switch between them with `cp`, but I believe it wouldn't work in dbt Cloud anyway.

PR #7928, opened 2023-06-22
Fix query comment tests

### LINKED ISSUE #7845: [CT-2686] Incorrectly set up tests in test_query_comment.py

### Current Behavior

Tests in [test_query_comment.py](https://github.com/dbt-labs/dbt-core/blob/v1.6.0b3/tests/adapter/dbt/tests/adapter/query_comment/test_query_comment.py) are incorrectly set up. They are always passing, no matter if query comments are correctly implemented.

Assertions are placed in functions named `matches_comment`. As pytest run functions with names prefixed with `test`, these assertions are never checked.

The only thing which this tests are doing, is invoking [test_comments](https://github.com/dbt-labs/dbt-core/blob/v1.6.0b3/tests/adapter/dbt/tests/adapter/query_comment/test_query_comment.py#LL35C9-L35C22) function, which invokes [run_assert_comments](https://github.com/dbt-labs/dbt-core/blob/v1.6.0b3/tests/adapter/dbt/tests/adapter/query_comment/test_query_comment.py#LL31C9-L31C28), which invokes  [run_get_json](https://github.com/dbt-labs/dbt-core/blob/v1.6.0b3/tests/adapter/dbt/tests/adapter/query_comment/test_query_comment.py#LL22C9-L22C21), which is capturing logs. There are no assertions in any of above functions.

Additionally, I think that it would be better to check in dbt_run result query if query-comments were added, instead of checking in logs.

### Expected Behavior

Tests should detect bugs instead of always pass.

### Steps To Reproduce

1. In [test_query_comment.py](https://github.com/dbt-labs/dbt-core/blob/v1.6.0b3/tests/adapter/dbt/tests/adapter/query_comment/test_query_comment.py) Change any assertion from any `matches_comment`, to wrong assertion e.g. `assert 0 == 1`. Run tests, notice that they are still passing.
2. Add some wrong assertion in [test_comments](https://github.com/dbt-labs/dbt-core/blob/v1.6.0b3/tests/adapter/dbt/tests/adapter/query_comment/test_query_comment.py#L35) function. Run tests, notice that all tests from this file are failing

### Relevant log output

_No response_

### Environment

```markdown
- OS: macOS
- Python: 3.10
- dbt: v1.6.0b3
```

postgres

### Additional Context

I'm willing to create PR for this issue, as I already overwrote these tests in [dbt-trino](https://github.com/starburstdata/dbt-trino) to test `query-comments` properly on this adapter.

damian3031:
resolves #7845 

### Description

Fixed query-comment tests, so they are actually checking this functionality, instead of always pass.

CC: @dbeatty10 @dataders

### DIFF
--- a/.changes/unreleased/Fixes-20230623-092933.yaml
+++ b/.changes/unreleased/Fixes-20230623-092933.yaml
@@ -1,0 +1,6 @@
+kind: Fixes
+body: Fix query comment tests
+time: 2023-06-23T09:29:33.225649+02:00
+custom:
+  Author: damian3031
+  Issue: "7845"
--- a/tests/adapter/dbt/tests/adapter/query_comment/test_query_comment.py
+++ b/tests/adapter/dbt/tests/adapter/query_comment/test_query_comment.py
@@ -28,13 +28,6 @@ class BaseDefaultQueryComments:
         assert len(res) > 0
         return raw_logs
 
-    def run_assert_comments(self):
-        logs = self.run_get_json()
-        return logs
-
-    def test_comments(self, project):
-        self.run_assert_comments()
-
 
 # Base setup to be inherited #
 class BaseQueryComments(BaseDefaultQueryComments):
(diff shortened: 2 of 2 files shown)


dataders (approved):
thanks @damian3031!!!

dbeatty10:
I also double-checked that these tests cases (actually!) ran in CI and passed:

image

Issue #5301, opened 2022-05-26
[CT-691] why isn't `postgres_get_relations` double-underscored?


dataders:
for dbt-greenplum, @markporoshin [wants to inherit this macro](https://github.com/markporoshin/dbt-greenplum/pull/6#discussion_r882158493) without copying it. is there a reason it isn't double-underscored? Currently, the workaround is to modify this line in `impl.py`

https://github.com/dbt-labs/dbt-core/blob/10b27b963392cce33950d9cef31c7da5c3e65592/plugins/postgres/dbt/adapters/postgres/impl.py#L15-L16

to be this:

```
GET_RELATIONS_MACRO_NAME = "greenplum_get_relations"
```

but I don't know why!

https://github.com/dbt-labs/dbt-core/blob/10b27b963392cce33950d9cef31c7da5c3e65592/plugins/postgres/dbt/include/postgres/macros/relations.sql#L1

dbeatty10:
@alzaar thanks for trying to find a `good_first_issue` to make your first contributions to this repo.

@dataders is this issue still relevant? If so, then @alzaar would like to try resolving it.

alzaar:
Thanks @dbeatty10 In terms of any pre-reqs, is there anything I need to know before I get started (from the dbt docs) ? I would also like to know how can I replicate the usage locally.

dbeatty10:
@alzaar you can set up your machine for local development using our [contributing](https://github.com/dbt-labs/dbt-core/blob/main/CONTRIBUTING.md) instructions (which include a local Postgres database using Docker).

Then you can create a local dbt project via [`dbt init`](https://docs.getdbt.com/reference/commands/init) or some other means. Run `dbt debug` to make sure the connection is working.

Create this file:
`analyses/test_macro.sql`
```
{% set result = postgres_get_relations() %}

{% if execute %}
{% do result.print_table() %}
{% endif %}
```

Run it and see the results:
```shell
dbt compile -s test_macro
```

After you've done your implementation, then you should be able to change `postgres_get_relations` within `analyses/test_macro.sql` to either be `get_relations` or `postgres__get_relations` and it should work.

dbeatty10:
@alzaar if you've never run a dbt project before, then [this repo](https://github.com/dbt-labs/jaffle_shop_duckdb) is a quick way to get a first experience.

PR #7944, opened 2023-06-25
Fixed double underline

lllong33:
resolves https://github.com/dbt-labs/dbt-core/issues/5301

### Checklist

image


mikealfare:
This macro is no longer used in `dbt-redshift` (in favor of a redshift-specific one). So we don't need to make any updates there. I agree with @jtcohen6's concern in the attached issue, that we need to maintain the name and point to the updated macro for backwards compatibility, so I'd like to see that in this PR as well.

lllong33:
@mikealfare thx for review, are the examples of backwards compatibility?

lllong33:
image

mikealfare (changes requested):

mikealfare:
Thanks for the updates @lllong33. I had a chance to read through this more thoroughly and provided some guidance. What we're effectively doing here is generalizing an adapter-specific macro. There are implementations of this in other adapters, so that makes sense. As a summary, here's the general pattern we use for doing such a thing:
```sql
-- in dbt-core
{% macro my_awesome_macro(arg1, arg2, ...) %}
    -- implement anything we want to happen all of the time, such as logging (we don't do that in this case)
    -- dispatch to the adapter-specific version
{% endmacro %}

{% macro default__my_awesome_macro(arg1, arg2, ...) %}
    -- provide a default in case the macro is not implemented for the adapter
    -- throw an exception to indicate that the macro is not implemented for the adapter unless there is a very universal default, but we tend to not want to assume anything to reduce coupling between adapters
{% endmacro %}

-- in dbt-postgres (or whatever adapter you're working on)
{% postgres__my_awesome_macro(arg1, arg2, ...) %}
    -- the adapter-specific implementation of `my_awesome_macro`
{% endmacro %}
```
The last macro is usually the only one that has code in it, the others are really just for opening an interface in the adapter framework.

lllong33:
Many thanks to @mikealfare for patient guidance.
Here's my latest tweak, but in the get_relations example in test_macro. I can't get the adapter postgres__get_relatins correctly.
My operating steps is to modify the DBT-core code, then run `make dev`, then run `dbt compile -s test_macro`
image

core/dbt/include/global_project/macros/adapters/metadata.sql
```diff
@@ -63,3 +63,12 @@
   {{ exceptions.raise_not_implemented(
     'list_relations_without_caching macro not implemented for adapter '+adapter.type()) }}
 {% endmacro %}
+
+{% macro get_relations() %}
+  {{ return(adapter.dispatch('get_relations')) }}
```
mikealfare, line 68: This is missing the 'dbt' parameter and it's not actually calling the macro, it's just passing it in. I think this line needs to look like this:
```sql
    {{ return(adapter.dispatch('get_relations', 'dbt')()) }}
```

lllong33:
good ! 
image

mikealfare (approved):

Issue #7940, opened 2023-06-23
[CT-2735] [Bug] Compile query error doesn't include node information


ChenyuLInx:
### Is this a new bug in dbt-core?

- [X] I believe this is a new bug in dbt-core
- [X] I have searched the existing issues, and I could not find an existing issue for this bug

### Current Behavior

This might be one symptom of a more general error handling problems in dbt-core.
The specific situation happened here is that when **compiling a inline dbt sql**, and **run into an execption**, dbt will fire an event that doesn't contain information about what node run into this issue. See Steps to Repro for command.

This is because uncatched exceptions are catched in `requires.postflight` linked below instead of in the place where we compile query.
https://github.com/dbt-labs/dbt-core/blob/38ca4fce2522aca496a72ccb704472d5bdf85908/core/dbt/cli/requires.py#L90

### Expected Behavior

I would expect the fired event contains information about which node run into issue.  for example, maybe the json log looks like
```json
{
    "data": {
            "exc": "Compilation Error in sql operation inline_query (from remote system.sql)\n  'a' is undefined. This can happen when calling a macro that does not exist. Check for typos and/or install package dependencies with \"dbt deps\"."
            "unique_id": "sqloperation.jaffle_shop.inline_query",
        },
    "info": {
            "category": "",
            "code": "Z002",
            "extra": {},
            "invocation_id": "780f6edf-c3f2-43a9-a742-eeefd1619c80",
            "level": "error",
            "msg": "Compilation Error in sql operation inline_query (from remote system.sql)\n  'a' is undefined. This can happen when calling a macro that does not exist. Check for typos and/or install package dependencies with \"dbt deps\".",
            "name": "compile error",
            "pid": 464,
            "thread": "somethread",
            "ts": "2023-06-16T19:58:11.410810Z"
    }
}
```

Current issue is because we do not have much exception handling logic in various tasks in general. This is a `dbt-core` general problem.
But specifically for this case, we can make sure we handle exception and fire proper event then re raise another exception at
https://github.com/dbt-labs/dbt-core/blob/aecbb4564cd8aee26deca1f381228f6006798c61/core/dbt/task/compile.py#L38
and 
https://github.com/dbt-labs/dbt-core/blob/aecbb4564cd8aee26deca1f381228f6006798c61/core/dbt/task/compile.py#L132
to fire event that contains node information in place.

### Steps To Reproduce

1. in a jaffle_shop project
2. run
    ```bash
[...]

PR #7960, opened 2023-06-27
fire proper event for inline query error

ChenyuLInx:
resolves https://github.com/dbt-labs/dbt-core/issues/7940

Before:
```
~/git/jaffle_shop (main*) » dbt compile --inline 'select * from {{ref("model_not_exisit")}}'  
21:44:36  Running with dbt=1.6.0-b7
21:44:36  Registered adapter: postgres=1.6.0-b7
21:44:36  Encountered an error:
Compilation Error
  Sql_Operation 'sql_operation.jaffle_shop.inline_query' (from remote system.sql) depends on a node named 'model_not_exisit' which was not found
```
After
```
~/git/jaffle_shop (main*) » dbt compile --inline 'select * from {{ref("model_not_exisit")}}'                                                  1 ↵ chenyuli@Chenyu-Li
21:43:57  Running with dbt=1.6.0-b7
21:43:58  Registered adapter: postgres=1.6.0-b7
21:43:58  Error while parsing node: inline_query
Sql_Operation 'sql_operation.jaffle_shop.inline_query' (from remote system.sql) depends on a node named 'model_not_exisit' which was not found
21:43:58  Encountered an error:
Error parsing inline query
```

The main difference here now is that the error we log are for the specific node, instead of a general error in main thread.


tests/functional/compile/test_compile.py
```diff
@@ -187,12 +187,23 @@ def test_compile_inline_not_add_node(self, project):
...
-            populate_cache=False,
-        )
+        dbt = dbtRunner()
         assert len(manifest.nodes) == 4
 
+    def test_compile_inline_syntax_error(self, project, mocker):
+        patched_fire_event = mocker.patch("dbt.task.compile.fire_event")
+        dbt = dbtRunner()
+        res = dbt.invoke(["compile", "--inline", "select * from {{ ref(1) }}"])
+        patched_fire_event.assert_called_once()
```
ChenyuLInx, line 197: I didn't really find a better way to test things here other than making sure we call fire event and re-raised a proper error.

core/dbt/task/compile.py
```diff
@@ -129,14 +134,26 @@ def defer_to_manifest(self, adapter, selected_uids: AbstractSet[str]):
...
+                block_parser = SqlBlockParser(
+                    project=self.config, manifest=self.manifest, root_project=self.config
+                )
+                sql_node = block_parser.parse_remote(self.args.inline, "inline_query")
+                process_node(self.config, self.manifest, sql_node)
+                # keep track of the node added to the manifest
+                self._inline_node_id = sql_node.unique_id
+            except CompilationError as exc:
+                fire_event(
+                    ParseNodeError(
```
gshank, line 147: I made this request before, but somehow github lost it... "ParseNodeError" is too generic of a name, it sounds like it might apply to any parsing error, but it's only fired for inline nodes. Could we call it something like "ParseInlineNodeError"?

gshank (changes requested):

core/dbt/task/compile.py
ChenyuLInx: The reason I didn't name this to be specific for `inline node` is the same error can happened for any node.
Given the exception I am catching here, do you think `NodeCompilationError` would be a better one that's not too generic? Or any other suggestions that doesn't contain `inline node`?

gshank (approved):

core/dbt/task/compile.py
ChenyuLInx: Per offline discussion with Gerda, we decided that raising a narrower scoped exception is fine because distinguishing between inline node parsing and project node parsing might be useful.

PR #7973, opened 2023-06-28
Do not log traceback when log format is json

### LINKED ISSUE #7972: [CT-2753] [Bug] dbt-core 1.3 and pre versions would create non-json log when

ChenyuLInx:
resolves #7972
This is only for 1.3 and pre versions of dbt-core.

When calling `logger.debug` with `exc_info` set to True, logger will actually print out the formatted log to stdout(?). We want to avoid it when we set log_format to `JSON`

Will backport to 1.2 and 1.1 and 1.0 once this is merged.

### DIFF
--- a/.changes/unreleased/Fixes-20230627-224355.yaml
+++ b/.changes/unreleased/Fixes-20230627-224355.yaml
@@ -1,0 +1,6 @@
+kind: Fixes
+body: Fix producing non json log when log format is json
+time: 2023-06-27T22:43:55.071123-07:00
+custom:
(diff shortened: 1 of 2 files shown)


jtcohen6:
@ChenyuLInx Change itself seems fine to me.

> Will backport to 1.2 and 1.1 and 1.0 once this is merged.

We've no longer been cutting patches of v1.0 and v1.1 (since they've reached EOL) - let's discuss offline if there's a need to do this?

peterallenwebb (approved):

PR #7984, opened 2023-06-28
CT-2707: Populate metric input measures

### LINKED ISSUE #7884: Derived metrics throw an error when using the run based semantic manifest.

### Current Behavior

I'm getting this error when querying a derived metric: `ins.ValueError: Unknown metric: MetricReference(element_name='order_gross_profit') in element set`. I noticed that the measures list in the semantic manifest is empty, which is likely causing the error: 

@QMalcolm per our conversation today in standup we need to update the way we populate measures. 
```
 {
            "name": "order_gross_profit",
            "description": "Gross profit from each order.",
            "type": "derived",
            "type_params": {
                "measure": null,
                "measures": [],
                "numerator": null,
                "denominator": null,
                "expr": "revenue - cost",
                "window": null,
                "grain_to_date": null,
                "metrics": [
                    {
                        "name": "food_order_amount",
                        "filter": null,
                        "alias": "revenue",
                        "offset_window": null,
                        "offset_to_grain": null
                    },
                    {
                        "name": "order_amount",
                        "filter": null,
                        "alias": "cost",
                        "offset_window": null,
                        "offset_to_grain": null
                    }
                ]
            },
            "filter": null,
            "metadata": null
        },
```

### Expected Behavior

I should be able to query derived metrics from the run based manifest. 

### Steps To Reproduce

1. Create a derived metric like the one in jaffle-sl-template: https://github.com/dbt-labs/jaffle-sl-template/blob/d7ce5ba300bbe809a9d36578912c36d1379175f5/models/metrics/example_metrics.yaml#L48
2. Kick of a run in dbt cloud
3. Run semantic_layer.query()

### Relevant log output

_No response_

### Environment

```markdown
- OS:
- Python:
- dbt:
```

_No response_

### Additional Context

_No response_

QMalcolm:
resolves #7884 

### Description

We weren't populating `MetricNode.type_params.input_measures` and now we do! It was problematic that we weren't populating the `input_measures` because the semantic layer depends on this field to be populated in order to properly generate queries. Now we populate it, and life is good (tests included 🙂)!

### Demo

<a href="https://www.loom.com/share/1e6533198b944664a6083587bc3c8e9f">
    
  </a>

### Checklist

### DIFF
--- a/core/dbt/parser/manifest.py
+++ b/core/dbt/parser/manifest.py
@@ -122,6 +122,8 @@ from dbt.version import __version__
 from dbt.dataclass_schema import StrEnum, dbtClassMixin
 from dbt.plugins import get_plugin_manager
 
+from dbt_semantic_interfaces.enum_extension import assert_values_exhausted
+from dbt_semantic_interfaces.type_enums import MetricType
 
 PARSING_STATE = DbtProcessState("parsing")
 PERF_INFO_FILE_NAME = "perf_info.json"
@@ -1064,16 +1066,15 @@ class ManifestLoader:
     # node, and updates 'depends_on.nodes' with the unique id
     def process_metrics(self, config: RuntimeConfig):
         current_project = config.project_name
+        for metric in self.manifest.metrics.values():
+            if metric.created_at < self.started_at:
+                continue
+            _process_metric_node(self.manifest, current_project, metric)
+            _process_metrics_for_node(self.manifest, current_project, metric)
         for node in self.manifest.nodes.values():
             if node.created_at < self.started_at:
                 continue
             _process_metrics_for_node(self.manifest, current_project, node)
-        for metric in self.manifest.metrics.values():
-            # TODO: Can we do this if the metric is derived & depends on
-            # some other metric for its definition? Maybe....
-            if metric.created_at < self.started_at:
-                continue
-            _process_metrics_for_node(self.manifest, current_project, metric)
         for exposure in self.manifest.exposures.values():
             if exposure.created_at < self.started_at:
                 continue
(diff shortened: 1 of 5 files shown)


core/dbt/parser/manifest.py
```diff
@@ -1439,6 +1440,63 @@ def _process_refs(
...
+        assert (
+            metric.type_params.measure is not None
+        ), f"{metric} should have a measure defined, but it does not."
+        metric.type_params.input_measures.append(metric.type_params.measure)
+
+    elif metric.type is MetricType.DERIVED or metric.type is MetricType.RATIO:
+        input_metrics = metric.input_metrics
+        if metric.type is MetricType.RATIO:
+            if metric.type_params.numerator is None or metric.type_params.denominator is None:
+                raise dbt.exceptions.ParsingError(
```
QMalcolm, line 1467: I wasn't wholly sure if this should be considered a `ParsingError` or `CompilationError` given the point at which we are in the manifest loading process. Happy to switch it up if a different error than what I went with is more appropriate.

core/dbt/parser/manifest.py
```diff
@@ -1439,6 +1440,63 @@ def _process_refs(
...
+    """Sets a metric's input_measures"""
+
+    # This ensures that if this metrics input_measures have already been set
+    # we skip the work. This could happen either due to recursion or if multiple
+    # metrics derive from another given metric.
+    # NOTE: This does not protect against infinite loops
+    if len(metric.type_params.input_measures) > 0:
+        return
+
+    if metric.type is MetricType.SIMPLE or metric.type is MetricType.CUMULATIVE:
```
QMalcolm, line 1457: Typically for an if statement like this I'd do more of a
```python
if metric.type in [MetricType.SIMPLE, MetricType.CUMULATIVE]:
...
```
However the type checker help of `assert_values_exhausted` doesn't work with that kind of formatting. Thus the individual `is` based checked.

core/dbt/parser/manifest.py
gshank: I'd call it a parsing error. The boundary isn't very clear because certain Jinja errors are always called compilation errors no matter whether they're in parsing or compilation.

gshank (approved):

PR #7985, opened 2023-06-29
Feature/materialized views/adap 608

mikealfare:

### Description

Pull more of the materialized views functionality out of the `dbt-postgres` space and into the `dbt-core` space.

### Checklist


mikealfare:
For any readers, as a preemptive strike, please note that I split out files much wider than we normally do...

core/dbt/adapters/relation/models/_relation_stub.py
mikealfare, line 1: Rename this `relation_ref`

mikealfare: Done

core/dbt/adapters/materialization/factory.py
```diff
@@ -0,0 +1,39 @@
...
+        if parser := self._get_parser(materialization_type):
+            assert self.relation_factory is not None  # mypy
+            return parser.from_runtime_config(
+                runtime_config=runtime_config,
+                relation_factory=self.relation_factory,
+                existing_relation_ref=existing_relation_ref,
+            )
+        return None
+
+    def _get_parser(
```
MichelleArk, line 36: why is this called _get_parser when it is returning an `Optional[Type[models.Materialization]]`

core/dbt/include/global_project/macros/materializations/materialized_view.sql
```diff
@@ -0,0 +1,68 @@
+{%- materialization materialized_view, default -%}
+
+    {%- set materialization = adapter.make_materialization_from_runtime_config(config, 'materialized_view') -%}
```
MichelleArk, line 3: This config variable is the one documented [here](https://docs.getdbt.com/reference/dbt-jinja-functions/config), correct? 
> The config variable exists to handle end-user configuration for custom [materializations](https://docs.getdbt.com/terms/materialization). Configs like unique_key can be implemented using the config variable in your own materializations.

If so, I don't believe it's accurate to call it the `runtime_config` as it is not an instance of the [RuntimeConfig class in core](https://github.com/dbt-labs/dbt-core/blob/main/core/dbt/config/runtime.py#L90), which is much broader in its scope and responsibilities. 

Perhaps just `materialization_config`? Is it possible to type this input at all through the python implementation?

core/dbt/adapters/materialization/README.md
```diff
@@ -0,0 +1,111 @@
...
+docstrings on these functions for more detail:
+
+- `conform_part`
+- `render_part`
+- `render`
+
+### Materialization
+This is a new class that serves as a service layer to expose `MaterializationConfig` functionality in the
+jinja context. We're effectively tucking away the modelling of database objects in python and only exposing
+class methods to serve as a basic API into `MaterializationConfig`.
```
MichelleArk, line 83: Currently dbt classes available in the jinja context are documented here: https://docs.getdbt.com/reference/dbt-classes. Let's open an issue to do the same for `Materialization` in https://github.com/dbt-labs/docs.getdbt.com/. I could see an argument in holding off on it until the pattern is adopted more broadly for other materialization types though.

core/dbt/adapters/materialization/models/_materialization.py
```diff
@@ -0,0 +1,143 @@
...
+        materialization = cls.from_dict(config_dict)
+        return materialization
+
+    @classmethod
+    def parse_runtime_config(
+        cls,
+        runtime_config,
+        relation_factory: RelationFactory,
+        existing_relation_ref: Optional[RelationRef] = None,
+    ) -> dict:
```
MichelleArk, line 125: Not the only place where we're doing this but it's more common in `dbt-core` to to use `typing.Dict` as opposed to `dict` -- it's more flexible and allows you to specify the type of the of keys and values as well

core/dbt/include/index.html
dataders, line 1: delete this?

core/dbt/adapters/relation/factory.py
```diff
@@ -0,0 +1,135 @@
...
+class RelationFactory:
+    """
+    Unlike other classes that get used by adapters, this class is not intended to be subclassed. Instead,
+    an instance should be taken from it that takes in all the required configuration (or defaults to
+    what is here).
+    """
+
+    # this configuration should never change
+    backup_suffix: str = "__dbt_backup"
+    intermediate_suffix: str = "__dbt_tmp"
```
MichelleArk, line 20: nit: uppercasing these constants would be more standard stylistically 

Reference: https://stackoverflow.com/a/45702782

tests/unit/relation_tests/conftest.py
```diff
@@ -0,0 +1,166 @@
...
+
+
+def test_materialization_factory(materialization_factory):
+    postgres_parser = materialization_factory.relation_factory._get_parser(
+        RelationType.MaterializedView
+    )
+    assert postgres_parser == MaterializedViewRelation
+
+
+def test_materialized_view_ref(materialized_view_ref):
```
colin-rogers-dbt, line 150: are these supposed to be in conftest?

colin-rogers-dbt: oh are these testing our fixtures?

core/dbt/include/global_project/macros/materializations/materialized_view.sql
MichelleArk: Recapping from 1:1 for visibility - this is a `RuntimeConfigObject` but could just be the `ModelNode` and associated `NodeConfig` to make this interface more narrow in scope. `config.get` does some nice safety handling to ensure there is a config on `config.model` that would be nice to maintain.

core/dbt/adapters/materialization/factory.py
mikealfare: I didn't want to confuse it with getting an instance of `Materialization`. It's returning the appropriate class for the type. And the class, at the end of the day, is really just a parser. I'm not tied to it, but I didn't want to call it `get_materialization` since it doesn't actually produce a `Materialization` instance.

core/dbt/include/global_project/macros/materializations/materialized_view.sql
mikealfare: This has been updated to `config.model`, which is a `ModelNode` most of the time from what I can tell. We only need a `CompiledNode` for a lot of the modeling, and less than that even (`ParsedNode`) in other spots. `adapter.make_materialization_from_runtime_config` will be updated to `adapter.make_materialization_from_node`. This is purposely ambiguous since `config.model` changes meaning over time. But the expectation in our use cases is that we will at least have a `CompiledNode`. Typing will be updated to reflect that `Relation` models need `CompiledNode` instances and `RelationComponent` models will need `ParsedNode` instances.

core/dbt/adapters/materialization/README.md
mikealfare: I added this to my backlog, associated with the MV epic.

core/dbt/adapters/materialization/models/_materialization.py
mikealfare: Agreed, I use `Dict` where I can. But sometimes the types are unwieldy. I suppose the vast majority fall into `Dict[str, Any]` and I could use that that to slightly more precise than `dict`, which is effectively `Dict[Any, Any]`.

core/dbt/include/index.html
mikealfare: Interesting, I must have caught that with a find/replace. I'll make sure there are no changes from `main`, but I can't delete it as it's in `main`. Nice catch though.

core/dbt/adapters/materialization/models/_materialized_view.py
```diff
@@ -0,0 +1,55 @@
...
+class MaterializedViewMaterialization(Materialization, ABC):
+    """
+    This config identifies the minimal materialization parameters required for dbt to function as well
+    as built-ins that make macros more extensible. Additional parameters may be added by subclassing for your adapter.
+    """
+
+    target_relation: Relation = None  # type: ignore
+    existing_relation_ref: RelationRef = None  # type: ignore
+    intermediate_relation: Relation = None  # type: ignore
+    backup_relation_ref: RelationRef = None  # type: ignore
```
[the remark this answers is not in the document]
mikealfare: That was probably when I went 12 rounds with `mypy` one time. I'll remove them if I can and either resolve the issue or have an explanation for why it's there.

core/dbt/adapters/relation/factory.py
mikealfare: True, I was thinking of them as default settings, but they should literally never change. I'll make the change.

tests/unit/relation_tests/conftest.py
mikealfare: Yup, make sure that the fixtures were actually successfully created so that we're hunting down the right problem. They are not thorough tests, just make sure the thing was created and has some identifying piece of information.

core/dbt/adapters/materialization/factory.py
mikealfare: Renaming this `_get_materialization_class` and the counterpart in `RelationFactory` `_get_relation_class`.

mikealfare:
This PR is way out of date. The branch will be saved for reference, but the PR is stale.

PR #8001, opened 2023-06-30
Add validate_sql method to base adapter with implementation for SQLAdapters

### LINKED ISSUE #7839: [CT-2681] [Feature] Add `explain` or `dry_run` method to BaseAdapter

### Describe the feature

Sometimes we want to validate queries for syntactical or attribute reference errors without triggering the query run itself. For example, in the MetricFlow integration, we want to be able to run warehouse validations that do a simple metric query verification to ensure all metrics and expressions are defined using legal SQL expressions, but we do not want to trigger warehouse queries as that would be too costly.

For most SQL engines, this can be achieved by prefixing the query with the `EXPLAIN` keyword. Unfortunately, BigQuery does not support this - instead, callers have to pass in a flag to the BQ client at query time, which is not currently possible with the dbt-bigquery adapters. 

We propose adding a method to BaseAdapter that will allow end users to issue dry run/explain queries so callers - such as MetricFlow - can have a consistent interface for doing so regardless of the underlying engine being used in the dbt project.

Here's a suggested mock up of the API:

```
def dry_run(self, sql: str) -> None:
    """Trigger a dry run of a SQL query against the underlying engine. Useful for things like validating query syntax."""
    raise NotImplementedError
```

In the SQLAdapter we could default this method to prefix the input query with the `EXPLAIN` keyword, and then add an internal implementation in the BigQueryAdapter that works for that.

This method as described would either return None on success or throw an exception on failure. We could also return a result object here instead of requiring callers to try/catch every invocation manually.

Questions:

1. Are there better alternatives?
2. Do we want to call this `dry_run` or `explain`? The reasoning behind `dry_run` is explain might not return a useful plan output result.
3. Do we prefer to let a failure propagate as an exception, or handle the exception and return the errors along with some kind of response object? The former is more typical, the latter is more in line with how adapters manage calls to execute().
4. Should we make this an `@abstractmethod` on the BaseAdapter? That way downstream implementers know they have to add support for this if they are not extending SQLAdapter (or BigQueryAdapter). The drawback is once this goes live existing implementations will probably not work with that version of core.
5. Should we leave room for bind parameters in the function signature? Those appear to only be supported through the add_query() method today, which is not implemented for BigQuery.

_No response_

### Who will this benefit?

_No response_

Yes, I will build this, I may need some guidance on test configuration.

### Anything else?

_No response_
```[tasklist]
### Tasks
- [ ] https://github.com/dbt-labs/dbt-snowflake/issues/682
- [ ] https://github.com/dbt-labs/dbt-redshift/issues/516
- [ ] https://github.com/dbt-labs/dbt-spark/issues/820
- [ ] https://github.com/dbt-labs/dbt-bigquery/issues/805
```

tlento:
resolves #7839

### Description

In the CLI integration, MetricFlow will issue dry run queries as
part of its warehouse-level validation of the semantic manifest,
including all semantic model and metric definitions.

In most cases, issuing an `explain` query is adequate, however,
BigQuery does not support the `explain` keyword and so we cannot
simply pre-pend `explain` to our input queries and expect the
correct behavior across all contexts.

This PR adds a validate_sql() method to the BaseAdapter which allows
for an implementation of the necessary functionality. It also includes
an implementation on the SQL adapter which delegates the SQL
expression adjustments to a macro which may be overridden as
needed for adapters supporting incompatible dialects. By default,
the macro pre-pends `explain` to the input SQL.

The current implementation should work out of the box with most
of our adapters. BigQuery will require us to implement a method on
the BigQueryConnectionManager for issuing queries with the dry-run
flag set, and community-maintained adapters can opt in by enabling
the test and ensuring their own implementations work as expected.

Note - we decided to make these concrete methods that throw runtime
exceptions for direct descendants of BaseAdapter in order to avoid
forcing community adapter maintainers to implement a method that does
not currently have any use cases in dbt proper.

### Checklist


core/dbt/adapters/sql/connections.py
```diff
@@ -152,6 +151,18 @@ def execute(
...
+    def dry_run(self, sql: str) -> AdapterResponse:
+        """Submit the given SQL to the engine for validation, but not execution.
+
+        By default we simply prefix the query with the explain keyword and allow the
+        exceptions thrown by the underlying engine on invalid SQL inputs to bubble up
+        to the exception handler.
+
+        :param sql str: The sql to validate
+        """
+        explain_sql = f"explain {sql}"
```
jtcohen6, line 163: Generally we would wrap this in a Jinja macro, and call that macro here (via `execute_macro`) instead of Python f-string. Something like:
```sql
{% macro get_explain_sql(query) %}
    explain {{ query }}
{% endmacro %}
```
That way it's "user space" code that someone could could override if they wanted to. Including an adapter maintainer, if the only thing that's different on their data platform is (e.g.) a synonym for `explain`.

There are legitimate cases for using Python templating over Jinja templating: Easier to write, to unit-test. In cases where we believe it's tricky-to-write & functionally uncontroversial, such that no user would realistically want/need to override, that approach can make more sense.

Not saying we need to go one way or the other — just offering my perspective & context on how we've done this in the past.

core/dbt/adapters/base/connections.py
```diff
@@ -412,3 +413,14 @@ def execute(
         :rtype: Tuple[AdapterResponse, agate.Table]
         """
         raise dbt.exceptions.NotImplementedError("`execute` is not implemented for this adapter!")
+
+    def dry_run(self, sql: str) -> AdapterResponse:
```
colin-rogers-dbt, line 417: since 'run' generally refers to a `dbt run` command (imagine if we develop a `--dry-run` flag for dbt at some point) should we just call this 'explain' or 'validate_sql'?

core/dbt/adapters/base/impl.py
```diff
@@ -785,7 +796,6 @@ def _make_match(
         schema: str,
         identifier: str,
     ) -> List[BaseRelation]:
-
```
tlento, line 788: Weird, must be some auto-formatter somewhere. :/

tlento (commented):
We've moved to the macro approach, but it's easy to revert if we change our minds. I'm als happy to do renames or whatever else is needed!

core/dbt/adapters/base/connections.py
jtcohen6: `validate_sql` makes sense to me! That captures the narrower goal of this method for now

tlento:
Renamed all the things, waiting for tests to pass. Are we ready here?

jtcohen6:
@tlento Good by me!

@colin-rogers-dbt Could you give this another look-over when you get a chance, and the ✅ if it lgty?

colin-rogers-dbt (approved):
LGTM

Issue #7969, opened 2023-06-27
[CT-2752] Run DSI SemanticManifest validations during parsing


QMalcolm:
In [dbt-semantic-interfaces](https://github.com/dbt-labs/dbt-semantic-interfaces/) there lives a tool called [SemanticManifestValidator](https://github.com/dbt-labs/dbt-semantic-interfaces/blob/adf7a992a231f5be6329e18b98c53f1d5b53db2a/dbt_semantic_interfaces/validations/semantic_manifest_validator.py#L57). The `SemanticManifestValidator` takes in a semantic manifest and runs a series of rules against it to ensure that it is valid. These are things like ensuring that either a `measure` has an `agg_time_dimension` set or that there is a `default_agg_time_dimension` set on the `semantic_model` that the `measure` belongs to.

Currently we don't run these validation rules in core. The only way someone using MetricFlow with core can run the validations is by doing
```bash
$ dbt parse
...
$ mf validate-configs
```
The issue with this is that validations are only run after the semantic manifest is written. This is problematic because validation issues should cause a failure before writing the semantic manifest. Additionally most people likely won't run `mf validate-configs` separately. Which means the issues would only be caught at query time 😬 We want to be informing customers of their issues when writing their configs, thus we need to ensure the `SemanticManifestValidator` is run during parsing.

Luckily, using the `SemanticManifestValidator` is fairly easy. It's a generic python class that requires that you declare an implementation of the `SemanticManifest` protocol on instantiation. It would look something like
```python
validator = SemanticManifestValidator[SemanticManifestImplementation]()
validation_results = validator.validate(semantic_manifest)
```

The other part of this is presently we only construct the `SemanticManifest`, and specifically a `PydanticSemanticManifest` right before we write out the semantic manifest. We can either reuse the [PydanticSemanticManifest builder](https://github.com/dbt-labs/dbt-core/blob/7156cc5c1da9da2dae9d7f2fd4d524e39773bb61/core/dbt/contracts/graph/manifest.py#L988C18-L988C19) to get a `PydanticSemanticManifest` and run that through a `SemanticManifestValidator` or we'll need to actually create a core implementation of the [SemanticManifest protocol](https://github.com/dbt-labs/dbt-semantic-interfaces/blob/main/dbt_semantic_interfaces/protocols/semantic_manifest.py) and then run that through a `SemanticManifestValidator`. The former is easier, but the latter is probably more in core's style.

# Acceptance Criteria

[...]

PR #8012, opened 2023-06-30
Extract Methods to New SemanticManifest Class

peterallenwebb:
This is a functionality-neutral refactor in to make the rest of the implementation of #7969 cleaner.

### Checklist

### DIFF
--- a/core/dbt/contracts/graph/semantic_manifest.py
+++ b/core/dbt/contracts/graph/semantic_manifest.py
@@ -1,0 +1,62 @@
+from dbt_semantic_interfaces.implementations.metric import PydanticMetric
+from dbt_semantic_interfaces.implementations.project_configuration import (
+    PydanticProjectConfiguration,
+)
+from dbt_semantic_interfaces.implementations.semantic_manifest import PydanticSemanticManifest
+from dbt_semantic_interfaces.implementations.semantic_model import PydanticSemanticModel
+from dbt_semantic_interfaces.implementations.time_spine_table_configuration import (
+    PydanticTimeSpineTableConfiguration,
+)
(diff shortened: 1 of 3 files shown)


core/dbt/contracts/graph/semantic_manifest.py
```diff
@@ -0,0 +1,62 @@
...
+                )
+            # Create time_spine_table_config, set it in project_config, and add to semantic manifest
+            time_spine_table_config = PydanticTimeSpineTableConfiguration(
+                location=model.relation_name,
+                column_name="date_day",
+                grain=TimeGranularity.DAY,
+            )
+            pydantic_semantic_manifest.project_configuration.time_spine_table_configurations = [
+                time_spine_table_config
+            ]
```
QMalcolm, line 60: I realize that this is just a copy and paste, but while we're here can we actually move this logic to be before the two for loops that populate the semantic_models and metrics for the semantic manifest? Basically there's no reason to go through all the work of converted and appending them if we're just gonna error out here.

QMalcolm (approved):
This is actually really exciting. We're getting exceedingly close to having fully implemented the `SemanticManifest` protocol [as defined in DSI](https://github.com/dbt-labs/dbt-semantic-interfaces/blob/main/dbt_semantic_interfaces/protocols/semantic_manifest.py). As a follow up I may even do that work. This will be really nice when we get to the stage of wanting to pass a python object directly to `dbt-metricflow` instead of going through the whole serialization/de-serialization handoff.

core/dbt/contracts/graph/semantic_manifest.py
peterallenwebb: Sure. Will take care of that in the next changeset for this issue.

PR #8049, opened 2023-07-07
Semantic Model Validation

### LINKED ISSUE #7969: [CT-2752] Run DSI SemanticManifest validations during parsing

peterallenwebb:
resolves #7969

### Description

Support validation of Semantic Models by issuing a new log message type on validation warnings and errors, and throwing an exception at parse time when there are errors.

### Checklist


core/dbt/events/types_pb2.py
```diff
@@ -16,3464 +15,10 @@
...
-_RUNRESULTWARNINGMESSAGEMSG = DESCRIPTOR.message_types_by_name['RunResultWarningMessageMsg']
-_DEBUGCMDOUT = DESCRIPTOR.message_types_by_name['DebugCmdOut']
-_DEBUGCMDOUTMSG = DESCRIPTOR.message_types_by_name['DebugCmdOutMsg']
-_DEBUGCMDRESULT = DESCRIPTOR.message_types_by_name['DebugCmdResult']
-_DEBUGCMDRESULTMSG = DESCRIPTOR.message_types_by_name['DebugCmdResultMsg']
-_LISTCMDOUT = DESCRIPTOR.message_types_by_name['ListCmdOut']
-_LISTCMDOUTMSG = DESCRIPTOR.message_types_by_name['ListCmdOutMsg']
-_NOTE = DESCRIPTOR.message_types_by_name['Note']
-_NOTEMSG = DESCRIPTOR.message_types_by_name['NoteMsg']
-EventInfo = _reflection.GeneratedProtocolMessageType('EventInfo', (_message.Message,), {
```
QMalcolm, line 454: It's weird that the next 3k lines were removed and not regenerated in any form 🤔

QMalcolm (commented):
This looks great! Thanks again for doing this work.

As a side note, do you have any idea what's going on in `types_pb2.py`? This file gets auto generated, which it looks like you did. But it removed (and didn't regenerate) a whole section which seems odd 🤔

QMalcolm:
We just merged #8050, pulling that in should fix the code quality check failure

core/dbt/events/types_pb2.py
peterallenwebb: I looked into this a bit. Although my changeset removes all these lines, they were not present in older versions of the file. It looks like they may have been introduced unintentionally by [this changeset](https://github.com/dbt-labs/dbt-core/commit/4c44c29ee4cec511f4751cc9fec668949ff9726b).

peterallenwebb:
@QMalcolm I'd argue this is ready to merge. I disabled the check we added while pairing because several existing tests have metrics without semantic models, and I'd like to move ahead with this and fix that separately. I'm happy to follow that up in this sprint if we have time.

QMalcolm:
> @QMalcolm I'd argue this is ready to merge. I disabled the check we added while pairing because several existing tests have metrics without semantic models, and I'd like to move ahead with this and fix that separately. I'm happy to follow that up in this sprint if we have time.

Yea in the `depends_on` work I'm doing I'm fixing those tests 😂 I'll rebase after this is merged, and then uncomment the check 👍

QMalcolm (approved):
I'm still confused about why we're generated so much less code in `types_pb2`, but everything seems happy so 🤷 Looks good!

PR #8050, opened 2023-07-07
pin click

### LINKED ISSUE #8048: [CT-2804] Exclude `click==8.1.4` from dependencies
## Problem
When `click==8.1.4` was released, our code quality workflows using `mypy` began failing. An issue has been created on the [click repository](https://github.com/pallets/click/issues/2558). 

## Solution
The solution is to exclude, for the time being, `click==8.1.4`. Currently our click dependency is set to `click>=7.0,<9.0`, this should become on main `click>=8.1.1,<8.1.4`.

## Backports
We need to backport this fix to `1.3.latest`, `1.4.latest`, and `1.5.latest`. For the backports we should update the dependency from `click>=7.0,<9.0` to `click>=7.0,<8.1.4`. The reason for the different specification in the backports is that we already support `click 7.x` in these earlier versions. Dropping support for `click 7.x` could be problematic if people are installing dbt-core alongside other dependencies which limit click to `7.x.`, then dropping support for `click 7.x` would represent a breaking change (and we shouldn't do this in a patch version).

emmyoop:
resolves #8048 

### Description

Pin main to `click>=8.1.1,<8.1.4`

### Checklist

### DIFF
--- a/.changes/unreleased/Dependencies-20230707-104052.yaml
+++ b/.changes/unreleased/Dependencies-20230707-104052.yaml
@@ -1,0 +1,6 @@
+kind: Dependencies
+body: Pin click>=8.1.1,<8.1.4
+time: 2023-07-07T10:40:52.565603-05:00
+custom:
+  Author: emmyoop
+  PR: "8050"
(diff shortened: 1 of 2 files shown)


QMalcolm (approved):
Thank you again for taking this on! Looks poifect!

PR #8032, opened 2023-07-05
partial parse file path

ChenyuLInx:
resolves #7911

### Description
This adds the ability to specify a env var,  cli args(hidden), kwargs in dbtRunner to specify a path file to partial parse file.

### Checklist

- [x] I have read [the contributing guide](https://github.com/dbt-labs/dbt-core/blob/main/CONTRIBUTING.md) and understand what's expected of me
- [x] I have signed the [CLA](https://docs.getdbt.com/docs/contributor-license-agreements)
- [x] I have run this code in development and it appears to resolve the stated issue
- [x] This PR includes tests, or tests are not required/relevant for this PR
- [x] I have run `changie new` to [create a changelog entry](https://github.com/dbt-labs/dbt-core/blob/main/CONTRIBUTING.md#adding-a-changelog-entry)

PR #8066, opened 2023-07-10
Fixes #7785: fail-fast behavior

### LINKED ISSUE #7785: [CT-2638] Fix `--fail-fast` behavior
A/C:
1. Convert output of `--fail-fast` from single element to list
2. Inherit `--fail-fast` flag in `dbt retry` for consistency
3. Record skipped nodes in `run_results.json` (see https://github.com/dbt-labs/dbt-core/pull/7763#discussion_r1218248760, copied below)

> One thing that @MichelleArk noticed while investigating https://github.com/dbt-labs/dbt-core/issues/7744 (fix in https://github.com/dbt-labs/dbt-core/pull/7767): If a model is cancelled because of --fail-fast, its downstream dependencies will not be recorded as "skipped" in run_results.json, and so it won't be retried by dbt retry.
> (Repro case: model_a -> model_b (with error) -> model_c. dbt run -x && dbt retry. model_c never runs.)
> Would it make sense to address that here, or to open as a new/separate issue?
```[tasklist]
### Tasks
- [ ] Convert output of --fail-fast from single element to list
- [ ] Inherit --fail-fast flag in dbt retry for consistency
- [ ] Record skipped nodes in run_results.json (see Fixes #7299: dbt retry #7763 (comment), copied below)
```

aranke:
resolves #7785

### Checklist

### DIFF
--- a/core/dbt/task/base.py
+++ b/core/dbt/task/base.py
@@ -192,20 +192,6 @@ class ExecutionContext:
         self.node = node
 
 
-def mark_node_as_skipped(node, message):
-    thread_id = threading.current_thread().name
-    return RunResult(
-        status=RunStatus.Skipped,
-        thread_id=thread_id,
-        execution_time=0,
-        timing=[],
-        message=message,
-        node=node,
-        adapter_response={},
-        failures=None,
-    )
-
-
 class BaseRunner(metaclass=ABCMeta):
     def __init__(self, config, adapter, node, node_index, num_nodes):
         self.config = config
@@ -484,7 +470,7 @@ class BaseRunner(metaclass=ABCMeta):
                     )
                 )
 
-        node_result = mark_node_as_skipped(self.node, error_message)
+        node_result = RunResult.from_node(self.node, RunStatus.Skipped, error_message)
         return node_result
 
     def do_skip(self, cause=None):
--- a/core/dbt/task/runnable.py
+++ b/core/dbt/task/runnable.py
@@ -14,7 +14,13 @@ from dbt.adapters.base import BaseRelation
 from dbt.adapters.factory import get_adapter
 from dbt.contracts.graph.manifest import WritableManifest
 from dbt.contracts.graph.nodes import ResultNode
-from dbt.contracts.results import NodeStatus, RunExecutionResult, RunningStatus
+from dbt.contracts.results import (
+    NodeStatus,
+    RunExecutionResult,
+    RunningStatus,
+    RunResult,
+    RunStatus,
+)
 from dbt.contracts.state import PreviousState
 from dbt.events.contextvars import log_contextvars, task_contextvars
 from dbt.events.functions import fire_event, warn_or_error
@@ -47,7 +53,7 @@ from dbt.logger import (
     NodeCount,
 )
 from dbt.parser.manifest import write_manifest
-from dbt.task.base import ConfiguredTask, mark_node_as_skipped
+from dbt.task.base import ConfiguredTask
 from .printer import (
     print_run_result_error,
     print_run_end_messages,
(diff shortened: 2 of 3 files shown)


core/dbt/task/base.py
```diff
@@ -192,6 +192,20 @@ def __init__(self, node):
         self.node = node
 
 
+def skip_result(node, message):
```
ChenyuLInx, line 195: Nitty nits: This is from previous code, but `skip_result` function name seems like an action instead of result, wanna consider updating it to something along the line of `skipped_result` or `result_skipped`?

core/dbt/task/runnable.py
```diff
@@ -360,21 +357,25 @@ def execute_nodes(self):
         pool = ThreadPool(num_threads)
         try:
             self.run_queue(pool)
-
         except FailFastError as failure:
             self._cancel_connections(pool)
+
+            node_results_ids = [r.node.unique_id for r in self.node_results]
```
ChenyuLInx, line 363: Nitty Nitty Nits: It took me a sec to understand self.node_results only contain info about nodes that got executed here. Maybe updating the name to `executed_node_ids` will make it easier to read?

core/dbt/task/runnable.py
```diff
@@ -360,21 +357,25 @@ def execute_nodes(self):
...
             raise
-
-        pool.close()
-        pool.join()
-
-        return self.node_results
+        finally:
+            pool.close()
+            pool.join()
+            return self.node_results
```
ChenyuLInx, line 378: Out of curiosity, why return is also moved into the `finally` block?

tests/functional/retry/test_retry.py
```diff
@@ -180,3 +164,57 @@ def test_removed_file_leaf_node(self, project):
...
+    @pytest.fixture(scope="class")
+    def models(self):
+        return {
+            "sample_model.sql": models__sample_model,
+            "second_model.sql": models__second_model,
+            "union_model.sql": models__union_model,
+            "final_model.sql": "select * from {{ ref('union_model') }};",
+        }
+
+    def test_fail_fast(self, project):
```
ChenyuLInx, line 179: Love the functional test here!

ChenyuLInx (approved):
❤️ Thanks for adding this patch and great work on the functional test! I have some readability nits but will leave the choice to you.

Regarding tests I am wondering if we can try to mock things and add just unit test for the newly added exception handling logic by mocking stuff out. I did [one](https://github.com/dbt-labs/dbt-core/pull/8032/files#diff-23e0ece663df2f9f26baa12ea7d5d58d9082b51c090cca3522a4d7cf1fd04200R108) for adding the partial parse file path. This is slightly tricky to write, and I am not saying it is required for this PR. But the more unittest we write, the less functional test that require actually running queries agains postgres we have, and our CI can be faster(or at least doesn't get much slower). Just something to think about!

core/dbt/task/base.py
```diff
@@ -192,6 +192,20 @@ def __init__(self, node):
         self.node = node
 
 
+def mark_node_as_skipped(node, message):
```
MichelleArk, line 195: nit: missing typing

MichelleArk, line 195: Is there anywhere else that was considered to define this method? Having it straggling at the top-level of the base task doesn't feel like a good spot. Its responsibility is in creating a `RunResult` object from a node + message. Perhaps a classmethod on `RunResult`? For example, `RunResult.from_node(node, message, status)`?

MichelleArk (approved):

core/dbt/task/runnable.py
ChenyuLInx: But doesn't it mean we no longer return `self.node_results` when no exception happens?

MichelleArk: I believe `finally` blocks are executed whether an exception is raised or not

PR #8076, opened 2023-07-12
Update validate sql test classes to new nomenclature

tlento:
resolves # 7839

### Problem
The original implementation of validate_sql was called dry_run,
but in the rename the test classes and much of their associated
documentation still retained the old naming.

This is mainly cosmetic, but since these test classes will be
imported into adapter repositories we should fix this now before
the wrong name proliferates.

### Solution
Renamed the relevant symbols and updated docblocks as needed

### Checklist


mikealfare (approved):

tlento:
Thanks @mikealfare ! I don't have merge permissions, would you mind pushing the button?

PR #8078, opened 2023-07-12
Add semantic_models to tracked resource counts

### LINKED ISSUE #8077: [CT-2818] Anonymous usage tracking for semantic object definitions
https://github.com/dbt-labs/dbt-core/blob/b78d23f68d6e9124b74c07e0b101bed00c4171ed/core/dbt/compilation.py#L43-L64

- Tracking should already be in place for `metrics`
- It would be straightforward to add `semantic_models`
- While I'm not sure, it seems like it would be a bigger lift to track counts of `dimensions` and `entities`

**This issue can be closed when:**
- We can guarantee that tracking is working for `metrics` + `semantic_models`
- We have investigated whether there is a straightforward way to extend this tracking to `dimensions` + `entities`

jtcohen6:
resolves #8077
~no docs changes needed~

### Problem

We want to be reporting on & tracking the count of semantic models defined in a project

### Solution

Add `semantic_models` to the list of resources being counted

We _might_ be able to do this for `dimensions` & `entities` too, but it's a lot less nice & straighforward

### While here

Aesthetic preference: Don't include resources in the logged "stat line" that aren't actually used/defined in the project.
```
$ dbt compile
14:01:29  Running with dbt=1.6.0-b8
14:01:29  Registered adapter: postgres=1.6.0-b8
14:01:30  Found 2 models, 1 seed, 349 macros, 1 semantic model
```
versus
```
 $ dbt compile
14:01:39  Running with dbt=1.6.0-b8
14:01:39  Registered adapter: postgres=1.6.0-b8
14:01:39  Found 2 models, 0 tests, 0 snapshots, 0 analyses, 349 macros, 0 operations, 1 seed, 0 sources, 0 exposures, 1 semantic model, 0 metrics, 0 groups
```

### Checklist

### DIFF
--- a/core/dbt/compilation.py
+++ b/core/dbt/compilation.py
@@ -84,26 +84,17 @@ def _generate_stats(manifest: Manifest):
         if _node_enabled(node):
             stats[node.resource_type] += 1
 
-    for source in manifest.sources.values():
-        stats[source.resource_type] += 1
-    for exposure in manifest.exposures.values():
-        stats[exposure.resource_type] += 1
-    for metric in manifest.metrics.values():
-        stats[metric.resource_type] += 1
-    for macro in manifest.macros.values():
-        stats[macro.resource_type] += 1
-    for group in manifest.groups.values():
-        stats[group.resource_type] += 1
+    # REVIEW: Why are these counted if disabled, when nodes in the general node
+    # collection are not?
+    stats[NodeType.Source] += len(manifest.sources)
+    stats[NodeType.Exposure] += len(manifest.exposures)
+    stats[NodeType.Metric] += len(manifest.metrics)
+    stats[NodeType.Macro] += len(manifest.macros)
+    stats[NodeType.Group] += len(manifest.groups)
+    stats[NodeType.SemanticModel] += len(manifest.semantic_models)
 
     # TODO: should we be counting dimensions + entities?
-    # dimensions = set()
-    # entities = set()
-    for semantic_model in manifest.semantic_models.values():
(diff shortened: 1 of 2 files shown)


core/dbt/compilation.py
```diff
@@ -82,16 +84,17 @@ def _generate_stats(manifest: Manifest):
...
-    for exposure in manifest.exposures.values():
-        stats[exposure.resource_type] += 1
-    for metric in manifest.metrics.values():
-        stats[metric.resource_type] += 1
-    for macro in manifest.macros.values():
-        stats[macro.resource_type] += 1
-    for group in manifest.groups.values():
-        stats[group.resource_type] += 1
+    # REVIEW: Why are these counted if disabled, when nodes in the general node
+    # collection are not?
```
MichelleArk, line 88: Poked into this -- I believe it's because the disabled mapping on the manifest stores a broader set of node types ([GraphMemberNode](https://github.com/dbt-labs/dbt-core/blob/6604b9ca31726098e3a8455e008ea12ef46bac7d/core/dbt/contracts/graph/nodes.py#L1611-L1616)) than the general node collection. For example, disabled exposures would be omitted from `manifest.exposures` and instead find themselves in `manifest.disabled` and so would not get included in the count. It looks like the `_node_enabled` is really just handling `NodeType.Test` resources which presumably find themselves in `manifest.nodes` even if they are disabled.

peterallenwebb: That does make sense. I'll remove the comment.

MichelleArk (approved):
Nice refactor! ✨

PR #8087, opened 2023-07-13
gitignore .user.yml and profiles.yml

MichelleArk:
resolves #  

### Problem

running `pytest/unit` leaves behind a couple straggling files - .user.yml and profiles.yml. I'm not entirely sure which test is writing these, presumably something related to testing `dbt init` behaviour which writes a profiles.yml file.

### Solution

Ideally these files are cleaned up as part of the unit test execution, but they are not written by a unit test itself (during setup, for example) and instead are a side-effect of running certain codepaths. A quick fix and robust fix is to simply .gitignore these files to avoid erroneously checking them in. 
* e.g. writing .user.yml: https://github.com/dbt-labs/dbt-core/blob/main/core/dbt/tracking.py#L158

### Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,6 +29,8 @@ var/
 .mypy_cache/
 .dmypy.json
 logs/
+.user.yml
+profiles.yml
 
 # PyInstaller
 #  Usually these files are written by a python script from a template


QMalcolm (approved):
🚀 Makes sense! We should definitely follow up and figure out what is writing this 🙃

PR #8117, opened 2023-07-17
Ensure that target_schema from snapshot config is promoted to node level

### LINKED ISSUE #6745: [CT-1913] Snapshot target_schema & target_database used from dbt_project.yml

### Current Behavior

Profiles.yml
```
    dev:
      type: bigquery
      method: oauth
      project: gcp-project-test
      **dataset: holding**
      location: EU
```

snapshot config in dbt_project.yml 

```
snapshots:
    test_dbt_project
        +persist_docs:
            relation: true
            columns: true
        primary:
            snapshot_models:
                **+target_schema: snapshot_dataset**
```

schema.yml to add the descriptions to snapshots tables

```
snapshots:
    - name: customers
      config:
        enabled: true
      columns:
        - name: customer_id
           description: 'customer unique id'
```

snapshot folder structure:
```
-snapshots
  -primary
      -snapshot_models
         -customers.sql
         -orders.sql
         - schema.yml
```

When we add the schema.yml file under snapshots/primary/snapshots_models and run the below command

`dbt snapshot --select customers` then the model is getting created under holding dataset(profile.yml) instead of snapshot_dataset(dbt_project.yml)

-

### Expected Behavior

The snapshots should be created in the dataset defined in dbt_project.yml file. For now have removed the schema.yml file so that it can create the model's under dataset provided in dbt_project.yml file. This is stopping to add the descriptions to table columns

### Steps To Reproduce

details provide in descriptions

### Relevant log output

_No response_

### Environment

```markdown
- OS: Airflow 2.2 composer 2.2
- Python:3.8
- dbt: dbt bigquery 1.2.0
```

bigquery

### Additional Context

_No response_

gshank:
resolves #6745 

### Problem

Snapshot target_schema and target_database were being processed when the snapshot nodes were processed, but those config fields were not being promoted  to the node level schema and database if set in config in dbt_proejct.yml or a schema yaml file.

### Solution

Added code to the "update_parsed_node_relation_names" to promote target_schema and target_database for snapshot nodes.

### Checklist

### DIFF
--- a/core/dbt/parser/base.py
+++ b/core/dbt/parser/base.py
@@ -102,8 +102,7 @@ class RelationUpdate:
         self.package_updaters = package_updaters
         self.component = component
 
-    def __call__(self, parsed_node: Any, config_dict: Dict[str, Any]) -> None:
-        override = config_dict.get(self.component)
+    def __call__(self, parsed_node: Any, override: Optional[str]) -> None:
         if parsed_node.package_name in self.package_updaters:
             new_value = self.package_updaters[parsed_node.package_name](override, parsed_node)
         else:
@@ -280,9 +279,18 @@ class ConfiguredParser(
     def update_parsed_node_relation_names(
         self, parsed_node: IntermediateNode, config_dict: Dict[str, Any]
     ) -> None:
-        self._update_node_database(parsed_node, config_dict)
-        self._update_node_schema(parsed_node, config_dict)
-        self._update_node_alias(parsed_node, config_dict)
+
+        if parsed_node.resource_type == "snapshot":
+            if "target_database" in config_dict and config_dict["target_database"]:
+                parsed_node.database = config_dict["target_database"]
+            if "target_schema" in config_dict and config_dict["target_schema"]:
+                parsed_node.schema = config_dict["target_schema"]
+        else:
+            # These call the RelationUpdate callable to go through generate_name macros
+            self._update_node_database(parsed_node, config_dict.get("database"))
+            self._update_node_schema(parsed_node, config_dict.get("schema"))
+            self._update_node_alias(parsed_node, config_dict.get("alias"))
+
         self._update_node_relation_name(parsed_node)
 
     def update_parsed_node_config(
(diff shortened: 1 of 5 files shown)


gshank:
Note: the original ticket said that there was also an issue with "meta" however I could not find one and there wasn't actually a reproduction case for that. There is no difference in the way that "meta" is handled with snapshots or other nodes. There *is* a difference with target_schema and target_database. The original example was not setting "meta" as a config, but as a node level attribute. I'm not sure what people were looking at to determine that setting "meta" was failing.

gshank:
I was misinterpreting the original ticket. It was adding meta to a schema file which caused the bug to show up, there wasn't an actual problem with "meta". Never mind...

core/dbt/parser/base.py
```diff
@@ -280,9 +279,18 @@ def update_parsed_node_config_dict(
     def update_parsed_node_relation_names(
         self, parsed_node: IntermediateNode, config_dict: Dict[str, Any]
     ) -> None:
-        self._update_node_database(parsed_node, config_dict)
-        self._update_node_schema(parsed_node, config_dict)
-        self._update_node_alias(parsed_node, config_dict)
+
+        if parsed_node.resource_type == "snapshot":
```
MichelleArk, line 283: small nit: let's use the `NodeType.Snapshot` enum here instead if possible

core/dbt/parser/base.py
```diff
@@ -280,9 +279,18 @@ def update_parsed_node_config_dict(
...
+            if "target_database" in config_dict and config_dict["target_database"]:
+                parsed_node.database = config_dict["target_database"]
+            if "target_schema" in config_dict and config_dict["target_schema"]:
+                parsed_node.schema = config_dict["target_schema"]
```
MichelleArk, line 287: For my own understanding: Previously we were calling also `self._update_node_[database|schema|alias]` on snapshots nodes. Were those no-ops and are no longer necessary?

core/dbt/parser/base.py
gshank: Done

MichelleArk (approved):

PR #8150, opened 2023-07-19
Rearrange pp_versioned_models test

gshank:
resolves # 7781

### Problem

The partial_parsing test TestVersionedModels was intermittently failing, from 1 to 3 times per 500 runs.  From the logs from the failing test runs it appeared to happen in the view materialization macro when "existing_relation" was retrieved from "load_cached_relation" but no longer existed by the time the rename was done later: 

```
  {% if existing_relation is not none %}
    {{ adapter.rename_relation(existing_relation, backup_relation) }}
```

This is probably a race condition or something similar, and might possibly be due to the testing environment.

### Solution

I separated out the test that was failing into its own file, test_versioned_models.py and refactored a bit to avoid using "run_dbt_and_capture" which was making it harder to read the test output. This doesn't address any possibly underlying race condition, but so far the failure has not occurred with the flaky_test workflow. The test is commented so that if such an error occurs again, we could reduce the window for the race condition.

### Checklist

### DIFF
--- a/.changes/unreleased/Under the Hood-20230719-124611.yaml
+++ b/.changes/unreleased/Under the Hood-20230719-124611.yaml
@@ -1,0 +1,6 @@
+kind: Under the Hood
+body: Refactor flaky test pp_versioned_models
+time: 2023-07-19T12:46:11.972481-04:00
+custom:
+  Author: gshank
+  Issue: "7781"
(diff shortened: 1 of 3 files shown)


MichelleArk (approved):

MichelleArk:
Should we backport to 1.5 and 1.6 to avoid breaking automated workflows (e.g: https://dbt-labs.slack.com/archives/C02MC5GKCKX/p1689798984364629)?

gshank:
Yeah, we can do that. It's isolated to tests so ought to be okay.

PR #8151, opened 2023-07-19
Hotfix for 372: Use JSONEncoder in json.dumps

aranke:
Hotfix for https://app.incident.io/dbt-labs/incidents/372

### Problem

JSON serialization in dbt Cloud is failing on complex data types like `datetime`.

### Solution

Encode data with `dbt.utils.JSONEncoder`

### Checklist

### DIFF
--- a/core/dbt/events/eventmgr.py
+++ b/core/dbt/events/eventmgr.py
@@ -175,7 +175,7 @@ class _JsonLogger(_Logger):
         from dbt.events.functions import msg_to_dict
 
         msg_dict = msg_to_dict(msg)
-        raw_log_line = json.dumps(msg_dict, sort_keys=True, cls=dbt.utils.JSONEncoder)
+        raw_log_line = json.dumps(msg_dict, sort_keys=True, cls=dbt.utils.ForgivingJSONEncoder)
         line = self.scrubber(raw_log_line)  # type: ignore
         return line
 
--- a/core/dbt/events/functions.py
+++ b/core/dbt/events/functions.py
@@ -201,7 +201,7 @@ def stop_capture_stdout_logs():
 # the message may contain secrets which must be scrubbed at the usage site.
 def msg_to_json(msg: EventMsg) -> str:
     msg_dict = msg_to_dict(msg)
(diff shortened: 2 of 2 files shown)


core/dbt/utils.py
```diff
@@ -337,15 +337,16 @@ class JSONEncoder(json.JSONEncoder):
...
+        elif isinstance(obj, jinja2.Undefined):
             return ""
-        if hasattr(obj, "to_dict"):
+        elif hasattr(obj, "to_dict"):
             # if we have a to_dict we should try to serialize the result of
             # that!
             return obj.to_dict(omit_none=True)
-        return super().default(obj)
+        else:
+            return super().default(obj)
```
MichelleArk, line 349: thank you for leaving the campground cleaner than you found it ✨⛺

MichelleArk (approved):
Looks good! Non blocking comment - could you please fill in the problem context in the PR description for posterity?

aaronraff (commented):
Should this be using the forgiving encoder, or will the regular encoder be sufficient?
https://github.com/dbt-labs/dbt-core/blob/a1b067c683212bccd9bc10f6e52bcdd168ccf7ec/core/dbt/utils.py#L351

peter-bertuglia (approved):

aaronraff (approved):

gshank (commented):
It looks like the JSONEncoder will handle decimal, datetime, jinja undefined and Exception types, but the ForgivingJSONEncoder will take any field that gets a TypeError and stringify it.  I haven't seen a list of the things which are blowing up in the json logs... do they all fall in that list of things handled by the JSONEncoder? For current code those things aren't going to make it through the logging event init code, but for older code wouldn't we want the Forgiving JSONEncoder? Seems like that was probably the reason that it was used in dbt-rpc.

gshank:
If some other non-serializable class is passed to the old logging code, we'd end up with the same issue that this is supposed to address. I think it would be safer on older versions to used the ForgivingJSONEncoder. We'd rather not see Round 3 of json logging error whack-a-mole.

gshank (approved):

PR #8187, opened 2023-07-22
[CT-2860] Add IAM Authentication to dbt-postgres

### LINKED ISSUE #8186: [Feature] Add AWS IAM Authentication to dbt-postgres

### Describe the feature

As with dbt-redshift, I would like to enable dbt-postgres to use AWS IAM Authentication. I use IAM Authentication for nearly everything and it just bugs me that it is available with dbt-redshift but not with dbt-postgres.

I envision the same setup as dbt-redshift. With a connection factory that returns a connection function.
This would easily allow other implementations as well.

The following profile configurations would change
- `password` becomes optional
- `method` is added: default `database`; options: `database`, `iam`
- `iam_profile` is added: default `None`, used to overwrite default AWS IAM profile
- `region` is added: default `None`, used to overwrite default AWS Region

If IAM is chosen as method, then before connecting to the PostgreSQL db using `psycopg2.connect`, it would
- open boto3 session
- create rds client
- get the auth token by calling generate_db_auth_token and pass it to kwargs as password

Instead of implementing it in dbt, you could create a wrapper for dbt, which gets the token and passes it as a environment variable to dbt. Using env_var in the profile I could use this token as password.
But this token could only be used once. If during the run, the connection has to be reopened, this wouldn't work. Therefore I really don't like this this idea.

### Who will this benefit?

Everybody who is running dbt in AWS and is using IAM authentication for RDS PostgreSQL.

Yes; Since this is my first contribution to dbt, please give me some guidance.

### Anything else?

There is a closed issue #3122, which was closed because it became stale. The author is asking for a implementation for IAM Authentication with Amazon Aurora PostgreSQL.
There is no difference in the IAM Authentication implementation for Amazon RDS Aurora PostgreSQL and Amazon RDS for PostgreSQL. Both work exactly the same.

christopherscholz:
resolves dbt-labs/dbt-postgres#14
docs [#3798](https://github.com/dbt-labs/docs.getdbt.com/issues/3798)

### Problem

It is not possible to use AWS IAM Authentication with dbt-postgres.

### Solution

This adds IAM Authentication to dbt-postgres in order to connect to Amazon RDS Aurora PostgreSQL and Amazon RDS for PostgreSQL.
It does it in the same fashion as dbt-redshift. It is refactoring the connection function by providing it through a connection factory.

The following profile configurations change
- `password` becomes optional
- `method` is added: default `database`; options: `database`, `iam`
- `iam_profile` is added: default `None`, used to overwrite default AWS IAM profile
- `region is` added: default `None`, used to overwrite default AWS Region

If `iam` is chosen as `method`, then before connecting to the PostgreSQL db using `psycopg2.connect`, it would
- open boto3 session
- create rds client
- get the auth token by calling `generate_db_auth_token` and pass it to kwargs as password

if `method` is not set or set to `database`, the current password connection is used

### Checklist

### DIFF
--- a/.changes/unreleased/Features-20230722-201041.yaml
+++ b/.changes/unreleased/Features-20230722-201041.yaml
@@ -1,0 +1,6 @@
+kind: Features
+body: add IAM Authentication to dbt-postgres
+time: 2023-07-22T20:10:41.9286694+02:00
+custom:
+  Author: christopherscholz
+  Issue: "8186"
--- a/plugins/postgres/dbt/adapters/postgres/connections.py
+++ b/plugins/postgres/dbt/adapters/postgres/connections.py
@@ -3,18 +3,27 @@ from contextlib import contextmanager
 import psycopg2
 from psycopg2.extensions import string_types
 
+import boto3
+
 import dbt.exceptions
 from dbt.adapters.base import Credentials
 from dbt.adapters.sql import SQLConnectionManager
 from dbt.contracts.connection import AdapterResponse
+from dbt.dataclass_schema import StrEnum
+from hologram.helpers import StrLiteral
 from dbt.events import AdapterLogger
 
 from dbt.helper_types import Port
 from dataclasses import dataclass
-from typing import Optional
+from typing import Any, Optional
 
 
 logger = AdapterLogger("Postgres")
+
+
+class PostgresConnectionMethod(StrEnum):
+    DATABASE = "database"
+    IAM = "iam"
 
 
 @dataclass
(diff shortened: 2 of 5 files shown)


christopherscholz:
I have fixed the tests. Also tested it against a real Amazon RDS for PostgreSQL using IAM Auth, which worked for me without any issue.

christopherscholz:
I have changed the `Credentials` validation flow a bit. My goal was to keep it as restrictive as before without changing too much code.

The classmethod `PostgresCredentials.validate` (`hologram.JsonSchemaMixin.validate`) validates a `Dict` against the class. 

I have changed it, so that the dataclass `PostgresCredentials` is a union of all specific credential methods.  But instead of only validate against this one, I have changed the validate classmethod to 

1. validate against itself as before
2. if no error is risen then check which authentication method is chosen
3. validate also against the extenden PostgresCredentials dataclass for the specific method

dbeatty10:
Thanks for taking the time to open this PR @christopherscholz! Since opening, we've [decoupled dbt Adapters from dbt Core](https://github.com/dbt-labs/dbt-core/discussions/9171), and this code now lives in a separate repo: [dbt-postgres](https://github.com/dbt-labs/dbt-postgres).

PR #8279, opened 2023-08-02
Fix #7179

### LINKED ISSUE #7179: [CT-2310] [Bug] The test_no_introspect test fails intermittently
The Tests and Code Checks action has been failing on the main branch of dbt-core due to intermittent failure of the test_no_introspect test. We need to improve the reliablity of this test.

Initial investigation uncovered log lines suggesting that the issue may be related to closing an already-closed connection:

```
[DEBUG] file_log: 14:50:53.859407 [debug] [Thread-1  ]: Traceback (most recent call last):
  File "/home/runner/work/dbt-core/dbt-core/plugins/postgres/dbt/adapters/postgres/connections.py", line 66, in exception_handler
    yield
  File "/home/runner/work/dbt-core/dbt-core/core/dbt/adapters/sql/connections.py", line 80, in add_query
    cursor = connection.handle.cursor()
psycopg2.InterfaceError: connection already closed
```

However, there is another issue, which is that the resulting DbtRuntimeException is constructed with the wrong type in postgres/connections.py at line 89. The parameter is an Exception rather than a str:

```python
raise dbt.exceptions.DbtRuntimeError(e) from e
```

As a result, a secondary exception is generated during exception handling.

aranke:
resolves #7179

### Checklist

### DIFF
--- a/.changes/unreleased/Under the Hood-20230802-153143.yaml
+++ b/.changes/unreleased/Under the Hood-20230802-153143.yaml
@@ -1,6 +1,0 @@
-kind: Under the Hood
-body: flaky test test_no_introspect
-time: 2023-08-02T15:31:43.27438+01:00
-custom:
-  Author: aranke
-  Issue: "7179"


.changes/unreleased/Under the Hood-20230802-153143.yaml
emmyoop, line 1: No changelog needed for this.

emmyoop (approved):
Just remove the changelog and add the label to skip the check in CI and this looks good to me.

Issue #7965, opened 2023-06-27
[CT-2749] [Spike] Investigate merging `merge_from_artifact` and `add_from_artifact`


aranke:
Thinking more about this: Do we actually need separate methods for `merge_from_artifact` and `add_from_artifact`? Or could it simply be that:
- for nodes which are "being deferred," we completely clobber/replace this manifest's version with the other manifest's version
- for nodes which _aren't_ "being deferred" (because they're being selected, or already exist in this schema), we keep the existing node entry and add the `defer_relation` attribute

Then, all tasks could call this same method. We don't need the divergent behavior between `clone` and `run`, and we set ourselves up more nicely for future work around contract inference, dev/prod diff, ...

Pseudo code:
```python
    def merge_from_artifact(
        self,
        adapter,
        other: "WritableManifest",
        selected: AbstractSet[UniqueID],
        favor_state: bool = False,
    ) -> None:
        """Given the selected unique IDs and a writable manifest, update this
        manifest by replacing any unselected nonexisting nodes with their counterpart.
        For any selected/existing nodes, add information about each node's location
        in the other manifest.

        Only non-ephemeral refable nodes are examined.
        """
        refables = set(NodeType.refable())
        merged = set()
        for unique_id, node in other.nodes.items():
            current = self.nodes.get(unique_id)
            
            # do the replace thing
            if current and (
                node.resource_type in refables
                and not node.is_ephemeral
                and unique_id not in selected
                and (
                    not adapter.get_relation(current.database, current.schema, current.identifier)
                    or favor_state
                )
            ):
                merged.add(unique_id)
                self.nodes[unique_id] = node.replace(deferred=True)
           
           # do the additive thing
           elif current and (node.resource_type in refables and not node.is_ephemeral):
                defer_state_relation = RelationalNode(node.database, node.schema, node.alias)
                self.nodes[unique_id] = current.replace(defer_relation=defer_relation)

        # Rebuild the flat_graph, which powers the 'graph' context variable,
        # now that we've deferred some nodes
        self.build_flat_graph()

        # log up to 5 items
        sample = list(islice(merged, 5))
        fire_event(MergedFromState(num_merged=len(merged), sample=sample))
```

[...]

PR #8339, opened 2023-08-08
spike: inferred contracts

MichelleArk:
resolves https://github.com/dbt-labs/dbt-core/issues/7432

### Checklist

:tophat:

model.sql
```sql
select 'test' as test, 2 as test2, 3 as test3
```

schema.yml
```yml
version: 2
models:
  - name: model1
    config:
      contract:
        enforced: true
        inferred: true
    columns: 
      - name: test3
        data_type: int
        constraints:
          - type: unique
```

👟 
1. Set `enforced: false` (& optionally comment `columns` out)
2. `dbt run`  to create a prod version of the model
3. Set `enforced: true`  (& optionally comment `columns` back in)
4. `dbt run --defer --state target/`


core/dbt/context/providers.py
```diff
@@ -1382,7 +1382,8 @@ def sql(self) -> Optional[str]:
         if self.model.language == ModelLanguage.sql:  # type: ignore[union-attr]
             # If the model is deferred and the adapter doesn't support zero-copy cloning, then select * from the prod
             # relation
-            if getattr(self.model, "defer_relation", None):
+            # TODO: avoid routing on args.which if possible
+            if getattr(self.model, "defer_relation", None) and self.config.args.which == "clone":
```
jtcohen6, line 1386: Definitely want to avoid the need for this. I had previously wondered if we'd be able to consolidate the deferral methods used by `clone` and other commands, by:
- clobbering unselected resources with the production versions (to support "traditional" deferral)
- adding `defer_relation` for selected resources only (to support `clone`)

(this comment/issue: https://github.com/dbt-labs/dbt-core/issues/7965)

core/dbt/include/global_project/macros/materializations/models/table/columns_spec_ddl.sql
```diff
@@ -8,8 +8,22 @@
 
 {% macro table_columns_and_constraints() %}
   {# loop through user_provided_columns to create DDL with data types and constraints #}
+    {%- if defer_relation and config.get('contract').inferred -%}
+      {% set prod_columns = get_column_schema_from_query(get_empty_subquery_sql("select * from " ~ defer_relation)) %}
+      {% set raw_columns = {} %}
+      {% for column in prod_columns -%}
+        {# TODO: reconsider this mutating operation, could be done with an additional warehouse call #}
```
jtcohen6, line 15: 👍 At minimum, we should probably be deepcopying to avoid mutating the user-provided value stored in the context... right?

jtcohen6 (commented):
Not as bad as I thought! Now we just need to think through, is it something we actually want?

core/dbt/context/providers.py
jtcohen6: Discussing live with @MichelleArk: We should overwrite `compiled_code` (also aliased as `sql`) within the `clone` task. Then we don't need to have this task-aware conditional logic live in the context method.

core/dbt/include/global_project/macros/materializations/models/table/columns_spec_ddl.sql
jtcohen6: We can either:
- plumb it through (but this takes some time, and tightly couples the methods)
- make the warehouse call twice (and consider caching/optimization possibilities in the future)

jtcohen6:
Ah, I knew I'd been forgetting something. From the issue:
> Crucially, we should **allow** column additions, but raise a `ContractBreakingChangeError` if an existing column is renamed/removed/retyped.

I think this is two things.

### (1) Adding new columns
Right now, to add a column to a model with an inferred contract, you need to explicitly define its name + data type. Otherwise, you get a `ContractError`.
```
13:35:49    Compilation Error in model model1 (models/model1.sql)
  This model has an enforced contract that failed.
  Please ensure the name, data_type, and number of columns in your contract match the columns in your model's definition.

  | column_name | definition_type | contract_type | mismatch_reason     |
  | ----------- | --------------- | ------------- | ------------------- |
  | test4       | INT             |               | missing in contract |
```
I think we have two options:
- (a) We figure out some way (not sure how) to combine the column names/types inferred from prod with the column names/types inferred from `get_column_schema_from_query`. The problem then is we'd need to do type translation between the cursor types and the DWH types (gross).
- (b) We say, c'est comme ça. Once the model has an enforced contract, _all new columns must be declared explicitly_. (Once the updated model has been run in prod, you could then remove those column specifications, and have them inferred instead.) I'm leaning toward this option.

### (2) Detecting breaking changes
We also need to update our logic for detecting `ContractBreakingChangeError`. Basically, we cannot rely on the contract being fully specified within the static yaml config. So for any models with an "inferred" contract, I think we need to:
- No-op the `same_contract` check within `state:modified`
- Instead perform the detailed comparison while the model is being run. If there's a column in prod that's missing in dev/CI, or the same column name with different data types, that should be a `ContractBreakingChangeError`.

PR #8355, opened 2023-08-10
Fix #8031: Call materialization macro from adapter dispatch

### LINKED ISSUE #8031: Support dispatch for materialization macros
In conversation with Kshitij just now, it seems that the main changes need to be in core/dbt/context/macros.py, in MacroNamespace.get_from_package where we can add special processing of "materialization" macros. The "get_from_package" method is called from adapter.dispatch, which comes from the adapter "contextproperty" of the ProviderContext in core/dbt/context/providers.py. The actual "dispatch" method is in the BaseDatabaseWrapper.

The materialization macro can be found in the "execute" method of the clone task (and probably the same method in runnable.py) by doing context.adapter.dispatch from the constructed context. This should already be a MacroGenerator callable, so we wouldn't need to construct a MacroGenerator as we do further down in "execute".

_Originally posted by @gshank in https://github.com/dbt-labs/dbt-core/issues/7799#issuecomment-1614996026_

---

### Motivation

This would enable us to refactor the implementation in the `clone` materialization here:

https://github.com/dbt-labs/dbt-core/blob/a58b5ee8fb9fafd8bbf9da52d3b9e8603725592c/core/dbt/include/global_project/macros/materializations/models/clone/clone.sql#L49-L58

It might also unblock an approach for this desirable UX improvement to model versions:
- #7442

aranke:
resolves dbt-labs/dbt-core#8031 

### Checklist

### DIFF
--- a/core/dbt/context/providers.py
+++ b/core/dbt/context/providers.py
@@ -13,12 +13,16 @@ from typing import (
     Mapping,
     Tuple,
 )
+
+import agate
 from typing_extensions import Protocol
 
+from dbt import selected_resources
 from dbt.adapters.base.column import Column
 from dbt.adapters.factory import get_adapter, get_adapter_package_names, get_adapter_type_names
 from dbt.clients import agate_helper
 from dbt.clients.jinja import get_rendered, MacroGenerator, MacroStack
+from dbt.config import IsFQNResource
 from dbt.config import RuntimeConfig, Project
 from dbt.constants import SECRET_ENV_PREFIX, DEFAULT_ENV_PLACEHOLDER
 from dbt.context.base import contextmember, contextproperty, Var
@@ -30,6 +34,7 @@ from dbt.context.macros import MacroNamespaceBuilder, MacroNamespace
 from dbt.context.manifest import ManifestContext
 from dbt.contracts.connection import AdapterResponse
 from dbt.contracts.graph.manifest import Manifest, Disabled
+from dbt.contracts.graph.metrics import MetricReference, ResolvedMetricReference
 from dbt.contracts.graph.nodes import (
     Macro,
     Exposure,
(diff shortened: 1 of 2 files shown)


gshank:
Pretty simple so far, and glad that it's working.  

Now that those basics are in, there are a few things to look at doing... You should use the same type of context.adapter.dispatch call in the run task, and look for breakage.

 @jtcohen6 was interested in allowing the same naming convention for materialization macros that we use for other macros, for ease of overriding. Right now we can't override easily. That's the naming convention that starts with the adapter, like "postgres__drop_schema". You should attempt to allow both forms of materialization naming and get them to work, with the "<adapter_name>__" style name overriding the earlier format.

I suspect that when we use context.adapter.dispatch in the run task we'll get breakage due to the special context that's used by the code that detects where the python code is being called from. You'll probably need to set that where the MacroGenerator objects are created.

core/dbt/clients/jinja.py
```diff
@@ -259,22 +256,8 @@ def call_macro(self, *args, **kwargs):
...
-        return len(self.call_stack)
-
-    def push(self, name):
-        self.call_stack.append(name)
-
-    def pop(self, name):
-        got = self.call_stack.pop()
-        if got != name:
-            raise DbtInternalError(f"popped {got}, expected {name}")
+class MacroStack(deque):
```
mikealfare: I love seeing changes like this.

core/dbt/context/providers.py
```diff
@@ -174,30 +175,34 @@ def dispatch(
...
-                    # any package
-                    macro = None
-
-                if package_name is None:
-                    attempts.append(search_name)
-                else:
-                    attempts.append(f"{package_name}.{search_name}")
-
-                if macro is not None:
+            if macro_name.startswith("materialization_"):
```
mikealfare, line 183: I think we made a fix recently for scenarios where macros that started with "materialization_" were incorrectly being picked up as materializations. For example, this was being flagged as a materialization instead of a macro:
```sql
{% macro materialization_setup() %}
    ...
{% endmacro %}
```
Have you verified this doesn't reintroduce that issue?

PR #8357, opened 2023-08-10
Register Postgres UUID and CIDR datatypes when creating a connection

### LINKED ISSUE #8353: [CT-2957] [Bug] Cannot specify column contract of type UUID in PostgreSQL with DBT 1.5.3

### Current Behavior

I have a source table with the following DDL, populated with three records:

```sql
CREATE TABLE src_onestream_raw_table.confirm_rule (
    conf_group_key uuid NOT NULL,
    display_order integer NOT NULL,
    display_type integer NOT NULL,
    failure_action integer NOT NULL,
    failure_message character varying NOT NULL,
    rule_formula character varying NOT NULL,
    rule_frequency integer NOT NULL,
    rule_name character varying NOT NULL,
    rule_text character varying NOT NULL,
    time_filter_for_reqt_freq character varying NOT NULL,
    unique_id uuid PRIMARY KEY,
    warning_message character varying NOT NULL,
    xml_data character varying NOT NULL
);
CREATE UNIQUE INDEX confirm_rule_pkey ON src_onestream_raw_table.confirm_rule(unique_id uuid_ops);
```

I have created the following staging model:
```sql
WITH

src AS (
    SELECT * FROM {{ source('onestream_raw_table', 'confirm_rule') }}
),

final AS (
    SELECT
        conf_group_key,
        display_order,
        display_type,
        failure_action,
        failure_message::TEXT,
        rule_formula::TEXT,
        rule_frequency,
        rule_name::TEXT,
        rule_text::TEXT,
        time_filter_for_reqt_freq::TEXT,
        unique_id,
        warning_message::TEXT,
        xml_data::TEXT
    FROM src
)

SELECT * FROM final
```

I have configured it this way:

```yaml
version: 2

models:
  # ...
  - name: stg_onestream_raw_table__confirm_rule
    constraints:
      - type: primary_key
        columns: [unique_id]
    columns:
      - name: conf_group_key
        data_type: UUID
      - name: display_order
        data_type: INTEGER
      - name: display_type
        data_type: INTEGER
      - name: failure_action
        data_type: INTEGER
      - name: failure_message
        data_type: TEXT
      - name: rule_formula
        data_type: TEXT
      - name: rule_frequency
        data_type: INTEGER
      - name: rule_name
        data_type: TEXT
      - name: rule_text
        data_type: TEXT
      - name: time_filter_for_reqt_freq
        data_type: TEXT
      - name: unique_id
        data_type: UUID
      - name: warning_message
        data_type: TEXT
      - name: xml_data
        data_type: TEXT
    config:
      contract:
        enforced: true
    # ...
```

When I run DBT, I get the following error, with no extra context and no other information:

```
2023-08-10 12:57:36 15:57:36  Unhandled error while executing 
2023-08-10 12:57:36 2950
```

It looks like `2950` is the `oid` of the UUID data type in postgres: https://github.com/postgres/postgres/blob/5e0c761d0a13c7b4f7c5de618ac38560d74d74d0/src/include/catalog/pg_type.dat#L403

If I cast all my UUIDs to TEXT, and change my contracts to TEXT instead of UUID, the error disappears. Without the contracts, the model runs just fine and the UUID type works as expected.

### Expected Behavior

See previous section

### Steps To Reproduce

See previous section

### Relevant log output

```shell
2023-08-10 12:57:36 15:57:36  Unhandled error while executing 
2023-08-10 12:57:36 2950
```

No further context or any other information is output with this error. It looks like `2950` is the `oid` of the UUID data type in postgres: https://github.com/postgres/postgres/blob/5e0c761d0a13c7b4f7c5de618ac38560d74d74d0/src/include/catalog/pg_type.dat#L403

### Environment

```
- Docker image "python:3.10.8" running on a mac
- Python: 3.10.8
- dbt: 1.5.3
```

postgres

### Additional Context

_No response_

gmontanola:
resolves dbt-labs/dbt-postgres#54

### Problem

Contracts checks for data types and when it finds stuff like UUID it fails because it's not registered by default.

### Solution

Registering some extra data types - UUID and IP ADDR

### Checklist

### DIFF
--- a/plugins/postgres/dbt/adapters/postgres/connections.py
+++ b/plugins/postgres/dbt/adapters/postgres/connections.py
@@ -34,7 +34,6 @@ class PostgresCredentials(Credentials):
     sslrootcert: Optional[str] = None
     application_name: Optional[str] = "dbt"
     retries: int = 1
-    register_extra_data_types: bool = False
 
     _ALIASES = {"dbname": "database", "pass": "password"}
 
@@ -63,7 +62,6 @@ class PostgresCredentials(Credentials):
             "sslrootcert",
             "application_name",
             "retries",
-            "register_extra_data_types",
         )
 
 
@@ -133,9 +131,6 @@ class PostgresConnectionManager(SQLConnectionManager):
         if credentials.application_name:
             kwargs["application_name"] = credentials.application_name
 
-        if credentials.register_extra_data_types:
-            kwargs["register_extra_data_types"] = credentials.register_extra_data_types
-
         def connect():
             handle = psycopg2.connect(
                 dbname=credentials.database,
@@ -147,9 +142,9 @@ class PostgresConnectionManager(SQLConnectionManager):
                 **kwargs,
             )
 
-            if credentials.register_extra_data_types:
-                register_uuid(conn_or_curs=handle)
-                register_ipaddress(conn_or_curs=handle)
+            # Register extra data types to avoid dbt contract errors
+            register_uuid(conn_or_curs=handle)
+            register_ipaddress(conn_or_curs=handle)
 
             if credentials.role:
                 handle.cursor().execute("set role {}".format(credentials.role))
--- a/plugins/postgres/dbt/adapters/postgres/connections.py
+++ b/plugins/postgres/dbt/adapters/postgres/connections.py
@@ -143,8 +143,8 @@ class PostgresConnectionManager(SQLConnectionManager):
             )
 
             # Register extra data types to avoid dbt contract errors
-            register_uuid(conn_or_curs=handle)
-            register_ipaddress(conn_or_curs=handle)
+            register_uuid()
+            register_ipaddress()
 
             if credentials.role:
                 handle.cursor().execute("set role {}".format(credentials.role))
--- a/plugins/postgres/dbt/adapters/postgres/connections.py
+++ b/plugins/postgres/dbt/adapters/postgres/connections.py
@@ -33,6 +33,7 @@ class PostgresCredentials(Credentials):
     sslrootcert: Optional[str] = None
     application_name: Optional[str] = "dbt"
     retries: int = 1
+    register_extra_data_types: bool = False
 
     _ALIASES = {"dbname": "database", "pass": "password"}
 
@@ -61,6 +62,7 @@ class PostgresCredentials(Credentials):
             "sslrootcert",
             "application_name",
             "retries",
+            "register_extra_data_types",
         )
 
 
@@ -129,6 +131,9 @@ class PostgresConnectionManager(SQLConnectionManager):
 
         if credentials.application_name:
             kwargs["application_name"] = credentials.application_name
+
+        if credentials.register_extra_data_types:
+            kwargs["register_extra_data_types"] = credentials.register_extra_data_types
 
         def connect():
             handle = psycopg2.connect(
--- a/plugins/postgres/dbt/adapters/postgres/connections.py
+++ b/plugins/postgres/dbt/adapters/postgres/connections.py
@@ -2,6 +2,7 @@ from contextlib import contextmanager
 
 import psycopg2
 from psycopg2.extensions import string_types
+from psycopg2.extras import register_uuid, register_ipaddress
 
 import dbt.exceptions
 from dbt.adapters.base import Credentials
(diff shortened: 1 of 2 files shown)


gmontanola:
Some points:

1. Should this be a adapter configuration option?
    a. If so I can modify the docs.
    b. And maybe add basic tests.
3. Are tests needed here? Feels like I'll be testing `psycopg2` if I try to check for the registered types.
4. AFAIC there are some additional types like `hstore` but I checked the code and it was not straightforward as UUID/IP. Should we register it?

gmontanola:
Well, I'll make this simpler since using this inside credentials won't work :smile:

plugins/postgres/dbt/adapters/postgres/connections.py
```diff
@@ -140,6 +141,11 @@ def connect():
                 connect_timeout=credentials.connect_timeout,
                 **kwargs,
             )
+
+            # Register extra data types to avoid dbt contract errors
+            register_uuid()
+            register_ipaddress()
```
jtcohen6, line 147: If I understand right -

This change will allow users to specify `UUID`, `CIDR`, `INET`, etc as data types within model contracts. These are standard types that were added in later versions of Postgres / psycopg2, which is why they're opt-in here. Is there any risk / downside to registering them here?

This still won't support [totally custom data types](https://www.postgresql.org/docs/current/sql-createtype.html) that users may define within their Postgres instances.

gmontanola: Yeap! **UUID** + **IP** stuff probably will cover most cases!

Not sure if `dbt` should handle edge or user specific cases right now (like custom data types) since contracts are not widespread and this will need a plugin-like functionality on the adapter to extend the range of accepted types.

colin-rogers-dbt:
> Are tests needed here? Feels like I'll be testing psycopg2 if I try to check for the registered types.

If we're going to say that dbt-postgres data contracts support something I think we want to test the integration of psycopg2 and data contracts

> What do you think? Can we make this always-on or should we keep this as adapter configuration?

I wonder if we can turn this on if needed (i.e. a callback function if a data contract specifies these types). Not saying this needs to be a blocker since we can probably keep it on by default

dataders:
### Assumptions

lmk if i'm wrong about any of the following

- `psycopg2.extras`'s `register_uuid` and `register_ipaddress` were added to psycopg2 for version `2.7` ([released](https://pypi.org/project/psycopg2/2.7/) March 1, 2017).
- dbt-postgres today is compat-pinned to `~=2.8`
- both postgres's `UUID` and `CIDR` data types were introduced in Postgres 11 (October 18, 2018)
 
### questions

- what happens if a user defines a contract on a column and gives one of these types and but is not using postgres 11 or newer?
- do we have documented somewhere the data types for which contracts are supported in postgres?  
- do we hope to support the long tail of data types across data platforms? (e.g. `hstore` for postgres)
- if a new data platform introduces a new data type, is there an explicit list of steps to ensure that they work within contracts?

gmontanola:
>If we're going to say that dbt-postgres data contracts support something I think we want to test the integration of psycopg2 and data contracts

I'll look into adding some tests then.

> What happens if a user defines a contract on a column and gives one of these types and but is not using postgres 11 or newer?

Nice question. I think a weird exception will appear but I'll test it using an older PG version. Then I can try/except graciously and emit a log message in case of failure.

---

I'm a bit swamped right now but I'll try to improve this ASAP.

jenna-jordan:
Hi! I just wanted to note here that I ran into the same issue with postgres's `money` datatype (as well as the `uuid` datatype).

mowsen:
Just wanted to come here and say thank you! Your PR fixed [my problem](https://discourse.snowplow.io/t/dbt-snowplow-web-not-working-unhandled-error-while-executing/9210) with snowplow + dbt + uuid. Thanks !

hkorhola:
seconding @ondraz and @mowsen , I also had to tweak the Python code back then to make it work properly with Postgres and would be good if this would be fixed in the main branch - whether with some configs or then that register -style

ondraz:
Just for the sake of completeness, this has been fixed by [this PR](https://github.com/MaterializeInc/materialize/pull/23885). I tested it in dbt-core version 1.7.10 and it works. Thank you.

dbeatty10:
Thanks for taking the time to open this PR @gmontanola! Since opening, we've [decoupled dbt Adapters from dbt Core](https://github.com/dbt-labs/dbt-core/discussions/9171), and this code now lives in a separate repo: [dbt-postgres](https://github.com/dbt-labs/dbt-postgres).

PR #8379, opened 2023-08-12
Ct 2599/childrens parents depth

### LINKED ISSUE #7703: [Feature] childrens_parents depth

### Describe the feature

Currently the `@` graph operator does not allow for depth control in any way and will always include *all children* of a model and their parents. 
I would like to be able to specify the depth of the children selection in the same way the children/parent operators allow for depth specification.

example:
`dbt ls -s 1@my_model` would result in the selection of the models children up to depth 1 + all the parents of those children.
The same should be possible with a YAML selector definition
```yml
definition:
  method: fqn
  value: my_model
  childrens_parents: true
  children_depth: 1
```

I tried finding a workaround with complex yaml selector definitions but since we can't "layer" definitions (take all the 1 depth children of this model + then take all the 1 depth parents of all the models in the selection) I could not get it to work as I wanted. 

### Who will this benefit?

For our specific use case we want to run all seeds that are used in the tests of a specified model.

Currently we can achieve this with the `@` operator and a resource-type filter but this also includes any seeds that are used in tests of any of the child models of the specified model, which are not necessary for us and increases run times.

sure, I haven't looked at the codebase structure but this feels like a good first issue.

### Anything else?

Some extra context on where/why this would help us:
https://github.com/duneanalytics/spellbook/pull/3408#issuecomment-1562445109

kentkr:
resolves #7703

### Problem

This feature adds the ability to specify the maximum childs parent depth like `1@model`. It functions similarly to `1+model`. 

### Solution

A more thorough discussion can be found in #7703. The solution was to add to existing functionality like spec parsing and child/parent depth restrictions when creating the DAG. 

### Checklist


kentkr:
This commit is a bit premature because some of the requested functionality in #7703 doesn't quite work. 

> * `1@changed_model+` (equivalent as`1@changed_model`)
>   all children, and their first-order parents

The above will not work due to

https://github.com/dbt-labs/dbt-core/blob/b7aee3f5a428aa69241839196289eeed166e24f1/core/dbt/graph/selector_spec.py#L75C10-L75C10

I don't know why the prefix`@` and suffix `+` would be incompatible but there must be a reason out there. Before I go ahead and change it I wanted to get confirmation on that. @jtcohen6 any thoughts?

dataders:
@kentkr hey do you mind resolving these merge conflicts so that we can take a look?

kentkr:
Hey! Yep thanks for getting back to me. I fixed the merge conflicts.

kentkr:
@dbeatty10 thanks for jumping in on this! Let me know if you need anything else from me.

.changes/unreleased/Features-20230811-223109.yaml
```diff
@@ -0,0 +1,6 @@
+kind: Features
+body: Allows for specifying a childs parents max depth
+time: 2023-08-11T22:31:09.729033-07:00
+custom:
+  Author: kentkr
+  Issue: CT-2599
```
dbeatty10, line 6:
```suggestion
  Issue: 7703
```

Issue #6643, opened 2023-01-18
[CT-1845] [Feature] Write a file of installed versions during `dbt deps`


dbeatty10:
### Is this your first time submitting a feature request?

- [X] I have read the [expectations for open source contributors](https://docs.getdbt.com/docs/contributing/oss-expectations)
- [X] I have searched the existing issues, and I could not find an existing issue for this feature
- [X] I am requesting a straightforward extension of existing dbt functionality, rather than a Big Idea better suited to a discussion

#### Problem
It's hard to determine the exact package versions that were installed during the `dbt deps` process.

#### Multi-part proposal
We should consider a feature with three phases, each subsequent one being optional:
1. As part of `dbt deps`, write a file with the version of each of the installed packages
    - One possible name is `packages.json`
    - By default, write within a `.gitignore`'d directory (`dbt_packages`, `target`, or something else)
    - This file would be essentially saying, "I'm installing the X version of package Y from location Z"
    - ☝️ we do do this in structured logging
1. Use this file to make a more compelling user experience(s)
    - Ex. a frozen set of packages à la `pip freeze` or `pip-tools compile`
    - Ex. compare the difference between packages asserted to be currently installed and those to-be downloaded; we could then optionally clean out-of-date versions and/or just install only the not-yet-installed versions
1. Include `dbt deps` automatically as part of every relevant sub-command
    - e.g., if we can make `dbt deps` such a delightful and pleasurable experience, why not include it within every sub-command?
    - This would implement the following:
        - https://github.com/dbt-labs/dbt-core/issues/4557
        - https://github.com/dbt-labs/dbt-core/issues/6570
    - Note: if we did this today, it would **not** be a great user experience

The advantage provided by the 1st phase is that users (and systems!) would know the exact package versions that were installed during the `dbt deps` process.

I'm initially keeping these three phases together to preserve their common context, but we can split these out into separate issues at any time.

The 1st phase described above is actionable right now, but by the time we get to the 3rd one, we're definitely in speculative territory.

### Describe alternatives you've considered

[...]

PR #6735, opened 2023-01-25
:sparkles: restructure deps command with package-lock.yml

justbldwn:
resolves #6643

### Description

this PR restructures the `dbt deps` command, aiming to add a few new key features:
- [x] create a new `package-lock.yml` file that will keep record of all packages installed during the last `dbt deps` run
- [x] use newly created `package-lock.yml` as the default file to install packages from, which helps reduce install time
- [x] adds 2 new `dbt deps` subcommands: `dbt deps lock` and `dbt deps add`
- [x] refresh `dbt_packages/` directory each time `dbt deps` is run (no more persisting old packages left in the directory)

the below will provide an overview of the changes, what they accomplish, and ~~some~~ **a lot** of detail on implementation and examples. sorry in advance for the length! 😅 

---------------------------

### `dbt deps lock` 
1. the new lock command will take the packages in `packages.yml` and list them and their transitive dependencies in a new `package-lock.yml` file.
2. package versions should be resolved for each dependency, else, error thrown.
3. `package-lock.yml` will get written to the project_root alongside `packages.yml`.

**example command:**
```bash
dbt deps lock
```

---------------------------

### `dbt deps add` 

1. allows a user to add a package to their `packages.yml` file with configurable version and location of a given package.
2. added a `--dry-run` flag to the `dbt deps add` sub-command that, when `False`, will re-create the lock file after a new package is added to the `packages.yml` file. (this is setup as `False` by default in `argparse`, a user can use the flag and set to True to **not** persist changes down to the `package-lock.yml`
3. overwrite package info in `packages.yml` if a package name already exists (either explicit or fuzzy name matching)
4. added a quality check to ensure a version is passed in whenever the `source != "local"`

**example command:**
```bash
# add package from hub (--source arg defaults to "hub")
dbt deps add --package dbt-labs/dbt_utils --version 1.0.0

# add package from hub with semantic version
dbt deps add --package dbt-labs/snowplow --version ">=0.7.0,<0.8.0"

# add package from git
dbt deps add --package https://github.com/fivetran/dbt_amplitude --version v0.3.0 --source git 

# add package from local (--version not required for local)
dbt deps add --package /opt/dbt/redshift --source local

# add package to packages.yml WITHOUT updating package-lock.yml
dbt deps add --package dbt-labs/dbt_utils --version 1.0.0 --dry-run True
```

### `dbt deps` 🚩(this is now changed as of 2/10/2023, changes noted in strikethrough) 🚩
[...]

PR #8408, opened 2023-08-15
Deps lock by justbldwn

ChenyuLInx:
resolves https://github.com/dbt-labs/dbt-core/issues/6643
This is work from @justbldwn. original PR at #6735. 
We(core team) are working on getting it up-to-date with the changes made recently about dependencies.yml and updating some small things to get it into ready to merge state.

We would like this to be a non breaking change, which means `dbt deps` will behave the same(-ish, caveat below). With this in mind, considering the constrain of how click commands are structured, also talked to Jerco about the philosophy of CLI commands of dbt-core, we are updating the `dbt deps add` and `dbt deps lock` to just `dbt deps --add` and `dbt deps --lock`. @justbldwn  let us know if you feel strongly about it should be sub command vs a flag.

The caveat with the current implementation:
Previously, running `dbt deps` will upgrade packages is version is specified as range and new version come out. For example, local package A is version 0.1.1, specified version [> 0.1.0, < 0.2.0]. Latest version is 0.1.5, running deps previously will install 0.1.5, but now with the lock file approach it will actually still install 0.1.1. @dbeatty10  how you feel about it?

## Description of behavior
- running `dbt deps` first time will generate `package-lock.yml` that contains all resolved packages
- running `dbt deps` with package spec in `depenedencies.yml` or `packages.yml` not updated, dbt-core will just install from `package-lock.yml`.(hence the caveat above)
- update package spec and run deps again, both package lock and package get updated accordingly
- There's the `--add` flag, when specified, along with `--package` and `--version`, dbt-core will try to update the package spec and run deps depending on whether `--dry-run` is specified. The original PR provide better examples. Remember `--dry-run` is not a flag and you do not need to say `--dry-run True`, just `--dry-run`.
- There's also the flag `--update` which can be used when you have packaged specified in version range and want to try to update to the new package available. Previous Deps behavior is actually `dbt deps --update` after this change.
- There's another flag `--lock` which will generate the new lock file but not actually install any packages.

## TODO for this PR
- [x] Add logic to store the hash of the previous defined packages there to automatically run deps when content packages got updated.
- [x] Add a `--upgrade` flag to upgrade the locks file if user specified version bound and there's new versions released
- [x] Add a `--lock` flag  to only generate a new lock file without actually install
- [x] make sure `--add` works with `dependencies.yml`
- [x] reorganize lock and add as args so `deps` will not be a breaking change, and we keep the consistent way of CLI commands where the commands are mostly verbs, and we alter the behavior of command with flags.
- [x] reformat the saved yml file
- [x] confirm how many times we will have to pull down a git repo when running deps with packages-lock not exist, and determine optimizing/not (Behavior here is the same as before, it will pull down git packages during resolution).

- [x] fix unittest

## TODO for opening follow up PR
There's a set of optimizations mentioned in #6643 that is enabled by the new lock yml. We should open follow up PRs about it
- [ ] install only diff package
- [ ] run deps before every command

### Checklist


core/dbt/task/deps.py
```diff
@@ -59,40 +121,166 @@ def track_package_install(
...
+            current_hash = _create_sha1_hash(self.project.packages.packages)
+            previous_hash = load_yml_dict(lock_file_path).get(PACKAGE_LOCK_HASH_KEY, None)
+            if previous_hash != current_hash:
+                self.lock()
 
+        # Early return when dry run or lock only.
+        if self.args.dry_run or self.args.lock:
+            return
+
+        if system.path_exists(self.project.packages_install_path):
```
ChenyuLInx, line 239: This is a change, update

ChenyuLInx:
Take a look at the resolvers and see if we can refactor

core/dbt/cli/params.py
```diff
@@ -465,6 +493,22 @@
...
+    default="hub",
+)
+
+source = click.option(
+    "--source",
+    envvar=None,
+    help="Source to download page from, must be one of hub, git, or local. Defaults to hub.",
+    type=click.Choice(["hub", "git", "local"], case_sensitive=True),
+    default="hub",
+)
```
MichelleArk, line 510: This looks like an exact duplicate of the above lines.

core/dbt/task/deps.py
jtcohen6: Yes! IMO this is a big improvement (sorry I never saw/responded to this sooner)

core/dbt/cli/params.py
```diff
@@ -248,6 +269,13 @@
     type=click.Path(file_okay=True, dir_okay=False, writable=True),
     default=None,
 )
+package = click.option(
+    "--package", envvar=None, help="Name of package to add to packages.yml.", type=click.STRING
+)
+
+package_version = click.option(
+    "--package-version", envvar=None, help="Version of the package to install.", type=click.STRING
+)
```
MichelleArk, line 278: have we considered removing these two options in favor with providing a `package@version` (or `package=version` or just `package`) as the option value for `--add`? I think that would be more intuitive from a user perspective (fewer flags to reason over), and could simplify option validation. It would just require a bit of parsing internally to get the package and package_version out of the provided `--add` option value.

core/dbt/task/deps.py
```diff
@@ -27,6 +36,59 @@
...
+    Returns:
+        str: SHA1 hash of the packages list
+    """
+    package_strs = [json.dumps(package.to_dict(), sort_keys=True) for package in packages]
+    package_strs = sorted(package_strs)
+
+    return sha1("\n".join(package_strs).encode("utf-8")).hexdigest()
+
+
+def _create_packages_yml_entry(package, version, source):
```
MichelleArk, line 61: nit: missing type annotations. Same for above methods on return types

core/dbt/task/deps.py
```diff
@@ -59,40 +121,166 @@ def track_package_install(
...
+        packages_yml_filepath = (
+            f"{self.project.project_root}/{self.project.packages_specified_path}"
+        )
+
+        if not system.path_exists(packages_yml_filepath):
+            raise dbt.exceptions.DbtRuntimeError("No packages definition file.")
+
+        if not self.args.package_version and self.args.source != "local":
+            raise dbt.exceptions.DbtRuntimeError(
+                f"Version is required to add a package when source is {self.args.source}"
```
MichelleArk, line 155: nit: this could be a little more actionable of we specified which flag was missing (e.g. --package-version) instead of 'Version'

core/dbt/cli/params.py
```diff
@@ -6,6 +6,12 @@
 from dbt.cli.resolvers import default_project_dir, default_profiles_dir
 from dbt.version import get_version_information
 
+add_package = click.option(
+    "--add",
+    help="Add a package to current package spec with `--package` and `--package-version`",
```
MichelleArk, line 11: is --source also only valid in the context of `--add`?

core/dbt/task/deps.py
```diff
@@ -59,40 +121,166 @@ def track_package_install(
...
+            self.project.packages.packages
+        )
+
+        with open(lock_filepath, "w") as lock_obj:
+            yaml.safe_dump(packages_installed, lock_obj)
+
+        fire_event(DepsLockUpdating(lock_filepath=lock_filepath))
+
+    def run(self) -> None:
+        if self.args.ADD:
```
MichelleArk, line 213: super nit: using varying conventions for accessing attributes on self.args in this method. The rest of the method uses lowercase, and I believe that's generally the case in the repo. so I'd use `self.args.add` here as well

core/dbt/cli/params.py
ChenyuLInx: Used @ in the end

core/dbt/cli/params.py
MichelleArk: I don't _love_ the location of the validation here (all the way in task running), but it's where we're doing a similar check for the dry-run flag: https://github.com/dbt-labs/dbt-core/pull/8408/files#diff-77c52d1632e496415c4b241f4eb7ced5e8f7c6cbe207955177b2e8c6c6a54227R217-R220

I think it's tricky to do this validation in click.

Some alternative locations for doing this validation would be either: 
1. in the `deps` click command method itself in main.py, or
2.  implementing something similar to `Flags._assert_mutually_exclusive` (perhaps `assert_dependency`- keeping click/arg parsing concerns happening at a higher level. 

I think ideally we could leverage click to do this validation by structuring `dbt deps` as a group and `dbt deps add` as a subcommand, but I understand we're not going down that path for backwards compatibility reasons - which is fair! So I'd bias to doing the simpler thing if anything (option 1) - to avoid legitimizing this pattern much, favoring subcommands when possible.

MichelleArk (approved):

PR #8433, opened 2023-08-16
Modify run results test to wait on timer instead of file write

peterallenwebb:
### Problem

This test previously waited on a file write which was not guaranteed to happen.

### Solution

We now wait on a timer, which is not ideal, but should suffice for now.

### Checklist


gshank:
Now it looks like the first integration batch is hanging.

Issue #7119, opened 2023-03-03
[CT-2268] [Bug] dbt-core >= 1.4.2 manifests not passing v8 schema validation


dlawin:
### Is this a new bug in dbt-core?

- [X] I believe this is a new bug in dbt-core
- [X] I have searched the existing issues, and I could not find an existing issue for this bug

### Current Behavior

I'm noticing that manifests generated by dbt-core versions 1.4.2, 1.4.3, and 1.4.4 are not passing json schema validation based on the schema here: 
- https://schemas.getdbt.com/dbt/manifest/v8.json

### Expected Behavior

I would expect this validation to pass or for a v9 manifest to be available.

### Steps To Reproduce

1. Generate manifest.json for one of the mentioned versions (I used jaffle_shop)
2.
```
import json
from jsonschema import validate
import yaml
import requests
manifest_path = "/Users/dan/Desktop/manifest.json"
r = requests.get(url = "https://schemas.getdbt.com/dbt/manifest/v8.json")

schema_str = r.content

schema = json.loads(schema_str)
with open(manifest_path, "r", encoding="utf-8") as fp:
            manifest_dict = yaml.safe_load(fp)

validate(
    instance=manifest_dict
    , schema=schema
)
```

### Relevant log output

```shell
'seed' is not one of ['analysis']

Failed validating 'enum' in schema[0]['properties']['resource_type']:
    {'enum': ['analysis'], 'type': 'string'}

On instance['resource_type']:
    'seed'
```

### Environment

```markdown

- OS: OSX 13.2
- Python 3.10.9
- dbt-core: 1.4.2, 1.4.3, 1.4.4
```

### Which database adapter are you using with dbt?

other (mention it in "Additional Context")

### Additional Context

Noticed this with any adapter I tested: snowflake, redshift, postgres, databricks

jtcohen6:
There are some known problems with the library (`hologram`) we're currently using to auto-generate JSONSchemas from our Python dataclasses. This was a previous bug report that sounds similar:
- https://github.com/dbt-labs/dbt-core/issues/4657

More discussion:
- #6776
- https://github.com/dbt-labs/dbt-core/discussions/6915#discussioncomment-4918721

dlawin:
For context, I contribute to a couple open source tools that utilize these artifacts and their json schemas:
https://github.com/yu-iskw/dbt-artifacts-parser

Subsequently using that in:
https://github.com/datafold/data-diff

[...]

Issue #4657, opened 2022-02-01
[CT-119] [Bug] run_results.json schema validation error with oneOf constraint on status


indyyyyy:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Trying to validate the run_results.json artifact produced by the `dbt test` command, I used the jsonschema python package and the schema you expose [here](https://schemas.getdbt.com/dbt/run-results/v3.json) (run-results/v3.json).
The validation fails with the error below.

### Expected Behavior

As the artifact and the json schema are produced by dbt, I expect the schema validation to be OK.
"oneOf" constraint seems to be unadapted as a status could appears in several enum.

### Steps To Reproduce

1. Run `dbt test`
2. In a python3 console, run the following lines:
```
import jsonschema
import requests

dbt_runresults_schema = requests.get(url='https://schemas.getdbt.com/dbt/run-results/v3.json').json()
runresults_rawjson=json.load(open('path/to/target/run_results.json'), encoding="UTF-8"))

validate(instance=runresults_rawjson, schema=dbt_runresults_schema)
```

### Relevant log output

```shell
>>> validate(instance=runresults_rawjson, schema=dbt_runresults_schema)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python3.9/site-packages/jsonschema/validators.py", line 934, in validate
    raise error
jsonschema.exceptions.ValidationError: 'warn' is valid under each of {'type': 'string', 'enum': ['pass', 'warn', 'error', 'runtime error']}, {'type': 'string', 'enum': ['pass', 'error', 'fail', 'warn', 'skipped']}

Failed validating 'oneOf' in schema['properties']['results']['items']['properties']['status']:
    {'oneOf': [{'enum': ['success', 'error', 'skipped'], 'type': 'string'},
               {'enum': ['pass', 'error', 'fail', 'warn', 'skipped'],
                'type': 'string'},
               {'enum': ['pass', 'warn', 'error', 'runtime error'],
                'type': 'string'}]}

On instance['results'][0]['status']:
    'warn'
```
```


### Environment

```markdown
- OS:Debian GNU/Linux 10 (buster)
- Python:3.9.6
- dbt:0.21.1
- jsonschema:3.1.1
```

### What database are you using dbt with?

redshift

### Additional Context

[...]

PR #8437, opened 2023-08-16
Convert to using mashumaro jsonschema with acceptable performance

### LINKED ISSUE #8426: [CT-3000] Handle performance/caching issues with using mashumaro jsonschema generation
### Description

In #6776, pull request #8132 we switched from using hologram to mashumaro for generating jsonschemas, but the lack of caching caused performance issues (CI jobs took more than twice as long to complete).

Using that work as a base, develop some kind of caching which will match the performance of hologram.

gshank:
resolves #8426

### Problem

Original conversion performed in #8132, but with performance issues. Use caching to improve performance.

See the comments in #8132 for additional context for code reviews.

### Checklist


core/dbt/contracts/graph/model_config.py
```diff
@@ -447,11 +436,11 @@ class NodeConfig(NodeAndTestConfig):
     persist_docs: Dict[str, Any] = field(default_factory=dict)
     post_hook: List[Hook] = field(
         default_factory=list,
-        metadata=MergeBehavior.Append.meta(),
+        metadata={"merge": MergeBehavior.Append, "alias": "post-hook"},
     )
     pre_hook: List[Hook] = field(
         default_factory=list,
-        metadata=MergeBehavior.Append.meta(),
+        metadata={"merge": MergeBehavior.Append, "alias": "pre-hook"},
```
emmyoop, line 443: Why are aliases needed for Append now?  Why does packages not need it on line 466?

core/dbt/contracts/project.py
```diff
@@ -25,12 +26,7 @@ def _deserialize(cls, value: str) -> "SemverString":
         return SemverString(value)
 
 
-# this supports full semver,
-# but also allows for 2 group version numbers, (allows '1.0').
-register_pattern(
-    SemverString,
-    r"^(0|[1-9]\d*)\.(0|[1-9]\d*)(\.(0|[1-9]\d*)(?:-((?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\+([0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?)?$",  # noqa
-)
+sem_ver_pattern = r"^(0|[1-9]\d*)\.(0|[1-9]\d*)(\.(0|[1-9]\d*)(?:-((?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\+([0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?)?$"
```
emmyoop, line 29: Can you add back in the comment so we know what the regex should match?

core/setup.py
```diff
@@ -72,12 +72,12 @@
         # ----
         # These are major-version-0 packages also maintained by dbt-labs. Accept patches.
         "dbt-extractor~=0.5.0",
-        "hologram~=0.0.16",  # includes transitive dependencies on python-dateutil and jsonschema
```
emmyoop, line 75: 🎉

core/dbt/contracts/graph/model_config.py
gshank: The "alias" is for handling the dashes in the names properly. Most of the other field definitions use that kind of hacky metadata=MergeBehavior.DictKeyAppend.meta() thing, which doesn't allow setting additional metadata.

core/dbt/contracts/project.py
gshank: Thanks! Comment has been restored.

gshank:
@jtcohen6 @graciegoheen I tagged you on this because the output jsonschema is different than what was generated by hologram in a number of ways. Do people actually read it? Do we have any concerns there?

For example, the resource_type shows up as a const, and the use of OneOf vs AnyOf is different.

core/setup.py
peterallenwebb: 🎊

jtcohen6:
@gshank I think that's fine, as long as this is a forward-looking change for new versions of `dbt-core`, and not a change to the existing published & versioned schemas.

We know that the jsonschemas generated by hologram were not always even technically correct, which could lead to edge cases if used for programmatic validation (e.g. https://github.com/dbt-labs/dbt-core/issues/4657). I am hoping that the ones produced by `mashumaro` can achieve better correctness!

I just want to clarify that:
- The actual contents of our contracted artifacts (`manifest.json`, `run_results.json`, `catalog.json`, `sources.json`) are not changing
- The new JSONSchemas (produced by `mashumaro`) can be used to successfully validate those artifacts, in the ways that we know some users try to do programmatically (e.g. https://github.com/dbt-labs/dbt-core/issues/7119)
- We will still be able to publish these JSONSchemas, and the "human-readable" versions, at schemas.getdbt.com

cc @dbt-labs/cloud-artifacts for visibility

gshank:
That's right, the other schemas will change too. Should I update the other schemas too? Nothing has probably changed as far as validation... Or should we wait for an actual change and just verify that newly generated schemas still work?

core/dbt/dataclass_schema.py
```diff
@@ -1,97 +1,133 @@
...
-        return self.to_dict(**args)
+    # This is called by the mashumaro to_dict in order to handle
+    # nested classes. We no longer do any munging here, but leaving here
+    # so that subclasses can leave super() in place for possible future needs.
+    def __post_serialize__(self, data):
+        return data
+
+    @classmethod
+    @functools.lru_cache
+    def json_schema(cls):
```
mikealfare, line 78: Are the remaining methods here (lines 78-130) for backwards compatibility due to removing hologram classes and related methods? For example, are we keeping `clean_dictionary_keys_for_validate` because there's a concern that folks might have defined that downstream?

tests/unit/test_graph.py
```diff
@@ -19,14 +19,16 @@
 from dbt.contracts.graph.manifest import MacroManifest, ManifestStateCheck
 from dbt.graph import NodeSelector, parse_difference
 from dbt.events.functions import setup_event_logger
+from dbt.flags import set_from_args
+from argparse import Namespace
```
mikealfare, line 23: nit: this should really be back up where it was with the built-in imports in the first section of the imports. Likewise, `from .utils ...` should be moved down here.

tests/unit/utils.py
```diff
@@ -199,6 +199,9 @@ def assert_from_dict(obj, dct, cls=None):
         obj_from_dict.created_at = 1
         obj.created_at = 1
 
+    if obj_from_dict != obj:
```
mikealfare, line 202: nit: This could arguably be put in the assert statement:
```python
assert obj_from_dict == obj, f"--- obj_from_dict: {obj_from_dict}\n--- obj: {obj}"
```

mikealfare (commented):
I have a few questions, no serious concerns, and multiple nits (take them or leave them, just things I noticed).

core/dbt/context/context_config.py
```diff
@@ -189,9 +189,21 @@ def initial_result(self, resource_type: NodeType, base: bool) -> C:
...
         translated = self._active_project.credentials.translate_aliases(partial)
-        return result.update_from(
+        translated = self.translate_hook_names(translated)
+        updated = result.update_from(
             translated, self._active_project.credentials.type, validate=validate
         )
+        return updated
+
+    def translate_hook_names(self, project_dict):
+        # This is a kind of kludge because the fix for #6411 specifically allowed misspelling
```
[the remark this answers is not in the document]
gshank: That ticket specifically allowed the "incorrect" spellings, so it's now a feature.

core/dbt/contracts/util.py
```diff
@@ -192,11 +195,12 @@ class VersionedSchema(dbtClassMixin):
     dbt_schema_version: ClassVar[SchemaVersion]
 
     @classmethod
-    def json_schema(cls, embeddable: bool = False) -> Dict[str, Any]:
-        result = super().json_schema(embeddable=embeddable)
-        if not embeddable:
-            result["$id"] = str(cls.dbt_schema_version)
-        return result
+    @functools.lru_cache
+    def json_schema(cls) -> Dict[str, Any]:
```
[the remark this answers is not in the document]
gshank: "embeddable" was a hologram thing and we're no longer using hologram. Anything passing in an invalid parameter would have failed in testing.

core/dbt/dataclass_schema.py
```diff
@@ -1,97 +1,133 @@
...
-
-        return dct
-
-    # This is called by the mashumaro _from_dict method, before
-    # performing the conversion to a dict
+    # This is called by the mashumaro from_dict in order to handle
+    # nested classes. We no longer do any munging here, but leaving here
+    # so that subclasses can leave super() in place for possible future needs.
     @classmethod
     def __pre_deserialize__(cls, data):
```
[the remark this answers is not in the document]
mikealfare: oof, ok

core/dbt/context/context_config.py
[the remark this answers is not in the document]
gshank: You'd have to ask product and Doug :).  If you want to open a ticket, go ahead. Not in scope for this one though...

core/dbt/contracts/graph/model_config.py
```diff
@@ -376,15 +374,6 @@ def finalize_and_validate(self: T) -> T:
         self.validate(dct)
         return self.from_dict(dct)
 
-    def replace(self, **kwargs):
```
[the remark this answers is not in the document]
gshank: This was removed because it did extra munging of the keys using the field_mapping, which is no longer needed. It's not currently used anywhere. If somebody wanted equivalent functionality they could use the "replace" function. I'm not a fan of the "Replaceable" class; I think it adds confusion and bloat and actually did a branch with a trial removal, but decided it wasn't worth it at that time. Since this wasn't used, I don't think the behavior with defaults mattered :)  But I suspect it was accidental.

core/dbt/context/context_config.py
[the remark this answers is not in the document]
jtcohen6: The misspelling here we mean is, we'll accept either kebab case or snake case for these two configs, in the several places they could be potentially defined:
- `post-hook` or `post_hook`
- `pre-hook` or `pre_hook`

core/dbt/dataclass_schema.py
```diff
@@ -1,97 +1,133 @@
...
+        validator = jsonschema.Draft7Validator(json_schema)
+        error = next(iter(validator.iter_errors(data)), None)
+        if error is not None:
+            raise ValidationError.create_from(error) from error
+
+    # This method was copied from hologram. Used in model_config.py and relation.py
+    @classmethod
+    def _get_fields(cls) -> List[Tuple[Field, str]]:
+        if cls._mapped_fields is None:
+            cls._mapped_fields = {}
```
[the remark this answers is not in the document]
gshank: I think the behavior might be subtly different with initializing to {} instead of None, but maybe not enough to make a difference. This was copied straight from hologram and is only used in a couple of places (which it might be good to revisit), but I'd rather leave this just as it is instead of introducing additional changes in a pull request that's already big.

core/dbt/dataclass_schema.py
```diff
@@ -1,97 +1,133 @@
...
+                    continue
+
+                # Note fields() doesn't resolve forward refs
+                f.type = type_hints[f.name]
+
+                # hologram used the "field_mapping" here, but we use the
+                # the field's metadata "alias". Since this method is mainly
+                # just used in merging config dicts, it mostly applies to
+                # pre-hook and post-hook.
+                field_alias = f.metadata.get("alias")
```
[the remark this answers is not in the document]
gshank: Okay

core/dbt/dataclass_schema.py
```diff
@@ -1,97 +1,133 @@
...
+                field_name = field_alias or f.name
+                mapped_fields.append((f, field_name))
+            cls._mapped_fields[cls.__name__] = mapped_fields
+        return cls._mapped_fields[cls.__name__]
+
+    # copied from hologram. Used in tests
+    @classmethod
+    def _get_field_names(cls):
+        fields = cls._get_fields()
+        field_names = []
```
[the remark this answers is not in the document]
gshank: Okay

tests/unit/test_graph.py
gshank: Removed the addition of set_from_args. It can be fixed with the ticket to do that to all of the tests.

tests/unit/utils.py
gshank: Removed.

core/dbt/parser/base.py
```diff
@@ -189,7 +189,7 @@ def _create_error_node(
         # message reasons
         return UnparsedNode(
             name=name,
-            resource_type=self.resource_type,
+            resource_type=self.resource_type,  # type: ignore
```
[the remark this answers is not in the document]
gshank: It was complaining that the type here didn't match the type defined in UnparsedNode. I changed the type in UnparsedNode to NodeType, because changing the other direction didn't work because of abtractmethod signatures in the base parser class.

core/dbt/dataclass_schema.py
gshank: Removed "clean_dictionary_keys_for_validate". I think it was an abandoned approach to some issue I've forgotten. The other methods are necessary and used in multiple places for validation, mostly of yaml dictionaries.

core/dbt/context/context_config.py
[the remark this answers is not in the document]
mikealfare: > The misspelling here we mean is, we'll accept either kebab case or snake case for these two configs

Agreed, I'm asking if we ever want to back out of that ditch, or support that for the foreseeable future.

tests/functional/artifacts/test_docs_generate_defer.py
```diff
@@ -29,5 +29,12 @@ def test_generate_defer(
...
 
         # defer test, it succeeds
-        results = run_dbt(["docs", "generate", "--state", "./state", "--defer"])
-        assert results.nodes["model.test.model"]
+        catalog = run_dbt(["docs", "generate", "--state", "./state", "--defer"])
+        assert catalog.nodes["model.test.model"]
+
+        # Check that catalog validates with jsonschema
+        catalog_dict = catalog.to_dict()
+        try:
```
mikealfare, line 37: I can't explain it, but this feels like an odd flow to me. Would something like this work?
```python
assert catalog.validate(catalog_dict), "Catalog validation failed"
```
or even
```python
assert catalog.validate(catalog.to_dict()), "Catalog validation failed"
```

tests/functional/sources/test_source_fresher_state.py
```diff
@@ -81,6 +82,10 @@ def _assert_freshness_results(self, path, state):
         with open(path) as fp:
             data = json.load(fp)
 
+        try:
```
mikealfare, line 85: Same comment as in `test_docs_generate_defer`.

peterallenwebb (approved):

PR #8449, opened 2023-08-17
ADAP-814: Add support for replacing materialized views with tables/views and vice versa

mikealfare:
### Problem

Materialized view support for dbt-postgres does not allow for replacing a materialized view with a table/view and vice versa.

### Solution

- Make the drop and rename statements generic across relations
- Dispatch to the relation-specific statement for drop/rename to allow for variations in the statement
- Allow a table/view to trigger the drop/rename of a materialized view
- Allow a materialized view to trigger the drop/rename of a table/view

### Checklist

- [x] I have read [the contributing guide](https://github.com/dbt-labs/dbt-core/blob/main/CONTRIBUTING.md) and understand what's expected of me  
- [x] I have run this code in development and it appears to resolve the stated issue  
- [x] This PR includes tests, or tests are not required/relevant for this PR
- [x] This PR has no interface changes (e.g. macros, cli, logs, json artifacts, config files, adapter interface, etc) or this PR has already received feedback and approval from Product or DX

PR #8459, opened 2023-08-18
ADAP-814: Refactor prep for MV updates

mikealfare:
### Problem

There is significant refactoring required for #8449, which will make reviewing that PR difficult. This is due to the fact that the new macros will cut across all relation types. Grouping macros by what they do and dispatching to relation-specific macros makes it easier to organize and follow logic flows.

### Solution

This PR will only move macros around. There will be no functional changes, no changes in signature, etc. This will leave #8449 with only the changes required to implement the change.

### Checklist


core/dbt/include/global_project/macros/adapters/relations/materialized_view/create.sql
```diff
@@ -1,9 +1,10 @@
 {% macro get_create_materialized_view_as_sql(relation, sql) -%}
-    {{- log('Applying CREATE to: ' ~ relation) -}}
```
colin-rogers-dbt, line 2: why delete this log?

mikealfare: It will get put back in in the general `get_create_sql` macro. I can leave it here though and then make the update with the other change.

colin-rogers-dbt (approved):

PR #8477, opened 2023-08-23
8295 unit testing artifacts

### LINKED ISSUE #8295: [CT-2922] Produce unit testing artifacts for metadata about unit testing
### Description

We should produce json artifacts which can be ingested and used as metadata about unit testing. There will be at least two types of artifacts, including a "result" artifact and a more manifest-like artifact.

The output in the run_result should organize test output by the unique_id of the model that they're running on, should include test result output, and a lightweight unit test structure.

Developing these artifacts will require settling on a "node-like" representation of unit tests, which will resemble the unparsed yaml objects, but have some additional data, but will not include everything in the UnitTestNode object.

### Acceptance criteria

We should have a run result structure and an artifact structure that we can share with other teams for feedback.

gshank:
resolves #8295

### Problem

Produce artifacts in the manifest (or elsewhere) with metadata about existing unit tests.

### Solution

A dictionary of UnitTestCases is stored in manifest.unit_tests.

Selection happens against a standard manifest (filtering for UnitTestCases in manifest.unit_tests), then the unit test manifest is built with UnitTestNodes in manifest.nodes plus input models.

This pull request also implements selection of a single unit test and use of "standard" selection syntax for models.

In addition it implements partial parsing for the manifest.unit_tests.

### Checklist


core/dbt/graph/selector.py
```diff
@@ -31,6 +31,8 @@ def can_select_indirectly(node):
     """
     if node.resource_type == NodeType.Test:
         return True
+    elif node.resource_type == NodeType.Unit:
+        return True
```
MichelleArk, line 35: What are the implications of this? If I do a `dbt run --select model` on `model` that has unit tests defined, do the unit tests actually get run by the run task? Or is it a no-op because the run manifest does not contain `UnitTestNode`s? Either way, I think we should leave unit test execution out of the `dbt run` functionality since that command is typically used in production jobs in addition to development ones.

MichelleArk, line 35: My understanding is that this is to power selection for the unit test task, but does not actually include unit tests in other task execution because `UnitTestNodes` are not in the manifest. Is that correct?

core/dbt/graph/selector_methods.py
```diff
@@ -431,19 +436,27 @@ def search(self, included_nodes: Set[UniqueId], selector: str) -> Iterator[Uniqu
...
     __test__ = False
 
     def search(self, included_nodes: Set[UniqueId], selector: str) -> Iterator[UniqueId]:
-        for node, real_node in self.parsed_nodes(included_nodes):
-            if real_node.resource_type == NodeType.Test and hasattr(real_node, "test_metadata"):
-                if fnmatch(real_node.test_metadata.name, selector):  # type: ignore[union-attr]
-                    yield node
+        command = get_command_name()
+        # Is this too kludgey?
+        if command == "unit-test":
```
MichelleArk, line 450: It may also be desirable to use this selector for other commands, such as `list`. Thinking through the issues with searching for elements in `manifest.unit_tests` in addition to `self.parsed_nodes` as opposed to conditionally based on the commands: 
* duplicate names across unit tests and other TestNode resources
  * this actually feels okay given we are yielding the full unique_id which will be disambiguated by resource type
* selecting nodes that aren't included in the graph queue for commands other than `unit-test`
  * is there a precedent for something like this? perhaps ephemeral models?
  * I think if we can handle this case gracefully by omitting selected unit test cases from the graph queue, this should be okay.

core/dbt/graph/selector.py
gshank: In this case the objects with NodeType.Unit are actually the UnitTestCases in manifest.unit_tests, which are in the manifest. They're not included in other tasks because each task has a Selector which filters out other resource types. So unit test cases aren't included in the "run" task job_queue because it only executes NodeType.Model.

core/dbt/graph/selector_methods.py
gshank: The UnitTestCases in manifest.unit_tests had to be included in the graph in order for selection to work. I have been thinking that it ought to work okay to include for other cases because of filter_selection (the task node_selector objects).

core/dbt/parser/manifest.py
```diff
@@ -1693,7 +1695,11 @@ def write_semantic_manifest(manifest: Manifest, target_path: str) -> None:
 
 
 def write_manifest(manifest: Manifest, target_path: str):
-    path = os.path.join(target_path, MANIFEST_FILE_NAME)
+    if get_command_name() == "unit-test":
+        file_name = UNIT_TEST_MANIFEST_FILE_NAME
+    else:
+        file_name = MANIFEST_FILE_NAME
+    path = os.path.join(target_path, file_name)
```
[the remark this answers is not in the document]
gshank: You mean set a field in the unit_test task?  "write_manifest" is called from the run method in runnable.py.

core/dbt/cli/main.py
```diff
@@ -879,14 +879,12 @@ def test(ctx, **kwargs):
 @requires.project
 @requires.runtime_config
 @requires.manifest
-@requires.unit_test_collection
```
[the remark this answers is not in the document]
gshank: A requires decorator runs before the task starts. The unit test manifest can't be created until after selection is handled, which happens during task.run.

core/dbt/parser/manifest.py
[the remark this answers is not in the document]
gshank: "write_manifest" is called in the outer_wrapper/wrapper method in core/dbt/cli/requires.py. It wouldn't have access to something set in the unit test task...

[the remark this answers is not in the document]
gshank: If we make the unit_test_manifest a different class of object, we could check that... Maybe we can defer this until after we look at that change?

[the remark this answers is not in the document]
MichelleArk: Perhaps we could use `flags.which` in the requires wrapper. It still is routing based on the command which doesn't feel great but in a less deeply nested part of the code. Looks like we're already doing something similar (accessing flags) in that wrapper as well: https://github.com/dbt-labs/dbt-core/blob/main/core/dbt/cli/requires.py#L252-L253

core/dbt/graph/selector_methods.py
MichelleArk: > thinking that it ought to work okay to include for other cases
same here - could we validate this by removing the `if` statement on the command and instead looping over both `manifest.unit_tests` and `self.parsed_nodes` (or parsed_and_unit_nodes) here instead?

core/dbt/parser/manifest.py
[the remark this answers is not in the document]
gshank: Actually it turns out that we don't want to write out the unit test manifest in the cli requires wrapper function, because that's the full parsed manifest.  We want it written out in the runnable.py code.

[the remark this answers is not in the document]
gshank: Done.

core/dbt/parser/manifest.py
```diff
@@ -1692,8 +1693,15 @@ def write_semantic_manifest(manifest: Manifest, target_path: str) -> None:
...
-def write_manifest(manifest: Manifest, target_path: str):
-    path = os.path.join(target_path, MANIFEST_FILE_NAME)
+def write_manifest(manifest: Manifest, target_path: str, which: Optional[str] = None):
+    if which and which == "unit-test":
+        file_name = UNIT_TEST_MANIFEST_FILE_NAME
+    else:
+        file_name = MANIFEST_FILE_NAME
+    import traceback
+
+    traceback.print_stack()
```
MichelleArk, line 1703: is this intentional?

gshank: Oops. Removed :)

core/dbt/parser/manifest.py
```diff
@@ -1692,8 +1693,15 @@ def write_semantic_manifest(manifest: Manifest, target_path: str) -> None:
     semantic_manifest.write_json_to_file(path)
 
 
-def write_manifest(manifest: Manifest, target_path: str):
-    path = os.path.join(target_path, MANIFEST_FILE_NAME)
+def write_manifest(manifest: Manifest, target_path: str, which: Optional[str] = None):
+    if which and which == "unit-test":
+        file_name = UNIT_TEST_MANIFEST_FILE_NAME
+    else:
+        file_name = MANIFEST_FILE_NAME
```
MichelleArk, line 1700: not blocking but: i think this method would be more generalized if the optional argument was `file_name` instead of `which` -- assuming its possible to plumb the UNIT_TEST_MANIFEST_FILE_NAME itself through from the requires decorator

gshank: We don't need it in the write_manifest call in the requires decorator. We need it in the write_manifest call from runnable.py. So we'd just have to move the "if" to the runnable "run" method, which we could certainly do, but doesn't seem like that much of an improvement.  And we'd still need an "if" in the write_manifest method to determine which file name to use unless we pass the file name in every place that we call "write_manifest" which feels like overkill.

MichelleArk (approved):

Issue #7911, opened 2023-06-20
[CT-2721] [Feature] support specify partial parsing file location


ChenyuLInx:
### Is this your first time submitting a feature request?

- [X] I have read the [expectations for open source contributors](https://docs.getdbt.com/docs/contributing/oss-expectations)
- [X] I have searched the existing issues, and I could not find an existing issue for this feature
- [X] I am requesting a straightforward extension of existing dbt functionality, rather than a Big Idea better suited to a discussion

### Describe the feature

All inputs for `dbt-core` should be configurable independently, with sane default when necessary. This makes running dbt-core on cloud easier to reason about. Because you can clearly configure all of your input for a `dbt` invocation.

Currently `dbt-core` will try to load up `partial_parse.msgpack` from the target dir([link](https://github.com/dbt-labs/dbt-core/blob/39e0c22353d2230fe1bc8b4852bbb0afa27fd2c0/core/dbt/parser/manifest.py#L949)). There's no way to configure it without configure target dir to the same place.

We should provide a parameter(for all commands that does parsing) to being able to specify location to load `partial_parse.msgpack`, and if not specified `dbt-core` will fall back to use the current target dir.

### Describe alternatives you've considered

Not do anything.

jtcohen6:
@ChenyuLInx Given how internal this is to `dbt-core`, I'd love if there were a way to support configuring this without another flag/parameter. Ideas:
- Support passing this into `dbtRunner` &rarr; `@requires.manifest` programmatically
- Support this as an env var (without associated CLI flag) - similar to [configurable location for hub / package registry](https://github.com/dbt-labs/dbt-core/blob/de65697ff9e0746e0b7dec39880381decebd3a85/core/dbt/clients/registry.py#L20-L23)

ChenyuLInx:
@jtcohen6  How does create [hidden parameter](https://github.com/dbt-labs/dbt-core/blob/4cdeff11cdeece7c6c28edb682477fd674cb9e04/core/dbt/cli/params.py#L272-L279) like what we do for `deprecated` params sounds?

Issue #6259, opened 2022-11-16
A general solution for accessing attributes in the flag object safely


ChenyuLInx:
### Is this your first time submitting a feature request?

- [X] I have read the [expectations for open source contributors](https://docs.getdbt.com/docs/contributing/oss-expectations)
- [X] I have searched the existing issues, and I could not find an existing issue for this feature
- [X] I am requesting a straightforward extension of existing dbt functionality, rather than a Big Idea better suited to a discussion

### Describe the feature

Right now in the click cli, we are trying to front load everything in [Flags](https://github.com/dbt-labs/dbt-core/blob/feature/click-cli/core/dbt/cli/main.py#L60) and then trying to use it at the root level CLI function. This makes accessing attributes defined in lower level cli functions possible, but this operation itself is not safe.

For example, right now if we try to run `docs` command with the click cli(`python core/dbt/cli/main.py docs`), we will run into error
```
  File "/Users/chenyuli/git/dbt-core/core/dbt/cli/main.py", line 70, in cli
    initialize_from_flags(flags.ANONYMOUS_USAGE_STATS, flags.PROFILES_DIR)
AttributeError: 'Flags' object has no attribute 'PROFILES_DIR'
```
since `PROFILES_DIR` is a parameter that is not defined in any of the command under `docs`, it never got set in `Flags`.

## potential Exit Criteria:
1. Rethink about whether construct flags at root level cli and do everything there is the method we want to go next
2. If answer to 1 is yes, then come up with solutions of making accessing potential non-existence attributes raise warning if there are cases that is not thought through(maybe through auto generated tests?). If the answer to 1 is no, then come up with new way to organize parameters and test that properly.

jtcohen6:
Thanks @ChenyuLInx! I added to the list in #5527

PR #8482, opened 2023-08-23
compile --suppress-ephemeral-ctes flag

### LINKED ISSUE #8480: [CT-3023] [Feature] compile --suppress-ephemeral-ctes

### Describe the feature

When linting with SQLFluff, it compiles the model to SQL (2x) and then linting is executed against the true SQL of the model, instead of the raw Jinja template.

The first compilation compiles the model as-is, and the second compilation is of a synthetic model built out of the Jinja blocks and expressions, and some tracking values in between. The result thereof is used to "trace" which parts of the compiled model come from what expressions, and to detect loops, etc.

When compiling a model with direct references to ephemeral models, the resulting injected CTEs violate some assumptions of the tracing and result in a failure to lint the model.

Proposal: for the `compile` command only, if a new flag `--suppress-ephemeral-ctes` is provided, the CTE injection does not happen and the expression referencing to the ephemeral model is replaced with a raw identifier. 

This resulting SQL is valid syntax, and allows the tracing to work as it is presently implemented.

I attempted to strip out the resulting CTEs from both compilations within the Cloud SQLFluff templater, but this was fraught with typical [regex peril](https://xkcd.com/1171/).

When using SQLFluff's CLI templater today, I believe it works around this by altering the manifest nodes for any ephemeral models to appear to be materialized to table instead, during the mid-lint compilation.

### Who will this benefit?

Cloud IDE usage of SQLFluff will immediately benefit from using this flag.

The SQLFluff maintainers may also be able to benefit from this flag in future versions of their CLI templater implementation.

yes! provided a proposed implementation in #8482 

### Anything else?

_No response_

benmosher:
...for inline compilation of models during linting only.

resolves #8480  

### Problem

Injection of ephemeral model CTEs breaks SQLFluff linting in Cloud.

### Solution

Optionally suppressing the injection of CTEs (intentionally added to the `compile` command only!) enables SQLFluff linting of models with direct ephemeral model references.

### Checklist

### DIFF
--- a/.changes/unreleased/Features-20230823-140407.yaml
+++ b/.changes/unreleased/Features-20230823-140407.yaml
@@ -1,0 +1,6 @@
+kind: Features
+body: Add --suppress-ephemeral-ctes flag for `compile` command, for usage by linting.
+time: 2023-08-23T14:04:07.617476-04:00
+custom:
+  Author: benmosher
+  Issue: "8480"
--- a/core/dbt/cli/params.py
+++ b/core/dbt/cli/params.py
@@ -40,6 +40,14 @@ compile_docs = click.option(
     default=True,
 )
 
+compile_suppress_ephemeral_ctes = click.option(
+    "--suppress-ephemeral-ctes/--no-suppress-ephemeral-ctes",
+    envvar="DBT_SUPPRESS_EPHEMERAL_CTES",
+    help="Internal flag for whether to supress injecting referenced ephemeral models' CTEs during `compile`.",
+    hidden=True,
+    default=False,
+)
+
 config_dir = click.option(
     "--config-dir",
     envvar=None,
--- a/core/dbt/compilation.py
+++ b/core/dbt/compilation.py
@@ -320,6 +320,10 @@ class Compiler:
         if model.compiled_code is None:
             raise DbtRuntimeError("Cannot inject ctes into an uncompiled node", model)
 
+        flags = get_flags()
+        if getattr(flags, "SUPPRESS_EPHEMERAL_CTES", False):
+            return (model, [])
+
         # extra_ctes_injected flag says that we've already recursively injected the ctes
         if model.extra_ctes_injected:
             return (model, model.extra_ctes)
--- a/core/dbt/cli/main.py
+++ b/core/dbt/cli/main.py
@@ -330,6 +330,7 @@ def docs_serve(ctx, **kwargs):
 @p.state
 @p.defer_state
 @p.deprecated_state
+@p.compile_suppress_ephemeral_ctes
 @p.target
 @p.target_path
 @p.threads
--- a/tests/functional/materializations/test_ephemeral_compilation.py
+++ b/tests/functional/materializations/test_ephemeral_compilation.py
@@ -1,3 +1,5 @@
+from dbt.contracts.graph.nodes import ModelNode
+from dbt.contracts.results import RunExecutionResult, RunResult
 import pytest
 from dbt.tests.util import run_dbt
 
@@ -53,6 +55,16 @@ models:
 """
 
 
+SUPPRESSED_CTE_EXPECTED_OUTPUT = """-- fct_eph_first.sql
+
+
+with int_eph_first as(
+    select * from __dbt__cte__int_eph_first
+)
+
+select * from int_eph_first"""
+
+
 class TestEphemeralCompilation:
     @pytest.fixture(scope="class")
     def models(self):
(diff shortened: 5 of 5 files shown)


dbeatty10:
@cla-bot check

benmosher:
@MichelleArk -- I see your :eyes:; I left this in draft because the last checkbox is unchecked but it's otherwise ready to go AFAIK. anything else I should do?

core/dbt/cli/params.py
```diff
@@ -40,6 +40,14 @@
     default=True,
 )
 
+compile_suppress_ephemeral_ctes = click.option(
+    "--suppress-ephemeral-ctes/--no-suppress-ephemeral-ctes",
```
MichelleArk, line 44: From review with @graciegoheen, we discussed naming this option more consistently with existing naming conventions, where the default behaviour (in this case - injecting ephemeral ctes) is effectively the flag name, which also has a negated version. For example `--partial-parse/--no-partial-parse`. 

In this case, we could consider: 
`--inject-ephemeral-ctes/--no-inject-ephemeral-ctes`

@benmosher - What do you think?

core/dbt/compilation.py
```diff
@@ -320,6 +320,10 @@ def _recursively_prepend_ctes(
         if model.compiled_code is None:
             raise DbtRuntimeError("Cannot inject ctes into an uncompiled node", model)
 
+        flags = get_flags()
```
MichelleArk, line 323: `self.config.args` is an alias to the same `flags` global object obtained via `get_flags` and is generally preferred to the global access via `get_flags` where possible. 

Similar to flags, self.config.args does not have a[ safe mechanism to access attributes that are not set in the context of the running command](https://github.com/dbt-labs/dbt-core/issues/6259). 

So: `self.config.args.suppress_ephemeral_ctes` would be unsafe in the context of a non-compiled command, but given this is part of the `Compiler` -- it feels like a relatively sane thing to do. Alternatively, we could still do `getattr(self.config.args.suppress_ephemeral_ctes, False)` as you've implemented so far.

MichelleArk:
@benmosher - with regards to that last checkbox, @graciegoheen will be helping out with this review from the Product perspective so we can get approval on that front before merging :)

core/dbt/compilation.py
MichelleArk: On second thought - `Compiler` is used in tasks that inherit from the compile task - not just the `compile` command. Which you've probably already realized :) 

Let's go with `getattr(self.config.args.suppress_ephemeral_ctes, False)` and add a comment with the tech_debt issue (#6259) for future reference/clean-up.

graciegoheen:
Let's hide the flag from the CLI `--help` text, since I think it will be confusing clutter for 99% of end users. Same as we did for https://github.com/dbt-labs/dbt-core/issues/7911

benmosher:
@graciegoheen agreed! I believe I've hidden it via [this `hidden=True`](https://github.com/dbt-labs/dbt-core/pull/8482/files#diff-f328aa466db7d10c562b1d34db497f911c05b6d1210988f127e07fa198a14763R47) 👍

core/dbt/cli/params.py
graciegoheen: I'd like to go with `--inject-ephemeral-ctes/--no-inject-ephemeral-ctes`, so that it's clear what the default is `--inject-ephemeral-ctes` vs. the override `--no-inject-ephemeral-ctes`. 

I worry with something like `--lint-mode`, folks may get confused that dbt is doing the actual linting.

core/dbt/compilation.py
```diff
@@ -320,6 +320,10 @@ def _recursively_prepend_ctes(
         if model.compiled_code is None:
             raise DbtRuntimeError("Cannot inject ctes into an uncompiled node", model)
 
+        # tech debt: safe flag/arg access (#6259)
+        if not getattr(self.config.args, "inject_ephemeral_ctes", True):
```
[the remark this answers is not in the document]
MichelleArk: Hmm, that's a fair point.. but I feel confident enough that we have automated testing that covers the ephemeral compilation that we'd detect issue(s) arising from this type of change

.changes/unreleased/Features-20230823-140407.yaml
```diff
@@ -0,0 +1,6 @@
+kind: Features
+body: Add --suppress-ephemeral-ctes flag for `compile` command, for usage by linting.
```
MichelleArk, line 2: nit: renamed to `--inject-ephemeral-ctes`

MichelleArk (approved):

benmosher:
I'd like to revalidate this in the `ide-server` before merging after changing the flag name. Should be fine since the test still passes but just want to be sure!

benmosher:
Revalidation complete 👍

MichelleArk:
🚢 🇮🇹

benmosher:
@MichelleArk I think you need to push the button -- I'm not authorized to merge to Core 😅

PR #8485, opened 2023-08-23
ADAP-845: Update BaseAdapter to meet mypy standards

### LINKED ISSUE #8396: [CT-2980] Fix untyped functions in adapters/base/ (mypy warning)
### Problem
When we hit a mypy error the console is flooded with mypy warnings as well.  Clean up the warning to improve developer experience. 

### Reproduction
Simply force a mypy error (change types) and the warnings will be generated.

### Relevant Output
```
core/dbt/adapters/base/query_headers.py:29: note: By default the bodies of untyped functions are not checked, consider using --check-untyped-defs  [annotation-unchecked]
core/dbt/adapters/base/impl.py:229: note: By default the bodies of untyped functions are not checked, consider using --check-untyped-defs  [annotation-unchecked]
```

mikealfare:
resolves #8396

### Problem

`mypy` throws a lot of warnings that should be addressed.

### Solution

Address the warnings by adding/updating types, checking types in code, or moving abstract methods towards the base class.

### Checklist

### DIFF
--- a/core/dbt/adapters/base/connections.py
+++ b/core/dbt/adapters/base/connections.py
@@ -400,7 +400,7 @@ class BaseConnectionManager(metaclass=abc.ABCMeta):
 
     @abc.abstractmethod
     def execute(
-        self, sql: str, auto_begin: bool = False, fetch: bool = False
+        self, sql: str, auto_begin: bool = False, fetch: bool = False, limit: Optional[int] = None
     ) -> Tuple[AdapterResponse, agate.Table]:
         """Execute the given SQL.
 
@@ -408,7 +408,28 @@ class BaseConnectionManager(metaclass=abc.ABCMeta):
         :param bool auto_begin: If set, and dbt is not currently inside a
             transaction, automatically begin one.
         :param bool fetch: If set, fetch results.
+        :param int limit: If set, limits the result set
         :return: A tuple of the query status and results (empty if fetch=False).
         :rtype: Tuple[AdapterResponse, agate.Table]
         """
         raise dbt.exceptions.NotImplementedError("`execute` is not implemented for this adapter!")
+
+    def add_select_query(self, sql: str) -> Tuple[Connection, Any]:
+        """
+        This was added here because base.impl.BaseAdapter.get_column_schema_from_query expects it to be here.
+        That method wouldn't work unless the adapter used sql.impl.SQLAdapter, sql.connections.SQLConnectionManager
+        or defined this method on <Adapter>ConnectionManager before passing it in to <Adapter>Adapter.
+
+        See https://github.com/dbt-labs/dbt-core/issues/8396 for more information.
+        """
+        raise dbt.exceptions.NotImplementedError(
+            "`add_select_query` is not implemented for this adapter!"
+        )
+
+    @classmethod
+    def data_type_code_to_name(cls, type_code: Union[int, str]) -> str:
+        """Get the string representation of the data type from the type_code."""
+        # https://peps.python.org/pep-0249/#type-objects
+        raise dbt.exceptions.NotImplementedError(
+            "`data_type_code_to_name` is not implemented for this adapter!"
+        )
--- a/core/dbt/adapters/base/impl.py
+++ b/core/dbt/adapters/base/impl.py
@@ -43,7 +43,7 @@ from dbt.exceptions import (
     UnexpectedNullError,
 )
 
-from dbt.adapters.protocol import AdapterConfig, ConnectionManagerProtocol
+from dbt.adapters.protocol import AdapterConfig
 from dbt.clients.agate_helper import empty_table, merge_tables, table_from_rows
 from dbt.clients.jinja import MacroGenerator
 from dbt.contracts.graph.manifest import Manifest, MacroManifest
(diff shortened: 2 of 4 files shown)


core/dbt/adapters/sql/connections.py
```diff
@@ -131,14 +131,6 @@ def get_result_from_cursor(cls, cursor: Any, limit: Optional[int]) -> agate.Tabl
 
         return dbt.clients.agate_helper.table_from_data_flat(data, column_names)
 
-    @classmethod
-    def data_type_code_to_name(cls, type_code: Union[int, str]) -> str:
```
mikealfare, line 135: This is already an abstract method. Move it up to ConnectionManager since BaseAdapter expects it to exist for some methods.

core/dbt/adapters/base/impl.py
```diff
@@ -315,14 +315,21 @@ def get_column_schema_from_query(self, sql: str) -> List[BaseColumn]:
 
     @available.parse(lambda *a, **k: ("", empty_table()))
     def get_partitions_metadata(self, table: str) -> Tuple[agate.Table]:
```
mikealfare, line 317: This should probably live in dbt-bigquery, based on the comment. But to preserve backwards compatibility, I checked that the method exists prior to calling it. This makes mypy happy while preserving backwards compatibility.

core/dbt/adapters/base/impl.py
```diff
@@ -453,9 +460,10 @@ def _relations_cache_for_schemas(
         # it's possible that there were no relations in some schemas. We want
         # to insert the schemas we query into the cache's `.schemas` attribute
         # so we can check it later
-        cache_update: Set[Tuple[Optional[str], Optional[str]]] = set()
+        cache_update: Set[Tuple[Optional[str], str]] = set()
         for relation in cache_schemas:
-            cache_update.add((relation.database, relation.schema))
+            if relation.schema:
```
mikealfare, line 465: mypy balked at relation.schema potentially not existing. That's because BaseRelation lists all of its parts as optional (though at least one must exist). In this scenario, we would expect schema to exist, as the type above suggests. Adding this check satisfies mypy and doesn't deviate much from the existing code.

core/dbt/adapters/base/impl.py
```diff
@@ -453,9 +460,10 @@ def _relations_cache_for_schemas(
         # it's possible that there were no relations in some schemas. We want
         # to insert the schemas we query into the cache's `.schemas` attribute
         # so we can check it later
-        cache_update: Set[Tuple[Optional[str], Optional[str]]] = set()
+        cache_update: Set[Tuple[Optional[str], str]] = set()
```
mikealfare, line 463: This type aligns with the change below and with what update_schemas expects (which is the only reason for this variable).

core/dbt/adapters/base/connections.py
```diff
@@ -400,15 +400,36 @@ def _add_query_comment(self, sql: str) -> str:
...
+        or defined this method on <Adapter>ConnectionManager before passing it in to <Adapter>Adapter.
+
+        See https://github.com/dbt-labs/dbt-core/issues/8396 for more information.
+        """
+        raise dbt.exceptions.NotImplementedError(
+            "`add_select_query` is not implemented for this adapter!"
+        )
+
+    @classmethod
+    def data_type_code_to_name(cls, type_code: Union[int, str]) -> str:
```
mikealfare, line 430: Moved up from SQLConnection.

core/dbt/adapters/base/connections.py
```diff
@@ -400,15 +400,36 @@ def _add_query_comment(self, sql: str) -> str:
...
         :param bool auto_begin: If set, and dbt is not currently inside a
             transaction, automatically begin one.
         :param bool fetch: If set, fetch results.
+        :param int limit: If set, limits the result set
         :return: A tuple of the query status and results (empty if fetch=False).
         :rtype: Tuple[AdapterResponse, agate.Table]
         """
         raise dbt.exceptions.NotImplementedError("`execute` is not implemented for this adapter!")
+
+    def add_select_query(self, sql: str) -> Tuple[Connection, Any]:
```
mikealfare, line 417: The comment says it all. This is overridden in SQLConnectionManager, but is used in BaseAdapter, which uses ConnectionManager. This is not expected to exist today, and the BaseAdapter method only works if this method is defined anyway.

core/dbt/adapters/base/connections.py
```diff
@@ -400,15 +400,36 @@ def _add_query_comment(self, sql: str) -> str:
 
     @abc.abstractmethod
     def execute(
-        self, sql: str, auto_begin: bool = False, fetch: bool = False
+        self, sql: str, auto_begin: bool = False, fetch: bool = False, limit: Optional[int] = None
```
mikealfare, line 403: This was leftover from the implementation of limit in a previous change. The type here matches that of the original change.

mikealfare:
> Looks good! Amazing how many changes were required just by adding a return type...

More than half of them were the protocol that you identified. I replaced it with the actual class (`ConnectionManager`). We can fix the protocol later since the only thing it specified was `TYPE`, which as you described it, was not very useful. I didn't need to import any new modules, just a new class from an existing module, so no concern for circular references.

Issue #8327, opened 2023-08-06
[CT-2944] [Bug] Ambiguous reference error for tests when model name is duplicated across packages


jtcohen6:
### Is this a new bug in dbt-core?

- [X] I believe this is a new bug in dbt-core
- [X] I have searched the existing issues, and I could not find an existing issue for this bug

### Current Behavior

When:
- a model with the same name is defined in multiple projects
- those projects are installed in the same runtime environment (via packages)
- at least one of those models has a generic test defined

dbt raises an ambiguous reference error

### Expected Behavior

dbt should successfully parse my project, and resolve tests to reference the model in the same project/package where they are defined

### Steps To Reproduce

Create a project named `my_dbt_project`, and another local project under a subdirectory named `mypkg/`.

Within the root project:
```yml
# dbt_project.yml
name: my_dbt_project
profile: default
```
```yml
# dependencies.yml OR packages.yml
packages:
  - local: mypkg
```
```sql
-- models/model_a.sql
select 1 as id
```
```yml
# models/config.yml
models:
  - name: model_a
    columns:
      - name: id
        tests:
          - unique
```

Within the package `mypkg`, add the exact same model with the exact same config:
```yml
# mypkg/dbt_project.yml
name: mypkg
```
```sql
-- mypkg/models/model_a.sql
select 1 as id
```
```yml
# mypkg/models/config.yml
models:
  - name: model_a
    columns:
      - name: id
        tests:
          - unique
```

From the root project:
```
$ dbt deps
14:10:31  Running with dbt=1.6.0
14:10:31  Installing mypkg
14:10:31  Installed from <local @ mypkg>

$ dbt --no-partial-parse parse
14:10:43  Running with dbt=1.6.0
14:10:43  Registered adapter: postgres=1.6.0
14:10:44  Encountered an error:
Compilation Error
  When referencing 'model_a', dbt found nodes in multiple packages: 'model.my_dbt_project.model_a', 'model.mypkg.model_a'
  To fix this, use two-argument 'ref', with the package name first: 'my_dbt_project' or 'mypkg'
```

Remove the `unique` tests, then parse again (this time successfully):
```
$ dbt --no-partial-parse parse
14:11:23  Running with dbt=1.6.0
14:11:24  Registered adapter: postgres=1.6.0
14:11:24  Performance info: /Users/jerco/dev/scratch/testy/target/perf_info.json
```

### Environment

```markdown
- OS: macOS Ventura 13.4.1
[...]

PR #7374, opened 2023-04-17
Allow duplicate refable node names across packages

MichelleArk:
resolves: https://github.com/dbt-labs/dbt-core/issues/7446
resolves: https://github.com/dbt-labs/dbt-core/issues/1269

### Description
* removes DuplicateResourceNameError call in manifest parsing
* uses _find_unique_ids_for_package in RefableLookup
* raises AmbiguousResourceNameRefError in RefableLookup.get_unique_id if multiple unique_ids found
* plumbs caller node (source_node) through RefableLookup for node-aware CompilationError message

### Checklist

- [x] I have read [the contributing guide](https://github.com/dbt-labs/dbt-core/blob/main/CONTRIBUTING.md) and understand what's expected of me
- [x] I have signed the [CLA](https://docs.getdbt.com/docs/contributor-license-agreements)
- [x] I have run this code in development and it appears to resolve the stated issue
- [x] This PR includes tests, or tests are not required/relevant for this PR
- [x] I have [opened an issue to add/update docs](https://github.com/dbt-labs/docs.getdbt.com/issues/new/choose), or docs changes are not required/relevant for this PR
- [x] I have run `changie new` to [create a changelog entry](https://github.com/dbt-labs/dbt-core/blob/main/CONTRIBUTING.md#adding-a-changelog-entry)

```
Compilation Error in model model_downstream (models/model_downstream.sql)
  dbt found multiple potential nodes when referencing "model" - ['model.project_b.model', 'model.project_a.model'].
  To fix this, specify which package to use as the first argument to 'ref' (one of: ['project_b', 'project_a']).
```

PR #8488, opened 2023-08-24
Fix ambiguous reference error for duplicate model names across packages with tests

MichelleArk:
resolves https://github.com/dbt-labs/dbt-core/issues/8327
resolves https://github.com/dbt-labs/dbt-core/issues/8493
[docs](https://github.com/dbt-labs/docs.getdbt.com/issues/new/choose) dbt-labs/docs.getdbt.com/#  N/A

### Problem
When generic test nodes are constructed, an 'attached_node' attribute is set by looking up the attached node in the manifest via `manifest.ref_lookup.get_unique_id`. Currently, this call passes `None` as the value for `package_name` -- resulting in ambiguous reference errors if the model exists with a duplicate name across packages. 

This is unexpected behaviour since the restriction to have unique model names across packages was lifted in https://github.com/dbt-labs/dbt-core/pull/7374.

Similarly, looking up the unique id from ref_lookup for versioned node parsing also passes in `None` for package_name, resulting in an ambiguous ref name error described in #8493.

### Solution

Pass the target package_name to ref lookup for attached_node, as generic test nodes should only ever be attached to nodes defined in the same package

### Checklist

### DIFF
--- a/.changes/unreleased/Fixes-20230824-161024.yaml
+++ b/.changes/unreleased/Fixes-20230824-161024.yaml
@@ -1,7 +1,7 @@
 kind: Fixes
-body: fix ambiguous reference error for tests when model name is duplicated across
+body: fix ambiguous reference error for tests and versions when model name is duplicated across
   packages
 time: 2023-08-24T16:10:24.437362-04:00
 custom:
   Author: michelleark
-  Issue: "8327"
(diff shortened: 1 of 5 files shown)


gshank (approved):
Looks like some unit tests need to be fixed, but otherwise looks good.

PR #8492, opened 2023-08-24
Include Compiled Node Attributes in run_results.json

### LINKED ISSUE #7519: [CT-2537] Include "compiled" node attributes in run results
### Proposed change

Let's include attributes that can vary at compile/runtime. Those attributes are:
- `compiled: bool = True` (always `False` after parsing)
- `compiled_code` (always empty after parsing) - this is **most** important
- `relation_name` - this is set after parsing, but it can be updated for tests that have `--store-failures` enabled
- `depends_on.macros` (https://github.com/dbt-labs/dbt-core/issues/5079) - since additional macro dependencies are required/registered while a model is being materialized, versus just compiled (e.g. `create_table_as`)

([Here](https://github.com/dbt-labs/dbt-core/blob/f1dddaa6e96b743498440925a55e66a7d5cdacfa/core/dbt/compilation.py#L361-L407) is where we set many of those attributes, during compilation)

The `run_results.json` produced by `compile`, `docs generate`, `run`/`build`/`test`/etc should include these fields (and only these fields) in addition to the node's `unique_id`. All other fields can be accessed from the manifest, and represent logical state at a point in time.

### Background

- Way back in v0.19, we removed the full node entry from `run_results.json`, and moved the `compiled_code` attribute into the manifest, so as to power `dbt-docs` from _only_ manifest + catalog.
- Starting in v1.4 (https://github.com/dbt-labs/dbt-core/pull/6384), we narrowed down the set of fields that differ between "uncompiled" and "compiled" nodes.

### Rationale

- The manifest represents "logical" state (parsed from project code + configs)
- Run results + catalog represent "applied" state (as materialized by dbt, as the objects exist in the data warehouse)

Compiled code can be a form of "applied" state. If a model's SQL depends on the results of an introspective query, it can vary given different inputs from the data warehouse.

### Acceptance Criteria

- The `run_results.json` produced by `compile`, `docs generate`, `run`, `build`, and `test` should include the `compiled`, `compiled_code`, `relation_name`, and `depends_on.macros` fields (and only these fields) in addition to the node's `unique_id`.
- The same information should also be included in the results returned by dbtRunner when is used to invoke dbt as a library.

peterallenwebb:
resolves #7519

### Checklist


gshank (commented):
tests/functional/invariants/invariants.py is kind of odd naming. Why "invariants"?  Why the doubled up name? Is there some reason you put "get_run_results" in that file instead of in dbt/tests/util.py along with "get_manifest"? 

Looks like artifacts tests need to be fixed...

peterallenwebb:
> tests/functional/invariants/invariants.py is kind of odd naming. Why "invariants"? Why the doubled up name? 

@gshank I agree this was awkwardly named. The conditions I wound up calling "assertions" were initially "invariants" [for nerdy reasons](https://en.wikipedia.org/wiki/Invariant_(mathematics)#Invariants_in_computer_science), but I thought calling them assertions was more friendly, and I should have made the naming consistent.

Let me know what you think of the new names I chose.

> Is there some reason you put "get_run_results" in that file instead of in dbt/tests/util.py along with "get_manifest"?

Only that it didn't occur to me. I've moved the code to tests/util.py.

> Looks like artifacts tests need to be fixed...

Fixed!

gshank (approved):
Looks good!

PR #8502, opened 2023-08-28
Semantic model configs - enable/disable + groups

### LINKED ISSUE #7968: [CT-2751] Support Configuration and Enabling/Disabling Semantic Models
**Acceptance Criteria**

- A `semantic_model` can be enabled and disabled by YAML configuration, both individually and at the project level.
- A `semantic_model` can be configured to belong to a `group`, both individually and at the project level, so that it can `ref` a private model within the same group.

```
# models/semantic.yml
```yml
semantic_models:
  - name: orders
    config:
      enabled: true | false
      group: some_group
```
```yml
# dbt_project.yml
semantic_models:
  my_project_name:
    +enabled: true | false
    +group: some_group
```

emmyoop:
resolves #7968 

### Problem

Semantic models cannot be enabled/disabled and also cannot belong to groups.

### Solution

Enable configs to allow semantic models to be disabled and be in groups.
When a semantic model is disabled but the defined metric for the measure is not, it throws a ParsingError.

### Checklist

### DIFF
--- a/.changes/unreleased/Features-20230828-092100.yaml
+++ b/.changes/unreleased/Features-20230828-092100.yaml
@@ -1,0 +1,6 @@
+kind: Features
+body: Support configuration of semantic models with the addition of enable/disable and group enablement.
+time: 2023-08-28T09:21:00.551633-05:00
+custom:
+  Author: emmyoop
+  Issue: "7968"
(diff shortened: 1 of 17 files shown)


QMalcolm (approved):
Thank you for doing this work! It's looking great 🙂 One thing that caught my attention is that we're handling disabled semantic models during resolution differently than we handle resolution of all other nodes. It confused me for awhile, and then I recognized that it was necessary due to our indirect referencing from the `measure` being the intermediary for the lookup (making the `DisabledLookup` not fit this use case).

It feels weird to me. Seems like we might need a refactoring here sometime, but I think that's out of scope for this.

PR #8306, opened 2023-08-03
update `Number` class to handle integer values

dave-connors-3:
resolves #8153

### Problem

Agate tables, which power `dbt show`, and therefore the dbt cloud IDE, show integer values as decimals, leading to user confusion when columns seem to have decimals when they don't! this overrides that behavior in the dbt-core customization of the agate `Number` class

### Solution

Ensure that the `Number.cast()` and `Number.jsonify()` methods properly handle integer values. 

### Checklist

- [x] I have read [the contributing guide](https://github.com/dbt-labs/dbt-core/blob/main/CONTRIBUTING.md) and understand what's expected of me  
- [x] I have run this code in development and it appears to resolve the stated issue  
- [x] This PR includes tests, or tests are not required/relevant for this PR
- [ ] This PR has no interface changes (e.g. macros, cli, logs, json artifacts, config files, adapter interface, etc) or this PR has already received feedback and approval from Product or DX

Issue #8153, opened 2023-07-19
[CT-2848] [Bug] `dbt show` adding decimal places to non-decimal values


dave-connors-3:
### Is this a regression in a recent version of dbt-core?

- [X] I believe this is a regression in dbt-core functionality
- [X] I have searched the existing issues, and I could not find an existing issue for this regression

### Current Behavior

`dbt show` is adding decimal places to numeric/integer values incorrectly. This is causing problems for users in the dbt Cloud IDE, as the IDE uses that command to render the preview tab. 

### Expected/Previous Behavior

Previously (we think!) this was not an issue, and decimal places were properly handled by dbt show

### Steps To Reproduce

1. use dbt-snowflake>=1.5.0
2. create a model in your project:

```sql
# in my_model.sql
select
  cast(1 as numeric) as _numeric,
  cast(1 as integer) as _integer,
  cast(1 as decimal) as _decimal
```
3. run `dbt show -s my_model --output JSON`
4. get this!
```zsh
❯ dbt show -s my_model --output json
20:33:15  Running with dbt=1.5.3
20:33:16  Registered adapter: snowflake=1.5.2
20:33:16  Unable to do partial parsing because of a version mismatch
20:33:17  Found 3 models, 4 tests, 0 snapshots, 0 analyses, 436 macros, 0 operations, 0 seed files, 0 sources, 0 exposures, 0 metrics, 0 groups
20:33:17  
20:33:18  Concurrency: 8 threads (target='dev')
20:33:18  
20:33:19  {
  "node": "my_model",
  "show": [
    {
      "_NUMERIC": 1.0,
      "_INTEGER": 1.0,
      "_DECIMAL": 1.0
    }
  ]
}
```

### Environment

```markdown
- OS: mac
- Python: Python 3.9.16
- dbt (working version): unknown
- dbt (regression version): 1.5.3
```

### Which database adapter are you using with dbt?

snowflake

dave-connors-3:
did some more local testing on other versions of 1.5 and observed the same behavior - -this may not be a regression, but just a regular bug!

jtcohen6:
Copying from #8167 (great minds think alike @dave-connors-3):

## What's going on?

We're just using the `agate` library's type system and JSON serialization:

https://github.com/dbt-labs/dbt-core/blob/eeb057085c19301f9803d1620d79645f27d501b5/core/dbt/task/show.py#L78-L83

[...]

PR #8525, opened 2023-08-30
Fix untyped functions in `core/dbt/context/base.py`

### LINKED ISSUE #8394: [CT-2978] Fix untyped functions in context/base.py (mypy warning)
### Problem
When we hit a mypy error the console is flooded with mypy warnings as well.  Clean up the warning to improve developer experience. 

### Reproduction
Simply force a mypy error (change types) and the warnings will be generated.

### Relevant Output
```
core/dbt/context/base.py:188: note: By default the bodies of untyped functions are not checked, consider using --check-untyped-defs  [annotation-unchecked]

```

QMalcolm:
resolves #8394 

### Problem

We have a lot of untyped functions in `core/dbt/context/base.py` this means that
1. There is a lot of type checking that is not happening
2. When a type issue occurs in the file a bunch of `can't type check untyped functions` warnings appear

### Solution
ADD TYPES TO ALL THE THINGS 🎉 

### Checklist

### DIFF
--- a/core/dbt/context/base.py
+++ b/core/dbt/context/base.py
@@ -181,13 +181,17 @@ class Var:
 
 
 class BaseContext(metaclass=ContextMeta):
-    # subclass is TargetContext
-    def __init__(self, cli_vars):
-        self._ctx = {}
-        self.cli_vars = cli_vars
-        self.env_vars = {}
+    # Set by ContextMeta
+    _context_members_: Dict[str, Any]
+    _context_attrs_: Dict[str, Any]
 
-    def generate_builtins(self):
+    # subclass is TargetContext
+    def __init__(self, cli_vars: Dict[str, Any]) -> None:
+        self._ctx: Dict[str, Any] = {}
+        self.cli_vars: Dict[str, Any] = cli_vars
+        self.env_vars: Dict[str, Any] = {}
+
+    def generate_builtins(self) -> Dict[str, Any]:
         builtins: Dict[str, Any] = {}
         for key, value in self._context_members_.items():
             if hasattr(value, "__get__"):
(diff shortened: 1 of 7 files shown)


QMalcolm:
I'm very confused on how the functions [contextmember](https://github.com/dbt-labs/dbt-core/blob/f94fdadde0f63ff1bccedcffb28e3b8e6a6441b4/core/dbt/context/base.py#L101) and [contextproperty](https://github.com/dbt-labs/dbt-core/blob/f94fdadde0f63ff1bccedcffb28e3b8e6a6441b4/core/dbt/context/base.py#L107) operate. The are function decorators, and in python function decorators are required to return a callable. However... in cases where the `value` passed into the decorator isn't a string, we return an object (`ContextMember`). Thus typing these two functions is neigh impossible because currently we have to state that one of the return types must be a bare `ContextMember` object, which then every invocation of the decorator complains about.

gshank (approved):

McKnight-42:
Out of curiosity is this a different way of approaching the change we reverted here https://github.com/dbt-labs/dbt-core/pull/8306?

QMalcolm:
> Out of curiosity is this a different way of approaching the change we reverted here #8306?

@McKnight-42 I'm not wholly sure, but I don't think so. I could absolutely be wrong about that though. My understanding of issue #8153 is that all numeric values were being forced into floats, and the solution (#8306) was to actually add an `Integer` object for creating integer numeric types. In this PR we're not adding or changing any functionality of the code itself (with the exception of [a613ef0](https://github.com/dbt-labs/dbt-core/pull/8525/commits/a613ef09dbd5de68c529acc2352c2a9ae8266883)). The primary focus of this PR is defining what the expected return type of a function should be. That way when we run mypy, it can properly check all of our typing. These function return type definitions are actually thrown away at run time and don't actually affect the operation of the code.

McKnight-42:
ah makes sense thanks for clarification @QMalcolm

PR #8536, opened 2023-08-31
More Type Annotations

### LINKED ISSUE #8537: [CT-3066] Add More Type Annotations
This is an umbrella issue for opportunistically adding more type annotations to our code.

### Acceptance criteria

There are more type annotations than previously.

peterallenwebb:
resolves #8537

Add more type annotations in the events module, and to __init__ methods more widely.

### Checklist


core/dbt/events/base_types.py
```diff
@@ -100,9 +100,9 @@ def to_dict(self):
             self.pb_msg, preserving_proto_field_name=True, including_default_value_fields=True
         )
 
-    def to_json(self):
+    def to_json(self) -> str:
         return MessageToJson(
-            self.pb_msg, preserving_proto_field_name=True, including_default_valud_fields=True
+            self.pb_msg, preserving_proto_field_name=True, including_default_value_fields=False
```
peterallenwebb, line 105: The addition of type annotations here uncovered a typo in the named parameter. Because the named parameter was previously defaulting to False, I have changed the supplied value from True to False in order to retain existing behavior. @gshank Perhaps you'll have an opinion on that?

gshank: Oops.  I suspect that the only types of places that this would matter are in strings set to '' and ints set to 0. Mostly structured fields have just been used for constructing messages so they haven't been used so much so far, but if people do start using them more we probably want them. I'd set it to True...

PR #8538, opened 2023-08-31
fix: override materialization python models

### LINKED ISSUE #8520: [CT-3049] [Bug] Unable to override default materialization for python models (dbt >1.4)/Hierarchical Configuration Not Applied

### Current Behavior

When working within my company's workflow, we outline all configurations, metadata, etc. in the `schema.yml` file, with broader configurations being described in the `dbt_projects.yml`. However, the materialization of Python models cannot be controlled through these files. This requires us to set the configuration within the model itself, as demonstrated below:

```python
def model(dbt, session):
    dbt.config(materialization='incremental')
    ...
```

If this isn't done, the materialization always defaults to 'table' in the generated `manifest.json`.

commit that introduces this behavior: https://github.com/dbt-labs/dbt-core/blame/48d04e81417195393af5af1f78ef695f3398f193/core/dbt/parser/base.py#L217

### Expected Behavior

I would expect that the materialization of Python models could be controlled through these configuration files, just as it is observed with SQL models. Ideally, without having to specify the materialization within the model itself. When defining configurations in these files, they should be respected, and the materialization shouldn't default to 'table' in the generated manifest.json unless explicitly set to do so. The current behavior introduces an inconsistency in the way dbt models are configured and managed.

### Steps To Reproduce

1. Ensure you are using dbt version 1.4 or newer.

2. Create a Python model with a materialization setting other than 'table' specified either in dbt_projects.yml or schema.yml.
3. Execute dbt run. Observe in the terminal that the materialization is displayed as 'table'.
4. Upon inspecting the manifest.json, you'll find that all other configurations are respected, except for the materialization which defaults to 'table'.

### Relevant log output

_No response_

### Environment

```markdown
- OS: Ubuntu 20.04.6 LTS
- Python: 3.10.9
- dbt: (tested with multiple versions: 1.4, 1.5, and 1.6.1)
```

spark, other (mention it in "Additional Context")

### Additional Context

database adatpter: dbt-databricks

commit that introduces this behavior: https://github.com/dbt-labs/dbt-core/blame/48d04e81417195393af5af1f78ef695f3398f193/core/dbt/parser/base.py#L217

I've some workarounds to fix this. If this issue is accepted I can send a PR.

### LINKED ISSUE #9703: [Bug] Custom materialization not respected when passed via dbt_project.yaml or model.yaml

### Current Behavior

models/dummy/test.py
```python
def model(dbt, session):
    dbt.config(materialized="test_mat")
    return None
```

dbt_project.yml
```yaml
...
models:
  dummy:
    test:
      +materialized: test_mat
```

When the `dbt.config(materialized="test_mat")` is removed from models/dummy/test.py the materialization code is not called

### Expected Behavior

materialized config is respected when specified in dbt_project.yml and model.yml not only in the actual model

### Steps To Reproduce

https://gist.github.com/rbs392/5e9d4a29aed63e385839c4f354ce0b56

### Relevant log output

```shell
when run with materialized config from dbt_project.yml

12:43:33  Running with dbt=1.7.9
12:43:34  Registered adapter: databricks=1.7.8
12:43:34  Unable to do partial parsing because a project config has changed
12:43:35  Found 1 model, 0 sources, 0 exposures, 0 metrics, 536 macros, 0 groups, 0 semantic models
12:43:35
12:43:36  Concurrency: 1 threads (target='dev')
12:43:36
12:43:36  1 of 1 START python table model dummy.test ...................................... [RUN]
12:43:36  Databricks adapter: Job submission response={'id': '1079e76b1aa64c14b84d407a13fbfe4f'}
12:43:47  1 of 1 ERROR creating python table model dummy.test ............................. [ERROR in 10.99s]
12:43:47
12:43:47  Finished running 1 table model in 0 hours 0 minutes and 12.51 seconds (12.51s).
12:43:47
12:43:47  Completed with 1 error and 0 warnings:
```

when run with materialized set in model file
```
12:44:03  Running with dbt=1.7.9
12:44:05  Registered adapter: databricks=1.7.8
12:44:05  Found 1 model, 0 sources, 0 exposures, 0 metrics, 536 macros, 0 groups, 0 semantic models
12:44:05
12:44:05  Concurrency: 1 threads (target='dev')
12:44:05
12:44:05  1 of 1 START python test_mat model dummy.test ................................... [RUN]
12:44:05  1 of 1 ERROR creating python test_mat model dummy.test .......................... [ERROR in 0.04s]
12:44:06
12:44:06  Finished running 1 test_mat model in 0 hours 0 minutes and 1.40 seconds (1.40s).
12:44:06
12:44:06  Completed with 1 error and 0 warnings:
```

### Environment

```markdown
- OS: Mac os
- Python: 3.9.5
- dbt: 1.7.9
- dbt-databricks: 1.7.8
```

### Additional Context

dbt-databricks

devmessias:
resolves #8520  , resolves #9703

#### Current Behavior
When working within my company's workflow, we outline all configurations, metadata, etc. in the schema.yml file, with broader configurations being described in the dbt_projects.yml. However, the materialization of Python models cannot be controlled through these files. This requires us to set the configuration within the model itself, as demonstrated below:

```python
def model(dbt, session):
    dbt.config(materialization='incremental')
```

If this isn't done, the materialization always defaults to 'table' in the generated manifest.json.

commit that introduces this behavior: https://github.com/dbt-labs/dbt-core/blame/48d04e81417195393af5af1f78ef695f3398f193/core/dbt/parser/base.py#L217

#### Expected Behavior

I would expect that the materialization of Python models could be controlled through these configuration files, just as it is observed with SQL models. Ideally, without having to specify the materialization within the model itself. When defining configurations in these files, they should be respected, and the materialization shouldn't default to 'table' in the generated manifest.json unless explicitly set to do so. The current behavior introduces an inconsistency in the way dbt models are configured and managed.

### Solution

I've considered three approaches to address this bug. Firstly, I'll outline the two solutions I discarded, followed by an explanation of why I chose the third.

#### First Solution: (discarded)

Remove the behavior introduced by commit 48d04e8141719539.

While this is a straightforward, it could break many existing pipelines. This would be unfavorable for many dbt users because it's a behavior available since v1.4.

#### Second Solution: (discarded)

The proposal here is to add a condition where the  default materialization value would only used if it hasn't been predefined in a .yml file or within the model's .py file itself.

```python
# core/dbt/parser/base.py:L192
if block.path.relative_path.endswith(".py"):
    language = ModelLanguage.python
    config.add_config_call({"_dbt_default_arg_materialized": "table"})
```

```python
# core/dbt/context/context_config.py:L140
if patch_config_dict:
    result = self._update_from_config(result, patch_config_dict)

default_args = [k for k in config_call_dict if k.startswith("_dbt_default_arg_")]
result_keys = dir(result)
for key in default_args:
    config_key = key.replace("_dbt_default_arg_", "")
    if config_key in result_keys:
        del config_call_dict[key]
    else:
        config_call_dict[config_key] = config_call_dict[key]
```
However, this approach has two major issues:

1- The solution is messy and has a high cognitive load.
2- If the configuration for this materialization isn't defined in the model's .py or any .yml file, it defaults to creating a "view" materialization. This is because the NodeConfig class defaults to "view" as its materialization.
https://github.com/dbt-labs/dbt-core/blob/e5e1a272ffbb1072332b7f2d71e31b805296ddcb/core/dbt/contracts/graph/model_config.py#L438-L441

It's just a hypothesis, but I believe this might be the very reason for commit 48d04e8141719539 since "view" is not supported for Python models.

#### Third Solution (preferred):

In base.py, the proposal is to add a key-value pair to the config_call_dict.

https://github.com/devmessias/dbt-core/blob/21696c70c5c3c22003f4282405df9fad65a96ce5/core/dbt/parser/base.py#L215-L224,

Parameters described in _dbt_node_type_configs are used to overwrite the initial node configuration:

 https://github.com/devmessias/dbt-core/blob/21696c70c5c3c22003f4282405df9fad65a96ce5/core/dbt/context/context_config.py#L136

Since we don't want the "_dbt_node_type_configs" to persist, I opted to remove it in the latest update.

https://github.com/devmessias/dbt-core/blob/21696c70c5c3c22003f4282405df9fad65a96ce5/core/dbt/context/context_config.py#L152-L156

### Checklist

### DIFF
--- a/core/dbt/context/context_config.py
+++ b/core/dbt/context/context_config.py
@@ -168,6 +168,7 @@ class BaseContextConfigGenerator(Generic[T]):
             for fqn_config in self._active_project_configs(fqn, resource_type):
                 result = self._update_from_config(result, fqn_config)
 
+        # pytest.set_trace()
         # this is mostly impactful in the snapshot config case
         # TODO CT-211
         return result  # type: ignore[return-value]
--- a/core/dbt/context/context_config.py
+++ b/core/dbt/context/context_config.py
@@ -92,8 +92,9 @@ class RenderedConfig(ConfigSource):
 
 
 class BaseContextConfigGenerator(Generic[T]):
-    def __init__(self, active_project: RuntimeConfig):
+    def __init__(self, active_project: RuntimeConfig, language: Optional[ModelLanguage] = None):
         self._active_project = active_project
+        self._language = language
 
     def get_config_source(self, project: Project) -> ConfigSource:
         return RenderedConfig(project)
@@ -184,15 +185,16 @@ class BaseContextConfigGenerator(Generic[T]):
 
 
 class ContextConfigGenerator(BaseContextConfigGenerator[C]):
-    def __init__(self, active_project: RuntimeConfig):
+    def __init__(self, active_project: RuntimeConfig, language: Optional[ModelLanguage] = None):
         self._active_project = active_project
+        self._language = language
 
     def get_config_source(self, project: Project) -> ConfigSource:
         return RenderedConfig(project)
 
     def initial_result(self, resource_type: NodeType, base: bool) -> C:
         # defaults, own_config, config calls, active_config (if != own_config)
-        config_cls = get_config_for(resource_type, base=base)
+        config_cls = get_config_for(resource_type, base=base, language=self._language)
         # Calculate the defaults. We don't want to validate the defaults,
         # because it might be invalid in the case of required config members
         # (such as on snapshots!)
@@ -306,6 +308,7 @@ class ContextConfig:
         *,
         rendered: bool = True,
         patch_config_dict: Optional[dict] = None,
+        language: Optional[ModelLanguage] = None,
     ) -> Dict[str, Any]:
         if rendered:
             # TODO CT-211
--- a/core/dbt/context/context_config.py
+++ b/core/dbt/context/context_config.py
@@ -168,7 +168,6 @@ class BaseContextConfigGenerator(Generic[T]):
             for fqn_config in self._active_project_configs(fqn, resource_type):
                 result = self._update_from_config(result, fqn_config)
 
-        # pytest.set_trace()
         # this is mostly impactful in the snapshot config case
         # TODO CT-211
         return result  # type: ignore[return-value]
--- a/core/dbt/context/context_config.py
+++ b/core/dbt/context/context_config.py
@@ -92,9 +92,8 @@ class RenderedConfig(ConfigSource):
 
 
 class BaseContextConfigGenerator(Generic[T]):
-    def __init__(self, active_project: RuntimeConfig, language: Optional[ModelLanguage] = None):
+    def __init__(self, active_project: RuntimeConfig):
         self._active_project = active_project
-        self._language = language
 
     def get_config_source(self, project: Project) -> ConfigSource:
         return RenderedConfig(project)
@@ -148,6 +147,7 @@ class BaseContextConfigGenerator(Generic[T]):
         patch_config_dict: Optional[Dict[str, Any]] = None,
     ) -> BaseConfig:
         own_config = self.get_node_project(project_name)
+
         result = self.initial_result(resource_type=resource_type, base=base)
 
         project_configs = self._project_configs(own_config, fqn, resource_type)
@@ -185,16 +185,15 @@ class BaseContextConfigGenerator(Generic[T]):
 
 
 class ContextConfigGenerator(BaseContextConfigGenerator[C]):
-    def __init__(self, active_project: RuntimeConfig, language: Optional[ModelLanguage] = None):
+    def __init__(self, active_project: RuntimeConfig):
         self._active_project = active_project
-        self._language = language
 
     def get_config_source(self, project: Project) -> ConfigSource:
         return RenderedConfig(project)
 
     def initial_result(self, resource_type: NodeType, base: bool) -> C:
         # defaults, own_config, config calls, active_config (if != own_config)
-        config_cls = get_config_for(resource_type, base=base, language=self._language)
+        config_cls = get_config_for(resource_type, base=base)
         # Calculate the defaults. We don't want to validate the defaults,
         # because it might be invalid in the case of required config members
         # (such as on snapshots!)
@@ -308,7 +307,6 @@ class ContextConfig:
         *,
         rendered: bool = True,
         patch_config_dict: Optional[dict] = None,
-        language: Optional[ModelLanguage] = None,
     ) -> Dict[str, Any]:
         if rendered:
             # TODO CT-211
--- a/core/dbt/context/context_config.py
+++ b/core/dbt/context/context_config.py
@@ -147,8 +147,9 @@ class BaseContextConfigGenerator(Generic[T]):
         patch_config_dict: Optional[Dict[str, Any]] = None,
     ) -> BaseConfig:
         own_config = self.get_node_project(project_name)
-
         result = self.initial_result(resource_type=resource_type, base=base)
+        if "_dbt_node_type_configs" in config_call_dict:
+            result = self._update_from_config(result, config_call_dict["_dbt_node_type_configs"])
 
         project_configs = self._project_configs(own_config, fqn, resource_type)
         for fqn_config in project_configs:
@@ -162,6 +163,9 @@ class BaseContextConfigGenerator(Generic[T]):
 
         # config_calls are created in the 'experimental' model parser and
         # the ParseConfigObject (via add_config_call)
+        config_call_dict = {
+            k: config_call_dict[k] for k in config_call_dict.keys() - {"_dbt_node_type_configs"}
+        }
         result = self._update_from_config(result, config_call_dict)
 
         if own_config.project_name != self._active_project.project_name:
@@ -193,6 +197,7 @@ class ContextConfigGenerator(BaseContextConfigGenerator[C]):
 
     def initial_result(self, resource_type: NodeType, base: bool) -> C:
         # defaults, own_config, config calls, active_config (if != own_config)
+        # aqui vai puxar o node config com default view
         config_cls = get_config_for(resource_type, base=base)
         # Calculate the defaults. We don't want to validate the defaults,
         # because it might be invalid in the case of required config members
--- a/core/dbt/context/context_config.py
+++ b/core/dbt/context/context_config.py
@@ -148,8 +148,6 @@ class BaseContextConfigGenerator(Generic[T]):
     ) -> BaseConfig:
         own_config = self.get_node_project(project_name)
         result = self.initial_result(resource_type=resource_type, base=base)
-        if "_dbt_node_type_configs" in config_call_dict:
-            result = self._update_from_config(result, config_call_dict["_dbt_node_type_configs"])
 
         project_configs = self._project_configs(own_config, fqn, resource_type)
         for fqn_config in project_configs:
@@ -163,9 +161,6 @@ class BaseContextConfigGenerator(Generic[T]):
 
         # config_calls are created in the 'experimental' model parser and
         # the ParseConfigObject (via add_config_call)
-        config_call_dict = {
-            k: config_call_dict[k] for k in config_call_dict.keys() - {"_dbt_node_type_configs"}
-        }
         result = self._update_from_config(result, config_call_dict)
 
         if own_config.project_name != self._active_project.project_name:
@@ -197,7 +192,6 @@ class ContextConfigGenerator(BaseContextConfigGenerator[C]):
 
     def initial_result(self, resource_type: NodeType, base: bool) -> C:
         # defaults, own_config, config calls, active_config (if != own_config)
-        # aqui vai puxar o node config com default view
         config_cls = get_config_for(resource_type, base=base)
         # Calculate the defaults. We don't want to validate the defaults,
         # because it might be invalid in the case of required config members
--- a/core/dbt/contracts/graph/model_config.py
+++ b/core/dbt/contracts/graph/model_config.py
@@ -58,9 +58,17 @@ RESOURCE_TYPES: Dict[NodeType, Type[BaseConfig]] = {
 BASE_RESOURCE_TYPES: Dict[NodeType, Type[BaseConfig]] = RESOURCE_TYPES.copy()
 
 
-def get_config_for(resource_type: NodeType, base=False) -> Type[BaseConfig]:
+def get_config_for(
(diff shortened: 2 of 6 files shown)


devmessias:
Hi @gshank,

Could I please receive feedback on this PR? One aspect I'm unable to understand is why the codecov is reducing the percentage of coverage. This PR did not introduce any branch conditions inside the code.  I've noticed a similar decrease in some other PRs as well; should I be concerned about this at the moment?

gshank:
I think the original "fix" (setting the materialization in the config_call_dict) was wrong, but I think this approach just doubles down on that wrong approach. This is a tricky area. I suspect that what's needed is some special setting in "update_parsed_node_config" in core/dbt/parser/base.py.  The problem is detecting the difference between somebody actually setting the materialization vs it just defaulting. I'll have to poke around a bit more to see if I can find a good way to do this.

devmessias:
> I think the original "fix" (setting the materialization in the config_call_dict) was wrong, but I think this approach just doubles down on that wrong approach. This is a tricky area. I suspect that what's needed is some special setting in "update_parsed_node_config" in core/dbt/parser/base.py. The problem is detecting the difference between somebody actually setting the materialization vs it just defaulting. I'll have to poke around a bit more to see if I can find a good way to do this.

Thank you for your feedback, @gshank. I agree that this change may not be the ideal one.

It's a great idea to bring @aranke into this discussion, since they were responsible for the original commit that introduced this behavior.

**Note:**  As I mentioned in the PR, it seems to me that the solution from @aranke is easier to reconcile the differences between Python and SQL models. I've tried to rethink the behavior that @aranke introduced and try to place it elsewhere but couldn't, maybe there really isn't a cleaner solution yet, especially since it's a behavior that breaks the convention of all versions post 1.4.

Could you give us your opinion, @aranke?

devmessias:
A big issue is that removing the default table behavior could break numerous pipelines, as views are not supported in Python models.

devmessias:
One option I am considering, given that @aranke only wanted to change the default materialization for Python (so I assume there is no point in considering future default changes), and as @gshank suggested it to be something in `update_parsed_node_config`, we could simply check if the language in parsed_node within the same function is a Python model and at the same time verify if the materialization is set as view, then change it to table. Since this is what causes the issue. What do you think? It would also have the advantage of being much simpler.

gshank:
@devmessias I agree with your proposed solution. In update_parsed_node_config check that the language is python and that materialization is not set in the config dict, and if both of those are true, set it in the config dict.

devmessias:
I've updated the PR with the proposed solution @gshank. Could you review the changes to ensure we're on the same page?

devmessias:
Ah, I forgot to mention one thing. The default is set before the update_node_config. As a result, the following test created by @aranke , which was passing in my previous suggestion, won't pass now. Would it be better to remove this test and create a new one that checks only after the update_node_config?


image

gshank:
Yes, that would be better.  If you can find a way to do it in unit tests, that would be fine, but it might be easier to just look at the manifest for the python node in a functional test.

devmessias:
Hi @gshank. I've managed to determine how to retain the unit test.
My previous approach was somewhat messy as it also had to manage `config.build_config_dict(patch_config_dict=patch_config_dict)` to address all issues within `update_parsed_node_config`. However, I've modified `update_parsed_node_config` to pass the model language as an argument to `ContextConfigGenerator`, thereby creating an appropriate NodeConfig for the Python model. I believe this method is superior. Could you please review my new commit and share your thoughts?

devmessias:
Hi @aranke . What do you think about this new approach?

devmessias:
Hi @gshank, just circling back on this pull request. Would appreciate your feedback when it's convenient for you. Thanks!

devmessias:
Hi @dbeatty10, @gshank and @aranke, any updates on this bug/MR? I'm not sure about the dbt-core roadmap, or if bug #8520 was resolved in another MR, because if so, I can close this PR.

dbeatty10:
@graciegoheen and @peterallenwebb when you take a look at this, could you also check if it resolves https://github.com/dbt-labs/dbt-core/issues/9703?

The reprex is here: https://github.com/dbt-labs/dbt-core/issues/9703#issuecomment-1979983639

core/dbt/contracts/graph/model_config.py
```diff
@@ -658,9 +663,17 @@ def finalize_and_validate(self):
...
+) -> Type[BaseConfig]:
     if base:
         lookup = BASE_RESOURCE_TYPES
     else:
         lookup = RESOURCE_TYPES
-    return lookup.get(resource_type, NodeConfig)
+
+    resource = lookup.get(resource_type, NodeConfig)
+
+    if language == "python" and (resource == NodeConfig):
```
peterallenwebb, line 676: Let's use the enum value here instead of a string literal.

core/dbt/context/context_config.py
```diff
@@ -82,8 +82,9 @@ def get_config_dict(self, resource_type: NodeType) -> Dict[str, Any]:
 
 
 class BaseContextConfigGenerator(Generic[T]):
-    def __init__(self, active_project: RuntimeConfig):
+    def __init__(self, active_project: RuntimeConfig, language: Optional[ModelLanguage] = None):
```
peterallenwebb, line 85: @gshank @ChenyuLInx Does this seem like the best way to do this? It seems odd to me that language would be the one element that needs to be passed separately.

tests/unit/test_parser.py
```diff
@@ -973,7 +973,10 @@ def test_python_model_parse(self):
             raw_code=python_model,
             checksum=block.file.checksum,
             unrendered_config={"materialized": "table", "packages": python_packages},
-            config_call_dict={"materialized": "table", "packages": python_packages},
+            config_call_dict={
```
peterallenwebb, line 976: In order to accept this PR, we will need an effective test.

peterallenwebb (dismissed):

core/dbt/context/context_config.py
ChenyuLInx: @devmessias Thanks for the contribution and engaging discussion, sorry we have been a behind on this patch.

I have the same feeling when reading this code.
I can see the appeal of having a proper PythonNode with default materialization being `table`.
But because this node is buried deep under layers functions, I don't think having this language passed through is ideal.
I would recommend going back to what Gerda initially proposed ([link](https://github.com/dbt-labs/dbt-core/commit/8f95385262d24ca68e3328e5dcdd7daf7061bc0a))

> In update_parsed_node_config check that the language is python and that materialization is not set in the config dict, and if both of those are true, set it in the config dict.

That would make this PR easier to follow. What do you think?

devmessias: Hi, thank you for your feedback

It’s been quite some time since I worked on this, but if memory serves me right, the approach you’re referring to is indeed what I implemented in this commit: https://github.com/dbt-labs/dbt-core/pull/8538/commits/8f95385262d24ca68e3328e5dcdd7daf7061bc0a. However, this change introduced another bug which was caught by a unit test, as documented here: [unit test issue comment](https://github.com/dbt-labs/dbt-core/pull/8538#issuecomment-1744871154)

devmessias: I believe the need to have a Python node is because the default materialization type in dbt is not supported for Python models (at least when I made this PR). This issue was not apparent before because it was being obscured by a materialization bug.

ChenyuLInx: I think this the unit test is failing because the config_dict generated by `config_dict = config.build_config_dict(patch_config_dict=patch_config_dict)` is what's being added to the parsed node in the end.
And I see why you have the current solution which is creating a place to define a default node.

I think passing the language through multiple layers of calls is not ideal, but if we want to do something better, we have two ways:
- do some large refactors to have this config consolidation function moved to a node and that would give us an easier way to define a default materialization on a python model node
- slightly twist the patch_config_dict(or create a new version of it) with any config that is different from the default default, and apply it. We should add some comments to it also. (I personally prefer this for the current case, and we can refactor later on if we see a pattern developing here).


So this would look like adding the following logic [here](https://github.com/dbt-labs/dbt-core/commit/8f95385262d24ca68e3328e5dcdd7daf7061bc0a#diff-0afffd02471ddca604b25dd73c0b0c411bafab7cb5a48f3d640a3cc9e613c7ceR327), and update the comment a bit.
```
if not patch_config_dict:
    patch_config_dict = {}
if parsed_node.resource_type == NodeType.Model and parsed_node.language == ModelLanguage.python:
    if "materialized" not in patch_config_dict:
        patch_config_dict["materialized"] = "table"
```

@devmessias @gshank what do you think?

ChenyuLInx: @devmessias are you still interested in making an update to this? No worries if not, we can take it across the finish line. Thanks either way for putting so much effort into this!

devmessias: Yes, I’m still interested even though I’ve switched jobs. Now it’s a matter of honor :laughing: . I will make the modifications this week and do a rebase.

core/dbt/contracts/graph/model_config.py
dbeatty10: Here's the enum:

https://github.com/dbt-labs/dbt-core/blob/a7af3b383139851d4ad6f4e3859cdd961330a384/core/dbt/artifacts/resources/types.py#L54-L56

Here's an example of using it:

https://github.com/dbt-labs/dbt-core/blob/a7af3b383139851d4ad6f4e3859cdd961330a384/core/dbt/parser/base.py#L25

https://github.com/dbt-labs/dbt-core/blob/a7af3b383139851d4ad6f4e3859cdd961330a384/core/dbt/parser/base.py#L218

core/dbt/context/context_config.py
graciegoheen: Hi @devmessias just wanted to check if you're still interested in completing this PR!

tests/unit/test_parser.py
devmessias: Yes! The issue was that `test_python_model_custom_materialization` was testing with `materialized = view`, which is the default value, so nothing was actually tested.  I have updated the PR

devmessias:
ping everyone!

core/dbt/context/context_config.py
```diff
@@ -168,6 +168,7 @@ def calculate_node_config(
             for fqn_config in self._active_project_configs(fqn, resource_type):
                 result = self._update_from_config(result, fqn_config)
 
+        # pytest.set_trace()
```
ChenyuLInx, line 171: Nit: probably left from testing?

ChenyuLInx (approved):
Thanks for adding the patch!!

core/dbt/context/context_config.py
ChenyuLInx: Took it out, hope you don't mind

PR #8539, opened 2023-08-31
ADAP-869: Support atomic replace in replace macro

mikealfare:
resolves #  

### Problem

The current replace macro does not support replacing an existing relation with a target relation of the same type using an atomic "create and replace". Instead it tries to do the stage/backup/swap method which uses at least two statements (one to drop the existing and one to create the target).

### Solution

Update the replace macro to first check if the types are the same, and if so, if the type is a replaceable type. Some relation types are not replaceable (e.g. materialized views in Postgres). If this check fails, then head down the stage/backup/swap route.

### Checklist


mikealfare:
Since postgres does not use replace in this way yet (it's not implemented for tables and views and materialized views don't support create or replace), it was tested against `dbt-snowflake`. The tests that used to use drop/create methods are now using create or replace.

core/dbt/include/global_project/macros/materializations/models/materialized_view/get_materialized_view_configuration_changes.sql
mikealfare, line 1: Moved to `relations/materialized_view/alter.sql`

core/dbt/include/global_project/macros/relations/materialized_view/_replace.sql
mikealfare, line 1: Replaced with `relations/materialized_view/replace.sql`, should have shown as a rename. That file was then updated to what shows below.

core/dbt/include/global_project/macros/materializations/models/view/helpers.sql
mikealfare, line 1: Contents were moved to `relations/view/replace.sql`.

plugins/postgres/dbt/include/postgres/macros/relations/materialized_view/_replace.sql
mikealfare, line 1: Replaced with `relations/materialized_view/replace.sql`, should have shown as a rename. The contents of that file were then updated to the below.

plugins/postgres/dbt/include/postgres/macros/relations/view/replace.sql
mikealfare, line 1: This was taken from the default create view statement in `dbt-core`. Postgres did not overwrite it. It was then updated to run a `create or replace` instead of just `replace` and whitespace was adjusted for readability; otherwise it was left untouched.

plugins/postgres/dbt/include/postgres/macros/materializations/materialized_view.sql
mikealfare, line 1: Contents of this file were moved to `relations/materialized_view/alter.sql`.

McKnight-42 (approved):

PR #8548, opened 2023-09-05
Copy dir if symlink fails (#7447)

aranke:
Backport for 1.6.2

### DIFF
--- a/core/dbt/deps/local.py
+++ b/core/dbt/deps/local.py
@@ -51,19 +51,15 @@ class LocalPinnedPackage(LocalPackageMixin, PinnedPackage):
         src_path = self.resolve_path(project)
         dest_path = self.get_installation_path(project, renderer)
 
-        can_create_symlink = system.supports_symlinks()
-
         if system.path_exists(dest_path):
             if not system.path_is_symlink(dest_path):
                 system.rmdir(dest_path)
             else:
                 system.remove_file(dest_path)
-
-        if can_create_symlink:
+        try:
             fire_event(DepsCreatingLocalSymlink())
             system.make_symlink(src_path, dest_path)
-
-        else:
+        except OSError:
             fire_event(DepsSymlinkNotAvailable())
             shutil.copytree(src_path, dest_path)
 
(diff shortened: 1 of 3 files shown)


core/dbt/deps/local.py
```diff
@@ -51,19 +51,15 @@ def install(self, project, renderer):
...
             else:
                 system.remove_file(dest_path)
-
-        if can_create_symlink:
+        try:
             fire_event(DepsCreatingLocalSymlink())
             system.make_symlink(src_path, dest_path)
-
-        else:
+        except OSError:
```
jtcohen6, line 62: Could someone explain this logic to me? It doesn't seem quite right.

Previously: if the system does not [support symlinks](https://github.com/dbt-labs/dbt-core/blob/main/core/dbt/clients/system.py#L161), we would use `copytree` instead of calling `make_symlink`.

Now, we _always_ try to call `make_symlink` first. If it encounters an `OSError` (e.g. due to permissions), then we fall back to calling `copytree`.

But within [`make_symlink`](https://github.com/dbt-labs/dbt-core/blob/main/core/dbt/clients/system.py#L150-L158), we first check to see if the system supports symlinks, and we raise a [`SymbolicLinkError`](https://github.com/dbt-labs/dbt-core/blob/7eedfcd2742fcf789200a4b829c8c2eb98369089/core/dbt/exceptions.py#L718) if it does not. This exception is a subclass of `CompilationError`, _not_ `OSError`, so I don't think we'd fall back to calling `copytree`!

We could update this `except` to:
```python
    except OSError or SymbolicLinkError:
```
I think that might do the trick? I haven't tested locally at all, just trying to read through the relevant codepaths.

jtcohen6 (changes requested):
One blocking comment/question before we backport this

core/dbt/deps/local.py
jtcohen6: Ok - I think you're right!

https://github.com/dbt-labs/dbt-core/blob/7eedfcd2742fcf789200a4b829c8c2eb98369089/core/dbt/clients/system.py#L161-L162

Per Python docs, it sounds like [`os.symlink`](https://docs.python.org/3/library/os.html#os.symlink) is supported on both Unix + Windows.

I like your idea of removing this code, and just allowing the more generic `OSError` to flow through. Do you think it make more sense to do that as a separate change in the `main` branch, and keep the backported change here more narrowly scoped?

PR #8551, opened 2023-09-05
Fix #8544: Parse the correct schema version from manifest

### LINKED ISSUE #8544: [Bug] failed to resolve schema version

### Current Behavior

The latest `get_manifest_schema_version` will get wrong versions and drop the metrics data:

https://github.com/dbt-labs/dbt-core/blob/v1.6.1/core/dbt/contracts/graph/manifest.py

```py
def get_manifest_schema_version(dct: dict) -> int:
    schema_version = dct.get("metadata", {}).get("dbt_schema_version", None)
    if not schema_version:
        raise ValueError("Manifest doesn't have schema version")
    return int(schema_version.split(".")[-2][-1])
```

This input will get `0`

```
https://schemas.getdbt.com/dbt/manifest/v10.json
```

And it will make scheam upgrader to drop metrics data
https://github.com/dbt-labs/dbt-core/blob/v1.6.1/core/dbt/contracts/graph/manifest_upgrade.py#L65-L69

the 

### Expected Behavior

The `metrics` data should be kept

### Steps To Reproduce

1. use the dbt 1.6 project with metrics (https://github.com/InfuseAI/jaffle-sl-template)
2. create the state directory `dbt parse --target-path baseline`
3. modify some metrics criteria
4. select modified and nothing selected `dbt list -s 'state:modified' --state baseline`

image

### Relevant log output

_No response_

### Environment

```markdown
- OS: OSX
- Python: 3.10.6
- dbt:


Core:
  - installed: 1.6.1
  - latest:    1.6.1 - Up to date!

Plugins:
  - snowflake: 1.6.1 - Update available!
  - duckdb:    1.6.0 - Up to date!

```
```


### Additional Context

_No response_

### PR DESCRIPTION
**Pull request by aranke**
resolves #8544 

### Checklist

### DIFF
--- a/.changes/unreleased/Under the Hood-20230906-164901.yaml
+++ b/.changes/unreleased/Under the Hood-20230906-164901.yaml
@@ -1,0 +1,6 @@
+kind: Under the Hood
+body: Add typing to __init__ in base.py
+time: 2023-09-06T16:49:01.150713+01:00
+custom:
+  Author: aranke
+  Issue: "8398"
--- a/core/dbt/task/base.py
+++ b/core/dbt/task/base.py
@@ -193,7 +193,7 @@ class ExecutionContext:
 
 
 class BaseRunner(metaclass=ABCMeta):
-    def __init__(self, config, adapter, node, node_index, num_nodes):
+    def __init__(self, config, adapter, node, node_index, num_nodes) -> None:
         self.config = config
         self.adapter = adapter
         self.node = node

### CODE REVIEW

File: core/dbt/contracts/graph/manifest.py [hunk 1]
```diff
@@ -1528,7 +1528,8 @@ def get_manifest_schema_version(dct: dict) -> int:
     schema_version = dct.get("metadata", {}).get("dbt_schema_version", None)
     if not schema_version:
         raise ValueError("Manifest doesn't have schema version")
-    return int(schema_version.split(".")[-2][-1])
+
+    return int(schema_version.split("/")[-1].split(".")[0][1:])
```
QMalcolm, line 1532: You've explained to me how this works, and it looks good. We should probably add a comment here documenting the format / an example of the string we're expecting. That will make it more obvious why this specific sequence of parsing actually does what we want it to do 🙂

QMalcolm (approved):

PR #8578, opened 2023-09-06
Unskip and rename test_expression_metric

### LINKED ISSUE #8134: [CT-2837] Stop skipping `test_expression_metric`
We currently `@pytest.mark.skip` [test_expression_metric](https://github.com/dbt-labs/dbt-core/blob/867534c1f49a3daccd6aae83dd32772173479f9d/tests/functional/metrics/test_metric_helper_functions.py#L20). We should at this time be able to un-skip this test with some minor tweaks. Additionally, if the test is broken for real reasons, this issue should be converted to fixing the functionality as well.

## Acceptance Criteria

- `test_expression_metric` is no longer skipped
- `test_expression_metric` renamed to `test_derived_metric`
- `test_derived_metric` passes (and not because it's rigged)

QMalcolm:
resolves #8134 

### Problem

Previously we were skipping `TestMetricHelperFunctions.test_expression_metric` leaving all the helper metric helper functions untested. Additionally because this test was being skipped, we didn't realize all the helper functions were broken 😬 

### Solution

Correct, simplify, and add typing to the metric helper functions. Unskip `TestMetricHelperFunctions.test_expression_metric` and rename it to `TestMetricHelperFunctions.test_derived_metric`.

### Note
I'm not a huge fan of the `parent_metrics` functions and dependency tree functions returning the originating metric. That seems unnecessary. It'd also simplify the code to not return the originating function. However, as I don't know who depends on these functions, I didn't want to alter their behavior unexpectedly.

### Checklist

### DIFF
--- a/core/dbt/contracts/graph/metrics.py
+++ b/core/dbt/contracts/graph/metrics.py
@@ -66,12 +66,11 @@ class ResolvedMetricReference(MetricReference):
         """
         if metric_node.type in DERIVED_METRICS:
             yield {metric_node.name: metric_depth_count}
-            metric_depth_count = metric_depth_count + 1
 
-        for parent_unique_id in metric_node.depends_on.nodes:
-            metric = manifest.metrics.get(parent_unique_id)
-            if metric is not None and metric.type in DERIVED_METRICS:
-                yield from cls.reverse_dag_parsing(metric, manifest, metric_depth_count)
+            for parent_unique_id in metric_node.depends_on.nodes:
+                metric = manifest.metrics.get(parent_unique_id)
+                if metric is not None:
+                    yield from cls.reverse_dag_parsing(metric, manifest, metric_depth_count + 1)
 
     def full_metric_dependency(self):
         """Returns a unique list of all upstream metric names."""
--- a/core/dbt/contracts/graph/metrics.py
+++ b/core/dbt/contracts/graph/metrics.py
@@ -31,6 +31,7 @@ class ResolvedMetricReference(MetricReference):
 
     @classmethod
     def parent_metrics(cls, metric_node, manifest):
+        """For a given metric, yeilds all upstream metrics."""
         yield metric_node
 
         for parent_unique_id in metric_node.depends_on.nodes:
@@ -40,6 +41,7 @@ class ResolvedMetricReference(MetricReference):
 
     @classmethod
     def parent_metrics_names(cls, metric_node, manifest):
+        """For a given metric, yeilds all upstream metric names"""
         yield metric_node.name
 
         for parent_unique_id in metric_node.depends_on.nodes:
@@ -49,6 +51,10 @@ class ResolvedMetricReference(MetricReference):
 
     @classmethod
     def reverse_dag_parsing(cls, metric_node, manifest, metric_depth_count):
+        """For the given metric, yeilds dictionaries having {<metric_name>: <depth_from_initial_metric} of upstream derived metrics.
+
+        This function is intended as a helper function for other metric helper functions.
+        """
         if metric_node.calculation_method == "derived":
             yield {metric_node.name: metric_depth_count}
             metric_depth_count = metric_depth_count + 1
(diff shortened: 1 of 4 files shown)


tests/functional/metrics/test_metric_helper_functions.py
```diff
@@ -3,30 +3,35 @@
...
         self,
         project,
     ):
 
         # initial parse
-        run_dbt(["compile"])
+        run_dbt(["parse"])
 
         # make sure all the metrics are in the manifest
         manifest = get_manifest(project.project_root)
```
MichelleArk, line 33: small nit: I think we can remove this line in favor of grabbing the manifest from the result of `run_dbt(["parse"])`, e.g. `manifest = run_dbt(["parse"])`

core/dbt/contracts/graph/metrics.py
```diff
@@ -17,7 +24,7 @@ class ResolvedMetricReference(MetricReference):
     for working with metrics (ie. __str__ and templating functions)
     """
 
-    def __init__(self, node, manifest, Relation):
+    def __init__(self, node: Metric, manifest: Manifest, Relation=None):
```
MichelleArk, line 27: nit: `Relation` is not typed. we could also include the `-> None` return type here to avoid mypy warnings such as the ones cleaned up in https://github.com/dbt-labs/dbt-core/issues/8387.

Related, but not introduced in this PR, so could be addressed in a follow-on: it feels pretty nonstandard to have this variable name be capitalized - it looks like this is propagated from `db_wrapper.Relation` in `RuntimeMetricResolver` -- perhaps we could rename to `relation_cls` instead?

tests/functional/metrics/test_metric_helper_functions.py
QMalcolm: 👍 Will do!

core/dbt/contracts/graph/metrics.py
QMalcolm: Happy to type `Relation`  and rename it , as well as add the `-> None:`. Interestingly `Relation` isn't used at all in this class and thus far it seems dbt-core instantiations of this class pass `None` as the third variable. I wanted to rip it out very badly while looking at this code, but didn't want to break any third parties instantiating the class for whatever reason.

core/dbt/contracts/graph/metrics.py
```diff
@@ -17,76 +24,74 @@ class ResolvedMetricReference(MetricReference):
...
+    def parent_metrics(cls, metric_node: Metric, manifest: Manifest) -> Iterator[Metric]:
+        """For a given metric, yeilds all upstream metrics."""
         yield metric_node
 
         for parent_unique_id in metric_node.depends_on.nodes:
-            node = manifest.metrics.get(parent_unique_id)
-            if node and node.resource_type == NodeType.Metric:
-                yield from cls.parent_metrics(node, manifest)
+            metric = manifest.metrics.get(parent_unique_id)
+            if metric is not None:
```
MichelleArk, line 45: is there a valid scenario where it'd be valid for a unique_id to exist in `metric_node.depends_on.nodes` but not be in `manifest.metric`? If not, we could handle this less gracefully by getting metric from `metric = manifest.expect(parent_unique_id)` to ensure an issue is not silently handled.

QMalcolm: `depends_on` contains metrics and things that aren't metrics (like semantic models). So if we were to use `manifest.expect(...)` we'd have to bring back checking if the returned node is a metric.

MichelleArk: I see! but once we've got what we know is a metric -- should we expect it to ever be `None`? If not, we could assert/raise an internal error here to make future investigation easier

MichelleArk (approved):

core/dbt/contracts/graph/metrics.py
QMalcolm: Oooo expect does a look up against all the manifest.x objects. I though this would be a bad runtime cost, but these are dictionaries not lists, so doing an `in` lookup wouldn't be too costly. I'll switch to that as you suggested. That function handles the `None` case, and thus all we need to do on return is check if it's a metric node.

PR #8638, opened 2023-09-12
Fix #8160: Warn when --state == --target

### LINKED ISSUE #8160: [Feature] Improve the error message for `dbt clone` when source = destination

### Describe the feature

Mentioned by @aranke [here](https://github.com/dbt-labs/docs.getdbt.com/pull/3742#discussion_r1263839942) and [here](https://github.com/dbt-labs/docs.getdbt.com/pull/3742#discussion_r1267161625):

> Something else to note here is that you cannot use the same target directory for both the source and target

> You get a very weird error that source and target states are the same, and so it's a no-op.
> 
> I ran into this during local development, so think this warning needs to be surfaced loudly.

_No response_

### Who will this benefit?

_No response_

_No response_

### Anything else?

_No response_

aranke:
resolves #8160 

A/C from ticket:

- [x] Raise a warning whenever --state path == --target-path
- [x] Update logic of create_or_replace_clone as described

### Checklist

### DIFF
--- a/core/dbt/events/types.py
+++ b/core/dbt/events/types.py
@@ -1252,7 +1252,8 @@ class WarnStateTargetEqual(WarnLevel):
 
     def message(self) -> str:
         return yellow(
-            f"Warning: The state and target directories are the same: '{self.state_path}'"
+            f"Warning: The state and target directories are the same: '{self.state_path}'. "
+            f"This could lead to missing changes due to overwritten state including non-idempotent retries."
         )
 
 


core/dbt/events/types.py
```diff
@@ -1247,6 +1246,16 @@ def message(self) -> str:
         )
 
 
+class WarnStateTargetEqual(WarnLevel):
+    def code(self):
+        return "I072"
+
+    def message(self) -> str:
+        return yellow(
+            f"Warning: The state and target directories are the same: '{self.state_path}'"
```
MichelleArk, line 1255: I like this message because its actionable/hints at how the user could prevent the warnings, but I do think it's missing something to explain why this is a warning in the first place. Could we work in "what could go wrong/unexpectedly because state and target directories are the same" to the message? 

My understanding from the issue is the potential for missing changes in state-based selection or retries that are not idempotent

gshank (approved):

PR #8647, opened 2023-09-13
Loosen type on replaceable_relation and renameable_relation and provide guidance in docstrings

mikealfare:
### Problem

The typing on `renameable_relations` and `replaceable_relations` is overly prescriptive. The only requirement is that the relations in this class variable are contained in an object that defines `__contains__` and that said object is immutable (frozen dataclass). `Tuple` and `FrozenSet` are both valid examples, and an adapter maintainer may prefer one or the other, or some third option.

### Solution

Update the type to `Iterable[str]` and make updates in the methods to account for `None`.

### Checklist

### DIFF
--- a/core/dbt/adapters/base/relation.py
+++ b/core/dbt/adapters/base/relation.py
@@ -1,6 +1,6 @@
 from collections.abc import Hashable
 from dataclasses import dataclass, field
-from typing import Optional, TypeVar, Any, Type, Dict, Iterator, Tuple, Set, Iterable
+from typing import Optional, TypeVar, Any, Type, Dict, Iterator, Tuple, Set, Union, FrozenSet
 
 from dbt.contracts.graph.nodes import SourceDefinition, ManifestNode, ResultNode, ParsedNode
 from dbt.contracts.relation import (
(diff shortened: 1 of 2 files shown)


core/dbt/adapters/base/relation.py
```diff
@@ -291,11 +300,11 @@ def create(
 
     @property
     def can_be_renamed(self) -> bool:
-        return self.type in self.renameable_relations
+        return self.type in self.renameable_relations if self.type else False
 
     @property
     def can_be_replaced(self) -> bool:
-        return self.type in self.replaceable_relations
+        return self.type in self.replaceable_relations if self.type else False
```
colin-rogers-dbt, line 307: not sure we need this, if `self.type is None` it should evaluate to `False` anyways

colin-rogers-dbt (approved):
one nit but otherwise LGTM

core/dbt/adapters/base/relation.py
mikealfare: I was appeasing the `mypy` overlord. However, I no longer need it since I backed off of `Iterable` to `Union[Tuple, FrozenSet]`. I did that because a general Iterable is not serializable (for mashumaro).

PR #8653, opened 2023-09-15
ADAP-850: Support test results as a view

### LINKED ISSUE #6914: [CT-2070] [Feature] Materialized tests
This one is about making sure tests stay relevant as we expand into low latency scenarios with [Materialized Views](https://github.com/dbt-labs/dbt-core/issues/6911) (MVs).

Prior art can be found in [`dbt-materialize`](https://github.com/MaterializeInc/materialize/blob/main/misc/dbt-materialize/dbt/include/materialize/macros/materializations/test.sql). More details on the benefits [here](https://materialize.com/blog/real-time-data-quality-tests-using-dbt-and-materialize/).

Let's say you materialized a model with a MV, this is `agg_daily_active_users`:

```YAML

{{
    config(
        materialized='materialized_view',
        auto_refresh=true
    )
}}

select
    date_trunc('day', event_at) as date_day,
    count(distinct user_id) as daily_active_users

from raw_app_data.events
```

I'm defining a [test](https://docs.getdbt.com/docs/build/tests#singular-tests) to ensure data quality:

```SQL
select
    date_day,
    daily_active_users,
    current_date() as test_date
from {{ ref('agg_daily_active_users' )}}
where daily_active_users < 0
```

Currently, when `dbt run` is triggered (Wednesday the 8th, 2023 at 4pm Pacific Time), dbt will execute the underlying SQL query and returns the result in the log.

|date_day|daily_active_users|test_date|
|-|-|-|
|2023-02-07|-2|2023-02-08|

Ok I know about this one, there was some returns that were processed in a bizarre way, it should be solved in the coming day so I'm happy to ignore it.

That's excellent when you know the data is not moving when dbt is not looking. But what happens when it is, like it does with MVs? Then I want my tests to be run continuously. What I want is to make that test queryable so I can get the current status of my MV

I could run `dbt test --select agg_daily_active_users` instead. But then I'm forced to switch context between where my data is flowing (my database engine, via the materialized view) and where my test results appear. The best experience for me is to stay in the database for that.

I could run `dbt test --select agg_daily_active_users --store-failures` instead. But that may be slow and expensive, while making the operation a 2 step process (refresh the table, query the table).

What would be nice is to tell dbt that I want my test query materialized as a view (or a MV) instead, via a [configuration](https://docs.getdbt.com/reference/test-configs#test-specific-configurations):

```SQL

{{ config(
    materialized="view"
) }}

select
    date_day,
    daily_active_users,
    current_date() as test_date
from {{ ref('agg_daily_active_users' )}}
where daily_active_users < 0
```

Which allows me to issue to debug my data with the following query:

```SQL
select * from dbt_test__audit.agg_daily_active_users_negative_count
```
Which when I query it tomorrow will give me the following alarming situation, without having to use dbt.

|date_day|daily_active_users|test_date|
|-|-|-|
|2023-02-25|-547|2023-02-09|
|2023-02-07|-2|2023-02-09|
|2023-02-06|-1500|2023-02-09|

### Should we deprecate [`store-failures`](https://docs.getdbt.com/docs/build/tests#storing-test-failures)?

The run level flag? Certainly not. The intent is different. With `--store-failures` I'm deciding at run time to persist my test results, independently of my test materializations.

The test level config? I think setting `materialized="table"` is equivalent to `store_failures = true`. I don't want to start the deprecation process of `store_failures = true` yet though. Let's punt that for later.

In terms of **precedence of configurations:**
- the `materialized` configuration is top level, it takes precedence over `store_failures` and the flag (they are ignored)
  - if the test `materialized` configuration is omitted, the `store_failures` config will take precedence over the presence or absence of the `--store-failures` flag (it is ignored)
    - If the `materialized` configuration is omitted, or the `store_failures` config is `none` or is omitted, the resource will use the value of the `--store-failures` flag (current [behavior](https://docs.getdbt.com/reference/resource-configs/store_failures#description))

### Available materializations

We should support `view` and `table` for now.

Out of scope at the moment, we will support `materialized_view` at some point. Because warehouses may be able to surface the history of MVs (change log), which would allow me to get a list of all the transient errors that happened when I was not looking at the test MV.

### Scope of the configuration

This configuration is a ["tests config"](https://docs.getdbt.com/reference/test-configs). It should be available at every level where a test configuration is offered and respect normal [inheritance rules](https://docs.getdbt.com/reference/configs-and-properties#config-inheritance).

Just use a model instead of a test. Because that's basically what we're discussing here.

But in my opinion tests are tests not because their output is a collection of rows, a csv, a table or a view. They are tests because of their specific intent, their life cycle, and the fact they don't belong on the DAG. 
I still want to be able to `dbt test`, even when all my tests are materialized as views.

mikealfare:
resolves #6914

### Problem

We now support materialized views (and dynamic tables in Snowflake). These new objects can update without running dbt. However, the test results will stay static if we run them with `--store-failures`, or just won't exist at all if we don't specify `--store-failures`. This is philosophically inconsistent behavior.

### Solution

Support test results as a view, so that they will always show the current failures without the need to run dbt. We should also support table as a relation type since this is effectively `--store-failures`. Users can specify this setting in the config block in their tests by setting `store_failures_as="view"`.

Approach:
- Add a new parameter to the `TestConfig` called `store_failures_as` (was `strategy` in previous iterations)
- Default `store_failures_as` to `"table"` to align with existing behavior for when `store_failures = True`
- Catch `store_failures_as` in the `test` materialization
- Replace the hard coded `'table'` with the contents of `store_failures_as`
- Update the create statement to allow for a generic create instead of specifying "create table"
- Update expected values for existing tests to reflect the new parameter `store_failures_as` with a non-empty default (`"table"`)
- Add tests to demonstrate both views and tables can be created
- Ensure existing `store_failures` tests still pass as-is
- Ensure dbt-managed adapters reflect the changes:
  - https://github.com/dbt-labs/dbt-snowflake/pull/783
  - https://github.com/dbt-labs/dbt-bigquery/pull/932
  - https://github.com/dbt-labs/dbt-redshift/pull/614
  - https://github.com/dbt-labs/dbt-spark/pull/889

### Checklist


tests/adapter/dbt/tests/adapter/store_test_failures_tests/test_persist_results.py
mikealfare, line 1: I just started with a test that demonstrates this runs. I manually verified that the views are actually created, but I want to add tests for that.

mikealfare: I updated tests to check for views and tables, and to show that the views will update with results whereas the tables don't.

core/dbt/contracts/graph/model_config.py
```diff
@@ -572,6 +572,7 @@ def same_contents(cls, unrendered: Dict[str, Any], other: Dict[str, Any]) -> boo
             "warn_if",
             "error_if",
             "store_failures",
+            "strategy",
```
colin-rogers-dbt, line 575: I think strategy is a little vague, maybe something like: `store_materialization`? That way it's clear that it's related to `store_failures`

core/dbt/contracts/graph/nodes.py
```diff
@@ -984,7 +983,9 @@ def language(self):
 class TestShouldStoreFailures:
     @property
     def should_store_failures(self):
-        if self.config.store_failures:
+        if self.config.strategy in ["view", "table"]:
```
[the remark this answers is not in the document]
mikealfare: I agree that there is some conflict between `store_failures` and what we're calling `strategy`. Per the [reqs](https://github.com/dbt-labs/dbt-core/issues/6914):

> In terms of precedence of configurations:
>  - the materialized configuration is top level, it takes precedence over store_failures and the flag (they are ignored)
>    - if the test materialized configuration is omitted, the store_failures config will take precedence over the presence or absence of the --store-failures flag (it is ignored)
>      - If the materialized configuration is omitted, or the store_failures config is none or is omitted, the resource will use the value of the --store-failures flag (current [behavior](https://docs.getdbt.com/reference/resource-configs/store_failures#description))

I could see this going the other way. In other words, if `store_failures` is `False`, then we don't store them. If `store_failures` is `True`, then `strategy` tells dbt _how_ to store the failures. If it's missing, then it defaults to `table`, which is what happens today. @Fleid, I would be interested in your thoughts on this, given you're the source of the reqs above.

core/dbt/contracts/graph/model_config.py
mikealfare: Agreed. When we first picked `strategy`, I didn't realize how close we were to the generic `Config` object. Maybe `store_failures_as`? Again, I would want input from @fleid on this, and maybe @dataders as well.

dataders (commented):
yo so just starting to scratch the surface on the user-interface conundrum. I don't have time to do a full write up, but plan to do one tomorrow and get sign off from the rest of my DX team.

core/dbt/contracts/graph/model_config.py
colin-rogers-dbt: `store_failures_as` seems pretty intuitive to me

core/dbt/include/global_project/macros/materializations/tests/test.sql
```diff
@@ -1,8 +1,33 @@
...
+
+        {% do relations.append(target_relation) %}
+
+        {% set main_sql %}
+            select * from {{ target_relation }}
+        {% endset %}
+
+        {{ adapter.commit() }}
+
+  {% elif should_store_failures() %}
```
dataders, line 30: here's my read on the logic

1. if `relation_type` is configured, then depending on the argument materialize a table or view
2. else, if `should_store_failures` do all that jazz.

the first thing reads to me effectively as a sub-materialization, that is independent of that. so might i be able to pass `should_store_failures() and relation_type='table'`? what happens then?

I'm down with a deprecation in the future, but in the meantime might be deprecate the implementation and make `config.get('store_failures')` imply `relation_type='table'`? then there's only one code path for creating a table and storing test results?

dataders (commented):
left a comment on the new test materialization workflow

core/dbt/include/global_project/macros/materializations/tests/test.sql
mikealfare: > so might i be able to pass should_store_failures() and relation_type='table'? what happens then?

It will create it as a table. This is the same as omitting the 'table' piece entirely.

> but in the meantime might be deprecate the implementation and make config.get('store_failures') imply relation_type='table'?

That's the conclusion that @colin-rogers-dbt and I reached as well. That alters the feature request slightly, but it makes much more intuitive sense. The feature request is now effectively:

I know `store_failures` creates the results as as table, but I want to make it create the results as a view.

Instead of implementing a feature alongside `store_failures`, we're augmenting `stored_failures` itself.

core/dbt/include/global_project/macros/materializations/tests/test.sql
```diff
@@ -6,15 +6,17 @@
...
-        identifier=identifier, schema=schema, database=database, type='table') -%} %}
+        identifier=identifier, schema=schema, database=database, type=relation_type) -%} %}
 
     {% if old_relation %}
         {% do adapter.drop_relation(old_relation) %}
     {% endif %}
 
     {% call statement(auto_begin=True) %}
-        {{ create_table_as(False, target_relation, sql) }}
+        {{ get_create_sql(target_relation, sql) }}
```
mikealfare, line 19: It's worth noting this potentially provides shadow support for materialized views with default configurations. This is not tested, and is left as a future.

mikealfare:
> yo so just starting to scratch the surface on the user-interface conundrum. I don't have time to do a full write up, but plan to do one tomorrow and get sign off from the rest of my DX team.
> 
> when last we spoke, @dbeatty10 and I landed on something like `results_materialized` `test_materialized` as the ideal config. in this scenario the existing default is `ephemeral`, though it can changed to be `table` or `materialized_view`. stay tuned

I updated this to `store_failures_as` since it directly modifies behavior arising from `store_failures`. We were against using "materialize" or its variations here since this is a relation and not a materialization. We would like to draw a line in the sand here and actively make decisions that help disambiguate the two concepts. We're open to other variations but we felt like `store_failures_as` was a decent placeholder (if not final decision).

In order to support backwards compatibility the default needs to be `table` since that is the current behavior of `store_failures` (which obviously omits the `store_failures_as` argument). I think setting `store_failures_as` to `ephemeral` is effectively setting `store_failures=False`. Perhaps that is your argument. Either way, we need a default for `store_failures_as` when `store_failures=True`, and `ephemeral` doesn't align with `store_failures=True`.

colin-rogers-dbt (approved):

dataders:
@mikealfare @colin-rogers-dbt want to call your attention to a prototype that @dbeatty10 made with #8748. Obviously not ready to be merged, but useful perhaps as a proof of concept.

Our DX opinion is that `materialized` is the ideal config over `strategy` or `store_failures_as`, but leave both of the following to engineering:
- implementation, and
- sequencing (e.g. we introduce `store_failures_as` as a short-term solution to address the existing specification, then phase it out in favor of the more ideal interface)

dbeatty10:
> Our DX opinion is that `materialized` is the ideal config over `strategy` or `store_failures_as`, but leave both of the following to engineering:

:point_right: If you take a look at https://github.com/dbt-labs/dbt-core/pull/8748, here's the commit that actually matters as it relates to `materialized`: https://github.com/dbt-labs/dbt-core/pull/8748/commits/468c85c282e98086690e076d598418768cca1215

i.e., change just a couple lines of code, and `materialized=view` is allowed for tests :sparkles:

colin-rogers-dbt:
> opinion is that `materialized` is the ideal config over `strategy` or `store_failures_as`

@dataders / @dbeatty10 I think there is further discussion to be had here as long term we would like to draw a meaningful and consistent distinction between _materialization_ and _relation_. I hark back to @mikealfare's thoughts here: https://github.com/dbt-labs/dbt-core/discussions/8177

dbeatty10 (changes requested):
## TLDR
- The precedence of configurations appears different than described in [#6914](https://github.com/dbt-labs/dbt-core/issues/6914).
- Ran into some issues. Proposed fix in https://github.com/dbt-labs/dbt-core/commit/0baf090c3e030c80d3927120a5f7eb1a88482c2c.

## Summary

#### Four things I checked for this PR:

1. Precedence of configurations in [#6914](https://github.com/dbt-labs/dbt-core/issues/6914)
1. Singular test in `tests/singular/check_something.sql` file 
1. Generic tests in `models/_models.yml` file
1. Config in `dbt_project.yml`

#### Not covering in this review:
- User experience of `store_failures_as` vs. `materialized`, which we can cover in a separate discussion.

I hear you about your goal of making decisions that help [disambiguate](https://github.com/dbt-labs/dbt-core/pull/8653#issuecomment-1740177267
) between the concepts of "relations" and "materializations"! Let's set up some time to discuss that portion further.

## 1. Precedence of configurations

The issue description in [#6914](https://github.com/dbt-labs/dbt-core/issues/6914) described the **precedence of configurations**. To make it easier to discuss in the context of this PR, I've replaced "`materialized`" (how it appeared originally) with "`store_failures_as`".

> In terms of **precedence of configurations:**
> - the `store_failures_as` configuration is top level, it takes precedence over `store_failures` and the flag (they are ignored)
>   - if the test `store_failures_as` configuration is omitted, the `store_failures` config will take precedence over the presence or absence of the `--store-failures` flag (it is ignored)
>     - If the `store_failures_as` configuration is omitted, or the `store_failures` config is `none` or is omitted, the resource will use the value of the `--store-failures` flag (current [behavior](https://docs.getdbt.com/reference/resource-configs/store_failures#description))

@mikealfare Is this the precedence that was implemented?

Or did the design of the user experience get intentionally altered at some point?

Note: See "Config in `dbt_project.yml`" below for additional precedence considerations.

## 2. Singular tests

Worked in my local testing ✅ 

## 3. Generic tests

Didn't work in my local testing ❌

<details>
<summary>Reprex</summary>

1. Create these two files:
    
    `models/my_model.sql`
    
    ```sql
    select 1 as id union all
    select 2 as id
    ```
    
    `models/_models.yml`
    
    ```yaml
    models:
      - name: my_model
        columns:
          - name: id
            tests:
              - not_null:
                  config:
                    store_failures_as: view
    ```

2. Run the model and its generic tests

    ```shell
    dbt run my_model
    dbt test --select fqn:my_model,test_type:generic --store-failures
    ```

3. Then check either the logs or the database to see if the results were stored as a view or not.

</details>

Proposed fix in in https://github.com/dbt-labs/dbt-core/commit/0baf090c3e030c80d3927120a5f7eb1a88482c2c.

@mikealfare What do you think about adding some pytests that cover [generic tests](https://docs.getdbt.com/docs/build/tests#example)?

See "Next steps" below for a reproducible example and proposed next steps.

## 4. Config in `dbt_project.yml`

Didn't work in my local testing ❌

<details>
<summary>Reprex</summary>

```yaml
name: "my_project"
version: "1.0.0"
config-version: 2
profile: "sandcastle"

tests:
  my_project:
    +store_failures_as: table
    my_subfolder_1:
      +store_failures_as: view
    my_subfolder_2:
      +store_failures_as: view
```

`models/my_subfolder_1/my_model_1.sql`

```sql
select 1 as id union all
select 2 as id
```

`models/my_subfolder_1/_models.yml`
```yaml
models:
  - name: my_model_1
    columns:
      - name: id
        tests:
          - not_null
```

`models/my_subfolder_2/my_model_2.sql`

```sql
select 1 as id union all
select 2 as id
```

`models/my_subfolder_2/_models.yml`
```yaml
models:
  - name: my_model_2
    columns:
      - name: id
        tests:
          - not_null
```

```shell
dbt run
dbt test --select test_type:generic --store-failures
```

</details>

Proposed fix in in https://github.com/dbt-labs/dbt-core/commit/0baf090c3e030c80d3927120a5f7eb1a88482c2c.

Hierarchical configurations in `dbt_project.yml` have their own order of precedence:
- [Test configurations](https://docs.getdbt.com/reference/test-configs#related-documentation)
    - [Test-specific configurations](https://docs.getdbt.com/reference/test-configs#test-specific-configurations)
- [Configuring directories of models in dbt_project.yml](https://docs.getdbt.com/reference/model-configs#configuring-directories-of-models-in-dbt_projectyml)
- [Config inheritance](https://docs.getdbt.com/reference/configs-and-properties#config-inheritance)

## Key questions

Here's a summary of the key questions that appear inline above:
1. Is the original precedence in the issue the one that got implemented? Or did the design of the user experience get intentionally altered at some point?
2. What do you think about adding some pytests that cover [generic tests](https://docs.getdbt.com/docs/build/tests#example)?

mikealfare:
The first thing I'll call out on the most recent comments, this is all good conversation that should probably be in something like a GH discussion. We want it to be on-going and not get archived away with this PR (which kind of by definition should be a small change). I'll address a few of the points here for now, but would encourage that we migrate.

> Our DX opinion is that `materialized` is the ideal config over strategy or store_failures_as...

This is not currently possible without breaking backwards compatibility. The value for `materialized` on every test is currently `test`. Updating that would be a giant lift and we should think about how we do that (if we decide to do it at all).

> 👉 If you take a look at https://github.com/dbt-labs/dbt-core/pull/8748, here's the commit that actually matters as it relates to materialized: https://github.com/dbt-labs/dbt-core/commit/468c85c282e98086690e076d598418768cca1215

The implementation here is to call a materialization from another materialization. We had a similar conversation last month and decided we will not this. One of the results of that conversation was that this is further evidence that we're blurring the lines between a relation and a materialization, and that we want to actively fight that.

> @mikealfare Is this the precedence that was implemented?
Or did the design of the user experience get intentionally altered at some point?

The precedence that was implemented differs from that in the feature request. Engineering made this decision after seeing what the implemented requirements resulted in, and determined that it was both confusing and potentially not backwards compatible. The intention was to cover this version with @Fleid to ensure that we were still delivering the spirit of the feature.

> 3. Generic tests
Didn't work in my local testing ❌

This is a miss. The tests for this PR only touch on custom tests. I will add a test for generic tests and review/merge your commit that addresses this.

> 4. Config in dbt_project.yml
Didn't work in my local testing ❌

Same as above.

@dbeatty10, to make sure I cover all of your concerns, what is the change in particular that warranted the blocking review?

dbeatty10:
### Blocking review

> @dbeatty10, to make sure I cover all of your concerns, what is the change in particular that warranted the blocking review?

I'm open to feedback if it is more helpful and useful to utilize a different mechanism than a blocking review!

I put on the blocking review to avoid accidentally merging before these three things were covered:
- 1\. Precedence of configurations
- 3\. Generic tests
- 4\. Config in dbt_project.yml

Just let me know if you feels like there's a better way to communicate and approach these types of things.

### Precedence of configurations

> The intention was to cover this version with @Fleid to ensure that we were still delivering the spirit of the feature.

Since this config UX and behavior is at the Core/Adapters intersection, we should include DX/PM feedback from the Core team (me and/or @grace). Could you add one or both of us to that conversation with @Fleid so we can make sure all of us are on the same page?

The original precedence that @Fleid outlined made perfect sense to me from a DX perspective and seems like the ideal. But totally understand that you may have discovered some things during the implementation that might affect our DX design -- if so, let's collaboratively discuss and figure out what our options are.

colin-rogers-dbt:
Just want to confirm the following behavior @Fleid :
We are going forward with having `store_failures_as` take precedence over `store_failures`, which means we are comfortable with the following scenarios:
* Users will be unable to turn _off_ storing failures in the model config if they specify `store_failures_as` at a higher level
* If a user specifies `store_failures` as false but specifies `store_failures_as` we will still store failures

Fleid:
We had talked about the fact that the `materialized` keyword was not ideal, and I think we decided on `strategy`. 
I'm not against changing it again. There is a tension between the current model - where users will most likely understand and expect`materialized` - and the one we are aiming forward - with a clearer delineation between relation and materialization. Up to us to decide together on which we want to optimize for here.

My initial reaction is the later, I like setting up future-us for success ;)

But that keyword change was discussed in the context of a 1:1 replacement of my initial proposal.
If we also need to change the order of precedence (do we really?) then we'll need to re-consider also the keyword itself.

We should chat :)

mikealfare:
@dbeatty10 @dataders @Fleid @colin-rogers-dbt Please review the test cases in particular to make sure they align with the intention of the feature. The functional code change itself is small; I'm just using the `__post_init__` to effectively swap the prioritization. The test cases contain the pointy edges that we discussed.

dbeatty10:
> @dbeatty10 @dataders @Fleid @colin-rogers-dbt Please review the test cases in particular to make sure they align with the intention of the feature.

@mikealfare Did you add any test cases that cover the generic tests described in the reprex of "3. Generic tests" [here](https://github.com/dbt-labs/dbt-core/pull/8653#pullrequestreview-1652835272)?

mikealfare:
> > @dbeatty10 @dataders @Fleid @colin-rogers-dbt Please review the test cases in particular to make sure they align with the intention of the feature.
> 
> @mikealfare Did you add any test cases that cover the generic tests described in the reprex of "3. Generic tests" [here](https://github.com/dbt-labs/dbt-core/pull/8653#pullrequestreview-1652835272)?

I'm working on that today. I wanted to swap the existing state first and get that pushed up.

dbeatty10:
@colin-rogers-dbt identified an unspecified piece here:

> * Users will be unable to turn _off_ storing failures in the model config if they specify `store_failures_as` at a higher level


UX to address this: https://github.com/dbt-labs/dbt-core/issues/6914#issuecomment-1745401752

mikealfare:
Generic test test cases are in and passing. I merged your updates @dbeatty10 and they addressed this gap.

> Users will be unable to turn off storing failures in the model config if they specify store_failures_as at a higher level

This is possible, but not intuitive. I added a test case that demonstrates how to do this. Basically you set `store_failures = False` and `store_failures_as = None`: https://github.com/dbt-labs/dbt-core/blob/c09b81dce9a2a111f77142d76593557d259826a1/tests/adapter/dbt/tests/adapter/store_test_failures_tests/_files.py#L96-L101

colin-rogers-dbt (approved):
LGTM, doesn't look like we have covered these with unit tests in the past but worth considering if this change justifies some: https://github.com/dbt-labs/dbt-core/blob/main/tests/unit/test_model_config.py

dbeatty10:
> This is possible, but not intuitive.

@mikealfare Thanks for explaining how this is possible right now, even it if it isn't intuitive.

We would love to make this intuitive so that it's easier to document and easier to accomplish.

Do you have any concerns with the proposal in https://github.com/dbt-labs/dbt-core/issues/6914#issuecomment-1745401752? It proposes using `"ephemeral"` to mean "don't actually create any database object".

colin-rogers-dbt:
@dbeatty10 in terms of the config parameter naming are you comfortable with proceeding with `store_failures_as` or do you feel we should stick with `strategy`?

I prefer `store_failures_as` since it's clearer/more specific but I think it's more important to get this out the door so don't want this to be a blocker

mikealfare:
@dbeatty10, I'm comfortable moving forward with "ephemeral"; I'll make those changes and push it up.

I agree with @colin-rogers-dbt that we should stick with `store_failures_as`. My argument for `store_failures_as` is that it directly impacts `store_failures` (in the sense that it takes precedence). This behavior feels more intuitive if the names of the settings are similar. Thoughts @dbeatty10?

dbeatty10:
> @dbeatty10 in terms of the config parameter naming are you comfortable with proceeding with `store_failures_as` or do you feel we should stick with `strategy`?
> 
> I prefer `store_failures_as` since it's clearer/more specific but I think it's more important to get this out the door so don't want this to be a blocker

@colin-rogers-dbt and @mikealfare

### My order of preference

1. **Re-use** an existing config name that has **same** meaning (i.e., `materialized`)
2. Create a **new** config name that could be re-used **across** resource types (like `strategy`, etc.)
3. Create a **new** config name that can only be used for a **single** resource type (like `store_failures_as`, etc.)

This order is meant to capture what would be most easy for users.

I totally understand that there may be engineering considerations that could affect one or more of these. If so, would love to learn anything you've uncovered or anything you think should be taken into consideration.

To be clear, I think any of them are workable long-term from a user perspective. They are just ordered from most intuitive to least. Theoretically, an experienced dbt user wouldn't even need to consult the documentation for the first one, but would definitely need to read docs for the last one.

dbeatty10:
> And to confirm, is this all that's left for this PR? Or am I missing another issue?

@mikealfare

### Outstanding things to solve

From my perspective, here's the user experience items that we should finalize:
1. Naming of the config (`store_failures_as`, `strategy`, `materialized`, etc.)
1. Accepted values of the config (`table`, `view`, `ephemeral`, etc.)
1. How to behave if the user supplies a non-accepted value (`materialized_view`, `incremental`, `purple_people`, etc)

See above for my take on the first two, and see below for the last one.

### How to behave if the user supplies a non-accepted value
It seems like we should raise an error for invalid values. It could be something similar to this:
```
20:23:35  Completed with 1 error and 0 warnings:
20:23:35  
20:23:35    Compilation Error in test some_test_name_here (tests/some_file_name_here.sql)
  'purple_people' is not a valid value for configuration_name_here. Accepted values are ['table', 'view', and 'ephemeral']
```

What do you think about raising an error when the config includes an invalid value?

colin-rogers-dbt:
As @mikealfare described above `materialized` is problematic in our current implementation so I don't think we should use it here.

> Theoretically, an experienced dbt user wouldn't even need to consult the documentation for the first one, but would definitely need to read docs for the last one.

I like this principle a lot but are we concerned about reusing words in different contexts that accept different inputs and mean different things? In this case `strategy` means one thing when configuring snapshots another in incremental and now would be another for tests. 

Also, I don't think we do but any chance we're going to want to store success in the future? (i.e. would we need separate 'strategies' for successful /failing tests?) 

> How to behave if the user supplies a non-accepted value

+1 a compilation error is what a user should expect here

mikealfare:
> 1. **Re-use** an existing config name that has **same** meaning (i.e., `materialized`)

If we could go back and change how tests work, then I agree that `materialized` makes the most sense here. However that's already taken up with the value (and associated functionality) of `test`. I could see reworking that over time, but not before the 1.7 rc. The blast radius would be huge and I'd want to get a lot of input before a change like that.

> 2. Create a **new** config name that could be re-used **across** resource types (like `strategy`, etc.)

Given this is the next preference, I'll start making these updates. But I do have some questions:

1. What is meant by a "resource type"?
2. What is the valid value list for `strategy`? Would that list change for other resource types?

Regarding the second question, I could see the list getting shorter or longer across resource type (e.g. `table`/`view` for one and `table`/`view`/`ephemeral` for another). But I think it would get really hairy if we start accepting different concepts across resource type (e.g. `table`/`view` for one and `merge`/`delete+insert` for another).

@colin-rogers-dbt got this in while I was writing this up. So to include his piece here, he is highlighting exactly my concern with `strategy` and provides good context for my second question:

> I like this principle a lot but are we concerned about reusing words in different contexts that accept different inputs and mean different things? In this case strategy means one thing when configuring snapshots another in incremental and now would be another for tests.

----

> 1. Naming of the config (store_failures_as, strategy, materialized, etc.)

From the above, it seems like this is `strategy`.

> 2. Accepted values of the config (table, view, ephemeral, etc.)

The valid values are currently any value of `RelationType`. However, this is implicit and not strictly enforced. We only support `table` and `view` for now (and `ephemeral` after implementing what we mentioned above). Theoretically you could also pass in `materialized_view` and see what happens, YMMV. That being said, if we're going to raise an exception for invalid values as stated in 3., then it would be the three values you stated.

> 3. How to behave if the user supplies a non-accepted value (materialized_view, incremental, purple_people, etc)

I'm sure we don't fail gracefully here as the valid value list is implicit. I'll see if I can throw something in along the lines of your suggestion, but I can't promise that it will cover all cases.

mikealfare:
For those following along at home, ephemeral is now in, as well as the compilation error.

dbeatty10:
> I like this principle a lot but are we concerned about reusing words in different contexts that accept different inputs and mean different things? In this case strategy means one thing when configuring snapshots another in incremental and now would be another for tests.

Definitely hear you both that using `strategy` would be confusing because it is already in use by incremental materializations.

I'm not attached to the name `strategy` at all, was just using it as a stand-in for the idea of "Create a new config name that could be re-used across resource types".

By "resource types" (AKA "node types"), I mean [these](https://docs.getdbt.com/docs/dbt-cloud-apis/project-state#affected-states-by-node-type) which are listed in the code [here](https://github.com/dbt-labs/dbt-core/blob/5e1f0c5fbc63ffa298a48c971f7de89d29edf079/core/dbt/node_types.py#L20-L36).

### `store_failures_as`

In addition to @mikealfare & @colin-rogers-dbt being advocates of `store_failures_as`, @jtcohen6 stated preference for `store_failures_as` in dbt internal Slack [here](https://dbt-labs.slack.com/archives/C02HE19D51R/p1696251693057819?thread_ts=1695916353.788799&cid=C02HE19D51R). @Fleid also expressed preference for this name in a Zoom call this afternoon.

So while `materialized` seems like the most simple, elegant, and re-usable config name to me, I will disagree and commit on this one and also support `store_failures_as` as the name of a new config.

Sorry if this causes any churn for you @mikealfare!

core/dbt/include/global_project/macros/materializations/tests/test.sql
mikealfare: This is no longer the case as we raise an exception for unsupported values of `store_failures_as`.

dbeatty10 (approved):
@mikealfare

### Works as expected

I re-tried the use cases described [here](https://github.com/dbt-labs/dbt-core/pull/8653#pullrequestreview-1652835272) in a local environment and they worked 😎

Each of these values all behaved as expected:
```yaml
store_failures_as: ephemeral
store_failures_as: null
store_failures_as:
```

I also tried giving invalid values like the following, which also gave appropriate error messages:
```yaml
store_failures_as: materialized_view
store_failures_as: carrot
store_failures_as: none
store_failures_as: None
```

### One edge case

One discovered edge case was switching from `table` (or `view`) to `ephemeral` between successive `dbt run`s.

When switching from `table` to `view`, the 1st relation in dropped and the 2nd is created. But when switching from `table` to `ephemeral`, the table persists afterwards and needs to be manually dropped.

### Summary

Regardless of that edge case, approving this PR 👍

We can follow-up in a separate issue to decide what to do when switching from `table` (or `view`) to `ephemeral`.

PR #8654, opened 2023-09-15
Add semantic model test to `test_contracts_graph_parsed.py`

### LINKED ISSUE #8139: [CT-2842] Add SemanticModel tests to `test_contracts_graph_parsed`
## Context
We have a bunch of tests in [test_contract_graph_parsed.py](https://github.com/dbt-labs/dbt-core/blob/main/tests/unit/test_contracts_graph_parsed.py) which check the symmetry of dict representation of objects and the actual python objects as well as checking the required properties of graph nodes. Currently we have none of these tests for `SemanticModel`s

## Acceptance Criteria
- Tests exist in `test_contract_graph_parsed.py` for checking symmetry and required graph node properties of `SemanticModel`s
- Tests do not require hand-rolling of objects and dictionaries

QMalcolm:
resolves #8139

### Problem

The tests in `test_contracts_graph_parsed.py` are meant to ensure that we can go from objects to dictionaries and back without any changes. There are two problems this PR aims to solve

1. these tests are brittle, and are tedious to fix (we want a simpler style of test)
2. we lacked a test of this style for semantic models

### Solution

We added a test for semantic models symmetry. The test depends on a new package `hypothesis`. The `@given` parameterizes a test, with it generating the arguements it has following `strategies`. The main strategies we use is `builds` this takes in a callable passes any sub strategies for named arguements, and will try to infer any other arguments if the callable is typed. I found that even though the test was run many many times, some of the `SemanticModel` properties weren't being changed. For instance `dimensions`, `entities`, and `measures` were always empty lists. Because of this I defined sub strategies for some attributes of `SemanticModel`s.

### Checklist

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -5,6 +5,7 @@ docutils
 flake8
 flaky
 freezegun==0.3.12
+hypothesis
 ipdb
 mypy==1.4.1
 pip-tools
--- a/tests/unit/README.md
+++ b/tests/unit/README.md
@@ -1,1 +1,18 @@
 # Unit test README
+
+## test_contracts_graph_parsed.py
+
+### The Why
+We need to ensure that we can go from objects to dictionaries and back without any
+changes. If some property or property value of an object gets dropped, added, or modified
+while transitioning between its different possible representations, that is problematic.
+
+### The How 
+The easiest way to ensure things don't get droped, added, or modified is by starting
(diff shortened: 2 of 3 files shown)


emmyoop (changes requested):
Can you add some info to the existing README (that's empty...) about when we generally want to use hypothesis?

gshank (approved):

PR #8670, opened 2023-09-19
Fix #6497: Support global flags passed in after subcommands

### LINKED ISSUE #6497: Support arbitrary order for CLI flags
We currently distinguish between two types of flags:
- those which precede the subcommand ("global"), e.g. `--no-use-color`
- those which follow the subcommand, and may have different behavior depending on the subcommand, e.g. `--select`, `--full-refresh`

IMO this is pretty confusing for end users! We should really seek to support flags in any order (and resolve any resulting ambiguities by actually renaming flags). I know we implemented the new CLI modeling in accordance with the old CLI model — parity is a good goal to shoot for with any refactor — but it would be great to just support both of these, seamlessly and equivalently:
```
$ dbt --no-use-color run --select one_model
$ dbt run --select one_model --no-use-color
```

aranke:
resolves #6497 

### Checklist

### DIFF
--- a/core/dbt/cli/main.py
+++ b/core/dbt/cli/main.py
@@ -65,7 +65,7 @@ class dbtRunner:
         self,
         manifest: Optional[Manifest] = None,
         callbacks: Optional[List[Callable[[EventMsg], None]]] = None,
-    ):
+    ) -> None:
         self.manifest = manifest
 
         if callbacks is None:
--- a/.changes/unreleased/Features-20230914-074429.yaml
+++ b/.changes/unreleased/Features-20230914-074429.yaml
@@ -1,0 +1,7 @@
+kind: Features
+body: Add support for optional label in semantic_models, measures, dimensions and
+  entities.
+time: 2023-09-14T07:44:29.828199-05:00
+custom:
+  Author: emmyoop
+  Issue: "8595"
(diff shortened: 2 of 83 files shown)


core/dbt/cli/main.py
```diff
@@ -118,6 +119,44 @@ def invoke(self, args: List[str], **kwargs) -> dbtRunnerResult:
             )
 
 
+# approach from https://github.com/pallets/click/issues/108#issuecomment-280489786
+def global_flags(func):
```
jtcohen6, line 123: This is neat!

jtcohen6 (commented):
Mea culpa - I had missed adding the 1.5 + 1.6 backport labels to https://github.com/dbt-labs/dbt-core/issues/6497. Ideally we could offer the same consistency for all our CLIs starting with v1.5+. Given how "clean" this change looks, how would you feel about backporting the change (at least the approach if not the exact commit diff) to `1.5.latest` + `1.6.latest`?

jtcohen6:
@aranke That's odd - it's only those three flags which are being difficult? Do you have any sense of why?

jtcohen6:
@aranke Thanks for investigating & opening the ticket! We should try to get that tech debt ticket resolved — it's important for our ability to present consistent documentation & CLI behavior (https://github.com/dbt-labs/docs.getdbt.com/issues/4041) — but I agree that it shouldn't block merging the brunt of the work in this PR.

gshank (approved):

emmyoop (approved):
LGTM

ChenyuLInx (approved):
Thanks!

PR #8693, opened 2023-09-22
Add git info to context

### LINKED ISSUE #8690: [Feature] Make the current git branch (if any) available in the dbt Jinja context

### Describe the feature

I would like to be able to access the current git branch name from the dbt context in order to be able to run some Jinja code depending on its value.

If we had `current_branch` available as a Jinja variable, we could potentially generate models in schemas/database (by adding some logic in `generate_schema_name()`) that depend on this branch name. 

This could be useful for longer living branch where multiple developers will work on the same feature.

We would need to handle the case where dbt is run on code that is not a git repo/branch and maybe return an empty value then.

People could potentially define a `var` in `dbt_project.yml` to hard code what branch they are on. This variable would then have the same value as the current branch but the drawbacks are that:

- people need to remember to set this variable every time
- there will often be merge conflicts to solve on the line where the `var` is defined as each branch would define its own value

### Who will this benefit?

- people who work on multiple Feature Branches at the same time
- teams where multiple people work on the same feature branches
- people having long-living feature branches for specific purpose

Yes!

### Anything else?

_No response_

b-per:
resolves #8690 

### Problem

Today, it is not possible to know which branch and/or which commit dbt's current code is based on. This PR adds the variables `git_branch` and `git_sha` to the Jinja context

### Solution

Add objects to the context

### Still To be Done (and keen to get Core Team inputs on the 2 first ones)

- [x] Handle the case where `--project-dir` is set and not the current dir
- [x] Add tests
- [ ] Handle partial parsing gracefully, forcing a full parse if any of the [`special_override_macros`](https://github.com/dbt-labs/dbt-core/blob/81447496418e6ae71670b9aea73c7d3b54fff436/core/dbt/parser/partial.py#L53-L60) refers to git_branch` or `git_sha`.

### Checklist

### DIFF
--- a/.changes/unreleased/Features-20230925-182237.yaml
+++ b/.changes/unreleased/Features-20230925-182237.yaml
@@ -1,0 +1,6 @@
+kind: Features
+body: Add git_branch and git_sha to the Jinja context
+time: 2023-09-25T18:22:37.142445+02:00
+custom:
+  Author: b-per
+  Issue: "8690"
--- a/core/dbt/context/base.py
+++ b/core/dbt/context/base.py
@@ -208,12 +208,18 @@ class BaseContext(metaclass=ContextMeta):
 
     @contextproperty()
     def git_branch(self) -> str:
-        """The `git_branch` variable returns the current branch if the
+        """The `git_branch` variable returns the current branch name if the
         code is version controlled by git.
         Otherwise it returns an empty string.
         """
+
+        if hasattr(get_flags(), "PROJECT_DIR"):
+            project_dir = get_flags().PROJECT_DIR
+        else:
+            project_dir = "."
+
         try:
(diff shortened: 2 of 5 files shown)


b-per:
I have now added support for `project_dir`, updated existing tests and added a new one.

The last pending thing would be to invalidate the partial parsing cache in case `git_branch` is used there but I would most likely need a bit of help from the core team for that.

b-per:
Would there also be appetite to add this info to `manifest.json` and/or `run_results.json`?

b-per:
Thinking more about it. 

In order to identify that we changed branch (and that we might need to trigger a full parse in case we refer to `{{ branch_name }}` in one of the `generate_xxx_name()` ), would there be any other technical way except storing this information in the manifest?

PR #8721, opened 2023-09-26
Model contracts: raise warning for numeric types without specified scale

### LINKED ISSUE #8183: Check yaml numeric scale >= SQL numeric scale in model contracts
[slack thread](https://getdbt.slack.com/archives/CMZ2Q9MA9/p1689785924008589) & https://github.com/dbt-labs/dbt-core/issues/7824#issuecomment-1586217515

A few data platforms (Snowflake, Redshift) have `scale=0` for their numeric/decimal data type. (It's 9 by default on [BigQuery](https://cloud.google.com/bigquery/docs/reference/standard-sql/data-types#decimal_types).)

[Postgres' docs](https://www.postgresql.org/docs/current/datatype-numeric.html) say:

> The SQL standard requires a default scale of 0, i.e., coercion to integer precision. We find this a bit useless. If you're concerned about portability, always specify the precision and scale explicitly.

I agree!

### Reproduction case

Thanks @gnilrets for the clear repro case!

```
-- models/fct_test_contract.sql
select
   2/3 as two_thirds
```
Without contract enforcement, two_thirds is a number(7,6).  However, if I enforce the contract and fail to specify the precision:
```yml
models:
  - name: fct_test_contract
    config:
      contract:
        enforced: true

    columns:
      - name: two_thirds
        data_type: number
```
Then the build passes, but `two_thirds` is now a `number(38,0)`, so I’ve lost all precision and `two_thirds = 1`!

### Proposed resolution

Our contract check should ensure that the `scale` specified in the yaml contract is >= the maximum `scale` in the SQL definition. IMO they shouldn't have to match up _exactly_, but we should avoid unintended rounding/coercion, especially since that's an explicit anti-goal of contracts (#8035).

This will require us to start saving information about numeric scale here:

https://github.com/dbt-labs/dbt-core/blob/0d645c227f9e84116ee03c048bad22fce4fcc2b3/core/dbt/adapters/base/impl.py#L303-L314

And performing the scale comparison here:

https://github.com/dbt-labs/dbt-core/blob/0d645c227f9e84116ee03c048bad22fce4fcc2b3/core/dbt/include/global_project/macros/materializations/models/table/columns_spec_ddl.sql#L31-L35

emmyoop:
resolves #8183 

### Problem

Enforces contracts with numeric types that don't specify scale run the risk of failing contract enforcement.  This is actually confusing as they're not actually intentionally failing, the precision just needs to be specified so it can be enforced instead of applying the default.

### Solution

If the user has provided a numeric data type, which should specify precision/scale, without having specified anything except the type, raise a warning.

### Checklist

### DIFF
--- a/.changes/unreleased/Features-20230926-134728.yaml
+++ b/.changes/unreleased/Features-20230926-134728.yaml
@@ -1,0 +1,6 @@
+kind: Features
+body: Raise a wanring when a contracted model has a numeric field without scale defined
+time: 2023-09-26T13:47:28.645383-05:00
+custom:
+  Author: emmyoop
+  Issue: "8183"
--- a/.changes/unreleased/Features-20230926-134728.yaml
+++ b/.changes/unreleased/Features-20230926-134728.yaml
@@ -1,5 +1,5 @@
 kind: Features
-body: Raise a wanring when a contracted model has a numeric field without scale defined
+body: Raise a warning when a contracted model has a numeric field without scale defined
 time: 2023-09-26T13:47:28.645383-05:00
 custom:
   Author: emmyoop
--- a/core/dbt/include/global_project/macros/adapters/columns.sql
+++ b/core/dbt/include/global_project/macros/adapters/columns.sql
@@ -48,7 +48,7 @@
       {%- set col = columns[i] -%}
       {%- if col['data_type'] is not defined -%}
         {%- do col_err.append(col['name']) -%}
-      {#-- If this column's type is just 'numeric' missing precision/scale, raise a warning --#}
+      {#-- If this column's type is just 'numeric' then it is missing precision/scale, raise a warning --#}
       {#- elif api.Column.create(col['name'], col['data_type'].strip()).is_numeric() -#}
       {%- elif col['data_type'].strip().lower() in ('numeric', 'decimal', 'number') -%}
         {%- do col_naked_numeric.append(col['name']) -%}
--- a/core/dbt/include/global_project/macros/adapters/columns.sql
+++ b/core/dbt/include/global_project/macros/adapters/columns.sql
@@ -49,7 +49,6 @@
       {%- if col['data_type'] is not defined -%}
         {%- do col_err.append(col['name']) -%}
       {#-- If this column's type is just 'numeric' then it is missing precision/scale, raise a warning --#}
-      {#- elif api.Column.create(col['name'], col['data_type'].strip()).is_numeric() -#}
       {%- elif col['data_type'].strip().lower() in ('numeric', 'decimal', 'number') -%}
         {%- do col_naked_numeric.append(col['name']) -%}
       {%- endif -%}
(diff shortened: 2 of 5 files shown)


gshank (approved):

core/dbt/include/global_project/macros/adapters/columns.sql
```diff
@@ -36,24 +36,30 @@
...
     {%- set col_err = [] -%}
+    {%- set col_naked_numeric = [] -%}
     select
     {% for i in columns %}
       {%- set col = columns[i] -%}
       {%- if col['data_type'] is not defined -%}
-        {{ col_err.append(col['name']) }}
+        {%- do col_err.append(col['name']) -%}
+      {#-- If this column's type is just 'numeric' missing precision/scale, raise a warning --#}
+      {#- elif api.Column.create(col['name'], col['data_type'].strip()).is_numeric() -#}
```
colin-rogers-dbt, line 52: nit: delete commented code

core/dbt/include/global_project/macros/adapters/columns.sql
```diff
@@ -36,24 +36,30 @@
...
 {% macro default__get_empty_schema_sql(columns) %}
     {%- set col_err = [] -%}
+    {%- set col_naked_numeric = [] -%}
     select
     {% for i in columns %}
       {%- set col = columns[i] -%}
       {%- if col['data_type'] is not defined -%}
-        {{ col_err.append(col['name']) }}
+        {%- do col_err.append(col['name']) -%}
+      {#-- If this column's type is just 'numeric' missing precision/scale, raise a warning --#}
```
colin-rogers-dbt, line 51:
```suggestion
      {#-- If this column's type is just 'numeric' then it is missing precision/scale, raise a warning --#}
```

colin-rogers-dbt (commented):
just some nits otherwise LGTM

.changes/unreleased/Features-20230926-134728.yaml
```diff
@@ -0,0 +1,6 @@
+kind: Features
+body: Raise a wanring when a contracted model has a numeric field without scale defined
```
emmyoop, line 2:
```suggestion
body: Raise a warning when a contracted model has a numeric field without scale defined
```

core/dbt/include/global_project/macros/adapters/columns.sql
```diff
@@ -36,24 +36,30 @@
...
     {%- set col_err = [] -%}
+    {%- set col_naked_numeric = [] -%}
     select
     {% for i in columns %}
       {%- set col = columns[i] -%}
       {%- if col['data_type'] is not defined -%}
-        {{ col_err.append(col['name']) }}
+        {%- do col_err.append(col['name']) -%}
+      {#-- If this column's type is just 'numeric' then it is missing precision/scale, raise a warning --#}
+      {#- elif api.Column.create(col['name'], col['data_type'].strip()).is_numeric() -#}
```
emmyoop, line 52:
```suggestion
```

colin-rogers-dbt (approved):

PR #8743, opened 2023-09-28
Enable inline csv format in unit testing

### LINKED ISSUE #8626: [CT-3110] [Implementation] Support a configurable `format: csv` attribute on a `given` input and `expect` fixtures in unit tests
### Acceptance Criteria:
* extend the existing unit test spec to support `format: csv` 
  * by default, the format should be `dict` 
* support supplying __inline__ csv strings when format is `csv`
```
unit:
  - model: my_model
    tests:
      - name: test_my_model
        given:
          - input: ref('my_model_a')
            format: csv
            rows: |
            c
            3
          - input: ref('my_model_b')
            format: csv
            rows: |
            id, b
            1,  2
            2, 2
        expect:
          format: csv
          rows: |
          c
          3
```

### Technical Considerations: 
* this will need updates in both the `InputFixture` and `given` schemas. Currently they are different objects/types but we should consider unifying them to both inheriting from a `Fixture` class in order to share logic/structure for configurable formats. 

Question: should the default `format` be overridable? Perhaps at the project-level? It could be annoying to always have to specify `format: csv` in every `given` and `expect` configuration when the project has a general preference for `csv` over `dict` @graciegoheen. Perhaps a top-level `fixtures` config in dbt_project.yml?

gshank:
resolves #8626

### Problem

Users want to be able to use the csv format in addition to dictionaries in their unit tests.

### Solution

Add a "format" field to the given and expect structure, create an OutputFixture class (in addition to InputFixture). 

### Checklist


core/dbt/parser/unit_tests.py
```diff
@@ -138,6 +157,14 @@ def parse_unit_test_case(self, test_case: UnitTestDefinition):
             # Add unique ids of input_nodes to depends_on
             unit_test_node.depends_on.nodes.append(input_node.unique_id)
 
+    def _build_rows_from_csv(self, csv_string) -> List[Dict[str, Any]]:
```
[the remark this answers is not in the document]
gshank: Because it's not a file.

core/dbt/parser/unit_tests.py
```diff
@@ -59,14 +61,25 @@ def parse_unit_test_case(self, test_case: UnitTestDefinition):
         # for selection.
         # Note: no depends_on, that's added later using input nodes
         name = f"{test_case.model}__{test_case.name}"
+        if test_case.expect.format == "dict":
```
MichelleArk, line 64: nit: let's use the enum here in the comparison: 
```python
if test_case.expect.format == UnitTestFormat.Dict:
```

core/dbt/parser/unit_tests.py
```diff
@@ -118,16 +131,22 @@ def parse_unit_test_case(self, test_case: UnitTestDefinition):
             # TODO: package_name?
             input_name = f"{test_case.model}__{test_case.name}__{original_input_node.name}"
             input_unique_id = f"model.{package_name}.{input_name}"
+            rows: List[Dict[str, Any]]
+            if given.format == "csv":
```
MichelleArk, line 135: same comment re: using enum for comparison

core/dbt/parser/unit_tests.py
```diff
@@ -59,14 +61,25 @@ def parse_unit_test_case(self, test_case: UnitTestDefinition):
...
+            if isinstance(test_case.expect.rows, List):
+                expected_rows = test_case.expect.rows
+            else:
+                raise ParsingError("Wrong format for expected rows")
+        else:  # test_case.expect.format == "csv":
+            # build a dictionary from the csv string
+            if isinstance(test_case.expect.rows, str):
+                expected_rows = self._build_rows_from_csv(test_case.expect.rows)
+            else:
+                raise ParsingError("Wrong format for expected rows")
```
MichelleArk, line 74: I'm brainstorming ways that we could consolidate this logic and share it across `InputFixture` and `OutputFixture`. The simplest thing I can think of is adding a `parsed_rows` property on a shared `UnitTestFixture` base class that does this conditional logic to parse the raw `rows` given a `format` and convert it into a consistent internal representation. What do you think?

MichelleArk: An alternative could be to have a `UnitTestFixture` class that can be created from either an `InputFixture` or `OutputFixture` (making these just the external-facing interfaces). That way we could encapsulate the logic of parsing the rows based on the format, into a consistent internal fixture object and store the parsed rows as state on that object.

gshank: Yeah, I thought it would be a good idea to have some kind of combined logic but when I wrote this my brain was kind of offline :).   Since we know we're going to also support csv files soon, it should be a solution that makes that easy.  Let me think about whether a base class or derived class makes more sense...

core/dbt/contracts/graph/unparsed.py
```diff
@@ -741,15 +743,38 @@ class UnitTestFormat(StrEnum):
...
 
 
+class UnitTestFixture:
+    @property
+    def format(self) -> UnitTestFormat:
+        return UnitTestFormat.Dict
+
+    @property
+    def rows(self) -> Union[str, List[Dict[str, Any]]]:
+        return []
```
MichelleArk, line 753: is it necessary for these to be properties? Could they instead be attributes that are inherited by UnitTestInputFixture and UnitTestOutputFixture?

e.g.
```python
rows: Union[str, List[Dict[str, Any]]] = ""
format: UnitTestFormat = UnitTestFormat.Dict
```

core/dbt/parser/unit_tests.py
```diff
@@ -189,6 +190,26 @@ def parse(self) -> ParseResult:
...
+                    test.expect.format == UnitTestFormat.Dict
+                    and not isinstance(test.expect.rows, list)
+                ) or (
+                    test.expect.format == UnitTestFormat.CSV
+                    and not isinstance(test.expect.rows, str)
+                ):
+                    raise ParsingError(
+                        f"Expected rows in invalid format for unit test {test.name}"
+                    )
+
```
MichelleArk, line 212: I think ideally this kind of validation happens on the `UnitTestFixture` itself so callers/users of the instantiated class don't need to worry about it. What do you think about making it a post_init method on UnitTestFixture?

core/dbt/contracts/graph/unparsed.py
gshank: If we do that then we run into the frustrating issue of fields without defaults can't come after fields with defaults issue and have to split them out into a special class and do a different order. I'd kind of rather not.

core/dbt/parser/unit_tests.py
gshank: I made a separate "validate_fixture" class method and called it when parsing. A post_init method would work too, but I'm just not all that fond of doing validation there. We generally validate as a separate step, so this is a bit more like that.

core/dbt/contracts/graph/unparsed.py
```diff
@@ -736,10 +736,22 @@ def normalize_date(d: Optional[datetime.date]) -> Optional[datetime.datetime]:
...
     input: str
-    rows: List[Dict[str, Any]] = field(default_factory=list)
+    rows: Union[str, List[Dict[str, Any]]] = ""
+    format: UnitTestFormat = UnitTestFormat.Dict
+
+
+@dataclass
+class OutputFixture(dbtClassMixin):
+    rows: Union[str, List[Dict[str, Any]]] = ""
+    format: UnitTestFormat = UnitTestFormat.Dict
```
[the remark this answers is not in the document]
gshank: Names have been changed.

core/dbt/contracts/graph/unparsed.py
```diff
@@ -736,10 +738,54 @@ def normalize_date(d: Optional[datetime.date]) -> Optional[datetime.datetime]:
...
+            assert isinstance(self.rows, str)
+            dummy_file = StringIO(self.rows)
+            reader = csv.DictReader(dummy_file)
+            rows = []
+            for row in reader:
+                rows.append(row)
+            return rows
+
+    @classmethod
+    def validate_fixture(cls, fixture, fixture_type, test_name):
```
MichelleArk, line 769: What is the utility of this as a classmethod given it is passed an instance of itself? Perhaps a regular method would be more straightforward. Also, types on this signature would be great!

core/dbt/contracts/graph/unparsed.py
```diff
@@ -736,10 +738,54 @@ def normalize_date(d: Optional[datetime.date]) -> Optional[datetime.datetime]:
...
+        if (fixture.format == UnitTestFormat.Dict and not isinstance(fixture.rows, list)) or (
+            fixture.format == UnitTestFormat.CSV and not isinstance(fixture.rows, str)
+        ):
+            raise ParsingError(
+                f"Unit test {test_name} has {fixture_type} rows which do not match format"
```
MichelleArk, line 774: Could make this message a tiny bit clearer if we also provided the format, e.g "do not match format: {fixture.format}"

MichelleArk (approved):
some small suggestions, otherwise LGTM!

core/dbt/contracts/graph/unparsed.py
gshank: Oh, right. I originally created it as an override of the 'validate' class method, but that didn't work because it had to have a different signature.

core/dbt/contracts/graph/unparsed.py
gshank: Done.

PR #8747, opened 2023-09-29
fix duplication of docs issues

### LINKED ISSUE #8746: [CT-3171] [Bug] user docs autogenerated issues sometimes are duplicated

### Current Behavior

We're seeing some weirdness with the GitHub Action. See how this [Core PR](https://github.com/dbt-labs/dbt-core/pull/8589) created two bot comments that created two docs issues: [this one ](https://github.com/dbt-labs/docs.getdbt.com/issues/4133)and [this one](https://github.com/dbt-labs/docs.getdbt.com/issues/4134).

### Expected Behavior

Only 1 issue gets generated per PR - this can be accomplished by checking that the label added by the labeled event was `user docs`

### Steps To Reproduce

This is really a timing issue.  In the case of the above PR, a label was added right after the PR was merged causing the workflow to trigger twice and since they ran simultaneously neither saw the comment the other generated in time to know to skip creating a new issue in the docs.getdbt.com repo.

### Relevant log output

_No response_

### Environment

```markdown
- OS:
- Python:
- dbt:
```

_No response_

### Additional Context

_No response_

emmyoop:
resolves #8746 

### Problem

Workflow has double generated issues in the other repo

### Solution

Ensure the label added in the labeled event is `user docs` and only run if that's true or the PR is in a merged state with the correct label.

### Checklist

### DIFF
--- a/.github/workflows/docs-issue.yml
+++ b/.github/workflows/docs-issue.yml
@@ -27,7 +27,13 @@ permissions:
 
 jobs:
   open_issues:
-    if: contains( github.event.pull_request.labels.*.name, 'user docs') && github.event.pull_request.merged == true
+    # we only want to run this when the PR has been merged or the label in the labeled event is `user docs`.  Otherwise it runs the
+    # risk of duplicaton of issues being created due to merge and label both triggering this workflow to run and neither having
+    # generating the comment before the other runs.  This lives here instead of the shared workflow because this is where we
+    # decide if it should run or not.
+    if: |
+      (contains( github.event.pull_request.labels.*.name, 'user docs') && github.event.pull_request.merged == true) ||
+      (github.event.label.name == 'user docs' && github.event.action == 'labeled' )
     uses: dbt-labs/actions/.github/workflows/open-issue-in-repo.yml@main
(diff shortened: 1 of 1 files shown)


emmyoop:
> Can we change this logic to check for a docs issue instead, and only create one if it doesn't exist?

We can't because we don't know what the issue is.  The main action flow is:

1. Check if a templated comment exists on the PR
2. If the Comment does not exist open an issue in the designated repo
3. Leave a comment on the PR with a link to the newly opened issue

So that takes care of most cases where this is a problem.  The reason it generated 2 issues on a PR was because we merged the PR and immediately added a backport label.  It triggered this twice (for both closed and labeled) and they ran so close together that neither left the comment before the other checked for a comment.  

This fix doesn't solve for the case of someone merging a PR and specifically adding the `user docs` label immediately but I'm not sure there's much we can do about that based on how this kind of state exists in GitHub and wanting the flexibility to be able to add the label before or after a PR is merged.  The `user docs` label is also not generally a label we wait to add until the last second.  The backport labels are since they don't trigger when you add them before merge.

PR #8748, opened 2023-09-29
Prototype: `test` materialization that allows `materialization` to be set in the `config()`

dbeatty10:
regarding #6914

🚧 This is a prototype. It currently includes logging statements an inline comments for demo purposes.

### Problem

The failures of dbt tests can only be done ephemerally or stored in a table [stored in a table](https://docs.getdbt.com/reference/resource-configs/store_failures); there is no way to capture test results as a view or a materialized view. So users need to re-run `dbt test` (or `dbt build`) to see the results that are up-to-date. Of course those results can immediately go stale 😬

#### Big idea

Enable test results as a view or a materialized view via the [`materialized` config](https://docs.getdbt.com/docs/build/materializations) so that they will always show the current failures without the need to run dbt.
 
Users can specify the materialization within the config block in their tests like any the following:
- `materialized="table"` (isomorphic to [`--store-failures`](https://docs.getdbt.com/reference/resource-configs/store_failures))
- `materialized="view"` **NEW**
- `materialized="materialized_view"` **NEW**

#### Implementation details
Flesh out the ideas in [this Loom](https://www.loom.com/share/725ca16f9cab450b8c4c38ebd601a7b6) by combining https://github.com/dbt-labs/dbt-core/pull/8653 with the approach used by the [`clone` materialization](https://github.com/dbt-labs/dbt-core/blob/408a78985a647c52125e344c64747b21c02651f7/core/dbt/include/global_project/macros/materializations/models/clone/clone.sql#L54-L62) (introduced in #7881).

### Checklist

### DIFF
--- a/core/dbt/contracts/graph/model_config.py
+++ b/core/dbt/contracts/graph/model_config.py
@@ -583,12 +583,6 @@ class TestConfig(NodeAndTestConfig):
                     return False
         return True
 
-    @classmethod
-    def validate(cls, data):
-        super().validate(data)
-        if data.get("materialized") and data.get("materialized") != "test":
-            raise ValidationError("A test must have a materialized value of 'test'")
-
 
 @dataclass
 class EmptySnapshotConfig(NodeConfig):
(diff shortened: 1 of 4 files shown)


dbeatty10:
Closing since this was just a prototype to support https://github.com/dbt-labs/dbt-core/pull/8653

PR #8754, opened 2023-09-29
Add meta to SemanticModels

emmyoop:
resolves #8511 

### Problem

No meta attribute on SemanticModels

### Solution

Add meta under configs

### Checklist

- [ ] I have read [the contributing guide](https://github.com/dbt-labs/dbt-core/blob/main/CONTRIBUTING.md) and understand what's expected of me  
- [ ] I have run this code in development and it appears to resolve the stated issue  
- [ ] This PR includes tests, or tests are not required/relevant for this PR
- [ ] This PR has no interface changes (e.g. macros, cli, logs, json artifacts, config files, adapter interface, etc) or this PR has already received feedback and approval from Product or DX
- [ ] This PR includes [type annotations](https://docs.python.org/3/library/typing.html) for new and modified functions

Issue #8511, opened 2023-08-29
[CT-3042] Support `meta` attribute for `SemanticModel` nodes


QMalcolm:
It was recently noted by a user in a [issue](https://github.com/dbt-labs/metricflow/issues/739) created on the [MetricFlow project](https://github.com/dbt-labs/metricflow) that the `meta` attribute isn't supported for `SemanticModel` nodes. As `meta` is a manifest node attribute (and not needed currently in MetricFlow or MetricFlowServer), I've moved the request here.

# Acceptance Criteria
- A `meta` field which is an arbitrary dictionary can be specified for `SemanticModel`s and makes it to the corresponding `SemanticModel` node in the manifest
- A test exists for ensuring this functionality

QMalcolm:
Is `meta` an attribute we require for all manifest nodes? If so perhaps we should consider creating a `ManifestNode` protocol that defines the fields which all nodes must support, and then use that protocol for some type checking.

QMalcolm:
It's worth noting that the original [issue](https://github.com/dbt-labs/metricflow/issues/739) denoted a desire to have `meta` propagated to the `semantic_manifest.json`. This is not currently possible, this is because

1.  [dbt-semantic-interfaces](https://github.com/dbt-labs/dbt-semantic-interfaces/tree/ef8a893cc6e65f695db46356b8471a039246456b) de-serialization does not allow additional properties
2. [dbt-semantic-interfaces](https://github.com/dbt-labs/dbt-semantic-interfaces/tree/ef8a893cc6e65f695db46356b8471a039246456b) protocols only contain attributes needed by both `dbt-core` and `MetricFlow`
3. the `meta` attribute is not needed in `MetricFlow`

Thus because of (3), `meta` is not part of the  DSI protocol spec (2).  Which in turn means we can't put it in the `semantic_manifest.json` because (1) means de-serialization would break. It is possible that long term we can change the de-serialization process in `DSI` to allow additional properties to exist in the serialization without causing de-serialization exceptions, but that is a separate conversation.

jtcohen6:
@QMalcolm There's a little bit of history here:
- Originally, `meta` was a top-level attribute of (most/all) manifest nodes
- We then allowed users to start defining it as a "config": in `config:` blocks in yaml files, and for all nodes in a subdirectory of the resource type in `dbt_project.yml`
[...]

PR #8766, opened 2023-10-04
remove top level meta attribute from SemanticModels

emmyoop:
related to #8511

### Problem

`meta` is a top level attribute in #8754 but doesn't need to be

### Solution

Modify to only allow meta under the config key.

### Checklist

### DIFF
--- a/.changes/unreleased/Features-20230929-170945.yaml
+++ b/.changes/unreleased/Features-20230929-170945.yaml
@@ -1,5 +1,5 @@
 kind: Features
-body: Add meta attribute to SemanticModels
+body: Add meta attribute to SemanticModels config
 time: 2023-09-29T17:09:45.0354-05:00
 custom:
   Author: emmyoop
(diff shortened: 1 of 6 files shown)


emmyoop:
Code coverage went down but all I did was remove code.

QMalcolm (approved):
🚀 Looks great! Sorry about the thrash 🙃

emmyoop:
@QMalcolm it was a good question and the code is a tiny bit cleaner now!  Worth it.

dbeatty10:
The current docs for the [`meta`](https://docs.getdbt.com/reference/resource-configs/meta) config says the following:

> Depending on the resource you're configuring, `meta` may be available within the config property, or as a top-level key. (For backwards compatibility, `meta` is always supported as a top-level key, though without the capabilities of `config` inheritance.)

Since this is no longer accurate, I'll update it [here](https://github.com/dbt-labs/docs.getdbt.com/pull/4218) to something like the following:

> Depending on the resource you're configuring, `meta` may be available within the config property, and/or as a top-level key. (For backwards compatibility, `meta` is often (but not always) supported as a top-level key, though without the capabilities of `config` inheritance.)

PR #8772, opened 2023-10-04
Selectors in docs generate limits catalog generation

### LINKED ISSUE #6014: Respect node selection in catalog queries run by `docs generate`
When users specify `dbt docs generate --select <selection_criteria>`, that limits the set of project resources for which dbt will generate _compile SQL_. This is implicit behavior in the `GenerateTask`, because it inherits from the `CompileTask`, where that behavior is defined.

We should additionally make `docs generate` respect node selection for the purposes of generating catalogs, which can be very very expensive.

To really have an effect, this change would want to be paired with https://github.com/dbt-labs/dbt-core/issues/4997, which would filter down our catalog queries to only the ones that are actually selected. Without that change, while we would successfully limit ourselves to cataloging metadata on only the schema(s) containing the selected resource(s), we'd still catalog the entire schema, which could include hundreds/thousands of unrelated resources.

### Use case

- I'm working in a project with hundreds/thousands of models
- I just made changes to one model, e.g. by updating its SQL and adding descriptions
- I ran that one model: `dbt run --select my_one_model`
- Now I want to view documentation for _just_ that one model: `dbt docs generate --select my_one_model`
- I understand that catalog info (columns, table stats, etc) will be missing for all other models

### Technical implementation

- Pass in a set of relations (from [`get_node_selector`](https://github.com/dbt-labs/dbt-core/blob/7dbdfc88e001b59598af7f5902f6604eb30fb26f/core/dbt/task/compile.py#L44), plus selected `sources`) [into `adapter.get_catalog` here](https://github.com/dbt-labs/dbt-core/blob/7dbdfc88e001b59598af7f5902f6604eb30fb26f/core/dbt/task/generate.py#L240)
- Instead of [`adapter._get_catalog_schemas`](https://github.com/dbt-labs/dbt-core/blob/7dbdfc88e001b59598af7f5902f6604eb30fb26f/core/dbt/adapters/base/impl.py#L378) accessing `manifest.nodes` and `manifest.sources` directly, it should limit itself to the nodes/sources actually included in the selection logic

### Note

To be clear, this would _not_ accomplish what users tend to ask for when asking for more powerful `docs generate --select`, which is to produce a `manifest.json` with only the selected resources. Relevant discussions: https://github.com/dbt-labs/dbt-core/discussions/5096, https://github.com/dbt-labs/dbt-core/discussions/5244

gshank:
resolves #6014

### Problem

Catalog generation is expensive, so we want to limit catalog queries to only the selected nodes.

### Solution

Retrieve the selected node unique_ids and pass them to the methods used to get the catalog information. 

### Docs

In the past using a selector on docs generate would only affect which nodes were compiled. With this change, selectors will limit the information in the catalog.json.

### Checklist

### DIFF
--- a/.changes/unreleased/Features-20231004-170155.yaml
+++ b/.changes/unreleased/Features-20231004-170155.yaml
@@ -1,0 +1,6 @@
+kind: Features
+body: Selectors with docs generate limits catalog generation
+time: 2023-10-04T17:01:55.845479-04:00
+custom:
+  Author: gshank
+  Issue: "6014"
--- a/core/dbt/adapters/base/impl.py
+++ b/core/dbt/adapters/base/impl.py
@@ -434,9 +434,8 @@ class BaseAdapter(metaclass=AdapterMeta):
         return info_schema_name_map
 
     def _get_catalog_relations_by_info_schema(
-        self, manifest: Manifest
+        self, relations
     ) -> Dict[InformationSchema, List[BaseRelation]]:
-        relations = self._get_catalog_relations(manifest)
         relations_by_info_schema: Dict[InformationSchema, List[BaseRelation]] = dict()
         for relation in relations:
             info_schema = relation.information_schema_only()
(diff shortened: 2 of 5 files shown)


peterallenwebb (approved):
Looks good!

PR #8795, opened 2023-10-08
Add support for getting freshness from DBMS metadata

### LINKED ISSUE #8704: [CT-3149] [Implementation] allow source freshness to be evaluated from warehouse metadata tables, instead of running select max(loaded_at_field) ... SQL queries
### Housekeeping

### Short description

From https://github.com/dbt-labs/dbt-core/issues/7012

We should allow source freshness metadata to be collected from warehouse metadata tables when possible. 

In this case, the `loaded_at_field` becomes optional for source freshness configuration:
- if you provide `loaded_at_field`, we will use the SQL query `select max(loaded_at_field) ... from ...`
- if you don’t provide `loaded_at_field`, we will get the source freshness from the warehouse metadata table
- if the user omits the `loaded_at_field ` while using an adapter without metadata support we will raise a warning

A significant benefit is that it doesn't rely on a specific column being consistently available - it provides a more generic implementation for the masses.

### Acceptance criteria

- Add a new static boolean property to the base adapter called METADATA_FRESHNESS_SUPPORT. It will default to False.
- When an adapter implementation overrides this property and sets it to True, and the user has requested that freshness for a source be determined via metadata, dbt-core will look for a macro called get_relation_last_modified(information_schema, relations). The new macro will accept the name of the information schema and a List[BaseRelation]. It will issue and return the result of a query with the following output columns: `database`, `schema`, `identifier`, `last_modified`, and `snapshotted_at`. By accepting a list, we will allow core to batch metadata-freshness requests in the future.
- In the YML description of a source, freshness can be requested by metadata, by excluding `loaded_at_field` from freshness config.
- If the user omits the loaded at field while using an adapter without metadata support we will raise a warning
### Impact to Other Teams

Impact adapters

No

### Context

_No response_

peterallenwebb:
resolves #8704

### Checklist


core/dbt/adapters/base/impl.py
```diff
@@ -92,6 +95,14 @@ def _expect_row_value(key: str, row: agate.Row):
     return row[key]
 
 
+def _get_column_value_uncased(key: str, row: agate.Row) -> Any:
```
MichelleArk, line 98: this might be a better fit in core/dbt/clients/agate_helper.py

core/dbt/adapters/base/impl.py
```diff
@@ -1518,12 +1608,14 @@ def render_model_constraint(cls, constraint: ModelLevelConstraint) -> Optional[s
         else:
             return None
 
-    @classmethod
-    def has_feature(cls, feature: AdapterFeature) -> bool:
-        # The base adapter implementation does not implement any optional
-        # features, so always return false. Adapters which wish to provide
-        # optional features will have to override this function.
-        return False
+    def capabilities(self) -> List[CapabilitySupport]:
```
MichelleArk, line 1611: What do you think about having this method a `Dict[Capability, CapabilitySupport]` (perhaps wrapping it in our own type with a default lookup) instead of a `List[CapabilitySupport]`? That way, adapters just need to implement this method and `capability_support` could do a lookup on `capabilities`

This assumes that capabilities won't be costly to call, and will largely look like returning a static dictionary as you've done for the postgres implemention

core/dbt/adapters/base/impl.py
```diff
@@ -1518,12 +1608,14 @@ def render_model_constraint(cls, constraint: ModelLevelConstraint) -> Optional[s
...
+        return []
```
MichelleArk, line 1612: The default could also be each capability in `Capability` with an `Unkown` support

core/dbt/adapters/base/impl.py
```diff
@@ -1518,12 +1608,14 @@ def render_model_constraint(cls, constraint: ModelLevelConstraint) -> Optional[s
         else:
             return None
```
MichelleArk, line 1610: should this also be a classmethod? similar to its override in plugins/postgres/dbt/adapters/postgres/impl.py

core/dbt/adapters/base/impl.py
```diff
@@ -162,12 +175,62 @@ def submit(self, compiled_code: str) -> Any:
...
+    support: Support
+    first_version: Optional[str] = None
+
+    def __bool__(self):
+        return self.support == Support.Versioned or self.support == Support.Full
+
+
+# Make python 3.8 happy with this generic type, remove when we support 3.9+
+if TYPE_CHECKING:
+    CapabilityUserDict = UserDict[Capability, CapabilitySupport]
```
mikealfare, line 220: I'm still not clear on why we need a `dict` to assign a `CapabilitySupport` to a `Capability` when an attribute of `CapabilitySupport` is its `Capability`. I get the value in being able to look up the support by capability, but perhaps we don't need `capability` on `CapabilitySupport` then?

core/dbt/adapters/base/impl.py
```diff
@@ -1238,13 +1301,52 @@ def calculate_freshness(
...
+            "information_schema": source.information_schema_only(),
+            "relations": [source],
+        }
+        result = self.execute_macro(
+            GET_RELATION_LAST_MODIFIED_MACRO_NAME, kwargs=kwargs, manifest=manifest
+        )
+        adapter_response, table = result.response, result.table  # type: ignore[attr-defined]
+
+        try:
+            assert len(table) == 1  # There should be one row since we requested one relation
```
mikealfare, line 1326: I just learned this recently, `assert` does nothing in functional code. The test runner flips a switch under the hood to make it do something, but you would never want that switch on in production. If you want to catch an edge case here, then you need to use a regular if block.

core/dbt/adapters/base/impl.py
```diff
@@ -1238,13 +1301,52 @@ def calculate_freshness(
...
+            age = (snapshotted_at - max_loaded_at).total_seconds()
+
+            freshness: FreshnessResponse = {
+                "max_loaded_at": max_loaded_at,
+                "snapshotted_at": snapshotted_at,
+                "age": age,
+            }
+
+            return adapter_response, freshness
+        except Exception:
```
mikealfare, line 1347: This is a long `try` block and a generic `Exception`. Can we limit the `try` block to just the piece that we think will raise the exception, and then also be more specific in the exception that we're catching? For example, I think you're testing for when `len(table) != 1`. Between that and the assert comment, can we move those into an if block and move everything else out of the try block?:

```python
if len(table) != 1:
    raise MacroResultError(...)

row = table[0]
...
```

core/dbt/adapters/base/impl.py
peterallenwebb: > perhaps we don't need capability on CapabilitySupport then?

Agree that it's not necessary in the current design, and I'll get rid of that.

core/dbt/task/freshness.py
```diff
@@ -95,26 +99,42 @@ def from_run_result(self, result, start_time, timing_info):
...
+            elif self.adapter.capability_support(Capability.TableLastModifiedMetadata):
+                if compiled_node.freshness.filter is not None:
+                    fire_event(
+                        Note(
+                            f"A filter cannot be applied to a metadata freshness check on source '{compiled_node.name}'.",
+                            EventLevel.WARN,
+                        )
+                    )
+
+                adapter_response, freshness = self.adapter.calculate_freshness_from_metadata(
```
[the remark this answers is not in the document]
mikealfare: I would wrap the macro execution itself in `calculate_freshness_from_metadata` if you're guarding against a failed execution of the macro.

core/dbt/adapters/base/impl.py
peterallenwebb: My ambivalence around using static-methods/classmethods is that they make testing difficult. I don't mind using them for pure functions, but when working on a MockAdapter class for testing (not part of this review) I immediately hit the problem that I have to declare an entirely new class to override the features reported by the mock.

That said, enough people have voted for classmethod here that I'll change it for both

core/dbt/adapters/base/impl.py
```diff
@@ -162,12 +175,62 @@ def submit(self, compiled_code: str) -> Any:
         raise NotImplementedError("PythonJobHelper submit function is not implemented yet")
 
 
-class AdapterFeature(str, Enum):
+class Capability(str, Enum):
```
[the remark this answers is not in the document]
peterallenwebb: Doing this now.

core/dbt/adapters/base/impl.py
peterallenwebb: Done.

core/dbt/adapters/base/impl.py
peterallenwebb: Implemented this suggestion.

core/dbt/adapters/base/impl.py
peterallenwebb: Moved.

plugins/postgres/dbt/adapters/postgres/impl.py
```diff
@@ -148,6 +162,5 @@ def valid_incremental_strategies(self):
     def debug_query(self):
         self.execute("select 1 as id")
 
-    @classmethod
-    def has_feature(cls, feature: AdapterFeature) -> bool:
-        return feature in cls.SUPPORTED_FEATURES
+    def capabilities(self) -> CapabilityDict:
```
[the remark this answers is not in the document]
peterallenwebb: Done.

core/dbt/adapters/base/impl.py
peterallenwebb: Done.

core/dbt/adapters/base/impl.py
peterallenwebb: I at least tightened up the scope here.

core/dbt/adapters/base/impl.py
peterallenwebb: Got rid of this assertion.

core/dbt/task/freshness.py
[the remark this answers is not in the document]
peterallenwebb: We may want to think through how we handle permissions based failures in the future, but I'm just declaring it out of scope for this issue based on time constraints.

plugins/postgres/dbt/adapters/postgres/impl.py
```diff
@@ -1,9 +1,15 @@
 from datetime import datetime
 from dataclasses import dataclass
-from typing import Optional, Set, List, Any
+from typing import Any, Optional, Set, List
 
 from dbt.adapters.base.meta import available
-from dbt.adapters.base.impl import AdapterConfig, AdapterFeature, ConstraintSupport
+from dbt.adapters.base.impl import (
```
mikealfare, line 6: We should pull these from its new location `dbt.adapters.capability`

MichelleArk (approved):

core/dbt/parser/sources.py
```diff
@@ -1,6 +1,8 @@
 import itertools
 from pathlib import Path
 from typing import Iterable, Dict, Optional, Set, Any, List
+
+from dbt.adapters.base.impl import Capability
```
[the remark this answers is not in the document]
peterallenwebb: Done.

core/dbt/task/freshness.py
```diff
@@ -20,12 +21,15 @@
...
     LogStartLine,
     LogFreshnessResult,
+    Note,
 )
 from dbt.node_types import NodeType
 
 from dbt.graph import ResourceTypeSelector
 from dbt.contracts.graph.nodes import SourceDefinition
-
+from ..adapters.base.impl import Capability
```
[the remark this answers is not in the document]
peterallenwebb: Done.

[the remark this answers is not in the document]
peterallenwebb: Done.

plugins/postgres/dbt/adapters/postgres/impl.py
peterallenwebb: Done.

core/dbt/adapters/capability.py
```diff
@@ -0,0 +1,55 @@
...
+
+
+# Make python 3.8 happy with this generic type, remove when we support 3.9+
+if TYPE_CHECKING:
+    CapabilityUserDict = UserDict[Capability, CapabilitySupport]
+else:
+    CapabilityUserDict = UserDict
+
+
+class CapabilityDict(CapabilityUserDict):
```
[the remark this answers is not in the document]
peterallenwebb: This worked well!

PR #8798, opened 2023-10-09
Add SavedQuery nodes

### LINKED ISSUE #8594: [CT-3091] Add `SavedQuery` node
## Context
DSI will be adding a `SavedQuery` protocol which the semantic manifest should have a list of under `saved_queries`. This is outlined in https://github.com/dbt-labs/dbt-semantic-interfaces/issues/144. This is necessary to support the `saved query` functionality in MetricFlow (https://github.com/dbt-labs/metricflow/issues/765).

### Things we do know

- `SavedQuery` nodes will depend on `Metric` nodes
- `SavedQuery` nodes will have
  - a `name`
  - an optional `description`
  - a `query` object that has the following
    - `metrics` a list of strings (that they're valid metrics is handled by the DSI provided validator)
    - `group_bys` a list of strings (that they have a valid format is handled by the DSI provided validator)
    - `where` a `WhereFilter` object (should be specified as a string by the user, and we convert to an object when parsing the unparsed node)

### Necessary work
- Define an unparsed version of the node
- Define a manifest version of the node
- Define schema_yaml_reader process for going from the unparsed node to the manifest node
- Define a lookup utility for resolving`SavedQuery` nodes
- Handle `SavedQuery` unique_ids in `manifest.expect`
- partial parsing of `SavedQuery` nodes

QMalcolm:
resolves #8594 

### Problem
We need to support a new node type `SavedQuery` for 1.7 of core. This is work is to support MetricFlow and the Semantic Layer.

### Solution

We

- Added SavedQuery nodes to the manifest
- Enabled parsing of SavedQuery nodes
- Enabled partial parsing of SavedQuery nodes
- Enabled selector method of SavedQuery nodes
- Enabled `doc` support for SavedQuery node decsriptions
- Provided upgrade support from v10 (1.6) to v11 (1.7)
- Created a Lookup utility for SavedQuery nodes

### Checklist


core/dbt/contracts/graph/manifest.py
```diff
@@ -299,6 +300,41 @@ def perform_lookup(self, unique_id: UniqueID, manifest: "Manifest") -> Metric:
...
+
+        self.storage[saved_query.search_name][saved_query.package_name] = saved_query.unique_id
+
+    def populate(self, manifest):
+        for saved_query in manifest.saved_queries.values():
+            if hasattr(saved_query, "name"):
+                self.add_saved_query(saved_query)
+
+    def perform_lookup(self, unique_id: UniqueID, manifest: "Manifest") -> SavedQuery:
+        if unique_id not in manifest.metrics:
```
MichelleArk, line 331: should this be a lookup in manifest.saved_queries?

core/dbt/parser/manifest.py
```diff
@@ -1579,14 +1605,19 @@ def _process_metric_node(
...
-    node: Union[ManifestNode, Metric, Exposure],
+    node: Union[ManifestNode, Metric, Exposure, SavedQuery],
 ):
     """Given a manifest and a node in that manifest, process its metrics"""
 
+    metrics: List[List[str]]
     if isinstance(node, SeedNode):
         return
+    elif isinstance(node, SavedQuery):
+        metrics = [[metric] for metric in node.metrics]
```
MichelleArk, line 1616: for my own understanding - why is this necessary? what does node.metrics look like?

QMalcolm: This necessary because the for loop that begins on 1620 expects each "metric" in the list of "metrics" to be a list of either 1 or 2 strings (i.e. `List[List[str]]`. Where the first string is the target_metric_name and the second (if it exists) is the `target_metric_package`. Saved query node metrics don't support package targeting. and it's `metrics` is just a list of strings (i.e. `List[str]`), with each string being a `target_metric_name`. Thus instead of handling them differently within the for loop that resolves the metrics, it was easier to just massage the saved query metrics to have the same shape.

core/dbt/contracts/graph/manifest.py
QMalcolm: Yes indeed! Copy and paste strikes again 🙃

MichelleArk (approved):

peterallenwebb:
@QMalcolm I haven't looked into the intended semantics of this new node type, but one file you might have missed is [compilation.py](https://github.com/dbt-labs/dbt-core/blob/main/core/dbt/compilation.py). I also missed it in my initial implementation of `SemanticModel`. Take a look at the spots that the `semantic_models` collection is mentioned in that file. In particular, without changes to `link_graph` and `link_node` functions, the new node type probably will not be treated like it is "part of the DAG" for some purposes.

Issue #8751, opened 2023-09-29
[Bug] Postgres + Grants + Username-With-Dashes does not work


tomans-spirent:
### Is this a new bug in dbt-core?

- [X] I believe this is a new bug in dbt-core
- [X] I have searched the existing issues, and I could not find an existing issue for this bug

### Current Behavior

When I configure grants in DBT for a postgres database and a db role like 'foo-bar' I get postgres syntax errors because the name is not quoted. 

If I try to quote it like "\"foo-bar\"" it works for the grants part but breaks in the incremental grants part. Looks like some places do not strip and some places do strip those extra quotes I added to my db role name. 

### Expected Behavior

I should not have to put in extra quotes - DBT should be able to create the appropriate database syntax for supported role names. 

### Steps To Reproduce

Create a postgres role with dashes in the name. 

Try to have DBT assign grants to that user - it will break. 
Add manual quotes - it will work. 

Try to make a incremental model with those same grants - it will break. 

### Environment

```markdown
- OS: macOS ventura 13.6
- Python: 3.10.12
- dbt: 1.6.2
```

### Which database adapter are you using with dbt?

postgres

tomans-spirent:
As an update and workaround to this issue - I have been able to leverage the post-hook feature of models to apply the grants myself - where I just make sure I properly quote my db role name.

dbeatty10:
Thanks for reporting this @tomans-spirent !

Could you try [this workaround](https://github.com/dbt-labs/dbt-core/issues/6444#issuecomment-1354216431) and see if it also works for you?

tomans-spirent:
Seems like a similar issue - but no such luck if I add that revoke macro to my project.

Some extra context from my dbt run log output:

```
Database Error in model my_model (models/my_group/my_model.sql)
  syntax error at or near "-"
  LINE 4: ...db"."schema"."my_model" to grantee-with-dash;
```

With the following in my dbt_project.yml 

```yaml
models:
  my_group:
    +grants:
      select: ["grantee-with-dash"]
```

dbeatty10:
Ah, sorry I only gave you the revoke-side of the story @tomans-spirent ! 😅 

### Related issues

It looks to me like there are two different (but very related) things going on:
1. dbt-postgres/redshift doesn't quote username when granting ([#8751](https://github.com/dbt-labs/dbt-core/issues/8751))
[...]

PR #8809, opened 2023-10-10
Add quotes to usernames when revoking and granting table permissions

### LINKED ISSUE #6444: [Bug] Grants fails to correctly quote username when revoking

### Current Behavior

As part of the new grants feature, https://docs.getdbt.com/reference/resource-configs/grants dbt will attempt to revoke users not part of the grant before reapplying users listed in the grant. 
DBT does this (at least in the Redshift adaptor ) by pulling the table permissions and explicitly looping through every user then using a revoke command like 
revoke select on "database"."schema"."table" from user1, user2 ;
The user is not quoted so a user like user.name will fail with 
syntax error at or near "."
  LINE 2: ...user1, user.name, user2...

I am fairly certain the macro in https://github.com/dbt-labs/dbt-core/blob/main/core/dbt/include/global_project/macros/adapters/apply_grants.sql line 82 just needs some quotes

### Expected Behavior

the revoke command should quote users
revoke select on "database"."schema"."table" from "user1", "user2" ;

### Steps To Reproduce

Create a user with a period in the name for example in Redshift 
create user "first.last";

Create a simple dbt incremental model and user the grant syntax to give permission to that table within dbt.

Run DBT twice (on the first run the revoke macro is not run as the table is new)

### Relevant log output

_No response_

### Environment

```markdown
- OS:Ubuntu 20.04
- Python: 3.8.10 
- dbt: 1.3.1
```

redshift

### Additional Context

_No response_

barton996:
resolves dbt-labs/dbt-adapters#156
resolves dbt-labs/dbt-postgres#55

### Problem
For models with grants, where the corresponding table is not new (model has been run before), dbt revokes all grants on this table by looping through all users with table permissions.

The macro currently does not surround the user names that it reads from the table permissions with quotes, meaning that revoke statements on users with usernames like first.last will fail.

This is problematic when granting to a user GROUP, as you cannot control what is returned by the table permissions query using your grant yaml config e.g. grants: select: ["GROUP developers"].

### Solution
apply_grants.sql jinja template is updated so that usernames are surrounded with adapter.quote() character when granting and revoking.

### Checklist

- [] This PR has no interface changes (e.g. macros, cli, logs, json artifacts, config files, adapter interface, etc) or this PR has already received feedback and approval from Product or DX

### DIFF
--- a/.changes/unreleased/Fixes-20231010-193527.yaml
+++ b/.changes/unreleased/Fixes-20231010-193527.yaml
@@ -1,0 +1,6 @@
+kind: Fixes
+body: Add quotes to usernames when revoking and granting table permissions
+time: 2023-10-10T19:35:27.223761+01:00
+custom:
+  Author: barton996
+  Issue: 6444 8751
--- a/core/dbt/include/global_project/macros/adapters/apply_grants.sql
+++ b/core/dbt/include/global_project/macros/adapters/apply_grants.sql
@@ -70,7 +70,11 @@
 {% endmacro %}
 
 {%- macro default__get_grant_sql(relation, privilege, grantees) -%}
-    grant {{ privilege }} on {{ relation }} to {{ grantees | join(', ') }}
+    grant {{ privilege }} on {{ relation }} to
+    {% for grantee in grantees %}
+    {{ adapter.quote(grantee) }}
+    {% if not loop.last %},{% endif %}
+    {% endfor %}
 {%- endmacro -%}
 
 
@@ -79,7 +83,11 @@
 {% endmacro %}
 
 {%- macro default__get_revoke_sql(relation, privilege, grantees) -%}
-    revoke {{ privilege }} on {{ relation }} from {{ grantees | join(', ') }}
+    revoke {{ privilege }} on {{ relation }} from
+    {% for grantee in grantees %}
+    {{ adapter.quote(grantee) }}
+    {% if not loop.last %},{% endif %}
+    {% endfor %}
 {%- endmacro -%}
 
 


barton996:
@mikealfare @dbeatty10

mikealfare (dismissed):

graciegoheen:
One thing to callout here - I believe this will make usernames case-sensitive, which is technically not backwards compatible.

dbeatty10:
> One thing to callout here - I believe this will make usernames case-sensitive, which is technically not backwards compatible.

Good callout @graciegoheen. I think we should finish refining the behavior within https://github.com/dbt-labs/dbt-core/issues/8751 before merging this.

Specifically, https://github.com/dbt-labs/dbt-core/issues/8751#issuecomment-1747199588 mentions the following:

> One thing for us to consider is if we can/should add a `quote_grants` config for grants (or something else similar to [`quote_columns`](https://docs.getdbt.com/reference/resource-configs/quote_columns)). If we add some kind of quoting-related config, hopefully it would fit together with https://github.com/dbt-labs/dbt-core/issues/2986.

dbeatty10 (changes requested):
@barton996 and @graciegoheen adding a blocking review here to prevent an accidental merge until we decide if we can/should make this backwards-compatible or not. See discussion [here](https://github.com/dbt-labs/dbt-core/pull/8809#issuecomment-1756305771).

dbeatty10:
Thanks for taking the time to open this PR @barton996! Since opening, we've [decoupled dbt Adapters from dbt Core](https://github.com/dbt-labs/dbt-core/discussions/9171), and this code now lives in a separate repo: [dbt-adapters](https://github.com/dbt-labs/dbt-adapters).

PR #8817, opened 2023-10-11
Partial parsing issue when adding groups and updating models at the same time

### LINKED ISSUE #8697: [CT-3145] [Bug] partial parsing issue when adding private models to a new group

### Current Behavior

[loom explanation](https://www.loom.com/share/9c61cf7d28ff41348ca7a4bba98ac6f1)

adding multiple nodes that have a relationships to the same group with private access results in a parsing error:

```
Parsing Error
  Node model.mega_corp_big_co_inc.int_returns__unioned attempted to reference node model.mega_corp_big_co_inc.stg_tpcds_core__catalog_returns, which is not allowed because the referenced node is private to the 'sales_analytics' group.
```

in this case, `int_returns__unioned` is a child of `stg_tpcds_core__catalog_returns`, both were added to a newly-created group called `sales_analytics` and had access set to `private`, all of which happened between invocations of dbt. 

### Expected Behavior

everything should parse just fine! these are both private members of the same group and we love that

### Steps To Reproduce

see loom!

### Relevant log output

_No response_

### Environment

```markdown
- OS: mac
- Python: Python 3.9.16
- dbt: 1.6.0 and 1.6.3 (see loom)
```

snowflake

### Additional Context

i love dbt

gshank:
resolves #8697

### Problem

When processing changed models the partial parsing code was removing referencing nodes and updating pp_dict with the original model yaml dictionary, but the dictionary had changed so the patch information was lost when doing "merge_patch".

### Solution

The code was called in the order handle_schema_file_changes => delete_schema_mssa_links => schedule_referencing_nodes_for_parsing => schedule_nodes_for_parsing => remove_mssat_file => remove_node_in_saved => merge_patch.

The merge_patch call was saving the original patch from the saved schema file, so when the changed schema file patch was encountered it was skipped in merge_patch. An additional parameter was added to "merge_patch" so that if the merge_patch call comes from a schema file change, it will update the patch to replace the old saved patch.

### Checklist

### DIFF
--- a/.changes/unreleased/Fixes-20231010-202148.yaml
+++ b/.changes/unreleased/Fixes-20231010-202148.yaml
@@ -1,0 +1,6 @@
+kind: Fixes
+body: Partial parsing fix for adding groups and updating models at the same time
+time: 2023-10-10T20:21:48.154666-04:00
+custom:
+  Author: gshank
+  Issue: "8697"
--- a/core/dbt/parser/partial.py
+++ b/core/dbt/parser/partial.py
@@ -338,7 +338,10 @@ class PartialParsing:
             file_id = node.patch_path
             # it might be changed...  then what?
             if file_id not in self.file_diff["deleted"] and file_id in self.saved_files:
-                # schema_files should already be updated
+                # Schema files should already be updated if this comes from a node,
+                # but this code is also called when updating groups and exposures.
+                # This might save the old schema file element, so when the schema file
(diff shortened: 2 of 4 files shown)


emmyoop (approved):
Looks good.  A but more meat on the last test would be useful though.

martynydbt:
@gshank The patches went out so you can backport these now for the next patch. thanks!

gshank:
Removing the backport 1.5.latest label because this shouldn't happen in 1.5.

Issue #4637, opened 2022-01-27
[CT-103] Add support for sha1 to package dependancies


emmyoop:
### Is there an existing feature request for this?

- [X] I have searched the existing issues

### Describe the Feature

Add an optional `sha1` field to package definitions.  If it exists, verify the downloaded package matches.

More details to come.

### Describe alternatives you've considered

Not doing it.  

### Who will this benefit?

Anyone using packages.  Will allow more informative error messages when things go sideways.

### Anything else?

This will help catch corrupt files as in #4579.  #4205 is also proposing adding this functionality with the new download type.

jtcohen6:
For Hub-hosted packages, the Hub API includes/returns both:
- the URL of the tarball to be downloaded
- the `sha1` expected for that download

[Example](https://github.com/dbt-labs/hub.getdbt.com/blob/222ca090785388a57400bb6d62b5aa29d6abb3d5/data/packages/dbt-labs/dbt_utils/versions/0.8.0.json#L16):
```json
    "downloads": {
        "tarball": "https://codeload.github.com/dbt-labs/dbt-utils/tar.gz/0.8.0",
        "format": "tgz",
        "sha1": "9fb457bcec7b19a9e51241742cde64a01eb2d77e"
    }
```

Could we add that check somewhere [here](https://github.com/dbt-labs/dbt-core/blob/f3735187a620e3a40ee01b3fa04be191de2b5901/core/dbt/deps/registry.py#L61-L74), when we install registry packages? Then we'd benefit from the check, without users having to specify it in `packages.yml`

PR #8855, opened 2023-10-13
import nested tarball package dependencies

### LINKED ISSUE #8621: [CT-3107] [Bug] nested dependencies not installed when package is a tarball

### Current Behavior

when running `dbt deps` to install a package specified as a tarball, dbt doesn't install nested dependencies (i.e. packages specified in the imported package's `packages.yml` file) as it does when installing a package from local, git or the dbt hub.

### Expected Behavior

consistent behaviour across import methods regarding nested dependencies. dbt should install any dependencies specified in the tarball project's packages.yml file.

### Steps To Reproduce

this can be reproduced by importing the tarball of a package with nested dependencies. In this case, importing dbt_expectations should cause dbt_date to be installed, as its included in the package's dependencies here: https://github.com/calogica/dbt-expectations/blob/0.9.0/packages.yml

Steps:
1. create a `packages.yml` file in a project with the following structure:
``` yaml
packages:
  - tarball: "https://github.com/calogica/dbt-expectations/archive/refs/tags/0.9.0.tar.gz"
    name: "dbt_expectations"
```
2. run `dbt deps`

running dbt deps will only install dbt_expectations:
```
20:08:55  Running with dbt=1.5.6
20:08:55  Installing dbt_expectations
20:08:56  Installed from tarball (url: https://github.com/calogica/dbt-expectations/archive/refs/tags/0.9.0.tar.gz)
```
compare this to installing the same package from dbt hub, with the following `packages.yml`:
``` yaml
packages:
  - package: calogica/dbt_expectations
    version: "0.9.0"
```
```
20:14:24  Running with dbt=1.5.6
20:14:24  Installing calogica/dbt_expectations
20:14:25  Installed from version 0.9.0
20:14:25  Up to date!
20:14:25  Installing calogica/dbt_date
20:14:25  Installed from version 0.8.1
20:14:25  Updated version available: 0.9.1
20:14:25  
20:14:25  Updates available for packages: ['calogica/dbt_date']                 
Update your versions in packages.yml, then run dbt deps
```

### Relevant log output

_No response_

### Environment

```markdown
- OS: Mac OS 13.5.2 (22G91)
- Python: 3.9
- dbt: 1.5.6
```

snowflake

### Additional Context

_No response_

adamlopez:
resolves #8621

### Problem

when running dbt deps to install a package specified as a tarball, dbt doesn't install nested dependencies (i.e. packages specified in the imported package's packages.yml file) as it does when installing a package from local, git or the dbt hub.

### Solution

This PR solves the problem above by adapting the git dependency installation method, which downloads the dbt project into the DOWLOADS_PATH directory and pulls the metadata from the project once downloaded. In my opinion this is preferable to the current implementation of TarballPackage's `_fetch_metadata` method which recycles RegistryPackageMetadata's `download_and_untar` functionality after fetching the metadata. The downside of thte current approach is that metadata can't be pulled from the project since it hasn't yet been untarred at the time `_fetch_metadata` is called.

### Checklist


adamlopez:
changie new

adamlopez:
Hi @MichelleArk! What's the expected timeline for reviewing this change? let me know if you have any questions/concerns with this fix

jtcohen6:
@adamlopez I'm going to mark this PR `ready_for_review` and pull it into our triage/review queue!

core/dbt/deps/tarball.py
```diff
@@ -42,23 +48,31 @@ def nice_version_name(self):
...
 
     def install(self, project, renderer):
-        self._install(project, renderer)
+        dest_path = self.get_installation_path(project, renderer)
+        if os.path.exists(dest_path):
+            if system.path_is_symlink(dest_path):
+                system.remove_file(dest_path)
+            else:
+                system.rmdir(dest_path)
+        system.move(self.untarred_path, dest_path)
```
jtcohen6, line 75: This assumes that `_fetch_metadata` has always already been run. That might not be the case with the changes to `dbt deps` in v1.7 — the introduction of `package-lock.json` — which separates these two steps:
1. Resolving package metadata (to generate/update the lock file)
2. Installing package contents (from the lock file specification)

core/dbt/deps/tarball.py
```diff
@@ -42,23 +48,31 @@ def nice_version_name(self):
...
-        dct = {
-            "name": self.package,
-            "packages": [],  # note: required by RegistryPackageMetadata
-            "downloads": {"tarball": self.tarball},
-        }
-
-        return RegistryPackageMetadata.from_dict(dct)
+        """Download and untar the project and parse metadata from the project folder."""
+        system.download(self.tarball, self.tar_path)
+        system.untar_package(self.tar_path,self.untarred_path)
```
jtcohen6, line 53: We should use the same `download_and_untar` + `connection_exception_retry` methods here. That way, if we need to add more things ([like `sha1` comparison for tarball contents](https://github.com/dbt-labs/dbt-core/issues/4637)), we can do it in one place.

IMO it also makes more sense for this method to include the check of the downloads directory path, to see whether the package has already been downloaded there. (The [default behavior](https://github.com/dbt-labs/dbt-core/blob/e24f9b3da722328e24ee10a06fa2e8d2ca73b3e6/core/dbt/deps/base.py#L31-L35) is to clear the downloads directory after every `dbt deps` run.)

core/dbt/deps/tarball.py
```diff
@@ -42,23 +48,31 @@ def nice_version_name(self):
...
-        """
-        recycle RegistryPackageMetadata so that we can use the install and
-        download_and_untar from RegistryPinnedPackage next.
-        build RegistryPackageMetadata from info passed via packages.yml since no
-        'metadata' service exists in this case.
-        """
-
-        dct = {
-            "name": self.package,
-            "packages": [],  # note: required by RegistryPackageMetadata
```
jtcohen6, line 54: An alternative approach, more in keeping with the previous implementation: Allow users to manually specify the dependent packages that need to be installed.
```yml
# packages.yml
packages:
  - tarball: "https://github.com/calogica/dbt-expectations/archive/refs/tags/0.9.0.tar.gz"
    name: "dbt_expectations"
    # manually specified
    packages:
      - package: calogica/dbt_date
        version: [">=0.9.0", "<1.0.0"]
```
But that doesn't feel any better than just manually specifying both packages:
```yml
# packages.yml
packages:
  - tarball: "https://github.com/calogica/dbt-expectations/archive/refs/tags/0.9.0.tar.gz"
    name: "dbt_expectations"
  - package: calogica/dbt_date
    version: [">=0.9.0", "<1.0.0"]
```

core/dbt/deps/tarball.py
```diff
@@ -42,23 +48,31 @@ def nice_version_name(self):
...
     def _fetch_metadata(self, project, renderer):
-        """
-        recycle RegistryPackageMetadata so that we can use the install and
-        download_and_untar from RegistryPinnedPackage next.
-        build RegistryPackageMetadata from info passed via packages.yml since no
-        'metadata' service exists in this case.
-        """
-
-        dct = {
-            "name": self.package,
```
jtcohen6, line 53: In theory, if we're downloading the package to read its metadata, it would no longer be required for users to provide the package `name`

jtcohen6 (commented):
_summarizing from live review with @emmyoop_

`dbt deps` has two steps:
1. **Resolution:** Resolve package dependencies to understand which packages need to be installed
2. **Installation:** Actually install the contents of those packages.

Historically, `dbt deps` always did both things every time it ran. dbt-core v1.7 introduced a separation between those steps, with `package-lock.json` as the saved state in between them.

The behavior here varies by installation method:
- Hub packages can perform step (1) just by calling an API.
- `git` and `local` packages need to actually download/read the contents of the package to determine which other packages it depends on.
- `tarball` packages do not perform any dependency resolution. We decided this was [out of scope for the initial implementation](https://github.com/dbt-labs/dbt-core/issues/4205#issuecomment-962611921); you must manually specify their `name` along with any dependencies. The goal of this PR is to make this possible.

I think the implementation here could be quite similar to the implementation of `git` packages: both the `_fetch_metadata` and `install` methods call another method that actually copies down the files. This is `_checkout` for git packages, but it could be 
`download_and_untar` for tarball packages:

https://github.com/dbt-labs/dbt-core/blob/e24f9b3da722328e24ee10a06fa2e8d2ca73b3e6/core/dbt/deps/git.py#L75-L112

I think that will serve us better than assuming `_fetch_metadata` and `install` are always both run, one before the other.

adamlopez (commented):
Hi @jtcohen6 thanks for the feedback! I've updated the PR to use `download_and_untar`, and call it in both `_fetch_metadata` and `install`

core/dbt/deps/tarball.py
jtcohen6: Legit! Okay, let's call this out of scope for the present PR. @emmyoop made a really good point that requiring the user to type the `name` in `packages.yml` is probably better for readability anyway.

.changes/unreleased/Fixes-20231013-130943.yaml
```diff
@@ -0,0 +1,6 @@
+kind: Fixes
+body: resolve nested dependencies for packages installed using tarball method
```
jtcohen6, line 2: Recommendation for clarity (make it easier for people to see right away if this change is relevant to them):
```suggestion
For packages installed with tarball method, fetch metadata to resolve nested dependencies
```

jtcohen6 (approved):
Thank you for the contribution @adamlopez! :)

This looks good from a functional perspective. I tested out locally, and it's working well.

```yml
# packages.yml
packages:
  - tarball: "https://github.com/calogica/dbt-expectations/archive/refs/tags/0.9.0.tar.gz"
    name: "dbt_expectations"
```
```yml
package-lock.yml
packages:
- name: dbt_expectations
  tarball: https://github.com/calogica/dbt-expectations/archive/refs/tags/0.9.0.tar.gz
- package: calogica/dbt_date
  version: 0.8.1
sha1_hash: 234a682225e1bdd4d81d10fc0d902b015b2dc26c
```
```
$ dbt deps --upgrade
16:45:34  Running with dbt=1.8.0-a1
16:45:36  Updating lock file in file path: /.../package-lock.yml
16:45:36  Installing dbt_expectations
16:45:37  Installed from tarball (url: https://github.com/calogica/dbt-expectations/archive/refs/tags/0.9.0.tar.gz)
16:45:37  Installing calogica/dbt_date
16:45:37  Installed from version 0.8.1
16:45:37  Updated version available: 0.10.0
16:45:37
16:45:37  Updates available for packages: ['calogica/dbt_date']
Update your versions in packages.yml, then run dbt deps
$ dbt deps
16:45:41  Running with dbt=1.8.0-a1
16:45:41  Installing dbt_expectations
16:45:42  Installed from tarball (url: https://github.com/calogica/dbt-expectations/archive/refs/tags/0.9.0.tar.gz)
16:45:42  Installing calogica/dbt_date
16:45:42  Installed from version 0.8.1
16:45:42  Updated version available: 0.10.0
16:45:42
16:45:42  Updates available for packages: ['calogica/dbt_date']
Update your versions in packages.yml, then run dbt deps
```

emmyoop (approved):
@adamlopez this looks great.  Thanks for wading through the deps code to make the experience using tarballs better for everyone!

adamlopez:
Thanks @emmyoop @jtcohen6! Let me know if anything else is needed on my end for this to be merged 🙂

.changes/unreleased/Fixes-20231013-130943.yaml
```diff
@@ -0,0 +1,6 @@
+kind: Fixes
+For packages installed with tarball method, fetch metadata to resolve nested dependencies
```
emmyoop, line 2: on us - accidentally deleted the key!
```suggestion
body: For packages installed with tarball method, fetch metadata to resolve nested dependencies
```

adamlopez:
Hey @emmyoop @jtcohen6, when can we expect to see this change included in a release? Just wondering if there's anything I need to do to get this included in the next minor release, and couldn't see anything in the contributing docs. Thanks 🙂

emmyoop:
@adamlopez this will be included in our next minor release, 1.8.  You can read more about the timing of it in our roadmap post [dbt: The Next Generation (November 2023)](https://github.com/dbt-labs/dbt-core/blob/main/docs/roadmap/2023-11-dbt-tng.md)!

CodeGeek1212:
Is it also possible to use a tar file which does not end on .tar.gz? 

For example, we want to save the tar file on a databricks workspace and refer it in the packages.yml.
But if we upload it to the workspace the url is handled by Databricks.

packages:
  - tarball: https://[worskace[].azuredatabricks.net/?0=121212121212122#files=122112112
    name: 'dbt_utils'

When exuting this I get: 
External connection exception occurred: not a gzip file

PR #8856, opened 2023-10-13
Fix issues around new get_catalog_by_relations macro

### LINKED ISSUE #8846: [CT-3216] [Regression] Calls to `get_catalog` are not all wrapped in the feature flag

### Current Behavior

Calls to the new `get_catalog` method fail due to providing too many arguments (the instance plus two arguments). This line causes the error:

https://github.com/dbt-labs/dbt-core/blob/140b655f0d7663468066cfcb2c77e9b8ca9a3d38/core/dbt/task/generate.py#L248

Calls to `get_catalog` work so that `docs generate` runs properly.

### Steps To Reproduce

Try to run `docs generate` to trigger a call to `get_catalog`.

### Relevant log output

```shell
https://github.com/databricks/dbt-databricks/actions/runs/6499031208/job/17651537811
```

### Environment

```markdown
- OS:
- Python:
- dbt (working version): dbt-core/dbt-databricks 1.7.0rc1
- dbt (regression version): dbt-core/dbt-databricks 1.6
```

### Additional Context

dbt-databricks 1.7.0rc1

peterallenwebb:
resolves #8846

### Problem

An optional parameter was added to BaseAdapter.get_catalog() in 1.7.0rc1, which was unfortunately overridden by databricks, and may also be overridden in other adapter implementations.

### Solution

This changeset reverts the function to its original signature, and adds a new function called BaseAdapter.get_filtered_catalog(). It also reworks the `docs generate` code which had relied previously relied on the new parameter. 

Significantly, the filtering is now performed against a set of BaseRelation objects passed to the adapter, instead of a set of nodes. This reduces the number of concepts that leak into the adapter, which should not need to "know" about the dag or nodes.

### Checklist

### DIFF
--- a/.changes/unreleased/Fixes-20231024-155400.yaml
+++ b/.changes/unreleased/Fixes-20231024-155400.yaml
@@ -1,0 +1,6 @@
+kind: Fixes
+body: Rework get_catalog implementation to retain previous adapter interface semantics
+time: 2023-10-24T15:54:00.628086-04:00
+custom:
+  Author: peterallenwebb
+  Issue: "8846"
--- a/core/dbt/adapters/base/impl.py
+++ b/core/dbt/adapters/base/impl.py
@@ -455,30 +455,16 @@ class BaseAdapter(metaclass=AdapterMeta):
 
         return relations_by_info_schema
 
-    def _get_catalog_relations(
-        self, manifest: Manifest, selected_nodes: Optional[Set] = None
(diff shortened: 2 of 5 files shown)


gshank (approved):

mikealfare:
I think we might have introduced a pre-regression with this PR. See the details here: https://github.com/dbt-labs/dbt-snowflake/pull/816.

PR #8887, opened 2023-10-24
Dbeatty/8877 unknown data type code

### LINKED ISSUE #8353: [CT-2957] [Bug] Cannot specify column contract of type UUID in PostgreSQL with DBT 1.5.3

### LINKED ISSUE #8877: [CT-3243] [Bug] Unrecognized data types break `get_column_schema_from_query` macro (and contracts)

### Current Behavior

Currently the `get_column_schema_from_query` macro, which is used as part of contracts, but also is available for any users to call (although not documented) will fail when it encounters an unknown data type code. This has already been raised as a specific case for postgres `uuid` type in #8353 and a fix is in place for both this and the `ip_address` type in #8357. However this issue is more generic to support any potential non-defined type rather than specifics.

My concern is that if postgres adds any more types in the future this issue will persist and a new PR is going to need to be raised and merged in before this macro can be used or a contract can be enforced on the model. This may also impact other warehouses although I am not sure exactly how their column type mappings are implemented. Already in #8357 someone has mentioned the `money` type has the same issue.

In addition to failing, it does so in a very unhelpful way, with the error message literally just being the type code from the warehouse, if the behaviour won't change it would be nice to have a more specific error message.

### Expected Behavior

I see a few options that may be possible in the case of getting a type code that is not in the type dictionary that dbt is using.

1. When no match is returned simply return the type code itself, this would allow the macro to complete and contracts to be enforced, however the user would need to define the contract using the type code instead of the type name. The dbt docs could list all supported (postgres?) types and detail what to do for non-supported types. This could lead to issues later if those types are added later and the user upgrades. It would be more work, but if contract checks supported both type codes and type names (auto converting from name to code as codes would always be available) this would solve that issue, but would require changing some of the contract code I suspect?
2. Return something like "unknown", however this would go against the idea of contracts and causes the same later supported issues.
3. The macro would still fail, but the [data_type_code_to_name](https://github.com/dbt-labs/dbt-core/blob/d571dd19ae0049b80e526682bd7ca52c7d15b416/plugins/postgres/dbt/adapters/postgres/connections.py#L203) method would produce a more useful error message. This is not ideal as it will still fail if any new types are in a table, but at least it is more obvious why it is failing.

### Steps To Reproduce

1. Create a new dbt project, using a postgres connection
2. Create a new model with the following content:
    ```
    {% set uuid_col = get_column_schema_from_query( "select * from (select 'f76838a4-7411-4f04-a960-4fc77fd3107b'::uuid as test_col) a") %}
    
    select 1
    ```
3. Do a `dbt run`, get the `2950` error despite no contract being used in the package.

### Relevant log output

```shell
09:05:33  Running with dbt=1.6.3
09:05:33  Registered adapter: postgres=1.6.3
09:05:34  Found 1 model, 0 sources, 0 exposures, 0 metrics, 349 macros, 0 groups, 0 semantic models
09:05:34  
09:05:34  Concurrency: 1 threads (target='postgres')
09:05:34  
09:05:34  1 of 1 START sql table model dbt_ryan.uuid_read_test ........................... [RUN]
09:05:34  Unhandled error while executing 
2950
09:05:34  1 of 1 ERROR creating sql table model dbt_ryan.uuid_read_test .................. [ERROR in 0.12s]
09:05:34  
09:05:34  Finished running 1 table model in 0 hours 0 minutes and 0.39 seconds (0.39s).
09:05:34  
09:05:34  Completed with 1 error and 0 warnings:
09:05:34  
09:05:34    2950
09:05:34  
09:05:34  Done. PASS=0 WARN=0 ERROR=1 SKIP=0 TOTAL=1
```

### Environment

```markdown
- OS: MacOS
- Python: 3.8.13
- dbt: 1.6.3
```

postgres

### Additional Context

I think this was somewhat discussed in the original PR https://github.com/dbt-labs/dbt-core/pull/6986 but it was decided to progress anyway.

For context the reason we use this macro in our packages explicitly is to get all the columns in a table so that we can rename some, but without having to manually list out all names (because postgres doesn't support `exclude` syntax, and we don't necessarily know all the columns in the table at this point). https://github.com/snowplow/dbt-snowplow-web/blob/e4ae6b297728abfb2dce5673d540e556b21f68df/models/base/scratch/default/snowplow_web_base_events_this_run.sql#L62

dbeatty10:
resolves #8353
resolves #8877

### Problem

Postgres has built-in data types like `money`, `uuid`, etc that aren't recognized by default. In addition, Postgres has the ability to create user-defined data types. In both cases, the `psycopg2` Python driver won't know how to do a reverse lookup from the `type_code` into a human-readable data type string (like `money` or `uuid`).

Other database platforms may have similar situations.

### Solution

The proposed solution is to check if a lookup is available. If not, then provide a stand-in of the form `unknown data_type 4678`. This stand-in only applies if a data contract is broken. If the data contract is valid, then this text label isn't necessary. 

### TODO
- [ ] Put screenshots of error messages above
- [ ] Add applicable test cases

### Checklist

- [ ] 👈 This PR includes tests
- [x] This PR has no interface changes (e.g. macros, cli, logs, json artifacts, config files, adapter interface, etc)
- [x] This PR does **not** include [type annotations](https://docs.python.org/3/library/typing.html) because there are no new or modified functions

### DIFF
--- a/plugins/postgres/dbt/adapters/postgres/connections.py
+++ b/plugins/postgres/dbt/adapters/postgres/connections.py
@@ -204,4 +204,7 @@ class PostgresConnectionManager(SQLConnectionManager):
 
     @classmethod
     def data_type_code_to_name(cls, type_code: int) -> str:
-        return string_types[type_code].name
+        if type_code in string_types:
+            return string_types[type_code].name
+        else:
+            return f"unknown type_code {type_code}"
--- a/.changes/unreleased/Fixes-20231024-110151.yaml
+++ b/.changes/unreleased/Fixes-20231024-110151.yaml
@@ -1,0 +1,6 @@
+kind: Fixes
+body: Handle unknown `type_code` for model contracts
+time: 2023-10-24T11:01:51.980781-06:00
+custom:
+  Author: dbeatty10
+  Issue: "8877" "8353"


rlh1994:
Would it be useful to add a debug level warning when it fails to find a type? In particular if someone is using the `column.Data_type` from a `get_column_schema_from_query` call it might not be obvious why something is failing without directly outputting all the column data types?

plugins/postgres/dbt/adapters/postgres/connections.py
```diff
@@ -204,4 +204,7 @@ def get_response(cls, cursor) -> AdapterResponse:
 
     @classmethod
     def data_type_code_to_name(cls, type_code: int) -> str:
-        return string_types[type_code].name
+        if type_code in string_types:
+            return string_types[type_code].name
+        else:
+            return f"unknown type_code {type_code}"
```
dbeatty10, line 210: @rlh1994 [brought up](https://github.com/dbt-labs/dbt-core/pull/8887#issuecomment-1777754121) the idea of raising a debug-level warning in this case.

Seems like a valuable idea to explore. Maybe something like this?

```python
            logger.debug("unknown type_code: {}", type_code)
```

rlh1994:
Just for complete visibility on this PR, I have confirmed that this works for user created types in postgres as well.

dbeatty10:
See below for a prototype of converting an unrecognized `type_code` into the desired `data_type`.

1\. Create user-space macros to do lookups:

`macros/type_code.sql`

```sql

{# Define your lookup table here #}
{% macro get_type_code_lookup_table() %}
  {% set lookup_table = {
      '790': 'MONEY',
      '2950': 'UUID',
  } %}
  {{ return(lookup_table) }}
{% endmacro %}


{# Look up an unrecognized `type_code` for a Column #}
{% macro lookup_type_code(data_type) %}

    {% set lookup_table = get_type_code_lookup_table() %}
    
    {# Attempt to match the pattern in the provided data_type string. #}
    {% set pattern = 'unknown type_code ' %}
    {% if data_type.startswith(pattern) %}
        {# Extract the type code and trim any leading/trailing whitespace. #}
        {% set type_code = data_type.replace(pattern, '').strip() %}
        
        {# Look up the extracted type code in the dictionary. #}
        {% if type_code in lookup_table %}
            {% set data_type = lookup_table[type_code] %}
        {% endif %}
    {% endif %}

  {{ return(data_type) }}

{% endmacro %}
```

2\. Define an analysis to try it out:

`analyses/get_column_schema_from_query.sql`
```sql
{#- Get the column names and data types -#}
{%- set cols = get_column_schema_from_query(sql) -%}

SQL:

  {{ sql | indent(2) }}

Columns:
{% for col in cols %}
  {{ col }}
{%- endfor %}

User-provided `type_code` translations:
{% for col in cols %}
  {{ col.name }} ({{ lookup_type_code(col.data_type) }})
{%- endfor %}
```

3\. Compile the analysis:
```shell
dbt compile
```

4\. Examine the output:

`target/compiled/my_project/analyses/get_column_schema_from_query.sql`
```
SQL:

  select
    '12.34'::numeric as numeric_col,
    '12.34'::money as money_col,
    'f76838a4-7411-4f04-a960-4fc77fd3107b'::text as text_col,
    'f76838a4-7411-4f04-a960-4fc77fd3107b'::uuid as uuid_col

Columns:

  <Column numeric_col (DECIMAL)>
  <Column money_col (unknown type_code 790)>
  <Column text_col (TEXT)>
  <Column uuid_col (unknown type_code 2950)>

User-provided `type_code` translations:

  numeric_col (DECIMAL)
  money_col (MONEY)
  text_col (TEXT)
  uuid_col (UUID)
```

MichelleArk (approved):

plugins/postgres/dbt/adapters/postgres/connections.py
dbeatty10: @MichelleArk how would you feel about adding this ☝️ debug output when the `type_code` isn't found?

MichelleArk: It may be a little redundant given `unknown type_code` will appear in the error message either way. So I'd either omit it or write a more informative message like"type_code not found in adapter-provided lookup, defaulting to 'unknown type_code'"

And we may actually need to use fire_event to fire a debuglevel event as opposed not logger.debug

dbeatty10: 👍 Splitting this into its own issue:
- https://github.com/dbt-labs/dbt-core/issues/8912 

That way we can ship this fix while still giving that debug message enough attention.

clementdavies56:
I'm interested in the release status of this feature on PyPI. I haven't found any details about its release in the recent tags or notes. Could you please update me on its current status?

Issue #8422, opened 2023-08-16
[CT-2997] Ability to test incremental models


gshank:
### Description

We should allow folks to add unit tests to incremental models

### Acceptance criteria

* Ability to test incremental models in both full refresh and incremental mode
* Ability to specify the current status of an incremental model to test the incremental behavior

More on the above two points in https://github.com/dbt-labs/dbt-core/discussions/8275#discussioncomment-6825650: 
> To support this, dbt's unit testing framework would need to provide the ability to:
* override the is_incremental macro to return True (p0), and provide the ability to specify an input fixture for {{ this }} (p1) in order to mock out the previous state of an incremental model. The expect output would be the set of rows that dbt's incremental materialization would insert/merge into the existing relation in production, as opposed to the resulting updated representation of the model.

To be moved into a separate issue: 
* Ability to test the incremental behavior of models using incremental strategy that requires partitioning (e.g. insert overwrite in BigQuery): https://github.com/dbt-labs/dbt-core/issues/8664 (blocked on spike #8499)

A "current_actual" could define the current state of the model we're incrementing on, testing whether the incrementing works as expected.

martynydbt:
@gshank please pull Ability to test the incremental behavior of models using incremental strategy that requires partitioning (e.g. insert overwrite in BigQuery) out into a separate ticket

jtcohen6:
2x2 = 4 acceptance tests for us to write as part of implementation:
- {{ is_incremental() }} : True + False
- {{ this }} fixture: provided or not

MichelleArk:
Here's how i'm imagining the spec for `this`:
```
unit-tests: 
 - model: my_model
   tests: 
    - name: test_name
      given: 
      - input: this
         rows: ... 
```

PR #8891, opened 2023-10-25
Support unit testing incremental models

MichelleArk:
resolves https://github.com/dbt-labs/dbt-core/issues/8422

### Problem

Unit testing incremental models requires support for both: 
1. overriding the `is_incremental` macro to return True or False (depending on which scenario is being tested) and provide  (already done)
2. the ability to specify an input fixture for `{{ this }}` in order to mock out the previous state of the tested incremental model (this PR!)

### Solution
* accept `this` as a `given` input yml name, in addition to `ref` and `source` calls
* storing the model node of on a `UnitTestNode` when `this` is provided for reference downstream in `UnitTestContext.this` 
* overriding `this` in `UnitTestContext` to add an ephemeral prefix to the unit test sql if `this` is specified

### Checklist

### DIFF
--- a/core/dbt/context/providers.py
+++ b/core/dbt/context/providers.py
@@ -1543,10 +1543,10 @@ class UnitTestContext(ModelContext):
 
     @contextproperty()
     def this(self) -> Optional[str]:
-        if self.model.this:
-            # TODO: RuntimeRefResolver.set_cte also passes None as second argument.
-            self.model.set_cte(self.model.this.unique_id, None)  # type: ignore
-            return self.adapter.Relation.add_ephemeral_prefix(self.model.this.name)
+        if self.model.this_model_fixture_unique_id in self.manifest.nodes:
+            this_node = self.manifest.expect(self.model.this_model_fixture_unique_id)
+            self.model.set_cte(self.model.this_model_fixture_unique_id, None)  # type: ignore
+            return self.adapter.Relation.add_ephemeral_prefix(this_node.name)
         return None
 
 
--- a/core/dbt/context/providers.py
+++ b/core/dbt/context/providers.py
@@ -1543,9 +1543,9 @@ class UnitTestContext(ModelContext):
 
     @contextproperty()
     def this(self) -> Optional[str]:
-        if self.model.this_model_fixture_unique_id in self.manifest.nodes:
-            this_node = self.manifest.expect(self.model.this_model_fixture_unique_id)
-            self.model.set_cte(self.model.this_model_fixture_unique_id, None)  # type: ignore
+        if self.model.this_input_node_unique_id:
+            this_node = self.manifest.expect(self.model.this_input_node_unique_id)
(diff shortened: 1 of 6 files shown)


core/dbt/context/providers.py
```diff
@@ -1541,6 +1541,14 @@ def env_var(self, var: str, default: Optional[str] = None) -> str:
         else:
             return super().env_var(var, default)
 
+    @contextproperty()
+    def this(self) -> Optional[str]:
```
gshank, line 1545: Elsewhere "this" is a Relation. Here it's just a model name plus the ephemeral_prefix. Might be confusing. Maybe call it something else?

core/dbt/contracts/graph/nodes.py
```diff
@@ -1068,6 +1068,7 @@ class UnitTestNode(CompiledNode):
     attached_node: Optional[str] = None
     overrides: Optional[UnitTestOverrides] = None
     config: UnitTestNodeConfig = field(default_factory=UnitTestNodeConfig)
+    this: Optional[ModelNode] = None
```
gshank, line 1071: I'm wondering if using "this" in this way might not be confusing. I think "this" in the rest of the context is actually a Relation, right?  What do you think of calling it "original_input_node"?

gshank (approved):
Overall I think this is very clever. I'm not convinced that using "this" as the name in several places is the best choice though. Not only do the two places differ from each other, they're not the same as the original "this"...

core/dbt/context/providers.py
MichelleArk: I don't think we can we can call it anything else in the provider specifically, because need the `{{ this }}` call in the jinja context to resolve to the CTE name of the fixture we create from `this` provided in the unit test yaml.

core/dbt/contracts/graph/nodes.py
MichelleArk: `attached_node` above (should maybe be renamed to `tested_node_unique_id`(I can go ahead and rename that here) already represents that but this (no pun intended) is actually different.

MichelleArk: I'm going to test something out to see if they can be used interchangeably though..

MichelleArk: I've updated this to explicitly include `this_input_node_unique_id` and updated `attached_node` to be `tested_node_unique_id`: https://github.com/dbt-labs/dbt-core/pull/8891/files#diff-42cfc99b2fcd89ab75957b8999bee2b1af822f7d39953ca99f1f285c170557c1R1068-R1069

the provider.this implementation needs to do a lookup now but the benefit is that we don't store the whole node on the UnitTestNode which may want to make its way to a serialized artifact

core/dbt/context/providers.py
gshank: Ah. Okay :)

gshank (approved):
Looks good!

Issue #8925, opened 2023-10-26
Remove usage of dbt.include.global_project in dbt/adapters


MichelleArk:

MichelleArk:
Closed by: https://github.com/dbt-labs/dbt-core/pull/8930

PR #8930, opened 2023-10-26
move include/global_project to adapters

MichelleArk:
resolves https://github.com/dbt-labs/dbt-core/issues/8925

### Checklist

  - 🎩 ran `dbt docs generate && dbt docs serve` off this branch & off main + jaffle_shop, manually inspected no differences. 
  - Did this explicitly for python 3.8 as well to ensure oldest supported python installations till works given changes related to include paths. 


core/dbt/task/docs/index.html
MichelleArk, line 1: this was moved from `include/index.html`. it was never in `include/global_project` but was there historically to simplify [setup.py](https://github.com/dbt-labs/dbt-core/blob/f60938aab029a863206d3e076bc401acef9a76e7/core/setup.py#L26) and is no longer necessary (ty @jtcohen6 for the sleuthing)

MichelleArk:
> However, I have not verified that the macros get correctly inherited from an adapter in another repo (e.g. dbt-snowflake). Do you expect there to be an issue there? Put another way, will we need to do the same migration at the adapter level? I'm guessing that's not the case given that dbt-postgres was not updated. But I want to call it out explicitly.

I don't think this should be an issue for adapters defined in other repos since they are installed in the same manner as dbt-postgres. Unfortunately it'd take a lot of effort to test this incrementally because the other adapters do not have corresponding migration branches yet.

mikealfare (approved):

PR #8957, opened 2023-10-31
Debug log when `type_code` fails to convert to a `data_type`

### LINKED ISSUE #8912: Debug log when `type_code` fails to convert to a `data_type`
Would it be useful to add a debug level warning when it fails to find a type? In particular if someone is using the `column.Data_type` from a `get_column_schema_from_query` call it might not be obvious why something is failing without directly outputting all the column data types?

_Originally posted by @rlh1994 in https://github.com/dbt-labs/dbt-core/issues/8887#issuecomment-1777754121_

dbeatty10:
resolves dbt-labs/dbt-postgres#38

### Problem

If someone is using the `Column.data_type` from a [`get_column_schema_from_query`](https://github.com/dbt-labs/dbt-core/blob/7317de23a3199fe2f9bb212406dc523f134e7bfb/core/dbt/adapters/base/impl.py#L323) call (like described in https://github.com/dbt-labs/dbt-core/issues/8877), it might not be obvious when a translation from `type_code` wasn't available.

### Solution

When a translation isn't possible, fire a debug-level logging event that is specific to this situation. Namely, a logging event like this:
```
[0m15:55:59.634717 [debug] [Thread-1  ]: The `type_code` 790 was not recognized, which may affect error messages for enforced contracts that fail as well as `Column.data_type` values returned by `get_column_schema_from_query`
```

### Checklist

- [x] This PR includes tests
- [ ] This PR has already received some feedback from Product or DX, but still needs final approval 
- [x] ~This PR includes [type annotations](https://docs.python.org/3/library/typing.html) for new and modified functions~


dbeatty10:
Closing in favor of these two:
- https://github.com/dbt-labs/dbt-adapters/pull/135
- https://github.com/dbt-labs/dbt-postgres/pull/39

PR #8966, opened 2023-11-01
convert to use unit test name at top level key

### LINKED ISSUE #8698: [CT-3146] [Implementation] updating the unit-tests block (to start with test name instead of model)
### Housekeeping

### Short description

From https://github.com/dbt-labs/dbt-core/issues/8606

Current the unit test spec is as follows:

```
unit:
  - model: my_model # name of the model I'm unit testing
    tests:
     - name: test_is_valid_email_address # this is the unique name of the test
       given: # optional: list of inputs to provide as fixtures
        - input: ref('users')
          rows:
           - {user_id: 1, email: cool@example.com,     email_top_level_domain: example.com}
           - {user_id: 2, email: cool@unknown.com,     email_top_level_domain: unknown.com}
           - {user_id: 3, email: badgmail.com,         email_top_level_domain: gmail.com}
           - {user_id: 4, email: missingdot@gmailcom,  email_top_level_domain: gmail.com}
         - input: ref('top_level_domains')
           rows:
            - {tld: example.com}
            - {tld: gmail.com}
       expect: # required: the expected output given the inputs above
        - {user_id: 1, is_valid_email_address: true}
        - {user_id: 2, is_valid_email_address: false}
        - {user_id: 3, is_valid_email_address: false}
        - {user_id: 4, is_valid_email_address: false}
    - name: test_another_column
      ...
```

we should adjust it so that the block is
1. named `unit-tests:` instead of `unit:`
2. starts with the test name instead of the model name

the final spec should look like this:

```
unit-tests:
  - name: test_is_valid_email_address # this is the unique name of the test
    model: my_model # name of the model I'm unit testing
    given: # optional: list of inputs to provide as fixtures
      - input: ref('users')
        rows:
         - {user_id: 1, email: cool@example.com,     email_top_level_domain: example.com}
         - {user_id: 2, email: cool@unknown.com,     email_top_level_domain: unknown.com}
         - {user_id: 3, email: badgmail.com,         email_top_level_domain: gmail.com}
         - {user_id: 4, email: missingdot@gmailcom,  email_top_level_domain: gmail.com}
      - input: ref('top_level_domains')
        rows:
         - {tld: example.com}
         - {tld: gmail.com}
    expect: # required: the expected output given the inputs above
      - {user_id: 1, is_valid_email_address: true}
      - {user_id: 2, is_valid_email_address: false}
      - {user_id: 3, is_valid_email_address: false}
      - {user_id: 4, is_valid_email_address: false}
  - name: test_another_column
    ...
```

### Acceptance criteria

- unit test block is called `unit-tests:`
- unit test block starts with test name (instead of model name)

### Impact to Other Teams

None

No

### Context

_No response_

emmyoop:
resolves #8698 
* this is only half what the above issue wants.  renaming top level key from `unit` to `unit-tests` will be a separate PR 

### Problem

Uses test name instead of model as top level key, resulting in the removal of test suites under a single key.

### Solution

- Use name as top level key
- remove logic for test suite
- also update unique key to match other node type patterns

### Checklist

### DIFF
--- a/core/dbt/parser/partial.py
+++ b/core/dbt/parser/partial.py
@@ -711,7 +711,6 @@ class PartialParsing:
     # Take a "section" of the schema file yaml dictionary from saved and new schema files
     # and determine which parts have changed
     def get_diff_for(self, key, saved_yaml_dict, new_yaml_dict):
-        dict_name = "name"
         if key in saved_yaml_dict or key in new_yaml_dict:
             saved_elements = saved_yaml_dict[key] if key in saved_yaml_dict else []
             new_elements = new_yaml_dict[key] if key in new_yaml_dict else []
@@ -722,9 +721,9 @@ class PartialParsing:
         new_elements_by_name = {}
         # sources have two part names?
         for element in saved_elements:
-            saved_elements_by_name[element[dict_name]] = element
+            saved_elements_by_name[element["name"]] = element
         for element in new_elements:
-            new_elements_by_name[element[dict_name]] = element
+            new_elements_by_name[element["name"]] = element
 
         # now determine which elements, by name, are added, deleted or changed
         saved_element_names = set(saved_elements_by_name.keys())
@@ -754,7 +753,6 @@ class PartialParsing:
     # flag indicates that we're processing a schema file, so if a matching
     # patch has already been scheduled, replace it.
     def merge_patch(self, schema_file, key, patch, new_patch=False):
-        elem_name = "name"
         if schema_file.pp_dict is None:
             schema_file.pp_dict = {}
         pp_dict = schema_file.pp_dict
@@ -764,7 +762,7 @@ class PartialParsing:
             # check that this patch hasn't already been saved
             found_elem = None
             for elem in pp_dict[key]:
-                if elem["name"] == patch[elem_name]:
+                if elem["name"] == patch["name"]:
                     found_elem = elem
             if not found_elem:
                 pp_dict[key].append(patch)
@@ -773,7 +771,7 @@ class PartialParsing:
                 pp_dict[key].remove(found_elem)
                 pp_dict[key].append(patch)
 
-        schema_file.delete_from_env_vars(key, patch[elem_name])
+        schema_file.delete_from_env_vars(key, patch["name"])
         self.add_to_pp_files(schema_file)
 
     # For model, seed, snapshot, analysis schema dictionary keys,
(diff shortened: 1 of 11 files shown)


core/dbt/parser/unit_tests.py
```diff
@@ -179,52 +179,51 @@ def __init__(self, schema_parser: SchemaParser, yaml: YamlBlock) -> None:
...
-                    expect=test.expect,
-                    description=test.description,
-                    overrides=test.overrides,
-                    depends_on=DependsOn(nodes=[tested_model_node.unique_id]),
-                    fqn=unit_test_fqn,
-                    config=unit_test_config,
-                )
-                self.manifest.add_unit_test(self.yaml.file, unit_test_definition)
+            unit_test_case_unique_id = (
+                f"{NodeType.Unit}.{self.project.project_name}.{unit_test.model}.{unit_test.name}"
```
emmyoop, line 187: Is there a reason we were hard coding `unit` here?  It doesn't match how we assign the unique ids for other node types.

tests/functional/unit_testing/test_unit_testing.py
```diff
@@ -146,6 +151,7 @@ def test_basic(self, project):
 
         # Partial parsing... remove test
         write_file(test_my_model_yml, project.project_root, "models", "test_my_model.yml")
+        # breakpoint()
```
emmyoop, line 154:
```suggestion
```

core/dbt/parser/unit_tests.py
MichelleArk: I don't think there's an explicit reason. What would be a more conventional way to build the unique_id here?

core/dbt/parser/partial.py
```diff
@@ -711,7 +711,6 @@ def _handle_element_change(
     # Take a "section" of the schema file yaml dictionary from saved and new schema files
     # and determine which parts have changed
     def get_diff_for(self, key, saved_yaml_dict, new_yaml_dict):
-        dict_name = "model" if key == "unit" else "name"
```
MichelleArk, line 714: ✨

MichelleArk (approved):

core/dbt/parser/unit_tests.py
emmyoop: Using `{NodeType.Unit}` instead of a hardcoded `unit` is more conventional.  I went ahead and made the change because it didn't seem intentional but wanted to check in case.

Issue #8378, opened 2023-08-12
[CT-2967] [Feature] run dbt in --sample mode


MichelleArk:
> [!NOTE]
> **UPDATE Jan 8, 2025**
> See the latest discussion [#11200 Sample Mode (for faster Development and CI 🚀)](https://github.com/dbt-labs/dbt-core/discussions/11200)

### Describe the feature

Introduce a new boolean flag, `--sample/--no-sample`, that informs whether dbt should run in 'sample' mode. 

In sample mode, dbt would resolve `ref` and `source` identifiers as `select * from <identifier> limit <sample-size>`. 

A default sample size could be used (perhaps 100?), or set in `config` at the model-level to inform the size of the sample to select when referencing the model. 

```
models:
  - name: my_model
     config: 
       sample: 
         size: 10
```

This would also enable setting a sample size using the hierarchical config syntax in dbt_project.yml:
```
--- dbt_project.yml
models: 
  <resource-path>:
    + sample: {<dictionary>}
```

It would also be interesting to explore providing column-level filters as configs, so that samples would resolve as `select * from <identifier> where <filters> limit <sample-size>`. This way, it would be possible to tune samples to ensure referential integrity within a run in sample mode.

We may also want to create an equivalent env_var, so folks could set a sample for a given environment

### Describe alternatives you've considered

It's possible to implement something like this with overrides of ref & source macros, but that makes dbt upgrades more difficult.

Also considered a simpler version of this where `--sample` is an integer argument representing the sample size in  https://github.com/dbt-labs/dbt-core/pull/8337. This doesn't leave design space for configuring filters on model/source samples however.

### Who will this benefit?

any dbt user looking for speedier (at the cost of completeness/accuracy) development & CI runs!

### Are you interested in contributing this feature?

yep!

### Anything else?

Took a first pass at this in a spike here: https://github.com/dbt-labs/dbt-core/pull/8337

jaypeedevlin:
Hey @MichelleArk, excited about this enhancement, as it intersects with some planned work that we were going to execute using ref/source overrides!

We definitely would love to see this in a really configurable form, like the `where` config you've suggested, or even additionally an `order by` (ie `order by random()`).  Essentially, the more customizable the better from our perspective!

jtcohen6:
Thanks for opening @MichelleArk!

I love how simple the `--sample` boolean flag is, with all the actual configuration defined in code.

[...]

PR #8971, opened 2023-11-01
[spike] --dry-run flag

MichelleArk:
resolves #

### Checklist

### DIFF
--- a/core/dbt/adapters/base/relation.py
+++ b/core/dbt/adapters/base/relation.py
@@ -1,6 +1,7 @@
 from collections.abc import Hashable
 from dataclasses import dataclass, field
 from typing import Optional, TypeVar, Any, Type, Dict, Iterator, Tuple, Set, Union, FrozenSet
+import uuid
 
 from dbt.contracts.graph.nodes import SourceDefinition, ManifestNode, ResultNode, ParsedNode
 from dbt.contracts.relation import (
@@ -36,6 +37,7 @@ class BaseRelation(FakeAPIObject, Hashable):
     include_policy: Policy = field(default_factory=lambda: Policy())
     quote_policy: Policy = field(default_factory=lambda: Policy())
     dbt_created: bool = False
+    sample: Optional[int] = None
 
     # register relation types that can be renamed for the purpose of replacing relations using stages and backups
     # adding a relation type here also requires defining the associated rename macro
@@ -192,7 +194,11 @@ class BaseRelation(FakeAPIObject, Hashable):
 
     def render(self) -> str:
         # if there is nothing set, this will return the empty string.
-        return ".".join(part for _, part in self._render_iterator() if part is not None)
+        rendered_parts = ".".join(part for _, part in self._render_iterator() if part is not None)
+        if self.sample and rendered_parts:
+            alias = f"_dbt_sample_{uuid.uuid4().hex.upper()[:6]}"
+            return f"(select * from {rendered_parts} limit {self.sample}) {alias}"
+        return rendered_parts
 
     def quoted(self, identifier):
         return "{quote_char}{identifier}{quote_char}".format(
--- a/core/dbt/cli/params.py
+++ b/core/dbt/cli/params.py
@@ -415,6 +415,14 @@ include_saved_query = click.option(
     hidden=True,
 )
 
+sample = click.option(
+    "--sample",
+    envvar="DBT_SAMPLE",
+    help="Limit by sample rows when resolving dbt ref and sources.",
+    type=click.INT,
+    default=None,
+)
+
 model_decls = ("-m", "--models", "--model")
 select_decls = ("-s", "--select")
 select_attrs = {
--- a/core/dbt/cli/main.py
+++ b/core/dbt/cli/main.py
@@ -342,6 +342,7 @@ def docs_serve(ctx, **kwargs):
 @p.profile
 @p.profiles_dir
 @p.project_dir
+@p.sample
 @p.select
 @p.selector
 @p.inline
@@ -599,6 +600,7 @@ def parse(ctx, **kwargs):
 @p.profile
 @p.profiles_dir
 @p.project_dir
+@p.sample
 @p.select
 @p.selector
 @p.state
--- a/core/dbt/context/providers.py
+++ b/core/dbt/context/providers.py
@@ -531,9 +531,13 @@ class RuntimeRefResolver(BaseRefResolver):
     def create_relation(self, target_model: ManifestNode) -> RelationProxy:
         if target_model.is_ephemeral_model:
             self.model.set_cte(target_model.unique_id, None)
-            return self.Relation.create_ephemeral_from_node(self.config, target_model)
+            return self.Relation.create_ephemeral_from_node(
+                self.config, target_model, sample=self.config.args.sample
+            )
         else:
-            return self.Relation.create_from(self.config, target_model)
+            return self.Relation.create_from(
+                self.config, target_model, sample=self.config.args.sample
+            )
 
     def validate(
         self,
@@ -590,7 +594,7 @@ class RuntimeSourceResolver(BaseSourceResolver):
                 target_kind="source",
                 disabled=(isinstance(target_source, Disabled)),
             )
-        return self.Relation.create_from_source(target_source)
+        return self.Relation.create_from_source(target_source, sample=self.config.args.sample)
 
 
 # metric` implementations


MichelleArk:
**Spike Report**

The crux of this issue is in resolving `ref` and `source` calls to a subquery that wraps the relation identifier to limit its read to 0 rows. The trickiness stems from wanting to do this in a way that is overridable by adapter, because each adapter may have a slightly different syntax for achieving this. There are a few options I've identified: 
1. Try to make use of the existing adapter-dispatchable `get_empty_subquery_sql` macro, which consolidates this adapter-specific syntax. However, relation rendering (`Relation.render`) currently occurs without a context, and it would be pretty circular to provide a context to a `Relation` since the ref and source resolvers themselves are part the context. So the only way I can imagine doing this would be to override `ref` and `source` macros in the `global_project`, but this adds a good amount of complexity and will be cumbersome to maintain (changes to ref/source will need to consider the override, potentially making changes across both) 
2. implement a default `empty_subquery_template` in python on the `BaseRelation`, so it's overridable across adapters (albeit in python). This has the downside of duplication across the `get_empty_subquery_sql` macro and the python template. But on the flip side, there should only be one correct way of writing this template so it doesn't need to be all the way in user-space (jinja), as long as its overwritable by adapter maintainers (in their BaseRelation subclass).

I've opted for (2) in this spike for its simplicity and given there is some precedence for doing a small amount of sql templating in the adapter python module.

core/dbt/adapters/base/relation.py
```diff
@@ -194,6 +195,15 @@ def render(self) -> str:
         # if there is nothing set, this will return the empty string.
         return ".".join(part for _, part in self._render_iterator() if part is not None)
 
+    def render_limited(self) -> str:
+        rendered = self.render()
+        if self.limit is None:
```
colin-rogers-dbt, line 200: Do we need this if we're also doing this check in `__str__`?

core/dbt/cli/params.py
```diff
@@ -90,6 +90,12 @@
     is_flag=True,
 )
 
+empty = click.option(
+    "--empty/--no-empty",
```
colin-rogers-dbt, line 94: what does `--no-empty` mean?

colin-rogers-dbt (commented):
a couple questions but generally LGTM

core/dbt/adapters/base/relation.py
MichelleArk: technically it's fine to remove, but this keeps the methods independent from one another and `render_limited` could be called from other places as well. If we remove this & make it less defensive / more context-aware, i'd just say we make it a 'private' method `_render_limited`

tests/adapter/dbt/tests/adapter/empty/test_empty.py
```diff
@@ -0,0 +1,75 @@
...
+        # run without empty - 3 expected rows in output - 1 from each input
+        run_dbt(["run"])
+        self.assert_row_count(project, "model", 3)
+
+        # run with empty - 0 expected rows in output
+        run_dbt(["run", "--empty"])
+        self.assert_row_count(project, "model", 0)
+
+
+class TestEmpty(BaseTestEmpty):
```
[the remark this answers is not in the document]
MichelleArk: What are some cases you're imaging here? I'm not sure there are any expected failure cases - --empty should work across any ref-able or source-able inputs.

core/dbt/adapters/base/relation.py
```diff
@@ -227,13 +237,11 @@ def create_ephemeral_from_node(
...
         node: ManifestNode,
+        **kwargs: Any,
     ) -> Self:
         # Note that ephemeral models are based on the name.
         identifier = cls.add_ephemeral_prefix(node.name)
-        return cls.create(
-            type=cls.CTE,
-            identifier=identifier,
-        ).quote(identifier=False)
+        return cls.create(type=cls.CTE, identifier=identifier, **kwargs).quote(identifier=False)
```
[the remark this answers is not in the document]
MichelleArk: good call! removed

MichelleArk:
> Does using a limit of sys.maxsize instead of None make the code a little cleaner?

I can see how that could clean this up but it would mean that _every_ rendered relation would be wrapped in a (select * limit max_size) - whether we are using the `--empty` flag or not (we could add some handling for this but then the code doesn't get much cleaner). I think this functionality should be limited to when `--empty` is provided to avoid changing the compiled sql under the head on existing surface area (all ref/source calls) to lower its risk.

core/dbt/cli/params.py
MichelleArk: Ah, unfortunately it's necessary because of the way that `dbt retry` reconstructs the previous command's args. It sees an `empty:False` and tries to map that to a non-existent `--no-empty` flag explicitly, causing failures.

tests/adapter/dbt/tests/adapter/empty/test_empty.py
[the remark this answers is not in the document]
MichelleArk: Hmm, yes that represents a legit failure case with `dbt run`, and should fail with a database error but I'm not sure that case is actually introduced by or related to  the --empty flag. I want to avoid re-testing all of `dbt run` and just test the new paths that `--empty` introduces at the functional level.

colin-rogers-dbt (approved):

mwstanleyft:
Adapters need to be provided the ability to override the [SQL statement here](https://github.com/dbt-labs/dbt-core/pull/8971/files#diff-8b301fa58dbc12c519f4bf6e3056605c09aea9da46b21d2ec5fd979cfd4f38adR203-R205) at minimum, because using a LIMIT clause in this way in BigQuery means that users will still be billed the full cost as if they had run their model for real and they will be very confused by this. BigQuery does provide some facility to avoid this but it would require either overriding the SQL statement here, or overriding the behaviour of `--empty` completely.

github-christophe-oudar:
I just happened to land that PR randomly looking at the latest commits and I think I've an interesting opened discussion that correlate with your intent here:
https://github.com/dbt-labs/dbt-core/discussions/8560

That feature is a great step toward easily dry run.
I've set up a similar custom approach with 2 macros in my codebase
```
{% macro limit_ci() -%}
{% if target.name == 'ci' -%}
LIMIT 0
{%- endif %}
{%- endmacro %}

{% macro limit_dev(rows = 100) -%}
{% if target.name == 'dev' -%}
LIMIT {{rows}}
{%- endif %}
{%- endmacro %}
```

Yet it's very manual (but working fairly well because developers love copy pasting existing models).

However I would like to do the following on CI:
- Run a dry run (with that empty approach)
- Run models that have unit tests

It feels like to do so properly, having `render` method that is overridable through a macro would open to a lot of custom solutions 🙌

jtcohen6:
@github-christophe-oudar That's exactly the motivation behind this `--empty` feature: to support unit testing in CI, without building any "real" data. We do see extensions for limiting/sampling data in development as well — check out this issue:
- https://github.com/dbt-labs/dbt-core/issues/8378

@mwstanleyft In my experience, while `limit 0` is not sufficient on BigQuery, `where false limit 0` avoids scanning any data and billing for bytes processed. There have been a one or two edge cases where this doesn't work (Christophe has found a few of them!), but these have tended to be bugs on BigQuery's side, and generally within more complex DML statements, not simple `select` / CTA.

github-christophe-oudar:
I finally tried this feature and found one major caveat:
I suspect, like me, a lot of dbt users are using `FROM {{ source('schema', 'table') }} my_table_alias`
Then using `--empty` will result in 
```FROM  (select * from `db`.`schema`.`table` where false limit 0) _dbt_limit_subq my_table_alias```
Which obviously won't work.

There are at least few solutions that could be explored to solve that problem:
- update source macro to support an alias parameter
- update source macro to support adding or not a default alias `_dbt_limit_subq`
- set no alias (but then I think some DBs requires an alias and then it would fail for them with default setup)

seub:
> I finally tried this feature and found one major caveat: I suspect, like me, a lot of dbt users are using `FROM {{ source('schema', 'table') }} my_table_alias` Then using `--empty` will result in `` FROM (select * from `db`.`schema`.`table` where false limit 0) _dbt_limit_subq my_table_alias `` Which obviously won't work.
> 

Hey @github-christophe-oudar, I'm curious about your issue but I'm not following, can you clarify? Why would `--empty` add `_dbt_limit_subq` in your query?

Issue #8356, opened 2023-08-10
[Feature] Postgres Query Timeouts


rchui:
### Is this your first time submitting a feature request?

- [X] I have read the [expectations for open source contributors](https://docs.getdbt.com/docs/contributing/oss-expectations)
- [X] I have searched the existing issues, and I could not find an existing issue for this feature
- [X] I am requesting a straightforward extension of existing dbt functionality, rather than a Big Idea better suited to a discussion

### Describe the feature

When running with an orchestrator like Airflow, users can set DAG timeouts that will issue SIGTERM to running processes. The SIGTERM does not signal to DBT to cancel a running query like SIGINT will and will cause the query to continue running even if it has exceeded the Airflow specified timeout.

This can cause a thundering herd situation when you have an incremental model that runs very frequently. The task will timeout killing the DBT process, restart and run the query again with the old one running. On and on and on.

It would be useful to specify a timeout for a Postgres query to be killed if it exceeds a certain amount of time. This can be achieved by injecting a SQL header to set `statement_timeout` like:
```sql
"SET statement_timeout TO '15min' ; "
```

This could be configured globally in profiles.yml or as a model property
```yml
type: postgres
...
execution_timeout: 15min
```

```sql
{{
  config(
    execution_timeout='15min'
  )
}}
```

### Describe alternatives you've considered

* Handle SIGTERM correctly and treat it like SIGINT. Upon receiving a SIGTERM cancel running queries and exit gracefully

### Who will this benefit?

Anyone who uses a Postgres data base and an external orchestrator that allows setting timeouts external to DBT.

rchui:
@jtcohen6 

I don't believe there is a way to have Airflow issue a SIGINT instead of a SIGTERM on task timeout. Or at least not a method that I am aware of.

I tried to remedy this on my own. I first read this article on graceful signal handling in K8s: https://medium.com/@brianepohl/terminating-dbt-in-dagster-kubernetes-job-c53c3bc26012. This led me to another article about graceful shutdowns in ECS: https://aws.amazon.com/blogs/containers/graceful-shutdowns-with-ecs/.

I implemented a variation of the python signal handler:
```python
import signal, time, os

def shutdown(signum, frame):
    print('Caught SIGTERM, shutting down')
    # Finish any outstanding requests, then...
    exit(0)

if __name__ == '__main__':
    # Register handler
    signal.signal(signal.SIGTERM, shutdown)
    # Main logic goes here
```

[...]

Issue #6217, opened 2022-11-04
[CT-1475] [Bug] Automatic Backport fails


emmyoop:
### Is this a new bug in dbt-core?

- [X] I believe this is a new bug in dbt-core
- [X] I have searched the existing issues, and I could not find an existing issue for this bug

### Current Behavior

We currently had to do some backports to `1.1.latest`, `1.2.latest` and `1.3.latest` ([original PR](#6204))

The `1.3.latest` backport worked flawlessly.

However the `1.1.latest` and the `1.2.latest` both failed.  When I manually cherry picked the commits I was able to do it without conflict so it's unclear why the cherry picks failed.

Additionally, I was told that if the labels are added _before_ the PR has been merged, the backport will not happen.  the labels must be added after the merge is complete.  Unsure if this is a bug or just how it works but wanted to note it here as well.

### Expected Behavior

Backports just work - always!

### Steps To Reproduce

See #6204

emmyoop:
@McKnight-42 did you experience something similar with any of the adapters?

emmyoop:
Inof: backport to `1.2.latest` worked for https://github.com/dbt-labs/dbt-core/pull/6202 today

McKnight-42:
@emmyoop yeah 1.2. and 1.1 both fails with `dbt-snowflake` https://github.com/dbt-labs/dbt-redshift/pull/211

emmyoop:
More examples in #6405

emmyoop:
The most stright forward solution may be to write our own version of the backport so we can be more in control of how it functions and see the actual reason it's failing.

PR #8994, opened 2023-11-03
[IDE] also treat SystemExit exception as an interrupt.

benmosher:
IDE worker process is sent SIGTERM in multiple scenarios, including user abort of a command. IDE task runner (Celery) has a signal handler that raises `SystemExit` when SIGTERM is received. 

Because `SystemExit` is raised instead of `KeyboardInterrupt`, cancelation is not being sent to the data warehouse.

Likely resolves JIRA:TRIAGE-667 -- I will attempt to validate this but wanted to open the PR and get feedback in parallel.

**Update**: testing in a custom IDE build when backported to 1.6-latest* does resolve the JIRA:TRIAGE-667 issue with nodes appearing to run forever, when tested in a Postgres-backed project. I have not yet fully validated with Snowflake specifically.

**Note:** would like to back port this to 1.6 as well.

*I was unable to test on latest because of errors raised about duplicate macros.

### Problem

When IDE commands are canceled, the data warehouse is not being sent cancelation commands.

### Solution

Any system exit caused by raising `SystemExit` will now send cancelation commands.

### Checklist

    - I did not find existing tests for the cancelation on `KeyboardInterrupt`

### DIFF
--- a/.changes/unreleased/Under the Hood-20231106-080422.yaml
+++ b/.changes/unreleased/Under the Hood-20231106-080422.yaml
@@ -1,0 +1,6 @@
+kind: Under the Hood
+body: Treat SystemExit as an interrupt if raised during node execution.
+time: 2023-11-06T08:04:22.022179-05:00
+custom:
+  Author: benmosher
+  Issue: n/a
--- a/core/dbt/task/runnable.py
+++ b/core/dbt/task/runnable.py
@@ -371,7 +371,7 @@ class GraphRunnableTask(ConfiguredTask):
             print_run_result_error(failure.result)
             # ensure information about all nodes is propagated to run results when failing fast
             return self.node_results
-        except KeyboardInterrupt:
+        except (KeyboardInterrupt, SystemExit):
             run_result = self.get_result(
                 results=self.node_results,
                 elapsed_time=time.time() - self.started_at,


jtcohen6:
Related issue from a few months ago:
- https://github.com/dbt-labs/dbt-core/issues/8356

Would this fix address that issue in the general case, or only in cases where `dbt-core` is being invoked by a wrapping process that converts SIGTERM &rarr; `SystemExit`?

benmosher:
@jtcohen6 -- @rockstar and I did some digging and it appears that this PR is only half the work because SIGTERM does not raise `SystemExit` without installing a signal handler [like the one from your linked issue](https://github.com/dbt-labs/dbt-core/issues/8356#issuecomment-1690422876). That + this should give that user the cancelation-on-SIGTERM they're looking for.

In order to not disturb IDE, I'd advocate for installing the signal handler in a code path that won't be hit by `dbtRunner.invoke` but I'm not familiar enough with Core to know exactly where that might be.

jtcohen6:
@benmosher Thanks for the follow-up! I think it should be up to the program wrapping `dbt-core` to do that signal handling.

Discussing live with @ChenyuLInx: For the immediate scope of just this change, if dbt-core has active queries, and receives `SystemExit`, it will try to cancel the queries before the process exits. I buy it.

ChenyuLInx (approved):
Thanks Ben

benmosher:
🙌 thanks for merging this y'all. 

wondering though -- I put the 1.6 and 1.7 backport tags on here while it was open, but AFAICT nothing ported it back yet.

a) are y'all supportive of back porting this to 1.6 and 1.7 for IDE, and if so, 
b) how can I get that rolling? (remove tags and re-add perhaps?)

jtcohen6:
Ah, it looks like a failure of our automated backport process (https://github.com/dbt-labs/dbt-core/issues/6217).

You (or we) can open manual backports against those two release branches, by checking out the `1.6.latest` + `1.7.latest` branches, opening new branches, and cherry-picking this commit (https://github.com/dbt-labs/dbt-core/commit/931b2dbe4069cabe33bb03ad27452a9771d814e1).

benmosher:
oof, I tried jiggling the tags at the same time you did but you won the race 😅 thanks!

PR #9004, opened 2023-11-06
Support new agate Integer data_type in adapter code

### LINKED ISSUE #8895: [CT-3259] [Regression] Running empty seed file raises unhandled exception

### Current Behavior

We have an empty seed file in one of our dbt projects (Sometimes we need to manually intervene and add lines to it) that has been running successfully for a while now but after upgrading to dbt 1.6.6 you get an unhandled exception when compiling:

```
12:21:14  Running with dbt=1.6.6
12:21:15  Registered adapter: bigquery=1.6.7
12:21:15  Unable to do partial parsing because of a version mismatch
12:21:16  Found 3 models, 2 seeds, 0 sources, 0 exposures, 0 metrics, 394 macros, 0 groups, 0 semantic models
12:21:16  
12:21:25  Concurrency: 8 threads (target='integration-local')
12:21:25  
12:21:25  2 of 2 START seed file dry_run.my_seed ............................................................................................................................................. [RUN]
12:21:25  1 of 2 START seed file dry_run.empty_seed .......................................................................................................................................... [RUN]
12:21:25  Unhandled error while executing 
'NoneType' object has no attribute 'upper'
12:21:25  1 of 2 ERROR loading seed file dry_run.empty_seed .................................................................................................................................. [ERROR in 0.09s]
12:21:30  2 of 2 OK loaded seed file dry_run.my_seed ......................................................................................................................................... [INSERT 2 in 4.28s]
12:21:30  
12:21:30  Finished running 2 seeds in 0 hours 0 minutes and 13.56 seconds (13.56s).
12:21:30  
12:21:30  Completed with 1 error and 0 warnings:
12:21:30  
12:21:30    'NoneType' object has no attribute 'upper'
12:21:30  
12:21:30  Done. PASS=1 WARN=0 ERROR=1 SKIP=0 TOTAL=2

```

In `dbt 1.5.x` it works fine:

```
12:24:09  Running with dbt=1.5.8
12:24:10  Registered adapter: bigquery=1.5.7
12:24:10  Unable to do partial parsing because of a version mismatch
12:24:10  Found 3 models, 0 tests, 0 snapshots, 0 analyses, 360 macros, 0 operations, 2 seed files, 0 sources, 0 exposures, 0 metrics, 0 groups
12:24:10  
12:24:11  Concurrency: 8 threads (target='integration-local')
12:24:11  
12:24:11  1 of 2 START seed file dry_run.empty_seed .......................................................................................................................................... [RUN]
12:24:11  2 of 2 START seed file dry_run.my_seed ............................................................................................................................................. [RUN]
12:24:15  2 of 2 OK loaded seed file dry_run.my_seed ......................................................................................................................................... [INSERT 2 in 3.76s]
12:24:16  1 of 2 OK loaded seed file dry_run.empty_seed ...................................................................................................................................... [INSERT 0 in 4.34s]
12:24:16  
12:24:16  Finished running 2 seeds in 0 hours 0 minutes and 5.52 seconds (5.52s).
12:24:16  
12:24:16  Completed successfully
12:24:16  
12:24:16  Done. PASS=2 WARN=0 ERROR=0 SKIP=0 TOTAL=2

``` 

### Steps To Reproduce

1. In dbt 1.6.x (I am using Big Query) create a project and create a seed with just headings:

```
#my_seed.csv
a,b,c
```

Then when you run `dbt build/seed` you will see the error.

### Relevant log output

```shell
Already posted.
```

### Environment

```markdown
- OS: Mac OS
- Python: 3.8.16
- dbt (working version): 1.5.8
- dbt (regression version): 1.6.6
```

bigquery

### Additional Context

_No response_

gshank:
resolves #8895

### Problem

In #8153 we added an agate Integer data type, but did not include conversion methods that are used when constructing a csv table.

### Solution

Add Integer type to 'convert_agate_type' and add new 'convert_integer_type' method.

### Checklist


.changes/unreleased/Fixes-20231031-144837.yaml
```diff
@@ -0,0 +1,6 @@
+kind: Fixes
+body: Fix exception running empty seed file
```
colin-rogers-dbt, line 2: The PR and this changelog indicate we're solving multiple issues, can we add a second changelog to cover new data_type?

colin-rogers-dbt:
do we need to beef up our tests around dbt-show or docs to catch this? https://github.com/dbt-labs/dbt-core/blob/main/tests/functional/show/test_show.py#L104C3-L104C3

.changes/unreleased/Fixes-20231031-144837.yaml
gshank: This doesn't solve multiple issues, it's all the same issue. The problem with the empty seed file was happening because of the lack of support for the new Integer data type.

gshank:
I think the "show" tests are adequate. The problem was that the show command was fixed by changing infrastructure that also affected the code used to create an empty table for an empty seed. I added a new seed test, but probably that test should be moved to the adapter zone seed tests.

.changes/unreleased/Fixes-20231031-144837.yaml
gshank: Oh, you mean additional info in the existing changelog? Sure.

mikealfare:
I want to consider different scenarios with how adapter maintainers may have implemented this, and see how this solution would potentially affect them. (I'm basically processing out loud here.)

1. The adapter has overwritten `convert_agate_type`.
In this scenario, the adapter remains stable because they don't actually pick up this change.
2. The adapter has not overwritten `convert_agate_type`, but has an existing method `convert_integer_type`.
This scenario is extremely unlikely. And in the event that this occurred, I think it's fair to assume what `convert_integer_type` will be used for given the existing naming convention. If this breaks an adapter, I think it's fair to assume the adapter maintainer should update the adapter.
3. The adapter has not overwritten `convert_agate_type`, nor is there an existing method `convert_integer_type`. Hence these changes are fully implemented. And `'integer'` is a valid type for the database platform.
This is a likely scenario, and is the scenario we're solving for. We want this.
4. The adapter has not overwritten `convert_agate_type`, nor is there an existing method `convert_integer_type`. Hence these changes are fully implemented. However, `'integer'` is not a valid type for the database platform.
This is a less likely scenario, but not an unlikely scenario. This will break the adapter unexpectedly as `Integer` types will all of a sudden get converted whereas in the past they would have just flowed through. I'm concerned about this scenario.

@gshank / @colin-rogers-dbt, are there any other scenarios that you can think of? And in particular, are we comfortable with scenario 4 above, where we will break adapters which have not tried to implement something like this in the past and which do not support "integer" as a valid type?

mikealfare:
I'll focus again on scenario 4 above, since I think that's the only problematic scenario. Given the following observations:

- this is broken (hence we need to do something)
- `Integer` is only present in 1.6+
- the adapter would need to have implemented `Integer` to be impacted by scenario 4
- in the event scenario 4 is an issue, the fix is trivial (overwrite `convert_integer_type` to return the appropriate data type)
- combining `convert_integer_type` with `convert_number_type` does not address the concern of attempting to use an invalid type of "integer" (and makes the code more complicated)

I think I'm comfortable with the approach. To confirm, are the observations above valid?

.changes/unreleased/Fixes-20231031-144837.yaml
colin-rogers-dbt: yup! not questioning the PR just want to make sure our changelogs cover everything that's being changed

colin-rogers-dbt (approved):

colin-rogers-dbt:
> To confirm, are the observations above valid?

@mikealfare I _believe_ so, I'm comfortable with this approach

mikealfare (approved):

gshank:
Yes, those observations are valid.  

Although it's not just that combining the integer and number types makes the code complicated, with the current method of calling MaxPrecision it can't be gotten to work...  We'd have to update every "convert_number_type" method to do ... something else.

Thanks for all of the attention!

Issue #9000, opened 2023-11-05
[CT-3320] [Regression] DBT deps no longer working on git repository with subdirectory


DanMawdsleyBA:
### Is this a regression in a recent version of dbt-core?

- [X] I believe this is a regression in dbt-core functionality
- [X] I have searched the existing issues, and I could not find an existing issue for this regression

### Current Behavior

Install dbt deps with a github repository using a subdirectory no longer works. I was using 1.5.5 previously but since upgrading to 1.7 deps is no longer working.

### Expected/Previous Behavior

To be able to run deps from a GitHub repository like previously.

### Steps To Reproduce

Run the dbt deps command with a packages file having a dependency on github repository with a subdirectory

### Relevant log output

```shell
[0m12:55:54.901755 [debug] [MainThread]: Checking out revision BADICREW-1-inc-0143754-easa-cabin-flight-crew-dashboard-data-hermes.
[0m12:55:54.903756 [debug] [MainThread]: Executing "git remote set-branches origin BADICREW-1-inc-0143754-easa-cabin-flight-crew-dashboard-data-hermes"
[0m12:55:55.090351 [debug] [MainThread]: STDOUT: "b''"
[0m12:55:55.090351 [debug] [MainThread]: STDERR: "b''"
[0m12:55:55.092917 [debug] [MainThread]: Executing "git fetch origin --depth 1 --tags BADICREW-1-inc-0143754-easa-cabin-flight-crew-dashboard-data-hermes"
[0m12:56:04.606659 [debug] [MainThread]: STDOUT: "b''"
[0m12:56:04.607668 [debug] [MainThread]: STDERR: "b'From https://github.com/BritishAirways-Ent/olympus-oprn-crw-ops-planning\n * branch            BADICREW-1-inc-0143754-easa-cabin-flight-crew-dashboard-data-hermes -> FETCH_HEAD\n'"
[0m12:56:04.607668 [debug] [MainThread]: Executing "git tag --list"
[0m12:56:04.766131 [debug] [MainThread]: STDOUT: "b''"
[0m12:56:04.766131 [debug] [MainThread]: STDERR: "b''"
[0m12:56:04.766131 [debug] [MainThread]: Executing "git reset --hard origin/BADICREW-1-inc-0143754-easa-cabin-flight-crew-dashboard-data-hermes"
[0m12:56:05.199503 [debug] [MainThread]: STDOUT: "b'HEAD is now at b47a4a3 Trivial change\n'"
[...]

PR #9019, opened 2023-11-07
Fix lock for git subduer

ChenyuLInx:
#9000
Fix lock file for git with subdirectories.
Issue was that when writing out the lock file we didn't write out the complete information.


peterallenwebb (approved):
Let's add type annotations. Otherwise looks good.

Issue #6013, opened 2022-10-06
[CT-1301] Behavior change: `full_refresh: true` config should respect `--full-refresh, -f` flag


jtcohen6:
https://docs.getdbt.com/reference/resource-configs/full_refresh

### Current behavior
When configured, `full_refresh: true|false` means: "always/never full-refresh, and always ignore the --full-refresh flag"

Added way back in https://github.com/dbt-labs/dbt-core/pull/2438 / v0.18

### Proposed change
- `full_refresh: true` should mean "always respect the --full-refresh flag"
- `full_refresh: false` should mean "never full refresh, under any circumstances"

That's trickier for intuition, and harder to explain, but definitely more desirable in terms of the behavior it enables. I'm not sure why you'd always want to full-refresh an `incremental` model; at that point it's really just a `table` materialization.

If this is set at the project level:
```yml
models:
  +full_refresh: false
```
It's not possible to override that setting for one particular model ("actually I want this model to respect the `--full-refresh` flag"). This doesn't work as I'd hoped it might:
```sql
{{ config(materialized = 'incremental', full_refresh = none) }}
```

### Is this a breaking change?

Only for people who currently make use of `full_refresh: true`. Again, I'm not actually sure what the use case for doing that would be!

b-per:
The only reason I could think of people using `full_refresh: true` would be from some `env_var()`, e.g. always doing a full-refresh in Dev (and maybe limiting the number of rows) but keeping it incremental in Prod. This looks very niche though. Most of the time that I need this parameter it is to set it to `false` to prevent FR.

b-per:
I think that it is worth keeping this issue active

b-per:
Commenting to keep it on! I still think that we could improve incremental models handling.

jtcohen6:
@b-per I agree! I think we'd want to find some way to change the default here, without breaking existing users' projects. If you can think of a clever way to do that, without requiring _another_ opt-in config (`full_refresh_config_respect_flag: true`), I'm all ears.

This behavior would then also diverge from `--store-failures` / `store_failures`, where the node-level config always takes precedence over the flag. Should we change that behavior too, to match?

andrii0yerko:
I had faced the same problem.

Did anyone find any workarounds to achieve respect for the `--full-refresh` flag?

I tried things like
```
{{
    config(
        materialized='incremental',
        full_refresh = flags.FULL_REFRESH,
        ...
}}
```

[...]

PR #9020, opened 2023-11-07
add flag --no-skip-on-failure for command build, run, retry, seed

### LINKED ISSUE #2142: Option to disable "skip downstream models on failure"
### Describe the feature
When a model run fails, all dependent, downstream model runs are skipped. This is a great default, but there can be some cases where you still want to run the downstream model.

For example, one case where I would actually want this default behaviour overridden is for jobs that rely on an infrequently updated dimension. Something like a *country dimension* or *currency dimension* which changes very rarely. Say you have a daily summary/rollup model that uses *order transaction data* and a *country dimension*. If *country dimension* starts failing i’d still want my model to keep running off the new *order transaction data*, and not have to wait until the responsible team fixes it.

Individually running models and keeping track of which ones ran successfully and which ones failed. This would be quite time consuming.

### Who will this benefit?
Can't comment on how many people this will benefit, but the added flexibility is always a plus.

leo-schick:
resolves #2142

### Problem

Downstream models are skipped. Sometimes e.g. when working in big data warehouses, it might be desirable to proceed with downstream models even a upstream model failed.

### Solution

Adds a new flag `--no-skip-on-failture` for the dbt commands `build`, `run`, `retry`, `clone` and `seed`. If set, the downstream models will not be added to `self._skipped_children` and therefore will not be skipped.

Note: user docs might need to be updated.

### Testing

I created a initial project with two additional models:
* `must_fail` - always fails
* `dependent_model` - uses the failing model

#### Test 1 - run without the new flag

image

The model `dependent_model` is still skipped.

#### Test 2 - run with flag `--no-skip-on-failture`

image

The model `dependent_model` is now executed, but fails since the table for `must_fail` does not exist. (This is expected behavior. In a real world scenario the failing model table might already exist from a previous execution and the model `dependent_model` would process successfully).

#### Test 3 - run retry with flag `--no-skip-on-failture`

image

The failing models are executed as expected.

### Checklist

- [x] This PR includes tests, or tests are not required/relevant for this PR *(IMO not required; if wanted please assist how to build the tests)*
- [x] This PR includes [type annotations](https://docs.python.org/3/library/typing.html) for new and modified functions *(not required)*

### DIFF
--- a/.changes/unreleased/Features-20231107-132842.yaml
+++ b/.changes/unreleased/Features-20231107-132842.yaml
@@ -1,5 +1,5 @@
 kind: Features
-body: add flag --no-skip-on-failture
+body: add flag --no-skip-on-failure
 time: 2023-11-07T13:28:42.420727773+01:00
 custom:
   Author: leo-schick
--- a/.changes/unreleased/Features-20231107-132842.yaml
+++ b/.changes/unreleased/Features-20231107-132842.yaml
@@ -1,0 +1,6 @@
+kind: Features
+body: add flag --no-skip-on-failture
+time: 2023-11-07T13:28:42.420727773+01:00
+custom:
+  Author: leo-schick
+  Issue: "2142"
--- a/core/dbt/cli/main.py
+++ b/core/dbt/cli/main.py
@@ -174,7 +174,7 @@ def cli(ctx, **kwargs):
 @p.export_saved_queries
 @p.full_refresh
 @p.deprecated_include_saved_query
-@p.no_skip_on_failture
+@p.no_skip_on_failure
 @p.profiles_dir
 @p.project_dir
 @p.resource_type
@@ -542,7 +542,7 @@ def parse(ctx, **kwargs):
 @global_flags
 @p.exclude
 @p.full_refresh
-@p.no_skip_on_failture
+@p.no_skip_on_failure
 @p.profiles_dir
 @p.project_dir
 @p.empty
@@ -580,7 +580,7 @@ def run(ctx, **kwargs):
 @p.profiles_dir
 @p.vars
 @p.target_path
-@p.no_skip_on_failture
+@p.no_skip_on_failure
 @p.threads
 @p.full_refresh
 @requires.postflight
@@ -677,7 +677,7 @@ def run_operation(ctx, **kwargs):
 @global_flags
 @p.exclude
 @p.full_refresh
-@p.no_skip_on_failture
+@p.no_skip_on_failure
 @p.profiles_dir
 @p.project_dir
 @p.select
--- a/core/dbt/cli/main.py
+++ b/core/dbt/cli/main.py
@@ -174,6 +174,7 @@ def cli(ctx, **kwargs):
 @p.export_saved_queries
 @p.full_refresh
 @p.deprecated_include_saved_query
+@p.no_skip_on_failture
 @p.profiles_dir
 @p.project_dir
 @p.resource_type
@@ -541,6 +542,7 @@ def parse(ctx, **kwargs):
 @global_flags
 @p.exclude
 @p.full_refresh
+@p.no_skip_on_failture
 @p.profiles_dir
 @p.project_dir
 @p.empty
@@ -578,6 +580,7 @@ def run(ctx, **kwargs):
 @p.profiles_dir
 @p.vars
 @p.target_path
+@p.no_skip_on_failture
 @p.threads
 @p.full_refresh
 @requires.postflight
@@ -674,6 +677,7 @@ def run_operation(ctx, **kwargs):
 @global_flags
 @p.exclude
 @p.full_refresh
+@p.no_skip_on_failture
 @p.profiles_dir
 @p.project_dir
 @p.select
--- a/core/dbt/cli/main.py
+++ b/core/dbt/cli/main.py
@@ -609,6 +609,7 @@ def retry(ctx, **kwargs):
 @global_flags
 @p.exclude
 @p.full_refresh
+@p.no_skip_on_failure
 @p.profiles_dir
 @p.project_dir
 @p.resource_type
--- a/core/dbt/cli/params.py
+++ b/core/dbt/cli/params.py
@@ -120,6 +120,13 @@ fail_fast = click.option(
     help="Stop execution on first failure.",
 )
 
+no_skip_on_failture = click.option(
+    "--no-skip-on-failture",
+    envvar="DBT_NO_SKIP_ON_FAILTURE",
+    help="If specified, dbt will proceed with downstream models even the dependent model failed.",
+    is_flag=True,
+)
+
 favor_state = click.option(
     "--favor-state/--no-favor-state",
     envvar="DBT_FAVOR_STATE",
--- a/core/dbt/cli/params.py
+++ b/core/dbt/cli/params.py
@@ -120,9 +120,9 @@ fail_fast = click.option(
     help="Stop execution on first failure.",
 )
 
-no_skip_on_failture = click.option(
-    "--no-skip-on-failture",
-    envvar="DBT_NO_SKIP_ON_FAILTURE",
+no_skip_on_failure = click.option(
+    "--no-skip-on-failure",
+    envvar="DBT_NO_SKIP_ON_FAILURE",
     help="Proceed with downstream nodes even if an upstream node fails.",
     is_flag=True,
 )
--- a/core/dbt/cli/params.py
+++ b/core/dbt/cli/params.py
@@ -123,7 +123,7 @@ fail_fast = click.option(
 no_skip_on_failture = click.option(
     "--no-skip-on-failture",
     envvar="DBT_NO_SKIP_ON_FAILTURE",
-    help="If specified, dbt will proceed with downstream models even the dependent model failed.",
+    help="Proceed with downstream nodes even if an upstream node fails.",
     is_flag=True,
 )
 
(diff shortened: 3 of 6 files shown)


chancekbarkley:
@leo-schick do you know where this PR stands today?  Looks like would be great for our use case.

i-brahimi:
> It works

@leo-schick you said it works but you didn't specify on which version of DBT

leo-schick:
@i-brahimi It works on the main development branch which uses version tag `1.8.0a1` which is not released yet. I don't know why the dbt core team did not respond on this PR yet since so many people where requesting this feature. Maybe someone can bump them to take a look at this and possibly merge it for the next upcoming release.

Aarboush:
There are so many people using DBT waiting for this feature !

.changes/unreleased/Features-20231107-132842.yaml
```diff
@@ -0,0 +1,6 @@
+kind: Features
+body: add flag --no-skip-on-failture
```
cwcartmell, line 2:
```suggestion
body: add flag --no-skip-on-failure
```

"failture" should be changed to "failure" here and at all other usages.

core/dbt/cli/main.py
```diff
@@ -189,6 +189,7 @@ def cli(ctx, **kwargs):
 @p.exclude
 @p.full_refresh
 @p.include_saved_query
+@p.no_skip_on_failture
```
cwcartmell, line 192:
```suggestion
@p.no_skip_on_failure
```

I think that this flag should be added to `dbt clone` on line 637 as well.

core/dbt/cli/params.py
```diff
@@ -112,6 +112,13 @@
     help="Stop execution on first failure.",
 )
 
+no_skip_on_failture = click.option(
+    "--no-skip-on-failture",
+    envvar="DBT_NO_SKIP_ON_FAILTURE",
+    help="If specified, dbt will proceed with downstream models even the dependent model failed.",
```
cwcartmell, line 118:
```suggestion
    help="Proceed with downstream nodes even if an upstream node fails.",
```

This language is more in line with that of the other option help messages.

cwcartmell (commented):
@leo-schick, thank you so much for opening this PR! I just ran into the same issue so I'm thrilled to see that someone is already working on it.

leo-schick:
@cwcartmell all requested changes are done now.

BenLiyanage:
Has there been any word on getting this feature into dbt?

cwcartmell (commented):
I've had some more time on my hands so I came back to this PR post-dbt 1.8.0. I think that [changes](https://github.com/dbt-labs/dbt-core/pull/9956) that have been made recently have broken this PR. When I pull down the fork and pull `upstream main` into the branch, I get errors when trying to run dbt commands:

```console
$ dbt test               
01:42:42  Running with dbt=1.9.0-a1
01:42:42  Registered adapter: postgres=1.9.0-a1
01:42:42  Unable to do partial parsing because profile has changed
01:42:43  Found 2 models, 4 data tests, 413 macros
01:42:43  
01:42:44  Concurrency: 4 threads (target='dev')
01:42:44  
01:42:44  1 of 4 START test not_null_my_first_dbt_model_id ............................... [RUN]
01:42:44  2 of 4 START test not_null_my_second_dbt_model_id .............................. [RUN]
01:42:44  3 of 4 START test unique_my_first_dbt_model_id ................................. [RUN]
01:42:44  4 of 4 START test unique_my_second_dbt_model_id ................................ [RUN]
01:42:44  2 of 4 ERROR not_null_my_second_dbt_model_id ................................... [ERROR in 0.12s]
01:42:44  1 of 4 ERROR not_null_my_first_dbt_model_id .................................... [ERROR in 0.13s]
01:42:44  4 of 4 ERROR unique_my_second_dbt_model_id ..................................... [ERROR in 0.09s]
01:42:44  3 of 4 ERROR unique_my_first_dbt_model_id ...................................... [ERROR in 0.10s]
Exception in thread Thread-7:
  File "/home/conrad/.pyenv/versions/3.8.13/lib/python3.8/threading.py", line 932, in _bootstrap_inner
    self.run()
  File "/home/conrad/.pyenv/versions/3.8.13/lib/python3.8/threading.py", line 870, in run
    self._target(*self._args, **self._kwargs)
  File "/home/conrad/.pyenv/versions/3.8.13/lib/python3.8/multiprocessing/pool.py", line 592, in _handle_results
    cache[job]._set(i, obj)
  File "/home/conrad/.pyenv/versions/3.8.13/lib/python3.8/multiprocessing/pool.py", line 776, in _set
    self._callback(self._value)
  File "/home/conrad/Documents/dbt-core/core/dbt/task/runnable.py", line 281, in callback
    self._handle_result(result)
  File "/home/conrad/Documents/dbt-core/core/dbt/task/compile.py", line 132, in _handle_result
    super()._handle_result(result)
  File "/home/conrad/Documents/dbt-core/core/dbt/task/runnable.py", line 338, in _handle_result
    self._mark_dependent_errors(node.unique_id, result, cause)
  File "/home/conrad/Documents/dbt-core/core/dbt/task/runnable.py", line 421, in _mark_dependent_errors
    no_skip_on_failure = get_flags().NO_SKIP_ON_FAILURE
  File "/home/conrad/Documents/dbt-core/core/dbt/cli/flags.py", line 360, in __getattr__
    return super().__getattribute__(name)  # type: ignore
AttributeError: 'Flags' object has no attribute 'NO_SKIP_ON_FAILURE'
```

BenLiyanage:
Seems pretty unfortunate.  This is a pretty critical feature to prevent one failure from having everything fail.

jtcohen6 (commented):
@leo-schick I'm sorry for the delayed response here. This issue had been open for a long time; your PR has helped it gain renewed traction.

The last time I thought about this in detail was [~2 years ago](https://github.com/dbt-labs/dbt-core/issues/2142#issuecomment-1125917049), and we’ve been trying to define a clearer, narrower, more maintainable scope for `dbt-core` since then. I’ve become warier of feature requests that feel they go against some of dbt’s core principles, one of which is the integrity of the DAG. Whether to skip downstream nodes on upstream errors is less of a dogmatic concern than allowing node-level cycles, or running the DAG out of order, but it’s one of the simplifying assumptions that’s gotten us a lot over time.

That being said:
- It’s already possible to achieve the end state described here, by retrying a failed job while simply excluding the failing models from the previous run. Users have pointed out that this is already possible (retry from point of failure, skip point of failure, keep building downstream) — but it requires quite a bit of custom work(around).
- This is analogous to marking a test as “warn”: notify me that it detected failures, but don’t stop the DAG for it.

I reread #2142 to remind myself of the use cases we’re solving for, and the majority of the examples provided deal with issues with upstream data ingestion that can cause flakey model failure. People would rather have stale data coming from one place than stale transformations across the board. There is a judgment call here, model by model. To that end, I am more convinced by the value of this as a model-level config, defined on parent model, than as a runtime-level config set for all models at once.

Previously I said:

> If we decide to support both, I'd see this being similar to how the --full-refresh flag / full_refresh model config work today ([docs](https://docs.getdbt.com/reference/resource-configs/full_refresh)), where the model-level config (if set) takes precedence over the run-level flag.

That doesn’t feel right to me. It good insofar as it’s consistent with full-refresh and store-failures, but I don’t think [we nailed the interplay of full-refresh flag and model-level config](https://github.com/dbt-labs/dbt-core/issues/6013).

I coul see this as behavior that wants to differ across environments: dbt should always skip (strict) in dev/CI, but allow specific downstream models to build in deployment environments. It would be possible to accomplish that with conditional Jinja configs, but probably better to have some rule that combines both a model configuration and a flag setting.

So here are the options I see, in order of preference (1 then 2 then 3 then 4):
1. By default, any model in any environment skips children on failures. A specific flaky model can opt into on_error: continue for a specific environment.
2. By default, any model in any environment skips children on failure. A specific model can opt into on_error: continue for all environments (dev, prod, CI, etc).
3. By default, any model in any environment skips children on failures. In some invocations/environments, you can optionally switch this behavior for all models to on_error: continue.
4. By default, any model in any environment skips children on failures. For some models, you can set them to always skip or always continue. In some invocations/environments, can also optionally switch the behavior for all other (unset) models.

This PR implements (3), with the option to add in node-level config later on (4). I don’t think we could go from 3 -> 2 without a breaking change, since the continue-on-skip behavior becomes “AND” instead of “OR.”

Where I'm landing is: I'm open to adding this as a node-level config (exception not the rule), but I don't think we should start with the global runtime flag. If we decide a flag is appropriate to support conditional per-environment behavior ("strict" in dev/CI, "forgiving" in Prod), we can do that later on.

Next steps:
- Do you (and the folks who have been following this thread) agree with my line of reasoning above?
- Would you (or someone else) be interested in carrying that work forward? Again, apologies for the delay on our end, and I understand if it means you're no longer able or interested.

mchurley:
For my team our need is specifically to have a global runtime flag. The majority of our pipeline is constructed in a way where we still want to run the entire pipeline regardless of the state of parent models and leave the decision making regarding the validity of that data to a later point. As an example, there might be one column in a later model that is impacted by a parent having incorrect data/not running due to an error but that does not mean we would not want to refresh the table as a whole anyway. It is more beneficial to report that inconsistency in a different manner rather than simply not have that child model run.

As a direct consequence of not having this we are making the decision to maintain specific DAG tasks per model/test which is incredibly inefficient and butchers a lot of the great features dbt provides.

core/dbt/task/runnable.py
```diff
@@ -450,8 +450,10 @@ def _mark_dependent_errors(
     ) -> None:
         if self.graph is None:
             raise DbtInternalError("graph is None in _mark_dependent_errors")
-        for dep_node_id in self.graph.get_dependent_nodes(UniqueId(node_id)):
-            self._skipped_children[dep_node_id] = cause
+        no_skip_on_failure = get_flags().NO_SKIP_ON_FAILURE
+        if not no_skip_on_failure:
```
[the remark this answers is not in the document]
hope-wat: strong agree

PR #9032, opened 2023-11-08
Unit test support for `state:modified` and `--defer`

### LINKED ISSUE #8517: [CT-3046] Enable use of state:modified for detecting changed unit tests
### Description

We can select unit tests using `state:modified` for changes to existing models. 

In addition we'd like to be able to select unit tests that have changed using `state:modified`.

Changes include:

- Definition of the unit test
- Changes to the inputs(ie: fixtures)
- Changes to the logic of the model the unit test is for (depending on [indirect_selection](https://docs.getdbt.com/reference/node-selection/test-selection-examples#indirect-selection) config)

### Acceptance Criteria

Change only the contents of a unit test definition, and do `dbt unit-test --select state:modified` and the changed unit test should run.

Change the SQL for a given model and do `dbt unit-test --select state:modified`, the unit test on the changed model should run (depending on [indirect_selection](https://docs.getdbt.com/reference/node-selection/test-selection-examples#indirect-selection) config)

### Impact to other teams
                  
None                 
                                
No

jtcohen6:
resolves #8517 and #9053

### Problem

Unit tests should be selectable via `state:modified`. This includes cases where the unit test definition itself has changed, and where the underlying model has changed (following the existing behavior of `--indirect-selection`).

Unit tests should _also_ respect `--defer`. This enables column type checking against the production versions of unbuilt models upstream of the model being unit tested.

### Solution

- Implement `same_contents` for `UnitTestDefinition`, and add it as a resource type to `state:modified`.
- I added a `checksum` for faster comparison — but if this is over-engineered, I'm happy to do something simpler.
- Unit tests are not yet selected by `list` or `build`

### Checklist


core/dbt/task/unit_test.py
```diff
@@ -206,13 +207,25 @@ def build_unit_test_manifest(self):
...
         # We have the selected models from the "regular" manifest, now we switch
         # to using the unit_test_manifest to run the unit tests.
         self.using_unit_test_manifest = True
         self.manifest = self.build_unit_test_manifest()
         self.compile_manifest()  # create the networkx graph
         self.job_queue = self.get_graph_queue()
 
+    def before_run(self, adapter, selected_uids: AbstractSet[str]) -> None:
+        # We already did cache population + deferral earlier (above)
+        # and we don't need to create any schemas
```
MichelleArk, line 226: 👍

MichelleArk:
This was implemented before CSV file fixtures were supported (https://github.com/dbt-labs/dbt-core/pull/9044) and unfortunately doesn't 'just work' in combination with file-based fixtures. Crucially if a unit test fixture is modified, _all_ downstream unit tests that build on top of this fixture should be re-run. Otherwise, it'd be possible to introduce a change to a fixture that surfaces errors in later CI builds, when the test/model is actually modified. 

I tried a couple naive approaches, without success, to get them working together: 
1. Add `given.inputs::fixture` to the unit test checksum
    a. This felt promising and the way this _should_ work but unfortunately does not because of partial parsing. File-based fixtures are not considered as part of the diff so the node is not re-parsed, meaning no checksum is built on subsequent runs even if a file-based fixture has changes to it. We'd need to include a file-based fixtures as linked files during partial parsing for this to work as expected.
2. Mark any unit test that depends on a file-based fixture as having been modified between runs
    This is super defensive and feels overly crude - having a single unit test that uses a file-based fixture would mean it's impossible for state:modified selection to return 0 results. 

I've implemented the beginnings of (1) including a test (that uses `--no-partial-parse` explicitly and have made a follow-up (p0) issue to wrap up the partial parsing component here: https://github.com/dbt-labs/dbt-core/issues/9067

MichelleArk (approved):

core/dbt/task/unit_test.py
gshank: After moving the calling of unit tests into the test task, this is going to be harder to do differently like this since running unit tests will hook into the standard runnable flow.

gshank (approved):
I think that moving unit testing into the test and build tasks will probably require moving where we're populating the adatper cache in this.  But I can deal with that in my pull request. The important thing is probably to have the tests...

core/dbt/task/unit_test.py
MichelleArk: yeah, exactly. happy to brainstorm alternative approaches in the test task work

Issue #2740, opened 2020-09-03
Automating Non Regression Test : How to get a ref() to the deferred version of the selected model ?


fabrice-etanchaud:
### Describe the feature

Based on the new --defer feature, one would like to automate non regression test (NRT), comparing local model data with the same deferred model data. How one could get the ref() of the deferred model to be given to a NRT macro comparing the common columns on the common primary key values ?

### Describe alternatives you've considered
No alternative so far

### Who will this benefit?
All dbt users who want to automate non regression testing.

### Are you interested in contributing this feature?
Yes !

jtcohen6:
Interesting idea, @fabrice-etanchaud! Based on how we've defined `--defer` to work in v0.18.0, a model can _either_ be selected _or_ deferred at runtime. I don't think the current behavior would support ref'ing the same node across multiple manifests.

That said, the foundational work here around artifact/state comparison—the surface of which we've just barely scratched—could totally make something like this possible in the future. Maybe we add a "defer" argument to `ref`, like "use the other one if it's available," such that we could code up a new [audit-helper](https://hub.getdbt.com/fishtown-analytics/audit_helper/latest/) macro ???

fabrice-etanchaud:
Hi @jtcohen6 , 

currently I am using the macro below to check for non regression.
I use it in deferred mode, the unmanaged model is the dev model, and in order to ref the managed version of the same model (in production, for example), I have to mention the database, schema and identifier explicitly.
Having a way to use ref() instead would be imho a terrific feature. Is there any plan ?
Can I help ?

Best,
Fabrice

```jinja
{% macro non_regression_test(unmanaged_model, managed_model_database, managed_model_schema, managed_model_identifier, pk) %}
-- depends_on: {{ ref(unmanaged_model) }}
{% set unmanaged_model = ref(unmanaged_model) %}
{% set managed_model = api.Relation.create(database=managed_model_database, schema=managed_model_schema, identifier=managed_model_identifier) %}
{% set managed_cols = adapter.get_columns_in_relation(managed_model) %}
{% set managed_colnames = managed_cols | map(attribute='name') | list %}
{% set unmanaged_cols = adapter.get_columns_in_relation(unmanaged_model) %}
{% set unmanaged_colnames = unmanaged_cols | map(attribute='name') | list %}
{% set pk_cols = managed_cols | selectattr('name', 'in', pk) | list %}
{% set pk_colnames = pk_cols | map(attribute='name') | list %}
[...]

Issue #7036, opened 2023-02-23
[CT-2169] [Bug] All global configs should also be settable in ProjectFlags


jtcohen6:
_Opening from https://github.com/dbt-labs/dbt-core/issues/6903#issuecomment-1437439141_

There are some "global configs" that are supported as params (flag / env var), but missing in [`ProjectFlags`](https://github.com/dbt-labs/dbt-core/blob/3e2ec1601b4c41abe4640124be88ee38bd0569e2/core/dbt/contracts/project.py#L297-L318).

Acceptance criteria:
- All "global configs" should be settable in three places (CLI flag, env var, `dbt_project.yml`). [These decorators](https://github.com/dbt-labs/dbt-core/blob/a1b067c683212bccd9bc10f6e52bcdd168ccf7ec/core/dbt/cli/main.py#L129-L158), documented here:
https://docs.getdbt.com/reference/global-configs
- EXCEPT configs whose type is `click.Path` (`--log-path`, `--target-path`), which do not make sense to define in a separate file

Relevant, but very much out of scope for this issue:
- #6207

jtcohen6:
But maybe not `--log-path` or `--target-path`. Not sure any "path"-specific configs make a whole lot of sense! :)

jtcohen6:
Switching this from `tech_debt` to `bug`, because it's actually unexpected behavior, given our current documentation

dbeatty10:
> [!CAUTION]
> This issue is in the same area of code as #9183 / #9289.
> 
> Any solutions to #9183 and #7036 should be compared to each other to make sure one doesn't accidentally stomp on / overwrite the solution of the other.

jtcohen6:
The remaining scope of this issue is to decide whether these "global" flags should be settable in `dbt_project.yml`:
- `print`
- `quiet`
- `log_cache_events`

The question being: Do these feel _different enough_ from all other global configs to justify the inconsistency? Right now, we only disallow setting file-path flags (`--log-path`, `--target-path`, `--state`) in the project `flags`.

dbeatty10:
`print`,  `quiet`, and `log_cache_events` don't seem different enough to me to justify the inconsistency.

At various times, I've found myself tying to set each of these within `profiles.yml` at different times. Granted, `print` and `log_cache_events` would be most likely to be used while troubleshooting. And my uses of `quiet` were specifically to get parseable output from [`dbt list`](https://docs.getdbt.com/reference/commands/list).

[...]

PR #9040, opened 2023-11-08
Refactor deferral: always `merge_from_artifact` & support all commands

### LINKED ISSUE #7965: [CT-2749] [Spike] Investigate merging `merge_from_artifact` and `add_from_artifact`

### LINKED ISSUE #8715: [CT-3159] [spike] Globalize remaining Click flags
### Housekeeping

- [x] I am a maintainer of `dbt-core`

### Short description

Follow-up from https://github.com/dbt-labs/dbt-core/issues/8670, globalize the following flags so that they can be passed in after the sub command and still work:

- [ ] @p.warn_error
- [ ] @p.warn_error_options
- [ ] @p.log_format

### Acceptance Criteria

The flags above are part of the following method

https://github.com/dbt-labs/dbt-core/blob/549dbf3390885bbca5ed6b112cc88f74d7663b73/core/dbt/cli/main.py#L123

Capture issues related to globalizing these flags

### Impact to Other Teams

N/A

No

jtcohen6:
resolves #7965
resolves #8715
throwback: #2740

### Problem

- Weird logical divergence between `clone` and other tasks in terms of how deferral works. (I borrowed some of @MichelleArk's spike work from #8339.)
- Some commands don't support `--defer` & `--state`, but they shouldn't raise an error. That's the case currently (e.g. if you submit `run-operation` within the IDE while "Defer to production" is enabled)

### Solution

- Consolidate how deferral works for all runnable tasks
- Support `--defer` & friends as "global" flags

### Checklist


tests/unit/test_manifest.py
```diff
@@ -1041,7 +1041,8 @@ def test_add_from_artifact(self):
 
         original_manifest = Manifest(nodes=original_nodes)
         other_manifest = Manifest(nodes=other_nodes)
-        original_manifest.add_from_artifact(other_manifest.writable_manifest())
+        adapter = mock.MagicMock()
+        original_manifest.merge_from_artifact(adapter, other_manifest.writable_manifest(), {})
```
MichelleArk, line 1045: nit: can rename test to `test_merge_from_artifact`

core/dbt/context/providers.py
```diff
@@ -1410,7 +1410,8 @@ def sql(self) -> Optional[str]:
         if self.model.language == ModelLanguage.sql:  # type: ignore[union-attr]
             # If the model is deferred and the adapter doesn't support zero-copy cloning, then select * from the prod
             # relation
-            if getattr(self.model, "defer_relation", None):
+            # TODO: avoid routing on args.which if possible
+            if getattr(self.model, "defer_relation", None) and self.config.args.which == "clone":
```
MichelleArk, line 1414: out of scope for this PR, but one way to do this could be to have the clone task further mutate the manifest after `merge_from_artifact` is called to overwrite `compiled_code` based on `defer_relation`. that way we could get rid of this if block entirely

core/dbt/context/providers.py
```diff
@@ -1410,7 +1410,8 @@ def sql(self) -> Optional[str]:
         if self.model.language == ModelLanguage.sql:  # type: ignore[union-attr]
             # If the model is deferred and the adapter doesn't support zero-copy cloning, then select * from the prod
             # relation
-            if getattr(self.model, "defer_relation", None):
+            # TODO: avoid routing on args.which if possible
```
MichelleArk, line 1413: out of scope for this PR, but one way to do this could be to have the clone task further mutate the manifest after merge_from_artifact is called to overwrite compiled_code based on defer_relation. that way we could get rid of this if block entirely

core/dbt/task/runnable.py
```diff
@@ -130,9 +130,22 @@ def get_selection_spec(self) -> SelectionSpec:
...
+                "Expected to defer to manifest, but there is no runtime manifest to defer from!"
+            )
+        self.manifest.merge_from_artifact(
+            adapter=adapter,
+            other=deferred_manifest,
+            selected=selected_uids,
+            favor_state=bool(self.args.favor_state),
+        )
+        # TODO: is it wrong to write the manifest here? I think it's right...
+        write_manifest(self.manifest, self.config.project_target_path)
```
MichelleArk, line 148: we may want to gate on self.config.args.write_json here, similar to what's done when the manifest is initially written here: https://github.com/dbt-labs/dbt-core/blob/main/core/dbt/cli/requires.py#L277

core/dbt/context/providers.py
jtcohen6: True. The risk(?) of that approach is, the `clone` task would produce a genuinely different (compiled) manifest from other tasks, and overwrite it at the end of the task [here](https://github.com/dbt-labs/dbt-core/blob/5488dfb9926c3f81e6a80bd0d8b01fe7f0ac7a60/core/dbt/task/runnable.py#L488-L491) with potentially surprising implications for metadata consumers. (I don't think there would be any effect for `state:modified`, which doesn't use `compiled_code` at all.)

That might still be preferable to the approach in this PR.

dbeatty10:
This covers 3 out of the 4 [UserConfig missing from global configs](https://github.com/dbt-labs/dbt-core/issues/7036) that showed up in a separate and independent audit.

What about `indirect_selection`? It is the only 1 of the 4 that showed up there but is not covered here.

core/dbt/context/providers.py
MichelleArk: I'm okay to leave this as a TODO and consider alternative approaches down the line to keep risk minimal. Removing the command-based check here also feels less satisfying if downstream metadata consumers will need to put something similar in place to preserve their expected functionality.

core/dbt/cli/main.py
```diff
@@ -121,11 +121,18 @@ def invoke(self, args: List[str], **kwargs) -> dbtRunnerResult:
...
     @p.debug
+    @p.defer
+    @p.deprecated_defer
+    @p.defer_state
+    @p.deprecated_favor_state
     @p.deprecated_print
+    @p.deprecated_state
     @p.enable_legacy_logger
     @p.fail_fast
+    @p.favor_state
```
MichelleArk, line 132: @jtcohen6, re: https://github.com/dbt-labs/dbt-core/pull/9040#issuecomment-1889709518

is this just a matter of adding indirect_selection here? took it for a spin locally and it spun

core/dbt/task/runnable.py
```diff
@@ -130,9 +130,22 @@ def get_selection_spec(self) -> SelectionSpec:
...
+            raise DbtInternalError(
+                "Expected to defer to manifest, but there is no runtime manifest to defer from!"
+            )
+        self.manifest.merge_from_artifact(
+            adapter=adapter,
+            other=deferred_manifest,
+            selected=selected_uids,
+            favor_state=bool(self.args.favor_state),
+        )
+        # TODO: is it wrong to write the manifest here? I think it's right...
```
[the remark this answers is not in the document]
MichelleArk: updated the comment

tests/unit/test_manifest.py
MichelleArk: updated

core/dbt/cli/main.py
dbeatty10: I'm guessing it would also involve removing `@p.indirect_selection` from each of the sub commands.

MichelleArk (approved):

core/dbt/task/seed.py
```diff
@@ -62,10 +62,6 @@ def print_result_line(self, result):
 
 
 class SeedTask(RunTask):
-    def defer_to_manifest(self, adapter, selected_uids):
```
jtcohen6, line 65: Discussed offline with @MichelleArk: Should we actually still disable deferral for tasks like seed/freshness/list, since those tasks never (in theory) resolve runtime references?

> It's technically possible to have a seed reference another node, via a pre/post hook. Or write custom code in a project (on-run-*) hook that wants to use the `defer_relation` property of nodes in the graph. I don't think that's _good_, but it is possible.
>
> At this point it feels like a choice between consistency, or a (small) performance optimization. Either way, I want all commands to support the flags as CLI options. We could keep the overrides in for now, until the follow-up refactor (https://github.com/dbt-labs/dbt-core/pull/9199) that removes deferral logic from tasks entirely, and makes it part of manifest loading. That does feel like the right direction, and this is a step on the way there.

PR #9054, opened 2023-11-10
Decouple adapter constraints from core

### LINKED ISSUE #9024: [CT-3334] Remove usage of dbt.contracts.graph.nodes.ConstraintType, ColumnLevelConstraint, ModelLevelConstraint in dbt/adapters
### Housekeeping

### Short description

`dbt.contracts.graph.nodes.ConstraintType` is used in dbt/adapters in:
* `BaseAdapter.CONSTRAINT_SUPPORT`
* `BaseAdapter.process_parsed_constraint`
* `PostgresAdapter.CONSTRSAINT_SUPPORT`
for the purposes of rendering python constraints objects to DDL during materializations. `ConstraintType` is also used in parsing and 

`ColumnLevelConstraint` and `ModelLevelConstraint` are used for similar purpose in `dbt/adapters`

Additionally, these classes are used in parsing and are part of the serialized manifest in core

Implementation options: 
* define these `dbt.contracts.graph.nodes` outside of core - either in `dbt.common.contracts` or `dbt.adapters.contracts` + update import paths
* keep the existing node classes in core, but define separate protocols or dataclasses in adapters for passing data between core and adapters - this will involve some duplication but may be beneficial if the adapters module does not need _everything_ functional in these classes provided by core, but just the data / schemas.

### Acceptance criteria

The following adapter interfaces no longer rely on `dbt.contracts.graph.nodes.ConstraintType`:
* `BaseAdapter.CONSTRAINT_SUPPORT`
* `BaseAdapter.process_parsed_constraint`
* `PostgresAdapter.CONSTRSAINT_SUPPORT`

### Impact to Other Teams

1p adapter imports will need to be updated, but those can be done separately as part of a bundled migration of 1p for https://github.com/dbt-labs/dbt-core/issues/8917 if the work is done on a feature branch: https://github.com/dbt-labs/dbt-core/pull/8906

No

### Context

_No response_

peterallenwebb:
resolves #9024

### Problem

Adapters code relied on constraint class definitions in core code

### Solution

I moved the classes into dbt.common

### Checklist


core/dbt/common/constraints.py
```diff
@@ -0,0 +1,43 @@
+from dataclasses import dataclass, field
```
colin-rogers-dbt, line 1: Should we consider constraints a 'contract'? If so we should move it to common/contracts

peterallenwebb: @MichelleArk What do you think about this?

MichelleArk: agree - common/contracts makes sense as a home for these based on the existing patterns we have in dbt & dbt/adapters

MichelleArk (approved):
LGTM! (after the file rename + changelog)

colin-rogers-dbt (dismissed):
@MichelleArk feel free to chime in if I'm off base but I think we've been trying to add changelogs as we go to preserve a record of everything that's changed, can you add one?

Issue #9015, opened 2023-11-06
[CT-3329] [Bug] Unit testing errors with long CTE name given reasonable test name length


MichelleArk:
### Is this a new bug in dbt-core?

- [X] I believe this is a new bug in dbt-core
- [X] I have searched the existing issues, and I could not find an existing issue for this bug

### Current Behavior

Long unit test names (in combination with long model name + input names) can lead to a "Relation name '{}' is longer than 63 characters" if the sum of lengths is > 63 (on postgres) characters.

### Expected Behavior

No error

### Steps To Reproduce

Changing the `test_my_model_incremental_yml` fixture from in `test_unit_testing.py`:
```
test_my_model_incremental_yml = """
unit_tests:
  - name: incremental_false
...
```
to: 
```
test_my_model_incremental_yml = """
unit_tests:
  - name: test_incremental_false
...
```

Results in: 
```
15:58:28  Completed with 1 error and 0 warnings:
15:58:28  
15:58:28    Runtime Error in unit_test my_incremental_model__test_incremental_false (models/test_my_incremental_model.yml)
  Relation name '__dbt__cte__my_incremental_model__test_incremental_false__events' is longer than 63 characters
15:58:28  
15:58:28  Done. PASS=1 WARN=0 ERROR=1 SKIP=0 TOTAL=2
```

### Environment

```markdown
- OS:
- Python:
- dbt:
```

### Additional Context

This is happening because the ephemeral fixture nodes are named based on the concatenation of: the model being tested, the name of the unit test, and the name of the input itself. This ensures they have globally unique names in the unit test manifest but is overkill for the purpose of naming the CTEs, which are local to a unit test case.

Options: 
Shorten the name of the ModelNode for input fixtures to only include the name of the original input node -- assuming that node names don't need to be unique within a DAG. - These aren't used in a DAG, so this shouldn't be a concern.
Consider hashing to add extra layer of caution to avoid name duplication.
These are ephemeral names only they are not executable.

### Testing Considerations:

- Test having 2 input models with same names and different packages.

jtcohen6:
We have some prior art for this:

https://github.com/dbt-labs/dbt-core/blob/bb35b3eb87ab2543247fd6ea51f6043bb89a1f89/plugins/postgres/dbt/include/postgres/macros/adapters.sql#L141-L178

graciegoheen:
[...]

Issue #8757, opened 2023-10-02
[CT-3176] [spike] infer fixture data types for unit testing


graciegoheen:
### Housekeeping

- [X] I am a maintainer of dbt-core

### Short description

From https://github.com/dbt-labs/dbt-core/issues/8649

Instead of requiring folks to run dbt run --select +my_model --exclude my_model before executing unit tests on my_model or requiring the use of --deferral (which would cause issues in a CI context when you're deferring to a production environment and may have changes to your inputs as part of the PR)...

we should be able to run dbt unit-test --select my_model in a new environment without having previously run dbt run --select +my_model --exclude my_model or dbt docs generate

- we will use the data_types defined in the direct parent's yml config
- if no data_type is provided...
    - we will have the warehouse "presume" the data type based on the mock data provided (users could put cast statements in their fixture)
    - if no mock data is provided for a column, we will guess that the data type is a string
raise a warning during unit test execution if an input fixture specifies columns that don’t have user-provided data_type values.

We will need to confirm when we should add '' around the mocked data.

We can get all column names from `adapter.get_column_schema_from_query` so users don't have to provide a yml name for all columns of direct parents.

### Acceptance criteria

- I can run `dbt unit-test --select my_model` in a new environment without having previously run `dbt run --select +my_model --exclude my_model` or `dbt docs generate`
- edit test to remove the build prior to executing unit tests https://github.com/dbt-labs/dbt-core/blob/unit_testing_feature_branch/tests/functional/unit_testing/test_unit_testing.py#L123-L124

### Impact to Other Teams

No

### Will backports be required?

No

graciegoheen:
Let's make sure we get feedback on this during the beta!

graciegoheen:
The goal of this spike is to determine an appropriate path forward - additional ideas tracked here https://github.com/dbt-labs/dbt-core/issues/8649

graciegoheen:
Opened - https://github.com/dbt-labs/dbt-core/issues/8979 and https://github.com/dbt-labs/dbt-core/issues/8980

graciegoheen:
Our assumption:

> We can get all column names from adapter.get_column_schema_from_query so users don't have to provide a yml name for all columns of direct parents.

was incorrect. So closing in favor of:
- use `--dry-run` to build "empty" versions of models
- `dbt build` includes unit tests to materialize upstream nodes before running unit tests

PR #9059, opened 2023-11-12
Allow use of sources as unit testing inputs

### LINKED ISSUE #8507: [CT-3038] Support input sources, create test cases
### Description

Currently unit tests can take inputs from models: `ref('my_model')`. We also want to allow sources as inputs: `source('source_dataset', 'source_name')`.  The big difference between these two things is that dbt builds the relations for models, whereas sources already exist as relations, outside of dbt.

Do we want to use the source as-is?  This is probably simplest. It might also be possible to construct an ephemeral model which only contains the columns to be tests.

It's possible that this will require different jinja implementation than the current ephemeral model implementation of input models.

### Acceptance Criteria

A unit test can reference a source node and "just work".

### Impact to other teams
                  
None                 
                                
No

gshank:
resolves #8507

### Problem

We want to support the use of sources as inputs in unit test cases.

### Solution

Created a UnitTestSourceDefinition object, which acts as a source for purposes of resolving "source" calls, but acts as a model for purpose of executing the test case.

### Checklist

### DIFF
--- a/core/dbt/adapters/base/relation.py
+++ b/core/dbt/adapters/base/relation.py
@@ -2,7 +2,13 @@ from collections.abc import Hashable
 from dataclasses import dataclass, field
 from typing import Optional, TypeVar, Any, Type, Dict, Iterator, Tuple, Set, Union, FrozenSet
 
-from dbt.contracts.graph.nodes import SourceDefinition, ManifestNode, ResultNode, ParsedNode
+from dbt.contracts.graph.nodes import (
+    SourceDefinition,
+    ManifestNode,
+    ResultNode,
+    ParsedNode,
+    UnitTestSourceDefinition,
+)
 from dbt.contracts.relation import (
     RelationType,
     ComponentName,
@@ -201,7 +207,9 @@ class BaseRelation(FakeAPIObject, Hashable):
         )
 
     @classmethod
-    def create_from_source(cls: Type[Self], source: SourceDefinition, **kwargs: Any) -> Self:
+    def create_from_source(
+        cls: Type[Self], source: Union[SourceDefinition, UnitTestSourceDefinition], **kwargs: Any
+    ) -> Self:
         source_quoting = source.quoting.to_dict(omit_none=True)
         source_quoting.pop("column", None)
         quote_policy = deep_merge(
@@ -263,8 +271,10 @@ class BaseRelation(FakeAPIObject, Hashable):
         node: ResultNode,
         **kwargs: Any,
     ) -> Self:
-        if node.resource_type == NodeType.Source:
-            if not isinstance(node, SourceDefinition):
+        if node.resource_type == NodeType.Source or isinstance(node, UnitTestSourceDefinition):
+            if not (
+                isinstance(node, SourceDefinition) or isinstance(node, UnitTestSourceDefinition)
+            ):
                 raise DbtInternalError(
                     "type mismatch, expected SourceDefinition but got {}".format(type(node))
                 )
--- a/core/dbt/context/providers.py
+++ b/core/dbt/context/providers.py
@@ -606,6 +606,34 @@ class RuntimeSourceResolver(BaseSourceResolver):
         return self.Relation.create_from_source(target_source)
 
 
+class RuntimeUnitTestSourceResolver(RuntimeSourceResolver):
+    def resolve(self, source_name: str, table_name: str):
+        target_source = self.manifest.resolve_source(
+            source_name,
+            table_name,
+            self.current_project,
+            self.model.package_name,
+        )
+        if target_source is None or isinstance(target_source, Disabled):
+            raise TargetNotFoundError(
+                node=self.model,
+                target_name=f"{source_name}.{table_name}",
+                target_kind="source",
+                disabled=(isinstance(target_source, Disabled)),
+            )
+        # For unit tests, this isn't a "real" source, it's a ModelNode taking
+        # the place of a source. We want to get the relation of the fake model
+        # node, not the original source relation.
+        return self.create_relation(target_source)
+
+    def create_relation(self, target_source: ManifestNode) -> RelationProxy:
+        if target_source.is_ephemeral_model:
+            self.model.set_cte(target_source.unique_id, None)
+            return self.Relation.create_ephemeral_from_node(self.config, target_source)
+        else:
+            return self.Relation.create_from(self.config, target_source)
+
+
 # metric` implementations
 class ParseMetricResolver(BaseMetricResolver):
     def resolve(self, name: str, package: Optional[str] = None) -> MetricReference:
@@ -746,7 +774,7 @@ class RuntimeUnitTestProvider(Provider):
     DatabaseWrapper = RuntimeDatabaseWrapper
     Var = UnitTestVar
     ref = RuntimeUnitTestRefResolver
-    source = RuntimeSourceResolver  # TODO: RuntimeUnitTestSourceResolver
+    source = RuntimeUnitTestSourceResolver
     metric = RuntimeMetricResolver
 
 
--- a/core/dbt/context/providers.py
+++ b/core/dbt/context/providers.py
@@ -622,16 +622,11 @@ class RuntimeUnitTestSourceResolver(RuntimeSourceResolver):
                 disabled=(isinstance(target_source, Disabled)),
             )
         # For unit tests, this isn't a "real" source, it's a ModelNode taking
-        # the place of a source. We want to get the relation of the fake model
-        # node, not the original source relation.
-        return self.create_relation(target_source)
-
-    def create_relation(self, target_source: ManifestNode) -> RelationProxy:
-        if target_source.is_ephemeral_model:
-            self.model.set_cte(target_source.unique_id, None)
-            return self.Relation.create_ephemeral_from_node(self.config, target_source)
-        else:
-            return self.Relation.create_from(self.config, target_source)
+        # the place of a source. We don't really need to return the relation here,
+        # we just need to set_cte, but skipping it confuses typing. We *do* need
+        # the relation in the "this" property.
+        self.model.set_cte(target_source.unique_id, None)
+        return self.Relation.create_ephemeral_from_node(self.config, target_source)
 
 
 # metric` implementations
--- a/core/dbt/context/providers.py
+++ b/core/dbt/context/providers.py
@@ -606,7 +606,7 @@ class RuntimeSourceResolver(BaseSourceResolver):
         return self.Relation.create_from_source(target_source)
 
 
-class RuntimeUnitTestSourceResolver(RuntimeSourceResolver):
+class RuntimeUnitTestSourceResolver(BaseSourceResolver):
     def resolve(self, source_name: str, table_name: str):
         target_source = self.manifest.resolve_source(
             source_name,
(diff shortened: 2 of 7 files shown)


core/dbt/adapters/base/relation.py
```diff
@@ -263,8 +271,10 @@ def create_from(
         node: ResultNode,
         **kwargs: Any,
     ) -> Self:
-        if node.resource_type == NodeType.Source:
-            if not isinstance(node, SourceDefinition):
+        if node.resource_type == NodeType.Source or isinstance(node, UnitTestSourceDefinition):
```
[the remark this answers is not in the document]
gshank: How?

core/dbt/parser/unit_tests.py
```diff
@@ -187,10 +210,13 @@ def _get_original_input_node(self, input: str, tested_node: ModelNode):
                         name, package, version, self.manifest
                     )
             elif statically_parsed["sources"]:
-                input_package_name, input_source_name = statically_parsed["sources"][0]
-                original_input_node = self.manifest.source_lookup.find(
-                    input_source_name, input_package_name, self.manifest
-                )
+                for source in statically_parsed["sources"]:
```
MichelleArk, line 213: would we expect to get more than one result back from `statically_parsed["sources"]`? If so this loop could lead to surprising behaviour because original_input_node is overwritten on each iteration.

I'd expect that we should only ever get a single source back from this parsing because the input looks something like `source('test', 'test_name')`, in which case `source = statically_parsed["sources"][0]` would make more sense here conceptually

gshank: That doesn't work because statically_parsed["sources"] is a set. So I did it the same way that the ref call above is handled. We could convert the set to a list and index it though, which might be better.

MichelleArk: ah, I see! yeah I think that's pretty subtle and converting to a list + grabbing the first (& only!) element would be more readable here

core/dbt/parser/unit_tests.py
```diff
@@ -105,44 +106,66 @@ def parse_unit_test_case(self, test_case: UnitTestDefinition):
...
             original_input_node = self._get_original_input_node(given.input, tested_node)
 
-            original_input_node_columns = None
-            if (
-                original_input_node.resource_type == NodeType.Model
-                and original_input_node.config.contract.enforced
-            ):
-                original_input_node_columns = {
-                    column.name: column.data_type for column in original_input_node.columns
-                }
```
MichelleArk, line 121: thanks for cleaning this up! we're not going down this route any longer 🔥

core/dbt/parser/unit_tests.py
```diff
@@ -105,44 +106,66 @@ def parse_unit_test_case(self, test_case: UnitTestDefinition):
...
+                # but that shouldn't matter since this acts as an ephemeral model which just
+                # wraps a CTE around the unit test node.
+                input_node = UnitTestSourceDefinition(
+                    **common_fields,
+                    name=original_input_node.name,  # must be the same name for source lookup to work
+                    path=input_name + ".sql",  # for writing out compiled_code
+                    raw_code=self._build_fixture_raw_code(
+                        given.get_rows(
+                            self.root_project.project_root, self.root_project.fixture_paths
+                        ),
```
MichelleArk, line 160: Comparing this to setting [raw_code above](https://github.com/dbt-labs/dbt-core/pull/9059/files#diff-4164ae2da9549b426a9e1a37f2d15509fef9bed35de3139273a506c0ba8c8f5cR141-R144), I believe raw_code could be moved to common_fields

core/dbt/parser/unit_tests.py
```diff
@@ -105,44 +106,66 @@ def parse_unit_test_case(self, test_case: UnitTestDefinition):
...
+                "fqn": original_input_node.fqn,
+                "checksum": FileHash.empty(),
+            }
+
+            if original_input_node.resource_type == NodeType.Model:
+                original_input_node_columns = None
+                if original_input_node.config.contract.enforced:
+                    original_input_node_columns = {
+                        column.name: column.data_type for column in original_input_node.columns
+                    }
```
MichelleArk, line 136: ah, just seeing this. I guess we're not just removing it as above here: https://github.com/dbt-labs/dbt-core/pull/9059/files#r1391606187

I think at this point with what we've landed on from https://github.com/dbt-labs/dbt-core/issues/8757#issuecomment-1796297194, we could actually remove this attempt to infer columns from here (perhaps doing the cleanup in macros in a follow-on issue). That way we can consolidate `raw_code` to `common_fields` as well.

core/dbt/parser/unit_tests.py
gshank: Done.

gshank:
I looked at taking out the special casing of UnitTestSourceDefinition, but unfortunately there are subtle differences in the specification of quoting between sources and models, and so I think it's best to actually use the relation.create_from_source to get the quoting right.

core/dbt/parser/unit_tests.py
```diff
@@ -105,44 +106,52 @@ def parse_unit_test_case(self, test_case: UnitTestDefinition):
...
+                # but that shouldn't matter since this acts as an ephemeral model which just
+                # wraps a CTE around the unit test node.
+                input_node = UnitTestSourceDefinition(
+                    **common_fields,
+                    name=original_input_node.name,  # must be the same name for source lookup to work
+                    path=input_name + ".sql",  # for writing out compiled_code
+                    source_name=original_input_node.source_name,  # needed for source lookup
+                )
+                # Sources need to go in the sources dictionary in order to create the right lookup
+                self.unit_test_manifest.sources[input_node.unique_id] = input_node  # type: ignore
```
MichelleArk, line 152: do we anticipate any issues by having the sources dictionary contain a unique_id key that is prefixed with `model` instead of `source` here?

core/dbt/parser/unit_tests.py
```diff
@@ -105,44 +106,52 @@ def parse_unit_test_case(self, test_case: UnitTestDefinition):
...
+
+            # Both ModelNode and UnitTestSourceDefinition need to go in nodes dictionary
```
MichelleArk, line 154: for my own understanding - is this because to enable cte injection?

core/dbt/context/providers.py
```diff
@@ -606,6 +606,29 @@ def resolve(self, source_name: str, table_name: str):
         return self.Relation.create_from_source(target_source)
 
 
+class RuntimeUnitTestSourceResolver(RuntimeSourceResolver):
```
MichelleArk, line 609: I think we're overriding the only method `RuntimeSourceResolver` provides & not using super(), so it may make more sense to subclass `BaseSourceResolver` here.

core/dbt/parser/unit_tests.py
```diff
@@ -105,44 +106,52 @@ def parse_unit_test_case(self, test_case: UnitTestDefinition):
...
-                    given.get_rows(
-                        self.root_project.project_root, self.root_project.fixture_paths
-                    ),
-                    original_input_node_columns,
+            project_root = self.root_project.project_root
+            common_fields = {
+                "resource_type": NodeType.Model,
+                "package_name": package_name,
+                "original_file_path": original_input_node.original_file_path,
+                "unique_id": f"model.{package_name}.{input_name}",
```
MichelleArk, line 123: I think we may need to include `source_name` in `input_name` to avoid clobbering sources with the same table_name but different source_names when they are inserted to `manifest.nodes` and `manifest.sources` below.

core/dbt/parser/unit_tests.py
```diff
@@ -178,18 +189,21 @@ def _get_original_input_node(self, input: str, tested_node: ModelNode):
...
+                    name, package, version, self.manifest
+                )
             elif statically_parsed["sources"]:
-                input_package_name, input_source_name = statically_parsed["sources"][0]
+                source = list(statically_parsed["sources"])[0]
+                input_source_name, input_name = source
                 original_input_node = self.manifest.source_lookup.find(
-                    input_source_name, input_package_name, self.manifest
+                    f"{input_source_name}.{input_name}",
+                    self.root_project.project_name,
```
MichelleArk, line 205: I believe it should be possible to write a unit test where a source input is coming from an imported package (cc @graciegoheen to confirm)

Does this setting restrict that capability? I think it should be `None` to find a source in any package

graciegoheen: Yes, I think we should make it possible to add a unit test to a model that depends on a source defined in a package.

core/dbt/parser/unit_tests.py
gshank: Yeah. There's code in compilation.py that looks up the existence of the cte in the nodes dictionary: `if cte.id not in manifest.nodes:`.

gshank: In theory we could also check for a UnitTestSourceDefinition and in sources, but that didn't feel like an improvement.

core/dbt/parser/unit_tests.py
gshank: It doesn't seem to care. We don't actually check the unique_id prefix that I can recall. If somebody starts parsing the unit_test_manifest, I suppose it might be confusing.  But right now we're putting it in two places, so one of them will be wrong.

core/dbt/parser/unit_tests.py
gshank: Good point.

core/dbt/parser/unit_tests.py
MichelleArk: This probably isn't worth spending tons of time on.. but I think it could be possible to get around having to add the node to manifest.sources and do the lookup from the .nodes collection in UnitTestRuntimeSourceResolver since the unique_id will include source_name. kind of like what's done here: https://github.com/dbt-labs/dbt-core/blob/unit_testing_feature_branch/core/dbt/context/providers.py#L578

Not entirely sure what's more readable or less complex in this case. I can imagine having to maintain UnitTestSourceDefinitions across both dictionaries could be error-prone though..

core/dbt/parser/unit_tests.py
gshank: Will change package to None.

core/dbt/context/providers.py
gshank: Done.

core/dbt/parser/unit_tests.py
gshank: I've changed this to include the source_name. This does make for pretty long unique_ids.  Do we have any concerns about that?  It's not like we're using that name to construct tables or anything...

core/dbt/parser/unit_tests.py
gshank: Done.

core/dbt/parser/unit_tests.py
gshank: I think the lookup behavior of sources and nodes is subtly different with regard to the meaning of package=None, so I don't think looking up sources as though they were nodes is worth it.

core/dbt/parser/unit_tests.py
MichelleArk: So far I've noticed this issue creep up in https://github.com/dbt-labs/dbt-core/issues/9015. I think we could shorten the node name for CTE generation (since it doesn't need to be unique) but keep the unique_id longer

MichelleArk (approved):

PR #9064, opened 2023-11-13
Fix #8652: Use seed file from disk for unit testing if rows not specified in YAML config

### LINKED ISSUE #8652: Handle input seeds in unit tests

### Describe the feature

Our proposed [unit testing spec](https://github.com/dbt-labs/dbt-core/discussions/8275), doesn't yet take into account if one of the inputs to the model you want to add a unit test to is a `seed`. 

Let's say `model_a` depends on `my_seed`. If you want to add a unit test to `model_a` - do you need to define an input for `my_seed`? That feels redundant since seeds don't often change and are already small data inputs.

How should we handle inputs for seeds?

_No response_

### Who will this benefit?

_No response_

_No response_

### Anything else?

_No response_

aranke:
Resolves #8652 

### Problem

Seed values must be specified for unit testing.

### Solution

Read seed from disk if no rows are specified in YAML config, instead of throwing an error.

### Checklist

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -1,6 +1,6 @@
 black==23.3.0
 bumpversion
-ddtrace
+ddtrace==2.1.7
 docutils
 flake8
 flaky


tests/functional/unit_testing/test_unit_testing.py
```diff
@@ -100,3 +100,139 @@ def test_basic(self, project):
...
+        return {
+            "my_new_model.sql": my_new_model,
+            "my_favorite_model.sql": my_favorite_model,
+            "schema.yml": schema_yml_implicit_seed,
+        }
+
+    def test_basic(self, project):
+        run_dbt(["seed"])
+        run_dbt(["run"])
+        # assert len(results) == 1
```
MichelleArk, line 212: ditto

core/dbt/parser/unit_tests.py
```diff
@@ -203,6 +206,22 @@ def __init__(self, schema_parser: SchemaParser, yaml: YamlBlock) -> None:
         self.schema_parser = schema_parser
         self.yaml = yaml
 
+    def load_rows_from_seed(self, seed_name):
```
MichelleArk, line 209: missing type annotations, a docstring would be helpful here too

core/dbt/parser/unit_tests.py
```diff
@@ -203,6 +206,22 @@ def __init__(self, schema_parser: SchemaParser, yaml: YamlBlock) -> None:
...
+
+        try:
+            seed_node = self.manifest.ref_lookup.perform_lookup(
+                f"seed.{self.project.project_name}.{seed_name}", self.manifest
+            )
+            seed_path = Path(seed_node.root_path) / seed_node.original_file_path
+            with open(seed_path, "r") as f:
+                for row in DictReader(f):
+                    rows.append(row)
+        except DbtInternalError:
```
MichelleArk, line 220: can we catch this further up - after `self.manifest.ref_lookup`? that way the error handling remains localized and precise

core/dbt/parser/unit_tests.py
```diff
@@ -203,6 +206,22 @@ def __init__(self, schema_parser: SchemaParser, yaml: YamlBlock) -> None:
         self.schema_parser = schema_parser
         self.yaml = yaml
 
+    def load_rows_from_seed(self, seed_name):
+        rows = []
+
+        try:
+            seed_node = self.manifest.ref_lookup.perform_lookup(
+                f"seed.{self.project.project_name}.{seed_name}", self.manifest
```
MichelleArk, line 214: manifest.ref_lookup.find can be used to avoid having to reconstruct the seed unique id like this. It returns None instead of raises an error, so the handling downstream would change a bit.

tests/functional/unit_testing/test_unit_testing.py
```diff
@@ -100,3 +100,139 @@ def test_basic(self, project):
...
+            "my_favorite_model.sql": my_favorite_model,
+            "schema.yml": schema_yml_nonexistent_seed,
+        }
+
+    def test_basic(self, project):
+        run_dbt(["seed"])
+        run_dbt(["run"])
+
+        # Select by model name
+        with pytest.raises(AttributeError):
```
MichelleArk, line 237: would `AttributeError` make its way to an end-user facing error?

core/dbt/parser/unit_tests.py
```diff
@@ -203,6 +206,30 @@ def __init__(self, schema_parser: SchemaParser, yaml: YamlBlock) -> None:
...
 
+    def _get_seed_name_from_ref(self, ref: str) -> str:
+        """Extract seed name from ref string."""
+        return py_extract_from_source("{{ " + ref + " }}")["refs"][0]["name"]
+
+    def _load_rows_from_seed(self, seed_name: str) -> List[Dict[str, Any]]:
+        """Read rows from seed file on disk if not specified in YAML config. If seed file doesn't exist, return empty list."""
+        rows: List[Dict[str, Any]] = []
+
+        package_name = self.project.project_name
```
MichelleArk, line 217: I think we should be accepting an optional[str] package_name to this method from extraction in _get_seed_name_from_ref, and defaulting to self.project.project_name if not provided.

core/dbt/parser/unit_tests.py
```diff
@@ -203,6 +206,30 @@ def __init__(self, schema_parser: SchemaParser, yaml: YamlBlock) -> None:
...
+
+        seed_node = self.manifest.ref_lookup.find(seed_name, package_name, None, self.manifest)
+
+        if not seed_node:
+            raise ParsingError(
+                f"Unable to find seed '{package_name}.{seed_name}' for unit tests in directories: {self.project.seed_paths}"
```
MichelleArk, line 223: this may be misleading because the seed (if found) could come from seed directories from packages as well (e.g. seed_nodes.original_file_path)

core/dbt/parser/unit_tests.py
MichelleArk: related to this earlier comment: https://github.com/dbt-labs/dbt-core/pull/9064#discussion_r1392900244

core/dbt/parser/unit_tests.py
```diff
@@ -203,6 +206,30 @@ def __init__(self, schema_parser: SchemaParser, yaml: YamlBlock) -> None:
...
+
+    def _load_rows_from_seed(self, seed_name: str) -> List[Dict[str, Any]]:
+        """Read rows from seed file on disk if not specified in YAML config. If seed file doesn't exist, return empty list."""
+        rows: List[Dict[str, Any]] = []
+
+        package_name = self.project.project_name
+
+        seed_node = self.manifest.ref_lookup.find(seed_name, package_name, None, self.manifest)
+
+        if not seed_node:
```
MichelleArk, line 221: could be expanded to `if not seed_node or seed_node.resource_type != ResourceType.Seed` to ensure the node is a seed node in downstream processing.

I think this is related to the failing schema test as well

core/dbt/parser/unit_tests.py
```diff
@@ -219,6 +222,33 @@ def __init__(self, schema_parser: SchemaParser, yaml: YamlBlock) -> None:
         self.schema_parser = schema_parser
         self.yaml = yaml
 
+    def _load_rows_from_seed(self, ref: Dict[str, str]) -> List[Dict[str, Any]]:
```
MichelleArk, line 225: at this point the structure of `ref` feels fairly specific - the caller will have to do some preprocessing to make use of the method correctly. I would just roll the `py_extract_from_source` call directly in this method and accept the raw input `ref_str` instead.

core/dbt/parser/unit_tests.py
```diff
@@ -219,6 +222,33 @@ def __init__(self, schema_parser: SchemaParser, yaml: YamlBlock) -> None:
...
+
+        if not seed_node or seed_node.resource_type != NodeType.Seed:
+            # Seed not found in custom package specified
+            if package_name != self.project.project_name:
+                raise ParsingError(
+                    f"Unable to find seed '{package_name}.{seed_name}' for unit tests in '{package_name}' package"
+                )
+            else:
+                raise ParsingError(
+                    f"Unable to find seed '{package_name}.{seed_name}' for unit tests in directories: {self.project.seed_paths}"
```
MichelleArk, line 242: nice! appreciate the level of clarity in these error messages depending on the case. What do you think about refactoring this into its own error that subclasses ParsingError to keep this method more succinct?

PR #8950, opened 2023-10-31
DSI 0.4.0 and Saved Query Exports

QMalcolm:
resolves #8892 

### Problem
dbt-semantic-interfaces recently released 0.4.0. We want to support this for 1.7.0 of dbt-core. dbt-semantic-interfaces 0.4.0 has a few changes from 0.3.0

- `where`, `group_by`, and `metrics` are now nested under the `query_params` property on a `SavedQuery`
- The `SavedQuery` now has a concept of `exports`

### Solution

Begin supporting the changes. 

### Other things to flag
In 0.4.0 the names of metrics begin being validated by dbt-semantic-interfaces. This was always supposed to be happening, but wasn't 😬 This might break for some people. To be clear it was likely breaking in the semantic layer anyways. This just ensures that the issue is caught at parse time rather than query time.

### Checklist

- [X] I have read [the contributing guide](https://github.com/dbt-labs/dbt-core/blob/main/CONTRIBUTING.md) and understand what's expected of me  
- [X] I have run this code in development and it appears to resolve the stated issue  
- [X] This PR includes tests, or tests are not required/relevant for this PR
- [X] This PR has no interface changes (e.g. macros, cli, logs, json artifacts, config files, adapter interface, etc) or this PR has already received feedback and approval from Product or DX
- [X] This PR includes [type annotations](https://docs.python.org/3/library/typing.html) for new and modified functions

Issue #8892, opened 2023-10-25
[CT-3255] [implementation] interface changes to dbt-core saved_query spec to add export


graciegoheen:
### Housekeeping

- [X] I am a maintainer of dbt-core

### Short description

Add exports to saved_query

```
saved_queries:
  - name: my_query
    description:
    query:
      metrics:
        - 
      group_bys:
        - TimeDimension()
        - ...
      where:
        - 
    exports:
      - name: my_query 
        config:
          export_as: table # options: table, view - FUTURE: incremental_table, window_table, file
          schema: my_schema  # [optional - DEFAULT to deployment schema]
          alias: some_table_name # [optional - DEFAULT to export name]
```

### Acceptance criteria

- add exports to saved_query

### Impact to Other Teams

SL

### Will backports be required?

1.7 latest

QMalcolm:
The hierarchical config calculation part of this has been pushed to https://github.com/dbt-labs/dbt-core/issues/8956

PR #9065, opened 2023-11-13
Support hierarchical config setting for SavedQueryExport configs

### LINKED ISSUE #8956: [CT-3299] [SL] Support Hierarchical Configs for Saved Queries and their Exports
### Housekeeping

### Short description

In #8950 we added `Exports` to `SavedQueries`. `Exports` have configs and we want these configs to be setable from a project's `dbt_project.yml`, i.e. hierarchical. Originally this work was to be part of #8892 in #8950, however due to deadlines and some pesky edge cases, we separated out this work. With that said, we've thought through (and tried some) approaches

1. Write an entire bespoke config compiler just for `Exports`
2. Make an `Export` a node type (but don't put it in the DAG), and then use [calculate_node_config](https://github.com/dbt-labs/dbt-core/blob/bb21403c9e906097c4913edcf7a013d404e93d80/core/dbt/context/context_config.py#L134) to generate the config
3. Abstract pretty much all of `calculate_node_config` except the config class getter into a separate function which would then be called by `calculate_node_config` and directly from the export parsing

We haven't attempted option (1) because it seemed wrong to write a whole bunch of bespoke code for this. Options (2) and (3) suffer from an issue where-in the `ExportConfig` has a `schema_name` attr, but the `UnparsedExportConfig` uses `schema` (this is necessary for other reasons). Additionally option (2) would make an export a pseudo node, which we don't want to do unintentionally.

As it stands it seems like there are two paths forward: option (3) but use the `UnrenderedExportConfig` during the config generation process and then after the fact map things to the `ExportConfig` or a forth option would be to refactor our config parsing process to be less clever.

### Acceptance criteria

We have implemented a way for `Export` configs to be set hierarchically AND it is tested

### Impact to Other Teams

Semantic Layer, maybe  also Cloud Artifacts?

Yes 1.7

### Context

_No response_

QMalcolm:
resolves #8956 

### Problem
In https://github.com/dbt-labs/dbt-core/pull/8950 we added Exports to SavedQueries. Exports have configs and we want these configs to be setable from a project's dbt_project.yml, i.e. hierarchical. Originally this work was to be part of https://github.com/dbt-labs/dbt-core/issues/8892 in https://github.com/dbt-labs/dbt-core/pull/8950, however due to deadlines and some pesky edge cases, we separated out this work.

### Solution

In #8956 we outlined some paths forward, and we ended up taking none of them. We did not make `ExportConfig` a typical `BaseConfig` like all other configs, and we didn't use the `calculate_node_config` nor create any abstraction of it. Instead we decided to continue treating `ExportConfig` like a non-config object. An `ExportConfig` is mapped from an `Export.config` dictionary. We've taken the extra step to merge in any relevant `SavedQueryConfig` attributes with a preference for attributes defined directly on the `ExportConfig`. Additionally because the `SavedQueryConfig` is produced from the `calculate_node_config` process, we also get relevant configs in the dbt_project.yml.

### Checklist

### DIFF
--- a/core/dbt/contracts/graph/model_config.py
+++ b/core/dbt/contracts/graph/model_config.py
@@ -15,6 +15,7 @@ from dbt.contracts.util import Replaceable, list_str
 from dbt.exceptions import DbtInternalError, CompilationError
 from dbt import hooks
 from dbt.node_types import NodeType, AccessType
+from dbt_semantic_interfaces.type_enums.export_destination_type import ExportDestinationType
 from mashumaro.jsonschema.annotations import Pattern
 
 
@@ -407,6 +408,8 @@ class SavedQueryConfig(BaseConfig):
         default_factory=dict,
         metadata=MergeBehavior.Update.meta(),
     )
+    export_as: Optional[ExportDestinationType] = None
+    schema: Optional[str] = None
 
 
 @dataclass
--- a/core/dbt/parser/schema_yaml_readers.py
+++ b/core/dbt/parser/schema_yaml_readers.py
@@ -1,3 +1,4 @@
+from copy import deepcopy
 from dbt.parser.schemas import YamlReader, SchemaParser, ParseResult
 from dbt.parser.common import YamlBlock
 from dbt.node_types import NodeType
@@ -6,7 +7,6 @@ from dbt.contracts.graph.unparsed import (
     UnparsedDimensionTypeParams,
     UnparsedEntity,
     UnparsedExport,
-    UnparsedExportConfig,
     UnparsedExposure,
     UnparsedGroup,
     UnparsedMeasure,
@@ -57,7 +57,7 @@ from dbt_semantic_interfaces.type_enums import (
     MetricType,
     TimeGranularity,
 )
-from typing import List, Optional, Union
+from typing import Any, Dict, List, Optional, Union
 
 
 def parse_where_filter(
@@ -669,12 +669,11 @@ class SavedQueryParser(YamlReader):
 
         return config
 
-    def _get_export_config(self, unparsed: UnparsedExportConfig) -> ExportConfig:
-        return ExportConfig(
-            export_as=unparsed.export_as,
-            schema_name=unparsed.schema,
-            alias=unparsed.alias,
-        )
+    def _get_export_config(self, unparsed: Dict[str, Any]) -> ExportConfig:
+        parsed = deepcopy(unparsed)
+        if parsed.get("schema") is not parsed.get("schema_name") is None:
+            parsed["schema_name"] = parsed["schema"]
+        return ExportConfig.from_dict(parsed)
 
     def _get_export(self, unparsed: UnparsedExport) -> Export:
         return Export(name=unparsed.name, config=self._get_export_config(unparsed.config))
--- a/core/dbt/parser/schema_yaml_readers.py
+++ b/core/dbt/parser/schema_yaml_readers.py
@@ -670,11 +670,11 @@ class SavedQueryParser(YamlReader):
         return config
 
     def _get_export_config(
-        self, unparsed: Dict[str, Any], saved_query_config: SavedQueryConfig
+        self, unparsed_export_config: Dict[str, Any], saved_query_config: SavedQueryConfig
     ) -> ExportConfig:
         # Combine the two dictionaries using dictionary unpacking
         # the second dictionary is the one whose keys take priority
-        combined = {**saved_query_config.__dict__, **unparsed}
+        combined = {**saved_query_config.__dict__, **unparsed_export_config}
         # `schema` is the user facing attribute, but for DSI protocol purposes we track it as `schema_name`
         if combined.get("schema") is not None and combined.get("schema_name") is None:
             combined["schema_name"] = combined["schema"]
--- a/core/dbt/parser/schema_yaml_readers.py
+++ b/core/dbt/parser/schema_yaml_readers.py
@@ -1,4 +1,3 @@
-from copy import deepcopy
 from dbt.parser.schemas import YamlReader, SchemaParser, ParseResult
 from dbt.parser.common import YamlBlock
 from dbt.node_types import NodeType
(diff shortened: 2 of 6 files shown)


core/dbt/contracts/graph/model_config.py
```diff
@@ -407,6 +408,8 @@ class SavedQueryConfig(BaseConfig):
         default_factory=dict,
         metadata=MergeBehavior.Update.meta(),
     )
+    export_as: Optional[ExportDestinationType] = None
+    schema: Optional[str] = None
```
MichelleArk, line 412: are we dropping support for [alias](https://github.com/dbt-labs/dbt-core/pull/9065/files#diff-42f07118add483c2e34f2e1b77388819659667339ab74066353882b4adfed07eL738)?

core/dbt/parser/schema_yaml_readers.py
```diff
@@ -669,16 +669,25 @@ def _generate_saved_query_config(
...
-            schema_name=unparsed.schema,
-            alias=unparsed.alias,
+    def _get_export_config(
+        self, unparsed: Dict[str, Any], saved_query_config: SavedQueryConfig
+    ) -> ExportConfig:
+        # Combine the two dictionaries using dictionary unpacking
+        # the second dictionary is the one whose keys take priority
+        combined = {**saved_query_config.__dict__, **unparsed}
+        # `schema` is the user facing attribute, but for DSI protocol purposes we track it as `schema_name`
+        if combined.get("schema") is not combined.get("schema_name") is None:
```
MichelleArk, line 679: I'm finding this predicate a bit tricky to read: `A is not B is None`? is something missing after `is not`?

QMalcolm: That is indeed a bit wonky, I think something got shortened 😬 Will revisit, and maybe write an additional test

core/dbt/contracts/graph/model_config.py
QMalcolm: I intentionally left out alias because including it felt weird 🙃 If a saved query has multiple exports, I don't think we'd want them to all get aliased to to same thing. 

I didn't feel there was a great place to leave a comment about it (feels weird to comment in a class about why it doesn't have an attribute), so I put the informartion in the commit message
> We didn't add the `ExportConfig` `alias` property to the `SavedQueryConfig`. This
is because `alias` will always be specific to a single export, and thus it doesn't
make sense to allow defining it on the `SavedQueryConfig` to then apply to all
`Exports` belonging to the `SavedQuery`

Happy to add a comment to the class if that feels like the correct approach. Alternatively, if we do want to add alias we can, although I do worry that could cause other problems.

core/dbt/parser/schema_yaml_readers.py
QMalcolm: Feels like I when I write a sentence, and then edit it multiple times, and then it becomes incoherent 🙃

QMalcolm: Fixed in [dd3be16](https://github.com/dbt-labs/dbt-core/pull/9065/commits/dd3be1640c3ecd65598ec8caa356c5580df61221)

core/dbt/contracts/graph/model_config.py
MichelleArk: Makes sense. Have we documented `alias` in user documentation yet? If so it might make sense to support some no-op backwards-compatible interface here given we are backporting this change.

core/dbt/parser/schema_yaml_readers.py
```diff
@@ -669,16 +669,25 @@ def _generate_saved_query_config(
 
         return config
 
-    def _get_export_config(self, unparsed: UnparsedExportConfig) -> ExportConfig:
-        return ExportConfig(
-            export_as=unparsed.export_as,
-            schema_name=unparsed.schema,
-            alias=unparsed.alias,
+    def _get_export_config(
+        self, unparsed: Dict[str, Any], saved_query_config: SavedQueryConfig
```
MichelleArk, line 673: super-nit: perhaps `unparsed` could be renamed to `unparsed_export_config`? might make more clear what we're combining to get an export config.

tests/functional/saved_queries/test_configs.py
```diff
@@ -0,0 +1,186 @@
...
+        export1 = next(
+            (export for export in saved_query.exports if export.name == "my_export"), None
+        )
+        assert export1 is not None
+        assert export1.config.export_as == ExportDestinationType.VIEW
+        assert export1.config.export_as != saved_query.config.export_as
+        assert export1.config.schema_name == "my_custom_export_schema"
+        assert export1.config.schema_name != saved_query.config.schema
+
+        # assert Export `my_export` has its configs defined from the saved_query because they should take priority
```
MichelleArk, line 126: ✨

tests/functional/saved_queries/test_configs.py
```diff
@@ -0,0 +1,186 @@
...
+
+        # parse with default fixture project config
+        result = runner.invoke(["parse"])
+        assert result.success
+        assert isinstance(result.result, Manifest)
+        assert len(result.result.saved_queries) == 1
+        saved_query = result.result.saved_queries["saved_query.test.test_saved_query"]
+        assert saved_query.config.export_as == ExportDestinationType.VIEW
+
+        # change export's `export_as` to `TABLE` via project config
```
MichelleArk, line 174: ✨

MichelleArk (approved):
✨

core/dbt/parser/schema_yaml_readers.py
QMalcolm: Absolutely 🙂

core/dbt/contracts/graph/model_config.py
QMalcolm: I will run through it with @graciegoheen / @jtcohen6. As of right now exports as a whole don't appear to be documented https://docs.getdbt.com/docs/build/saved-queries 🙃

graciegoheen: That's right! We closed out this docs issue until exports is fully ready to go https://github.com/dbt-labs/docs.getdbt.com/issues/4381

Relevant slack thread here -> https://dbt-labs.slack.com/archives/C05K4R7KZ5Z/p1699478482262759

dbeatty10: @QMalcolm Here's one new PR for saved queries:
https://github.com/dbt-labs/docs.getdbt.com/pull/4469

PR #7920, opened 2023-06-22
feat: move `--profile` and `--target` to global config

rexledesma:
Resolves https://github.com/dbt-labs/dbt-core/issues/7798.

### Description

Move `--target` and `--profile` command line flags to global config. Also, allow these flags to be set by environment variable (`DBT_TARGET`, and `DBT_PROFILE`).

### Checklist

- [x] I have read [the contributing guide](https://github.com/dbt-labs/dbt-core/blob/main/CONTRIBUTING.md) and understand what's expected of me
- [ ] I have signed the [CLA](https://docs.getdbt.com/docs/contributor-license-agreements)
- [x] I have run this code in development and it appears to resolve the stated issue
- [x] This PR includes tests, or tests are not required/relevant for this PR
- [x] I have [opened an issue to add/update docs](https://github.com/dbt-labs/docs.getdbt.com/issues/new/choose), or docs changes are not required/relevant for this PR
- [x] I have run `changie new` to [create a changelog entry](https://github.com/dbt-labs/dbt-core/blob/main/CONTRIBUTING.md#adding-a-changelog-entry)

PR #9081, opened 2023-11-15
Add --target and --profile to global config

### LINKED ISSUE #7798: [CT-2649] [Feature] Move `--target` and `--profile` command line flags to global config

### Describe the feature

When attempting to dynamically configure a user's profile and target, We should prefer `dbt --target TARGET_NAME` over `dbt run --target TARGET_NAME`, and similarly for `--profile`.

`--profiles-dir` is essentially already a global config with [its exposure as an environment variable](https://docs.getdbt.com/docs/core/connect-data-platform/connection-profiles#2-use-the-dbt_profiles_dir-environment-variable-to-change-the-default-location), so I believe other associated configurations with profiles (e.g. `--target` and `--profile`) should also be a global config.

To achieve dynamic behavior, environment variables (e.g. `DBT_TARGET` and `DBT_PROFILE`) can be used to configure `--target` and `--profile` by templating these values into `profiles.yml`. 

However: 

- These environment variables are less explicit than configuring CLI flags, and
- This requires the user to use `env_var(...)` for their `profile.yml`. Ideally, this feature should come out of the box.

### Who will this benefit?

_No response_

_No response_

### Anything else?

_No response_

barton996:
resolves #7798
Related to [pull/7920](https://github.com/dbt-labs/dbt-core/pull/7920)

### Problem
`--profile` and `--target` are not part of the global config and cannot be set once at env var level without using `env_var()` in `profiles.yml`. This feature should be provided out of the box as is the case for `--profiles-dir`.

### Solution
`--profile` and `--target` have been elevated to global config and can now be set by the `DBT_PROFILE` and `DBT_TARGET` env vars.

### Checklist


.changes/unreleased/Features-20231115-092005.yaml
```diff
@@ -0,0 +1,7 @@
+kind: Features
+body: Move --target and --profile command line flags to global config. Also, allow
+  these flags to be set by environment variable (DBT_TARGET, and DBT_PROFILE).
```
dbeatty10, line 3:
```suggestion
body: Global config for --target and --profile CLI flags and DBT_TARGET and DBT_PROFILE environment variables.
```

core/dbt/cli/main.py
```diff
@@ -139,11 +139,13 @@ def global_flags(func):
     @p.populate_cache
     @p.print
     @p.printer_width
+    @p.profile
     @p.quiet
     @p.record_timing_info
     @p.send_anonymous_usage_stats
     @p.single_threaded
     @p.static_parser
+    @p.target
```
jtcohen6, line 148: I think you'll also need to remove these (`@p.profile` and `@p.target`) where they're specified on specific subcommands down below. By adding them to `global_flags`, they are automatically supported both at the root level (`dbt --profile <subcommand>`) and after every subcommand (`dbt subcommand --profile`).

dbeatty10: @barton996 https://github.com/dbt-labs/dbt-core/pull/9081/commits/d404c9de4b8d40ec11105654eb0bae453f943f30 is intended to cover removing `@p.profile` and `@p.target` from all the subcommands, so hopefully you don't need to worry about this piece now.

barton996: @dbeatty10 Thanks. I've also added a unit test to protect against this now

https://github.com/dbt-labs/dbt-core/blob/e925d2b0879c2116a8511a70870e5281ded735f7/tests/unit/test_cli.py#L56

dbeatty10:
### Question

[source](https://github.com/dbt-labs/dbt-core/pull/9081/files#r1399448755)

> @jtcohen6 in the vein of #7036, does the following need to be added to [UserConfig](https://github.com/dbt-labs/dbt-core/blob/1c9cec17878ca0fec537a4cc13812f91ee97b29b/core/dbt/contracts/project.py#L282)?
> 
> ```python
>     profile: Optional[str] = None
>     target: Optional[str] = None
> ```

### Answer

Outside of GitHub, @jtcohen6 and I discussed this, and he replied with the following:

> I think it would be confusing to set those in [UserConfig](https://github.com/dbt-labs/dbt-core/blob/1c9cec17878ca0fec537a4cc13812f91ee97b29b/core/dbt/contracts/project.py#L282), given that:
> 
> - profile is already set in `dbt_project.yml`
> - default target is already set within each profile

### Decision

In light of those considerations, we're going to add `profile` and `target` the list of exceptions of [global flags](https://github.com/dbt-labs/dbt-core/blob/9bb970e6ef26907b993d4e51de6dc6ef95fe4aa0/core/dbt/cli/main.py#L123) **not** included within [UserConfig](https://github.com/dbt-labs/dbt-core/blob/1c9cec17878ca0fec537a4cc13812f91ee97b29b/core/dbt/contracts/project.py#L282).

tests/unit/test_cli_flags.py
```diff
@@ -349,6 +349,38 @@ def test_log_file_settings_from_config(self):
...
+    ):
+        cli_params = ["run"]
+
+        if cli_profile:
+            cli_params = ["--profile", cli_profile] + cli_params
+
+        if cli_target:
+            cli_params = ["--target", cli_target] + cli_params
+
+        context = self.make_dbt_context("run", cli_params)
```
dbeatty10, line 377: To be backwards-compatible, we should make sure that the opposite order works as well.

e.g., we need this to still work:

```shell
dbt run --profile my_profile_name
```

So we'd probably want to make sure this works fine too:

```python
        if cli_profile:
            cli_params = cli_params + ["--profile", cli_profile]

        if cli_target:
            cli_params = cli_params + ["--target", cli_target]
```

dbeatty10 (changes requested):

jaklan:
Any updates on that?

In [docs](https://docs.getdbt.com/reference/global-configs/about-global-configs#available-flags) I can see:
image
but it doesn't seem to work, especially the type `path` is pretty confusing here.

tests/unit/test_cli_flags.py
barton996: Sorry for the glacial reply. I've added this scenario to the unit test and it doesn't appear to be working correctly.

I'm failing to get the TARGET and PROFILE flags to be set when they're passed after `run` as you suggest.

Target and profile are added to the global flags and the global flags are in the decorator for the run context in main. The test assertions are failing because `flags.PROFILE` and `flags.TARGET` are always coming back as `None`.

```
@pytest.mark.parametrize(
        "cli_profile, command_profile, flag_profile",
        [
            (None, None, None),
            ("cli_profile", None, "cli_profile"),
            (None, "command_profile", "command_profile"),
            ("cli_profile", "command_profile", "cli_profile"),
        ],
    )
    @pytest.mark.parametrize(
        "cli_target, command_target, flag_target",
        [
            (None, None, None),
            ("cli_target", None, "cli_target"),
            (None, "command_target", "command_target"),
            ("cli_target", "command_target", "cli_target"),
        ],
    )
    def test_profile_settings_interaction_1(
        self, cli_profile, cli_target, command_profile, command_target, flag_profile, flag_target
    ):
        cli_params = ["run"]

        if cli_profile:
            # dbt level --profile flag goes before the run command
            cli_params = ["--profile", cli_profile] + cli_params

        if cli_target:
            # dbt level --target flag goes before the run command
            cli_params = ["--target", cli_target] + cli_params

        if command_profile:
            # command level --profile flag goes after the run command
            cli_params = cli_params + ["--profile", command_profile]

        if command_target:
            # command level --target flag goes after the run command
            cli_params = cli_params + ["--target", command_target]

        context = self.make_dbt_context("run", cli_params)

        flags = Flags(context, None)

        assert flags.PROFILE == flag_profile
        assert flags.TARGET == flag_target
```

dbeatty10:
> Any updates on that?

> it doesn't seem to work, especially the type `path` is pretty confusing here.

Thanks for calling this out @jaklan -- you are right, these docs will need to be updated until this PR is merged+released.

I just opened this issue to update the docs in the meantime: https://github.com/dbt-labs/docs.getdbt.com/issues/5111

tests/unit/test_cli_flags.py
dbeatty10: No worries @barton996

### TLDR
After looking at the pre-existing test cases more closely, we might not even need this at all.

### More detail
i.e., it looks like the general case is already covered by [`test_global_flag_at_child_context`](https://github.com/dbt-labs/dbt-core/blob/c5eb6d70b44f966d17e76e834703bbc6ad05549e/tests/unit/test_cli_flags.py#L356). And the case of specifying it both at the before and after the subcommand is covered by [`test_duplicate_flags_raises_error`](https://github.com/dbt-labs/dbt-core/blob/c5eb6d70b44f966d17e76e834703bbc6ad05549e/tests/unit/test_cli_flags.py#L349). 

So I don't think we need new tests here after all.

If we _do_ end up wanting to add tests specific to `--profile` and `--target`, I think they can be covered in separate test cases since they aren't designed or expected to have interaction effects (unlike the [`test_log_format_interaction`](https://github.com/dbt-labs/dbt-core/blob/c5eb6d70b44f966d17e76e834703bbc6ad05549e/tests/unit/test_cli_flags.py#L296) test that this originally appeared like it was patterned after).

### Potential tests instead

If we do want to add new tests after moving `profile` and `target` to be global flags, here's a couple tests that worked when I tried them out:

```python
    @pytest.mark.parametrize(
        "flag_value",
        [
            (None),
            ("my_profile"),
        ],
    )
    def test_profile_settings(self, flag_value):
        # CLI parameters _before_ subcommand
        parent_context_a = self.make_dbt_context("parent_context_a", ["--profile", flag_value])
        child_context_a = self.make_dbt_context("child_context_a", ["run"], parent_context_a)
        flags_a = Flags(child_context_a)

        # CLI parameters _after_ subcommand
        parent_context_b = self.make_dbt_context("parent_context_b", ["run"])
        child_context_b = self.make_dbt_context(
            "child_context_b", ["--profile", flag_value], parent_context_b
        )
        flags_b = Flags(child_context_b)

        assert flags_a.PROFILE == flags_b.PROFILE
        assert flags_a.PROFILE == flag_value
        assert flags_b.PROFILE == flag_value

    @pytest.mark.parametrize(
        "flag_value",
        [
            (None),
            ("my_target"),
        ],
    )
    def test_target_settings(self, flag_value):
        # CLI parameters _before_ subcommand
        parent_context_a = self.make_dbt_context("parent_context_a", ["--target", flag_value])
        child_context_a = self.make_dbt_context("child_context_a", ["run"], parent_context_a)
        flags_a = Flags(child_context_a)

        # CLI parameters _after_ subcommand
        parent_context_b = self.make_dbt_context("parent_context_b", ["run"])
        child_context_b = self.make_dbt_context(
            "child_context_b", ["--target", flag_value], parent_context_b
        )
        flags_b = Flags(child_context_b)

        assert flags_a.TARGET == flags_b.TARGET
        assert flags_a.TARGET == flag_value
        assert flags_b.TARGET == flag_value
```

peterallenwebb (approved):

tests/unit/test_cli_flags.py
dbeatty10: Per discussion with @peterallenwebb, we feel good about removing this test altogether since:
1. It's covered by [test_global_flag_at_child_context](https://github.com/dbt-labs/dbt-core/blob/c5eb6d70b44f966d17e76e834703bbc6ad05549e/tests/unit/test_cli_flags.py#L356).
2. Doug verified all of this manually also

tests/unit/test_cli.py
```diff
@@ -52,3 +52,14 @@ def test_commands_in_enum_and_dict(self):
...
+    def test_global_flags_not_on_subcommands(self):
+        def run_test(command):
+            # For each subcommand
+            for command_name, command in command.commands.items():
+                # Get string representation of all parameter options for this subcommand e.g. "['-t', '--target']"
+                parameter_options = [str(sorted(param.opts)) for param in command.params]
+                # Check there are no parameter options that have been applied twice
+                assert sorted(list(set(parameter_options))) == sorted(parameter_options)
+
+        run_test(cli)
```
dbeatty10, line 65: Per discussion with @peterallenwebb, we don't think we need this, and we can add it back in the future if needed.

So I'm going to remove it.

dbeatty10, line 65:
```suggestion
```

dbeatty10 (approved):
Approved from a product and user experience perspective

yu-iskw:
@barton996 @dbeatty10 I have a question about the behavior of the environment variables, though the case might be an edge case.  If we set `DBT_TAGET` and `--target` at the same time, `DBT_TAGET` is used isn't it? Personally, it would sound natural that `--target` can have higher priority.

```
$ DBT_TARGET=no_such_target dbt build --target dev --threads 20

02:46:21  Running with dbt=1.8.0
02:46:21  Encountered an error:
Runtime Error
  The profile 'jaffle_shop' does not have a target named 'no_such_target'. The valid target names for this profile are:
   - dev
   - prod
```

dbeatty10:
@yu-iskw It is _supposed_ to use the following precedence order:

> ### Global config precedence
> 
> dbt will pick the config in the following order (lower takes priority):
> 1. [user config](https://docs.getdbt.com/reference/global-configs/yaml-configurations)
> 1. [environment variable](https://docs.getdbt.com/reference/global-configs/environment-variable-configs)
> 1. [CLI flag](https://docs.getdbt.com/reference/global-configs/command-line-flags)
> 
> i.e., if all three are provided, then the CLI flag takes precedence.

If this is not the case for [`target`](https://docs.getdbt.com/docs/core/connect-data-platform/connection-profiles#about-profiles), would you be willing to [open a bug report ](https://github.com/dbt-labs/dbt-core/issues/new?assignees=&labels=bug%2Ctriage&projects=&template=bug-report.yml&title=%5BBug%5D+%3Ctitle%3E)?

PR #9101, opened 2023-11-15
Add test asserting GraphRunnableTasks attempt to cancel connections on SystemExit

### LINKED ISSUE #9088: [CT-3390] [IDE] Add test case for treating SystemExit exception as an interrupt
### Housekeeping

### Short description

In #8994 we added functionality that ensured that `SystemExit` exceptions sent the cancellation command of running operations to the data warehouse. However, we didn't add a test for this, and we should.

### Acceptance criteria

A test exists which checks that `SystemExit` exceptions send the cancellation command to data warehouses.

### Impact to Other Teams

This is a test addition, not functionality. And thus this shouldn't impact other teams

#8994 is being backported to 1.7.latest and 1.6.latest, in which case this test should also be backported.

### Context

_No response_

QMalcolm:
resolves #9088 

### Problem

In #8994 we added functionality for cancelling connections on caught `SystemExit` exceptions, but did not add a test for checking this functionality. 

### Solution
To test this without running a full integration test, we created a custom implementation of the `GrapheRunnableTask` abstract class which guarantees a `SystemExit` will be raised on execution. I'm sure a mocking library could have helped, but this seemed slim enough to not need something more powerful.

*Note*: There is not changelog entry because this only adds a test

### Checklist

### DIFF
--- a/tests/unit/test_graph_runnable_task.py
+++ b/tests/unit/test_graph_runnable_task.py
@@ -63,3 +63,13 @@ def test_graph_runnable_task_cancels_connection_on_keyboard_interrupt():
 
     # If `did_cancel` is True, that means `_cancel_connections` was called
     assert task.did_cancel is True
+
+
+def test_graph_runnable_task_doesnt_cancel_connection_on_generic_exception():
+    task = MockRunnableTask(exception_class=Exception)
+
+    with pytest.raises(Exception):
+        task.execute_nodes()
+
+    # If `did_cancel` is True, that means `_cancel_connections` was called
+    assert task.did_cancel is False
--- a/tests/unit/test_graph_runnable_task.py
+++ b/tests/unit/test_graph_runnable_task.py
@@ -1,0 +1,52 @@
+import pytest
+
+from dataclasses import dataclass
+from dbt.task.runnable import GraphRunnableTask
+from typing import AbstractSet, Any, Dict, Optional
+
+
+@dataclass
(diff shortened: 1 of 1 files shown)


MichelleArk:
> I'm sure a mocking library could have helped, but this seemed slim enough to not need something more powerful.

Fair! It may have required a good number of patches on an actual runnable task, I've gone down a similar rabbit hole before... 🐰

tests/unit/test_graph_runnable_task.py
```diff
@@ -0,0 +1,52 @@
...
+
+    def defer_to_manifest(self, adapter, selected_uids: AbstractSet[str]):
+        """This is an `abstract_method` on `GraphRunnableTask`, thus we must implement it"""
+        return None
+
+
+def test_graph_runnable_task_cancels_connection_on_system_exit():
+    task = CustomRunnableTask(args=MockArgs(), config=MockConfig(), manifest=None)
+
+    with pytest.raises(SystemExit):
```
MichelleArk, line 48: while we're here let's also test for [KeyboardInterrupt](https://github.com/dbt-labs/dbt-core/pull/8994/files#diff-f01af78b7eaa1bacc301afa75eea4c7cdbe03064b3913bc2ce3f6bf01d0bc125R374)?

MichelleArk (approved):
small comment, otherwise LGTM! thank you 🎉

PR #9121, opened 2023-11-18
Fix #8618: Refactor `ManifestLoader.load`

### LINKED ISSUE #8618: Refactor ManifestLoader.load method
This method is quite long and difficult to understand its overall logic - let's start refactoring it by extracting methods and using them in ManifestLoader.load

aranke:
resolves #8618 

### Problem

`ManifestLoader.load` is unwieldy and complicated. 

### Solution

Break up the code into multiple methods.

### Checklist


core/dbt/parser/manifest.py
```diff
@@ -333,15 +332,214 @@ def get_full_manifest(
...
+            fire_event(PartialParsingSkipParsing())
+        else:
+            self.load_macros_and_tests(project_parser_files, orig_project_parser_files)
+
+        self.load_external_nodes()
+        self.check_for_model_deprecations()
+
+        return self.manifest
+
+    def rebuild_manifest_pp(
```
ChenyuLInx, line 468: Thoughts on naming this: get_partial_parsing files?
And maybe return whether we need to do skip_parsing?
Returning the value instead of modifying a global attribute makes the code easier to understand.

core/dbt/parser/manifest.py
```diff
@@ -361,215 +559,27 @@ def load(self) -> Manifest:
...
-                self.manifest.rebuild_ref_lookup()
-                self.process_refs(
-                    self.root_project.project_name,
-                    self.root_project.dependencies,
-                )
-                # parent and child maps will be rebuilt by write_manifest
-
-        if not skip_parsing or external_nodes_modified:
+            # parent and child maps will be rebuilt by write_manifest
+        if not self.skip_parsing or external_nodes_modified:
```
ChenyuLInx, line 579: I don't think this writing out manifest should belong to this function given the function name.
If we move it out to the load layer then I think we don't have to add the global attribute `skip_parsing`?

core/dbt/parser/manifest.py
```diff
@@ -333,15 +332,214 @@ def get_full_manifest(
 
         return manifest
 
+    def load_macros_and_tests(
+        self,
+        project_parser_files: Dict[str, Dict[str, List[str]]],
+        orig_project_parser_files: Dict[str, Dict[str, List[str]]],
+    ) -> None:
+        """
+        Load Macros and tests
```
ChenyuLInx, line 341: Can we add a bit more in the description about what objects will be changed after running this function?

core/dbt/parser/manifest.py
```diff
@@ -333,15 +332,214 @@ def get_full_manifest(
...
+        self._perf_info.process_manifest_elapsed = time.perf_counter() - start_process
+        self._perf_info.static_analysis_parsed_path_count = (
+            self.manifest._parsing_info.static_analysis_parsed_path_count
+        )
+        self._perf_info.static_analysis_path_count = (
+            self.manifest._parsing_info.static_analysis_path_count
+        )
+
     # This is where the main action happens
     def load(self) -> Manifest:
```
gshank, line 450: Could you move this method up above "load_macros_and_tests" (or whatever the name will be)?  It's better to be able to read down the file in the approximate order of execution instead of having to jump back and forth.

gshank (changes requested):
Do we really want to keep "get_full_manifest" (a class method) as the main entry point to the ManifestLoader?  Did you consider just constructing the ManifestLoader object and doing an instance method, like ManifestLoader->load?

PR #9125, opened 2023-11-20
Forward default implementation of rename_relation to get_rename_sql

### LINKED ISSUE #9124: rename_relation macro should call get_rename_sql instead of defaulting to alter table
### Housekeeping

### Short description

_Sorry, could not really find a good match in the issue templates for this._

The issue is about [this line](https://github.com/dbt-labs/dbt-core/blob/1c9cec17878ca0fec537a4cc13812f91ee97b29b/core/dbt/include/global_project/macros/relations/rename.sql#L33C1-L34C1) in the `default__rename_relation` macro.

The name of the macro implies that it would be able to rename a relation, but in fact, it's only able to rename tables as it's calling `alter table`.

We do have a better version of this in the codebase already, namely the `get_rename_sql` macro.
[That version](https://github.com/dbt-labs/dbt-core/blob/1c9cec17878ca0fec537a4cc13812f91ee97b29b/core/dbt/include/global_project/macros/relations/rename.sql#L7) is much more extensive and is capable of renaming all supported relations.

My proposal would be to forward the `alter table` call in `rename_relation` to `get_rename_sql`, so that we end up with something like this:

```sql+jinja
{% macro default__rename_relation(from_relation, to_relation) -%}
  {% set target_name = adapter.quote_as_configured(to_relation.identifier, 'identifier') %}
  {{ log('from_relation: ' ~ from_relation, info=True) }}
  {% call statement('rename_relation') -%}
    {{ get_rename_sql(from_relation, target_name) }}
  {%- endcall %}
{% endmacro %}
```

Right now, adapter maintainers who'd like to simply rename relations without having to deal with the hassle of calling `get_rename_sql` directly, have to create such an override in their own adapter. That is a bit tedious and probably not intended.

### Acceptance criteria

- All current tests still succeed

### Impact to Other Teams

I don't foresee any impact except that adapter maintainers can drop their custom override. The original behavior where the macro calls `alter table` in the end will still work. The difference in behavior would be that it no longer calls `alter table` for views, but `alter view` instead and likewise for materialized views.

no

### Context

_No response_

sdebruyn:
resolves dbt-labs/dbt-adapters#152

### Problem

See dbt-labs/dbt-adapters#152 for more details.

### Solution

See dbt-labs/dbt-adapters#152 for more details.

### Checklist

### DIFF
--- a/.changes/unreleased/Under the Hood-20231120-233336.yaml
+++ b/.changes/unreleased/Under the Hood-20231120-233336.yaml
@@ -1,0 +1,6 @@
+kind: Under the Hood
+body: Forward default implementation of rename_relation to get_rename_sql
+time: 2023-11-20T23:33:36.103869+01:00
+custom:
+  Author: sdebruyn
+  Issue: "9124"
--- a/core/dbt/include/global_project/macros/relations/rename.sql
+++ b/core/dbt/include/global_project/macros/relations/rename.sql
@@ -30,6 +30,6 @@
 {% macro default__rename_relation(from_relation, to_relation) -%}
   {% set target_name = adapter.quote_as_configured(to_relation.identifier, 'identifier') %}
   {% call statement('rename_relation') -%}
-    {{ get_rename_sql(from_relation, target_name) }}
+    {{ get_rename_sql(from_relation, target_name) -}}
   {%- endcall %}
 {% endmacro %}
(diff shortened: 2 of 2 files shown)


sdebruyn:
From the failing tests:

```
E       Expected: [call('/* dbt */\nalter table "postgres"."test_schema".table_a rename to table_b', None)]
E       Actual: [call('BEGIN', None),
E        call('/* dbt */\n\n    alter table "postgres"."test_schema".table_a rename to table_b\n', None)]
```

Whitespace differences are possible with this change, not as straightforward to fix them though. ☹️

But what I don't get is why suddenly there is an extra call `BEGIN`. Where does it come from? There are no extra statement blocks AFAIK.

dbeatty10:
Thanks for taking the time to open this PR @sdebruyn! Since opening, we've [decoupled dbt Adapters from dbt Core](https://github.com/dbt-labs/dbt-core/discussions/9171), and this code now lives in a separate repo: [dbt-adapters](https://github.com/dbt-labs/dbt-adapters).

PR #9134, opened 2023-11-22
remove adapter.get_compiler

### LINKED ISSUE #9148: [CT-3422] Remove usage of dbt.compilation in dbt/adapters
BaseAdapter.get_compiler imports the core compiler to provide a compiler for tasks that require it. This pluggability was initially intended to support, but the interface is much to broad (requires a full manifest to be passed around) and was not used. If we need this capability in the future, we should have much finer-combed interfaces to support this capability.

MichelleArk:
resolves #9148

### Problem

* BaseAdapter.get_compiler has a circular dependency on `dbt.compilation.Compiler`
* the compiler protocol is overly broad - it requires passing a full manifest which is also a circular dependency. if we want to support this type of override in the future, it should be more granular and narrowly scoped to set of adapter-specific interfaces. 

### Solution

* remove CompilerProtocol and BaseAdapter.get_compiler method + usages. 

### Checklist

  - confirmed no override of get_compiler method in sqlserver adapter: https://github.com/dbt-msft/dbt-sqlserver/blob/master/dbt/adapters/sqlserver/sql_server_adapter.py
  - bigquery: https://github.com/dbt-labs/dbt-bigquery/blob/main/dbt/adapters/bigquery/impl.py
  - snowflake:  https://github.com/dbt-labs/dbt-snowflake/blob/main/dbt/adapters/snowflake/impl.py
  - spark: https://github.com/dbt-labs/dbt-spark/blob/main/dbt/adapters/spark/impl.py
  - redshift: https://github.com/dbt-labs/dbt-redshift/blob/main/dbt/adapters/redshift/impl.py

### DIFF
--- a/.changes/unreleased/Breaking Changes-20231127-114757.yaml
+++ b/.changes/unreleased/Breaking Changes-20231127-114757.yaml
@@ -1,0 +1,6 @@
+kind: Breaking Changes
+body: Remove adapter.get_compiler interface
+time: 2023-11-27T11:47:57.443202-05:00
+custom:
+  Author: michelleark
+  Issue: "9148"
--- a/core/dbt/adapters/base/impl.py
+++ b/core/dbt/adapters/base/impl.py
@@ -1370,11 +1370,6 @@ class BaseAdapter(metaclass=AdapterMeta):
         """
         pass
 
-    def get_compiler(self):
-        from dbt.compilation import Compiler
-
-        return Compiler(self.config)
-
     # Methods used in adapter tests
     def update_column_sql(
         self,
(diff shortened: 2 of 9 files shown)


MichelleArk:
@cla-bot check

jtcohen6 (approved):
For posterity: The original motivation for doing this was, what if your database doesn't support nested CTEs, and your adapter wants to support ephemeral models?
- https://github.com/dbt-labs/dbt-core/pull/2712

This is the case for MSFT SQLServer. Way back in the day, @dataders and I hacked on what a solution here might look like:
- https://github.com/dbt-msft/dbt-sqlserver/issues/137

Then and now, the far better answer is: _Simply demand that your DWH vendor implement support for nested CTEs._ The all-time classic treatise (by Anders in 2021): https://dbt-msft.github.io/dbt-msft-docs/docs/nested_CTES/

In the meantime, we should absolutely remove the `Compiler` and `Manifest` from the common interface boundary shared by `dbt-core` and adapters.

PR #9136, opened 2023-11-22
Add a GitHub action for performance checks.

### LINKED ISSUE #8323: [Applied State] Develop a Performance Analysis GitHub Action
### Housekeeping

### Short description

Develop a performance analysis process to monitor changes in performance over time, and quickly catch performance regressions.

### Acceptance critera

- A new GitHub action called Performance Analysis has been added to dbt-core
- The new action runs several different commands across projects of different sizes, and with different performance characteristics
- For each command we note, at a minimum: (1) Peak memory use, (2) run time, (3) stats on total block storage read/writes.
- Minimum performance criteria should be set, and the action should fail if the criteria are not met.

### Context

_No response_

peterallenwebb:
resolves #8323

### Problem

We don't have a way to systematically monitor core for performance regressions.

### Solution

Create a GitHub action to measure performance as a starting point for further experimentation. When run, the action produces a perf_results.json file, and the script has an option to compare two runs against each other.

### Checklist

### DIFF
--- a/scripts/perf-check.py
+++ b/scripts/perf-check.py
@@ -64,13 +64,12 @@ def run_jobs(projects):
             print(f"running job {job_name}")
             cwd = project["path"] if "path" in project else project_name
             start = time.perf_counter()
-            res = subprocess.run(job["command"], cwd=cwd, capture_output=True)
+            res = subprocess.run(job["command"], cwd=cwd)
             end = time.perf_counter()
             if res.returncode != 0:
                 results[job_name] = {"succeeded": False}
             else:
                 results[job_name] = {"succeeded": True, "time": end - start}
-                pass
 
     return results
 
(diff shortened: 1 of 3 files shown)


scripts/perf-check.py
```diff
@@ -0,0 +1,129 @@
...
+    with open("perf_check.json", "w") as w:
+        json.dump(results, w, indent=4)
+
+
+if len(sys.argv) < 2:
+    print_usage()
+elif sys.argv[1] == "baseline":
+    baseline(projects)
+elif sys.argv[1] == "compare":
+    compare(sys.argv[2], sys.argv[3])
```
MichelleArk, line 127: It'd be great if we could run the comparison utility in CI for PR commit against main to detect regressions. I'm imagining that would look like building a baseline off of main and uploading the artifact to a known path, later pulling it down as a step in the PR workflow above to providing the path to this method. 

Is that an extension you were imagining as well? Or was this intended to be used locally in the near-term?

PR #9149, opened 2023-11-27
Decouple macro generator from adapters

MichelleArk:
resolves #https://github.com/dbt-labs/dbt-core/issues/9205

Removed two usages of core imports in dbt/adapters: 
1. core/dbt/adapters/base/query_headers.py:
    - dbt.clients.jinja::QueryStringGenerator
2. core/dbt/adapters/base/impl.py:
    - dbt.clients.jinja::MacroGenerator

### Problem

dbt.clients.jinja was used in 2 places: query header generation & macro execution. In both cases, subclasses of `BaseMacroGenerator` were imported in order to generate callables out of manifest macros. 

### Solution

- Refactored `BaseMacroGenerator` to dbt/common
- Defined `QueryStringGenerator` in dbt/adapters since it was only used by adapters
- Split MacroGenerator into `CallableMacroGenerator` (common) and `MacroGenerator` (which subclasses CallableMacroGenerator, in core). 
  - dbt/adapters uses `CallableMacroGenerator`, which does not provide any dependency tracking `manifest.nodes.depends_on.append` during macro execution.
### Checklist


core/dbt/common/utils/jinja.py
```diff
@@ -0,0 +1,33 @@
+from dbt.common.exceptions import DbtInternalError
+
+
+MACRO_PREFIX = "dbt_macro__"
```
colin-rogers-dbt, line 4: Should this just be folded into `clients.jinja`? It seems like we're defining app logic in a utils package which exceeds the implied scope of a "utility" module.

colin-rogers-dbt (approved):

core/dbt/common/utils/jinja.py
MichelleArk: Created a follow-on issue for further refactoring in dbt/common: https://github.com/dbt-labs/dbt-core/issues/9206

PR #9155, opened 2023-11-28
Fix ensuring we produce valid jsonschema artifacts for manifest, catalog, sources, and run-results

### LINKED ISSUE #8991: [CT-3316] [Bug] Invalid JSON schema for manifest v11

### Current Behavior

Manifest v11 [JSON schema](https://schemas.getdbt.com/dbt/manifest/v11.json ) has been released with dbt 1.7.0. However, it's not a valid JSON schema: 

```json
{
  "$ref": "#/$defs/WritableManifest",
  "$defs": {
    "ManifestMetadata": {
      "type": "object",
      "title": "ManifestMetadata",
      ...
  }
}
```

Seems like `$defs.ManifestMetadata` contains the actual JSON schema that's supposed to be published.

### Expected Behavior

A valid JSON schema to be published.

### Steps To Reproduce

1. Copy https://schemas.getdbt.com/dbt/manifest/v11.json
2. Paste it into a JSON schema validator, e.g. https://www.jsonschemavalidator.net/

### Relevant log output

_No response_

### Environment

```markdown
- OS:
- Python:
- dbt:
```

_No response_

### Additional Context

_No response_

QMalcolm:
resolves #8991 

### Problem

In #8991 it was noted that our the JsonSchema schema we produced for our manifest wasn't a valid schema. The underlying cause ended up being that we were creating an embed-able/fractal/partial schema. That is to say what we were producing wasn't a full fledged schema, but one that was anticipated to be ref'd by a schema and never by itself. This would have been fine-ish if we hadn't been including `$id`, but we were. And specifying `$id` says "hey I'm a full fledged schema" when it wasn't. This in turn made some JsonSchema tools not like our schemas.

### Solution

Stop producing everything as `ref` by dropping `all_refs=True`. Additionally we made some other enhancements for sanity. We now explicitly say which dialect to use to produce the schemas, and encode the used dialect into the schema itself.

**_Note_**: We don't regenerate the v11 manifest in this PR, I plan to do that as part of the backport. This is to guarantee the generated jsonschema manifest schema matches the `WritableManifest` object that exists in `1.7.latest`

<div>
    <a href="https://www.loom.com/share/acce7f154e4a4b8bb2d122bb4ad660fd">
      
    </a>
</div>

### Checklist


core/dbt/contracts/graph/manifest.py
```diff
@@ -1550,7 +1550,7 @@ def __init__(self, macros) -> None:
 
 
 @dataclass
-@schema_version("manifest", 11)
+@schema_version("manifest", 12)
```
emmyoop, line 1553: Why bump the manifest version but not the run_results version?

.changes/unreleased/Fixes-20231127-165244.yaml
```diff
@@ -0,0 +1,7 @@
+kind: Fixes
+body: Ensure we produce validate jsonschema schemas for manifest, catalog, run-results,
```
emmyoop, line 2:
```suggestion
body: Ensure we produce valid jsonschema schemas for manifest, catalog, run-results,
```

core/dbt/contracts/util.py
```diff
@@ -197,7 +198,7 @@ class VersionedSchema(dbtClassMixin):
     @classmethod
     @functools.lru_cache
     def json_schema(cls) -> Dict[str, Any]:
-        json_schema_obj = build_json_schema(cls, all_refs=True)
+        json_schema_obj = build_json_schema(cls, dialect=DRAFT_2020_12, with_dialect_uri=True)
```
emmyoop, line 201: Interesting that this isn't a default in mashumaro.

core/dbt/contracts/graph/manifest.py
QMalcolm: I couldn't find documentation on when the different schema versions should be bumped. I knew that we always bumped the manifest version on minor release (and that we yet haven't done so in main since cutting 1.7 into it's release branch). I didn't know when `run-results`, `catalog`, and `sources` should be bumped, and didn't find documentation with that information

gshank: Don't we want to fix the v11 json schema?  Otherwise we're leaving a broken schema out there. This will have to be backported to 1.7. I guess we could fix the v11 schema in 1.7.latest, but it seems like it would be simpler to just fix the v11 schema and then bump to v12 when there are actual schema changes. (Have there been schema changes since v11 was produced? If there have been, we should generate v11 schema from older class versions.)

QMalcolm: In my mind v11 manifest should be generated from the objects that exist on the 1.7.latest branch during the backport. Doing so guarantees that the generated manifest jsonschema matches the `WriteableManifest` class. I went through the commit history of `/core/dbt/contracts` of `main` and `1.7.latest`, and they do differ. Looking at the commit differences I don't think there should be a difference. However, the risk we run in regenerating the v11 manifest here is the v11 manifest jsonschema doesn't match the actual written artifacts. I believe partial parsing would then break as we won't be able to instantiate the `WritableManifest` for the `PreviousState`. Additionally, I believe in such a scenario dbt cloud jobs would begin failing because the artifacts will likely fail to pass jsonschema validation. It's fairly easy to reduce this risk to _zero_ by simply regenerating the v11 manifest in the backport itself. Given that it's fairly easy to reduce the risk to _zero_, I recommend we do so. This will be going out just before the holidays, and it would suck if we had to do another fix for this / have an incident.

That said, I'm happy to not generate the v12 manifest in this PR (and also not regenerate the v11 manifest). However that just makes it harder to prove that this PR to main is doing what is supposed to. If that's acceptable though, happy to drop those changes

gshank (approved):
Looks good

PR #9169, opened 2023-11-29
Remove `--dry-run` flag from `dbt deps`

### LINKED ISSUE #9100: [CT-3395] [Feature] Merge `--dry-run` and `--lock` into the same flag for dbt deps

### Describe the feature

> I don't see the need for both the flags `--dry-run` and `--lock`, they could be merged into a same flag as far as I can tell.

_Originally posted by @seub in https://github.com/dbt-labs/docs.getdbt.com/issues/4470#issuecomment-1811769591_
            

See https://github.com/dbt-labs/docs.getdbt.com/issues/4470#issuecomment-1811769591_ for context

### Who will this benefit?

See https://github.com/dbt-labs/docs.getdbt.com/issues/4470#issuecomment-1811769591_ for context

_No response_

### Anything else?

_No response_

jtcohen6:
resolves #9100

### Problem

In theory, `dbt deps --add-package --dry-run` only updates `packages.yml` with the package entry being added, without actually resolving packages, writing `package-lock.yml`, or installing packages.

It took me a while to understand this workflow:
```bash
# add a package to *just* packages.yml / dependencies.yml
$ dbt deps --add-package dbt-labs/dbt_utils@1.0.0 --dry-run
# now do 'package resolution' and update lock file
$ dbt deps --lock
# now actually install packages, using lock file
$ dbt deps
```
~It doesn't seem to work in practice.~ Now that I've understood the workflow, I understand why we added this flag — but I don't think it's essential that we support it. I'd rather remove the flag entirely than have it hanging out as extra weight. It's already possible to achieve what is (IMO) the important part of this workflow with `dbt deps --lock --add-package ...`.

### Solution

Remove the `dbt deps --dry-run` flag.

### Checklist

### DIFF
--- a/.changes/unreleased/Breaking Changes-20231129-091921.yaml
+++ b/.changes/unreleased/Breaking Changes-20231129-091921.yaml
@@ -1,0 +1,7 @@
+kind: Breaking Changes
+body: Rm --dry-run flag from 'dbt deps --add-package', in favor of just 'dbt deps
+  --lock'
+time: 2023-11-29T09:19:21.071212+01:00
+custom:
+  Author: jtcohen6
+  Issue: "9100"
--- a/core/dbt/cli/main.py
+++ b/core/dbt/cli/main.py
@@ -457,7 +457,6 @@ def debug(ctx, **kwargs):
 @p.target
 @p.vars
 @p.source
-@p.dry_run
 @p.lock
 @p.upgrade
 @p.add_package
@@ -482,12 +481,6 @@ def deps(ctx, **kwargs):
             raise BadOptionUsage(
                 message=f"Version is required in --add-package when a package when source is {flags.SOURCE}",
                 option_name="--add-package",
-            )
-    else:
-        if flags.DRY_RUN:
-            raise BadOptionUsage(
-                message="Invalid flag `--dry-run` when not using `--add-package`.",
-                option_name="--dry-run",
             )
     task = DepsTask(flags, ctx.obj["project"])
     results = task.run()
--- a/core/dbt/cli/params.py
+++ b/core/dbt/cli/params.py
@@ -83,13 +83,6 @@ deprecated_defer = click.option(
     hidden=True,
 )
 
-dry_run = click.option(
-    "--dry-run",
-    envvar=None,
-    help="Option to run `dbt deps --add-package` without updating package-lock.yml file.",
-    is_flag=True,
-)
-
 empty = click.option(
     "--empty/--no-empty",
     envvar="DBT_EMPTY",
--- a/core/dbt/task/deps.py
+++ b/core/dbt/task/deps.py
@@ -220,8 +220,9 @@ class DepsTask(BaseTask):
             if previous_hash != current_hash:
                 self.lock()
 
-        # Early return when dry run or lock only.
-        if self.args.dry_run or self.args.lock:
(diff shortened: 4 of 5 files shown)


seub:
Thanks @jtcohen6 !

> Now that I've understood the workflow, I understand why we added this flag

I don't, but maybe I'm missing something. Since `--dry-run` can only be used with `--add-package`, whatever effect it has can be the effect of using `--lock` with `--add-package`. Right?

Anyway I think that's what your PR is saying, it looks good to me :+1:

jtcohen6:
> Since `--dry-run` can only be used with `--add-package`, whatever effect it has can be the effect of using `--lock` with `--add-package`. Right?

@seub Technically, `--dry-run` will update only `packages.yml`/`dependencies.yml`, whereas `--lock` will also update `package-lock.yml`. So they are distinct, but not in a way that I think is important or particularly useful.

dbeatty10:
> Technically, `--dry-run` will update only `packages.yml`/`dependencies.yml`, whereas `--lock` will also update `package-lock.yml`.

Did `--dry-run` work for you @jtcohen6?

With dbt-core 1.7.3, `--dry-run` still adds/updates the lock file for me:
- Log output includes `Updating lock file in file path:`
- `package-lock.yml` is created if it doesn't exist, and it is updated otherwise

### Log output
```
$ dbt deps --add-package dbt-labs/dbt_utils@0.8.0 --dry-run

14:32:17  Running with dbt=1.7.3
14:32:17  Found duplicate package in packages.yml, removing: {'version': '1.0.0', 'package': 'dbt-labs/dbt_utils'}
14:32:17  Added new package dbt-labs/dbt_utils@0.8.0 to /dbt-core-9076/packages.yml
14:32:17  Updating lock file in file path: /dbt-core-9076/package-lock.yml
```

### Benefits of this PR

So I think there are two different issues that this PR resolves:
1. `dbt deps --dry-run` doesn't work as originally intended; it behaves no differently than `--lock` currently
2. `dbt deps --lock` is sufficient to add a new dependency without actually installing the dependency

It also avoids multiple types of confusion:
- Removing `--dry-run` avoids potential confusion with other "dry run" type functionality in dbt (e.g. [#8971](https://github.com/dbt-labs/dbt-core/pull/8971) or [#7839](https://github.com/dbt-labs/dbt-core/pull/7839) or the `--dry` flag that was removed in 0.7.0).
- My personal expectation is that dry runs don't do any modifications, so removing this flag will also avoid the non-intuitive behavior that `--dry-run` is actually modifying the `packages.yml` /`dependencies.yml` file.

### Trade-offs

By removing `dbt deps --dry-run ...`, users will get an error message that says "dbt.cli.exceptions.DbtUsageException: No such option: --dry-run".

There is no command that adds to `packages.yml` without also locking. The workaround will be to add the dependency to the file manually. If this turns out to be a desirable feature, then we can always add a new flag in the future.

dbeatty10 (approved):
Looks good! Full rationale for supporting the details in this PR are [here](https://github.com/dbt-labs/dbt-core/pull/9169#issuecomment-1850398325).

Not blocking suggestion:
- so that we can completely excise "dry_run", my only suggestion would be to change the name of the `test_deps_add_dry_run` test to be `test_deps_add_without_install` instead.

jtcohen6:
Thank you @dbeatty10 !

PR #9173, opened 2023-11-29
Update parser to support conversion metrics

WilliamDee:
resolves #

### Checklist


core/dbt/parser/manifest.py
```diff
@@ -1598,6 +1598,44 @@ def _process_metric_node(
...
+                f"A semantic model having a measure `{conversion_type_params.conversion_measure.name}` does not exist but was referenced.",
+                node=metric,
+            )
+        if target_conversion_semantic_model.config.enabled is False:
+            raise dbt.exceptions.ParsingError(
+                f"The measure `{conversion_type_params.conversion_measure.name}` is referenced on disabled semantic model `{target_conversion_semantic_model.name}`.",
+                node=metric,
+            )
+        metric.depends_on.add_node(target_base_semantic_model.unique_id)
+        metric.depends_on.add_node(target_conversion_semantic_model.unique_id)
```
QMalcolm, line 1638: This looks to follow the same process as what we do for `MetricType.CUMULATIVE` and `MetricType.SIMPLE` metric types, except we also do it twice. Could we de-duplicate this? I think we could do so by at the beginning of what would be `if metric.type is MetricType.SIMPLE or metric.type is MetricType.CUMULATIVE or metric.type is MetricType.CONVERSION`
We first build the list of measure inputs to lookup (for simple / cumulative it'd be a list with 1 measure input, for conversion it'd be a list of two). Then we just do necessary logic in a loop.

QMalcolm (changes requested):
This looks really good 🚀 Thank you for being so willing to jump into this. There is a bit of code in the processing of metric nodes that I think we can deduplicate. Besides that what we should do, is do an alpha or beta DSI release (0.5.0a1?) with the conversion metric updates, then we can include that in this work and add the appropriate tests 🙂

tests/functional/artifacts/test_previous_version_state.py
```diff
@@ -443,7 +443,7 @@ def test_compare_results_current(self, project):
             current_run_results_schema_version == self.CURRENT_EXPECTED_RUN_RESULTS_VERSION
         ), "Sounds like you've bumped the run_results version and need to update this test!"
         # If we need a newly generated run_results, uncomment the following line and commit the result
-        self.generate_latest_run_results(project, current_run_results_schema_version)
+        # self.generate_latest_run_results(project, current_run_results_schema_version)
```
QMalcolm, line 446: Did I accidentally leave it un-commented from my PR?... The answer is yes, yes I did 😬 🤦 Thank you for catching this.

QMalcolm (commented):
This looks great! Thank you for making the changes. Could you add a changelog entry? I disabled the changelog checker for you while this was a draft PR, but it needs a changelog entry before we merge. Gonna remove the `skip changelog` label now.

QMalcolm (approved):
Looks good to me! Thank you again for doing all this work 🙂

PR #9184, opened 2023-11-30
Enable unit testing in non-root packages

### LINKED ISSUE #8285: [CT-2913] Run unit tests in packages that are not the root package
### Description

If we want to handle running unit tests in packages that are not the root package, we need to make a few changes in the unit testing code.
* Construct a unique_id that uses the non-root package name
* Set the package name in the constructed UnitTestNode

### Acceptance Criteria

Non-root package unit tests are executed if selected.

### Impact to Other Teams

None

No 

### Context

gshank:
resolves #8285

### Problem

When tests run they should also run unit tests in non-root packages, like other tests.

### Solution

Ensure that the "package" in the unit test nodes comes from the right place.

### Checklist


tests/functional/unit_testing/test_ut_dependency.py
```diff
@@ -0,0 +1,84 @@
...
+    def test_unit_test_in_dependency(self, project):
+        run_dbt(["deps"])
+        run_dbt(["seed"])
+        results = run_dbt(["run"])
+        assert len(results) == 2
+
+        results = run_dbt(["test"])
+        assert len(results) == 2
+        unique_ids = get_unique_ids_in_results(results)
+        assert "unit_test.local_dep.dep_model.test_dep_model_id" in unique_ids
```
MichelleArk, line 84: could we add one more command run to this test to ensure selection works on tests in packages?

core/dbt/parser/unit_tests.py
```diff
@@ -227,35 +225,6 @@ def __init__(self, schema_parser: SchemaParser, yaml: YamlBlock) -> None:
         self.schema_parser = schema_parser
         self.yaml = yaml
 
-    def _load_rows_from_seed(self, ref_str: str) -> List[Dict[str, Any]]:
```
MichelleArk, line 230: thank you for moving this down :)

tests/functional/unit_testing/test_ut_dependency.py
gshank: Added some tests.

PR #9199, opened 2023-12-04
Move deferral resolution from `merge_from_artifact` to `RuntimeRefResolver`

jtcohen6:
resolves: https://github.com/dbt-labs/dbt-core/issues/10017
follow-up to #9040

### Problem

In https://github.com/dbt-labs/dbt-core/pull/9040#discussion_r1413279816, @MichelleArk wondered aloud:

> I wonder if it'd actually make more sense to do this deferral + merging business in the `requires.manifest` wrapper to centralize where manifest loading + writing happens.

### Solution

Within `requires.manifest`: Add `defer_relation` to all relevant nodes right after manifest loading

Within `RuntimeRefResolver`: Handle logic for picking the target relation versus `defer_relation`, based on `--favor-state` and adapter cache checks (does the target relation exist in the expected location?)

This seems like a promising way to streamline much of this logic (for all tasks), and move it earlier (outside of tasks entirely).

### Checklist


core/dbt/contracts/graph/nodes.py
```diff
@@ -342,7 +342,6 @@ class ParsedNode(NodeInfoMixin, ParsedNodeMandatory, SerializableType):
     docs: Docs = field(default_factory=Docs)
     patch_path: Optional[str] = None
     build_path: Optional[str] = None
-    deferred: bool = False
```
jtcohen6, line 345: As mentioned, I don't think this is necessary anymore. However:
- This does require updating a lot of tests
- It's exactly the sort of breaking change (removing a field) that we said we would try to avoid doing for artifacts in v1.8+ ...

core/dbt/parser/unit_tests.py
```diff
@@ -131,6 +131,7 @@ def parse_unit_test_case(self, test_case: UnitTestDefinition):
                 "fqn": original_input_node.fqn,
                 "checksum": FileHash.empty(),
                 "raw_code": self._build_fixture_raw_code(given.rows, None),
+                "defer_relation": original_input_node.defer_relation,
```
jtcohen6, line 134: This works in combination with a change to `get_fixture_sql` that prefers to get column types from `defer_relation` if available, otherwise falling back to `this`

https://github.com/dbt-labs/dbt-adapters/pull/60

jtcohen6:
`FAILED tests/functional/unit_testing/test_state.py::TestUnitTestDeferState::test_unit_test_defer_state` is blocked by https://github.com/dbt-labs/dbt-adapters/pull/94.

At the same time, I got myself into quite a tricky loop trying to debug an edge case with dbt unit tests, until I realized that this PR should actually _fix_ that edge case:

core/dbt/artifacts/resources/v1/components.py
```diff
@@ -181,7 +181,6 @@ class ParsedResource(ParsedResourceMandatory):
     docs: Docs = field(default_factory=Docs)
     patch_path: Optional[str] = None
     build_path: Optional[str] = None
-    deferred: bool = False
```
MichelleArk, line 184: Confirming this is a minor change and safe to make here from a runtime serialization/deserialization perspective 👍 

Additionally, we're not using this field for anything functional in dbt-core's implementation, it was previously just accessed to assert certain behaviour had occurred during deferral.

vadim82 (approved):

core/dbt/artifacts/resources/v1/components.py
vadim82: Approving minor schema evolution changes

core/dbt/parser/manifest.py
```diff
@@ -1886,7 +1886,12 @@ def write_manifest(manifest: Manifest, target_path: str, which: Optional[str] =
     write_semantic_manifest(manifest=manifest, target_path=target_path)
 
 
-def parse_manifest(runtime_config, write_perf_info, write, write_json):
+def parse_manifest(
+    runtime_config: RuntimeConfig,
+    write_perf_info: bool,
+    write: bool,
+    write_json: bool,
+) -> Manifest:
```
jtcohen6, line 1894: Before 6b63e31a4eee3272409eae0417cbad7d94b5384a, this was where I was calling `merge_from_artifact`. Now this file diff is just adding type annotations and an inline comment.

core/dbt/task/run.py
```diff
@@ -454,10 +454,10 @@ def print_results_line(self, results, execution_time) -> None:
 
     def before_run(self, adapter, selected_uids: AbstractSet[str]) -> None:
         with adapter.connection_named("master"):
+            self.defer_to_manifest()
```
jtcohen6, line 457: The purpose of this method is now to add `defer_relation` to all nodes in the project.

Previously, we needed to know at this point which objects exist in the target database locations (which requires the cache). Now that's cache check is happening within `RuntimeRefResolver`, meaning that we can consistently call `defer_to_manifest` _before_ `populate_adapter_cache` in every task.

core/dbt/task/runnable.py
```diff
@@ -479,8 +477,8 @@ def populate_adapter_cache(
 
     def before_run(self, adapter, selected_uids: AbstractSet[str]):
         with adapter.connection_named("master"):
+            self.defer_to_manifest()
```
jtcohen6, line 480: See comment on `RunTask.before_run`

MichelleArk (approved):

PR #9211, opened 2023-12-05
Remove ResultNode usage from connections

### LINKED ISSUE #9214: [CT-3454] Remove usage of dbt.contracts.graph.nodes.ResultNode in dbt/adapters

MichelleArk:
resolves #9214 

### Problem

Usage of dbt.contracts.graph.nodes.ResultNode in dbt/adapters
### Solution

Context:
- ResultNode was used as a type annotation on inputs to the connection_named method. 
- This method already wrapped a node in NodeWrapper for safe access - creating a safe layer of abstraction between core's ResultNode and the object necessar

Solution: 
- Pass Any to connection_named instead, renaming `node` to `query_header_context` (since that's what the node is actually used for)
- rename NodeWrapper to QueryContextWrapper - wrapping the `Any` object with a safe accessor method
- remove ResultNode imports

### Checklist

### DIFF
--- a/.changes/unreleased/Under the Hood-20231205-165812.yaml
+++ b/.changes/unreleased/Under the Hood-20231205-165812.yaml
@@ -1,0 +1,6 @@
+kind: Under the Hood
+body: Remove usage of dbt.contracts.graph.nodes.ResultNode in dbt/adapters
+time: 2023-12-05T16:58:12.932172+09:00
+custom:
+  Author: michelleark
+  Issue: "9214"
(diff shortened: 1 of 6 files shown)


core/dbt/common/contracts/config.py
colin-rogers-dbt, line 1: should this file be empty?

colin-rogers-dbt (approved):
small nit on empty config.py file but otherwise LGTM

core/dbt/common/contracts/config.py
MichelleArk: oops! I think this was just left in from some unrelated work locally. will delete

PR #9213, opened 2023-12-05
remove manifest from adapter.set_relations_cache signature

### LINKED ISSUE #9217: remove manifest from adapter.set_relations_cache signature

MichelleArk:
resolves #9217 

### Problem

Setting the relations cache requires an entire manifest currently, even though just a sliver of functionality is necessary for the task. 

### Solution

Pre-compute the manifest nodes to populate the cache for - passing in an iterable of RelationConfigs to the adapter instead of the entire manifest. 

### Checklist


core/dbt/task/runnable.py
```diff
@@ -406,11 +406,21 @@ def populate_adapter_cache(
         if not self.args.populate_cache:
             return
 
+        if self.manifest is None:
+            raise DbtInternalError("manifest was None in populate_adapter_cache")
```
colin-rogers-dbt, line 410: Any idea under what conditions we wouldn't have a manifest?

core/dbt/task/runnable.py
```diff
@@ -406,11 +406,21 @@ def populate_adapter_cache(
...
+
         start_populate_cache = time.perf_counter()
+        # the cache only cares about executable nodes
+        cachable_nodes = [
+            node
+            for node in self.manifest.nodes.values()
+            if (node.is_relational and not node.is_ephemeral_model and not node.is_external_node)
+        ]
```
colin-rogers-dbt, line 418: any advantage to having this be a method or just a calculated attribute of the manifest itself?

gshank (approved):

core/dbt/task/runnable.py
MichelleArk: This was primarily added out of defensiveness &  mypy appeasement , and is not an real expected condition -- it would be our own improper usage of populate_adapter_cache. Since GraphRunnableTask inherits from ConfiguredTask which has an optional manifest, mypy was flagging the None-value of manifest as potentially problematic. the manifest gets set earlier on during GraphRunnableTask execution and should not be None by this point.

MichelleArk:
closing in favor of: https://github.com/dbt-labs/dbt-core/pull/9242

PR #9289, opened 2023-12-14
Move UserConfig to dbt_project.yml and rename to ProjectFlags

### LINKED ISSUE #9183: [CT-3438] `UserConfig` in `profiles.yml` -> `flags` in `dbt_project.yml`
### Housekeeping

### Short description

We are deprecating `config` as a thing that is settable in `profiles.yml`. Instead, we will support `flags` in `dbt_project.yml`

For dbt-core v1.8, we should see if the user has defined `config` in `profiles.yml`.
- If they have, continue to respect it, and raise a deprecation warning.
- If they haven't, great. Instead, we will be reading `flags` from `dbt_project.yml`. The two are mutually exclusive.

In a future release, we will remove `UserConfig` from `profiles.yml` entirely.

### Acceptance criteria
In `dbt_project.yml`, it will look like:
```yml
# dbt_project.yml
flags:
  send_anonymous_usage_stats: True
```

Going forward, some flags will be settable in:
1. "Project flags" (`dbt_project.yml`) _only_. These flags define behaviors tied to the source code in a project, and should remain consistent across all users, environments, and invocations of that project.
2. CLI option / env var _only_. These flags are highly specific to the running invocation, and don't make sense to define in version-controlled code. Several of these are file system paths, such as `--project-dir` or `--log-path`. Some want to vary across commands, such as `--full-refresh`. Environment variables are best suited to flags that want different settings in different environments, such as turning `DBT_FAIL_FAST` on in dev/CI but not in prod.
3. Both places. In this case, a project wants to set a default behavior, but it may be appropriate to override it for different environments/invocations.

If set in both places, the inheritance hierarchy remains the same:
1. CLI var (handled by `click`)
2. Env var (handled by `click`)
3. ~User config~ Project flags

### Implementation Notes

- We should rename the internal `UserConfig` to `ProjectFlags`
- User config is read when [Flags is instantiated here](https://github.com/dbt-labs/dbt-core/blob/main/core/dbt/cli/flags.py#L206). This is probably still the right place to do it, but the logic now becomes: 
   - read_project_flags: 
      - tries to load a user config from profiles.yml. If found, read it and raise a deprecation warning
      - if not found, tries to load user config from dbt_project.yml (can be a minimal yaml parse for just the 'flags' key - no jinja parsing necessary)
      - otherwise, use default user config settings (Nones are appropriate to ensure flag precedence works as expected for flags that can be set via CLI or env var)
- this will also need to be updated in core.dbt.flags.set_from_args

Testing: 
* let's add a unit test that ensures that: every attribute of ProjectFlags that has a corresponding click option in params.py should be set to None by default (except for anon user tracking). Going forward, flags _**can**_ have non-None defaults if they do not have a corresponding CLI option/env var. These will be used to control backwards incompatible interface or behaviour changes.

### Impact to Other Teams

N/A

no :)

### Implications of deprecation

Why was this called `UserConfig` before? Well... hypothetically, if Michelle wants colorized logs, and Jeremy doesn't, they should be able to set user-specific preferences that apply across all their projects, without checking those preferences into those projects' version control (because `profiles.yml` lives in the HOME directory on each machine). While we are removing `UserConfig` from `profiles.yml`, it will remain possible to set user-specific preferences by setting environment variables (in `.bashrc`, `.zshrc`, etc).

gshank:
resolves #9183

### Problem

The flags that are in the current UserConfig object attached to the profile should be in dbt_project.yml.

### Solution

Rename UserConfig to ProjectFlags, and various variables and attributes from "user_config" to "project_flags".  Modify function called in core/dbt/cli/flags.py to retrieve flags to look in both profiles.yml and dbt_project.yml, and issue a deprecation warning if found in profiles.yml and raise an error if found in both places.

Note that we can't use the warn_errors and warn_error_options for this deprecation warning because the deprecation is encountered when loading the project flags which contains warn_errors and warn_error_options.

Remove skipped and unnecessary tests/unit/test_flags.py file.

### Checklist


core/dbt/config/project.py
```diff
@@ -709,7 +715,6 @@ def to_project_config(self, with_packages=False):
                 "exposures": self.exposures,
                 "vars": self.vars.to_dict(),
                 "require-dbt-version": [v.to_version_string() for v in self.dbt_version],
-                "config-version": self.config_version,
```
MichelleArk, line 712: is this just removing some dead code? or are we actively removing support for config-version in dbt_project.yml with this change?

core/dbt/config/project.py
```diff
@@ -773,3 +778,52 @@ def get_macro_search_order(self, macro_namespace: str):
...
+
+def read_project_flags(project_dir: str, profiles_dir: str) -> ProjectFlags:
+    try:
+        project_flags: Dict[str, Any] = {}
+        # Read project_flags from dbt_project.yml first
+        # Flags are instantiated before the project, so we don't
+        # want to throw an error for non-existence of dbt_project.yml here
+        # because it breaks things.
+        project_root = os.path.normpath(project_dir)
+        project_yaml_filepath = os.path.join(project_root, "dbt_project.yml")
```
MichelleArk, line 791: nit: could use dbt.constants.DBT_PROJECT_FILE_NAME here instead

core/dbt/config/project.py
gshank: We removed support for config-version sometime last year, but a few things were still hanging around.  It still exists in the class that validates because we don't want to cause a validation error, but there's no point in passing it along.

core/dbt/common/events/types.proto
```diff
@@ -280,6 +280,454 @@ message ProjectCreatedMsg {
     ProjectCreated data = 2;
 }
 
+// E - DB Adapter
+
+// E001
+message AdapterEventDebug {
```
colin-rogers-dbt, line 286: why move adapter events into common?

core/dbt/config/project.py
gshank: Fixed

core/dbt/common/events/types.proto
gshank: I didn't. Don't know why that's showing up as a change. I just merged main into the branch... Git might have gotten confused?

gshank: It looks like there was a merge error which needed to be resolved but wasn't, and nothing failed because types.proto wasn't rebuilt...

tests/unit/test_cli_flags.py
```diff
@@ -404,3 +406,38 @@ def test_from_dict_0_value(self):
         args_dict = {"log_file_max_bytes": 0}
         flags = Flags.from_dict(Command.RUN, args_dict)
         assert flags.LOG_FILE_MAX_BYTES == 0
+
+
+def test_project_flag_defaults():
```
MichelleArk, line 411: nice!

MichelleArk (approved):

colin-rogers-dbt (approved):

PR #9296, opened 2023-12-15
update dbt-extractor version

ChenyuLInx:
Seems like the tests would fail with dbt-core extractor 0.5.0, `tests/functional/retry/test_retry.py` is an example locally.

### DIFF
--- a/core/setup.py
+++ b/core/setup.py
@@ -71,7 +71,7 @@ setup(
         "sqlparse>=0.2.3,<0.5",
         # ----
         # These are major-version-0 packages also maintained by dbt-labs. Accept patches.
-        "dbt-extractor~=0.5.0",
+        "dbt-extractor~=0.5.1",
         "minimal-snowplow-tracker~=0.0.2",
         "dbt-semantic-interfaces~=0.5.0a2",
         # ----


emmyoop:
@ChenyuLInx I'm not sure what the issue is here?  Tests in core are currently passing.  The 0.5.1 distribution just included the source files I believe.

ChenyuLInx:
@emmyoop So I noticed this because I had extractor 0.5.0 installed previously, and did `make dev` with main branch. But afterwards the functional test listed would fail. And manually install extractor 0.5.1 would fix the problem.

core/setup.py
```diff
@@ -71,7 +71,7 @@
         "sqlparse>=0.2.3,<0.5",
         # ----
         # These are major-version-0 packages also maintained by dbt-labs. Accept patches.
-        "dbt-extractor~=0.5.0",
+        "dbt-extractor~=0.5.1",
```
emmyoop, line 74: the `~=` installs the latest patch for the `0.5` release so this should be installing `0.5.1` already.

emmyoop (commented):
I still don't think this is necessary.  I just ran `make-dev` and it installs `0.5.1` as expected.

PR #9313, opened 2023-12-22
Support custom expressions for incremental merge updates

### LINKED ISSUE #9312: [CT-3511] [Feature] Custom merge update expressions on incremental models

### Describe the feature

The current implementation of the `merge` incremental strategy always overwrites all columns whenever there are new rows with an existing `unique_key`. However, in many cases custom update expressions are necessary, especially when the previous value of a particular column needs to be taken into account. 

As an example, consider the following table (`orders`):

|order_id|user_id|order_ts|
|---|---|---|
|1|A|2023-12-14 10:01:00|
|2|A|2023-12-14 10:02:00|
|3|B|2023-12-14 10:03:00|
|4|B|2023-12-14 10:04:00|
|5|B|2023-12-14 10:05:00|
|6|A|2023-12-15 10:06:00|
|7|A|2023-12-15 10:07:00|

<br>

Let's assume we want to keep track of the number of orders per user and the timestamp of the last order from each user. The naive approach would be to group by `user_id` and aggregate:

<br>

```sql
{{ config(materialized='table') }}

select
    user_id,
    count(order_id) as order_count,
    max(order_ts) as last_order_ts
from orders
group by user_id
order by user_id
```

|user_id|order_count|last_order_ts|
|---|---|---|
|A|4|2023-12-15 10:07:00|
|B|3|2023-12-14 10:05:00|

<br>

As the number of orders and users grows by orders of magnitude, aggregating the entire table becomes more and more expensive and redundant, in which case the most straightforward solution is to switch to an `incremental` materialization.

<br>

```sql
{{ config(
    materialized='incremental',
    incremental_strategy='merge',
    unique_key='user_id'
)}}

select
    user_id,
    count(order_id) as order_count,
    max(timestamp) as last_order_ts
from orders
{%- if is_incremental() %}
where order_ts > ( select max(last_order_ts) from {{this}} )
{%- endif %}
group by user_id
order by user_id
```

<br>

This approach would work well for the `last_order_ts` field but... how about the `order_count` field?

It's clear the `MERGE INTO` statement needs to add the existing value of `order_count` with the one from the incoming chunk, otherwise the field will be overwritten solely with the `order_count` from the incoming chunk, which won't reflect reality.

To tackle these use cases, I propose adding an extra config option `merge_update_expressions` that applies only to incremental models with the `merge` strategy:

```sql
{{ config(
    materialized='incremental',
    incremental_strategy='merge',
    unique_key='user_id',
    merge_update_expressions={
        'order_count': 'DBT_INTERNAL_DEST.order_count + DBT_INTERNAL_SOURCE.order_count'
    }
)}}

select
    user_id,
    count(order_id) as order_count,
    max(timestamp) as last_order_ts
from orders
{%- if is_incremental() %}
where order_ts > ( select max(last_order_ts) from {{this}} )
{%- endif %}
group by user_id
order by user_id
```

<br>

I concede referring to `DBT_INTERNAL_DEST` and `DBT_INTERNAL_SOURCE` directly might not be the most elegant syntax. Nonetheless, it would be consistent with the [incremental_predicates](https://docs.getdbt.com/docs/build/incremental-models#about-incremental_predicates) option where these internal variables are referred to as well.

_No response_

### Who will this benefit?

_No response_

Yes

### Anything else?

Tangentially related to this, I've noticed https://github.com/dbt-labs/dbt-core/issues/9223
The latter seems to be more focused on the predicates/matching rather that the update expressions themselves, but both would be complementary

monti-python:
resolves #9312

### Checklist

### DIFF
--- a/core/dbt/include/global_project/macros/materializations/models/incremental/merge.sql
+++ b/core/dbt/include/global_project/macros/materializations/models/incremental/merge.sql
@@ -9,6 +9,7 @@
     {%- set dest_cols_csv = get_quoted_csv(dest_columns | map(attribute="name")) -%}
     {%- set merge_update_columns = config.get('merge_update_columns') -%}
     {%- set merge_exclude_columns = config.get('merge_exclude_columns') -%}
+    {%- set merge_update_expressions = config.get('merge_update_expressions') -%}
     {%- set update_columns = get_merge_update_columns(merge_update_columns, merge_exclude_columns, dest_columns) -%}
     {%- set sql_header = config.get('sql_header', none) -%}
 
@@ -39,7 +40,11 @@
     {% if unique_key %}
     when matched then update set
         {% for column_name in update_columns -%}
-            {{ column_name }} = DBT_INTERNAL_SOURCE.{{ column_name }}
+            {%- if column_name in merge_update_expressions %}
+                {{ column_name }} = {{ merge_update_expressions[column_name] }}
+            {%- else %}
+                {{ column_name }} = DBT_INTERNAL_SOURCE.{{ column_name }}
+            {%- endif %}
             {%- if not loop.last %}, {%- endif %}
         {%- endfor %}
     {% endif %}


dbeatty10:
Closing per https://github.com/dbt-labs/dbt-core/issues/9312#issuecomment-1885333505

PR #9322, opened 2023-12-31
added hook support for dbt source freshness

### LINKED ISSUE #5609: [CT-985] [Feature] on-run-end hook and results variable support for source freshness command

### Describe the feature

- Currently `dbt source freshness` does not support running an [on-run-end hook](https://docs.getdbt.com/reference/project-configs/on-run-start-on-run-end). 
- Other dbt commands like `dbt run`, `dbt test`, `dbt snapshot` and more already support running an on-run-end hook and [getting results jinja variable](https://docs.getdbt.com/reference/dbt-jinja-functions/on-run-end-context) as their context. 
- Using an on-run-end hook and the results variable in the `dbt source freshness` command, we can upload the source freshness results to the data warehouse. This could be used for displaying source freshness results on a dashboard, monitoring, and alerting.

- The `dbt source freshness` command already writes the freshness results to a file called sources.json. 
- A possible alternative is to process this json file to get the source freshness results. 
- This solution requires implementing an external solution that will process this json and automatically upload the results to the data warehouse after every execution of the `dbt source freshness` command. 
- This alternative is harder to implement and requires implementing code outside of the dbt project. 
- Additionally, this solution requires orchestrating this new tool and giving it access to the same project so it can extract the source freshness results from the sources.json file.
- Support for on-run-end and results variable will simplify this dramatically, enabling it to be implemented in the dbt project itself.

### Who will this benefit?

- dbt users who rely on on-run-end hook and the results variable to monitor their dbt commands, executions and results.
- dbt users who want to build dashboards to present their source freshness results.
- dbt users who rely on dbt packages like [Elementary](https://github.com/elementary-data/dbt-data-reliability) to upload their dbt artifacts and monitor their dbt projects. 

Yes

### Anything else?

_No response_

ofek1weiss:
resolves #5609

### Problem
Currently, `on-run-*` hooks are not supported in the `dbt source freshness` command, this PR makes it so it does.

### Solution
Changed the `FreshnessTask` class to inherit from `RunTask` (as all other task types that support hooks do).

Alternative solutions:
* Move the hook logic to `GraphRunnableTask` and disabling in `CompileTask` - feels a little messier (disabling parent functionality in inheriter).
* Create a `HookTask` in the middle of the inheritance tree - seems like a lot more work when every other hook-able task just inherits from `RunTask`.

### Checklist

### DIFF
--- a/.changes/unreleased/Features-20231231-171205.yaml
+++ b/.changes/unreleased/Features-20231231-171205.yaml
@@ -1,0 +1,6 @@
+kind: Features
+body: Added hook support for `dbt source freshness`
+time: 2023-12-31T17:12:05.587185+02:00
+custom:
+  Author: ofek1weiss
+  Issue: "5609"
--- a/core/dbt/contracts/results.py
+++ b/core/dbt/contracts/results.py
@@ -213,7 +213,6 @@ class RunResultOutput(BaseResult):
 
 
 def process_run_result(result: RunResult) -> RunResultOutput:
-
     compiled = isinstance(result.node, CompiledNode)
 
     return RunResultOutput(
@@ -282,7 +281,8 @@ class RunResultsArtifact(ExecutionResult, ArtifactMixin):
     @classmethod
     def upgrade_schema_version(cls, data):
         """This overrides the "upgrade_schema_version" call in VersionedSchema (via
-        ArtifactMixin) to modify the dictionary passed in from earlier versions of the run_results."""
+        ArtifactMixin) to modify the dictionary passed in from earlier versions of the run_results.
+        """
         run_results_schema_version = get_artifact_schema_version(data)
         # If less than the current version (v5), preprocess contents to match latest schema version
         if run_results_schema_version <= 5:
@@ -398,12 +398,12 @@ class FreshnessMetadata(BaseArtifactMetadata):
 @dataclass
 class FreshnessResult(ExecutionResult):
     metadata: FreshnessMetadata
-    results: Sequence[FreshnessNodeResult]
+    results: Sequence[Union[FreshnessNodeResult, BaseResult]]
 
     @classmethod
     def from_node_results(
         cls,
-        results: List[FreshnessNodeResult],
+        results: List[Union[FreshnessNodeResult, BaseResult]],
         elapsed_time: float,
         generated_at: datetime,
     ):
@@ -426,7 +426,9 @@ class FreshnessExecutionResultArtifact(
 
     @classmethod
     def from_result(cls, base: FreshnessResult):
-        processed = [process_freshness_result(r) for r in base.results]
+        processed = [
+            process_freshness_result(r) for r in base.results if isinstance(r, FreshnessNodeResult)
+        ]
         return cls(
             metadata=base.metadata,
             results=processed,
--- a/core/dbt/task/freshness.py
+++ b/core/dbt/task/freshness.py
@@ -7,7 +7,7 @@ from .base import BaseRunner
 from .printer import (
     print_run_result_error,
 )
-from .runnable import GraphRunnableTask
+from .run import RunTask
 
 from dbt.contracts.results import (
     FreshnessResult,
@@ -170,7 +170,7 @@ class FreshnessSelector(ResourceTypeSelector):
         return node.has_freshness
 
 
-class FreshnessTask(GraphRunnableTask):
+class FreshnessTask(RunTask):
     def defer_to_manifest(self, adapter, selected_uids):
         # freshness don't defer
         return
--- a/core/dbt/task/freshness.py
+++ b/core/dbt/task/freshness.py
@@ -23,6 +23,7 @@ from dbt.common.events.types import (
     LogFreshnessResult,
     Note,
 )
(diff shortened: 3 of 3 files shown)


ofek1weiss:
@cla-bot check

ofek1weiss:
@cla-bot check

ofek1weiss:
@ChenyuLInx @jtcohen6 does this solution seem fine to you, is there any reason not to inherit from `RunTask`?

jtcohen6:
@ofek1weiss Thanks for picking up the issue, and for opening the PR! I've added this to our review list. I also kicked off tests to run.

The change you're proposing is elegant & simple :)

This will have the effect of running `on-run-start/end` hooks before/after source freshness checks, unless users filter on `flags.WHICH == 'freshness'` — though that's exactly the intended change of this issue/PR, and it's not for any good or opinionated reason that the `freshness` task was previously excluded.

TODOs during code review:
- Decide if we need to gate the intended behavior change — hooks running during `dbt source freshness` — behind a project flag that would offer back-compatibility for the previous behavior
- Think through any other potential side-effects of having `FreshnessTask` inherit from `RunTask` (methods on the latter that weren't previously implemented on the former)

MichelleArk:
Hey @ofek1weiss, thanks again for opening this PR! @jtcohen6 and I had a chance to review it today with the two points from the comment above in mind:

> Think through any other potential side-effects of having FreshnessTask inherit from RunTask (and thereby `CompileTask`)

We went through the new methods that FreshnessTask would now be inheriting from `RunTask`, and found that most of them are safely overwritten in FreshnessTask already, but a handful were new. tl;dr -- the switch to `RunTask` looks like it won't introduce any unintended behaviour. Details on updated/new methods as a result of the change to `RunTask`:
  * CompileTask._runtime_initialize
      * This is safe because it is gated by a check for `args.inline` — which is not supported for `dbt freshness` and so shouldn’t introduce any new//unintended behaviour
  * CompileTask._handle_result:
        * This is also a no-op. It calls the `GraphRunnable._handle_result` and then does some additional processing for ephemeral nodes, which should not be reachable for the freshness task. 
  * RunTask.index_offset: 
      * This does something clever wrt to choosing nodes of the run queue based on the number of downstream nodes. But the node queue containing for `dbt freshness` is flat (all independent sources), so this should have no effect. 
  * RunTask.print_results_line
      * This was previously a no-op in GraphRunnableTask, but now emits a `FinishedRunningState` event as implemented in RunTask. @jtcohen6  ran this locally to see the effect, and decided it was acceptable (& preferable!) from a product perspective to start including this log-line to source freshness going forward: 

```sh
$ dbt source freshness
15:57:51  Running with dbt=1.8.0-a1
15:57:52  Registered adapter: postgres=1.8.0-a1
15:57:52  Found 4 models, 1 test, 3 sources, 0 exposures, 0 metrics, 538 macros, 0 groups, 0 semantic models
15:57:52
15:57:52  Concurrency: 5 threads (target='dev')
15:57:52
15:57:52  1 of 3 START freshness of dbt_jcohen.another_tbl ............................... [RUN]
...
15:57:52  Finished running 3 sources in 0 hours 0 minutes and 0.43 seconds (0.43s).  <--- this is the new line!!!
15:57:52
...
```

----- 
> Decide if we need to gate the intended behavior change — hooks running during dbt source freshness — behind a project flag that would offer back-compatibility for the previous behavior

We concluded this change _should_ be gated behind a project flag. Otherwise, the default behaviour of running source freshness will change in a way that could introduce unexpected performance regressions or cost spikes between minor versions of dbt. For example, something called `run_hooks_freshness: bool = False`, which gets configured at the project-level to preserve the existing behaviour.

Conveniently, we've recently reworked flags / user configs just for this purpose as part of https://github.com/dbt-labs/dbt-core/issues/9183. It's still a very new pattern in dbt-core, and unfortunately there aren't great examples to point you to yet. For that reason, I'd be happy to take this PR over to implement this and we can use it as a reference for future contributions. Same goes for adding tests to this work because we'd need to write tests for either flag value, and we don't yet have much documentation / examples on how to do so.

ofek1weiss:
Hi @MichelleArk Amazing! thank you so much.
How long do you think it will take? will it be a part of the next dbt release? I'd be happy to assist in any way i can to make sure it does :)

MichelleArk:
@ofek1weiss - I anticipate I'll get to it over the next couple weeks. We're [planning to release the next dbt version (1.8) in April ](https://github.com/dbt-labs/dbt-core/blob/main/docs/roadmap/2023-11-dbt-tng.md#the-next-six-months-stability--unit-testing)so I'm pretty confident at this point that this change will make it into the next release!

martynydbt:
Just wanted to chime in and thank you for submitting the PR to improve dbt-core!

MichelleArk:
@ofek1weiss - As an update, we've continued to take this work across the finish line with flag support + tests here: https://github.com/dbt-labs/dbt-core/pull/9366.

MichelleArk:
Hey @ofek1weiss - we merged your changes in as part of https://github.com/dbt-labs/dbt-core/pull/9366 today! So it should be available as part of the next dbt-core release in the spring.

Issue #9152, opened 2023-11-27
[CT-3425] Move common events only used in core to core


MichelleArk:

MichelleArk:
Note: We should probably refactor the test_events unit test as part of this work so that there is a separate test file for common, adapters, and core tests

PR #9326, opened 2024-01-03
move events only used by core to dbt/events

MichelleArk:
resolves https://github.com/dbt-labs/dbt-core/issues/9152

### Problem
The common subdirectly is being moved to a separate repo (dbt-common) shortly. Currently most events are defined in common, regardless of their usage. A large chunk of these only have usage in core, and should be moved out of common to avoid increasing development overhead (coordinating across multiple repos) when making changes to events only used in core. 

### Solution

Audit which events defined in common are only used in core (and not by common or adapters), and move those definitions to core. 

### Checklist

Ensuring we haven't regressed on cyclical dependencies: 
```
❯ python3 scripts/migrate-adapters.py core/dbt/adapters

migration progress: 100.0% of dbt imports are valid (from adapters or common)
remaining core imports: 0
```

(also was keeping an eye out as I was changing import paths)


core/dbt/common/helper_types.py
```diff
@@ -67,9 +68,13 @@ def _validate_items(self, items: List[str]):
...
 class WarnErrorOptions(IncludeExclude):
     def _validate_items(self, items: List[str]):
-        valid_exception_names = set(
+        valid_common_exception_names = set(
             [name for name, cls in dbt_event_types.__dict__.items() if isinstance(cls, type)]
         )
+        valid_core_exception_names = set(
+            [name for name, cls in core_dbt_event_types.__dict__.items() if isinstance(cls, type)]
+        )
+        valid_exception_names = valid_common_exception_names.union(valid_core_exception_names)
```
MichelleArk, line 77: thinking about this more.. I think we should also explicitly include adapter warnings here + add a test for it. 

we document the ability to use warn error options with an adapter event here: https://docs.getdbt.com/reference/global-configs/warnings

colin-rogers-dbt (approved):

Issue #9112, opened 2023-11-17
[CT-3401] [Bug] dbt retry fails to apply --full-refresh to downstream models and vars


lennons301:
### Is this a new bug in dbt-core?

- [X] I believe this is a new bug in dbt-core
- [X] I have searched the existing issues, and I could not find an existing issue for this bug

### Current Behavior

`dbt retry` fails to pass `--full-refresh` flag from the original invocation to all models in the retry DAG after the first one

Haven't used `dbt retry` in earlier versions so unable to say if this is a regression or not

### Expected Behavior

`dbt retry` creates a new DAG starting from the point of failure, continuing with everything downstream

All flags/args passed in the previous invocation should be applied to all models in the retry DAG

### Steps To Reproduce

1. Create 2 incremental models. The second model should reference the first. In both models log the value of `is_incremental` with `{{ log('is_incremental is' ~ is_incremental(), info=True)}}`

2. Create a compilation error in model 1 and invoke `dbt run --select +model_2 --full-refresh` 

This should set `is_incremental()` to `False` and

- produce a compilation error in model 1
- skip model 2

3. Correct your compilation error in model 1 and invoke `dbt retry`

Model one will now run, logging `is_incremental` = `False` (expected behaviour)
Model two will now run, logging `is_incremental `= `True` (not desired behaviour)

### Relevant log output

```shell
'''
dbt retry
10:49:37  Running with dbt=1.6.0
10:49:37  Registered adapter: trino=1.6.0
10:49:37  is_incremental isFalse
10:49:37  Found 2 models
10:49:37  
10:49:38  Concurrency: 1 threads (target='local')
10:49:38  
10:49:38  1 of 2 START sql incremental model model_1  [RUN]
10:49:38  is_incremental isFalse
10:49:41  1 of 2 OK created sql incremental model model_1 [SUCCESS in 2.77s]
10:49:41  2 of 2 START sql incremental model model_2  [RUN]
10:49:41  is_incremental isTrue
'''
```

### Environment

```markdown
- OS: Ubuntu 20.04.6
- Python:3.10.13
- dbt:1.6.0
```

### Which database adapter are you using with dbt?

other (mention it in "Additional Context")

### Additional Context

Using `dbt-trino 1.6.0`

graciegoheen:
Notes from refinement:
- we decided that you should be able to override the path configs the the `retry` command
- for every other flag though, whatever was the setting for the original invocation, use the same flags for `retry`
[...]

Issue #9325, opened 2024-01-02
consistent policy around cli args, env vars, etc. for `dbt retry`


graciegoheen:
### Housekeeping

- [X] I am a maintainer of dbt-core

### Short description

blah

### Acceptance criteria

blah

### Impact to Other Teams

none

### Will backports be required?

none

graciegoheen:
Notes from technical refinement:
- we may not currently have a way to determine _how_ a given arg was set (via CLI, env var, etc.)

anentropic:
I'm interested in scenario 1

I see older issue https://github.com/dbt-labs/dbt-core/issues/9112 was closed as fixed, and also https://github.com/dbt-labs/dbt-core/pull/9328

But I'm on 1.11 and `dbt retry --full-refresh` doesn't seem to work, I just get `dbt retry` behaviour i.e. it fails again and I seem to have to do `dbt build --full-refresh --select <model>` manually for the model(s) which failed

My context is models with config:

```
        materialized='incremental',
        on_schema_change='fail',
        incremental_strategy='merge'
```

TBH what I'd really like is an `on_schema_change='full_refresh'` option, given that currently:

> None of the `on_schema_change` behaviors backfill values in old records for newly added columns

PR #9328, opened 2024-01-04
Fix full-refresh and vars for retry

### LINKED ISSUE #9327: [CT-3518] [Bug] Retry should not print out warning when state path is not specified

### Current Behavior

We decided to add a warning when `--state path` == `--target-path` https://github.com/dbt-labs/dbt-core/issues/8160 

This would cause dbt core raise a warning in retry when dbt tries to load the previous state. But actually dbt do not need to load all previous state in this case and should not raise such warning

### Expected Behavior

We should only load the previous run results, and not print out a warning.

### Steps To Reproduce

...

### Relevant log output

_No response_

### Environment

```markdown
- OS:
- Python:
- dbt:
```

_No response_

### Additional Context

_No response_

ChenyuLInx:
resolves #9327 and #9112

This patch restructures retry so that it resolves all flags and config first, then does the parsing and execution.
It also changes to only load the last run results since that's the only thing needed for retry.


core/dbt/task/retry.py
```diff
@@ -28,6 +32,7 @@
     "defer_state",
     "deprecated_state",
     "target_path",
+    "warn_error",
```
ChenyuLInx, line 35: @graciegoheen how do we feel about `warn_error` being a thing that's being carried over?
Current behavior:
- `dbt run --warn-error` failed because of some warning seen as error
- 'dbt retry' will not treat those warnings as error
- 'dbt retry --warn-error' will be the way to still treat warnings as error.
This is the previous behavior we have. Do we think this is good? or do we want to change it?

ChenyuLInx: @peterallenwebb I think we talked about this became a behavior change in my version. But in the latest version this is actually not.

core/dbt/task/retry.py
```diff
@@ -57,59 +62,61 @@
...
         previous_args = {
-            k: v for k, v in self.previous_args.items() if k not in OVERRIDE_PARENT_FLAGS
+            k: v for k, v in self.previous_args.items() if k not in IGNORE_PARENT_FLAGS
+        }
+        click_context = get_current_context()
+        current_args = {
+            k: v
+            for k, v in args.__dict__.items()
+            if k in IGNORE_PARENT_FLAGS
+            or click_context.get_parameter_source(k) == ParameterSource.COMMANDLINE
```
ChenyuLInx, line 99: This is a key change, any parameter from the command line would overwrite the previous set parameter.
If it is not from command line and not in the `IGNORE_PARENT_FLAGS`, we will reuse things from previous invocation

core/dbt/task/retry.py
```diff
@@ -57,59 +62,61 @@
...
 class RetryTask(ConfiguredTask):
     def __init__(self, args, config, manifest) -> None:
-        super().__init__(args, config, manifest)
-
-        state_path = self.args.state or self.config.target_path
-
-        if self.args.warn_error:
-            RETRYABLE_STATUSES.add(NodeStatus.Warn)
-
-        self.previous_state = PreviousState(
```
ChenyuLInx, line 67: We do not need the full previous state here, just run results.
Loading the whole state could be memory intensive for large projects(manifest can get large.)

core/dbt/task/retry.py
```diff
@@ -17,9 +20,10 @@
 from dbt.task.seed import SeedTask
 from dbt.task.snapshot import SnapshotTask
 from dbt.task.test import TestTask
+from dbt.parser.manifest import parse_manifest
 
 RETRYABLE_STATUSES = {NodeStatus.Error, NodeStatus.Fail, NodeStatus.Skipped, NodeStatus.RuntimeErr}
-OVERRIDE_PARENT_FLAGS = {
+IGNORE_PARENT_FLAGS = {
```
ChenyuLInx, line 26: Updating the name to better reflect what this list is.

core/dbt/task/retry.py
peterallenwebb: @ChenyuLInx Doesn't this mean that any parameter supplied on the command line would be used? That seems different than what we discussed with Grace, which was that only a limited set of parameters could be overriden when using retry.

ChenyuLInx: Thanks for calling it out. Updated.

peterallenwebb: The code now looks better to me, but I think it also means we need to add some more parameters to the ALLOW_OVERRIDE_ARGS list. In particular the --target and --profile flags are overridden in production usage of dbt, as you can see using [this Datadog query](https://dbtlabsmt.datadoghq.com/logs?query=%22dbt%20client%20called%20with%22%20retry%20&cols=host%2Cservice&index=%2A&messageDisplay=inline&refresh_mode=sliding&stream_sort=desc&viz=stream&from_ts=1704403160572&to_ts=1704404060572&live=true).

peterallenwebb: A [different Datadog query](https://dbtlabsmt.datadoghq.com/logs?query=%22cli%20with%20the%20following%20command%22%20retry%20&cols=host%2Cservice&event=AgAAAYzWaiMFv2_IbAAAAAAAAAAYAAAAAEFZeldhaWo5QUFCVVpWR3pLS2hZRWdBTwAAACQAAAAAMDE4Y2Q2NmItN2UzZi00NGJhLTgzNWUtYTk5OWMwMmNkZTI1&index=%2A&messageDisplay=inline&refresh_mode=sliding&stream_sort=desc&viz=stream&from_ts=1704403160572&to_ts=1704404060572&live=true) shows that --log-format, --debug, and --profiles-dir are also specified sometimes in production.

ChenyuLInx: So we actually have two list now, and they behave slightly different.
There's a `IGNORE_PARENT_FLAGS` whose behavior is: no matter what the last command set, we actually don't try to reuse them, but just use the new ones,
then there is the `ALLOW_OVERRIDE_ARGS`, which is : use whatever the command set, unless user specify something new in the retry command.

Is the name not obvious enough? should I change to different name?

peterallenwebb: I think it would be slightly clearer if we called it ALLOW_CLI_OVERRIDE_FLAGS, but upon further review this looks okay.

peterallenwebb (approved):

core/dbt/task/retry.py
ChenyuLInx: Updated!

core/dbt/task/retry.py
graciegoheen: > - dbt run --warn-error failed because of some warning seen as error
> - 'dbt retry' will not treat those warnings as error

This is surprising to me. It ties back to the desire to have a [consistent](https://github.com/dbt-labs/dbt-core/issues/9325) policy here. I would assume that **all** flags passed to the original command are pulled through to `dbt retry`. Is this a specific exception for `--warn-error`? Ultimately, if we're worried about breaking something in a backport, we could address this for `1.8`.

ChenyuLInx: So `warn-error` is a flag, so if we don't ignore it in retry there's actually no way to specify `no-warn-error` in retry.


> dbt run --warn-error failed because of some warning seen as error
'dbt retry' will not treat those warnings as error

Just to be super clear this is the current behavior of retry.

graciegoheen:
https://github.com/dbt-labs/dbt-core/pull/9391

PR #9338, opened 2024-01-04
Decouple warn error options from core and adapters

### LINKED ISSUE #9337: [CT-3526] Remove import of dbt.adapters.events.types and dbt.events.types from dbt/common

MichelleArk:
resolves #9337 

### Problem

WarnErrorOptions is now defined in dbt/common, which is being moved to a separate upstream package/repo (dbt-common) shortly. It is defined in dbt/common because it is used both by dbt-common and dbt-core. Unfortunately, it depends on both dbt-core and dbt-adapter to import the event type module to validate items in its collection. This would represent a circular dependency once moved to a separate repo as dbt-common should be upstream to both dbt-adapter and dbt-core.

### Solution
* update WarnErrorOptions constructor to accept `valid_error_names` on instantiation, for use in validation
* provide `valid_error_names` during `WarnErrorOptions` construction in dbt-core with `ALL_EVENT_NAMES` (combines adapter+common+core events)
* moved test_helper_types unit tests to tests/common since `helper_types` is defined in common and only has dbt-common dependencies!

### Checklist

### DIFF
--- a/.changes/unreleased/Under the Hood-20240104-165248.yaml
+++ b/.changes/unreleased/Under the Hood-20240104-165248.yaml
@@ -1,0 +1,7 @@
+kind: Under the Hood
+body: Accept valid_error_names in WarnErrorOptions constructor, remove global usage
+  of event modules
+time: 2024-01-04T16:52:48.173716-05:00
+custom:
+  Author: michelleark
+  Issue: "9337"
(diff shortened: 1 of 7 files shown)


emmyoop (approved):
Looks good.

colin-rogers-dbt (approved):

PR #9345, opened 2024-01-06
Add flags columns and truncate for dbt show

touchmarine:
### Problem

`dbt show` is a useful tool for debugging your queries and as a quick preview. As such, it would be useful to be able to:
- print an untruncated table (with more than 6 columns and columns wider than 20 chars)
- specify which columns to print

### Solution

This PR adds two CLI flags for the `dbt show` command:

1. `--columns` - allows specifying columns to return
2. `--truncate/--no-truncate` - allows disabling the default behaviour of truncating the table (number of columns, width of columns, and the max precision of numbers)

## User Docs

The [dbt show docs](https://docs.getdbt.com/reference/commands/show) should be updated to reflect the new flags.

### Checklist

### DIFF
--- a/core/dbt/cli/params.py
+++ b/core/dbt/cli/params.py
@@ -39,6 +39,16 @@ introspect = click.option(
     default=True,
 )
 
+columns = click.option(
+    "--columns",
+    envvar=None,
+    help="Space-delimited list of columns returned by dbt show",
+    type=tuple,
+    cls=MultiOption,
+    multiple=True,
+    default=[],
+)
+
 compile_docs = click.option(
     "--compile/--no-compile",
     envvar=None,
--- a/core/dbt/cli/params.py
+++ b/core/dbt/cli/params.py
@@ -608,6 +608,16 @@ threads = click.option(
     type=click.INT,
 )
 
+truncate = click.option(
+    "--truncate/--no-truncate",
+    envvar=None,
+    help=(
+        "Truncate the columns returned by dbt show "
+        "(number of columns, width of columns, and max precision of numbers)"
+    ),
+    default=True,
+)
+
 use_experimental_parser = click.option(
     "--use-experimental-parser/--no-use-experimental-parser",
     envvar="DBT_USE_EXPERIMENTAL_PARSER",
--- a/.changes/unreleased/Features-20240106-210614.yaml
+++ b/.changes/unreleased/Features-20240106-210614.yaml
@@ -1,0 +1,6 @@
+kind: Features
+body: Add --columns and --truncate flags for dbt show
+time: 2024-01-06T21:06:14.998544+01:00
+custom:
+  Author: touchmarine
+  Issue: "9345"
--- a/core/dbt/cli/main.py
+++ b/core/dbt/cli/main.py
@@ -378,6 +378,7 @@ def compile(ctx, **kwargs):
 @cli.command("show")
 @click.pass_context
 @global_flags
+@p.columns
 @p.defer
 @p.deprecated_defer
 @p.exclude
--- a/core/dbt/cli/main.py
+++ b/core/dbt/cli/main.py
@@ -401,6 +401,7 @@ def compile(ctx, **kwargs):
 @p.target
 @p.target_path
 @p.threads
+@p.truncate
 @p.vars
 @requires.postflight
 @requires.preflight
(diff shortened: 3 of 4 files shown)


core/dbt/cli/params.py
```diff
@@ -39,6 +39,16 @@
     default=True,
 )
 
+columns = click.option(
```
joshuataylor, line 42: When using Snowflake, it seems the columns need to be in capitals, which is also how the columns are output anyway to the screen, but would be good to note that columns might need to be capitalised, atleast until it's fixed in adapters (if needed(?).

When using lower case:
```sh
16:00:58  Encountered an error:
tuple.index(x): x not in tuple
16:00:58  Traceback (most recent call last):
  File "dbt-core/core/dbt/cli/requires.py", line 91, in wrapper
    result, success = func(*args, **kwargs)
                      ^^^^^^^^^^^^^^^^^^^^^
  File "dbt-core/core/dbt/cli/requires.py", line 76, in wrapper
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "dbt-core/core/dbt/cli/requires.py", line 169, in wrapper
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "dbt-core/core/dbt/cli/requires.py", line 198, in wrapper
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "dbt-core/core/dbt/cli/requires.py", line 245, in wrapper
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "dbt-core/core/dbt/cli/requires.py", line 278, in wrapper
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "dbt-core/core/dbt/cli/main.py", line 420, in show
    results = task.run()
              ^^^^^^^^^^
  File "dbt-core/core/dbt/task/runnable.py", line 493, in run
    self.task_end_messages(result.results)
  File "dbt-core/core/dbt/task/show.py", line 88, in task_end_messages
    table = table.select(self.args.columns)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "venvs/dbtsnowflake/.venv/lib/python3.11/site-packages/agate/table/select.py", line 21, in select
    indexes = tuple(self._column_names.index(k) for k in key)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "venvs/dbtsnowflake/.venv/lib/python3.11/site-packages/agate/table/select.py", line 21, in <genexpr>
    indexes = tuple(self._column_names.index(k) for k in key)
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^
ValueError: tuple.index(x): x not in tuple
```

joshuataylor:
I've been playing with this, and it works really would. Was honestly surprised it wasn't out of the box with the release, and I really like your implementation.

core/dbt/cli/params.py
joshuataylor: 🤔 , just had a thought.

Just tried this, it works, but not sure if the variable should be moved out or it's fine to define here.

```python
                # convert columns to uppercase to match agate's behavior
                table = table.select([col.upper() for col in self.args.columns])
```

touchmarine: Uppercasing all columns would break other databases. We could accept case-insensitve columns and map them to exact column names using `table.colum_names`. But that wouldn't work for people using case-sensitive column names.

To avoid the error we can filter out unknown columns (which includes incorrect casing) and add a note that exact column names are required:

```python
valid_columns = [col for col in self.args.columns if col in table.column_names]
table = table.select(valid_columns)
```

hubdotcom:
+1 for this functionality

HampB:
+1 
this would be real nice

core/dbt/cli/params.py
PhilLab: @touchmarine @joshuataylor sounds like this is on the right track. Any chance you would be able to pick this up again?

The feature really looks promising!

touchmarine: @PhilLab the patch works as-is good enough, improving it depends on whether the dbt team wants to support an API like this or not.

quangbilly79:
Is this included in the final version? I need this!

glenn-christmas:
+1 for this as well

PR #9366, opened 2024-01-12
Feature/source freshness hooks

### LINKED ISSUE #5609: [CT-985] [Feature] on-run-end hook and results variable support for source freshness command

MichelleArk:
resolves #5609

### Problem
Currently, `on-run-*` hooks are not supported in the `dbt source freshness` command, this PR makes it so it does.

### Solution
Changed the `FreshnessTask` class to inherit from `RunTask` (as all other task types that support hooks do).

Alternative solutions:
* Move the hook logic to `GraphRunnableTask` and disabling in `CompileTask` - feels a little messier (disabling parent functionality in inheriter).
* Create a `HookTask` in the middle of the inheritance tree - seems like a lot more work when every other hook-able task just inherits from `RunTask`.

Additionally, this change has been made backwards-compatible with previous versions of dbt which did not include project hooks by default for `dbt source freshness` commands via a project flag + tests added across the different scenarios (default configuration, flag explicitly disabled, flag explicitly enabled)
### Checklist


MichelleArk:
Initial review covering first 2 commits of this change: https://github.com/dbt-labs/dbt-core/pull/9322#issuecomment-1879504718

gshank (approved):
This looks fine. As things for possible future improvements, that whole flags processing thing is getting complicated. We should look for an opportunity to refactor it. Also mixing in different Status enums is something that's not wonderful. We need to do a rethink of all of those status codes sometime.  It might be better to have a single enum instead of all of these different ones.

MichelleArk:
> As things for possible future improvements, that whole flags processing thing is getting complicated. We should look for an opportunity to refactor it.

Yeah, I agree. It's definitely got a lot of responsibilities and is quite aware of dbt-specific CLI flags even though it's meant to be more generalized. In the meantime, I've added an additional unit test for the project-only flag behaviour so it'll be safer to refactor in the future. 

>  Also mixing in different Status enums is something that's not wonderful. We need to do a rethink of all of those status codes sometime. It might be better to have a single enum instead of all of these different ones.

We may be able to take a stab at this with some of the ongoing dbt/artifacts refactoring, but agree it shouldn't block this work.

PR #9393, opened 2024-01-16
Ct 3539/yaml parsing

### LINKED ISSUE #9358: [Bug] Input following generic test name is ignored - hides incorrectly configured tests

### Current Behavior

For the not_null and unique generic tests, any value after the test name is ignored.

For example the two yamls below run the same test:

```yaml

- name: example_column
  description: "example column description"
  tests:
    - not_null anything_can_be_specified_here_and_ignored
```

```yaml

- name: example_column
  description: "example column description"
  tests:
    - not_null
```

This behaviour makes it easy for incorrectly specified test configs to go unnoticed. For example, if a user fails to specify a `:` before a test filter, the test filter will be ignored. In the example below, both tests will run but the `order_filter` will only be applied to the test for the `example_column_two_filter_applied` column. Since both tests run without warning or error it is easy for the missing `:` and associated filter to be undetected.
```yaml
models:
  - name: my_first_dbt_model
    config:
      order_filter: &order_filter
        config:
          where: "order_id > 21"
    columns:
      - name: example_column_filter_ignored
        tests:
          - not_null *order_filter
      - name: example_column_two_filter_applied
        tests:
          - not_null: *order_filter
```

### Expected Behavior

Warning or error when values after generic test name are ignored.

### Steps To Reproduce

1. Modify not_null or unique generic test specification to have any value after the test name
```yaml

- name: example_column
  description: "example column description"
  tests:
    - not_null anything_can_be_specified_here_and_ignored
```
2. `dbt test`

### Relevant log output

_No response_

### Environment

```markdown
- OS: macOS Sonoma 14.2.1
- Python: 3.11.1
- dbt: 1.5.1 & (dbt-bigquery==1.5.3)
```

bigquery

### Additional Context

_No response_

kentkr:
resolves #9358 

### Problem

Trailing characters in yaml after test names are ignored.  This can be unexpected behavior as these characters may have a purpose but be malformatted.

### Solution

Check str test names during parsing for trailing characters and raise a warning.

### Checklist


core/dbt/parser/schema_generic_tests.py
```diff
@@ -361,7 +363,11 @@ def parse_test(
         column: Optional[UnparsedColumn],
         version: Optional[NodeVersion],
     ) -> None:
+
         if isinstance(test, str):
+            # warn if test name has trailing characters which are unused
+            if len(test.split()) > 1:
+                warn_or_error(UnusedTrailingYaml(test=test))
```
kentkr, line 370: @dbeatty10 do you recommend putting the `warn_or_error` call somewhere else? I'd like to reference the node name in the warning, rather than the str `test`, but wasn't sure where to check for the issue after parsing.

dbeatty10: I don't have a good sense of this one way or the other.

But you could look through other locations where `warn_or_error` is utilized to see if any of those serve as good guidance.

Here's a listing:
https://github.com/search?q=repo%3Adbt-labs%2Fdbt-core%20warn_or_error&type=code

You'd probably want to start by looking at the examples within `core/dbt/parser/` and see if any of them are relevant to your use-case.

kentkr: I scoured the parser folder and don't see a better place for it. 

Really my comment was about if specifying the test name, and not the node name, would suffice. I think it would. Using the example from the issue the output would be:
```
[WARNING]: The test "not_null *where_filter" has trailing characters that will be ignored. If those characters are meaningful append a ":" after "unique". 
```

PR #9399, opened 2024-01-17
still run register adapter when manifest is provided.

ChenyuLInx:
@peterallenwebb I tried but couldn't write a test that would fail without this change.


gshank (approved):

ChenyuLInx:
@peterallenwebb I tried harder and added the test

ChenyuLInx:
Closing in favor of #9440 as my CLA check always fail

PR #9421, opened 2024-01-22
Enable unit testing versioned models

### LINKED ISSUE #9344: [CT-3529] [Unit Testing] Unit Testing Versioned Models
### Housekeeping

### Short description

This is the outcome of the spike+ https://github.com/dbt-labs/dbt-core/issues/8799.  Exact details of what input should look like can be found there.

### Outcomes of the spike
1.  We need to patch unit-tests to be able to determine what versions of a model exist since versions are defined in the schema files and schema files are parsed at the end.  This is largely done in https://github.com/dbt-labs/dbt-core/pull/9302.  
2. There will still only be one unit test node, even though we may be executing multiple unit tests.  The versioned models that the unit test can run against will be listed in the `depends_on` of the `UnitTestDefinition`
3. We may need to do something similar to what @gshank did in the build command [here](https://github.com/dbt-labs/dbt-core/pull/9273), where there are two 'selected' lists, one with unit tests and one without, but we would need two selection lists, one with models and one without (in order to account for run results)

### Acceptance criteria

1.  A unit test definition can define the model versions to include or exclude from a test
2. If no versions are defined in the unit test definition, but the target model is versioned, a unit test will be run for all versions of the model
3. When `dbt build` is run with a select for a versioned model, only the unit test for that specific model version will run, even if no version is defined in the schema file
4. When a command selects on a unit test that is for a versioned model, unit tests for all versions of that model will be run

### Impact to Other Teams

no

no

### Suggested Tests

1. test with no version specified, should create a separate unit test for each version
2. with with an exclude version specified, should create a separate unit test for each version except the excluded version
3. test with an include version specified, should create a single unit test for only the version specified
4. test with an include and exclude version specified, should get ValidationError
5. test with an include for an unversioned model, should error
6. partial parsing test: test with no version specified, then add an exclude version, then switch to include version and make sure the right unit tests are generated for each
7. test with no version specified in the schema file and use selection logic on a versioned model for a specific version
8. test with no version specified in the schema file and use selection logic on a unit test - expect unit tests for all versioned models
9. test specifying the fixture version with {{ ref(name, version) }}

gshank:
resolves #9344

### Problem

Versioned models aren't available at the time that unit tests are parsed, so the tested model needs to be fixed at the end of parsing. Partial parsing needs to work correctly when unit tests are against versioned models.

### Solution

A "process_unit_tests" step has been added to the ManifestLoader post-processing steps. For every changed unit test we check for versioned models and build a unit test definition for every applicable model.  We also had to defer the detection of a non-existing model to this point, because we couldn't differentiate between models that weren't found because they couldn't exist and models that weren't found because the versions hadn't been processed yet. 

### Checklist


tests/functional/unit_testing/test_ut_versions.py
```diff
@@ -0,0 +1,162 @@
...
+        assert len(results) == 1
+        unique_ids = get_unique_ids_in_results(results)
+        assert unique_ids == ["unit_test.test.my_model.test_my_model_v2"]
+
+        # with an exclude version specified, should create a separate unit test
+        # for each version except the excluded version (v2)
+        write_file(
+            test_my_model_exclude_versions_yml, project.project_root, "models", "unit_tests.yml"
+        )
+        # without re-running here, we get an error in the v3 unit test. Not sure why.
```
emmyoop, line 67: Is this still true?

core/dbt/contracts/graph/nodes.py
```diff
@@ -1073,6 +1074,9 @@ class UnitTestDefinition(NodeInfoMixin, GraphNode, UnitTestDefinitionMandatory):
     config: UnitTestConfig = field(default_factory=UnitTestConfig)
     checksum: Optional[str] = None
     schema: Optional[str] = None
+    created_at: float = field(default_factory=lambda: time.time())
+    versions: Optional[UnitTestNodeVersions] = None
+    version: Optional[NodeVersion] = None
```
emmyoop, line 1079: What is this version for?

core/dbt/parser/unit_tests.py
```diff
@@ -214,6 +222,9 @@ def _get_original_input_node(self, input: str, tested_node: ModelNode):
             else:
                 raise InvalidUnitTestGivenInput(input=input)
 
+        if not original_input_node:
+            raise InvalidUnitTestGivenInput(input=input)
```
emmyoop, line 226: How do we end up here?  Can we test this?

tests/functional/unit_testing/test_ut_versions.py
```diff
@@ -0,0 +1,162 @@
...
+        unique_ids = get_unique_ids_in_results(results)
+        expected_ids = [
+            "unit_test.test.my_model.test_my_model_v1",
+            "unit_test.test.my_model.test_my_model_v2",
+            "unit_test.test.my_model.test_my_model_v3",
+        ]
+        assert sorted(expected_ids) == sorted(unique_ids)
+
+        # Select tests for a single versioned model
+        results = run_dbt(["test", "--select", "my_model.v2"])
```
emmyoop, line 57: Should we also be able to select tests for all versions and expect all unit tests for each version of the model?

`run_dbt(["test", "--select", "my_model"])`

emmyoop (changes requested):
Looking great!  I mostly just had questions.

core/dbt/parser/unit_tests.py
MichelleArk: I think we'd hit this if a unit test case was specified against a model or source that did not exist in the manifest

core/dbt/contracts/graph/unparsed.py
```diff
@@ -830,3 +836,11 @@ class UnparsedUnitTest(dbtClassMixin):
     description: str = ""
     overrides: Optional[UnitTestOverrides] = None
     config: Dict[str, Any] = field(default_factory=dict)
+    versions: Optional[UnitTestNodeVersions] = None
```
MichelleArk, line 839: Is there a reason we couldn't use `dbt_common.helper_types.IncludeExclude` here instead? it would consolidate some of the validation we do below and provides an `.includes` method. Its syntax/error messaging would also be consistent with other features that use include/exclude

tests/functional/unit_testing/test_ut_versions.py
gshank: Yes. I don't think it's actually related to this work, as far as I can see. Something weird is happening in the SQL construction bit, which is difficult to debug. So I wanted to get this review going. If I don't get it solved soon, I was thinking we could open up a follow-on ticket to diagnose.

core/dbt/contracts/graph/nodes.py
gshank: Right now it's used to construct the name that's displayed on the terminal. But it also seems like we should have some way of distinguishing the related version for the unit test.

tests/functional/unit_testing/test_ut_versions.py
gshank: Yes, that should work. I can make another test for it.

core/dbt/contracts/graph/unparsed.py
gshank: Didn't even realize that was a thing :). I'll look into using it.

core/dbt/parser/unit_tests.py
gshank: If somebody has an invalid model ref in the given input, there would be an error.  It's not actually related to this pr, it's base functionality, but I can add a test.

core/dbt/contracts/graph/unparsed.py
gshank: The dbt_common IncludeExclude class only handles strings, while the UnitTestNodeVersions has include and exclude for NodeVersion which is a Union of str and float.  Also the dbt_common class handles include and exclude differently (include can be a single string but exclude can't) which wouldn't be necessary for this use case.

It might be possible to use generics to modify the class to use it here, but I'd prefer to investigate that separately. It would be nice to have the same behavior though -- do you think it's worth copying the functionality?

tests/functional/unit_testing/test_ut_versions.py
gshank: Done.

core/dbt/parser/unit_tests.py
gshank: Done.

.changes/unreleased/Features-20240122-145854.yaml
```diff
@@ -0,0 +1,6 @@
+kind: Features
+body: Enable unit testing versions
```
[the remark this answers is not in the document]
gshank: Changed.

emmyoop (approved):
Look good!

core/dbt/contracts/graph/unparsed.py
MichelleArk: Ahhh I see, thanks for digging into it. I think it'd only be worth it to mirror the existing syntax 1:1 but I don't think we'd need to make breaking changes to do that using a generalized IncludeExclude in the future.. so I don't have a strong opinion here

PR #9433, opened 2024-01-24
Support materialized views in `get_columns_in_relation`

### LINKED ISSUE #9419: [Bug] postgres `get_columns_in_relation` does not include MVs

### Current Behavior

calling `adapter.get_columns_in_relation` on a materialized view returns 0 columns.

### Expected Behavior

calling `adapter.get_columns_in_relation` on a materialized view returns all the columns in the materialized view.

### Steps To Reproduce

**Create a Materialized view**
```
{{ config(
    materialized="materialized_view"
) }}

(SELECT
    1 as int_col,
    'one' as string_col,
    FALSE as bool_col)
UNION
(SELECT
    2 as int_col,
    'two' as string_col,
    TRUE as bool_col)
UNION
(SELECT
    3 as int_col,
    'three' as string_col,
    FALSE as bool_col)
```

**Build the Model**
`dbt build -s mv_gene_yaml_test`

**Run the macro**
`{{ codegen.generate_model_yaml(model_names=['mv_gene_yaml_test']) }}`
or
`dbt run-operation codegen.generate_model_yaml --args '{"model_names": ["mv_gene_yaml_test"]}'`

(this is copied from a `dbt-codegen` [issue](https://github.com/dbt-labs/dbt-codegen/issues/155). This macro uses `adapter.get_columns_in_relation` under the hood. You could update the last step to call the macro directly. 

per @drewblinn in that issue:
> The macro queries `information_schema.columns` which does not store data about materialized views. Column information for MVs can be found in` pg_catalog.pg_attribute`

### Relevant log output

_No response_

### Environment

```markdown
- OS:
- Python:
- dbt: 1.7.4
```

postgres

### Additional Context

_No response_

dbeatty10:
resolves #9419

### Problem

As described in https://github.com/dbt-labs/dbt-core/issues/9419#issuecomment-1907067697, the [`information_schema.columns` view](https://www.postgresql.org/docs/current/infoschema-columns.html) in postgres includes these:
- o[r]dinary table, [v]iew, [f]oreign table, [p]artitioned table

But it does **not** include:
- [m]aterialized view

### Solution

Two things:
1. Add the full definition of the `information_schema.columns` view into dbt-postgres.
2. Add `'m'::"char"` to the definition so that MVs are included.

To get the full text definition of this view, run the following:
```sql
select definition 
from pg_catalog.pg_views 
where schemaname = 'information_schema' 
  and viewname = 'columns'
```

The output is a fairly gnarly query.

### To do

After a quick search, I couldn't find any relevant tests for `adapter.get_columns_in_relation`, so we'll probably want some that cover all the different types of relations:
- ephemeral
- table
- view
- materialized view

The tests would basically just need to:
1. Define a relation with a column or two.
2. Build it.
3. Make sure that those columns are returned by `adapter.get_columns_in_relation`.

I'm assuming that ephemeral will either raise an exception or return nothing, but we'll want to add applicable tests either way.

### Checklist

### DIFF
--- a/.changes/unreleased/Fixes-20240123-191339.yaml
+++ b/.changes/unreleased/Fixes-20240123-191339.yaml
@@ -1,0 +1,6 @@
+kind: Fixes
+body: Support materialized views in get_columns_in_relation
+time: 2024-01-23T19:13:39.538476-07:00
+custom:
+  Author: dbeatty10
+  Issue: "9419"
--- a/plugins/postgres/dbt/include/postgres/macros/adapters.sql
+++ b/plugins/postgres/dbt/include/postgres/macros/adapters.sql
@@ -184,7 +184,7 @@
      JOIN pg_namespace nco ON ((co.collnamespace = nco.oid))) ON (((a.attcollation = co.oid) AND ((nco.nspname <> 'pg_catalog'::name) OR (co.collname <> 'default'::name)))))
      LEFT JOIN (pg_depend dep
      JOIN pg_sequence seq ON (((dep.classid = ('pg_class'::regclass)::oid) AND (dep.objid = seq.seqrelid) AND (dep.deptype = 'i'::"char")))) ON (((dep.refclassid = ('pg_class'::regclass)::oid) AND (dep.refobjid = c.oid) AND (dep.refobjsubid = a.attnum))))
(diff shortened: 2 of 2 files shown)


zqureshi:

PR #9439, opened 2024-01-24
Define Macro and Group resources in dbt/artifacts

MichelleArk:
resolves https://github.com/dbt-labs/dbt-core/issues/9382
resolves https://github.com/dbt-labs/dbt-core/issues/9381

### Problem

* Moving Macro and Group schema definitions to dbt/artifacts/resources

### Solution

* Split `Macro` into resource (defined in dbt/artifcats/resources/v1) and internal representation
* Move dependency `Docs` to dbt/artifacts
* Renaming import of `DocumentationArtifact` to `DocumentationResource` for consistent naming
* Split `Group` into functional & resource classes
* rename `BaseArtifactNode` to `BaseResource`

### Checklist

### DIFF
--- a/core/dbt/artifacts/resources/v1/macro.py
+++ b/core/dbt/artifacts/resources/v1/macro.py
@@ -1,0 +1,40 @@
+from dataclasses import dataclass, field
+import time
+from typing import Literal, List, Dict, Optional, Any
+
+from dbt_common.contracts.util import Replaceable
+from dbt_common.dataclass_schema import dbtClassMixin
+from dbt.artifacts.resources.base import BaseArtifactNode
+from dbt.artifacts.resources.types import NodeType, ModelLanguage
+from dbt.artifacts.resources.v1.docs import Docs
+
+
+@dataclass
+class MacroArgument(dbtClassMixin):
+    name: str
+    type: Optional[str] = None
+    description: str = ""
+
+
+@dataclass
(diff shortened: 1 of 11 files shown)


core/dbt/artifacts/resources/v1/macro.py
```diff
@@ -0,0 +1,40 @@
...
+    type: Optional[str] = None
+    description: str = ""
+
+
+@dataclass
+class MacroDependsOn(dbtClassMixin, Replaceable):
+    macros: List[str] = field(default_factory=list)
+
+    # 'in' on lists is O(n) so this is O(n^2) for # of macros
+    def add_macro(self, value: str):
```
MichelleArk, line 24: Went back and forth on whether this method belongs in dbt/artifacts given it is mutating / not read-only. An alternative would be to define the method more directly on the internal `Macro` node class. 

 Decided to leave it as-is for now because `DependsOn` inherits from `MacroDependsOn`. We should revisit once we move `DependsOn` though.

QMalcolm: For those looking in @MichelleArk and I have talked about this a bit in DMs. In my PR #9460 I plan on moving `DependsOn` in full over to artifacts. We plan in the future to revist the design of both `MacroDependsOn` and `DependsOn` to perhaps simplify and improve them.

QMalcolm (approved):
Looks good to me!

PR #9440, opened 2024-01-24
still register adapter when manifest is provided

ChenyuLInx:
This is a fix that we did when backporting a change to 1.7

https://github.com/dbt-labs/dbt-core/commit/6e331833b0cb8680aa0b269298a7134492155046#diff-beabcbe3aef83fe254d714ef250f484abd1f1ef314dc9506059eb54a9bfbf9a0R274

### DIFF
--- a/core/dbt/cli/requires.py
+++ b/core/dbt/cli/requires.py
@@ -1,7 +1,7 @@
 import dbt.tracking
 from dbt_common.invocation import reset_invocation_id
 from dbt.version import installed as installed_version
-from dbt.adapters.factory import adapter_management
+from dbt.adapters.factory import adapter_management, register_adapter
 from dbt.flags import set_flags, get_flag_dict
 from dbt.cli.exceptions import (
     ExceptionExit,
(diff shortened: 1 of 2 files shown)


core/dbt/cli/requires.py
```diff
@@ -277,6 +278,8 @@ def wrapper(*args, **kwargs):
                 ctx.obj["manifest"] = parse_manifest(
                     runtime_config, write_perf_info, write, ctx.obj["flags"].write_json
                 )
+            else:
+                register_adapter(runtime_config, get_mp_context())
```
MichelleArk, line 282: I think we need to add: 

```py
adapter = get_adapter(runtime_config)      
adapter.set_macro_context_generator(generate_runtime_macro_context)
adapter.set_macro_resolver(ctx.obj["manifest"])
```

to get the tests passing

ChenyuLInx: Thanks! Done!

MichelleArk (approved):

PR #9443, opened 2024-01-24
fix retry as CLI

ChenyuLInx:
Resolves: #9444
Our integration test right now goes through dbtRunner, which will add the `manifest` as None in `ctx.obj`.
When running retry as a CLI command, current code would just fail.

We should add some test to run each command as CLI command just to make sure CLI works.

### DIFF
--- a/core/dbt/cli/main.py
+++ b/core/dbt/cli/main.py
@@ -608,7 +608,7 @@ def retry(ctx, **kwargs):
     task = RetryTask(
         ctx.obj["flags"],
         ctx.obj["runtime_config"],
-        ctx.obj["manifest"],
+        None,
     )
 
     results = task.run()
--- a/core/dbt/cli/main.py
+++ b/core/dbt/cli/main.py
@@ -608,7 +608,6 @@ def retry(ctx, **kwargs):
     task = RetryTask(
         ctx.obj["flags"],
         ctx.obj["runtime_config"],
-        None,
     )
 
     results = task.run()
--- a/core/dbt/cli/main.py
+++ b/core/dbt/cli/main.py
@@ -602,9 +602,9 @@ def run(ctx, **kwargs):
 @requires.profile
 @requires.project
 @requires.runtime_config
-@requires.manifest
 def retry(ctx, **kwargs):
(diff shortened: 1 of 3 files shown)


core/dbt/cli/main.py
```diff
@@ -608,7 +608,7 @@ def retry(ctx, **kwargs):
     task = RetryTask(
         ctx.obj["flags"],
         ctx.obj["runtime_config"],
-        ctx.obj["manifest"],
```
QMalcolm, line 611: Is it okay for the manifest to be `None`? Would it be safer to do something like `ctx.obj.get("manifest")` which returns either the object or `None`?

ChenyuLInx: So the last refactor actually refactored parsing for retry into the task itself. So it would be alright to always pass in a None(which is what our intergration tests where doing before)

core/dbt/cli/main.py
```diff
@@ -608,7 +608,7 @@ def retry(ctx, **kwargs):
     task = RetryTask(
```
MichelleArk, line 608: Let's also remove @requires.manifest & manifest from the `RetryTask` constructor

core/dbt/cli/main.py
MichelleArk: We don't even use the manifest passed to the RetryTask's constructor at all at all: https://github.com/dbt-labs/dbt-core/blob/main/core/dbt/task/retry.py#L66

This should be safe to entirely remove, we don't even need to pass `None`

ChenyuLInx: Done!

MichelleArk (approved):

core/dbt/cli/main.py
QMalcolm: Awesome! Thank y'all for the clarification 🙂

QMalcolm (approved):

core/dbt/cli/main.py
ChenyuLInx: Thanks for poking and resulting in better comment.

PR #9452, opened 2024-01-25
update dependencies

### LINKED ISSUE #8790: [CT-3190] Pinning detective work
### Housekeeping

### Short description

We recently pinned `types-requests<2.31.0` because it had a dependency conflict with `urllib3` which we have pinned to `~=1.0` because of another conflict with `requests` requiring `openssl`.

This ticket is to look into if those pins are still required and clean them up if not.  

### Acceptance criteria

We have confirmed that the pins are
- required to continue to work
_or_
- not required and we have re-pinned appropriately

### Impact to Other Teams

adapters - based on the notes it seems like `urllib3` is pinned for the snowflake adapter as well so we will want to ensure changing the dependencies does not adversely affect them

no

### Context

_No response_

emmyoop:
resolves #8790 

### Problem

We had to hard pin a type check because of dependency resolution

### Solution

Stop pinning it
add back requests dependency because core uses it
remove other type checkers we no longer need

### Checklist

### DIFF
--- a/core/setup.py
+++ b/core/setup.py
@@ -61,6 +61,7 @@ setup(
         # with major versions in each new minor version of dbt-core.
         "click>=8.0.2,<9",
         "networkx>=2.3,<4",
+        "requests<3.0.0",  # should match dbt-common
         # ----
         # These packages are major-version-0. Keep upper bounds on upcoming minor versions (which could have breaking changes)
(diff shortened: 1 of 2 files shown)


ChenyuLInx (approved):
Thanks！！

MichelleArk (approved):

PR #9456, opened 2024-01-25
TestGenerateCatalogWithExternalNodes

### LINKED ISSUE #9458: [Regression] External nodes not included in catalog during docs generation

### Current Behavior

External nodes should be included in the catalog as a result of `dbt docs generate`. Currently, projects with external nodes are not included in the resulting catalog.json

catalog.json includes catalog entry for external node when project contains external node.

### Steps To Reproduce

test passes 1.6.latest: https://github.com/dbt-labs/dbt-core/pull/9457
same test fails on main (and 1.7.latest, confirmed locally): https://github.com/dbt-labs/dbt-core/pull/9456

### Relevant log output

_No response_

### Environment

```markdown
- OS:
- Python:
- dbt (working version): 1.6.9
- dbt (regression version): 1.7.5, 1.8.0a2
```

_No response_

### Additional Context

_No response_

MichelleArk:
resolves #9458

### Problem
catalog from docs generate does not include external nodes, even though that was the case in 1.6. This is because starting from 1.7, the docs generate task started using a ResourceSelector which filters out empty nodes (e.g. external ones) as opposed to iterating over the manifest directly.

### Solution

* Set `include_empty_nodes=True` in the `ResourceTypeSelector` for the generate task.
  * This indicates to the selector that 'empty' (e.g. nodes without raw_code, like external nodes) should be included in selection and make their way to the catalog
  * Same approach as was used to select external nodes in the ListCommand, which needed to select external nodes for the result even though they are not 'runnable': https://github.com/dbt-labs/dbt-core/blob/main/core/dbt/task/list.py#L201
* Add a test that ensures generated catalogs include external nodes when those are available in the project

### Checklist

### DIFF
--- a/.changes/unreleased/Fixes-20240125-182243.yaml
+++ b/.changes/unreleased/Fixes-20240125-182243.yaml
@@ -1,0 +1,7 @@
+kind: Fixes
+body: Add TestGenerateCatalogWithExternalNodes, include empty nodes in node selection
+  during docs generate
+time: 2024-01-25T18:22:43.253228-05:00
+custom:
+  Author: michelleark
+  Issue: "9456"
--- a/core/dbt/task/docs/generate.py
+++ b/core/dbt/task/docs/generate.py
@@ -343,6 +343,7 @@ class GenerateTask(CompileTask):
             manifest=self.manifest,
             previous_state=self.previous_state,
             resource_types=EXECUTABLE_NODE_TYPES,
+            include_empty_nodes=True,
         )
 
     def get_catalog_results(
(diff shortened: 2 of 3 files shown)


graciegoheen:
@MichelleArk What's the reason to backport to `1.1.latest`? I think we only need to backport to `1.7.latest`?

MichelleArk:
Agreed! Think the 1.1.latest label was just fat-thumbed

MichelleArk:
@aranke - good question! Exposures have their own handling for selection: https://github.com/dbt-labs/dbt-core/blob/main/core/dbt/graph/selector.py#L168

This should not affect them.

MichelleArk:
Manual backport: https://github.com/dbt-labs/dbt-core/pull/9483

PR #9343, opened 2024-01-05
Ensure unit test models can depend on external nodes

QMalcolm:
resolves #8944

### Problem
We weren't sure if external nodes were already supported in unit test nodes. This PR added tests that identified that unit test nodes didn't have full support for external nodes, and then made the necessary changes to support them.

### Solution
* Update all models in a unit test node's `shadow manifest` have a `path` for file writing
* Update unit test parsing to ensure models keep their original package name
* Altered the `packages_for_node` to gracefully allow for unknown dependencies

### Checklist

- [X] I have read [the contributing guide](https://github.com/dbt-labs/dbt-core/blob/main/CONTRIBUTING.md) and understand what's expected of me  
- [X] I have run this code in development and it appears to resolve the stated issue  
- [X] This PR includes tests, or tests are not required/relevant for this PR
- [X] This PR has no interface changes (e.g. macros, cli, logs, json artifacts, config files, adapter interface, etc) or this PR has already received feedback and approval from Product or DX
- [X] This PR includes [type annotations](https://docs.python.org/3/library/typing.html) for new and modified functions

PR #9483, opened 2024-01-29
TestGenerateCatalogWithExternalNodes (#9456)

MichelleArk:
(cherry picked from commit 06e55bb93ed771921d656807cd6bfa2dc1995b20)

resolves #

### Checklist

### DIFF
--- a/core/dbt/context/providers.py
+++ b/core/dbt/context/providers.py
@@ -60,7 +60,6 @@ from dbt.exceptions import (
     MetricArgsError,
     MissingConfigError,
     OperationsCannotRefEphemeralNodesError,
-    PackageNotInDepsError,
     ParsingError,
     RefBadContextError,
     RefArgsError,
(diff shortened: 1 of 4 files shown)


core/dbt/context/providers.py
```diff
@@ -638,10 +637,8 @@ def packages_for_node(self) -> Iterable[Project]:
         package_name = self._node.package_name
 
         if package_name != self._config.project_name:
-            if package_name not in dependencies:
-                # I don't think this is actually reachable
-                raise PackageNotInDepsError(package_name, node=self._node)
-            yield dependencies[package_name]
+            if package_name in dependencies:
+                yield dependencies[package_name]
```
MichelleArk, line 641: from https://github.com/dbt-labs/dbt-core/pull/9343/files#r1461989833: 

> For posterity, thinking out loud about the implications of this change:
> * This method is used to load the vars available for a particular jinja (model-scoped) context via the vars macro.
> * This method falls back on the running project's vars if the node has no associated packages. If it returns the wrong package -- this could lead to a 'no var found' issue at runtime.
> * Before external models, it was not possible to have a model node exist in the project without a corresponding package, which is probably why this code path was not reachable
> * With external nodes, this is totally possible now -- so explicitly supporting both paths feels totally reasonable, and correct!

martynydbt (approved):

PR #9484, opened 2024-01-29
Move data portions of `Metric` node to `dbt/artifacts`

QMalcolm:
resolves #

### Checklist


core/dbt/artifacts/resources/v1/metric.py
```diff
@@ -0,0 +1,135 @@
...
+    created_at: float = field(default_factory=lambda: time.time())
+    group: Optional[str] = None
+
+    @property
+    def depends_on_nodes(self):
+        return self.depends_on.nodes
+
+    @property
+    def search_name(self):
+        return self.name
```
MichelleArk, line 123: I think these should stay in contracts/graph as they are not part of the dsi protocol

core/dbt/artifacts/resources/v1/metric.py
```diff
@@ -0,0 +1,135 @@
...
+
```
MichelleArk, line 124: A comment here pointing to the metric protocol in DSI would be helpful! 

https://github.com/dbt-labs/dbt-semantic-interfaces/blob/main/dbt_semantic_interfaces/protocols/metric.py

core/dbt/artifacts/resources/v1/metric.py
QMalcolm: Good catch! Guess I got a little indiscriminate in the final commit 🙃

MichelleArk (approved):
🚢 🇮🇹

PR #9503, opened 2024-02-01
Fix `compiled_code`, reimplement `sql` as wrapper

### LINKED ISSUE #9502: [Bug] Update command-based logic in `compiled_code` context member, to match `sql`

### Current Behavior

In https://github.com/dbt-labs/dbt-core/pull/9040, we added `defer_relation` for all tasks that perform deferral.

Previously, there was dynamic routing logic in the `sql` + `compiled_code` context members that checked for the existnece of `defer_relation` and overwrote the "compiled code" to use during `clone`.

#9040 updated the dynamic routing logic for the `sql` context member (based on the running task), but not the `compiled_code` context member.

These need to be the same. Otherwise, I might `build` a model with deferral and, depending on whether the materialization uses `{{ sql }}` or `{{ compiled_code }}`, get a different (inconsistent) result.

### Expected Behavior

There should be no difference between `{{ sql }}` and `{{ compiled_code }}` for SQL models.

### Steps To Reproduce

1. dbt-core `main` branch
2. Use a materialization that calls `compiled_code` instead of `sql`, e.g. [dbt-snowflake's incremental materialization](https://github.com/dbt-labs/dbt-snowflake/blob/4b2d93808e1b852aec3b1c281fa0af231b0cf297/dbt/include/snowflake/macros/materializations/incremental.sql#L83)

### Relevant log output

_No response_

### Environment

```markdown
dbt: main
```

_No response_

### Additional Context

_No response_

jtcohen6:
resolves #9502

### Problem

- #9040 added `defer_relation` to all nodes any time `--defer` was provided
- We had routing logic for the `clone` command that depended on the presence of `defer_relation`
- #9040 updated this logic for `sql` context member, but not for `compiled_code`

### Solution

- Update the logic for `compiled_code` context member
- Reimplement `sql` context member as a wrapper around `compiled_code`, to prevent future divergence

### Checklist

### DIFF
--- a/core/dbt/context/providers.py
+++ b/core/dbt/context/providers.py
@@ -1458,31 +1458,22 @@ class ModelContext(ProviderContext):
         ]
 
     @contextproperty()
-    def sql(self) -> Optional[str]:
-        # only doing this in sql model for backward compatible
-        if self.model.language == ModelLanguage.sql:  # type: ignore[union-attr]
-            # If the model is deferred and the adapter doesn't support zero-copy cloning, then select * from the prod
-            # relation
-            # TODO: avoid routing on args.which if possible
-            if getattr(self.model, "defer_relation", None) and self.config.args.which == "clone":
-                # TODO https://github.com/dbt-labs/dbt-core/issues/7976
-                return f"select * from {self.model.defer_relation.relation_name or str(self.defer_relation)}"  # type: ignore[union-attr]
-            elif getattr(self.model, "extra_ctes_injected", None):
-                # TODO CT-211
-                return self.model.compiled_code  # type: ignore[union-attr]
-            else:
-                return None
-        else:
-            return None
-
-    @contextproperty()
     def compiled_code(self) -> Optional[str]:
-        if getattr(self.model, "defer_relation", None):
+        # TODO: avoid routing on args.which if possible
+        if getattr(self.model, "defer_relation", None) and self.config.args.which == "clone":
             # TODO https://github.com/dbt-labs/dbt-core/issues/7976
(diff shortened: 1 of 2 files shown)


gshank (approved):

MichelleArk (approved):

core/dbt/context/providers.py
```diff
@@ -1486,6 +1469,14 @@ def compiled_code(self) -> Optional[str]:
         else:
             return None
 
+    @contextproperty()
+    def sql(self) -> Optional[str]:
+        # only set this for sql models, for backward compatibility
+        if self.model.language == ModelLanguage.sql:  # type: ignore[union-attr]
+            return self.compiled_code
```
mikealfare, line 1476: If I'm following the logic correctly, I think this previously resulted in `self.model.compiled_code`. I want to make sure this is intentional since both are syntactically correct.

jtcohen6:
🎩 Confirmed this fix locally by running an incremental model on `dbt-snowflake`

Before:
```sql
On model.my_project.my_model: /* {"app": "dbt", "dbt_version": "1.8.0a1", "profile_name": "snowflake", "target_name": "dev", "node_id": "model.my_project.my_model"} */
create or replace transient table analytics_dev.dbt_jcohen.my_model
         as
        (select * from analytics.analytics.my_model
        );
```

After:
```sql
On model.my_project.my_model: /* {"app": "dbt", "dbt_version": "1.8.0a1", "profile_name": "snowflake", "target_name": "dev", "node_id": "model.my_project.my_model"} */
create or replace transient table analytics_dev.dbt_jcohen.my_model
         as
        (

... tons of SQL ...
);
```

PR #9529, opened 2024-02-06
Initialize invocation context before test fixture creation.

peterallenwebb:
### Problem

The code used to set up test fixtures sometimes uses environment variables via the new InvocationContext mechanism, but it was not always initialized in time.

### Solution

Add an initialization function to the test fixtures, and set an invocation context in that function.

### Checklist


QMalcolm (approved):
Thank you so much for hunting down this issue 🙇

PR #9538, opened 2024-02-07
Move manifest nodes to dbt/artifacts

### LINKED ISSUE #9388: [CT-3558] Move all `ManifestNode`s to dbt/artifacts
### Housekeeping

### Short description

As part of an effort to move any `ArtifactMixin` subclasses upstream of other core modules into the top -level `dbt/artifacts` directory, we need to move all ManifestNode classes, which WritableManifest depends on.

### Acceptance criteria

All [ManifestNode](https://github.com/dbt-labs/dbt-core/blob/main/core/dbt/contracts/graph/nodes.py#L1844-L1848) classes should be split into data + functional components, where data contracts are defined upstream in dbt/artifacts. 

### Impact to Other Teams

N/A - should be a refactor. 

nope

### Context

_No response_

gshank:
resolves #9388

### Problem

As part of the initiative to make a package to handle artifact generation, serializing and validation we are moving things to a separate dbt/artifacts directory. This PR is to move the nodes/resources in the ManifestNodes union.

### Solution

Move the nodes in the ManifestNodes union to the artifacts package.

### Checklist


gshank:
Questions:

There's some validation logic in the configs that I'm not sure should be in artifacts. Should we attempt to move that out someplace else?

I put all of the "resources" into a manifest_nodes.py file, because a number of them were pretty small and it seemed like overkill to give each one separate files, but I don't have strong feelings about that.

Anything in the "base" resources dictionary will not be versioned, so we should be sure that it makes sense to put things there.

I did not put UnitTestNode into artifacts because right now that resource is never written out into json artifacts, it's only used internally in constructing the temporary unit testing manifest.

None of the embedded objects (i.e. Configs and things like RefArgs) can use the pattern of putting the fields and methods in different classes, because then there are different types in the artifacts and the nodes and mypy complains.  It works fine for top-level resources though.

gshank:
There's a couple of failing unit tests that I need to chase down, but wanted to give people on opportunity to review.

core/dbt/artifacts/resources/v1/components.py
```diff
@@ -35,3 +52,113 @@ def keyword_args(self) -> Dict[str, Optional[NodeVersion]]:
...
+    deferred: bool = False
+    unrendered_config: Dict[str, Any] = field(default_factory=dict)
+    created_at: float = field(default_factory=lambda: time.time())
+    config_call_dict: Dict[str, Any] = field(default_factory=dict)
+    relation_name: Optional[str] = None
+    raw_code: str = ""
+
+
+@dataclass
+class CompiledNode(ParsedNode):
```
MichelleArk, line 149: We could rename these to ParsedResource (& ParsedResourceMandatory) + CompiledResource and define them in https://github.com/dbt-labs/dbt-core/blob/main/core/dbt/artifacts/resources/base.py to be more consistent with the base + graph resource refactors.

core/dbt/artifacts/resources/v1/manifest_nodes.py
MichelleArk, line 1: > I put all of the "resources" into a manifest_nodes.py file, because a number of them were pretty small and it seemed like overkill to give each one separate files, but I don't have strong feelings about that.

An alternative organization could be to define these as separate files, but also pull in some of their (exclusive, not reused) components so its more clear what components participate in what resources. For example, I could imagine a `seed.py` that defines a `SeedResource` as well as `SeedConfig`. It could be a nice way to get the config.py file smaller as well.

MichelleArk: Separately, what do you think about renaming these resources to drop the 'Node' bit? e.g. just `Analysis`, `Hook`, `Sql(Operation)`, etc. That would be more consistent with the pattern that's shaking out in dbt/artifacts so far.

core/dbt/artifacts/resources/v1/components.py
gshank: We can't add anything to classes in base.py because they aren't versioned. Are we sure that this is what we want?

MichelleArk: Discussed offline: there isn't really a clear policy shaking out to discern whether these type of base resource classes should be defined in v1/ or one directory above in base.py. Let's revisit this in later on once we have backward/forward compatibility considered and tested, potentially as part of a more holistic 'final' reorganization push.

MichelleArk (approved):
Huge lift! Thank you 💪

PR #9539, opened 2024-02-07
Preserve model constraints when versioning models

### LINKED ISSUE #9364: [CT-3544] [Bug] Model level constraints aren't being applied to versioned models

### Current Behavior

If you're using model versions (https://docs.getdbt.com/docs/collaborate/govern/model-versions), model level constraints (https://docs.getdbt.com/reference/resource-properties/constraints#defining-constraints) don't seem to be applied to the model.

### Expected Behavior

Model level constraints should be applied as expected.

### Steps To Reproduce

First let's look at a well behaved model level constraints example (without model versioning).

```yaml
# dbt_project.yml
name: my_dbt_project
profile: all
config-version: 2
version: "1.0.0"

models:
  my_dbt_project:
    +materialized: table

# models/schema.yml
version: 2
models:
  - name: foo
    config:
      contract:
        enforced: true
    constraints:
      - type: primary_key
        columns: [id, user_name]
    columns:
      - name: id
        data_type: int
        constraints:
          - type: not_null
      - name: user_name
        data_type: text
```

```sql
-- models/foo.sql
select 1 as id, 'alice' as user_name
```

Do a run:

```sh
$ dbt --debug run -s foo
...
23:06:58  On model.my_dbt_project.foo: /* {"app": "dbt", "dbt_version": "1.7.4", "profile_name": "all", "target_name": "pg-local", "node_id": "model.my_dbt_project.foo"} */

  
    

  create  table "postgres"."public"."foo__dbt_tmp"
  
  
    
  
  (
    id int not null,
    user_name text,
    
    primary key (id, user_name)
    )
 ;
    insert into "postgres"."public"."foo__dbt_tmp" (
      id, user_name
    )
  
  (
    
    select id, user_name
    from (
        select 1 as id, 'alice' as user_name
    ) as model_subq
  );
  
23:06:58  SQL status: INSERT 0 1 in 0.0 seconds
...
```

We see the constraint `primary key (id, user_name)` is applied.

Now let's rename the file `foo.sql` > `foo_v1.sql` and edit the `schema.yml` files contents:

```yaml
# models/schema.yml
version: 2
models:
  - name: foo
    latest_version: 1
    config:
      contract:
        enforced: true
    constraints:
      - type: primary_key
        columns: [id, user_name]
    columns:
      - name: id
        data_type: int
        constraints:
          - type: not_null
      - name: user_name
        data_type: text
    versions:
      - v: 1
```

```sql
-- models/foo_v1.sql
select 1 as id, 'alice' as user_name
```

Do a run:

```sh
$ dbt clean && dbt --debug run -s foo
...
23:09:56  On model.my_dbt_project.foo.v1: /* {"app": "dbt", "dbt_version": "1.7.4", "profile_name": "all", "target_name": "pg-local", "node_id": "model.my_dbt_project.foo.v1"} */

  
    

  create  table "postgres"."public"."foo_v1__dbt_tmp"
  
  
    
  
  (
    id int not null,
    user_name text
    
    )
 ;
    insert into "postgres"."public"."foo_v1__dbt_tmp" (
      id, user_name
    )
  
  (
    
    select id, user_name
    from (
        select 1 as id, 'alice' as user_name
    ) as model_subq
  );
  
23:09:56  SQL status: INSERT 0 1 in 0.0 seconds
...
```

A versioned model of `foo` is created (`foo_v1`) and column level contraints are applied but the model level constraint is not applied.

### Relevant log output

_No response_

### Environment

```markdown
- OS: macOS
- Python: 3.11.2
- dbt: 
Core:
  - installed: 1.7.4
  - latest:    1.7.4 - Up to date!

Plugins:
  - bigquery:  1.7.2 - Up to date!
  - snowflake: 1.7.1 - Up to date!
  - postgres:  1.7.4 - Up to date!
```

postgres

### Additional Context

_No response_

gshank:
resolves #9364

### Problem

When changing a model from non-versioned to versioned, the model constraints were being lost.

### Solution

Constraints were only being copied when the model contract.enabled was True, but when parsing versioned models the config hadn't been processed before handling constraints.

### Checklist

### DIFF
--- a/core/dbt/parser/schemas.py
+++ b/core/dbt/parser/schemas.py
@@ -845,7 +845,9 @@ class ModelPatchParser(NodePatchParser[UnparsedModelUpdate]):
                 # Includes alias recomputation
                 self.patch_node_config(versioned_model_node, versioned_model_patch)
 
-                # Need to reapply this here, in the case that 'contract: {enforced: true}' was during config-setting
+                # Need to reapply setting constraints and contract checksum here,
+                # in the case that 'contract: {enforced: true}' was changed during config-setting
+                self.patch_constraints(versioned_model_node, versioned_model_patch.constraints)
                 versioned_model_node.build_contract_checksum()
                 source_file.append_patch(
                     versioned_model_patch.yaml_key, versioned_model_node.unique_id
@@ -869,6 +871,7 @@ class ModelPatchParser(NodePatchParser[UnparsedModelUpdate]):
                     unique_id=node.unique_id,
                     field_value=patch.access,
                 )
+        # These two will have to be reapplied after config is built for versioned models
         self.patch_constraints(node, patch.constraints)
         node.build_contract_checksum()
 
--- a/core/dbt/parser/schemas.py
+++ b/core/dbt/parser/schemas.py
@@ -845,8 +845,9 @@ class ModelPatchParser(NodePatchParser[UnparsedModelUpdate]):
                 # Includes alias recomputation
                 self.patch_node_config(versioned_model_node, versioned_model_patch)
 
-                # Need to reapply setting constraints and contract checksum here,
-                # in the case that 'contract: {enforced: true}' was changed during config-setting
+                # Need to reapply setting constraints and contract checksum here, because
+                # they depend on node.contract.enabled, which wouldn't be set when
+                # patch_node_properties was called if it wasn't set in the model file.
                 self.patch_constraints(versioned_model_node, versioned_model_patch.constraints)
                 versioned_model_node.build_contract_checksum()
                 source_file.append_patch(
--- a/tests/functional/configs/test_contract_configs.py
+++ b/tests/functional/configs/test_contract_configs.py
@@ -85,7 +85,6 @@ def model(dbt, _):
 """
 
 model_schema_yml = """
-version: 2
 models:
   - name: my_model
     config:
@@ -110,7 +109,6 @@ models:
 """
 
 model_schema_alias_types_false_yml = """
-version: 2
 models:
   - name: my_model
     config:
@@ -136,7 +134,6 @@ models:
 """
 
 model_schema_ignore_unsupported_yml = """
-version: 2
 models:
   - name: my_model
     config:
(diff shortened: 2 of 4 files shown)


tests/functional/configs/test_versioned_model_constraint.py
emmyoop, line 1: nit: It makes more sense for this test to live in /`constraints` instead of `/configs` in terms of discoverability.

tests/functional/configs/test_versioned_model_constraint.py
```diff
@@ -0,0 +1,76 @@
...
+        constraints:
+          - type: not_null
+      - name: user_name
+        data_type: text
+    versions:
+      - v: 1
+"""
+
+
+class TestVersionedModelConstraints:
```
emmyoop, line 51: Are we already testing versions column level constrains somewhere?

tests/functional/configs/test_contract_configs.py
```diff
@@ -85,7 +85,6 @@ def model(dbt, _):
 """
 
 model_schema_yml = """
-version: 2
```
emmyoop, line 88: ❤️

ChenyuLInx (approved):

core/dbt/parser/schemas.py
```diff
@@ -845,7 +845,9 @@ def parse_patch(self, block: TargetBlock[UnparsedModelUpdate], refs: ParserRef)
                 # Includes alias recomputation
                 self.patch_node_config(versioned_model_node, versioned_model_patch)
 
-                # Need to reapply this here, in the case that 'contract: {enforced: true}' was during config-setting
+                # Need to reapply setting constraints and contract checksum here,
+                # in the case that 'contract: {enforced: true}' was changed during config-setting
```
ChenyuLInx, line 849: Thanks for the commnet!

tests/functional/configs/test_versioned_model_constraint.py
gshank: I don't know. I don't see anything for it.

PR #9543, opened 2024-02-08
Move `SourceDefinition` to `dbt/artifacts`

### LINKED ISSUE #9384: [CT-3554] Define SourceDefinition contract in dbt/artifacts
### Housekeeping

### Short description

As part of an effort to move any `ArtifactMixin` subclasses upstream of other core modules into the top -level `dbt/artifacts` directory, we need to move Metric, which WritableManifest depends on.

### Acceptance criteria

SourceDefinition (and its necessary dependencies) are moved to a new dbt/artifacts directory. dbt/artifacts should have no additional dependencies on dbt.core except for from contracts.nodes which are actively being refactored. 

SourceDefinition should be split into data & functional interfaces, defined in dbt/artifacts in dbt/core respectively. Additionally, any inheritance hierarchy between concrete node classes should be flattened as part of this spike.

Should follow initial pattern set in https://github.com/dbt-labs/dbt-core/issues/9379

### Impact to Other Teams

N/A - should be a refactor. 

nope

### Context

_No response_

QMalcolm:
resolves #9384 

### Problem

We are moving data artifacts to `dbt/artifacts` in a piecewise fashion. We needed to move `SourceDefinition` as part of that.

### Solution

Moved data portion of `SourceDefinition` to `dbt/artifacts` (and all other classes that doing so required)

### Checklist


core/dbt/contracts/graph/components.py
```diff
@@ -0,0 +1,53 @@
+from dataclasses import dataclass, field
+from datetime import timedelta
+from dbt.artifacts.resources import (
+    FreshnessThreshold as FreshnessThresholdResource,
+    HasRelationMetadata as HasRelationMetadataResource,
+    Time as TimeResource,
+)
+from typing import Dict, Optional
+
+
```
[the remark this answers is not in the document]
MichelleArk: Reading through these, I'm in favor of defining the classes below in dbt/artifacts. Their methods are fairly slim and minimal, so I don't think it'd add much to the interface burden of dbt-artifacts to maintain for downstream consumers. None of them bring in dependencies on core and are fairly simple definitions based on the raw data. 

It would simplify things a good deal too instead of having this inheritance pattern for fairly intermediate components of reosurces.

core/dbt/contracts/graph/nodes.py
```diff
@@ -1249,30 +1209,14 @@ def tests(self) -> List[TestDef]:
...
-    identifier: str
-    resource_type: Literal[NodeType.Source]
-
-
-@dataclass
-class SourceDefinition(NodeInfoMixin, ParsedSourceMandatory):
-    quoting: Quoting = field(default_factory=Quoting)
-    loaded_at_field: Optional[str] = None
+class SourceDefinition(
+    NodeInfoMixin,
```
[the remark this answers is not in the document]
MichelleArk: Discussed offline: NodeInfoMixin makes a lot of sense to keep as an internal concern of dbt-core, since it is actively set in the internal node + removed prior to serialization. 

We could probably even remove the https://github.com/dbt-labs/dbt-core/pull/9543/files#diff-42cfc99b2fcd89ab75957b8999bee2b1af822f7d39953ca99f1f285c170557c1R1221 method!

The complexity is in ensuring that the SourceDefinition node is not serialized directly via Manifest.writable_manifest, but is first converted to a resource. We'll need to put a better pattern in place to avoid using nodes and resources interchangeably for the purposes of serialization

core/dbt/contracts/graph/manifest.py
```diff
@@ -1020,26 +1023,30 @@ def build_group_map(self):
                     group_map[node.group].append(node.unique_id)
         self.group_map = group_map
 
+    @classmethod
+    def _map_nodes_to_map_resources(cls, nodes_map: MutableMapping[str, NodeClassT]):
+        return {name: node.to_resource() for name, node in nodes_map.items()}
```
MichelleArk, line 1028: super nit: node_id would be more precise than 'name'

core/dbt/contracts/graph/manifest.py
```diff
@@ -1020,26 +1023,30 @@ def build_group_map(self):
...
             nodes=self.nodes,
-            sources=self.sources,
+            sources=self._map_nodes_to_map_resources(self.sources),
             macros=self.macros,
             docs=self.docs,
-            exposures=self.exposures,
-            metrics=self.metrics,
+            exposures=self._map_nodes_to_map_resources(self.exposures),
+            metrics=self._map_nodes_to_map_resources(self.metrics),
             groups=self.groups,
```
MichelleArk, line 1040: I believe we should also implement the resource_class method on Group in nodes.py and use self._map_nodes_to_map_resources on it here

core/dbt/contracts/graph/nodes.py
```diff
@@ -118,13 +114,19 @@
...
-class BaseNode(ABC, Generic[ResourceTypeT], BaseResource):
+class BaseNode(BaseResource):
     """All nodes or node-like objects in this file should have this as a base class"""
 
+    # In an ideal world this would be a class property. However, chaining @classmethod and
+    # @property was deprecated in python 3.11 and removed in 3.13. There are more
+    # complicated ways of making a class property, however a class method suits our
+    # purposes well enough
+    @classmethod
+    def resource_class(cls) -> Type[BaseResource]:
```
MichelleArk, line 126: ✨

MichelleArk (approved):
Small outstanding comment, otherwise looks amazing! Thank you ✨

core/dbt/contracts/graph/manifest.py
QMalcolm: Good catch!

Issue #9548, opened 2024-02-09
[Bug] semantic models always included as part of state:modified


MichelleArk:
### Is this a new bug in dbt-core?

- [X] I believe this is a new bug in dbt-core
- [X] I have searched the existing issues, and I could not find an existing issue for this bug

### Current Behavior

Currently, dbt always detects that semantic models have changed during state:modified, even if none of their model, description, defaults, entities, dimensions, measures, configs, primary entities, or groups have changed. 

### Expected Behavior

dbt only includes semantic models if one of the above listed properties has changed. 

### Steps To Reproduce

`dbt ls --select "state:modified"` on a project with unchanged semantic models

### Environment

```markdown
- OS:
- Python:
- dbt:
```

gshank:
I wasn't able to recreate this. Could you give me a brief reproduction test case?

MichelleArk:
I think it was already resolved by the combination of https://github.com/dbt-labs/dbt-core/pull/9549 and https://github.com/dbt-labs/dbt-core/pull/9561

PR #9549, opened 2024-02-09
Fix: SemanticModel.same_model compares model attribute

MichelleArk:
resolves https://github.com/dbt-labs/dbt-core/issues/9548

### Problem
* semantic models were always getting picked up as part of state:modifed because the SemanticModel.same_model was comparing the wrong things to each other (self.model == self.same_model (method)), which always returned false

### Solution
* Fix SemanticModel.same_model to actually compare the right attributes!

### Checklist


QMalcolm (approved):
Makes sense! Thank you for finding and fixing this ❤️

PR #9553, opened 2024-02-12
Fix #9385: Add a workflow to check if artifacts have changed

### LINKED ISSUE #9385: [CT-3555] Add detect artifact changes workflow
### Housekeeping

### Short description

We should detect whether a PR makes changes to an existing artifact schema, and fail CI if this is the case. https://github.com/marketplace/actions/prevent-file-change might do the trick!

Or, if we don't want to use an (overly-complex) public action, could do something manually: 
* check which files changed, and fail if any of those are modified
* allow new files in dbt/artifacts

### Acceptance criteria

* CI fails if any file is modified in dbt/artifacts
* failure message provides details 
* CI failure should either: not block merge, or is made to pass by setting a label (e.g. artifact_minor_upgrade)

Future work could be only failing the check if _breaking_ changes are introduced. Until then, we can manually inspect the changes and merge on red if the change is deemed safe. 

Additionally, tests in dbt/artifacts should provide signal as to whether a breaking change is made to an existing artifact because existing upgrade/downgrade logic would fail. 

### Impact to Other Teams

N/A - CI work in dbt-core

aranke:
resolves #9385 

### Problem

It isn't always clear when a file in the `core/dbt/artifacts` directory has been changed.

### Solution

This PR adds a CI check that:
1. Fails if any file in `core/dbt/artifacts` has been changed
2. Can be overridden by adding the `artifact_minor_upgrade` tag to the PR.

This solution currently uses `git diff` and doesn't rely on any external actions.

Run history: 
1. Example failure: https://github.com/dbt-labs/dbt-core/actions/runs/7866653246/job/21461212098?pr=9553
2. Example success (with label): https://github.com/dbt-labs/dbt-core/actions/runs/7866663319/job/21461233620?pr=9553
3. All runs: https://github.com/dbt-labs/dbt-core/actions/workflows/check-artifact-changes.yml

### Checklist

### DIFF
--- a/.github/workflows/check-artifact-changes.yml
+++ b/.github/workflows/check-artifact-changes.yml
@@ -2,10 +2,8 @@ name: Check Artifact Changes
 
 on:
   pull_request:
-    types:
-      - labeled
-    paths:
-      - 'core/dbt/artifacts/**'
+    types: [ opened, reopened, labeled, unlabeled, synchronize ]
+    paths-ignore: [ '.changes/**', '.github/**', 'tests/**', '**.md', '**.yml' ]
 
 jobs:
   check-artifact-changes:
--- a/.github/workflows/check-artifact-changes.yml
+++ b/.github/workflows/check-artifact-changes.yml
@@ -1,0 +1,40 @@
+name: Check Artifact Changes
+
+on:
+  pull_request:
+    paths:
+      - 'core/dbt/artifacts/**'
+
+jobs:
+  check-artifact-changes:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Check for changes in dbt/artifacts
+        id: check_changes
+        run: |
+          # Check for changes in dbt/artifacts
+          CHANGED_FILES=$(git diff --name-only origin/${GITHUB_BASE_REF} origin/${GITHUB_HEAD_REF} | grep 'core/dbt/artifacts/')
+          if [ ! -z "$CHANGED_FILES" ]; then
+            echo "CHANGED=true" >> $GITHUB_ENV
+            echo "Changed files in dbt/artifacts:"
+            echo "$CHANGED_FILES"
+          else
+            echo "CHANGED=false" >> $GITHUB_ENV
+          fi
+
+      - name: Fail CI if artifacts have changed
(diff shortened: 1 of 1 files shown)


.github/workflows/check-artifact-changes.yml
```diff
@@ -0,0 +1,42 @@
...
+            echo "$CHANGED_FILES"
+          else
+            echo "CHANGED=false" >> $GITHUB_ENV
+          fi
+
+      - name: Fail CI if artifacts have changed
+        if: env.CHANGED == 'true' && !contains(github.event.pull_request.labels.*.name, 'artifact_minor_upgrade')
+        run: |
+          echo "CI failure: Artifact changes checked in core/dbt/artifacts directory."
+          echo "To bypass this check, add the 'artifact_minor_upgrade' label to the PR."
```
MichelleArk, line 36: Could you update this message to include a link to the [breaking changes](https://github.com/dbt-labs/dbt-core/blob/main/core/dbt/artifacts/README.md#breaking-changes) section of the dbt/artifacts/README.md? 

Perhaps something like: "To bypass this check, confirm that the change is not breaking (link) and add the 'artifact_minor_upgrade' label to the PR."

.github/workflows/check-artifact-changes.yml
```diff
@@ -0,0 +1,42 @@
...
+      - name: Checkout code
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Check for changes in core/dbt/artifacts
+        id: check_changes
+        run: |
+          # Check for changes in dbt/artifacts
+          CHANGED_FILES=$(git diff --name-only origin/${GITHUB_BASE_REF} origin/${GITHUB_HEAD_REF} | grep 'core/dbt/artifacts/')
```
MichelleArk, line 23: Could we exclude matching `core/dbt/artifacts/README.md` as part of this search?

MichelleArk (approved):
couple small comments re: linking to and ignoring changes to core/dbt/artifacts/README.md, but otherwise looks great! thank you!

.github/workflows/check-artifact-changes.yml
aranke: We're ignoring all `*.md` files, so we should be fine:

https://github.com/dbt-labs/dbt-core/pull/9553/files#diff-a7bcbf79c3b395a73ebf1356d7c6a129c82f2f01d385cfbf82cfc4af21f5b3c0R6

PR #9561, opened 2024-02-12
Anther Fix for Semantic model same content

ChenyuLInx:
resolves another part of #9548
Per conversation with @graciegoheen and Jerco, we should not compare the node relations for same content


MichelleArk (approved):

jtcohen6:
thank you @ChenyuLInx !

PR #9585, opened 2024-02-15
Handle exceptions during node execution more elegantly.

### LINKED ISSUE #9583: Do Not Hang on Worker Thread Exceptions

### Current Behavior

For a while now, when there was an exception during the execution of one of dbt's implementations of BaseRunner, dbt would hang rather than crash or report the exception message.

### Expected Behavior

When a node fails to execute with an exception, we should record the failure as we would any other node error, and proceed normally.

peterallenwebb:
resolves #9583

### Problem

Because of a bug in Python's thread pool mechanism, an uncaught exception on dbt's task threads would cause it to hang.

### Solution

Eliminate a follow-up exception that caused a top-level exception to escape.

### Checklist


core/dbt/task/runnable.py
```diff
@@ -221,19 +221,36 @@ def call_runner(self, runner: BaseRunner) -> RunResult:
...
+                        fire_event(
+                            NodeFinished(
+                                node_info=runner.node.node_info,
+                                run_result=result.to_msg_dict(),
+                            )
                         )
-                    )
+                    else:
+                        msg = f"Exception on worker thread. {thread_exception}"
+                        fire_event(Note(msg=msg))
```
emmyoop, line 242: Should we use the `GenericExceptionOnRun` event instead here?

emmyoop (approved):

core/dbt/task/runnable.py
gshank: I'd vote for this

gshank (approved):

PR #9589, opened 2024-02-16
remove uses of Replaceable class

### LINKED ISSUE #7802: [CT-2652] Remove "Replaceable" class usage as much as possible

emmyoop:
resolves #7802 

### Problem

We pull in a Replaceable class which contains a "replace" method in a lot of our classes. This method is designed to replace the current object with a newly constructed one based on kwargs, and is in general not a very good idea. If we want to replace a current object it would be better to be a lot more explicit and intentional about what's happening.

### Solution

Remove the usage Replaceable class where it's found in dbt-core

### Checklist

### DIFF
--- a/.changes/unreleased/Under the Hood-20240216-104002.yaml
+++ b/.changes/unreleased/Under the Hood-20240216-104002.yaml
@@ -1,0 +1,6 @@
+kind: Under the Hood
+body: Remove uses of Replaceable class
+time: 2024-02-16T10:40:02.25455-06:00
+custom:
+  Author: emmyoop
+  Issue: "7802"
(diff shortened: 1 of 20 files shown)


MichelleArk (approved):
🔥

PR #9231, opened 2023-12-06
made class changing directory a context manager.

rariyama:
resolves #8997 

### Problem

There is a difference in behaviour between `dbt cli` and `dbt.invoke`.
For example, When executing `dbtRunner().invoke(["deps", "--project-dir", "project_dir"])` in python, directory
 changes to `project_dir`, even though it doesn't affect when executing `dbt deps`.

### Solution

The reason why directory changes is `dbt.task.base.move_to_nearest_project_dir()` is called in `CleanTask `, `DepsTask ` and `InitTask `. Directory changes to the argument of the function.  
It'll be possible to resolve this problem by using those classes as a context manager. The change will happen only in with statement.

### Checklist

- [ ] I have read [the contributing guide](https://github.com/dbt-labs/dbt-core/blob/main/CONTRIBUTING.md) and understand what's expected of me  
- [ ] I have run this code in development and it appears to resolve the stated issue  
- [ ] This PR includes tests, or tests are not required/relevant for this PR
- [ ] This PR has no interface changes (e.g. macros, cli, logs, json artifacts, config files, adapter interface, etc) or this PR has already received feedback and approval from Product or DX
- [ ] This PR includes [type annotations](https://docs.python.org/3/library/typing.html) for new and modified functions

PR #9596, opened 2024-02-19
Fix/changes current working dir when using a dbt project dir

### LINKED ISSUE #8997: [CT-3319] [Bug] Running dbt.invoke(['deps']) changes the current working directory when using a DBT_PROJECT_DIR

### Current Behavior

We use a nested directory structure like:
projects/
 - project_a
 - project_b
 
 When using the env var `DBT_PROJECT_DIR` and invoking dbt using python running `deps` the current working directory is modified to the be the `DBT_PROJECT_DIR` which causes issues for the remaining python script. 

### Expected Behavior

The working directory should not be affected by the invocation. 

### Steps To Reproduce

Use a multi project setup and invoke dbt deps using python whilst using the DBT_PROJECT_DIR env var. 

### Relevant log output

```shell
[MainThread] 2023-11-03 15:44:28,936 root         DEBUG    Current directory is /Users/josh.elston-green/repos/my_repo
15:44:28  Running with dbt=1.6.0
15:44:29  Installing calogica/dbt_expectations
15:44:33  Installed from version 0.8.5
15:44:33  Updated version available: 0.10.1
15:44:33  Installing calogica/dbt_date
15:44:33  Installed from version 0.7.2
15:44:33  Updated version available: 0.10.0
15:44:33  Installing dbt-labs/dbt_utils
15:44:33  Installed from version 1.1.0
15:44:33  Updated version available: 1.1.1
15:44:33  Installing dbt-labs/spark_utils
15:44:33  Installed from version 0.3.0
15:44:33  Up to date!
15:44:33  Installing dbt-labs/codegen
15:44:34  Installed from version 0.9.0
15:44:34  Updated version available: 0.11.0
15:44:34  Installing elementary-data/elementary
15:44:34  Installed from version 0.8.1
15:44:34  Updated version available: 0.12.1
15:44:34  Installing dbt-labs/dbt_project_evaluator
15:44:34  Installed from version 0.6.1
15:44:34  Updated version available: 0.8.0
15:44:34  
15:44:34  Updates available for packages: ['calogica/dbt_expectations', 'calogica/dbt_date', 'dbt-labs/dbt_utils', 'dbt-labs/codegen', 'elementary-data/elementary', 'dbt-labs/dbt_project_evaluator']                 
Update your versions in packages.yml, then run dbt deps
[MainThread] 2023-11-03 15:44:34,916 root         DEBUG    Current directory is /Users/josh.elston-green/repos/my_repo/projects/my_project
```

### Environment

```markdown
- OS: MacOS 14.0 (23A344)
- Python: 3.11
- dbt: 1.6.0
```

spark

### Additional Context

Databricks

rariyama:
resolves #8997 
I closed [the previous PR](https://github.com/dbt-labs/dbt-core/pull/9231) by mistake, then I submit this one.

### Problem

There is a difference in behaviour between `dbt cli` and `dbt.invoke`.
For example, When executing `dbtRunner().invoke(["deps", "--project-dir", "project_dir"])` in python, directory
 changes to `project_dir`, even though it doesn't affect when executing `dbt deps`.

### Solution

The reason why directory changes is `dbt.task.base.move_to_nearest_project_dir()` is called in `CleanTask `, `DepsTask ` and `InitTask `. Directory changes to the argument of the function.  
It'll be possible to resolve this problem by using those classes as a context manager. The change will happen only in with statement.

### Checklist


rariyama:
Hi @ChenyuLInx . Thanks for your comments.
I agree with your idea, then I updated my code as follows on this commit https://github.com/dbt-labs/dbt-core/pull/9596/commits/31ba979062d8c10d168a1847288c1aecc9fbeced.
- Added `__enter__` and `__exit__` to `BaseTask`
- each child TaskClass inherits that class and behaves as a context manager.

I would like to add test code for this update. I feel like it's good enough to add the following code to integration test but what do you think about?
```
    def test_dbtRunner_invoke(self):
        pd = "<test-project-path>"
        dr = dbtRunner()
        commands = ["deps", "clean", "init"]

        for command in commands:
            req = [command, "--project-dir", pd, "--profiles-dir", pd]
            before_dir = os.getcwd()
            res = dr.invoke(req)
            self.assertEqual(res.success, True)
            # MEMO: Test whether directory is not changed after command is invoked.

            self.assertEqual(before_dir, os.getcwd())
```

dbeatty10:
@rariyama I saw your message in https://github.com/dbt-labs/dbt-core/issues/8997#issuecomment-2290454692 and checked the comments in this PR.

Could you do these two things? After that we can have one of our engineers take another look:

1. Resolve the conflicts
1. Add the test below

```python
    def test_runner_invoke(self, project):
        project_dir = project.project_root
        runner = dbtRunner()
        commands = ["deps", "clean", "init"]
        
        # TODO: change to some other directory other than the dbt project root

        for command in commands:
            before_dir = os.getcwd()
            res = runner.invoke([command, "--project-dir", project_dir, "--profiles-dir", project_dir])
            self.assertEqual(res.success, True)
            self.assertEqual(before_dir, os.getcwd())
```

The key thing will be to make sure the test actually captures the scenario in the bug report. i.e., you'll want to make sure that the test fails without your changes, but the test passes after you changes.

rariyama:
Hi @dbeatty10. Thank you for checking my comment.
I resolved conflict in this commit https://github.com/dbt-labs/dbt-core/pull/9596/commits/fc22afddbd48b3938193345a300530393495925c and add test assertion in the commit https://github.com/dbt-labs/dbt-core/pull/9596/commits/c75835b937245c874d4a5e8a6356407c4eac6a33.

I confirmed the results of both unittest and integration test are as expected by following ways.
- Added test assertion without my changes and integration test fails.
    - Added test assertion to the source code of main branch and test failed.
- Added my changes and both unit test and integration test passes.
    - It means my changes can resolve the issue.

I run the test code by the command `make test` and `make integration`.

If there's no remaining tasks I have to do, could you please have the engineers review it?

rariyama:
Hello Team. Could someone please review my PR?

core/dbt/tests/util.py
```diff
@@ -91,7 +91,12 @@ def run_dbt(
     if profiles_dir and "--profiles-dir" not in args:
         args.extend(["--profiles-dir", profiles_dir])
     dbt = dbtRunner()
+
+    before_dir = os.getcwd()
     res = dbt.invoke(args)
+    after_dir = os.getcwd()
+    # The directory has not been changed after running dbt command.
+    assert before_dir == after_dir
```
dbeatty10, line 99: It seems like this assertion would be better if it were directly within a test rather than here. Maybe [here](https://github.com/dbt-labs/dbt-core/blob/84eb0ff6720ec82ce4015c2657d512bf51381732/tests/functional/dbt_runner/test_dbt_runner.py#L14) instead?

What do you think @ChenyuLInx?

core/dbt/task/deps.py
```diff
@@ -89,16 +89,14 @@ def _create_packages_yml_entry(package: str, version: Optional[str], source: str
...
         # N.B. This is a temporary fix for a bug when using relative paths via
         # --project-dir with deps.  A larger overhaul of our path handling methods
         # is needed to fix this the "right" way.
         # See GH-7615
-        project.project_root = str(Path(project.project_root).resolve())
-        self.project = project
-
-        move_to_nearest_project_dir(project.project_root)
         self.cli_vars = args.vars
+        self.project = project
```
dbeatty10, line 99: @rariyama Did you happen to check if preserving [this order](https://github.com/dbt-labs/dbt-core/blob/84eb0ff6720ec82ce4015c2657d512bf51381732/core/dbt/task/deps.py#L92-L101) worked or not?

I'm wondering if you changed the ordering intentionally or if it was just an accidental side-effect when resolving merge conflicts.

i.e., I'm curious if this would work (which is basically identical to [here](https://github.com/dbt-labs/dbt-core/blob/84eb0ff6720ec82ce4015c2657d512bf51381732/core/dbt/task/deps.py#L92-L101) except for removing the `move_to_nearest_project_dir(project.project_root)` line):

```python
        super().__init__(args=args)

        # N.B. This is a temporary fix for a bug when using relative paths via
        # --project-dir with deps.  A larger overhaul of our path handling methods
        # is needed to fix this the "right" way.
        # See GH-7615
        project.project_root = str(Path(project.project_root).resolve())
        self.project = project

        self.cli_vars = args.vars
```

core/dbt/tests/util.py
ChenyuLInx: I think this is great!! Also makes it checking all future commands.

ChenyuLInx (commented):
Thanks @rariyama for thinking through this and submit this PR. This change makes sense to me!
Since this is something that modifies the behavior of directory I would like to play with it a bit more locally before giving it the approve and merge.

core/dbt/tests/util.py
rariyama: I added test code to `tests/functional/dbt_runner/test_dbt_runner.py` in this commit.
https://github.com/dbt-labs/dbt-core/pull/9596/commits/26ec953a61bc819e1658457bfc328943a3ff276f

core/dbt/task/deps.py
rariyama: I fixed the order of variables as you wrote in this commit and it worked fine.
https://github.com/dbt-labs/dbt-core/pull/9596/commits/4191735ea2c32f83ed4c1ecc1f6816d1b462ea20

rariyama:
@ChenyuLInx 
Thank you for checking my PR. Please try to run on your local environment. If there are something you want to ask or I need to fix, please tell me about that.

core/dbt/task/deps.py
```diff
@@ -202,6 +200,7 @@ def lock(self) -> None:
         fire_event(DepsLockUpdating(lock_filepath=lock_filepath))
 
     def run(self) -> None:
+        move_to_nearest_project_dir(self.args.project_dir)
```
dbeatty10, line 203: Did `move_to_nearest_project_dir(self.args.project_dir)` _need_ to move to this location in order for things to work?

Or was it "nice-to-have" for some reason?

rariyama: There is a reason for moving the `move_to_nearest_project_dir()` function. The `DepsTask` class inherits from the `BaseTask` class and is designed to behave as a context manager. In `BaseTask`, the current directory is obtained within `__enter__()`, and `os.chdir()` is called within `__exit__()` to return to the previously obtained directory.

Normally, since `__init__()` is executed before `__enter__()`, if `move_to_nearest_project_dir()` is called inside `__init__()`, the directory after the move will be recorded, and it will not return correctly to the original directory.

The sequence of events can be described as follows:

1. `__init__().move_to_nearest_project_dir()` # Moves to the project directory.
2. `__enter__().os.getcwd()` # Gets the current directory, which is **the project directory.**
3. `__exit__().os.chdir()` # Moves back to the project directory obtained in step 2.

By moving the `move_to_nearest_project_dir()` function to the `run()` function, the order of execution changes as follows:

1. `__enter__().os.getcwd()` # Gets the current directory, i.e., the directory where `dbt run` is executed.
2. `DepsTask.run().move_to_nearest_project_dir()` # Moves to the project directory to execute `dbt deps`.
3. `__exit__().os.chdir()` # Returns to the directory obtained in step 1.

This is why it was necessary to move the `move_to_nearest_project_dir()` function.

dbeatty10: That is an amazing explanation -- thank you!

ChenyuLInx:
@rariyama looks like the new test you just introduced is failing on the deps command.(I think this is due to missing `project` fixture as input arg to the test. 
`def test_directory_does_not_change(self, dbt: dbtRunner, project) -> None:`
I pushed a commit to your branch as this is a small fix, hope that's okay with you.
I did some local testing with dbtRunner and the behavior looks good to me.

ChenyuLInx (approved):

ChenyuLInx:
@rariyama I went ahead and merged this change! Thank you so much for fixing it!

rariyama:
Hi @ChenyuLInx 
Thank you for the review. If I have another request for a review in the future, I would appreciate your assistance again.

PR #9614, opened 2024-02-21
Paw/restrict to protobuf 4

### LINKED ISSUE #9566: Google protobuf 5 is incompatible
### Description

We specify google.protobuf: "protobuf>=4.0.0".  Google has come out with a release candidate for version 5 which is incompatible with our code because keyword parameters to MessageToJson and MessageToDict have been renamed. There may be other incompatibilities.  We need to either change our upper limit to < 5.0.0 or support google.protobuf 5 in our code.

This was seen internally because a pipeline uses `pip install --pre`.

### Acceptance Criteria

An incompatible version of google protobuf cannot be installed.

Yes. We need to limit the protobuf version in all versions of dbt Core that use google protobuf functions.

peterallenwebb:
resolves #9566 

### Problem

Existing code is incompatible with upcoming protobuf 5, and there is no simple way to make it compatible with both 4 and 5.

### Solution

Restrict protobuf version to <5.0.0

### Checklist


ChenyuLInx (approved):
Should we create a follow up ticket to update our code to make it compatible with protobuf 5.0? Assuming people might install dbt-core with other things which might require 5.0?

QMalcolm:
> Should we create a follow up ticket to update our code to make it compatible with protobuf 5.0? Assuming people might install dbt-core with other things which might require 5.0?

We absolutely should

akan72:
Should the restriction have also been applied in `core/setup.py` https://github.com/dbt-labs/dbt-core/blob/main/core/setup.py#L82?

Google just released `protobuf==5.26.0` in the past hour and caused this issue https://github.com/dbt-labs/dbt-core/issues/9759

PR #9635, opened 2024-02-22
Fix `Check Artifact Changes` Action

### LINKED ISSUE #9613: [CI] check-artifact-changes workflow fails when no dbt/artifacts files modified

### Current Behavior

Currently, the check-artifact-changes workflow fails on 'Check for changes in core/dbt/artifacts' even when there are no core/dbt/artifacts files modified in the PR: 

Example failing run: https://github.com/dbt-labs/dbt-core/actions/runs/7984103257/job/21800375206?pr=9605

### Expected Behavior

check-artifact-changes should be a no-op when no core/dbt/artifacts files have been modified.

### Steps To Reproduce

Create a PR with no modifications to dbt/core/artifacts, run CI. Example: https://github.com/dbt-labs/dbt-core/pull/9605/files

### Relevant log output

```shell
# Check for changes in dbt/artifacts
  CHANGED_FILES=$(git diff --name-only origin/${GITHUB_BASE_REF} origin/${GITHUB_HEAD_REF} | grep 'core/dbt/artifacts/')
  if [ ! -z "$CHANGED_FILES" ]; then
    echo "CHANGED=true" >> $GITHUB_ENV
    echo "Changed files in core/dbt/artifacts:"
    echo "$CHANGED_FILES"
  else
    echo "CHANGED=false" >> $GITHUB_ENV
  fi
  shell: /usr/bin/bash -e {0}
Error: Process completed with exit code 1.
```

### Environment

```markdown
CI
```

_No response_

### Additional Context

_No response_

QMalcolm:
resolves #9613

### Problem

We were using `git diff` to check if any files had changed in `core/dbt/artifacts`. However, our `git diff` usage was including any changes that happened on `main` which the PR branch did not have. This caused the check to error as it was detecting changes which weren't actually there.

### Solution

We've switched the check from using `git diff` to `dorny/paths-filter`, which is what we use for checking for changelog existence as well. The `dorny/paths-fitler` includes logic for excluding changes that are on main but not the PR branch (which is what want to happen).

### Checklist

### DIFF
--- a/.github/workflows/schema-check.yml
+++ b/.github/workflows/schema-check.yml
@@ -26,7 +26,7 @@ permissions: read-all
 
 env:
   LATEST_SCHEMA_PATH: ${{ github.workspace }}/new_schemas
-  SCHEMA_DIFF_ARTIFACT: ${{ github.workspace }}//schema_schanges.txt
+  SCHEMA_DIFF_ARTIFACT: ${{ github.workspace }}/schema_schanges.txt
   DBT_REPO_DIRECTORY: ${{ github.workspace }}/dbt
   SCHEMA_REPO_DIRECTORY: ${{ github.workspace }}/schemas.getdbt.com
 
--- a/.github/workflows/check-artifact-changes.yml
+++ b/.github/workflows/check-artifact-changes.yml
@@ -10,6 +10,7 @@ on:
 jobs:
   check-artifact-changes:
     runs-on: ubuntu-latest
+    if: ${{ !contains(github.event.pull_request.labels.*.name, 'artifact_minor_upgrade') }}
     steps:
       - name: Checkout code
         uses: actions/checkout@v4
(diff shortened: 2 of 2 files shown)


QMalcolm:
Here is an example PR that is behind main that previously would have failed the artifact check but now passes: https://github.com/dbt-labs/dbt-core/actions/runs/8011901819/job/21886077928?pr=9637
Here is an example PR that has artifact changes and thus still (appropriately) fails with the updated check: https://github.com/dbt-labs/dbt-core/pull/9636

.github/workflows/schema-check.yml
```diff
@@ -26,7 +26,7 @@ permissions: read-all
 
 env:
   LATEST_SCHEMA_PATH: ${{ github.workspace }}/new_schemas
-  SCHEMA_DIFF_ARTIFACT: ${{ github.workspace }}//schema_schanges.txt
```
emmyoop, line 29: ❤️

emmyoop (approved):
Looks good!

PR #9677, opened 2024-02-26
Do not add duplicate input_measures when parsing metrics

### LINKED ISSUE #9360: [CT-3541] [Bug] Duplicate Metric InputMeasures cause query time errors in MetricFlow

### Current Behavior

After parsing a metric, we process a metric to populate it's `input_measures`. This happens in [_process_metric_node](https://github.com/dbt-labs/dbt-core/blob/84dfb22cd59edee0d3e2bca6041288449bfbe7c4/core/dbt/parser/manifest.py#L1563). It's possible for there to be _duplicate_ input measures in the `input_measures` list. When there is, a validation warning shows up in the output. Additionally on the MetricFlow side, this causes a query time error for the metric.

### Expected Behavior

There shouldn't be duplicate `InputMeasure`s in a `Metric`'s `Metric.type_params.input_measures` and thus no validation warning should be raised.

### Steps To Reproduce

1. Add a ratio metric that depends on two simple metrics that have the same underlying measure
  a. Alternatively pull down my [qmalcolm--duplicate-input-measure-repro](https://github.com/dbt-labs/jaffle-sl-template/compare/main...qmalcolm--duplicate-input-measure-repro) from [jaffle-sl-template](https://github.com/dbt-labs/jaffle-sl-template)
2. Run `dbt parse`
3. Observe validation warning in output

### Relevant log output

```shell
PydanticMetric ratio_new_customers has multiple identical input measures specifications for measure customers_with_orders. This might be hiding a semantic error. Input measure specification: name='customers_with_orders' filter=None alias=None join_to_timespine=False fill_nulls_with=None.
```

### Environment

```markdown
- OS: macOS (Sonoma 14.2)
- Python: 3.8
- dbt: 1.7.4
```

snowflake

### Additional Context

The Semantic Layer team, is going to work on mitigating this in MetricFlow as an immediate fix, however we should also fix this on the Core side.

gshank:
resolves #9360

### Problem

In some case metrics were ending up with duplicate input_measures when they came from different measures. This caused validation errors and runtime errors later on.

### Solution

Instead of just appending or extending input_measures, check that the input_measure is not equal to existing input_measures, in a "add_input_measure" method.

### Checklist


core/dbt/artifacts/resources/v1/metric.py
```diff
@@ -92,6 +92,12 @@ class MetricTypeParams(dbtClassMixin):
     metrics: Optional[List[MetricInput]] = None
     conversion_type_params: Optional[ConversionTypeParams] = None
 
+    def add_input_measure(self, input_measure) -> None:
```
MichelleArk, line 95: nit: missing type annotation on input_measure

MichelleArk: also, there's no 'node' equivalent for a MetricTypeParams, but I'm still not sure we want to have this type of mutating method implemented on a resource. 

It's more verbose, but what do you think about implementing this as either `add_input_measure_to_type_params` or `add_input_measure` (since i think its unambiguous where the input measures are) on the `Metric` node in contracts.graph.nodes directly?

gshank: I can move it to the Metric node.

QMalcolm: To push this farther, do we need a mutating method? An alternative would be to just use a `set` in `_process_metric_node` for constructing the list, and then at the end of the function convert via `type_params.input_measures=list(measures_set)`. I guess the drawback would be if anywhere else in the code base wanted to add additional measures to the list after the fact, but I'm not sure that's even a workflow we'd want to encourage or support.

core/dbt/contracts/graph/nodes.py
```diff
@@ -1400,6 +1401,12 @@ def same_contents(self, old: Optional["Metric"]) -> bool:
             and True
         )
 
+    def add_input_measure(self, input_measure: MetricInputMeasure) -> None:
```
QMalcolm, line 1404: Do we need a mutating method? An alternative would be to just use a `set` in `_process_metric_node` for constructing the list, and then at the end of the function convert via `type_params.input_measures=list(measures_set)`. I guess the drawback would be if anywhere else in the code base wanted to add additional measures to the list after the fact, but I'm not sure that's even a workflow we'd want to encourage or support.

QMalcolm (commented):
One thing worth discussing that I'm raising as a review for visibility.

core/dbt/artifacts/resources/v1/metric.py
gshank: I looked at this option, but it doesn't fit into that big if/elif/elif/elif/else very well.  The "add_input_measure" method is similar to other methods that we use such as "add_node", and it's usable in other places, should they occur. If you wanted to change the definition of "input_measures" to be a set...

QMalcolm (approved):
I think this is perfectly fine to move forward with as is, and thank you for doing the work 🙂 

---

I'd still prefer us to get away from creating additional setter methods for this workflow as I believe it should be a method of the underlying data type instead. This is partially because we're doing same thing in multiple places throughout the codebase (f.x. [add_macro](https://github.com/dbt-labs/dbt-core/blob/8a395e928d1016368712e80641642a57e59590b4/core/dbt/artifacts/resources/v1/components.py#L22), [add_node](https://github.com/dbt-labs/dbt-core/blob/8a395e928d1016368712e80641642a57e59590b4/core/dbt/artifacts/resources/v1/components.py#L31)) and partially because there are data types specifically geared for giving uniqueness, like a `Set`.

The "easiest" way to move away from this pattern be to use a `Set`, as the order of the inputs does not matter. Unfortunately, the protocol spec in `dbt-semantic-interfaces` defines it as a `Sequence` which is a subclass of `Collection`. The `List` type is a subclass of `Sequence` whereas `Set` is a direct subclass of `Collection` meaning we can't use `Set` as the base type (hence my previous perhaps contrived solution of using a `Set` and converting to `List`.

An alternative solution would be to just subclass `List` and override the `append` method via something like the following
```
class UniqueList(list):
    def append(self, item):
        for thing in self.__iter__():
            if thing == item:
                return
        super().append(item)
```

gshank:
The UniqueList idea is interesting. It does feel like it's a fuzzy substitute for a set, but in cases where the type can't easily be changed it makes sense.  We do have quite a few places where we do similar things, as you mentioned. This should maybe be added to the tech debt list.

PR #9682, opened 2024-02-27
Add factory wrappers to renamed_relations

### LINKED ISSUE #9681: [Bug] renamed_relations not providing overriding hook to adapters

### Current Behavior

Surfaced from debugging this transaction error in this [Redshift issue](https://github.com/dbt-labs/dbt-redshift/issues/693). 

In short, pre-`dbt-adapters` `renamed_relations` is [defined here in core](
https://github.com/dbt-labs/dbt-core/blob/0a6d0c158e5a1956bd48e53793ca3b28faee2b34/core/dbt/adapters/base/relation.py#L44).

It is used to know when a `rename` instead of a `get_drop_sql` macro can be called. This was implemented without mechanisms to support for adapter overrides leading to obscure behavior in Redshift adapters and presumably Bigquery, Databricks (they've been notified via Slack), Snowflake, and so on.

For verbose details, see [this lovely comment](https://github.com/dbt-labs/dbt-redshift/issues/693#issuecomment-1965526194) from @mikealfare .

### Expected Behavior

`renamed_relations` should not return `null`, but that's what it's doing. So boolean checks using it return "falsey" results.

### Steps To Reproduce

This test in Redshift or other relevant repos. 

```python
from dbt.adapters.redshift.relation import RedshiftRelation
from dbt.contracts.relation import RelationType # dbt.adapters.contracts.relation in dbt ~=1.8


def test_renameable_relation():
    relation = RedshiftRelation.create(
        database="my_db",
        schema="my_schema",
        identifier="my_table",
        type=RelationType.Table,
    )
    assert relation.renameable_relations == frozenset({
        RelationType.View,
        RelationType.Table
    })
```

### Relevant log output

_No response_

### Environment

```markdown
- OS: 13.2.1
- Python: 3.11.0
- dbt: 1.7.3
```

redshift

### Additional Context

Per Mike
> The fix needs to be implemented in dbt-core>=1.6 and dbt-adapters>=1.8. The tests need to be implemented in dbt-snowflake>=1.6, dbt-redshift>=1.6, and dbt-bigquery>=1.7.

VersusFacit:
resolves #9681 

### Problem

> (*Lt. Montgomery Scott voice*) Captain, they be missing their wrappers. I can't hold 'em together for much longer.

We need the `field` wrappers to allow these to be overridden correctly in the runtime.

### Solution

Wrap them, so that they are in compliance with other such members of the dataclass.

### Checklist

- [X] This PR includes tests, or tests are not required/relevant for this PR (they are in linked adapter PRs)


VersusFacit:
@check-cla

VersusFacit:
@cla-bot check

plugins/postgres/dbt/adapters/postgres/relation.py
mikealfare, line 1: I know in related PRs I mentioned that I don't think we need to make this change. However I'm pretty sure `dbt-postgres` gets imported by some other adapters, in which case, these adapters would need to override this setting. So I think we should keep this one.

mikealfare (approved):

plugins/postgres/dbt/adapters/postgres/relation.py
VersusFacit: It's 100% necessary. Chatted with Mike about this. If you look at earlier CI logs, you can see it failing as an empty `frozenset` until you add this wrapper.

mikealfare:
lol, the codecov check failed because we merged 1.7.latest into this branch, which triggered the check to look at all files, and somehow that was less than 1.7.latest. Magical.

PR #9700, opened 2024-02-28
Throw ParsingError for invalid PK definition on constraints

### LINKED ISSUE #9581: [Constraints] Add parse-time validation that ensures only a single primary_key constraint is defined
### Housekeeping

### Short description

Currently, no parse-time error is raised when dbt sees a contract that defines a primary_key at both the model-level and the column-level. Additionally, no parse-time error is raised when multiple primary_keys are defined at the column-level.

In both cases, this will lead to a database error across many (probably all) data warehouses. Validated here: https://github.com/dbt-labs/jaffle_shop/pull/106

Additionally, our documented guidance in dbt is to define tests/constraints that work across multiple columns at the model-level. If a warehouse supports defining multiple primary keys at the column-level, it is really working by coincidence from the perspective of dbt. 

### Acceptance criteria

We raise an error at parse-time if: 
1. a primary_key constraint is defined at both the model-level and column-level
2. a primary_key constraint is defined on multiple columns within a model

### Suggested Tests

1. Scenario where a primary key is defined across model and column level constraints
2. scenario where a primary key is defined on multiple columns
3. same as (1) and (2), but for versioned models

### Impact to Other Teams

N/A

No.

### Context

Additionally, [I believe we only parse + serialize constraints if the contract is enforced currently](https://github.com/dbt-labs/dbt-core/blob/main/core/dbt/parser/schemas.py#L889). So if config.contract is False, we don't even store the constraints as provided. This was probably done to avoid raising errors related to warehouse-specific errors related to constraint prerequisites at parse time. However, we should probably still be parsing the constraints (storing constraints defined in project spec onto internal node + serializing them) and running any parse-time validation against them at least.

emmyoop:
resolves #9581

### Problem

Users receive database errors when PKs are defined on multiple columns or at both the model  and column levels.

### Solution

Throw a parsing error before it hits the DB

### Checklist

### DIFF
--- a/core/dbt/parser/schemas.py
+++ b/core/dbt/parser/schemas.py
@@ -887,7 +887,7 @@ class ModelPatchParser(NodePatchParser[UnparsedModelUpdate]):
     def patch_constraints(self, node, constraints):
         contract_config = node.config.get("contract")
         if contract_config.enforced is True:
-            self._validate_constraint_prerequisites(node, constraints)
+            self._validate_constraint_prerequisites(node)
 
             if any(
                 c for c in constraints if "type" not in c or not ConstraintType.is_valid(c["type"])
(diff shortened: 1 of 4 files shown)


gshank (approved):

core/dbt/parser/schemas.py
```diff
@@ -927,6 +929,28 @@ def _validate_constraint_prerequisites(self, model_node: ModelNode):
         if str(model_node.language) != "sql":
             errors.append(f"Language Error: Expected 'sql' but found '{model_node.language}'")
 
+        # check for primary key constraints defined at the column level
+        pk_col: List[str] = []
+        for col in model_node.columns.values():
+            for constraint in col.constraints:
+                if constraint.type == "primary_key":
```
MichelleArk, line 936: nit: this could be `constraint.type == ConstraintType.primary_key` instead of a hard-coded string.

MichelleArk: ditto for the comparisons below

core/dbt/parser/schemas.py
```diff
@@ -927,6 +929,28 @@ def _validate_constraint_prerequisites(self, model_node: ModelNode):
         if str(model_node.language) != "sql":
```
MichelleArk, line 929: I actually think that from [line 923](https://github.com/dbt-labs/dbt-core/pull/9700/files#diff-d6c689f513ec4e0f5ddbae6bdc92a08dfcb930d5fa8b2fa70c87fe9c366a53e7R923) -- this should be a separate method that is called regardless of whether `contract_config.enforced` is true or not. Currently, `_validate_constraint_prerequisites` is only called when the contract enforcement is enabled.

The warn_or_error issues above seem directly related to constraint validation, whereas the constraint parsing is independent and should be consistent on every project regardless of whether the contract is being validated or not.

MichelleArk (approved):
code quality issue looks independent from these changes as its also failing on main: https://github.com/dbt-labs/dbt-core/actions/runs/8088531062/job/22102753563

PR #9725, opened 2024-03-02
Remove duplicate dependencies already specified in dbt-common

aranke:
resolves https://github.com/dbt-labs/dbt-common/issues/14

### Problem

Remove duplicate dependencies between `dbt-common` and `dbt-core`.
 
### Solution

Analysis here: https://docs.google.com/spreadsheets/d/1vl8-kFEk6z_xEpzh5tmtTRuYaMGmV22V9ZnrFDJuvfo/edit?usp=sharing

We can safely remove the `agate`, `Jinja2`, `mashumaro[msgpack]`, `requests`, `pathspec`, `protobuf`, `typing-extensions` dependencies in `dbt-core` since these should now come in transitively via `dbt-common`.

My idea here is to keep `dbt-common` as the single source of truth for dependencies, since every dependency required in `dbt-common` is also required in `dbt-core` (except maybe `typing-extensions`).

### Checklist

### DIFF
--- a/core/setup.py
+++ b/core/setup.py
@@ -47,12 +47,6 @@ setup(
     },
     install_requires=[
         # ----
-        # dbt-core uses these packages deeply, throughout the codebase, and there have been breaking changes in past patch releases (even though these are major-version-one).
-        # Pin to the patch or minor version, and bump in each new minor version of dbt-core.
-        "agate>=1.7.0,<1.8",
-        "Jinja2>=3.1.3,<4",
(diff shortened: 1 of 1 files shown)


emmyoop (changes requested):
For anything we import directly in core we should still define the dependency here, even though it's defined in common as well.  We've been bitten by removing dependencies or by bumping the pins without realizing it affects packages that also depend on it.

aranke:
Thanks, @emmyoop, I'm going to close this PR.

PR #9733, opened 2024-03-07
Scrub secret vars

### LINKED ISSUE #7247: [CT-2342] [Feature] Support secret `--vars`, by scrubbing their values from logs (like env vars)

### Current Behavior

When the `RequiredVarNotFoundError` exception is raised anywhere within the dbt execution it leaks secrets to stdout / logs since it does not mask `DBT_ENV_SECRET_*` vars in the exception message.

### Expected Behavior

`DBT_ENV_SECRET_*` vars should be masked in the exception message for `RequiredVarNotFoundError`

### Steps To Reproduce

1. Trigger a `RequiredVarNotFoundError` with `DBT_ENV_SECRET_*` vars
2. See plain text secrets in stdout

### Relevant log output

_No response_

### Environment

```markdown
- OS: any
- Python: any
- dbt: 1.4.3+
```

_No response_

### Additional Context

_No response_

nielspardon:
resolves #7247 

### Problem

Currently, dbt only supports secret env variables and only scrubs the values of those secret env variables from log messages. Secret variables being provided via the `--vars` flag are not being scrubbed and the plaintext values appear in a couple of places.

### Solution

With this PR secret variables being provided via the `--vars` flag are also scrubbed just like env variables reusing the same prefix as for env variables.

- Scrubs secret vars in RequiredVarNotFoundError
- Scrubs secret vars in StateCheckVarsHash event
- Scrubs secret vars in run results

### Checklist

### DIFF
--- a/tests/functional/context_methods/test_cli_vars.py
+++ b/tests/functional/context_methods/test_cli_vars.py
@@ -3,7 +3,13 @@ import yaml
 
 from tests.fixtures.dbt_integration_project import dbt_integration_project  # noqa: F401
 
-from dbt.tests.util import run_dbt, get_artifact, write_config_file
+from dbt.tests.util import (
+    run_dbt,
+    run_dbt_and_capture,
+    get_logging_events,
+    get_artifact,
+    write_config_file,
+)
 from dbt.tests.fixtures.project import write_project_files
 from dbt.exceptions import DbtRuntimeError, CompilationError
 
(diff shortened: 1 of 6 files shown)


nielspardon:
@dbeatty10 any reviewers available for this PR?

tests/functional/context_methods/test_cli_vars.py
```diff
@@ -206,3 +206,38 @@ def test_vars_in_selectors(self, project):
...
+        return {
+            "schema.yml": models_scrubbing__schema_yml,
+            "simple_model.sql": models_scrubbing__simple_model_sql,
+        }
+
+    def test__run_results_scrubbing(self, project):
+        results = run_dbt(["run", "--vars", "{DBT_ENV_SECRET_simple: abc, unused: def}"])
+        assert len(results) == 1
+        results = run_dbt(["test", "--vars", "{DBT_ENV_SECRET_simple: abc, unused: def}"])
+        assert len(results) == 1
```
dbeatty10, line 241: @nielspardon I looked at this along with @ChenyuLInx 

One question we have:
- Does it need to separate `run` and `test` here? Or would just `run` suffice? If it needs to be both `run` and `test`, then would `build` work instead?

tests/functional/context_methods/test_cli_vars.py
dbeatty10, line 1: In your PR description, you mentioned a few cases:
- Scrubs secret vars in RequiredVarNotFoundError
- Scrubs secret vars in StateCheckVarsHash event
- Scrubs secret vars in run results

This test covers the last one -- did you think about writing tests for the other two as well?

dbeatty10 (commented):
@nielspardon I looked at this along with @ChenyuLInx

The main question we have is regarding the test coverage. Could you take a look?

tests/functional/context_methods/test_cli_vars.py
nielspardon: originally, I tried triggering the `RequiredVarNotFoundError` with a test case in the `TestCLIVarsScrubbing` test class but was not able to surface any exception following the same test pattern as other tests in the class that are expecting `DbtRuntimeError` so I gave up on covering that one. If you have any ideas/pointers on how I could add a test I can take another stab at it.

tests/functional/context_methods/test_cli_vars.py
dbeatty10, line 241: Cool, let's just remove this one then:

```suggestion
```

tests/functional/context_methods/test_cli_vars.py
nielspardon: For `StateCheckVarsHash` I also did not know how to best add a test for this. I'm currently trying something with registering a callback for dbtRunner. Let's see how far I get.

nielspardon: I recalled that we found the issue with `StateCheckVarsHash` within a debug log line so I extended the existing test case to also check for the log line generated by the `StateCheckVarsHash` event. So this should be covered now.

tests/unit/test_parse_manifest.py
```diff
@@ -108,7 +108,7 @@ def _new_file(self, searched, name, match):
 class TestPartialParse(unittest.TestCase):
     def setUp(self) -> None:
         mock_project = MagicMock(RuntimeConfig)
-        mock_project.cli_vars = ""
+        mock_project.cli_vars = {}
```
nielspardon, line 111: @ChenyuLInx looks like you accidentally initialized this to a str but this should be a dict

tests/functional/context_methods/test_cli_vars.py
nielspardon: I rebased on current main and now the log line for `StateCheckVarsHash` does no longer show up on the CI system but still does on my local setup. Any ideas what might be causing this?

nielspardon: I added a test for `RequiredVarNotFoundError`.

nielspardon: I found the issue with the CI system: locally, Makefile sets the log format to JSON while on the CI system it's the default text format. I modified the tests to change the log format to JSON.

nielspardon: All tests pass. I now have tests for all three cases. Please take another look. Thanks.

ChenyuLInx (approved):
@nielspardon Thanks a bunch for adding this!!!

PR #9734, opened 2024-03-07
Avoid recursing when installing a project to a subproject via copytree

### LINKED ISSUE #9719: [Bug] dbt deps automatically recognizes projects in subdirectories

### Current Behavior

I'm attempting to build an integration_tests sub-project similar to dbt_utils, then install the parent project when running the integration tests. The file in `my_project/integration_tests/packages.yml` file is the same as in dbt_utils:
```
packages:
    - local: ../
```

This installs the parent project as a package, but that includes the integration_tests child project which dbt is recognizing as a project, and attempts to install its dependencies again, resulting in an endless call of dependencies. 

I end up with a directory structure of:
```
my_project
/integration_tests
/integration_tests/dbt_packages/my_project
/integration_tests/dbt_packages/my_project/integration_tests
/integration_tests/dbt_packages/my_project/integration_tests/dbt_packages/my_project
...
```

Eventually deps fails with an error:
`"[WinError 206] The filename or extension is too long: 'dbt_packages\\\\my_project\\\\integration_tests\\\\dbt_packages\\\\my_project\\\\integration_tests\\\\dbt_packages\\\\my_project\\\\integration_tests\\\\dbt_packages\\\\my_project\\\\integration_tests\\\\dbt_packages\\\\my_project\\\\integration_tests'"`

### Expected Behavior

When running dbt deps to a local project, only recognize the project.yml and packages.yml from the directly-referenced project and not sub-project directories. In this example, dbt should only look at `../packages.yml` and not be looking at `../integration_tests/packages.yml`.

### Steps To Reproduce

1. Create a project containing a sub-project
2. In the sub-project, add a package pointing to the parent project
3. Run `dbt deps`

### Relevant log output

_No response_

### Environment

```markdown
- OS: Windows 10
- Python: 3.11.7
- dbt: 1.7.9
```

_No response_

### Additional Context

_No response_

djbelknap:
resolves #9719

### Problem

When installing a local package that contains the current project as a subfolder (like the `integration_tests` project in dbt-utils), if the symlink fails the copytree recursively copies the current project repeatedly until the path length limit is reached.

### Solution

Use the `ignore` parameter of copytree to ignore the current project's packages directory to avoid repeated recursion.

### Checklist

- [ ] This PR includes tests, or tests are not required/relevant for this PR (Let me know if you'd like a test specific to this fix)


core/dbt/deps/local.py
```diff
@@ -67,7 +67,13 @@ def install(self, project, renderer):
...
         except OSError:
             fire_event(DepsSymlinkNotAvailable())
-            shutil.copytree(src_path, dest_path)
+            shutil.copytree(
+                src_path,
+                dest_path,
+                ignore=lambda directory, contents: project.packages_install_path
+                if directory == project.project_root
+                else [],
+            )
```
dbeatty10, line 76: ### What

"Don't install yourself into your own [`dbt_packages`](https://docs.getdbt.com/reference/project-configs/packages-install-path) folder."

### How

[`shutil.copytree`](https://docs.python.org/3/library/shutil.html#shutil.copytree) recursively copies an entire directory tree rooted at `src` to a directory named `dst`.

The new `ignore` parameter introduces a base case under which this recursion should stop. It essentially says:

"No dbt project ever needs to install itself within it's own `packages_install_path` folder 😄. So when the current directory being processed is the same as _this_ dbt project, then do **not** recursively copy the configured [`packages_install_path`](https://docs.getdbt.com/reference/project-configs/packages-install-path) directory because it's already covered. Just ignore it!"

### Why

The `ignore` parameter is being added here to prevent a dbt project from installing itself via `dbt deps`. As mentioned in https://github.com/dbt-labs/dbt-core/issues/9719, this can happen if:
1. a dbt project is a subfolder of a [`local` dependency listed within `packages.yml`](https://docs.getdbt.com/docs/build/packages#local-packages), and
2.  the host operating system doesn't support symlinks.

The 1st item is common for CI testing for packages like `dbt_utils`, and the 2nd is most common in legacy versions of Windows.

dbeatty10: Windows 10 users can possibly avoid this code path by turning on Developer Mode (which may allow creating Symbolic Links without requiring admin privileges) [[1]](https://stackoverflow.com/questions/58038683/allow-mklink-for-a-non-admin-user/58048888#58048888).

But this PR is specifically intended to cover the case where the user's system doesn't allow symlinks for whatever reason. So Developer Mode for Windows 10 is not a sufficient workaround for all cases.

Issue #4627, opened 2022-01-26
[CT-97] [Bug] dbt cli with an empty command takes more than a second to run


ajmarcus:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

When I run the `dbt` command in my terminal without additional arguments, it takes more than a second to print the help text:

```
% time dbt
...output...
dbt  1.67s user 0.13s system 98% cpu 1.828 total
```

Looks like this is a related issue with percieved run performance for specific commands:
https://github.com/dbt-labs/dbt-core/issues/4625

### Expected Behavior

Basic operations with the `dbt` cli would take less than one second to run.

### Steps To Reproduce

Install dbt:
```
pip install dbt-core
```

Run dbt:
```
% time dbt
...output...
dbt  1.67s user 0.13s system 98% cpu 1.828 total
```

### Relevant log output

This could deflinitely be an issue with my machine or setup. To see if I could find more information I used [`py-spy`](https://github.com/benfred/py-spy) to generate a flamegraph using these commands:

```shell
git clone 'git@github.com:dbt-labs/dbt-core.git'
cd dbt-core
pip install py-spy
sudo py-spy record -o profile_cli_empty.svg -- python core/scripts/dbt
```

There is nothing conclusive, but looks like there is a lot of time spent with calls to `importlib`:

https://gist.githubusercontent.com/ajmarcus/b077f4c8a94582df74c39258325addf5/raw/9af82bb55f5b322fb10bc0a959c20b175e689c3c/profile_cli_empty.svg

Which reminded me that Python eagerly loads imported dependencies. `importlib` does provide a lazy import loader but it comes with this warning:

https://python.readthedocs.io/en/latest/library/importlib.html#importlib.util.LazyLoader

```
For projects where startup time is critical, this class allows for potentially minimizing the cost of loading a module if it is never used. For projects where startup time is not essential then use of this class is heavily discouraged due to error messages created during loading being postponed and thus occurring out of context.
```

### Environment

```markdown
- OS: Mac 12.0.1
- Python: 3.7.8
- dbt: 1.0.1
```

### What database are you using dbt with?

bigquery

### Additional Context

![profile_cli_empty](https://user-images.githubusercontent.com/1408790/151216538-069af362-c1f6-4da3-944a-89b1b11543ea.svg)

ajmarcus:
[...]

PR #9744, opened 2024-03-09
Make `dbt` CLI cold start load up to 8% faster with lazy loading of task modules and `agate`

dwreeves:
The dbt CLI takes a while to load, especially before `.pyc` files are compiled. This PR makes things go much faster by making sure a handful of resources only get loaded when they are needed.

**In isolation, this PR speeds things up by about 5%. The additional 3% requires the following change to `dbt-adapters`: https://github.com/dbt-labs/dbt-adapters/pull/126**

## Lazy loading task modules

The first change was already discussed, but issue #4627 was closed without ever being fully resolved:

> Since the defaults are only used if a subcommand is being called, was thinking we could improve performance by only importing the relevant `dbt.task`. For example:
> 1. If the user uses the command `dbt build` we will only `import dbt.task.build as build_task`
> 2. If the user uses the command `dbt clean` we will only `import dbt.task.clean as clean_task`
> 3. If the user uses the command `dbt` without a subcommand we will not import any classes from `dbt.task`

Implementing this speeds up the CLI by about 5% on a cold start.

## Use `TYPE_CHECKING` to import `agate`

When I run `PYTHONPROFILEIMPORTTIME=1 dbt —help` to see where some additional free speed-ups may be, `agate` sticks out as taking a while to load; specifically it takes up about 3.5% of the total load time for the command `dbt —help`.

`agate` appears to only be used in three contexts across dbt (correct me if I am wrong): dbt unit tests, dbt docs generate, and dbt seeds.

Most `agate` imports are actually just for type annotations, meaning runtimes such as `dbt run —select foo` that do not require `agate` at all will end up loading it.

**Important note:** Gating `import agate` behind `if TYPE_CHECKING:` also requires making this change to the `dbt-adapters` library as well. Once both `dbt-core` and `dbt-adapters` use `if TYPE_CHECKING` to import agate in those contexts, then this change should shave another 3.5% off load times, and this speed improvement will impact not just `dbt —help` but _all_ dbt command invocations that don’t require `agate`.

## Before and after

I use the following script to test speeds before and after the changes. Note that this is only testing the lazy loading of modules, not `if TYPE_CHECKING: import agate` for reasons stated above.

In my testing, the before and after difference is about 5%.

```bash
#!/bin/bash
# OS X - requires `brew install coreutils` for gdate

for ((i=1; i<=60; i++))
do
  find "./venv" -type f -name "*.pyc" -delete
  find "./.venv" -type f -name "*.pyc" -delete
  find "./core" -type f -name "*.pyc" -delete

  start=$(gdate +%s%3N)
  dbt --help >/dev/null
  end=$(gdate +%s%3N)

  # Ignore first 5 runs
  if [ $i -gt 5 ]; then
    echo $((end - start))
  fi
done
```

If we compare the before and after outputs of the `PYTHONPROFILEIMPORTTIME=1` diagnostics, we can see what modules that are no longer avoided which account for the 5% speed-up:

```python
import csv
from pprint import pprint

before = []
after = []
with open(‘imports-before.txt’) as f:
    reader = csv.reader(f, delimiter='|')
    for row in reader:
        before.append(row[-1].strip())
with open(‘imports-after.txt') as f:
    reader = csv.reader(f, delimiter='|')
    for row in reader:
        after.append(row[-1].strip())

pprint([i for i in before if i not in after])
```

Output:

```python
['dbt.task',
 'dbt.parser.sql',
 'sqlparse.tokens',
 'sqlparse.utils',
 'sqlparse.sql',
 'sqlparse.exceptions',
 'sqlparse.cli',
 'sqlparse.engine.grouping',
 'sqlparse.keywords',
 'sqlparse.lexer',
 'sqlparse.engine.statement_splitter',
 'sqlparse.engine.filter_stack',
 'sqlparse.engine',
 'sqlparse.filters.others',
 'sqlparse.filters.output',
 'sqlparse.filters.tokens',
 'sqlparse.filters.reindent',
 'sqlparse.filters.right_margin',
 'sqlparse.filters.aligned_indent',
 'sqlparse.filters',
 'sqlparse.formatter',
 'sqlparse',
 'dbt.compilation',
 'dbt.task.printer',
 'dbt.task.base',
 'multiprocessing.dummy.connection',
 'multiprocessing.dummy',
 'dbt.task.runnable',
 'dbt.task.compile',
 'dbt.task.run',
 'dbt.task.snapshot',
 'dbt.task.seed',
 'daff',
 'dbt.task.test',
 'dbt.task.build',
 'dbt.task.clean',
 'dbt.task.clone',
 'dbt.links',
 'dbt.task.debug',
 'dbt.deps',
 'dbt.deps.base',
 'dbt.deps.local',
 'dbt.deps.tarball',
 'packaging',
 'packaging._structures',
 'packaging.version',
 'dbt.clients.git',
 'dbt.deps.git',
 'dbt.clients.registry',
 'dbt.deps.registry',
 'dbt.deps.resolver',
 'dbt.task.deps',
 'dbt.task.docs',
 'dbt.task.docs.generate',
 'socketserver',
 'shlex',
 'webbrowser',
 'http.server',
 'dbt.task.docs.serve',
 'dbt.contracts.results',
 'dbt.task.freshness',
 'dbt.task.init',
 'dbt.task.list',
 'dbt.task.run_operation',
 'dbt.task.retry',
 'dbt.task.show']
```


dwreeves:
It looks like this PR needs a `artifact_minor_upgrade` label because I add `if TYPE_CHECKING:` to a file in `core/dbt/artifacts`.

MichelleArk (approved):

peterallenwebb (approved):

peterallenwebb:
@dwreeves Thank you for making dbt better! We hope to have even more improvements to start-up time coming your way soon.

dwreeves:
Thanks @MichelleArk and @peterallenwebb!

The associated `dbt-adapters` PR is essential for realizing the speed boost across the dbt runtimes which don't use `agate`. Not sure who is in charge of that.

PR #9769, opened 2024-03-17
Fix #9593: Validation of unit test parsing for incremental models

### LINKED ISSUE #9593: [Bug] parse-time validation when unit testing incremental model

### Current Behavior

Suppose a model includes the `is_incremental()` macro.

If a unit test is added for that model, then an override like the following is required or else an error will occur:

```yaml
    overrides:
      macros:
        is_incremental: true
```

### Acceptance Criteria

We should have a parse time error when:
- unit testing an incremental model but not providing an override for `is_incremental`
- unit testing an incremental model overriding `is_incremental: true` but NOT providing an input for `this`

### Steps To Reproduce

`models/my_model.sql`

```sql
-- is_incremental: {{ is_incremental() }}
select
    1 as id
```

`models/_unit_tests.yml`

```yaml
unit_tests:

  - name: demo_unit_test_overrides

    given: []

    model: my_model

    expect:
      format: csv
      rows: |
        id
        1
```

```shell
dbt run -s models/my_model.sql
dbt build -s models/my_model.sql
```

💥 See error log below.

In order to work, the following needs to be added within the unit test YAML above:

```yaml
    overrides:
      macros:
        is_incremental: true
```

### Relevant log output

```shell
21:41:03    Compilation Error in unit_test my_model__demo_unit_test_overrides (models/_unit_tests.yml)
  'None' has no attribute 'database'
```

### Environment

```markdown
- OS:
- Python:
- dbt: 1.8 beta
```

_No response_

### Additional Context

_No response_

aranke:
resolves #9593

### Problem

Unclear error message when `is_incremental` override wasn't specified for a unit test 

### Solution

Add parse-time errors when:
* unit testing an incremental model but not providing an override for is_incremental
* unit testing an incremental model overriding is_incremental: true but NOT providing an input for this

### Checklist

### DIFF
--- a/core/dbt/parser/unit_tests.py
+++ b/core/dbt/parser/unit_tests.py
@@ -420,7 +420,6 @@ def find_tested_model_node(
 def process_models_for_unit_test(
     manifest: Manifest, current_project: str, unit_test_def: UnitTestDefinition, models_to_versions
 ):
-
     # If the unit tests doesn't have a depends_on.nodes[0] then we weren't able to resolve
     # the model, either because versions hadn't been processed yet, or it's not a valid model name
     if not unit_test_def.depends_on.nodes:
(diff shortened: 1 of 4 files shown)


core/dbt/parser/unit_tests.py
```diff
@@ -438,6 +437,31 @@ def process_models_for_unit_test(
...
     assert isinstance(target_model, ModelNode)
+
+    target_model_is_incremental = "macro.dbt.is_incremental" in target_model.depends_on.macros
+    unit_test_def_has_incremental_override = unit_test_def.overrides and isinstance(
+        unit_test_def.overrides.macros.get("is_incremental"), bool
+    )
+
+    if target_model_is_incremental and (not unit_test_def_has_incremental_override):
+        raise ParsingError(
+            f"Boolean override for 'is_incremental' must be provided for unit testing model '{target_model.name}'"
```
MichelleArk, line 448: suggestion -- let's also add the unit test name here for ease of user debugging since there can be multiple unit tests for a single model

core/dbt/parser/unit_tests.py
```diff
@@ -438,6 +437,31 @@ def process_models_for_unit_test(
...
+        )
+
+    unit_test_def_incremental_override_true = (
```
MichelleArk, line 451: do we actually need this in addition to `unit_test_def_has_incremental_override`? given we already have `unit_test_def_has_incremental_override` + are past the parsing error above, I think they are equivalent

aranke: Yes, since we want to test that the value of `unit_test_def_has_incremental_override` is explicitly `True` here.

Above, we're only testing that its value is not `None`.

We don't want this branch to be triggered if the value is `False`.

MichelleArk: ahh gotcha! thank you

PR #9772, opened 2024-03-18
Fix retry of a command with defer

### LINKED ISSUE #9770: [Regression] Retry of command with defer error: Could not find manifest in --state path
### Current Behavior

Executing a "retry" command on a command with "defer" sees error: Could not find manifest in --state path

### Expected Behavior

Retry command completes successfully

### Steps to reproduce

Do a retry of a command with a "--defer" component.

### Relevant log output

15:20:51  Encountered an error:
Runtime Error
  Could not find manifest in --state path: "<dbt.contracts.state.PreviousState object at 0x112b56170>"

### Additional context

This regression was caused by moving the "defer_state" parameter to global parameters. The existing code caused that flag to be instantiated as a "PosixPath('None')" object instead of Python None.

gshank:
resolves #9770

### Problem

Retry of a command that contained "defer" was failing with an error message: Could not find manifest in --state path:

### Solution

Include DEFER_STATE in the dictionary of args with defaults so that it won't change to PosixPath('None')

### Checklist

### DIFF
--- a/.changes/unreleased/Fixes-20240318-153338.yaml
+++ b/.changes/unreleased/Fixes-20240318-153338.yaml
@@ -1,0 +1,6 @@
+kind: Fixes
+body: Fix use of retry command on command using defer
+time: 2024-03-18T15:33:38.90058-04:00
+custom:
+  Author: gshank
+  Issue: "9770"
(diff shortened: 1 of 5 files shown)


MichelleArk (commented):
Fix looks great! Do you think its worth introducing a functional test for that retries a command that uses deferral, or is the unit test sufficient?

MichelleArk (approved):

PR #9794, opened 2024-03-22
Fix #9534: Add NodeRelation to SavedQuery Export

### LINKED ISSUE #9534: Saved queries should always include relation spec (`database.schema.alias`), and respect `generate_x_name` macros
### We should:
- always include `NodeRelation` objects for each export (including `database`, `schema_name`, `alias`)
- resolve the `database` / `schema_name` / `alias` with respect to `generate_x_name` macros (possibly by calling [this method](https://github.com/dbt-labs/dbt-core/blob/15dcb9a19d7549dbb5c4dbf250f6d915578cbfe9/core/dbt/parser/base.py#L285))

### Testing Considerations:

Will need a migration from previously saved manifests. 
Needs to be non-breaking change to artifacts.

### Considerations:
May not be able to backport these.

### Discussion Summarization:
* Might need to refactor `generate_x_name` functions to be usable outside of initial parsing
* Might not be able to _migrate_ existing manifests, i.e. they won't get regenerated until reparsed

aranke:
resolves #9534 

### Problem

For export configs:
1. `schema_name` should default to project schema if not set in config
2. `database` should be set to project database

### Solution

Run `export` nodes through the [update_parsed_node_relation_names](https://github.com/dbt-labs/dbt-core/blob/1e4e15c023926cfb1571fe0e192b9877c496bc57/core/dbt/parser/base.py#L286) function.

### Checklist


QMalcolm (approved):
Looks good to me! Let's get it merged in.

As a side note, wow is `update_parsed_node_relation_names` is a doozy of a function. Hopefully we can find some segment of work in the future to make how that function operates easier to follow.

PR #9802, opened 2024-03-22
ensure the upload name is unique

### LINKED ISSUE #9801: [Bug] time collisions on actions/upload-artifact

### Current Behavior

CI for funtional tests randomly fail on the upload_artfact step.  it generally succeeds on rerun.

### Expected Behavior

Don't fail due to a name already existing

### Steps To Reproduce

This happens randomly in CI. 

example run: https://github.com/dbt-labs/dbt-core/actions/runs/8377594563/job/22940086388

### Relevant log output

```shell
Error: Failed to CreateArtifact: Received non-retryable error: Failed request: (409) Conflict: an artifact with this name already exists on the workflow run
```

### Environment

```markdown
n/a
```

_No response_

### Additional Context

When we run tests in CI, we upload the results to an artifact.  Along with the update to Node20 in `actions/upload-artifact@v4` they also made a breaking change that the artifact names in a job must be unique.  We name the artifact with a timestamp so we sometimes get name collisions since we run in parallel and the timestamps may be equal.  Adding in the plit number should resolve this error.

More details in https://github.com/actions/upload-artifact/issues/478

emmyoop:
resolves #9801 

### Problem

CI is flaky due to actions/upload-artifact breaking change

### Solution

Ensure uniqueness of artifact name

### Checklist

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -216,7 +216,7 @@ jobs:
       - uses: actions/upload-artifact@v4
         if: always()
         with:
-          name: logs_${{ matrix.python-version }}_${{ matrix.os }}_${{ steps.date.outputs.date }}
+          name: logs_${{ matrix.python-version }}_${{ matrix.os }}_${{ matrix.split-group }}_${{ steps.date.outputs.date }}
           path: ./logs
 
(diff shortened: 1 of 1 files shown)


MichelleArk (approved):
thank you!

PR #9810, opened 2024-03-23
Only create the `packages-install-path` / `dbt_packages` folder during `dbt deps`

### LINKED ISSUE #6985: [CT-2127] [Bug] Running dbt commands before deps results in dbt_packages folder creation at invocation location

### Current Behavior

On a `dbt run` issued through the `dbtRunner.invoke`, dbt is attempting to create an empty default `dbt_packages`  directory at the location from which the command is being run, instead of at the project root.

### Expected Behavior

Any creation of a `dbt_packages` dir should default to the project root

### Steps To Reproduce

This behavior was observed through use of the `dbtRunner` class in 1.5.0+, though it likely existed prior.
1. In a python shell anywhere with access to dbt other than your project, do a run using the `dbtRunner`:
```
dbt = dbtRunner()
dbt.invoke(["run", "--project-dir", "../jaffle_shop"])
```
2. Observe a `dbt_packages` dir at the root of wherever this was invoked (if you have packages you needed to install, the error will be thrown but this dir will still be created)
3. Delete this dir and repeat, but run a `deps` first and then a `run`
4. Observer `dbt_packages` creation at the project root, no duplicate dir at the location of invocation

### Relevant log output

_No response_

### Environment

```markdown
- OS: Ventura 13.1
- Python: 3.8.13
- dbt: 1.4.0b1 branch `cl/server-on-merged`, soon to be 1.5.0a1
```

postgres, snowflake

### Additional Context

Everywhere that `packages_installation_path` is referenced in core, the value of that path is just `dbt_packages`. I checked that value on parse, run, deps — that value remains the same.

The absolute value of that path is what changes. Initially, and with any actions taken before doing a deps, the absolute value points to where I’m running it-- ~/dbt-core/dbt_packages. After running a deps and in all subsequent runs, the abspath points to the project-- ~/jaffle_shop/dbt_packages

[This is the function call](https://github.com/dbt-labs/dbt-core/blob/cl/server-on-merged/core/dbt/task/deps.py#L121) responsible for that change. That function calls an `os.chdir`, and we never change back (understandably). This function is not called for a run, so if deps is not called,[ this line](https://github.com/dbt-labs/dbt-core/blob/cl/server-on-merged/core/dbt/compilation.py#L165) on the `Compiler` object is creating an empty dir at ~/dbt-server/dbt_packages .

### LINKED ISSUE #9584: [Feature] Only create the `packages-install-path` / `dbt_packages` folder if there are packages to install

### Describe the feature

Even if you don't have a `packages.yml` file, dbt will create a `dbt_packages` folder whenever you run a dbt command.

Rather, it should only create this directory when running `dbt deps` (and only if there is at least one package to install).

It doesn't do much harm to have an empty `dbt_packages`, but it does feel like unnecessary cruft in the current working directory.

### Who will this benefit?

_No response_

_No response_

### Anything else?

See also: https://github.com/dbt-labs/dbt-core/issues/6985

Here's an example of the extra folder in the dbt Cloud IDE:

image

dbeatty10:
resolves #6985
resolves #9584

### Problem

dbt commands other than `dbt deps` are making sure a directory exists at this configured `packages-install-path` (which defaults to "dbt_packages").

This is leading to the surprising behavior reported in #6985 and #9584.

### Solution

Only create the [`packages-install-path`](https://docs.getdbt.com/reference/project-configs/packages-install-path) / `dbt_packages` folder when running `dbt deps`.

### Checklist

- [x] This PR has no interface changes (e.g. macros, cli, logs, json artifacts, config files, adapter interface, etc)

### DIFF
--- a/.changes/unreleased/Fixes-20240323-124558.yaml
+++ b/.changes/unreleased/Fixes-20240323-124558.yaml
@@ -1,0 +1,6 @@
+kind: Fixes
+body: Only create the packages-install-path / dbt_packages folder during dbt deps
+time: 2024-03-23T12:45:58.159017-06:00
+custom:
+  Author: dbeatty10
+  Issue: 6985 9584
--- a/core/dbt/compilation.py
+++ b/core/dbt/compilation.py
@@ -274,7 +274,6 @@ class Compiler:
 
     def initialize(self):
         make_directory(self.config.project_target_path)
-        make_directory(self.config.packages_install_path)
 
     # creates a ModelContext which is converted to
     # a dict for jinja rendering of SQL
--- a/tests/functional/list/test_list.py
+++ b/tests/functional/list/test_list.py
@@ -37,6 +37,13 @@ class TestList:
             },
         }
 
+    def test_packages_install_path_does_not_exist(self, project):
+        run_dbt(["list"])
+        packages_install_path = "dbt_packages"
+
+        # the packages-install-path should not be created by `dbt list`
+        assert not os.path.exists(packages_install_path)
+
     def run_dbt_ls(self, args=None, expect_pass=True):
         log_manager.stdout_console()
         full_args = ["ls"]


QMalcolm (approved):
Thank you for putting this together 🙂 It looks great! 

Of note, part of #9584 is the following line

> Rather, it should only create this directory when running dbt deps (and only if there is at least one package to install).

The in parentheses portion of that is to only create the directory when there are actually dependencies to install. I am absolutely okay with that not being part of this PR. However, we still want to implement that we can either create a new issue for just that part or I'm happy to pair and get that in as part of this PR.

dbeatty10:
> The in parentheses portion of that is to only create the directory when there are actually dependencies to install. I am absolutely okay with that not being part of this PR. However, we still want to implement that we can either create a new issue for just that part or I'm happy to pair and get that in as part of this PR.

I think the part in parentheses is already taken care of here in `DepsTask` whenever `dbt deps` is executed:
https://github.com/dbt-labs/dbt-core/blob/96f54264b4a4a185706051a58c801c9ea3b27385/core/dbt/task/deps.py#L234

Just let me know if it looks like something key is missing!

PR #9823, opened 2024-03-26
Fix info message for TestsConfigDeprecation

### LINKED ISSUE #9579: Wording of TestConfigDeprecation is incorrect
### Description

The wording of this deprecation refers to 'dbt_project.yml', but this deprecation is also issued for various "tests" configs in schema.yml files, which is really confusing.

### Acceptance Criteria

We issue non-confusing messages.  There will probably be multiple places with the "tests" config, and we only want to issue one deprecation message, so specifying the file might or might not be a good idea.

aranke:
resolves #9579 

### Problem

`TestsConfigDeprecation` refers to `dbt_project.yml`, which isn't related to fixing the problem.

### Solution

Link to docs on `data_tests` changing instead

### Checklist

### DIFF
--- a/core/dbt/events/types.py
+++ b/core/dbt/events/types.py
@@ -395,7 +395,7 @@ class TestsConfigDeprecation(WarnLevel):
     def message(self) -> str:
         description = (
             f"The `{self.deprecated_path}` config has been renamed to `{self.exp_path}`. "
-            "Please update your `dbt_project.yml` configuration to reflect this change."
+            "Please see https://docs.getdbt.com/docs/build/data-tests#new-data_tests-syntax for more information."
         )
         return line_wrap_message(warning_tag(f"Deprecated functionality\n\n{description}"))
 


core/dbt/events/types.py
```diff
@@ -395,7 +395,7 @@ def code(self) -> str:
     def message(self) -> str:
         description = (
             f"The `{self.deprecated_path}` config has been renamed to `{self.exp_path}`. "
-            "Please update your `dbt_project.yml` configuration to reflect this change."
+            "Please see https://docs.getdbt.com/docs/build/data-tests#new-data_tests-syntax for more information."
```
MichelleArk, line 398: Would this link: https://docs.getdbt.com/docs/build/data-tests#new-data_tests-syntax be more informative?

cc @graciegoheen

graciegoheen: Yes, but then we should update that page to also have an example of `dbt_project.yml` cc: @matthewshaver

graciegoheen: The places folks will need to update this is in their properties yml files:
```
models:
  - name: orders
    columns:
      - name: order_id
        data_tests:
          - unique
          - not_null
```
and `dbt_project.yml`:
```
data_tests:
  +store_failures: true
```

aranke: I confirmed that this error is raised only once, regardless of how many `tests` blocks there are (across both `properties.yml` and `dbt_project.yml` files), so I think it's fine to link to the docs page instead of including a file path (I'm not sure that we can get this, either)?

graciegoheen: https://github.com/dbt-labs/docs.getdbt.com/pull/5168

MichelleArk (approved):

PR #9828, opened 2024-03-27
Fix Docker release process for `dbt-postgres`

### LINKED ISSUE #9827: [Bug] Docker release for `dbt-postgres` is broken

### Current Behavior

Running the Docker release GHA for `dbt-postgres>=1.6` fails.

### Expected Behavior

This action should run successfully for `dbt-postgres<1.8`.

### Steps To Reproduce

- Run the action for `dbt-postgres==1.7.7`
- Observe the failure: https://github.com/dbt-labs/dbt-core/actions/runs/8456355316/job/23165975988

### Relevant log output

_No response_

### Environment

```markdown
- OS:
- Python:
- dbt:
```

postgres

### Additional Context

_No response_

mikealfare:
resolves #9827

### Problem

We updated the Docker release process for the new location of `dbt-postgres`. This broke the process for previous versions of `dbt-postgres`, which still live in `dbt-core`'s repo.

### Solution

- add the subdirectory back to the repo reference, which only works for `dbt-postgres<1.8`
- handle the docker releases for `dbt-postgres>=1.8` in the new `dbt-postgres` repo

### Checklist

### DIFF
--- a/.github/workflows/release-docker.yml
+++ b/.github/workflows/release-docker.yml
@@ -69,17 +69,9 @@ jobs:
         id: build_arg
         run: |
           repo_arg=$(echo ${{ inputs.package }} | sed 's/\-/_/g')
-          if [[ ${{ inputs.version_number }} < 1.8.0 ]];
-          then
-            repo_value=$(echo ${{ inputs.package }} | sed 's/postgres/core/g')
-            dbt_postgres_subdirectory="&subdirectory=plugins/postgres"
-          else
-            repo_value=${{ inputs.package }}
-            dbt_postgres_subdirectory=""
-          fi
+          repo_value=$(echo ${{ inputs.package }} | sed 's/postgres/core/g')
           echo "repo_arg=$repo_arg" >> $GITHUB_OUTPUT
-          echo "repo_value=$repo_value" >> $GITHUB_OUTPUT
-          echo "dbt_postgres_subdirectory=$dbt_postgres_subdirectory" >> $GITHUB_OUTPUT
+          echo "repo-value=$repo_value" >> $GITHUB_OUTPUT
 
       - name: Log in to GHCR
         uses: docker/login-action@v3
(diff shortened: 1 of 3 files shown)


.github/workflows/release-docker.yml
```diff
@@ -68,10 +68,18 @@ jobs:
       - name: Get docker build arg
         id: build_arg
         run: |
-          BUILD_ARG_NAME=$(echo ${{ inputs.package }} | sed 's/\-/_/g')
-          BUILD_ARG_VALUE=$(echo ${{ inputs.package }} | sed 's/postgres/core/g')
-          echo "name=$BUILD_ARG_NAME" >> $GITHUB_OUTPUT
-          echo "value=$BUILD_ARG_VALUE" >> $GITHUB_OUTPUT
+          repo_arg=$(echo ${{ inputs.package }} | sed 's/\-/_/g')
+          if [[ ${{ inputs.version_number }} < 1.8.0 ]];
```
mikealfare, line 72: I'm not sure if this will work when we get to 1.10.0.

VersusFacit: It won't. You need something:
```vercomp () {
    if [[ $1 == $2 ]]
    then
        return 0
    fi
    local IFS=.
    local i ver1=($1) ver2=($2)
    # fill empty fields in ver1 with zeros
    for ((i=${#ver1[@]}; i<${#ver2[@]}; i++))
    do
        ver1[i]=0
    done
    for ((i=0; i<${#ver1[@]}; i++))
    do
        if [[ -z ${ver2[i]} ]]
        then
            # fill empty fields in ver2 with zeros
            ver2[i]=0
        fi
        if ((10#${ver1[i]} > 10#${ver2[i]}))
        then
            return 1
        fi
        if ((10#${ver1[i]} < 10#${ver2[i]}))
        then
            return 2
        fi
    done
    return 0
}
```

VersusFacit: you could also mid build curl and run this: https://gist.github.com/Ariel-Rodriguez/9e3c2163f4644d7a389759b224bfe7f3#file-test-sh-L11

mikealfare: Yeah, I figured that was too easy.

mikealfare: I think it's going to be easier to just revert the change, keep this process for 1.7 and prior, and then handle the new releases in the `dbt-postgres` repo.

VersusFacit (approved):

PR #9833, opened 2024-03-28
automate docker release

emmyoop:
resolves https://github.com/dbt-labs/dbt-release/issues/109

### Problem

We miss releasing docker images often

### Solution

Delete Docker release workflow and move it to dbt-release since it's not just used by the core team https://github.com/dbt-labs/dbt-release/pull/108

Add docker generation to end of release

### Checklist


.github/workflows/release.yml
mikealfare, line 1: We should rename the workflow run to include Docker, since it does now.

.github/workflows/release.yml
```diff
@@ -174,6 +174,16 @@ jobs:
       PYPI_API_TOKEN: ${{ secrets.PYPI_API_TOKEN }}
       TEST_PYPI_API_TOKEN: ${{ secrets.TEST_PYPI_API_TOKEN }}
 
+  docker-release:
+    name: Docker Release
+    needs: [pypi-release]
+
+    uses: dbt-labs/dbt-release/.github/workflows/docker-release.yml@main
+    with:
+      package: ${{ github.event.repository.name }}
```
mikealfare, line 183: This is not true for `dbt-postgres`. But the release workflow releases both at the same time. So I think this should be two jobs and the package name is hardcoded. That being said, that will fail for versions 1.8+.

emmyoop: Good point - I rewrote this with a conditional based on minot version

.github/workflows/release-docker.yml
emmyoop, line 1: This file was moved to `dbt-release` in https://github.com/dbt-labs/dbt-release/pull/108 since it's shared by all adapters + core

PR #9837, opened 2024-03-29
[Tidy first] rename node, realnode to unique_id, node in selector_methods.py

gshank:
Tidy first change:  rename `node, realnode` to `unique_id, node` in selector_methods.py


emmyoop (approved):
Such a satisfying cleanup.

PR #9838, opened 2024-03-31
Removing unwanted dev dependency

### LINKED ISSUE #9830: [Bug] v1.5.10 & v1.6.10 versions are using protobuf > 5 in prod

### Current Behavior

[community Slack #adapter-ecosystem thread](https://getdbt.slack.com/archives/C030A0UF5LM/p1711466741203999)

As a part of #9630 PR, we restricted the protobuf to 4.x versions but the manual backport on v1.5 (#9676) & v1.6 (#9675) branches were incorrect because the restriction is only applied to dev-requirements rather than at the setup.py. As we can see in [v1.5.10](https://github.com/dbt-labs/dbt-core/blob/v1.5.10/core/setup.py#L67) and [v1.6.10](https://github.com/dbt-labs/dbt-core/blob/v1.6.10/core/setup.py#L82) This resulted in breaking the downstream adapters compatibilities as they started including protobuf > 5.x which is not compatible with dbt-core. It doesn't surface in dbt-core tests because dev-requirements always ensure that protobuf 4.x is installed. 
Eg for errors happen in adapters 
```
 def msg_to_dict(msg: EventMsg) -> dict:
        msg_dict = dict()
        try:
>           msg_dict = MessageToDict(
                msg, preserving_proto_field_name=True, including_default_value_fields=True  # type: ignore
            )
E           TypeError: MessageToDict() got an unexpected keyword argument 'including_default_value_fields'

/dbt/events/functions.py:227: TypeError
```

Another concern, we should not have duplicate dependencies listed in setup.py and dev-requirements.txt because these manual mistakes can happen again in the future.

### Expected Behavior

All the downstream adapters in v1.5.10 & v1.6.10 should be using the protobuf 4.x versions to avoid the above errors. 

### Steps To Reproduce

1. Use any adapter with latest v1.5 or v1.6 branches eg dbt-core v1.5.10
2. Install the dependencies
3. Check in the pip list if protobuf > 5. If it is then it is not compatible with dbt-core. 

### Relevant log output

_No response_

### Environment

```markdown
- OS: Ubuntu 20.04
- Python:3.8.16
- dbt:1.5
```

### Additional Context

_No response_

niteshy:
resolves #9830

Avoid causing problem similar to #9830 

### Problem

protobuf is listed as dependencies in both [setup.py](https://github.com/dbt-labs/dbt-core/blob/main/core/setup.py#L64) & [dev-requirements.txt](https://github.com/dbt-labs/dbt-core/blob/main/dev-requirements.txt#L17). It cause problem as shown in #9830 

### Solution
Remove the duplicate entry from dev-requirements, as it is used in the prod code already.

### Checklist

### DIFF
--- a/.changes/unreleased/Dependencies-20240331-103917.yaml
+++ b/.changes/unreleased/Dependencies-20240331-103917.yaml
@@ -1,6 +1,6 @@
 kind: Dependencies
-body: Remove unwanted dependency of protobuf in dev-requirements
+body: Remove duplicate dependency of protobuf in dev-requirements
 time: 2024-03-31T10:39:17.432017-07:00
 custom:
   Author: niteshy
-  PR: "9838"
+  Issue: "9830"
--- a/.changes/unreleased/Dependencies-20240331-103917.yaml
+++ b/.changes/unreleased/Dependencies-20240331-103917.yaml
@@ -1,0 +1,6 @@
+kind: Dependencies
+body: Remove unwanted dependency of protobuf in dev-requirements
+time: 2024-03-31T10:39:17.432017-07:00
+custom:
+  Author: niteshy
+  PR: "9838"
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -14,7 +14,6 @@ ipdb
 mypy==1.4.1
 pip-tools
 pre-commit
-protobuf>=4.0.0,<5.0.0
 pytest>=7.4,<8.0
 pytest-cov
 pytest-csv>=3.0,<4.0


.changes/unreleased/Dependencies-20240331-103917.yaml
```diff
@@ -0,0 +1,6 @@
+kind: Dependencies
+body: Remove unwanted dependency of protobuf in dev-requirements
```
QMalcolm, line 2: Small nit: 
Can we replace the word `unwanted` with `duplicate` here. The word `unwanted` without the context that this dependency is getting pulled in somewhere else makes it sound like we're dropping the dependency completely.

QMalcolm (approved):
Thank you for this contribution 🙂 This will 100% reduce accidental dev/released version discrepancies that we want to avoid. One small nit in the change log to make things a bit clearer, but then we're good to go 👍

.changes/unreleased/Dependencies-20240331-103917.yaml
```diff
@@ -0,0 +1,6 @@
...
+time: 2024-03-31T10:39:17.432017-07:00
+custom:
+  Author: niteshy
+  PR: "9838"
```
dbeatty10, line 6:
```suggestion
  Issue: "9830"
```

.changes/unreleased/Dependencies-20240331-103917.yaml
dbeatty10, line 2:
```suggestion
body: Remove duplicate dependency of protobuf in dev-requirements
```

QMalcolm:
Gonna get this merged! 

Of note, we don't need to backport this. We try to only do backports for bugs and security issues. The risk we incur by not backporting this in exceedingly low. The risk is that in the future we backport a change to the version specification of protobuf (which we've only ever had to do once) and only only update the dep in `dev-requirements.txt` (updating it in just `setup.py` or both `setup.py` and `dev-requirements.txt` isn't problematic). Given that we backport by cherry-picking, and now the main branch will no longer have the dependency in `dev-requirements.txt`, any backported version specification change to protobuf will include a change to `setup.py`.

PR #9839, opened 2024-04-01
 [#9791] Fix datetime.datetime.utcnow() is deprecated as of Python 3.12

### LINKED ISSUE #9791: [Bug] `datetime.datetime.utcnow()` is deprecated as of Python 3.12

### Current Behavior

Thanks to [Talla in #adapter-ecosytem Slack channel](https://getdbt.slack.com/archives/C030A0UF5LM/p1711010240341219) for the heads up

Any usage of `datetime.datetime.utcnow()`  throws the following deprecation warning in Python 3.12

> DeprecationWarning: datetime.datetime.utcnow() is deprecated and scheduled for removal in a future version. Use timezone-aware objects to represent datetimes in UTC: datetime.datetime.now(datetime.UTC).

related: https://github.com/dbt-labs/dbt-common/issues/99

### Expected Behavior

This shouldn't happen

### Steps To Reproduce

Run our unit test suite with Python 3.12

### Relevant log output

_No response_

### Environment

```markdown
- OS:
- Python: 3.12
- dbt: `1.7.10`
```

_No response_

### Additional Context

_No response_

slothkong:
Resolves #9791 

### Problem

Any usage of `datetime.datetime.utcnow()` throws a deprecation warning in Python 3.12 when unittests are executed.

### Solution

To get some results firsts, I just want to help with:

1. the execution of your recommendations @dbeatty10, e.g. `datetime.utcnow()` -> `datetime.now(timezone.utc).replace(tzinfo=None) ` 
2. bringing a bit of homogeneity to the ways datetime and its classes are being invoked, such that future development would require less effort

### Code review checklist

- [ ] Confirm that there are no breaking changes to dbt/artifacts
    - Otherwise, remove the `artifact_minor_upgrade` label. See [#9839 (comment)](https://github.com/dbt-labs/dbt-core/pull/9839#issuecomment-2029859392) for context.

### Checklist

### Additional Context

* I successfully ran unit and integration tests on `Python 3.12.0`, `Ubuntu 22.04`
* For good measure, I also built some models with both `dbt-postgres` and `dbt-bigguery` (1.8.0b1), then browsed over the resulting `manifest.json` files and verified that time references were accurate.
* Currently, the only deprecation warnings raised during unittesting are related to `dbt-common`  and `logbook` dependencies:

```
tests/unit/task/test_base.py: 2 warnings
  /path/to/my/dbt-core/.tox/py/lib/python3.12/site-packages/dbt_common/events/helpers.py:12: DeprecationWarning: datetime.datetime.utcnow() is deprecated and scheduled for removal in a future version. Use timezone-aware objects to represent datetimes in UTC: datetime.datetime.now(datetime.UTC).
    ts = datetime.utcnow()
```

```
tests/unit/task/test_base.py: 2 warnings
  /path/to/my/dbt-core/.tox/py/lib/python3.12/site-packages/dbt_common/events/logger.py:141: DeprecationWarning: datetime.datetime.utcnow() is deprecated and scheduled for removal in a future version. Use timezone-aware objects to represent datetimes in UTC: datetime.datetime.now(datetime.UTC).
    ts: str = datetime.utcnow().strftime("%H:%M:%S")
```
```
tests/unit/test_events.py::TestAdapterLogger::test_set_adapter_dependency_log_level
  /path/to/my/dbt-core/.tox/py/lib/python3.12/site-packages/logbook/compat.py:143: DeprecationWarning: datetime.datetime.utcfromtimestamp() is deprecated and scheduled for removal in a future version. Use timezone-aware objects to represent datetimes in UTC: datetime.datetime.fromtimestamp(timestamp, datetime.UTC).
    return datetime.utcfromtimestamp(timestamp)
```


slothkong:
Judging by what is the [README](https://github.com/dbt-labs/dbt-core/blob/main/core/dbt/artifacts/README.md#making-changes-to-dbtartifacts),  about changing code of `dbt` artifacts, adding the `artifact_minor_upgrade` tag to this PR should be acceptable, no?


Changes made to `core/dbt/artifacts/schemas/base.py`, `core/dbt/artifacts/schemas/results.py` and `core/dbt/artifacts/schemas/run/v5/run.py` do not fall in any of those categories. 
The returned type of `datetime.utcnow()` is equivalent to that of `datetime.now()` (e.g. `<class 'datetime.datetime'>`), and both methods work across python versions, including `3.8`.

slothkong:
I will try to submit an additional commit this week, to raise the patch coverage, specifically for `core/dbt/contracts/sql.py`~

dbeatty10:
> Judging by what is noted in the [README](https://github.com/dbt-labs/dbt-core/blob/main/core/dbt/artifacts/README.md#making-changes-to-dbtartifacts), about changing code of `dbt` artifacts, it should be acceptable to add the `artifact_minor_upgrade` tag to this PR, no?

@slothkong Since I'm not aware of any breaking changes, I've added the `artifact_minor_upgrade` label.

But to cover our bases, I've added this checklist item for the code review:

image

edgarrmondragon:
> I will try to submit an additional commit this week, to raise the patch coverage, specifically for `core/dbt/contracts/sql.py`~

Is that a blocker for merging this PR?

slothkong:
@edgarrmondragon , to your comment:
> > I will try to submit an additional commit this week, to raise the patch coverage, specifically for `core/dbt/contracts/sql.py`~
> 
> Is that a blocker for merging this PR?

In theory it never blocked the PR,  I was just trying to prevent any code coverage reduction, to give the core team more trust on my code changes 👼

jtcohen6:
@slothkong Sorry for the delay getting back to you. We'd like to get this merged in for dbt-core v1.10, and we have time set aside to take a look early next week.

If you're willing to try working through the merge conflicts, I'd welcome it. I also know this code has shifted a lot since last April, so it might be simpler to start with a new changeset and a fresh PR. If you're too busy, one of us will give it a go. Either way, let's make it happen!

slothkong:
Hello @jtcohen6,

Thanks for finding time to get back on this one!

As you pointed out, the code base was shuffled quite a bit since last April. Luckily,  it was not too difficult to fast-forward and include the upstream, in its state as of Saturday (4aa5169212d8256002095d44dc5f2505dca1b07c) . 


**What 's new?**
* Right after fast-forwarding, I patched new appearances of `utcnow()`. This was done for test scripts as well.

* In essence, I replaced the pattern:
 ```python
impor datetime
datetime.datetime.utcnow()
```
with the pattern:
```python
from datetime import datetime, timezone # and other classes whenever applicable
datetime.now(timezone.utc).replace(tzinfo=None)
```
* The noteworthy exception is `TestTargetConfigs`, which required the use of `modules` to test Jinja rendering:
https://github.com/dbt-labs/dbt-core/blob/7225496abb630a6d97ec57196c694a7bfceed349/tests/functional/configs/test_configs.py#L60

**What didn't change?**

* There are still places in the code base where `datetime` is imported as a module (`import datetime`), and not as a class (`from datetime import datetime`). Since the `utcnow()` method is not used in such files, for the sake of atomicity, they are kept untouched by the fix branch.
* Deprecation warnings coming from `dbt-common`  are still unaddressed (see https://github.com/dbt-labs/dbt-common/issues/99 ).

**Rational**
* As originally discussed in the issue, this approach was favored because it resolves the warnings and preserves of the current behavior -- specially important when touching the artifacts and contracts.
* This is done at the risk of 1) seeming pedantic, 2) slightly more verbosity in the import statements, and 3) inconsistent invocation of methods from the `datetime` module in some files.
*  Regarding risk 3, I am sure there are/will be cases where `import datetime` is  the convenient thing to do, for namespace clarity, or simply to align with how the module has been historically used. However,  incrementally moving towards more specific imports is in general good practice, and as long as unnecessary refactoring of working code is avoided.

**Tests**
* Besides the usual `make test` and `make integration`, I successfully built models with `dbt-bigquery` adapter and verified that compilation results were accurate.  
* Regarding CI tests results, I am still hesitant on how to deal with the alleged degradation flagged by Codecov :see_no_evil: 
* To be clear, effectively there were no new line additions, only replacement. It just so happens that some of the replacement are verbose and so `black` broke them into multiple lines, bloating up a bit the miss count. 
* The biggest offenders are still `contracts/sql.py` and `tasks/sql.py`, which so far lacked testing. Is it worth invest into coming up for a reasonable test suit from scratch in this case? Or how would be you deal with this if you were in my place? 

As usual, thank you for feedback and all the good work :pray: :cat:   

cc @zqureshi

zqureshi:
Thank you @slothkong!

We use a library called mashumaro for (de)serialization - that library is where `from_dict` and `to_dict` are implemented. It looks like mashumaro tries to [parse datetimes](https://github.com/Fatal1ty/mashumaro/blob/0e97dd77556fbed90303e242c260bb4fc9c0ecba/mashumaro/core/helpers.py#L29-L30) from strings, and if it cannot find a timezone, it presumes UTC (rather than naive).

@jtcohen6 and I ran your test on an existing dbt codebase and the test fails with the same output, that is to say this is expected behaviour and we should add this test and make sure the system continues to behave in the same way before and after your change.

It would be helpful to use [BaseArtifactMetadata](https://github.com/dbt-labs/dbt-core/blob/c6aeb4a291a30cc5b63dc0bf34c946c4c335ae30/core/dbt/artifacts/schemas/base.py#L55) instead of `RemoteCompileResult` in the test instead. I am more interested in the base artifact class as the use case we want to target is that dbt has serialized a manifest artifact, deserialized it, and the result is the same.

slothkong:
Hello @zqureshi, 

Thanks for taking a closer look and for the very useful observations :see_no_evil: 
I share below 2 remarks worth making (?), about the closing idea in your previous message. Would love to hear your thoughts before going any further:

> It would be helpful to use [BaseArtifactMetadata](https://github.com/dbt-labs/dbt-core/blob/c6aeb4a291a30cc5b63dc0bf34c946c4c335ae30/core/dbt/artifacts/schemas/base.py#L55) instead of RemoteCompileResult in the test instead. I am more interested in the base artifact class...

The reasoning behind that is of course understood. Nevertheless, I think some reconciliation with the following points is due:
1. **Favoring the test of concrete classes**: The [tests/unit/README.md](https://github.com/dbt-labs/dbt-core/blob/main/tests/unit/README.md) includes some guardrails about unittesing classes, with the TLDR being  "Test the concrete classes, not the ones they inherit from".
2. **Addressing coverage degradation**: Putting up additional unittests for [BaseArtifactMetadata](https://github.com/dbt-labs/dbt-core/blob/c6aeb4a291a30cc5b63dc0bf34c946c4c335ae30/core/dbt/artifacts/schemas/base.py#L55) wouldn't address the misses that are highlighted by the [Codecov report](https://app.codecov.io/gh/dbt-labs/dbt-core/pull/9839/tree?search=&displayType=list). As you might know, the file where the base class sits at around 90% and the miss count is not affected by the fix. On the other hand, `core/dbt/contracts/sql.py` and `core/dbt/tasks/sql.py` so far had 0% coverage and are flagged with the most number of misses.

If you agree both arguments have weight, then I would suggest to proceed with implementation for concrete classes only (i.e. `RemoteCompileResult`, `RemoteExecutionResult`,  `RemoteRunResult`, `SqlCompileRunner`, etc), starting with something like:

```python
# tests/unit/contracts/tests_sql.py
import pytest
from unittest.mock import patch

from tests.unit.fixtures import model_node
from hypothesis import given
from hypothesis.strategies import builds, lists, text

from dateutil.tz import tzutc
from dbt.artifacts.schemas.results import TimingInfo
from dbt.contracts.sql import RemoteCompileResult


@given(
    raw_code=text(),
    compiled_code=text(),
    timing=lists(builds(TimingInfo)),
)
def test_remote_compile_result_serialization(raw_code, compiled_code, timing):
    obj = RemoteCompileResult(
        raw_code=raw_code,
        compiled_code=compiled_code,
        node=model_node(),
        timing=timing,
    )

    obj_from_dict = RemoteCompileResult.from_dict(obj.to_dict())
    
    assert obj.generated_at.tzinfo is None

    assert obj_from_dict.generated_at.tzinfo == tzutc()
    assert obj_from_dict.generated_at.replace(tzinfo=None) == obj.generated_at

    assert obj_from_dict.raw_code == obj.raw_code
    assert obj_from_dict.compiled_code == obj.compiled_code
    assert obj_from_dict.node == obj.node
    assert obj_from_dict.timing == obj.timing
```
**NOTE**: This **2nd version of the new test** passes when ran  against the HEAD of main and also against this  fix branch. It makes explicit the fact that the concrete classes in question stick to naive datetimes, while the serialization/de-serialization process infers `tzinfo` in UTC (due to current behavior of `mashumaro` ).

cc: @jtcohen6

tests/functional/artifacts/test_run_execution_result.py
```diff
@@ -0,0 +1,41 @@
...
+    assert len(result_from_dict.results) == len(result.results)
+
+    assert result.generated_at.tzinfo is None
+    assert result_from_dict.generated_at.tzinfo == tzutc()
+    assert result_from_dict.generated_at.replace(tzinfo=None) == result.generated_at
+
+    for original in result.results:
+        for deserialized in result_from_dict.results:
+            if original.node.unique_id == deserialized.node.unique_id:
+                assert original.node.created_at == deserialized.node.created_at
```
zqureshi, line 41: nit: This equality assertion is a bit confusing because this is a float rather than a datetime type which is the intent of the test.

I am also not 100% sure if float equality checks would be fully accurate in this context.

zqureshi (approved):
Thank you so much @slothkong, this is excellent and comprehensive!

We're going to merge.

zqureshi:
Closing and reopening to trigger full CI again, one test seems to be stuck.

jtcohen6 (approved):

emmyoop (approved):

emmyoop:
We can force merge past the failing artifact check.  It's to enforce 2 reviews, which we have.  I will put up a fix for this check tomorrow.

tests/functional/artifacts/test_run_execution_result.py
slothkong: @zqureshi, sorry for the confusion. 
Indeed `created_at` was affected by code changes of the fix.  I opted for adding it in the new tests because it is the other time-related value that ends up in the `manifest` (an other artifacts) and because I could not find any explicit de-serialization check in the test suit.

In hindsight, this might comes across as  over cautious :/

Issue #6323, opened 2022-11-28
[CT-1556] Smarter handling of `--vars` in partial parsing


jtcohen6:
Just like https://github.com/dbt-labs/dbt-core/issues/3885, but for CLI `--vars`.

This would require us to capture, at parse time, which files depend on which `--vars`, via calls to the Jinja `{{ var() }}` function. That would also include macros that call `var()`, and are then called by models / other macros in turn.

For Python models, if we introduce a built-in `dbt.var()` function, we'd want to do the same. We're already doing something similar for configs, [to power `config.get()` at runtime](https://github.com/dbt-labs/dbt-core/blob/bce0e7c0964e67d5a463bba5ce498a3fe497e0ab/core/dbt/parser/models.py#L209-L213).

Whenever the `--vars` change, instead of triggering a full re-parse, we'd schedule just the files that depend on the `var` for re-parsing. Of course, if the `var` is used for a configuration within `dbt_project.yml`, that could still affect many many nodes.

ChenyuLInx:
If possible, can we separate this handling from parsing? What I am thinking is that parsing is going from everything in project file -> a representation, then in actual runtime, we apply configuration to the representation and do the execution.

jtcohen6:
@ChenyuLInx Supportive of this line of thinking! The biggest caveats here is that `vars` can be used to dynamically disable/enable models, or to conditionally affect relationships between models — so it is necessary to resolve some `vars` during parsing in order to know the shape of the DAG, and to support node selection.

During parsing, we could store pointers to those variables, and then conditionally reevaluate them just before each execution. That feels similar to the approach described in this issue (partial parsing), though with some subtle differences in implementation.

jtcohen6:
@gshank Do you know if the partial parsing manifest (`target/partial_parse.msgpack`) contains enough information (raw file contents & unrendered yaml configurations/attributions), such that we could support a re-parse when CLI `--vars` are supplied, without needing to go back to the actual file system?

I'm thinking:
- we can skip the need to re-read files (sooner)
- we can also add support for "partial" re-parsing of only those files/nodes which are actually affected by the `--vars` override (the original scope of this issue)

kp-tom-sc:
We use the cli --vars to pass in airflow datetime variables. These change on each run, so we can't partial parse.
Is there a better way of handling datetime variables?
[...]

PR #9844, opened 2024-04-02
[BUG]  Exclude password-like fields for considering reparse

### LINKED ISSUE #9795: [Bug] Exclude password-like fields (missing from `_connection_keys`) in partial parsing state check

### Current Behavior

A change in the value of a "hidden" (password-like) field will trigger a full re-parse — but it shouldn't, because that field is not accessible in the `{{ target }}` Jinja variable, and therefore cannot be used as an input to parse-time configurations. That's the whole motivation for including profile information in the partial parse state check in the first place.

--

Instead of hashing the full contents of the active profile, we should only hash the ["connection keys"](https://github.com/dbt-labs/dbt-adapters/blob/59cb168bbb0f7b01591016ab64f3fa1f2f3716cd/dbt/adapters/contracts/connection.py#L159) of the active profile. This is equivalent to what the user sees in the output of `dbt debug`, and what they can access in the `{{ target }}` Jinja variable.

So instead of:

https://github.com/dbt-labs/dbt-core/blob/c6c0c79216a3c8032d5ea5ed608ca2ef7bdacc9a/core/dbt/parser/manifest.py#L991-L994

A simple solution would be something like:
```python
connection_keys = config.credentials._connection_keys()
profile_hash = FileHash.from_contents(pprint.pformat(connection_keys))
```

**However**, I think this would _undo_ a workaround that's existing for a long time to avoid full re-parses for _non-credential_ values by setting them to a "secret" env var. This behavior isn't yet documented, but that's more our oversight than an intentional choice to leave it undocumented:
- https://github.com/dbt-labs/docs.getdbt.com/issues/1066

### Expected Behavior

**Acceptance criteria:**

1. A change to the `password` profile field (or similar) should not trigger a full re-parse
2. Users should retan the ability to "opt" out additional fields by setting them to "secret" env vars, the values of which dbt does not track 

### Steps To Reproduce

1. Create a profile that has a `password` field and another field set to `{{ env_var('DBT_SECRET_MYSECRET') }}`
1. Full parse (`dbt parse`)
2. Change password field
3. Rerun `dbt parse`, this is a full parse but it shouldn't be
4. Change value of `DBT_SECRET_MYSECRET`
5. This should not be a full reparse

### Relevant log output

_No response_

### Environment

```markdown
- OS: macOS 14.3
- Python: 3.10.11
- dbt: 1.7.10
```

_No response_

### Additional Context

[internal thread](https://dbt-labs.slack.com/archives/C05FWBP9X1U/p1711046660026059)

ChenyuLInx:
resolves #9795
related PR https://github.com/dbt-labs/dbt-snowflake/pull/950 has been merged and backported to dbt-snowflake 1.6.latest and 1.7.latest.

### Problem
Currently any change in profiles.yml would lead to full reparse.

### Solution
Generate the hash for determining rereparse with contents that are accessible in Jinja context instead of the whole profile.yml.

### Checklist


core/dbt/parser/manifest.py
```diff
@@ -988,10 +988,15 @@ def build_manifest_state_check(self):
...
-        with open(profile_path) as fp:
-            profile_hash = FileHash.from_contents(fp.read())
+        # Create a hash of the connection keys, which user has access to in
+        # jinja context. Thus keys here may affect the parsing result.
+
+        # Renaming this variable mean that we will have to do a whole lot more
+        # change to make sure the previous manifest can be loaded correctly.
+        # This is an example of naming should be chosen based on the functionality
+        # rather than the implementation details.
+        connection_keys = config.credentials._connection_keys()
```
MichelleArk, line 998: It looks like `credentials._connection_keys()` is just the set of names of credential keys, e.g: 
```python
('account', 'user', 'database', 'warehouse', 'role', 'schema', 'authenticator', 'oauth_client_id', 'query_tag', 'client_session_keep_alive', 'host', 'port', 'proxy_host', 'proxy_port', 'protocol', 'connect_retries', 'connect_timeout', 'retry_on_database_errors', 'retry_all', 'insecure_mode', 'reuse_connections')
```

If we hash just the key names without the values, then partial parsing won't retrigger even if one of the values change, as it should. We should use `list(config.credentials.credential_info())` instead. `with_aliases=False` looks appropriate here as well.

tests/unit/test_parse_manifest.py
```diff
@@ -122,6 +134,17 @@ def test_partial_parse_file_path(self, patched_open, patched_os_exist, patched_s
         # if specified in flags, we use the specified path
         patched_open.assert_called_with("specified_partial_parse_path", "rb")
 
+    def test_partial_parse_profile_change(self):
```
QMalcolm, line 137: The name of this test seems to imply that we're asserting that something with partial parsing did or didn't happen. However looking at the test it seems to only assert that the file hash has changed, which is not indicative itself of whether a full or partial parse happened.

core/dbt/parser/manifest.py
ChenyuLInx: Good catch! This also mean I probably should add an integration test to capture the behavior

QMalcolm (commented):
The change looks good! I think we can be doing more testing wise though 🤔 We don't actually seem to be testing what style of parsing is occurring based on change happened in `profiles.yml`. I think we need two more tests:

- change something in `profiles.yml` that isn't in `_connection_keys`, assert that only a partial parse occurred
- change something n `profiles.yml` that is in `_connection_keys`, assert that a full parse occurred

core/dbt/parser/manifest.py
QMalcolm: You beat me to my suggestion 😂

core/dbt/parser/manifest.py
```diff
@@ -988,10 +988,16 @@ def build_manifest_state_check(self):
...
-            profile_hash = FileHash.from_contents(fp.read())
+        # Create a hash of the connection_info, which user has access to in
+        # jinja context. Thus attributes here may affect the parsing result.
+        # Ideally we should not expose all of the connection info to the jinja.
+
+        # Renaming this variable mean that we will have to do a whole lot more
+        # change to make sure the previous manifest can be loaded correctly.
+        # This is an example of naming should be chosen based on the functionality
+        # rather than the implementation details.
+        connection_keys = list(config.credentials.connection_info())
```
gshank, line 999: We don't want to convert connection_info to a list, that removes the values from the connection_info dictionary. We need to preserve them, because it's changes in the values that force a re-parse.  In other places in the code we hash a dictionary for profiles_env_var_hash and and project_env_vars_hash, but any method that turns a dictionary into something hashable would work.

gshank (changes requested):
We need to include the values of the connection_info dictionary in the hash.

core/dbt/parser/manifest.py
gshank: Also, we don't really need profile_env_vars_hash anymore after doing this since the env_vars would be resolved. So we might want to remove that part, since it will cause extra churn in parts of the profile that may not be being used.

ChenyuLInx: So connection_info returns a iterator, after converting to a list it will look like [(key1, value1), (key2, value2)].
I will get the profile `env_var_hash` part adjusted

QMalcolm (approved):
Assuming @gshank's comments are addressed, this looks good to me 🙂 Thank for adding the extra test cases!

tests/functional/partial_parsing/test_pp_vars.py
```diff
@@ -314,40 +314,33 @@ def dbt_profile_target(self):
         # calls 'load_config' before the tests are run.
         # Note: only the specified profile is rendered, so there's no
         # point it setting env_vars in non-used profiles.
-        os.environ["ENV_VAR_USER"] = "root"
-        os.environ["ENV_VAR_PASS"] = "password"
+        os.environ["ENV_VAR_HOST"] = "localhost"
         return {
             "type": "postgres",
             "threads": 4,
-            "host": "localhost",
```
ChenyuLInx, line 322: @gshank I think the adjust to the test here makes sense since user and pass are not in `connection_info`

gshank (approved):
Looks good.

ChenyuLInx:
@MichelleArk @jtcohen6 I think we mentioned we will be using config to FF when a behavior change happens in order to maintain backward compatibility. Do you think this is considered a behavior change that should actually have a config for?
My guess is not, but it is a behavior change(small). We should track it so we have a deciding criteria for later.

jtcohen6:
@ChenyuLInx Thanks for thinking of it! I had a similar question. Here's my rationale:
- The behavior we're deprecating (you can avoid a full re-parse by setting a non-credential profile value to a "secret" env var) has never been documented (https://github.com/dbt-labs/docs.getdbt.com/issues/1066).
- It's not a behavior change that risks _correctness_, just means that there will be full parses instead of partial parses
- I'm aware of other workarounds... and we agree that the real answer is to be smarter with our detection of where `target` values, [`--vars`](https://github.com/dbt-labs/dbt-core/issues/6323), etc, are being used in the project.

So **no**, I don't think we need a behavior change flag here. This is our implementation detail, not a behavior that we've documented or guaranteed.

will-sargent-dbtlabs:
@ChenyuLInx For single tenants on 1.8, do you think the updated image is available nowish and will deploy during next weeks release?

ChenyuLInx:
@will-sargent-dbtlabs maybe not, I will reach out in slack about more detailed timeline

PR #9873, opened 2024-04-08
Support using sql in unit testing fixtures

### LINKED ISSUE #9405: [CT-3567] [Feature] unit testing support format: sql

### Describe the feature

We should allow support to set `format: sql` for unit test definition:

```
  - name: test_valid_email_address
    model: dim_wizards 
    given: 
      - input: ref('stg_wizards')
        format: sql
        rows: |
          select 1 as wizard_id, cool@example.com as email, example.com as email_top_level_domain union all 
          select 2 as wizard_id, cool@unknown.com as email, unknown.com as email_top_level_domain union all 
          select 3 as wizard_id, badgmail.com as email, gmail.com as email_top_level_domain union all 
          select 4 as wizard_id, missingdot@gmailcom as email, gmail.com as email_top_level_domain
      - input: ref('top_level_email_domains')
        format: sql
        rows: |
          select example.com as tld union all,
          select gmail.com as tld
    expect: 
      format: sql
      rows: |
        select 1 as wizard_id, true as is_valid_email_address union all
        select 2 as wizard_id, false as is_valid_email_address union all
        select 3 as wizard_id, false as is_valid_email_address union all
        select 4 as wizard_id, false as is_valid_email_address 
```

We should also allow folks to also use sql fixutres:

```
  - name: test_valid_email_address
    model: dim_wizards 
    given: 
      - input: ref('stg_wizards')
        format: sql
        fixture: wizard_emails_input
      - input: ref('top_level_email_domains')
        format: sql
        fixture: tld_input
    expect: 
      format: sql
      fixture: wizard_emails_output
```

### Anything else?

- One of the main motivations for this is this would **automatically allow support for complex data types** (if you want to use type variant/json/etc. just use `format: sql`) https://github.com/dbt-labs/dbt-core/issues/8423
- We would not allow for folks to use jinja in this SQL, the reasons being:
   - we don't think folks should `{{ ref() }}` a model from their **static** input data (we don't want to consider DAG order, we don't wan't to point to data that could change, etc.)
   - this would mean folks couldn't use cross-db macros in their SQL (we're ok with that sacrifice)
   - this would mean folks couldn't use macros in their SQL
  
### Scope

- Allow for SQL fixtures, but throw error with jinja present.

gshank:
resolves #9405

### Problem

In some cases there might be difficulties in getting csv or dictionary fixtures to work correctly for unit testing, so we want to allow the use of SQL fixtures.

### Solution

This allows setting SQL in the unit test definition or in fixture files.  At the current time it requires all columns to be specified.

### Checklist


ChenyuLInx (approved):

graciegoheen (approved):
Tested this out, and looks good to me!

PR #9915, opened 2024-04-12
search for dbt_core source distribution

MichelleArk:
resolves #N/A

### Problem

setuptools released a patch that enforcing a naming standard on source distributions to use '_' instead of '-'. this breaks our regex-based search in CI: https://github.com/dbt-labs/dbt-core/actions/runs/8665429967/job/23764102566?pr=9891

 https://setuptools.pypa.io/en/stable/history.html

https://github.com/pypa/setuptools/issues/4300

### Solution

update regex to search using `_` also updating to only search for dbt_core explicitly since we no longer build other distributions in this step. 

### Checklist

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -288,7 +288,7 @@ jobs:
       - name: Install source distributions
         # ignore dbt-1.0.0, which intentionally raises an error when installed from source
         run: |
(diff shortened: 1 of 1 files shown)


mikealfare (approved):
I assume this is only run on `main`?

PR #9954, opened 2024-04-16
Fix nightly release & Docker drafts

emmyoop:
resolves #

### Problem

Nightly releases failed for having the wrong permissions.

Additionally, Docker released fail on a draft release since the tag does not exist.

### Solution

Add the `package: write` permission to the nightly release
Do not release to Docker when we are running the release as a test
Allow the release to _only_ release to Docker when a flag is flipped.  This will allow us to release to Docker as needed if the final release already exists.

### Checklist

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -215,7 +215,7 @@ jobs:
     # We cannot release to docker on a test run because it uses the tag in GitHub as
     # what we need to release but draft releases don't actually tag the commit so it
     # finds nothing to release
-    if: ${{ !inputs.test_run || inputs.only_docker }}
+    if: ${{ !failure() && !cancelled() && (!inputs.test_run || inputs.only_docker) }}
     strategy:
       matrix: ${{fromJson(needs.determine-docker-package.outputs.matrix)}}
 
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -59,6 +59,11 @@ on:
         type: boolean
         default: false
         required: false
+      only_docker:
+        description: "Only release Docker image, skip GitHub & PyPI"
+        type: boolean
+        default: false
+        required: false
 
 permissions:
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -99,6 +105,7 @@ jobs:
   bump-version-generate-changelog:
     name: Bump package version, Generate changelog
     needs: [job-setup]
+    if: ${{ !inputs.only_docker}}
```
mikealfare, line 108: Minor nit, I think. There should be a space between the variable and the `}}`. I don't think this affects the parsing though.

.github/workflows/release.yml
```diff
@@ -114,7 +121,7 @@ jobs:
 
   log-outputs-bump-version-generate-changelog:
     name: "[Log output] Bump package version, Generate changelog"
-    if: ${{ !failure() && !cancelled() }}
+    if: ${{ !failure() && !cancelled() && !inputs.only_docker}}
```
mikealfare, line 124: Same space thing.

.github/workflows/release.yml
```diff
@@ -128,7 +135,7 @@ jobs:
 
   build-test-package:
     name: Build, Test, Package
-    if: ${{ !failure() && !cancelled() }}
+    if: ${{ !failure() && !cancelled() && !inputs.only_docker}}
```
mikealfare, line 138: Same space thing.

.github/workflows/release.yml
```diff
@@ -149,7 +156,7 @@ jobs:
 
   github-release:
     name: GitHub Release
-    if: ${{ !failure() && !cancelled() }}
+    if: ${{ !failure() && !cancelled() && !inputs.only_docker}}
```
mikealfare, line 159: Same space thing.

mikealfare (approved):
One minor nit (x4)

PR #9956, opened 2024-04-16
Add flags.deprecate_package_materialization_builtin_override

MichelleArk:
resolves https://github.com/dbt-labs/dbt-core/issues/10007

### Problem

Materializations defined in third-party packages are included in the search order for materialization macros. This is difficult to detect, difficult to prevent, and it works differently from other macros.

We should remove third-party packages from the search order.

This change needs to go behind a behavior-change flag:[Legacy behaviors | dbt Developer Hub](https://docs.getdbt.com/reference/global-configs/legacy-behaviors) . See https://github.com/dbt-labs/dbt-core/pull/9366 for an implementation of a behavior-change flag.

### Solution

Introduce flags.deprecate_package_materialization_builtin_override.

when the flag is False (legacy behaviour):
* fire a deprecation event and warning and allow the override
when the flag is True (soon-to-be default behaviour)
* get the next candidate, only searching for candidates with core and root localities

### Checklist


core/dbt/cli/flags.py
```diff
@@ -355,7 +355,7 @@ def set_common_global_flags(self):
     # This is here to prevent mypy from complaining about all of the
     # attributes which we added dynamically.
     def __getattr__(self, name: str) -> Any:
-        return super().__get_attribute__(name)  # type: ignore
+        return super().__getattribute__(name)  # type: ignore
```
MichelleArk, line 358: This appears to have been broken all along, but was never reached in testing

core/dbt/contracts/graph/manifest.py
```diff
@@ -946,11 +960,17 @@ def find_materialization_macro_by_name(
...
+            # preserve legacy behaviour - allow materialization override
+            if get_flags().deprecate_package_materialization_builtin_override is False:
+                deprecations.warn(
+                    "package-materialization-override",
+                    package_name=materialization_candidate.macro.package_name,
+                    materialization_name=materialization_name,
+                )
+            else:
+                materialization_candidate = candidates.last_candidate(
+                    valid_localities=[Locality.Core, Locality.Root]
```
MichelleArk, line 972: was not able to do this lower-level using the `_find_macros_by_name` `filter` parameter because that filter only has a view of the macro candidate being iterated over. For this check, we need to exclude imported materialization macros only if the materialization is builtin in order to continue supporting custom materializations from packages (tested)

MichelleArk:
Note: we'd like to backport these changes but it will very likely need to be manual since ProjectFlags does not exist in 1.6 or 1.7

.changes/unreleased/Features-20240422-173703.yaml
```diff
@@ -0,0 +1,6 @@
+kind: Features
+body: Add flags.deprecate_package_materialization_builtin_override
```
emmyoop, line 2: Can you reword this to be more human readable / not code?

emmyoop (approved):
Looks good - just a changelog comment.

MichelleArk:
codecov/project is hanging waiting for a response but I can see the results passing on: https://app.codecov.io/gh/dbt-labs/dbt-core/pull/9956
Screenshot 2024-04-23 at 3 54 03 PM

PR #9959, opened 2024-04-16
Materialized Views not updating cache

McKnight-42:
linked issue: #https://github.com/dbt-labs/adapters-team-private/issues/127
this pr would eventually need to be backported to 1.6 and have similar changes made in `dbt-adapters`

### Problem
We do not appear to be updating cache for MV functionality like drops, renames, adds which is causing catalog mismatch errors in runs especially between MV's that are dependent on each other

### Solution

Current:
we will remove `materialized_view` from the `renameable_relation` variable thus reverting us to the create and drop functionality that we were seeing before the introduction in https://github.com/dbt-labs/dbt-core/pull/9682

Attempted:
Update our various macros to either use commonly used Python functions from the base `impl.py` such as `drop_relation`, and `rename_relation` or use the more primitive version designed in the base adapter e.g `cache_dropped` and `cache_renamed` to ensure we are maintaining parity with database expectation.

  TODO:
- [x] Add changelog entry after a solution is found
- [ ] Create a version for dbt-postgres on [main](https://github.com/dbt-labs/dbt-postgres/pull/69)
- [ ] Double check on backport to 1.6 in core

### Checklist


plugins/postgres/dbt/include/postgres/macros/adapters.sql
```diff
@@ -39,7 +39,7 @@
   on {{ relation }} {% if index_config.type -%}
     using {{ index_config.type }}
   {%- endif %}
-  ({{ comma_separated_columns }});
```
McKnight-42, line 42: removing `;`as it doubles up with one at end of statement causing error

plugins/postgres/dbt/include/postgres/macros/relations/materialized_view/rename.sql
```diff
@@ -1,3 +1,11 @@
 {% macro postgres__get_rename_materialized_view_sql(relation, new_name) %}
+  {% if relation is not none %}
+    {% set database = relation.database %}
+    {% set schema = relation.schema %}
+    {% set to_relation = adapter.get_relation(database=database, schema=schema, identifier=new_name) %}
+    {% if to_relation is not none %}
+        {{ adapter.rename_relation(from_relation=relation, to_relation=to_relation) }}
+    {% endif %}
+  {% endif %}
```
McKnight-42, line 9: Currently this seems the most correct attempt we've made

When in a `debug` session of a `dbt run --full-refresh` run we see
```
ipdb> locals().keys()
dict_keys(['l_1_relation', 'l_1_new_name', 't_2', 'l_1_database', 'l_1_schema', 'l_1_adapter', 'l_1_to_relation', 'l_1_debug', 'concat', 'context', 'environment', 'missing', 'resolve', 't_1', 'undefined'])
ipdb> l_1_relation
<PostgresRelation "dbt"."dbt_mcknight"."orders">
ipdb> l_1_new_name
'orders__dbt_backup'
ipdb> l_1_database
'dbt'
ipdb> l_1_schema
'dbt_mcknight'
ipdb> l_1_to_relation
ipdb> l_1_to_relation
```

but on run end we are getting error 
```Internal Error
  Cache inconsistency detected: in rename, new key _ReferenceKey(database='dbt', schema='dbt_mcknight', identifier='orders') already in cach```

File: core/dbt/include/global_project/macros/relations/materialized_view/drop.sql [hunk 3]
```diff
@@ -11,4 +11,5 @@ actually executes the drop, and `get_drop_sql`, which returns the template.
 
 {% macro default__drop_materialized_view(relation) -%}
     drop materialized view if exists {{ relation }} cascade
+    {{ adapter.cache_dropped(relation) }}
```
mikealfare, line 14: I think this needs to go in the main macro that dispatches here, and then remove it from the postgres versions below. It seems like we're missing a drop scenario based on what's happening below. i.e. we're dropping the object, not updating the cache, doing all the replace work, and then trying to rename back to the thing that we dropped (but which is still in the cache).

McKnight-42: Can do, I initially started pulling them into these calls because I worried the dispatch macro wasn't going to get stopped especially on `drop` as it is an actual return on the call

core/dbt/include/global_project/macros/relations/materialized_view/rename.sql
```diff
@@ -1,4 +1,12 @@
 {% macro get_rename_materialized_view_sql(relation, new_name) %}
+    {% set database = relation.database %}
+    {% set schema = relation.schema %}
+    {% set to_relation = adapter.get_relation(database=database, schema=schema, identifier=new_name) %}
+    {% if relation is not none %}
+        {% if to_relation is not none %}
+            {{ adapter.cache_renamed(from_relation=relation, to_relation=to_relation) }}
+        {% endif %}
+    {% endif %}
```
McKnight-42, line 9: @mikealfare Will this work or be a issue with the flow of 
`get_create_backup_sql` and `get_rename_intermediate_sql` both calling `get_rename_sql` which then calls `get_rename_materialzied_view_sql`

so we would technically be hitting our addition twice?

mikealfare: This should go in the generic macro, I think it's called `get_rename_sql`.

McKnight-42: I've tried moving this up, into the generic default call, and initial dispatch call and `still get a cache inconsistency error for orders already existing`

McKnight-42:
The current fix seems to be to remove ability of `Postgers` to try and rename through the workflow we have in `get_replace_sql` this will send us back to the behavior of `create` and `drop` the user was experiencing before updating.

the cause seems to be a dependency between renaming `backup` temporary relation to the model name that causes an issue during drop and cascade when another model that is also a materialized view refs it.

mikealfare (approved):

PR #9969, opened 2024-04-17
[BACKPORT 1.5] bump sqlparse

emmyoop:
Backport df2bf132fd77a777875fe5a8f640fe3c11c28321 and 3f52cc4278f0e3fb456049a2b0e14f99b37da92a to 1.5.latest

### DIFF
--- a/.changes/unreleased/Dependencies-20240415-202426.yaml
+++ b/.changes/unreleased/Dependencies-20240415-202426.yaml
@@ -1,0 +1,6 @@
+kind: Dependencies
+body: Bump sqlparse to >=0.5.0, <0.6.0
+time: 2024-04-15T20:24:26.768707-05:00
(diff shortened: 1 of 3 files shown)


emmyoop:
Unfortunately, fixing this issue isn't straightforward for us for `dbt-core` v1.5.

While we'd love to update `sqlparse` to a safer version, it no longer supports Python 3.7, and `dbt-core` v1.5 relies on Python 3.7 compatibility. That means if we update `sqlparse`, we would break compatibility with our supported Python version.

Given that `dbt-core` v1.5 will [soon reach its end of life](https://docs.getdbt.com/docs/dbt-versions/core#:~:text=Critical%20%E2%80%94%20Apr%2027%2C%202024), we've decided not to update `sqlparse`.

Issue #7673, opened 2023-05-21
[Bug] Selectors in `selectors.yml` do not default to `--indirect-selection` or `DBT_INDIRECT_SELECTION`


dbeatty10:
### Is this a new bug in dbt-core?

- [X] I believe this is a new bug in dbt-core
- [X] I have searched the existing issues, and I could not find an existing issue for this bug

### Current Behavior

When a [YAML selector](https://docs.getdbt.com/reference/node-selection/yaml-selectors#indirect-selection) definition does not include `indirect_selection`, then neither the `--indirect-selection` CLI argument nor the `DBT_INDIRECT_SELECTION` environment variable have any effect.

Instead, these YAML selectors always default to `eager`, no matter what.

### Expected Behavior

[This](https://github.com/dbt-labs/docs.getdbt.com/blob/89f838049f13829312edf833f6cbdec5e604fa5e/website/docs/reference/node-selection/yaml-selectors.md?plain=1#L167) says:
> If provided, a YAML selector's `indirect_selection` value will take precedence over the CLI flag `--indirect-selection`. 

And [this](https://github.com/dbt-labs/dbt-core/blob/87ea28fe84edd65d589fd3cf658d8dd1f65c000a/core/dbt/graph/selector_spec.py#L120) says:
> If defined field in selector, override CLI flag

Both of those imply that the intent was that selectors would use `get_flags().INDIRECT_SELECTION` or `flags.INDIRECT_SELECTION` when `indirect_selection` is not defined within a selector in `selectors.yml`.

## Reprex

<details>
<summary>

### Project contents
</summary>

`models/_models.yml`
```yaml
version: 2

models:
  - name: model_b
    columns:
      - name: id
        tests:
          - relationships:
              to: ref('model_a')
              field: id
  - name: model_c
    columns:
      - name: id
        tests:
          - relationships:
              to: ref('model_b')
              field: id
```

`models/model_a.sql`
```sql
select 1 as id
```

`models/model_b.sql`
```sql
select * from {{ ref("model_a") }}
```

`models/model_c.sql`
```sql
select * from {{ ref("model_b") }}
```

`selectors.yml`
```yaml
selectors:
  - name: my_selector
    definition:
      method: fqn
      value: model_b
```

</details>

### Try it
Compare the output of these commands:
```shell
dbt ls -s model_b --indirect-selection eager
dbt ls -s model_b --indirect-selection cautious
dbt ls -s model_b --indirect-selection buildable
```
[...]

PR #10009, opened 2024-04-22
[Bugfix][Selection] yaml selector do not obey default overwrite

### LINKED ISSUE #9776: [Bug] indirect_selection in yaml selector definition doesn't work
### Current behavior

With a yaml selector in selectors.yml, indirect_selection doesn't work:
```
  - name: buildable_selector
    definition:
      union:
        - method: path
          value: models/some_output.sql
          children: false
          parents: true
          indirect_selection: buildable
```
If the same selector is defined using the CLI and --indirect-selection, it does work.

### Expected behavior

"indirect_selection" defined in a yaml selectors works as expected.

### Steps to reproduce

Create a selector with indirect_selection defined.

### Additional context

It looks like some code was added last year via a community contributor PR which doesn't function as expected for yaml selectors.

ChenyuLInx:
resolves #9776 #7673

### Problem
If user define a YAML selector and does not define indirect selection field on it. We would always use the default value `eager` instead of using the resolved default value of the CLI flag and env vars.(meaning the user cannot overwrite it)
See more detail in #7673

### Solution
This patch would cause a selector default to the resolved value of CLI, env var, specified by user.
This patch also removed the technical debt of always passing an indirect selection field across multiple layers where we should really just do one lookup when selector is created.

### Checklist

### DIFF
--- a/core/dbt/graph/selector_spec.py
+++ b/core/dbt/graph/selector_spec.py
@@ -4,11 +4,13 @@ from abc import ABCMeta, abstractmethod
 from dataclasses import dataclass
 from dbt_common.dataclass_schema import StrEnum, dbtClassMixin
 
+
 from typing import Set, Iterator, List, Optional, Dict, Union, Any, Iterable, Tuple
 from .graph import UniqueId
 from .selector_methods import MethodName
 from dbt_common.exceptions import DbtRuntimeError
 from dbt.exceptions import InvalidSelectorError
+from dbt.flags import get_flags
 
 
 RAW_SELECTOR_PATTERN = re.compile(
@@ -110,7 +112,6 @@ class SelectionCriteria:
         cls,
         raw: Any,
         dct: Dict[str, Any],
-        indirect_selection: IndirectSelection = IndirectSelection.Eager,
     ) -> "SelectionCriteria":
         if "value" not in dct:
             raise DbtRuntimeError(f'Invalid node spec "{raw}" - no search value!')
@@ -121,7 +122,7 @@ class SelectionCriteria:
 
         # If defined field in selector, override CLI flag
         indirect_selection = IndirectSelection(
-            dct.get("indirect_selection", None) or indirect_selection
+            dct.get("indirect_selection", get_flags().INDIRECT_SELECTION)
         )
 
         return cls(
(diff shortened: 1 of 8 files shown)


core/dbt/graph/selector_spec.py
```diff
@@ -121,7 +122,7 @@ def selection_criteria_from_dict(
 
         # If defined field in selector, override CLI flag
         indirect_selection = IndirectSelection(
-            dct.get("indirect_selection", None) or indirect_selection
+            dct.get("indirect_selection", get_flags().INDIRECT_SELECTION)
```
ChenyuLInx, line 125: This is the most important change that leads to resolving the bug and also removes all of the variable pass-through.

tests/unit/graph/test_selector_spec.py
```diff
@@ -0,0 +1,48 @@
...
+from unittest.mock import patch
+
+from dbt.graph.selector_spec import SelectionCriteria, IndirectSelection
+
+
+@pytest.mark.parametrize(
+    "indirect_selection_value,expected_value",
+    [(v, v) for v in IndirectSelection],
+)
+def test_selection_criteria_default_indirect_value(indirect_selection_value, expected_value):
```
ChenyuLInx, line 11: New unit test added for this

core/dbt/graph/selector_spec.py
MichelleArk: What is the motivation for removing the variable pass-through here? I think it could be preferable to call `get_flags().INDIRECT_SELECTION` in `runnable.py` to avoid lower-level methods accessing global state.

ChenyuLInx: I looked up all call paths using this variable and all of them are referring to `flags. INDIRECT_SELECTION`. We can either pass it all the way everywhere or call it once in this lower level function(since this is where the selector being created.).
I think calling once at the lower level is a cleaner implementation overall. Since dbt is built on the assumption of flags is the first thing being constructed during an invocation.

MichelleArk: Okay, that's fair. I recall from our 'de-globalization' conversations in the past that `flags` was more acceptable to access because its never mutated and initialized at the very beginning of an invocation, as you mentioned.

MichelleArk (approved):

adrian-pasek-prv:
I think there are still some issues with selectors and indirect-selection flags. When using `dbt build --selector my_selector --indirect-selection=cautious` neither CLI flag nor selector's `indirect-selection` section overrides `DBT_INDIRECT_SELECTION` env var. I'm always getting `'indirect_selection': 'eager'` in the logs during invocations.

PR #10031, opened 2024-04-25
Pin `macos` test runners to `macos-12`

mikealfare:
### Problem

`macos-latest` now points to `macos-14-arm64`; it used to point to `macos-12`. There is no `py39` nor `py38` on `macos-14-arm64`.

### Solution

Pin to `macos-12` for now and discuss long term options to use `macos-14-arm64`, and CI pinning in general.

### Checklist

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -135,7 +135,7 @@ jobs:
       - name: generate include
         id: generate-include
         run: |
-          INCLUDE=('"python-version":"3.8","os":"windows-latest"' '"python-version":"3.8","os":"macos-latest"' )
+          INCLUDE=('"python-version":"3.8","os":"windows-latest"' '"python-version":"3.8","os":"macos-12"' )
           INCLUDE_GROUPS="["
           for include in ${INCLUDE[@]}; do
               for group in $(seq 1 ${{ env.PYTHON_INTEGRATION_TEST_WORKERS }}); do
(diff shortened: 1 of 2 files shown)


.github/workflows/main.yml
```diff
@@ -135,7 +135,7 @@ jobs:
       - name: generate include
         id: generate-include
         run: |
-          INCLUDE=('"python-version":"3.8","os":"windows-latest"' '"python-version":"3.8","os":"macos-latest"' )
+          INCLUDE=('"python-version":"3.8","os":"windows-latest"' '"python-version":"3.8","os":"macos-12"' )
```
emmyoop, line 138:
```suggestion
          INCLUDE=('"python-version":"3.11","os":"windows-latest"' '"python-version":"3.8","os":"macos-latest"' )
```

.github/workflows/test-repeater.yml
```diff
@@ -36,7 +36,7 @@ on:
         type: choice
         options:
           - 'ubuntu-latest'
-          - 'macos-latest'
+          - 'macos-12'
```
emmyoop, line 39: This needs to stay macos-12 because we may need to test for flakiness with py-38

emmyoop (commented):
let's test with py-3.11 on macos

.github/workflows/test-repeater.yml
emmyoop:
```suggestion
          - 'macos-12'  # macos-latest doesn't support py38 or py39
```

PR #10058, opened 2024-04-26
Add ability to `silence` warnings AND allow `error` and `warn` to be used for `include` and `exclude`

### LINKED ISSUE #9644: [Feature] An option to silence warnings

### Describe the feature

https://docs.getdbt.com/reference/global-configs/warnings

```yml
warn_error_options:
  error: # previously called "include", promote to "error"
  warn: # previously called "exclude", keep as "warn"
  silence: # NEW: ignore entirely
```

(We'd retain the old names as well for backwards compatability)

Screenshot 2024-02-23 at 10 56 05

- A separate config, `WARN_SILENCE`, that takes a list. This would be better in the case where a user wants to `silence` warnings via project flags (version-controlled default for everyone), and also promote specific warnings as errors (via `include`/`exclude`) in different runtime environments, without needing to re-specify the `silence` list every time. While that's slightly more ergonomic, it's also less explicit, and it doesn't feel like a show-stopper.
- Only allow users to silence **deprecation** warnings. I think there's a valid reason to generalize this to all kinds of warnings, e.g. silence `ConstraintNotEnforced` once for the entire project rather than once per model/constraint in the project.
- Remove `TestsConfigDeprecation` (= give up in advance)

### Who will this benefit?

All users of dbt, especially since:
- In v1.8, we are renaming `tests` &rarr; to `data_tests`, for clearer distinction from `unit_tests`. Tens of thousands of projects will experience the `TestsConfigDeprecation`. They won't all be able to take action. We should give them a way to silence the deprecation (for now).
- We should still reword that deprecation message! https://github.com/dbt-labs/dbt-core/issues/9579

:)

### Anything else?

Sketched out a start:
- https://github.com/dbt-labs/dbt-common/commit/60a6ef125de49d98609550cd1ed3ba2a08ed71ea
- https://github.com/dbt-labs/dbt-core/commit/31b10023c7e56f2cc922e429c671485e9c8f1de4

Example:
```yml
$ cat dbt_project.yml
name: "my_dbt_project"

tests:
  +enabled: True

flags:
  warn_error_options:
    silence:
      - TestsConfigDeprecation
```

```
$ dbt --no-partial-parse parse
16:02:55  Running with dbt=1.8.0-a1
16:02:56  Registered adapter: duckdb=1.7.0
16:02:57  Performance info: /Users/jerco/dev/scratch/testy/target/perf_info.json
```

[Slack thread with more context](https://dbt-labs.slack.com/archives/C03SK9W08BC/p1708704701904679)

QMalcolm:
resolves #9644

### Problem

* People wanted to be able to silence warnings
* People wanted to be able to use `error` instead of `include`
* People wanted to be able to use `warn` instead of `exclude`

### Solution

* Bumped dbt_common to 1.0.2 which added `silence` to `WarnErrorOptions`
* Support `silence` arg to `warn_error_options` via CLI args and `dbt_project.yml`
* Support `error` and `warn` to `warn_error_options` as alternatives to `include` and `exclude

### Checklist

### DIFF
--- a/.changes/unreleased/Features-20240426-233126.yaml
+++ b/.changes/unreleased/Features-20240426-233126.yaml
@@ -1,5 +1,5 @@
 kind: Features
-body: Abillity to `silence` warnings via `warn_error_options`
+body: Ability to `silence` warnings via `warn_error_options`
 time: 2024-04-26T23:31:26.601057-05:00
 custom:
   Author: QMalcolm
--- a/.changes/unreleased/Features-20240426-233126.yaml
+++ b/.changes/unreleased/Features-20240426-233126.yaml
@@ -1,0 +1,6 @@
+kind: Features
+body: Abillity to `silence` warnings via `warn_error_options`
+time: 2024-04-26T23:31:26.601057-05:00
+custom:
+  Author: QMalcolm
+  Issue: "9644"
--- a/core/dbt/config/project.py
+++ b/core/dbt/config/project.py
@@ -26,8 +26,8 @@ from dbt_common.clients.system import path_exists, load_file_contents
 from dbt.clients.yaml_helper import load_yaml_text
 from dbt.adapters.contracts.connection import QueryComment
 from dbt.exceptions import (
-    DbtConfigError,
     DbtProjectError,
+    DbtWarnErrorOptionsError,
     ProjectContractBrokenError,
     ProjectContractError,
     DbtRuntimeError,
@@ -846,7 +846,7 @@ def read_project_flags(project_dir: str, profiles_dir: str) -> ProjectFlags:
 
             ProjectFlags.validate(project_flags)
             return ProjectFlags.from_dict(project_flags)
-    except (DbtProjectError, DbtConfigError) as exc:
+    except (DbtProjectError, DbtWarnErrorOptionsError) as exc:
         # We don't want to eat the DbtProjectError for UserConfig to ProjectFlags or
         # DbtConfigError for warn_error_options munging
         raise exc
--- a/core/dbt/config/project.py
+++ b/core/dbt/config/project.py
@@ -27,7 +27,7 @@ from dbt.clients.yaml_helper import load_yaml_text
 from dbt.adapters.contracts.connection import QueryComment
 from dbt.exceptions import (
     DbtProjectError,
-    DbtWarnErrorOptionsError,
+    DbtExclusivePropertyUseError,
     ProjectContractBrokenError,
     ProjectContractError,
     DbtRuntimeError,
@@ -840,12 +840,16 @@ def read_project_flags(project_dir: str, profiles_dir: str) -> ProjectFlags:
             # handle collapsing `include` and `error` as well as collapsing `exclude` and `warn`
             # for warn_error_options
             warn_error_options = project_flags.get("warn_error_options")
-            exclusive_primary_alt_value_setting(warn_error_options, "include", "error")
-            exclusive_primary_alt_value_setting(warn_error_options, "exclude", "warn")
+            exclusive_primary_alt_value_setting(
+                warn_error_options, "include", "error", "warn_error_options"
+            )
+            exclusive_primary_alt_value_setting(
+                warn_error_options, "exclude", "warn", "warn_error_options"
+            )
 
             ProjectFlags.validate(project_flags)
             return ProjectFlags.from_dict(project_flags)
-    except (DbtProjectError, DbtWarnErrorOptionsError) as exc:
+    except (DbtProjectError, DbtExclusivePropertyUseError) as exc:
         # We don't want to eat the DbtProjectError for UserConfig to ProjectFlags or
         # DbtConfigError for warn_error_options munging
         raise exc
--- a/core/dbt/config/project.py
+++ b/core/dbt/config/project.py
@@ -837,12 +837,11 @@ def read_project_flags(project_dir: str, profiles_dir: str) -> ProjectFlags:
             project_flags = profile_project_flags
 
         if project_flags is not None:
-            # if warn_error_options are set, handle collapsing `include` and `error` as well as
-            # collapsing `exclude` and `warn`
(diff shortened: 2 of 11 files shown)


core/setup.py
```diff
@@ -75,7 +75,7 @@
         "minimal-snowplow-tracker>=0.0.2,<0.1",
         "dbt-semantic-interfaces>=0.5.1,<0.6",
         # Minor versions for these are expected to be backwards-compatible
-        "dbt-common>=1.0.1,<2.0",
+        "dbt-common>=1.0.2,<2.0",
```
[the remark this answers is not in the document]
QMalcolm: The addition of the attr `silence` to the class `WarnErrorOptions` is new, and got released in dbt-common `1.0.2`. If we don't bump it, the minimum then it would be possible for us to get dbt-common `1.0.1` which doesn't support silence, and we'd break at run time. [Here is the work in dbt-common](https://github.com/dbt-labs/dbt-common/pull/112) which added the functionality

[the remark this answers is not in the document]
emmyoop: I started a thread in slack about this.  Unrelated to you actual PR - more around how this should have been a minor bump instead of a patch.

.changes/unreleased/Features-20240426-233126.yaml
```diff
@@ -0,0 +1,6 @@
+kind: Features
+body: Abillity to `silence` warnings via `warn_error_options`
```
MichelleArk, line 2: super nit: Abillity -> Ability

core/dbt/config/utils.py
```diff
@@ -23,3 +23,25 @@ def parse_cli_yaml_string(var_string: str, cli_option_name: str) -> Dict[str, An
...
+    Sometimes we allow setting something via TWO keys, but not at the same time. If both the primary
+    key and alt key have values, an error gets raised. If the alt key has values, then we update
+    the dictionary to ensure the primary key contains the values. If neither are set, nothing happens.
+    """
+
+    primary_options = dictionary.get(primary)
+    alt_options = dictionary.get(alt)
+
+    if primary_options and alt_options:
+        raise DbtWarnErrorOptionsError(
```
MichelleArk, line 42: This error indicates to me that the method should only be used in the context of `warn_error_options` configuration, and not more generally. However, it's defined generality in config.utils. Is there somewhere closer to WarnErrorOptions (perhaps on the class itself) that this method would be more appropriate? 

Alternatively, we could raise a more generic error here and have the caller catch + handle it for its specific context.

QMalcolm: I think we should probably go the more generic error here and have the the caller catch + handle it for its specific context. The question is what existing error to use or what new error to create. How about something like `DbtExclusivePropertyUseError`?

Currently intentionally skip a whole bunch of errors in [read_project_flags](https://github.com/dbt-labs/dbt-core/blob/49b66e10549190128a1a228a41a2d3f31c1f2ef1/core/dbt/config/project.py#L853). I originally wrote `exclusive_primary_alt_value_setting` to raise a `DbtRuntimeError`, but this was explicitly being swallowed by `read_project_flags` exception handling. I then tried `DbtConfigError`, which actually seemed more reasonable and better fitting. However it turns out that `DbtConfigError` is a subclass of `DbtRuntimeError` and thus was being swallowed by the same logic mentioned previously. Additionally if we start raising instead of swallowing `DbtConfigError`s some tests start failing, which means we actually depend on `DbtConfigError` being swallowed here 🙃 

---
Separately, we _could_ put this in `WarnErrorOptions` in `dbt-common`, and if we want to we can. However, I want to push back against doing so:
1. This likely won't be our last instance of changing terminology where we want to use this logic (thus having the generalized implementation is useful)
2. The conceptually the flexibility is on the input end, not on the object end. Thus by implementing the flexibility on the input end, the implementation matches the conceptual mental model.
3. (fuzzier) It's been discussed whether events should be more dynamic (i.e. the level -- `debug`, `info`, `warn`, `error` -- can configured for any event). If we move that way, we'd likely want to create a new object and deprecate `WarnErrorOptions`, and instead of having to duplicate the logic across objects, it exists separately in core prior to instantiation.

core/dbt/config/project.py
```diff
@@ -835,10 +837,18 @@ def read_project_flags(project_dir: str, profiles_dir: str) -> ProjectFlags:
             project_flags = profile_project_flags
 
         if project_flags is not None:
+            # if warn_error_options are set, handle collapsing `include` and `error` as well as
+            # collapsing `exclude` and `warn`
+            warn_error_options = project_flags.get("warn_error_options")
+            if warn_error_options:
```
MichelleArk, line 843: non-blocking suggestion: It could be worth having `exclusive_primary_alt_value_setting` handle `None` as an input so that callers don't have to concern themselves with this logic, given that the preprocessing of `warn_error_options` is happening outside the `WarnErrorOptions` class in dbt-common and may need to be duplicated

core/dbt/config/utils.py
MichelleArk: `DbtExclusivePropertyUseError` sounds good 👍 . We can also pass in the `config_name` (in this case, "warn_error_options") as a parameter to this method for templating the error message

MichelleArk (approved):
one outstanding comment but otherwise LGTM! Thank you for taking the extra care to make this change backward compatible ✨

core/dbt/config/utils.py
```diff
@@ -23,3 +23,25 @@ def parse_cli_yaml_string(var_string: str, cli_option_name: str) -> Dict[str, An
...
+            f"Only `{alt}` or `{primary}` can be specified in `warn_error_options`, not both"
+        )
```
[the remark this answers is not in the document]
QMalcolm: I finally read the question correctly 🙃 I don't think I had enough coffee when I first read it. Glad I managed to answer to answer the question in my sleep deprived state. Also github needs to start supporting the :melt: emoji.

Issue #10078, opened 2024-05-02
[Feature] support for local/package level and serializable python modules/macros.


woods-chen:
### Is this your first time submitting a feature request?

- [X] I have read the [expectations for open source contributors](https://docs.getdbt.com/docs/contributing/oss-expectations)
- [X] I have searched the existing issues, and I could not find an existing issue for this feature
- [X] I am requesting a straightforward extension of existing dbt functionality, rather than a Big Idea better suited to a discussion

### Describe the feature

Suppose that we need to remove (dedent) the same number of leading whitespaces in each line of a sql command or a string of documentation. I've been googling for jinja solutions however that's quite complicated. And it would be quite ease if we can invoke a python module/function.
The proposed method is, we can write a macro in a python file "macros/my_python_module.py":
```
from textwrap import dedent
```
and then we call call the defined module/function in our model "models/test1.sql":
```
{% set query_cmd %}
    select col_1,
              col_2
    from table_name
    where col3=1
        and col4=2
{% endset %}
{{ my_python_module.dedent(query_cmd) }}
````
What's more, as the implementation of this python macro is within the project locally, we can call the macro using `my_project.my_python_module.dedent` if this project is imported to other projects as a package, just as what we do with a jinja macro.

### Describe alternatives you've considered

The requirement for "dedent" is just an example, actually we need python functions in a lot of cases. For example:
    1. Generate a datetime range from "start" to "end", in which each element is the first day of each month.
    2. Create JDBC tables in spark or call the "mysql" function in clickhouse without hard coding the configurations (host/port/password) in the sql command.
    3. Have better integrations with airflow.
    4. ......

### Who will this benefit?

The ones who need more elegant way to implement a complex logic and haven't access to dbt cloud (offline/internal use cases).

### Are you interested in contributing this feature?

Yes, I've pushed a PR.

### Anything else?

Why do we need python macros:
1. I really agree with what Anders Hejlsberg (the creator of programming languages such as Turbo Pascal, Delphi, C and TypeScript) said about Low-Code/No-Code platforms or tools in this [interview](https://thenewstack.io/microsofts-programming-language-czar-anders-hejlsberg-the-interview/):

> I think the challenge has always been with these tools, how to put appropriate escape hatches in so that when you hit the glass ceiling of what this particular low-code environment can do, you aren’t stuck. There needs to be escape hatches that allow you to then move to a fuller or complete programming environment, which means there needs to be meaningful extensibility models.

[...]

PR #10080, opened 2024-05-02
support local/package level and serializable python modules/macros. (…

woods-chen:
…#10078)

resolves #

Suppose that we need to remove (dedent) the same number of leading whitespaces in each line of a sql command or a string of documentation. I've been googling for jinja solutions however that's quite complicated. And it would be quite ease if we can invoke a python module/function.
The proposed method is, we can write a macro in a python file "macros/my_python_module.py":
```
from textwrap import dedent
```
and then we call call the defined module/function in our model "models/test1.sql":
```
{% set query_cmd %}
    select col_1,
              col_2
    from table_name
    where col3=1
        and col4=2
{% endset %}
{{ my_python_module.dedent(query_cmd) }}
````
What's more, as the implementation of this python macro is within the project locally, we can call the macro using `my_project.my_python_module.dedent` if this project is imported to other projects as a package, just as what we do with a jinja macro.

### Solution

add a `PythonModule` node class in additional to the `Macro` node class, which can I parsed and serialized.
However there are some limitations:
1. the macro written in python is in package level which is the same as jinja macros, and should be called with the heading package name if we need to call the python macro in other projects.
2. Each directory (python module) should have a "init.py" file inside except for the "macros" directory.

### Checklist

### DIFF
--- a/.changes/unreleased/Features-20240502-201419.yaml
+++ b/.changes/unreleased/Features-20240502-201419.yaml
@@ -1,0 +1,6 @@
+kind: Features
+body: support local/package level and serializable python modules/macros. (#10078)
+time: 2024-05-02T20:14:19.303925629+08:00
+custom:
+  Author: woods-chen
+  Issue: "10078"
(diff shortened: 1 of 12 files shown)


dbeatty10:
Closing per https://github.com/dbt-labs/dbt-core/issues/10078#issuecomment-2090540226

Issue #8311, opened 2023-08-03
[CT-2930] test_custom_node_colors_configs.py: wrong exception raised


gshank:
It looks like in this case, instead of a ValidationError, we're getting a ConfigUpdateError, so this is probably a sequence of operations error?

Might be able to fix it by: `with pytest.raises( (ValidationError, ConfigUpdateError) ):`

```
/home/runner/work/dbt-core/dbt-core/core/dbt/tests/util.py:105: in run_dbt
    raise res.exception
/home/runner/work/dbt-core/dbt-core/core/dbt/cli/requires.py:87: in wrapper
    result, success = func(*args, **kwargs)
/home/runner/work/dbt-core/dbt-core/core/dbt/cli/requires.py:72: in wrapper
    return func(*args, **kwargs)
/home/runner/work/dbt-core/dbt-core/core/dbt/cli/requires.py:[143](https://github.com/dbt-labs/dbt-core/actions/runs/5750441184/job/15587233233#step:8:144): in wrapper
    return func(*args, **kwargs)
/home/runner/work/dbt-core/dbt-core/core/dbt/cli/requires.py:172: in wrapper
    return func(*args, **kwargs)
/home/runner/work/dbt-core/dbt-core/core/dbt/cli/requires.py:219: in wrapper
    return func(*args, **kwargs)
/home/runner/work/dbt-core/dbt-core/core/dbt/cli/requires.py:246: in wrapper
    manifest = ManifestLoader.get_full_manifest(
/home/runner/work/dbt-core/dbt-core/core/dbt/parser/manifest.py:306: in get_full_manifest
    manifest = loader.load()
/home/runner/work/dbt-core/dbt-core/core/dbt/parser/manifest.py:465: in load
    self.parse_project(
/home/runner/work/dbt-core/dbt-core/core/dbt/parser/manifest.py:660: in parse_project
    parser.parse_file(block)
/home/runner/work/dbt-core/dbt-core/core/dbt/parser/base.py:466: in parse_file
    self.parse_node(file_block)
/home/runner/work/dbt-core/dbt-core/core/dbt/parser/base.py:427: in parse_node
    self.render_update(node, config)
/home/runner/work/dbt-core/dbt-core/core/dbt/parser/models.py:333: in render_update
    super(ModelParser, model_parser_copy).render_update(
[...]

PR #10084, opened 2024-05-02
Better error message for Private package definition

### LINKED ISSUE #10083: [Feature] Add a private package definition

### Describe the feature

dbt-cloud is adding support to allow users to define private packages in their `dependencies.yml` or `packages.yml`. The format will be similar to git package definition except that the `git` field will be the `private` field and an additional optional provider field. Example:
```
packages:
  - private: dbt-labs/my-awesome-repo
    # optional configs - same as for 'git' packages
    revision: ... # tag, release, branch, or commit SHA
    subdirectory: ...
	provider: github (if there are multiple providers)
```
In dbt-core, when such a package is defined, we should provide clear error message and tell user to use git definition.

_No response_

### Who will this benefit?

_No response_

_No response_

### Anything else?

_No response_

ChenyuLInx:
resolves #10083

### Problem
We are introducing better support for private package in dbt-cloud, which allows people to define a new package.
If we don't do anything in core, this would lead to core raising
```
  Validator Error:
  {'private': 'dbt-labs/audit_helper', 'revision': 'HEAD', 'subdirectory': 'audit_helper', 'unrendered': {'private': 'dbt-labs/audit_helper', 'revision': 'HEAD', 'subdirectory': 'audit_helper'}} is not valid under any of the given schemas
```
Instead, we should have a clear message telling people to define git package.

### Checklist

### DIFF
--- a/core/dbt/deps/resolver.py
+++ b/core/dbt/deps/resolver.py
@@ -17,7 +17,7 @@ from dbt.deps.local import LocalUnpinnedPackage
 from dbt.deps.registry import RegistryUnpinnedPackage
 from dbt.deps.tarball import TarballUnpinnedPackage
 from dbt.exceptions import (
-    DbtInternalError,
+    DependencyError,
     DuplicateDependencyToRootError,
     DuplicateProjectDependencyError,
     MismatchedDependencyTypeError,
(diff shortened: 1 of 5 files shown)


tests/functional/configs/test_custom_node_colors_configs.py
```diff
@@ -304,7 +305,7 @@ def test__invalid_color_config_block(
         self,
         project,
     ):
-        with pytest.raises(ValidationError):
+        with pytest.raises((ValidationError, ConfigUpdateError)):
```
ChenyuLInx, line 308: This caused an issue in CI because of the execution order of tests. Since Gerda also [ran into this before](https://github.com/dbt-labs/dbt-core/issues/8311), I am just fixing it here.

core/dbt/deps/resolver.py
```diff
@@ -74,6 +75,10 @@ def update_from(self, src: List[PackageSpec]) -> None:
                 pkg = TarballUnpinnedPackage.from_contract(contract)
             elif isinstance(contract, GitPackage):
                 pkg = GitUnpinnedPackage.from_contract(contract)
+            elif isinstance(contract, PrivatePackage):
+                raise DbtInternalError(
```
MichelleArk, line 79: I don't think this should be a DbtInternalError if it can be caused by user-configuration and is resolvable given user action

MichelleArk (approved):

PR #10085, opened 2024-05-03
[Tidy First] Use `isort` to order/format/de-dupe python imports

QMalcolm:
resolves N/A

### Problem

Our imports haven't been ordered ever, probably because doing so manually can be annoying (especially when things are already a mess). However having them out of order actually slows us (at least me) down because when we want to import something its harder to visually figure out if the necessary module is already being imported from or not. 

This disorganization of imports caused instances where we were importing from the same module multiple times in the same file by accident. I wrote a short script to figure out how many distinct duplicate module import paths we had (this is when an `from x.y.z import` occurs more than once in a single file). The script found 60 instances across 50 files 🙃 One example of this for a visual representation is that `from dbt.artifacts.resources import` appeared _twice_ in [unparsed.py](https://github.com/dbt-labs/dbt-core/blob/dc744f6e8ffaff62a71b33562f2c59f4f9ee8aa9/core/dbt/contracts/graph/unparsed.py) ([Exhibit A](https://github.com/dbt-labs/dbt-core/blob/dc744f6e8ffaff62a71b33562f2c59f4f9ee8aa9/core/dbt/contracts/graph/unparsed.py#L5), [Exhibit B](https://github.com/dbt-labs/dbt-core/blob/dc744f6e8ffaff62a71b33562f2c59f4f9ee8aa9/core/dbt/contracts/graph/unparsed.py#L16)). In another file there was a module path that was imported from 3 separate times...

### Solution

We've been using [isort](https://pycqa.github.io/isort/) in other repositories such as [metricflow](https://github.com/dbt-labs/metricflow/blob/16e50baad4d798ff829a492638b4dbee49cdbde7/ruff.toml#L82) and [dbt-semantic-intefaces](https://github.com/dbt-labs/dbt-semantic-interfaces/blob/e4f029b29c2baab1acee4b9d0dc915a5574525c4/.pre-commit-config.yaml#L15) for this exact problem. So I've gone and added `isort` here as well, and it'll be invoked in pre-commit hooks.

### Checklist


QMalcolm:
I've added the `artifact-minor-upgrade` label because the files in `artifacts` have been changed, but it's just the imports.

QMalcolm:
I think it's debatable if a changie doc is needed for this. We've added a dependency, but it's only a dev dependency. We've made changes in the source code that we distribute, but it's just the imports that are touched. Seems almost less than an `Under the Hood` change.

.isort.cfg
mikealfare, line 1: Will `pre-commit` acknowledge these settings? I thought they would need to be in `pre-commit-config.yaml` as well.

PR #10094, opened 2024-05-06
Enable serialization context

### LINKED ISSUE #10093: Enable contexts for serialization
### Description

In some cases we'd like to serialize differently for creating artifacts and saving the internal manifest (and possibly other purposes too). Mashumaro has a feature to enable passing along a context dictionary to pre_serialize and post_serialize calls.

### Acceptance Criteria

We can use a context parameter in to_dict calls: `obj.to_dict(context={"test": True})`.

### Impact to other teams

We need to change the signature of all pre_ and post_serialize calls, so we need to verify that nobody else is using these.

### Will backports be required

No

gshank:
resolves #10093

### Problem

We want to be able to serialize differently in different places.

### Solution

Enable mashumaro serialization context feature.

### Checklist


core/dbt/artifacts/resources/v1/model.py
```diff
@@ -31,3 +31,9 @@ class Model(CompiledResource):
     latest_version: Optional[NodeVersion] = None
     deprecation_date: Optional[datetime] = None
     defer_relation: Optional[DeferRelation] = None
+
+    def __post_serialize__(self, dct: Dict, context: Optional[Dict] = None):
+        dct = super().__post_serialize__(dct, context)
+        if context and context.get("artifact") and "defer_relation" in dct:
```
MichelleArk, line 37: Is this strictly a refactor or would we be starting to include the `defer_relation` in the programmatic response from `dbtRunner` invocations with this change?

gshank: This just maintains current functionality, which is that when generating artifacts we remove "defer_relation" and "call_config_dict" from all nodes. We could easily change it now if we want, but in order to minimize disruption I went with preserving the way it currently works.

gshank: Both of those are included in the msgpack versions (for internal use/partial parsing)

MichelleArk (approved):

Issue #10122, opened 2024-05-10
[Bug] TypeError: __post_serialize__() takes 2 positional arguments but 3 were given


popcornylu:
### Is this a new bug in dbt-core?

- [X] I believe this is a new bug in dbt-core
- [X] I have searched the existing issues, and I could not find an existing issue for this bug

### Current Behavior

Cannot run any dbt subcommand. Exit with exception. Please see the log below.

### Expected Behavior

Should run successfully

### Steps To Reproduce

1. Update to the lastest dbt-core. `dbt-core==1.8.0`
2. Run the `dbt run` (or other dbt subcommand, list, parse)

### Relevant log output

```shell
[0m13:44:35.107219 [debug] [MainThread]: Sending event: {'category': 'dbt', 'action': 'invocation', 'label': 'start', 'context': [<snowplow_tracker.self_describing_json.SelfDescribingJson object at 0x102a11460>, <snowplow_tracker.self_describing_json.SelfDescribingJson object at 0x105909bb0>, <snowplow_tracker.self_describing_json.SelfDescribingJson object at 0x105909d60>]}


============================== 13:44:35.110287 | c1779630-42a6-4a47-842b-a5551cc89e80 ==============================
[0m13:44:35.110287 [info ] [MainThread]: Running with dbt=1.8.0
[0m13:44:35.110712 [debug] [MainThread]: running dbt with arguments {'printer_width': '80', 'indirect_selection': 'eager', 'log_cache_events': 'False', 'write_json': 'True', 'partial_parse': 'True', 'cache_selected_only': 'False', 'profiles_dir': '/Users/popcorny/Documents/infuseai/repo/dbt-infusetude', 'version_check': 'True', 'debug': 'False', 'log_path': '/Users/popcorny/Documents/infuseai/repo/dbt-infusetude/logs', 'fail_fast': 'False', 'warn_error': 'None', 'use_colors': 'True', 'use_experimental_parser': 'False', 'no_print': 'None', 'quiet': 'False', 'log_format': 'default', 'static_parser': 'True', 'warn_error_options': 'WarnErrorOptions(include=[], exclude=[])', 'introspect': 'True', 'target_path': 'None', 'invocation_command': 'dbt run', 'send_anonymous_usage_stats': 'True'}
[...]

PR #10123, opened 2024-05-10
Update minimum req for dbt-common + dbt-adapters

jtcohen6:
prevent #10122 from happening to more people

### Problem

If someone has older installs of `dbt-core` + `dbt-common` + `dbt-adapters`, and they `pip install dbt-core`, it won't automatically upgrade `dbt-common` + `dbt-adapters`.

But `dbt-core==1.8.0` actually requires `dbt-common>=1.0.4` and `dbt-adapters>=1.1.1`, given the change in #10094.

### Solution

Bump the minimum requirements for `dbt-common` + `dbt-adapters`

### Checklist

### DIFF
--- a/core/setup.py
+++ b/core/setup.py
@@ -75,8 +75,8 @@ setup(
         "minimal-snowplow-tracker>=0.0.2,<0.1",
         "dbt-semantic-interfaces>=0.5.1,<0.6",
         # Minor versions for these are expected to be backwards-compatible
-        "dbt-common>=1.0.2,<2.0",
-        "dbt-adapters>=0.1.0a2,<2.0",
+        "dbt-common>=1.0.4,<2.0",
+        "dbt-adapters>=1.1.1,<2.0",
         # ----
         # Expect compatibility with all new versions of these packages, so lower bounds only.
         "packaging>20.9",


core/setup.py
```diff
@@ -75,8 +75,8 @@
         "minimal-snowplow-tracker>=0.0.2,<0.1",
         "dbt-semantic-interfaces>=0.5.1,<0.6",
         # Minor versions for these are expected to be backwards-compatible
-        "dbt-common>=1.0.2,<2.0",
-        "dbt-adapters>=0.1.0a2,<2.0",
+        "dbt-common>=1.0.4,<2.0",
```
aranke, line 78: Do we want to set bounds to patch versions, or is minor version sufficient?

For external packages, it looks like we default to pinning to minor version (unless there’s a bug that got fixed in a patch version).

aranke (approved):
LGTM, 1 nit

core/setup.py
jtcohen6: @aranke `dbt-core==1.8.0` is actually incompatible with `dbt-common==1.0.3` and `dbt-adapters==1.1.0`, due to the `__post_serialize__` signature change. I think we need to put the specific patch numbers as the minima here to make clear that incompatibility. It wasn't our proudest semver moment, but we're learning from it and moving forward :)

Issue #7914, opened 2023-06-21
[CT-2723] [spike+] Maximally parallelize `dbt clone` operations, a different mechanism for processing a queue


aranke:
_Originally posted by @jtcohen6 in https://github.com/dbt-labs/dbt-core/pull/7881#discussion_r1236713450_

> One thing I noticed in my previous spike, and I'm seeing in this implementation as well: Even though each node's `clone` operation is completely independent of any other node, we are still running `clone` in DAG order. That significantly limits how fast this can be with `--threads 1000`.

> I'm not sure how to fix that, and it will require digging into our base task / graph queue logic — but I suspect that will be key to unlocking a lot of the value here.

> Repro
```
-- models/model_a.sql
select 1 as id

-- models/model_b.sql
select * from {{ ref('model_a') }} 
```
```
$ dbt run --target prod
$ mkdir state
$ mv target/manifest.json state/
$ dbt clone --target dev --state state --full-refresh
```
> From the logs: Notice that `model_b` runs in serial after `model_a`, even though they're on different threads. dbt understands that `model_b` depends on `model_a`, even though for the purposes of the `clone` task, it really doesn't.
```
============================== 11:36:51.356566 | 2fab3bf3-3220-47f5-a842-a8e7aee780d1 ==============================
[0m11:36:51.356566 [info ] [MainThread]: Running with dbt=1.6.0-b4
[0m11:36:51.360523 [debug] [MainThread]: running dbt with arguments {'printer_width': '80', 'indirect_selection': 'eager', 'write_json': 'True', 'log_cache_events': 'False', 'partial_parse': 'True', 'cache_selected_only': 'False', 'warn_error': 'None', 'version_check': 'False', 'debug': 'False', 'log_path': '/Users/jerco/dev/scratch/testy/logs', 'fail_fast': 'False', 'profiles_dir': '/Users/jerco/.dbt', 'use_colors': 'True', 'use_experimental_parser': 'False', 'no_print': 'None', 'quiet': 'False', 'warn_error_options': 'WarnErrorOptions(include=[], exclude=[])', 'static_parser': 'True', 'log_format': 'default', 'introspect': 'True', 'target_path': 'None', 'send_anonymous_usage_stats': 'False'}
[0m11:36:51.616469 [info ] [MainThread]: Registered adapter: snowflake=1.6.0-b3
[...]

PR #10129, opened 2024-05-10
Maximally parallelize dbt clone

MichelleArk:
resolves #https://github.com/dbt-labs/dbt-core/issues/7914

### Checklist

### DIFF
--- a/core/dbt/graph/queue.py
+++ b/core/dbt/graph/queue.py
@@ -25,8 +25,14 @@ class GraphQueue:
     the same time, as there is an unlocked race!
     """
 
-    def __init__(self, graph: nx.DiGraph, manifest: Manifest, selected: Set[UniqueId]) -> None:
-        self.graph = graph
+    def __init__(
+        self,
+        graph: nx.DiGraph,
+        manifest: Manifest,
+        selected: Set[UniqueId],
+        preserve_edges: bool = True,
+    ) -> None:
+        self.graph = graph if preserve_edges else nx.classes.function.create_empty_copy(graph)
         self.manifest = manifest
         self._selected = selected
         # store the queue as a priority queue.
--- a/core/dbt/graph/selector.py
+++ b/core/dbt/graph/selector.py
@@ -319,7 +319,7 @@ class NodeSelector(MethodManager):
 
         return filtered_nodes
 
-    def get_graph_queue(self, spec: SelectionSpec) -> GraphQueue:
+    def get_graph_queue(self, spec: SelectionSpec, preserve_edges: bool = True) -> GraphQueue:
         """Returns a queue over nodes in the graph that tracks progress of
         dependecies.
         """
@@ -330,7 +330,7 @@ class NodeSelector(MethodManager):
         # Construct a new graph using the selected_nodes
         new_graph = self.full_graph.get_subset_graph(selected_nodes)
         # should we give a way here for consumers to mutate the graph?
-        return GraphQueue(new_graph.graph, self.manifest, selected_nodes)
+        return GraphQueue(new_graph.graph, self.manifest, selected_nodes, preserve_edges)
 
 
 class ResourceTypeSelector(NodeSelector):
--- a/core/dbt/task/clone.py
+++ b/core/dbt/task/clone.py
@@ -91,6 +91,8 @@ class CloneRunner(BaseRunner):
 
 
 class CloneTask(GraphRunnableTask):
+    PRESERVE_EDGES = False
+
     def raise_on_first_error(self):
         return False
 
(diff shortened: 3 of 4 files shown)


MichelleArk:
I've done some local 🎩 benchmarking of `dbt clone` against bigquery (since it supports zero-copy cloning) to compare performance on this branch vs main. 

🎩  Setup: 
```
dbt run --target prod
mkdir state
mv target/manifest.json state/
```

### On a project with a chain of 5 models (i.e. 1 -> 2 -> 3 -> 4 -> 5):
Recording average execution time of `dbt clone --state state --full-refresh --threads 5` across 5 runs:
* main: 15.38s, 34% cpu utilization
* this branch: 10.17s, 48% cpu utilization

### On a project with a chain of 10 models:
Recording average execution time of `dbt clone --state state --full-refresh --threads 5` across 5 runs:
* main: 20.41s, 20% cpu utilization
* this branch: 9.02s, 57% cpu utilization

core/dbt/graph/queue.py
```diff
@@ -25,8 +25,14 @@ class GraphQueue:
...
-    def __init__(self, graph: nx.DiGraph, manifest: Manifest, selected: Set[UniqueId]) -> None:
-        self.graph = graph
+    def __init__(
+        self,
+        graph: nx.DiGraph,
+        manifest: Manifest,
+        selected: Set[UniqueId],
+        preserve_edges: bool = True,
+    ) -> None:
+        self.graph = graph if preserve_edges else nx.classes.function.create_empty_copy(graph)
```
gshank, line 35: Could you add a comment above the "create_empty_copy" saying something like "'create_empty_copy' will remove all of the edges in the graph"?  Because to me, "create_empty_copy" was not intuitive and I had to look it up.

gshank (approved):
Really nice that this was fairly simple. Would like just an additional comment as noted on the line.

core/dbt/graph/queue.py
MichelleArk: Good call! Will do 👍

tests/unit/task/test_runnable.py
```diff
@@ -23,12 +30,28 @@ class MockConfig:
...
+        nodes: Optional[List[MockNode]] = None,
+        edges: Optional[List[Tuple[str, str]]] = None,
+    ):
+        nodes = nodes or []
+        edges = edges or []
+
         self.forced_exception_class = exception_class
         self.did_cancel: bool = False
         super().__init__(args=MockArgs(), config=MockConfig(), manifest=None)
+        self.manifest = make_manifest(nodes=nodes)
```
ChenyuLInx, line 50: This feels like we are reimplementing the task logic in the test again. I view unit test as documenting behavior in this case,  so the only thing we need to test at the runnable task is: are we using the correct arguments used when calling `get_graph_queue`?

tests/unit/task/test_runnable.py
```diff
@@ -40,13 +63,25 @@ def _cancel_connections(self, pool):
 
     def get_node_selector(self):
         """This is an `abstract_method` on `GraphRunnableTask`, thus we must implement it"""
-        return None
+        selector = ResourceTypeSelector(
```
ChenyuLInx, line 66: ditto

tests/unit/task/test_runnable.py
```diff
@@ -40,13 +63,25 @@ def _cancel_connections(self, pool):
...
+        return selector
 
     def defer_to_manifest(self, adapter, selected_uids: AbstractSet[str]):
         """This is an `abstract_method` on `GraphRunnableTask`, thus we must implement it"""
         return None
 
 
+class MockRunnableTaskIndependent(MockRunnableTask):
+    def get_run_mode(self) -> GraphRunnableMode:
+        return GraphRunnableMode.Independent
```
ChenyuLInx, line 82: I think we are using inheritance where we should be using patch

tests/unit/utils/__init__.py
```diff
@@ -387,3 +388,17 @@ def replace_config(n, **kwargs):
         config=n.config.replace(**kwargs),
         unrendered_config=dict_replace(n.unrendered_config, **kwargs),
     )
+
+
+def make_manifest(nodes=[], sources=[], macros=[], docs=[]) -> Manifest:
```
ChenyuLInx, line 393: I think this is duplicating https://github.com/dbt-labs/dbt-core/blob/84456f50f687b352072f64867247d21bef2647ff/tests/unit/utils/manifest.py#L986 here, thoughts on how can we make that one easier to use/more visible?

ChenyuLInx (commented):
Late to the party and sorry about being nitpicking here.
Let's discuss a bit as this is probably going to be used as example for next set of tests

Issue #9907, opened 2024-04-12
[testing] Reduce flaky tests by retrying git failures


emmyoop:
### Housekeeping

- [X] I am a maintainer of dbt-core

### Short description

We have a lot of tests that are failing because of Git connection issues.  Sometimes tox fails to install all dependencies and that causes the entire test run to fail without actually running any tests.  This makes our monitoring noisy.

Suggested approach: leveraging something the `nick-fields/retry@v3` action ([example](https://github.com/dbt-labs/actions/blob/7738a62a39f411ddf53aa1fa104783f9b9eb007e/.github/workflows/release-branch-tests.yml#L135-L143) but in the tox invocation [here](https://github.com/dbt-labs/dbt-core/blob/fe28d9e115f9a31cedcc641f67fcd4d229ad0027/.github/workflows/main.yml#L204-L207))

### Acceptance criteria

Anytime we use git when testing, have retry logic

### Suggested Tests

This task is specifically for tests

-- can force a test to fail in a commit & observe the retry works as expected at the integration group level

### Impact to Other Teams

Adapters team won't be impacted but may be interested if we come up with a solution

### Will backports be required?

backport as far as we can to reduce this noise

### Context

log output from test failing on tox
```
Run tox -- --ddtrace
integration: install_deps> python -I -m pip install -r dev-requirements.txt -r editable-requirements.txt
  Running command git clone --filter=blob:none --quiet https://github.com/dbt-labs/dbt-adapters.git /tmp/pip-req-build-g9zkv3vu
  error: RPC failed; curl 16 Error in the HTTP2 framing layer
  fatal: expected 'packfile'
  fatal: could not fetch 22b2ad3f683cca452f28320c0aba8bb95933ca6e from promisor remote
Collecting git+https://github.com/dbt-labs/dbt-adapters.git@main (from -r dev-requirements.txt (line 1))
  Cloning https://github.com/dbt-labs/dbt-adapters.git (to revision main) to /tmp/pip-req-build-g9zkv3vu
integration: exit 1 (2.55 seconds) /home/runner/work/dbt-core/dbt-core> python -I -m pip install -r dev-requirements.txt -r editable-requirements.txt pid=1980
  warning: Clone succeeded, but checkout failed.
  You can inspect what was checked out with 'git status'
  and retry with 'git restore --source=HEAD :/'

  error: subprocess-exited-with-error
  
[...]

PR #10137, opened 2024-05-13
Fix #9907: Add retry to tox to reduce flaky tests due to network failures

aranke:
resolves #9907 

### Problem

Tox fails due to network/Git errors

### Solution

Retry tox commands to prevent flaky tests

### Checklist


.github/workflows/main.yml
```diff
@@ -201,8 +205,12 @@ jobs:
           python -m pip install tox
           tox --version
 
-      - name: Run tests
-        run: tox -- --ddtrace
+      - name: Run integration tests
+        uses: nick-fields/retry@v3
+        with:
+          timeout_minutes: 10
```
emmyoop, line 211: Should this be 30 since the timeout for this job is 30 minutes?

.github/workflows/structured-logging-schema-check.yml
```diff
@@ -94,7 +94,11 @@ jobs:
       # integration tests generate a ton of logs in different files. the next step will find them all.
       # we actually care if these pass, because the normal test run doesn't usually include many json log outputs
       - name: Run integration tests
-        run: tox -e integration -- -nauto
+        uses: nick-fields/retry@v3
+        with:
+          timeout_minutes: 10
```
emmyoop, line 99: Same thing here for 30 min

emmyoop (changes requested):

.github/workflows/main.yml
aranke: Good call, I doubled the max runtime of the last few runs (5 × 2 = 10 mins), but can set it to the timeout instead.

.github/workflows/main.yml
```diff
@@ -95,8 +95,12 @@ jobs:
           python -m pip install tox
           tox --version
 
-      - name: Run tox
-        run: tox
+      - name: Run unit tests
+        uses: nick-fields/retry@v3
+        with:
+          timeout_minutes: 5
```
ChenyuLInx, line 101: Do we need the time out here?

ChenyuLInx (approved):
I would prefer not having the timeout for unit test but otherwise looks good to me

.github/workflows/main.yml
aranke: Looks like we're setting it to 10 minutes on line 72, so I guess I'll change it to be consistent.

ChenyuLInx: Sounds like we write the exact number in different places, maybe we should consider consolidating it into one place, like a variable or something? Not a hard restriction

PR #10147, opened 2024-05-14
Repository unit tests for `ManifestLoader.get_full_manifest`

### LINKED ISSUE #9865: Write a unittest for get_full_manifest
# Description
Write unittest/s for [get_full_manifest](https://github.com/dbt-labs/dbt-core/blob/ebc22fa26c9cb26b6deef86cc0a7ceb1ee3fb642/core/dbt/parser/manifest.py#L281) function, this will act as an example for writing more unittest in parser directory 

You will likely need to
- constructing a [Manifest](https://github.com/dbt-labs/dbt-core/blob/ebc22fa26c9cb26b6deef86cc0a7ceb1ee3fb642/core/dbt/contracts/graph/manifest.py#L786) object using fixture/functions generated in #9665 and mock the `loader.load()` to return it directly.
- create some fixture/function to generate a `RuntimeConfig`.
- do some mocking for tracking.

## Non-functional requirements
Please organize tests related to `parser/manifest.py` and util functions related to make it easier for the next person to find it.

Potential existing arts: https://github.com/dbt-labs/dbt-core/blob/ebc22fa26c9cb26b6deef86cc0a7ceb1ee3fb642/tests/unit/test_manifest.py#L108

QMalcolm:
resolves #9865

### Problem

[ManifestLoader.get_full_manifest](https://github.com/dbt-labs/dbt-core/blob/d21ff38ba1aafdee67e0165856f59b31d824d144/core/dbt/parser/manifest.py#L263) is one of the central functions of dbt-core. It is what is generally used for loading the manifest. However, previously we had no unit tests for it.

### Solution

Add unit tests for `ManifestLoader.get_full_manifest`. So in reality, `get_full_manifest` is more of an orchestrator method and thus offloads a fair bit of the work to other functions. We should separately ensure the sub functions get appropriately unit tested. With that said, `get_full_manifest` has itself some logical branches that we want to guarantee with testing. The tests we've added test these logical branches
* usage of the `reset` argument
* usage of the `write_perf_info` argument
* usage of the `PARTIAL_PARSE_FILE_DIFF` flag

### Checklist

### DIFF
--- a/tests/unit/parser/test_manifest.py
+++ b/tests/unit/parser/test_manifest.py
@@ -10,6 +10,7 @@ from dbt.config import RuntimeConfig
 from dbt.contracts.graph.manifest import Manifest
 from dbt.flags import set_from_args
 from dbt.parser.manifest import ManifestLoader
+from dbt.parser.read_files import FileDiff
 from dbt.tracking import User
 
 
@@ -170,3 +171,30 @@ class TestGetFullManifest:
         ManifestLoader.get_full_manifest(config=mock_project, reset=True)
         assert mock_project.clear_dependencies.called
         assert mock_adapter.clear_macro_resolver.called
+
+    def test_partial_parse_file_diff_flag(
+        self,
+        manifest: Manifest,
+        mock_project: MagicMock,
+        mock_adapter: MagicMock,
+        mocker: MockerFixture,
+    ) -> None:
+        self.set_required_mocks(mocker, manifest, mock_adapter)
+
+        # FileDiff.from_dict is only called if PARTIAL_PARSE_FILE_DIFF == False
+        # So we can track this function call to check if setting PARTIAL_PARSE_FILE_DIFF
+        # works appropriately
+        mock_file_diff = mocker.patch("dbt.parser.read_files.FileDiff.from_dict")
+        mock_file_diff.return_value = FileDiff([], [], [])
+
+        set_from_args(Namespace(), {})
+        ManifestLoader.get_full_manifest(config=mock_project)
+        assert not mock_file_diff.called
+
+        set_from_args(Namespace(PARTIAL_PARSE_FILE_DIFF=True), {})
+        ManifestLoader.get_full_manifest(config=mock_project)
+        assert not mock_file_diff.called
+
+        set_from_args(Namespace(PARTIAL_PARSE_FILE_DIFF=False), {})
+        ManifestLoader.get_full_manifest(config=mock_project)
+        assert mock_file_diff.called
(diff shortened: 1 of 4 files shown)


tests/unit/parser/test_manifest.py
```diff
@@ -20,9 +25,25 @@ def mock_project():
     mock_project.profile_env_vars = {}
     mock_project.project_target_path = "mock_target_path"
     mock_project.credentials = MagicMock()
+    mock_project.clear_dependencies = MagicMock()
     return mock_project
 
 
+@pytest.fixture
```
ChenyuLInx, line 32: Shoude this live somewhere in the utils being used by other unittests?

tests/unit/parser/test_manifest.py
```diff
@@ -91,3 +112,89 @@ def test_partial_parse_safe_update_project_parser_files_partially(
         assert "full_reparse_reason" in exc_info
         assert "KeyError: 'Whoopsie!'" == exc_info["exception"]
         assert isinstance(exc_info["code"], str) or isinstance(exc_info["code"], type(None))
+
+
+class TestGetFullManifest:
+    def set_required_mocks(
+        self, mocker: MockerFixture, manifest: Manifest, mock_adapter: MagicMock
+    ):
+        mocker.patch("dbt.parser.manifest.get_adapter").return_value = mock_adapter
```
ChenyuLInx, line 121: Is it possible to do the patch inside the fixture? So users of those fixtures can just request the fixture then the patch is done?

ChenyuLInx (commented):
Some thoughts but otherwise LGTM

tests/unit/parser/test_manifest.py
QMalcolm: We can lift these 👍

tests/unit/parser/test_manifest.py
QMalcolm: I tried doing so but, as noted in the commit message of https://github.com/dbt-labs/dbt-core/pull/10147/commits/791ec1d64f569a59d817a9d5e0fb4b95a4b291cf, the mocks stopped propagating when I did so. I can take another stab at it, but can't make any promises. It definitely would be nice if we could do it as a fixture though.

QMalcolm: Converting it to a fixture does seem to work now. Not sure why it didn't before and why it is now 🤷

ChenyuLInx: I wonder whether it was not imported properly before

ChenyuLInx (approved):

PR #10153, opened 2024-05-15
Standardize event catching during repository unit tests

### LINKED ISSUE #9947: Create a standard way to check events being fired in Unittest
In dbt-core, we now use an event manager to handle logs. In unit tests, we want to check whether under certain conditions, dbt-core will emit certain type of logs. We have been doing this a lot in functional tests by using the `run_and_capture` function. In unittest, we should mock the event manager so we can check what `fire_event` has been called with.
AC:
- we have a standard way of checking events in unit test.
- a unit test using the new way to check events being fired.

QMalcolm:
resolves #9947

### Problem

* Any modifications to the global event manager during unit tests would affect other tests unless properly cleaned up
* There wasn't a good way to check what events were fired during unit tests

Previously if we modified the global event manager during unit tests, we then had to clean up the event manager at the end of the test.   No clean up necessary. We also moved the `EventCatcher` class 

### Solution

* In [67c87fd](https://github.com/dbt-labs/dbt-core/pull/10153/commits/67c87fdb1434920640279ad7295ab51421f84334) we added a new fixture `always_clean_event_manager` to the unit test suite
  * because we use `autouse=True` the fixture is used for every unit test. Thus every unit test will have clean state (so long as we aren't multi-threading our unit tests, which we currently don't)
* In bcb2efcf0f0f4bfb1a0c0e94ae7cdd329d5d4b68 we moved the `EventCatcher` class to `tests.utils`
  * The `EventCatcher.catch` can be added to the callbacks of the event manager to check that certain event types happen
  * Example [here](https://github.com/dbt-labs/dbt-core/blob/32a7f82772d90a4b9ab443908f5f4455d0aa2c73/tests/functional/manifest_validations/test_check_for_spaces_in_model_names.py#L18)

### Checklist

### DIFF
--- a/tests/unit/events/test_logging.py
+++ b/tests/unit/events/test_logging.py
@@ -1,7 +1,10 @@
+from argparse import Namespace
 from copy import deepcopy
 
+from pytest_mock import MockerFixture
+
 from dbt.events.logging import setup_event_logger
-from dbt.flags import get_flags
+from dbt.flags import get_flags, set_from_args
 from dbt_common.events.base_types import BaseEvent
 from dbt_common.events.event_manager_client import get_event_manager
 from dbt_common.events.logger import LoggerConfig
@@ -24,3 +27,16 @@ class TestSetupEventLogger:
         setup_event_logger(flags=flags)
         assert len(manager.loggers) == 0
         assert len(manager.callbacks) == 0
+
+    def test_specify_max_bytes(
+        self,
+        mocker: MockerFixture,
+        mock_global_event_manager,
+    ) -> None:
+        patched_file_handler = mocker.patch("dbt_common.events.logger.RotatingFileHandler")
+        args = Namespace(log_file_max_bytes=1234567)
+        set_from_args(args, {})
+        setup_event_logger(get_flags())
+        patched_file_handler.assert_called_once_with(
+            filename="logs/dbt.log", encoding="utf8", maxBytes=1234567, backupCount=5
+        )
--- a/tests/unit/events/test_logging.py
+++ b/tests/unit/events/test_logging.py
@@ -8,7 +8,7 @@ from dbt.flags import get_flags, set_from_args
 from dbt_common.events.base_types import BaseEvent
 from dbt_common.events.event_manager_client import get_event_manager
 from dbt_common.events.logger import LoggerConfig
-from tests.functional.utils import EventCatcher
+from tests.utils import EventCatcher
 
 
 class TestSetupEventLogger:
--- a/tests/unit/events/test_logging.py
+++ b/tests/unit/events/test_logging.py
@@ -1,0 +1,26 @@
+from copy import deepcopy
+
+from dbt.events.logging import setup_event_logger
+from dbt.flags import get_flags
+from dbt_common.events.base_types import BaseEvent
+from dbt_common.events.event_manager_client import get_event_manager
+from dbt_common.events.logger import LoggerConfig
+from tests.functional.utils import EventCatcher
+
+
+class TestSetupEventLogger:
+    def test_clears_preexisting_event_manager_state(self, mock_global_event_manager) -> None:
+        manager = get_event_manager()
+        manager.add_logger(LoggerConfig(name="test_logger"))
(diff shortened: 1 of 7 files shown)


QMalcolm:
Odd that the tests [are failing](https://github.com/dbt-labs/dbt-core/actions/runs/9103801494/job/25026420089?pr=10153) here but not locally. The unit test that is failing is a unit test we'd expect to fail if the event manager changes that happen in `test_specify_max_bytes` (previously `test_setup_event_logger_specify_max_bytes`) if the `cleanup_event_logger` call was deleted prior to moving to the mocked event logger. Moving to the mocked event logger solved that problem locally, but it doesn't seem to have worked in the github action 🤔

QMalcolm:
Okay so I've been doing some investigating. The tests start breaking with the change from [e08c448](https://github.com/dbt-labs/dbt-core/pull/10153/commits/e08c4481db60763508db380f8eb797eab8b7af51). This is odd though because the change in that commit is renaming the test `test_setup_event_logger_specify_max_bytes` to `test_specify_max_bytes` and moving it from `tests/unit/test_functions.py` to `tests/unit/events/test_logging.py`. You wouldn't think that moving a test would break an unrelated test `tests/unit/graph/test_selector.py::test_run_specs[a,b--]` nor alter how fixtures are working 🤔 Gonna take a closer look at [e08c448](https://github.com/dbt-labs/dbt-core/pull/10153/commits/e08c4481db60763508db380f8eb797eab8b7af51), and might possibly revert it.

QMalcolm:
Reverting [e08c448](https://github.com/dbt-labs/dbt-core/pull/10153/commits/e08c4481db60763508db380f8eb797eab8b7af51) does solve the problem 🤯 I mean this makes sense given that we identified the commit as the problem, but I still don't understand why the commit is problematic.

tests/unit/events/test_logging.py
```diff
@@ -1,7 +1,10 @@
+from argparse import Namespace
```
QMalcolm, line 1: Reverting this commit fixes our broken `tests/unit/graph/test_selector.py::test_run_specs[a,b--]` test problem. I don't understand why.

tests/unit/events/test_logging.py
```diff
@@ -24,3 +27,16 @@ def test_clears_preexisting_event_manager_state(self, mock_global_event_manager)
...
+
+    def test_specify_max_bytes(
+        self,
+        mocker: MockerFixture,
+        mock_global_event_manager,
+    ) -> None:
+        patched_file_handler = mocker.patch("dbt_common.events.logger.RotatingFileHandler")
+        args = Namespace(log_file_max_bytes=1234567)
+        set_from_args(args, {})
+        setup_event_logger(get_flags())
```
QMalcolm, line 39: One theory I had was that although this commit primarily just _moves_ the test, I also made one small change which is highlighted here. Specifically, in this test previously, these methods were called via the module instead of directly. That is these two lines were
```python
flags.set_from_args(args, {})
setup_event_logger(flags.get_flags())
```
However modifying these lines and the the associated import did not solve the broken test issue we're seeing 🤔

QMalcolm (commented):
Some investigation details

ChenyuLInx (approved):
Thanks for the detailed PR description!

ChenyuLInx:
For CI breaking I think it might be changing the name caused order of test running changed

QMalcolm:
> For CI breaking I think it might be changing the name caused order of test running changed

Good timing! I just reached the same conclusion 😂 `test_specify_max_bytes` is modifying the _global_ flags, and I think this is carrying over and causing the test failure. Previously this test happened in a different order, meaning one of two things was happening:
1. There was a test coming after `test_specify_max_bytes` but before the current breaking test which was resetting the flags previously
2. The test `test_specify_max_bytes` was happening after the breaking test previously

QMalcolm:
Turns out that mocking the global event manager wasn't isolating the event manager by test. Which is unfortunate. We now instead cleanup the event manager before every unit test. Which isn't as nice, but gets the job done.

PR #10177, opened 2024-05-19
Add more accurate RSS high water mark measurement for Linux

### LINKED ISSUE #10120: Make Memory Usage Reporting More Reliable
At present, dbt uses python's resource.getrusage() function to record the max RSS upon termination. This is not reliable in all of our use cases. Unfortunately this value persists across fork and exec calls on linux, which means that the process which kicks off dbt may set a memory floor that dbt never reaches, obscuring its actual memory usage.

To improve the situation, we can use the /proc/self/status file which should be present on Linux systems, extracting the value of the VmHWM line from that file, which reprsents the high water mark (HWM) of the resident set size. When this file is missing we should fall back to the current implementation.

peterallenwebb:
resolves #10120

### Problem

Memory measurement on Linux reflected the maximum RSS level of not just the current process, but all parent processes.

### Solution

Measure max RSS on Linux via a platform-specific mechanism that does not have this defect.

### Checklist

### DIFF
--- a/core/dbt/cli/requires.py
+++ b/core/dbt/cli/requires.py
@@ -32,6 +32,7 @@ from dbt.parser.manifest import parse_manifest
 from dbt.plugins import set_up_plugin_manager
 from dbt.profiler import profiler
 from dbt.tracking import active_user, initialize_from_flags, track_run
+from dbt.utils import try_get_max_rss_kb
 from dbt.version import installed as installed_version
 from dbt_common.clients.system import get_env
 from dbt_common.context import get_invocation_context, set_invocation_context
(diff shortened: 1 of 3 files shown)


core/dbt/utils.py
```diff
@@ -386,3 +386,21 @@ def strtobool(val: str) -> bool:
         return False
     else:
         raise ValueError("invalid truth value %r" % (val,))
+
+
+def try_get_max_rss_kb() -> Optional[int]:
+    """Attempts to get the high water mark for this process's memory use via
+    the most reliable and accurate mechanism available through the host OS.
+    Currently only implemented for Linux."""
+    try:
```
aranke, line 395: Maybe check for `sys.platform == 'linux'` in the first line to avoid the `try/catch`?

core/dbt/cli/requires.py
```diff
@@ -148,14 +149,17 @@ def wrapper(*args, **kwargs):
                 import resource
 
                 rusage = resource.getrusage(resource.RUSAGE_SELF)
+                max_rss = try_get_max_rss_kb()
```
aranke, line 152: Nit: `max_rss = try_get_max_rss_kb() or rusage.ru_maxrss`

aranke (approved):
Mostly LGTM, can be cleaned up a little

PR #10196, opened 2024-05-21
Add initial unit tests for `ModelRunner` class

### LINKED ISSUE #9867: [SPIKE+] Make a plan for how to write a unittest for ModelRunner
# Description
make a plan for how to write a unit test for the ModelRunner, and implement it if possible.
This may require:
- mocking adapter
- utils/fixtures to generate manifest that we already have
- Mocking logging system to capture logging behavior

This is not amid to have a completely tested ModelRunner. The goal should be to have a good understanding of what utils/fixtures are best for a runner's needs. And have one example to follow.

QMalcolm:
resolves #9867 

### Problem

We haven't had a way to _easily_ unit test things like the `ModelRunner` class previously. This is because there were A LOT of things to mock / build before getting to "testing". We've been building a fair amount of testing fixtures and utilities over the past few months. We wanted to investigate how close we are to being able to unit test the `ModelRunner` class.

### Solution

Attempt to unit test `ModelRunner` class. We were able to write a unit test the `print_result_line` method, which itself calls many `ModelRunner` methods. However we were unable to successfully test the _primary_ method `execute`.

### Checklist

### DIFF
--- a/tests/unit/task/test_run.py
+++ b/tests/unit/task/test_run.py
@@ -3,6 +3,7 @@ from unittest.mock import MagicMock, patch
 
 import pytest
 
+from dbt.adapters.postgres import PostgresAdapter
 from dbt.artifacts.schemas.results import RunStatus
 from dbt.artifacts.schemas.run import RunResult
 from dbt.config.runtime import RuntimeConfig
@@ -68,13 +69,13 @@ class TestModelRunner:
     @pytest.fixture
     def model_runner(
         self,
-        mock_adapter: MagicMock,
+        postgres_adapter: PostgresAdapter,
         table_model: ModelNode,
         runtime_config: RuntimeConfig,
     ) -> ModelRunner:
         return ModelRunner(
             config=runtime_config,
-            adapter=mock_adapter,
+            adapter=postgres_adapter,
             node=table_model,
             node_index=1,
             num_nodes=1,
@@ -120,7 +121,7 @@ class TestModelRunner:
         assert log_model_result_catcher.caught_events[0].data.status == EventLevel.ERROR
 
     @pytest.mark.skip(
-        reason="The mock_adapter can't be found during `FACTORY.lookup_adapter` causing a runtime error"
+        reason="The materialization table for the table_model can't be found during execution"
     )
     def test_execute(
         self, table_model: ModelNode, manifest: Manifest, model_runner: ModelRunner
--- a/tests/unit/task/test_run.py
+++ b/tests/unit/task/test_run.py
@@ -14,7 +14,7 @@ from dbt.flags import get_flags, set_from_args
 from dbt.task.run import ModelRunner, RunTask
 from dbt.tests.util import safe_set_invocation_context
 from dbt_common.events.base_types import EventLevel
-from dbt_common.events.event_manager_client import get_event_manager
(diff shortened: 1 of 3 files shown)


tests/unit/task/test_run.py
```diff
@@ -50,3 +59,72 @@ def test_run_task_preserve_edges():
...
+        )
+
+    def test_print_result_line(
+        self,
+        log_model_result_catcher: EventCatcher,
+        model_runner: ModelRunner,
+        run_result: RunResult,
+    ) -> None:
+        # Setup way to catch events
+        event_manager = get_event_manager()
```
ChenyuLInx, line 104: Should this be some setup steps vs just something in a single test?

tests/unit/task/test_run.py
```diff
@@ -50,3 +59,72 @@ def test_run_task_preserve_edges():
...
+        assert log_model_result_catcher.caught_events[0].info.level == EventLevel.ERROR
+        assert log_model_result_catcher.caught_events[0].data.status == EventLevel.ERROR
+
+    @pytest.mark.skip(
+        reason="Default and adapter macros aren't being appropriately populated, leading to a runtime error"
+    )
+    def test_execute(
+        self, table_model: ModelNode, manifest: Manifest, model_runner: ModelRunner
+    ) -> None:
+        model_runner.execute(model=table_model, manifest=manifest)
```
ChenyuLInx, line 129: Any thoughts on how we can actually test this? I think we probably want to think a bit more about what the `runner` layer logic is? Is it run a specific macro with correct parameters? or something else

tests/unit/task/test_run.py
QMalcolm: We can move this to the local `log_model_result_catcher` fixture. Additionally, the setup can be reduced by a few steps because, in the time since originally writing this, we've released a version of dbt-common which provides a `add_callback_to_manager` method. So the `log_model_result_catcher` fixture would become
```python
@pytest.fixture
def log_model_result_catcher(self) -> EventCatcher:
    catcher = EventCatcher(event_to_catch=LogModelResult)
    add_callback_to_manager(catcher.catch)
    return catcher
```

tests/unit/task/test_run.py
QMalcolm: A non-exhaustive list of things we could unit test about the `execute` function
1. The result on success
  a. `execute` should return a `RunResult`
    1. assert that the status is `success`
    2. assert that the `adapter_response` is as expected
2. If the `materialization_macro` is not found, the appropriate error is raised
  a. this is what we are unintentionally doing currently (and we want to be doing it intentionally)
3. If `config` is not in `context` the appropriate error is raised
4. If the materialization macro doesn't support the language of the model, the appropriate error is raised
5. Pre model hook is fired
6. Post model hook is fired

What is blocking most of this currently is (2.a)

ChenyuLInx: Should we open a follow up ticket about 2.a?

ChenyuLInx (approved):
LGTM!
Lets make sure we have follow up ticket to run execute(or maybe have some sort of scoped down running materialization)

tests/unit/task/test_run.py
QMalcolm: Yes indeededy 👍

PR #10242, opened 2024-05-29
Create `RuntimeConfig` fixture

QMalcolm:
resolves #10000

### Problem

We haven't had a RuntimeConfig pytest fixture for unit tests. This made writing a lot of unit tests hard because a lot of our code _depends_ on a RuntimeConfig existing. 

### Solution

Create a `runtime_config` pytest fixture and use it to rewrite some existing tests 🙂 

### Checklist

- [X] I have read [the contributing guide](https://github.com/dbt-labs/dbt-core/blob/main/CONTRIBUTING.md) and understand what's expected of me  
- [X] I have run this code in development and it appears to resolve the stated issue  
- [X] This PR includes tests, or tests are not required/relevant for this PR
- [X] This PR has no interface changes (e.g. macros, cli, logs, json artifacts, config files, adapter interface, etc) or this PR has already received feedback and approval from Product or DX
- [X] This PR includes [type annotations](https://docs.python.org/3/library/typing.html) for new and modified functions

PR #10243, opened 2024-05-29
Move to test concrete class over abstract class

### LINKED ISSUE #10001: An example unit test for GraphRunnableTask
## Update Description
Create guidance on how to test the graph runnable task and move the current test over.

## Original Description
As part of the Robust Testing initiative, we want to create examples of unit tests for people to follow.
Graph Runnable is one crucial task we have that other tasks like `ls`, `build`, `run` inheritance from.

### AC
- write 3 unit tests testing different part of the Graph RunnableTask

### Additional context
We have a few existing unit tests for task like [freshness task](https://github.com/dbt-labs/dbt-core/blob/29b83598e3530e99a441e4b6de51f9f9dace781f/tests/unit/task/test_freshness.py#L114). When working on this we would spend some additional time looking at all of the unit tests for task and come up with a more standard way of writing it.

ChenyuLInx:
resolves #10001

### Problem
We were trying to test the ABC classes which led to needing to implement logic in tests to help test the code.

### TODO
- [ ] switching to use runtime config fixture. https://github.com/dbt-labs/dbt-core/pull/10242

### Solution
Stop testing the ABC classes, instead, we should test the concrete class that uses those functionalities implemented in the ABC classes.
This patch added this guideline to README and adjusted the test_graphrunnable.py to this style.
reference: https://enterprisecraftsmanship.com/posts/how-to-unit-test-an-abstract-class/

### Checklist


tests/unit/task/test_run.py
```diff
@@ -0,0 +1,54 @@
...
+
+def test_run_task_preserve_edges():
+    mock_node_selector = MagicMock()
+    mock_spec = MagicMock()
+    with patch.object(RunTask, "get_node_selector", return_value=mock_node_selector), patch.object(
+        RunTask, "get_selection_spec", return_value=mock_spec
+    ):
+        task = RunTask(get_flags(), None, None)
+        task.get_graph_queue()
+        mock_node_selector.get_graph_queue.assert_called_with(mock_spec, True)
```
ChenyuLInx, line 54: @MichelleArk let me know what you think of this. This test is the third type of unit test mentioned [here](https://enterprisecraftsmanship.com/posts/styles-of-unit-testing/). So we don't need to validate the out put of the `get_graph_queue`, and probably should consider moving it to a private method.

tests/unit/task/test_run.py
```diff
@@ -0,0 +1,54 @@
...
+    safe_set_invocation_context()
+
+    def mock_run_queue(*args, **kwargs):
+        raise exception_to_raise("Test exception")
+
+    with patch.object(RunTask, "run_queue", mock_run_queue), patch.object(
+        RunTask, "_cancel_connections"
+    ) as mock_cancel_connections:
+
+        # TODO clean up this after we have a proper runtime config fixture
```
ChenyuLInx, line 29: Adjust this before merge

QMalcolm: https://github.com/dbt-labs/dbt-core/pull/10242 has been merged, so we should be able to clean this up now 👍

QMalcolm (commented):
Looks good to me! Nice amount of LOC reduction too 👌 I think we can clean up `test_run_task_cancel_connections` a bit now since https://github.com/dbt-labs/dbt-core/pull/10242 has been merged which added a runtime config fixture.

QMalcolm (approved):
Looks good to me!

PR #10245, opened 2024-05-30
[Feature] add generic config for generic tests

### LINKED ISSUE #10197: [Impl] Allow generic tests to accept arbitrary configs
### Housekeeping

### Short description

Currently, we only support a specific set of named configurations for generic tests. That list is defined [here](https://github.com/dbt-labs/dbt-core/blob/09243d1e5d178c7e3051bc63d33816cafaa4c98b/core/dbt/parser/generic_test_builders.py#L83-L99) and accessed [here](https://github.com/dbt-labs/dbt-core/blob/09243d1e5d178c7e3051bc63d33816cafaa4c98b/core/dbt/parser/generic_test_builders.py#L135).

The context is that we support two syntaxes for passing configs into generic tests:
```yml
# recommended since dbt v0.20
models:
  - name: my_model
    columns:
      - name: colour
        tests:
          - accepted_values:
              values: ['blue', 'red']
              config:
                severity: warn
```
```yml
# legacy (supported for backwards compatibility)
models:
  - name: my_model
    columns:
      - name: colour
        tests:
          - accepted_values:
              values: ['blue', 'red']
              severity: warn                
```
In this example, `values` is an argument being passed into `accepted_values`, whereas `severity` is a configuration being set for the test.

In the legacy syntax, there is no distinction between "arguments" and "configurations." With the new recommended syntax, there is a distinction, because all configs are nested under `config`.

The desirable outcome is that:
- Users can pass whatever configs they want into `config:` (new syntax), e.g. `snowflake_warehouse`
- We continue to use the explicit list of named configs to provide backwards compatibility for the legacy syntax

### Acceptance criteria

1. Users can pass whatever configs they want into `config:` (new syntax), e.g. `snowflake_warehouse`
2. We continue to use the explicit list of named configs to provide backwards compatibility for the legacy syntax

To make it concrete: In the following example, `accepted_values_my_model_colour__blue__red` should end up with both configurations `severity: warn` and `custom_config: custom_value`.
```yaml
models:
  - name: my_model
    columns:
      - name: colour
        tests:
          - accepted_values:
              values: ['blue', 'red']
              severity: warn
              config:
                custom_config: custom_value
```

### Suggested Tests

In the example above, the test should be parsed to have both configs: `custom_config` via recommended syntax, and `severity` via old syntax.

It should be possible to write a unit test for this, but the logic is baked into `TestBuilder`, which is quite complicated. Refactoring that is out of scope for this particular issue.

The closest existing functional test case is [this one](https://github.com/dbt-labs/dbt-core/blob/09243d1e5d178c7e3051bc63d33816cafaa4c98b/tests/functional/schema_tests/test_schema_v2_tests.py#L418-L452).

As such, it might be best to create a new functional test instead.

### Impact to Other Teams

Adapters

no

### Context

Original issue:
- https://github.com/dbt-labs/dbt-core/issues/5532

McKnight-42:
resolves #10197 

### Problem
We would like to update our `TestBuilder` class for generic tests to be able to accept any config that the user can think of needing. 

examples.
```
models:
  - name: my_model
    columns:
      - name: colour
        tests:
          - accepted_values:
              values: ['blue', 'red']
              severity: warn
              config:
                custom_config: custom_value
```

### Solution

in the `Testbuilder` class we first setup a variable with a ll defined accepted configs we can keep this a little further down we  can add a if condition to process the `config` beyond the original of searching for the `CONFIG_ARGS` variable.

### Checklist


mikealfare (commented):
I like the approach of breaking out the logic into separate methods for readability. And I provided comments that align with that approach as you've implemented it.

However, I also think this could be done even better by dispatching to methods that correspond to _stage_ instead of _legacy/preferred_. There's a lot of duplication in the two methods, which suggests that pivoting how it's split could be useful.

To make sure I completely understand the context, here's how I understand the intent and constraints:

- there is a legacy method of specifying config at the root, these expected config keys are captured in `self.CONFIG_ARGS`
- there is a preferred method of specifying config which will show up as `self.args["config"]` (as a result of processing `data_test` via `self.extract_test_args()`)
- the same key should not show up in both methods, otherwise raise `SameKeyNestedError()`
- we need to render `str` entries, and fail on a missing macro
- we need to filter `None` entries

Assuming the above is correct, this could be done by translating the above steps into the below pseudo-code:
```python
def __init__(...)
    ...
    raw_config = self._raw_config()
    self.config = self._rendered_config(raw_config)
    ...

def _raw_config(self) -> Dict[str, Optional[str]]:
    legacy_config = ...
    config = ...
    # identify issue where key is present in both
    raw_config = ...
    return raw_config

def _rendered_config(
    self,
    raw_config: Dict[str, Optional[str]],
    render_ctx: Dict[str, Any],
    column_name: Optional[str]
) -> Dict[str, Any]:
    rendered_config = {}
    for key, value in raw_config.items():
        # try to render, raising an error if the macro is missing
        # filter for `None` here since you're already looping, and you could make the argument that removing `None` values is a form of rendering
    return rendered_config
```

core/dbt/parser/generic_test_builders.py
```diff
@@ -125,62 +125,71 @@ def __init__(
...
                         name=self.name,
                         key=key,
                         err_msg=e.msg,
                     )
-
             if value is not None:
                 self.config[key] = value
 
         if "config" in self.args:
+            self.process_config_args(self.args["config"], render_ctx, column_name)
```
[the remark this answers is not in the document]
McKnight-42: I've made changes in this line of code to match feedback (updated the two methods for private, calling directly not nested, and pulled out the render_values bit into its own method as well.

gshank (approved):

McKnight-42:
> I like the approach of breaking out the logic into separate methods for readability. And I provided comments that align with that approach as you've implemented it.
> 
> However, I also think this could be done even better by dispatching to methods that correspond to _stage_ instead of _legacy/preferred_. There's a lot of duplication in the two methods, which suggests that pivoting how it's split could be useful.
> 
> To make sure I completely understand the context, here's how I understand the intent and constraints:
> 
> * there is a legacy method of specifying config at the root, these expected config keys are captured in `self.CONFIG_ARGS`
> * there is a preferred method of specifying config which will show up as `self.args["config"]` (as a result of processing `data_test` via `self.extract_test_args()`)
> * the same key should not show up in both methods, otherwise raise `SameKeyNestedError()`
> * we need to render `str` entries, and fail on a missing macro
> * we need to filter `None` entries
> 
> Assuming the above is correct, this could be done by translating the above steps into the below pseudo-code:
> 
> ```python
> def __init__(...)
>     ...
>     raw_config = self._raw_config()
>     self.config = self._rendered_config(raw_config)
>     ...
> 
> def _raw_config(self) -> Dict[str, Optional[str]]:
>     legacy_config = ...
>     config = ...
>     # identify issue where key is present in both
>     raw_config = ...
>     return raw_config
> 
> def _rendered_config(
>     self,
>     raw_config: Dict[str, Optional[str]],
>     render_ctx: Dict[str, Any],
>     column_name: Optional[str]
> ) -> Dict[str, Any]:
>     rendered_config = {}
>     for key, value in raw_config.items():
>         # try to render, raising an error if the macro is missing
>         # filter for `None` here since you're already looping, and you could make the argument that removing `None` values is a form of rendering
>     return rendered_config
> ```

When I tried to combine the levels/ or hit its version of code more it would hit macro errors in dbt-adapters to augment to that style seems like it would require a higher lift to supply more configs to other existing macros.

core/dbt/parser/generic_test_builders.py
```diff
@@ -182,6 +157,45 @@ def __init__(
...
 
+    def _process_legacy_args(self):
+        config = {}
+        for key in self.CONFIG_ARGS:
+            value = self.args.pop(key, None)
+            if value and "config" in self.args and key in self.args["config"]:
+                raise SameKeyNestedError()
+            if not value and "config" in self.args:
+                value = self.args["config"].pop(key, None)
+            if value is not None:
```
mikealfare, line 168: This check is also happening in `self._render_values`. I think we can always update the dict with this key/value pair.

core/dbt/parser/generic_test_builders.py
```diff
@@ -182,6 +157,45 @@ def __init__(
...
+                config[key] = value
+
+        return self._render_values(config)
+
+    def _process_config_args(self, config_dict: Dict[str, Any]) -> Dict[str, Any]:
+        config = {}
+        for key, value in config_dict.items():
+            if value is not None:
```
mikealfare, line 176: Same as above, this check is happening in `self._render_values`. And since that's all this method is doing, you could drop it then, and update line 133-135 to:
```python
self.config.update(self._render_values(self.args.pop("config", {}))
```

mikealfare (approved):
I have some small nits that would make the code a bit more succinct, but nothing that's worth stopping from moving forward. If you have the time, take a look, otherwise merge and move on.

PR #10258, opened 2024-06-03
Call adapter.pre_model_hook + adapter.post_model_hook within tests

### LINKED ISSUE #10198: [Impl] Call adapter.pre_model_hook + adapter.post_model_hook within TestTask
### Housekeeping

### Short description

Before running models, dbt runs an adapter-specified "hook." Note that this is *different* from user-configured model hooks. In `dbt-snowflake`, this is the mechanism we use for calling `use warehouse {snowflake_warehouse}`, and it's what we should also use for custom secondary roles (and other things in the future).

Here's where that happens within RunTask:
https://github.com/dbt-labs/dbt-core/blob/09243d1e5d178c7e3051bc63d33816cafaa4c98b/core/dbt/task/run.py#L285-L291

We want the same thing to happen within `TestTask`, for both data + unit tests:
https://github.com/dbt-labs/dbt-core/blob/09243d1e5d178c7e3051bc63d33816cafaa4c98b/core/dbt/task/test.py#L143-L146
https://github.com/dbt-labs/dbt-core/blob/09243d1e5d178c7e3051bc63d33816cafaa4c98b/core/dbt/task/test.py#L216-L225

This way, unit + data tests can use custom Snowflake warehouses.

### Acceptance criteria

`adapter.pre_model_hook` + `adapter.post_model_hook` are being called within `TestTask` for unit + data tests.

### Suggested Tests

I don't think we're doing a great job of testing that `pre_model_hook` and `post_model_hook` are called today.

Ideas:
- Patch the adapter to define a `pre_model_hook` and `post_model_hook` that logs, and ensure it logs during test execution.
- After https://github.com/dbt-labs/dbt-core/issues/10197 is complete: Add to [the functional tests in `dbt-snowflake` here](https://github.com/dbt-labs/dbt-snowflake/blob/5da6fdbfd0895cf86c49dd2b95d89d3d2a57aa70/tests/functional/warehouse_test/test_warehouses.py#L8) here. A test configured with a `snowflake_warehouse` that does not exist should fail. Today, it doesn't, because it doesn't actually use this config.

### Impact to Other Teams

Impact: Adapters

no

### Context

Original issue:
- https://github.com/dbt-labs/dbt-adapters/issues/212

However, compilation is out of scope. This issue is focused _only_ on adding these hook calls to `TestTask`.

McKnight-42:
resolves #10198

### Problem
Allow tests to take in pre and post hooks similar to our `pre/post_model_hook` to take in configs like `snowflake_warehouse` and other arbitrary configs.

### Solution
Potential Solutions:
1. Update unit and data test execute methods to trigger either the existing `pre/post_model_hoook` and add logging to 
adapter definitions as described or in the area right after we add the hooks in via this pr.
2. Create a  new `pre/post_test_hook` in BaseAdapter in `dbt-adapters` then have the adapters create their own version of the method as we do for `pre_model_hook` (logging could possibly happen in method more easily this way) but would also mean many more pr's across adapters to add the new methods.

### Checklist


.changes/unreleased/Features-20240606-112334.yaml
```diff
@@ -0,0 +1,6 @@
+kind: Features
+body: add pre_model and post_model hoook calls to data and unit tests to be able to provide extra config options
```
mikealfare, line 2: spelling error

core/dbt/task/test.py
```diff
@@ -213,16 +220,18 @@ def execute_unit_test(
                 "Invalid materialization context generated, missing config: {}".format(context)
             )
 
-        # generate materialization macro
-        macro_func = MacroGenerator(materialization_macro, context)
-        # execute materialization macro
         try:
+            # generate materialization macro
+            macro_func = MacroGenerator(materialization_macro, context)
```
mikealfare, line 225: Is there concern that generating `macro_func` would throw an exception that we want to catch (hence moving it inside the `try`)? If not, I'd leave this where it is. We want to minimize what runs inside of a `try`.

core/dbt/task/test.py
```diff
@@ -140,10 +142,13 @@ def execute_data_test(self, data_test: TestNode, manifest: Manifest) -> TestResu
                 "Invalid materialization context generated, missing config: {}".format(context)
             )
 
-        # generate materialization macro
-        macro_func = MacroGenerator(materialization_macro, context)
-        # execute materialization macro
-        macro_func()
+        try:
+            # generate materialization macro
+            macro_func = MacroGenerator(materialization_macro, context)
```
mikealfare, line 147: Per my comment below, would it make sense to move this outside of the `try`?

core/dbt/task/test.py
McKnight-42: I don't know of any outright concern about the intilizer having it in the try more so because we only call it the one time each time so it seemed like it should stay as close together as possible, if others feel differently can try pulling it out.

tests/functional/unit_testing/test_ut_adapter_hooks.py
```diff
@@ -0,0 +1,50 @@
...
+
+        mock_pre_model_hook = mock.Mock()
+        with mock.patch.object(type(project.adapter), "pre_model_hook", mock_pre_model_hook):
+            results = run_dbt(["test", "--select", "test_name:test_my_model"], expect_pass=True)
+
+            assert len(results) == 1
+            mock_pre_model_hook.assert_called_once()
+
+
+class TestUnitTestAdapterHookFails(BaseUnitTestAdapterHook):
```
McKnight-42, line 39: Do we feel we also need to have tests for `post_model_hook`?

MichelleArk: I think it's worth it since they are different codepaths and could regress for different reasons.

mattmcknight13: Will update

core/dbt/task/test.py
```diff
@@ -126,6 +126,8 @@ def before_execute(self):
     def execute_data_test(self, data_test: TestNode, manifest: Manifest) -> TestResultData:
         context = generate_runtime_model_context(data_test, self.config, manifest)
 
+        hook_ctx = self.adapter.pre_model_hook(context)
```
MichelleArk, line 129: let's also test this codepath (hooks before/after data test execution) in addition to hooks before/after unit test execution. 

We don't already have a non-adapter `tests/functional/data_tests/` but let's create one! 

Here's some example setup for a singular data test that should run through this codepath: https://github.com/dbt-labs/dbt-core/blob/main/tests/functional/test_singular_tests.py

Beyond the setup, I believe the tests should look pretty similar to what we've got in test_ut_adapter_hooks.py so far

McKnight-42: added tests please let me know what you think! and thanks for all the suggestions @MichelleArk

core/dbt/task/test.py
McKnight-42: pulled it out

core/dbt/task/test.py
```diff
@@ -142,8 +144,12 @@ def execute_data_test(self, data_test: TestNode, manifest: Manifest) -> TestResu
 
         # generate materialization macro
         macro_func = MacroGenerator(materialization_macro, context)
-        # execute materialization macro
-        macro_func()
+        try:
+            # execute materialization macro
+            macro_func()
+        finally:
+            self.adapter.post_model_hook(context, hook_ctx)
```
MichelleArk, line 151: To confirm, this is not a behaviour change because: 
* pre/post_model_hook are no-ops by default: https://github.com/dbt-labs/dbt-adapters/blob/8ff69435a753f0a94dd1274d0e8dad199805cd13/dbt/adapters/base/impl.py#L1415-L1438
* only [dbt-snowflake has custom implementation](https://github.com/dbt-labs/dbt-snowflake/blob/main/dbt/adapters/snowflake/impl.py#L101), which is based on the presence of a config and effectively a no-op otherwise (uses warehouse from profile). Is the assumption that users with a `snowflake_warehouse` would _always_ want to route tests to that warehouse? 

cc @jtcohen6

McKnight-42: I think i can help supply a answer here. with this pr and 

- https://github.com/dbt-labs/dbt-core/pull/10245 will allow us to configure the snowflake_warehosue even for tests as its a config that can be changed in various ways through the config:

MichelleArk: ah! gotcha -- so configs were not possible to provide for generic tests in the past, which means there should not be any currently set on them

McKnight-42: from my understanding yes, we could use a old pattern to set a set number of accepted configs linked in the generic_config issue, making this update to the newer syntax while maintaining the old will allow us to do this now.

jtcohen6: Thanks @MichelleArk @McKnight-42 for thinking of this!

As I see it, the only way in which this could be a "breaking" change is if someone had configured a singular test, or all tests (from `dbt_project.yml`), with a `snowflake_warehouse` that they didn't actually want to be used.

Something like:
```
# dbt_project.yml
tests:
  +snowflake_warehouse: does_not_exist
```
```
-- tests/my_singular_test.sql
{{ config(snowflake_warehouse = 'wrong_warehouse') }}

select ...
```

I do not see that as a breaking change; I see it as this feature in dbt (specifically `dbt-snowflake`) finally working as intended. As such, I do not believe we need to put it behind a legacy behavior migration flag.

MichelleArk (approved):

PR #10260, opened 2024-06-04
update dbt_utils version to 1.2.0

aranke:
### Problem

Tests were failing because of dbt-utils version bump to 1.2.0

### Solution

Update test to use dbt-utils version 1.2.0

### Checklist


ChenyuLInx (approved):
This works for now. Lets see how often this type of package update change version being locked happens. If it happens again we should modify the test so it doesn't break when new dbt_utils gets released.

PR #10264, opened 2024-06-05
Fix issues with inline nodes and selectors

### LINKED ISSUE #8943: [CT-3290] [Bug] Using --inline should exclude selectors.yml defaults

### Current Behavior

If a default selector is defined in `selectors.yml`, that selector will still applied when the `--inline` flag is used with `show` or `compile`. 

[See slack thread for additional context](https://dbt-labs.slack.com/archives/C05FWBP9X1U/p1698689854497189)

### Expected Behavior

- When we do `dbt show --inline ...` we expect default selectors to be ignored
- make `--inline` and `--select` mutually exclusive for `dbt show` (do this at the click level)

### Steps To Reproduce

1. Add a default selector to a local dbt project's `selectors.yml` : 
```
selectors:
  - name: test_selector
    description: Exclude everything
    default: true
    definition:
       method: package
       value: "foo"
```
2. Using dbt-core 1.5+ (for support for --inline), run `dbt compile --inline 'select 1'`
3. Observe that inline compile does not execute

### Relevant log output

```shell
(dbt-core-1.6-local) racheldaniel:jaffle_shop/ (master✗) $ dbt compile --inline 'select 1'                                                                                                       [14:22:20]
19:22:26  Running with dbt=1.6.6
19:22:26  Registered adapter: postgres=1.6.6
19:22:26  Found 10 models, 3 seeds, 1 operation, 18 tests, 0 sources, 0 exposures, 0 metrics, 661 macros, 1 group, 0 semantic models
19:22:26  Using default selector test_selector
19:22:26
19:22:26  Nothing to do. Try checking your model configs and model specification args
```

### Environment

```markdown
- OS: MacOS Ventura 13.5.2
- Python: 3.8.10
- dbt: 1.6.6
```

postgres

### Additional Context

See https://github.com/dbt-labs/dbt-core/issues/9269#issuecomment-2125718609 for several specific cases where `--inline` doesn't work as desired.

### LINKED ISSUE #9269: [CT-3487] [Bug] `dbt show --inline` does not work when all models are disabled in `dbt_project.yml`

### Current Behavior

When I run `dbt show --inline` it works. When I add a `--target` flag it does not work.

### Expected Behavior

`dbt show` works when a target is supplied.

### Steps To Reproduce

I'm running the latest version of `dbt-biguery` with this `profiles.yml`:
```yml
dbt_docs:
    target: dev
    outputs:
        dev:
            type: bigquery
            method: oauth
            project: <PROJECT_ID>
            schema: <ENV_VAR>
            location: europe-west4
            threads: 4
        dev2:
            type: bigquery
            method: oauth
            project: <PROJECT_ID>
            schema: <ENV_VAR>
            location: europe-west4
            threads: 4
```

Note that `dev2` is an exact copy of `dev`.

Run the following commands and observe differing behaviour:
```shell
$ dbt show --inline "SELECT 1 AS id"
$ dbt show --inline "SELECT 1 AS id" --target dev
$ dbt show --inline "SELECT 1 AS id" --target dev2
```

The first 2 commands return a result, the third command does not.

### Relevant log output

```shell
$ dbt show --inline "SELECT 1 AS id"
15:13:44  Running with dbt=1.7.3
15:13:44  Registered adapter: bigquery=1.7.2
15:13:44  Unable to do partial parsing because of a version mismatch
15:13:46  [WARNING]: Configuration paths exist in your dbt_project.yml file which do not apply to any resources.
There are 1 unused configuration paths:
- tests
15:13:47  Found 1 model, 1 seed, 0 sources, 0 exposures, 0 metrics, 949 macros, 0 groups, 0 semantic models
15:13:47  
15:13:48  Concurrency: 4 threads (target='dev')
15:13:48  
15:13:49  Previewing inline node:
| id |
| -- |
|  1 |

$ dbt show --inline "SELECT 1 AS id" --target dev
15:13:54  Running with dbt=1.7.3
15:13:55  Registered adapter: bigquery=1.7.2
15:13:55  Unable to do partial parsing because config vars, config profile, or config target have changed
15:13:57  [WARNING]: Configuration paths exist in your dbt_project.yml file which do not apply to any resources.
There are 1 unused configuration paths:
- tests
15:13:57  Found 1 model, 1 seed, 0 sources, 0 exposures, 0 metrics, 949 macros, 0 groups, 0 semantic models
15:13:57  
15:13:58  Concurrency: 4 threads (target='dev')
15:13:58  
15:13:59  Previewing inline node:
| id |
| -- |
|  1 |

$ dbt show --inline "SELECT 1 AS id" --target dev2
15:14:03  Running with dbt=1.7.3
15:14:03  Registered adapter: bigquery=1.7.2
15:14:03  Unable to do partial parsing because config vars, config profile, or config target have changed
15:14:06  Found 1 seed, 0 sources, 0 exposures, 0 metrics, 949 macros, 0 groups, 0 semantic models
15:14:06  
15:14:06  Nothing to do. Try checking your model configs and model specification args
```

### Environment

```markdown
- OS:Ubuntu 22.04
- Python: Python 3.10.12
- dbt:

Core:
  - installed: 1.7.3
  - latest:    1.7.3 - Up to date!

Plugins:
  - bigquery: 1.7.2 - Up to date!
```

bigquery

### Additional Context

_No response_

gshank:
resolves #8943
resolves #9269

### Problem

The inline node wasn't executing because of interactions with selection mechanisms.

### Solution

We shouldn't use selectors at all when inline nodes are running. Add code in "get_selector" to make an exception for inline nodes. Add error message for using --select or --selector with --inline. Add hack to move inline node out of disabled if someone has disabled all model nodes in their project.

### Checklist

### DIFF
--- a/core/dbt/cli/flags.py
+++ b/core/dbt/cli/flags.py
@@ -289,6 +289,9 @@ class Flags:
             params_assigned_from_default, ["WARN_ERROR", "WARN_ERROR_OPTIONS"]
         )
 
+        if hasattr(self, "SELECT") and hasattr(self, "INLINE"):
+            self._assert_mutually_exclusive(params_assigned_from_default, ["SELECT", "INLINE"])
+
         # Support lower cased access for legacy code.
         params = set(
             x for x in dir(self) if not callable(getattr(self, x)) and not x.startswith("__")
--- a/core/dbt/cli/flags.py
+++ b/core/dbt/cli/flags.py
@@ -319,7 +319,9 @@ class Flags:
         """
         set_flag = None
         for flag in group:
-            flag_set_by_user = hasattr(self, flag) and flag.lower() not in params_assigned_from_default
+            flag_set_by_user = (
+                hasattr(self, flag) and flag.lower() not in params_assigned_from_default
+            )
             if flag_set_by_user and set_flag:
                 raise DbtUsageException(
                     f"{flag.lower()}: not allowed with argument {set_flag.lower()}"
--- a/core/dbt/cli/flags.py
+++ b/core/dbt/cli/flags.py
@@ -289,8 +289,14 @@ class Flags:
             params_assigned_from_default, ["WARN_ERROR", "WARN_ERROR_OPTIONS"]
         )
 
-        if hasattr(self, "SELECT") and hasattr(self, "INLINE"):
-            self._assert_mutually_exclusive(params_assigned_from_default, ["SELECT", "INLINE"])
+        # Handle arguments mutually exclusive with INLINE
+        if hasattr(self, "INLINE"):
+            if hasattr(self, "SELECT"):
+                self._assert_mutually_exclusive(params_assigned_from_default, ["SELECT", "INLINE"])
+            if hasattr(self, "SELECTOR") and self.SELECTOR is not None:
+                self._assert_mutually_exclusive(
+                    params_assigned_from_default, ["SELECTOR", "INLINE"]
+                )
 
         # Support lower cased access for legacy code.
         params = set(
--- a/core/dbt/cli/flags.py
+++ b/core/dbt/cli/flags.py
@@ -290,13 +290,8 @@ class Flags:
         )
 
         # Handle arguments mutually exclusive with INLINE
-        if hasattr(self, "INLINE"):
-            if hasattr(self, "SELECT"):
-                self._assert_mutually_exclusive(params_assigned_from_default, ["SELECT", "INLINE"])
-            if hasattr(self, "SELECTOR") and self.SELECTOR is not None:
-                self._assert_mutually_exclusive(
-                    params_assigned_from_default, ["SELECTOR", "INLINE"]
-                )
+        self._assert_mutually_exclusive(params_assigned_from_default, ["SELECT", "INLINE"])
+        self._assert_mutually_exclusive(params_assigned_from_default, ["SELECTOR", "INLINE"])
 
         # Support lower cased access for legacy code.
         params = set(
@@ -324,7 +319,7 @@ class Flags:
         """
         set_flag = None
         for flag in group:
-            flag_set_by_user = flag.lower() not in params_assigned_from_default
+            flag_set_by_user = hasattr(self, flag) and flag.lower() not in params_assigned_from_default
             if flag_set_by_user and set_flag:
                 raise DbtUsageException(
                     f"{flag.lower()}: not allowed with argument {set_flag.lower()}"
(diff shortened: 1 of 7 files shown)


core/dbt/task/compile.py
```diff
@@ -104,6 +104,12 @@ def _runtime_initialize(self):
                 )
                 sql_node = block_parser.parse_remote(self.args.inline, "inline_query")
                 process_node(self.config, self.manifest, sql_node)
+                # Special hack to remove disabled, if it's there. This would only happen
+                # if all models are disabled in dbt_project
```
MichelleArk, line 108: So if _all_ models are disabled, we effectively re-enable all of them? I think I'm missing something here as to why this would be safe to do

core/dbt/cli/flags.py
```diff
@@ -289,6 +289,15 @@ def _assign_params(
             params_assigned_from_default, ["WARN_ERROR", "WARN_ERROR_OPTIONS"]
         )
 
+        # Handle arguments mutually exclusive with INLINE
+        if hasattr(self, "INLINE"):
+            if hasattr(self, "SELECT"):
```
MichelleArk, line 294: hmm, presumably we didn't need to do this handling for WARN_ERROR or WARN_ERROR_OPTIONS because they are global flags. 

What do you think about baking it into `_assert_mutually_exclusive`? 

e.g. 
```python
def _assert_mutually_exclusive(
        self, params_assigned_from_default: Set[str], group: List[str]
    ) -> None:
        """
        Ensure no elements from group are simultaneously provided by a user, as inferred from params_assigned_from_default.
        Raises click.UsageError if any two elements from group are simultaneously provided by a user.
        """
        set_flag = None
        for flag in group:
            flag_set_by_user = hasattr(self, flag.lower()) and flag.lower() not in params_assigned_from_default
...
```

and simplifying these lines to: 
```python
self._assert_mutually_exclusive(params_assigned_from_default, ["SELECT", "INLINE"])
self._assert_mutually_exclusive(params_assigned_from_default, ["SELECTOR", "INLINE"])
```
(I don't think the `self.SELECTOR is not None` is necessary since default handling already occurs in _assert_mutually_exclusive)

core/dbt/task/compile.py
gshank: The only way that an inline node would be disabled is if somebody has all models to enabled: false (like in the test case) because there's no place else for enabled: false to come from. This just enables and moves the temporary inline node that was just created. It doesn't affect anything else in the manifest.

core/dbt/cli/flags.py
gshank: Yes, that's better. Have changed.

MichelleArk (approved):

PR #10282, opened 2024-06-08
Migrate selector tests to pytest

### LINKED ISSUE #9868: [SPIKE+] Write a unittest for a simple selection
Creating this one as a Spike + because there is not a single unit test for selection, and selection is very involved.
By trying to write out one unit test for a simple selection, we aim to get a sense of
- what fixture do we need to properly start writing unit tests for selection
- how much effort it takes to refactor selection out of task into it's own module. So task can start with a selected set of nodes to run.

QMalcolm:
resolves #9868 

### Problem

In #9868 we wanted to ensure we write, and can write, some selector tests. After creating the ticket, we actually found some selector tests while we were reorganizing tests. Thus #9868 became about converting these tests to use pytest.

### Solution

Converted selector tests to use pytest, showing that we can write selector tests

### Checklist


tests/unit/graph/test_selector.py
```diff
@@ -82,7 +68,7 @@ def graph():
 
 
 @pytest.fixture
-def manifest(graph):
+def mock_manifest(graph):
     return _get_manifest(graph)
```
ChenyuLInx, line 72: Should we be more specific here that we are only mocking the graph functionality part?

tests/unit/graph/test_selector.py
```diff
@@ -231,350 +217,82 @@ def test_invalid_specs(invalid):
...
-    def test__two_models_package_ref(self):
-        self.use_models(
-            {
-                "model_one": "select * from events",
-                "model_two": "select * from {{ref('test_models_compile', 'model_one')}}",
-            }
-        )
+        expected_nodes = [model.unique_id for model in models]
+        assert linker.nodes() == set(expected_nodes)
+        assert list(linker.edges()) == [tuple(expected_nodes)]
```
ChenyuLInx, line 247: Nitty Nit: It is better to just write out the actual values in the test if it is not too complex. This way reader can understand this edges will yield a list of string without knowing what `model.unique_id`(which is the implementation detail of `model`) is.

tests/unit/utils/manifest.py
```diff
@@ -963,25 +968,58 @@ def macros(
 
 
 @pytest.fixture
-def unit_tests(unit_test_table_model) -> list:
+def unit_tests(unit_test_table_model) -> List[UnitTestDefinition]:
```
ChenyuLInx, line 971: Thanks for all the type annotation!

ChenyuLInx (approved):
Have two optional nits otherwise looks good to me!

tests/unit/graph/test_selector.py
QMalcolm: Yea I can change the name to something like `manifest_with_mocked_graph`.

tests/unit/graph/test_selector.py
QMalcolm: Would a compromise of adding a type hint be sufficient? For example:
```python
expected_nodes: List[str] = [model.unique_id for model in models]
```

My issue with hand writing it out is that I find that it doesn't scale well and makes things more brittle. From a perspective of making it easier to write tests, I think this approach will suit us better.

ChenyuLInx: I should have added this point in the last comment.
So in some way, we are writing out some form of implementation in the tests(it is fairly small in this case). IMHO unit tests should be defining input and output explicitly without (or as little as possible) dependencies between them.
I don't think writing out an explicit value makes writing tests harder.

This is not a very strongly held opinion and I am happy that you make the final call here. We can always adjust later on if we find out more.

PR #5488, opened 2022-07-18
Feature CT-665 Allow control over the outer CTE identifier generated in ephemeral materialization

miro-ur:
resolves CT-665 #5273 

### Description

Add support for ephemeral models using a dot in file naming convention.
Prioritize alias if specified when creating node identifier and sanitize the identifier to make sure no illegal characters are used when creating CTEs that would cause a runtime error.

### Checklist

- [x] I have read [the contributing guide](https://github.com/dbt-labs/dbt-core/blob/main/CONTRIBUTING.md) and understand what's expected of me
- [ ] I have signed the [CLA](https://docs.getdbt.com/docs/contributor-license-agreements)
- [ ] I have run this code in development and it appears to resolve the stated issue
- [x] This PR includes tests, or tests are not required/relevant for this PR
- [ ] I have [opened an issue to add/update docs](https://github.com/dbt-labs/docs.getdbt.com/issues/new/choose), or docs changes are not required/relevant for this PR
- [ ] I have run `changie new` to [create a changelog entry](https://github.com/dbt-labs/dbt-core/blob/main/CONTRIBUTING.md#Adding-CHANGELOG-Entry)

PR #10290, opened 2024-06-10
Use model alias for the CTE identifier generated during ephemeral materialization

### LINKED ISSUE #5273: [CT-665] [Feature] Allow control over the outer CTE identifier generated in ephemeral materialization

### Describe the Feature

We use a dot notation to group and organize dbt model files with a naming pattern of `{group}.{model}` such as `ingest.source` and `dw.dimension` etc. With specially complex models this provides a level of organization and consistency that keeps files organized, unique and faster to identify and find.

We rely on config aliases to then properly name relations - otherwise adding schema would be an invalid fully qualified relation name like `schema.layer.table`.  This works for both **table** and **view** materialization, however causes the generated sql to be invalid when using the **ephemeral** materialization due to the outer CTE identifier being generated as `__dbt__cte__ingest.source`.

If model `alias` is configured it should be used everywhere as the identifier of the output instead of the file name - regardless of materialization used to give end user better control over generated SQL code and avoid errors or potential invalid code being generated when using an _unconventional_ naming of model files.

An alternative might be some form of sanitization of CTE name during generation to avoid generating invalid CTE identifier - like stripping non alphanumeric characters or generating a non model name linked name - like an incremental `__dbt__cte__1` - as long as its a valid and unique CTE name to enable stacking CTEs.

### Who will this benefit?

This feature would introduce consistency in behavior compared to other materialization strategies (alias being used as table or view name), give uses control over naming of models vs files and allow end users to come up with creative naming convention and organization of model files without having to worry that switching materialization will cause a failure because of the model file name.

Happy to contribute, looking for feedback and some pointers on this from the team

### Anything else?

_No response_

jeancochrane:
Resolves #5273. Reimplementation of https://github.com/dbt-labs/dbt-core/pull/5488 to match the new split between adapter and core repos.

See https://github.com/dbt-labs/dbt-adapters/pull/236 for the corresponding adapter PR.

### Problem

Unit tests and ephemeral materialization do not currently support referencing models that use special characters like dots in file names, even if these special characters are sanitized using model aliases.

### Solution

Prioritize using the model alias (if specified) via `model.identifier` when creating CTE identifiers for unit tests and ephemeral materialization.

This change requires a corresponding change to `dbt-adapters` to fully fix the bug, which is implemented in this PR: https://github.com/dbt-labs/dbt-adapters/pull/236 This second PR is necessary because the adapter is used to construct the CTE identifier at the time of CTE reference, while the core code is used to construct the CTE identifier at the time of CTE definition.

### Checklist

### DIFF
--- a/core/dbt/compilation.py
+++ b/core/dbt/compilation.py
@@ -294,10 +294,10 @@ class Compiler:
 
         return context
 
-    def add_ephemeral_prefix(self, alias: str):
+    def add_ephemeral_prefix(self, name: str):
         adapter = get_adapter(self.config)
         relation_cls = adapter.Relation
-        return relation_cls.add_ephemeral_prefix(alias)
+        return relation_cls.add_ephemeral_prefix(name)
 
     def _recursively_prepend_ctes(
         self,
@@ -371,7 +371,7 @@ class Compiler:
 
             _extend_prepended_ctes(prepended_ctes, new_prepended_ctes)
 
-            new_cte_name = self.add_ephemeral_prefix(cte_model.alias)
+            new_cte_name = self.add_ephemeral_prefix(cte_model.identifier)
             rendered_sql = cte_model._pre_injected_sql or cte_model.compiled_code
             sql = f" {new_cte_name} as (\n{rendered_sql}\n)"
 
--- a/core/dbt/compilation.py
+++ b/core/dbt/compilation.py
@@ -294,10 +294,10 @@ class Compiler:
 
         return context
 
-    def add_ephemeral_prefix(self, name: str):
+    def add_ephemeral_prefix(self, alias: str):
         adapter = get_adapter(self.config)
         relation_cls = adapter.Relation
-        return relation_cls.add_ephemeral_prefix(name)
+        return relation_cls.add_ephemeral_prefix(alias)
 
     def _recursively_prepend_ctes(
         self,
@@ -371,7 +371,7 @@ class Compiler:
 
             _extend_prepended_ctes(prepended_ctes, new_prepended_ctes)
 
-            new_cte_name = self.add_ephemeral_prefix(cte_model.name)
+            new_cte_name = self.add_ephemeral_prefix(cte_model.alias)
             rendered_sql = cte_model._pre_injected_sql or cte_model.compiled_code
             sql = f" {new_cte_name} as (\n{rendered_sql}\n)"
 
--- a/tests/functional/schema/test_custom_schema.py
+++ b/tests/functional/schema/test_custom_schema.py
@@ -17,6 +17,7 @@ from tests.functional.schema.fixtures.sql import (
     _TABLE_TWO,
     _TABLE_TWO_DOT_MODEL,
     _TABLE_TWO_DOT_MODEL_NAME,
+    _TABLE_TWO_DOT_MODEL_SCHEMA,
     _TABLE_TWO_SCHEMA,
     _VALIDATION_SQL,
 )
@@ -201,14 +202,17 @@ class TestCustomSchemaWithCustomMacroFromModelName(BaseTestCustomSchema):
         project.run_sql(_VALIDATION_SQL)
         run_dbt(["seed"])
         results = run_dbt(["run"])
-        # Model 2 is ephemeral, so it should not show up in the results list
-        assert len(results) == 2
+        assert len(results) == 3
         table_results = {r.node.name: r.node.schema for r in results.results}
 
         assert table_results[_TABLE_ONE_DOT_MODEL_NAME] == _TABLE_ONE_DOT_MODEL_SCHEMA
+        assert table_results[_TABLE_TWO_DOT_MODEL_NAME] == _TABLE_TWO_DOT_MODEL_SCHEMA
         assert table_results["table_3"] == f"{project.test_schema}"
         check_relations_equal(
             adapter=project.adapter, relation_names=("seed", _TABLE_ONE_DOT_MODEL_NAME)
+        )
+        check_relations_equal(
+            adapter=project.adapter, relation_names=("seed", _TABLE_TWO_DOT_MODEL_NAME)
         )
         check_relations_equal(
             adapter=project.adapter, relation_names=("agg", f"{project.test_schema}.table_3")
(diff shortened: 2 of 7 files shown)


core/dbt/compilation.py
```diff
@@ -371,7 +371,7 @@ def _recursively_prepend_ctes(
 
             _extend_prepended_ctes(prepended_ctes, new_prepended_ctes)
 
-            new_cte_name = self.add_ephemeral_prefix(cte_model.name)
+            new_cte_name = self.add_ephemeral_prefix(cte_model.alias)
```
jeancochrane, line 374: It strikes me as a bit fragile that this line of code and the `BaseRelation.add_ephemeral_prefix()` implementation in `dbt-adapters` are both responsible for generating the CTE identifier for ephemeral models; ideally this bugfix could have been made in one PR against one of either `dbt-core` or `dbt-adapters` without requiring PRs against both repos. I wonder if perhaps there's a way to transfer responsibility over this call to the adapter? Happy to make that change if you can think of a reasonable approach.

tests/functional/schema/test_custom_schema.py
```diff
@@ -202,18 +201,15 @@ def test__postgres__custom_schema_from_model_name(
         project.run_sql(_VALIDATION_SQL)
         run_dbt(["seed"])
         results = run_dbt(["run"])
-        assert len(results) == 3
+        # Model 2 is ephemeral, so it should not show up in the results list
+        assert len(results) == 2
```
jeancochrane, line 205: I hope this is the right place to test support for ephemeral models with dots in their names, but let me know if there's a better option!

jeancochrane:
@MichelleArk @QMalcolm @gshank Sorry for the wide tag radius, I'm not sure who is best equipped to help here! This should be ready for review, but `cla-bot` seems convinced that I haven't signed the CLA even though I have. (I may have done so incorrectly, but I'm not sure how to confirm 😅 ) Any idea how I can move forward here?

QMalcolm:
Howdy @jeancochrane!

Sorry for the noise from the CLA bot 🙃 I've looked at our CLA database and the issue that we filter out any submitted usernames that begin with `@`. Unfortunately, I don't have edit access to the underlying database, just view access. Could you submit your CLA again but without the `@` prefix on your username?

jeancochrane:
Done, thanks for the tip @QMalcolm!

jeancochrane:
Thank you for summoning the bot @QMalcolm! If you could also summon it on https://github.com/dbt-labs/dbt-adapters/pull/236 I would be very grateful 🙏🏻

dbeatty10:
@jeancochrane with the latest version of https://github.com/dbt-labs/dbt-adapters/pull/236, do we need this PR in dbt-core anymore?

jeancochrane:
> @jeancochrane with the latest version of https://github.com/dbt-labs/dbt-adapters/pull/236, do we need this PR in dbt-core anymore?

Unfortunately I believe it's still needed in order to fix unit testing in our case @dbeatty10! While https://github.com/dbt-labs/dbt-adapters/pull/236 successfully gets ephemeral materializations working for projects that use aliases to strip out special characters, unit tests appear to use `Compiler._recursively_prepend_ctes()` to construct the CTE identifiers that represent input data, and `_recursively_prepend_ctes()` uses `model.name` when constructing these CTE identifiers:

https://github.com/dbt-labs/dbt-core/blob/4c7d922a6da0d4c63980115f0fc46296b80e191d/core/dbt/compilation.py#L374

In practice, this means that with the current main branch of `dbt-core` installed alongside https://github.com/dbt-labs/dbt-adapters/pull/236, a unit test query can end up looking something like this (adapted from an example in our project that uses dots to namespace model filenames) -- note how the definition of the CTE at the top does not use the alias `final_model_raw`, while the reference at the bottom does:

```sql
with __dbt__cte__model.final_model_raw as (
    -- Fixture for model.final_model_raw
    select try_cast(null as varchar) as year, ...
)
select
    ...
from __dbt__cte__final_model_raw AS fmr
```

So I think that the current PR title "Use model alias for the CTE identifier generated during ephemeral materialization" is actually misleading; this particular PR is in fact fixing the CTE identifier generated during unit testing, not during ephemeral materialization. I'll go ahead and change that!

Taking a step back, however, it strikes me as a potential code smell that the CTE identifier is generated one way during definition (`with {identifier} as (...)`) and another way during reference (`... from {identifier} as fmr`). Perhaps that's an intentional design decision, or perhaps it's a sign that a better fix than this one would be to consolidate the code that generates CTE identifiers into one place. I leave that decision up to you, however, since I don't have as much insight into this code as you do! Just let me know what direction you think is appropriate to go and I'd be happy to take a stab at it.

jeancochrane:
Actually @dbeatty10, my previous comment was incorrect -- the discrepancy between the CTE identifier used during definition vs. reference that I described above _does_ in fact affect ephemeral materialization as well, not just unit tests. I should have tested more thoroughly prior to commenting! I've reverted the title of this PR and will edit my previous comment to reflect this mistake.

dbeatty10:
Closing and re-opening to try to restart CI tests.

dbeatty10:
> the discrepancy between the CTE identifier used during definition vs. reference that I described above _does_ in fact affect ephemeral materialization as well

Okay, makes sense @jeancochrane !

Could you do a couple things?

1. Update from `alias` to `identifier` as-needed to pass CI
2. Instead of modifying an existing test in `tests/functional/schema/fixtures/sql.py` and tests/functional/schema/test_custom_schema.py, could you add a new one?

ChenyuLInx (approved):
Thanks a lot for contributing @jeancochrane !
This change looks good, I think we need to merge the one in dbt-adapters, release a patch version, bump requirements here then merge.

jeancochrane:
Sounds great, thanks so much @ChenyuLInx! Let me know if you'd like a hand with any of that, otherwise I'll hang tight.

jeancochrane:
@dbeatty10 @ChenyuLInx Seems like these changes didn't get added to the changelog for [v1.8.6](https://github.com/dbt-labs/dbt-core/releases/tag/v1.8.6), any idea why not? I'm assuming I just didn't format the changelog metadata correctly, I'll try to do that right next time!

jeancochrane:
Makes sense, thanks for all your help getting this over the line @dbeatty10!

PR #10291, opened 2024-06-10
create a happy path project fixture

### LINKED ISSUE #10256: Create "happy path" project fixture and use it for `tests/functional/list/test_list.py`
### Housekeeping

### Short description
When we write functional tests we generally end up writing more fixtures. A lot of these fixtures we write are either duplicates of fixtures for other functional tests that already exist (or are very close copies). Additionally because we don't have a standard "happy path" fixture, it has happened a few times that test which checks specific functionality fails to detect that it is broken when a certain type of node is involved because it wasn't included in the project fixture. By creating a happy path project fixture we solve both issues. Tests should try to use existing nodes in the happy path fixture, and when they cannot, add a new node to it (so all tests get it). Additionally because the happy path fixture will be broad, it should help guarantee node types existing don't break certain functionalities.

### Acceptance criteria
- There exists a `happy_path_project` fixture for our functional tests
- We use it for tests in `tests/functional/list/test_list.py`

ChenyuLInx:
resolves #10256

### Problem
We are missing a central place to define a happy path project that contains a bit of everything to be reused across different tests.

### Solution
We are adding a `happy_path_project` fixture that will give integration tests a project to use
- I had to refactor the project construction a little bit to make using happy path project a bit more obvious and intuiative
- I choose to use normal file folder instead of using string defined in code to improve local testing experience and readability.


core/dbt/tests/fixtures/project.py
```diff
@@ -282,8 +282,8 @@ def adapter(
     project_root,
     profiles_root,
     profiles_yml,
-    dbt_project_yml,
     clean_up_logging,
+    dbt_project_yml,
```
QMalcolm, line 286: At first I was confused about why this fixture argument moved. At first I thought order of operations was important for some reason here. Looking through the commits though, I think what happened is that it got removed in one commit, and then re-added in another.

tests/functional/list/test_list.py
```diff
@@ -1,42 +1,32 @@
...
+    #         "config-version": 2,
+    #         "analysis-paths": [self.dir("analyses")],
+    #         "snapshot-paths": [self.dir("snapshots")],
+    #         "macro-paths": [self.dir("macros")],
+    #         "seed-paths": [self.dir("seeds")],
+    #         "test-paths": [self.dir("tests")],
+    #         "seeds": {
+    #             "quote_columns": False,
+    #         },
+    #     }
```
QMalcolm: Are we keeping this around for a specific reason, or can it be deleted?

tests/functional/list/test_list.py
```diff
@@ -56,7 +42,7 @@ def expect_given_output(self, args, expectations):
                 else:
                     assert got == expected
 
-    def expect_snapshot_output(self, project):
+    def expect_snapshot_output(self, happy_path_project):  # noqa: F811
```
QMalcolm, line 45: In this test we're actually using the `happy_path_project` fixture argument in the function. Is the `# noqa: F811` necessary for another reason?

tests/functional/list/test_list.py
```diff
@@ -724,12 +710,12 @@ def expect_resource_type_env_var(self):
         }
         del os.environ["DBT_EXCLUDE_RESOURCE_TYPES"]
 
-    def expect_selected_keys(self, project):
+    def expect_selected_keys(self, happy_path_project):  # noqa: F811
```
QMalcolm, line 713: In this test we're actually using the happy_path_project fixture argument in the function. Is the # noqa: F811 necessary for another reason?

core/dbt/tests/fixtures/project.py
```diff
@@ -598,7 +599,7 @@ def project_setup(
 # to pull in the other fixtures individually to access their information.
 @pytest.fixture(scope="class")
 def project(
-    project_files,
     project_setup: TestProjInfo,
+    project_files,
```
QMalcolm: I'm not asking for a change here, but I would like to understand the context. Does the ordering of the fixture arguments here mean something? Is there a reason the `project_setup` fixture has to be come before `project_files`? This is the kind of thing I would love to see in a commit comment so that I can better understand why something is changing while reviewing.

tests/functional/test_selection/fixtures.py
```diff
@@ -84,14 +82,3 @@ def models():
...
-
-@pytest.fixture(scope="class")
-def project_files(
-    project_root,
-    tests,
-    models,
-    selectors_yml,
-):
-    write_project_files(project_root, "tests", tests)
-    write_project_files(project_root, "models", models)
```
QMalcolm: Were we able to remove this because of the existing more widely available `project_files` satisfied the relevant tests? I'm trying to understand if this is extraneous cleanup or if it was necessary for some reason in regards to the happy path testing we are setting up in this branch.

QMalcolm (approved):
Looks good to me! Thank you for doing this work 🙂 

I do have some questions in regards to how things changed. They aren't blocking questions, but I'd like to discuss them so I can better understand how the code is functioning. Most of these occur in the `fix` commits. It'd be amazing to have some of that context in the commit messages themselves.

tests/functional/fixtures/happy_path_project/macros/macro_stuff.sql
```diff
@@ -0,0 +1,8 @@
+
```
MichelleArk, line 1: super nit: looks like a lot of these files have an extra empty line at the top of the file, could we clean that up?

tests/functional/list/test_list.py
MichelleArk: +1, also curious about this

ChenyuLInx: I think for whatever reason flake8 was complaining about all of these in my local dev environment.
Let me try again and see what happened here

core/dbt/tests/fixtures/project.py
ChenyuLInx: Yes, adding a comment in the following commit.
The order of the argument determines the order of fixtures being executed(if they need to).
So we do all setups, then write the project files. Setup should be independent of project file write.

Another reason is for the happy path fixture, we do not use the `dbt_project.yml` fixture, instead, we want to use the files in the fixture folder, changing the order makes sure that the tests run with the intended files

tests/functional/test_selection/fixtures.py
ChenyuLInx: Yes, they are covered by the project_files.
And the reason they are being removed is they failed to write certain files(like dbt_project.yml), they might not be relevant anymore with the order of writing project files but I think it is still good to remove.

MichelleArk (approved):
Not necessarily blocking but I would like to understand the necessity for the new `# noqa: F811` usages with `happy_path_project` since we'll probably start seeing those all over the place once we migrate more tests to using `happy_path_project`.

tests/functional/list/test_list.py
ChenyuLInx: This is the problem with pytest, I don't see a easy way of getting rid of it.
https://stackoverflow.com/questions/50268284/why-im-getting-f811-error-in-this-code

PR #10295, opened 2024-06-11
Implement state:modified for saved queries

### LINKED ISSUE #10294: saved queries are not included in the "modified" selector
A community member is reporting [the exact opposite behavior](https://getdbt.slack.com/archives/C02CCBBBR1D/p1717530937516939) - saved queries do not show up in state:modified or state:new even when they should. Here's the repro step sequence they provided in the linked Slack thread:

```
Steps used for testing:

1. Upgrade existing projects to dbt v1.8
2. Run dbt parse for project containing semantic models
3. Rename target to old-target
4. Edit semantic model (edit existing saved query, add new saved query)
5. Make edits to SQL model as a control for the test
6. Run dbt parse again
7. Run dbt ls -s "state:modified state:new" --target old-target

Result: no semantic model changes detected, SQL model change detected as expected. 
```

_Originally posted by @tlento in https://github.com/dbt-labs/dbt-core/issues/10138#issuecomment-2155685545_

gshank:
resolves #10294

### Problem

Doing state:modified for saved queries didn't work.

### Solution

Implement state:modified for saved queries.

### Checklist


QMalcolm (approved):
This looks great! Thank you for fixing it 🙂

ChenyuLInx (approved):
LGTM!
Two things that are not blocker:
- is it possible to unit test this change?
- should we consider move this `select modified` test to the test for list command and use the new [`happy path fixture`](https://github.com/dbt-labs/dbt-core/pull/10291) to test it? You can create a follow up ticket and give it to me.

Oh and another thing is if we want to list modified `semantic model` and metrics, do we need to do similar change? CC @aliceliu  since she is working on something related

gshank:
I think that semantic_models and metrics are already listable. What is Alice working on?

gshank:
@ChenyuLInx I think I'll make a separate ticket for unit tests and happy path tests, since this probably needs to be backported to 1.8.

PR #10337, opened 2024-06-18
add a check to make sure unit tests are considered when changing code

### LINKED ISSUE #10329: CI check whether a unit test is introduced while code is being modified
This should be able to skip with a tag.
The goal is to keep track of things that are hard to test so we can know what to improve next

ChenyuLInx:
resolves #10329

For new code checked in, we require 75% unittest coverage. We can always adjust if it becomes too noisy.
Docs about component https://about.codecov.io/blog/codecov-components-breaking-down-coverage-by-filters/


.github/workflows/unittest-existence.yml
ChenyuLInx, line 1: I think we should refactor https://github.com/dbt-labs/actions/blob/main/.github/workflows/changelog-existence.yml#L2 a bit to make it more generalized and use it here.
@emmyoop  WDYT?

ChenyuLInx: https://github.com/dbt-labs/actions/pull/210

.github/workflows/unittest-existence.yml
```diff
@@ -35,43 +35,12 @@ permissions:
...
-          gh issue comment ${{ github.event.pull_request.number }} --repo ${{ github.repository }} --body "${{ env.comment }}"
-        env:
-          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
-      - name: Fail job if unittest is missing and required
-        if: steps.unittest_existence.outputs.exists == 'false'
-        uses: actions/github-script@v7
-        with:
-          script: core.setFailed('Unit Test required to merge.')
+    uses: dbt-labs/actions/.github/workflows/check-change-and-comment.yml@main
```
ChenyuLInx, line 43: With the new action things should look much easier.
@emmyoop thoughts on this refactor and how to best test it?

MichelleArk (approved):

PR #10350, opened 2024-06-20
Add `CumulativeTypeParams` & sub-daily granularities to semantic manifest

### LINKED ISSUE #10360: [Feature] Support `cumulative_type_params` in semantic manifest

### Describe the feature

DSI has recently moved fields related to cumulative metrics from the `type_params` field to `type_params.cumulative_type_params`. This creates better organization in the metric spec, and also includes a new field called `type_params.cumulative_type_params.period_agg` which is used to re-aggregate cumulative metrics for non-default granularities. Add support for this field in core.

_No response_

### Who will this benefit?

_No response_

_No response_

### Anything else?

_No response_

courtneyholcomb:
resolves #10360

### CumulativeTypeParams
DSI has recently moved fields related to cumulative metrics (specifically, `window` and `grain_to_date`) from `type_params` to `type_params.cumulative_type_params`. This creates better organization in the metric spec, and also adds a new field called `type_params.cumulative_type_params.period_agg` which is used to re-aggregate cumulative metrics for non-default granularities. This PR adds core support for the new fields.

Note that the old fields `type_params.window` and `type_params.grain_to_date` will still work, so this is not a breaking change. There is a transformation that runs in MetricFlow to copy the old fields into the new fields if the new fields are not set, and sets the default `period_agg` value to `PeriodAggregation.FIRST` if it's not already set in the manifest.

Soon, setting the old fields will trigger a validation warning prompting users to move those values to the new fields. The plan is to eventually deprecate those fields.

### Sub-Daily Granularities
Since the new sub-daily granularity options are included in the DSI version upgrade, this PR also adds support for those. 

### Notes
- The related schema docs update is [here](https://github.com/dbt-labs/schemas.getdbt.com/pull/42), and I'll need to merge that to get CI passing.
- I noticed a couple of schema changes were generated that did not appear related to my changes. I'm assuming someone else's changes never got updated in the schema.
- This is my first time making a schema update here, so please LMK if I missed anything or handled something incorrectly!

### Checklist


Jstein77:
Pulling out the yaml spec here:
```
    - name: cumulative_orders
      label: Rolling total of orders (all time)
      type: cumulative
      type_params:
        measure: num_orders
        cumulative_type_params:
          period_agg: last
```

schemas/dbt/manifest/v12.json
```diff
@@ -4388,18 +4388,10 @@
...
-                    "items": {
-                      "type": "string"
-                    }
-                  },
-                  {
-                    "type": "null"
-                  }
-                ],
-                "default": null
+                "type": "array",
```
courtneyholcomb, line 4391: Not sure what this change is - I don't think it's related to my changes, might be that someone else did not update the schema in their commit?

schemas/dbt/manifest/v12.json
```diff
@@ -17930,26 +18188,7 @@
                       "type": "string"
                     },
                     "resource_type": {
```
courtneyholcomb, line 18190: This change also seems unrelated to my PR.

QMalcolm: I think this is a fix. Basically, the `resource_type` of a saved query isn't actually an enum, it's should _always_ be `saved_query`

QMalcolm: I think what's happening is we aren't running the `scripts/collect-artifact-schema.py` as often as we should. I'll bring this up to the rest of the core team. Maybe we can have a CI check for it.

core/dbt/parser/schema_yaml_readers.py
```diff
@@ -325,12 +349,15 @@ def _get_metric_type_params(self, type_params: UnparsedMetricTypeParams) -> Metr
...
+            window=self._get_optional_time_window(type_params.window),
             grain_to_date=grain_to_date,
             metrics=self._get_metric_inputs(type_params.metrics),
             conversion_type_params=self._get_optional_conversion_type_params(
                 type_params.conversion_type_params
-            )
+            ),
+            cumulative_type_params=self._get_optional_cumulative_type_params(
+                type_params.cumulative_type_params
+            ),
```
QMalcolm, line 360: I think either here or in `_get_optional_cumulative_type_params` we want to fall back on the deprecated values `type_params.window` and `type_params.grain_to_date` to populate the new values `type_params.cumulative_type_params.window` and `type_params.cululative_type_params.grain_to_date` respectively, if the new paths haven't been specified but the old paths have.

QMalcolm (changes requested):
This looks great! Thank you for doing this work ❤️ I'm not sure on everything happening in `manifest/v12.json` but it looks reasonable.

There is one thing we probably should change though. In the DSI counterpart (https://github.com/dbt-labs/dbt-semantic-interfaces/pull/288) we [discussed the migration plan](https://github.com/dbt-labs/dbt-semantic-interfaces/pull/288#pullrequestreview-2111135930). I think from that we are missing `2.b` where in we populate the new paths from the old paths (iff the new paths aren't specified and the old paths are)

courtneyholcomb:
> There is one thing we probably should change though. In the DSI counterpart (https://github.com/dbt-labs/dbt-semantic-interfaces/pull/288) we https://github.com/dbt-labs/dbt-semantic-interfaces/pull/288#pullrequestreview-2111135930. I think from that we are missing 2.b where in we populate the new paths from the old paths (iff the new paths aren't specified and the old paths are)

@QMalcolm ah ok! We are running the transformation in MF that should handle this. I didn't realize we would need to handle that in core, too. What's the reason for needing it in both places?

QMalcolm:
> @QMalcolm ah ok! We are running the transformation in MF that should handle this. I didn't realize we would need to handle that in core, too. What's the reason for needing it in both places?

In a vacuum, if all we care about is the functionality working, then that would be correct. However, that isn't our only goal. Our additional goal is to move in the direction of deprecating old paths. In order for DSI to remove the old paths, we first need to reasonably guarantee that nearly all semantic manifests being consumed have the new paths set when applicable. There are four possible states for the new paths and old paths:

1. Old paths not set, new paths not set
2. Old paths not set, new paths set
3. Old paths set, new paths not set
4. Old paths set, new paths set

Currently, any core produced semantic manifest will produce (3). In order for DSI to ever remove the old paths (and the related transformations), we have to guarantee that semantic manifests being consumed by the MetricFlow / SL don't have state (3). So, although we don't technically _need_ to do it now, we need to do it eventually. However, the longer we wait to do it, the longer DSI and Metricflow have to handle both states. Thus, doing it now means the tech debt we incur can be shorter lived and arguably more contained.

courtneyholcomb:
@QMalcolm Updated, and added tests for the old paths too.

mirnawong1:
Affects PR https://github.com/dbt-labs/docs.getdbt.com/pull/5688/files

tests/functional/metrics/test_metrics.py
```diff
@@ -402,6 +406,66 @@ def test_conversion_metric(
...
+        result = runner.invoke(["parse"])
+        assert result.success
+        assert isinstance(result.result, Manifest)
+
+        manifest = get_manifest(project.project_root)
+        metric_ids = set(manifest.metrics.keys())
+        expected_metric_ids_to_cumulative_type_params = {
+            "metric.test.weekly_visits": CumulativeTypeParams(
+                window=MetricTimeWindow(count=7, granularity=TimeGranularity.DAY),
+                period_agg=PeriodAggregation.AVERAGE,
```
QMalcolm, line 435: This could _probably_ be converted to unit test. I don't think it's necessary to do that as part of this PR though. I've been working on being able to unit test the reader clases in `schema_yaml_readers.py`, and it should be possible, but I think there is one kink to still work out first.

QMalcolm (approved):
LGTM ❤️ Thanks again!

QMalcolm:
Hmmm so the `artifact_minor_upgrade` label skips `Check Artifact Changes` CI but not `Artifact Schema Check`. I should double check what that means

courtneyholcomb:
> Hmmm so the `artifact_minor_upgrade` label skips `Check Artifact Changes` CI but not `Artifact Schema Check`. I should double check what that means

@QMalcolm I thought that was what I need the schema docs change for? Linked in the PR description but also here! https://github.com/dbt-labs/schemas.getdbt.com/pull/42

QMalcolm:
@courtneyholcomb I looked into the failing check. Basically, if it fails, we know that this change impacts the cloud artifacts team. I've added the label `impact: CA` which hopefully notifies them somehow. I'm also figuring out the appropriate way to notify them in slack.

QMalcolm:
@courtneyholcomb Additionally, in talking to the core team about the check. It is specifically not _blocking_, and we are good to proceed.

courtneyholcomb:
> @courtneyholcomb I looked into the failing check. Basically, if it fails, we know that this change impacts the cloud artifacts team. I've added the label `impact: CA` which hopefully notifies them somehow. I'm also figuring out the appropriate way to notify them in slack.

@QMalcolm awesome thank you! Merging now

courtneyholcomb:
> > @courtneyholcomb I looked into the failing check. Basically, if it fails, we know that this change impacts the cloud artifacts team. I've added the label `impact: CA` which hopefully notifies them somehow. I'm also figuring out the appropriate way to notify them in slack.
> 
> @QMalcolm awesome thank you! Merging now

@QMalcolm just kidding, I don't have permission to merge 🙃 Would you be able to merge for me?

PR #10352, opened 2024-06-21
Warn if timestamp updated_at field uses incompatible timestamp

### LINKED ISSUE #10234: Warn if `updated_at` field for snapshot is not same datatype as what's returned in `snapshot_get_time()`
### Description

There are 4 different semantics for timestamp datatypes - for simplicity let’s say 2 categories:

*  naive timestamp - no timezone information, no UTC offset “noon on january 1st, 2024”
*  aware timestamp - is aware of timezone
    * UTC offset like “-2 hours”; 
    * specific named timezone “America/Los Angeles”; 
    * absolute timestamp `ltz` uses your session timezone

One of the trickiest things that can happen within snapshots is when there are [naive](https://docs.python.org/3/library/datetime.html#aware-and-naive-objects) timestamps (rather than [aware](https://docs.python.org/3/library/datetime.html#aware-and-naive-objects)). In those cases, we need clear ways for the user to configure a "mutual agreement" with the dbt system how to interpret those timestamps when they are actively involved in the snapshot configuration. 

**Up until this point, the approach has been that naive timestamp must be UTC** 
- there is no way to specify anything else
- we don't document this
- there's no warning if you supply a data type that's not a UTC timestamp

Some users may want to use an adjacent column that contains the relevant UTC offset or time zone (or configure the time zone globally for the dbt project (like [this](https://github.com/calogica/dbt-date/blob/ca326b2d6d946edd3b3bfea98c88767840c37765/README.md#variables)) or specifically for one model).

Example of snapshot - `updated_at` when using `strategy=timestamp`:

```
{% snapshot orders_snapshot %}

{{
    config(
      target_database='analytics',
      target_schema='snapshots',
      unique_key='id',

      strategy='timestamp',
      updated_at='updated_at_field'
    )
}}

select * from {{ source('jaffle_shop', 'orders') }}

{% endsnapshot %}
```

### PROBLEM 

When the initial snapshot occurs, the `dbt_valid_from` and `dbt_valid_to` columns inherit the data type of whatever you've provided for `updated_at` config (could be a `date`, could be a naive timestamp, etc.). But once a record is changed, dbt updates the `dbt_valid_to` column to `snapshot_get_time()` which returns (in snowflake) a `timestamp_ntz` data type in the UTC time zone. This is then implicitly converted to the original inherited data type, **which can lead to incorrect data** if their `updated_at` field is NOT `timestamp_ntz` data type in the UTC time zone (incorrect comparisons; may coerce timestamp differently than user expects). 

Naive vs. aware timestamps in Snowflake:
https://docs.google.com/presentation/d/14qrx8YbnGnoP2FgnxSD5Z8fxcnoYq4DKRrfHBDanbfA/edit#slide=id.g20f6efbe802_0_34

### SOLUTION

Let’s go with Idea 1 for now, maybe some magic in the future

#### Idea 1 - put burden of using correct data type on user:

- go look up what data type `updated_at` field is, see if it’s compatible
- if it’s not throw an warning (or error but that would be backwards incompatible, so would need to use a flag) to tell user to convert field to correct data type (adapter dependent)
- `updated_at` already accepts a SQL statement [though we should add docs for this]:
```
    {% snapshot orders_snapshot %}
    
    {{
        config(
          target_database='analytics',
          target_schema='snapshots',
          unique_key='id',
    
          strategy='timestamp',
          updated_at='convert_timezone('UTC', updated_at_field),'YYYY-MM-DD HH24:MI:SS.FF3 TZHTZM')'
        )
    }}
    
    select * from {{ source('jaffle_shop', 'orders') }}
    
    {% endsnapshot %}
```
    
#### Idea 2 - magic stuff to automatically convert:

- go look up what data type `updated_at_field` is, see if it’s compatible
- based on what it is, automatically do the conversion to the correct data type (adapter dependent)

gshank:
resolves #10234

### Problem

Snapshot tables use the 'get_snapshot_time' macro, which may be incompatible with the data type of the field specified as the 'updated_at' field, causing unexpected results.

### Solution

Warn if the two sources of timestamp columns have different data types.

### Checklist


.changes/unreleased/Features-20240621-164400.yaml
```diff
@@ -0,0 +1,6 @@
+kind: Features
+body: Warn on incompatible snapshot timestamps
+time: 2024-06-21T16:44:00.023992-04:00
+custom:
+  Author: gshank
+  Issue: "10234"
```
MichelleArk, line 6: nit: are two changelog entries intentional?

gshank: Removed duplicate

core/dbt/events/core_types_pb2.py
```diff
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 # Generated by the protocol buffer compiler.  DO NOT EDIT!
 # source: core_types.proto
-# Protobuf Python Version: 5.26.1
+# Protobuf Python Version: 4.25.3
```
MichelleArk, line 4: Let's ensure we're using 5.26.1 if possible to generate the protos: https://github.com/dbt-labs/dbt-core/pull/10456

MichelleArk (approved):
Small note on [core/dbt/events/core_types_pb2.py](https://github.com/dbt-labs/dbt-core/pull/10352/files/5f15e2876712126552649cacba8cdf4f41288cba#diff-5482d82eecf21674d78be2bc404d17375cd9fd1cab15a9cc3bf07a79b43e5704), otherwise looks good!

MichelleArk:
Added a tech debt ticket re: protoc versions; https://github.com/dbt-labs/dbt-core/issues/10658

PR #10358, opened 2024-06-24
pin versions so CI and local dev match

emmyoop:
resolves #

### Problem

Requirements for pre-commit hooks were unpinned in `dev-requirements.txt` resulting in differences for local development and CI.

### Solution

Pin the requirements in dev-requirements so that local development matches CI.

Bump isort because there was no breaking change.

black breaks and requires other code changes for major version 24 so I left the pin as is for now
flake8 also finds more problems on 1.10 so left that as well

### Checklist


aranke (approved):
LGTM!

PR #10364, opened 2024-06-25
pin twine version

ChenyuLInx:
the latest twine is broken
https://github.com/pypa/twine/issues/1125

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -268,7 +268,7 @@ jobs:
       - name: Install python dependencies
         run: |
           python -m pip install --user --upgrade pip
-          python -m pip install --upgrade setuptools wheel twine check-wheel-contents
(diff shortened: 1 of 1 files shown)


aliceliu (approved):

.github/workflows/main.yml
```diff
@@ -268,7 +268,7 @@ jobs:
       - name: Install python dependencies
         run: |
           python -m pip install --user --upgrade pip
-          python -m pip install --upgrade setuptools wheel twine check-wheel-contents
+          python -m pip install --upgrade setuptools wheel twine==5.0.0 check-wheel-contents
```
aranke, line 271: Let's avoid pinning inline here.

Also looks like `twine==5.1.0` was yanked: https://pypi.org/project/twine/5.1.0/

PR #10366, opened 2024-06-26
Support for redshift 821

### LINKED ISSUE #10365: [Bug] Unit test fixture null values cause typing issues

### Current Behavior

Unit test fixtures are parsed correctly, but they leave typing up to the target database. This will result in unhandled database failures when there is no one row Null free (this handles indirectly the all values in a column Null failure case). The database won't be able to determine which type to use and default to something like VARCHAR(1) which may not be what the source table expects -- this leads to a type failure. 

This change is motivated by a redshift change but could affect other adapters some day.

### Expected Behavior

Let's sort this to give fixtures the best chance to succeed or in unwinnable situations, error out with a known exception.

### Steps To Reproduce

See reprex [here](https://github.com/dbt-labs/dbt-redshift/issues/821#issuecomment-2189925914)

### Relevant log output

_No response_

### Environment

```markdown
- OS:
- Python:
- dbt:
```

redshift

### Additional Context

See redshift [PR here](https://github.com/dbt-labs/dbt-redshift/pull/852)

VersusFacit:
Resolves #10365 

### Solution

Sort and do some like validation for nulls.

### Checklist


core/dbt/parser/unit_tests.py
```diff
@@ -388,6 +388,18 @@ def _validate_and_normalize_rows(self, ut_fixture, unit_test_definition, fixture
                     ut_fixture.fixture, self.project.project_name, unit_test_definition.unique_id
                 )
 
+        # Some databases like redshift will default the type of a column
+        # where the first value is None to a type like VARCHAR(1)
+        if ut_fixture.rows and (
```
colin-rogers-dbt, line 393: this feels like it should be unit testable (I realize the semantic ridiculousness of unit testing `unit_tests.py` but here we are)

VersusFacit: Isn't it already covered by existing "unit test" functional tests in core?

ChenyuLInx: Coverage does not tell the whole story. It just shows some test is executing this code. It does not show functionality is correct

We have 5 actual lines here, two of them are evaluation, one is a side effect, one is assignment and one is raising an error(not covered, and seems like an important behavior). The first 4 lines just happened to get executed by other tests.

For bug fixes, we should have a test to repro the bug, it will fail without code change, then with the code change test pass. This both prevents future regression and acts as the documentation for the code.

Side note: this is a private function, so if you write a test, you can test the public functions using it instead of private function([reference](https://enterprisecraftsmanship.com/posts/unit-testing-private-methods/))

VersusFacit: Hey @ChenyuLInx , thanks for chiming in!

> For bug fixes, we should have a test to repro the bug
Absolutely. Unfortunately, because this is a Redshift surfacing issue, I don't think I can provide the usual postgres reprex-fix-test combo.

#### Allow me to offer some context:

Among our 1p adapters, this bug is unique to Redshift's type interpolation behavior. It does not occur in Postgres. However, Redshift nor the base adapter has a straightforward way -- that I know of -- to intercept and modify the order of these rows.

Also for reference, the redshift PR does (indirectly) test this code [in this PR here](https://github.com/dbt-labs/dbt-redshift/pull/852/files#diff-fa16d6a4b96751c43394815126f09d409c56cc89baff1a089af16c15e55118baR48) both in the success and failure case.

> one is raising an error(not covered, and seems like an important behavior). 
see last sentence 🙏

colin-rogers-dbt:
@MichelleArk curious if this issue might be a code smell for us to eventually move some of this functionality into the base adapter? That would enable different adapters to handle these kinds of edge cases on their own

VersusFacit:
Colin, if we could move even just the final sql rendering into the base adapter, I could keep all this logic out of core and provide more logically interfaces for testing. So that's an enthusiastic +1 from me. 

I did explore some additional ways to getting at the prerendered sql in the base adapter but no dice.

core/dbt/parser/unit_tests.py
ChenyuLInx: So the contract you are looking for from of this [parse](https://github.com/dbt-labs/dbt-core/blob/0671f85af62136c4014e88e0e9fc80f711c302e6/core/dbt/parser/unit_tests.py#L248) function is : the rows are sorted. and will raise an error in certain cases.
I would write a unit test to check that(and not involve adapters). Me or @QMalcolm are happy to pair if you run into a problem trying to write that. We want to make writing unit test easier.

VersusFacit: Okay, that sounds like a solid way to go about it. I'll jump in.

I'll take a look at some existing unit tests and write one. Will reach out as needed!

VersusFacit:
circling back here, I added several unit tests here for coverage 🖖 Thanks for @ChenyuLInx for the support in using the existing test framework.

Redshift tests added in [this PR](https://github.com/dbt-labs/dbt-redshift/pull/852). That PR branch is pointing to the changes in this core branch. The change here will resolve the ticket that led to the Redshift PR.
image

ChenyuLInx (approved):
Thanks for adding this!

PR #10371, opened 2024-06-26
Incremental models with a contract don't need their columns modified

### LINKED ISSUE #10362: [Bug] Incorrect column data type with incremental contracted model and varchar data_type
### Current Behavior

A temporary table is created for incremental models when the table already exists. When this temporary table is created and one of the column contracted data_types is "character varying(1)" (size is not significant) the string size is lost in the creation of the temporary relation and instead the column is altered after the temporary table is created like: alter table "xxx"."xxxx"."xxx" add column "<column_name>__dbt_alter" character varying(256);

The column is created with the wrong string size.

### Expected Behavior

The column is created with the string size that's specified in the contract. 

### Steps to recreate

Create an incremental model, update it with additional rows (causing the creation of a temporary table) and observe the the wrong string size is used.

### Additional Context

Reported internally by Lee Bond-Kennedy.

gshank:
resolves #10362

### Problem

Incremental models with existing tables had their columns updated by "expand_target_column_types", but this is unnecessary when models have a contract.

### Solution

Do not call "expand_target_column_types" when an incremental model has an enforced contract.

### Checklist


dev-requirements.txt
```diff
@@ -1,4 +1,4 @@
-git+https://github.com/dbt-labs/dbt-adapters.git@main
+git+https://github.com/dbt-labs/dbt-adapters.git@incremental_contract_varchar
```
mikealfare, line 1: Reminder to revert this prior to merging.

mikealfare (approved):
Approved, conditional on reverting the feature branch reference in `dev-requirements.txt`.

PR #10394, opened 2024-07-02
[TIDY-FIRST] Fix isort for dbt-semantic-interfaces

QMalcolm:
resolves #N/A

### Problem

We've been grouping internal package imports by using `isort`. However, imports from `dbt_semantic_interfaces` weren't being grouped because I typo'd it as `dbt_semantic_interfaces` in the isort config.

### Solution

Fix the typo and rerun isort

### Checklist

### DIFF
--- a/.isort.cfg
+++ b/.isort.cfg
@@ -1,4 +1,4 @@
 [settings]
 profile=black
 extend_skip_glob=.github/*,third-party-stubs/*,scripts/*
-known_first_party=dbt,dbt_adapters,dbt_common,dbt_extractor,dbt_semantic_interface
+known_first_party=dbt,dbt_adapters,dbt_common,dbt_extractor,dbt_semantic_interfaces
--- a/core/dbt/artifacts/resources/v1/metric.py
+++ b/core/dbt/artifacts/resources/v1/metric.py
@@ -2,14 +2,6 @@ import time
 from dataclasses import dataclass, field
 from typing import Any, Dict, List, Literal, Optional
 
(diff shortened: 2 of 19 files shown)


.isort.cfg
```diff
@@ -1,4 +1,4 @@
 [settings]
 profile=black
 extend_skip_glob=.github/*,third-party-stubs/*,scripts/*
-known_first_party=dbt,dbt_adapters,dbt_common,dbt_extractor,dbt_semantic_interface
+known_first_party=dbt,dbt_adapters,dbt_common,dbt_extractor,dbt_semantic_interfaces
```
emmyoop, line 4: Should `snowplow_tracker` also be included here?  We maintain our own fork of it (minimal-snowplow-tracker)

emmyoop (changes requested):

.isort.cfg
QMalcolm: That's a good question 🤔 I do not know. We can, but it feels in a gray area. It is a fork that we maintain, that we don't appear to sync. However, it also overrides the name space of a non-internal package. If you think it's cut and dry that it should be included, happy to make the change. However, if you also feel that it's in a gray area, maybe it warrants more discussion and shouldn't be considered blocking for this PR.

emmyoop: After talking offline I agree we should solve this in #8409

QMalcolm:
@cla-bot check

QMalcolm:
Good bot

PR #10423, opened 2024-07-09
Global CLI flag should override env var flag

### LINKED ISSUE #10304: [Bug] Environment variables take precedence over CLI flags that appear after the subcommand

### Current Behavior

[As requested by](https://dbt-labs.slack.com/archives/C05CJT5MJF2/p1718202630252599?thread_ts=1718034218.496729&cid=C05CJT5MJF2) @graciegoheen 

[The docs say](https://docs.getdbt.com/reference/global-configs/about-global-configs#setting-flags)
>The most specific setting “wins.” If you set the same flag in all three places, the CLI option will take precedence, followed by the environment variable, and finally, the value in dbt_project.yml. If you set the flag in none of those places, it will use the default value defined within dbt.

But the observed precedence distinguishes between flags before the subcommand vs. after the subcommand

### Acceptance criteria

1. CLI flags take precedence over environment variables regardless if they appear before the subcommand vs. after the subcommand.
2. These commands give identical output:
    ```shell
    DBT_QUIET=0 dbt --quiet list
    DBT_QUIET=0 dbt list --quiet
    ```

### Steps To Reproduce

See Doug's comment below

gshank:
resolves #10304

### Problem

When both an environment variable and a command line flag are provided, the cli flag should override the environment variable. This wasn't happening for "global" flags, which can occur both before and after the sub command, when the flag came after the command.

### Solution

The click context contains a "child" context for command options and a "parent" context for flags which occur before the command. If there was no "cli" setting for a flag, it would use the env var. When processing the parent context flags we add code to check if the source of this parameter is an env var, and do not overwrite the value of the param if it has already been set by the user.

### Checklist

### DIFF
--- a/tests/unit/cli/test_flags.py
+++ b/tests/unit/cli/test_flags.py
@@ -380,6 +380,22 @@ class TestFlags:
 
         assert flags_a.USE_COLORS == flags_b.USE_COLORS
 
+    def test_global_flag_with_env_var(self, monkeypatch):
+        # The environment variable is used for whichever parent or child
+        # does not have a cli command.
+        # Test that "child" global flag overrides env var
+        monkeypatch.setenv("DBT_QUIET", "0")
+        parent_context = self.make_dbt_context("parent", ["--no-use-colors"])
+        child_context = self.make_dbt_context("child", ["--quiet"], parent_context)
+        flags = Flags(child_context)
+        assert flags.QUIET is True
+
+        # Test that "parent" global flag overrides env var
+        parent_context = self.make_dbt_context("parent", ["--quiet"])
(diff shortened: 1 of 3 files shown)


tests/unit/cli/test_flags.py
```diff
@@ -380,6 +380,22 @@ def test_global_flag_at_child_context(self):
 
         assert flags_a.USE_COLORS == flags_b.USE_COLORS
 
+    def test_global_flag_with_env_var(self, monkeypatch):
```
ChenyuLInx, line 383: Love this test

ChenyuLInx (approved):
Thanks!

PR #10426, opened 2024-07-10
Dc/expoure and source groups

### LINKED ISSUE #9339: [CT-3527] [implementation] extend access and groups to exposures & source
### Housekeeping

### Short description

From the discussion [here](https://github.com/dbt-labs/dbt-core/issues/7750)...

We should extend `group` configurations to all resource types. The current outstanding ones are `sources` and `exposures`:
- exposures should be able to reference models within the same group
- sources should be able to be private to a given group

We should also enable `sources` to configure `access` to `private` or `protected` (default). We **should not** allow folks to set `access` to `public` for `sources`.

### Acceptance criteria

- [ ] you can set `group` config for an exposure
- [ ] you can set `group` config for a source
- [ ] you can set `access` to `private` or `protected` (default) for a source
- [ ] you can **not** set `access` for an exposure (they are inherently public)

As described [here](https://docs.getdbt.com/reference/resource-configs/access#definition), referencing a resource outside of its supported access should raise an error (but otherwise should not).

### Impact to Other Teams

none

none

### Context

Notes from technical refinement:
- would allow folks to set this for the source, and also for a specific table within a source
- it's a config (enabled is the only one right now https://docs.getdbt.com/reference/source-configs), may be more complicated
- alt approach discussed [here](https://github.com/dbt-labs/dbt-core/issues/12212) and [here](https://github.com/dbt-labs/dbt-core/issues/10132#issuecomment-2107432551)

dave-connors-3:
resolves #9339

### Checklist

### DIFF
--- a/core/dbt/artifacts/resources/v1/exposure.py
+++ b/core/dbt/artifacts/resources/v1/exposure.py
@@ -3,7 +3,7 @@ from dataclasses import dataclass, field
 from typing import Any, Dict, List, Literal, Optional
 
 from dbt.artifacts.resources.base import GraphResource
-from dbt.artifacts.resources.types import NodeType
+from dbt.artifacts.resources.types import AccessType, NodeType
 from dbt.artifacts.resources.v1.components import DependsOn, RefArgs
 from dbt.artifacts.resources.v1.owner import Owner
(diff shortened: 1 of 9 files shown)


zqureshi:

PR #10446, opened 2024-07-15
Revert "Support for redshift 821"

ChenyuLInx:
Reverts dbt-labs/dbt-core#10366
Since this will introduce parsing failure to projects that currently do not fail.

### DIFF
--- a/core/dbt/parser/unit_tests.py
+++ b/core/dbt/parser/unit_tests.py
@@ -389,42 +389,6 @@ class UnitTestParser(YamlReader):
                     ut_fixture.fixture, self.project.project_name, unit_test_definition.unique_id
                 )
 
-        # sanitize order of input
-        if ut_fixture.rows and (
-            ut_fixture.format == UnitTestFormat.Dict or ut_fixture.format == UnitTestFormat.CSV
-        ):
-            self._promote_first_non_none_row(ut_fixture)
-
-    def _promote_first_non_none_row(self, ut_fixture):
-        """
-        Promote the first row with no None values to the top of the ut_fixture.rows list.
-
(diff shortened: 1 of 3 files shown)


aranke (approved):
LGTM!

MichelleArk (approved):

core/dbt/parser/unit_tests.py
```diff
@@ -389,42 +389,6 @@ def _validate_and_normalize_rows(self, ut_fixture, unit_test_definition, fixture
...
-        for index, row in enumerate(ut_fixture.rows):
-            # Check if all values in the row are not None
-            if all(value is not None for value in row.values()):
-                non_none_row_index = index
-                break
-
-        if non_none_row_index is None:
-            raise ParsingError(
-                "Unit Test fixtures require at least one row free of Nones to ensure consistent column types."
-            )
```
MichelleArk, line 421: I'd suggest isolating this exception only to adapters that will actually fail to execute if this is the case to make this backward-compatible in terms of expected behaviour. 

For example, we could introduce a `validate_row(row, row_idx)` macro and call it here in the base implementation: https://github.com/dbt-labs/dbt-adapters/blob/782a32b530e25afe34af6896392d9fb081dce9d1/dbt/include/global_project/macros/unit_test_sql/get_fixture_sql.sql#L26-L27

By default, the macro would simply pass but adapters could implement the macro to raise an exception (or warning if they wouldn't fail 100% of the time) if the row. 

Just proposing a high-level approach here, don't meant to be overly-prescriptive about the suggested interfaces (e.g. maybe validate_rows (plural) makes more sense instead of passing an index along).

PR #10448, opened 2024-07-16
Support for redshift 821

VersusFacit:
take 2 on #10366 

### Problem

Merge of the original code with an exception for a case the team thought to be unlikely to exist, did in fact exist in one of our downstream analytics projects.

I conferred with the team and we have landed on instead logging a warning and letting the code move ahead after making an effort to float non-`None` value columns to the first row inserted from an `excepted` fixture.

### Solution

Borrowing the existing `dbt_common` event to avoid adding yet another message type for a one-off error. Also, we will propose floating this logic to dbt adapters at some point, so we don't want to bake in too much logic right now.

### Checklist


tests/unit/parser/test_unit_tests.py
```diff
@@ -173,3 +226,62 @@ def test_multiple_unit_tests(self):
...
+
+    def test_expected_promote_non_none_row_sql(self):
+        expected_rows = "select null\n" + "select 1"
+        self._parametrize_test_promote_non_none_row(
+            UNIT_TEST_NONE_ROWS_SORT_SQL, UnitTestFormat.SQL, expected_rows
+        )
+
+    def test_no_full_row_does_not_raise_exception(self):
+        block = self.yaml_block_for(UNIT_TEST_NONE_ROWS_SORT_FAILS, "test_my_model.yml")
+        UnitTestParser(self.parser, block).parse()
```
MichelleArk, line 287: This test can more precisely test the expected event was fired with something like 
```py
from dbt_common.events.event_manager_client import add_callback_to_manager
from dbt_common.events.types import SystemStdErr
from tests.utils import EventCatcher
...

def test_no_full_row_does_not_raise_exception(self):
  catcher = EventCatcher(SystemStdErr)
  add_callback_to_manager(catcher.catch)
  ... 
  UnitTestParser(self.parser, block).parse()
  assert len(catcher.caught_events) == 1
```

tests/unit/parser/test_unit_tests.py
```diff
@@ -173,3 +226,62 @@ def test_multiple_unit_tests(self):
         for unit_test in self.parser.manifest.unit_tests.values():
             self.assertEqual(len(unit_test.depends_on.nodes), 1)
             self.assertEqual(unit_test.depends_on.nodes[0], "model.snowplow.my_model")
+
+    def _parametrize_test_promote_non_none_row(
```
MichelleArk, line 230: took me a while to realize the purpose of this method wasn't just to parametrize / set up a test but also to run assertions. non-blocking naming suggestion would be to call this `_assert_fixture_yml_matches_expected_rows`, or something similar?

MichelleArk: (more examples in tests/unit/parser/test_manifest.py)

core/dbt/parser/unit_tests.py
```diff
@@ -389,6 +391,44 @@ def _validate_and_normalize_rows(self, ut_fixture, unit_test_definition, fixture
...
+        # Iterate through each row and its index
+        for index, row in enumerate(ut_fixture.rows):
+            # Check if all values in the row are not None
+            if all(value is not None for value in row.values()):
+                non_none_row_index = index
+                break
+
+        if non_none_row_index is None:
+            fire_event(
+                SystemStdErr(
```
MichelleArk, line 422: This looks like a debug-level event, so it won't be visible to the user unless they go looking in the debug logs.

I think we do should actually make this a user-visible warn event, and that worth making a new event in core so that users can promote it to error or silence it using warn_error_options.

MichelleArk: Ideally we can also include the fixture name that caused the warning in the message so users can action on it easily

MichelleArk: We should also use `warn_or_error` so that warn_error_options is respected and the warning can be promoted/silenced using that config

core/dbt/parser/unit_tests.py
```diff
@@ -389,6 +391,44 @@ def _validate_and_normalize_rows(self, ut_fixture, unit_test_definition, fixture
...
+                    bmsg="Unit Test fixtures benefit from having at least one row free of Null values to ensure consistent column types. Failure to meet this recommendation can result in type mismatch errors between unit test source models and `expected` fixtures."
```
MichelleArk, line 423: this is very clear from my perspective reading this diff but I'm worried it's somewhat verbose for an end-user reading this. What about something more straightforward like: 

"Unit test fixutre <fixture-name> in <unit-test-name> does not have any row free of null values, which may cause type mismatch errors during unit test execution."

VersusFacit: I'll take this suggestion for the user facing exception to be raised in redshift

MichelleArk (approved):

PR #10451, opened 2024-07-16
Fix multiple semantic models with generated metrics

### LINKED ISSUE #10450: Partial parsing issue with generated metrics on multiple semantic models in the same file
### Current Behavior

If multiple semantic models containing generated metrics (the "create_metric" flag is set to true on a measure) are in the same file and one of the semantic models is modified, the generated metrics for the other semantic model are removed and not replaced. This can result in error messages such as "Saved_Query ... depends on a metric named '...' which was not found". 

### Expected Behavior

Generated metrics referenced by saved queries are found when they should be.

### Steps to Reproduce

* Create multiple semantic models in the same file, each of which has "create_metric" set for a measure.
* Modify one of the semantic models and re-parse.
* Attempt to run a saved query referring to the generated metric from the non-changed semantic model.

### Additional Context

This problem can be avoided by separating out semantic models with generated metrics into separate files.

gshank:
resolves #10450

### Problem

When partial parsing, all generated metrics in a file were being deleted instead of just for the changed semantic model.

### Solution

Use a dictionary in SchemaSourceFile, with the semantic model name for the key and a list containing metric unique ids for the value.

### Checklist

### DIFF
--- a/core/dbt/contracts/files.py
+++ b/core/dbt/contracts/files.py
@@ -275,7 +275,7 @@ class SchemaSourceFile(BaseSourceFile):
             parts = metric_unique_id.split(".")
             metric_name = parts[-1]
             if "semantic_models" in self.dict_from_yaml:
-                for sem_model in self.dict_from_yaml["semantic_model"]:
+                for sem_model in self.dict_from_yaml["semantic_models"]:
                     if "measures" in sem_model:
                         for measure in sem_model["measures"]:
                             if measure["name"] == metric_name:
--- a/core/dbt/contracts/files.py
+++ b/core/dbt/contracts/files.py
@@ -194,6 +194,7 @@ class SchemaSourceFile(BaseSourceFile):
     metrics: List[str] = field(default_factory=list)
     # metrics generated from semantic_model measures
     generated_metrics: List[str] = field(default_factory=list)
+    metrics_from_measures: Dict[str, Any] = field(default_factory=dict)
     groups: List[str] = field(default_factory=list)
     # node patches contain models, seeds, snapshots, analyses
     ndp: List[str] = field(default_factory=list)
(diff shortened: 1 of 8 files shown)


core/dbt/contracts/files.py
```diff
@@ -259,6 +260,28 @@ def get_tests(self, yaml_key, name):
...
+        if self.generated_metrics:
+            self.fix_metrics_from_measures()
+        if semantic_model_name not in self.metrics_from_measures:
+            self.metrics_from_measures[semantic_model_name] = []
+        self.metrics_from_measures[semantic_model_name].append(metric_unique_id)
+
+    def fix_metrics_from_measures(self):
+        # Loop through yaml dictionary looking for matching generated metrics
+        generated_metrics = self.generated_metrics
+        self.generated_metrics = []
```
ChenyuLInx, line 273: Are we using this manipulation of generated_metrics to avoid `fix_metrics_from_measures` and `fix_metrics_from_measures` infinite looping? Can we try to organize the function so it is easier to read?

core/dbt/contracts/files.py
ChenyuLInx, line 1: Is it possible to add some unittest to test this new logic added? I see the coverage is only 10%-ish.

core/dbt/parser/partial.py
```diff
@@ -968,13 +968,15 @@ def delete_schema_semantic_model(self, schema_file, semantic_model_dict):
...
-        metrics = schema_file.generated_metrics.copy()
-        for unique_id in metrics:
-            if unique_id in self.saved_manifest.metrics:
-                self.saved_manifest.metrics.pop(unique_id)
-                schema_file.generated_metrics.remove(unique_id)
-            elif unique_id in self.saved_manifest.disabled:
-                self.delete_disabled(unique_id, schema_file.file_id)
+        if schema_file.generated_metrics:
+            schema_file.fix_metrics_from_measures()
+        if semantic_model_name in schema_file.metrics_from_measures:
```
ChenyuLInx, line 973: Can we add a bit of comment here?

tests/functional/semantic_models/test_semantic_model_parsing.py
```diff
@@ -146,3 +147,28 @@ def test_semantic_model_flipping_create_metric_partial_parsing(self, project):
         # Verify the metric originally created by `create_metric: true` was removed
         metric = result.result.metrics[generated_metric]
         assert metric.name == "txn_revenue"
+
+
+class TestSemanticModelPartialParsingGeneratedMetrics:
```
ChenyuLInx, line 152: Nice re-pro case!!!

ChenyuLInx (approved):
Thanks for reproducing this and adding the fix!
Understanding what's going on here is a bit tricky, can we add a bit more comments? And maybe do a live walk through with the team to pass the knowledge along?
Maybe this is because I am not familiar with how this section of the logic works.

core/dbt/parser/partial.py
```diff
@@ -968,13 +968,17 @@ def delete_schema_semantic_model(self, schema_file, semantic_model_dict):
             elif unique_id in self.saved_manifest.disabled:
                 self.delete_disabled(unique_id, schema_file.file_id)
 
-        metrics = schema_file.generated_metrics.copy()
```
ChenyuLInx, line 971: So here we were deleting all generated metrics instead of just the one related to a semantic model

ChenyuLInx (approved):
I am following this time! Thanks a lot for the comment and explanation!

core/dbt/parser/partial.py
gshank: Correct.

PR #10453, opened 2024-07-16
Warn error options invalid when warn or error set to none

### LINKED ISSUE #10452: warn_error_options with empty warn or error fields does not use silence
### Current Behavior

If you use the example in the docs for silencing warnings:
```
  warn_error_options:
    error: # Previously called "include"
    warn: # Previously called "exclude"
    silence: # To silence or ignore warnings
      - TestsConfigDeprecation
      - NoNodesForSelectionCriteria
```
it doesn't work to silence TestsConfigDeprecation. You have to remove the empty error: and warn: fields.

### Expected Behavior

The example should work for silencing warnings.

### Steps to Reproduce

Copy the config into dbt_project.yml in a project with a "tests:" config.

### Relevant log output

```
18:43:15  [WARNING]: Deprecated functionality
The `tests` config has been renamed to `data_tests`. Please see
https://docs.getdbt.com/docs/build/data-tests#new-data_tests-syntax for more
information.
```

### Additional Context

gshank:
resolves #10452

### Problem

If a field in warn_error_options was set to None, a valid WarnErrorOptions object was not constructed.

### Solution

Fix the function to normalize the warn_error_options dictionary.

### Checklist

### DIFF
--- a/.changes/unreleased/Fixes-20240716-171427.yaml
+++ b/.changes/unreleased/Fixes-20240716-171427.yaml
@@ -1,0 +1,6 @@
+kind: Fixes
+body: Fix error constructing warn_error_options
+time: 2024-07-16T17:14:27.837171-04:00
+custom:
+  Author: gshank
+  Issue: "10452"
--- a/core/dbt/cli/option_types.py
+++ b/core/dbt/cli/option_types.py
@@ -1,6 +1,6 @@
 from click import Choice, ParamType
 
-from dbt.config.utils import exclusive_primary_alt_value_setting, parse_cli_yaml_string
+from dbt.config.utils import normalize_warn_error_options, parse_cli_yaml_string
 from dbt.events import ALL_EVENT_NAMES
 from dbt.exceptions import OptionNotYamlDictError, ValidationError
 from dbt_common.exceptions import DbtValidationError
@@ -51,12 +51,7 @@ class WarnErrorOptionsType(YAML):
     def convert(self, value, param, ctx):
         # this function is being used by param in click
         include_exclude = super().convert(value, param, ctx)
-        exclusive_primary_alt_value_setting(
-            include_exclude, "include", "error", "warn_error_options"
-        )
-        exclusive_primary_alt_value_setting(
-            include_exclude, "exclude", "warn", "warn_error_options"
-        )
+        normalize_warn_error_options(include_exclude)
 
         return WarnErrorOptions(
             include=include_exclude.get("include", []),
--- a/core/dbt/config/project.py
+++ b/core/dbt/config/project.py
@@ -10,7 +10,7 @@ from dbt import deprecations
 from dbt.adapters.contracts.connection import QueryComment
 from dbt.clients.yaml_helper import load_yaml_text
 from dbt.config.selectors import SelectorDict
-from dbt.config.utils import exclusive_primary_alt_value_setting
+from dbt.config.utils import normalize_warn_error_options
 from dbt.constants import (
     DBT_PROJECT_FILE_NAME,
     DEPENDENCIES_FILE_NAME,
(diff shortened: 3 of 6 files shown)


QMalcolm (changes requested):
First, thank you for doing this work ❤️  This is well done 🙂 However, I worry the changes as is if merged would remove a generalized utility that we'll likely continually need.

Specifically, `exclusive_primary_alt_value_setting` was created specifically to handle general name changes gracefully. Although it is only being used for `exclude/warn` and `include/error` at this time, it is not unlikely that there will be other keys we want to rename in the future. As such, altering the function to only work with `warn_error_options` seems like a step back.

I think there are two simpler changesets to fix the problem we are trying to address: extending `exclusive_primary_alt_value_setting` to default the associated value OR create an orchestrating function.

### Alt (1): extending `exclusive_primary_alt_value_setting`

First the function would take an additional optional argument making the function signature look like
```python
def exclusive_primary_alt_value_setting(
    dictionary: Optional[Dict[str, Any]],
    primary: str,
    alt: str,
    parent_config: Optional[str] = None,
    default_value: Optional[Any] = None,
) -> None:
```

Then at the end of the function we'd conditionally set the default via

```python
if dictionary.get(primary) is None:
    dictionary[primary] = default_value
```

### Alt (2): Orchestrating Function
If we didn't want to modify `exclusive_primary_alt_value_setting` at all, then we could put the defaulting logic into an orchestrating function specific for `warn_error_options` munging. Somthing like
```python
def normalize_warn_error_options(dictionary: Dict[str, Any]) -> None:
    exclusive_primary_alt_value_setting(
        warn_error_options, "include", "error", "warn_error_options"
    )
    exclusive_primary_alt_value_setting(
        warn_error_options, "exclude", "warn", "warn_error_options"
    )
    if dictionary.get("include") is None:
        dictionary["include"] = []
    if dictionary.get("exclude") is None:
        dictionary["exclude"] = []
```

gshank:
I see that you intended "exclusive_primary_alt_value_setting" as a general purpose utility, but it looks like premature generalization to me.  What I've found is that when/if you actually get to some other use case that falls kind of in the same category, it will have other constraints or conditions and the intended "general" utility won't actually work at all without complicating changes.

I think it makes more sense to derive a general use case utility when there is more than one use case.

QMalcolm:
> I see that you intended "exclusive_primary_alt_value_setting" as a general purpose utility, but it looks like premature generalization to me. What I've found is that when/if you actually get to some other use case that falls kind of in the same category, it will have other constraints or conditions and the intended "general" utility won't actually work at all without complicating changes.
> 
> I think it makes more sense to derive a general use case utility when there is more than one use case.

I'm not sure that I agree that it's a pre-mature generalization, but I'm absolutely open to discussing it further. Can you talk more about your concern that makes the existing function pre-mature? In my mind, there is more than one use case. The existing function addresses the use case of "A key in a dictionary is being renamed to another, but we want to support both as inputs". It is being used for two keys. It renames "error" to "include" and "warn" to "exclude". 

I do agree that there could be more constraints. The issue that this PR aims to address is one of those, the ability to have a default value association. This could simply be addressed as outlined above, extending the existing function or external to it. There are additional constraints that could be imagined, such as supporting multiple "alt" key names for a given "primary". I don't recommend we do work to handle multiple "alts" as we don't have an existing or planned use case where that is necessary.

QMalcolm:
From your comment and my comments, my understanding is that we view the underlying discrete cases differently. I also few them both as valid. The work I did on renaming dictionaries keys in dictionaries has more than one use case. However, we need to default all the values for keys in `warn_error_options` to an empty list if the value is `None`. I think this could be included or excluded as part of the `exclusive_primary_alt_value_setting`, and which way it goes isn't too important in my mind. I think for the sake of defaulting all three values in the same manner though, that we should do it externally. A compromise solution being

```python
def normalize_warn_error_options(
    dictionary: Dict[str, Any],
) -> None:
    convert = {
        "error": "include",
        "warn": "exclude",
        "silence": None,
    }

    for key, alt convert.items():
        if alt is not None:
            exclusive_primary_alt_value_setting(warn_error_options, key, alt, "warn_error_options")
        
        if dictionary.get(key) is None:
            dictionary[key] = []
```

QMalcolm (approved):
LGTM

PR #10473, opened 2024-07-22
Support passing `args` as keyword argument for `run-operation` in programmatic invocations

### LINKED ISSUE #10355: [Bug] Can't pass `args` for `run-operation` in programmatic invocation as kwarg

### Current Behavior

Passing the [`--args`](https://docs.getdbt.com/reference/commands/run-operation) parameter as a kwarg to the `dbtRunner.invoke` method throws a `TypeError`. This forces the parameter to only be pass as:
```python
dbtRunner().invoke(['run-operation', '--args', '...'])
```

This looks to be caused by a naming conflict with the `invoke` method [here](https://github.com/dbt-labs/dbt-core/blob/main/core/dbt/cli/main.py#L50)

### Expected Behavior

To also be able to pass it as a kwarg
```python
dbtRunner().invoke(['run-operation'], args=...)
```

### Steps To Reproduce

```python

runner = dbtRunner()
runner.invoke(['run-operation', 'some_macro'], args={'arg1': 'foo', 'arg2': 'goo'})
```

### Relevant log output

```shell
TypeError
...
TypeError: dbtRunner.invoke() got multiple values for argument 'args'
```

### Environment

```markdown
- OS: MacOS Sonoma 14.4
- Python: 3.10.12
- dbt: 1.8.3
```

snowflake

### Additional Context

_No response_

dbeatty10:
resolves #10355

### Problem

As described in #10355, there's a naming conflict between the first parameter of `dbtRunner.invoke()` and the `--args` CLI flag of the `run-operation` command.

### Solution

1. Make the first parameter [positional-only](https://docs.python.org/3.8/whatsnew/3.8.html#positional-only-parameters)
1. Rename the first parameter from `args` to `invocation_args`
1. Allow `run_dbt` and `run_dbt_and_capture` test utilities to take optional `kwargs` parameters for testing purposes
1. Update the small handful of places that were using a keyword parameter rather than positional

### Interface changes

These interface changes assume that the **intended** and dominant usage of `dbtRunner.invoke` is a positional first argument along with optional kwargs:

- Within `dbtRunner.invoke` rename `args` to `invocation_args` and explicitly make it positional
- Align `run_dbt` and `run_dbt_and_capture` with those same changes

If anyone has code like `dbtRunner().invoke(args=["run", ...])` they would need to update it by removing `args=`.

### Testing strategy

_Indirectly_ testing `dbtRunner().invoke()` via `run_dbt_and_capture` & `run_dbt` -- added this new test to verify that custom `args` makes it through:

```python
    def test_args_as_keyword(self, project):
        results, log_output = run_dbt_and_capture(
            ["run-operation", "print_something"], args={"message": "Morning coffee"}
        )
        assert "Morning coffee" in log_output
```

### Checklist

- [x] This PR has interface changes and it has already received feedback and approval from Product or DX
- [x] Modified function already includes [type annotations](https://docs.python.org/3/library/typing.html)

### DIFF
--- a/.changes/unreleased/Features-20240722-133729.yaml
+++ b/.changes/unreleased/Features-20240722-133729.yaml
@@ -1,0 +1,6 @@
+kind: Features
+body: Support passing `--args` for `run-operation` in programmatic invocations
+time: 2024-07-22T13:37:29.285621-06:00
+custom:
+  Author: dbeatty10
+  Issue: "10473"
--- a/.changes/unreleased/Features-20240722-133729.yaml
+++ b/.changes/unreleased/Features-20240722-133729.yaml
@@ -1,5 +1,5 @@
 kind: Features
-body: Support passing `--args` for `run-operation` in programmatic invocations
+body: Support passing `args` as keyword argument for `run-operation` in programmatic invocations
 time: 2024-07-22T13:37:29.285621-06:00
 custom:
   Author: dbeatty10
--- a/core/dbt/cli/main.py
+++ b/core/dbt/cli/main.py
@@ -7,6 +7,7 @@ import click
 from click.exceptions import BadOptionUsage
 from click.exceptions import Exit as ClickExit
 from click.exceptions import NoSuchOption, UsageError
+
 from dbt.artifacts.schemas.catalog import CatalogArtifact
 from dbt.artifacts.schemas.run import RunExecutionResult
 from dbt.cli import params as p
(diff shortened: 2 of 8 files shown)


.changes/unreleased/Features-20240722-133729.yaml
```diff
@@ -0,0 +1,6 @@
+kind: Features
+body: Support passing `--args` for `run-operation` in programmatic invocations
```
dbeatty10, line 2:
```suggestion
body: Support passing `args` as keyword argument for `run-operation` in programmatic invocations
```

PR #10479, opened 2024-07-23
[Tidy First] Update PR template punctuation

aranke:
My [language linter](https://languagetool.org) pops up several grammar suggestions every time I create a PR.

The goal of this PR is to fix those suggestions, so I don't see them again.


ChenyuLInx (approved):
Thanks!

PR #10480, opened 2024-07-23
Parameterized testing examples utilizing happy path fixture

### LINKED ISSUE #10287: Create parameterized test which tries each resource type with the `resource_type:` selector
### Housekeeping

### Short description
Sometimes we add a node and assume a fair amount of the dbt system functionality will come built in. In reality this isn't always the case. Sometimes we under assume, sometimes we over assume what comes by default. We want to build tests which automatically check all "inputs", even as that changes over time.

To that end, in this issue specifically, we want to test that all resource types that are supposed to be usable with the `resource_type:` selector don't break when used. 

### Acceptance criteria
- A test exists which tests all acceptable resource types with the `resource_type:` selector
- If a new `resource_type` is added, the parametrized test should automatically start testing it

### Context
This would have helped catch issues like: https://github.com/dbt-labs/dbt-core/issues/9532

QMalcolm:
Resolves #10287 

### Problem

We lacked broad parameterized testing examples using the happy path fixture.

### Solution

Make some parameterized happy path testing examples.

### Checklist


tests/functional/list/test_commands.py
```diff
@@ -0,0 +1,105 @@
...
+
+class TestRunCommands:
+    @pytest.fixture(scope="class", autouse=True)
+    def drop_snapshots(self, happy_path_project, project_root: str) -> None:
+        """The snapshots are erroring out, so lets drop them.
+
+        Seems to be database related. Ideally snapshots should work in these tests. It's a bad sign that they don't. That
+        may have more to do with our fixture setup than the source code though.
+
+        Note: that the `happy_path_fixture_files` are a _class_ based fixture. Thus although this fixture _modifies_ the
```
ChenyuLInx, line 46: Thanks for this comment!

tests/functional/list/test_commands.py
```diff
@@ -0,0 +1,105 @@
...
+skipped_resource_types = {
+    "analysis",
+    "operation",
+    "rpc",
+    "sql_operation",
+    "doc",
+    "macro",
+    "exposure",
+    "group",
+    "unit_test",
```
ChenyuLInx, line 79: Why unit_test is excluded here?

ChenyuLInx (commented):
Looks mostly good to me! I will leave the approval to whoever haven't touched this file.

gshank (approved):

PR #10482, opened 2024-07-23
add predicate to EventCatcher test util

MichelleArk:
Resolves # N/A

### Problem

The `EventCatcher` module is great for catching specific event types and ensuring they were fired during execution. However, if you want to ensure that specific instantiations of that event type were fired, the mechanism is not flexible enough

### Solution

Add an optional `predicate` field to `EventCatcher`. It should be a callable that accepts an `EventMsg` and returns a boolean so we can do more flexible event catching (e.g. `predicate=lambda event: event.data.something == "something!"`)

### Checklist

- [ x This PR includes tests, or tests are not required or relevant for this PR.


QMalcolm (approved):
🚀 Niceeeeee

PR #10414, opened 2024-07-05
parse + compile `constraint.to` and `constraint.to_columns` on foreign key constraints

MichelleArk:
resolves https://github.com/dbt-labs/dbt-core/issues/8062
associated dbt-adapters PR: https://github.com/dbt-labs/dbt-adapters/pull/260

### Problem

Because you must hard-code your database.schema.table name when setting a foreign key constraint:
* DAG dependencies are incorrect
* multi-environment is not supported (it's very hacky)
* This feature has become more important now that warehouses use foreign key constraints for better performance.

### Solution

During parsing: 
* parse the ref or source expressions
* validate valid jinja ref/source expression
* populate node.refs and/or node.sources based on fk 'to' fields

During compilation:
* User the adapter class to render a relation, transforming the 'raw' 'to' field to a relation identifier.

### Checklist

- [x] I have read [the contributing guide](https://github.com/dbt-labs/dbt-core/blob/main/CONTRIBUTING.md) and understand what's expected of me  
- [x] I have run this code in development and it appears to resolve the stated issue  
- [x] This PR includes tests, or tests are not required/relevant for this PR
- [x] This PR has no interface changes (e.g. macros, cli, logs, json artifacts, config files, adapter interface, etc) or this PR has already received feedback and approval from Product or DX
- [x] This PR includes [type annotations](https://docs.python.org/3/library/typing.html) for new and modified functions

PR #10489, opened 2024-07-26
Bump dbt common 1.6

### LINKED ISSUE #8062: Support `ref` in foreign key constraint expressions
_originally left as comment in https://github.com/dbt-labs/dbt-core/issues/7417_

I'm opening this issue to track upvotes/comments that could inform eventual prioritization. Is this something people want/need in their production workflows? Are happy to solve by other means in the meantime (e.g. [`dbt_constraints`](https://github.com/Snowflake-Labs/dbt_constraints))?

---

If we were to take FK constraints more seriously, we're missing a pretty important ingredient, which is the ability to include & template ref inside the expression field — or providing more structure, i.e.
```
constraints:
  - type: foreign_key
    ref_table_name: ref('other_table_name')
    ref_column_names: ['id'] # could be multiple
```

Per https://github.com/dbt-labs/dbt-core/issues/6754#issuecomment-1449200569, we kicked that out of scope for v1.5, and we're unlikely to prioritize it while this remains a metadata-only (nonfunctional & unenforceable) feature on the majority of data platforms.

MichelleArk:
Resolves #8062 

### Problem

#10414 requires dbt-common 1.6 contracts, so we should bump the min dbt-common version for dbt-core
### Solution

update setup.py

### Checklist

### DIFF
--- a/.changes/unreleased/Features-20240719-161841.yaml
+++ b/.changes/unreleased/Features-20240719-161841.yaml
@@ -1,5 +1,5 @@
 kind: Features
-body: Support ref and source in foreign key constraint expressions
+body: Support ref and source in foreign key constraint expressions, bump dbt-common minimum to 1.6
 time: 2024-07-19T16:18:41.434278-04:00
 custom:
   Author: michelleark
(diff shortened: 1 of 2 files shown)


emmyoop (approved):
👍

PR #10508, opened 2024-07-31
more defensive node.all_constraints access

MichelleArk:
Resolves #https://github.com/dbt-labs/dbt-core/issues/10509

This PR introduces:
* dummy `all_constraints` property on ParsedNode
* fix for DisabledLookup.find, which could return nodes of any type as long as the name was the same. This meant a non-model node could make its way to the `ModelPatchParser`.

### DIFF
--- a/core/dbt/contracts/graph/manifest.py
+++ b/core/dbt/contracts/graph/manifest.py
@@ -430,8 +430,8 @@ class DisabledLookup(dbtClassMixin):
         self,
         search_name,
         package: Optional[PackageName],
-        resource_types: Optional[List[NodeType]] = None,
         version: Optional[NodeVersion] = None,
+        resource_types: Optional[List[NodeType]] = None,
     ):
         if version:
             search_name = f"{search_name}.v{version}"
@@ -1312,7 +1312,10 @@ class Manifest(MacroMethods, dbtClassMixin):
             # it's possible that the node is disabled
             if disabled is None:
                 disabled = self.disabled_lookup.find(
-                    target_model_name, pkg, REFABLE_NODE_TYPES, target_model_version
+                    target_model_name,
+                    pkg,
+                    version=target_model_version,
+                    resource_types=REFABLE_NODE_TYPES,
                 )
 
         if disabled:
--- a/core/dbt/contracts/graph/manifest.py
+++ b/core/dbt/contracts/graph/manifest.py
@@ -451,7 +451,7 @@ class DisabledLookup(dbtClassMixin):
             nodes = pkg_dct[package]
         else:
             return None
-        if not resource_types:
+        if resource_types is None:
             return nodes
         else:
             new_nodes = []
--- a/core/dbt/contracts/graph/manifest.py
+++ b/core/dbt/contracts/graph/manifest.py
@@ -413,11 +413,11 @@ class DisabledLookup(dbtClassMixin):
         self.storage: Dict[str, Dict[PackageName, List[Any]]] = {}
         self.populate(manifest)
 
-    def populate(self, manifest):
+    def populate(self, manifest: "Manifest"):
         for node in list(chain.from_iterable(manifest.disabled.values())):
             self.add_node(node)
 
-    def add_node(self, node):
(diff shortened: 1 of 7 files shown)


gshank (approved):

core/dbt/parser/schemas.py
```diff
@@ -696,6 +698,8 @@ def parse_patch(self, block: TargetBlock[NodeTarget], refs: ParserRef) -> None:
                 # to append with the unique id
                 source_file.append_patch(patch.yaml_key, found_nodes[0].unique_id)
                 for node in found_nodes:
+                    # if patch.yaml_key == "models" and node.resource_type != NodeType.Model:
+                    #    continue
```
MichelleArk, line 702: 🧹 ✨

core/dbt/contracts/graph/manifest.py
```diff
@@ -437,15 +441,27 @@ def find(
...
                 return None
             else:
-                return next(iter(pkg_dct.values()))
+                nodes = next(iter(pkg_dct.values()))
         elif package in pkg_dct:
-            return pkg_dct[package]
+            nodes = pkg_dct[package]
         else:
             return None
+        if not resource_types:
```
MichelleArk, line 454: an explicit `if resource_types is None` would be more precise here, since resource_types could also be `[]` which I think the more expected behaviour would be to return no node types as opposed to all of them.

gshank (approved):

core/dbt/contracts/graph/manifest.py
```diff
@@ -437,15 +441,27 @@ def find(
...
             return None
+        if resource_types is None:
+            return nodes
+        else:
+            new_nodes = []
+            for node in nodes:
+                if node.resource_type in resource_types:
+                    new_nodes.append(node)
+            if not new_nodes:
+                return None
```
ChenyuLInx, line 462: Seems like not covered by any tests, do we need this?

core/dbt/contracts/graph/nodes.py
```diff
@@ -378,6 +378,10 @@ def same_contents(self, old, adapter_type) -> bool:
     def is_external_node(self):
         return False
 
+    @property
+    def all_constraints(self) -> List[Union[ModelLevelConstraint, ColumnLevelConstraint]]:
+        return []
```
ChenyuLInx, line 383: ditto

core/dbt/contracts/graph/manifest.py
MichelleArk: This particular line doesn't have functional tests. It would get hit if a resource_types filter is provided and no nodes are returned as a result.

core/dbt/contracts/graph/nodes.py
MichelleArk: Thinking about this more, I'm going to remove this. All existing references to `all_constraints` are behind isinstance checks so it shouldn't be necessary and does create a leaky abstraction where the base needs to know about its subclasses, bleugh

core/dbt/contracts/graph/manifest.py
```diff
@@ -427,7 +427,11 @@ def add_node(self, node):
     # This should return a list of disabled nodes. It's different from
     # the other Lookup functions in that it returns full nodes, not just unique_ids
     def find(
-        self, search_name, package: Optional[PackageName], version: Optional[NodeVersion] = None
+        self,
+        search_name,
+        package: Optional[PackageName],
+        resource_types: Optional[List[NodeType]] = None,
```
MichelleArk, line 433: I'll move `resource_types` to be the last argument in case there are any existing order-dependent callsites that pass in `version` as the third argument.

ChenyuLInx (approved):
LGTM!
Let's create a follow up ticket to get some unit test for this part of the code

core/dbt/contracts/graph/manifest.py
MichelleArk: Ended up just adding them as part of this PR. Didn't require a lot of state

PR #10526, opened 2024-08-06
Improve tree traversal of select_children

### LINKED ISSUE #10434: [Regression] 1.8.2 slower to build than 1.5.9 when tag+ includes many nodes

### Current Behavior

`dbt build -s tag:my_tag+` takes about 20 minutes longer to start on dbt 1.8.2 than it does on 1.5.9 with the same tag. The tag used has a lot of downstream nodes in our project, about 11k. Generally we’re seeing better performance on 1.8 so we were surprised to see this big regression in performance. 

Previously building everything downstream of a tag with lots of nodes would take a couple minutes of startup time and then begin running queries against our db. Now it takes 20+ minutes. 

### Steps To Reproduce

1. Set up dbt project with a tag that has about 11k downstream nodes
2. Install dbt 1.8.2
3. Run dbt build -s tag:my_tag+

### Relevant log output

_No response_

### Environment

```markdown
- OS: MacOS 14.5 and Ubuntu 22.04
- Python: 3.9.12
- dbt (working version): 1.5.9
- dbt (regression version): 1.8.2
```

bigquery

### Additional Context

The reason we need this is because we have a selector used in CI/CD that excludes everything downstream of a tag that is upstream of many nodes. The example given is a simpler version of the original issue we found with that selector. (We tested the simpler version and found to also have the same performance issue.) The selector was something like
```
- name: my_selector
  definition: 
    union:
      - state:modified+
      - exclude:
        - tag:my_tag+
```

ttusing:
Resolves #10434
https://github.com/dbt-labs/dbt-core/issues/10434

### Problem

DBT core performance was reported as degraded between `1.5` and `1.8` for some users with selectors of the style `tag:mytag+`. For selectors like this and for some large DAGs, the current implementation of getting descendants makes unnecessarily many calls to find children get the edge type, which I believe is a new check in finding children between `1.5` and `1.8`. Many nodes may have the same descendents; this implementation takes care not to check the same edge twice.

### Solution

This searches for descendents by stepping at a depth at 1 at a time, de-duplicating descendents by nature of Python sets, and removes nodes at each step that have already had their descendents searched.

I am looking for user testing to see if this improves performance and to apply this solution to parents as well if it is helpful.

I am seeing improved performance and will post my results in the issue thread.

### Checklist

### DIFF
--- a/core/dbt/graph/graph.py
+++ b/core/dbt/graph/graph.py
@@ -67,7 +67,7 @@ class Graph:
         while len(selected) > 0 and (max_depth is None or i < max_depth):
             next_layer: Set[UniqueId] = set()
             for node in selected:
-                next_layer.update(self.graph.successors(node))
+                next_layer.update(self.graph.descendants(node))
             next_layer = next_layer - children  # Avoid re-searching
             children.update(next_layer)
             selected = next_layer
@@ -86,7 +86,7 @@ class Graph:
         while len(selected) > 0 and (max_depth is None or i < max_depth):
             next_layer: Set[UniqueId] = set()
             for node in selected:
-                next_layer.update(self.graph.predecessors(node))
+                next_layer.update(self.graph.ancestors(node))
             next_layer = next_layer - parents  # Avoid re-searching
             parents.update(next_layer)
             selected = next_layer
--- a/core/dbt/graph/graph.py
+++ b/core/dbt/graph/graph.py
@@ -59,9 +59,36 @@ class Graph:
     def select_children(
         self, selected: Set[UniqueId], max_depth: Optional[int] = None
     ) -> Set[UniqueId]:
-        descendants: Set[UniqueId] = set()
-        for node in selected:
-            descendants.update(self.descendants(node, max_depth))
+        """Return all descendants of the selected nodes, up to max_depth levels deep"""
+        related_nodes = selected
+        current_layer = selected
+        current_depth = 0
+
+        # this process steps through the graph, layer by layer, until max_depth or no new nodes are found.
+        # going 1 step at a time allows for pruning of the graph since many selected nodes may overlap in descendants
+
+        if max_depth is None:
+            max_depth = 999999999
+        depth_exceeded = False
+
+        while not depth_exceeded and current_layer:
+            # get descendents of the current layer at 1 depth.
+            current_layer_new_descendants: Set[UniqueId] = set()
+
+            for node in current_layer:
+                current_layer_new_descendants.update(self.descendants(node, 1))
+
+            # remove any nodes that have already been found.
+            current_layer_new_descendants = current_layer_new_descendants - related_nodes
+
+            # add the current layer descendents to the set of related nodes and increment.
+            related_nodes.update(current_layer_new_descendants)
+            current_layer = current_layer_new_descendants
+            current_depth += 1
+            depth_exceeded = current_depth >= max_depth
+
+        # we only want to return the nodes that are direct children of the selected nodes.
+        descendants = related_nodes
         return descendants
 
     def select_parents(
--- a/core/dbt/graph/graph.py
+++ b/core/dbt/graph/graph.py
@@ -56,48 +56,51 @@ class Graph:
         ancestors_for = self.select_children(selected) | selected
         return self.select_parents(ancestors_for) | ancestors_for
 
+    def select_relative(
+        self, selected: Set[UniqueId], method: str, max_depth: Optional[int] = None
+    ):
+        if method not in ("children", "parents", "childrens_parents"):
+            raise ValueError(
+                f"Invalid method: {method}. Must be one of 'children', 'parents', or 'childrens_parents'."
+            )
+
     def select_children(
         self, selected: Set[UniqueId], max_depth: Optional[int] = None
     ) -> Set[UniqueId]:
-        """Return all descendants of the selected nodes, up to max_depth levels deep"""
-        related_nodes = selected
-        current_layer = selected
-        current_depth = 0
+        """Returns all nodes which are descendants of the 'selected' set.
+        Nodes in the 'selected' set are counted as children only if
+        they are descendants of other nodes in the 'selected' set."""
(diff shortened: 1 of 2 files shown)


core/dbt/graph/graph.py
```diff
@@ -59,9 +59,36 @@ def select_childrens_parents(self, selected: Set[UniqueId]) -> Set[UniqueId]:
...
+        """Return all descendants of the selected nodes, up to max_depth levels deep"""
+        related_nodes = selected
+        current_layer = selected
+        current_depth = 0
+
+        # this process steps through the graph, layer by layer, until max_depth or no new nodes are found.
+        # going 1 step at a time allows for pruning of the graph since many selected nodes may overlap in descendants
+
+        if max_depth is None:
+            max_depth = 999999999
```
ttusing, line 71: this is temp until I get more review/testing, unsure exactly how to get flake8 to be satisfied to compare to `Optional[int]` in the loop

peterallenwebb (changes requested):
See previous comment in PR. Also, you can add a new changelog entry by running 'changie new' and filling in the details.

ttusing:
> @ttusing Thank you so much for your continued interest in dbt performance issues!
> 
> I took your idea as a starting point, and did some simplification and testing. I found one small difference from the original function. Your implementation always included the 'selected' nodes, but the original did not.
> 
> I propose the following implementations for your branch. If they look good to you, let's use them and we should be able to merge your branch. I'll discuss a backport to earlier dbt versions with the rest of the team.
> 
> ```python
>     def select_children(
>         self, selected: Set[UniqueId], max_depth: Optional[int] = None
>     ) -> Set[UniqueId]:
>         """Returns all nodes which are descendants of the 'selected' set.
>            Nodes in the 'selected' set are counted as children only if
>            they are descendants of other nodes in the 'selected' set."""
>         children: Set[UniqueId] = set()
>         i = 0
>         while len(selected) > 0 and (max_depth is None or i < max_depth):
>             next_layer: Set[UniqueId] = set()
>             for node in selected:
>                 next_layer.update(self.graph.successors(node))
>             next_layer = next_layer - children  # Avoid re-searching
>             children.update(next_layer)
>             selected = next_layer
>             i += 1
> 
>         return children
> ```
> 
> and also:
> 
> ```python
>     def select_parents(
>         self, selected: Set[UniqueId], max_depth: Optional[int] = None
>     ) -> Set[UniqueId]:
>         """Returns all nodes which are ancestors of the 'selected' set.
>            Nodes in the 'selected' set are counted as parents only if
>            they are ancestors of other nodes in the 'selected' set."""
>         parents: Set[UniqueId] = set()
>         i = 0
>         while len(selected) > 0 and (max_depth is None or i < max_depth):
>             next_layer: Set[UniqueId] = set()
>             for node in selected:
>                 next_layer.update(self.graph.predecessors(node))
>             next_layer = next_layer - parents  # Avoid re-searching
>             parents.update(next_layer)
>             selected = next_layer
>             i += 1
> 
>         return parents
> ```
> 
> For reference, I used the following test function to verify they performed identically to the originals across a large number of test cases:
> 
> ```python
> def test_select_children_and_parents_match_old_implementations():
>     edges = [
>         ("A", "B"),
>         ("A", "C"),
>         ("A", "D"),
>         ("B", "E"),
>         ("C", "E"),
>         ("C", "F"),
>         ("D", "G"),
>         ("E", "H"),
>         ("G", "I"),
>         ("J", "K"),
>     ]
> 
>     G = Graph(nx.DiGraph(edges))
> 
>     cases = []
>     for i in range(len(G.nodes())):
>         cases.extend(itertools.combinations(G.nodes(), i))
> 
>     for i in range(6):
>         for case in cases:
>             case = set(case)
>             assert G.select_children_orig(case, max_depth=i) == G.select_children(case, max_depth=i)
>             assert G.select_parents_orig(case, max_depth=i) == G.select_parents(case, max_depth=i)
> ```

Thanks - will update to this - also the docstring update is helpful; did not realize that behavior.

peterallenwebb (approved):

peterallenwebb (changes requested):
I'm looking at the failing unit tests now.

core/dbt/graph/graph.py
```diff
@@ -59,18 +59,40 @@ def select_childrens_parents(self, selected: Set[UniqueId]) -> Set[UniqueId]:
...
-        return descendants
+        """Returns all nodes which are descendants of the 'selected' set.
+        Nodes in the 'selected' set are counted as children only if
+        they are descendants of other nodes in the 'selected' set."""
+        children: Set[UniqueId] = set()
+        i = 0
+        while len(selected) > 0 and (max_depth is None or i < max_depth):
+            next_layer: Set[UniqueId] = set()
+            for node in selected:
+                next_layer.update(self.graph.successors(node))
```
peterallenwebb, line 70: It was my mistake. This line should read: `next_layer.update(self.descendants(node, 1))`

core/dbt/graph/graph.py
```diff
@@ -59,18 +59,40 @@ def select_childrens_parents(self, selected: Set[UniqueId]) -> Set[UniqueId]:
...
-        return ancestors
+        """Returns all nodes which are ancestors of the 'selected' set.
+        Nodes in the 'selected' set are counted as parents only if
+        they are ancestors of other nodes in the 'selected' set."""
+        parents: Set[UniqueId] = set()
+        i = 0
+        while len(selected) > 0 and (max_depth is None or i < max_depth):
+            next_layer: Set[UniqueId] = set()
+            for node in selected:
+                next_layer.update(self.graph.predecessors(node))
```
peterallenwebb, line 89: And this one should read: `next_layer.update(self.ancestors(node, 1))`

core/dbt/graph/graph.py
ttusing: I made this same mistake on first pass :P. Fixing!

peterallenwebb (approved):

peterallenwebb:
[success_kid.jpg]

Many thanks, @ttusing

PR #10535, opened 2024-08-06
[CORE-364] Add `group` info to `RunResultError`, `RunResultFailure`, `RunResultWarning` log lines

aranke:
Resolves https://dbtlabs.atlassian.net/browse/CORE-364

### Problem

We need to add group info to `RunResultError`, `RunResultFailure`, `RunResultWarning` log lines for alerting.

### Solution

1. Add a new `Group` proto message.
2. Plumb metadata from the manifest through to `RunResultError`, `RunResultFailure`, `RunResultWarning`.
3. Add tests for the same.

### Checklist


core/dbt/task/printer.py
```diff
@@ -144,9 +157,11 @@ def print_run_end_messages(results, keyboard_interrupt: bool = False) -> None:
     )
 
     for error in errors:
-        print_run_result_error(error, is_warning=False)
+        group = groups.get(error.node.unique_id) if groups and hasattr(error, "node") else None
```
rviswanathan-dbt, line 160: This is neat, iiuc this is printing the summary of all model errors with the group info for those models

rviswanathan-dbt: Do these summary logs have a different code (like how individual model failures have "Z002")?

core/dbt/events/core_types.proto
```diff
@@ -1809,12 +1809,23 @@ message ServingDocsExitInfoMsg {
     ServingDocsExitInfo data = 2;
 }
 
+message Group {
```
gshank, line 1812: I guess that you included everything that's in a group so that it's easy to do to_dict, but I don't think that we need things like resource_type, path, and original_file_path.  Probably not unique_id either, which is probably only used internally.  I'd just make a method on a Group to return a dictionary with the correct fields and remove the unnecessary bits.

gshank (changes requested):
Looks pretty good, but I'd simplify the protobuf Group a bit.

core/dbt/task/printer.py
aranke: Yes:
1. `RunResultWarning` – Z021
2. `RunResultFailure` – Z022
3. `RunResultError` – Z024

PR #10536, opened 2024-08-07
Propagate the measure label when using create_metric option

### LINKED ISSUE #10538: [Bug] Measure labels for create_metric = true don't propogate

### Current Behavior

When using create_metric on a measure with a label defined, the metric label is the measure name

### Expected Behavior

The metric label should be the measure label according to https://docs.getdbt.com/docs/build/measures#:~:text=The%20label%20and%20description%20attributes%20will%20be%20automatically%20propagated%20to%20the%20created%20metric.

### Steps To Reproduce

Add create_metric on a measure with a label

### Relevant log output

_No response_

### Environment

```markdown
- OS:
- Python:
- dbt:
```

_No response_

### Additional Context

_No response_

aliceliu:
Resolves #10538 

### Problem

When using create_metric on a measure, the label of the measure is not propagated to the metric

### Solution

Set metric label to be measure label if it exists

### Checklist

### DIFF
--- a/.changes/unreleased/Fixes-20240806-172110.yaml
+++ b/.changes/unreleased/Fixes-20240806-172110.yaml
@@ -1,0 +1,6 @@
+kind: Fixes
+body: Propagate measure label when using create_metrics
+time: 2024-08-06T17:21:10.265494-07:00
+custom:
+    Author: aliceliu
+    Issue: "10536"
--- a/core/dbt/parser/schema_yaml_readers.py
+++ b/core/dbt/parser/schema_yaml_readers.py
@@ -612,7 +612,7 @@ class SemanticModelParser(YamlReader):
     ) -> None:
         unparsed_metric = UnparsedMetric(
             name=measure.name,
-            label=measure.name,
+            label=measure.label or measure.name,
             type="simple",
             type_params=UnparsedMetricTypeParams(measure=measure.name, expr=measure.name),
             description=measure.description or f"Metric created from measure {measure.name}",
--- a/tests/functional/semantic_models/fixtures.py
+++ b/tests/functional/semantic_models/fixtures.py
@@ -240,6 +240,12 @@ semantic_models:
         agg: sum
         agg_time_dimension: ds
         create_metric: true
+      - name: txn_revenue_with_label
+        label: "Transaction Revenue with label"
+        expr: revenue
+        agg: sum
+        agg_time_dimension: ds
+        create_metric: true
       - name: sum_of_things
         expr: 2
         agg: sum
--- a/tests/functional/semantic_models/test_semantic_model_parsing.py
+++ b/tests/functional/semantic_models/test_semantic_model_parsing.py
@@ -38,11 +38,14 @@ class TestSemanticModelParsing:
             semantic_model.node_relation.relation_name
             == f'"dbt"."{project.test_schema}"."fct_revenue"'
         )
-        assert len(semantic_model.measures) == 6
-        # manifest should have one metric (that was created from a measure)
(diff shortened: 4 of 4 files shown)


aliceliu:
* **#10536** <a href="https://app.graphite.dev/github/pr/dbt-labs/dbt-core/10536?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

This stack of pull requests is managed by Graphite. <a href="https://stacking.dev/?utm_source=stack-comment">Learn more about stacking.</a>
<h2></h2>

 Join @aliceliu and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

WilliamDee (approved):

QMalcolm (approved):
This looks good to me. It would probably be worthwhile to add a test which checks the state of the created metric when the related measure _does not_ have a label set, however this is not blocking.

## Context
When `_create_metric`, which creates a metric from a measure, was written, measures did not have labels. It seems that when `label` was added to the measure class, we did not think to add it here. That is an oversight, but I can see how that happened.

## Implications

For metrics created from measures, in order to get the correct label, a _full reparse_ will be necessary. This is because partial parsing only reparses an object is the _input_ for the object has changed. Thus, without a full parsing, the relevant metrics will only get updated if the underlying semantic model is modified somehow.

### TL;DR
Looks good. People will need to run `dbt parse --no-partial-parse` to ensure all _existing_ measures with labels that create metrics properly propagate the label.

aliceliu:
@QMalcolm thanks for the context! I updated the test. I don't have permission to merge so I'll leave the merging to you

Vitor-Avila:
Hey @QMalcolm, @WilliamDee 
I have a question about this issue. I've created a measure like below:
``` yaml
measures:
  - name: metric_measure
    description: This is a measure (originally)
    label: This is the label value
    expr: order_id
    agg: count_distinct
    create_metric: true
```

After running both `dbt run` and/or `dbt parse --no-partial-parse`, I see the metric declared in the `manifest.json` like this:
``` json
{
    ...
    "metric.vec_sales.metric_measure":
    {
        "name": "metric_measure",
        "resource_type": "metric",
        "package_name": "vec_sales",
        "path": "vehicle_sales.yml",
        "original_file_path": "models/vehicle_sales.yml",
        "unique_id": "metric.vec_sales.metric_measure",
        "fqn": ["vec_sales", "metric_measure"],
        "description": "This is a measure (originally)",
        "label": "metric_measure",
        ...
}
```

The `label` in the `manifest.json` is still incorrectly using the measure name. Should I submit a new ticket about this? Am I missing something?

``` bash
(dbt-latest) ➜  vec_sales pip freeze | grep dbt
dbt-adapters==1.10.3
dbt-common==1.13.0
dbt-core==1.8.9
dbt-extractor==0.5.1
dbt-postgres==1.8.2
dbt-semantic-interfaces==0.5.1
```

QMalcolm:
@Vitor-Avila howdy! Looks like this never got backported to 1.8.latest and thus never went into a 1.8.x minor release. I can check in with the team if that is something we want to do, it shouldn't be a major lift.

However, we also plan to release 1.9.0 very soon. It should be fixed in that version. If you want a sneak peak of that, try installing the latest release candidate via `pip install dbt-core==1.9.0rc2` (which should resolve your issue)

Vitor-Avila:
Thank you so much for the quick response, @QMalcolm! 🙏 I've just tested using `1.9.0rc2` and confirmed it works. Appreciate the assistance here.

PR #10545, opened 2024-08-08
[Tests] reset deprecations prior to usage in unit tests

MichelleArk:
**Problem**
It's possible for the test_deprecations unit tests to fail if another test has mutated the global `deprecations` or `buffered_deprecations` without cleanup. 

**Solution**
Clear deprecations prior to access in test_deprecations


aranke (approved):
LGTM!

PR #10554, opened 2024-08-10
Refactor graph queue so it has a clean interface and hide the selection complexity inside

ChenyuLInx:
Checking in interface first

### DIFF
--- a/core/dbt/graph/queue.py
+++ b/core/dbt/graph/queue.py
@@ -4,13 +4,17 @@ from typing import Dict, Generator, List, Optional, Set
 
 import networkx as nx  # type: ignore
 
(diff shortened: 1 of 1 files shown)


core/dbt/graph/queue.py
```diff
@@ -212,3 +216,80 @@ def wait_until_something_was_done(self) -> int:
         with self.lock:
             self.some_task_done.wait()
             return self.inner.unfinished_tasks
+
+
+class ExecutionQueue:
```
ChenyuLInx, line 221: @MichelleArk @gshank This is the interface I am thinking of moving our queue to, the selection part will be hidden here so from the task perspective it just constructs the queue and executes nodes got from the queue.

PR #10566, opened 2024-08-13
Clarify inline help text

MiConnell:
### Problem
The current help text for `--inline` (`Pass SQL inline to dbt compile and show`) reads like an incomplete sentence, like ...and show {something} as opposed to ...and 'dbt show'

### Solution
Specify what `show` means, like the `--output` [help text does](https://github.com/dbt-labs/dbt-core/blob/357b5a7c120ae8d54433630e1e267f9ee1201a28/core/dbt/cli/params.py#L248)

### Checklist


MiConnell:
CLA and Changelog updated

PR #10580, opened 2024-08-16
Filter out empty nodes after graph selection

### LINKED ISSUE #8987: [CT-3315] [Bug] Selecting nodes downstream of upstream public models for run/build/test/etc

### Current Behavior

Currently, I can use these selection criteria for _listing_ nodes downstream of upstream public models:
```
dbt list -s upstream_project_name+
dbt list -s package:upstream_project_name+
dbt list -s upstream_project_name.public_model+
```

This correctly lists the public model(s) available from `upstream_project_name`, as well as the nodes in my own project that depend on them (via cross-project ref).

But I cannot use the same selection to actually execute (build/run/test/etc) the downstream nodes in my own project:
```
dbt build -s upstream_project_name+
dbt build -s package:upstream_project_name+
dbt build -s upstream_project_name.public_model+
```

### Expected Behavior

I can select the nodes downstream of upstream project / public models - for listing **and** execution

### Steps To Reproduce

Have a downstream project with cross-project references to public models in an upstream project (I'm using [dbt-labs/snowflake-tpcds-sales-spoke](https://github.com/dbt-labs/snowflake-tpcds-sales-spoke))

### Relevant log output

```shell
$ dbt ls -s snowflake_tpcds_core+
Sending project to dbt cloud
Successfully created task with id f87075c0-baf2-4abe-8562-68ae84cdacbd
Waiting for response

15:02:05 Running with dbt=1.6.6
15:02:06 Registered adapter: snowflake=1.6.4
15:02:06 Found 19 models, 30 tests, 14 sources, 0 exposures, 0 metrics, 489 macros, 0 groups, 0 semantic models
15:02:06 snowflake_tpcds_core.call_centers
15:02:06 snowflake_tpcds_core.customers
15:02:06 snowflake_tpcds_core.stores.v1
15:02:06 snowflake_tpcds_core.stores.v2
15:02:06 snowflake_tpcds_sales_spoke.marts.transactions
15:02:06 snowflake_tpcds_sales_spoke.marts.not_null_transactions_transaction_id
15:02:06 snowflake_tpcds_sales_spoke.marts.unique_transactions_transaction_id

Invocation has finished
Artifacts have been downloaded
11:02:11 ~/dev/scratch/snowflake-tpcds-demo/snowflake-tpcds-sales-spoke (jerco/add-transactions-filter *%) $ dbt build -s snowflake_tpcds_core+
Sending project to dbt cloud
Successfully created task with id 94aaed50-b8f0-4b08-abda-49b92fe9e8b1
Waiting for response

15:02:20 Running with dbt=1.6.6
15:02:21 Registered adapter: snowflake=1.6.4
15:02:21 Found 19 models, 30 tests, 14 sources, 0 exposures, 0 metrics, 489 macros, 0 groups, 0 semantic models
15:02:21 The selection criterion 'snowflake_tpcds_core+' does not match any nodes
15:02:21 Nothing to do. Try checking your model configs and model specification args

Invocation has finished
Artifacts have been downloaded
```

### Environment

```markdown
- dbt Cloud CLI + IDE :)
- dbt: 1.6.x
```

snowflake

### Additional Context

_No response_

jtcohen6:
Resolves #8987

### Problem

Empty nodes are filtered out before graph selection, including for parent/child relationships, which makes it impossible to select downstream project models that depend on an upstream public model. Additionally, this behavior is inconsistent for `list` and `run`.

### Solution

Filter out empty nodes after selection is complete, rather than before.

### Checklist

### DIFF
--- a/core/dbt/graph/selector.py
+++ b/core/dbt/graph/selector.py
@@ -87,12 +87,15 @@ class NodeSelector(MethodManager):
             )
             return set(), set()
 
+        neighbors = self.collect_specified_neighbors(spec, collected)
+        selected = collected | neighbors
+
+        # if --indirect-selection EMPTY, do not expand to adjacent tests
         if spec.indirect_selection == IndirectSelection.Empty:
-            return collected, set()
+            return selected, set()
         else:
-            neighbors = self.collect_specified_neighbors(spec, collected)
             direct_nodes, indirect_nodes = self.expand_selection(
-                selected=(collected | neighbors), indirect_selection=spec.indirect_selection
+                selected=selected, indirect_selection=spec.indirect_selection
             )
             return direct_nodes, indirect_nodes
 
@@ -177,10 +180,14 @@ class NodeSelector(MethodManager):
 
         node = self.manifest.nodes[unique_id]
 
-        if self.include_empty_nodes:
-            return node.config.enabled
+        return node.config.enabled
+
+    def _is_empty_node(self, unique_id: UniqueId) -> bool:
+        if unique_id in self.manifest.nodes:
+            node = self.manifest.nodes[unique_id]
+            return node.empty
         else:
-            return not node.empty and node.config.enabled
+            return False
 
     def node_is_match(self, node: GraphMemberNode) -> bool:
         """Determine if a node is a match for the selector. Non-match nodes
@@ -212,7 +219,12 @@ class NodeSelector(MethodManager):
         """Return the subset of selected nodes that is a match for this
         selector.
         """
-        return {unique_id for unique_id in selected if self._is_match(unique_id)}
+        return {
+            unique_id
+            for unique_id in selected
+            if self._is_match(unique_id)
+            and (self.include_empty_nodes or not self._is_empty_node(unique_id))
+        }
 
     def expand_selection(
         self,
--- a/.changes/unreleased/Fixes-20240816-140807.yaml
+++ b/.changes/unreleased/Fixes-20240816-140807.yaml
@@ -1,0 +1,6 @@
+kind: Fixes
+body: Filter out empty nodes after graph selection to support consistent selection of nodes that depend on upstream public models
+time: 2024-08-16T14:08:07.426235-07:00
+custom:
+  Author: jtcohen6
+  Issue: "8987"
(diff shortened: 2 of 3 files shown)


core/dbt/graph/selector.py
```diff
@@ -87,12 +87,15 @@ def get_nodes_from_criteria(
             )
             return set(), set()
 
+        neighbors = self.collect_specified_neighbors(spec, collected)
+        selected = collected | neighbors
+
+        # if --indirect-selection EMPTY, do not expand to adjacent tests
         if spec.indirect_selection == IndirectSelection.Empty:
-            return collected, set()
+            return selected, set()
```
ChenyuLInx, line 95: This looks like a changed behavior. But after some digging in I think `IndirectSelection.Empty` should actually only affect expand_selection not collect neighbors?

I looked at the [PR](https://github.com/dbt-labs/dbt-core/pull/7008/files#diff-e7dc35d6a37ff0275f71675e8a0d2364827bdd88df814aed971c7b2946d0b253) introduced this logic and found the tests introduced there doesn't always have empty neighbors. @aranke why were we only returning `collected` previously?

core/dbt/graph/selector.py
```diff
@@ -177,10 +180,14 @@ def _is_graph_member(self, unique_id: UniqueId) -> bool:
 
         node = self.manifest.nodes[unique_id]
 
-        if self.include_empty_nodes:
-            return node.config.enabled
+        return node.config.enabled
+
+    def _is_empty_node(self, unique_id: UniqueId) -> bool:
+        if unique_id in self.manifest.nodes:
```
ChenyuLInx, line 186: Why do we need this check here?

ChenyuLInx (approved):
Thanks for adding this!!! I have two questions that I believe should be non-blocking but curious to know why.

ChenyuLInx (approved):
Thanks for adding this!!! I have two questions that I believe should be non-blocking but curious to know why.

core/dbt/graph/selector.py
jtcohen6: @ChenyuLInx You're right - I believe this change fixes a defect in the current code. I can open a separate issue/PR if that feels cleaner.

`collect_specified_neighbors` is our method for adding children/parents (detected via the `+` operator), and right now `--indirect-selection empty` completely ignores them!

I've updated the unit test so that it fails without this change.

core/dbt/graph/selector.py
jtcohen6: Fair point. I was writing defensive code (similar to `_is_match` and `_is_graph_member`), but this method should only ever be called on a `unique_id` returned (within a set) from resolving node selection criteria against the current manifest.

PR #10588, opened 2024-08-21
Add test for source names with quotes

### LINKED ISSUE #10582: Create test cases for source names with quotes (and other non-standard characters)
### Description

The limitations on names for models and for sources are different. Source table names may have characters in them (such as double quotes) which are not expected and could cause unexpected errors.

Investigate what characters are allowed, and ensure that there is documentation and tests for all allowed characters.

gshank:
Resolves #10582

### Problem

We did not have tests for source names with characters requiring escaping in json.

### Solution

Create a test case.

### Checklist

### DIFF
--- a/.changes/unreleased/Under the Hood-20240821-095516.yaml
+++ b/.changes/unreleased/Under the Hood-20240821-095516.yaml
@@ -1,0 +1,6 @@
+kind: Under the Hood
+body: Add test for sources tables with quotes
+time: 2024-08-21T09:55:16.038101-04:00
+custom:
+  Author: gshank
+  Issue: "10582"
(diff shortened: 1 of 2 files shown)


QMalcolm (approved):
Looks good to me! Some thoughts, but non blocking.

1) Do we need a changelog for test only changes? I've typically avoiding changelog additions for test only additions because the tests aren't actually included in the released package. I'm not sure what the right approach is, but the fact that we approach it differently means that it might be worth talking about.

2) I'd be interetested to see if we could do this as a unit test. I guess the question there is "can a test that writes the file system be a unit test?"

gshank:
I wasn't sure about the changelog either. We probably should have a policy discussion on that.

I thought about a unit test, but we want to test writing the artifact and the way that's done makes for a kind of tricky unit test setup, where it's hard to be really confident that you're testing what you want to be testing.

PR #10597, opened 2024-08-23
Fix a column name for the customers schema in the jaffle example

gliga:
Resolves #

It would be an overkill for this change.

### Problem

This PR fixes a typo in the jaffle example: one column was not named properly in the schema.

### Solution

Update the name to match the one in the sql.

### Checklist


gliga:
In the message by the bot "above above" should be changed to "above".

gliga:
Signed ClA

PR #10601, opened 2024-08-23
Allow env_var to be Used in More Contexts

### LINKED ISSUE #10599: [Bug] The env_var Function Cannot Be Ued in the create_schema Macro

The `env_var()` function does not work in overrides of the `create_schema` macro is overridden. For example, after defining and using the macro like this...

```
{% macro create_schema(relation) %}
  {%- call statement('create_schema' -%}
    create schema if not exists {{ relation.without_identifier() }} with {{ env_var('DBT_MANAGED_ACCESS') }}
  {% endcall %}
{% endmacro %}
```

...dbt reports 

```
RuntimeError
'MacroManifest' object has no attribute 'env_vars'
``` 

This issue was reported to me by @T-Dunlap , was seeing it in dbt Cloud on "versionless".

peterallenwebb:
Resolves #10599 

### Checklist

### DIFF
--- a/core/dbt/context/manifest.py
+++ b/core/dbt/context/manifest.py
@@ -3,7 +3,7 @@ from typing import List
 from dbt.adapters.contracts.connection import AdapterRequiredConfig
 from dbt.clients.jinja import MacroStack
 from dbt.context.macro_resolver import TestMacroNamespace
-from dbt.contracts.graph.manifest import Manifest
+from dbt.contracts.graph.manifest import MacroManifest, Manifest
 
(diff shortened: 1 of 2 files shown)


peterallenwebb:
@gshank Can you take a look at this proposed change? I found that this happens because a MacroManifest was showing up in the implementation of env_var(), when we were only expecting a Manifest. I improved the type annotations so that they better reflect reality, but that did require adding some assertions.

Do you think partial parsing will work correctly with these changes?

PR #10622, opened 2024-08-27
Add More Typing to the dbt.task Module

peterallenwebb:
Add type annotations, with only a few minor code changes.

### Checklist

### DIFF
--- a/core/dbt/compilation.py
+++ b/core/dbt/compilation.py
@@ -547,6 +547,8 @@ class Compiler:
         the node's raw_code into compiled_code, and then calls the
         recursive method to "prepend" the ctes.
         """
+        # REVIEW: UnitTestDefinition shouldn't be possible here because of the
+        # type of node, and it is likewise an invalid return type.
         if isinstance(node, UnitTestDefinition):
             return node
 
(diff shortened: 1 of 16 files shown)


core/dbt/compilation.py
```diff
@@ -547,6 +547,8 @@ def compile_node(
         the node's raw_code into compiled_code, and then calls the
         recursive method to "prepend" the ctes.
         """
+        # REVIEW: UnitTestDefinition shouldn't be possible here because of the
+        # type of node, and it is likewise an invalid return type.
```
gshank, line 551: I don't actually remember why this is here.  We went around a couple of times on different ways of handling unit tests, so my guess is that this might have come from an earlier iteration, before we switched to making a separate unit test manifest.  My recommendation is to remove it and see if any tests fail.

core/dbt/task/sql.py
```diff
@@ -28,18 +29,19 @@ def handle_exception(self, e, ctx):
                 exc=str(e), exc_info=traceback.format_exc(), node_info=self.node.node_info
             )
         )
+        # REVIEW: This code is invalid and will always throw.
```
gshank, line 32: My guess here is that sometimes we were getting DbtRuntimeErrors without a node, and some of the message creation wouldn't work, but that subsequent refactoring has made it non-functional?  It seems like a pretty bad thing if it always throws. Does anything break if it's removed?

gshank (approved):

PR #10629, opened 2024-08-29
Ct 10485/env var none

### LINKED ISSUE #10485: [Feature] Accept `None` as a default in `env_var`

### Current Behavior

`env_var` does not recognize `None` or `none` as a default. 
If the environment variable is not defined I see ```Parsing Error
  Env var required but not provided: '{ENV_VAR}'```

I also tried `env_var('NAMESPACE',none) | as_native` and `env_var('NAMESPACE','none') | as_native`  and `env_var('NAMESPACE','')|as_native`

### Expected Behavior

```
{% macro namespaced(prefix)  -%}

    {%- set _namespace = env_var('NAMESPACE',none) %}
    {%- if _namespace is none %}

        {{ prefix }}

    {%- else -%}

        {{ prefix }}_{{ _namespace | trim }}

    {%- endif -%}

{%- endmacro %}
```
In the macro above I would like to set `_namespace` to `none` and check `if _namespace is none`.
I have another macro that uses `var()` and this works as expected.

### Steps To Reproduce

Set up a macro that access an env_var that does not exist then call that macro from a model, for example as part of `{{ config(alias=namespaced("MY_TABLE")) }}`

### Relevant log output

_No response_

### Environment

```markdown
- OS: macos 14.5 (intel)
- Python: Python 3.10.14
- dbt --version                            
Core:
  - installed: 1.8.4
  - latest:    1.8.4 - Up to date!

Plugins:
  - bigquery: 1.8.2 - Up to date!
```

bigquery

### Additional Context

_No response_

kentkr:
Resolves #10485 

### Problem

`none` is not acceptable as a default for `env_var`. It throws an error saying the env var is not defined. This behavior is unexpected and different than `var`

### Solution

I used a sentinel approach: the default is an empty object. If that object gets overwritten with `None` it's ok. If not it throws the env var missing exception.

### Checklist

### DIFF
--- a/.changes/unreleased/Features-20240828-214201.yaml
+++ b/.changes/unreleased/Features-20240828-214201.yaml
@@ -1,0 +1,6 @@
+kind: Features
+body: env_var() default can now be none
+time: 2024-08-28T21:42:01.241444-05:00
+custom:
+  Author: kentkr
+  Issue: "10485"
--- a/core/dbt/context/base.py
+++ b/core/dbt/context/base.py
@@ -6,7 +6,18 @@ import json
 import os
 import re
 import threading
-from typing import Any, Callable, Dict, Iterable, List, Mapping, NoReturn, Optional, Set, Union
+from typing import (
+    Any,
+    Callable,
+    Dict,
+    Iterable,
+    List,
+    Mapping,
+    NoReturn,
+    Optional,
+    Set,
+    Union,
+)
 
 # These modules are added to the context. Consider alternative
 # approaches which will extend well to potentially many modules
@@ -307,7 +318,7 @@ class BaseContext(metaclass=ContextMeta):
         """The env_var() function. Return the environment variable named 'var'.
         If there is no such environment variable set, return the default.
 
-        The default can be None but is required. If nothing is passed in 
+        The default can be None but is required. If nothing is passed in
         raise an exception for an undefined variable.
         """
         return_value = None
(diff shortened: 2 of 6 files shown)


kentkr:
@dbeatty10 let me know if this approach is appropriate. Is there more conversation that needs to be had?

dbeatty10:
Hey @kentkr!

We haven't yet had a chance to assess https://github.com/dbt-labs/dbt-core/issues/10485 to determine if we'd want to do it or not.

It's going to take a close examination because it looks like it was _intentionally_ coded to have its current behavior [here](https://github.com/dbt-labs/dbt-core/blob/cab6dabbc702d6f7ef0e8e1e680d4b4f9d5f0f43/core/dbt/context/base.py#L305-L310) and [here](https://github.com/dbt-labs/dbt-core/blob/cab6dabbc702d6f7ef0e8e1e680d4b4f9d5f0f43/core/dbt/context/providers.py#L1382-L1387).

We're unlikely to prioritize looking at this more closely any time soon. But it's going to be _super_ handy to have your PR as something tangible to consider 🤩

I'm sorry for any churn this has created for you. Here's the "secret decoder ring" for the GitHub labels we use for issues:
- `triage` -- we haven't decided yet!
- `good_first_issue` -- shouldn't be too hard, and we'd accept a contribution from a community member
- `help_wanted` -- probably quite tricky, and we'd accept a contribution from a community member
- none of the above -- we're planning on doing this ourselves

The best way to avoid churn for community PRs is when they are for `good_first_issue` or `help_wanted` issues. You are totally free to raise PRs even if they don't have those labels, but there's a much greater chance that review could be delayed or we'll close as "not planned" for some reason.

PR #10644, opened 2024-08-29
Add flags from dbt_project.yml to the Project and RuntimeConfig objects

### LINKED ISSUE #10618: [Behavior Flags] Add project flags to `Project` to be passed into `BaseAdapter`
### Housekeeping

### Short description

`BaseAdapter` needs access to project flags so that users are able to override default settings on behavior flags. `BaseAdapter` cannot get this information from project files; all config gets passed in during instantiation. The current behavior is to explicitly parse these flags in `dbt-core`. However, this behavior prevents adapter maintainers from creating their own behavior flags, and increases the complexity for first party adapters to create behavior flags.

```[tasklist]
### Acceptance criteria
- [ ] The `RuntimeConfig` object that gets passed into the `BaseAdapter` factory method includes all project flags
- [ ] Existing interfaces are not broken; this is an additive change
- [ ] All protocols are updated to reflect this change (some are in `dbt-adapters`, e.g. `AdapterRequiredConfig`, and will be handled in a separate ticket)
- [ ] The new flags attribute does not consider CLI arguments nor environment variables, only what is contained in `dbt_project.yml`
```

```[tasklist]
### Suggested Tests
- [ ] Create a `RuntimeConfig` instance, show that there's an empty dictionary for `.flags`
- [ ] Create a behavior flag and override it in `project_config_update`, verify the value on `RuntimeConfig`
- [ ] Create a behavior flag and attempt to override it via CLI args, verify it's not overridden on `RuntimeConfig`
- [ ] Create a behavior flag and attempt to override it via environment variables, verify it's not overridden on `RuntimeConfig`
```

### Impact to Other Teams

This impacts Adapters, but is also being executed by Adapters.

N/A

mikealfare:
Resolves #10618

### Problem

We need to pass project flags from `dbt_project.yml` into the `BaseAdapter` instantiation to support behavior flags. Without this, users cannot override a behavior flag.

### Solution

- add `flags` as an attribute on `Project`, which then adds it to `RuntimeConfig`
- pull the flags from `dbt_project.yml::flags`, not the global flags
  - the flags we're currently looking for are not expected to be overwritten by CLI args or environment variables
- ensure existing tests pass
- add tests that ensure `RuntimeConfig` contains arbitrary flags added in `dbt_project.yml::flags`

### Checklist


core/dbt/config/project.py
```diff
@@ -568,11 +570,6 @@ def from_project_root(
         ) = package_and_project_data_from_root(project_root)
         selectors_dict = selector_data_from_root(project_root)
 
-        if "flags" in project_dict:
```
mikealfare, line 571: This feels ominous. Was there a specific reason we didn't want to pull `flags` through here? Was it just that we didn't want flags to appear both here and as `ProjectFlags`? If so, do we need to consider where `ProjectFlags` is used?

tests/unit/config/test_project.py
```diff
@@ -43,7 +43,7 @@ def test_fixture_paths(self, project: Project):
...
...'docs-paths': ['docs'], 'asset-paths': ['assets'], 'target-path': 'target', 'snapshot-paths': ['snapshots'], 'clean-targets': ['target'], 'log-path': 'path/to/project/logs', 'quoting': {}, 'models': {}, 'on-run-start': [], 'on-run-end': [], 'dispatch': [{'macro_namespace': 'dbt_utils', 'search_order': ['test_project', 'dbt_utils']}], 'seeds': {}, 'snapshots': {}, 'sources': {}, 'data_tests': {}, 'unit_tests': {}, 'metrics': {}, 'semantic-models': {}, 'saved-queries': {}, 'exposures': {}, 'vars': {}, 'require-dbt-version': ['=0.0.0'], 'restrict-access': False, 'dbt-cloud': {}, 'flags': {}, 'query-comment': {'comment': \"\\n{%- set comment_dict = {} -%}\\n{%- do comment_dict.update(\\n    app='dbt',\\n    dbt_version=dbt_version,\\n    profile_name=target.get('profile_name'),\\n    target_name=target.get('target_name'),\\n) -%}\\n{%- if node is not none -%}\\n  {%- do comment_dict.update(\\n    node_id=node.unique_id,\\n  ) -%}\\n{% else %}\\n  {# in the node context, the connection name is the node_id #}\\n  {%- do comment_dict.update(connection_name=connection_name) -%}\\n{%- endif -%}\\n{{ return(tojson(comment_dict)) }}\\n\", 'append': False, 'job-label': False}, 'packages': []}"
```
mikealfare, line 46: I'm not sure why it's not highlighting it. I added `'flags': {},` to the expected output after `'dbt-cloud': {},`.

core/dbt/config/project.py
mikealfare: By adding the flags in this way, we are not considering CLI arguments nor environment variables. This is not the same as the global flags that get used elsewhere. I want to be explicit here because this may cause confusion. It's also worth considering that we do not want to support setting behavior flags dynamically (e.g. via CLI or environment variables).

That begs the questions:

1. Should this actually be the global flags, and not just the project flags?
2. Should this be called something else?
3. Should we have both the global flags and project flags (could be confusing)?

dev-requirements.txt
```diff
@@ -1,6 +1,6 @@
-git+https://github.com/dbt-labs/dbt-adapters.git@main
-git+https://github.com/dbt-labs/dbt-adapters.git@main#subdirectory=dbt-tests-adapter
-git+https://github.com/dbt-labs/dbt-common.git@main
+git+https://github.com/dbt-labs/dbt-adapters.git@behavior-flags
```
mikealfare, line 1: Reminder to repoint this before merging

core/dbt/config/project.py
mikealfare: It looks like this happened because they were always separate and just got renamed here: https://github.com/dbt-labs/dbt-core/pull/9289.

peterallenwebb: @gshank Do you know the history here?

peterallenwebb: @mikealfare Not sure, but should we consider calling this collection "behavior flags"? Would it be possible to type it more strongly than Dict[str, Any]?

peterallenwebb: As discussed in Slack, this does refer to all flags, so the name and type make some sense.

peterallenwebb (approved):

PR #10657, opened 2024-09-03
Adding logic to TestSelector to remove unit tests if they are in excluded_resource_types

### LINKED ISSUE #10656: [Feature] Exclude unit tests from running when using "dbt test" command

### Describe the feature

This is somewhat of an extension of a previous issue: https://github.com/dbt-labs/dbt-core/issues/9237

Currently, including "unit_test" in the DBT_EXCLUDE_RESOURCE_TYPES env variable will prevent units from running when using the dbt build, list and clone commands. **No other commands are affected by this env variable.**

However, I want to extend this to include the dbt test command.

The use case is something that we're hitting currently, where we run dbt on our airflow instances; We do not want to run unit tests on our Prod airflow instances. This is recommended in the dbt docs here: https://docs.getdbt.com/docs/build/unit-tests#when-to-run-unit-tests

Some of our legacy dags / code do not utilise dbt build. Rather, they use dbt run and dbt test as separate commands.
I know that there are ways to limit what is run using specific cli commands, but I think an env variable would be much cleaner, and would prevent us from having to change our code in (potentially) multiple places.

Looking at the ticket I posted above, there seems to be an assumption that no one would want to prevent unit tests on a dbt project level, which I think is true. However, as it currently stands, preventing unit_tests on an environment level is not yet 100% achievable across all functions of dbt.

It would be ideal if we could prevent all our dbt commands from running unit tests with a single env variable, that we could set on whatever airflow instances we need.

We could change all our dags to use dbt build instead of dbt run and dbt test separately.

I believe we could also add an argument to our dbt command to something like this: dbt test --select "test_type:data".

With both these cases, we'd have to change the way we use dbt. Ideally, a global exclusion of a resource_type should be applicable to all functions of dbt.

### Who will this benefit?

Anyone who wants to prevent unit tests from running on a per environment basis.

I am just about to create a PR for this.

### Anything else?

_No response_

TowardOliver:
Resolves #10656 

### Problem

Currently, including "unit_test" in the `DBT_EXCLUDE_RESOURCE_TYPES` env variable (or passing via `--exclude-resource-type`) will prevent unit tests from running when using the `dbt build`, `list` and `clone` commands **ONLY**. 

I want to extend this to also prevent unit tests from running when using the `dbt test` command.

### Solution

I've added a check to the TestSelector, to check the if unit_test has been set in the exclusion list. If it has, then it only allows data tests to run.

Default behaviour is to run both unit and data tests.

It's only a small / simple change. But it will get the job done of preventing unit tests from running on our Prod environments when running `dbt test`.

### Checklist

### DIFF
--- a/tests/functional/unit_testing/test_unit_testing.py
+++ b/tests/functional/unit_testing/test_unit_testing.py
@@ -76,6 +76,13 @@ class TestUnitTests:
         )
         assert len(results) == 1
 
+        # Run test command but specify no unit tests
+        results = run_dbt(
+            ["test", "--select", "my_model", "--exclude-resource-types", "unit_test"],
+            expect_pass=True,
+        )
+        assert len(results) == 0
+
         # Exclude unit tests with environment variable for build command
         os.environ["DBT_EXCLUDE_RESOURCE_TYPES"] = "unit_test"
         results = run_dbt(["build", "--select", "my_model"], expect_pass=True)
--- a/tests/functional/unit_testing/test_unit_testing.py
+++ b/tests/functional/unit_testing/test_unit_testing.py
@@ -76,10 +76,14 @@ class TestUnitTests:
         )
         assert len(results) == 1
 
-        # Exclude unit tests with environment variable
+        # Exclude unit tests with environment variable for build command
         os.environ["DBT_EXCLUDE_RESOURCE_TYPES"] = "unit_test"
         results = run_dbt(["build", "--select", "my_model"], expect_pass=True)
         assert len(results) == 1
+
+        # Exclude unit tests with environment variable for test command
+        results = run_dbt(["test", "--select", "my_model"], expect_pass=True)
+        assert len(results) == 0
 
         del os.environ["DBT_EXCLUDE_RESOURCE_TYPES"]
 
--- a/.changes/unreleased/Features-20240903-132428.yaml
+++ b/.changes/unreleased/Features-20240903-132428.yaml
@@ -1,0 +1,6 @@
+kind: Features
+body: Adding logic to TestSelector to remove unit tests if they are in excluded_resource_types
+time: 2024-09-03T13:24:28.592837+01:00
+custom:
+    Author: TowardOliver
+    Issue: "10656"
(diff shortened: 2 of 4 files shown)


TowardOliver:
I've just signed the CLA, so hopefully this now proceeds.

tests/functional/unit_testing/test_unit_testing.py
```diff
@@ -76,11 +76,15 @@ def test_basic(self, project):
...
-        # Exclude unit tests with environment variable
+        # Exclude unit tests with environment variable for build command
         os.environ["DBT_EXCLUDE_RESOURCE_TYPES"] = "unit_test"
         results = run_dbt(["build", "--select", "my_model"], expect_pass=True)
         assert len(results) == 1
 
+        # Exclude unit tests with environment variable for test command
+        results = run_dbt(["test", "--select", "my_model"], expect_pass=True)
+        assert len(results) == 0
+
```
dbeatty10, line 87: Nice @TowardOliver !

Could you do something similar just a little earlier in the file for when `dbt test --exclude-resource-types unit_test`? You can copy-paste-modify the section that starts with "# Run build command but specify no unit tests".

dbeatty10:
@cla-bot check

TowardOliver:
I've the specific unit test you mentioned @dbeatty10 to test `dbt test --exclude-resource-types unit_test`.

Although I see that you've raised your own PR here:  #10706
So hopefully that one goes through!

dbeatty10:
@TowardOliver Your PR provided the map of everything that needed to be done, plus the unit tests to verify it. 🤩 That empowered me to see how we could solve this though a refactor. Thank you!

https://github.com/dbt-labs/dbt-core/pull/10706 includes your commits and also lists you as co-author to preserve the credit for your contributions.

dbeatty10:
Thanks again for proposing this feature in https://github.com/dbt-labs/dbt-core/issues/10656 and then "showing the way" in this PR @TowardOliver 🏆 !

We just merged https://github.com/dbt-labs/dbt-core/pull/10706, so closing this PR now that the feature is implemented.

PR #10663, opened 2024-09-04
[Backport 1.8.latest] Warn if timestamp updated_at field uses incompatible timestamp (#10352)

gshank:
Backport of #10352 from c28cb92af51d7f2cb27618aeb43705ba951aa3ef


graciegoheen:
I feel ok about only having this for 1.9+ if we want to close out this backport

PR #10664, opened 2024-09-05
Add custom granularities to YAML spec

### LINKED ISSUE #9265: [CT-3485] Semantic Layer custom calendar config support
### Housekeeping

### Short description
Details unknown.

The spec for this is still in the ideation phase.

### Acceptance criteria
People can specify the attributes necessary for custom calendar support 👀 

### Impact to Other Teams

* Semantic Layer - The semantic layer can't begin using this until we support it
* Cloud artifacts - This change will change the v12 manifest that goes out with 1.8

Not sure if required, but the SL team would prefer a backport to 1.7

### Context

This issue should not be considered ready to work on until the work is done in dbt-semantic-interfaces, because the schema we need to implement isn't finalized until that point. Currently there does exist an issue for this work in dbt-semantic-interfaces.

courtneyholcomb:
Resolves #9265

### Description
The semantic layer is in the process of adding a custom calendar feature that will allow users to query using custom granularities like `retail_month`. This requires an addition to the YAML spec for models that looks like this:
```
models:
 - name: my_time_spine
   description: "my favorite time spine"
   time_spine:
      standard_granularity_column: date_day
      custom_granularities:
        - name: fiscal_year
          column_name: fiscal_year_column
```
The only new piece here is `custom_granularities` and everything nested under that key. This spec has been approved by the core product team. 

The corresponding change to schemas.getdbt.com is [here](https://github.com/dbt-labs/schemas.getdbt.com/pull/51).

### Checklist

### DIFF
--- a/core/dbt/artifacts/resources/v1/model.py
+++ b/core/dbt/artifacts/resources/v1/model.py
@@ -23,8 +23,15 @@ class ModelConfig(NodeConfig):
 
 
 @dataclass
+class CustomGranularity(dbtClassMixin):
+    name: str
+    column_name: Optional[str] = None
+
+
+@dataclass
 class TimeSpine(dbtClassMixin):
     standard_granularity_column: str
+    custom_granularities: List[CustomGranularity] = field(default_factory=list)
 
 
 @dataclass
--- a/core/dbt/contracts/graph/semantic_manifest.py
+++ b/core/dbt/contracts/graph/semantic_manifest.py
@@ -24,6 +24,7 @@ from dbt_semantic_interfaces.implementations.semantic_manifest import (
 from dbt_semantic_interfaces.implementations.semantic_model import PydanticSemanticModel
 from dbt_semantic_interfaces.implementations.time_spine import (
     PydanticTimeSpine,
+    PydanticTimeSpineCustomGranularityColumn,
     PydanticTimeSpinePrimaryColumn,
 )
 from dbt_semantic_interfaces.implementations.time_spine_table_configuration import (
@@ -105,6 +106,12 @@ class SemanticManifest:
                     name=time_spine.standard_granularity_column,
                     time_granularity=standard_granularity_column.granularity,
                 ),
+                custom_granularities=[
+                    PydanticTimeSpineCustomGranularityColumn(
+                        name=custom_granularity.name, column_name=custom_granularity.column_name
+                    )
+                    for custom_granularity in time_spine.custom_granularities
+                ],
             )
             pydantic_time_spines.append(pydantic_time_spine)
             if (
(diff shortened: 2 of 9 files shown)


core/dbt/artifacts/resources/v1/model.py
```diff
@@ -22,9 +22,16 @@ class ModelConfig(NodeConfig):
...
+@dataclass
+class CustomGranularity(dbtClassMixin):
+    name: str
+    column_name: Optional[str] = None
+
+
 @dataclass
 class TimeSpine(dbtClassMixin):
     standard_granularity_column: str
+    custom_granularities: List[CustomGranularity] = field(default_factory=list)
```
QMalcolm, line 34: Perfect, thank you for the defaulting behavior

core/dbt/contracts/graph/semantic_manifest.py
```diff
@@ -105,6 +106,12 @@ def _get_pydantic_semantic_manifest(self) -> PydanticSemanticManifest:
                     name=time_spine.standard_granularity_column,
                     time_granularity=standard_granularity_column.granularity,
                 ),
+                custom_granularities=[
+                    PydanticTimeSpineCustomGranularityColumn(
+                        name=custom_granularity.name, column_name=custom_granularity.column_name
+                    )
+                    for custom_granularity in time_spine.custom_granularities
+                ],
```
QMalcolm, line 114: The more we do this, the more I wish we had a better way to automatically translate them. Alas unless we're literally using the same object (or subclassing) we'll need the translation logic somewhere

core/dbt/contracts/graph/unparsed.py
```diff
@@ -207,9 +207,16 @@ class UnparsedNodeUpdate(HasConfig, HasColumnTests, HasColumnAndTestProps, HasYa
     access: Optional[str] = None
 
 
+@dataclass
+class UnparsedCustomGranularity(dbtClassMixin):
+    name: str
+    column_name: Optional[str] = None
+
+
```
QMalcolm, line 215: Instead of redefining this class, can we instead just reuse the `CustomGranularity` class defined in core/dbt/artifacts/resources/v1/model.py. The classes are exactly the same _except_ in their class name (unless I'm misreading something, which is absolutely possible). We could actually do the same with the `UnparsedTimeSpine` vs `TimeSpine` class, but that is a separate matter.

QMalcolm (changes requested):
This looks really great! Thank you for doing this work ❤️ My only request is that I believe we're doing some unnecessary class duplication across `artifiacts/resources/v1/model.py` and `contracts/graph/unparsed.py`. Tests look good and everything else looks good.

core/dbt/contracts/graph/unparsed.py
courtneyholcomb: Good call! Removed duplicates for both!

QMalcolm (approved):
Looks good to me 🚀

PR #10670, opened 2024-09-05
Remove deprecation warning to change "tests:" config to "data_tests:"

### LINKED ISSUE #10564: Remove deprecation warnings for tests to data_tests configs
### Description

We are not going to actually deprecate the use of the "tests:" config.  Although we prefer "data_tests:" because it's more precise and less confusing, and contrasts more logically with our newer "unit_tests:" definitions, the use of "tests:" will continue to work, so we should remove the actual deprecation warning.

Instead we will move this to the project auditing feature: https://github.com/dbt-labs/dbt-project-evaluator/issues/489

### Acceptance Criteria

No more warnings are issued and people who have set up a "warn_error_options" config for TestsConfigDeprecation do not get errors.

gshank:
Resolves #10564

### Problem

The decision was made that we are not deprecating the "tests:" config and it will not be removed in the future, so issuing a deprecation warning is no longer functional.

### Solution

Remove the code that issues the warning. 

### Checklist

### DIFF
--- a/core/dbt/events/types.py
+++ b/core/dbt/events/types.py
@@ -388,6 +388,9 @@ class ConfigTargetPathDeprecation(WarnLevel):
         return line_wrap_message(warning_tag(f"Deprecated functionality\n\n{description}"))
 
 
+# Note: this deprecation has been removed, but we are leaving
+# the event class here, because users may have specified it in
+# warn_error_options.
 class TestsConfigDeprecation(WarnLevel):
     def code(self) -> str:
         return "D012"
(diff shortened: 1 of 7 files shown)


core/dbt/events/types.py
```diff
@@ -388,6 +388,9 @@ def message(self) -> str:
         return line_wrap_message(warning_tag(f"Deprecated functionality\n\n{description}"))
 
 
+# Note: this deprecation has been removed, but we are leaving
+# the event class here, because users may have specified it in
+# warn_error_options.
```
MichelleArk, line 393: good call!

MichelleArk (approved):

core/dbt/events/types.py
dbeatty10: 🧠

PR #10677, opened 2024-09-06
Microbatch: batched execution

MichelleArk:
Resolves https://github.com/dbt-labs/dbt-core/issues/10700

### Problem

For microbatch models, it should be possible to: 
* Split the execution of a microbatch model into batches, regardless of whether it is a full-refresh or incremental run
* Run a single adapter (main) query per batch (i.e. partition)
* Fail gracefully if a single batch fails

### Solution
* Move start/end time computation from provider to MicrobatchBuilder
* Build start/end + batches during ModelRun execution for microbatch models
* Execute each batch: 
  * Populate jinja context vars for recompilation of refs for each batch
  * Re-compile model + update jinja context vars like is_incremental + should_full_refresh
  * Create run result

### Checklist

🎩 Example batch-level failure:
Screenshot 2024-09-16 at 1 22 32 AM


core/dbt/task/run.py
```diff
@@ -284,18 +354,168 @@ def execute(self, model, manifest):
...
+            )
+            if exception:
+                print(exception)
+
+            batch_results.append(batch_run_result)
+
+        return batch_results
+
+    def _build_end_time(self) -> Optional[datetime]:
+        return datetime.now(tz=pytz.utc)
```
MichelleArk, line 458: I removed the `is_incremental` part of this check because I think it's an edge case (and only really one we see in testing) where the dataset has event_time data that is _ahead_ of the current time. 

Additionally, returning `None` here when `is_incremental` was False meant that chunked rebuilds had no upper bound to create batches against and would require a full query (no filters) for an initial run unless `end-time` was specified.

MichelleArk:
🤷 clicking into the failing codecov check, things look okay!

Screenshot 2024-09-16 at 1 39 39 AM

pairing.md
```diff
@@ -0,0 +1,21 @@
+### Requirements
```
QMalcolm, line 1: TODO: We should drop this file

core/dbt/task/run.py
```diff
@@ -284,18 +374,95 @@ def execute(self, model, manifest):
...
+    def _execute_microbatch_materialization(self, model, manifest, context, materialization_macro):
+        batch_results = []
+        microbatch_builder = MicrobatchBuilder(
+            model=model,
+            is_incremental=self._is_incremental(model),
+            event_time_start=getattr(self.config.args, "EVENT_TIME_START", None),
+            event_time_end=getattr(self.config.args, "EVENT_TIME_END", None),
+        )
+        end = microbatch_builder.build_end_time()
+        start = microbatch_builder.build_start_time(end)
```
QMalcolm, line 410: When I was reading through `MicrobatchBuilder` I found it odd that `build_end_time` and `build_start_time` weren't private functions that the `__init__` would then call to default the `event_time_start`/`event_time_end`. With this bit of code here, I find myself still thinking so. Basically, the start/end are always going to be specific to a given `MicrobatchBuilder` instance. Perhaps as a fast follow to this PR we should investigate if this can be reduced to that. The code here would then become
```
...
microbatch_builder = MicrobatchBuilder(
  model=model,
  is_incremental=self._is_incremental(model),
  event_time_start=getattr(self.config.args, "EVENT_TIME_START", None),
  event_time_end=getattr(self.config.args, "EVENT_TIME_END", None),
)
batches = microbatch_builder.build_batches()
...
```

The alternative would be to take the `MicrobatchBuilder` class to be less specific to a model, which also seems like a valid approach. Right now though we seem to be somewhere in the middle with a class that is specific to a model, but methods that we're expected to call which should never change their return given the associated model.

core/dbt/task/run.py
```diff
@@ -284,18 +374,95 @@ def execute(self, model, manifest):
...
+            self.print_batch_result_line(
+                batch_run_result, batch[0], batch_idx + 1, len(batches), exception
+            )
+            batch_results.append(batch_run_result)
+
+        return batch_results
+
+    def _is_incremental(self, model) -> bool:
+        # TODO: Remove. This is a temporary method. We're working with adapters on
+        # a strategy to ensure we can access the `is_incremental` logic without drift
```
QMalcolm, line 454: We should follow up with this on @mikealfare about the possible timeline

core/dbt/task/run.py
```diff
@@ -284,18 +374,95 @@ def execute(self, model, manifest):
...
             self.adapter.post_model_hook(context_config, hook_ctx)
 
-        for relation in self._materialization_relations(result, model):
-            self.adapter.cache_added(relation.incorporate(dbt_created=True))
+        if batch_results:
+            return self._build_run_microbatch_model_result(model, batch_results)
 
         return self._build_run_model_result(model, context)
 
+    def _execute_microbatch_materialization(self, model, manifest, context, materialization_macro):
```
QMalcolm, line 401: I think we can safely add typing to this function 🤔 But also maybe that's a lot harder than I think in this case 😅

core/dbt/task/run.py
```diff
@@ -284,18 +374,95 @@ def execute(self, model, manifest):
...
+            ):
+                batch_results = self._execute_microbatch_materialization(
+                    model, manifest, context, materialization_macro
+                )
+            else:
+                result = MacroGenerator(
+                    materialization_macro, context, stack=context["context_macro_stack"]
+                )()
+                for relation in self._materialization_relations(result, model):
+                    self.adapter.cache_added(relation.incorporate(dbt_created=True))
```
QMalcolm, line 392: I know we talked about this work, and I absolutely understand why we're doing it this way. Something feels smelly about it though, and I can't exactly put my finger on it. My best guess is that we're doing a conditional, exiting the conditional, and then basically re-entering the conditional on lines 396-399. For instance line 399 should _never_ be hit if we enter line 384. However because of the split conditionals, this isn't immediately apparent. I wonder if we should be calling into two separate private functions just before the `try` on line 378, and only one or the other function would ever be called depending on if we're doing microbatch stuff or not.

core/dbt/task/run.py
```diff
@@ -239,6 +302,33 @@ def _build_run_model_result(self, model, context):
...
 
+    def _build_run_microbatch_model_result(self, model, batch_run_results):
+        failures = sum([result.failures for result in batch_run_results if result.failures])
+
+        return RunResult(
+            node=model,
+            status=RunStatus.Success,
+            timing=[],
+            thread_id=threading.current_thread().name,
+            # TODO -- why isn't this getting propagated to logs?
```
QMalcolm, line 313: The execution_time isn't making it to the logs? 🤔 That's odd....

core/dbt/materializations/incremental/microbatch.py
```diff
@@ -0,0 +1,130 @@
...
+from typing import List, Optional, Tuple
+
+import pytz
+
+from dbt.artifacts.resources.types import BatchSize
+from dbt.contracts.graph.nodes import ModelNode, NodeConfig
+from dbt.exceptions import DbtInternalError, DbtRuntimeError
+
+
+class MicrobatchBuilder:
```
QMalcolm, line 11: I think we should add some docstrings to the class and functions in this file. It will set a good precedence, and also allows us to document some edge cases / assumptions that the functions make.

QMalcolm (changes requested):
Requesting changes and am gonna follow up by making said changes 😅

core/dbt/task/run.py
QMalcolm: Done! 🙂

QMalcolm (approved):
My remaining open comments can be addressed at a later date. Let's move forward 🚀

Issue #8286, opened 2023-08-02
[CT-2914] [Feature] Use PostgreSQL 15 for CI tests


dbeatty10:
### Is this your first time submitting a feature request?

- [X] I have read the [expectations for open source contributors](https://docs.getdbt.com/docs/contributing/oss-expectations)
- [X] I have searched the existing issues, and I could not find an existing issue for this feature
- [X] I am requesting a straightforward extension of existing dbt functionality, rather than a Big Idea better suited to a discussion

### Describe the feature

Use postgresql 15.x within CI via GitHub Actions rather than [14.8](https://github.com/actions/runner-images/blob/1fb9925a9ff67c2803a854c7094b2b58d342a2ef/images/linux/Ubuntu2004-Readme.md?plain=1#L190-L191).

You can discover the current version of Postgres that is being used with [this](https://github.com/search?q=repo%3Aactions%2Frunner-images%20PostgreSQL&type=code) code search.

### Describe alternatives you've considered

We can leave the `merge` incremental strategy untested for dbt-postgres.

Alternatively, we can rely upon manual testing performed off-line to test this strategy with postgres.

### Who will this benefit?

The [merge](https://github.com/dbt-labs/dbt-core/blob/a433c31d6ebae79bd5b281653551899b7369067c/tests/functional/materializations/test_incremental.py#L45-L46) incremental strategy only works for [Postgres >= 15](https://www.postgresql.org/about/news/postgresql-15-released-2526/).

To test this strategy in https://github.com/dbt-labs/dbt-core/pull/8268, we will need postgresql >= 15.

PR #10594, opened 2024-08-22
Microbatch: event_time ref + source filtering

MichelleArk:
Resolves #9490 #10635 #10637 #10638 #10636 #10662 #10639

### Details

We're working to introduce a new feature, Microbatch Incremental Models! The goal being that microbatch incremental models will break updates to microbatch models into batched updates for a given time frame. There's more to microbatch incremental models than that, but that's the gist of if. 

This PR doesn't actually do the breaking up into batches. However, what this PR does is set up the fundamentals to do so. There are a few new model configs we need: `event_time`, `batch_size`, `lookback`. Additionally, there are some CLI flags we need: `--event-time-start` and `--event-time-end`. This PR makes those available and uses them to set the foundation of generating event time filters and determining _when_ the event time filters _should_ be generated. Finally, all this foundational functionality is gated by a new environment variable `DBT_EXPERIMENTAL_MICROBATCH`.

### Checklist

- [X] I have read [the contributing guide](https://github.com/dbt-labs/dbt-core/blob/main/CONTRIBUTING.md) and understand what's expected of me.
- [X] I have run this code in development, and it appears to resolve the stated issue.
- [X] This PR includes tests, or tests are not required or relevant for this PR.
- [X] This PR has no interface changes (e.g., macros, CLI, logs, JSON artifacts, config files, adapter interface, etc.) or this PR has already received feedback and approval from Product or DX.
- [X] This PR includes [type annotations](https://docs.python.org/3/library/typing.html) for new and modified functions.

PR #10687, opened 2024-09-10
install postgres-16 in CI

MichelleArk:
Resolves https://github.com/dbt-labs/dbt-core/issues/8286

### Problem

We need postgres >= 15 to test dbt-core code that depend on merge operations in dbt-postgres. Specifically this is blocking testing the new microbatch incremental strategy: https://github.com/dbt-labs/dbt-core/pull/10594

Additionally, dbt-postgres is already using postgres 16 in CI so there is some inconsistency there. 

### Solution

Using: https://computingforgeeks.com/install-and-configure-postgresql-on-ubuntu/

Install postgres 16 onto ubuntu 20.04. We can also consider a service-based approach as what is done in dbt-postgres, but that will likely be a bigger refactor (https://github.com/dbt-labs/dbt-core/pull/10688) + container services are only supported for linux runners. We'd still need to roll our own updates for macos + windows

### Checklist

### DIFF
--- a/.github/actions/setup-postgres-linux/action.yml
+++ b/.github/actions/setup-postgres-linux/action.yml
@@ -5,6 +5,7 @@ runs:
   steps:
     - shell: bash
       run: |
+        sudo apt-get --purge remove postgresql postgresql-*
         sudo apt update -y
         sudo apt install gnupg2 wget vim -y
         sudo sh -c 'echo "deb https://apt.postgresql.org/pub/repos/apt $(lsb_release -cs)-pgdg main" > /etc/apt/sources.list.d/pgdg.list'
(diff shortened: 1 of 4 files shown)


gshank (approved):

MichelleArk:
Ideas to speed up windows runs: 
* create a separate image for windows that pre-installs postgres16 
* uninstall existing postgres package, use choco install (which appears to be faster), and then figure out how to start postgres as a (or start postgres up on on our own)
  * uninstall guide: https://stackoverflow.com/questions/113542/how-can-i-uninstall-an-application-using-powershell 
  * starting the service will likely take a good amount of time to figure out
  * can take inspiration from https://github.com/ikalnytskyi/action-setup-postgres/blob/master/action.yml, or try to use the action directly (first unsuccessful attempt here: https://github.com/dbt-labs/dbt-core/pull/10689)

Will make an issue shortly 👍

PR #10706, opened 2024-09-12
Enable `--resource-type` and `--exclude-resource-type` CLI flags and environment variables for `dbt test`

### LINKED ISSUE #10656: [Feature] Exclude unit tests from running when using "dbt test" command

dbeatty10:
Resolves #10656

Alternative to #10657

### Problem

Currently, the [resource type](https://docs.getdbt.com/reference/global-configs/resource-type) CLI flags and environment variables apply to the `dbt build`, `dbt list` and `dbt clone` subcommands **ONLY** -- they do not extend to `dbt test`.

We want to extend these to apply to the `dbt test` subcommand also.

### Solution

Simplify by standardizing upon the `ResourceTypeSelector` which already provides the desired `resource_types` parameter.

This was accomplished by:
1. adding the `resource_type` and `exclude_resource_type` click args to the `dbt test` subcommand
1. adding a `TEST_NODE_TYPES` list of valid test types (data tests and unit tests)
1. 👉 adding `resource_types` property to the `TestTask` (which uses `TEST_NODE_TYPES`) 👈 **this was the crux!**

### Backstory

Branched from https://github.com/dbt-labs/dbt-core/pull/10657 to try an alternative approach that resolves the issue via a refactor.

Did these PRs leading up to this one:
- https://github.com/dbt-labs/dbt-core/pull/10730
- https://github.com/dbt-labs/dbt-core/pull/10718

### Code examples

dbt `test` now enables passing values like `unit_tests`, `tests`, etc. in the `--resource-type` / `--exclude-resource-type` CLI flags (or in the  environment variables `DBT_RESOURCE_TYPES` / `DBT_EXCLUDE_RESOURCE_TYPES`).

#### Example 1 - CLI flags

```shell
dbt test --resource-type unit_test
dbt test --exclude-resource-type unit_test
```

#### Example 2 - environment variables

```shell
DBT_RESOURCE_TYPES=unit_test dbt test
DBT_EXCLUDE_RESOURCE_TYPES=unit_test dbt test
```

### Documentation to update

At minimum (there may be other relevant pages):
- https://docs.getdbt.com/reference/global-configs/resource-type

### Checklist


.changes/unreleased/Features-20240903-132428.yaml
```diff
@@ -0,0 +1,6 @@
+kind: Features
+body: Adding logic to TestSelector to remove unit tests if they are in excluded_resource_types
+time: 2024-09-03T13:24:28.592837+01:00
+custom:
+    Author: TowardOliver
```
dbeatty10, line 5:
```suggestion
    Author: TowardOliver dbeatty10
```

.changes/unreleased/Features-20240903-132428.yaml
```diff
@@ -0,0 +1,6 @@
+kind: Features
+body: Adding logic to TestSelector to remove unit tests if they are in excluded_resource_types
```
dbeatty10, line 2:
```suggestion
body: Enable `--resource-type` and `--exclude-resource-type` CLI flags and environment variables for `dbt test`
```

core/dbt/task/test.py
```diff
@@ -387,14 +397,24 @@ class TestTask(RunTask):
...
+    @property
+    def resource_types(self) -> List[NodeType]:
+        resource_types: Collection[NodeType] = resource_types_from_args(
+            self.args, set(TEST_NODE_TYPES), set(TEST_NODE_TYPES)
+        )
+
+        # filter out any non-test node types
+        resource_types = [rt for rt in resource_types if rt in TEST_NODE_TYPES]
+        return list(resource_types)
+
```
dbeatty10, line 409: This is basically the same thing as [here](https://github.com/dbt-labs/dbt-core/blob/3e437a67344469d51d7f2d4a4d0338a45c1de261/core/dbt/task/list.py#L179-L188) and [here](https://github.com/dbt-labs/dbt-core/blob/3e437a67344469d51d7f2d4a4d0338a45c1de261/core/dbt/task/build.py#L85-L94).

core/dbt/task/test.py
```diff
@@ -387,14 +397,24 @@ class TestTask(RunTask):
...
     def get_node_selector(self) -> ResourceTypeSelector:
         if self.manifest is None or self.graph is None:
             raise DbtInternalError("manifest and graph must be set to get perform node selection")
         return ResourceTypeSelector(
             graph=self.graph,
             manifest=self.manifest,
             previous_state=self.previous_state,
-            resource_types=[NodeType.Test, NodeType.Unit],
+            resource_types=self.resource_types,
```
dbeatty10, line 417: Basically same thing as [here](https://github.com/dbt-labs/dbt-core/blob/3e437a67344469d51d7f2d4a4d0338a45c1de261/core/dbt/task/list.py#L214) and [here](https://github.com/dbt-labs/dbt-core/blob/3e437a67344469d51d7f2d4a4d0338a45c1de261/core/dbt/task/build.py#L210).

gshank (approved):

PR #10721, opened 2024-09-16
Add Snowplow tracking for behavior flags

### LINKED ISSUE #10552: [Behavior Flags] Track behavior flags in Snowplow
### Housekeeping

## Short description

As a maintainer, I need to track what proportion of users are using legacy behavior, so that I know when it's safe to retire that behavior. We already fire a warning event when the flag evaluates to `False`. We need to track how often this occurs for each flag as a count of invocation id.

```[tasklist]
## Acceptance criteria
- [x] Any time a behavior flag evaluates to `False`, this gets tracked in our Snowplow instance
- [x] No events for behavior flags are tracked if the user has turned off anonymous tracking (this is existing behavior for all other tracking and should continue to be observed)
```

```[tasklist]
## Suggested testing
- [x] Create a scenario where the flag evaluates to `False`, verify that the callback is fired
- [x] Create a scenario where the flag evaluates to `True`, verify that the callback is not fired
- [x] Fire an event other than a `BehaviorDeprecationEvent` and verify that the callback is not fired
```

### Impact to other teams

There's no immediate impact to any team. However, it would make sense for existing behavior flags in `dbt-core` to migrate to the new framework at some point.

No backports are required; this is new functionality.

mikealfare:
resolves #10552

### Problem

We need to track deprecated states of behavior flags in Snowplow so that we get telemetry for both Cloud and Core users.

### Solution

Handle the event that is already firing for logging with a callback that tracks the event in Snowplow.

### Checklist

### DIFF
--- a/core/dbt/events/logging.py
+++ b/core/dbt/events/logging.py
@@ -2,7 +2,7 @@ import os
 from functools import partial
 from typing import Callable, List
 
-from dbt.tracking import track_behavior_deprecation_warn
+from dbt.tracking import track_behavior_change_warn
 from dbt_common.events.base_types import EventLevel, EventMsg
 from dbt_common.events.event_manager_client import (
     add_callback_to_manager,
@@ -70,7 +70,7 @@ def setup_event_logger(flags, callbacks: List[Callable[[EventMsg], None]] = []) 
     make_log_dir_if_missing(flags.LOG_PATH)
     event_manager = get_event_manager()
     event_manager.callbacks = callbacks.copy()
-    add_callback_to_manager(track_behavior_deprecation_warn)
+    add_callback_to_manager(track_behavior_change_warn)
 
     if flags.LOG_LEVEL != "none":
         line_format = _line_format_from_str(flags.LOG_FORMAT, LineFormat.PlainText)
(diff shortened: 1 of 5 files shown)


mikealfare:
@cla-bot check

core/dbt/events/logging.py
```diff
@@ -68,6 +70,7 @@ def setup_event_logger(flags, callbacks: List[Callable[[EventMsg], None]] = [])
     make_log_dir_if_missing(flags.LOG_PATH)
     event_manager = get_event_manager()
     event_manager.callbacks = callbacks.copy()
+    add_callback_to_manager(track_behavior_deprecation_warn)
```
mikealfare, line 73: I think we want snowplow tracking for behavior flags regardless of the log setting. It should only be turned off when the user turns off tracking. However, this is handled outside of the event manager. The downside is that tests need to be updated for this, though it appears to only be one test.

gshank (approved):

mikealfare:
I think these tests are failing due to updates in `dbt-common` on which this is dependent. I'll rerun the tests once that PR is merged; it's currently queued.

PR #10735, opened 2024-09-18
Fix Calls to pre_model_hook()

peterallenwebb:
This fixes two calls to pre_model_hook() which pass objects of the wrong type. The type used in other places this function is called is RuntimeConfigObject, but here a context dictionary was passed. This was causing errors for me in a very specific testing scenario, but looks like a more general problem.

### Checklist


gshank (approved):
Yes, I think you're correct.

PR #10737, opened 2024-09-18
Split out model vs microbatch execution

QMalcolm:
Resolves #N/A

### Description

In #10677 we added microbatch model execution logic to the `ModelRunner.execute`. In doing so we changed the order of operations of some normal model execution steps. This PR splits the logic so that we can return the order of operations for normal model execution back to what it was previously while still supporting microbatch model execution. We don't know if the change of order operations that happened in #10677 was problematic, but moving it back is being done out caution.

### Checklist

### DIFF
--- a/core/dbt/task/run.py
+++ b/core/dbt/task/run.py
@@ -350,6 +350,47 @@ class ModelRunner(CompileRunner):
         )
         raise CompilationError(msg, node=model)
 
+    def _execute_model(
+        self,
+        hook_ctx: Any,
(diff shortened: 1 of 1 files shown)


MichelleArk (approved):

QMalcolm:
The failing unit test, `test_false_evaluation_triggers_snowplow_tracking`, is also failing on `main` right now. Looks like this was a recently added test via #10721. Unfortunately, the failing unit test is blocking merges.

PR #10743, opened 2024-09-19
Write microbatch compiled + run code to separate target files

MichelleArk:
Resolves https://github.com/dbt-labs/dbt-core/issues/10714

### Problem

During execution of microbatch models, each batch is re-compiled with batch-level time filters. Currently, this results in the compiled/run .sql files becoming clobbered on each write when there should be one file per compiled/run batch.

### Solution

Pass an additional argument, `split_suffix` (naming suggestions welcome!) that splits the write path by the suffix and writes to a top-level directory without the suffix.

### Checklist


tests/functional/microbatch/test_microbatch.py
```diff
@@ -442,3 +443,78 @@ def test_run_with_event_time(self, project):
...
+
+
+class TestMicrobatchCompiledRunPaths(BaseMicrobatchTest):
+    @mock.patch.dict(os.environ, {"DBT_EXPERIMENTAL_MICROBATCH": "True"})
+    def test_run_with_event_time(self, project):
+        # run all partitions from start - 2 expected rows in output, one failed
+        with patch_microbatch_end_time("2020-01-03 13:57:00"):
+            run_dbt(["run", "--event-time-start", "2020-01-01"])
+
+        # Compiled paths - compiled model without filter only
```
MichelleArk, line 455: After discussion with @graciegoheen, we decided it could still be useful to see the non-batched (no filters applied) model file, and at the top-level is where users would expect it. I've added a test to formalize this expected behaviour, but it's something that would be easy to change if we get beta feedback!

core/dbt/contracts/graph/nodes.py
```diff
@@ -559,6 +572,16 @@ def infer_primary_key(self, data_tests: List["GenericTestNode"]) -> List[str]:
...
 
+    def format_batch_start(self, batch_start: Optional[datetime]) -> Optional[str]:
+        if batch_start is None:
+            return batch_start
+
+        return str(
+            batch_start.date()
+            if (batch_start and self.config.batch_size != BatchSize.hour)
+            else batch_start
+        )
```
QMalcolm, line 583: ### nit
Feels somewhat weird that this is on the `ParsedNode` class instead of the [MicrobatchBuilder](https://github.com/dbt-labs/dbt-core/blob/aa23af98e57157d99b86be1571f544fe6b37c1d0/core/dbt/materializations/incremental/microbatch.py#L11) class. Just feels out of place.

QMalcolm (approved):
Looks good! Thanks for doing this work. Found one aesthetic thing, but not blocking.

core/dbt/contracts/graph/nodes.py
MichelleArk: totally agree! I think at first I didn't do this because it's necessary in providers.py which didn't Added it as a static method. Good call 👍

PR #10749, opened 2024-09-20
Allow snapshots to be defined in YAML.

### LINKED ISSUE #10246: [Feature] No more jinja block for snapshots - new snapshot design ideas

### Current State

To configure a snapshot currently, you must nest your configuration and SQL within a snapshot jinja block like so:
```
{% snapshot orders_snapshot %}

{{
    config(
      target_database='analytics',
      target_schema='snapshots',
      unique_key='id',
      strategy='timestamp',
      updated_at='updated_at'
    )
}}

select * from {{ source('jaffle_shop', 'orders') }}

{% endsnapshot %}
```

Why? (you might ask)

**The story begins…**

Snapshots are a really *ancient* dbt feature -- implemented as `dbt archive` in https://github.com/dbt-labs/dbt-core/pull/183 and first released in [0.5.1](https://github.com/dbt-labs/dbt-core/blob/41c6177ae2c85343426c5af53f9e23451e762619/CHANGELOG.md#dbt-release-051), just two days shy of dbt's 6 month anniversary.

There were no `snapshot` blocks and  `snapshots/*.sql` files in these early days.

Instead, they were originally declared within `dbt_project.yml` like this:
```
archive:
    - source_schema: synced_production_data  # schema to look for tables in (declared below)
      target_schema: dbt_archive             # where to archive the data to
      tables:
        - source_table: some_other_table
           target_table: some_other_table_archive
           updated_at: "updatedAt"
           unique_key: "id"
```

**A glow up**

https://github.com/dbt-labs/dbt-core/issues/1175 and https://github.com/dbt-labs/dbt-core/pull/1361 allowed snapshots to escape YAML Land and become:
> `select` statements, defined within a `snapshot` block in a `.sql` file

```
{% snapshot your_snapshot_name_here %}

{{ config(
          target_database='<optional database name>',
          target_schema='<schema name>',
          target_table='<table name>',
          strategy='check',
          unique_key='id',
          check_cols=['object_status', 'object_name'],
) }}

-- your select statement here

{% endsnapshot %}
```

At the time the thought was, “we should/will reimplement all the resources like this” (so that you could define multiple “model blocks” in a single file).

Turns out that defining multiple resources in one file makes
- parsing slower
- IDE interactive compile/preview *much* harder

and so in the leadup to v1.0, it wasn’t a priority to do this rework — and finally decided it wasn’t really even desirable.

### Future State

WE ARE GOING WITH OPTION 1

**Option 1: Snapshots are just yml configs, they contain no logic (like exposures, sources, tests, etc.)**
```
# snapshots/my_snapshots.yml
snapshots:
  - name: orders_snapshot
    config:
      schema: snapshots
      database: analytics
    relation: source('jaffle_shop', 'orders') # name TBD could be a model or a source... ideas - input, on, relation
    unique_key: id
    strategy: timestamp
    updated_at: updated_at
```
- transformation is just `select * from {{ source('jaffle_shop', 'orders') }}`
  - best practice is “staging” layer on top of your snapshots for cleanup (could even be an ephemeral model if you don’t want clutter in your warehouse)
  - exceptions where you’d want to do light transformation on your source, then snapshot (or bake logic into snapshot):
    - types of logic: filters, deduplication, surrogate → unique key, etc.
    - best practice → snapshot an ephemeral model that contains the logic
    - This method also helps with development because you’ll be able to run the query rather than running a snapshot every time
- “snapshots are a config” on a model (like tests, etc.)
- [ideally] snapshots could be defined in yml in models folder (like sources) to allow for better organization (currently silo-d in snapshots folder)

**Option 2: Snapshots are just models, they are a materialization (like incremental, view, table, etc.)**
```
# models/my_snapshot.sql

{{
    config(
      target_database='analytics',
      target_schema='snapshots',
      unique_key='id',
      strategy='timestamp',
      updated_at='updated_at',
      materialized='snapshot'
    )
}}

select * from {{ source('jaffle_shop', 'orders') }}
```
  
- if snapshots are just models, would they now be included in `dbt run`? vs. `dbt snapshot`?
- how would we handle `snapshot-paths`? would you be able to put them in your models folder? or only snapshots folder?
- what would happen to their `resource_type` - model or snapshot (incremental is model)? if we went with model, that would be an id change, adjustment to selector syntax for dbt build, `DBT_EXCLUDE_RESOURCE_TYPE` wouldn’t work for excluding snapshots, etc.
- this would be the only (?) model materialization where dbt injects its own meta-fields into the final table
- how would a migration path look like from one materialization to another? table -> snapshot? incremental -> snapshot? etc?
- `materialized='snapshot'` or `materialized='scd'`

**Option 3: Snapshots are just sql files in the snapshots folder, but they don’t use jinja blocks (one .sql file per snapshot)**
```
# snapshots/my_snapshot.sql

{{
    config(
      target_database='analytics',
      target_schema='snapshots',
      unique_key='id',
      strategy='timestamp',
      updated_at='updated_at'
    )
}}

select * from {{ source('jaffle_shop', 'orders') }}
```
    
- snapshot name is same as .sql file name, just like models

**Which is best?**
- I think the question between the options above comes down to:
  - how strongly do we believe in the “snapshots should only be `select *`” best practice
  - how much do we care about migration paths
  - what other things would we need to change (commands, configs, etc.)
  - what’s the technical lift for each of these (can we map the new → old)

  | Option 1: Snapshots are just yml configs | Option 2: Snapshots are just models | Option 3: Snapshots are just sql files
-- | -- | -- | --
enforce best practice that snapshot is just select * | X |   |  
can have SQL logic being snapshotted | X (snapshot ephemeral model with logic) | X | X
easy migration path for users (does this matter if we keep prior art for backwards compatibility?) |   | X | X
mapping new to old (define multiple snapshots in 1 file, etc.) | X |   |  
minimal consequences of other things we’d need to change / consider (commands, configs, etc.) | X |   | X
define your snapshots in your models directory | X | X |   |

### Notes
- we will keep the prior art, for backwards compatibility

### Related issues
https://github.com/dbt-labs/dbt-core/issues/4761
https://github.com/dbt-labs/dbt-core/issues/9033

peterallenwebb:
Resolves #10246

### Problem

Users had to awkwardly create jinja snapshot blocks in SQL files, even for very simple snapshots.

### Solution

Delight users by allowing snapshots to be defined intuitively, with just a few lines of YAML.

### Checklist

### DIFF
--- a/.changes/unreleased/Features-20240903-132428.yaml
+++ b/.changes/unreleased/Features-20240903-132428.yaml
@@ -1,0 +1,6 @@
+kind: Features
+body: Enable `--resource-type` and `--exclude-resource-type` CLI flags and environment variables for `dbt test`
+time: 2024-09-03T13:24:28.592837+01:00
+custom:
+    Author: TowardOliver dbeatty10
+    Issue: "10656"
--- a/.changes/unreleased/Features-20240920-110447.yaml
+++ b/.changes/unreleased/Features-20240920-110447.yaml
@@ -1,0 +1,6 @@
+kind: Features
+body: Allow snapshots to be defined in YAML.
+time: 2024-09-20T11:04:47.703117-04:00
+custom:
+  Author: peterallenwebb
+  Issue: "10246"
--- a/.changes/unreleased/Under the Hood-20240911-162730.yaml
+++ b/.changes/unreleased/Under the Hood-20240911-162730.yaml
@@ -1,0 +1,6 @@
+kind: Under the Hood
+body: Add Snowplow tracking for behavior flag deprecations
+time: 2024-09-11T16:27:30.293832-04:00
+custom:
+  Author: mikealfare
+  Issue: "10552"
--- a/.changes/unreleased/Under the Hood-20240916-102201.yaml
+++ b/.changes/unreleased/Under the Hood-20240916-102201.yaml
@@ -1,0 +1,6 @@
+kind: Under the Hood
+body: Replace `TestSelector` with `ResourceTypeSelector`
+time: 2024-09-16T10:22:01.339462-06:00
+custom:
+  Author: dbeatty10
+  Issue: "10718"
--- a/.changes/unreleased/Under the Hood-20240918-170325.yaml
+++ b/.changes/unreleased/Under the Hood-20240918-170325.yaml
@@ -1,0 +1,7 @@
+kind: Under the Hood
+body: Standardize returning `ResourceTypeSelector` instances in `dbt list` and `dbt
+  build`
+time: 2024-09-18T17:03:25.639516-06:00
+custom:
+  Author: dbeatty10
+  Issue: "10739"
--- a/core/dbt/cli/main.py
+++ b/core/dbt/cli/main.py
@@ -785,6 +785,8 @@ cli.commands["source"].add_command(snapshot_freshness, "snapshot-freshness")  # 
 @click.pass_context
 @global_flags
 @p.exclude
+@p.resource_type
+@p.exclude_resource_type
 @p.profiles_dir
 @p.project_dir
 @p.select
--- a/core/dbt/events/logging.py
+++ b/core/dbt/events/logging.py
@@ -2,8 +2,10 @@ import os
 from functools import partial
 from typing import Callable, List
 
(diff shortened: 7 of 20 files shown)


peterallenwebb:
These changes were merged in a different PR.

PR #10755, opened 2024-09-22
Fix unit tests for incremental models with alias

### LINKED ISSUE #10754: [Regression] Unit tests generate an sql with wrong CTE  `this` referrence on incremental model with alias

### Current Behavior

Unit tests fail due to wrong `this` cte referrence on incremental model with alias.

### Expected Behavior

I expect that the name of `this` referrence will be consistent with the cte, regardless of alias.

### Steps To Reproduce

1. I have a incremental model with alias

```sql
{{ 
    config(
        materialized='incremental',
        alias='aliased_example',
    )
}}

with source_data as (

    select 1 as id
    union all
    select null as id

)

select *
from source_data
{% if is_incremental() %}
where id > (select max(id) from {{ this }} )
{% endif %}
```

2. And a unit test on a table which overrides is_incremental macro is true

```yml
unit_tests:
  - name: case_incremental
    model: example
    overrides:
      macros:
        is_incremental: true
    given:
      - input: this
        rows:
          - {id: 0}
    expect:
      rows:
        - {id: 1}
```

3. Then I am running the unit test `dbt test --s example`, I am getting an error:

```
15:04:35  Finished running 1 unit test in 0 hours 0 minutes and 2.35 seconds (2.35s).
15:04:35  
15:04:35  Completed with 1 error and 0 warnings:
15:04:35  
15:04:35    Runtime Error in unit_test case_incremental (models/example/schema.yml)
  An error occurred during execution of unit test 'case_incremental'. There may be an error in the unit test definition: check the data types.
   Database Error
    Table "__dbt__cte__example" must be qualified with a dataset (e.g. dataset.table).
```

because the generated SQL has different name of the CTE and the from referrence:

```sql
select * from (
        

with  __dbt__cte__aliased_example as (

-- Fixture for example
select safe_cast(0 as INT64) as id
), source_data as (

    select 1 as id
    union all
    select null as id

)

select *
from source_data

where id > (select max(id) from __dbt__cte__example )

    ) as __dbt_sbq
    where false and current_timestamp() = current_timestamp()
    limit 0

    )
```

### Relevant log output

_No response_

### Environment

```markdown
- OS: Ubuntu 24.04
- Python: 3.12.3
- dbt: 1.8.6
```

bigquery

### Additional Context

_No response_

### LINKED ISSUE #10763: [Regression] Incorrect CTE name in unit tests using `incremental` mode for versioned incremental models
### Is this a new bug?

- [X] I believe this is a new bug

### Current Behavior

When defining a unit test for the incremental load on a versioned incremental model, the generated SQL is invalid. The name of the CTE "injected" at the top contains the version name, when `this` is replaced by the CTE name, but without the version name.

e.g., with the versioned model `my_model` we get

```
with  __dbt__cte__my_model_v1 as (
...
)

select * from ...
where my_date > (select max(my_date) from __dbt__cte__my_model)
```

This is not valid SQL

### Expected Behavior

`this` is replaced by the CTE name that includes the version number

### Steps To Reproduce

1. create an incremental model 
2. add a version to the model
3. add a unit test to the model with `is_incremental: true` and some rows for `this`
4. run the unit test

### Relevant log output

```shell
Not relevant
```

### Environment

```markdown
- dbt-adapters:1.6.1
```

### Additional Context

_No response_

katsugeneration:
Resolves #10754 
Resolves #10763

### Problem

- Unit tests fail due to wrong input cte referrence name on incremental `this` models with alias
- Because cte name is based on alias name on ephemeral model. This change on dbt-adapters 1.4.1 by https://github.com/dbt-labs/dbt-adapters/pull/236

### Solution

- Fix `this` referrence name in UnitTestContext to based on alias name on incremental model

### Checklist

### DIFF
--- a/core/dbt/context/providers.py
+++ b/core/dbt/context/providers.py
@@ -1636,7 +1636,7 @@ class UnitTestContext(ModelContext):
         if self.model.this_input_node_unique_id:
             this_node = self.manifest.expect(self.model.this_input_node_unique_id)
             self.model.set_cte(this_node.unique_id, None)  # type: ignore
-            return self.adapter.Relation.add_ephemeral_prefix(this_node.name)
+            return self.adapter.Relation.add_ephemeral_prefix(this_node.identifier)  # type: ignore
         return None
 
 
--- a/tests/functional/unit_testing/test_unit_testing.py
+++ b/tests/functional/unit_testing/test_unit_testing.py
@@ -8,6 +8,7 @@ from fixtures import (  # noqa: F401
     external_package,
     external_package__accounts_seed_csv,
     my_incremental_model_sql,
+    my_incremental_model_versioned_yml,
     my_incremental_model_with_alias_sql,
     my_model_a_sql,
     my_model_b_sql,
@@ -279,6 +280,24 @@ class TestUnitTestIncrementalModelWithAlias:
             "my_incremental_model.sql": my_incremental_model_with_alias_sql,
             "events.sql": event_sql,
             "schema.yml": test_my_model_incremental_yml_basic,
+        }
+
+    def test_basic(self, project):
+        results = run_dbt(["run"])
+        assert len(results) == 2
+
+        # Select by model name
+        results = run_dbt(["test", "--select", "my_incremental_model"], expect_pass=True)
+        assert len(results) == 2
+
+
+class TestUnitTestIncrementalModelWithVersion:
+    @pytest.fixture(scope="class")
+    def models(self):
+        return {
+            "my_incremental_model.sql": my_incremental_model_sql,
+            "events.sql": event_sql,
+            "schema.yml": my_incremental_model_versioned_yml + test_my_model_incremental_yml_basic,
         }
 
     def test_basic(self, project):
--- a/tests/functional/unit_testing/test_unit_testing.py
+++ b/tests/functional/unit_testing/test_unit_testing.py
@@ -8,6 +8,7 @@ from fixtures import (  # noqa: F401
     external_package,
     external_package__accounts_seed_csv,
     my_incremental_model_sql,
+    my_incremental_model_with_alias_sql,
     my_model_a_sql,
     my_model_b_sql,
     my_model_sql,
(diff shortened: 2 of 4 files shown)


katsugeneration:
@cla-bot check

dbeatty10 (commented):
Thanks for the targeted solution and the nice functional test @katsugeneration 🤩 

I can confirm that running the following test fails without the fix, but succeeds with the test in place:

```shell
python -m pytest tests/functional/unit_testing/test_unit_testing.py::TestUnitTestIncrementalModelWithAlias```

File: tests/functional/unit_testing/test_unit_testing.py [hunk 1]
```diff
@@ -271,6 +273,42 @@ def test_no_this_input(self, project):
...
+            "schema.yml": test_my_model_incremental_yml_basic,
+        }
+
+    def test_basic(self, project):
+        results = run_dbt(["run"])
+        assert len(results) == 2
+
+        # Select by model name
+        results = run_dbt(["test", "--select", "my_incremental_model"], expect_pass=True)
+        assert len(results) == 2
```
dbeatty10, line 291: This tests the regression reported in https://github.com/dbt-labs/dbt-core/issues/10754.

I confirmed this test fails without the fix but succeeds with the fix in place:
```shell
python -m pytest tests/functional/unit_testing/test_unit_testing.py::TestUnitTestIncrementalModelWithAlias
```

tests/functional/unit_testing/test_unit_testing.py
```diff
@@ -271,6 +273,42 @@ def test_no_this_input(self, project):
...
+            "schema.yml": my_incremental_model_versioned_yml + test_my_model_incremental_yml_basic,
+        }
+
+    def test_basic(self, project):
+        results = run_dbt(["run"])
+        assert len(results) == 2
+
+        # Select by model name
+        results = run_dbt(["test", "--select", "my_incremental_model"], expect_pass=True)
+        assert len(results) == 2
```
dbeatty10, line 309: This tests the regression reported in https://github.com/dbt-labs/dbt-core/issues/10763.

I confirmed this test fails without the fix but succeeds with the fix in place:
```shell
python -m pytest tests/functional/unit_testing/test_unit_testing.py::TestUnitTestIncrementalModelWithVersion
```

core/dbt/context/providers.py
```diff
@@ -1636,7 +1636,7 @@ def this(self) -> Optional[str]:
         if self.model.this_input_node_unique_id:
             this_node = self.manifest.expect(self.model.this_input_node_unique_id)
             self.model.set_cte(this_node.unique_id, None)  # type: ignore
-            return self.adapter.Relation.add_ephemeral_prefix(this_node.name)
+            return self.adapter.Relation.add_ephemeral_prefix(this_node.identifier)  # type: ignore
```
dbeatty10, line 1639: The root causes of the regressions reported in https://github.com/dbt-labs/dbt-core/issues/10754 and https://github.com/dbt-labs/dbt-core/issues/10763 was that:

1. the update from `this_node.name` to `this_node.identifier` was accidentally overlooked during the implementation and code review of https://github.com/dbt-labs/dbt-core/pull/10290/files, and
2. there were no functional tests to catch those particular cases

This PR addresses _both_ of those root causes ✅ ✅

MichelleArk (approved):

rmbborges:
Hey @dbeatty10 , will this fix be included in `1.8.8`?

PR #10752, opened 2024-09-20
Microbatch Config Validation

QMalcolm:
Resolves #10709

### Problem

We either don't validate configs for microbatch models or validate them during execution. We need parse-time validation to ensure that required fields are present and typed correctly:
* batch_size (day, hour, month, year)
* lookback (int)
* event_time (str)
* begin (datetime)

Additionally, any inputs to a microbatch model that have an event_time config must have that event_time config be a string.

### Solution

* implement parse-time  validation as part of final manifest validation since we need all ref/sources resolved to scan inputs of microbatch model.

### Checklist

- [x] I have read [the contributing guide](https://github.com/dbt-labs/dbt-core/blob/main/CONTRIBUTING.md) and understand what's expected of me.
- [x] I have run this code in development, and it appears to resolve the stated issue.
- [x] This PR includes tests, or tests are not required or relevant for this PR.
- [x] This PR has no interface changes (e.g., macros, CLI, logs, JSON artifacts, config files, adapter interface, etc.) or this PR has already received feedback and approval from Product or DX.
- [x] This PR includes [type annotations](https://docs.python.org/3/library/typing.html) for new and modified functions.

PR #10756, opened 2024-09-22
Add required 'begin' config support for microbatch models

MichelleArk:
Resolves https://github.com/dbt-labs/dbt-core/issues/10701

### Problem

We need to know what the "beginning of time" is for a microbatch model, to know when to start for these two cases:
* The first time this model is run
* If you run in “full-refresh-mode”

This is distinct from the CLI parameter --event-time-start which can also be used when running in "incremental" mode.

### Solution

* continue to return early if --event-time-start is provided
* For non-incremental runs (or if there is no checkpoint to lookback from), return the `begin` config! (raise exception if its not available, and further parse-time validation will be introduced as part of https://github.com/dbt-labs/dbt-core/pull/10752
* update schemas.getdbt.com v12 manifest
* update tests to provide `begin` config.

### Checklist


MichelleArk:
CI should pass once https://github.com/dbt-labs/schemas.getdbt.com/pull/57 is merged

QMalcolm (approved):
Woooo! This looks great 😄 I made one comment on updating a doc string, and we should probably fix it before merging, but also didn't want my review to be blocking as you'll be back online before I am.

PR #10767, opened 2024-09-24
add builder config to test node config

### LINKED ISSUE #10484: [Bug] generic tests to accept arbitrary configs does not appear to be pulling in config options for certain tests

### Current Behavior

while trying respond to a slack conversation [thread](https://dbt-labs.slack.com/archives/C017GDLAF7D/p1721696763124499)

about implementing the new general config options for `data_tests`

found that local testing showed that some tests types are not accepting the `config` object it seems 
```
version: 2
models:
  - name: my_test_model  # Replace with your actual model name
    columns:
      - name: id
        data_tests:
          - not_null:
              config:
                snowflake_warehouse: analytics
```

logs test may pass but logs will not show `snowflake_warehouse` being set

best guess is tests that just trigger a macro without having any real additional config are for some reason now not able to pull in the general config.

### Expected Behavior

all test types should be able to have config options added based on the work done around https://github.com/dbt-labs/dbt-core/issues/10197

### Steps To Reproduce

create various simple models to test against

schema.yml

```
version: 2
models:
  - name: my_test_model # Replace with your actual model name
    columns:
      - name: color
        tests:
          - accepted_values:
              values: ['blue', 'red', 'green', 'yellow']
              config:
                snowflake_warehouse: analytics

# can sub out for a not_null test to match above

 - name: my_test_model  # Replace with your actual model name
    columns:
      - name: id
        data_tests:
          - not_null:
              config:
                snowflake_warehouse: analytics
            

  - name: child_table
    description: "Child table containing records referencing the parent table"
    columns:
      - name: parent_id
        data_tests:
          - relationships:
              to: ref('parent_table')
              field: id
              config:
                snowflake_warehouse: dbt_snowflake_warehouse
```

models
```
child_table.sql
select * from {{ ref('child') }}

parent_table.sql
select * from {{ ref('parent') }}

my_test_model.sql
select * from {{ ref('seed') }}
```
seeds
```
child.csv

id,parent_id,value
1,1,Value 1
2,1,Value 2
3,2,Value 3
4,3,Value 4  # Corrected parent_id to 3 which exists in parent_table

parent.csv

id,name
1,Parent A
2,Parent B
3,Parent C

seed.csv

id,color,value
1,blue,10
2,red,20
3,green,30
4,yellow,40
5,blue,50
6,red,60
7,blue,70
8,green,80
9,yellow,90
10,blue,100
```

### Relevant log output

_No response_

### Environment

```markdown
- OS:
- Python:
- dbt:
```

_No response_

### Additional Context

_No response_

colin-rogers-dbt:
Resolves #10484 

### Checklist


colin-rogers-dbt:
Failing dbt-snowflake test against main: https://github.com/dbt-labs/dbt-snowflake/actions/runs/11042144022/job/30673904052#step:8:575

Succesful tests running against this PR branch: https://github.com/dbt-labs/dbt-snowflake/actions/runs/11041928244/job/30673264899#step:8:572

gshank (changes requested):
Prior to this "add_config_call" was generally used for SQL nodes to store the extracted config for re-application after applying schema changes.  But generic nodes are a fundamentally different thing making it a bit hard to reason about.  Since they are constructed only from schema file definitions, there is no "base" sql file to apply later. The config_call_dict on a node is also stored in the internal manifest (for a partial parsing reason that I remember knowing at one point but am not sure really applies anymore...).  But generic tests are very leaf-y things that will always be re-generated when anything above them changes, so having it stored on the generic test node probably isn't an issue.

I think this is fine, but please add some comments to the place where the "add_config_call" is done about why (and how it's different than other uses of that call), and add some tests.

colin-rogers-dbt:
> ic nodes are a fundamentally different thing making it a bit hard to reason about. Since they are constructed only from schema file definitions, there is no "base" sql file t

I have added a short code comment but can elaborate further, I have tried to add a test that validates this but all I have achieved is narrowing down the scope of impact: it appears this _only_ ensure the config is available when the pre_model_hook is run. Since dbt-postgres doesn't definine a generic pre_model_hook (like dbt-snowflake) we can't test this behavior here

gshank (approved):
There are some failing tests, but I'm approving so you don't have to come back again.

PR #10770, opened 2024-09-24
Add '--inline-direct' parameter to 'dbt show'.

### LINKED ISSUE #10778: [Feature] Add a `--inline-direct` flag for `dbt show`

### Describe the feature

Introduce a new `--inline-direct` flag for `dbt show` that:
- avoids loading the entire manifest, and
- skips rendering any Jinja templates

i.e. speed up execution time of doing the equivalent of `dbt show --inline` when there is no Jinja templating that is needed. For example, the following doesn't have any Jinja templating and doesn't need the manifest, so we should be able to skip both of those:

```shell
dbt show --inline-direct "select * from my_table"
```

### Usage example

With this feature, you can call dbt show with `--inline-direct` rather than `--inline`:

```shell
dbt show --inline-direct "select * from my_table"
```

_No response_

### Who will this benefit?

_No response_

_No response_

### Anything else?

_No response_

peterallenwebb:
resolves #10778

### Problem

The 'dbt show' command loaded the manifest and ran jinja templating, even when these steps were not needed.

### Solution

Add a new "--inline-direct" parameter to skip the manifest load and templating steps.

### Checklist

- [] This PR includes [type annotations](https://docs.python.org/3/library/typing.html) for new and modified functions.

### DIFF
--- a/core/dbt/cli/main.py
+++ b/core/dbt/cli/main.py
@@ -8,12 +8,15 @@ from click.exceptions import BadOptionUsage
 from click.exceptions import Exit as ClickExit
 from click.exceptions import NoSuchOption, UsageError
 
+from dbt.adapters.factory import register_adapter
 from dbt.artifacts.schemas.catalog import CatalogArtifact
 from dbt.artifacts.schemas.run import RunExecutionResult
 from dbt.cli import params as p
 from dbt.cli import requires
 from dbt.cli.exceptions import DbtInternalException, DbtUsageException
+from dbt.cli.requires import setup_manifest
 from dbt.contracts.graph.manifest import Manifest
+from dbt.mp_context import get_mp_context
 from dbt_common.events.base_types import EventMsg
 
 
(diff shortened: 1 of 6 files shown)


peterallenwebb:
Note that I have disabled several type checks because we need to make several changes to AdapterProtocol before types in this changeset can be adequately annotated.

ChenyuLInx (approved):

core/dbt/cli/main.py
```diff
@@ -362,17 +366,26 @@ def compile(ctx, **kwargs):
...
-    task = ShowTask(
-        ctx.obj["flags"],
-        ctx.obj["runtime_config"],
-        ctx.obj["manifest"],
-    )
+    from dbt.task.show import ShowTask, ShowTaskDirect
+
+    if ctx.obj["flags"].inline_direct:
+        # Issue the inline query directly, wit no templating. Does not require
+        # loading the mainfest.
```
dbeatty10, line 376:
```suggestion
        # Issue the inline query directly, with no templating. Does not require
        # loading the manifest.
```

aranke (approved):
Please add a failure test case too, otherwise LGTM!

PR #10780, opened 2024-09-25
Enable setting datetime value for dbt_valid_to when the record is current

### LINKED ISSUE #10187: Allow custom date for dbt_valid_to in snapshots
### Description

Right now, `dbt_valid_to` is set to `NULL` for current records.

Some people, want to set `dbt_valid_to` to an arbitrary future date (so that their snapshot will work for a join, corporate standard, etc.).

The way they do this today, is create a view on top of their snapshot:

```sql
select 
	coalesce(dbt_valid_to, date(9999, 12, 31)) as fixed_dbt_valid_to,
...
from {{ ref('my_snapshot') }}
```

### Acceptance criteria

- new config to define what you want to set `dbt_valid_to` for the current records in your snapshot
    - default is `NULL` (current behavior)
    ```
    snapshots:
      my_project:
        +dbt_valid_to_current_indicator: "to_date('2024-05-10')"
    ```
- can set config as project default or for individual snapshot
- config can be equal to:
    - SQL statement ('NULL', hard coded date/timestamp, etc.) `"to_date('2024-05-10')"`
    - [ideally also] project variable that returns a SQL statement `'{{ var('my_future_date') }}'`
    - [ideally also] macro that returns a SQL statement `'{{ dbt.date(9999, 12, 31) }}'`
- if someone updates this config, we:
    - because we rely on knowing what `dbt_valid_to` is equal to for current records for the insertion logic when snapshotting, the existing snapshot **must be updated** before we can proceed
    - how can we throw an error:
      - if there was a `dbt_is_current` column (as has been requested in another feature request) we could run a validation check to say "is every row with `is_current = true` also have `dbt_valid_to` equal to whatever I have set for my `dbt_valid_to_for_current_records` config); could have a performance implication
      - [not everyone uses CI] similar to how we handle [versioned model contract changes](https://docs.getdbt.com/reference/resource-properties/versions#detecting-breaking-changes), if someone updates this config and runs their snapshot using deferral and `state:modified` we should throw a warning/error that this config has been updated (make them go in and manually update their existing snapshot)

### Notes
Ideas for how to handle config changes:
- ["silent failure"] we just use the new config for all future inserts 
- [out of scope, but potential future enhancement] update all historic "current" records to have the new value (would this be possible?)
- ideas for config name: `dbt_valid_to_for_current_records`, `dbt_valid_to_current_records`, `dbt_valid_to_default`, `dbt_valid_to_current_indicator`, `dbt_valid_to_current`, `dbt_valid_to_current_date`, `dbt_valid_to_current_marker`, `coalesce_dbt_valid_to`

Screenshot 2024-06-27 at 3 51 31 PM

gshank:
Resolves #10187

### Problem

Currently we use a null to indicate that a snapshot record is current. Users would like to use some arbitrary future date to indicate current instead, because it makes some SQL easier.

### Solution

Add a new attribute "dbt_valid_to_current", and implement changes in adapters to use it.

### Checklist


peterallenwebb (approved):

peterallenwebb:
**Reminder**, we need to unpin the dependency branches before merge. Otherwise this looks good.

PR #10782, opened 2024-09-25
Improve performance of infer primary key

### LINKED ISSUE #10781: Improve performance of "process_model_inferred_primary_keys"
### Description

The method "process_model_inferred_primary_keys" method that was added by [Add `primary_key` to manifest by dave-connors-3 · Pull Request #10096 · dbt-labs/dbt-core](https://github.com/dbt-labs/dbt-core/pull/10096)  does not process only models that have been changed, but every single model in the manifest, which is not how it ought to work. 

There may be other issues with gathering the generic tests to infer the primary key, but that's at the top of the list.

gshank:
Resolves #10781

### Problem
Poor performance of inferring primary key.

### Solution

Do not process primary keys for unchanged models. When inferring a primary key, start by making a map of models to generic tests.

### Checklist

### DIFF
--- a/core/dbt/parser/manifest.py
+++ b/core/dbt/parser/manifest.py
@@ -1158,7 +1158,9 @@ class ManifestLoader:
                 continue
             if not model_to_generic_test_map:
                 model_to_generic_test_map = self.build_model_to_generic_tests_map()
-            generic_tests = model_to_generic_test_map[node.unique_id]
+            generic_tests: List[GenericTestNode] = []
+            if node.unique_id in model_to_generic_test_map:
+                generic_tests = model_to_generic_test_map[node.unique_id]
             primary_key = node.infer_primary_key(generic_tests)
             node.primary_key = sorted(primary_key)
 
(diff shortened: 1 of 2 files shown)


core/dbt/parser/manifest.py
```diff
@@ -1150,10 +1150,17 @@ def process_saved_queries(self, config: RuntimeConfig):
 
     def process_model_inferred_primary_keys(self):
         """Processes Model nodes to populate their `primary_key`."""
+        model_to_generic_test_map: Dict[str, List[GenericTestNode]] = {}
         for node in self.manifest.nodes.values():
             if not isinstance(node, ModelNode):
                 continue
-            generic_tests = self._get_generic_tests_for_model(node)
+            if node.created_at < self.started_at:
```
ChenyuLInx, line 1157: Why we can continue here because this node is from partial parting which means it didn't change? maybe a short comment?

ChenyuLInx (approved):
LGTM! A small nit that is not blocking

core/dbt/parser/manifest.py
gshank: Yes, because it hasn't changed on this parse. That equivalent line occurs in 19 other places, so I'm not going to bother documenting it here :)

PR #10784, opened 2024-09-26
Attempt to skip saved query processing when no semantic manifest changes

### LINKED ISSUE #10563: Do not execute process_saved_queries if nothing in semantic manifest has changed
### Description

Currently we run process_saved_queries for every saved query in a manifest, even if nothing has changed in the semantic manifest. In the right circumstances this can be a performance problem.

### Acceptance Criteria

Check for actual changes to the semantic manifest before executing process_saved_queries for every saved query.

gshank:
Resolves #10563

### Problem

We run "process_saved_queries" even when nothing has changed in the semantic manifest.

### Solution

Check for changes to saved queries, semantic models, and metrics before executing process saved queries. This will have false positives because of cascading re-parsing, but will enable us to skip at least some unnecessary saved query processing.

### Checklist

### DIFF
--- a/core/dbt/parser/manifest.py
+++ b/core/dbt/parser/manifest.py
@@ -1145,12 +1145,18 @@ class ManifestLoader:
         # Note: This will also capture various nodes which have been re-parsed
         # because they refer to some other changed node, so there will be
         # false positives. Ideally we would compare actual changes.
+        semantic_manifest_changed = False
         semantic_manifest_nodes: chain[SemanticManifestNode] = chain(
             self.manifest.saved_queries.values(),
             self.manifest.semantic_models.values(),
             self.manifest.metrics.values(),
         )
-        if any(node.created_at > self.started_at for node in semantic_manifest_nodes):
+        for node in semantic_manifest_nodes:
+            # Check if this node has been modified in this parsing run
+            if node.created_at > self.started_at:
+                semantic_manifest_changed = True
+                break  # as soon as we run into one changed node we can stop
+        if semantic_manifest_changed is False:
             return
 
         current_project = config.project_name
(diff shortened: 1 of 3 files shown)


core/dbt/parser/manifest.py
```diff
@@ -1141,6 +1141,24 @@ def process_metrics(self, config: RuntimeConfig):
 
     def process_saved_queries(self, config: RuntimeConfig):
         """Processes SavedQuery nodes to populate their `depends_on`."""
+        # Note: This will also capture various nodes which have been re-parsed
+        # because they refer to some other changed node, so there will be
+        # false positives. Ideally we would compare actual changes.
+        semantic_manifest_changed = False
+        SemanticManifestNode = Union[SavedQuery, SemanticModel, Metric]
```
aranke, line 1148: Nit: Can we put this definition in a constants file somewhere?

core/dbt/parser/manifest.py
```diff
@@ -1141,6 +1141,24 @@ def process_metrics(self, config: RuntimeConfig):
...
+        semantic_manifest_nodes: chain[SemanticManifestNode] = chain(
+            self.manifest.saved_queries.values(),
+            self.manifest.semantic_models.values(),
+            self.manifest.metrics.values(),
+        )
+        for node in semantic_manifest_nodes:
```
aranke, line 1154: Nit: This can be simplified to

```
if any(node.created_at > self.started_at for node in semantic_manifest_nodes):
    return
```

aranke (approved):
LGTM

QMalcolm (approved):

core/dbt/parser/manifest.py
gshank: The logic of "any" is backward, we want to execute the process metrics for node code instead of returning, but doing it that way would make merging into dbt-mantle more muddy because it executes a different function, so I'm going to leave this as it is.

core/dbt/parser/manifest.py
gshank: Moved it to core/dbt/contracts/graph/nodes.py

PR #10788, opened 2024-09-26
[CT-10785] Microbatch models  should respect `full_refresh` model config

### LINKED ISSUE #10785: Ensure microbatch models respect model level `full_refresh` config during batch creation
In dbt-core, it is generally expected that values passed via CLI flags take precedence over model level configs. However, `full_refresh` on a model is an exception to this rule, where in the model config takes precedence. This config exists specifically to _prevent_ accidental full refreshes of large incremental models, as doing so can be costly. **_It is actually best practice to set `full_refresh=False` on incremental models_.**

Currently, for microbatch models the above is not happening. The CLI flag `--full-refresh` is taking precedence over the model config `full_refresh`. That means that currently if `--full-refresh` is supplied then the microbatch model **_will be refreshed_** even if `full_refresh=False` is set on the model. This needs to not be the case.

QMalcolm:
Resolves #10785 

### Problem

In dbt-core, it is generally expected that values passed via CLI flags take
precedence over model level configs. However, `full_refresh` on a model is an
exception to this rule, where in the model config takes precedence. This
config exists specifically to _prevent_ accidental full refreshes of large
incremental models, as doing so can be costly. **_It is actually best
practice_** to set `full_refresh=False` on incremental models.

Prior to this commit, for microbatch models, the above was not happening. The
CLI flag `--full-refresh` was taking precedence over the model config
`full_refresh`. That meant that if `--full-refresh` was supplied, then the
microbatch model **_would full refresh_** even if `full_refresh=False` was
set on the model. This commit solves that problem.

### Solution

Update `_is_incremental` to prefer the `model.config.full_refresh` value over the `RuntimeConfig.args.FULL_REFRESH` value provided by CLI flags.

### Checklist

### DIFF
--- a/.changes/unreleased/Features-20240926-153210.yaml
+++ b/.changes/unreleased/Features-20240926-153210.yaml
@@ -1,0 +1,6 @@
+kind: Features
+body: Ensure microbatch models respect `full_refresh` model config
+time: 2024-09-26T15:32:10.202789-05:00
+custom:
+  Author: QMalcolm
+  Issue: "10785"
--- a/core/dbt/task/run.py
+++ b/core/dbt/task/run.py
@@ -550,12 +550,17 @@ class ModelRunner(CompileRunner):
         relation = self.adapter.get_relation(
             relation_info.database, relation_info.schema, relation_info.name
         )
-        return (
+        if (
             relation is not None
             and relation.type == "table"
             and model.config.materialized == "incremental"
-            and not (getattr(self.config.args, "FULL_REFRESH", False) or model.config.full_refresh)
-        )
+        ):
+            if model.config.full_refresh is not None:
+                return not model.config.full_refresh
+            else:
+                return not getattr(self.config.args, "FULL_REFRESH", False)
+        else:
+            return False
 
 
 class RunTask(CompileTask):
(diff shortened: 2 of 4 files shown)


ChenyuLInx (dismissed):
The code looks good but I have some questions about the new behavior.

~Is this a behavior change for the existing incremental model also?
My understanding is with the new overwrite order, the user can not do a full refresh without changing the code. Is this correct? If yes is this what we want? @graciegoheen 

Feel free to dismiss my review if my understanding is wrong.~

My understanding was wrong

QMalcolm:
> The code looks good but I have some questions about the new behavior.
> 
> Is this a behavior change for the existing incremental model also? My understanding is with the new overwrite order, the user can not do a full refresh without changing the code. Is this correct? If yes is this what we want? @graciegoheen
> 
> Feel free to dismiss my review if my understanding is wrong.

This is _not_ new behavior for incremental models. This is already how incremental models operate on the `full_refresh` model config. This is only a change to make _microbatch_ models following the same pattern. [Here is the documentation](https://docs.getdbt.com/reference/resource-configs/full_refresh#description) on the `full_refresh` model config.

ChenyuLInx (approved):
🚢

PR #10792, opened 2024-09-27
[Round 2] Fix #9005: Allow singular tests to be documented in properties.yml

### LINKED ISSUE #9005: [CT-3323] Allow singular tests to be documented
### Description

Currently, you can document pretty much everything but a [data test](https://docs.getdbt.com/docs/build/data-tests#overview) - if you try, it fails on a parsing error. This ticket is for [singular data tests](https://docs.getdbt.com/docs/build/data-tests#singular-data-tests). A separate ticket (#2578) addresses [generic data tests](https://docs.getdbt.com/docs/build/data-tests#generic-data-tests).

Currently we don't have a "tests" section in schema files. In order to allow setting descriptions in yaml files we should add one. (As a side issue, the SingularTestNodes use the TestConfig, which is not particularly valid for SingularTestNodes. Maybe we need a separate SingularTestConfig.)

https://docs.getdbt.com/docs/build/tests#singular-tests

### Acceptance Criteria

* You can add a description to a singular test in a yaml file
* You can use a docs block to capture your test description
* That description shows up in the docs site

### Considerations

* Do we need to exclude this from state modified? 
* New section in schema files and new schema parser
* Don't have a test schema parser 
* Need yml to be able to have a description

aranke:
Resolves #

### Checklist

### DIFF
--- a/tests/functional/data_test_patch/test_singular_test_patch.py
+++ b/tests/functional/data_test_patch/test_singular_test_patch.py
@@ -1,0 +1,67 @@
+import os
+
+import pytest
+from functional.data_test_patch.fixtures import tests__malformed_schema_yml
+
+from dbt.tests.util import get_artifact, run_dbt, run_dbt_and_capture
+from tests.functional.data_test_patch.fixtures import (
+    tests__doc_block_md,
+    tests__invalid_name_schema_yml,
+    tests__my_singular_test_sql,
+    tests__schema_yml,
+)
+
+
+class TestPatchSingularTest:
+    @pytest.fixture(scope="class")
+    def tests(self):
+        return {
+            "my_singular_test.sql": tests__my_singular_test_sql,
+            "schema.yml": tests__schema_yml,
+            "doc_block.md": tests__doc_block_md,
+        }
+
+    def test_compile(self, project):
+        run_dbt(["compile"])
+        manifest = get_artifact(project.project_root, "target", "manifest.json")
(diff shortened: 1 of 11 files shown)


gshank (approved):
This looks good, but I do wonder if we might want to give a more detailed explanation in the warning, mentioning that "data_tests" is a newly parsed key? Or something...

tests/functional/data_test_patch/test_singular_test_patch.py
```diff
@@ -0,0 +1,69 @@
...
+        }
+
+    def test_compile(self, project):
+        _, log_output = run_dbt_and_capture(["compile"])
+
+        file_path = (
+            "tests\\schema_with_invalid_name.yml"
+            if os.name == "nt"
+            else "tests/schema_with_invalid_name.yml"
+        )
```
dbeatty10, line 50: Did you happen to try a cross-platform implementation like [this](https://docs.python.org/3/library/pathlib.html#basic-use) instead of needing specific conditionals like `if os.name == "nt"`?

```python
from pathlib import Path

file_path = Path("tests/schema_with_invalid_name.yml")
```

I just noticed this while helping write the documentation for this awesome feature within https://github.com/dbt-labs/docs.getdbt.com/pull/6147, so feel free to disregard this suggestion!

aranke: Good call-out, I did not. 
Trying now :)

PR #10793, opened 2024-09-30
state:modified vars, behind "state_modified_compare_vars" behaviour flag

### LINKED ISSUE #4304: [vars] [false negatives] Changes in used variables do not cause models to be selected by `state:modified`
Hi there,

In https://docs.getdbt.com/reference/node-selection/state-comparison-caveats, it is stated:

> ### Vars

> If a model uses a `var` or `env_var` in its definition, dbt is unable today to identify that lineage in such a way that it can include the model in `state:modified` because the `var` or `env_var` value has changed.

I was wondering if there is an issue tracking this limitation; I searched for one and couldn't find one, so I hope it's OK that I opened this issue so that our team can follow its status. (If there was one and I missed it, apologies!)

Also, I apologize for the naive question, but would it not be easier, at least for the case where the desired effect is to re-run all changed models and their children, to check if the source code *after* preprocessing (Jinja macro expansion) has changed, instead of *before*? I would expect that this would automatically cause changes in variables to propagate to changes to the compiled code of the models, which would then indicate a need to re-run them.

Thanks!

MichelleArk:
Resolves #4304

### Problem

With https://github.com/dbt-labs/dbt-core/pull/10675, we removed the capability for dbt to (sometimes) detect changes in vars as modifying models/dbt resources that depended on them (behind a behaviour flag).

### Solution

This behaviour is actually desired (distinction between env_var and var), so let's introduce state:modified.vars as a standalone selector method as well as a default. This is also behind a behaviour flag in case existing users rely on the current var state:modified selection behaviour (flawed as it is!). 

### Checklist


MichelleArk:
Associated schemas PR: https://github.com/dbt-labs/schemas.getdbt.com/pull/64

gshank (approved):

core/dbt/context/configured.py
```diff
@@ -58,19 +70,21 @@ def __call__(self, var_name, default=Var._VAR_NOTSET):
             all_vars.add(my_config.vars.vars_for(lookup, adapter_type))
         all_vars.add(active_vars)
 
-        if var_name in all_vars:
-            return all_vars[var_name]
+        if not var_found and var_name in all_vars:
```
MichelleArk, line 73: This refactor was primarily for the happiness of our linting tools. There was no final return statement on the method in the past.

PR #10799, opened 2024-09-30
Replace environment variable with a project flag to gate microbatch functionality

### LINKED ISSUE #10798: Replace DBT_EXPERIMENTAL_MICROBATCH env var with behaviour flag
Prior to the 1.9 beta, we should: 
1.  Replace usage of env var `DBT_EXPERIMENTAL_MICROBATCH` with a project flag, `require_builtin_microbatch_strategy` which defaults to `False` (@graciegoheen approves the name :) )

2. Make changes to dbt-adapters:
* this needs to use the project flag instead of os env var: https://github.com/dbt-labs/dbt-adapters/blob/b414668cd046ce9b68c03b68ca7fc94f190b4e99/dbt/adapters/base/impl.py#L1573
* base test needs to use the new project flag instead of env var: [https://github.com/dbt-labs/dbt-adapters/blob/b414668cd046ce9b68c03b68ca7fc94f190b[…]er/dbt/tests/adapter/incremental/test_incremental_microbatch.py](https://github.com/dbt-labs/dbt-adapters/blob/b414668cd046ce9b68c03b68ca7fc94f190b4e99/dbt-tests-adapter/dbt/tests/adapter/incremental/test_incremental_microbatch.py#L66)
* example: https://github.com/dbt-labs/dbt-core/pull/10799/files#diff-ac70d9689da91b61cf649606ebbc0cd875e46dfd8c3627dd8f8d124e08f60fd5R86-R93

3. Make changes to dbt-bigquery:
* a couple additional tests have been added beyond the base test
the additional tests patch the os env var directly
* https://github.com/dbt-labs/dbt-bigquery/blob/main/tests/functional/adapter/incremental/test_incremental_microbatch.py#L33
* https://github.com/dbt-labs/dbt-bigquery/blob/main/tests/functional/adapter/incremental/test_incremental_microbatch.py#L48
they'll need to be updated to use the project_config_update fixture instead

dbt-postgres and dbt-snowflake use the base test only and don't need any manual updates

MichelleArk:
Resolves #10798

### Problem

We had gated the new microbatch feature behind an environment variable in the initial implementation of microbatch (https://github.com/dbt-labs/dbt-core/pull/10594). However, for a better experience, we want people to be setting a project flag (AKA behavior flag).

### Solution

Gate microbatch functionality behind a project flag (AKA behavior flag) instead of an environment variable.

### Checklist


dev-requirements.txt
```diff
@@ -1,5 +1,5 @@
-git+https://github.com/dbt-labs/dbt-adapters.git@main
-git+https://github.com/dbt-labs/dbt-adapters.git@main#subdirectory=dbt-tests-adapter
+git+https://github.com/dbt-labs/dbt-adapters.git@microbatch-behavior-flag
+git+https://github.com/dbt-labs/dbt-adapters.git@microbatch-behavior-flag#subdirectory=dbt-tests-adapter
```
QMalcolm, line 2: This is needed to get the tests to pass. This is only necessary until the changes in dbt-adapters are merged. Said another way: we need to merge the changes to dbt-adapters first, and then revert these dependency changes before merging this PR's changes.

MichelleArk, line 2: reminder: let's make sure we bump the lower bound pin of `dbt-adapters` in setup.py as part of this PR once the dbt-adapters branches are merged.

core/dbt/events/types.py
```diff
@@ -486,6 +486,16 @@ def message(self) -> str:
         return line_wrap_message(warning_tag(description))
 
 
+class MicrobatchMacroOutsideOfBatchesDeprecation(WarnLevel):
+    def code(self) -> str:
+        return "D020"
+
+    def message(self) -> str:
+        description = "The use of a custom microbatch macro outside of batched execution is deprecated. To use it with batched execution, set `flags.require_batched_execution_for_custom_microbatch_strategy` to `True` in `dbt_project.yml`. In the future this will be the default behavior."
```
MichelleArk, line 494: super nit, just because this is user-facing: "In the future this will be the default behavior." -> "In the future batched execution will be the default behaviour."

MichelleArk: We could also mention the name of the custom macro that exists (e.g. `get_incremental_microbatch_sql` as part of the message in case the user that's seeing this doesn't know about the custom strategy macro naming convention offhand.

MichelleArk (commented):
✅ Can't approve a PR I "authored" but here is my psuedo-approval!

Thank you for taking this across the finish line @QMalcolm, clean + precise work ✨

QMalcolm (approved):
Self approving because I didn't open the PR 🙈

mirnawong1:
hey @QMalcolm @MichelleArk , the docs team received [this github issue](https://github.com/dbt-labs/docs.getdbt.com/issues/6447).  

i was reviewing this [dbt-adapters PR](https://github.com/dbt-labs/dbt-adapters/pull/323/files) and https://github.com/dbt-labs/dbt-core/pull/10799 -- am i correct in thinking that the following needs to be added to the [behavior flag doc](https://docs.getdbt.com/reference/global-configs/behavior-changes#behavior-change-flags):

require_batched_execution_for_custom_microbatch_strategy flag defaults to False, which will raise a deprecation warning. Set it to True to allow custom microbatch strategies (like get_incremental_microbatch_sql) to use batched execution.

QMalcolm:
> hey @QMalcolm @MichelleArk , the docs team received [this github issue](https://github.com/dbt-labs/docs.getdbt.com/issues/6447).
> 
> i was reviewing this [dbt-adapters PR](https://github.com/dbt-labs/dbt-adapters/pull/323/files) and #10799 -- am i correct in thinking that the following needs to be added to the [behavior flag doc](https://docs.getdbt.com/reference/global-configs/behavior-changes#behavior-change-flags):
> 
> > the `require_batched_execution_for_custom_microbatch_strategy` flag defaults to False, which will raise a deprecation warning. Set it to True to allow custom microbatch strategies (like get_incremental_microbatch_sql) to use batched execution.

@mirnawong1 It's actually a little bit different than that. The flag `require_batched_execution_for_custom_microbatch_strategy` does default to `False`. However, a user _only_ needs to set `require_batched_execution_for_custom_microbatch_strategy` to `True` if they have a _custom_ microbatch macro. If they don't have a custom microbatch macro, everything will work. Additionally, they'll only get a depreciation warning IFF they have a custom microbatch strategy and the flag is `False`.

mirnawong1:
hey @QMalcolm oh ok i see! so `require_batched_execution_for_custom_microbatch_strategy` is for custom microbatch macros only. 

so does this also mean we have another flag: `require_builtin_microbatch_strategy` ([pr here](https://github.com/dbt-labs/dbt-core/issues/10798)), which replaces the DBT_EXPERIMENTAL_MICROBATCH env var? and they should set to `true` if they had the env_var already enabled?

QMalcolm:
@mirnawong1 There is no other flag. I recognize that as slightly confusing. For context, the only reason we gated microbatch in the first place was the fear of breaking people's projects who already had a custom microbatch macro. What we've done here is make it so one _only_ needs to set a flag if they have a custom microbatch macro, instead of making _everyone_ set a flag because someone else's project might have a microbatch macro.

For a person setting up a microbatch model for the first time, there is no longer a flag or environment variable they need to set (assuming they don't have a custom microbatch macro)

mirnawong1:
oh ok got it! so i can update the callout [here](https://docs.getdbt.com/docs/build/incremental-microbatch) and replace it with content informing users of the behavior flag _should they have a custom microbatch macro only_. all other users that are setting up incremental microbatch strategy for the first time w no existing custom macro don't need to do anything apart from their config. 

thank you so much for clarifying this!

PR #10802, opened 2024-09-30
When retrying microbatch models, propagate prior successful state

### LINKED ISSUE #10800: Ensure prior batch successes are propagated in RunResults during multiple `dbt retry` scenarios
In https://github.com/dbt-labs/dbt-core/pull/10751/commits/a6a1ef87b172972062cbe539576b246d5479af53 of https://github.com/dbt-labs/dbt-core/pull/10751 we added some edge case handling logic such that if _every_ batch for a microbatch model failed, the batches would be regenerated on retry and true/false on `full-refresh` would be properly applied. This was to address the following edge case:

1. a microbatch model had a `begin` of `2020-01-01` and has been running this way for a while
2. the `begin` config has changed to `2024-01-01` and `dbt run --full-refresh` gets run
3. every batch for a microbatch model fails
4. on `dbt retry` the relation for the model is said to exist, and the now out of range data (2020-01-01 through 2023-12-31) is never purged

The solution in https://github.com/dbt-labs/dbt-core/pull/10751/commits/a6a1ef87b172972062cbe539576b246d5479af53 did solve this edge case, however it also created another:

1. On `dbt run` a microbatch model has 2 batches that succeed and 1 batch that fails
2. On the first invocation of `dbt retry` the 1 previously failed batch is retried and fails again
3. `dbt retry` is run a _second time_ and now 3 batches are run (the initial two that succeeded, and the failing one)

At step (3) _only_ the failing batch should be run. However, presently our logic is that if a prior run has _only_ failed batches, then we rerun the microbatch model as it did on `dbt run`. We need on the second retry to know that we had previously successful batches for the initial run being retried. However, we also must continue to solve the edge case we resolved in https://github.com/dbt-labs/dbt-core/pull/10751/commits/a6a1ef87b172972062cbe539576b246d5479af53.
​

QMalcolm:
Resolves #10800 

### Problem

If you invoked `dbt retry` on a microbatch model twice, where on the first `dbt retry` all the retried batches failed, then on the second `dbt retry` invocation the microbatch model would rerun all the batches that were run on the initial `dbt run` being retried.

Prior bad behavior

### Solution

When going through multiple `dbt retry` invocations, ensure that the prior `successful` batch information continues to be passed.

New good behavior

### Checklist

### DIFF
--- a/core/dbt/contracts/graph/nodes.py
+++ b/core/dbt/contracts/graph/nodes.py
@@ -60,7 +60,7 @@ from dbt.artifacts.resources import SourceDefinition as SourceDefinitionResource
 from dbt.artifacts.resources import SqlOperation as SqlOperationResource
 from dbt.artifacts.resources import TimeSpine
 from dbt.artifacts.resources import UnitTestDefinition as UnitTestDefinitionResource
-from dbt.artifacts.schemas.batch_results import BatchType
+from dbt.artifacts.schemas.batch_results import BatchResults
 from dbt.contracts.graph.model_config import UnitTestNodeConfig
 from dbt.contracts.graph.node_args import ModelNodeArgs
 from dbt.contracts.graph.unparsed import (
@@ -454,7 +454,7 @@ class HookNode(HookNodeResource, CompiledNode):
 
 @dataclass
 class ModelNode(ModelResource, CompiledNode):
-    batches: Optional[List[BatchType]] = None
+    batch_info: Optional[BatchResults] = None
 
     @classmethod
     def resource_class(cls) -> Type[ModelResource]:
--- a/core/dbt/task/run.py
+++ b/core/dbt/task/run.py
@@ -478,6 +478,9 @@ class ModelRunner(CompileRunner):
     ) -> List[RunResult]:
         batch_results: List[RunResult] = []
 
+        # Note currently (9/30/2024) model.batch_info is only ever _not_ `None`
+        # IFF `dbt retry` is being run and the microbatch model had batches which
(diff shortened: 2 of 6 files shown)


core/dbt/contracts/graph/nodes.py
```diff
@@ -454,7 +454,7 @@ def resource_class(cls) -> Type[HookNodeResource]:
 
 @dataclass
 class ModelNode(ModelResource, CompiledNode):
-    batches: Optional[List[BatchType]] = None
+    batch_info: Optional[BatchResults] = None
```
ChenyuLInx, line 457: Will this cause a loading error for a previous manifest?

core/dbt/task/run.py
```diff
@@ -481,8 +489,8 @@ def _execute_microbatch_materialization(
             start = microbatch_builder.build_start_time(end)
             batches = microbatch_builder.build_batches(start, end)
         else:
```
ChenyuLInx, line 491: Can we add a comment here mentioning this only happens during retry?

ChenyuLInx (approved):
One question and a nit but the change looks good to me!

core/dbt/contracts/graph/nodes.py
QMalcolm: This has no effect on the written manifest, which is intentional. That is the case because these do not exist on the artifact class for the object

core/dbt/task/run.py
QMalcolm: Can do!

PR #10820, opened 2024-10-03
Validate `--event-time-start` is before `--event-time-end`

### LINKED ISSUE #10786: Validate --event-time-start is earlier than --event-time-end

QMalcolm:
Resolves #10786 

### Problem

A person could specify an `--event-time-start` that was after `--event-time-end` or the current time. This would cause bad things to happen ™️ 

### Solution

Validate the cli options such that
* `--event-time-start` is _before_ `--event-time-end`
* `--event-time-start` is _before_ the current time if `--event-time-end` is not set

### Checklist

### DIFF
--- a/core/dbt/cli/flags.py
+++ b/core/dbt/cli/flags.py
@@ -7,6 +7,7 @@ from pathlib import Path
 from pprint import pformat as pf
 from typing import Any, Callable, Dict, List, Optional, Set, Union
 
+import pytz
 from click import Context, Parameter, get_current_context
 from click.core import Command as ClickCommand
 from click.core import Group, ParameterSource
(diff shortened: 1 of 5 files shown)


aranke (changes requested):
Can you please move flag validation logic here to be consistent with other flags? That'll avoid the need to create a new Click option as well.

https://github.com/dbt-labs/dbt-core/blob/75a09621cd0f3baaf01bdf4c409671f26ee45e67/core/dbt/cli/flags.py#L311

core/dbt/cli/flags.py
```diff
@@ -349,6 +353,24 @@ def _assert_mutually_exclusive(
...
+        if event_time_start is not None:
+            event_time_end = (
+                getattr(self, "EVENT_TIME_END") if hasattr(self, "EVENT_TIME_END") else None
+            )
+
+            if event_time_end is not None and event_time_start >= event_time_end:
+                raise DbtUsageException(
+                    "Value for `--event-time-start` must be less than `--event-time-end`"
+                )
+            elif event_time_start >= datetime.now():
```
aranke, line 369: These are two separate conditions, so let's do two `if`s?
Also, do we need to specify UTC timezone for `datetime.now()`?

core/dbt/cli/flags.py
```diff
@@ -349,6 +353,24 @@ def _assert_mutually_exclusive(
             elif flag_set_by_user:
                 set_flag = flag
 
+    def _validate_event_time_configs(self) -> None:
+        event_time_start = (
+            getattr(self, "EVENT_TIME_START") if hasattr(self, "EVENT_TIME_START") else None
+        )
+        if event_time_start is not None:
```
aranke, line 360: Nit: do we need this level of nesting?

aranke (approved):
Couple nits, but otherwise LGTM!

core/dbt/cli/flags.py
QMalcolm: To the first part, they are not two separate `if`s. This was intentional as the commit message for ee994c101e610556a398db78b05a08da428632f1 states. We only need to do the second check if `event_time_end` is not set. Said another way, the relation of `event_time_start` to `datetime.now()` is irrelevant if `event_time_end` is set, and thus doesn't need to be checked. 

---

As for the second part, maybe? The issue is that the click parsing of datetimes creates naive datetimes (lacking timezone information). It won't even parse something like `--event-time-start "2024-10-07 13:45:00+00:00"`. It'll give you an error like:

```
Invalid value for '--event-time-start': '2024-10-07 00:00:00+00:00' does not match the formats '%Y-%m-%d', '%Y-%m-%dT%H:%M:%S', '%Y-%m-%d %H:%M:%S'.
```

So our options are to either assume what they give us is UTC, and give the passed in `--event-time-start` a timezone, or compare to a naive datetime. Given the rest of our system assumes UTC, you are right, we should probably do the former instead of what we are currently doing (the latter).

core/dbt/cli/flags.py
QMalcolm: I think so? We need to check whether or not `event_time_start` is `None`, as it only makes sense to do the validation if `event_time_start` is a datetime value. We could get around this layer of nesting by doing something like:
```python
if not hasattr(self, "EVENT_TIME_START"):
  return

event_time_start = getattr(self, "EVENT_TIME_START")
...
```

I'm personally not a huge fan of multiple returns. I can see how doing an early return would reduce our level of if's from 2 down to 1. However, given we're only at 2 levels of if's that seems unnecessary.

---
Side note that's not applicable here, but a corollary to my above "I avoid multiple returns".

If there were more levels of nesting, I still likely woudn't add early returns. I'd instead end up looking at splitting it into multiple functions, thus continuing allowing to maintain low levels of if context nesting and avoiding early multiple returns in a function.

PR #10826, opened 2024-10-04
Stop making microbatch batches with filters that will never have any rows

### LINKED ISSUE #10824: Better batching when using `--event-time-end`

QMalcolm:
Resolves #10824 

Our logic previously created batches for each batch period where batch_start <= event_end_time. This was problematic when a batch_start equaled the event_end_time because a batch would be produced with the filter like `WHERE event_time >= '2024-01-01 00:00:00' AND event_time < '2024-01-01 00:00:00'`. The two statements in that filter would logicially exclude each other meaning that 0 rows would be selected _always_. Thus we've changed the batch creation logic to be batch_start `<` event_end_time (as opposed to `<=`), which stops the bad batch filter from being a possibility.

### Checklist


QMalcolm:
One outstanding question is...

what should happen if you have a `lookback=2`, `batch_size=day`, an `--event-time-end "2024-05-26"`, and don't specify an `--event-time-start`? Would you expect:
1. 3 batches
  a. 2024-05-23 00:00:00 <= event_time < 2024-05-24 00:00:00
  b. 2024-05-24 00:00:00 <= event_time < 2024-05-25 00:00:00
  c. 2024-05-25 00:00:00 <= event_time < 2024-05-26 00:00:00
 
2. 2 batches
  a. 2024-05-24 00:00:00 <= event_time < 2024-05-25 00:00:00
  b. 2024-05-25 00:00:00 <= event_time < 2024-05-26 00:00:00

With the change made in [30c9aea](https://github.com/dbt-labs/dbt-core/pull/10826/commits/30c9aeabb1580e75dfb48789e0302fa8ad5a3a20) we are currently doing (2). If we want (1) to happen we'll need to make a change to the start time calculation [here](https://github.com/dbt-labs/dbt-core/blob/6b9c1da1aedda98a9ad798fc56e191c64629714d/core/dbt/materializations/incremental/microbatch.py#L68). Basically we'd do something like
```python
if MicrobatchBuilder.truncate_timestamp(checkpoint, batch_size) == checkpoint:
  lookback += 1

start = MicrobatchBuilder.offset_timestamp(checkpoint, batch_size, -1 * lookback)
```

QMalcolm:
> One outstanding question is...
> 
> what should happen if you have a `lookback=2`, `batch_size=day`, an `--event-time-end "2024-05-26"`, and don't specify an `--event-time-start`?

Thinking about it more, I think we should go with option (1) that I laid out. That being three batches:
1. 2024-05-23 00:00:00 <= event_time < 2024-05-24 00:00:00
2. 2024-05-24 00:00:00 <= event_time < 2024-05-25 00:00:00
3. 2024-05-25 00:00:00 <= event_time < 2024-05-26 00:00:00

The reason I think we should go that direction is because I asked myself what I would expect to happen if an automated job kicked off that model `2024-05-26 00:00:00`. In that scenario I would expect it to run the previous day `2024-05-25 00:00:00 <= event_time < 2024-05-26 00:00:00` and then. the prior two days as well. If that is what we'd expect to happen in that scenario, then it is reasonable to extend that expectation if one specifies that timestamp via `--event-time-end`.

MichelleArk (approved):

PR #10828, opened 2024-10-04
Track batch execution time for microbatch models

### LINKED ISSUE #10825: [Feature] timing for individual batches in logs

### Describe the feature

it looks like the logs for running individual batches always say in 0.00s - can we instead have that be how long it took for the batch to run

Screenshot 2024-10-03 at 5 34 11 PM

_No response_

### Who will this benefit?

_No response_

_No response_

### Anything else?

_No response_

QMalcolm:
Resolves #10825

### Problem

We weren't tracking the execution time of batches for microbatch models. Thus, we were always logging that they took no time to run (0.00 seconds) which didn't actually match reality.

### Solution

Begin tracking the time it takes for a batch to run, and propagate that information to the batch run result.

### Checklist

### DIFF
--- a/core/dbt/tests/util.py
+++ b/core/dbt/tests/util.py
@@ -5,7 +5,7 @@ from contextlib import contextmanager
 from contextvars import ContextVar, copy_context
 from datetime import datetime
 from io import StringIO
-from typing import Any, Dict, List, Optional
+from typing import Any, Callable, Dict, List, Optional
 from unittest import mock
 
 import pytz
@@ -17,7 +17,7 @@ from dbt.cli.main import dbtRunner
 from dbt.contracts.graph.manifest import Manifest
 from dbt.materializations.incremental.microbatch import MicrobatchBuilder
 from dbt_common.context import _INVOCATION_CONTEXT_VAR, InvocationContext
-from dbt_common.events.base_types import EventLevel
+from dbt_common.events.base_types import EventLevel, EventMsg
 from dbt_common.events.functions import (
     capture_stdout_logs,
     fire_event,
@@ -76,6 +76,7 @@ from dbt_common.events.types import Note
 def run_dbt(
     args: Optional[List[str]] = None,
     expect_pass: bool = True,
+    callbacks: Optional[List[Callable[[EventMsg], None]]] = None,
 ):
     # reset global vars
     reset_metadata_vars()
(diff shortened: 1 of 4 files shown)


core/dbt/tests/util.py
```diff
@@ -93,7 +94,7 @@ def run_dbt(
         args.extend(["--project-dir", project_dir])
     if profiles_dir and "--profiles-dir" not in args:
         args.extend(["--profiles-dir", profiles_dir])
-    dbt = dbtRunner()
+    dbt = dbtRunner(callbacks=callbacks)
```
aranke, line 97: It looks like this pattern is used in many places, let's refactor them all at once?

For now, this should be fine?

```
@pytest.fixture(scope="function")
def runner(catcher: EventCatcher) -> dbtRunner:
    return dbtRunner(callbacks=[catcher.catch])
```

tests/functional/microbatch/test_microbatch.py
```diff
@@ -186,10 +188,16 @@ class TestMicrobatchCLI(BaseMicrobatchTest):
...
     def test_run_with_event_time(self, project):
         # run without --event-time-start or --event-time-end - 3 expected rows in output
+        catcher = EventCatcher(event_to_catch=LogModelResult)
+
         with patch_microbatch_end_time("2020-01-03 13:57:00"):
-            run_dbt(["run"])
+            run_dbt(["run"], callbacks=[catcher.catch])
         self.assert_row_count(project, "microbatch_model", 3)
 
+        for caught_event in catcher.caught_events:
```
aranke, line 197: Maybe assert the number of events here?

aranke (approved):
LGTM, couple nits

tests/functional/microbatch/test_microbatch.py
QMalcolm: Can do!

core/dbt/tests/util.py
QMalcolm: That's a good idea. However, we wouldn't be able to use that in this util. I think that fixture would likely make most sense `tests/` not `core/dbt/tests/`. This utility function (`run_dbt`) wouldn't have access to it. That doesn't mean we shouldn't do that work, but it's OOO for this PR and should be its own segment of work.

PR #10830, opened 2024-10-07
Ensure consistent `current_time` across microbatch models in an invocation

### LINKED ISSUE #10819: Set a singular "current_time" for microbatch models per invocation
Currently, microbatch models get a "current_time" (`datetime.datetime.now(pytz.UTC)`) when they are executed. Notably, each microbatch model gets a _different_ "current_time". This works, but has some funkiness.

Consider the following:
1. There is a source, `source_1`, which is constantly being updated by an external process
2. Microbatch model `model_a` pulls from `source_1`
3. Microbatch model `model_b` pulls from `source_1`

Regardless if we're in a small or large project, given _any_ delay in the execution of `model_a` and `model_b`, if there is any new data in `source_1`, the result of `model_a` will be _different_ from `model_b`. An example would be:

1. `source_1` has 3 rows with event times: `2024-10-07 12:17:00`, `2024-10-07 12:16:00`, `2024-10-07 12:15:00`
2. the "current_time" is `2024-10-07 12:18:00`
3. `model_a` is executed, picking up the 3 rows
4. new data arrives to `source_1`: `2024-10-07 12:18:30`
5. the "current_time" is `2024-10-07 12:19:00`
6. `model_b` is executed, picking up 4 rows.

Is the discrepancy okay?

QMalcolm:
Resolves #10819

### Problem

Different microbatch models in the same `dbt run` could end up with different `current_time`. This could cause a situation where two microbatch models operating on the same inputs could end up with more/less data if one microbatch model was executed significantly later than the other.

### Solution

Make it so each microbatch model in an invocation is using the same `current_time`

### Checklist

### DIFF
--- a/core/dbt/config/runtime.py
+++ b/core/dbt/config/runtime.py
@@ -101,7 +101,7 @@ class RuntimeConfig(Project, Profile, AdapterRequiredConfig):
     profile_name: str
     cli_vars: Dict[str, Any]
     dependencies: Optional[Mapping[str, "RuntimeConfig"]] = None
-    invocated_at: datetime = field(default_factory=lambda: datetime.now(pytz.UTC))
+    invoked_at: datetime = field(default_factory=lambda: datetime.now(pytz.UTC))
 
     def __post_init__(self):
         self.validate()
--- a/core/dbt/config/runtime.py
+++ b/core/dbt/config/runtime.py
@@ -1,7 +1,8 @@
 import itertools
 import os
 from copy import deepcopy
-from dataclasses import dataclass
+from dataclasses import dataclass, field
+from datetime import datetime
 from pathlib import Path
 from typing import (
     Any,
@@ -14,6 +15,8 @@ from typing import (
     Tuple,
     Type,
 )
+
+import pytz
 
 from dbt import tracking
 from dbt.adapters.contracts.connection import (
@@ -98,6 +101,7 @@ class RuntimeConfig(Project, Profile, AdapterRequiredConfig):
     profile_name: str
     cli_vars: Dict[str, Any]
     dependencies: Optional[Mapping[str, "RuntimeConfig"]] = None
+    invocated_at: datetime = field(default_factory=lambda: datetime.now(pytz.UTC))
 
     def __post_init__(self):
         self.validate()
--- a/core/dbt/materializations/incremental/microbatch.py
+++ b/core/dbt/materializations/incremental/microbatch.py
@@ -36,9 +36,7 @@ class MicrobatchBuilder:
             event_time_start.replace(tzinfo=pytz.UTC) if event_time_start else None
         )
         self.event_time_end = event_time_end.replace(tzinfo=pytz.UTC) if event_time_end else None
-        self.batch_current_time = (
-            batch_current_time if batch_current_time is not None else datetime.now(pytz.UTC)
-        )
+        self.batch_current_time = batch_current_time or datetime.now(pytz.UTC)
 
     def build_end_time(self):
         """Defaults the end_time to the current time in UTC unless a non `None` event_time_end was provided"""
--- a/core/dbt/materializations/incremental/microbatch.py
+++ b/core/dbt/materializations/incremental/microbatch.py
@@ -18,6 +18,7 @@ class MicrobatchBuilder:
         is_incremental: bool,
         event_time_start: Optional[datetime],
         event_time_end: Optional[datetime],
+        batch_current_time: Optional[datetime] = None,
     ):
         if model.config.incremental_strategy != "microbatch":
             raise DbtInternalError(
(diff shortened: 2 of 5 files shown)


core/dbt/materializations/incremental/microbatch.py
```diff
@@ -35,10 +36,13 @@ def __init__(
             event_time_start.replace(tzinfo=pytz.UTC) if event_time_start else None
         )
         self.event_time_end = event_time_end.replace(tzinfo=pytz.UTC) if event_time_end else None
+        self.batch_current_time = (
+            batch_current_time if batch_current_time is not None else datetime.now(pytz.UTC)
+        )
```
aranke, line 41: Nit: can be simplified to 

```suggestion
        self.batch_current_time = (
            batch_current_time or datetime.now(pytz.UTC)
        )
```

core/dbt/config/runtime.py
```diff
@@ -98,6 +101,7 @@ class RuntimeConfig(Project, Profile, AdapterRequiredConfig):
     profile_name: str
     cli_vars: Dict[str, Any]
     dependencies: Optional[Mapping[str, "RuntimeConfig"]] = None
+    invocated_at: datetime = field(default_factory=lambda: datetime.now(pytz.UTC))
```
aranke, line 104: Can we simplify this to `invoked_at`?

core/dbt/materializations/incremental/microbatch.py
```diff
@@ -18,6 +18,7 @@ def __init__(
         is_incremental: bool,
         event_time_start: Optional[datetime],
         event_time_end: Optional[datetime],
+        batch_current_time: Optional[datetime] = None,
```
aranke, line 21: Can this be `batch_invoked_at` for consistency?

core/dbt/materializations/incremental/microbatch.py
```diff
@@ -18,6 +18,7 @@ def __init__(
...
     ):
         if model.config.incremental_strategy != "microbatch":
```
aranke, line 23: Are we in the right branch for `batch_current_time`?

core/dbt/materializations/incremental/microbatch.py
```diff
@@ -35,10 +36,13 @@ def __init__(
...
         )
         self.event_time_end = event_time_end.replace(tzinfo=pytz.UTC) if event_time_end else None
+        self.batch_current_time = (
+            batch_current_time if batch_current_time is not None else datetime.now(pytz.UTC)
+        )
 
     def build_end_time(self):
         """Defaults the end_time to the current time in UTC unless a non `None` event_time_end was provided"""
-        return self.event_time_end or datetime.now(tz=pytz.utc)
+        return self.event_time_end or self.batch_current_time
```
aranke, line 45: Under which situations will we return `self.batch_current_time` here?
Probably worth adding to the description.

QMalcolm: From the function perspective, we return `self.batch_current_time` when `self.event_time_end` is `None`. From a system perspective this currently happens when the dbt runner is not invoked with an `--event-time-end <my-desired-end-value>`. I don't think the system context is relevant to this function though. If anything I think we should separately push that context to the instantiation of the class in `run.py` and drop this `or`-ing from the class itself.

I do have a planned refactor of this class for this week, this context shifting would be more appropriate as part of that.

core/dbt/materializations/incremental/microbatch.py
QMalcolm: I'm not sure I understand, can you elaborate on "in the right branch"?

core/dbt/materializations/incremental/microbatch.py
QMalcolm: I'm down to rename the variable. However, I think `batch_invoked_at` would give the wrong impression. This datetime is not when the batch was actually invoked. The datetime will be before the batch is actually invoked, by how much is dependent on how long the project takes to run and when the microbatch model is being worked on within that.

core/dbt/config/runtime.py
QMalcolm: We can 🙂 I went back and for on this rename myself. Will get it changed.

core/dbt/materializations/incremental/microbatch.py
QMalcolm: How about `batch_default_end_time`?

QMalcolm: or `default_end_time`? 🤔 Saying `batch` feels kinda redundant

core/dbt/materializations/incremental/microbatch.py
QMalcolm: If this is in reference to `batch` vs `microbatch`. For better or worse, we've been kinda using them interchangeably 🙈

QMalcolm: ### TL;DR
`batch_current_time` (now renamed to `default_end_time`) is _only_ relevant to microbatch models

---

This file (`microbatch.py`) and the class itself (`MicrobatchBuilder`) should only ever be invoked when running on a microbatch model. That logic for separating whether we in a "default" (a.k.a. non-microbatch) model or a microbatch model lives [in run.py in the execute() method](https://github.com/dbt-labs/dbt-core/blob/fc8eb820aa97aca6da8bafe603f4007167686eb6/core/dbt/task/run.py#L459). Thus, if we're in this code (microbatch.py), we're inherently in a microbatch context / logic branch. Hope that all makes sense. I sometimes forget that the rest of the team doesn't automatically have the context of how microbatch models work 😅 Let me know if you want a walk through the microbatch code at large if that'd be useful 🙂

aranke (approved):
LGTM!

PR #10849, opened 2024-10-14
fix: dbt unit tests feat without proper context

### LINKED ISSUE #10353: [Bug] Unit test compilation error 'var is undefined' when combining a variable with math operators

### Current Behavior

When running a unit test for a model that contains a Jinja math operation on a variable, a compilation error occurs 

Example operation
```
select {{ var('foo', 10 ) * 1}} as result
```
Error:
```
Compilation Error in unit_test test_unit_test (models\staging\test.yml)
  'var' is undefined. This can happen when calling a macro that does not exist. Check for typos and/or install package dependencies with "dbt deps".
```
Running the unit test on a Jinja variable without the math operation works successfully. Ex: `select {{ var('foo', 10 ) }} as result`

### Expected Behavior

Expected behavior would be to compile unit test and executing the math operation on the variable. 

### Steps To Reproduce

1. In any dbt project, create a model called `test` with the following SQL:
```
select {{ var('foo', 10 ) * 1}} as result
```
2. In that model's .yml file, create a unit test as follows (the 'input' is not used, but required):
```
version: 2

unit_tests:
  - name: test_unit_test
    model: test
    given:
      - input: ref('stg_ga4__events')
        format: csv
        rows: |
          event_key
          A
    expect:
      format: csv
      rows: |
        result
        10
```
3. Run the unit test `dbt test -s test_unit_test`

### Relevant log output

```shell
(python312env) C:\GitHub\velir-ga4-test-project>dbt test -s test_unit_test
22:04:52  Running with dbt=1.8.2
22:04:53  Registered adapter: bigquery=1.8.1
22:04:54  Found 59 models, 28 data tests, 1 seed, 1 source, 604 macros, 11 unit tests
22:04:54  
22:04:55  Concurrency: 4 threads (target='dev')
22:04:55  
22:04:55  1 of 1 START unit_test test::test_unit_test .................................... [RUN]
22:04:55  1 of 1 ERROR test::test_unit_test .............................................. [ERROR in 0.01s]
22:04:55  
22:04:55  Finished running 1 unit test in 0 hours 0 minutes and 1.22 seconds (1.22s).
22:04:55  
22:04:55  Completed with 1 error and 0 warnings:
22:04:55
22:04:55    Compilation Error in unit_test test_unit_test (models\staging\test.yml)
  'var' is undefined. This can happen when calling a macro that does not exist. Check for typos and/or install package dependencies with "dbt deps".
22:04:55
22:04:55  Done. PASS=0 WARN=0 ERROR=1 SKIP=0 TOTAL=1
```

### Environment

```markdown
- OS: Windows
- Python: 3.12
- dbt-core: 1.8.3, bigquery: 1.8.3
```

bigquery

### Additional Context

_No response_

### LINKED ISSUE #10410: [Bug] Unit Test fails when macro is used with jinja filter

### Current Behavior

When calling calling a macro passed to a jinja filter, the unit tests will fail, while the model can be compiled and run correctly.
Removing the jinja filter fixes the issue.
Creating a new macro to apply the jinja filter also works.

### Expected Behavior

Using a macro with a jinja filter would not cause the unit test to fail.

### Steps To Reproduce

my_macro.sql
```
{%- macro my_macro() -%}
    column_a
{%- endmacro -%}
```

my_model.sql
```
select {{ my_macro() | indent(4) }}
```

unit_test.sql
```
A unit test...
```

- Compiling the model will succeed.
- Running the model will succeed.
- Running the unit test will fail.
- Building the model will fail.

Error message:
```
Compilation Error in unit_test my_unit_test (<path>)
  'my_macro' is undefined. This can happen when calling a macro that does not exist. Check for typos and/or install package dependencies with "dbt deps".
```

### Relevant log output

_No response_

### Environment

_No response_

_No response_

### Additional Context

Work around:

Wrap the jinja filter in another macro.

indent.sql
```
{%- macro indent(text, width) -%}
    {{ text|indent(width=width)}}
{%- endmacro -%}
```

Then update my_model.sql
```
select {{ indent(my_macro(), 4) }}
```

### LINKED ISSUE #10539: [Bug] Unit test compilation error 'var is undefined' when macro is used with project variables as input parameters

### Current Behavior

When calling calling a macro that takes project variables as input, the configured unit tests fail with `var is undefined` error. 
Overriding the macro or project variables doesn't help but removing the variables resolves the issue. No issue with running the model without unit test.

### Expected Behavior

Expecting unit test to work with macro taking project variables as input, or at least end user can override the macro or the project variables used as inputs for the macro.

### Steps To Reproduce

get_columns.sql

```sql
{%- macro get_columns(include=[]) -%}
    columns
{%- endmacro -%}
```

my_model.sql

```sql
{{
  config(
  materialized='table'
)
}}
select 
  {{ get_columns(
    include=var('columns')
  )}}
from model
```

dbt_project.yml
```yml
columns:
  - column_a
  - column_b
  - column_c
```

unit_test__my_model.yml
Tried without override
```yml
```
Override macro:
```yml
    overrides:
      vars:
        get_columns: override
```
Override input:
```yml
    overrides:
      vars:
        columns: ['column_a','column_b','column_c']
```

### Relevant log output

```shell
'var' is undefined. This can happen when calling a macro that does not exist. Check for typos and/or install package dependencies with "dbt deps".
```

### Environment

```markdown
- OS: Apple M1 Max
- Python: 3.11
- dbt: 1.8.2
```

snowflake

### Additional Context

_No response_

devmessias:
Resolves #10539, resolves #10353 , resolves #10410

I made a very simple change to the code just to  solve a similar problem 10353, but this change is obviously not the correct solution. I’ll wait for a discussion before writing any more advanced code.

### Problem

absence of `env_var` and `var` in the `ctx` dictionary, which must be passed to the Jinja render method. 

### Solution

If a model has a unit test, the `get_rendered` method is called twice, each time with a different context. The issue arises with `build_unit_test_manifest_from_test`. When this method is called (https://github.com/dbt-labs/dbt-core/blob/f6088db25b4a386db49dfda5de613d1e70bb2445/core/dbt/parser/unit_tests.py#L116), the context doesn’t contain information about the `var` and environment variables (`env_var`).

Later, when the `get_rendered` method from `clients/jinja.py` calls `render_template` from `dbt_common/clients/jinja.py`, we encounter a compilation error due to the absence of `env_var` and `var` in the `ctx` dictionary, which must be passed to the Jinja render method. **The error will not appear if we don't have any kind of mixing with Nones, for example `select {{ var('something,1) }} as result` instead of `select {{ var('something,1)*2 }} as result`

Some questions:

Is this the expected behavior? Shouldn't the `ctx` used to render the string for the unit test include these variables?

Is there a reason why the same model is rendered more than once — once to build the model and again to render the unit test? Why is this string rendered a second time if it doesn’t seem to be used for any data manipulation?

models/_unit_tests.yml
```
version: 2

unit_tests:
  - name: test_unit_test
    model: test
    given: []
    expect:
      format: csv
      rows: |
        result
        10
```

models/test.sql
```
select {{ var('foo', 10) * 1 }} as result
```

This will only work for #10539 if ctx is constructed using `env_var`  and `var`,  but before coding any stuff I'll wait for the dbt-core team

### Checklist

### DIFF
--- a/core/dbt/context/providers.py
+++ b/core/dbt/context/providers.py
@@ -257,7 +257,9 @@ class BaseResolver(metaclass=abc.ABCMeta):
         return event_time_filter
 
     @abc.abstractmethod
-    def __call__(self, *args: str) -> Union[str, RelationProxy, MetricReference]:
+    def __call__(
+        self, *args: str
+    ) -> Optional[Union[str, int, float, RelationProxy, MetricReference]]:
         pass
 
 
@@ -1788,6 +1790,61 @@ class ExposureRefResolver(BaseResolver):
         return ""
 
 
+class ExposureVarResolver(BaseResolver):
+    def __call__(self, *args) -> Any:  # ?? -> Any??
+        n_args = len(args)
+        if n_args == 0 or n_args > 2:
+            raise RefArgsError(node=self.model, args=args)
+        var = args[0]
+        default = None if n_args == 1 else args[1]
+        # context is None using the same decision at
+        # generate_runtime_unit_test_context
+        return UnitTestVar(context={}, config=self.config, node=self.model)(var, default)
+
+
+# Should  we create a new copy of the `envvar` method, similar to what was done for
+# `TextContext` and `ProviderContext`?
+class ExposureEnvVarResolver(BaseResolver):
+    def __call__(self, *args) -> Optional[Union[int, float, str]]:
+        n_args = len(args)
+        if n_args == 0 or n_args > 2:
+            raise RefArgsError(node=self.model, args=args)
+        var = args[0]
+        default = None if n_args == 1 else args[1]
+
+        return_value = None
+        if var.startswith(SECRET_ENV_PREFIX):
+            raise SecretEnvVarLocationError(var)
+
+        env = get_invocation_context().env
+        if var in env:
+            return_value = env[var]
+        elif default is not None:
+            return_value = default
+
+        if return_value is not None:
+            # Save the env_var value in the manifest and the var name in the source_file
+            if self.model:
+                # If the environment variable is set from a default, store a string indicating
+                # that so we can skip partial parsing.  Otherwise the file will be scheduled for
+                # reparsing. If the default changes, the file will have been updated and therefore
+                # will be scheduled for reparsing anyways.
+                self.manifest.env_vars[var] = (
+                    return_value if var in env else DEFAULT_ENV_PLACEHOLDER
+                )
+                # the "model" should only be test nodes, but just in case, check
+                # TODO CT-211
+                if self.model.resource_type == NodeType.Test and self.model.file_key_name:  # type: ignore[union-attr] # noqa
+                    source_file = self.manifest.files[self.model.file_id]
+                    # TODO CT-211
+                    (yaml_key, name) = self.model.file_key_name.split(".")  # type: ignore[union-attr] # noqa
+                    # TODO CT-211
+                    source_file.add_env_var(var, yaml_key, name)  # type: ignore[union-attr]
+            return return_value
+        else:
+            raise EnvVarMissingError(var)
+
+
 class ExposureSourceResolver(BaseResolver):
     def __call__(self, *args) -> str:
         if len(args) != 2:
@@ -1830,6 +1887,8 @@ def generate_parse_exposure(
             project,
             manifest,
         ),
+        "var": ExposureVarResolver(None, exposure, project, manifest),
+        "env_var": ExposureEnvVarResolver(None, exposure, project, manifest),
     }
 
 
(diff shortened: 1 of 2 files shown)


devmessias:
Sorry to bother you @gshank, I looked into the git history in those methods, and it seems you're probably the best person to ask these questions.

devmessias:
Hi @MichelleArk , sorry for to bother you, but since  you're keeping an eye on the issues related to unit tests, could you help me with some questions I have in this PR?

urkle:
What is needed to wrap up this PR?

I tested locally and this fixes the issue with running unit tests against models referencing vars.

I'd be more than willing to recreate this PR and clean it up against current mainline. As this is currently stopping us from utilizing unit_tests on several models

PR #10857, opened 2024-10-15
Behavior change for mf timespine without yaml configuration

DevonFulcher:
Linear issue #SL-2935
Github issue #10959
docs: https://github.com/dbt-labs/docs.getdbt.com/issues/6387

### Problem

The configuration for the MetricFlow timespine has changed in a previous release. This deprecation warning was previously surfaced in DSI, but that is non-standard for dbt-core usage.

### Solution

This PR implements a behavior change to inform users of the deprecation of the old time spine config.

Tests:
- I ran `dbt parse` on `jaffle-sl-template` and saw no warnings
- I updated `jaffle-sl-template` to have a `metricflow_time_spine.sql` model and saw the deprecation warning.
- I changed the time spine configuration to have a granularity other than day and I saw no warnings.
- I added this to `dbt_project.yaml` in `jaffle-sl-template` and saw no warning for `dbt parse`:
```
flags:
  require_yaml_configuration_for_mf_time_spines: True
```

### Checklist


core/dbt/contracts/graph/semantic_manifest.py
```diff
@@ -58,6 +60,19 @@ def validate(self) -> bool:
         semantic_manifest = self._get_pydantic_semantic_manifest()
         validator = SemanticManifestValidator[PydanticSemanticManifest]()
         validation_results = validator.validate_semantic_manifest(semantic_manifest)
+        new_time_spines = semantic_manifest.project_configuration.time_spines
+        old_time_spines = semantic_manifest.project_configuration.time_spine_table_configurations
+        new_time_spines_contain_day = any(
+            c for c in new_time_spines if c.primary_column.time_granularity == TimeGranularity.DAY
+        )
+        if (
```
courtneyholcomb, line 68: Not related to this PR, but seeing this makes me realized that some of our older parsing errors are likely in the wrong place and should be moved here! i.e., all the parsing errors raised in `_get_pydantic_semantic_manifest()`, which disrupt the parsing flow.  Maybe we can have a task to fix that after this one is finished.

courtneyholcomb: Actually, ignore this. Just looked into it more and I think we are actually following the same logic for how most core objects are validated

core/dbt/contracts/graph/semantic_manifest.py
```diff
@@ -58,6 +60,20 @@ def validate(self) -> bool:
         semantic_manifest = self._get_pydantic_semantic_manifest()
         validator = SemanticManifestValidator[PydanticSemanticManifest]()
         validation_results = validator.validate_semantic_manifest(semantic_manifest)
+        new_time_spines = semantic_manifest.project_configuration.time_spines
```
MichelleArk, line 63: can we just name this as `time_spines`, and rename `old_time_spines` to `legacy_time_spines`? 'new' will have little semantic meaning outside the context of this PR. we also do a bunch of state comparison operations throughout the codebase so a reader may mistake it for something like that at first.

core/dbt/contracts/project.py
```diff
@@ -344,6 +344,7 @@ class ProjectFlags(ExtensibleDbtClassMixin):
     skip_nodes_if_on_run_start_fails: bool = False
     state_modified_compare_more_unrendered_values: bool = False
     state_modified_compare_vars: bool = False
+    allow_mf_time_spines_without_yaml_configuration: bool = False
```
MichelleArk, line 347: What do you think about the naming, `require_yaml_configuration_for_mf_time_spines: False` instead?

It would be a more conventional nam and better follow our guides: https://github.com/dbt-labs/dbt-core/blob/main/docs/guides/behavior-change-flags.md#rules-for-introducing-a-new-flag

core/dbt/events/README.md
```diff
@@ -1,5 +1,5 @@
 # Events Module
-The Events module is responsible for communicating internal dbt structures into a consumable interface. Because the "event" classes are based entirely on protobuf definitions, the interface is really clearly defined, whether or not protobufs are used to consume it. We use Betterproto for compiling the protobuf message definitions into Python classes.
+The Events module is responsible for communicating internal dbt structures into a consumable interface. Because the "event" classes are based entirely on protobuf definitions, the interface is really clearly defined, whether or not protobufs are used to consume it. We use protoc for compiling the protobuf message definitions into Python classes.
```
MichelleArk, line 2: thank you for these updates <3

core/dbt/contracts/graph/semantic_manifest.py
```diff
@@ -58,6 +60,20 @@ def validate(self) -> bool:
...
+            c for c in new_time_spines if c.primary_column.time_granularity == TimeGranularity.DAY
+        )
+        if (
+            get_flags().allow_mf_time_spines_without_yaml_configuration is False
+            and old_time_spines
+            and not new_time_spines_contain_day
+        ):
+            deprecations.warn(
+                "mf-timespine-without-yaml-configuration",
+            )
```
MichelleArk, line 76: Presumably this warning would not be terribly noisy -- just one per invocation?

MichelleArk:
Could you please create/link a corresponding [issue in docs.getdbt.com](https://github.com/dbt-labs/docs.getdbt.com/issues?q=is%3Aissue+is%3Aopen) so the docs team can work on documenting the new behaviour once this is merged?

core/dbt/contracts/graph/semantic_manifest.py
DevonFulcher: Yes, this is validation for the semantic manifest which there is only one of per project. I can't think of a way that this would fire more than once per invocation.

MichelleArk (approved):

PR #10859, opened 2024-10-16
Add adapter telemetry to snowplow event.

VersusFacit:
Resolves https://github.com/dbt-labs/dbt-adapters/issues/301

### Solution

We propose exposing a method on the base adapter that will use existing snowplow infrastructure for tracking user statistics for any adapter. 

We designed the json schema and event data to make it easy for us to add additional per-adapter information as use-cases emerge.

### Checklist


core/dbt/task/run.py
```diff
@@ -133,6 +133,7 @@ def track_model_run(index, num_nodes, run_model_result):
             "contract_enforced": contract_enforced,
             "access": access,
             "versioned": versioned,
+            "adapter": BaseAdapter.get_adapter_run_info(run_model_result.node.config),
```
VersusFacit, line 136: I need to alter this method invocation slightly but the final idea is captured here!

core/dbt/task/run.py
```diff
@@ -98,7 +99,7 @@ def get_execution_status(sql: str, adapter: BaseAdapter) -> Tuple[RunStatus, str
         return status, message
 
 
-def track_model_run(index, num_nodes, run_model_result):
+def track_model_run(index, num_nodes, run_model_result, adapter=BaseAdapter):
```
VersusFacit, line 102: This is a defensive move in case somehow any other company package relies on this method for telemetry (doubt it, but I'm being safely paranoid ;) )!

core/dbt/task/run.py
```diff
@@ -133,6 +139,7 @@ def track_model_run(index, num_nodes, run_model_result):
             "contract_enforced": contract_enforced,
             "access": access,
             "versioned": versioned,
+            "adapter_info": asdict(adapter_info),
```
VersusFacit, line 142: I considered that this should be something JSON safe, which a Python dict is 🫡

core/dbt/task/run.py
colin-rogers-dbt: Fair point that it's possible to not have an adapter, in that case should this be optional? We could handle it as: if `adapter=None` populate `adapter_info` with an empty dict `{}`?

VersusFacit: I don't think it's possible to not have the base adapter right? So you'd get the in-scope base adapter version which right now would include (1) name (2) version (3) an empty dict for adapter-specific details

Does that make sense or would you prefer None? I actually thought about None, but that felt like an illegal state to reach this code in

VersusFacit: I can agree with returning a default result of `{}` in the no adapter scenario. That's sufficiently defensive to quell my anxiety.

peterallenwebb (approved):

PR #10876, opened 2024-10-17
Change microbatch `lookback` default from `0` to `1`

### LINKED ISSUE #10867: [Bug] Defaulting lookback to `0` results in consistently incomplete batches
### We should switch the default for `lookback` to `1`.

Currently the lookback value defaults to `0`. The problem with this is that a lookback of `0` means that the dataset will never have a “complete” batch. This is because when microbatch is run, the current time is used for the latest batch. We do that because we’re favoring freshness over ensuring “only complete batches”. Unfortunately, this combined with `lookback=0` makes it such that no batch is ever "complete".

For an example, consider a microbatch model with a `batch_size` of `day`, and it's run at noon everyday (12:00:00). If our `lookback` is `0` then when the microbatch model is run today it’ll get data from today 00:00:00 to 12:00:00. Then tomorrow when my microbatch model it’ll get data for tomorrow 00:00:00 to 12:00:00, but it won’t go back and get the rest of “today’s” data (because the lookback is `0`).

Thus the “default” valid behavior should be a `lookback` of `1`. This ensures that batches are complete. The only caveats being when there is regularly late arriving data for which one can set a larger `lookback` value, or when there is one off late arriving data  using `—event-time-start` + `—event-time-end` to backfill the specific range.

QMalcolm:
Resolves #10867 

### Problem

A `lookback` value of `0` results in microbatch models consistently having incomplete data.

### Solution

Change the `lookback` default to `1`. This guarantees that batches will eventually get completed (assuming the job is run once per `batch_size`)

### Checklist


QMalcolm:
Changes to schemas.getdbt.com https://github.com/dbt-labs/schemas.getdbt.com/pull/68

gshank (approved):

mirnawong1:
hey @QMalcolm , when this pr is merged, am i right in changing the docs here from `0` to`1`? https://docs.getdbt.com/docs/build/incremental-microbatch#relevant-configs

QMalcolm:
> hey @QMalcolm , when this pr is merged, am i right in changing the docs here from `0` to`1`? https://docs.getdbt.com/docs/build/incremental-microbatch#relevant-configs

That is correct 🙂

mirnawong1:
docs pr here to address this: https://github.com/dbt-labs/docs.getdbt.com/pull/6351/files

QMalcolm:
Something weird happened. Some commits from another one of my branches got onto this one. I assume in my delirium of spinning plates, I did something dumb.

PR #10879, opened 2024-10-17
Ensure `KeyboardInterrupt` halts microbatch model execution

### LINKED ISSUE #10862: [Bug] canceling a run of a microbatch model should cause remaining batches to be skipped

### Current Behavior

Cancelling a run doesn’t immediately stop all the related batch runs. The remaining batches continue to build.

### Expected Behavior

All the batches cancel out regardless of which batch we initiate the cancellation action/request.

### Steps To Reproduce

See [here](https://www.notion.so/dbtlabs/Testing-out-the-Microbatch-Incrementals-121bb38ebda78066bdeff09870d39d31?pvs=4#121bb38ebda780f4991ed93455bad1c9)

### Relevant log output

_No response_

### Environment

```markdown
- OS:
- Python:
- dbt:
```

snowflake

### Additional Context

_No response_

QMalcolm:
Resolves #10862 

### Problem

A `KeyboardInterrupt` during the execution of microbatches would only cause the currently executing batch to stop, and then the rest of the batches (and job) would continue.

### Solution

Make `KeyboardInterrupt`s halt batch execution.

### Checklist

### DIFF
--- a/core/dbt/task/run.py
+++ b/core/dbt/task/run.py
@@ -568,6 +568,9 @@ class ModelRunner(CompileRunner):
                 # At least one batch has been inserted successfully!
                 incremental_batch = True
 
+            except (KeyboardInterrupt, SystemExit):
+                # reraise it for GraphRunnableTask.execute_nodes to handle
+                raise
             except Exception as e:
                 exception = e
                 batch_run_result = self._build_failed_run_batch_result(
--- a/.changes/unreleased/Fixes-20241017-153022.yaml
+++ b/.changes/unreleased/Fixes-20241017-153022.yaml
@@ -1,0 +1,6 @@
+kind: Fixes
+body: Ensure KeyboardInterrupt/SystemExit halts microbatch model execution
+time: 2024-10-17T15:30:22.781854-07:00
+custom:
+  Author: QMalcolm
+  Issue: "10862"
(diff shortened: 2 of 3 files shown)


QMalcolm:
Need to figure out how to test this. I think we can mock a function called during `ModelRunner._execute_microbatch_materialization` to raise a `KeyboardInterrupt`, which we could then check causes execution to halt.

core/dbt/task/run.py
```diff
@@ -568,6 +568,9 @@ def _execute_microbatch_materialization(
                 # At least one batch has been inserted successfully!
                 incremental_batch = True
 
+            except (KeyboardInterrupt, SystemExit):
+                # reraise it for GraphRunnableTask.execute_nodes to handle
```
ChenyuLInx, line 572: Should we be writing out run results in this case?

ChenyuLInx (commented):
I have one question but otherwise LGTM!

core/dbt/task/run.py
QMalcolm: The KeyboardInterrupt handler in GraphRunnableTask.execute_nodes handles that

ChenyuLInx (approved):
Thanks for clarify! LGTM

moseleyi:
Has this been working for anybody? I can't cancel microbatch via CTRL+C in VSCode terminal at all

PR #10699, opened 2024-09-11
wip: refactor safe_run_hooks and add more metadata in

aranke:
Resolves #

### Checklist

- [ ] I have read [the contributing guide](https://github.com/dbt-labs/dbt-core/blob/main/CONTRIBUTING.md) and understand what's expected of me.
- [ ] I have run this code in development, and it appears to resolve the stated issue.
- [ ] This PR includes tests, or tests are not required or relevant for this PR.
- [ ] This PR has no interface changes (e.g., macros, CLI, logs, JSON artifacts, config files, adapter interface, etc.) or this PR has already received feedback and approval from Product or DX.
- [ ] This PR includes [type annotations](https://docs.python.org/3/library/typing.html) for new and modified functions.

PR #10885, opened 2024-10-18
exclude hook results from results in on-run-end context

ChenyuLInx:
This brings the results in on-run-end back to before https://github.com/dbt-labs/dbt-core/pull/10699/files was introduced.
We can introduce this behind a behavior flag as needed later on.

### Checklist


ChenyuLInx:
@graciegoheen We should backport this to 1.9 since this and the previous change to hook behavior should be shipped together.

peterallenwebb:
Looks good.

peterallenwebb (approved):

PR #10889, opened 2024-10-21
fix: unit tests with versioned refs

### LINKED ISSUE #10528: [Bug] Unit Test fails when using a versioned model as an input

### Current Behavior

Unit tests fail if an input reference is a versioned model and the version is not explicitly defined in both the model AND the unit test. 

### Expected Behavior

The unit test should default to the latest version specified in the yml config for the input versioned model.

### Steps To Reproduce

1. In the dbt Cloud IDE, create a versioned mode (e.g. `stg_tpch_orders`, with versions `1` and `2`)  and second model (e.g. `customer_tier`) that references that versioned model, and a unit test on that second model (`customer_tier`) that references that versioned model as an input
2. Run a `dbt build --select customer_tier`

- If the `customer_tier` model **does not** reference a specific version, the unit test will always fail, regardless if a version is specified or not. For this scenario, see code below.
- If the `customer_tier` model **does** reference a specific version, the unit test will fail if no version is specified. The unit test does work if you explicitly specify `ref('stg_tpch_orders', v =2)`  in both places

Code for reference

schema.yml for stg_tpch_orders
```
- name: stg_tpch_orders
    description: staging layer for orders data
    versions:
      - v: 1
        columns:
          - include: all
            exclude: [comment]
        deprecation_date: 2024-8-31 00:00:00.00+00:00
      - v: 2
        columns:
          - include: all
    latest_version: 2
```

customer_tier.sql 
```
{{
    config(
        materialized='table'
    )
}}

with customer as (
    select * from {{ ref('stg_tpch_customers') }}
),

orders as (
    select * from {{ ref('stg_tpch_orders') }}
),
final as (
    select
        customer.customer_key,
        sum(orders.total_price) as lifetime_value,
        case 
            when lifetime_value <= 200000 then 'tier1'
            when lifetime_value > 2000000 then 'tier2'
            when lifetime_value between 1000000 and 1999999 then 'tier3'
            when lifetime_value between 0 and 999999 then 'tier4' 
        end as tier_name, 
        max(orders.comment) as comment
    from customer
        inner join orders
            on customer.customer_key = orders.customer_key
    group by 1
)

select * from final
```
unit test: 
```
unit_tests:
  - name: tiers_are_working
    description: "check if the logic for tiering is working correctly"
    model: customer_tier
    given:
      - input: ref('stg_tpch_customers')
        format: dict
        rows:
          - {customer_key: 629}
          - {customer_key: 4}
          - {customer_key: 1}
          - {customer_key: 26}

      - input: ref('stg_tpch_orders') 
        format: dict
        rows:
          - {customer_key: 629, total_price: 163443}
          - {customer_key: 4, total_price: 4134568}
          - {customer_key: 1,  total_price: 1428872}
          - {customer_key: 26, total_price: 418512}

    expect:
      rows:
        - {customer_key: 629,  tier_name: tier1}
        - {customer_key: 4,    tier_name: tier2}
        - {customer_key: 1,    tier_name: tier3}
        - {customer_key: 26,   tier_name: tier4}    
```

### Relevant log output

```shell
14:20:52 Compilation Error in unit_test tiers_are_working (models/marts/intermediate/intermediate.yml)
  Unit_Test 'unit_test.analytics.customer_tier.tiers_are_working' (models/marts/intermediate/intermediate.yml) depends on a node named 'stg_tpch_orders' which was not found
```

### Environment

```markdown
- OS:
- Python:
- dbt: Latest Version in dbt Cloud IDE
```

_No response_

### Additional Context

_No response_

### LINKED ISSUE #10623: [Bug] Runtime Error in unit_test when using versioned models

## The problem I'm having
DBT Unit Test is not able to find the version 2 of my model.

Let's take an example of a model I'm trying to test.
```
with

source as (

    select
        *
    from {{ ref('model') }}

)

, model_with_version as (

    select
        col1
        , col2

    from {{ ref('model_with_version', v=2) }}

)

select * from source
```
(This is an example, I know `model_with_version` is not selected in any way in this query)

With this unit test
```
unit_tests:
  - name: filter_test
    model: my_model_name
    overrides:
      macros:
        is_incremental: false
    given:
      - input: ref('model_with_version')
        rows:
          - {col1: test, col2: test2}
          - {col1: test2, col2: test2}
    expect:
      rows:
          - {col1: test, col2: test2}
```

If I run the command `dbt test --select "my_model_name,test_type:unit"` dbt will return 
```
  Unit_Test 'filter_test` depends on a node named `model_with_version` with version '2' which was not found.
```

I've make sure the version 2 exist.
I can easily run `dbt run -m model_with_version.v2` and it run successfully.

Note that the `latest` version is currently version 1, in my YAML file.

I've also tried to add the pin version to the `ref` in the unit test yml file. 
Like this `      - input: ref('model_with_version', v=2)`
But ended up with this error.

```
  An error occurred during execution of unit test 'filter_test'. There may be an error in the unit test definition: check the data types.
   Database Error
    002003 (42S02): SQL compilation error:
    Object '__DBT__CTE__MODEL_WITH_VERSION_V2' does not exist or not authorized.
```

## The context of why I'm trying to do this
I'm trying to configure a unit test for a model that uses DBT version.

## What I've already tried 
1. Configuring the models version in the `input`.
2. Trying without version

## Some example code or error messages
```
  Unit_Test 'filter_test` depends on a node named `model_with_version` with version '2' which was not found.
```

```
  An error occurred during execution of unit test 'filter_test'. There may be an error in the unit test definition: check the data types.
   Database Error
    002003 (42S02): SQL compilation error:
    Object '__DBT__CTE__MODEL_WITH_VERSION_V2' does not exist or not authorized.
```

## More Informations
If I configure the `input` as `- input: ref('model_with_version')` then in the query I can see that the query generate the static inputs
```
__dbt__cte__model_with_version
```
but my code want to select from 
```
    select
        address
        , validator_commission_rate

    from __dbt__cte__model_with_version_v2
```
Which fails.

### Expected Behavior

I shouldn't receive any errors as the `__DBT__CTE__STG_FIGMENT_VALIDATOR_REPOSITORY_V2` CTE should be created instead of `__DBT__CTE__STG_FIGMENT_VALIDATOR_REPOSITORY`

### Steps To Reproduce

1. Create 2 version of a single models.
2. In my scenario the version 1 is the `latest_version`.
3. In my scenario v1 has an alias with the same name, but WITHOUT the version ex alias:model_with_version.  (not `alias:model_with_version_v1`)
4. Version 2 is called `model_with_version_v2`.
5. Create a yml for the unit_tests.
6. add the required inputs. In my situation I have 2. (One without version and one referencing version 2 of my model)
7. The input is configure like this `- input: ref('model_with_version', v='2')`
8. This will generate the error 
```
  An error occurred during execution of unit test 'my_test'. There may be an error in the unit test definition: check the data types.
   Database Error
    002003 (42S02): SQL compilation error:
    Object '__DBT__CTE__MODEL_WITH_VERSION_V2' does not exist or not authorized.
```
9. The reason behind this error, when you look at the SQL query that was runned is that the CTE that DBT creates to inject the inputs values is name `__DBT__CTE__MODEL_WITH_VERSION`, but the one in the actual query that `select * from` has the version number after it `__DBT__CTE__MODEL_WITH_VERSION_V2`.
10. Since `select * from __DBT__CTE__MODEL_WITH_VERSION_V2` doesn't exist because the CTE that was created is called `__DBT__CTE__MODEL_WITH_VERSION` this break the test.
11. I've tried with DBT 1.8.0 and 1.8.5 and got the same error.

### Relevant log output

```shell
An error occurred during execution of unit test 'my_test'. There may be an error in the unit test definition: check the data types.
   Database Error
    002003 (42S02): SQL compilation error:
    Object '__DBT__CTE__MODEL_WITH_VERSION_V2' does not exist or not authorized.
```

### Environment

```markdown
- OS: MacOS (Intel) Sonoma 14.6.1. 
- Python:Python 3.9.19
- dbt:1.8.0 and 1.8.5
```

snowflake

### Additional Context

_No response_

### LINKED ISSUE #10880: [Bug] Unit test fails with versioned inputs

### Current Behavior

Unit tests fail if an input reference is a versioned model even though it is explicitly defined in both the model & the unit test input. The following is the error when the test fails:
`Compilation Error in unit_test unit_test_mart__model_b (models/mart/unit_test.yml)
  Unit_Test 'unit_test.repo.mart__model_b.unit_test_mart__model_b’ (models/mart/unit_test.yml) depends on a node named ‘input_versioned_model’ which was not found`

### Expected Behavior

Unit test should run with specified versioned model.

### Steps To Reproduce

1. Create a model with at least 1 version and have it saved with `_v#`, e.g. input_versioned_model_v1
2. This input_versioned_model_v1 has model.yml configured with `latest_version: 1` 
Code for reference:
```
  - name: input_versioned_model_v1 
    latest_version: 1
    columns:
      - name: id
        ...
    versions:
      - v: 1
```

3. Create mart__model_b that references this input_versioned_model without the versioning
Code for reference:
```
with iv as (
    select *
    from
        {{ ref('input_versioned_model') }}
),
im as (
    select *
    from
        {{ ref('input_model2') }}
)
select 
    iv.id,
    iv.category,
    im.profile_id,
    im.type
from 
    iv
join im
    on iv.id = im.customer_id
```
4. mart__model_b builds correctly with `dbt build -s +mart__model_b`
5. Create a dbt unit test for mart__model_b
Code for reference:
```
  - name: unit_test_mart__model_b
    model: mart__model_b
    given:
      - input: ref('input_versioned_model', v='1')
        rows:
        - {id: 123, category: electronics}
      - input: ref('input_model2')
        rows:
        - {customer_id: 123, profile_id: 2356, type: onetime}
    expect:
      rows:
      - {id: 123, category: electronics, profile_id: 2356, type: onetime}

```
Note: I tried the following versioned inputs:
`input: ref('input_versioned_model', v='1')`
`-input: ref('input_versioned_model', v=1)`
7. Run `dbt test --select "mart__model_b,test_type:unit"` will give you the following error:
`Compilation Error in unit_test unit_test_mart__model_b (models/mart/unit_test.yml)
  Unit_Test 'unit_test.repo.mart__model_b.unit_test_mart__model_b’ (models/mart/unit_test.yml) depends on a node named ‘input_versioned_model’ which was not found`

### Relevant log output

```shell
`Compilation Error in unit_test unit_test_mart__model_b (models/mart/unit_test.yml)
  Unit_Test 'unit_test.repo.mart__model_b.unit_test_mart__model_b’ (models/mart/unit_test.yml) depends on a node named ‘input_versioned_model’ which was not found`
```

### Environment

```markdown
- OS:
- Python: 3.12.6
- dbt-core: 1.8.3 & tried with 1.8.6 & 1.8.7
Plugins:
  - redshift: 1.8.1
```

redshift: 1.8.1

### Additional Context

Found following github issues that are related: [10528](https://github.com/dbt-labs/dbt-core/issues/10528) and [10623](https://github.com/dbt-labs/dbt-core/issues/10623) and tried the solutions provided.

devmessias:
Resolves #10880; Resolves #10528; Resolves #10623 

### Problem
The error raises due the method `RefableLookup.find` at  dbt-my/core/dbt/contracts/graph/manifest.py: 

Tracking the storage object, without specifying a version in the model we will obtain something like that for the Unit Test Node running phase 

 
```python
{'upstream_model_versioned': {}, 'upstream_model_versioned.v<LATEST>': {'demo_dbt_py': 'model.demo_dbt_py.upstream_model_versioned'}} 
```
 

Which it seems to me that is not right, because if I specified the `latest_version` for my model, I should get   a Storage like this 
```python
{
    'upstream_model_versioned': {'demo_dbt_py': 'model.demo_dbt_py.upstream_model_versioned'}, 
    'upstream_model_versioned.v <LATEST> ': {'demo_dbt_py': 'model.demo_dbt_py.upstream_model_versioned'} 
} 
```

I run some different cases and obtain the following table 

| Status | Model Ref | Model Latest | Test Ref | Test Latest | Expectation | v on find | Compilation Message | 
|--------|-----------|--------------|----------|-------------|-------------|---------------|---------------------| 
| ❌     | true      | true         | true     | false       | fail        | 1             | depends on a node named 'upstream_model_versioned' with version '1' which was not found | 
| ❌     | true      | false        | false    | None        | fail        | 2             | depends on a node named 'upstream_model_versioned' with version '2' which was not found | 
| ❌     | false     | None         | false    | false       | success     | None          | depends on a node named 'upstream_model_versioned' which was not found | 
| ❌     | false     | None         | true     | true        | success     | None          | depends on a node named 'upstream_model_versioned' which was not found | 
| ✅     | true      | true         | false    | None        | success     | 1             | model.demo_dbt_py.upstream_model_versioned | 
| ✅     | true      | true         | true     | true        | success     | 1             | model.demo_dbt_py.upstream_model_versioned | 
| ✅     | true      | false        | true     | false       | success     | 2             | model.demo_dbt_py.upstream_model_versioned | 

However, even in cases where the tests were successful, the storage objects it seems that not properly constructed. For example, when using `ref('upstream_model_versioned', v=1)` I got this:

```python
{'upstream_model_versioned': {}, 
 'upstream_model_versioned.v1': {'demo_dbt_py': 'model.demo_dbt_py.upstream_model_versioned'}}
```

This worked not because the storage structure was correct, but because there was a key named `upstream_model_versioned.v1`.

### Solution

The only problem that I found is that the Node object for the UnitTest was not copying the value from latest_version prop if it was available

### Checklist

### DIFF
--- a/core/dbt/parser/unit_tests.py
+++ b/core/dbt/parser/unit_tests.py
@@ -110,7 +110,6 @@ class UnitTestManifestLoader:
         # unit_test_node now has a populated refs/sources
 
         self.unit_test_manifest.nodes[unit_test_node.unique_id] = unit_test_node
-
         # Now create input_nodes for the test inputs
         """
         given:
@@ -132,7 +131,6 @@ class UnitTestManifestLoader:
                 given.input, tested_node, test_case.name
             )
             input_name = original_input_node.name
-
             common_fields = {
                 "resource_type": NodeType.Model,
                 # root directory for input and output fixtures
@@ -149,21 +147,25 @@ class UnitTestManifestLoader:
                 "name": input_name,
                 "path": f"{input_name}.sql",
             }
+            resource_type = original_input_node.resource_type
 
-            if original_input_node.resource_type in (
+            if resource_type in (
                 NodeType.Model,
                 NodeType.Seed,
                 NodeType.Snapshot,
             ):
+
                 input_node = ModelNode(
                     **common_fields,
                     defer_relation=original_input_node.defer_relation,
                 )
-                if (
-                    original_input_node.resource_type == NodeType.Model
-                    and original_input_node.version
-                ):
+
+                if not resource_type == NodeType.Model:
+                    continue
+                if original_input_node.version:
                     input_node.version = original_input_node.version
+                if original_input_node.latest_version:
+                    input_node.latest_version = original_input_node.latest_version
 
             elif original_input_node.resource_type == NodeType.Source:
                 # We are reusing the database/schema/identifier from the original source,
--- a/core/dbt/parser/unit_tests.py
+++ b/core/dbt/parser/unit_tests.py
@@ -159,16 +159,11 @@ class UnitTestManifestLoader:
                     **common_fields,
                     defer_relation=original_input_node.defer_relation,
                 )
-
-                if original_input_node.version:
-                    input_node.version = original_input_node.version
-                if original_input_node.latest_version:
-                    input_node.latest_version = original_input_node.latest_version
-            elif resource_type in (NodeType.Seed, NodeType.Snapshot):
-                input_node = ModelNode(
-                    **common_fields,
-                    defer_relation=original_input_node.defer_relation,
-                )
+                if resource_type == NodeType.Model:
+                    if original_input_node.version:
+                        input_node.version = original_input_node.version
+                    if original_input_node.latest_version:
+                        input_node.latest_version = original_input_node.latest_version
 
             elif resource_type == NodeType.Source:
                 # We are reusing the database/schema/identifier from the original source,
--- a/core/dbt/parser/unit_tests.py
+++ b/core/dbt/parser/unit_tests.py
@@ -160,12 +160,15 @@ class UnitTestManifestLoader:
                     defer_relation=original_input_node.defer_relation,
                 )
 
-                if not resource_type == NodeType.Model:
-                    continue
                 if original_input_node.version:
                     input_node.version = original_input_node.version
                 if original_input_node.latest_version:
                     input_node.latest_version = original_input_node.latest_version
+            elif resource_type in (NodeType.Seed, NodeType.Snapshot):
+                input_node = ModelNode(
+                    **common_fields,
+                    defer_relation=original_input_node.defer_relation,
+                )
 
             elif resource_type == NodeType.Source:
                 # We are reusing the database/schema/identifier from the original source,
--- a/core/dbt/parser/unit_tests.py
+++ b/core/dbt/parser/unit_tests.py
@@ -167,7 +167,7 @@ class UnitTestManifestLoader:
                 if original_input_node.latest_version:
                     input_node.latest_version = original_input_node.latest_version
 
-            elif original_input_node.resource_type == NodeType.Source:
+            elif resource_type == NodeType.Source:
                 # We are reusing the database/schema/identifier from the original source,
                 # but that shouldn't matter since this acts as an ephemeral model which just
                 # wraps a CTE around the unit test node.
--- a/tests/functional/unit_testing/test_unit_testing.py
+++ b/tests/functional/unit_testing/test_unit_testing.py
@@ -309,7 +309,7 @@ class TestUnitTestIncrementalModelWithVersion:
         assert len(results) == 2
 
 
-schema_ref_with_version = """
+schema_ref_with_versioned_model = """
 models:
   - name: source
     latest_version: {latest_version}
@@ -338,7 +338,7 @@ class TestUnitTestRefWithVersion:
             "model_to_test.sql": "select result from {{ ref('source')}}",
             "source.sql": "select 2 as result",
             "source_v2.sql": "select 2 as result",
-            "schema.yml": schema_ref_with_version.format(
+            "schema.yml": schema_ref_with_versioned_model.format(
                 **{"latest_version": 1, "input": "ref('source')"}
             ),
         }
@@ -357,7 +357,7 @@ class TestUnitTestRefMissingVersionModel:
             "model_to_test.sql": "select result from {{ ref('source')}}",
             "source_v1.sql": "select 2 as result",
             "source_v2.sql": "select 2 as result",
-            "schema.yml": schema_ref_with_version.format(
+            "schema.yml": schema_ref_with_versioned_model.format(
                 **{"latest_version": 1, "input": "ref('source', v=1)"}
             ),
         }
@@ -376,7 +376,7 @@ class TestUnitTestRefWithMissingVersionRef:
             "model_to_test.sql": "select result from {{ ref('source', v=1)}}",
             "source_v1.sql": "select 2 as result",
             "source_v2.sql": "select 2 as result",
-            "schema.yml": schema_ref_with_version.format(
+            "schema.yml": schema_ref_with_versioned_model.format(
                 **{"latest_version": 1, "input": "ref('source')"}
             ),
         }
@@ -395,7 +395,7 @@ class TestUnitTestRefWithVersionLatestSecond:
             "model_to_test.sql": "select result from {{ ref('source')}}",
             "source_v1.sql": "select 2 as result",
             "source_v2.sql": "select 2 as result",
-            "schema.yml": schema_ref_with_version.format(
+            "schema.yml": schema_ref_with_versioned_model.format(
                 **{"latest_version": 2, "input": "ref('source')"}
             ),
         }
@@ -414,16 +414,21 @@ class TestUnitTestRefWithVersionMissingRefTest:
             "model_to_test.sql": "select result from {{ ref('source', v=2)}}",
             "source_v1.sql": "select 2 as result",
             "source_v2.sql": "select 2 as result",
-            "schema.yml": schema_ref_with_version.format(
+            "schema.yml": schema_ref_with_versioned_model.format(
                 **{"latest_version": 1, "input": "ref('source')"}
             ),
         }
 
     def test_basic(self, project):
         results = run_dbt(["run"])
+
         assert len(results) == 3
-        # TODO: How to capture an compilation Error? pytest.raises(CompilationError) not working
-        run_dbt(["test", "--select", "model_to_test"], expect_pass=False)
+        # run_dbt(["test", "--select", "model_to_test"], expect_pass=False)
+        exec_result, _ = run_dbt_and_capture(
+            ["test", "--select", "model_to_test"], expect_pass=False
+        )
+        msg_error = exec_result[0].message
+        assert msg_error.lower().lstrip().startswith("compilation error")
 
 
 class TestUnitTestRefWithVersionDiffLatest:
@@ -433,7 +438,7 @@ class TestUnitTestRefWithVersionDiffLatest:
             "model_to_test.sql": "select result from {{ ref('source', v=2)}}",
             "source_v1.sql": "select 2 as result",
             "source_v2.sql": "select 2 as result",
-            "schema.yml": schema_ref_with_version.format(
+            "schema.yml": schema_ref_with_versioned_model.format(
                 **{"latest_version": 1, "input": "ref('source', v=2)"}
             ),
         }
--- a/tests/functional/unit_testing/test_unit_testing.py
+++ b/tests/functional/unit_testing/test_unit_testing.py
@@ -291,6 +291,24 @@ class TestUnitTestIncrementalModelWithAlias:
         assert len(results) == 2
 
 
+class TestUnitTestIncrementalModelWithVersion:
+    @pytest.fixture(scope="class")
+    def models(self):
+        return {
+            "my_incremental_model.sql": my_incremental_model_sql,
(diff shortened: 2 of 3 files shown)


devmessias:
Hello @dbeatty10. This is a PR with a very simple fix that closes 3 open issues. It's covered by functional tests, but I’m not sure why this class doesn’t have any unit tests, so I left that aside.

core/dbt/parser/unit_tests.py
```diff
@@ -147,23 +145,27 @@ def parse_unit_test_case(self, test_case: UnitTestDefinition):
...
                     **common_fields,
                     defer_relation=original_input_node.defer_relation,
                 )
-                if (
-                    original_input_node.resource_type == NodeType.Model
-                    and original_input_node.version
-                ):
+
+                if not resource_type == NodeType.Model:
+                    continue
```
MichelleArk, line 162: This is a little tricky to read given we're already in a `resource_type in (NodeType.Model,NodeType.Seed, NodeType.Snapshot)` block.

I'd suggest refactoring to be more straightforward to read, something like: 

```python
if resource_type == NodeType.model:
  input_node = ModelNode(
    **common_fields,
    defer_relation=original_input_node.defer_relation,
    latest_version=original_input_node.latest_version,
    version=original_input_node.version
  )
elif resource_type in (NodeType.Seed, NodeType.Snapshot):
  ...
elif resource_type == NodeType.Source:
  ... 
```

core/dbt/parser/unit_tests.py
```diff
@@ -147,23 +145,27 @@ def parse_unit_test_case(self, test_case: UnitTestDefinition):
                 "name": input_name,
                 "path": f"{input_name}.sql",
             }
+            resource_type = original_input_node.resource_type
```
MichelleArk, line 148: nice refactor! thank you

tests/functional/unit_testing/test_unit_testing.py
```diff
@@ -309,6 +309,141 @@ def test_basic(self, project):
...
+            "source_v2.sql": "select 2 as result",
+            "schema.yml": schema_ref_with_version.format(
+                **{"latest_version": 1, "input": "ref('source')"}
+            ),
+        }
+
+    def test_basic(self, project):
+        results = run_dbt(["run"])
+        assert len(results) == 3
+        # TODO: How to capture an compilation Error? pytest.raises(CompilationError) not working
```
MichelleArk, line 425: Yeah, this can be tricky to do in certain contexts depending on where the `CompilationError` is thrown internally. If pytest.raises didn't capture it, `expect_pass=False` is fine. 

However, you could change `run_dbt` to `run_dbt_and_capture` and assert the CompilationError message was found in the stdout logs. 

For example: https://github.com/dbt-labs/dbt-core/blob/bdf28d7eff13edcbe0d41746e50ffa8edcfa3cf0/tests/functional/defer_state/test_modified_state.py#L560C1-L568C39

MichelleArk:
Thank you for the contribution @devmessias! 

This is looking good overall, but it looks like this PR is failing on unrelated functional tests with the same 'strategy' error for tests related to snapshots. We've been doing a lot of work on snapshots recently so it may just be an intermittent issue that has since been resolved. Could you please merge `main` into this branch and re-run the tests?

Otherwise, just a refactoring suggestion for clarity, as well as some examples of testing for expected errors in logs.

tests/functional/unit_testing/test_unit_testing.py
devmessias: This worked! Thanks

core/dbt/parser/unit_tests.py
devmessias: done

devmessias:
Hi @MichelleArk , the branch is now updated

core/dbt/parser/unit_tests.py
```diff
@@ -149,23 +147,30 @@ def parse_unit_test_case(self, test_case: UnitTestDefinition):
...
 
-            if original_input_node.resource_type in (
+            if resource_type in (
                 NodeType.Model,
                 NodeType.Seed,
                 NodeType.Snapshot,
```
MichelleArk, line 155: NodeType.Seed and NodeType.Snapshot are covered by the `elif` statement on line 167 and can be removed. 

I believe sending snapshot/seed nodes down this codepath is what is causing the test failures since seeds and snapshots don't have an `.version` attribute: 

```sh
21:17:51  1 of 1 START unit_test customers::test_is_valid_email_address .................. [RUN]
21:17:51  Unhandled error while executing 
'SnapshotNode' object has no attribute 'version'
21:17:51  1 of 1 ERROR customers::test_is_valid_email_address ............................ [ERROR in 0.01s]
21:17:51  
21:17:51  Finished running 1 unit test in 0 hours 0 minutes and 0.08 seconds (0.08s).
...
FAILED tests/functional/unit_testing/test_ut_snapshot_dependency.py::TestUnitTestSnapshotDependency::test_snapshot_dependency
```

Failing tests: https://github.com/dbt-labs/dbt-core/actions/runs/11786025311/job/32828487620?pr=10889

devmessias: @MichelleArk, the unit/functional tests are all good now

MichelleArk (approved):

PR #10905, opened 2024-10-22
Fix: Source quoting ignores global configuration

### LINKED ISSUE #10892: [Regression] Quoting behavior for sources

### Current Behavior

In dbt-core 1.7 and earlier, quoting behavior for sources was not controlled by the quoting configuration in dbt_project.yml, but newer versions of dbt also apply those settings to sources.

In older versions of dbt, this was rendered without quotes around the schema name. The change seems to be unintentional, and we would like to return to the previous behavior.

### Steps To Reproduce

```yaml
# dbt_project.yml
quoting:
  schema: True
```

```yaml
# models/sources.yml
sources:
  - name: <sourcename>
...
```

Create a model that references the source, and note that source(<sourcename>, <tablename>) is rendered in recent versions as:

```
with source as (

    select * from <database>."<schema>".<tablename>

)
```

We can work around with a source which specifically overrides this setting:

```yaml
# models/sources.yml
sources:
  - name: <sourcename>
    quoting:
      schema: False
...
```

### Relevant log output

```shell
.
```

### Environment

```markdown
- OS: Ubuntu/macos
- Python: 3.11
- dbt: 1.8.latest
```

snowflake

### Additional Context

_No response_

MichelleArk:
Resolves #10892 (regression)

### Problem

In dbt-core 1.7 and earlier, quoting behavior for sources was not controlled by the quoting configuration in dbt_project.yml, but newer versions of dbt also apply those settings to sources.

### Solution

For sources, pass along an 'empty' quoting config as the base config, ignoring anything that's in self.config.quoting (ie dbt_project.yml). This was effectively the case in versions <1.7, which used [create_from_source](https://github.com/dbt-labs/dbt-core/blob/1.7.latest/core/dbt/adapters/base/relation.py#L207-L211) which did not incorporate the config quoting like [create_from_node](https://github.com/dbt-labs/dbt-core/blob/1.7.latest/core/dbt/adapters/base/relation.py#L241) did.

⚠️ Note: 
We are not putting this behind a behaviour flag because: 
* This behaviour was previously intentional and documented ([here](https://docs.getdbt.com/reference/project-configs/quoting) and [here](https://docs.getdbt.com/reference/resource-properties/quoting)). The quoting: config in the dbt_project.yml file is just for creating relations / resolving a ref, and this is truly addressing a regression
* It is highly unlikely that restoring the previous behaviour will cause any issues in existing projects that may be depending on the quoting config being ignored. [By default and in most adapters other than snowflake, quoting is True by default](https://docs.getdbt.com/reference/project-configs/quoting), so sources getting opted into quoting has not been problematic. It's only on snowflake that the regression has been raised because it is `False` by default, and adding quotes has been unsafe to do by default for sources when they otherwise wouldn't have been.
   * In either case, the current behaviour goes counter to our documentation and would not have been safe to rely on.
* We don't want to phase in the new behaviour / deprecate the old behavior. From @jtcohen6: 
  * people should be specifically configuring sources with nonstandard quoting, so for those people this should be no change (and the previous behavior is more desirable)

### Checklist

### DIFF
--- a/core/dbt/context/providers.py
+++ b/core/dbt/context/providers.py
@@ -684,12 +684,18 @@ class RuntimeSourceResolver(BaseSourceResolver):
                 target_kind="source",
                 disabled=(isinstance(target_source, Disabled)),
             )
-        return self.Relation.create_from(
-            self.config,
+
+        class SourceQuotingConfig:
+            quoting: Dict[str, Any] = {"database": False, "schema": False, "identifier": False}
+
+        relation = self.Relation.create_from(
+            SourceQuotingConfig(),
             target_source,
             limit=self.resolve_limit,
             event_time_filter=self.resolve_event_time_filter(target_source),
         )
+
+        return relation
 
 
 class RuntimeUnitTestSourceResolver(BaseSourceResolver):
--- a/core/dbt/context/providers.py
+++ b/core/dbt/context/providers.py
@@ -688,11 +688,11 @@ class RuntimeSourceResolver(BaseSourceResolver):
         # Source quoting does _not_ respect global configs in dbt_project.yml, as documented here:
         # https://docs.getdbt.com/reference/project-configs/quoting
         # Use an object with an empty quoting field to bypass any settings in self.
-        class SourceQuotingConfig:
+        class SourceQuotingBaseConfig:
             quoting: Dict[str, Any] = {}
 
         return self.Relation.create_from(
-            SourceQuotingConfig(),
+            SourceQuotingBaseConfig(),
             target_source,
             limit=self.resolve_limit,
             event_time_filter=self.resolve_event_time_filter(target_source),
--- a/core/dbt/context/providers.py
+++ b/core/dbt/context/providers.py
@@ -688,8 +688,15 @@ class RuntimeSourceResolver(BaseSourceResolver):
         class SourceQuotingConfig:
             quoting: Dict[str, Any] = {"database": False, "schema": False, "identifier": False}
 
+        # Preserve legacy behaviour where sources do not consider quoting configuration from dbt_project.yml
+        base_quoting_config = (
+            self.config
+            if self.config.args.source_quoting_respects_project_config
+            else SourceQuotingConfig()
+        )
+
         relation = self.Relation.create_from(
-            SourceQuotingConfig(),
+            base_quoting_config,
             target_source,
             limit=self.resolve_limit,
             event_time_filter=self.resolve_event_time_filter(target_source),
--- a/core/dbt/context/providers.py
+++ b/core/dbt/context/providers.py
@@ -688,15 +688,8 @@ class RuntimeSourceResolver(BaseSourceResolver):
         class SourceQuotingConfig:
             quoting: Dict[str, Any] = {}
 
-        # Preserve legacy behaviour where sources do not consider quoting configuration from dbt_project.yml
-        base_quoting_config = (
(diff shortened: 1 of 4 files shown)


core/dbt/context/providers.py
```diff
@@ -684,8 +684,15 @@ def resolve(self, source_name: str, table_name: str):
...
+
+        # Source quoting does _not_ respect global configs in dbt_project.yml, as documented here:
+        # https://docs.getdbt.com/reference/project-configs/quoting
+        # Use an object with an empty quoting field to bypass any settings in self.
+        class SourceQuotingConfig:
+            quoting: Dict[str, Any] = {}
+
         return self.Relation.create_from(
-            self.config,
+            SourceQuotingConfig(),
```
MichelleArk, line 695: An alternative approach would be to extend `BaseRelation.create_from` to accept an `Optional[HasQuoting]` value instead of just `HasQuoting`. That said, the method in dbt-adapters is:
* abstracted to the point where it does not concern itself with the node type (e.g. model vs seed vs source, etc), and this kind of business logic feels better-suited in dbt-core
* overridable across adapters, so could lead to inconsistent handling + need updates across multiple adapters for consistent behaviour

So I went with this approach. Open to feedback :) I opted to define `SourceBaseQuotingConfig` inline here since it quite specialized and likely won't need to be used elsewhere in the codebase.

tests/functional/relation_quoting/test_relation_quoting.py
```diff
@@ -0,0 +1,64 @@
...
+            "sources.yml": _SOURCES_YML,
+            "model.sql": "select * from {{ source('source_name', 'customers') }}",
+        }
+
+    def test_sources_ignore_global_quoting_configs(self, project):
+        run_dbt(["compile"])
+
+        generated_sql = read_file("target", "compiled", "test", "models", "model.sql")
+        assert generated_sql == 'select * from "source_database"."source_schema"."customers"'
+
```
MichelleArk, line 39: confirmed the above test does fail on main as expected: 

```sh
tests/functional/relation_quoting/test_relation_quoting.py F.                                                                                         [100%]

========================================================================= FAILURES ==========================================================================
_________________________________________ TestSourceQuotingGlobalConfigs.test_sources_ignore_global_quoting_configs _________________________________________

self = <test_relation_quoting.TestSourceQuotingGlobalConfigs object at 0x114bdba30>
project = <dbt.tests.fixtures.project.TestProjInfo object at 0x115b1e9b0>

    def test_sources_ignore_global_quoting_configs(self, project):
        run_dbt(["compile"])
    
        generated_sql = read_file("target", "compiled", "test", "models", "model.sql")
>       assert generated_sql == 'select * from "source_database"."source_schema"."customers"'
E       assert 'select * fro...ema.customers' == 'select * fro..."."customers"'
E         - select * from "source_database"."source_schema"."customers"
E         ?               -               - -             - -         -
E         + select * from source_database.source_schema.customers

/Users/michelleark/src/dbt-core/tests/functional/relation_quoting/test_relation_quoting.py:38: AssertionError
------------------------------------------------------------------- Captured stdout setup -------------------------
```

MichelleArk: Have also confirmed these tests pass with no code changes on 1.7.latest: https://github.com/dbt-labs/dbt-core/pull/10906

peterallenwebb (approved):

PR #10909, opened 2024-10-23
Behavior change cumulative type param

### LINKED ISSUE #10960: [Feature] Behavior change for updating cumulative type param fields

### Describe the feature

The location of cumulative type param fields has moved, so there should be a behavior change to indicate that the old yaml spec has been deprecated.

_No response_

### Who will this benefit?

This will benefit MetricFlow users.

Yes

### Anything else?

_No response_

DevonFulcher:
[Edited after my last commit]

Resolves #10960
Linear issue: SL-2935
docs: https://github.com/dbt-labs/docs.getdbt.com/issues/6393

### Problem & Solution
 
The type params for cumulative metrics have moved, so a behavior change needs to be implemented to surface this for users. This is what the warning looks like:
```
15:36:22  [WARNING]: Cumulative fields `type_params.window` and
`type_params.grain_to_date` have been moved and will soon be deprecated. Please
nest those values under `type_params.cumulative_type_params.window` and
`type_params.cumulative_type_params.grain_to_date`. See documentation on
behavior changes:
https://docs.getdbt.com/reference/global-configs/behavior-changes.
```
Here is what the error looks like when the user flips the behavior flag:
```
21:39:18  Cumulative fields `type_params.window` and `type_params.grain_to_date` should be nested under `type_params.cumulative_type_params.window` and `type_params.cumulative_type_params.grain_to_date`. Invalid metrics: orders_last_7_days. See documentation on behavior changes: https://docs.getdbt.com/reference/global-configs/behavior-changes.
```

Testing:
- I ran `dbt parse` in `jaffle-sl-template` and I saw the warning message due to this metric defined there:
```
    type: cumulative
    type_params:
      measure: order_count
      window: 7 days
```
- I updated this metric and didn't see the deprecation warning:
```
    type: cumulative
    type_params:
      measure:
        name: order_count
      cumulative_type_params:
        window: 7 days
```
- I changed the metric back to the deprecated version, added this to `dbt_project.yaml`, and saw the error message and that `dbt parse` failed:
```
flags:
  require_nested_cumulative_type_params: True
```

### Checklist


ChenyuLInx (commented):
This change looks good to me except for one behavior.
- I would expect when you set `require_nested_cumulative_type_params`, the warning from using `unnested param` would be an error. Since now user is required to set that param in nested fashion

DevonFulcher:
> This change looks good to me except for one behavior. Also thanks for enumerate the tests you did in the PR description!
> 
> * I would expect when you set `require_nested_cumulative_type_params`, the warning from using `unnested param` would be an error. Since now user is required to set that param in nested fashion

@ChenyuLInx So, this follows other behavior changes like [here](https://github.com/dbt-labs/dbt-core/blob/02d1902f185aee24b522497433633e8c2e9e2d50/core/dbt/contracts/graph/manifest.py#L1027) by using a warning message, not an error message. Since the legacy behavior is still supported, I think it makes sense to keep as a warning. The message is to notify to the user to upgrade their legacy yaml spec, not to indicate that there was a failure during invocation.

ChenyuLInx:
The case there is if you change the flag to true, the warning will be gone, but dbt will actually behave differently(change the macro overwrite order). The changing behavior part is an important piece of this flag.
I didn't quite follow will the behavior of the SL command change if the user flip the flag to true but has old definition. Happy to sync live really quick

ChenyuLInx:
We talked on Slack, and @DevonFulcher is going to change the behavior to failure if the schema is not adjusted but the flag is flipped to true.

ChenyuLInx (approved):

core/dbt/contracts/graph/semantic_manifest.py
```diff
@@ -60,6 +66,35 @@ def validate(self) -> bool:
...
+                    [
+                        ValidationError(
+                            context=MetricContext(
+                                # We don't have the file context at this point.
+                                file_context=FileContext(),
+                                metric=MetricModelReference(metric_name=m),
+                            ),
+                            message=f"Cumulative metric {m} fields `type_params.window` and `type_params.grain_to_date` should be nested under `type_params.cumulative_type_params.window` and `type_params.cumulative_type_params.grain_to_date`. See documentation on behavior changes: https://docs.getdbt.com/reference/global-configs/behavior-changes.",
+                        )
+                        for m in metrics_using_old_params
```
courtneyholcomb, line 94: Can we update this to be just one error that lists all the metrics with deprecated fields? We previously had complaints from customers about this because it was extremely noisy to have a warning for each metric.

core/dbt/events/types.py
```diff
@@ -476,6 +476,16 @@ def message(self) -> str:
         return line_wrap_message(warning_tag(description))
 
 
+class MFCumulativeTypeParamsDeprecation(WarnLevel):
+    def code(self) -> str:
+        return "D019"
+
+    def message(self) -> str:
```
courtneyholcomb, line 483: Would it be easy to add the names of the outdated metrics in this message? Otherwise it might be hard for users to find.
But if that doesn't work with the behavior change objects easily, it probably isn't worth it.

courtneyholcomb (approved):
One suggestion but otherwise LGTM! Thank you!

core/dbt/events/types.py
DevonFulcher: Ya, I don't think that is a straightforward change, given that it would change the interface. I think that could be an enhancement in the future.

PR #10919, opened 2024-10-25
update adapter version messages

### LINKED ISSUE #10230: [Bug] `dbt -v` doesn't know that adapters are decoupled from core

### Current Behavior

I installed dbt-core from the main branch, and ran `dbt -v`. 

I got this output: 

```
(dbt-dev) joel@Joel-Labes dbt-audit-helper % dbt -v
Core:
  - installed: 1.9.0-a1
  - latest:    1.8.1    - Ahead of latest version!

Plugins:
  - snowflake: 1.8.2   - Not compatible!
  - postgres:  1.9.0a1 - Ahead of latest version!

```

### Expected Behavior

As of 1.8, the fact that the Snowflake adapter has a different minor version isn't a problem. dbt run etc works fine, it's just the warnings that are coming from the version checking code. 

### Steps To Reproduce

Install a version of dbt calling itself 1.9 alongside a 1.8 vintage of an adapter
run `dbt -v`

### Relevant log output

_No response_

### Environment

```markdown
- OS: 
- Python:
- dbt: 1.9.0-a1
```

_No response_

### Additional Context

_No response_

dave-connors-3:
Resolves #10230 

### Problem

The messages in `dbt --version` don't reflect the fact that the minor version of an adapter could be ahead of the minor version of dbt core 

### Solution

update the semver check in `version.py`!

I added the assertions for mypy's sake, but can be convinced to do that a different way

### Checklist

### DIFF
--- a/core/dbt/version.py
+++ b/core/dbt/version.py
@@ -130,7 +130,11 @@ def _get_plugin_msg_info(
 
     needs_update = False
 
-    if plugin.major != core.major or plugin.minor != core.minor:
+    assert plugin.major is not None
+    assert plugin.minor is not None
+    assert core.major is not None
+    assert core.minor is not None
+    if plugin.major != core.major or plugin.minor < core.minor:
         compatibility_msg = red("Not compatible!")
         needs_update = True
         return (compatibility_msg, needs_update)
--- a/core/dbt/version.py
+++ b/core/dbt/version.py
@@ -8,7 +8,7 @@ from typing import Iterator, List, Optional, Tuple
 import requests
 
 import dbt_common.semver as semver
-from dbt_common.ui import green, red, yellow
+from dbt_common.ui import green, yellow
 
 PYPI_VERSION_URL = "https://pypi.org/pypi/dbt-core/json"
 
@@ -19,7 +19,7 @@ def get_version_information() -> str:
 
     core_msg_lines, core_info_msg = _get_core_msg_lines(installed, latest)
     core_msg = _format_core_msg(core_msg_lines)
-    plugin_version_msg = _get_plugins_msg(installed)
+    plugin_version_msg = _get_plugins_msg()
 
     msg_lines = [core_msg]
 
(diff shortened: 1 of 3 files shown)


core/dbt/version.py
```diff
@@ -130,7 +130,11 @@ def _get_plugin_msg_info(
 
     needs_update = False
 
-    if plugin.major != core.major or plugin.minor != core.minor:
+    assert plugin.major is not None
```
colin-rogers-dbt, line 133: `asserts` shouldn't be used in production, see: https://snyk.io/blog/the-dangers-of-assert-in-python/

core/dbt/version.py
```diff
@@ -130,7 +130,11 @@ def _get_plugin_msg_info(
...
+    assert plugin.minor is not None
+    assert core.major is not None
+    assert core.minor is not None
+    if plugin.major != core.major or plugin.minor < core.minor:
```
QMalcolm, line 137: So I talked with @colin-rogers-dbt + @mikealfare about this. I think we can get safely rid of this check entirely. Basically, we've decoupled the plugin (adapter specific) version and the core version. The majors no longer need to be equal, and the minor can be behind the core minor. The later might not actually be that uncommon.

All of this is the case because PyPI now handles this via `dbt-adapters`. Essentially, `dbt-core` specifies what `dbt-adapters` versions it is compatible with, and each dbt adapter also specifies what `dbt-adapters` versions it is compatible with. The result being, you'd have to intentionally force install incompatible versions for this check to be relevant. Thus, this check is unneeded.

QMalcolm (changes requested):
We should actually just rip this version checking logic out 🤔

core/dbt/version.py
dave-connors-3: even better! you want to do the honors or should i?

MichelleArk (approved):

QMalcolm (approved):

Issue #10926, opened 2024-10-28
[Feature] Emit a warning if none of a microbatch model's ref/source inputs have an event_time defined


graciegoheen:
### Describe the feature

An incremental model only filters it's upstream ref/source inputs if the upstream has an `event_time` config defined. We should emit a warning if **none** of a microbatch model's ref/source inputs have an `event_time` defined.

### Why

We've found the following to not be an uncommon issue:
1. Someone creates a microbatch model, `model_a`
2. `model_a` has one or more input model/sources
3. an `event_time` config is not added to _any_ of `model_a`'s inputs
4. `model_a` gets run, but every batch has the same number of rows (which matches the combined size of the inputs)

This is a problem because it isn't immediately apparent to the person that created `model_a` what happened / went wrong.

PR #10929, opened 2024-10-28
raise MicrobatchModelNoEventTimeInputs warning when no microbatch input has event_time config

MichelleArk:
Resolves https://github.com/dbt-labs/dbt-core/issues/10926

### Problem

It can be unintuitive for users that they need to configure microbatch models by adding `event_time` fields to input models. For microbatch models with no `event_time` configuration, this can lead to unexpected duplicate data given it is likely not the intended use case to have microbatch but no event time filters.

### Solution

Emit a warning only when no inputs to a microbatch model have an event_time config.

### Checklist

🎩 
Screenshot 2024-10-28 at 5 36 49 PM


core/dbt/events/types.py
```diff
@@ -924,6 +924,19 @@ def message(self) -> str:
...
 
 
+class MicrobatchModelNoEventTimeInputs(WarnLevel):
+    def code(self) -> str:
+        return "I074"
+
+    def message(self) -> str:
+        msg = (
+            f"The microbatch model '{self.model_name}' has no 'ref' or 'source' input with an 'event_time' configuration. "
+            "This can result in unexpected duplicate records in the resulting microbatch model."
```
MichelleArk, line 934: cc @graciegoheen for any input on the warning message

MichelleArk: much better!

Issue #10934, opened 2024-10-29
Missing `run_results.json` from failed runs (versionless)


taylorterwin:
### Is this a new bug in dbt-core?

- [X] I believe this is a new bug in dbt-core
- [X] I have searched the existing issues, and I could not find an existing issue for this bug

### Current Behavior

Raising this issue on behalf of customer - they are reporting no `run_results.json` were generated for failed job runs on versionless. Downgrading to dbt version 1.7 does the issue with the missing artifact. Their job is invoking `dbt --fail-fast run` and the issue started on Thu, 03 Oct 2024 11:43:48 UTC. They are using dbt API's to query for the `artifacts/run_results.json` file for dbt runs. This is causing broken pipelines for the customer. 

This is also causing the following error from dbt Cloud CLI as well: 
`Error: Cloud CLI detected incompatible schemas. This issue is likely occurring because your core version is lower than the production job that provides the deferred manifest. Please consider upgrading your core version to match or exceed the version of the production job.`

This occurrence is related to deferral within dbt Cloud CLI. 
For Cloud CLI - in order to let dev job recognize defer schema, the dev core version must >= prod run core version.

However, removing the `--no-deferral` flag does not resolve the issue with missing artifact; failed jobs on versionless are still missing the `run_results.json`.

Worth noting that manifest and logs were both present for failed runs (and logs indicate that execution finished with one model having a database error). 

### Expected Behavior

run_results.json should be uploaded and reviewable in artifacts 

### Steps To Reproduce

This has not been reproduced yet but this is occurring on CI jobs with and without deferral on versionless 

### Environment

```markdown
- dbt: 2024.10.258
- snowflake=1.9.0-post8+eea98443b21f6f3bdc3d297b727a0a9424d54d7b
```

### Which database adapter are you using with dbt?

snowflake

dbeatty10:
From @QMalcolm:

> We have reproduced the issue in core by forcing a database exception during the execution of on-run-end hooks

dbeatty10:
### Reprex

Create these files:

`macros/rce.sql`

```sql
{% macro rce(relation) %}

{% if execute %}

{{ exceptions.raise_compiler_error("Always raise a compiler error in execute") }}

{% endif %}

{% endmacro %}

```

`dbt_project.yml`

```yaml
on-run-end:
  - "{{ rce() }}"
```

`models/my_model.sql`

```sql
select 1 as id
```

Run these commands:

```shell
dbt clean
[...]

PR #10936, opened 2024-10-29
Qmalcolm  add event for debugging artifact writing

QMalcolm:
Resolves #

### Checklist


.changes/unreleased/Features-20241029-181728.yaml
```diff
@@ -0,0 +1,6 @@
+kind: Features
+body: Emit debug logging event whenever artifacts are written
```
ChenyuLInx, line 2: Do we still need this?

core/dbt/task/run.py
```diff
@@ -93,11 +93,16 @@ def get_execution_status(sql: str, adapter: BaseAdapter) -> Tuple[RunStatus, str
         response, _ = adapter.execute(sql, auto_begin=False, fetch=False)
         status = RunStatus.Success
         message = response._message
+    except (KeyboardInterrupt, SystemExit):
```
ChenyuLInx, line 96: Does this mean if we cancel during get_execution_status no result file will be written?

core/dbt/task/run.py
```diff
@@ -645,7 +650,6 @@ def safe_run_hooks(
             return RunStatus.Success
 
         status = RunStatus.Success
-        failed = False
```
ChenyuLInx, line 648: Thanks for making this logic more concise. 

If I understand it correctly, we want the behavior to be: if a node fails, we want to skip the rest?

I think the direction of `status` and `failed` is duplicate is spot on, but I am wondering should we get rid of status and rename failed to something like `skip_rest_due_to_failure` instead? In the current structure I have to do some guess and validation in order to understand what the code is trying to do.

ChenyuLInx (commented):
Generally looks good to me!
I believe the adding more events part should belong to a separate PR if that's useful, thoughts?

.changes/unreleased/Features-20241029-181728.yaml
QMalcolm: I think "need" is a strong word. We wanted to add this event because we couldn't tell if the writing of the artifact was actually happening when debugging the issue of #10934 in cloud. Having this log could thus reduce debug time in the future if something like this happens again. Additionally, it's possible that throughout execution we end up writing an artifact multiple times, which is likely wasteful. Having a log like this would be helpful in identifying if that is happening. Additionally, the fix for the #10934 isn't guaranteed by the fix in [0223f3a](https://github.com/dbt-labs/dbt-core/pull/10936/commits/0223f3a16ea0344bcc56ebc8ce4ef867164587a1). That fix is fairly narrowly scoped. It is possible if an exception gets raised elsewhere, we might run into a situation again where an artifact is not being written (thus the log is useful).

---
Tangentially, I just admitted the fix in [0223f3a](https://github.com/dbt-labs/dbt-core/pull/10936/commits/0223f3a16ea0344bcc56ebc8ce4ef867164587a1) is pretty limited. This is unfortunately necessary. I tried moving the exception handling up some levels, but quickly you run into a labrynth of catching too much in instances where failure is expected 🫠

core/dbt/task/run.py
QMalcolm: Correct. We could re-catch the `KeyboardInterrupt / SystemExit` in [after_run](https://github.com/dbt-labs/dbt-core/blob/dd77210756e6432ef2737ab59a65c238e3abfdda/core/dbt/task/run.py#L775) and do something similar to what is done in [execute_nodes](https://github.com/dbt-labs/dbt-core/blob/dd77210756e6432ef2737ab59a65c238e3abfdda/core/dbt/task/runnable.py#L421).

core/dbt/task/run.py
QMalcolm: That is correct, if one _hook_ fails, the rest are expected to be skipped. 

I can clean up the code a bit more, I was trying to avoid too much refactoring in this PR as it's generally best practice to separate feature/bug work from refactors. I actually walked back some refactoring work I did in this regard.

---

To the keen eyed observer, [5e9d3d6](https://github.com/dbt-labs/dbt-core/pull/10936/commits/5e9d3d6bf5f7ad139d24e3e21421f7837550284f) foreshadowed some work that I actually dropped before pushing my branch:
> If it isn't apparent, I'm trying to reduce the need for tracking variables
to make it easier extract the execution code to a separate private function
to make it easier to see what is happening.

QMalcolm:
@ChenyuLInx what if I actually split this into 3 PRs? It'll be easier to see what is happening and separates the work streams appropriately
1) Emitting events
2) Fixing `safe_run_hooks`
3) Refactor focused on simplification

QMalcolm:
Closing in favor of:
* https://github.com/dbt-labs/dbt-core/pull/10941
* https://github.com/dbt-labs/dbt-core/pull/10940

PR #10949, opened 2024-10-30
Only try to create log directory if file logging is enabled

### LINKED ISSUE #10948: [Bug] DBT tries to create a log directory even when file logging is disabled

### Current Behavior

I ran into this while deploying DBT core in an environment where dbt's current working directory is not writeable. I had set `DBT_LOG_PATH` to a separate writeable directory, but then realized recently that we actually want to entirely disable DBT's file logging because we store the stdout/stderr output separately.

After some digging through the source (because I couldn't find documentation of how to disable the file logging), I found this line:

https://github.com/dbt-labs/dbt-core/blob/dd77210756e6432ef2737ab59a65c238e3abfdda/core/dbt/events/logging.py#L95

A bit more digging and I found that I could set a `DBT_LOG_LEVEL_FILE=none` environment variable to effectively disable the file logging. I figured that after disabling file logging, I would no longer need to set `DBT_LOG_PATH` so I removed that environment variable.

Upon testing, I was surprised to see that `dbt debug` now provided no output and only exited with error code 2. I'd run into this issue before when it couldn't open the log file for writing, so I added `DBT_LOG_PATH=/tmp` back to the environment, and it ran successfully.

### Expected Behavior

It seems to me that if file logging is disabled, dbt should not try to make the logging directory.

### Steps To Reproduce

1. In an environment where DBT cannot make a directory in its current working directory
2. With env vars:
    ```sh
    DBT_TARGET_PATH=/tmp # or some writeable directory
    DBT_PACKAGES_DIR=/tmp # or some writeable directory
    DBT_LOG_LEVEL_FILE=none
    ```
3. Run `dbt debug`
4. See no output and exit code 2 (via `echo $?`)

Instead, I would expect `dbt debug` to succeed and print its output. Nothing should be logged to the file system.

### Relevant log output

_No response_

### Environment

```markdown
- OS: Amazon Linux 2023
- Python: 3.11.6
- dbt: 1.8.7
```

postgres

### Additional Context

I dug back through the git history a bit and I think the `make_log_dir_if_missing` call should have been moved under the if statement when the `none` option was added in commit a2213abbc0ba932ab164160ce98f8b2c63e74d5c

ericfreese:
Resolves #10948

### Problem

I ran into this while deploying DBT core in an environment where dbt's current working directory is not writeable. I had set `DBT_LOG_PATH` to a separate writeable directory, but then realized recently that we actually want to entirely disable DBT's file logging because we store the stdout/stderr output separately.

After some digging through the source (because I couldn't find documentation of how to disable the file logging), I found this line:

https://github.com/dbt-labs/dbt-core/blob/dd77210756e6432ef2737ab59a65c238e3abfdda/core/dbt/events/logging.py#L95

A bit more digging and I found that I could set a `DBT_LOG_LEVEL_FILE=none` environment variable to effectively disable the file logging. I figured that after disabling file logging, I would no longer need to set `DBT_LOG_PATH` so I removed that environment variable.

Upon testing, I was surprised to see that `dbt debug` now provided no output and only exited with error code 2. I'd run into this issue before when it couldn't open the log file for writing, so I added `DBT_LOG_PATH=/tmp` back to the environment, and it ran successfully.

It seems to me that if file logging is disabled, dbt should not try to make the logging directory.

### Solution

As mentioned above, I can work around the problem for now by setting `DBT_LOG_PATH=/tmp` so that it will just create an empty directory in `/tmp` but that feels kind of silly.

I think it would be best to only try to make the logging directory if file logging is enabled, as the changes in my PR implement.

### Checklist

### DIFF
--- a/.changes/unreleased/Fixes-20241030-130239.yaml
+++ b/.changes/unreleased/Fixes-20241030-130239.yaml
@@ -1,0 +1,6 @@
+kind: Fixes
+body: Don't try to create the log directory when file logging is disabled
+time: 2024-10-30T13:02:39.206813142-06:00
+custom:
+    Author: ericfreese
+    Issue: "10948"
--- a/core/dbt/events/logging.py
+++ b/core/dbt/events/logging.py
@@ -67,7 +67,6 @@ def _logfile_filter(log_cache_events: bool, line_format: LineFormat, msg: EventM
 
 def setup_event_logger(flags, callbacks: List[Callable[[EventMsg], None]] = []) -> None:
     cleanup_event_logger()
-    make_log_dir_if_missing(flags.LOG_PATH)
     event_manager = get_event_manager()
     event_manager.callbacks = callbacks.copy()
     add_callback_to_manager(track_behavior_change_warn)
@@ -93,6 +92,8 @@ def setup_event_logger(flags, callbacks: List[Callable[[EventMsg], None]] = []) 
         add_logger_to_manager(console_config)
 
     if flags.LOG_LEVEL_FILE != "none":
+        make_log_dir_if_missing(flags.LOG_PATH)
+
         # create and add the file logger to the event manager
         log_file = os.path.join(flags.LOG_PATH, "dbt.log")
         log_file_format = _line_format_from_str(flags.LOG_FORMAT_FILE, LineFormat.DebugText)


.changes/unreleased/Fixes-20241030-130239.yaml
```diff
@@ -0,0 +1,6 @@
+kind: Fixes
```
dbeatty10, line 1: We're considering this a [feature rather than a fix](https://github.com/dbt-labs/dbt-core/issues/10948#issuecomment-2452219657), so could you make this update?

```suggestion
kind: Features
```

dbeatty10 (commented):
Thanks for opening this PR @ericfreese !

We haven't finished evaluating https://github.com/dbt-labs/dbt-core/issues/10948 as a feature request yet (i.e. whether we want to commit to this behavior or not).

If we do choose to adopt it, then we'd need to add applicable testing to this PR.

Issue #10928, opened 2024-10-28
[Bug] Microbatch:  Calling ref() in a macro with a microbatch model name results in error 


brian-franklin:
### Is this a new bug in dbt-core?

- [X] I believe this is a new bug in dbt-core
- [X] I have searched the existing issues, and I could not find an existing issue for this bug

### Current Behavior

When running a macro that executes ref() with a microbatch model name, an error occurs: `'Macro' object has no attribute 'config'`

### Expected Behavior

ref() should resolve to a relation with the 3-part name of the object.

### Steps To Reproduce

1. using dbt-snowflake adapter version 1.9.0-b1 and dbt-core version 1.9.0-b2
2. Create a model using the microbatch incremental strategy
3. Create a new macro file with the code:
```sql
{% macro test_ref_microbatch(node) %}
  {{ log('Running for node: ' ~ node, info=True) }}
  {% set relation=ref(node) %}
  {{ log('relation: ' ~ relation, info=True) }}
{% endmacro %}
```
4. Call the macro passing the name of themicrobatch model created in step 2 (replace model_name
```shell
dbt run-operation test_ref_microbatch --args '{node: model_name}'
```

### Relevant log output

```shell
============================== 15:06:47.975382 | ad60fa66-5fe6-4d70-a9ae-1cb7092779f7 ==============================
[0m15:06:47.975382 [info ] [MainThread]: Running with dbt=1.9.0-b2
[0m15:06:47.975382 [debug] [MainThread]: running dbt with arguments {'printer_width': '80', 'indirect_selection': 'eager', 'log_cache_events': 'False', 'write_json': 'True', 'partial_parse': 'True', 'cache_selected_only': 'False', 'warn_error': 'None', 'fail_fast': 'False', 'debug': 'False', 'log_path': 'C:\\repos\\hmi_dbt_lib\\logs', 'profiles_dir': 'C:\\Users\\BFranklin\\.dbt', 'version_check': 'True', 'use_colors': 'True', 'use_experimental_parser': 'False', 'no_print': 'None', 'quiet': 'False', 'empty': 'None', 'log_format': 'default', 'static_parser': 'True', 'introspect': 'True', 'invocation_command': 'dbt run-operation test_ref_microbatch --args {node: content}', 'target_path': 'None', 'warn_error_options': 'WarnErrorOptions(include=[], exclude=[])', 'send_anonymous_usage_stats': 'True'}
[...]

PR #10975, opened 2024-11-05
[Fix] assert resolver.model is ModelNode prior to resolving event_time_filter

MichelleArk:
Resolves https://github.com/dbt-labs/dbt-core/issues/10928

### Problem

From the issue:
> When running a macro that executes ref() with a microbatch model name, an error occurs: 'Macro' object has no attribute 'config'

This error could surface from any resolver context that has a non-model node, since event_time_filters should only be resolved for microbatch incremental model nodes.

### Solution

add an additional isinstance check on resolver.model to ensure it's actually a model prior to trying to resolve an event time fitler.

### Checklist


QMalcolm (approved):
:shipit:

PR #10990, opened 2024-11-12
Update core to support DSI 0.8.3

WilliamDee:
### Problem
Upgrade to DSI 0.8.3 to support passing custom time grain in the yaml spec

schema change PR: https://github.com/dbt-labs/schemas.getdbt.com/pull/70

### Checklist

### DIFF
--- a/core/dbt/artifacts/resources/v1/metric.py
+++ b/core/dbt/artifacts/resources/v1/metric.py
@@ -48,6 +48,14 @@ class MetricTimeWindow(dbtClassMixin):
     count: int
     granularity: str
 
+    @property
+    def window_string(self) -> str:  # noqa: D
+        return f"{self.count} {self.granularity}"
+
+    @property
+    def is_standard_granularity(self) -> bool:  # noqa: D
+        return self.granularity.casefold() in {item.value.casefold() for item in TimeGranularity}
+
 
 @dataclass
 class MetricInput(dbtClassMixin):
--- a/core/dbt/artifacts/resources/v1/metric.py
+++ b/core/dbt/artifacts/resources/v1/metric.py
@@ -46,7 +46,7 @@ class MetricInputMeasure(dbtClassMixin):
 @dataclass
 class MetricTimeWindow(dbtClassMixin):
     count: int
-    granularity: TimeGranularity
+    granularity: str
 
 
 @dataclass
@@ -55,7 +55,7 @@ class MetricInput(dbtClassMixin):
     filter: Optional[WhereFilterIntersection] = None
     alias: Optional[str] = None
     offset_window: Optional[MetricTimeWindow] = None
-    offset_to_grain: Optional[TimeGranularity] = None
+    offset_to_grain: Optional[str] = None
 
     def as_reference(self) -> MetricReference:
         return MetricReference(element_name=self.name)
@@ -83,7 +83,7 @@ class ConversionTypeParams(dbtClassMixin):
 @dataclass
 class CumulativeTypeParams(dbtClassMixin):
     window: Optional[MetricTimeWindow] = None
-    grain_to_date: Optional[TimeGranularity] = None
+    grain_to_date: Optional[str] = None
     period_agg: PeriodAggregation = PeriodAggregation.FIRST
 
 
@@ -95,7 +95,9 @@ class MetricTypeParams(dbtClassMixin):
     denominator: Optional[MetricInput] = None
     expr: Optional[str] = None
     window: Optional[MetricTimeWindow] = None
-    grain_to_date: Optional[TimeGranularity] = None
+    grain_to_date: Optional[TimeGranularity] = (
+        None  # legacy, use cumulative_type_params.grain_to_date
+    )
     metrics: Optional[List[MetricInput]] = None
     conversion_type_params: Optional[ConversionTypeParams] = None
     cumulative_type_params: Optional[CumulativeTypeParams] = None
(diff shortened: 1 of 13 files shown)


core/dbt/artifacts/resources/v1/semantic_model.py
```diff
@@ -72,6 +80,7 @@ class Dimension(dbtClassMixin):
     type_params: Optional[DimensionTypeParams] = None
     expr: Optional[str] = None
     metadata: Optional[SourceFileMetadata] = None
+    config: Optional[SemanticLayerElementConfig] = None
```
QMalcolm, line 83: ✅ These are safe "minor" artifact changes as they default to `None`

core/dbt/artifacts/resources/v1/metric.py
```diff
@@ -95,7 +103,9 @@ class MetricTypeParams(dbtClassMixin):
     denominator: Optional[MetricInput] = None
     expr: Optional[str] = None
     window: Optional[MetricTimeWindow] = None
-    grain_to_date: Optional[TimeGranularity] = None
+    grain_to_date: Optional[TimeGranularity] = (
```
QMalcolm, line 106: ✅ I sanity checked that this one _didn't change_ in the manifest v12.json. Happy to report that it correctly stayed the same 🙂

core/dbt/artifacts/resources/v1/metric.py
```diff
@@ -55,7 +63,7 @@ class MetricInput(dbtClassMixin):
     filter: Optional[WhereFilterIntersection] = None
     alias: Optional[str] = None
     offset_window: Optional[MetricTimeWindow] = None
-    offset_to_grain: Optional[TimeGranularity] = None
+    offset_to_grain: Optional[str] = None
```
QMalcolm, line 66: ✅ This change of the type from `Optional[TimeGranularity]` to `Optional[str]` is safe and only a _minor_ artifact change, as the new type is a _superset_ of the old type.

core/dbt/parser/schema_yaml_readers.py
```diff
@@ -228,19 +228,11 @@ def _get_input_measures(
...
-            return None
-
-        return TimeGranularity(unparsed_grain_to_date)
-
     def _get_optional_time_window(
         self, unparsed_window: Optional[str]
     ) -> Optional[MetricTimeWindow]:
         if unparsed_window is not None:
-            parts = unparsed_window.split(" ")
+            parts = unparsed_window.lower().split(" ")
```
QMalcolm, line 235: Does DSI/Metricflow _require_ these be lowercase now? If so, this is breaking as for already parsed manifests 🤔

QMalcolm (commented):
This looks good! I had one question about MFs/DSIs expectations on case. Depending on the answer to that, we might need to make changes, or we might just be good to go 🙂

core/dbt/parser/schema_yaml_readers.py
courtneyholcomb: Nope! It should not be case sensitive. This is lowercasing it before we get it so that the user doesn't need to worry about lowercasing it but we can expect lowercase on our side. Will added tests for case sensitivity in DSI, too.

QMalcolm: The words that concern me are
> but we can expect lowercase on our side
I don't think you can unfortunately, but I could be wrong about that. Could you point me to the tests around case sensitivity in DSI?

courtneyholcomb: Yep!
https://github.com/dbt-labs/dbt-semantic-interfaces/blob/f3a50f02b962ce7193b801784740e244057ddb69/tests/parsing/test_metric_parsing.py#L215-L216
and
https://github.com/dbt-labs/dbt-semantic-interfaces/blob/f3a50f02b962ce7193b801784740e244057ddb69/tests/parsing/test_metric_parsing.py#L240-L241

courtneyholcomb: Let me go back and make sure all other fields have this covered, too

courtneyholcomb: Alright @QMalcolm - sorry for the delay on this, had some other high prios and then didn't want to mess with code freeze! 🥶 

We bumped the version of DSI one more time with some changes to fix the issues with case sensitivity. [This is the PR](https://github.com/dbt-labs/dbt-semantic-interfaces/pull/370) with those changes, but these commits are most relevant: [1](2083a249889252823b097098cb03471f2db0d00e), [2](00abab69a8b4537abcc0ea9e544547e99a3c27a2).
Let us know if you think there are any other issues here! The thorough review was much appreciated 🙏 
Once approved, we shouldn't merge quite yet - just because we want to coordinate the merge timing so that we will have time to update mantle between when this merges and the next sync to mantle.

cc @WilliamDee

QMalcolm: I think github screwed up those links somehow. They point to https://github.com/dbt-labs/dbt-core/pull/2083a249889252823b097098cb03471f2db0d00e and https://github.com/dbt-labs/dbt-core/pull/00abab69a8b4537abcc0ea9e544547e99a3c27a2. Which are commits, but on DSI, not core.

QMalcolm: For the public observer, the correct links are
1. https://github.com/dbt-labs/dbt-semantic-interfaces/pull/370/commits/2083a249889252823b097098cb03471f2db0d00e
2. https://github.com/dbt-labs/dbt-semantic-interfaces/pull/370/commits/00abab69a8b4537abcc0ea9e544547e99a3c27a2

QMalcolm (approved):
Looks good to me! Thank you for resolving the case sensitivity issue ❤️

PR #11008, opened 2024-11-15
[QOL] add index.html to .gitignore

MichelleArk:
Resolves #N/A

Running unit tests currently leaves an `index.html` artifact that you need to be mindful of checking in. Let's .gitignore it!


QMalcolm (approved):
💯

PR #11031, opened 2024-11-21
Add `batch` context object to microbatch jinja context

### LINKED ISSUE #11025: Begin supplying a `batch_id` in the mode context for batches given to adapters
In https://github.com/dbt-labs/dbt-adapters/issues/360, we're trying to ensure each batch ends up making a distinct `temp_relation` in the data warehouse (as they are currently using the _same_ `temp_relation` which is bad). In order for each batch to have a different `temp_relation`, we need to ensure that there is some sort of batch identifier information in the model context. We believe it would be best to start supplying a `batch_id` which for a given microbatch model is guaranteed to be unique, but is also human-readable to aid in debugging.

QMalcolm:
Resolves #11025

### Problem

We need to get an `id` for a batch into the jinja context for batches of models

### Solution

Add a `BatchContext` object under the `batch` key for a `ModelNodel`. The `BatchContext` contains an `id`, `event_time_start`, and `event_time_end`. The `id` field sets out the original problem we set out to solve. The `event_time_start` and `event_time_end`, formalize and supercede what we previously used `__dbt_internal_event_time_start` and `__dbt_internal_event_time_end`

As part of this work, we also _stopped_ double compiling each batch. Previously we were
(1) Compiling for the batch prior to instantiating the jinja context for the batch
(2) Compiling for the batch after instantiating the jinja context for the batch

We have stopped doing (1). As a side effect of this, the compiled `.sql` file for the microbatch model that had no event time filtering will no longer be outputted. Each batch specific compiled `.sql` file will still be produced. This _should_ be safe as `microbatch` is a new kind of model node being added in 1.9 and 1.9 GA has not been released yet. As such, nobody should be depending on its existence. If it does prove to be problematic, we've already spiked out some work on how to bring it back.

### Checklist

### DIFF
--- a/.changes/unreleased/Features-20241121-125630.yaml
+++ b/.changes/unreleased/Features-20241121-125630.yaml
@@ -1,5 +1,5 @@
 kind: Features
-body: Add `batch_id` to model context
+body: Add `batch` context object to model jinja context
 time: 2024-11-21T12:56:30.715473-06:00
 custom:
   Author: QMalcolm
--- a/.changes/unreleased/Features-20241121-125630.yaml
+++ b/.changes/unreleased/Features-20241121-125630.yaml
@@ -1,0 +1,6 @@
+kind: Features
+body: Add `batch_id` to model context
+time: 2024-11-21T12:56:30.715473-06:00
+custom:
+  Author: QMalcolm
+  Issue: "11025"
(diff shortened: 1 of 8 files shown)


QMalcolm:
16 commits for `+90`/`-41`, great job me 🫠

MichelleArk:
Kicked off dbt-snowflake tests, since Snowflake's microbatch implementation depends on the now-legacy (but still supported) model.config attributes: https://github.com/dbt-labs/dbt-snowflake/actions/runs/12057315923

MichelleArk (approved):

PR #11041, opened 2024-11-24
fix/ParsedNode for seed node dup

### LINKED ISSUE #11040: [Bug]  ParsedNode for seed node dup

### Current Behavior

```python
    @classmethod
    def _deserialize(cls, dct: Dict[str, int]):
        # The serialized ParsedNodes do not differ from each other
        # in fields that would allow 'from_dict' to distinguis
        # between them.
        resource_type = dct["resource_type"]
        if resource_type == "model":
            return ModelNode.from_dict(dct)
        elif resource_type == "analysis":
            return AnalysisNode.from_dict(dct)
        elif resource_type == "seed":
            return SeedNode.from_dict(dct)
        elif resource_type == "sql":
            return SqlNode.from_dict(dct)
        elif resource_type == "test":
            if "test_metadata" in dct:
                return GenericTestNode.from_dict(dct)
            else:
                return SingularTestNode.from_dict(dct)
        elif resource_type == "operation":
            return HookNode.from_dict(dct)
        elif resource_type == "seed":
            return SeedNode.from_dict(dct)
        elif resource_type == "snapshot":
            return SnapshotNode.from_dict(dct)
        else:
            return cls.from_dict(dct)
```

https://github.com/dbt-labs/dbt-core/blob/main/core/dbt/contracts/graph/nodes.py

### Expected Behavior

we need seed node only once

### Steps To Reproduce

https://github.com/dbt-labs/dbt-core/blob/main/core/dbt/contracts/graph/nodes.py

### Relevant log output

_No response_

### Environment

```markdown
- OS: Ubuntu20.04
- Python: 3.10
- dbt: 1.1.1
```

_No response_

### Additional Context

_No response_

JamesBonddu:
Resolves #11040

### Problem

clean code, Remove Duplicate seedNode from ParsedNode

### Solution

remove it

### DIFF
--- a/core/dbt/contracts/graph/nodes.py
+++ b/core/dbt/contracts/graph/nodes.py
@@ -305,8 +305,6 @@ class ParsedNode(ParsedResource, NodeInfoMixin, ParsedNodeMandatory, Serializabl
                 return SingularTestNode.from_dict(dct)
         elif resource_type == "operation":
             return HookNode.from_dict(dct)
-        elif resource_type == "seed":
-            return SeedNode.from_dict(dct)
         elif resource_type == "snapshot":
             return SnapshotNode.from_dict(dct)
         else:


dbeatty10:
Good eyes @JamesBonddu -- looks like this logic was accidentally duplicated in https://github.com/dbt-labs/dbt-core/pull/3364. Makes sense to remove the 2nd one.

JamesBonddu:
> Good eyes @JamesBonddu -- looks like this logic was accidentally duplicated in #3364. Makes sense to remove the 2nd one.

THK, haha!

PR #11051, opened 2024-11-25
Fix code to work with mashumaro 3.15

### LINKED ISSUE #11044: mashumaro incompatibility: node.external.partitions returned as string instead of dict

### Current Behavior

In dbt 1.8.9 it seems like the behavior of node.external.partitions has changed. It used to return a list of dicts, but now it seems to return a list of json strings.
This seems to be causing problems for dbt-external-tables when trying to create external tables with partitions defined.

I'm using this macro to compare the behaviors:
```jinja-sql
{% macro get_partitions() -%}
    {% set source_nodes = graph.sources.values() if graph.sources else [] %}
    {% for node in source_nodes %}
        {% if node.external %}
            {% if node.external.partitions %}
                {{print(node.external.partitions)}}
            {% endif %}
        {% endif %}
    {% endfor %}
{%- endmacro %}
```
dbt 1.8.8:
```cmd
dbt run-operation get_partitions
11:23:26  Running with dbt=1.8.8
11:23:27  Registered adapter: snowflake=1.8.4
11:23:31  Found 1412 models, 2 snapshots, 6 seeds, 5 operations, 129 data tests, 619 sources, 2 exposures, 730 macros
[{'name': 'dl_partition', 'data_type': 'string', 'expression': "<redacted>"}]
```

dbt 1.8.9:
```cmd
dbt run-operation get_partitions
11:21:38  Running with dbt=1.8.9
11:21:38  Registered adapter: snowflake=1.8.4
11:21:43  Found 1396 models, 2 snapshots, 6 seeds, 5 operations, 129 data tests, 611 sources, 2 exposures, 735 macros
['{\'name\': \'dl_partition\', \'data_type\': \'string\', \'expression\': "<redacted>"}']
```

I expect node.external.partitions to be a list of dicts so that the values can be accessed by looping the list and addressing with i.e. `partition.name`

### Steps To Reproduce

1. Install dbt 1.8.9
2. Create a yaml source file called `testtable.yml` with external and partition set. I.e.:
```yml
version: 2
sources:
- name: TEST
  schema: STAGE
  tables:
  - name: TABLE
    external:
      partitions:
      - name: dl_partition
        data_type: string
        expression: split_part(METADATA$FILENAME, '/', 2)
```
3. Create a test macro called get_partitions.sql:
```sql-jinja
{% macro get_partitions() -%}
    {% set source_nodes = graph.sources.values() if graph.sources else [] %}
    {% for node in source_nodes %}
        {% if node.external %}
            {% if node.external.partitions %}
                {{print(node.external.partitions)}}
            {% endif %}
        {% endif %}
    {% endfor %}
{%- endmacro %}
```
4. Run the macro:
`dbt run-operation get_partitions`

### Relevant log output

_No response_

### Environment

```markdown
- OS: Windows 11 23H2/Ubuntu 22.04.5 LTS
- Python: Python 3.12.4
- dbt (working version): 1.8.8
- dbt (regression version): 1.8.9
```

snowflake

### Additional Context

_No response_

gshank:
Resolves #11044

### Problem

Mashumaro 3.15 caused some breakages in our code.

### Solution

In a couple of cases, Unions had to be adjusted to de-serialize and serialize in the right order. Most of the effort went into refactoring the update node config code so that "from_dict" was not performed until after validation, so that config values were not converted to valid forms prior to being validated.  In general the pattern is that all sources of configs should be combined as dictionaries, then validated, then used to construct a valid config object. 

dbt-common BaseConfig.update_from method had to be converted to a class method and the "from_dict" removed. 

Since there were already a bunch of changes in core/dbt/context/context_config.py, I also took the opportunity to rename ContextConfig to ConfigBuilder, and replace variables and parameters using it in the rest of the parsing codebase.

Since the hooks needed to be fixed prior to validation in generate_node_config, I moved the hook fixing method to context_config.py and removed it from parser/base.py. 

### Checklist


gshank:
Also see https://github.com/dbt-labs/dbt-common/pull/228

gshank:
https://github.com/Fatal1ty/mashumaro/issues/269

core/dbt/contracts/graph/model_config.py
```diff
@@ -53,14 +53,5 @@ class UnitTestNodeConfig(NodeConfig):
 }
 
 
-# base resource types are like resource types, except nothing has mandatory
-# configs.
-BASE_RESOURCE_TYPES: Dict[NodeType, Type[BaseConfig]] = RESOURCE_TYPES.copy()
-
-
 def get_config_for(resource_type: NodeType, base=False) -> Type[BaseConfig]:
```
MichelleArk, line 56: Is the optional `base` parameter still necessary? It looks unused. 

And for my own understanding, are we no longer calling this method for resource types not in `RESOURCE_TYPES`?

core/dbt/context/context_config.py
```diff
@@ -131,85 +71,112 @@ def _active_project_configs(
...
         fqn: List[str],
         resource_type: NodeType,
         project_name: str,
-        base: bool,
         patch_config_dict: Optional[Dict[str, Any]] = None,
-    ) -> BaseConfig:
-        own_config = self.get_node_project(project_name)
+    ) -> Dict[str, Any]:
+        """This method takes resource configs from the project, the model (if applicable),
+        and the patch, and combines them into one config dictionary."""
```
MichelleArk, line 83: this whole method is so much more readable now, thank you ✨

core/dbt/context/context_config.py
```diff
@@ -222,69 +189,107 @@ def translate_hook_names(self, project_dict):
...
             project_name=project_name,
-            base=base,
             patch_config_dict=patch_config_dict,
         )
+        fix_hooks(config_dict)
         try:
-            finalized = config.finalize_and_validate()
-            return finalized.to_dict(omit_none=True)
+            config_cls.validate(config_dict)
+            config_obj = config_cls.from_dict(config_dict)
```
MichelleArk, line 213: I'm noticing we are [no longer omitting `None` values](https://github.com/dbt-labs/dbt-core/pull/11051/files#diff-5b85f0049476b4454944daa5f50db50088ff1eb3b095d59ce6f1e393a088bc06L76) during construction here. 

Is the implication of that that even `None` configurations will make their way onto the parsed node representation + get serialized out? I don't think this is a huge deal, just trying to wrap my head around any potential downsides of the change.

core/dbt/contracts/graph/model_config.py
gshank: It's never been used. As far as I can see the result for base=True and base=False has always been identical. I removed the other instances but missed it in the signature. Now removed.

core/dbt/parser/base.py
```diff
@@ -396,61 +393,53 @@ def update_parsed_node_config(
...
             return
         if not context:
-            context = self._context_for(parsed_node, config)
+            context = self._context_for(parsed_node, config_builder)
         for hook in hooks:
             get_rendered(hook.sql, context, parsed_node, capture_macros=True)
 
-    def initial_config(self, fqn: List[str]) -> ContextConfig:
-        config_version = min([self.project.config_version, self.root_project.config_version])
-        if config_version == 2:
```
MichelleArk, line 430: Is it safe to remove this check entirely at this point?

gshank: The config_version == 2?  Yeah, we haven't required that or checked in at least a year and a half. I think pre-version-2 was way back in... version 14?  Before I started anyway.

core/dbt/context/context_config.py
gshank: The config_dict is the merged dicts from all the various sources, and the method actually returns a config obj. It doesn't really matter whether you've omitted the None fields when constructing an object. This is called in build_config_dict, which converts it to a dictionary with omit_none, so it should actually be the same dictionary that was being returned before.

jboutwell-rva:
I'm wondering if there any any particular blocker remaining for this getting merged.  The version constraint is conflicting with other libraries we use and preventing upgrading.

honnix:
Another half year has passed since the last comment. I'm wondering whether this should be prioritised a bit given mashumaro is claimed as "dbt-core uses these packages deeply, throughout the codebase". I think for such a critical dependency, it is better to address blocking issues as soon as possible so new code will not contribute more to the already complex problem.

moreaupascal56:
Hi after some investigations I was about to open a PR to remove the snapshot finalize_and_validate function as this makes impossible to use the snowflake_warehouse config in yaml snapshot files. 
Would love to see this merged as well

prakharcode:
any update on this? There is still a problem when using dbt-core with Python 3.14.

mfatemipour:
Hey, Any update on this?

MatthijsKok:
> Hey, Any update on this?

mashumaro 3.15 compat was merged in `dbt-common` (not `dbt-core`!) yesterday https://github.com/dbt-labs/dbt-common/pull/363

MatthijsKok:
See also https://github.com/dbt-labs/dbt-core/pull/12828

ash2shukla:
mashumaro compatibility is addressed here: https://github.com/dbt-labs/dbt-core/pull/12828 Context config refactor can be taken up in a separate tech debt. Closing this as I think corresponding tidying work would be in a separate PR.

PR #11092, opened 2024-12-03
Improve the Performance Characteristics of add_test_edges()

### LINKED ISSUE #10950: [SPIKE+] Improve the Performance Characteristics of add_test_edges()
### Housekeeping

### Short description

The [add_test_edges()](https://github.com/dbt-labs/dbt-core/blob/dd77210756e6432ef2737ab59a65c238e3abfdda/core/dbt/compilation.py#L197) function is called during the `dbt build` command, and inserts edges into the execution graph which are meant to ensure that models downstream from a node will not run until all the tests on that node have passed.

The function is slow in certain projects, and recent data from the field show that it inflates the number of edges in the graph by a factor of six. It is slow enough that it often shows up in performance profiles, but is even more problematic in terms of memory consumption, as memory use is high enough to cause OOM crashes.

### Acceptance criteria

1. If possible, implement a new version of this function which adds edges to achieve the desired test-dependency behavior but inserts fewer edges and runs more quickly.
2. Add a new behavior flag which causes the new function to be used, while retaining the old function on the default code path.
3. Follow up by gathering data about the relative performance of the two implementations and monitoring for regressions.

### Suggested Tests

Existing tests should suffice, but we should add additional tests to reduce the risks associated with the new implementation.

### Impact to Other Teams

None.

No.

### Context

_No response_

peterallenwebb:
Resolves #10950 

### Checklist

### DIFF
--- a/core/dbt/compilation.py
+++ b/core/dbt/compilation.py
@@ -1,8 +1,9 @@
+import dataclasses
 import json
 import os
 import pickle
-from collections import defaultdict
-from typing import Any, Dict, List, Optional, Tuple
+from collections import defaultdict, deque
+from typing import Any, Dict, Iterable, List, Optional, Set, Tuple
 
 import networkx as nx  # type: ignore
 import sqlparse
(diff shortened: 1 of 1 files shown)


gshank (approved):
Yay! Looks great. Good comments. Nice optimizations.

PR #11108, opened 2024-12-07
Use xdg

### LINKED ISSUE #2515: On Linux: respect XDG_CONFIG_HOME and don't store config in user's home directory
### Describe the feature
The XDG Base Directory standard states that user config files should respect the `XDG_CONFIG_HOME` environment variable, which defaults `$HOME/.config`.

Currently, dbt defaults the PROFILE_DIR to `~/.dbt` and doesn't attempt to read from XDG_CONFIG_HOME or ~/.config.
```
mv ~/.dbt ~/.config/dbt
dbt debug
...
profiles.yml file [ERROR not found]
```

I can locally symlink my ~/.config/dbt directory to ~/.dbt, but it's not my preference to put additional directories in my home directory.  Alternately, I could set a global `DBT_PROFILES_DIR` variable.

### Additional context
This is only relevant to Linux users.  The [`appdirs` pypi package](https://pypi.org/project/appdirs/) makes the XDG_Base_directory standard simple to implement.

### Who will this benefit?
Linux users who have far too many files/folders in their home directories.

dradetsky:
Resolves #2515 

### Problem

We hardcode the profile dir to `~/.dbt` instead of using xdg.

On a related note, there was a circular dep issue that caused us to have to copypaste `default_profiles_dir()` instead of just importing it (which is annoying if we want to add more complexity to it), so we have to fix that.

### Solution

Import & use appdirs.

### Considerations

Per #2515, this has different effects on different platforms. I'm only thinking about it in terms of Linux, but maybe we don't want to use this on win/macos. Somebody else more familiar with those should comment.

### Checklist

### DIFF
--- a/.changes/unreleased/Fixes-20241207-155958.yaml
+++ b/.changes/unreleased/Fixes-20241207-155958.yaml
@@ -1,0 +1,6 @@
+kind: Fixes
+body: Use xdg and also fix circular dep
+time: 2024-12-07T15:59:58.137140008-08:00
+custom:
+    Author: dradetsky
+    Issue: "2515"
--- a/core/dbt/cli/__init__.py
+++ b/core/dbt/cli/__init__.py
@@ -1,1 +1,0 @@
-from .main import cli as dbt_cli  # noqa
--- a/core/dbt/cli/resolvers.py
+++ b/core/dbt/cli/resolvers.py
@@ -16,7 +16,7 @@ def default_profiles_dir() -> Path:
     if (Path.cwd() / "profiles.yml").exists():
         return Path.cwd()
     else:
-        return user_config_dir() / Path("dbt")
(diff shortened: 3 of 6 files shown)


dradetsky:
Fun Fact: pre-commit didn't work for me locally; most things were fine, but flake8 broke with

`AttributeError: 'EntryPoints' object has no attribute 'get'`

Looking into this error, the flake8 team just replies "use the new version". The project uses 4.0.1, and flake8 is on 7.1.1. It might be time to upgrade.

Additional Consideration: do we need to auto-migrate configs from ~/.dbt to ~/.config/dbt?

dradetsky:
@cla-bot check

dbeatty10:
@dradetsky re-assessing this as a feature request beginning here: https://github.com/dbt-labs/dbt-core/issues/2515#issuecomment-2536301326

PR #11113, opened 2024-12-09
Fix erroneous additional batch execution

### LINKED ISSUE #11112: [Bug] Microbatch models with only one batch raise unhandled error

### Current Behavior

Currently on 1.9.latest if you run a microbatch model with only one batch (can be achieved by setting `lookback=0` for the model), an unhandled exception gets raised (but doesn't actually break anything?)

### Expected Behavior

There shouldn't be an unhandled exception for simply having a microbatch model with `lookback=0`

### Steps To Reproduce

1. Ensure you're running on dbt-core 1.9.latest via `pip install -e <path>/<to>/<core>/<repo>/dbt-core/core`
    * repository must be checked out on `1.9.latest` branch
3. Create a dbt project with a microbatch incremental model (with `lookback=0` in its config)
4. Run the project via `dbt run`
5. Observe the log weirdness

### Relevant log output

```shell
(venv) quigleymalcolm@Quigley-Malcolm microbatch_demo % dbt run                               
18:10:03  Running with dbt=1.10.0-a1
18:10:03  Registered adapter: postgres=1.9.0-rc1
18:10:03  Found 3 models, 429 macros
18:10:03  
18:10:03  Concurrency: 1 threads (target='dev')
18:10:03  
18:10:03  1 of 3 START sql table model microbatch_demo.input_model ....................... [RUN]
18:10:03  1 of 3 OK created sql table model microbatch_demo.input_model .................. [SELECT 4 in 0.08s]
18:10:03  2 of 3 START sql microbatch model microbatch_demo.microbatch_model ............. [RUN]
18:10:03  Batch 1 of 1 START batch 2024-12-09 of microbatch_demo.microbatch_model .............. [RUN]
18:10:03  Batch 1 of 1 OK created batch 2024-12-09 of microbatch_demo.microbatch_model ......... [MERGE 0 in 0.09s]
18:10:03  Unhandled error while executing 
1
18:10:03  2 of 3 OK created sql microbatch model microbatch_demo.microbatch_model ........ [SUCCESS in 0.12s]
18:10:03  3 of 3 START sql view model microbatch_demo.dependent_model .................... [RUN]
18:10:03  3 of 3 OK created sql view model microbatch_demo.dependent_model ............... [CREATE VIEW in 0.05s]
18:10:03  
18:10:03  Finished running 1 incremental model, 1 table model, 1 view model in 0 hours 0 minutes and 0.39 seconds (0.39s).
18:10:03  
18:10:03  Completed successfully
18:10:03  
18:10:03  Done. PASS=3 WARN=0 ERROR=0 SKIP=0 TOTAL=3
```
```


### Environment

```markdown
- OS: macOS 15.1
- Python: 3.10.15
- dbt branch: 1.9.latest (and also main)
```

postgres

### Additional Context

Screenshot 2024-12-09 at 12 43 50

QMalcolm:
Resolves #11112

### Problem

Microbatch models that only ran on batch (either due to `lookback=0` or setting `--event-time-start` + `--event-time-end`) would raise an `Unhandled exception` like the following:
Screenshot 2024-12-09 at 12 43 50

### Solution

Ensure we skip the last batch execution path when only one batch needs to be run in total for a microbatch model
Screenshot 2024-12-09 at 13 50 14

### Checklist

### DIFF
--- a/.changes/unreleased/Fixes-20241209-133317.yaml
+++ b/.changes/unreleased/Fixes-20241209-133317.yaml
@@ -1,0 +1,6 @@
+kind: Fixes
+body: Fix running of extra "last" batch when there is only one batch
+time: 2024-12-09T13:33:17.253326-06:00
+custom:
+  Author: QMalcolm
+  Issue: "11112"
--- a/core/dbt/task/run.py
+++ b/core/dbt/task/run.py
@@ -768,18 +768,21 @@ class RunTask(CompileTask):
         # Wait until all submitted batches have completed
         while len(batch_results) != batch_idx:
             pass
-        # Final batch runs once all others complete to ensure post_hook runs at the end
-        self._submit_batch(
-            node=node,
-            adapter=runner.adapter,
-            relation_exists=relation_exists,
-            batches=batches,
-            batch_idx=batch_idx,
-            batch_results=batch_results,
-            pool=pool,
-            force_sequential_run=True,
-            skip=skip_batches,
-        )
+
+        # Only run "last" batch if there is more than one batch
+        if len(batches) != 1:
+            # Final batch runs once all others complete to ensure post_hook runs at the end
+            self._submit_batch(
+                node=node,
+                adapter=runner.adapter,
+                relation_exists=relation_exists,
+                batches=batches,
+                batch_idx=batch_idx,
+                batch_results=batch_results,
+                pool=pool,
+                force_sequential_run=True,
+                skip=skip_batches,
+            )
 
         # Finalize run: merge results, track model run, and print final result line
         runner.merge_batch_results(result, batch_results)
--- a/tests/functional/microbatch/test_microbatch.py
+++ b/tests/functional/microbatch/test_microbatch.py
@@ -1057,11 +1057,20 @@ class TestWhenOnlyOneBatchRunBothPostAndPreHooks(BaseMicrobatchTest):
 
         return EventCatcher(event_to_catch=JinjaLogDebug, predicate=pre_or_post_hook)  # type: ignore
 
+    @pytest.fixture
+    def generic_exception_catcher(self) -> EventCatcher:
+        return EventCatcher(event_to_catch=GenericExceptionOnRun)  # type: ignore
+
     def test_microbatch(
-        self, mocker: MockerFixture, project, batch_log_catcher: EventCatcher
+        self,
+        project,
+        batch_log_catcher: EventCatcher,
+        generic_exception_catcher: EventCatcher,
     ) -> None:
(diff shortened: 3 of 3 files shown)


ChenyuLInx (commented):
The change looks mostly good!

From organizing code perspective what's your thoughts on hide all logic here in the `_submit_batch` function? It seems like we also have some special logic to handle the last batch in that function. I think it's good to have all these in a single logic layer vs multiple. WDYT?  https://github.com/dbt-labs/dbt-core/blob/5942aa59c29f0b5d9cf47e3da3a653460be94afd/core/dbt/task/run.py#L741-L785

QMalcolm:
> The change looks mostly good!
> 
> From organizing code perspective what's your thoughts on hide all logic here in the `_submit_batch` function? It seems like we also have some special logic to handle the last batch in that function. I think it's good to have all these in a single logic layer vs multiple. WDYT?
> 
> https://github.com/dbt-labs/dbt-core/blob/5942aa59c29f0b5d9cf47e3da3a653460be94afd/core/dbt/task/run.py#L741-L785

@ChenyuLInx I'm down to refactor this code 🙂 I'd definitely like to simplify things here a fair bit. However, this change is going to need to be backported to 1.9.latest for 1.9.1. As such I think a refactor here is out of scope, as such work should probably be forward facing.

ChenyuLInx:
My preference is do the refactor as you change the code if it is not too complex
> going to need to be backported to 1.9.latest

Why this changes whether we should do the refactor?

QMalcolm:
> My preference is do the refactor as you change the code if it is not too complex
> 
> > going to need to be backported to 1.9.latest
> 
> Why this changes whether we should do the refactor?

My preference is to keep refactors as separate as possible from changes in logic. I've found it's desirable to not do a refactor as part of work which is being backported. This is because:
1. The refactor is irrelevant to the fix, and thus doesn't have a need to be backported
2. The more changes we introduce, the greater the likelihood of introducing a new edge case or bug

Now, I'm generally of the opinion that refactors should be separated from logical changes or bug fixes, unless it is impossible to separate the two. I'm a little looser on that when it is a forward-looking change only (i.e. solely going into an alpha/beta). With backports though, I take it a bit more seriously 

If the worry is that the refactor won't get done unless it comes alongside the bug fix / logic change, then I'll promise to do the refactor as the next PR I open 🙂

ChenyuLInx:
I am happy this goes in as is with a fast follow-up PR of the refactor.

One thing to note is If you don't do the refactor as part of this change and get it backported, when we have to have some other fixes around this go in, you will have to backport the refactor and then the fix(or do a custom fix).

ChenyuLInx (approved):

PR #11137, opened 2024-12-11
Fix circular import

### LINKED ISSUE #11142: [Tech debt] Remove duplicated definition of `default_profiles_dir` (without adding circular dependency)

### Describe the feature

The following code duplicated the definition of [`default_profiles_dir`](https://github.com/dbt-labs/dbt-core/blob/7df04b0fe461dcad3c2abcb67004178efaf445ce/core/dbt/cli/resolvers.py#L13-L14) within https://github.com/dbt-labs/dbt-core/pull/6788 in order to avoid a circular dependency issue.

https://github.com/dbt-labs/dbt-core/blob/7df04b0fe461dcad3c2abcb67004178efaf445ce/core/dbt/config/profile.py#L167-L174

We'd like to remove this duplication while still avoiding any circular dependencies.

_No response_

### Who will this benefit?

_No response_

See proposed implementation in https://github.com/dbt-labs/dbt-core/pull/11137

### Anything else?

_No response_

dradetsky:
resolves #11142

### Problem

There was a circular import which required duplicating a method in `core/dbt/config/profile.py`. This was both ugly and an impediment to improving that method if required (as in #11108).

### Solution

Fix the circular import and remove the duplication

### Checklist

### DIFF
--- a/.changes/unreleased/Fixes-20241211-132203.yaml
+++ b/.changes/unreleased/Fixes-20241211-132203.yaml
@@ -3,4 +3,4 @@ body: Fix circular dependency
 time: 2024-12-11T13:22:03.637443979-08:00
 custom:
     Author: dradetsky
-    Issue: none
+    Issue: 11142
--- a/.changes/unreleased/Fixes-20241211-132203.yaml
+++ b/.changes/unreleased/Fixes-20241211-132203.yaml
@@ -1,0 +1,6 @@
+kind: Fixes
+body: Fix circular dependency
+time: 2024-12-11T13:22:03.637443979-08:00
+custom:
+    Author: dradetsky
+    Issue: none
--- a/core/dbt/cli/__init__.py
+++ b/core/dbt/cli/__init__.py
@@ -1,1 +1,0 @@
-from .main import cli as dbt_cli  # noqa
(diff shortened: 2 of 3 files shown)


.changes/unreleased/Fixes-20241211-132203.yaml
```diff
@@ -0,0 +1,6 @@
+kind: Fixes
+body: Fix circular dependency
+time: 2024-12-11T13:22:03.637443979-08:00
+custom:
+    Author: dradetsky
+    Issue: none
```
dbeatty10, line 6:
```suggestion
    Issue: 11142
```

dradetsky:
wow, I didn't know I fixed an issue! good job me!

dradetsky:
The one check CI is uhappy with is code coverage, here: https://github.com/dbt-labs/dbt-core/pull/11137/checks?check_run_id=34330125418

> 86.17% (-0.12%) compared to 6621015

So probably what's happened here is that we've added 1 line of imports (which is not considered covered even if it is actually exercised by tests), and then removed ~10 lines of bugs & duplication. These removed lines _were_ covered, and thus coverage has gone down (it's also possible that we're accidentally counting changelog lines as uncovered code, but this is less likely).

Anyway, the moral of the story is that code coverage is not an ultra-precise measure of code quality. We could add a few lines of no-ops to an otherwise-covered function to "fix" this, but I don't recommend this.

ChenyuLInx (approved):
This looks good to me, thanks for adding this @dradetsky !!

PR #11138, opened 2024-12-12
add allow additional property for Model and SourceDefinition

ChenyuLInx:
Allowing additional properties on Model and SourceDefinition so we are more forgiving at reading artifacts.
This change doesn't change user contract. User still cannot add additional fields to project freely

### Checklist

### DIFF
--- a/core/dbt/artifacts/schemas/manifest/v12/manifest.py
+++ b/core/dbt/artifacts/schemas/manifest/v12/manifest.py
@@ -180,3 +180,13 @@ class WritableManifest(ArtifactMixin):
         if manifest_schema_version < cls.dbt_schema_version.version:
             data = upgrade_manifest_json(data, manifest_schema_version)
         return cls.from_dict(data)
+
+    @classmethod
+    def validate(cls, _):
+        # When dbt try to load an artifact with additional optional fields
+        # that are not present in the schema, from_dict will work fine.
+        # As long as validate is not called, the schema will not be enforced.
+        # This is intentional, as it allows for safer schema upgrades.
+        raise RuntimeError(
+            "The WritableManifest should never be validated directly to allow for schema upgrades."
+        )
--- a/core/dbt/artifacts/schemas/manifest/v12/manifest.py
+++ b/core/dbt/artifacts/schemas/manifest/v12/manifest.py
@@ -28,6 +28,7 @@ from dbt.artifacts.schemas.base import (
     schema_version,
 )
 from dbt.artifacts.schemas.upgrades import upgrade_manifest_json
+from dbt_common.exceptions import DbtInternalError
 
 NodeEdgeMap = Dict[str, List[str]]
 UniqueID = str
@@ -187,6 +188,6 @@ class WritableManifest(ArtifactMixin):
         # that are not present in the schema, from_dict will work fine.
         # As long as validate is not called, the schema will not be enforced.
         # This is intentional, as it allows for safer schema upgrades.
-        raise RuntimeError(
+        raise DbtInternalError(
             "The WritableManifest should never be validated directly to allow for schema upgrades."
         )
--- a/schemas/dbt/manifest/v12.json
+++ b/schemas/dbt/manifest/v12.json
@@ -13,7 +13,7 @@
         },
         "dbt_version": {
           "type": "string",
-          "default": "1.9.0b4"
+          "default": "1.10.0a1"
         },
         "generated_at": {
           "type": "string"
@@ -9121,19 +9121,7 @@
                         "type": "integer"
                       },
                       "granularity": {
-                        "enum": [
-                          "nanosecond",
-                          "microsecond",
-                          "millisecond",
-                          "second",
-                          "minute",
-                          "hour",
-                          "day",
-                          "week",
-                          "month",
-                          "quarter",
-                          "year"
-                        ]
+                        "type": "string"
                       }
                     },
                     "additionalProperties": false,
@@ -18712,19 +18700,7 @@
                                   "type": "integer"
                                 },
                                 "granularity": {
-                                  "enum": [
-                                    "nanosecond",
-                                    "microsecond",
-                                    "millisecond",
-                                    "second",
-                                    "minute",
-                                    "hour",
-                                    "day",
-                                    "week",
-                                    "month",
-                                    "quarter",
-                                    "year"
-                                  ]
+                                  "type": "string"
                                 }
                               },
                               "additionalProperties": false,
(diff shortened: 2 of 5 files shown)


schemas/dbt/manifest/v12.json
```diff
@@ -9121,19 +9133,7 @@
                         "type": "integer"
                       },
                       "granularity": {
-                        "enum": [
```
ChenyuLInx, line 9124: Not sure why this gets changed

tests/functional/artifacts/expected_manifest.py
```diff
@@ -372,6 +372,7 @@ def expected_seeded_manifest(project, model_database=None, quote_model=False):
                 "version": None,
                 "latest_version": None,
                 "time_spine": None,
+                "batch": None,
```
ChenyuLInx, line 375: I am not sure why I would have to add this field in expected manifest

tests/functional/artifacts/test_artifacts.py
```diff
@@ -503,6 +503,8 @@ def verify_manifest(project, expected_manifest, start_time, manifest_schema_path
         elif key in ["nodes", "sources", "exposures", "metrics", "disabled", "docs"]:
             for unique_id, node in expected_manifest[key].items():
                 assert unique_id in manifest[key]
+                if "_event_status" in manifest[key][unique_id]:
```
ChenyuLInx, line 506: Not sure why `_event_status` starts to pop up

schemas/dbt/manifest/v12.json
ChenyuLInx: I checked out main branch and generated a schema, these changes are also there, so it is not particular to this PR

tests/unit/contracts/graph/test_manifest.py
```diff
@@ -96,7 +96,6 @@
         "deprecation_date",
         "defer_relation",
         "time_spine",
-        "batch",
```
ChenyuLInx, line 99: @QMalcolm This is because we deleted batch when writing.

gshank (approved):

core/dbt/contracts/graph/nodes.py
```diff
@@ -475,6 +475,8 @@ def __post_serialize__(self, dct: Dict, context: Optional[Dict] = None):
             del dct["_has_this"]
         if "previous_batch_results" in dct:
             del dct["previous_batch_results"]
+        if "batch" in dct:
```
ChenyuLInx, line 478: We actually can't do this given `batch` is needed for jinja context

core/dbt/artifacts/resources/v1/model.py
```diff
@@ -51,4 +52,10 @@ def __post_serialize__(self, dct: Dict, context: Optional[Dict] = None):
         dct = super().__post_serialize__(dct, context)
         if context and context.get("artifact") and "defer_relation" in dct:
             del dct["defer_relation"]
+
+        # delete extra keys that should't be serialized
```
ChenyuLInx, line 56: @gshank @QMalcolm I figured out the reason `batch` suddenly started to show up is because we allow extra on Model, so when we convert ModelNode to Model when writing an artifact, it will be kept as extra and convert into keys in dct.

QMalcolm: Ohhhhhhhhh that makes sense. Thank you for figuring this out and removing it with the below process 👍

core/dbt/contracts/graph/manifest.py
```diff
@@ -936,6 +938,31 @@ class Manifest(MacroMethods, dbtClassMixin):
...
+    def validate(cls, data: Any) -> None:
+        json_schema = cls.json_schema()
+        cls._allow_additional_recursive(cls, json_schema)
+        validator = jsonschema.Draft7Validator(json_schema)
+        error = next(iter(validator.iter_errors(data)), None)
+        if error is not None:
+            raise ValidationError.create_from(error) from error
+
+    @staticmethod
+    def _allow_additional_recursive(cls, dct: Dict[str, Any]):
```
ChenyuLInx, line 953: I prefer a smaller interface to start with and if we need to make it generic it is pretty easy and we know more what it should look like at the time

core/dbt/contracts/graph/manifest.py
```diff
@@ -936,6 +938,31 @@ class Manifest(MacroMethods, dbtClassMixin):
         metadata={"serialize": lambda x: None, "deserialize": lambda x: None},
     )
 
+    # This and the following methods can be refacong into a mixin
```
MichelleArk, line 941: nit: typo in comment on "refactoring"

core/dbt/contracts/graph/manifest.py
```diff
@@ -936,6 +938,31 @@ class Manifest(MacroMethods, dbtClassMixin):
...
+    # and we should do so next time we do it for another artifact.
+    @classmethod
+    def validate(cls, data: Any) -> None:
```
MichelleArk, line 944: I think this should be on the `WritableManifest` class in dbt/artifacts, not the `Manifest` internal object, since it is defining behaviour related to ser/de which dbt/artifacts should be responsible for. This will also make it easier to extend the pattern to other artifact schemas (e.g. making a `ExtensibleArtifactMixin`, or just making `ArtifactMixin` have this `validate` implementation)

tests/unit/contracts/graph/test_nodes.py
```diff
@@ -236,10 +236,11 @@ def test_basic_compiled_model(basic_compiled_dict, basic_compiled_model):
...
 
-def test_invalid_extra_fields_model(minimal_uncompiled_dict):
-    bad_extra = minimal_uncompiled_dict
-    bad_extra["notvalid"] = "nope"
-    assert_fails_validation(bad_extra, ModelNode)
+def test_extra_fields_model_okay(minimal_uncompiled_dict):
+    extra = minimal_uncompiled_dict
+    extra["notvalid"] = "nope"
+    # Model still load fine with extra fields
+    ModelNode.from_dict(extra)
```
MichelleArk, line 243: might be overkill, but we could also test that the ModelNode does not have the extra attribute on the class, or any `_extra` property.

MichelleArk: I guess this test may not be necessary with the new `validate` implementation though...

core/dbt/artifacts/schemas/manifest/v12/manifest.py
```diff
@@ -180,3 +180,13 @@ def upgrade_schema_version(cls, data):
...
             data = upgrade_manifest_json(data, manifest_schema_version)
         return cls.from_dict(data)
+
+    @classmethod
+    def validate(cls, _):
+        # When dbt try to load an artifact with additional optional fields
+        # that are not present in the schema, from_dict will work fine.
+        # As long as validate is not called, the schema will not be enforced.
+        # This is intentional, as it allows for safer schema upgrades.
+        raise RuntimeError(
```
MichelleArk, line 190: This should probably be a dbtInternalError

MichelleArk (approved):
Small comment, otherwise LGTM! Thank you for adjusting the v12.json for future CI runs 👍

PR #11145, opened 2024-12-12
Warn if `concurrent_batches` config is set to `True`, but the available adapter doesn't support it

### LINKED ISSUE #11122: Emit warning if node-level `batch_concurrency` is set to `True` while adapter does not support MicrobatchConcurrency
Ideally this is one warning per project (attributed to misconfigured models), not per model

QMalcolm:
Resolves #11122

### Problem

Batches of microbatch models can be executed sequentially or concurrently. We try to figure out which to do intelligently. As part of that, we implemented an override, the model config `concurrent_batches`, to allow the user to bypass _some_ of our automatic detection. However, a user _cannot_ for batches to run concurrently if the adapter doesn't support concurrent batches (declaring support is opt in). Thus, if an adapter _doesn't_ support running batches concurrently, and a user tries to force concurrent execution via `concurrent_batches`, then we need to warn the user that that isn't happening.

### Solution

After parse/load of manifest, check for whether the adapter supports concurrent batch execution _and_ whether there are any microbatch models setting `concurrent_batches=True`. If the adapter _doesn't_ support concurrent batches and if there are _any_ models with `concurrent_batches=True`, fire a warning.

Example:
Screenshot 2024-12-12 at 14 39 35

### Checklist


tests/unit/parser/test_manifest.py
```diff
@@ -238,3 +241,59 @@ def test_warn_for_unused_resource_config_paths(
...
+    @patch("dbt.parser.manifest.ManifestLoader.build_manifest_state_check")
+    @patch("dbt.parser.manifest.os.path.exists")
+    @patch("dbt.parser.manifest.open")
+    def manifest_loader(
+        self, patched_open, patched_os_exist, patched_state_check
+    ) -> ManifestLoader:
+        mock_project = MagicMock(RuntimeConfig)
+        mock_project.project_target_path = "mock_target_path"
+        mock_project.project_name = "mock_project_name"
+        # patched_os_exist.return_value = True
```
MichelleArk, line 257: nit: this can probably be cleaned up

MichelleArk (approved):
small nit, otherwise LGTM!

PR #11161, opened 2024-12-17
Check modified contracts when doing state:modified

### LINKED ISSUE #11034: [Bug] `state:modified` should include `check_modified_contract`

### Current Behavior

Following the implementation in:
- https://github.com/dbt-labs/dbt-core/pull/10221

If I delete/rename a contracted model, I expect to see a warning/error when I `dbt <command> --select state:modified`.

But I only see this warning/error with `dbt <command> --select state:modified.contracts`

### Expected Behavior

`state:modified` should include `check_modified_contract`: https://github.com/dbt-labs/dbt-core/commit/14453eb225654f023a559e556ed4cd12c3aab039

@T-Dunlap:
>  It's going to be really difficult to educate customers that they have to explicitly call out `state:modified.contract` in their CI jobs. It's best if this is baked directly into `state:modified`

@joellabes:
> my understanding was that state:modified contained everything the more granular state:modified.* methods offered (and then some) and would not at all expect that there was something only available with a submethod

### Steps To Reproduce

1. I deleted the model and all references to the model in the project.
2. I deleted the contract, which contained two versions
3. My downstream project was pointed to v1, which was technically deprecated.
4. No downstream models were pointed to v2. 
5. No error or warning.

### Relevant log output

_No response_

### Environment

```markdown
dbt: 1.9 / Versionless
```

_No response_

### Additional Context

**Would this require a behavior-change flag?** I don't think so, because it feels in keeping with the original intent of the issue, with [what we've documented](https://docs.getdbt.com/reference/resource-properties/versions#detecting-breaking-changes), and with what users would expect to happen

It also generally crops up in CI jobs, so unlikely to suddenly break someone's prod runs unless they're actually proposing code changes. (I think they'd need to be doing something wacky with dynamically enabling/disabling contracted versioned models)

gshank:
Resolves #11034

### Problem

Checking modified:contract needed to be run separately, which users found unexpected.

### Solution

Call check_modified_contract when running check_modified.

### Checklist

### DIFF
--- a/.changes/unreleased/Fixes-20241217-154848.yaml
+++ b/.changes/unreleased/Fixes-20241217-154848.yaml
@@ -1,0 +1,6 @@
+kind: Fixes
+body: Run check_modified_contract for state:modified
+time: 2024-12-17T15:48:48.053054-05:00
+custom:
+  Author: gshank
+  Issue: "11034"
--- a/core/dbt/graph/selector_methods.py
+++ b/core/dbt/graph/selector_methods.py
@@ -680,17 +680,24 @@ class StateSelectorMethod(SelectorMethod):
     def check_modified_content(
         self, old: Optional[SelectorTarget], new: SelectorTarget, adapter_type: str
     ) -> bool:
+        different_contents = False
         if isinstance(
             new,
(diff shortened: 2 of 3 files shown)


ChenyuLInx (approved):
Thanks!

PR #11162, opened 2024-12-17
Parse `catalogs.yml`

MichelleArk:
Resolves [XPLAT-242](https://dbtlabs.atlassian.net/browse/XPLAT-242)

### Problem

We need to parse `catalogs.yml` to enable write integrations for X-Plat.

Spec is here (dbt Internal): https://www.notion.so/dbtlabs/RFC-94-Cross-Platform-Mesh-142bb38ebda780b4befdee9b7e09902b?pvs=4#150bb38ebda780e49bafd6fbd0a1651d

### Solution

- [x] Add `@requires.catalogs` to run and build commands.
- [x] Add tests for write integrations.
- [x] A lot of nitty-gritty tests (e.g. validating `table_format` will be handled by the Adapter directly).

### Checklist

[XPLAT-242]: https://dbtlabs.atlassian.net/browse/XPLAT-242?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


aranke:
Order of operations:
- [ ] dbt-labs/dbt-adapters#790
- [ ] #11162 
- [ ] dbt-labs/dbt-adapters#792
- [ ] dbt-labs/dbt-adapters#810

colin-rogers-dbt (approved):

aranke (dismissed):
Can I approve my own PR?

core/dbt/cli/requires.py
```diff
@@ -346,15 +370,22 @@ def setup_manifest(ctx: Context, write: bool = True, write_perf_info: bool = Fal
 
     runtime_config = ctx.obj["runtime_config"]
 
+    catalogs = ctx.obj["catalogs"] if "catalogs" in ctx.obj else []
+    active_write_integrations = [get_active_write_integration(catalog) for catalog in catalogs]
```
mikealfare, line 374: Shouldn't there only be one (or none) active write integrations?

core/dbt/config/catalogs.py
```diff
@@ -0,0 +1,88 @@
...
+                raise DbtValidationError(f"The file at {path} is empty")
+
+            return yaml_content
+        except DbtValidationError as e:
+            raise YamlLoadError(project_name=project_name, path=CATALOGS_FILE_NAME, exc=e)
+
+    return {}
+
+
+def load_single_catalog(raw_catalog: Dict[str, Any], renderer: SecretRenderer) -> Catalog:
```
mikealfare, line 31: Will this handle getting the built-in/default integration for Snowflake? We need to add an integration for the user to be backwards compatible with the existing Iceberg implementation. Even without solving for backwards compatibility, we should add the default catalog for the user. This means the adapter will add an integration that `dbt-core` will not know about because it's not in the user's config.

aranke: I think so? 
In the tests, `dbt-core` has no idea about `TestCatalogIntegration`; it is coming purely from the adapter.

core/dbt/cli/requires.py
aranke: Per catalog, yes. But there may be more than 1 catalog like so:

```yaml
catalogs:
  - name: catalog1
    ...
  - name: catalog2
    ...
  ...
```

mikealfare (approved):

tests/functional/catalogs/test_catalogs_parsing.py
```diff
@@ -0,0 +1,143 @@
...
+
+class TestSingleWriteIntegration(BaseCatalogTest):
+    @pytest.fixture
+    def integrations(self):
+        return [FIRST_INTEGRATION]
+
+    def test_catalog_parsing(self, catalogs, project):
+        with self.setup_test(catalogs, project):
+            run_dbt(["run"])
+            write_integration = project.adapter.get_catalog_integration("my_write_integration")
```
mikealfare, line 75: This needs to be updated with the `catalog_name` updates.
```suggestion
            write_integration = project.adapter.get_catalog_integration("test_catalog")
```

tests/functional/catalogs/test_catalogs_parsing.py
```diff
@@ -0,0 +1,143 @@
...
+            assert isinstance(write_integration, TestCatalogIntegration)
+            assert write_integration.name == "my_write_integration"
```
mikealfare, line 77: Same as above:
```suggestion
            assert write_integration.name == "test_catalog"
            assert write_integration.catalog_name == "my_write_integration"
```

tests/functional/catalogs/test_catalogs_parsing.py
```diff
@@ -0,0 +1,143 @@
...
+
+class TestValidActiveWriteIntegration(TestMultipleWriteIntegrations):
+    @pytest.fixture
+    def active_integration(self):
+        return "my_second_write_integration"
+
+    def test_catalog_parsing_adapter_initialialization(self, catalogs, project):
+        with self.setup_test(catalogs, project):
+            run_dbt(["build"])
+            with pytest.raises(DbtCatalogIntegrationNotFoundError):
```
mikealfare, line 98: This will actually pass now since both integrations are in the same catalog and the catalog is the reference.

tests/functional/catalogs/test_catalogs_parsing.py
```diff
@@ -0,0 +1,143 @@
...
+                project.adapter.get_catalog_integration("my_write_integration")
+
+            write_integration = project.adapter.get_catalog_integration(
+                "my_second_write_integration"
```
mikealfare, line 102:
```suggestion
                "test_catalog"
```

tests/functional/catalogs/test_catalogs_parsing.py
```diff
@@ -0,0 +1,143 @@
...
+            )
+            assert isinstance(write_integration, TestCatalogIntegration)
+            assert write_integration.name == "my_second_write_integration"
```
mikealfare, line 105:
```suggestion
            assert write_integration.name == "test_catalog"
            assert write_integration.catalog_name == "my_second_write_integration"
```

core/dbt/config/catalogs.py
```diff
@@ -0,0 +1,88 @@
...
+    raw_catalogs = read_catalogs(project_dir, project_name).get("catalogs", [])
+    catalogs_renderer = SecretRenderer(cli_vars)
+
+    return [load_single_catalog(raw_catalog, catalogs_renderer) for raw_catalog in raw_catalogs]
+
+
+def get_active_write_integration(catalog: Catalog) -> Optional[CoreCatalogIntegrationConfig]:
+    for integration in catalog.write_integrations:
+        if integration.name == catalog.active_write_integration:
+            return integration
```
mikealfare, line 86: We need to update the existing field `name` on `integration` to `catalog_name` and flatten the catalog `name` onto `integration` (as `name`) before returning here. While there can be multiple catalog integrations with the same catalog name in `dbt-core`, only the active one gets sent to the adapter, and it gets referenced by the catalog name, not the integration name. The integration name aligns with the name of the integration in the platform.

```suggestion
            active_integration = deepcopy(integration)
            active_integration.catalog_name = active_integration.name
            active_integration.name = catalog.name
            return active_integration
```

core/dbt/artifacts/resources/v1/catalog.py
```diff
@@ -0,0 +1,21 @@
+from dataclasses import dataclass, field
+from typing import Any, Dict, List, Optional
+
+from dbt.adapters.catalogs import CatalogIntegrationConfig
+from dbt_common.dataclass_schema import dbtClassMixin
+
+
+@dataclass
+class CoreCatalogIntegrationConfig(CatalogIntegrationConfig):
+    name: str
```
mikealfare, line 10: This protocol now has a `catalog_name` in addition to `name`. `name` should be the name of the catalog and `catalog_name` should be the name of the catalog integration. In other words, I think `name` here and `name` on `Catalog` below are the same thing.

```suggestion
    name: str
    catalog_name: str
```

mikealfare (commented):
I tried to address the various updates required to align with those in `dbt-adapters`. I left code suggestions to try and make it less tedious, but feel free to toss them if you have a better way.

mikealfare (approved):

PR #11177, opened 2024-12-24
Make MAXIMUM_SEED_SIZE_MIB configurable

### LINKED ISSUE #7117: [Feature] Make MAXIMUM_SEED_SIZE configurable

### LINKED ISSUE #7124: [Feature] Compute seed file hashes incrementally

jtcohen6:
resolves https://github.com/dbt-labs/dbt-core/issues/7117
resolves https://github.com/dbt-labs/dbt-core/issues/7124

Reapply changes from #7125. (This proved easier than rebasing the commits directly.)

### Problem

We apply an arbitrary limit of 1 MiB to seeds (CSVs), for the specific purpose of hashing contents and comparing those hashed contents during `state:modified` comparison. (That's a _mebibyte_, not a megabyte, for those who care to distinguish between the two).

That is, dbt doesn't raise an error if it detects a seed larger than `MAXIMUM_SEED_SIZE_MIB`, but certain features become unavailable and dbt does not make any guarantees about acceptable performance.

We could adjust this for inflation (1 MiB in 2020 is worth ~1.2 MiB today), but this PR takes the more forward-looking approach of making the value configurable by end users. Users can instruct dbt to compare the contents of larger seeds, so long as they're willing to "pay the price" of hashing the contents of large seeds.

We will need to update docs: https://docs.getdbt.com/reference/node-selection/state-comparison-caveats

### Solution

From #7125:

> Increasing the maximum size of seed files where the content is hashed for state comparison will enable a greater use of deferred runs with updated seed file contents. By making this a configuration with an environment variable users are able to override the default 1 MiB limit.

> Furthermore, to support reading larger files in memory constrained environments, a new method is added to read and compute file contents incrementally. As this is performed on bytes for better performance we continue using the previous UTF-8 content method for small files to not mess up current states where seed files are not stored with UTF-8.

### Checklist


core/dbt/parser/read_files.py
```diff
@@ -123,12 +124,14 @@ def validate_yaml(file_path, dct):
...
+    # Users can configure the maximum seed size (MiB) that will be hashed for state comparison
+    maximum_seed_size = get_flags().MAXIMUM_SEED_SIZE_MIB * 1024 * 1024
+    # maximum_seed_size = 0 means no limit
+    if match.file_size() > maximum_seed_size and maximum_seed_size != 0:
         # We don't want to calculate a hash of this file. Use the path.
         source_file = SourceFile.big_seed(match)
     else:
-        file_contents = load_file_contents(match.absolute_path, strip=True)
-        checksum = FileHash.from_contents(file_contents)
+        checksum = FileHash.from_path(match.absolute_path)
```
jtcohen6, line 134: I have confirmed that this is not a "breaking" change, insofar as the same seed produces the same checksum before and after this change.

core/dbt/artifacts/resources/base.py
```diff
@@ -60,6 +61,27 @@ def from_contents(cls, contents: str, name="sha256") -> "FileHash":
         checksum = hashlib.new(name, data).hexdigest()
         return cls(name=name, checksum=checksum)
 
+    @classmethod
+    def from_path(cls, path: str, name="sha256") -> "FileHash":
```
jtcohen6, line 65: Non-breaking change in the `dbt/artifacts/` directory, so I will add the `artifact_minor_upgrade` label to this PR

jtcohen6 (commented):
🎩 Given a `my_small_seed.csv` (<1 MB) and `my_large_seed.csv` (~3 MB)

Using `dbt-core@main`:
```
% dbt parse && mv target/manifest.json state && dbt ls -s state:modified --state state    
16:29:42  Running with dbt=1.10.0-a1
16:29:42  Registered adapter: duckdb=1.9.1
16:29:42  Performance info: /Users/jerco/dev/scratch/testy/target/perf_info.json
16:29:43  Running with dbt=1.10.0-a1
16:29:43  Registered adapter: duckdb=1.9.1
16:29:43  Found 1 operation, 1 model, 2 seeds, 424 macros
16:29:43  Found a seed (testy.my_large_seed) >1MB in size at the same path, dbt cannot tell if it has changed: assuming they are the same
16:29:43  The selection criterion 'state:modified' does not match any enabled nodes
16:29:43  No nodes selected!
```
Switching to `dbt-core@jerco/redo-pr-7125` (without recreating `state/manifest.json`):
```
% dbt ls -s state:modified --state state    
16:30:03  Running with dbt=1.10.0-a1
16:30:03  Registered adapter: duckdb=1.9.1
16:30:03  Found 1 operation, 1 model, 2 seeds, 424 macros
16:30:03  Found a seed (testy.my_large_seed) >1MiB in size at the same path, dbt cannot tell if it has changed: assuming they are the same
16:30:03  The selection criterion 'state:modified' does not match any enabled nodes
16:30:03  No nodes selected!
```
This proves that the `checksum` of `my_small_seed` has not changed.

Now, let's set the config and redo:
```
% export DBT_MAXIMUM_SEED_SIZE_MIB=10
% dbt parse && mv target/manifest.json state && dbt ls -s state:modified --state state
16:30:45  Running with dbt=1.10.0-a1
16:30:45  Registered adapter: duckdb=1.9.1
16:30:45  Performance info: /Users/jerco/dev/scratch/testy/target/perf_info.json
16:30:46  Running with dbt=1.10.0-a1
16:30:46  Registered adapter: duckdb=1.9.1
16:30:46  Found 1 operation, 1 model, 2 seeds, 424 macros
16:30:46  The selection criterion 'state:modified' does not match any enabled nodes
16:30:46  No nodes selected!
```
Manually edit one row in the large seed:
```
% dbt ls -s state:modified --state state
16:36:58  Running with dbt=1.10.0-a1
16:36:58  Registered adapter: duckdb=1.9.1
16:36:58  Found 1 operation, 1 model, 2 seeds, 424 macros
testy.my_large_seed
```

core/dbt/parser/read_files.py
jtcohen6: Update: The failing test [on Windows](https://github.com/dbt-labs/dbt-core/actions/runs/12484224257/job/34841299473?pr=11177) seems to indicate that the seed checksum does indeed change, as a result of this PR, but only on Windows. The contributor mentioned [this code comment](https://github.com/dbt-labs/dbt-core/blob/710600546a22d6fffa40f522250940e35c3a2d38/tests/functional/artifacts/expected_manifest.py#L195-L205) as indication that we expect actually different checksums on Windows versus MacOS / Linux.

In that test, the checksum for `'seed.test.seed'`:
- on MacOS + Linux (before/after): `'381eb2f...'`
- on Windows (before): `'54a28a3...'`
- on Windows (after): `'381eb2f...'`

I think our options are:
1. Restore the behavior to return actually different checksums on different systems. That might be easiest, but I agree it's more desirable to return the same checksum regardless of OS.
2. Place this behind a behavior-change flag, which we very quickly flip to "True" by default (in the next minor version of dbt-core), with targeted advisory for users on Windows: "When this change rolls out, all your seeds will be detected as modified, if comparing against a manifest produced before this code change."
3. Roll this out in the next minor version of dbt-core, _without_ a behavior-change flag. It won't affect the vast majority of users (who are not running on Windows). We mention it in the upgrade guide.

I lean toward option (2) for thoroughness.

ChenyuLInx: I wonder weather we want to just do a logic of: 
- if we find the hash is different on Windows(maybe also if the manifest was computed with a previous version of dbt), compute with the other method to see if it changed, if not, we return not changed but persist the new hash in artifact.

This should provide a seamless upgrade experience and we don't need to have a flag at all.

ChenyuLInx (dismissed):
LGTM with one thought around migration.

noppaz:
Hey @jtcohen6, happy to see some attention here again. While I'm a bit late to the party know that I'd be available to assist further in getting this on the road.

core/dbt/parser/read_files.py
noppaz: As long as option 2 is a minor addition of work in the PR I think that's great for completeness. But as you say with option 3 the impact should be minimal.

core/dbt/contracts/files.py
```diff
@@ -67,9 +66,8 @@ def absolute_path(self) -> str:
     def original_file_path(self) -> str:
         return os.path.join(self.searched_path, self.relative_path)
 
-    def seed_too_large(self) -> bool:
-        """Return whether the file this represents is over the seed size limit"""
-        return os.stat(self.full_path).st_size > MAXIMUM_SEED_SIZE
+    def file_size(self) -> int:
```
MichelleArk, line 72: this could be a @property to conform nicely with the rest of the `FilePath` implementation

core/hatch.toml
```diff
@@ -111,6 +111,7 @@ lint = [
 flake8 = "pre-commit run flake8-check --hook-stage manual --all-files"
 mypy = "pre-commit run mypy-check --hook-stage manual --all-files"
 black = "pre-commit run black-check --hook-stage manual --all-files"
+ft = "python -m pytest -x -nauto ../tests/functional/artifacts/test_artifacts.py::TestVerifyArtifacts::test_run_and_generate"
```
MichelleArk: is this intentional?

core/dbt/cli/params.py
```diff
@@ -322,6 +322,14 @@ def _create_option_and_track_env_var(
     default="eager",
 )
 
+maximum_seed_size_mib = click.option(
+    "--maximum-seed-size-mib",
+    envvar="DBT_MAXIMUM_SEED_SIZE_MIB",
```
MichelleArk: this should be prefixed with `DBT_ENGINE_` a opposed to `DBT_` now

core/dbt/cli/params.py
```diff
@@ -322,6 +322,14 @@ def _create_option_and_track_env_var(
...
+    help="Specify max size (MiB) for seed files that will be hashed for state comparison.",
```
MichelleArk: Would be good to include the interpretation of '0' to mean 'no limit' in the help text here. 

Re: https://github.com/dbt-labs/dbt-core/pull/11177/changes#diff-be2db70c9235cc4e5e40a643431183ffde088141ab6a61e5758cbb386fa31680R137

core/dbt/contracts/graph/nodes.py
```diff
@@ -949,6 +949,21 @@ def same_seeds(self, other: "SeedNode") -> bool:
...
+        # Seamless migration fallback: when upgrading from the old binary-mode
+        # hashing to the new text-mode (from_path) hashing, Windows line
+        # endings (\r\n vs \n) cause checksums to differ even though the file
+        # hasn't changed.  Re-compute using the legacy method and compare.
+        if (
+            not result
+            and self.checksum.name == other.checksum.name
+            and self.checksum.name not in ("path", "none")
+            and self.root_path
+        ):
```
MichelleArk, line 1006: Am I understanding correctly that we're always using the legacy path implementation anytime there is a checksum difference on a seed with any contents (or one that is a large seed)?

core/hatch.toml
sriramr98: No. My bad. I use this method to test specific tests and I forgot to remove it before pushing the changes.

sriramr98: No. My bad. I use this method to test specific tests and I forgot to remove it before pushing the changes.

core/dbt/contracts/graph/nodes.py
sriramr98: Yes. Now that I think about it more, I should probably add `os.name == "nt"` check to only re-compute on windows :D

noppaz:
Hey dbt team, I'm very happy to see some progress here. This is still a pain point for me and my team so I'd be super happy to get this out in 1.12. Let me know if I can help.

core/dbt/artifacts/resources/base.py
```diff
@@ -60,6 +61,40 @@ def from_contents(cls, contents: str, name="sha256") -> "FileHash":
...
+        """Create a file hash from the file at given path. The hash is always the
+        utf-8 encoding of the contents which is stripped to give similar hashes
+        as `FileHash.from_contents`.
+        """
+        path = convert_path(path)
+        chunk_size = 1 * 1024 * 1024
+        file_hash = hashlib.new(name)
+        with open(path, "r") as handle:
```
tauhid621: We should open with utf-8 encoding 
```suggestion
        with open(path, "r", encoding="utf-8") as handle:
```

tauhid621 (approved):

Issue #8872, opened 2023-10-18
[CT-3236] [Bug] When adding a new version of model `foo` - partial parsing runs into a Compilation error 'model.my_dbt_project.bar' depends on 'model.my_dbt_project.foo' which is not in the graph!


jeremyyeo:
### Is this a new bug in dbt-core?

- [X] I believe this is a new bug in dbt-core
- [X] I have searched the existing issues, and I could not find an existing issue for this bug

### Current Behavior

> Looks similar to #8859 

When adding a new model version (i.e.`foo_v2.sql`) - partial parsing appears to not be able to find the previous version of the model even though the `foo.sql` file exist. Things work correctly when doing a full parse (i.e. `dbt clean` / delete `target/` folder first).

### Expected Behavior

Partial parsing should be able to detect the previous model file.

### Steps To Reproduce

1. Project setup.

```yml
# dbt_project.yml
name: my_dbt_project
profile: all
config-version: 2
version: "1.0.0"

models:
  my_dbt_project:
    +materialized: table

# models/schema.yml
version: 2
models:
  - name: bar
  - name: foo
```

```sql
-- models/bar.sql
select * from {{ ref('foo') }}

-- models/foo.sql
select 1 as id
```

2. Build project to create initial `target/partial_parse.msgpack` file:

```sh
$ ls target

ls: target: No such file or directory

$ dbt clean && dbt build

21:27:02  Running with dbt=1.6.6
21:27:02  Checking target/*
21:27:02  Cleaned target/*
21:27:02  Finished cleaning all paths.
21:27:07  Running with dbt=1.6.6
21:27:07  Registered adapter: postgres=1.6.6
21:27:07  Unable to do partial parsing because saved manifest not found. Starting full parse.
21:27:08  Found 2 models, 0 sources, 0 exposures, 0 metrics, 352 macros, 0 groups, 0 semantic models
21:27:08  
21:27:08  Concurrency: 1 threads (target='pg-local')
21:27:08  
21:27:08  1 of 2 START sql table model public.foo ........................................ [RUN]
21:27:08  1 of 2 OK created sql table model public.foo ................................... [SELECT 1 in 0.17s]
21:27:08  2 of 2 START sql table model public.bar ........................................ [RUN]
21:27:08  2 of 2 OK created sql table model public.bar ................................... [SELECT 1 in 0.07s]
21:27:08  
[...]

PR #11184, opened 2025-01-02
fix: partial parsing - reparse downstream nodes when adding versioning

d-cole:
Resolves # https://github.com/dbt-labs/dbt-core/issues/8872

### Problem

Adding versioning to a model with dependencies requires a full-parse. See [workaround](https://github.com/dbt-labs/dbt-core/issues/8872#issuecomment-2071146875).

### Solution

Schedule referencing nodes for parsing if the model in the new schema file is versioned ([here](https://github.com/dbt-labs/dbt-core/pull/11184/files#diff-2fd42940aaee0e200a257863155b747faf735f6b49ae49ac9229c17be33db82aR865)). 

### Reviewer Questions
 - For this work and similar PRs, should unit and functional tests be added? Based on,  ["functional tests (tests/functional) cover anything that interacts with a database, namely adapter"](https://github.com/dbt-labs/dbt-core/blob/0bf38ce294e21c6e4056ed36ba8f35538c10604a/CONTRIBUTING.md#unit-integration-functional)",  I initially planned on only adding unit tests. However, after seeing the existing tests in  tests/functional/partial_parsing it seemed appropriate to add a functional test as well. 
- Type annotations were added to `delete_schema_mssa_links` but I ignored hints on the assignments to avoid including a larger refactor in this PR. Is there a preference for how type changes like this are handled? I see a mix of blanket ignores and type specific ignores (e.g. `# type: ignore[assignment]`). 

### Checklist

### DIFF
--- a/.changes/unreleased/Fixes-20250102-140543.yaml
+++ b/.changes/unreleased/Fixes-20250102-140543.yaml
@@ -1,0 +1,6 @@
+kind: Fixes
+body: Make partial parsing reparse referencing nodes of newly versioned models.
+time: 2025-01-02T14:05:43.629959-05:00
+custom:
+    Author: d-cole
+    Issue: "8872"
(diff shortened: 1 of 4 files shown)


gshank (approved):
Looks good!

PR #11195, opened 2025-01-07
Create `LogNodeResult` event

aranke:
### Problem

We need a new event to log generic node result events with dynamic messages.

### Solution

Create a new `LogNodeResult` event

### Checklist


gshank (commented):
It seems odd that we'd load up the event with a dictionary of status_to_message_map, when at the time we do that, we'd in theory already have the status. Is there some reason to do it this way instead of simply loading the appropriate message?

aranke:
@gshank Good catch, looks like I can simplify further which I have done. 
Could you please re-review this PR? Thanks!

gshank (approved):

PR #11205, opened 2025-01-09
Re-cast definition of unique_key in SnapshotConfig

gshank:
Resolves #

### Problem

Python caching of unique_key definition in SnapshotConfig causes deserialization errors because of wrong order.

### Solution

Change definition from `Optional[Union[str, List[str]]]` to `Union[str, List[str], None]`.

### Checklist


aranke (approved):
Thanks for finding and fixing this!

PR #11236, opened 2025-01-23
[TIDY FIRST] Click option defintiion organization (alphabetization)

QMalcolm:
Resolves #N/A because it's a tidy first

### Problem

The organization of click option definitions in `cli/params.py` was a weird mixture of some alphabetization, new options getting added to the EOF, and grouping by feature. This mean the only way to find an option was to do a text search which was annoying. It also made it confusing as to where a _new_ option should get added. Alphabetically? To the end of the file? Was there some similar option that the new option should be "near"?

### Solution

Alphabetize the click option definitions.

### Checklist

### DIFF
--- a/core/dbt/cli/params.py
+++ b/core/dbt/cli/params.py
@@ -7,6 +7,18 @@ from dbt.cli.options import MultiOption
 from dbt.cli.resolvers import default_profiles_dir, default_project_dir
 from dbt.version import get_version_information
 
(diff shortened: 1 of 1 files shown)


QMalcolm:
A quick way to "validate" all the options are still there is to do a text search on the old and new file for `click.option(`. They should have the same number of hits.

ChenyuLInx (approved):

PR #11260, opened 2025-01-30
Don't automatically run Artifact Schema Check on PRs

aranke:
Sample run: https://github.com/dbt-labs/dbt-core/actions/runs/13058262920

Input option:

image

### DIFF
--- a/.github/workflows/schema-check.yml
+++ b/.github/workflows/schema-check.yml
@@ -18,6 +18,12 @@ on:
   #    paths-ignore: [ '.changes/**', '.github/**', 'tests/**', '**.md', '**.yml' ]
 
   workflow_dispatch:
+    inputs:
+      target_branch:
+        description: "The branch to check against"
+        type: string
+        default: 'main'
+        required: true
 
 # no special access is needed
 permissions: read-all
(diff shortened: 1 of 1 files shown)


.github/workflows/schema-check.yml
```diff
@@ -9,13 +9,13 @@
...
 
 on:
-  pull_request:
-    types: [ opened, reopened, labeled, unlabeled, synchronize ]
-    paths-ignore: [ '.changes/**', '.github/**', 'tests/**', '**.md', '**.yml' ]
+  #  pull_request:
+  #    types: [ opened, reopened, labeled, unlabeled, synchronize ]
+  #    paths-ignore: [ '.changes/**', '.github/**', 'tests/**', '**.md', '**.yml' ]
 
   workflow_dispatch:
```
emmyoop, line 20: Let's add a branch input here that defaults to main.  Then we could run it for PRs we have open to generate the file.  Then just add a `ref: ${{ inputs.branch }}` on line 45 and we'll check out the right branch

emmyoop (changes requested):
Let's just make it a bit more flexible.

emmyoop (approved):
🎉  LGTM

PR #11284, opened 2025-02-07
Update ShowTaskDirect to correctly handle --limit -1

WilliamDee:
### Problem
In dbt show, if we don't pass --limit it defaults to 5 and renders LIMIT 5 at the end of the SQL. You can get around it by adding --limit -1 so that it doesn't render the LIMIT. However, it seems like that only works for --inline and if you try doing dbt show --inline-direct <> --limit -1 it errors with an adapter error showing cannot limit by a negative row number for certain adapters. In the `ShowRunner` we handle this case by `limit = None if self.config.args.limit < 0 else self.config.args.limit`, but this isn't the case for `ShowTaskDirect`.

### Solution
update `ShowTaskDirect` so that the limit gets set to `None` if it's a negative int when passed to the adapter.

### Checklist


thiagosalvatore (approved):

aranke (approved):
LGTM, thanks!

PR #11287, opened 2025-02-09
Fix default_profiles_dir

### LINKED ISSUE #11286: [Bug] Error occurs even if profiles_dir is specified on dbt cli when the home/.dbt directory does not exist

### Current Behavior

Error occurs even if profiles_dir is specified in kwargs on dbt cli when the home/.dbt directory does not exist.
But it worked when I specified --profiles-dir in the args or when I specified it in an environment variable.

I don't think this is the expected result.

### Expected Behavior

Even if home/.dbt does not exist, no error will occur if profiles_dir is specified.

### Steps To Reproduce

1. Delete home/.dbt.
2. Run the code below.
```python
DBT_DIR = "/path/to/dbt_dir"
res: dbtRunnerResult = dbt.invoke(
    args=args + ["ls"],
    project_dir=DBT_DIR,
    profiles_dir=DBT_DIR,
)
if res.success == False:
    logger.error(f"Command failed with error: {res.exception}")        
```
3. Error occured.
```
Command failed with error: Path '/Users/tnk-ysk/.dbt' does not exist.
```

### Relevant log output

```shell

```

### Environment

```markdown
- OS: macOS 14.4
- Python: 3.11.10
- dbt: 1.9.1
```

_No response_

### Additional Context

_No response_

tnk-ysk:
Resolves #11286

### Problem
Avoid error if [`default_profiles_dir`](https://github.com/dbt-labs/dbt-core/blob/300aa09fc562fef6dade08cc5cd7297d636efa3f/core/dbt/cli/resolvers.py#L13) does not exist if profiles_dir is specified.

### Solution
Check current directory and home/.dbt and use current directory if profiles.yml was not present in either.
This allows to avoid [`click.Path(exists=True)`](https://github.com/dbt-labs/dbt-core/blob/300aa09fc562fef6dade08cc5cd7297d636efa3f/core/dbt/cli/params.py#L460) in default_profiles_dir if profiles_dir is specified.

This solution is similar to [`default_project_dir`](https://github.com/dbt-labs/dbt-core/blob/300aa09fc562fef6dade08cc5cd7297d636efa3f/core/dbt/cli/resolvers.py#L7).

### Checklist

### DIFF
--- a/.changes/unreleased/Fixes-20250210-000107.yaml
+++ b/.changes/unreleased/Fixes-20250210-000107.yaml
@@ -1,0 +1,6 @@
+kind: Fixes
+body: Error occurs if default_profiles_dir does not exist
+time: 2025-02-10T00:01:07.477255+09:00
+custom:
+    Author: tnk-ysk
+    Issue: "11286"
--- a/.changes/unreleased/Fixes-20250210-000107.yaml
+++ b/.changes/unreleased/Fixes-20250210-000107.yaml
@@ -1,5 +1,5 @@
 kind: Fixes
-body: Error occurs if default_profiles_dir does not exist
+body: Fallback to current working directory (CWD) for `profiles_dir` if `profiles.yml` not found in CWD and `$HOME/.dbt`
 time: 2025-02-10T00:01:07.477255+09:00
 custom:
     Author: tnk-ysk
--- a/core/dbt/cli/resolvers.py
+++ b/core/dbt/cli/resolvers.py
@@ -10,7 +10,8 @@ def default_project_dir() -> Path:
 
 
 def default_profiles_dir() -> Path:
-    return Path.cwd() if (Path.cwd() / "profiles.yml").exists() else Path.home() / ".dbt"
+    paths = [Path.cwd(), Path.home() / ".dbt"]
+    return next((x for x in paths if (x / "profiles.yml").exists()), Path.cwd())
 
 
 def default_log_path(project_dir: Path, verify_version: bool = False) -> Path:


.changes/unreleased/Fixes-20250210-000107.yaml
```diff
@@ -0,0 +1,6 @@
+kind: Fixes
+body: Error occurs if default_profiles_dir does not exist
```
dbeatty10, line 2:
```suggestion
body: Fallback to current working directory (CWD) for `profiles_dir` if `profiles.yml` not found in CWD and `$HOME/.dbt`
```

core/dbt/cli/resolvers.py
```diff
@@ -10,7 +10,8 @@ def default_project_dir() -> Path:
 
 
 def default_profiles_dir() -> Path:
-    return Path.cwd() if (Path.cwd() / "profiles.yml").exists() else Path.home() / ".dbt"
+    paths = [Path.cwd(), Path.home() / ".dbt"]
+    return next((x for x in paths if (x / "profiles.yml").exists()), Path.cwd())
```
dbeatty10, line 14: It looks to me like the **true** problem is that executing the `runner.py` script [here](https://github.com/dbt-labs/dbt-core/issues/11286#issuecomment-2757855349) is behaving differently than `dbt ls --profiles-dir dbt_dir` for some reason.

We should get to the root cause of that discrepancy (which might trace back to [here](https://github.com/dbt-labs/dbt-core/blob/300aa09fc562fef6dade08cc5cd7297d636efa3f/core/dbt/cli/main.py#L53)). Or maybe [here](https://github.com/dbt-labs/dbt-core/blob/300aa09fc562fef6dade08cc5cd7297d636efa3f/core/dbt/cli/flags.py#L253-L254).

This solution looks like it would just mask the underlying discrepancy, so I do **not** think we should adopt it until we fully understand the root cause.

PR #11294, opened 2025-02-11
Round 2: Add doc_blocks to manifest for nodes and columns

aranke:
### Problem

#11224 (Add doc_blocks to manifest for nodes and columns) would have failed in a scenario where the user had already specified `doc_blocks` at the node or column level as outlined in #11282 (quick functional test on doc_blocks issue).

### Solution

- [x] Set `doc_blocks=[]` during serialization if either:
  - [x] `doc_blocks` is not a list
  - [x] not every element in `doc_blocks` is a `str`
- [x] Add tests for the scenarios above, which should satisfy the Mashumaro type of `list[str]`

### Checklist

### DIFF
--- a/.changes/unreleased/Features-20250122-170328.yaml
+++ b/.changes/unreleased/Features-20250122-170328.yaml
@@ -1,0 +1,6 @@
+kind: Features
+body: Add doc_blocks to manifest for nodes and columns
+time: 2025-01-22T17:03:28.866522Z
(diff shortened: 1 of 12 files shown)


emmyoop (approved):
Looks good - let just add a bit more context as to why we're updating to an empty list in some cases so future us has a better idea of what's happening.

Issue #11228, opened 2025-01-22
Remove `dbt_experimental_sample_mode` environment variable gate


QMalcolm:
### Housekeeping

- [X] I am a maintainer of dbt-core

### Short description

Once "sample mode" is relatively stable, we should remove the environment variable `dbt_experimental_sample_mode` gating of the functionality. We should do this either as part of the `beta`, but at the latest needs to happen before the GA of dbt-core `1.10.0`.

### Acceptance criteria

- The environment variable gating for sample mode has been removed

### Suggested Tests

- the environment variable `dbt_experimental_sample_mode` has _no_ effect on one's ability to run in "sample_mode"

### Impact to Other Teams

N/A

### Will backports be required?

No, it's a new feature

### Context

https://github.com/dbt-labs/dbt-core/issues/11227

PR #11307, opened 2025-02-14
Allow for running `sample mode` with `build` command

### LINKED ISSUE #11298: allow for sample mode during `build` command

QMalcolm:
Resolves #11298

### Problem

Sample mode could only be run with `run` command, we want to also use it with `build` command

### Solution

Allow for specification of `--sample=SAMPLE_SPEC` on `build` command

### Checklist

### DIFF
--- a/.changes/unreleased/Features-20250213-182932.yaml
+++ b/.changes/unreleased/Features-20250213-182932.yaml
@@ -1,0 +1,6 @@
+kind: Features
+body: Enable sample mode for 'build' command
+time: 2025-02-13T18:29:32.238857-06:00
+custom:
+  Author: QMalcolm
+  Issue: "11298"
--- a/core/dbt/cli/main.py
+++ b/core/dbt/cli/main.py
@@ -179,6 +179,7 @@ def cli(ctx, **kwargs):
 @p.project_dir
 @p.resource_type
 @p.exclude_resource_type
+@p.sample
 @p.select
 @p.selector
 @p.show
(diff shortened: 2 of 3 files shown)


aranke (commented):
Can you add a test for a node that cannot be sampled? E.g. exposures

QMalcolm:
> Can you add a test for a node that cannot be sampled? E.g. exposures

@aranke Hmmm I'm not sure I agree. 

In the specific context of exposures, there isn't a way I'm aware of to test this. Exposures are leaf nodes and thus can't be ref'd, which doesn't seem in the purview of this work. There is an argument of, we should test whether the `depends_on` ref's and source's should be tested for exposures. However, exposures are `no-op`'s, and so I'm not aware of a way to test them.

There is an argument for the general "test for a node that cannot be sampled". However, I don't think it is actually time for that yet. Sample mode is still in alpha/beta, and using it is gated behind an environment variable. Our general goal for GA is that for anything that render's `ref()` and `source()` into `select` SQL statements, those `ref`/`source` statements should be sampleable. As we add specific support for different node types being `ref`'d, we add positive tests to assert that they are supported. If we end up specifically not supporting sampling of `ref`s of certain node types, we should add tests for that. We have not identified anything that fits that yet though. If we do identify anything like that, it'll likely be tied to #11228

MichelleArk (approved):

aranke (approved):
Clarified with @QMalcolm offline, this only affects nodes that can be called via `ref` or `source`.

PR #11308, opened 2025-02-14
Add secondary_profiles to `profile.py`

aranke:
Resolves [XPLAT-241](https://dbtlabs.atlassian.net/browse/XPLAT-241).

### Problem

We need to parse `secondary_profiles` associated with a given profile.

### Solution

- Create a new `secondary_profiles` dict on the `Profile` class.
- Validate and load each secondary profile as if they were primary profiles.
- Allow `target_name` to be optional iff a secondary profile has a single target.
- Add tests.

### Checklist

[XPLAT-241]: https://dbtlabs.atlassian.net/browse/XPLAT-241?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/core/dbt/config/profile.py
+++ b/core/dbt/config/profile.py
@@ -1,6 +1,8 @@
 import os
 from dataclasses import dataclass
-from typing import Any, Dict, Optional, Self, Tuple
+from typing import Any, Dict, Optional, Tuple
+
+from typing_extensions import Self
 
 from dbt.adapters.contracts.connection import Credentials, HasCredentials
 from dbt.clients.yaml_helper import load_yaml_text
--- a/core/dbt/config/profile.py
+++ b/core/dbt/config/profile.py
@@ -324,11 +324,10 @@ defined in your profiles.yml file. You can find profiles.yml here:
             raw_profile, profile_name, target_override, renderer, is_secondary=is_secondary
         )
 
-        if is_secondary:
-            if "secondary_profiles" in profile_data:
-                raise DbtProfileError(
-                    f"Secondary profile '{profile_name}' cannot have nested secondary profiles"
-                )
+        if is_secondary and "secondary_profiles" in profile_data:
+            raise DbtProfileError(
+                f"Secondary profile '{profile_name}' cannot have nested secondary profiles"
+            )
 
         # valid connections never include the number of threads, but it's
         # stored on a per-connection level in the raw configs
--- a/core/dbt/config/profile.py
+++ b/core/dbt/config/profile.py
@@ -2,8 +2,6 @@ import os
 from dataclasses import dataclass
 from typing import Any, Dict, Optional, Tuple
 
-from typing_extensions import Self
-
 from dbt.adapters.contracts.connection import Credentials, HasCredentials
 from dbt.clients.yaml_helper import load_yaml_text
 from dbt.contracts.project import ProfileConfig
@@ -62,7 +60,7 @@ class Profile(HasCredentials):
     credentials: Credentials
     profile_env_vars: Dict[str, Any]
     log_cache_events: bool
-    secondary_profiles: Dict[str, Self]
+    secondary_profiles: Dict[str, "Profile"]
 
     def __init__(
         self,
@@ -233,7 +231,7 @@ defined in your profiles.yml file. You can find profiles.yml here:
         threads: int,
         profile_name: str,
         target_name: str,
-    ) -> Self:
+    ) -> "Profile":
         """Create a profile from an existing set of Credentials and the
         remaining information.
 
@@ -304,7 +302,7 @@ defined in your profiles.yml file. You can find profiles.yml here:
         target_override: Optional[str] = None,
         threads_override: Optional[int] = None,
         is_secondary: bool = False,
-    ) -> Self:
+    ) -> "Profile":
         """Create a profile from its raw profile information.
 
          (this is an intermediate step, mostly useful for unit testing)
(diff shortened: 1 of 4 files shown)


aranke:
Converted to draft temporarily to see if I can address feedback from @amychen1776 re making `target` optional for secondary profiles.

core/dbt/config/profile.py
```diff
@@ -325,13 +331,53 @@ def from_raw_profile_info(
...
             profile_name=profile_name,
             target_name=target_name,
             threads=threads,
         )
 
+        raw_secondary_profiles = profile_data.pop("secondary_profiles", [])
+
+        if raw_secondary_profiles:
+            for p in raw_secondary_profiles:
+                for secondary_profile_name, secondary_raw_profile in p.items():
```
mikealfare, line 345: The contents of this `for` loop are so similar to the original method that I would consider calling this method again to process the secondary profile (instead of duplicating the code). It would require doing something like adding an `is_secondary: bool = False` to this method's signature, and potentially updating `infer_target_name` to `is_secondary` in `cls.render_profile`, but I think it might be cleaner. I'm anticipating it could be boiled down to something like:
```python
    @classmethod
    def from_raw_profile_info(
        cls,
        raw_profile: Dict[str, Any],
        profile_name: str,
        renderer: ProfileRenderer,
        target_override: Optional[str] = None,
        threads_override: Optional[int] = None,
        is_secondary: bool = False,
    ) -> Self:
    ...
    
    target_name, profile_data = cls.render_profile(
        raw_profile,
        profile_name,
        target_override,
        renderer,
        is_secondary,
    )
    
    ...
    
    for secondary_profile, secondary_raw_profile in p.items():
        if secondary_profile_name in profile.secondary_profiles:
            raise DbtProfileError(f"Secondary profile '{secondary_profile_name}' is already defined")
    
        profile.secondary_profiles[secondary_profile_name] = cls.from_raw_profile_info(
            secondary_raw_profile,
            secondary_profile_name,
            renderer,
            target_override=target_override,
            threads_override=threads_override,
            is_secondary=True,
        )

    return profile
```

core/dbt/config/profile.py
```diff
@@ -60,6 +60,7 @@ class Profile(HasCredentials):
     credentials: Credentials
     profile_env_vars: Dict[str, Any]
     log_cache_events: bool
+    secondary_profiles: Dict[str, "Profile"]
```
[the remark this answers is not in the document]
aranke: Yes, TIL

core/dbt/config/profile.py
```diff
@@ -273,6 +278,11 @@ def render_profile(
         elif "target" in raw_profile:
             # render the target if it was parsed from yaml
             target_name = renderer.render_value(raw_profile["target"])
+        elif is_secondary and len(raw_profile["outputs"]) == 1:
```
MichelleArk, line 279: Is the "output" key guaranteed to exist? Let's make a .get("outputs", []) call here to be safe.

core/dbt/config/profile.py
```diff
@@ -273,6 +278,11 @@ def render_profile(
...
+            # if we only have one target, we can infer the target name
+            # currently, this is only used for secondary profiles
+            target_name = next(iter(raw_profile["outputs"]))
+            fire_event(MissingProfileTarget(profile_name=profile_name, target_name=target_name))
```
MichelleArk, line 283: Why is this event necessary in this case?

core/dbt/config/profile.py
MichelleArk: Separately -- is this case mutually exclusive from the above? Or does this is_secondary handling still need to happen even if "target" is in raw_profile?

aranke: > Is the "output" key guaranteed to exist?
I think so, an error gets thrown earlier – `test_no_outputs` and `test_secondary_profiles_no_outputs` are checking for this scenario.

> Let's make a .get("outputs", []) call here to be safe.
Agreed, I'll do this regardless.

aranke: > Separately -- is this case mutually exclusive from the above?

I believe so, yes. If user specifies `target: foo` but `outputs` only has `target: bar`, we should raise an error instead of defaulting to `bar`.

> Or does this is_secondary handling still need to happen even if "target" is in raw_profile?

I'm not sure I understand the question – if `"target" in raw_profile == True`, then `is_secondary and len(raw_profile["outputs"]) == 1` is never evaluated?

core/dbt/config/profile.py
aranke: The event name is confusing, but the message is useful for debugging purposes to indicate that we inferred a `target_name`:

https://github.com/dbt-labs/dbt-core/blob/e60b41d9fabb72fccd5e94a8ceb2a4f413c6f161/core/dbt/events/types.py#L68-L73

I'll add a comment for the same.

core/dbt/config/profile.py
MichelleArk: > if "target" in raw_profile == True, then is_secondary and len(raw_profile["outputs"]) == 1 is never evaluated

That's the current behaviour as-implemented. Just confirming that it's intentional. Sounds like it is

aranke: Yup, it's intentional. Thanks for checking!

core/dbt/config/profile.py
aranke: Added a comment here, let me know if any other changes are needed here.

MichelleArk (approved):

PR #11332, opened 2025-02-24
Rewrite execution of microbatch models to avoid blocking the main thread

### LINKED ISSUE #11243: [Bug] Microbatch models shouldn't block the main thread in multi-threaded dbt runs.
### Housekeeping

### Short description

Microbatch models currently block the main thread when running dbt with multiple threads. This affects microbatch models that have their batches run concurrently or sequentially, but has greater impacts when the batches are being run sequentially. In essence, the issue is that the scheduling of batch execution for a microbatch model happens on the main thread. This means that the scheduling of other models becomes blocked until the microbatch model is complete.

For example, consider we have a microbatch model in our project. The microbatch model has many sibling/cousin/nibling/etc nodes. Our dbt project is configured to have multiple threads. Once the main thread reaches the microbatch model, work already being done on worker threads for other nodes _will continue_. However, no new worker threads for other nodes will be spun up as this scheduling is handled by the main thread, which is now blocked by the microbatch model. When the batches for microbatch models are run sequentially, this means all the batches are being run on the main thread, basically reducing the dbt invocation to being single threaded until the completion of the microbatch model. When batches for microbatch models are run concurrently, worker threads will be spun up for the batches. However, the main thread is still blocked by the microbatch model until _all the batches_ have completed. In this case, the worker threads will be saturated _for the most part_. However, if there is one long-running batch where all the other work thread batches have finished, the remaining threads will sit idle until the last batch finishes for the microbatch model.

We should instead delegate the scheduling of microbatch batches _to a worker thread itself_. This would mean that when batches are being run sequentially, only one worker thread would be occupied, and would not block the main thread. When batches are being run concurrently, one worker thread will do the scheduling and saturate the other worker threads with batches. This does mean that one worker thread won't be doing batch work, which isn't great. However, if there are some long-running batches that are holding only a subset of the worker threads, then main thread can continue to allocate other nodes to the freed up worker nodes.

### Acceptance criteria

* microbatch batch execution doesn't block the main thread in multi-threaded dbt environments

### Suggested Tests

* microbatch batch execution doesn't block the main thread in multi-threaded dbt environments

### Impact to Other Teams

N/A

Possibly 1.9, although this might be a large enough change that it isn't safe to do so.

### Context

_No response_

### LINKED ISSUE #11306: [Bug] Microbatch can cause dbt runs to hang

### Current Behavior

You can cause microbatch models to hang the dbt runtime. This is most easily reproduced by providing the `--fail-fast` flag with a microbatch that has an error in it, full reproduction steps can be found https://github.com/dbt-labs/TMP_MICROBATCH_HANGING_REPRO. The hanging has been observed in other manners, but they are not as easily reproduced.

### Expected Behavior

Microbatch models shouldn't cause the dbt runtime to hang.

### Steps To Reproduce

https://github.com/dbt-labs/TMP_MICROBATCH_HANGING_REPRO

### Relevant log output

```shell

```

### Environment

```markdown
- OS: MacOS 15.1.1
- Python: 3.11.9
- dbt: 1.9.2
```

snowflake

### Additional Context

_No response_

QMalcolm:
Resolves #11243
Resolves #11306

### Problem

There are two problems
1. Executing microbatch model batches concurrently would block the main thread from scheduling other nodes as described in #11243
2. In certain scenarios, a microbatch model would _hang_ indefinitely as described in #11306

### Checklist

### DIFF
--- a/core/dbt/graph/thread_pool.py
+++ b/core/dbt/graph/thread_pool.py
@@ -1,0 +1,18 @@
+from __future__ import annotations
+
+from multiprocessing.pool import ThreadPool
+
+
+class DbtThreadPool(ThreadPool):
+    """A ThreadPool that tracks whether or not it's been closed"""
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.closed = False
+
+    def close(self):
+        self.closed = True
+        super().close()
+
+    def is_closed(self):
+        return self.closed
--- a/core/dbt/task/run.py
+++ b/core/dbt/task/run.py
@@ -345,28 +345,6 @@ class MicrobatchModelRunnerOLD(ModelRunner):
         self.batches: Dict[int, BatchType] = {}
         self.relation_exists: bool = False
 
-    def on_skip(self):
-        # TODO: Split into two method
-        # The first part of the if statement should move to the microbatch orchestration runner
-        # The second should move to the batch runner
-        # If node.batch is None, then we're dealing with skipping of the entire node
-        if self.batch_idx is None:
-            return super().on_skip()
-        else:
-            result = RunResult(
-                node=self.node,
-                status=RunStatus.Skipped,
-                timing=[],
-                thread_id=threading.current_thread().name,
-                execution_time=0.0,
-                message="SKIPPED",
-                adapter_response={},
-                failures=1,
-                batch_results=BatchResults(failed=[self.batches[self.batch_idx]]),
-            )
-            self.print_batch_result_line(result=result)
-            return result
-
     def _build_succesful_run_batch_result(
         self,
         model: ModelNode,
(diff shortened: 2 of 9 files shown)


core/dbt/graph/thread_pool.py
```diff
@@ -0,0 +1,18 @@
+from __future__ import annotations
+
+from multiprocessing.pool import ThreadPool
+
+
+class DbtThreadPool(ThreadPool):
```
QMalcolm, line 6: We created `DbtThreadPool` so that we can have visibility on whether `.close()` has been called on the pool. This class is now used instead of `ThreadPool`.

core/dbt/task/run.py
```diff
@@ -683,136 +611,241 @@ def _is_incremental(self, model) -> bool:
...
+        runner = self.get_runner(node)
+        # we finally know what we're running! Make sure we haven't decided
+        # to skip it due to upstream failures
+        if runner.node.unique_id in self._skipped_children:
+            cause = self._skipped_children.pop(runner.node.unique_id)
+            runner.do_skip(cause=cause)
+
+        if isinstance(runner, MicrobatchModelRunner):
+            runner._parent_task = self
+            runner._pool = pool
```
MichelleArk, line 844: nit: these should probably go through setters instead of updating the 'private' attrs directly.

MichelleArk (approved):

core/dbt/task/run.py
QMalcolm: Fixed in [a32a7ae](https://github.com/dbt-labs/dbt-core/pull/11332/commits/a32a7ae553b8a8a19051c39e90dbfb031d8da82c) 🙂

PR #11366, opened 2025-03-06
test fix

MichelleArk:
Resolves #

### Checklist

### DIFF
--- a/core/dbt/contracts/graph/manifest.py
+++ b/core/dbt/contracts/graph/manifest.py
@@ -519,7 +519,7 @@ def _packages_to_search(
     elif current_project == node_package:
         return [current_project, None]
     else:
-        return [current_project, node_package, None]
+        return [node_package, current_project, None]
 
 
 def _sort_values(dct):


jtcohen6:
_discussion with @zqureshi_

This feels like a correct change to me - strongly agree with the rationale that @dbeatty10 @TomSteenbergen laid out in the other thread.

This change would negatively impact someone IFF:
- They have `model_x` defined in both their root project and an installed package
- They are counting on the fact that (single-argument) references to `model_x` within that package will resolve to the root project implementation, rather than the package installation

This was not possible prior to dbt-core v1.6 (released 1.5-2 years ago), when we added support for model namespacing (i.e. models by the same name in different projects/packages). The previous approach to accomplishing ^this use case, [which we documented for many years](https://docs.getdbt.com/reference/resource-configs/enabled#disable-a-model-in-a-package-in-order-to-use-your-own-version-of-the-model), was to disable the package `model_x` and reimplement `model_x` in the root project.

So I think the risk here is relatively small, but it's also deeply nested logic, and we don't know exactly whom it might affect. (I did a little looking, and I did remember that this is a fairly common pattern for "template" models/packages that are reconfigured across multiple clients/regions/etc.)

Let's take a more-cautious approach and put this change behind a behavior flag, planned for `True` in v1.10. (I don't know if there's a good way for us to fire a log / deprecation warning without slowing down `ref` resolution, so we can track potential impact of this change, but that's worth some time-boxed investigation.)

TomSteenbergen:
@MichelleArk Any update on this PR? IIUC, @jtcohen6's comment [here](https://github.com/dbt-labs/dbt-core/pull/11366#issuecomment-2776810478) still requires a follow-up

core/dbt/contracts/graph/manifest.py
```diff
@@ -558,7 +558,10 @@ def _packages_to_search(
     elif current_project == node_package:
         return [current_project, None]
     else:
-        return [current_project, node_package, None]
+        if get_flags().require_ref_searches_node_package_before_root:
+            return [node_package, current_project, None]
+        else:
+            return [current_project, node_package, None]
```
QMalcolm, line 564: 👍🏻

core/dbt/contracts/project.py
```diff
@@ -367,6 +367,7 @@ class ProjectFlags(ExtensibleDbtClassMixin):
     require_all_warnings_handled_by_warn_error: bool = False
     require_generic_test_arguments_property: bool = True
     require_unique_project_resource_names: bool = False
+    require_ref_searches_node_package_before_root: bool = False
```
QMalcolm, line 370: 👍🏻 👍🏻

core/dbt/events/types.py
```diff
@@ -1271,6 +1271,19 @@ def message(self) -> str:
...
+    def code(self) -> str:
+        return "I077"
+
+    def message(self) -> str:
+        msg = (
+            f"The node '{self.node_name}'in package '{self.package_name}' depends on the root project node '{self.root_project_unique_id}'."
+            "This may lead to unexpected cycles downstream. Please set the 'require_ref_prefers_node_package_to_root' behavior change flag to True to avoid this issue."
+            "For more information, see the documentation at https://docs.getdbt.com/reference/global-configs/behavior-changes#require_ref_prefers_node_package_to_root"
+        )
+        return warning_tag(msg)
```
QMalcolm, line 1283: 👍🏻 👍🏻 👍🏻

core/dbt/parser/manifest.py
```diff
@@ -1629,6 +1630,19 @@ def invalid_target_fail_unless_test(
         )
 
 
+def warn_if_package_node_depends_on_root_project_node(
+    node: ManifestNode, target_model: ManifestNode, current_project: str
+) -> None:
+    if node.package_name != current_project and target_model.package_name == current_project:
```
QMalcolm: Do we want this to be slightly more targeted? My understanding is we're moving to a world where the package model will be preferred. That is the behavior flag, when flipped to true, will cause a matching model in the package to be referenced over the root project model IFF there is a matching model in the package. Is it guaranteed that the package has an implementation of the referenced model? If not, do we want to add another gate here that the `node.package_name` package must have a matching node? 🤔

QMalcolm:
Eeek I meant to "comment" instead of "request changes" 🤦🏻

core/dbt/parser/manifest.py
MichelleArk: > do we want to add another gate here that the node.package_name package must have a matching node? 

Hmmm, although I can't think of a use case where you'd want this, I also don't see any reason to restrict that given that core will already raise an error for problematic behavior like cycles or unresolved refs.

QMalcolm: I'm thinking specifically of the shape
1. package has model_b with `{{ ref("model_a") }}`
2. package doesn't have a `model_a`
3. root project has a `model_a`

I believe the above is possible? In which case the resolution in either state of the behavior flag is the same (`model_a` from root project), but the warning will be fired

QMalcolm: And specifically I'm thinking about the validity of the warning in different scenarios

QMalcolm (approved):
Thank you for doing this work! Lets do itttt 🙂

PR #11389, opened 2025-03-14
Macro Annotations and Inference

### LINKED ISSUE #11274: Macro Annotations and Inference
#### Arguments

We support [documenting macros in yml files](https://docs.getdbt.com/faqs/Docs/documenting-macros). Here is the standard example:

```yaml
macros:
  - name: cents_to_dollars
    description: A macro to convert cents to dollars
    arguments:
      - name: column_name
        type: string
        description: The name of the column you want to convert
      - name: precision
        type: integer
        description: Number of decimal places. Defaults to 2.
```

Currently, the `type` field is not validated and used only as documentation. We will implement opt-in validation and functionality for this field.

The `type` field will now support any value given by the following rule.

```
<T> := 
    str
    string
    bool
    int
    integer
    float
    any
    list[<T>]
    dict[<T>, <T>]
    optional[<T>]
    relation
    column
```

So, `str`, `bool`, `list[str]`, and `dict[str, list[int]]`, would all be valid types. Although this naming convention follows Python style conventions, these types should not be confused with Python types. Note that `string` is a synonym for `str` and `integer` is a synonym for `int`. This flexibility will allow users to avoid revising existing documentation which may have used older, inconsistent conventions.

### Acceptance Criteria

- We have added a new [behavior change flag](https://docs.getdbt.com/reference/global-configs/behavior-changes#behavior-change-flags) to optionally validate macro annotations
  - If the flag is set to false (default), we continue to permit any value for `type` and `name`
  - If flag is set to true (opt-in), we:
    - Raise a warning if the argument names you've added in yml do not match the argument names you have in your macro
    - Raise a warning if the argument types you've added in yml are not valid according to the rule above
- If no argument names are documented in yml, we infer them based on what you have in the macro and include. them in the manifest
- There will be no changes to manifest and no change to schema

### More info on types during refinement

- `optional[<T>]` -> `none` in jinja
- `relation` -> https://docs.getdbt.com/reference/dbt-classes#relation
- `column` -> https://docs.getdbt.com/reference/dbt-classes#column

### To do in a follow up 

- eventually, go to most popular packages / macros (dbt_utils) and add argument `type` and `description` annotations
- eventually, hygenic macros 

### Cut from scope

SQL Syntax Hints - In addition to the types specified above, we will also support a special variation on the str type, providing additional metadata in the very common case that a string argument or return type represents a fragment of SQL syntax. 

For example:
```
      - name: column_name
        type: str::sql_identifier
        description: The name of the column you want to convert
```

Regardless of which notation we use, the initial supported "type hint" values will be only for `type: str`:
 - `sql identifier`
 - `sql value expression` (general expression) 
 - `sql bool value expression` (general condition) 
 - `sql query expression` (any SELECT query) 

We may also include other SQL syntax types, guided by a review of the most common pieces of syntax dbt macros work with, and attempting to conservatively use syntactic constructs common to a vast majority of SQL dialects.

peterallenwebb:
Resolves #11274 

### Problem

In the past:
- dbt has not had a controlled vocabulary for the type annotations permitted on macro arguments in YAML
- dbt has not included macro type names in the manifest, even when they could be inferred directly from the jinja definition
- dbt has not validated that the types annotated in YAML match the names and number of types in the jinja definition

### Solution

This PR adds warnings for all of the above problems, gated behind the new `validate_macro_args` behavior flag.

### Checklist


core/dbt/parser/macros.py
```diff
@@ -94,11 +96,21 @@ def parse_unparsed_macros(self, base_node: UnparsedMacro) -> Iterable[Macro]:
...
+
             # get supported_languages for materialization macro
             if block.block_type_name == "materialization":
                 node.supported_languages = get_supported_languages(macro)
             yield node
 
+    def _extract_args(self, macro) -> List[MacroArgument]:
+        try:
+            return list([MacroArgument(name=arg.name) for arg in macro.args])
+        except Exception:
```
QMalcolm, line 111: Do we know in what cases an exception gets raised? If so, can we add a contextual comment about it?

core/dbt/parser/schemas.py
```diff
@@ -1265,4 +1279,101 @@ def parse_patch(self, block: TargetBlock[UnparsedMacroUpdate], refs: ParserRef)
         macro.created_at = time.time()
         macro.meta = patch.meta
         macro.docs = patch.docs
-        macro.arguments = patch.arguments
+
+        if getattr(get_flags(), "validate_macro_args", False):
+            self._check_patch_arguments(macro, patch)
+            macro.arguments = patch.arguments if patch.arguments else macro.arguments
+        else:
+            macro.arguments = patch.arguments
```
QMalcolm, line 1287: Nit: I believe this is logically equivalent to
```
if getattr(get_flags(), "validate_macro_args", False):
  self._check_patch_arguments(macro, patch)

if patch.arguments
  macro.arguments = patch.arguments
```
which I think is easier to follow, but eh. Also, if it's not logically equivalent I'd like to understand under which circumstances to better understand what's going on 🙂

core/dbt/parser/schemas.py
```diff
@@ -1265,4 +1279,101 @@ def parse_patch(self, block: TargetBlock[UnparsedMacroUpdate], refs: ParserRef)
...
+
+    def _check_patch_arguments(self, macro: Macro, patch: ParsedMacroPatch) -> None:
+        if not patch.arguments:
+            return
+
+        for macro_arg, patch_arg in zip(macro.arguments, patch.arguments):
+            if patch_arg.name != macro_arg.name:
+                msg = f"Argument {patch_arg.name} in yaml for macro {macro.name} does not match the jinja definition."
+                fire_event(Note(msg=msg), EventLevel.WARN)
```
QMalcolm, line 1296: Are we intentionally doing `fire_event` here or should we be doing `warn_or_error`?

core/dbt/parser/schemas.py
```diff
@@ -1265,4 +1279,101 @@ def parse_patch(self, block: TargetBlock[UnparsedMacroUpdate], refs: ParserRef)
...
+
+        if len(patch.arguments) != len(macro.arguments):
+            msg = f"The number of arguments in the yaml for macro {macro.name} does not match the jinja definition."
+            fire_event(Note(msg=msg), EventLevel.WARN)
```
QMalcolm, line 1300: Are we intentionally doing fire_event here or should we be doing warn_or_error?

core/dbt/parser/schemas.py
```diff
@@ -1265,4 +1279,101 @@ def parse_patch(self, block: TargetBlock[UnparsedMacroUpdate], refs: ParserRef)
...
+
+        for patch_arg in patch.arguments:
+            arg_type = patch_arg.type
+            if arg_type is not None and arg_type.strip() != "" and not is_valid_type(arg_type):
+                msg = f"Argument {patch_arg.name} in the yaml for macro {macro.name} has an invalid type."
+                fire_event(Note(msg=msg), EventLevel.WARN)
```
QMalcolm, line 1306: Are we intentionally doing fire_event here or should we be doing warn_or_error?

QMalcolm (commented):
So cool to see this! ❤️ 

I have a few comments. Happy to chat through it if need be 🙂

core/dbt/parser/macros.py
peterallenwebb: No. I simply did not want to add any risk at all to this code path if, for instance, the args list might ever be `None`.

core/dbt/parser/schemas.py
peterallenwebb: It is not logically equivalent because (in principle) the marco.arguments might have been set in the past before this code runs, and then overwritten. I don't have any reason to believe that was happening, but there is also no reason to take the risk of changing an existing behavior since, if they were not set, this is indeed logically equivalent.

core/dbt/events/core_types_pb2.py
```diff
@@ -1,12 +1,22 @@
 # -*- coding: utf-8 -*-
 # Generated by the protocol buffer compiler.  DO NOT EDIT!
+# NO CHECKED-IN PROTOBUF GENCODE
 # source: core_types.proto
-# Protobuf Python Version: 5.26.1
```
QMalcolm, line 4: Have we confirmed that this version change is okay?

QMalcolm (commented):
Question on changed protobuf version used for generation

QMalcolm (approved):

PR #11396, opened 2025-03-18
Support `project` as alias for `package` selection method

### LINKED ISSUE #8986: [CT-3314] [Feature] Support `project` as an alias for `package` selection method

### Describe the feature

Let's support `project` as an alias for `package` when selecting nodes from a specific project.

https://docs.getdbt.com/reference/node-selection/methods#the-package-method
https://docs.getdbt.com/docs/collaborate/govern/project-dependencies

If you're in a dbt project that's taken a dependency on another dbt project:
- via "package"-type dependency: this will select all nodes in the package
- via "project"-type dependency: this will select only public models from the upstream project, imported as metadata only

Not doing this, sticking with the current language

### Who will this benefit?

Less confusion for users of dbt Mesh. Why do I need to type `dbt list -s package:upstream_project+`, if I've intentionally chosen to depend on `upstream_project` as a "project" rather than a "package"?

sure!

### Anything else?

https://github.com/dbt-labs/dbt-core/blob/7fddd6e448d9873da282b2ff7835ab39738d9663/core/dbt/graph/selector_methods.py#L437-L442

smeyerre:
Resolves #8986 

### Problem
A user needs to run `dbt list -s package:upstream_project+` even if they have intentionally chosen to depend on `upstream_project` as a "project" rather than a "package". This can lead to confusion.

### Solution
This PR implements support for the `project:` method as an alias for `package:` in node selection, but with a filter applied to only include public models from the upstream project.

  The implementation:
  1. Adds logic to detect when the `project:` method is used
  2. Applies the existing `package:` selection logic
  3. Adds a filter to only return nodes that are marked with `access='public'`

  This provides a more intuitive experience for users of dbt Mesh who establish project-type dependencies. Now, instead of having to use `package:upstream_project+`, users can use `project:upstream_project+` which matches the dependency type they've defined.

Manual testing confirms the feature correctly filters out protected and private models while allowing public models through.

### Checklist

### DIFF
--- a/tests/unit/graph/test_selector_methods.py
+++ b/tests/unit/graph/test_selector_methods.py
@@ -20,6 +20,7 @@ from dbt.graph.selector_methods import (
     MetricSelectorMethod,
     PackageSelectorMethod,
     PathSelectorMethod,
+    ProjectSelectorMethod,
     QualifiedNameSelectorMethod,
     SavedQuerySelectorMethod,
     SemanticModelSelectorMethod,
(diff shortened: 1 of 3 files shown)


smeyerre:
First time contributing here so I'll include my local testing process as well in case I missed anything!

## Test Environment Setup

- Created two test projects:
  - `upstream_project`: Contains models with explicit access controls
  - `my_project`: Depends on the upstream project

## Test Configuration

### upstream_project
Set up models with explicit access controls:
   - `public_model.sql` with `access='public'`
   - `private_model.sql` with `access='private'`
   - Default example models with protected access

### my_project
Created `dependencies.yml` with project-type dependency:
   ```yaml
   projects:
     - name: 'upstream_project'
   ```

## Test Scenarios

### Test 1: Package Selection vs Project Selection

**Command:**
```bash
dbt list -s package:upstream_project+
```

**Result:** 
All models and tests from upstream_project included, regardless of access level.
```
upstream_project.example.my_first_dbt_model
upstream_project.example.my_second_dbt_model
upstream_project.private_model
upstream_project.public_model
upstream_project.example.not_null_my_first_dbt_model_id
upstream_project.example.not_null_my_second_dbt_model_id
upstream_project.example.unique_my_first_dbt_model_id
upstream_project.example.unique_my_second_dbt_model_id
```

**Command:**
```bash
dbt list -s project:upstream_project+
```

**Result:**
Only public model included.
```
upstream_project.public_model
```

**Note:** The project selector correctly filtered out private and protected models, displaying only the public model. Tests are not included as they do not have an access property.

### Test 2: Model Resource Type Verification

**Command:**
```bash
dbt list --resource-type model
```

**Result:**
```
my_project.example.my_first_dbt_model
my_project.example.my_second_dbt_model
upstream_project.example.my_first_dbt_model
upstream_project.example.my_second_dbt_model
upstream_project.private_model
upstream_project.public_model
```

ash2shukla (changes requested):
Thanks for the PR ! and apologies for taking so much time to review !
The functional changes look good, I have some suggestions for improving readability of ut.

smeyerre:
> Thanks for the PR ! and apologies for taking so much time to review ! The functional changes look good, I have some suggestions for improving readability of ut.

No problem, happy to adjust, what do you suggest?

tests/unit/graph/test_selector_methods.py
```diff
@@ -374,6 +375,48 @@ def test_select_package_this(manifest):
     }
 
 
+def test_select_project(manifest, view_model):
```
ash2shukla, line 378: This test can be more readable, I have a few suggestions - 
1. probably move setup to a separate function
2. use add_node, change_node instead of editing nodes dict directly
3. u can use table_model and ext_model directly as a fixture directly from utils/manifest.py

ash2shukla (changes requested):
> No problem, happy to adjust, what do you suggest?

Sorry forgot to submit the review !

PR #11442, opened 2025-03-28
Enable silencing warnings for `UnpinnedRefNewVersionAvailable` messages

### LINKED ISSUE #11221: [Bug] `UnpinnedRefNewVersionAvailable` messages cannot be silenced

### Current Behavior

When leveraging pre-releases for dbt model versioning, dbt-core helpfully prints a message during compilation when a message contains a reference to an unpinned upstream model if that upstream model has a pre-release version. While useful on a small scale, these messages can quickly become noisy if there are multiple models with pre-release versions. Unfortunately, these messages cannot be silenced, which means there is no way to prevent the logging of these messages (a couple of hundred messages in the case of one of my employer's dbt projects).

### Expected Behavior

As a dbt-core user, I would expect that I would be able to silence `UnpinnedRefNewVersionAvailable` messages using the documented warning and error silencing process[^1]. As such, I should be able to add the following flag to my `dbt_project.yml` to silence these messages entirely:

```yml
flags:
  warn_error_options:
    silence:
      - UnpinnedRefNewVersionAvailable
```

[^1]: https://docs.getdbt.com/reference/global-configs/warnings

### Steps To Reproduce

1. Install dbt-core
2. Setup a dbt project
3. Create a versioned dbt model, such that there exists both a latest version of the model and a pre-release version of the model
4. Build the project and observe the presence of an unpinned reference message
5. Add `UnpinnedRefNewVersionAvailable` to the `warn_error_options` flags config.
6.  Build the project and observe the continued presence of an unpinned reference message

### Relevant log output

```shell
$ dbt --warn-error-options '{"silence": ["UnpinnedRefNewVersionAvailable"]}' compile --select [REDACTED]

14:26:31  Running with dbt=1.8.9
14:26:32  Registered adapter: snowflake=1.8.4
14:26:44  Found 5560 models, 149 snapshots, 1 analysis, 8632 data tests, 308 seeds, 1 operation, 1859 sources, 15 exposures, 1221 macros, 12 groups, 4 unit tests
14:26:44  
14:26:51  Concurrency: 8 threads (target='dev')
14:26:51  
14:26:51  While compiling '[MODEL]':
Found an unpinned reference to versioned model '[UPSTREAM_MODEL]' in project '[PROJECT]'.
Resolving to latest version: [UPSTREAM_MODEL].v1
A prerelease version 2 is available. It has not yet been marked 'latest' by its maintainer.
When that happens, this reference will resolve to [UPSTREAM_MODEL].v2 instead.

  Try out v2: {{ ref('[PROJECT]', '[UPSTREAM_MODEL]', v='2') }}
  Pin to  v1: {{ ref('[PROJECT]', '[UPSTREAM_MODEL]', v='1') }}

Compiled node '[MODEL]' is:

[REDACTED]
```

### Environment

```markdown
- OS: MacOS 14.7
- Python: 3.11.5
- dbt: 1.8.9
```

snowflake

### Additional Context

I suspect the cause of the issue is that `UnpinnedRefNewVersionAvailable` inherits from `InfoLevel` and not `WarnLevel`. If this is the case, it should be relatively easy to resolve this issue.

dbeatty10:
Resolves #11221

### Problem

From https://github.com/dbt-labs/dbt-core/issues/11221:

> As a dbt-core user, I would expect that I would be able to silence `UnpinnedRefNewVersionAvailable` messages using the documented warning and error silencing process[^1]. As such, I should be able to add the following flag to my `dbt_project.yml` to silence these messages entirely:
> 
> ```yml
> flags:
>   warn_error_options:
>     silence:
>       - UnpinnedRefNewVersionAvailable
> ```

But the YAML above doesn't work to silence the warnings.

### Solution

From https://github.com/dbt-labs/dbt-core/issues/11221:

> I suspect the cause of the issue is that `UnpinnedRefNewVersionAvailable` inherits from `InfoLevel` and not `WarnLevel`[^2]. If this is the case, it should be relatively easy to resolve this issue.
> 
> [^2]: https://github.com/dbt-labs/dbt-core/blob/3de3b827bfffdc43845780f484d4d53011f20a37/core/dbt/events/types.py#L832

### Checklist


QMalcolm:
So this is _partially_ because the issue wasn't a warning. It's also because the warning isn't fired via warn_error, but instead [fire_event](https://github.com/dbt-labs/dbt-core/blob/7cca8470e028221c583bbe4995663aa86d2b308a/core/dbt/contracts/graph/manifest.py#L233). Interestingly it _was_ the case that only events via `warn_error` were controllable via `--warn-error-options`, however, we're making it the case that all warnings will be handleable via `--warn-error-options`, though gated behind a behavior flag. We'll have to wait for that work to be merged (#11483) in order to solve this completely. And at that point people will have to flip the behavior flag for their project to handle it.

zqureshi:

zqureshi:
Actually, we could review this.

PR #11446, opened 2025-03-28
Fix microbatch model edge case of having only one batch (that failed) but counting as success

### LINKED ISSUE #11390: [Bug] `raise_compiler_error` does not properly fail microbatch models

## Issue Description
When using `raise_compiler_error` in a pre_hook for a microbatch incremental model, the model execution incorrectly reports SUCCESS despite the error being detected. This behavior differs from other materialization strategies like 'append' where errors properly cause the model to fail.

### Expected Behavior

Both models should fail with ERROR status when validation fails in the pre_hook. The microbatch model should not report SUCCESS when a batch fails due to a compiler error.

### Steps To Reproduce

1. Create the following validation macro:
```sql
{% macro validate_param(valid_param) %}
    {% set current_param = config.get('meta')['param'] %}

    {% if execute %}
        {% if current_param != valid_param %}
            {% set exceptions_message = "Invalid param: " ~ current_param ~ ", valid param: " ~ valid_param %}
            {% do exceptions.raise_compiler_error(exceptions_message) %}
        {% endif %}
    {% endif %}
{% endmacro %}
```

2. Create two test models:

**Microbatch model (test1.sql)**:
```sql
{{
    config(
        materialized='incremental',
        incremental_strategy='microbatch',
        unique_key='tmp',
        event_time='tmp',
        begin='2025-03-17',
        lookback=1,
        batch_size='day',
        meta={'param': 'invalid_param'},
        pre_hook=[
            validate_param('param1')
        ]
    )
}}

SELECT current_date as tmp
```

**Append model (test2.sql)**:
```sql
{{
    config(
        materialized='incremental',
        incremental_strategy='append',
        meta={'param': 'invalid_param'},
        pre_hook=[
            validate_param('param2')
        ]
    )
}}

SELECT current_date as tmp
```

3. Run both models:
```
dbt run --select test1 test2 --full-refresh
```

## Current Behavior
The microbatch model (test1) shows an ERROR at the batch level but ultimately reports SUCCESS for the model execution:

```
1 of 2 START sql microbatch model public.test1 ................................. [RUN]
Batch 1 of 1 START batch 2025-03-17 of public.test1 .................................. [RUN]
Batch 1 of 1 ERROR creating batch 2025-03-17 of public.test1 ......................... [ERROR in 0.01s]
1 of 2 OK created sql microbatch model public.test1 ............................ [SUCCESS in 0.02s]
```

The append model (test2) correctly fails with an ERROR as expected:

```
2 of 2 START sql incremental model public.test2 ................................ [RUN]
2 of 2 ERROR creating sql incremental model public.test2 ....................... [ERROR in 0.01s]
```

### Relevant log output

```shell

```

### Environment

```markdown
- dbt version: 1.9.3
- Adapter: postgres 1.9.0
```

_No response_

### Additional Context

This issue makes it difficult to implement proper validation for microbatch models using pre_hooks, as errors are not properly propagated to the model execution status. This could lead to silent failures in production pipelines where invalid configurations are not caught.

QMalcolm:
Resolves #11390 

### Problem
An error can get raised raised [here during compilation](https://github.com/dbt-labs/dbt-core/blob/729caf0d5ec8017b8aa22b8f88db3076cf0bc9e3/core/dbt/compilation.py#L611). This eventually produces a run result for the batch [here](https://github.com/dbt-labs/dbt-core/blob/729caf0d5ec8017b8aa22b8f88db3076cf0bc9e3/core/dbt/task/base.py#L379). Unfortunately, the produced run result doesn't include the batch results. And since there is only one batch being run, this results in the batch results for the microbatch model having no failures, which we count as a success [here](https://github.com/dbt-labs/dbt-core/blob/729caf0d5ec8017b8aa22b8f88db3076cf0bc9e3/core/dbt/task/run.py#L651).

### Solution

Created an override method `MicrobatchBatchRunner.error_result` which ensures the batch information is included in the run result.

### Checklist

### DIFF
--- a/tests/functional/microbatch/test_microbatch.py
+++ b/tests/functional/microbatch/test_microbatch.py
@@ -1282,5 +1282,5 @@ class TestCompilationErrorOnSingleBatchRun(BaseMicrobatchTest):
         }
 
     def test_microbatch(self, project) -> None:
-        _, console_output = run_dbt_and_capture(["run"], expect_pass=True)
+        _, console_output = run_dbt_and_capture(["run"], expect_pass=False)
         assert "Completed with 1 error, 0 partial successes, and 0 warnings" in console_output
--- a/tests/functional/microbatch/test_microbatch.py
+++ b/tests/functional/microbatch/test_microbatch.py
@@ -1,4 +1,5 @@
 import json
+from typing import Dict
 from unittest import mock
 
 import pytest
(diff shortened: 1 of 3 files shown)


tests/functional/microbatch/test_microbatch.py
```diff
@@ -1231,3 +1232,55 @@ def test_microbatch(
...
+
+    @pytest.fixture(scope="class")
+    def macros(self) -> Dict[str, str]:
+        return {
+            "validate_param.sql": validate_param_macro_sql,
+        }
+
+    def test_microbatch(self, project) -> None:
+        _, console_output = run_dbt_and_capture(["run"], expect_pass=False)
+        assert "Completed with 1 error, 0 partial successes, and 0 warnings" in console_output
```
aranke, line 1286: Can we check the run results here instead of relying on logs?

aranke (approved):
One comment, otherwise LGTM

PR #11451, opened 2025-04-01
Stop dismissing workflow runs

emmyoop:
### Problem

Our artifact review check is failing.

https://github.blog/changelog/2025-02-12-notice-of-upcoming-deprecations-and-breaking-changes-for-github-actions/#changes-to-check-run-status-modification

### Solution

Only trigger on PR reviews.  This means it won't fail until there's at least 1 review but will still ensure artifact changes get at least 2 reviews.

Only GH itself can now dismiss workflows.  

### Checklist

### DIFF
--- a/.github/workflows/artifact-reviews.yml
+++ b/.github/workflows/artifact-reviews.yml
@@ -6,60 +6,27 @@
 # multiple reviews on a single PR based on files changed, so we need to enforce this manually.
 
 # **when?**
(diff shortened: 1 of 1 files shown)


emmyoop:
Note that the failing workflow will continue to fail because the trigger I removed is `pull_request_target` which means it runs the workflow currently on `main`.  An admin will be able to bypass it and merge anyways once there is an approval.

zqureshi (approved):
LGTM 🚢

PR #11488, opened 2025-04-11
Add package 'name' to lock file

### LINKED ISSUE #11487: [Impl] Include each package's project `name` in `package-lock.yml`
### Housekeeping

### Short description

dbt already gets the project `name` for each configured package during version/dependency resolution, and checks to make sure there are no duplicates. Then, it throws that information away.

If we instead wrote each package's (project) `name` to `package-lock.yml`, it could remove the need to go back to github or the package hub to find the package name of the dependency listed in the `packages.yml`.

@akbog says:
> It's extremely useful to us for understanding whether the state of dbt packages is correct, and for resolving the package dependencies and load order.

### Acceptance criteria

The immediate need is just to start storing each package's project `name` in `package-lock.yml`, and make sure that dbt works the same.

We may be able to unlock additional optimizations within dbt's package resolver, by plumbing the `name` throughout the various package config classes, but I'd say that's out of scope for the time being.

If possible, let's keep `sha1_hash` consistent, to avoid "busting" the package-lock cache if the `packages` config content has not actually changed.

### Suggested Tests

[`tests/unit/deps/test_deps`](https://github.com/dbt-labs/dbt-core/blob/main/tests/unit/deps/test_deps.py)
[`tests/functional/dependencies/test_dependency_options`](https://github.com/dbt-labs/dbt-core/blob/main/tests/functional/dependencies/test_dependency_options.py)

### Impact to Other Teams

None, if we play our cards right

I don't think so

### Context

_No response_

jtcohen6:
Resolves #11487

### Problem

dbt gets the true project `name` for each configured package (and their dependencies), but it doesn't store that information anywhere, requiring multiple trips to git providers / hub packages.

### Solution

- Write each package's true project `name` in `package-lock.yml`
- Without changing the `sha1_hash` (if possible)

### Checklist


zqureshi:
👀

core/dbt/contracts/project.py
```diff
@@ -42,13 +42,19 @@ class Quoting(dbtClassMixin, Mergeable):
 
 @dataclass
 class Package(dbtClassMixin):
-    pass
+
+    def __post_serialize__(self, data, context: Optional[Dict]):
```
jtcohen6, line 46: The intent is to avoid changing the `sha1_hash` in `package-lock.yml` just because there's now an optional `name` attribute in the contract:
```
{'git': 'https://github.com/dbt-labs/dbt-utils', 'revision': None, 'warn-unpinned': None, 'subdirectory': None, 'unrendered': {'git': 'https://github.com/dbt-labs/dbt-utils'}, 'name': None}
```

But `name` actually _is_ a required field for `TarballPackage`

core/dbt/task/deps.py
```diff
@@ -204,7 +207,7 @@ def run(self) -> None:
         if self.args.add_package:
             self.add()
 
-        # Check lock file exist and generated by the same pacakges.yml
+        # Check lock file exist and generated by the same packages.yml
```
zqureshi, line 210: I was wondering why this is a diff... until I saw it.

core/dbt/contracts/project.py
```diff
@@ -42,13 +42,19 @@ class Quoting(dbtClassMixin, Mergeable):
...
+            data.pop("name")
+            return data
+        return data
 
 
 @dataclass
 class LocalPackage(Package):
     local: str
     unrendered: Dict[str, Any] = field(default_factory=dict)
+    name: Optional[str] = None
```
jtcohen6, line 57: In the `deps` task [here](https://github.com/dbt-labs/dbt-core/blob/7cca8470e028221c583bbe4995663aa86d2b308a/core/dbt/task/deps.py#L231-L236), dbt reads `package-lock.yml` and validates its contents using these contracts. So if `name` is in there, it needs to be an allowed (optional) field in all `Package` contracts.

The alternative would be to strip out `name` when dbt reads that file — but leave it in for `TarballPackage` only? — which feels trickier.

core/dbt/task/deps.py
```diff
@@ -186,13 +186,16 @@ def lock(self) -> None:
...
         # this loop is to create the package-lock.yml in the same format as original packages.yml
         # package-lock.yml includes both the stated packages in packages.yml along with dependent packages
+        renderer = PackageRenderer(self.cli_vars)
         for package in resolved_deps:
-            packages_installed["packages"].append(package.to_dict())
-        packages_installed[PACKAGE_LOCK_HASH_KEY] = _create_sha1_hash(
-            self.project.packages.packages
-        )
+            package_dict = package.to_dict()
+            package_dict["name"] = package.get_project_name(self.project, renderer)
```
zqureshi, line 195: ✅

tests/functional/dependencies/test_dependency_options.py
```diff
@@ -34,9 +34,11 @@ def test_deps_lock(self, clean_start):
         with open("package-lock.yml") as fp:
             contents = fp.read()
 
-        fivetran_package = "- package: fivetran/fivetran_utils\n    version: 0.4.7"
+        fivetran_package = (
```
zqureshi, line 37: nit: I wonder if this would be clearer to read with a heredoc.

zqureshi (approved):
LGTM but I would have @peterallenwebb or @emmyoop rubber stamp since they are doing infra things.

core/dbt/contracts/project.py
zqureshi: This would be great to leave as a comment for the method.

tests/functional/dependencies/test_dependency_options.py
jtcohen6: qq chose comma ça ?
```python
        fivetran_package = """
  - name: fivetran_utils
    package: fivetran/fivetran_utils
    version: 0.4.7
"""

zqureshi: If you also feel that it increases comprehension, I'm 50/50.

core/dbt/contracts/project.py
jtcohen6: I can definitely add a comment. I'm also open to moving this logic elsewhere - perhaps [`_create_sha1_hash`](https://github.com/dbt-labs/dbt-core/blob/7cca8470e028221c583bbe4995663aa86d2b308a/core/dbt/task/deps.py#L42)?

tests/functional/dependencies/test_dependency_options.py
jtcohen6: I do think it's easier to read!

gshank (approved):

zqureshi:
🚀🚀🚀

owlas:
Not sure if intended but this change is backwards incompatible.

If I compile my project with dbt `1.10.*` it will generate a `package-lock.yml` file with the `name` attribute. If I then try to compile that project with dbt `1.9.*` with the new `package-lock.yml` file, it'll throw an error saying:

`The packages.yml file is malformed`

owlas:
Awesome! Thanks @jtcohen6

PR #11549, opened 2025-04-28
feat: Improve Semantic Manifest validation error messages with detailed error info

### LINKED ISSUE #9849: [Feature] Improve Semantic Manifest parsing error message

### Describe the feature

Improve [this](https://github.com/dbt-labs/dbt-core/blob/c95b1ea5e69ce578738bda165d467f2bb86589d6/core/dbt/parser/manifest.py#L491) error message to provide more details as to what is wrong with the Semantic Manifest. This was brought up [here](https://dbt-labs.slack.com/archives/C05K4R7KZ5Z/p1712099434602579) (internal to dbt Labs).

_No response_

### Who will this benefit?

Users who are trying to parse a semantic manifest.

_No response_

### Anything else?

_No response_

crStiv:
Resolves #9849

### Problem

The current error message for Semantic Manifest validation failures is not informative enough. When validation fails, users only see a generic message "Semantic Manifest validation failed" without any details about what specific validation errors occurred. This makes it difficult for users to understand and fix issues with their Semantic Manifest.

### Solution

I've improved the error message by including the specific validation errors in the failure message. Now when validation fails, the error message will contain all the individual validation errors with descriptive messages for each issue detected.

Specifically:
1. Modified the `validate()` method in `SemanticManifest` class to return a list of validation errors instead of just a boolean
2. Updated the code in the manifest parser to use these validation errors in the error message
3. Added tests to verify the improved error handling

This change makes it much easier for users to understand and fix validation errors in their Semantic Manifests.

### Checklist

### DIFF
--- a/core/dbt/contracts/graph/semantic_manifest.py
+++ b/core/dbt/contracts/graph/semantic_manifest.py
@@ -1,4 +1,4 @@
-from typing import List, Optional, Set
+from typing import List, Optional, Set, Tuple, Union
 
 from dbt import deprecations
 from dbt.constants import (
@@ -47,7 +47,7 @@ class SemanticManifest:
     def __init__(self, manifest: Manifest) -> None:
         self.manifest = manifest
 
-    def validate(self) -> bool:
+    def validate(self) -> Union[List[ValidationError], None]:
 
         # TODO: Enforce this check.
         # if self.manifest.metrics and not self.manifest.semantic_models:
@@ -60,7 +60,7 @@ class SemanticManifest:
         #    return False
 
         if not self.manifest.metrics or not self.manifest.semantic_models:
-            return True
+            return None
 
         semantic_manifest = self._get_pydantic_semantic_manifest()
         validator = SemanticManifestValidator[PydanticSemanticManifest]()
@@ -116,7 +116,7 @@ class SemanticManifest:
         for error in validation_result_errors:
             fire_event(SemanticValidationFailure(msg=error.message), EventLevel.ERROR)
 
-        return not validation_result_errors
+        return validation_result_errors if validation_result_errors else None
 
     def write_json_to_file(self, file_path: str):
         semantic_manifest = self._get_pydantic_semantic_manifest()
--- a/tests/unit/contracts/graph/test_semantic_manifest.py
+++ b/tests/unit/contracts/graph/test_semantic_manifest.py
@@ -1,4 +1,4 @@
-from unittest.mock import patch
+from unittest.mock import patch, MagicMock
 
 import pytest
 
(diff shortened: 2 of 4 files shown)


core/dbt/contracts/graph/semantic_manifest.py
```diff
@@ -60,7 +60,7 @@ def validate(self) -> bool:
         #    return False
 
         if not self.manifest.metrics or not self.manifest.semantic_models:
-            return True
+            return None
```
QMalcolm:
```suggestion
            return []
```

QMalcolm (commented):
Hi @crStiv 👋🏻 Thank you for putting together this PR ❤️ Sorry we didn't respond sooner to this PR or been more active on #9849. We _do_ need to improve the logs here, especially around the failures. But I don't think the approach of putting all validation errors in _one_ message is one we want to take. I've [added context](https://github.com/dbt-labs/dbt-core/issues/9849#issuecomment-3544183140) on #9849.

crStiv:
@QMalcolm hi! Forgive me as well. The mix of a lot of work and winter holidays took me off guard, therefore I did it only now, I hope I did everything as it should be

core/dbt/contracts/graph/semantic_manifest.py
```diff
@@ -116,7 +116,7 @@ def validate(self) -> bool:
         for error in validation_result_errors:
             fire_event(SemanticValidationFailure(msg=error.message), EventLevel.ERROR)
 
-        return not validation_result_errors
+        return validation_result_errors if validation_result_errors else None
```
QMalcolm, line 119: Because of the recent changes, I don't believe we need to be returning the validation results anymore, we can go back to returning the boolean value. As such the diff in this file can/should be undone 😅

tests/unit/contracts/graph/test_semantic_manifest.py
```diff
@@ -41,7 +42,37 @@ def test_validate(self, manifest):
         with patch("dbt.contracts.graph.semantic_manifest.get_flags") as patched_get_flags:
             patched_get_flags.return_value.require_yaml_configuration_for_mf_time_spines = True
             sm_manifest = SemanticManifest(manifest)
-            assert sm_manifest.validate()
+            assert sm_manifest.validate() is None
```
QMalcolm, line 45: Same comment to the changes in this test 😅

QMalcolm (changes requested):
Howdy! Sorry for the delay. With the most recent update, some of the other changes are no longer necessary. We should back those out so that we make the smallest change necessary. Otherwise I'm ready to approve 🙂 Thank you again for sticking with this ❤️

PR #11574, opened 2025-05-01
Make codecov informational-only

aranke:
### Problem

It's hard to know if a test failed because Codecov patch often fails.

### Solution

Make Codecov checks informational only.
Codecov docs here: https://docs.codecov.com/docs/quick-start#tips-and-tricks

### Checklist

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -8,10 +8,12 @@ coverage:
       default:
         target: auto
         threshold: 0.1% # Reduce noise by ignoring rounding errors in coverage drops
+        informational: true
     patch:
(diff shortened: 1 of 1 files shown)


codecov.yml
```diff
@@ -8,10 +8,12 @@ coverage:
       default:
         target: auto
         threshold: 0.1% # Reduce noise by ignoring rounding errors in coverage drops
+        informational: true
```
emmyoop, line 11: so easy.

emmyoop (approved):

PR #11589, opened 2025-05-08
Fix #11428: Add tags and meta config to exposures

### LINKED ISSUE #11428: [Bug] Add support for setting tags and meta for exposures in dbt_project.yml

### Current Behavior

i've tested whether the `dbt_project.yml` supports using `tag` and `meta` du to [this docs github issue](https://github.com/dbt-labs/docs.getdbt.com/issues/7079#issuecomment-2748856626),  and it looks like you can set it at a project-level config but it looks like it isn't promoted to the top-level tags field used for model selection. so in order to select the `tags` or `meta` , i need to configure it in the exposures.yml file.

@jtcohen6 also performed further tests with `meta` and `tags` and while the exposure does pick up those configurations, it doesn't apply them to the top-level tags or meta attributes that would enable selection:

```
{"name": "weekly_jaffle_metrics", "resource_type": "exposure", "package_name": "testy", "original_file_path": "models/exposures.yml", "unique_id": "exposure.testy.weekly_jaffle_metrics", "tags": [], "config": {"enabled": true, "tags": ["some_tag"], "meta": {"some_field": "some_value"}}, "depends_on": {"macros": [], "nodes": ["model.testy.my_model"]}}
```

So the only config that fully works by setting it at the project level right now is `enabled`.

### Expected Behavior

Have `meta` and `tags` set it at a project-level config and be promoted to the top-level tags field used for model selection.

### Steps To Reproduce

1. In my latest environment
2. Set `tags` or `meta` in the dbt_project.yml file
3. Run ` dbt run --select tag:tag_name`
4. See logs where `"tags": []` returns nothing

### Relevant log output

```shell

```

### Environment

```markdown
- OS: Mac
- dbt: latest
```

snowflake

### Additional Context

https://github.com/dbt-labs/docs.getdbt.com/issues/7079#issuecomment-2748856626

aranke:
Resolves #11428

### Problem

`meta` and `tags` specified in `dbt_project.yml` weren't flowing down to the exposure

### Solution

Combine `meta` and `tags` between global config in `dbt_project.yml` and local config in `properties.yml`.

### Checklist

### DIFF
--- a/core/dbt/parser/schema_yaml_readers.py
+++ b/core/dbt/parser/schema_yaml_readers.py
@@ -117,8 +117,8 @@ class ExposureParser(YamlReader):
                 f"Calculated a {type(config)} for an exposure, but expected an ExposureConfig"
             )
 
-        tags = sorted(set(self.project.exposures.get("tags", []) + config.tags + unparsed.tags))
-        meta = {**self.project.exposures.get("meta", {}), **config.meta, **unparsed.meta}
+        tags = sorted(set(self.project.exposures.get("tags", []) + unparsed.tags + config.tags))
+        meta = {**self.project.exposures.get("meta", {}), **unparsed.meta, **config.meta}
 
         config.tags = tags
         config.meta = meta
--- a/core/dbt/parser/schema_yaml_readers.py
+++ b/core/dbt/parser/schema_yaml_readers.py
@@ -117,8 +117,8 @@ class ExposureParser(YamlReader):
                 f"Calculated a {type(config)} for an exposure, but expected an ExposureConfig"
             )
 
-        tags = sorted(set(self.project.exposures.get("tags", []) + config.tags))
-        meta = {**self.project.exposures.get("meta", {}), **config.meta}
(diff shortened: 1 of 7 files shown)


core/dbt/parser/schema_yaml_readers.py
```diff
@@ -117,6 +117,12 @@ def parse_exposure(self, unparsed: UnparsedExposure) -> None:
                 f"Calculated a {type(config)} for an exposure, but expected an ExposureConfig"
             )
 
+        tags = sorted(set(self.project.exposures.get("tags", []) + unparsed.tags + config.tags))
+        meta = {**self.project.exposures.get("meta", {}), **unparsed.meta, **config.meta}
```
QMalcolm, line 121: Sweet, if my understanding is corrected, this ordering is such that  `config.x` will take precedence over `unparsed.x` for any key that is defined in both.

QMalcolm (approved):
:shipit:

core/dbt/parser/schema_yaml_readers.py
aranke: yes, correct.

PR #11605, opened 2025-05-12
fix command test

emmyoop:
### Problem

Tests were failing with "command not found" errors for commands that definitely exist.

### Solution

Use `to_list()` instead of `value`

### Checklist

### DIFF
--- a/tests/functional/list/test_commands.py
+++ b/tests/functional/list/test_commands.py
@@ -50,13 +50,12 @@ class TestRunCommands:
 
         shutil.rmtree(f"{project_root}/snapshots")
 
-    @pytest.mark.parametrize("dbt_command", commands)
     def test_run_commmand(
         self,
         happy_path_project,
-        dbt_command,
     ):
-        run_dbt(dbt_command)
+        for command in commands:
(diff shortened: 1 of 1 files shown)


tests/functional/list/test_commands.py
```diff
@@ -29,10 +29,11 @@
     "show",
     "snapshot",
     "freshness",
+    "serve",  # this was actually serving up docs and caused issues locally
```
QMalcolm, line 32: 👍🏻

tests/functional/list/test_commands.py
```diff
@@ -29,10 +29,11 @@
...
 }
 
 # Commands to happy path test
-commands = [command.value for command in Command if command.value not in commands_to_skip]
+commands = [command.to_list() for command in Command if command.value not in commands_to_skip]
```
QMalcolm, line 36: I was confused how `command.to_list()` worked in converting an enum entry into a list like `['build']`, but it does! Cool! Weird, but cool.

QMalcolm (approved):
:shipit:

PR #11596, opened 2025-05-08
fix query comment tests

emmyoop:
### Problem

Fixes broken tests related to https://github.com/dbt-labs/dbt-adapters/pull/1064

### Solution

Default is now None instead of having no default for a bool (False).

### Checklist

- [ ] I have read [the contributing guide](https://github.com/dbt-labs/dbt-core/blob/main/CONTRIBUTING.md) and understand what's expected of me.
- [ ] I have run this code in development, and it appears to resolve the stated issue.
- [ ] This PR includes tests, or tests are not required or relevant for this PR.
- [ ] This PR has no interface changes (e.g., macros, CLI, logs, JSON artifacts, config files, adapter interface, etc.) or this PR has already received feedback and approval from Product or DX.
- [ ] This PR includes [type annotations](https://docs.python.org/3/library/typing.html) for new and modified functions.

PR #1064, opened 2018-10-16
Connection managers

beckjake:
This is a pretty large one, fixes #963, fixes #1035, and fixes #962
Clean up adapters, with an end goal of adapters as plugins in mind:
 - Created a more consistent class hierarchy where bigquery adapter does not derive from postgres, snowflake and postgres have a common ancestor, etc
 - Defined what methods subclasses must override, use python's abstract classes to enforce them
 - Gave the DBWrapper its list of methods to expose via a decorator + metaclass model because I broke those arrays like 20 times during development and wow, what a pain to maintain with subclassing.
 - Split adapters into two components: A connection manager that doesn't understand relations/the manifest/anything about dbt, and the "real" adapter that does.
 - Remove deprecated methods
 - Connections are now managed on a per-adapter-instance basis. As adapters are generally singletons, that should be no different than before. This means that adapters have full control over their connection management,  which seems right.

#### Assorted thoughts
This is based off `dev/stephen-girard`, which I assume will become 0.13.x.

We also may want to add a deprecation warning to 0.12.x for `get_columns_in_table`, and then we could change this branch to remove it entirely. I've replaced it with `get_columns_in_relation`, which takes a `Relation`, and `get_columns_in_table` builds one and passes it in (like `drop`/`rename`/`truncate` did).

One downside of this model is a number of connection-manager methods have to be exposed to the adapter and they feel a bit silly. I considered two alternatives with connection managers: something like the dbwrapper/adapter relationship where a decorator tags methods as being exposed, or making adapters subclasses of their connection managers. I think if we wanted to change things, the latter is better, but thinking about method resolution orders was very difficult. I prefer the clunky but explicit model.

It wouldn't be too hard to just assign one connection per thread, as it is connections and threads are related (connection count is directly related to thread count) but disjoint (good luck mapping connections to threads or vice-versa). We could force that, by making the thread pool responsible for connections, so each thread gets one connection at start time. Then we would never have to track the existence of connections, just their state.

The connection management does leave me with this vague feeling that I should be able to do something like this:

```
with adapter.transaction() as conn:
   # connection created or reopened, begin called on __enter__
   conn.execute(...) # or add_query(), whatever
# commit/rollback + close called as appropriate on __exit__
# connection is now back in the pool, in a closed state
```

As it is I'm never really confident that all connections have been closed properly.

PR #11613, opened 2025-05-12
do not have DbtProject unit tests rely on default QueryCommment settings

MichelleArk:
Resolves # N/A

### Problem

[#1064](https://github.com/dbt-labs/dbt-adapters/pull/1064) caused dbt-core to start failing some tests because it tests off dbt-adapters which were subsequently fixed in [#11596](https://github.com/dbt-labs/dbt-core/pull/11596). However, these test fixes rely on a dbt-adapters bump for changes that should ideally be decoupled.
### Solution

Fixing the tests such that they do not depend on particular settings of QueryComment attribute defaults.

### Checklist

### DIFF
--- a/tests/unit/config/test_project.py
+++ b/tests/unit/config/test_project.py
@@ -402,7 +402,7 @@ class TestProjectInitialization(BaseConfigTest):
         )
         self.assertEqual(project.query_comment.comment, "run by user test")
         self.assertEqual(project.query_comment.append, True)
-    
+
     def test_default_query_comment_append_False(self):
         self.default_project_data.update(
             {
--- a/tests/unit/config/test_project.py
+++ b/tests/unit/config/test_project.py
@@ -350,7 +350,7 @@ class TestProjectInitialization(BaseConfigTest):
 
         assert "Cycle detected" in str(exc.exception)
 
-    def test_query_comment_disabled(self):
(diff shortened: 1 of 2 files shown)


tests/unit/config/test_project.py
```diff
@@ -350,7 +350,7 @@ def test_cycle(self):
 
         assert "Cycle detected" in str(exc.exception)
 
-    def test_query_comment_disabled(self):
```
MichelleArk, line 353: DbtProject itself doesn't have any business logic to set query_comment.append to False if the query-comment is `None` or `""`. It is entirely driven by the QueryComment() contract which is now defined in dbt-adapters. So this test was really just testing empty comment values.

tests/unit/config/test_project.py
```diff
@@ -402,6 +402,30 @@ def test_custom_query_comment_append(self):
         )
         self.assertEqual(project.query_comment.comment, "run by user test")
         self.assertEqual(project.query_comment.append, True)
+    
+    def test_default_query_comment_append_False(self):
```
MichelleArk, line 406: added explicit tests for when dbt-core drives the setting of "append" to False in the project

emmyoop (approved):
This makes more sense than the previous test changes I made!

PR #11628, opened 2025-05-15
Fix #11506: Add freshness config to sources

### LINKED ISSUE #11506: move `freshness` under `config` so you can configure defaults for all sources / models
### Housekeeping

### Short description

Freshness is currently a property, which means defaults can't be configured in the `dbt_project.yml`:

```
sources: 
  - name: ecom
    schema: raw
    description: E-commerce data for the Jaffle Shop
    freshness:
      warn_after:
        count: 24
        period: hour
```

```
models:
  - name: stg_orders
    freshness:
      build_after: 
        count: 4
        period: hour
```

### Acceptance criteria

- We should move `freshness` (for `sources` and `models`) under `config`

```
sources: 
  - name: ecom
    schema: raw
    description: E-commerce data for the Jaffle Shop
    config:
      freshness:
        warn_after:
          count: 24
          period: hour
```

```
models:
  - name: stg_orders
    config:
      freshness:
        build_after: 
          count: 4
          period: hour
```

- And add a deprecation warning for `freshness` as a property (not nested)

aranke:
Implements #11506 

### Problem

`freshness` doesn't live under `config` for sources

### Solution

Merge source freshness from 5 locations (in ascending order of precedence):
1. `dbt_project.yml`
2. `source` top-level key
3. `source` config
4. `table` top-level key
5. `table` config

Fire a deprecation warning if `freshness` is a top-level key (2 and 4).

### Checklist


core/dbt/parser/sources.py
```diff
@@ -154,7 +155,48 @@ def parse_source(self, target: UnpatchedSourceDefinition) -> SourceDefinition:
...
+        try:
+            project_freshness = FreshnessThreshold.from_dict(
+                self.root_project.sources.get("freshness", {})
+            )
+        except ValueError:
+            fire_event(
+                FreshnessConfigProblem(
+                    msg="Could not parse default freshness config in 'dbt_project.yml', ignoring.",
+                )
+            )
```
QMalcolm, line 168: Could we give the user slightly more information on why we couldn't parse the freshness config? I worry that without more context the user might not know what needs to change.

core/dbt/parser/sources.py
```diff
@@ -154,7 +155,48 @@ def parse_source(self, target: UnpatchedSourceDefinition) -> SourceDefinition:
...
+            project_freshness = None
+
+        source_freshness = source.freshness
+        if source_freshness:
+            deprecations.warn(
+                "custom-key-in-object-deprecation",
+                key="freshness",
+                file=target.path,
+                key_path=source.name,
```
QMalcolm, line 177: Can we create a more specific deprecation for this? My concern is that the `custom-key-in-object-deprecation` basically indicates that we don't know what the property is. It might be more useful to the user if we create something like a `freshness-property-moved-deprecation`. Where we give a message like.
```
[WARNING]: Deprecated functionality
Found `freshness` as a top level property of `<resource_name>` in file `<relative_file_path>`. The `freshness` specification should be moved into the `config` of the resource.
```
The idea being, we want to make it clear to the user that defining `freshness` isn't deprecated, but where one specifies it has been.

QMalcolm (changes requested):
Thank you for doing this work ❤️ I think we should take a pass at improving the messaging we're sending out about this to make the end user's life easier 🙂

QMalcolm (approved):
Looks good! I think we just need to add the event to `tests/unit/test_events.py`

britt-allen:
I believe this is creating downstream CI issues for users.

PR #11635, opened 2025-05-16
Extend Eligible Nodes for ConfigSelectorMethod to include Saved Query Nodes

### LINKED ISSUE #11607: [Feature] Extend Eligible Nodes for ConfigSelectorMethod to include Saved Query Nodes

- [x] I have searched the existing issues, and I could not find an existing issue for this feature
- [x] I am requesting a straightforward extension of existing dbt functionality, rather than a Big Idea better suited to a discussion

### Describe the feature

Today, our [`ConfigSelectorMethod`](https://github.com/dbt-labs/dbt-mantle/blob/f7f148f5c91c4dcdf75f7ec4922e3e3bcf8efa97/core/dbt/graph/selector_methods.py#L524C7-L524C27) supports selection from parsed and source nodes only.

As an example, you can do this `dbt ls -s config.meta.contains_pii:true`, but this command won't return any saved queries that have this `meta` key in the config, despite us using configs for other purposes (as [documented](https://docs.getdbt.com/reference/resource-properties/config)).

_No response_

### Who will this benefit?

From a customer directly:

`Our main use case here is tying our business definitions back to actual saved queries/exports/metrics etc. We're hoping to leverage this with dbt Explorer to help our end users understand the full lineage for a given business metric definition (these can be a one-to-many relationship in terms of semantic models, measures, metrics, etc). I'm also hoping to eventually use it in our metric processing jobs, as some of the metadata parsing we're doing could be simplified if we were able to select saved queries and their parents using an inbuilt method vs manually parsing to find it.`

Sure! I believe it *should* be a pretty straightforward extension but looking for guidance on whether or not that's the case before I get going.

### Anything else?

_No response_

trouze:
Resolves #11607 

### Problem

Today, our [ConfigSelectorMethod](https://github.com/dbt-labs/dbt-mantle/blob/f7f148f5c91c4dcdf75f7ec4922e3e3bcf8efa97/core/dbt/graph/selector_methods.py#L524C7-L524C27) supports selection from parsed and source nodes only.

As an example, you can do this `dbt ls -s config.meta.contains_pii:true`, but this command won't return any semantic layer saved queries that have this meta key in the config, despite us using configs for other purposes (as [documented](https://docs.getdbt.com/reference/resource-properties/config)).

### Solution

This PR changes the nodes used by the `ConfigSelectorMethod` to use `all_nodes` method instead of `configurable_nodes`. It's the case that `all_nodes` in that method are actually configurable.

`configurable_nodes` method was only used in the `ConfigSelectorMethod`, so this change renders the method unused so we can remove it.

### Checklist

### DIFF
--- a/.changes/unreleased/Features-20250627-094000.yaml
+++ b/.changes/unreleased/Features-20250627-094000.yaml
@@ -1,0 +1,6 @@
+kind: Features
+body: Let ConfigSelectorMethod search through all nodes to return matching selections
+time: 2025-06-27T09:40:00.485654-05:00
+custom:
+    Author: trouze
+    Issue: "11607"
--- a/core/dbt/graph/selector_methods.py
+++ b/core/dbt/graph/selector_methods.py
@@ -22,7 +22,6 @@ from dbt.contracts.graph.nodes import (
     ManifestNode,
     Metric,
     ModelNode,
-    ResultNode,
     SavedQuery,
     SemanticModel,
     SingularTestNode,
--- a/core/dbt/graph/selector_methods.py
+++ b/core/dbt/graph/selector_methods.py
@@ -536,7 +536,7 @@ class ConfigSelectorMethod(SelectorMethod):
         # search sources is kind of useless now source configs only have
         # 'enabled', which you can't really filter on anyway, but maybe we'll
         # add more someday, so search them anyway.
-        for unique_id, node in self.configurable_nodes(included_nodes):
+        for unique_id, node in self.all_nodes(included_nodes):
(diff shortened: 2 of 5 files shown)


QMalcolm (changes requested):
❤️ This looks great! One thing though, we should add some tests! I think the most appropriate place would be to add them to `tests/functional/selectors/`

trouze:
@QMalcolm added some unit & functional tests!

QMalcolm (approved):
Amazing, thank you so much!

Issue #11606, opened 2025-05-12
update build_after model freshness


graciegoheen:
### Housekeeping

- [x] I am a maintainer of dbt-core

### Short description

update `build_after` for model `freshness` from the existing:

```
      build_after: 
        count: 4
        period: hour
        depends_on: all
```

to 

```
      build_after: 
        count: 4
        period: hour
        updates_on: all
```

### Acceptance criteria

- rename `depends_on` to `updates_on`
- default for `updates_on` should be `any`

FishtownBuildBot:
Opened a new issue in dbt-labs/docs.getdbt.com: https://github.com/dbt-labs/docs.getdbt.com/issues/8148

PR #11644, opened 2025-05-20
Fix #11506: Add freshness config to models

### LINKED ISSUE #11506: move `freshness` under `config` so you can configure defaults for all sources / models

aranke:
Implements #11506 for models

### Problem

`freshness` doesn't live under config for `models`

### Solution

Merge model freshness from 3 locations (in ascending order of precedence):
1. `dbt_project.yml`
2. `model` top-level key
3. `model` config

Fire a deprecation warning if `freshness` is a top-level key (2).

### Checklist


graciegoheen:
Do we actually need a deprecation warning for _model_ freshness? Since this is new as of 1.10?

graciegoheen:
Also - @aranke if it's a small lift, could you make [this](https://github.com/dbt-labs/dbt-core/issues/11606) change as well?

aranke:
>Do we actually need a deprecation warning for model freshness? Since this is new as of 1.10?

Good catch, removing

> Could you make [this](https://github.com/dbt-labs/dbt-core/issues/11606) change as well?

Yup, can do

core/dbt/artifacts/resources/v1/model.py
```diff
@@ -12,15 +12,68 @@
...
+    return current_merged_value
+
+
 @dataclass
 class ModelConfig(NodeConfig):
     access: AccessType = field(
         default=AccessType.Protected,
         metadata=MergeBehavior.Clobber.meta(),
     )
+    freshness: Optional[ModelFreshness] = None
```
QMalcolm, line 76: I can't comment on the relevant line, but on 103 there is a top level freshness property for models. This was recently added, never released, and never documented. Can we safely rip it out?

QMalcolm (approved):
Looks good to me, one comment, but not necessary to address in order to proceed

core/dbt/artifacts/resources/v1/model.py
graciegoheen: we should rip out the top-level-property for freshness for models, and only support it nested under config

PR #11662, opened 2025-05-23
[1.9.latest] Support reading package 'name' in package-lock.yml, but do not write

### LINKED ISSUE #11661: `package-lock.yml` changes in 1.10.0 are not backwards compatible
Not sure if intended but this change: https://github.com/dbt-labs/dbt-core/pull/11488 is not backwards compatible.

If I compile my project with dbt `1.10.*` it will generate a `package-lock.yml` file with the `name` attribute. If I then try to compile that project with dbt `1.9.*` with the new `package-lock.yml` file, it'll throw an error saying:

`The packages.yml file is malformed`

_Originally posted by @owlas in https://github.com/dbt-labs/dbt-core/issues/11488#issuecomment-2900673481_

jtcohen6:
Resolves #11661

### Problem

Starting in dbt-core v1.10, `package-lock.yml` includes the `name` attribute for each package. This causes incompatibility errors if the user then downgrades back to v1.9, or runs the project on different versions in different environments.

### Solution

For v1.9: Support reading `name` in `package-lock.yml`, but do not actually write it

### Checklist


QMalcolm (approved):
Thanks for doing this work ❤️

PR #11673, opened 2025-05-26
no type-related deprecations yet, allow additional properties on colu…

MichelleArk:
Ensure we're not being too noisy on deprecations related to: 
* types (just focus on schema shapes)
* column_types should allow additional properties.


QMalcolm (approved):
🚀 Tysm!

PR #11686, opened 2025-05-28
Update changelog to point to 1.8 and 1.9

b-per:
Minor docs/DX update to point to all previous changelogs in the latest one


emmyoop:
This was done in #11693

PR #11729, opened 2025-06-11
deprecation warnings for --models, --model, -m

### LINKED ISSUE #11561: deprecate --models, -m flag

MichelleArk:
Resolves #11561

### Problem

--models, --model, and -m are getting deprecated! but they aren't straightforward to intercept via click because they are aliases for --select in many commands, and all get represented the same in click's internal params post-parsing.

### Solution

Do a simple check within argv during the rest of deprecation firing when parsing flags.

### Checklist

🎩 
```
❯ dbt compile -m microbatch_input
19:57:16  Running with dbt=1.11.0-a1
19:57:16  [WARNING]: Deprecated functionality
Usage of `--models`, `--model`, and `-m` is deprecated in favor of
`--resource-type model --select`
19:57:16  target not specified in profile 'snowflake', using 'default'
19:57:17  Registered adapter: snowflake=1.9.4
19:57:17  Unable to do partial parsing because saved manifest not found. Starting full parse.
...
19:57:18  [WARNING][DeprecationsSummary]: Deprecated functionality
Summary of encountered deprecations:
- ModelParamUsageDeprecation: 1 occurrence
To see all deprecation instances instead of just the first occurrence of each,
run command again with the `--show-all-deprecations` flag. You may also need to
run with `--no-partial-parse` as some deprecations are only encountered during
parsing.
```

```
❯ dbt compile --models microbatch_input
20:04:52  Running with dbt=1.11.0-a1
20:04:52  [WARNING]: Deprecated functionality
Usage of `--models`, `--model`, and `-m` is deprecated in favor of
`--resource-type model --select`
20:04:52  target not specified in profile 'snowflake', using 'default'
20:04:53  Registered adapter: snowflake=1.9.4
20:04:54  Unable to do partial parsing because saved manifest not found. Starting full parse.
...
20:04:55  [WARNING][DeprecationsSummary]: Deprecated functionality
Summary of encountered deprecations:
- ModelParamUsageDeprecation: 1 occurrence
To see all deprecation instances instead of just the first occurrence of each,
run command again with the `--show-all-deprecations` flag. You may also need to
run with `--no-partial-parse` as some deprecations are only encountered during
parsing.
```

### DIFF
--- a/core/dbt/cli/flags.py
+++ b/core/dbt/cli/flags.py
@@ -399,7 +399,7 @@ class Flags:
         # Handle firing deprecations of CLI aliases separately using argv
         # because click because it difficult to disambiguite which CLI option was used
         # and only preserves the 'canonical' representation.
-        for deprecated_flags, warning in DEPRECATED_FLAGS_TO_WARNINGS.item():
+        for deprecated_flags, warning in DEPRECATED_FLAGS_TO_WARNINGS.items():
             for deprecated_flag in deprecated_flags:
                 if deprecated_flag in sys.argv:
                     warn(warning)
--- a/core/dbt/cli/flags.py
+++ b/core/dbt/cli/flags.py
@@ -387,7 +387,7 @@ class Flags:
                     "Value for `--event-time-start` must be less than `--event-time-end`"
                 )
 
-    def fire_deprecations(self, ctx):
+    def fire_deprecations(self, ctx: Optional[Context] = None):
         """Fires events for deprecated env_var usage."""
         [dep_fn() for dep_fn in self.deprecated_env_var_warnings]
         # It is necessary to remove this attr from the class so it does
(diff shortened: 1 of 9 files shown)


tests/functional/deprecations/test_deprecations.py
```diff
@@ -491,3 +493,57 @@ def test_weo_include_exclude_deprecation(
...
+        assert len(event_catcher.caught_events) == 0
+        run_dbt(
+            ["ls", "--model", "some_model"],
+            callbacks=[event_catcher.catch],
+        )
+        assert len(event_catcher.caught_events) == 1
+
+
+class TestMParamUsageDeprecation:
+    @mock.patch.object(sys, "argv", ["dbt", "ls", "-m", "some_model"])
```
MichelleArk, line 547: not ideal to have to patch this but I don't see a way around and and it's better than having no test at all IMO

core/dbt/cli/flags.py
```diff
@@ -394,6 +396,14 @@ def fire_deprecations(self):
 
         fire_buffered_deprecations()
 
+        # Handle firing deprecations of CLI aliases separately using argv
+        # because click because it difficult to disambiguite which CLI option was used
```
QMalcolm: Nit
```suggestion
        # because click makes it difficult to disambiguite which CLI option was used
```
The wording "because click because" just felt a little odd

QMalcolm (approved):
Thank you for doing this work ❤️ 

Is the drawback of this approach that it only catches CLI invocations, and not programmatic invocations? A programmatic invocation being one that uses [dbtRunner.invoke](https://github.com/dbt-labs/dbt-core/blob/f4988c62e316998ea8f754615ea1c8d705b1985d/core/dbt/cli/main.py#L53)? If so we may want to follow up by checking the list of deprecated flags against the flags passed into `invoke`'s `args` parameter.

MichelleArk:
> If so we may want to follow up by checking the list of deprecated flags against the flags passed into invoke's args parameter.

Yes, that is the case and this is good suggestion! I found a pretty reasonable way to handle programmatic invocations as part of this work and just pushed that up!

QMalcolm (approved):
Approval round two: electric boogaloo

PR #11765, opened 2025-06-24
bring in latest properties yaml and dbt project yaml jsonschemas

MichelleArk:
Resolves #N/A

### Problem

* jsonschemas were missing a few fields / small typos

### Solution

* update jsonschemas for schema yaml and dbt project yaml files
* update model config validation to account for rename
* update happy path fixture (remove undocumented parse functionality) so that deprecations on happy path can be unskipped!

### Checklist

### DIFF
--- a/tests/functional/fixtures/happy_path_project/tests/schema.yml
+++ b/tests/functional/fixtures/happy_path_project/tests/schema.yml
@@ -3,9 +3,6 @@ version: 2
 data_tests:
   - name: t
     description: "Single Data Test in happy path fixture"
-    docs:
-      show: true
-      node_color: "blue"
     config:
       alias: "test_alias"
       database: "dbt"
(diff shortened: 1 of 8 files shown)


tests/functional/fixtures/happy_path_project/tests/schema.yml
```diff
@@ -3,9 +3,6 @@ version: 2
 data_tests:
   - name: t
     description: "Single Data Test in happy path fixture"
-    docs:
```
MichelleArk, line 6: We don't actually document support for the `docs` property in data_tests: https://docs.getdbt.com/reference/resource-configs/docs

I think this was unintentionally supported in dbt-core and currently is not included in the (what I believe to be correct) jsonschemas. I'm leaning towards going with our documentation here unless we get data that indicates docs on data_tests is a widely-relied upon (undocumented) capability.

MichelleArk: going to implement explicit support in a follow-up issue: https://github.com/dbt-labs/dbt-core/issues/11774

QMalcolm (approved):
LGTM 🚀

PR #11793, opened 2025-07-02
update ModelParamUsageDeprecation message

MichelleArk:
Resolves #N/A

Updating description message for --models deprecation warning, as it was confusing for customers applying `--resource-type` to commands that don't support it (e.g. dbt run).

### DIFF
--- a/core/dbt/events/types.py
+++ b/core/dbt/events/types.py
@@ -709,7 +709,9 @@ class ModelParamUsageDeprecation(WarnLevel):
         return "D032"
 
     def message(self) -> str:
-        description = "Usage of `--models`, `--model`, and `-m` is deprecated in favor of `--resource-type model --select`"
+        description = (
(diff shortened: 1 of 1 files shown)


core/dbt/events/types.py
```diff
@@ -709,7 +709,9 @@ def code(self) -> str:
         return "D032"
 
     def message(self) -> str:
-        description = "Usage of `--models`, `--model`, and `-m` is deprecated in favor of `--resource-type model --select`"
+        description = (
+            "Usage of `--models`, `--model`, and `-m` is deprecated in favor of `--select`."
```
jeremyyeo: Oh yeah I think it should include the shortcut...

```
Usage of `--models`, `--model`, and `-m` is deprecated in favor of `--select`, `-s`.
```

:P

jeremyyeo (approved):

QMalcolm (approved):

PR #11795, opened 2025-07-03
Create and protect dbt engine environment variable namespace via prefix `DBT_ENGINE`

### LINKED ISSUE #11340: Deprecate unconstrained user space of environment variables

### LINKED ISSUE #11384: [Feature] Make sure that any new configs/flags (e.g. `DBT_FLAG`) use names/namespaces that are not likely to clash with existing use of the same name

- [x] I have searched the existing issues, and I could not find an existing issue for this feature
- [x] I am requesting a straightforward extension of existing dbt functionality, rather than a Big Idea better suited to a discussion

### Describe the feature

We have on occasion introduced the setting of dbt flags/configs with env vars like `DBT_<SOMETHING>` which caused issues because users themselves are already using that same named env var for something else in their project. This usually materializes itself in broken dbt runs - which would be even worse in our new version-free dbt since we may just introduced a brand new `DBT_NEW_FLAG` all of a sudden...

Most recently, this is `DBT_SAMPLE`, `DBT_TARGET`, etc. 

We should figure out a way moving forward where we pick a name that is unlikely to clash with what users may already be using in their projects - perhaps some pattern that is obscure enough - e.g. `DBT___<SOMETHING>___DBT` - or what have you. Or maybe pattern reservation `DBT__CONFIG__<...>` - release guidance that such pattenrns will always be used for "dbt config/flags"...

_No response_

### Who will this benefit?

_No response_

_No response_

### Anything else?

This has happened more often for env vars that dbt Cloud suddenly decides to reserve as opposed to dbt-core but thought we should be prudent as well in core land :)

QMalcolm:
Resolves #11340
Resolves #11384

### Problem

Any time we added a new dbt engine environment variable using the `DBT_` environment variable (like `DBT_SAMPLE`) we were potentially _breaking_ projects that already has a custom environment variable for their own purposes which the same name. We _need_ to be able to add new environment variables _without_ breaking people's projects.

### Solution

Create a namespace that can _only_ contain dbt engine environment variables. Specifically, the prefix `DBT_ENGINE` is now _reserved_. People can still _set_ these environment variables, but only ones which are defined as existing.

### Checklist

### DIFF
--- a/core/dbt/env_vars.py
+++ b/core/dbt/env_vars.py
@@ -7,6 +7,8 @@ from dbt.deprecations import warn
 from dbt_common.constants import ENGINE_ENV_PREFIX
 
 # These are env vars that are not in the params module, but are still allowed to be set.
+# New additions to this list should use the new naming scheme, unless they are being added because
+# they already existed, but we didn't know about them previously.
 # TODO: Should at least some of these become (undocumented) cli param options?
 _ADDITIONAL_ENGINE_ENV_VARS: List[str] = [
     "DBT_INVOCATION_ENV",
--- a/core/dbt/env_vars.py
+++ b/core/dbt/env_vars.py
@@ -1,10 +1,10 @@
-import os
 from dataclasses import dataclass
 from typing import List, Optional
 
 from dbt.cli import params
 from dbt.deprecations import warn
 from dbt_common.constants import ENGINE_ENV_PREFIX
+from dbt_common.context import get_invocation_context
 
 # These are env vars that are not in the params module, but are still allowed to be set.
 # New additions to this list should use the new naming scheme, unless they are being added because
@@ -51,7 +51,8 @@ def validate_engine_env_vars() -> None:
     """
     Validate that any set environment variables that begin with the engine prefix are allowed.
     """
-    for env_var in os.environ.keys():
+    env_vars = get_invocation_context()._env
+    for env_var in env_vars.keys():
         if env_var.startswith(ENGINE_ENV_PREFIX) and env_var not in _ALLOWED_ENV_VARS:
             warn(
                 "environment-variable-namespace-deprecation",
--- a/core/dbt/env_vars.py
+++ b/core/dbt/env_vars.py
@@ -37,7 +37,7 @@ def _create_engine_env_var(name: str) -> EngineEnvVar:
 
 
 # Here we are creating a set of all known engine env vars. This is used in this moduleto create an allow list of dbt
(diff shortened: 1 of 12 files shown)


core/dbt/env_vars.py
```diff
@@ -0,0 +1,61 @@
...
+]
+
+
+@dataclass(frozen=True)
+class EngineEnvVar:
+    name: str
+    old_name: Optional[str] = None
+
+
+def _create_engine_env_var(name: str) -> EngineEnvVar:
```
MichelleArk: nit: perhaps this would be better-suited as a post_init method on EngineEnvVar

core/dbt/env_vars.py
```diff
@@ -0,0 +1,61 @@
...
+    if name.startswith(ENGINE_ENV_PREFIX):
+        return EngineEnvVar(name=name)
+    elif name.startswith("DBT"):
+        return EngineEnvVar(name=name.replace("DBT", f"{ENGINE_ENV_PREFIX}"), old_name=name)
+    else:
+        raise RuntimeError(
+            f"Invalid environment variable: {name}, this will only happen if we add a new option to dbt that has an envvar that doesn't start with DBT_ or {ENGINE_ENV_PREFIX}"
```
MichelleArk: Sounds like this reflects an issue with the dbt engine itself. Should this be a DbtInternalError instead?

core/dbt/events/types.py
```diff
@@ -722,6 +722,15 @@ def message(self) -> str:
         return line_wrap_message(deprecation_tag(description))
 
 
+class EnvironmentVariableNamespaceDeprecation(WarnLevel):
+    def code(self) -> str:
+        return "D036"
+
+    def message(self) -> str:
+        description = f"Found custom environment variable `{self.env_var}` in the environment. The prefix `{self.reserved_prefix}` reserved for dbt engine environment variables. Custom environment variables with the prefix `{self.reserved_prefix}` may cause collisions and runtime errors."
```
MichelleArk: nit: "The prefix `{self.reserved_prefix}` _is_ reserved..."

MichelleArk (approved):

core/dbt/env_vars.py
QMalcolm: I think I'm actually gonna make this one an `__init__` instead of `__post_init__`. The reason being we're partially determining whether the passed in argument should be the `name` or the `old_name`, and that feels more `__init__`-y

PR #11815, opened 2025-07-10
Make `GenericJSONSchemaValidationDeprecation` a "preview" deprecation

### LINKED ISSUE #11814: [Deprecations] Make `GenericJSONSchemaValidationDeprecation` a `preview` type deprecation
### Housekeeping

### Short description

Currently, there are some inaccuracies with the JSONSchema we use for validation. These are geting raised as a `GenericJSONSchemaValidationDeprecation`. This is causing confusion because it's getting highlighted in the deprecation summary. We should make it a preview type deprecation which will
1. Remove it from the summary
2. Emit it as a `Note` event instead of the actual deprecation event
  a. This does mean it'll still be in the logs

We should additionally update the message for `GenericJSONSchemaValidationDeprecation` to be a little more useful, like call it a "possible deprecation"

### Acceptance criteria

* `GenericJSONSchemaValidationDeprecation` is in "preview"
* `GenericJSONSchemaValidationDeprecation` has better wording

### Suggested Tests

N/A

### Impact to Other Teams

N/A

1.10.latest

### Context

_No response_

QMalcolm:
Resolves #11814

### Problem

Currently, there are some inaccuracies with the JSONSchema we use for validation. These are getting raised as a `GenericJSONSchemaValidationDeprecation` deprecation. This is causing confusion because it's getting highlighted in the deprecation summary, but it's not always clear what to do.

### Solution

Make the deprecation a "preview"

### Checklist

- [X]] This PR has no interface changes (e.g., macros, CLI, logs, JSON artifacts, config files, adapter interface, etc.) or this PR has already received feedback and approval from Product or DX.

### DIFF
--- a/core/dbt/events/types.py
+++ b/core/dbt/events/types.py
@@ -593,9 +593,11 @@ class GenericJSONSchemaValidationDeprecation(WarnLevel):
 
     def message(self) -> str:
         if self.key_path == "":
-            description = f"{self.violation} at top level in file `{self.file}`"
+            description = (
+                f"{self.violation} at top level in file `{self.file}` is possibly a deprecation"
+            )
         else:
-            description = f"{self.violation} in file `{self.file}` at path `{self.key_path}`"
+            description = f"{self.violation} in file `{self.file}` at path `{self.key_path}` is possibly a deprecation"
 
         return line_wrap_message(deprecation_tag(description, self.__class__.__name__))
 
--- a/core/dbt/events/types.py
+++ b/core/dbt/events/types.py
@@ -592,12 +592,12 @@ class GenericJSONSchemaValidationDeprecation(WarnLevel):
         return "D022"
 
     def message(self) -> str:
+        possible_causes = "This generally means that either we failed to catch this as a more specific deprecation type OR our JSONSchema had a regression (and this deprecation was erroneous)."
+
         if self.key_path == "":
-            description = (
-                f"{self.violation} at top level in file `{self.file}` is possibly a deprecation"
-            )
+            description = f"{self.violation} at top level in file `{self.file}` is possibly a deprecation. {possible_causes}"
         else:
-            description = f"{self.violation} in file `{self.file}` at path `{self.key_path}` is possibly a deprecation"
+            description = f"{self.violation} in file `{self.file}` at path `{self.key_path}` is possibly a deprecation. {possible_causes}"
 
         return line_wrap_message(deprecation_tag(description, self.__class__.__name__))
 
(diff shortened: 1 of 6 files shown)


core/dbt/events/types.py
```diff
@@ -593,9 +593,11 @@ def code(self) -> str:
 
     def message(self) -> str:
         if self.key_path == "":
-            description = f"{self.violation} at top level in file `{self.file}`"
+            description = (
+                f"{self.violation} at top level in file `{self.file}` is possibly a deprecation"
```
colin-k-rogers: nit: might be worth adding some more context or just link to a github issue/discussion with more context

tests/functional/deprecations/test_deprecations.py
```diff
@@ -10,6 +10,7 @@
 from dbt import deprecations
 from dbt.cli.main import dbtRunner
 from dbt.clients.registry import _get_cached
+from dbt.deprecations import GenericJSONSchemaValidationDeprecation as GJVD
```
colin-k-rogers: unless this acronym is a new/current pattern we plan to introduce can we just use the full name?

tests/functional/deprecations/test_deprecations.py
```diff
@@ -306,17 +308,27 @@ def models(self):
...
         except:  # noqa
             assert (
                 True
             ), "Expected an exception to be raised, because a model object can't be created with a deprecation_date as an int"
 
-        assert len(event_catcher.caught_events) == 1
-        assert (
-            "1 is not of type 'string', 'null' in file" in event_catcher.caught_events[0].info.msg
-        )
+        if GJVD()._is_preview:
```
colin-rogers-dbt, line 322: if `_is_preview` is being accessed outside of it's class isn't the leading underscore kind of a misnomer and it should just be `is_preview`?

tests/functional/deprecations/test_deprecations.py
QMalcolm: There are two `GenericJSONSchemaValidationDeprecation` classes confusingly. There is the core deprecation class, and the core event type class which maps to a dbt-protos proto. We're importing them _both_ in this file. The import of the other can be found on line 21 of this file. Thus, one of them needs to be aliased

core/dbt/events/types.py
QMalcolm: Added more detail 🙂

tests/functional/deprecations/test_deprecations.py
colin-k-rogers: oh that makesmore sense can we name the class something like `GenericCoreJSONSchemaValidationDeprecation`?

MichelleArk (approved):

tests/functional/deprecations/test_deprecations.py
QMalcolm: I can change the alias to that. I don't think we should rename the class directly, as that would break an established pattern and it's only in testing that we have to import _both_ classes, and only specifically because we've put the core class into "preview"

PR #11825, opened 2025-07-14
raise anyOf type deprecations when missing plus-prefix, otherwise skip

MichelleArk:
Resolves #

### Checklist

### DIFF
--- a/core/dbt/jsonschemas.py
+++ b/core/dbt/jsonschemas.py
@@ -120,24 +120,43 @@ def jsonschema_validate(schema: Dict[str, Any], json: Dict[str, Any], file_path:
                             file=file_path,
                             key_path=key_path,
                         )
-        elif error.validator == "type" and "deprecation_date" not in error_path:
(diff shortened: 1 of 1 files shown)


QMalcolm (dismissed):
This looks reasonable to me!

We may want to sanity check what happens if somebody specifies a _custom_ "plus config" via the dbt_project.yml. That is what happens if we have the following:
```yaml
# dbt_project.yml
...
models:
  my_project:
    +my_custom_config: true
```
Also I'm not sure if we were even handling that properly before 😬 If we:
1. weren't handling it before, we could leave doing so to another PR ⏭️ 
2. were handling it before, then we should ensure we continue to do so 🔍 
3. were handling it before, and still are, then awesome 🚀

MichelleArk:
> We may want to sanity check what happens if somebody specifies a custom "plus config" via the dbt_project.yml. That is what happens if we have the following:

Great call! 🧠 

> weren't handling it before, we could leave doing so to another PR ⏭️
This was the case, but the logic we introduced covers it now! I've added a test case and 🎩'd locally to be sure.

PR #11840, opened 2025-07-16
accept generic test args under 'args'

MichelleArk:
Resolves https://github.com/dbt-labs/dbt-core/issues/11847

* Raise `MissingArgumentsPropertyInGenericTestDeprecation` if non-`config`, non-`column_name` property found at top-level of generic test without `arguments` - behind `require_generic_test_arguments`
* merge `arguments` with existing arguments for backcompat - behind `require_generic_test_arguments`
* raise `ArgumentsPropertyInGenericTestDeprecation` if `arguments` detected as property and `require_generic_test_arguments` is not enabled (legacy behavior)
* bring in jsonschema updates

### Checklist

### DIFF
--- a/core/dbt/events/types.py
+++ b/core/dbt/events/types.py
@@ -747,7 +747,7 @@ class ArgumentsPropertyInGenericTestDeprecation(WarnLevel):
         return "D038"
 
     def message(self) -> str:
-        description = f"Found `arguments` property in test definition of `{self.test_name}` without usage of `require_generic_test_arguments` behavior change flag. The `arguments` property is deprecated for custom usage and will be used to nest keyword arguments in future versions of dbt."
+        description = f"Found `arguments` property in test definition of `{self.test_name}` without usage of `require_generic_test_arguments_property` behavior change flag. The `arguments` property is deprecated for custom usage and will be used to nest keyword arguments in future versions of dbt."
         return line_wrap_message(deprecation_tag(description, self.__class__.__name__))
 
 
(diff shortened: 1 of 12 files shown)


core/dbt/events/types.py
```diff
@@ -742,6 +742,24 @@ def message(self) -> str:
         return line_wrap_message(deprecation_tag(description, self.__class__.__name__))
 
 
+class ArgumentsPropertyInGenericTestDeprecation(WarnLevel):
+    def code(self) -> str:
+        return "D038"
+
+    def message(self) -> str:
```
venkaa28, line 749: This is temporary till we remove the flag from autofix, right?

core/dbt/parser/generic_test_builders.py
```diff
@@ -227,6 +229,20 @@ def extract_test_args(data_test, name=None) -> Tuple[str, Dict[str, Any]]:
...
+
+        # Extract kwargs when they are nested under new 'arguments' property separately from 'config' if require_generic_test_arguments_property is enabled
+        if get_flags().require_generic_test_arguments_property:
+            arguments = test_args.pop("arguments", {})
+            if not arguments and any(k not in ("config", "column_name") for k in test_args.keys()):
+                deprecations.warn(
+                    "missing-arguments-property-in-generic-test-deprecation", test_name=test_name
+                )
+            test_args = {**test_args, **arguments}
+        elif "arguments" in test_args:
```
venkaa28, line 241: same as earlier comment

core/dbt/events/types.py
MichelleArk: No, these deprecations are independent from autofix and don't assume it has been run. The deprecations should get raised:
* any time core encounters deprecated legacy syntax ('arguments') without the flag that requires it (since it will be a reserved property soon). 
* any time core encounters top-level properties that aren't under 'arguments' with the behavior flag enabled

colin-rogers-dbt (approved):

Issue #11075, opened 2024-11-29
[Regression] BigQuery: Unit tests using sources with sharded tables no longer work after CTE naming change


antitoine:
### Is this a regression in a recent version of dbt-core?

- [X] I believe this is a regression in dbt-core functionality
- [X] I have searched the existing issues, and I could not find an existing issue for this regression

### Current Behavior

Unit tests fail due to an invalid subquery name, as soon as the `identifier` contains an asterisk (`*`).

### Expected/Previous Behavior

I expect the CTE name to use the `name` and not the`identifier` field (as before).
Alternatively :
- specify the CTE name
- escape CTE name
- enclose the CTE name in quotes

### Steps To Reproduce

1. Setup a `schema.yml` for referencing a sharded table:
```yml
version: 2

sources:
  - name: my_dataset
    database: my_database
    tables:
      - name: my_table
        identifier: my_table_*
```
2. Use it in a model named `my_model.sql`:
```sql
{{
    config(
        materialized = 'incremental',
        incremental_strategy = 'insert_overwrite',
        on_schema_change = 'append_new_columns'
    )
}}

SELECT t.my_field
FROM {{ source('my_dataset', 'my_table') }} t
```
3. Add unit tests on it:
```yml
version: 2
unit_tests:
  - name: my_failing_cte_test
    description: "Not working test because of CTE name"
    model: my_model
    given:
      - input: source('my_dataset', 'my_table')
        rows:
          - { my_field: 10 }
          - { my_field: 20 }
    expect:
      rows:
          - { my_field: 10 }
          - { my_field: 20 }
```

### Relevant log output

Generated query:
```shell
create or replace table `my-project`.`testing`.`my_failing_cte_test__dbt_tmp__fcd73216-184a-43b9-a5e7-a3759c4ea2f7`
OPTIONS(
  expiration_timestamp=TIMESTAMP_ADD(CURRENT_TIMESTAMP(), INTERVAL 12 hour)
)
as (
  select * from (
    with __dbt__cte__my_table_* as ( -- <<<<<<<<<<<<< ISSUE HERE
...
```

```shell
  An error occurred during execution of unit test 'my_failing_cte_test'. There may be an error in the unit test definition: check the data types.
   Database Error
    Syntax error: Expected keyword AS but got "*" at [19:41].
```

### Environment

```markdown
- OS: Debian Bullseye
- Python: 3.11.3
- dbt (working version): 1.9.0-a1
[...]

PR #11841, opened 2025-07-16
Fix trailing comma bug in unit test CTE injection

byapparov:
Similar Issues Found:
  1. Issue #11075: "Unit tests no longer work after CTE naming change" - CTE naming mismatches
  2. Issue #10754: "Unit tests generate SQL with wrong CTE 'this' reference on incremental model with alias"
  3. Issue #10763: "Incorrect CTE name in unit tests using incremental mode for versioned incremental models"

### Solution

The inject_ctes_into_sql function was always adding a trailing comma after injected CTEs when there was an existing WITH clause, even when there were no actual existing CTEs to separate from. This caused invalid SQL syntax like:
  with injected_cte as (select 1), select * from table

The fix adds logic to distinguish between actual CTE definitions (containing "as" or being non-main SQL keywords) and main SQL statements (SELECT, INSERT, etc.), only adding commas when there are genuine existing CTEs that need separation.

Fixes the trailing comma issue in dbt-core's unit test compilation for incremental models while maintaining backward compatibility.

### Checklist

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/.changes/unreleased/Fixes-20250716-202700.yaml
+++ b/.changes/unreleased/Fixes-20250716-202700.yaml
@@ -1,0 +1,6 @@
+kind: Fixes
+body: Fix trailing comma bug in unit test CTE injection that caused invalid SQL syntax when injecting CTEs into SQL with empty WITH clauses
+time: 2025-07-16T20:27:00.000000-00:00
+custom:
+    Author: byapparov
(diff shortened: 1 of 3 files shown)


byapparov:
This looks like a bug in dbt model code after all

PR #11859, opened 2025-07-25
fix: Properly quote event_time column names in sample mode filters

### LINKED ISSUE #11858: [Bug] Event time filtering with --sample flag doesn't respect column-level quote settings from sources.yml

### Current Behavior

When using the --sample flag with dbt, the event time filtering doesn't respect the column-level quote settings defined in sources.yml. This causes SQL compilation errors when using reserved keywords as column names, even when they are explicitly marked as quote: true. The errors arrive when the column name is in camel case or with spaces.

### Expected Behavior

The generated SQL should respect the quote: true setting and use the column name with quotes:
```
-- Expected SQL filter
where "createdAt" >= '2024-01-01' and "createdAt" < '2024-01-31'
```

### Steps To Reproduce

sources.yml 
```
version: 2

sources:
  - name: source_name
    tables:
      - name: table_name
        columns:
          - name: "createdAt"
            quote: true
        config:
          event_time: "createdAt"
```

dbt command:
`dbt build --target sampling --sample '1 year'`

### Relevant log output

```shell
21:26:36  Failure in model model_name (models/model_name.sql)
21:26:36    Database Error in model model_name (models/model_name.sql)
  column "createdat" does not exist
  LINE 24: ... * from "testdb"."source_name"."table_name" where createdAt ...
                                                                ^
  HINT:  Perhaps you meant to reference the column "table_name.createdAt".
  compiled code at target/run/test_dbt_sampling/models/model_name.sql


Compiled code:

  create  table "testdb"."sampling"."table_na__dbt_tmp"
 
    as
  
  (

select * from (select * from "testdb"."source_name"."table_name" where createdAt >= '2024-07-24 21:26:36.233275+00:00' and createdAt < '2025-07-24 21:26:36.233275+00:00') _dbt_et_filter_subq_table_name;
```

### Environment

```markdown
- OS: MacOS
- Python: 3.11.13
I've seen the issue both on dbt-core and dbt Cloud.
dbt-core version: 1.10.4
dbt Cloud version: 2025.7.24+00c7a0e
Adapter type: PostgreSQL
Adapter Version: 1.9.1post20+f492c919d3bd415bf5065b3cd8cd1af23562feb0
```

postgres

### Additional Context

By modifying the `core/dbt/context/providers.py`, specifically the `resolve_event_time_filter` method from `BaseResolver` class, I managed to get the appropriate compilation.
```
def resolve_event_time_filter(self, target: ManifestNode) -> Optional[EventTimeFilter]:
        event_time_filter = None
        sample_mode = getattr(self.config.args, "sample", None) is not None

        # TODO The number of branches here is getting rough. We should consider ways to simplify
        # what is going on to make it easier to maintain

        # Only do event time filtering if the base node has the necessary event time configs
        if (
            isinstance(target.config, (NodeConfig, SeedConfig, SourceConfig))
            and target.config.event_time
            and isinstance(self.model, (ModelNode, SnapshotNode))
        ):
            quote_config = target.config._extra.get('quote', None)
            if quote_config:
                field_name = f'"{target.config.event_time}"'
            else:
                field_name = target.config.event_time

            # Handling of microbatch models
            if (
                isinstance(self.model, ModelNode)
                and self.model.config.materialized == "incremental"
                and self.model.config.incremental_strategy == "microbatch"
                and self.manifest.use_microbatch_batches(project_name=self.config.project_name)
                and self.model.batch is not None
            ):
                # Sample mode microbatch models
                if sample_mode:
                    start = (
                        self.config.args.sample.start
                        if self.config.args.sample.start > self.model.batch.event_time_start
                        else self.model.batch.event_time_start
                    )
                    end = (
                        self.config.args.sample.end
                        if self.config.args.sample.end < self.model.batch.event_time_end
                        else self.model.batch.event_time_end
                    )
                    event_time_filter = EventTimeFilter(
                        field_name=field_name,
                        start=start,
                        end=end,
                    )

                # Regular microbatch models
                else:
                    event_time_filter = EventTimeFilter(
                        field_name=field_name,
                        start=self.model.batch.event_time_start,
                        end=self.model.batch.event_time_end,
                    )

            # Sample mode _non_ microbatch models
            elif sample_mode:
                event_time_filter = EventTimeFilter(
                    field_name=field_name,
                    start=self.config.args.sample.start,
                    end=self.config.args.sample.end,
                )

        return event_time_filter
```

Note the retrieval of the quote configuration and the rough double quoting at the top of the method, which provides the appropriate compilation.

pablomc87:
Resolves #11858 

# Problem

When using the `--sample` flag with models that have camel case or spaced column names as their `event_time` field, the generated SQL would fail because the column name wasn't properly quoted. This is especially problematic for PostgreSQL which requires quoted identifiers for column names with spaces, special characters or camel case.

# Solution

The fix introduces a robust quoting system that properly handles column-level and source-level quoting configurations:

## Key Improvements

1. **Dedicated Method**: Extracted `_resolve_event_time_field_name()` to centralize the quoting logic
2. **Proper Precedence**: Column-level quote settings take precedence over source-level settings
3. **Leverages Existing Infrastructure**: Uses the `Column.create()` and `column.quoted` functionality
4. **Eliminates Duplication**: Single source of truth for field name resolution across all code paths
5. **Maintainable Design**: Cleaner, more readable code that's easier to maintain

## Technical Details

- **Column-level configuration**: Checks for `column_info.quote` setting first
- **Source-level fallback**: Uses `target.quoting.column` as fallback
- **Proper quoting**: Uses `column.quoted` for quoted names, `column.name` for unquoted
- **Consistent application**: Same logic applied across all three event time filter scenarios

## Changes Made

- Added `_resolve_event_time_field_name()` method in `BaseResolver` class
- Refactored `resolve_event_time_filter()` to use the new method
- Eliminated duplicate quoting logic across three code paths
- Improved code maintainability and readability

## Testing

The fix ensures that:
- Models with simple column names continue to work as before
- Models with camel case or spaced column names now work correctly with the `--sample` flag
- Column-level quote settings are properly respected
- Source-level quote settings are used as fallback
- PostgreSQL adapters can properly handle quoted column names in event time filters

## Breaking Changes

None. This is a backward-compatible fix that only adds proper quoting when needed.

# Checklist

### DIFF
--- a/.changes/unreleased/Fixes-20250725-112903.yaml
+++ b/.changes/unreleased/Fixes-20250725-112903.yaml
@@ -1,0 +1,6 @@
+kind: Fixes
+body: Quoting the event_time field when the configuration says so
+time: 2025-07-25T11:29:03.342884+02:00
+custom:
+    Author: pablomc87
+    Issue: "11858"
--- a/core/dbt/context/providers.py
+++ b/core/dbt/context/providers.py
@@ -264,7 +264,7 @@ class BaseResolver(metaclass=abc.ABCMeta):
         if should_quote:
             return column.quoted
         else:
-            return column.name
+            return target.config.event_time
         
     def resolve_event_time_filter(self, target: ManifestNode) -> Optional[EventTimeFilter]:
         event_time_filter = None
--- a/core/dbt/context/providers.py
+++ b/core/dbt/context/providers.py
@@ -242,9 +242,34 @@ class BaseResolver(metaclass=abc.ABCMeta):
     def resolve_limit(self) -> Optional[int]:
         return 0 if getattr(self.config.args, "EMPTY", False) else None
 
+
+    def _resolve_event_time_field_name(self, target: ManifestNode) -> str:
+        """Get the event time field name with proper quoting based on configuration."""
+        # Default to False for quoting
+        should_quote = False
+        
+        # Check column-level quote configuration first (overrides source-level)
+        for _, column_info in target.columns.items():
+            if column_info.name == target.config.event_time:
+                # Column-level quote setting takes precedence
+                column = Column.create(column_info.name, column_info.data_type if column_info.data_type else '')                
+                if hasattr(column_info, 'quote') and column_info.quote is not None:
+                    should_quote = column_info.quote
+                # Fallback to source-level quote setting
+                elif hasattr(target, 'quoting') and hasattr(target.quoting, 'column') and target.quoting.column is not None:
+                    should_quote = target.quoting.column
+                break
+        
+        # Apply quoting logic
+        if should_quote:
+            return column.quoted
+        else:
+            return column.name
+        
     def resolve_event_time_filter(self, target: ManifestNode) -> Optional[EventTimeFilter]:
         event_time_filter = None
         sample_mode = getattr(self.config.args, "sample", None) is not None
+        field_name = self._resolve_event_time_field_name(target)
 
         # TODO The number of branches here is getting rough. We should consider ways to simplify
         # what is going on to make it easier to maintain
@@ -277,7 +302,7 @@ class BaseResolver(metaclass=abc.ABCMeta):
                         else self.model.batch.event_time_end
                     )
                     event_time_filter = EventTimeFilter(
-                        field_name=target.config.event_time,
+                        field_name=field_name,
                         start=start,
                         end=end,
                     )
@@ -285,7 +310,7 @@ class BaseResolver(metaclass=abc.ABCMeta):
                 # Regular microbatch models
                 else:
                     event_time_filter = EventTimeFilter(
-                        field_name=target.config.event_time,
+                        field_name=field_name,
                         start=self.model.batch.event_time_start,
                         end=self.model.batch.event_time_end,
                     )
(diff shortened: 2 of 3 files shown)


QMalcolm (changes requested):
@pablomc87 Thank you for doing this work! This looks great ❤️ The one thing we need to do is add tests 🙂 

Also of note. This issue highlighted how dbt-core is doing a fair bit of work that the dbt-adapters package should handle. In this instance, we'll likely add some extra column API's such lines 252-261 shouldn't be necessary. However, we leave that for a future execerise for us the maintainers 😅 Primarily because it requires cross repository changes. As such we're going to move forward with this change (as soon as we get tests added)

pablomc87:
@QMalcolm Thank you for your review! I added the tests and needed to improve a bit the method for it to handle some edge cases.
Let me know if this works!

pablomc87:
Sorry about the styling, didn't think about running the pre-commit hook!

QMalcolm (approved):
Thank you again for doing this work!

QMalcolm:
It looks like our `[Enforce Additional Reviews on Artifact and Validations Changes / Validate Additional Reviews](https://github.com/dbt-labs/dbt-core/actions/runs/17278015298/job/49039723511?pr=11859)` action is having issues. I'm going to bypass it and get this merged 🙂

PR #11889, opened 2025-08-04
call adapter.add_catalog_integration during parse_manifest

colin-rogers-dbt:
Discovered as part of improving support in databricks for using write catalogs (see: [this PR for dbt-databricks](https://github.com/databricks/dbt-databricks/pull/1129)) we sometimes need to update the database for a node if there is an external catalog attached to it. That macro is currently called _before_ we actually call `add_catalog_integration` on the adapter so this will always result in an exception from the adapter that the catalog is not found. 

### Checklist

### DIFF
--- a/core/dbt/cli/requires.py
+++ b/core/dbt/cli/requires.py
@@ -428,5 +428,3 @@ def setup_manifest(ctx: Context, write: bool = True, write_perf_info: bool = Fal
         adapter.set_macro_resolver(ctx.obj["manifest"])
         query_header_context = generate_query_header_context(adapter.config, ctx.obj["manifest"])  # type: ignore[attr-defined]
         adapter.connections.set_query_header(query_header_context)
-        for integration in active_integrations:
-            adapter.add_catalog_integration(integration)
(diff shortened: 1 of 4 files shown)


core/dbt/cli/requires.py
```diff
@@ -424,6 +428,5 @@ def setup_manifest(ctx: Context, write: bool = True, write_perf_info: bool = Fal
         adapter.set_macro_resolver(ctx.obj["manifest"])
         query_header_context = generate_query_header_context(adapter.config, ctx.obj["manifest"])  # type: ignore[attr-defined]
         adapter.connections.set_query_header(query_header_context)
-
-    for integration in active_integrations:
-        adapter.add_catalog_integration(integration)
+        for integration in active_integrations:
+            adapter.add_catalog_integration(integration)
```
MichelleArk: is this still necessary given we are doing it as part of parse_manifest below?

colin-k-rogers: good point, we really only need to do this the first time we load the adapter

MichelleArk (approved):

PR #11911, opened 2025-08-11
flip require_generic_test_arguments_property behavior change flag

MichelleArk:
Resolves #n/a

### Problem

* Need to start raising deprecation warnings regarding legacy un-strict generic test syntax

### Solution

* flip `require_generic_test_arguments_property ` to true - this means that by default any custom arguments 
* add additional safety handling in case an existing `arguments` property is not a dictionary.
### Checklist

### DIFF
--- a/.changes/unreleased/Features-20250811-150347.yaml
+++ b/.changes/unreleased/Features-20250811-150347.yaml
@@ -1,0 +1,6 @@
+kind: Features
+body: Flip require_generic_test_arguments_property flag
+time: 2025-08-11T15:03:47.369377+02:00
+custom:
+    Author: michelleark
+    Issue: "11911"
--- a/.changes/unreleased/Features-20250811-150347.yaml
+++ b/.changes/unreleased/Features-20250811-150347.yaml
@@ -1,5 +1,5 @@
 kind: Features
-body: Flip require_generic_test_arguments_property flag
+body: Default require_generic_test_arguments_property flag to True - The 'arguments' property will be parsed as keyword arguments to data tests, if provided
 time: 2025-08-11T15:03:47.369377+02:00
 custom:
     Author: michelleark
--- a/core/dbt/parser/generic_test_builders.py
+++ b/core/dbt/parser/generic_test_builders.py
@@ -233,7 +233,9 @@ class TestBuilder(Generic[Testable]):
         # Extract kwargs when they are nested under new 'arguments' property separately from 'config' if require_generic_test_arguments_property is enabled
         if get_flags().require_generic_test_arguments_property:
             arguments = test_args.pop("arguments", {})
-            if not arguments and any(k not in ("config", "column_name") for k in test_args.keys()):
+            if not arguments and any(
+                k not in ("config", "column_name", "description", "name") for k in test_args.keys()
+            ):
                 deprecations.warn(
                     "missing-arguments-property-in-generic-test-deprecation", test_name=test_name
                 )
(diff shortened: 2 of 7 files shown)


tests/functional/fixtures/happy_path_project/seeds/s.yml
```diff
@@ -3,7 +3,8 @@ seeds:
     description: 'test_description'
     data_tests:
       - expression_is_true:
-          expression: "b = 2"
+          arguments:
+            expression: "b = 2"
```
MichelleArk, line 7: the test was expecting a particular log string and the test raising the arguments-related deprecation was causing it to break.

tests/functional/retry/fixtures.py
```diff
@@ -14,26 +14,29 @@
...
   - name: union_model
     columns:
       - name: sum3
         data_tests:
           - accepted_values:
-              values: [3]
-              quote: false
+              arguments:
+                values: [3]
+                quote: false
```
MichelleArk, line 39: ditto here - test was expecting a failure from --warn-error for a particular reason and new deprecation was not isolating the scenario under test.

emmyoop: I'm slightly surprised there were only 2 cases of this.

.changes/unreleased/Features-20250811-150347.yaml
```diff
@@ -0,0 +1,6 @@
+kind: Features
+body: Flip require_generic_test_arguments_property flag
```
emmyoop: Can you modify this message to be more specific for users as to what this change actually does?  Maybe even just "Defailt require_generic_test_arguments_property flag to true"?

core/dbt/parser/generic_test_builders.py
```diff
@@ -233,11 +233,14 @@ def extract_test_args(data_test, name=None) -> Tuple[str, Dict[str, Any]]:
         # Extract kwargs when they are nested under new 'arguments' property separately from 'config' if require_generic_test_arguments_property is enabled
         if get_flags().require_generic_test_arguments_property:
             arguments = test_args.pop("arguments", {})
-            if not arguments and any(k not in ("config", "column_name") for k in test_args.keys()):
+            if not arguments and any(
+                k not in ("config", "column_name", "description", "name") for k in test_args.keys()
```
emmyoop, line 237: Do we anticipate there being any issues changing the keys while also flipping the flag to true?

emmyoop (changes requested):
Should the [docs](https://docs.getdbt.com/reference/global-configs/behavior-changes) get updated some amount of time before this goes live?  They still say TBD for when it will flip.

tests/functional/retry/fixtures.py
MichelleArk: me too - but specifically these were tests that were parsing exception log contents in assertions and had data tests in the project. so not completely generalized.

core/dbt/parser/generic_test_builders.py
MichelleArk: Nope - this makes the check as to whether to fire a deprecation more precise, and if anything should lead to fewer deprecation warnings.

Either way the back-compat logic is applied to `arguments` (if present + a dictionary)

MichelleArk:
Corresponding docs changes are here: https://github.com/dbt-labs/docs.getdbt.com/pull/7758

Although I'll be going through that PR and adding more updates + examples of the new spec (even though the old one is still very much supported with this flag on)

PR #11916, opened 2025-08-12
upgrade protobuf to 6.0

colin-rogers-dbt:
dependent on:
*  https://github.com/dbt-labs/dbt-common/pull/300
* https://github.com/dbt-labs/dbt-adapters/pull/1259

contributes to solving https://github.com/dbt-labs/dbt-core/issues/11908

### Checklist

- [ ] I have read [the contributing guide](https://github.com/dbt-labs/dbt-core/blob/main/CONTRIBUTING.md) and understand what's expected of me.
- [ ] I have run this code in development, and it appears to resolve the stated issue.
- [ ] This PR includes tests, or tests are not required or relevant for this PR.
- [ ] This PR has no interface changes (e.g., macros, CLI, logs, JSON artifacts, config files, adapter interface, etc.) or this PR has already received feedback and approval from Product or DX.
- [ ] This PR includes [type annotations](https://docs.python.org/3/library/typing.html) for new and modified functions.

PR #11931, opened 2025-08-18
Backport 11916 to 1.7.latest

QMalcolm:
Manually backporting #11916 to 1.7.latest


QMalcolm:
Hmmm instead of installing iteslf it seems to be attempting to install dbt-core==1.7.19 and is therefore getting contradictory protobuf specs 🤔

PR #11945, opened 2025-08-21
first pass

MichelleArk:
Resolves #

### Checklist

### DIFF
--- a/core/dbt/contracts/graph/manifest.py
+++ b/core/dbt/contracts/graph/manifest.py
@@ -1165,6 +1165,7 @@ class Manifest(MacroMethods, dbtClassMixin):
                 selector_id: selector
                 for selector_id, selector in writable_manifest.selectors.items()
             },
+            metadata=writable_manifest.metadata,
         )
 
         return manifest
--- a/core/dbt/artifacts/resources/v1/source_definition.py
+++ b/core/dbt/artifacts/resources/v1/source_definition.py
@@ -25,6 +25,7 @@ class SourceConfig(BaseConfig):
     loaded_at_query: Optional[str] = None
     meta: Dict[str, Any] = field(default_factory=dict, metadata=MergeBehavior.Update.meta())
     tags: List[str] = field(default_factory=list)
+    description: str = ""
 
 
 @dataclass
--- a/core/dbt/graph/selector_methods.py
+++ b/core/dbt/graph/selector_methods.py
@@ -607,10 +607,25 @@ class TestTypeSelectorMethod(SelectorMethod):
                 yield unique_id
 
 
+class StateSelectorManifestUpgrade:
+    def __call__(self, manifest: Manifest) -> None:
+        pass
+
+
+class SourceConfigDescriptionUpgrade(StateSelectorManifestUpgrade):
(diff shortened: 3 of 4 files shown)


tests/functional/fixtures/happy_path_project/dbt_project.yml
```diff
@@ -3,6 +3,7 @@ analysis-paths:
 config-version: 2
 flags:
   send_anonymous_usage_stats: false
+  state_modified_compare_more_unrendered_values: true
```
MichelleArk, line 6: This was necessary to ensure state:modified comparisons were not state-aware since we are comparing between CI & local envs.

tests/functional/fixtures/happy_path_project/models/schema.yml
```diff
@@ -127,7 +127,7 @@ models:
         columns: ["id"]
         expression: "id > 0"
         name: "Check that id is greater than 0"
-    deprecation_date: "2052-05-01"
+    deprecation_date: "2052-05-01 00:00:00-04:00"
```
MichelleArk, line 130: This was necessary because the CI run was using its local timezone (-4:00) while local was not! And it was triggering a state:modified result

core/dbt/parser/base.py
```diff
@@ -401,15 +401,18 @@ def update_parsed_node_config(
...
             if patch_file_id:
                 patch_file = self.manifest.files.get(patch_file_id, None)
                 if patch_file and isinstance(patch_file, SchemaSourceFile):
-                    schema_key = resource_types_to_schema_file_keys[parsed_node.resource_type]
-                    if unrendered_patch_config := patch_file.get_unrendered_config(
-                        schema_key, parsed_node.name, getattr(parsed_node, "version", None)
-                    ):
-                        patch_config_dict = deep_merge(patch_config_dict, unrendered_patch_config)
+                    schema_key = resource_types_to_schema_file_keys.get(parsed_node.resource_type)
+                    if schema_key:
```
MichelleArk, line 409: this should be a super-safe refactor and avoids the key error that was possible before when the `state_modified_compare_more_unrendered_values` was set to true. I had observed the issue with happy path project when turning the flag on.

core/dbt/contracts/graph/manifest.py
```diff
@@ -1170,6 +1170,7 @@ def from_writable_manifest(cls, writable_manifest: WritableManifest) -> "Manifes
                 selector_id: selector
                 for selector_id, selector in writable_manifest.selectors.items()
             },
+            metadata=writable_manifest.metadata,
```
MichelleArk, line 1173: necessary to load the metadata as written instead of defaulting it, which meant that the default dbt_version was always the running version

core/dbt/parser/base.py
```diff
@@ -401,15 +401,18 @@ def update_parsed_node_config(
...
+                        if unrendered_patch_config := patch_file.get_unrendered_config(
```
QMalcolm, line 410: I forgot about the walrus operator, took me a second to realize what was going on 🤦🏻

MichelleArk: :=3

Issue #11984, opened 2025-09-02
[Bug] When using the new format for tags (tags under config), column tests do not inherit the column level tags - causing differences in test selection


Daniel-Wiszowaty:
### Is this a new bug in dbt-core?

- [x] I believe this is a new bug in dbt-core
- [x] I have searched the existing issues, and I could not find an existing issue for this bug

### Current Behavior

When having column level tags, the selection does not get picked up through `dbt test -s "tag:tag"`

### Expected Behavior

The tests do get picked up.

### Steps To Reproduce

```yaml
version: 2

models:
  - name: orders
    columns:
      - name: order_id
        config:
          tags: [my_column_tag] # does not work
        data_tests:
          - unique
```

Run `dbt test -s "tag:my_column_tag"`

`The selection criterion 'tag:my_column_tag' does not match any enabled nodes`

**The solution for now is to have test level tags**

```yaml
version: 2

models:
  - name: orders
    columns:
      - name: order_id
        data_tests:
          - unique:
            config:
              tags: [my_test_tag] # does somehow work but it's 2x code for unique and not_null
```

### Relevant log output

```shell

```

### Environment

```markdown
- OS: MacOS
- Python:
- dbt: 1.10.10 Versionless Cloud
```

### Which database adapter are you using with dbt?

snowflake

### Additional Context

We got the issue after dbt 1.10 & dbt-autofix
[Docs](https://docs.getdbt.com/reference/node-selection/test-selection-examples#run-tests-on-tagged-columns) 

If we remove the `config:` and indentation it seems to work, so something is off with the new config

Possible regression in #11839 ?

jeremyyeo:
I would rightly call this issue:

> When using the new format for tags (`tags` under `config`), column tests do not inherit the column level tags.

```yml
# models/schema.yml
models:
  - name: foo
    columns:
      - name: c1
        tags: ["yet_another_tag"]
        tests:
          - not_null
      - name: c2
        config:
          tags: ["yet_another_tag"]
        tests:
          - not_null
```

* Column `c1` is tagged with the old way.
* Column `c2` is tagged with the new way.

<details>

<summary>Compile project and inspect the test nodes in `manifest.json`...</summary>

```json
    "test.analytics.not_null_foo_c1.22e255e9e3": {
      "database": "postgres",
      "schema": "public_dbt_test__audit",
      "name": "not_null_foo_c1",
      "resource_type": "test",
      "package_name": "analytics",
[...]

Issue #11973, opened 2025-08-28
[Bug] `loaded_at_query` compiles incorrectly when defined in a source's `config` dictionary


ndrewwm:
### Is this a new bug in dbt-core?

- [x] I believe this is a new bug in dbt-core
- [x] I have searched the existing issues, and I could not find an existing issue for this bug

### Current Behavior

In the docs for source freshness, we see an example for the `loaded_at_query` being established at the source level, making use of `{{ this }}` ([link](https://docs.getdbt.com/reference/resource-properties/freshness#loaded_at_query)). This does not appear to function as I'd expect; instead the compiled SQL omits the relation entirely, resulting in a syntax error. The `loaded_at_query` will work as expected if you place it within a table's `config` dict, but this is cumbersome if you have a source that encompasses many tables.

### Expected Behavior

If the `loaded_at_query` is defined under the source's `config` dict, I would expect the query to be compiled consistently for each table declared within the source.

### Steps To Reproduce

Here are some materials I used to reproduce the error using a simple DuckDB sandbox. For this, I created a virtual environment using `venv`, but I've omitted those details.

```bash
pip install "dbt-core==1.10.3"
pip install "dbt-duckdb"

dbt init sandbox
duckdb sandbox/sandbox.duckdb \
  -c "create table my_source as select '2025-08-28 11:00'::datetime as loaded_at;"
```

```yaml
# profiles.yml
sandbox:
  target: sandbox
  outputs:
    sandbox:
      type: duckdb
      path: sandbox.duckdb

---
# models/sources/sources.yml
version: 2

sources:
  - name: raw_data
    config:
      freshness:
        error_after:
          count: 5
          period: minute
      loaded_at_query: "select loaded_at from {{ this }}"
    tables:
      - name: my_source
        columns:
          - name: loaded_at
```

### Relevant log output

```shell
dbt source freshness
17:27:55  Running with dbt=1.10.3
17:27:55  Registered adapter: duckdb=1.9.4
17:27:55  Found 2 models, 4 data tests, 1 source, 429 macros
17:27:55  
17:27:55  Concurrency: 1 threads (target='sandbox')
17:27:55  
17:27:55  1 of 1 START freshness of raw_data.my_source ................................... [RUN]
17:27:55  1 of 1 ERROR freshness of raw_data.my_source ................................... [ERROR in 0.02s]
17:27:55  
17:27:55  Finished running 1 source in 0 hours 0 minutes and 0.24 seconds (0.24s).
[...]

PR #11992, opened 2025-09-05
propagate meta/tags on columns to top-level and config

MichelleArk:
Resolves https://github.com/dbt-labs/dbt-core/issues/11984

### Problem

Adding `tags` / `meta` at the column config-level does not make the it through to `tag` selection for tests.

### Solution

🎩 for selection
```
models:
    name: my_model
    columns:
      - name: customer_id
        description: The unique key of the orders mart.
        data_tests:
          - not_null
          - unique
        config:

          tags: ["test_tag"]
```
```
(core) ➜  jaffle-shop git:(main) ✗ dbt ls --select tag:test_tag

19:39:47  Running with dbt=1.11.0-a1
19:39:47  Registered adapter: postgres=1.9.1-a0
19:39:48  Found 93 models, 7 seeds, 2 operations, 59 data tests, 6 sources, 1 exposure, 19 metrics, 1586 macros, 6 semantic models, 3 saved queries, 3 unit tests
jaffle_shop.marts.not_null_customers_customer_id
jaffle_shop.marts.unique_customers_customer_id
```

🎩 for false positives
In both these situations, running `dbt ls state:modified --no-partial-parse`

1. config/meta top-level to start
* State manifest: with tags/meta at top-level + generated from dbt-core@main
* Run against: no file changes, on this branch

2. config/meta config-level to start
* State manifest: with tags/meta at top-level + generated from dbt-core@main
* Run against: moving tags/meta to config-level, on this branch

I get 'No nodes selected!', as expected. I've also tested this both on model and source columns, as well as added a test scenario in happy_path_project and this is validated through test.

I've confirmed the manifest.json does change such that config-level vs column-level tags/meta are propagated in either direction, neither influences state:modified selection. `column.tags` and `column.meta` does not currently influence state:modified for models or sources (only their own tags/meta do) - even when under config. and changing `tags` on columns also does not influence state:modified selection of the tests on that column.

### Checklist

### DIFF
--- a/core/dbt/parser/schema_generic_tests.py
+++ b/core/dbt/parser/schema_generic_tests.py
@@ -383,7 +383,11 @@ class SchemaGenericTestParser(SimpleParser):
             should_quote = column.quote or (column.quote is None and target_block.quote_columns)
             if should_quote:
                 column_name = get_adapter(self.root_project).quote(column_name)
-            column_tags = list(set(column.tags + column.config.get("tags", [])))
+
+            column_config_tags = column.config.get("tags", [])
+            if isinstance(column_config_tags, str):
+                column_config_tags = [column_config_tags]
+            column_tags = list(set(column.tags + column_config_tags))
 
         block = GenericTestBlock.from_test_block(
             src=target_block,
--- a/tests/functional/test_selection/fixtures.py
+++ b/tests/functional/test_selection/fixtures.py
@@ -56,6 +56,12 @@ models:
           - relationships:
               to: source('my_src', 'my_tbl')
               field: fun
+      - name: fun_2
+        tags: [column_level_tag]
+        config:
+          tags: column_config_level_tag
+        data_tests:
+          - unique
 """
 
 models__model_b_sql = """
--- a/tests/functional/test_selection/test_selection_expansion.py
+++ b/tests/functional/test_selection/test_selection_expansion.py
@@ -75,6 +75,7 @@ class TestSelectionExpansion:
             "source_unique_my_src_my_tbl_fun",
             "source_unique_my_src_my_tbl_fun_2",
             "unique_model_a_fun",
+            "unique_model_a_fun_2",
         ]
 
         self.list_tests_and_assert(select, exclude, expected)
@@ -93,6 +94,7 @@ class TestSelectionExpansion:
             "relationships_model_a_fun__fun__ref_model_b_",
             "relationships_model_a_fun__fun__source_my_src_my_tbl_",
             "unique_model_a_fun",
+            "unique_model_a_fun_2",
         ]
 
         self.list_tests_and_assert(select, exclude, expected)
@@ -109,6 +111,7 @@ class TestSelectionExpansion:
             "cf_a_src",
             "just_a",
             "unique_model_a_fun",
+            "unique_model_a_fun_2",
             "relationships_model_a_fun__fun__ref_model_b_",
             "relationships_model_a_fun__fun__source_my_src_my_tbl_",
         ]
@@ -127,6 +130,7 @@ class TestSelectionExpansion:
             "cf_a_src",
             "just_a",
             "unique_model_a_fun",
+            "unique_model_a_fun_2",
             "source_unique_my_src_my_tbl_fun",
             "source_unique_my_src_my_tbl_fun_2",
             "relationships_model_a_fun__fun__ref_model_b_",
@@ -149,6 +153,7 @@ class TestSelectionExpansion:
             "source_unique_my_src_my_tbl_fun",
             "source_unique_my_src_my_tbl_fun_2",
             "unique_model_a_fun",
+            "unique_model_a_fun_2",
         ]
 
         self.list_tests_and_assert(select, exclude, expected)
@@ -166,6 +171,7 @@ class TestSelectionExpansion:
             "just_a",
             "relationships_model_a_fun__fun__ref_model_b_",
             "relationships_model_a_fun__fun__source_my_src_my_tbl_",
+            "unique_model_a_fun_2",
         ]
 
         self.list_tests_and_assert(select, exclude, expected)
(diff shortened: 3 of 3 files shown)


jeremyyeo:
## On the topic of source column tests appear to be missing out on this fix

Noticed something interesting when attempting to test "state:modified"...

```yaml
# models/schema.yml
# Full project: https://github.com/jeremyyeo/dbt-basic/tree/qa-dbt-core-11992
sources:
  - name: public
    tables:
      - name: source_old_way
        columns:
          - name: id
            tags: ["my_tag"]
            tests:
              - not_null
      - name: source_new_way
        columns:
          - name: id
            config:
              tags: ["my_tag"]
            tests:
              - not_null

models:
  - name: mdl_from_old
    columns:
      - name: id
        tags: ["my_tag"]
        tests:
          - not_null
  - name: mdl_from_new
    columns:
      - name: id
        config:
          tags: ["my_tag"]
        tests:
          - not_null

seeds:
  - name: seed_old
    columns:
      - name: id
        tags: ["my_tag"]
        tests:
          - not_null
  - name: seed_new
    columns:
      - name: id
        config:
          tags: ["my_tag"]
        tests:
          - not_null

snapshots:
  - name: snap_old
    relation: ref('seed_old')
    config:
      unique_key: id
      strategy: check
      check_cols: all
    columns:
      - name: id
        tags: ["my_tag"]
        tests:
          - not_null
  - name: snap_new
    relation: ref('seed_new')
    config:
      unique_key: id
      strategy: check
      check_cols: all
    columns:
      - name: id
        config:
          tags: ["my_tag"]
        tests:
          - not_null
```

4 resource types (source/model/seed/snapshot) - each with a "new way" (tags in config) and an "old way" (tags as top level).

Using `main` branch:

```sh
$ dbt ls -s 'tag:my_tag'
04:13:57  Running with dbt=1.11.0-a1
04:13:57  Registered adapter: postgres=1.9.1-a0
04:13:57  Found 2 models, 2 seeds, 8 data tests, 2 snapshots, 2 sources, 434 macros
analytics.not_null_mdl_from_old_id
analytics.not_null_seed_old_id
analytics.not_null_snap_old_id
analytics.source_not_null_public_source_old_way_id
```

Using `propagate-column-meta-tags` branch:

```sh
$ dbt ls -s 'tag:my_tag'
04:16:53  Running with dbt=1.11.0-a2
04:16:54  Registered adapter: postgres=1.9.1-a0
04:16:54  Unable to do partial parsing because of a version mismatch
04:16:54  Found 2 models, 2 seeds, 8 data tests, 2 snapshots, 2 sources, 434 macros
analytics.not_null_mdl_from_new_id
analytics.not_null_mdl_from_old_id
analytics.not_null_seed_new_id
analytics.not_null_seed_old_id
analytics.not_null_snap_new_id
analytics.not_null_snap_old_id
analytics.source_not_null_public_source_old_way_id
```

^ We can see that we list 7 now - missing out on the source_new_way column test...

> Ps. I rolled the version to `1.11.0a2` cause my venvs were messed up and I couldn't get this PR to be "installed" properly unless I install and set the version to be different to what `main` has (`1.11.0a1`).

Checking the `manifest.json` generated by `1.11.0a2` (fake bump for this `propagate-column-meta-tags` branch):

The source-using-the-new-way's test is still missing the top level tag:

```json
    "test.analytics.source_not_null_public_source_new_way_id.4c9bb957a0": {
...
        "tags": [],
...
      },
      "tags": [],
      "description": "",
```

But (models|seeds|snapshots)-using-the-new-way's test now have that propagated:

```json
    "test.analytics.not_null_mdl_from_new_id.9c4f287810": {
...
        "tags": [],
...
      },
      "tags": ["my_tag"],
      "description": "",
```

## On the topic of "state:modified" false positive

With the caveat of the above issue, it did not appear that this causes any false positives:

```sh
$ dbt ls -s state:modified+ --defer --state target_old

04:45:05  Running with dbt=1.11.0-a2
04:45:05  Registered adapter: postgres=1.9.1-a0
04:45:05  Unable to do partial parsing because saved manifest not found. Starting full parse.
04:45:06  Found 2 models, 2 seeds, 8 data tests, 2 snapshots, 2 sources, 434 macros
04:45:06  The selection criterion 'state:modified+' does not match any enabled nodes
04:45:06  No nodes selected!
```

^ `manifest.json` in `target_old` was generated with `main`.

## Other musings

I'm wondering if this is because "source" is a special type of node which we don't parse (since sources aren't really nodes that dbt "builds" per se (dbt doesn't actually materialize sources unlike models/snapshots/seeds))...  which is linked to #11973 

I put a `breakpoint()` here:

https://github.com/dbt-labs/dbt-core/blob/propagate-column-meta-tags/core/dbt/parser/schema_generic_tests.py#L375-L386

Just to test some stuff. When I exclusively have source tests in my project... I don't seem to hit this breakpoint - but I do hit it with model tests.

MichelleArk:
@jeremyyeo 

>^ We can see that we list 7 now - missing out on the source_new_way column test...
Checking the manifest.json generated by 1.11.0a2 (fake bump for this propagate-column-meta-tags branch):
The source-using-the-new-way's test is still missing the top level tag:

Thank you for doing this validation in addition to the false positive check! Turns out source tests were parsed separately and I added some handling for the tag propagation and confirmed + tested selection is working as expected now.

core/dbt/parser/sources.py
```diff
@@ -262,6 +262,11 @@ def parse_source_test(
         tags_sources = [target.source.tags, target.table.tags]
         if column is not None:
             tags_sources.append(column.tags)
+            if column_config_tags := column.config.get("tags", []):
+                if isinstance(column_config_tags, list):
```
QMalcolm, line 266: Had to sanity check that `isinstance("example_string", list)` would indeed return false 😅 Good news, it does! 

Related NIT: We aren't currently testing in the test suite an instance of the column config tag being just a string instead of a list of strings

QMalcolm (dismissed):
Minor NIT, but good to send it in my opinion 🚀

core/dbt/parser/sources.py
MichelleArk: Good point, added more tests!

PR #12008, opened 2025-09-11
Core handles lifecycle of function nodes

### LINKED ISSUE #11965: [UDFs] Core handles lifecycle of UDFs
### Housekeeping

### Short description

When there is a UDF in the DAG, core needs to work with dbt-adapters to ensure the UDF is created in the data warehouse.

### Acceptance criteria

* Core creates UDFs during `run` and `build` commands

### Suggested Tests

* After doing a `dbt run` in a project with a UDF, the UDF is callable in the data warehouse

### Impact to Other Teams

N/A

No

### Context

_No response_

QMalcolm:
Resolves #11965
related https://github.com/dbt-labs/dbt-adapters/pull/1332

### Problem

Core needs to create / replace function nodes when they're found in the graph.

### Solution

Create a function task runner, and add it to the map of runners during the build task

### Checklist

### DIFF
--- a/core/dbt/task/function.py
+++ b/core/dbt/task/function.py
@@ -79,7 +79,8 @@ class FunctionRunner(CompileRunner):
             status=RunStatus.Success,
             timing=[],
             thread_id=threading.current_thread().name,
-            execution_time=0.0,  # TODO: add execution time
+            # This gets set later in `from_run_result` called by `BaseRunner.safe_run`
+            execution_time=0.0,
             message=str(loaded_result.response),
             adapter_response=loaded_result.response.to_dict(omit_none=True),
             failures=loaded_result.get("failures"),
@@ -92,10 +93,6 @@ class FunctionRunner(CompileRunner):
         context = generate_runtime_function_context(compiled_node, self.config, manifest)
 
         MacroGenerator(materialization_macro, context=context)()
-
-        # TODO: Should we be caching something here?
-        # for relation in self._materialization_relations(result, compiled_node):
-        #     self.adapter.cache_added(relation.incorporate(dbt_created=True))
 
         return self.build_result(compiled_node, context)
 


core/dbt/node_types.py
```diff
@@ -18,6 +18,7 @@
     NodeType.Documentation,
     NodeType.RPCCall,
     NodeType.SqlOperation,
+    NodeType.Function,
```
QMalcolm, line 21: Gates whether the node type is selected for the `CompileTask` (or any subclass of `CompileTask` that doesn't declare its own `get_node_selector`). Doesn't technically gate it for `BuildTask` as the `BuildTask` defines its own `get_node_selector`, and we've done separate work to ensure function node selection.

core/dbt/task/build.py
```diff
@@ -43,6 +44,7 @@ class BuildTask(RunTask):
         NodeType.Unit: test_runner,
         NodeType.SavedQuery: saved_query_runner,
         NodeType.Exposure: exposure_runner,
+        NodeType.Function: function_runner,
```
QMalcolm, line 47: The build task only runs nodes that have a mapped runner for the node type, this ensures functions get executed

core/setup.py
```diff
@@ -74,7 +74,7 @@
         # Minor versions for these are expected to be backwards-compatible
         "dbt-common>=1.27.0,<2.0",
         "dbt-adapters>=1.15.5,<2.0",
-        "dbt-protos>=1.0.346,<2.0",
+        "dbt-protos>=1.0.375,<2.0",
```
QMalcolm, line 77: This bump gets us the new `LogFunctionResult` event type

dev-requirements.txt
```diff
@@ -1,7 +1,7 @@
-git+https://github.com/dbt-labs/dbt-adapters.git@main#subdirectory=dbt-adapters
-git+https://github.com/dbt-labs/dbt-adapters.git@main#subdirectory=dbt-tests-adapter
+git+https://github.com/dbt-labs/dbt-adapters.git@qmalcolm--1268-base-adapter-and-postgres-support-function-lifecycle#subdirectory=dbt-adapters
+git+https://github.com/dbt-labs/dbt-adapters.git@qmalcolm--1268-base-adapter-and-postgres-support-function-lifecycle#subdirectory=dbt-tests-adapter
 git+https://github.com/dbt-labs/dbt-common.git@main
-git+https://github.com/dbt-labs/dbt-adapters.git@main#subdirectory=dbt-postgres
+git+https://github.com/dbt-labs/dbt-adapters.git@qmalcolm--1268-base-adapter-and-postgres-support-function-lifecycle#subdirectory=dbt-postgres
```
QMalcolm, line 4: These are temporary. If we were merging to `main` I'd say we should fix them before merging. However, because we're merging to a feature branch. It's actually safe to leave it as is. We have one more segment of work to do on the feature branch after this. After that is merged, then we should fix this up to point back to `main`

MichelleArk, line 4: presumably the rollout plan here is dbt-adapters first and then we reset these prior to merge! leaving a comment to make sure we don't accidentally forget

core/dbt/task/function.py
```diff
@@ -0,0 +1,124 @@
...
+
+    def build_result(self, compiled_node: FunctionNode, context: Dict[str, Any]) -> RunResult:
+        loaded_result = context["load_result"]("main")
+
+        return RunResult(
+            node=compiled_node,
+            status=RunStatus.Success,
+            timing=[],
+            thread_id=threading.current_thread().name,
+            execution_time=0.0,  # TODO: add execution time
```
MichelleArk: Is this left as a follow-up?

core/dbt/task/function.py
```diff
@@ -0,0 +1,124 @@
...
+
+    def execute(self, compiled_node: FunctionNode, manifest: Manifest) -> RunResult:
+        materialization_macro = self._get_materialization_macro(compiled_node, manifest)
+        self._check_lang_supported(compiled_node, materialization_macro)
+        context = generate_runtime_function_context(compiled_node, self.config, manifest)
+
+        MacroGenerator(materialization_macro, context=context)()
+
+        # TODO: Should we be caching something here?
+        # for relation in self._materialization_relations(result, compiled_node):
```
MichelleArk: I think we have relations that were actually created from the materialization, this shouldn't be necessary as its a relation cache.

dev-requirements.txt
QMalcolm: Yes aaaand. This PR is being a merged to a feature branch, and that feature branch won't be merged until we get one more PR in which adds the `function` macro. As such, the sibling adapters PR is also being merged to a feature branch, because merging that one to main would break the CI in dbt-adapters until we merge the feature branch. Thus, changing these backs will happen on the feature branch before we merge the feature branch

core/dbt/task/function.py
QMalcolm: I should remove the TODO. A separate method on the BaseRunner/CompileRunner updates the execution time before we get to `print_result_line`. I can find which method it is and update the comment to reflect what's happening.

core/dbt/task/function.py
QMalcolm: Okay I'll drop this 🙂

dev-requirements.txt
MichelleArk: okay awesome! Just realized this was to a feature branch 💪

Issue #11690, opened 2025-05-28
[Bug] Showing unnecessary warning for legacy time spine with non-day grain


courtneyholcomb:
### Is this a new bug in dbt-core?

- [x] I believe this is a new bug in dbt-core
- [x] I have searched the existing issues, and I could not find an existing issue for this bug

### Current Behavior

- I had a metricflow_time_spine.sql model
- I configured a metricflow_time_spine.ymlfile appropriately with time_spine config using hour grain
- I got expected deprecation warnings
- I just renamed model and configs to all_days and no more deprecation warnings

### Expected Behavior

There should be no warning once a `time_spine` config is added for the model

### Steps To Reproduce

Add a model like this:
```
models:
- name: metricflow_time_spine
  time_spine:
    standard_granularity_column: date_hour
  columns:
  - name: date_hour
    granularity: hour
```
And run a dbt parse

### Relevant log output

```shell

```

### Environment

```markdown
- OS:
- Python:
- dbt:
```

PR #12018, opened 2025-09-15
Fix legacy time spine deprecation warning logic

courtneyholcomb:
Resolves https://github.com/dbt-labs/dbt-core/issues/11690

### Problem
Users still seeing a deprecation warning for their `metricflow_time_spine` model even when they have added a YAML config for the time spine.

### Solution
Previously, we were checking only the `node.relation_name` in one place, and using `ref_lookup.find()` in another. That mismatched logic was the source of the bug, and this updates the logic to be more unified.
This PR also adds better testing in this area and fixes a potential `UnboundLocalError`.

### Checklist


QMalcolm (approved):
🚀

PR #12029, opened 2025-09-18
Fix: convert kwargs to cli params, add aliases

### LINKED ISSUE #12028: [Bug] On programatic invocations CLI params with type checks cause failures

### Current Behavior

On programatic invocations, i.e. using `dbtRunner.invoke()`, the runner fails when e.g. a `profiles_dir` is specified. This appears to be because, when this is provided as a keyword argument (e.g. `dbtRunner.invoke(["build"], profiles_dir='./my_profiles')` the kwargs are only passed to the context _after_ creation, meaning any validation on creating the context fails. And this is exactly what happens, because the profiles_dir option checks for `type=click.Path(exists=True)`.

Why is this annoying? Because everytime you run this in a Docker container (or e.g. through an orchestrator) it will fall back to the default directory for this option (`~/username/.dbt`) which it of course doesn't find.

### Expected Behavior

Any additional parameters provided as kwargs to `dbtRunner.invoke()` should be added (and validated) on creation of the Click context.

### Steps To Reproduce

1. Make sure the default directory `~/YourName/.dbt` does not exists.
2. Run the following
```python
dbtRunner.invoke(["build"], profiles_dir='./my_profiles')
```

### Relevant log output

```shell

```

### Environment

```markdown
- OS: MacOS
- Python: 3.12
- dbt: 1.9
```

_No response_

### Additional Context

On [this line the context is created](https://github.com/dbt-labs/dbt-core/blob/main/core/dbt/cli/main.py#L55), but params are only set after

dumkydewilde:
Resolves #12028

### Problem
On programatic invocations, i.e. using dbtRunner.invoke(), the runner fails when e.g. a profiles_dir is specified. This appears to be because, when this is provided as a keyword argument (e.g. dbtRunner.invoke(["build"], profiles_dir='./my_profiles') the kwargs are only passed to the context after creation, meaning any validation on creating the context fails. And this is exactly what happens, because the profiles_dir option checks for type=click.Path(exists=True).

Why is this annoying? Because everytime you run this in a Docker container (or e.g. through an orchestrator) it will fall back to the default directory for this option (~/username/.dbt) which it of course doesn't find.

### Solution
Params can not be added to the Click context before it is created, which means they have to be added as string arguments. `flags.py` already provides functionality for this with the `command_params()` function. What we can do is take the kwargs through this function, and combine them with provided args so that params can be set _before_ the context creation and validation happens. 

### Checklist

### DIFF
--- a/.changes/unreleased/Fixes-20250918-152852.yaml
+++ b/.changes/unreleased/Fixes-20250918-152852.yaml
@@ -1,0 +1,6 @@
+kind: Fixes
+body: On programmatic invocations set CLI parameters before creating the context to prevent failure due to type validations.
+time: 2025-09-18T15:28:52.584487+02:00
+custom:
+    Author: dumkydewilde
+    Issue: "12028"
--- a/core/dbt/cli/flags.py
+++ b/core/dbt/cli/flags.py
@@ -538,6 +538,7 @@ def command_args(command: CliCommand) -> ArgsList:
         CliCommand.DEPS: cli.deps,
         CliCommand.INIT: cli.init,
         CliCommand.LIST: cli.list,
+        CliCommand.LS: cli.list,
         CliCommand.PARSE: cli.parse,
         CliCommand.RUN: cli.run,
         CliCommand.RUN_OPERATION: cli.run_operation,
--- a/core/dbt/cli/main.py
+++ b/core/dbt/cli/main.py
@@ -15,6 +15,8 @@ from dbt.cli import params as p
 from dbt.cli import requires
 from dbt.cli.exceptions import DbtInternalException, DbtUsageException
 from dbt.cli.requires import setup_manifest
+from dbt.cli.flags import command_params
+from dbt.cli.types import Command
 from dbt.contracts.graph.manifest import Manifest
 from dbt.mp_context import get_mp_context
 from dbt_common.events.base_types import EventMsg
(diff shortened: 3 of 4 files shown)


QMalcolm:
@cla-bot check

QMalcolm (changes requested):
Sorry for taking so long to get to this. We do need to fix the problem, but unfortunately I don't think we can move forward with this PR as is. The party stopper here is 
```python
            cli_params = command_params(command=Command.from_str(args[0]), args_dict=kwargs)
```
Specifically, the `Command.from_str(args[0])` portion. This assumes that the command will be the first argument in the list, and that is not guaranteed. I was actually kind of surprised by this. I myself assumed the command had to come first, but it does not. We actually do this a fair bit in the tests (which is how I discovered the problem). If we were to make the change as is, there is no telling how many projects we'd break unfortunately.

As I see it there are two paths forward:
1. We figure out a way to make this fix without the assumption that the command is the first item in the args
2. We put this behind a behavior flag like `require_command_is_first_arg`. Default it to `False`. When `False`, issue a deprecation warning telling the user how to update their execution command args (and they don't get the fix). When `True` do the fixed behavior.

graciegoheen:
> This assumes that the command will be the first argument in the list, and that is not guaranteed. I was actually kind of surprised by this. I myself assumed the command had to come first, but it does not. We actually do this a fair bit in the tests (which is how I discovered the problem). If we were to make the change as is, there is no telling how many projects we'd break unfortunately.

@QMalcolm This is something we [document](https://docs.getdbt.com/reference/global-configs/command-line-options) as "legacy behavior". I wonder if we should start issuing deprecation warnings for it and introduce a behavior change flag?

Screenshot 2026-02-23 at 3 00 02 PM

QMalcolm:
> @QMalcolm This is something we [document](https://docs.getdbt.com/reference/global-configs/command-line-options) as "legacy behavior". I wonder if we should start issuing deprecation warnings for it and introduce a behavior change flag?

Ah! If we're documenting it as legacy behavior, yea I think the correct path forward would be a behavior change flag + deprecation warning (i.e. option 2). @dumkydewilde Are you comfortable doing such work? If not let me know and I can get this across the line.

PR #12031, opened 2025-09-18
Add a new macro `function()` to the jinja context for using functions

### LINKED ISSUE #11972: [UDFs] New `function` macro for utilizing functions in other nodes
### Housekeeping

### Short description

We _shouldn't_ be using `ref` to utilize functions in other nodes. A function is inherently _different_ from a table/view/etc. A function is a callable in the data warehouse, a table/view is a relational object with columns in the data warehouse. As such, functions _should_ not ref-able. Instead, they should have a separate macro `function` for getting functions. This makes it clearer to the user what is happening, "are you getting a table or a function", and it makes it easier for us to do appropriate compilation. 

### Acceptance criteria

* `ref` does not work for getting functions
* `function` macro is available for getting functions

### Suggested Tests

* `ref` does not work for getting functions
* `function` macro is available for getting functions

### Impact to Other Teams

N/A

No

### Context

_No response_

QMalcolm:
Resolves #11972

### Problem

We need a way for functions to be gettable in the context of other nodes. However, functions _should not_ be `ref`-able. They shouldn't be ref-able because they are not tables/views/etc in the data warehouse AND ref-able things support doing things like `sample` and `limit` whereas references to functions _cannot_. 

### Solution

* Create a new function macro
* Manifest now has `functions` property
* Nodes now have `functions` properties
* Functions get added to the `depends_on` of a node

### Examples

In a model
```sql
# my_model.sql

SELECT id, {{ function("my_function") }}(price) FROM {{ ref("products") }}
```

Inline query
```bash
dbt show --inline "SELECT {{ function('my_function') }}(2)"
```

### Checklist


core/dbt/artifacts/schemas/manifest/v12/manifest.py
```diff
@@ -163,6 +163,10 @@ class WritableManifest(ArtifactMixin):
             description="The unit tests defined in the project",
         )
     )
+    functions: Mapping[UniqueID, Function] = field(
+        default_factory=dict,
```
QMalcolm, line 167: I'm not sure if this is actually how we want to do this. However, without doing this, the `WritableManifest` will break on loading of older manifests that don't have `functions`. The alternative to this would be to bump the schema version (v12 -> v13) and create an upgrade in `upgrade_manifest.py`.

gshank: Which part are you not sure that we want to do? Putting "functions" in as a separate dictionary as opposed to putting functions in the "nodes" dictionary?

QMalcolm: Setting the `default_factory`. That's not something we've done for any other property on the `WritableManifest`

core/dbt/artifacts/resources/v1/exposure.py
```diff
@@ -48,4 +48,5 @@ class Exposure(GraphResource):
     refs: List[RefArgs] = field(default_factory=list)
     sources: List[List[str]] = field(default_factory=list)
     metrics: List[List[str]] = field(default_factory=list)
+    functions: List[List[str]] = field(default_factory=list)
```
gshank: Exposures shouldn't need a function list. Is this here to solve some problem?

core/dbt/artifacts/resources/v1/metric.py
```diff
@@ -156,6 +156,7 @@ class Metric(GraphResource):
     depends_on: DependsOn = field(default_factory=DependsOn)
     refs: List[RefArgs] = field(default_factory=list)
     metrics: List[List[str]] = field(default_factory=list)
+    functions: List[List[str]] = field(default_factory=list)
```
gshank: Same question for metrics - is it even possible for metrics to have a function call?

core/dbt/artifacts/schemas/manifest/v12/manifest.py
gshank: I think we don't need the default_factory because the writable manifest is only created in one place where it's basically copied from the regular Manifest, the "def writable_manifest(..)" method in core/dbt/contracts/graph/manifest.py.

gshank (changes requested):
Mostly fine, just a couple of questions.

core/dbt/artifacts/schemas/manifest/v12/manifest.py
QMalcolm: The issue arises when loading a manifest that was created before `functions` was added. We load the previous manifest as a [WritableManifest in state.py](https://github.com/dbt-labs/dbt-core/blob/faeee357b1e5d366160e6de7d112ac6ebbbbcf3d/core/dbt/contracts/state.py#L41). Only after that do we then create a regular [manifest from the WritableManifest](https://github.com/dbt-labs/dbt-core/blob/faeee357b1e5d366160e6de7d112ac6ebbbbcf3d/core/dbt/contracts/state.py#L42).

Without setting the `default_factory` we get an error like `WritableManifest missing property 'functions'` in that the manifest file we're loading from doesn't specify a `functions` property. One of the ways to solve this is to do what we have here, a `default_factory`. The other way would be to bump the manifest version to `v13` and then add upgrade logic [here](https://github.com/dbt-labs/dbt-core/blob/faeee357b1e5d366160e6de7d112ac6ebbbbcf3d/core/dbt/artifacts/schemas/upgrades/upgrade_manifest.py#L125) such that `functions` gets defaulted to an empty dict on manifests loaded with a version less than `v13`.

core/dbt/artifacts/resources/v1/exposure.py
QMalcolm: They shouldn't. I was worried that not having them might cause some issues because although the function parser should never be called for exposures, the function parser does technically exist when parsing any node type based on our parser architecture. I'll test that nothing complains (mypy or otherwise) if exposures don't have the property.

core/dbt/artifacts/resources/v1/metric.py
QMalcolm: Ditto what I said to the exposures question

core/dbt/artifacts/resources/v1/exposure.py
QMalcolm: Removing it should be safe, pushed commit 🙂

core/dbt/artifacts/resources/v1/metric.py
QMalcolm: Removing it should be safe, pushed commit 🙂

core/dbt/contracts/graph/manifest.py
```diff
@@ -1397,6 +1451,29 @@ def resolve_source(
...
+        )
+        disabled: Optional[List[FunctionNode]] = None
+        for package in package_candidates:
+            function = self.function_lookup.find(target_function_name, package, self)
+            if function is not None and function.config.enabled:
+                return function
+
+            # it's possible that the function is disabled
+            if disabled is None:
+                disabled = self.disabled_lookup.find(f"{target_function_name}", package)
```
MichelleArk: super nit: f-string isnt necessary here

core/dbt/parser/manifest.py
```diff
@@ -2074,6 +2089,49 @@ def _process_sources_for_node(manifest: Manifest, current_project: str, node: Ma
...
+
+        target_function = manifest.resolve_function(
+            target_function_name,
+            target_function_package,
+            current_project,
+            node.package_name,
+        )
+
+        if target_function is None or isinstance(target_function, Disabled):
+            node.config.enabled = False
```
MichelleArk, line 2120: Am I understanding correctly that we are marking the referencing node as disabled if its referencing a function that does not exist or is disabled? Why would that be necessary?

core/dbt/artifacts/schemas/manifest/v12/manifest.py
MichelleArk: Is there any downside to providing a default value here? It seems reasonable to me and the way that we introduce minor schema evolutions to new node-level properties.

QMalcolm: In a technical sense, I don't believe there is a downside. The main thing is that traditionally, this is not how we've done it. I personally think this is a better/easier approach that what've done traditionally

core/dbt/parser/manifest.py
QMalcolm: This is the same pattern we use in `_process_sources_for_node` as well as `_process_refs`. That is to say, this is standard. I understood it to be a node that depends on a disabled node cascades. However, the comment context on the original definition is
``` python 
# This may raise. Even if it doesn't, we don't want to add
# this node to the graph b/c there is no destination node
```

MichelleArk: Thank you!

MichelleArk (approved):

core/dbt/artifacts/schemas/manifest/v12/manifest.py
gshank: Yeah, the traditional approach is that way because somebody did it that way 6 years ago and it's been copied ever since.

PR #12047, opened 2025-09-24
Add run_started_at to manifest.json metadata

MichelleArk:
Adding `run_started_at` to manifest.metadata

### Checklist

### DIFF
--- a/.changes/unreleased/Features-20250924-130923.yaml
+++ b/.changes/unreleased/Features-20250924-130923.yaml
@@ -1,0 +1,6 @@
+kind: Features
+body: Adding run_started_at to manifest.json metadata
+time: 2025-09-24T13:09:23.50087-04:00
+custom:
+    Author: michelleark
+    Issue: "12047"


core/dbt/config/runtime.py
```diff
@@ -303,6 +303,9 @@ def get_metadata(self) -> ManifestMetadata:
                 identifier=self.quoting.get("identifier", None),
                 column=self.quoting.get("column", None),
             ),
+            run_started_at=(
+                tracking.active_user.run_started_at if tracking.active_user is not None else None
```
colin-k-rogers, line 307: probably a silly question but are we sure `run_started_at` is always populated if there is an `active_user`?

colin-rogers-dbt (approved):

core/dbt/config/runtime.py
MichelleArk: Not a silly question, and yes I believe so! [`active_user` is a `User` instance:](https://github.com/dbt-labs/dbt-core/blob/faeee357b1e5d366160e6de7d112ac6ebbbbcf3d/core/dbt/tracking.py#L192), [which is initialized with a `run_started_at` property:  ](https://github.com/dbt-labs/dbt-core/blob/faeee357b1e5d366160e6de7d112ac6ebbbbcf3d/core/dbt/tracking.py#L133)

PR #12074, opened 2025-10-06
Partial Parsing support for function nodes

### LINKED ISSUE #12072: [UDFs] Partial parsing support for functions
### Housekeeping

### Short description

Functions don’t current support partial parsing. If an attempt is made to partial parse them, an error is raised.

### Acceptance criteria

Explicit partial parsing support for functions 

### Suggested Tests

Can partial parse a function

### Impact to Other Teams

N/A

No

### Context

_No response_

QMalcolm:
Resolves #12072 

### Problem

Functions previously broke if they were included during partial parsing 🫠 

### Solution

Plumb functions through partial parsing! 🚀 

### Checklist

### DIFF
--- a/core/dbt/parser/partial.py
+++ b/core/dbt/parser/partial.py
@@ -58,6 +58,7 @@ special_override_macros = [
     "generate_schema_name",
     "generate_database_name",
     "generate_alias_name",
+    "function",
 ]
 
 
@@ -295,6 +296,10 @@ class PartialParsing:
         if saved_source_file.parse_file_type == ParseFileType.Fixture:
             self.delete_fixture_node(saved_source_file)
 
+        # functions
+        if saved_source_file.parse_file_type == ParseFileType.Function:
+            self.delete_function_node(saved_source_file)
+
         fire_event(PartialParsingFile(operation="deleted", file_id=file_id))
 
     # Updates for non-schema files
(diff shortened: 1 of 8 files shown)


core/dbt/parser/partial.py
```diff
@@ -630,6 +646,31 @@ def delete_fixture_node(self, source_file):
...
+
+        # Remove the function node from the saved manifest
+        function_node = self.saved_manifest.functions.pop(function_unique_id)
+
+        # Remove the function node from the source file so that it's not viewed as a
+        # duplicate when it's re-added
+        source_file.nodes.remove(function_unique_id)
+
+        # schedule function for parsing.
+        # Note: We _don't_ need to schedule referencing nodes for reparsing, because a change in a function changes nothing
```
gshank: If the number of parameters has changed, is that something that's detected at parse time? Or only compilation time? If it's detected at parse time, then we should schedule referencing nodes in order to get that error. If it's not detected at parse time, this is fine.

gshank (approved):
Looks good, just one question.

core/dbt/parser/partial.py
QMalcolm: Nothing like responding to a comment a month and a half later 😅 

I think this comment stemmed from confusion of how the function parameters are specified/rendered. The specification of parameters for a function in a referencing model happen outside the `{{ }}` that is a function call with parameters would look like `{{ function("my_function") }}(my_arg1, my_arg2)`. As such, the rendering of the `function` macro is _only_ rendering the three part name of the function in the data warehouse. So a change to the function doesn't change the rendering.

I guess there is a question of if parsing should be rescheduled if the `database`, `schema`, or `alias` changes? 🤔

QMalcolm: I've added a test that confirms reparsing of the downstreams is not required if the 3 part name of the function changes

gshank (approved):

PR #11082, opened 2024-12-02
Create a no-op exposure runner

aranke:
### Problem

We want to make exposures executable with a no-op status in run results and log line, similar to saved queries.

### Solution

- [x] Create a new `NoOp` run status
- [x] Make `Exposure` extend `NodeInfoMixin` to make it executable in `build`
- [x] Create a new `NoOpRunner` in a new `dbt.runner` namespace, which uses the `NoOp` run status above
- [x] Make `SavedQueryRunner` inherit from `NoOpRunner` and move it to the `dbt.runner` namespace
- [x] Create a new `ExposureRunner` that also inherits from the `NoOpRunner`
- [x] Add new tests for run status for both exposures and saved queries
- [x] Add tests for selectors for exposures

#### Logs

```
❯ dbt build
12:39:45  Running with dbt=1.10.0-a1
12:39:45  Registered adapter: duckdb=1.9.1-dev28
12:39:46  Unable to do partial parsing because a project config has changed
12:39:46  Found 5 models, 3 seeds, 20 data tests, 1 exposure, 422 macros, 1 group
...
12:39:47  26 of 29 NO-OP exposure weekly_jaffle_metrics .................................. [NO-OP in 0.00s]
...
12:39:47  Finished running 1 exposure, 3 seeds, 2 table models, 20 data tests, 3 view models in 0 hours 0 minutes and 0.44 seconds (0.44s).
12:39:47  
12:39:47  Completed successfully
12:39:47  
12:39:47  Done. PASS=28 WARN=0 ERROR=0 SKIP=0 NO-OP=1 TOTAL=29
```

#### Run Results

```
❯ jq .results target/run_results.json
[
  ...
  {
    "status": "no-op",
    "timing": [
      {
        "name": "compile",
        "started_at": "2024-12-03T12:39:47.125472Z",
        "completed_at": "2024-12-03T12:39:47.125474Z"
      },
      {
        "name": "execute",
        "started_at": "2024-12-03T12:39:47.129151Z",
        "completed_at": "2024-12-03T12:39:47.129158Z"
      }
    ],
    "thread_id": "Thread-17 (worker)",
    "execution_time": 0.01341700553894043,
    "adapter_response": {},
    "message": "NO-OP",
    "failures": 0,
    "unique_id": "exposure.jaffle_shop.weekly_jaffle_metrics",
    "compiled": null,
    "compiled_code": null,
    "relation_name": null,
    "batch_results": null
  },
  ...
]
```

### Checklist

[...]

PR #12105, opened 2025-10-22
Don't return that exposures are _always_ part of the graph

QMalcolm:
Resolves #N/A

### Problem

In https://github.com/dbt-labs/dbt-core/pull/11082 (merged 10 months ago) we [added some lines such that](https://github.com/dbt-labs/dbt-core/pull/11082/files#r2453092558) an exposure should be considered part of the graph if it is enabled. But [higher up](https://github.com/dbt-labs/dbt-core/blob/ef3d8d47540110414f3c66433c6702b83f418dcf/core/dbt/graph/selector.py#L167) in that stack of if/elif/else change were lines that _always_ returned true for exposures. These lines higher up were never removed, so exposures continued to always return true (that they are part of the graph). This probably meant we were over selecting exposures the whole time 😬 

### Solution

Remove the old lines that always return true

### Checklist

### DIFF
--- a/.changes/unreleased/Fixes-20251022-140240.yaml
+++ b/.changes/unreleased/Fixes-20251022-140240.yaml
@@ -1,0 +1,6 @@
+kind: Fixes
+body: Don't return true for exposures being part of graph if they are disabled
+time: 2025-10-22T14:02:40.764424-05:00
+custom:
+  Author: QMalcolm
+  Issue: N/A


gshank (commented):
I thought disabled exposures would be in the disabled dictionary, not the exposures dictionary?

QMalcolm:
@gshank I've discussed this with @MichelleArk and @colin-rogers-dbt.
> I thought disabled exposures would be in the disabled dictionary, not the exposures dictionary?

You are right. This function seems to guard against a node being disabled but not actually making it to the `disabled` dictionary 🤔 This seems to being a doing a lot of work to simply ensure that we didn't make a mistake elsewhere. We're going to instead try ripping out and see what bugs fall out 🙈

PR #12135, opened 2025-11-03
Properly propagate config for metric when created from measure

WilliamDee:
### Problem
Currently, metrics can be created by measures from proxy through the `create_metric:true` flag. It was flagged that the `config.meta` was not being propagated correctly to that metric even though when `config.meta` is listed in the semantic model all downstream elements gets that and in this case if the metric is a proxy to the measure, we should also get that value. Similarly, the `expr` should take from `measure.expr` first before using the `measure.name`, otherwise it is incorrect logic. 

**Note that** `metric.expr` doesn't get used for measure proxy metrics currently so it won't change any behaviour atm, but will start to get used when measures gets removed in the coming quarter.
 

### Solution
Update the parser to pass the `config.meta` field during when handling `create_metric=True`
 

### Checklist

### DIFF
--- a/core/dbt/parser/schema_yaml_readers.py
+++ b/core/dbt/parser/schema_yaml_readers.py
@@ -635,7 +635,7 @@ class SemanticModelParser(YamlReader):
         semantic_model_name: str,
         meta: Optional[Dict[str, Any]] = None,
     ) -> None:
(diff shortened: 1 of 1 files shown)


QMalcolm (approved):
LGTM 🚀

PR #12138, opened 2025-11-03
[1.10.latest] Pin dbt-adapters to dev-requirements to fix python 3.9 support

MichelleArk:
Fixing CI builds now that python 3.9 has been dropped from dbt-adapters and dbt-test-adapter

Example failing run: https://github.com/dbt-labs/dbt-core/actions/runs/19045344894


QMalcolm (commented):
The failing check is just codecov. Looks good to me 🚀

PR #12157, opened 2025-11-07
Add new-syle YAML unparsed metric

theyostalservice:
Resolves [LINEAR](https://linear.app/dbt-labs/issue/DI-4407/parse-metrics-outside-of-models) [JIRA](https://dbtlabs.atlassian.net/browse/SEMANTIC-2983)

Depends on #12386 (Update: MERGED :-) )

### Problem

We need to support new yaml specs for metrics in the new V2 semantic layer YAML, although there is a lot of overlap.  This PR addresses creating the spec for the YAML itself.

### Solution

I broke the unparsed metric class into a base class and two different sub-classes based for v1 and v2.  I did the same thing for the tests for unparsed metrics.  (Sharing a base class and test is especially reasonable in this case because they share validation logic.)

The design goal here is to DRY as much as possible while also making it reasonable to do maintenance on only one version of YAML at once (including eventually surgically removing the old YAML format one day).

### Checklist


theyostalservice:
* **#12401** <a href="https://app.graphite.com/github/pr/dbt-labs/dbt-core/12401?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#12397** <a href="https://app.graphite.com/github/pr/dbt-labs/dbt-core/12397?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#12161** <a href="https://app.graphite.com/github/pr/dbt-labs/dbt-core/12161?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#12395** <a href="https://app.graphite.com/github/pr/dbt-labs/dbt-core/12395?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#12157** <a href="https://app.graphite.com/github/pr/dbt-labs/dbt-core/12157?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/dbt-labs/dbt-core/12157?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#12386** <a href="https://app.graphite.com/github/pr/dbt-labs/dbt-core/12386?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

core/dbt/contracts/graph/unparsed.py
```diff
@@ -141,11 +141,49 @@ class HasConfig:
...
+    label: Optional[str] = None
+    is_partition: bool = False
+    config: Dict[str, Any] = field(default_factory=dict)
+
+
+@dataclass
+class UnparsedDimensionTypeParams(dbtClassMixin):
+    """Used for dbt Semantic Layer dimensions (v1 YAML)."""
+
+    time_granularity: str  # TimeGranularity enum
```
WilliamDee, line 159: Comment isnt necessarily true here since we have custom grains

theyostalservice: I'm not sure how that works, tbh.  The comment isn't original to me - it was part of this line of code when I moved it.  However, if you look down at [this line](https://app.graphite.com/github/pr/dbt-labs/dbt-core/12157/Add-new-syle-YAML-unparsed-metric?utm_source=gt-slack-notif&pmgid=3435776#lines-core/dbt/parser/common.py-R226-R226), it seems like it would always become None if the value here wasn't a valid enum value.  It's also used to populate [this dsi field](https://github.com/dbt-labs/dbt-semantic-interfaces/blob/main/dbt_semantic_interfaces/implementations/elements/dimension.py#L39), which is an enum value.

WilliamDee: ah ok, i think the custom granularity here is specified under the timespine columns

QMalcolm: Yea the comment could have been more clear. Sorry about that. The comment is meant to mean "this is a string here, but really it's a TimeGranularity enum (in that that's what the user can specify). I forget _why_ we couldn't type it as an enum on the UparsedDimensionTypeParams but 🤷🏻 🫠

PR #12158, opened 2025-11-08
Stop allowing hyphens in metric names and add name validation tests

theyostalservice:
### Problem

Problems:
1. The validation for metric names should not allow hyphens.  They are already disallowed in dbt-semantic-interfaces and metricflow, and the error message here correctly identifies that they're not allowed, but a mistake in the regex allows them to slip through.
2. These name errors are validation errors, not parsing errors, and should be treated as such.
3. We don't have any tests showing that these work and protecting them during maintenance.

### Solution

Fix the regex, change the error type that we raise, and add tests!

### Checklist

### DIFF
--- a/core/dbt/contracts/graph/unparsed.py
+++ b/core/dbt/contracts/graph/unparsed.py
@@ -638,11 +638,11 @@ class UnparsedMetric(dbtClassMixin):
                 errors.append("cannot contain more than 250 characters")
             if not (re.match(r"^[A-Za-z]", data["name"])):
                 errors.append("must begin with a letter")
-            if not (re.match(r"[\w-]+$", data["name"])):
+            if not (re.match(r"[\w]+$", data["name"])):
                 errors.append("must contain only letters, numbers and underscores")
 
             if errors:
-                raise ParsingError(
+                raise ValidationError(
                     f"The metric name '{data['name']}' is invalid.  It {', '.join(e for e in errors)}"
                 )
 
--- a/.changes/unreleased/Fixes-20251107-160515.yaml
+++ b/.changes/unreleased/Fixes-20251107-160515.yaml
@@ -1,0 +1,6 @@
+kind: Fixes
+body: For metric names, fix bug allowing hyphens (not allowed in metricflow already), make validation throw ValidationErrors (not ParsingErrors), and add tests.
+time: 2025-11-07T16:05:15.946331-08:00
+custom:
+    Author: theyostalservice
+    Issue: n/a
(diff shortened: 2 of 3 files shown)


theyostalservice:
* **#12161** <a href="https://app.graphite.com/github/pr/dbt-labs/dbt-core/12161?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#12157** <a href="https://app.graphite.com/github/pr/dbt-labs/dbt-core/12157?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#12158** <a href="https://app.graphite.com/github/pr/dbt-labs/dbt-core/12158?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/dbt-labs/dbt-core/12158?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

core/dbt/contracts/graph/unparsed.py
```diff
@@ -638,11 +638,11 @@ def validate(cls, data):
                 errors.append("cannot contain more than 250 characters")
             if not (re.match(r"^[A-Za-z]", data["name"])):
                 errors.append("must begin with a letter")
-            if not (re.match(r"[\w-]+$", data["name"])):
+            if not (re.match(r"[\w]+$", data["name"])):
```
theyostalservice, line 641: tbh, I can't help but think it might make more sense to try to unify this regex and pull it directly from dbt-semantic-interfaces, but with all the changes afoot with that repo and metricflow right now, it seems better to avoid the added workflow complexity.

But, for reference, [here](https://github.com/dbt-labs/dbt-semantic-interfaces/blob/main/dbt_semantic_interfaces/validations/unique_valid_name.py#L69-L86) is the different but definitely-not-allowing hyphens regex fron dsi:
```
    NAME_REGEX = re.compile(r"\A[a-z]((?!__)[a-z0-9_])*[a-z0-9]\Z")

    @staticmethod
    def check_valid_name(  # noqa: D
        name: str, context: Optional[ValidationContext] = None
    ) -> Sequence[ValidationIssue]:
        issues: List[ValidationIssue] = []

        if not UniqueAndValidNameRule.NAME_REGEX.match(name):
            issues.append(
                ValidationError(
                    context=context,
                    message=f"Invalid name `{name}` - names may only contain lower case letters, numbers, "
                    f"and underscores. Additionally, names must start with a lower case letter, cannot end "
                    f"with an underscore, cannot contain dunders (double underscores, or __), and must be "
                    f"at least 2 characters long.",
                )
            )
```

QMalcolm (commented):
@theyostalservice is my understanding correct that these were raising an error previously on the MetricFlow side, but just not in core? I ask because if so, this should be good as is. If no error was being raised anywhere then we'd have to do some additional handling in code as it'd be a breaking change

theyostalservice:
Yes, that's the correct understanding! :-)

PR #12169, opened 2025-11-15
Add test coverage for SL metric parsing

theyostalservice:
### Problem

We don't have tests describing the expected behaviors for metrics in our current parsing logic, and that makes it harder to reason about said behaviors with confidence.  More urgently, it is making it hard for me to change that logic for the newer type of Semantic Layer YAML with confidence that I'm not disturbing the logic that controls parsing for the older yaml.

### Solution

Add tests!  :)

There's a particular focus on making sure the input measures, metrics, and depends_on (only used in core, not in metricflow I believe) are being handled as expected since these are most relevant to the work I'm doing elsewhere.

There are several things to note:
* I broke the tests into separate classes.  This is how tests are handled elsewhere in the codebase, and it seems necessary for proper isolation while we're reading and writing test files.
* I discovered a bug in how partial parsing is handled when removing large swathes of semantic layer objects (we don't order the removals in a way that necessarily respects the dependencies, which causes errors), so you will see the old test commented out (it no longer runs reliably now that the test yaml is more complex) as well as a new version of the test that removes metrics iteratively and shows that that at least is safe.

### Checklist

### DIFF
--- a/core/dbt/parser/manifest.py
+++ b/core/dbt/parser/manifest.py
@@ -1940,7 +1940,7 @@ def _process_metric_node(
 
             if target_metric is None:
                 raise dbt.exceptions.ParsingError(
-                    f"The metric `{input_metric.name}` does not exist but was referenced.",
+                    f"The metric `{input_metric.name}` does not exist but was referenced by metric `{metric.name}`.",
                     node=metric,
                 )
             elif isinstance(target_metric, Disabled):
--- a/tests/functional/semantic_models/test_semantic_model_parsing.py
+++ b/tests/functional/semantic_models/test_semantic_model_parsing.py
@@ -2,21 +2,29 @@ from typing import List
 
 import pytest
 
+from dbt.artifacts.resources.v1.semantic_model import MetricType
 from dbt.contracts.graph.manifest import Manifest
 from dbt.tests.util import run_dbt, write_file
 from dbt_common.events.base_types import BaseEvent
+from dbt_semantic_interfaces.type_enums.conversion_calculation_type import (
+    ConversionCalculationType,
+)
+from dbt_semantic_interfaces.type_enums.period_agg import PeriodAggregation
 from dbt_semantic_interfaces.type_enums.time_granularity import TimeGranularity
 from tests.functional.assertions.test_runner import dbtTestRunner
 from tests.functional.semantic_models.fixtures import (
+    base_schema_yml,
+    conversion_metric_yml,
     fct_revenue_sql,
     metricflow_time_spine_sql,
     multi_sm_schema_yml,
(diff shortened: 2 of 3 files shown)


theyostalservice:
* **#12169** <a href="https://app.graphite.com/github/pr/dbt-labs/dbt-core/12169?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/dbt-labs/dbt-core/12169?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

core/dbt/parser/manifest.py
```diff
@@ -1940,7 +1940,7 @@ def _process_metric_node(
 
             if target_metric is None:
                 raise dbt.exceptions.ParsingError(
-                    f"The metric `{input_metric.name}` does not exist but was referenced.",
+                    f"The metric `{input_metric.name}` does not exist but was referenced by metric `{metric.name}`.",
```
theyostalservice, line 1943: this is just a small improvement in error message that helped me when debugging. :-)

tests/functional/semantic_models/test_semantic_model_parsing.py
```diff
@@ -60,7 +78,155 @@ def test_semantic_model_error(self, project):
...
+        assert conversion_params.calculation == ConversionCalculationType.CONVERSION_RATE
+
+        assert len(metric.type_params.input_measures) == 2
+        assert "count_of_things" in [im.name for im in metric.type_params.input_measures]
+        assert "count_of_things_2" in [im.name for im in metric.type_params.input_measures]
+
+        # Not sure where dbt uses this, but for now, let's just make sure
+        # we're keeping the contract that these metrics are marked as depending on
+        # the semantic model.
+        assert "semantic_model.test.revenue" in metric.depends_on.nodes
```
QMalcolm, line 150: The depends_on of a node powers a few things
* node selection
* skipping nodes down stream of a failed node
* powering lineage in dbt docs / dbt explorer

The details are gonna dive mostly into exports because the metics.depends_on is mostly important when thinking about how it affects other nodes that create things 😅 

In the cloud IDE exports can be run with `build` so if I do `dbt build --select +my_export` everything upstream necessary for that export is "run". Exports point to metrics and semantic models, metrics point to semantic models, semantic models point to models, and etc. So the `--select +my_export` ensures all those upstream things that are dependend on get run.

Similarly, if I do a `dbt build`, and some of my models fail (but not all), then any export that depends on through its lineage will be skipped.

Finally, docs / explorer. To show how things relate, we need to know how they relate. The depends on tells us how things relate 🙂

